PR #1034, opened 2019-03-08
Validate branch sizes for all experiment types fixes #1030

### LINKED ISSUE #1030: The branch size can wrongly be saved with whole numbers larger or smaller than 100
**[Affected versions]:**
- Firefox Release 65.0.2 and above
- Chrome 72.0.3626.121

**[Affected Platforms]:** 	 	
- All Windows
- All Mac 	 	
- All Linux

**[Prerequisites]:**
- You are logged with your LDAP account on the "https://experimenter.stage.mozaws.net/" website.
- Have an experiment created on the Experimenter stage website.

**[Steps to reproduce]:**
2. Open the created experiment from prerequisites.
3. Click the "Edit" button from "Branches" section.
4. Enter a whole number larger or smaller than 100 in the "Branch Size" input field.
5. Click the "Save Draft" button and observe the behavior.

**[Expected result]:**
- The "! The size of all branches must add up to 100 " error message is displayed.

**[Actual result]:**
- The draft is saved.

**[Notes]:**
- This issue is also reproducible if multiple Branches are created and saved with different branch sizes.  The entered numbers sum can result in values smaller or larger than 100.
branchsize

jaredkerim:


app/experimenter/experiments/tests/test_forms.py
```diff
@@ -630,6 +630,26 @@ def test_form_checks_uniqueness_for_single_experiment_not_all(self):
         experiment2 = form2.save()
         self.assertEqual(experiment2.variants.count(), 3)
 
+    def test_form_valid_if_sizes_sum_to_100(self):
+        form = ExperimentVariantsAddonForm(
+            request=self.request, data=self.data
+        )
+        self.assertTrue(form.is_valid())
```
mythmon: So, I can understand why this test works. On the other hand, the data that makes this test valid and interesting is very far away from this test. I think it would be better to include the experiment sizes data directly in the test, even if it is redundant.

mythmon: This comment also applies to the other tests, which again rely on data that isn't clear and is far away from the test. I think that this hurts the readability of the tests, which makes it harder to maintain.

peterbe:
There's some repetition now. In the whole test suite that `!= 100` happens in:

* `TestExperimentVariantsPrefForm.test_form_invalid_if_sizes_sum_to_more_than_100`
* `TestExperimentVariantsPrefForm.test_form_invalid_if_sizes_sum_to_less_than_100`
* `TestExperimentVariantsAddonForm.test_form_invalid_if_sizes_sum_to_more_than_100`
* `TestExperimentVariantsAddonForm.test_form_invalid_if_sizes_sum_to_less_than_100`
* `TestExperimentVariantsFormSet.test_formset_invalid_if_sizes_sum_to_more_than_100`
* `TestExperimentVariantsFormSet.test_formset_invalid_if_sizes_sum_to_less_than_100`
* `TestExperimentVariantsFormSet.test_empty_branch_size_raises_validation_error`

The last one isn't actually about `!= 100` but actually a test of it being `== 0`. 

IMHO, it's a mistake to even have unit tests against `TestExperimentVariantsFormSet` since that's not the interface. The form submission goes to `TestExperimentVariantsPrefForm` or `TestExperimentVariantsAddonForm` which internally *implement* `ExperimentVariantsFormSet`.

jaredkerim:
@peterbe I see your point, I was thinking of it as the tests against `ExperimentVariantsFormSet` are unit tests which test its logic in solation, and the tests against `ExperimentVariantsPrefForm` and `ExperimentVariantsAddonForm` are integration tests.  This bug was able to sneak into staging because the unit tests existed but not the integration tests.  If you think it's more prudent to reduce the number of redundant tests, I agree we could get away with only the integration tests and not the unit tests.

jaredkerim:
Also @peterbe with regards to validating before the form is bound, can you be more specific?

peterbe:
> with regards to validating before the form is bound, can you be more specific?

I might have been wrong on that. A form instance is bound as soon as any data is passed to it. I.e.

```python
>>> form = Form()
>>> form.is_bound
False
>>> form = Form({'some': 'data'})
>>> form.is_bound
True
```

However, the `form.cleaned_data` is not available until `form.is_valid()`.

jaredkerim:
@peterbe Yes that's true but when validating a formset the forms are individually validated before the formset clean method is called:

https://docs.djangoproject.com/en/2.1/topics/forms/formsets/#custom-formset-validation

jaredkerim:
@peterbe @mythmon I used my good friend @wolever 's handy package https://github.com/wolever/parameterized to parameterize the tests, and added the explicit setup data to each one to make them more self contained.

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -256,70 +255,97 @@ def test_empty_slug_raises_error(self):
         self.assertFalse(form.is_valid())
 
 
-class TestExperimentVariantsFormSet(TestCase):
+@parameterized_class(
+    ["form_class"],
+    [[ExperimentVariantsAddonForm], [ExperimentVariantsPrefForm]],
+)
+class TestExperimentVariantsFormSet(MockRequestMixin, TestCase):
```
peterbe: This name isn't great anymore.

jaredkerim: I still think it applies because that's the name of the class that implements the logic that it's testing.

PR #1038, opened 2019-03-11
Run CI tests with pytest instead of Django's test runner fixes #1037

### LINKED ISSUE #1037: Run CI tests with pytest instead of Django's test runner
* pytest is less noisy, no stdout unless a test fails
* ability to control env vars without having to pass `--settings=experimenter.settings_test`
* With `pytest-cov` ability to configure and control coverage with a `.coveragerc` file
* Better output in general on errors
* Generally more powerful than Django's test runner and more options. 
* There's already a `Makefile` directive (`test-watch` that uses `ptw`)

peterbe:
Fixes #1037


jaredkerim:
@peterbe So does this mean we can remove settings_test.py altogether?

peterbe:
> So does this mean we can remove settings_test.py altogether?

[No, we still need it. For now at least.](https://github.com/mozilla/experimenter/blob/2ac057186304816e06a4a65b1277f8d2cc70ae7b/app/pytest.ini#L3)

There has to be *some* place where you hardcode a list of settings geared specifically for running tests. 

It's probably too much of a change at the moment but ideally we should be switching to using `django-configurations`. Both [Tecken](https://github.com/mozilla-services/tecken/blob/bc17c36e10a39c62144eb15e1d7ea0abb9f62530/tecken/settings.py#L668) and [Normandy](https://github.com/mozilla/normandy/blob/8fae42c766ebd6dc1267dbe51c68179d70347c9d/normandy/settings.py#L480) uses it. It simply takes the `settings.py` and breaks it up into classes instead of being just one long flat list of setting after setting. But there are other bigger changes required for that too. 

Another option is to use `pytest-env`. See this example: https://github.com/mozilla-services/remote-settings-lambdas/pull/251/files#diff-513c9b6861fc981fb2c708db477f542cR7 
It basically reads the `env` section from the `pytest.ini` and force in those settings before the python-decouple in `settings.py` kicks in. 

What I'd really like is to have a way to confidently and strictly make sure the local `.env` file is NEVER EVER read when running tests.

jaredkerim:
@peterbe I'm confused, how is `settings_test.py` used if this PR lands?

peterbe:
See https://github.com/mozilla/experimenter/blob/2ac057186304816e06a4a65b1277f8d2cc70ae7b/app/pytest.ini#L3

Since `pytest` doesn't run `manage.py` it can't pick any defaults there. And since it's not using `manage.py` it doesn't support the `--settings` argument. 
Instead, in `pytest.ini` it sets the `DJANGO_SETTINGS_MODULE`.

peterbe:
Which reminds me, we don't need to set `DEGUG = False` in there. Django does that automatically for you.

PR #1040, opened 2019-03-12
Add normandy slug fixes #582

### LINKED ISSUE #582: Auto format slugs correctly
When an experiment is created, we need to create a slug of the form

[type]-[name]-[population]-[bug id]

This means we need to have two slugs, one the experimenter slug, and the other the normandy slug, because we won't have the bugzilla id until after the experiment is created in experimenter.

The normandy slug is what will get communicated to normandy and then used for monitoring.

jaredkerim:
- Added the `normandy_slug` field
- Populate it when the experiment moves to `Ready to Ship`


app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1047,17 +1047,31 @@ def test_sets_bugzilla_id_when_draft_becomes_review(self):
...
             data={"status": experiment.STATUS_SHIP},
             instance=experiment,
         )
+
         self.assertTrue(form.is_valid())
         experiment = form.save()
+
+        self.assertEqual(
+            experiment.normandy_slug,
+            "pref-experiment-slug-Nightly-57.0-bug-12345",
```
mythmon: This is fine, but slugs *should* be all lower case.

jaredkerim: Good call.

jaredkerim:
@mythmon I wasn't planning on making it editable by users, but if that comes up we can do it in a separate PR.

PR #1043, opened 2019-03-14
group Risks questions/responses better fixes #1004

### LINKED ISSUE #1004: "Risks" details is hard to read - group the question/response better
* Unless your eyes/focus is on the first or last item, it's not sure which Yes/No goes with which question, 
* Normally green is the "universal" color for "yes" and red for "no". (Also, we could consider using emojis/unicode characters that look like checkboxes/crosses) 

screen shot 2019-03-04 at 2 51 01 pm

peterbe:
Fixes #1004

Screen Shot 2019-03-14 at 10 38 01 AM

### DIFF
--- a/app/experimenter/templates/experiments/section_risks.html
+++ b/app/experimenter/templates/experiments/section_risks.html
@@ -5,68 +5,44 @@
 {% endblock %}
 
 {% block section_content %}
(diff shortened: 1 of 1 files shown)


jaredkerim:
@shell1 How does this look to you?

shell1:
Love it!

Issue #1062, opened 2019-03-18
Danger of state conflicts


peterbe:
I've seen a couple of these. Here's one example: https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/forms.py#L859

It looks like this:

```python
class SomeModelForm:

    class Meta:
        model = Experiment

    def save(self, *args, **kwargs):
        experiment = super().save(*args, **kwargs)

        if some_condition:
             some_instance.save()
        if other_condition:
            SomeModel.objects.create(...)
```

So far nothing terrible has happened, that I'm aware of, in production. But I've definitely been bitten by it when dev'ing on something such that an exception is raised further down in the "meat" of the `save` method. Because the whole, default, implementation is *not* atomic, the first `save()` might work but the latter won't. What you end up with is potentially a state conflict. 

The relative risk is small because, I think, most cases of this is something like:

1. Save changes to experiment
2. Make a notification
3. Send an instruction to Celery to send an email
4. Send an instruction to Celery to update Bugzilla

In other words, many times the stuff that happens after that critical experimenter save isn't super important.

jaredkerim:
We are undergoing major refactors for the existing bugzilla task logic which I think should address this.

PR #1064, opened 2019-03-18
Store addon info fixes #500

### LINKED ISSUE #500: Collect additional Addon information
We need to store a URL to the signed XPI for Addon studies.  

If they select a "Pref Flip Study' (the very first question when creating a new experiment) - there is a "Firefox Pref" section that displays on the Population wizard page (below Population Filtering and above Control Branch).  

We need a similar section that only applies only if they answer "Add-on Study".  It should show up in the same location (between Population Filtering and above Control Branch).  Section should be called "Add-on Information".  It has 4 fields.

- Add-on Name (optional)
- Experiment ID (optional, experiment identifier as it appears in the addon)
- Testing Signed Add-on Location (optional)
- Release Signed Add-on Location (optional, but marked as incomplete - so this field shows as missing info that is needed when users click "ready to ship").
- There should be **help** for the Test Signing and Release Signing.  They should link here: https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-Add-ons

jaredkerim:
Screenshot 2019-03-19 14 54 45

Screenshot 2019-03-18 14 11 39
Screenshot 2019-03-18 14 12 42


app/experimenter/experiments/forms.py
```diff
@@ -434,19 +430,47 @@ def is_valid(self):
         return super().is_valid() and self.variants_formset.is_valid()
 
     def save(self, *args, **kwargs):
-        experiment = super().save(*args, **kwargs)
+        self.variants_formset.save()
```
peterbe: Two saves should be done within the same transaction. You can take a big bite out of https://github.com/mozilla/experimenter/issues/1062 by doing a...:

```python
with transaction.atomic():
    self.variants_formset.save()
    return super().save(*args, **kwargs)
```

It is admittedly a bit off-topic but that's how it's supposed by be done anyway and it'd be a shame to not tackle it on edited code.

app/experimenter/experiments/forms.py
```diff
@@ -434,19 +430,47 @@ def is_valid(self):
...
+    )
+    addon_release_url = forms.URLField(
+        required=False,
+        label="Add-On Signed Release Location",
+        help_text=Experiment.ADDON_RELEASE_URL_HELP_TEXT,
+    )
+
+    class Meta:
+        model = Experiment
+        fields = ExperimentVariantsBaseForm.Meta.fields + [
```
peterbe: Nice!

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -832,85 +806,166 @@ def test_clean_unrecognized_countries(self):
...
+class TestExperimentVariantsAddonForm(MockRequestMixin, TestCase):
 
     def setUp(self):
         super().setUp()
 
-        self.branch0_name = "control name"
-        self.branch1_name = "branch 1 name"
-        self.branch2_name = "branch 2 name"
-
         self.data = {
```
peterbe: Can we move all the common things from these now three massive dicts into just one. E.g.

```python
class TestExperimentVariantsBaseForm:
    FIXTURE_DATA = {
        "population_percent": "10",
        "firefox_version": Experiment.VERSION_CHOICES[-1][0],
        "firefox_channel": Experiment.CHANNEL_NIGHTLY,
        "client_matching": "en-us only please",
        "locales": [],
        "countries": [],
        ...
    }
    def setUp(self):
        self.data = copy.copy(self.FIXTURE_DATA)

class TestExperimentVariantsAddonForm:
    def setUp(self):
        self.data = copy.copy(TestExperimentVariantsBaseForm.FIXTURE_DATA)
        self.data.update({
            "addon_name": "Addon Name",
            "addon_experiment_id": "addon_experiment_id",
            "addon_testing_url": "https://www.example.com/testing.xpi",
            "addon_release_url": "https://www.example.com/release.xpi",
            ...anything else unique to this form class...
        })


class TestExperimentVariantsPrefForm:
    def setUp(self):
        self.data = copy.copy(TestExperimentVariantsBaseForm.FIXTURE_DATA)
        self.data.update({
            ...all things unique to this form class...
        })
```

app/experimenter/experiments/tests/test_models.py
```diff
@@ -601,6 +601,16 @@ def test_population_is_complete_when_values_set(self):
         experiment = ExperimentFactory.create()
         self.assertTrue(experiment.completed_population)
 
+    def test_addons_is_not_complete_when_release_url_not_set(self):
+        experiment = ExperimentFactory.create(addon_release_url=None)
+        self.assertFalse(experiment.completed_addon)
+
+    def test_adodns_is_complete_when_release_url_set(self):
```
peterbe:
```suggestion
    def test_addons_is_complete_when_release_url_set(self):
```

app/experimenter/experiments/tests/test_models.py
```diff
@@ -709,6 +719,30 @@ def test_completed_all_sections_true_when_complete(self):
...
 
+    def test_completed_all_sections_true_for_pref_study_without_addon(self):
+        experiment = ExperimentFactory.create_with_status(
+            Experiment.STATUS_REVIEW,
+            type=Experiment.TYPE_PREF,
+            addon_release_url=None,
+        )
+        self.assertTrue(experiment.completed_all_sections)
+
+    def test_completed_all_sections_false_for_addon_study_without_addon(self):
```
peterbe: This test duplicates the test in `test_addons_is_complete_when_release_url_set` above. I think keeping this one is OK.

app/experimenter/experiments/constants.py
```diff
@@ -358,6 +359,62 @@ class ExperimentConstants(object):
         )
     )
 
+    ADDON_NAME_HELP_TEXT = """
+      <p>
+        Enter the name of your addon.
```
peterbe:
```suggestion
        Enter the name of your add-on.
```

Perhaps it should be capitalized to "Add-On" to be most consistent.

app/experimenter/experiments/constants.py
```diff
@@ -358,6 +359,62 @@ class ExperimentConstants(object):
...
+        url=(
+            "https://mana.mozilla.org/wiki/display/FIREFOX/"
+            "Pref-Flip+and+Add-On+Experiments"
+            "#Pref-FlipandAdd-OnExperiments-Add-ons"
+        )
+    )
+
+    ADDON_EXPERIMENT_ID_HELP_TEXT = """
+      <p>
+        Enter the Experiment ID as it appears in the addon.
```
peterbe: Ditto

app/experimenter/experiments/constants.py
```diff
@@ -358,6 +359,62 @@ class ExperimentConstants(object):
...
+        url=(
+            "https://mana.mozilla.org/wiki/display/FIREFOX/"
+            "Pref-Flip+and+Add-On+Experiments"
+            "#Pref-FlipandAdd-OnExperiments-Add-ons"
+        )
+    )
+
+    ADDON_TESTING_URL_HELP_TEXT = """
+      <p>
+        Enter the URL where the test build of your addon can be found.
```
peterbe: ditto

app/experimenter/experiments/constants.py
```diff
@@ -358,6 +359,62 @@ class ExperimentConstants(object):
...
+        url=(
+            "https://mana.mozilla.org/wiki/display/FIREFOX/"
+            "Pref-Flip+and+Add-On+Experiments"
+            "#Pref-FlipandAdd-OnExperiments-Add-ons"
+        )
+    )
+
+    ADDON_RELEASE_URL_HELP_TEXT = """
+      <p>
+        Enter the URL where the release build of your addon can be found.
```
peterbe: ditto

app/experimenter/experiments/forms.py
```diff
@@ -434,19 +430,47 @@ def is_valid(self):
...
+        help_text=Experiment.ADDON_NAME_HELP_TEXT,
+    )
+    addon_experiment_id = forms.CharField(
+        required=False,
+        label="Add-On Experiment ID",
+        help_text=Experiment.ADDON_EXPERIMENT_ID_HELP_TEXT,
+    )
+    addon_testing_url = forms.URLField(
+        required=False,
+        label="Add-On Signed Testing Location",
```
peterbe: Out of curiosity, why the word "Location" and not URL? I tend to believe "Location" could add more confusion than "URL" but I'm also very web techy and perhaps blind to the nuances here.

app/experimenter/experiments/models.py
```diff
@@ -76,6 +76,13 @@ class Experiment(ExperimentConstants, models.Model):
         validators=[MaxValueValidator(ExperimentConstants.MAX_DURATION)],
     )
 
+    addon_name = models.CharField(max_length=255, blank=True, null=True)
+    addon_experiment_id = models.CharField(
+        max_length=255, blank=True, null=True
+    )
+    addon_testing_url = models.URLField(blank=True, null=True)
```
peterbe: Are you confident this default length is enough? It is 200 chars as of today in Django 2. 
I have, so many times, fallen for that default only to realize later that it wasn't enough. 
Especially when the URLs are quite unpredictable and especially if the URLs *might* contain Unicode which when turned into quoted hex becomes super long. 

I'm not super familiar with what these URLs are but something like `max_length=400` would give me a lot better sleep at night.

app/experimenter/experiments/tests/test_models.py
jaredkerim: What do you mean?  One is testing the `completed_addon` property, the other is testing the `completed_all_sections` property's optional dependence on the addon fields for addon studies.

peterbe: I mean that you don't need to have a dedicated unit test just for the property method. If the `completed_all_sections` test works, it'll travel all code branches. 

But if you want to keep both, that's fine too. Not a blocker for me.

jaredkerim: It's specifically testing the case where it's an addon type study and the completed_addon field becomes required here: https://github.com/mozilla/experimenter/pull/1064/files#diff-562e6046de58c57f7e5e7f6ecdb15efbR473

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -354,7 +355,47 @@ def test_empty_branch_size_raises_validation_error(self):
...
+    "client_matching": "en-us only please",
+    "locales": [],
+    "countries": [],
+    "pref_key": "browser.test.example",
+    "pref_type": Experiment.PREF_TYPE_STR,
+    "pref_branch": Experiment.PREF_BRANCH_DEFAULT,
+    "addon_name": "Addon Name",
+    "addon_experiment_id": "addon_experiment_id",
+    "addon_testing_url": "https://www.example.com/testing.xpi",
+    "addon_release_url": "https://www.example.com/release.xpi",
```
peterbe: Not a blocker but I think you should break this apart. Right now, if I'm not mistaken, this monster dict contains the union of all the possible data. Either something like this:

```python
VARIANTS_FORM_FIXTURE_BASE_DATA = {
    "population_percent": "10",
    "firefox_version": Experiment.VERSION_CHOICES[-1][0],
    "firefox_channel": Experiment.CHANNEL_NIGHTLY,
    "client_matching": "en-us only please",
    "locales": [],
    "countries": [],
    ...
}
VARIANTS_FORM_FIXTURE_ADDONS_DATA = {
    "addon_name": "Addon Name",
    "addon_experiment_id": "addon_experiment_id",
    "addon_testing_url": "https://www.example.com/testing.xpi",
    "addon_release_url": "https://www.example.com/release.xpi",
    ...
}
VARIANTS_FORM_FIXTURE_PREFS_DATA = {
    ...
}
```
...or...

reduce it down to the bare minimum that is common for all formsets and in each's `setUp` do something like:
```python
    def setUp(self):
        super().setUp()
        self.data = copy.copy(VARIANTS_FORM_FIXTURE_DATA)
        self.data.update({
             "specific": "to",
             "justthis": "Class",
             ...
        })
```

Also, since it's a "constant" (uppercase variable), it should sit at the top of the module. Either that, or even better, put it where it belongs; inside the first test case class that needs it.

jaredkerim: I don't think we can do this, because we parameterize the tests across all the form classes, the data needs to be a valid superset for all the forms.

app/experimenter/experiments/constants.py
```diff
@@ -151,9 +152,9 @@ class ExperimentConstants(object):
         for users that enroll into the experiment.
       </p>
       <p>
-        An <strong>{[1]}</strong> sends a Firefox addon which
+        An <strong>{[1]}</strong> sends a Firefox Add-On which
```
peterbe: I don't know if it matters, and we can always fix it some other day but the spelling seems to be "Add-ons" (not "Add-Ons").
Screen Shot 2019-03-19 at 3 17 21 PM
Screen Shot 2019-03-19 at 3 18 42 PM

jaredkerim: Oh yeah, looks like `add-on` is actually valid: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons

app/experimenter/experiments/forms.py
```diff
@@ -434,19 +431,48 @@ def is_valid(self):
         return super().is_valid() and self.variants_formset.is_valid()
 
     def save(self, *args, **kwargs):
-        experiment = super().save(*args, **kwargs)
+        with transaction.atomic():
+            self.variants_formset.save()
+            return super().save(*args, **kwargs)
```
peterbe: I know it was my suggestion initially, to use a context manager, but another option is to use it as a decorator which is pretty neat because they meat of the method doesn't have to be indented just because of this. And there's really no reason to NOT run the whole method in a single transaction.

```python
    @transaction.atomic
        self.variants_formset.save()
        return super().save(*args, **kwargs)
```

app/experimenter/experiments/models.py
```diff
@@ -241,6 +248,16 @@ def test_tube_url(self):
             ).format(slug=self.slug)
 
     def generate_normandy_slug(self):
+        if self.is_addon_study:
+            if not self.addon_experiment_id:
+                raise ValueError(
+                    (
+                        "An Add-On experiment requires a unique Active "
```
peterbe:
```suggestion
                        "An Add-On experiment requires an Active "
```

app/experimenter/experiments/models.py
```diff
@@ -452,13 +478,18 @@ def completed_required_reviews(self):
 
     @property
     def completed_all_sections(self):
-        return (
+        completed = (
             self.completed_population
             and self.completed_variants
             and self.completed_objectives
             and self.completed_risks
```
peterbe: Perhaps simpler, instead of requiring a mutating variable, would be:
```python
return (
    self.completed_population
    and self.completed_variants
    and self.completed_objectives
    and self.completed_risks
    and (not self.is_addon_study or self.completed_addon)
)
```

app/experimenter/experiments/models.py
jaredkerim: @peterbe After talking to @mythmon and Sunah yesterday we realized that we need to capture the experiment id to be passed to Normandy for monitoring and that it must be unique.

app/experimenter/experiments/models.py
jaredkerim: Yes that's definitely valid but it seems more a matter of preference.

app/experimenter/experiments/forms.py
jaredkerim: Yeah I saw that too, very convenient!

app/experimenter/templates/experiments/section_addon.html
```diff
@@ -0,0 +1,17 @@
...
+{% endblock %}
+
+{% block section_content %}
+  <h4>{{ experiment.addon_name }}</h4>
+  <p>
+    <strong class="mr-2">Experiment ID:</strong>
+    <code>{{ experiment.addon_experiment_id }}</code>
+  </p>
+  <p><strong>Test Build</strong></p>
+  <p><a href="{{ experiment.addon_testing_url }}" target="_blank" rel="noreferrer noopener">{{ experiment.addon_testing_url }}</a></p>
```
peterbe: It's a bit weird to display these and assume that they're valid URLs. The fields, in the model, are nullable. 
I'd rather it was something like

```django
{% if experiment.addon_testing_url %}
  <a href="{{ experiment.addon_testing_url }}" target="_blank" rel="noreferrer noopener">{{ experiment.addon_testing_url }}</a>
{% else %}
  <i>Not set</i>
{% endif %}
```

jaredkerim: That's [implicit here](https://github.com/mozilla/experimenter/pull/1064/files/49e3d4a07ac0d29c36dde4c990eee0978e805c2f#diff-b19c51dfc50ef6ccf00d88b6777481c2R62) by checking `completed_addon` before we show the content of this section.

app/experimenter/experiments/models.py
peterbe: The *code* doesn't doesn't do a check of uniqueness. It does a check of the `addon_experiment_id` being truthy. 

If it was something like:

```python
if Experiment.objects.filter(addon_experiment_id=addon_experiment_id).exclude(id=self.id).exists():
    raise ValueError('addon_experiment_id is not unique')
```
...the error message would match the code.

app/experimenter/experiments/forms.py
```diff
@@ -433,20 +430,49 @@ def clean_population_percent(self):
...
+    FORMSET_CLASS = ExperimentVariantsFormSet
+
+    addon_name = forms.CharField(
+        required=False,
+        label="Add-on Name",
+        help_text=Experiment.ADDON_NAME_HELP_TEXT,
+    )
+    addon_experiment_id = forms.CharField(
+        required=False,
+        label="Active Experiment Name",
```
peterbe: Why is the label "Name"? and not "Identifier" or "ID"?

jaredkerim: Yeah I changed that to reflect the way it's represented in add-ons: https://github.com/mozilla/shield-studies-addon-utils/blob/master/test-addon/src/studySetup.js#L23

jaredkerim: And added some context to the help text: https://github.com/mozilla/experimenter/pull/1064/files#diff-00de22e6c754b3d410f1d3e631e1825eR378

app/experimenter/experiments/models.py
peterbe: Also, **uniqueness** is already protected by the constraint in postgres. Except it's `django.db.exceptions.IntegrityError` (me thinks) exception.

jaredkerim: Oh I see what you mean, yes that is true, we're only checking that it's set.  Django field validation checks the actual uniqueness, so if this case it's not failing a uniqueness validation it's just failing a presence check.  You're right I'll update the message to accurately reflect what is being checked.

PR #1066, opened 2019-03-18
fix #847 make error message for enrollment duration match help text

### LINKED ISSUE #847: There is a inconsistency in the "Proposed Enrollment Duration" section between the description an the error message displayed regarding the enrollment duration
**[Affected versions]:**
- Firefox Release 65.0 and above
- Chrome 71.0.3578.98

**[Affected Platforms]:** 	 	
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a new Firefox profile created.
- Have an experiment created on the "https://experimenter.stage.mozaws.net/" website.
- You are logged in your LDAP account.

**[Steps to reproduce]:**
2. Open the experiment from prerequisites and click the "Edit" button from "Timeline" section.
3. Insert a higher number in the "Proposed Enrollment Duration" input field then the one displayed in the "Proposed Experiment Duration" input field.
4. Click the "Save Draft" button.
5. Click the "Help" button from the "Proposed Enrollment Duration".
6. Observe the duration of enrollment described in the error message and the one in the description displayed under the error message. 

**[Expected result]:**
- The descriptions provided and the error message imply the same enrollment duration.

**[Actual result]:**
- The error message implies that the duration must be "less than or equal"  to the experiment duration and the description implies that the duration must be "shorter" than the experiment duration.

**[Notes]:**
- Attached is a screen shot of the issue: 
enrollment

mostlygeek:
Fixes #847 text so wording is more consistent

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -186,8 +186,8 @@ class ExperimentOverviewForm(
 
         if enrollment > cleaned_data.get("proposed_duration", 0):
             msg = (
-                "The enrollment duration must be less than "
-                "or equal to the experiment duration."
+                "The enrollment duration must be shorter than "
+                "the experiment duration."
             )
             self._errors["proposed_enrollment"] = [msg]
 


jaredkerim:
Oh I didn't see @shell1 's comments, if we change the wording to this then we also have to change the logic, becuase the logic specifies that it should be less than or equal to.  That's the case where enrollment duration == experiment duration, which is the same as not specifying enrollment duration at all.  So I guess we should make it so that if they specify the enrollment duration at all it MUST be less than the experiment duration, but then we also have to update the logic and tests.  @mostlygeek That's probably more than you signed on for, do you want me to close this and I can make those logic updates?

mostlygeek:
The logic change is like one character ... in #847 it appears @shell1 left herself a note: 

"Shell find out if it is until the study end or until it hits the target population size or both - how does this really work :)"

I think since this fix is not fully undefined we should take this out of V2 and have shell follow up what she wants here.   I can do the work.  It's a good easy bug that lets me get a feel for the guts of experimenter :)

jaredkerim:
@mostlygeek Okay we'll block this until @shell1 is back from PTO to weigh in on the behaviour.

jaredkerim:
Also I'm going to close this PR until we figure out what the change should actually be so it doesn't clog the PR list.

PR #1073, opened 2019-03-19
Store normandy id fixes #910

### LINKED ISSUE #910: Add Field for the Normandy Recipe ID Number
Adding a field to enter the Normandy Recipe number makes connecting from the Experiment to the Normandy JSON much simpler.

Having this number makes it simpler for admins to move between Experimenter and Normandy.

This would be editing after it is in "Ready to Ship" state - so may be blocked on #874

jaredkerim:


app/experimenter/experiments/migrations/0034_auto_20190319_2119.py
```diff
@@ -0,0 +1,24 @@
...
+            model_name="experiment",
+            name="normandy_id",
+            field=models.PositiveIntegerField(blank=True, null=True),
+        ),
+        migrations.AlterField(
+            model_name="experiment",
+            name="client_matching",
+            field=models.TextField(
+                blank=True,
+                default="Prefs:\n\nStudies:\n\nAny additional filters:\n    ",
```
mythmon: This doesn't look right. Did this get left out of a previous PR?

jaredkerim: It suuuuurrreeeee did.  https://github.com/mozilla/experimenter/pull/1068

jaredkerim: Maybe we should add a step to continuous integration to check that no migrations got missed.

jaredkerim: Filed #1079

mythmon: I think we should, yeah. We [did it on Normandy](https://github.com/mozilla/normandy/pull/1591) recently, and it was a pretty easy check.

Do you want to leave this here and merge as is, or do you want to do something to fix this?

jaredkerim: Well It has to go in and I don't see the point of splitting it into its own tiny PR so I'm fine with this as is.

PR #1076, opened 2019-03-19
Use status labels for changelogs fixes #748

### LINKED ISSUE #748: Replace status change messages with status labels
We have this set of descriptions to describe status changes but we can just as easily use the status label in the change log and then remove those labels altogether, I think that simplifies things.

jaredkerim:
Screenshot 2019-03-19 16 34 44

### DIFF
--- a/app/experimenter/experiments/constants.py
+++ b/app/experimenter/experiments/constants.py
@@ -27,7 +27,7 @@ class ExperimentConstants(object):
     STATUS_DRAFT_LABEL = "Draft"
     STATUS_REVIEW_LABEL = "Ready for Sign-Off"
     STATUS_SHIP_LABEL = "Ready to Ship"
-    STATUS_ACCEPTED_LABEL = "Accepted by Shield"
+    STATUS_ACCEPTED_LABEL = "Accepted by Normandy"
     STATUS_LIVE_LABEL = "Live"
     STATUS_COMPLETE_LABEL = "Complete"
     STATUS_REJECTED_LABEL = "Rejected"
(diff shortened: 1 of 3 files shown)


peterbe:
I still don't think I understand how this works. Not talking about the python code, but from a *user* point of view. 

Screen Shot 2019-03-20 at 12 34 40 PM

Within the day, March 20, I assume the changes to the experiment is: most recent first. Because, that's how the dates are listed. 
This experiment is a **draft** right now but it says `Status: Ready for Sign-Off`??

Also, there are two **actions** (almost aka. "verbs") there ("Added sign-offs: ..."), and two **states** (almost aka. nouns) there ("Status: ...").

My gut feeling is that it would be easier to parse if it said something like `Changed status: <i>Draft</i> to <i>Ready for Sign-off</i>`.
Then it would all be *actions/changes*, or rather states/snapshots.

jaredkerim:
It's ordered by date, but within a given date it's the unique set of changes, not within a specific order.  I originally had tried to use more narrative statements like 'Marked as Ready to Ship', but I didn't find that it added any benefit after a while which is why I decided to change it to just stating the status, maybe if i add the verb 'Set' it'll be a little clearer.

I prefer to keep this in the model so that anywhere we use it we can expect the same representations anywhere we refer to changes.

peterbe:
> I prefer to keep this in the model so that anywhere we use it we can expect the same representations anywhere we refer to changes.

I don't agree so it's hard to approve the PR but I guess it doesn't matter. Feel free to merge. 

Also, the comments above about the History sidebar being confusing was just a gut-reaction dump of my impressions. We don't have to change it. Definitely not now! My intention was to write down what I saw with my naive eyes (today was the first time I properly even looked at the History sidebar area). We could copy-n-paste those comments to a new issue or just leave it as a lingering thought for a day when we have time to prove the UX. 

The "Set " prefix is a great improvement. 👍

jaredkerim:
@peterbe Thanks for the review, if you don't have any other blocking feedback can you approve the PR?

PR #1088, opened 2019-03-20
Add missing migration for changelogs fixes #1087

### LINKED ISSUE #1087: Missing migrations for change log fields
We merged a PR into master that had pending migrations that weren't included, which should have been caught by the new migration check in CI but possibly it wasn't included in that branch before we merged.

jaredkerim:
This renames 'Accepted by Shield' to 'Accepted by Normandy'


jaredkerim:
@peterbe it's possible I created the branch before I merged that check in.

PR #1090, opened 2019-03-20
BUGZILLA_HOST is not a host fixes #1086

### LINKED ISSUE #1086: BUGZILLA_HOST is not a host
I typed in `bugzilla-dev.allizom.org` in my `.env` and got weird errors from requests. 
The mistake in thinking it's a hostname. It's supposed to be the start of a URL otherwise the `urljoin()` doesn't work.

peterbe:
Fixes #1086

This grew surprisingly but it's nice because it gets us started on having checks. Now, next time when we want to do more system checks, we can build on this. Granted, the next time it might be `experimenter/experiments/checks.py` but at least we can copy from this.


Makefile
```diff
@@ -35,10 +35,13 @@ black_fix: test_build
 code_format: black_fix
 	echo "Code Formatted"
 
+check_checks: test_build
```
jaredkerim: I think this should be called "check_django"?

.env.test
```diff
@@ -19,6 +19,6 @@ EMAIL_HOST=
 EMAIL_PORT=
 EMAIL_HOST_USER=
 EMAIL_HOST_PASSWORD=
-BUGZILLA_HOST=
-BUGZILLA_API_KEY=
+BUGZILLA_HOST=https://bugzilla-dev.allizom.org
```
jaredkerim: Test shouldn't actually make any external calls, so I don't think we need this here.

Makefile
```diff
@@ -35,10 +35,13 @@ black_fix: test_build
...
+	docker-compose -f docker-compose-test.yml run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py check"
+
 check_migrations: test_build
 	docker-compose -f docker-compose-test.yml run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py makemigrations --check"
 
-check: test_build check_migrations black_check lint test
+check: test_build check_migrations check_checks black_check lint test
```
jaredkerim: Actually now that I think about it I don't think this belongs here at all.  This is statically checking configuration of your local development environment, it should probably go under `make up`, not here under the CI checks?

jaredkerim: Won't it run these static checks when the dev server starts up?

peterbe: Checks are automatically run when you run `./manage.py migrate` and `./manage.py runserver`. But if you're not doing any migrations, it won't run.

peterbe: That the checks are done during `migrate` is good because it ultimately checks that the env vars set up by CloudOps are correct/sane. I don't know exactly when the `migrate` command is executed but the fact that it does means that we probably raise awareness before rolling it out. 

However, we don't use `./manage.py runserver` in CI and but I believe we do execute all the migrations during `pytest` so that'll check stuff in CI. 

In other words, I agree with you. We shouldn't be mentioned ANYWHERE in the Makefile and or `.circleci/config.yml` (or any of its descendents). It's run anyway.

peterbe:
I think we need our `.env.test` to be a lot more "realistic". Otherwise the Django checks won't pass. The `.env.test` is just used straight up by `docker-compose-test.yml`.

jaredkerim:
@peterbe By 'realistic' do you mean should include some mock/valid-looking configuration for things like bugzilla?  I don't think that that's true.  If we omit those settings altogether, and some tests look to depend on them on a way that isn't properly mocked in the test and raises an exception, that's a good signal that the tests weren't set up correctly.  

We don't ever expect tests to make any actual external calls so I'm not sure why we'd want that configuration in the test settings?

peterbe:
My head is elsewhere at the moment but I quick comment...

I think what's going on is that the CircleCI runs `docker-compose.test.yml` with `.env.test` merged with `settings.py` and then it runs `python manage.py makemigrations --check --dry-run --noinput` which also triggers the checks. At that point, the `settings` object is a merge of defaults in `settings.py` + `.env.test`. 

The unit tests comes in later (as part of the `make check`) and that has the ability to override with custom for-the-sake-of-testing settings from `settings_test.py`. 

A human dev does it differently. They do something like `git clone ... && cd experimenter && cp .env.sample .env && ./manage.py migrate`. 
A human dev, would get an error about the bugzilla settings not being nice but I think I fixed that developer experience by improving the `.env.sample`. 
What happens in CircleCi is: `git clone ... && cd experimenter && docker-compose.test.yml -e .env.test ./manage.py makemigrations --check ...`

jaredkerim:
Thank you for submitting this @peterbe but I'm not comfortable adding these checks in case it adds more complexity to testing workflows.  I think the case of devs getting their local environment working is a restricted enough use case that I don't think it makes sense to operationalize for now.  If it becomes a larger problem in the future we can reopen and revisit this.

PR #1092, opened 2019-03-20
migration check creates files fixes #1091

### LINKED ISSUE #1091: migration check creates files
We missed the `--noinput` (CircleCI has no tty) and the `--dry-run` which would require a working filesystem.

peterbe:
Fixes #1091


peterbe:
I'm expecting this to fail in the first CI run because I edited the models without making a migration. Naughty!

PR #1093, opened 2019-03-20
Enforce maximum normandy slug length fixes #1078

### LINKED ISSUE #1078: Normandy slugs are max 80 characters
Normandy slugs can be no longer than 80 characters, so we need to modify the Normandy slug generator to enforce this.  It seems if we don't do this they will be truncated leading to potential collisions across experiments.

jaredkerim:
- Check the Addon Experiment ID since that will be used as the Normandy Slug verbatim for Add-on experiments
- Truncate the Experimenter slug so that the Normandy Slug is within the max for pref experiments

I didn't add a hard limit to the actual model field because there are past experiments that do have slugs that are longer than this limit and the behaviour as it relates to normandy is poorly defined, but I do want to be able to capture them accurately in our database, so we should enforce it for new experiments going forward.

### DIFF
--- a/app/experimenter/experiments/tests/test_forms.py
+++ b/app/experimenter/experiments/tests/test_forms.py
@@ -3,6 +3,7 @@ import decimal
 import json
 
 from django import forms
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.test import TestCase
 from django.utils import timezone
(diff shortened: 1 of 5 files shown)


app/experimenter/experiments/tests/test_forms.py
```diff
@@ -901,6 +902,25 @@ def test_addon_experiment_id_is_unique(self):
         self.assertFalse(form.is_valid())
         self.assertIn("addon_experiment_id", form.errors)
 
+    def test_addon_experiment_id_is_within_normandy_slug_max_len(self):
```
peterbe: "_length"?

mythmon:
> I didn't add a hard limit to the actual model field because there are past experiments that do have slugs that are longer than this limit and the behaviour as it relates to normandy is poorly defined, but I do want to be able to capture them accurately in our database, so we should enforce it for new experiments going forward.

I think this is a good pattern. :+1:

PR #1098, opened 2019-03-21
Use ChangeLog labels and add missing ones fixes #1097

### LINKED ISSUE #1097: Add additional changelog labels
Rather than ditching the idea of ditching changelog labels and just printing the name of the status, we should keep those labels because we don't want to lose useful ones like 'Created Draft' and 'Edited Draft'.  So let's revert #748 and then add additional messages to account for changes that don't currently have labels.

jaredkerim:
I changed my mind.  Rather than ditching the ChangeLog labels altogether and just emitting the status, we should keep it and just add the ones that were missing.  I thought it would be simpler to just emit the target status, but that loses two really important states: Created Experiment, Edited Experiment.  We shouldn't lose those.  So I'm reverting yesterdays change and adding the missing labels.  Sorry for all the back and forth.

### DIFF
--- a/app/experimenter/experiments/constants.py
+++ b/app/experimenter/experiments/constants.py
@@ -27,7 +27,7 @@ class ExperimentConstants(object):
     STATUS_DRAFT_LABEL = "Draft"
     STATUS_REVIEW_LABEL = "Ready for Sign-Off"
     STATUS_SHIP_LABEL = "Ready to Ship"
-    STATUS_ACCEPTED_LABEL = "Accepted by Shield"
+    STATUS_ACCEPTED_LABEL = "Accepted by Normandy"
     STATUS_LIVE_LABEL = "Live"
     STATUS_COMPLETE_LABEL = "Complete"
     STATUS_REJECTED_LABEL = "Rejected"
--- a/app/experimenter/experiments/constants.py
+++ b/app/experimenter/experiments/constants.py
@@ -27,7 +27,7 @@ class ExperimentConstants(object):
     STATUS_DRAFT_LABEL = "Draft"
     STATUS_REVIEW_LABEL = "Ready for Sign-Off"
     STATUS_SHIP_LABEL = "Ready to Ship"
-    STATUS_ACCEPTED_LABEL = "Accepted by Normandy"
+    STATUS_ACCEPTED_LABEL = "Accepted by Shield"
     STATUS_LIVE_LABEL = "Live"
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/constants.py
```diff
@@ -27,7 +27,7 @@ class ExperimentConstants(object):
     STATUS_DRAFT_LABEL = "Draft"
     STATUS_REVIEW_LABEL = "Ready for Sign-Off"
     STATUS_SHIP_LABEL = "Ready to Ship"
-    STATUS_ACCEPTED_LABEL = "Accepted by Normandy"
+    STATUS_ACCEPTED_LABEL = "Accepted by Shield"
```
peterbe: Wasn't it better with "Normandy"?

jaredkerim: The second commit changes it back: https://github.com/mozilla/experimenter/pull/1098/commits/e0906a0cf29f32579fe64119304456632c83e2f7#diff-00de22e6c754b3d410f1d3e631e1825eR30

peterbe: Oops. I just have clicked on the wrong one.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -845,15 +849,23 @@ def test_latest_returns_most_recent_changelog(self):
...
-            message=None, new_status=Experiment.STATUS_REVIEW
-        )
-        self.assertEqual(str(change), "Set Status: Ready for Sign-Off")
+        for old_status in ExperimentChangeLog.PRETTY_STATUS_LABELS.keys():
+            for new_status in ExperimentChangeLog.PRETTY_STATUS_LABELS[
+                old_status
+            ].keys():
+                expected_label = ExperimentChangeLog.PRETTY_STATUS_LABELS[
+                    old_status
+                ][new_status]
```
peterbe: Could simplify...

```python
for old_status, new_statuses in ExperimentChangeLog.PRETTY_STATUS_LABELS.items():
    for new_status, expected_label in new_statuses.items()
        changelog = ExperimentChangeLogFactory.create(
            experiment=experiment,
            old_status=old_status,
            new_status=new_status,
        )
        self.assertEqual(changelog.pretty_status, expected_label)
```

At least drop the redundant `.keys()`.

jaredkerim: Wow I don't know what I was on when I wrote that, you're right that's much more straight forward.

PR #1102, opened 2019-03-21
Make timeline optional fixes #901

### LINKED ISSUE #901: Allow null values in Overview Wizard when creating the Draft
History: The goal is to be able to start and save the draft before you know all the answers - capturing information you have as you go.  Today the validation on each section requires putting in dummy information (if you don't yet know the answers to the most critical fields) - in order to pass section by section validation on Required information.   

Ticket work: Please make it so in the "Overview Wizard" there is a way to leave the required "date" field and the "Proposed Experiment Duration (days)" field blank or "null" and still "save" the data that is there.    

Not sure if this should be part of this bug or bug #900:   Please make any Null required fields Red in the Wizard mode.  

in Bug #900 we want to throw a validation error if required fields are left empty when they click "ready for sign-offs".  It would be nice if they could go into edit mode and easily see the "red" fields.  User experience works if null fields are only "red" when they fail validation at "ready to ship" - or if they are "red from the first save that they were blank".    Putting here as not sure if that is easiest done as part of this issue or the final validation.

jaredkerim:
Screenshot 2019-03-21 16 49 17
Screenshot 2019-03-21 16 47 58

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -130,6 +130,7 @@ class ExperimentOverviewForm(
         widget=forms.Textarea(attrs={"rows": 3}),
     )
     proposed_start_date = forms.DateField(
+        required=False,
         label="Proposed Start Date",
         help_text=Experiment.PROPOSED_START_DATE_HELP_TEXT,
         widget=forms.DateInput(
@@ -137,6 +138,7 @@ class ExperimentOverviewForm(
         ),
     )
     proposed_duration = forms.IntegerField(
+        required=False,
         min_value=1,
         label="Proposed Experiment Duration (days)",
         help_text=Experiment.PROPOSED_DURATION_HELP_TEXT,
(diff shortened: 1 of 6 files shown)


app/experimenter/experiments/forms.py
```diff
@@ -184,9 +186,10 @@ def clean(self):
         cleaned_data = super().clean()
 
         # enrollment may be None
-        enrollment = cleaned_data.get("proposed_enrollment", 0) or 0
+        enrollment = cleaned_data.get("proposed_enrollment", None)
+        duration = cleaned_data.get("proposed_duration", None)
 
-        if enrollment > cleaned_data.get("proposed_duration", 0):
+        if (enrollment and duration) and enrollment > duration:
```
peterbe: nit: excess brackets

app/experimenter/experiments/forms.py
```diff
@@ -184,9 +186,10 @@ def clean(self):
         cleaned_data = super().clean()
 
         # enrollment may be None
-        enrollment = cleaned_data.get("proposed_enrollment", 0) or 0
+        enrollment = cleaned_data.get("proposed_enrollment", None)
+        duration = cleaned_data.get("proposed_duration", None)
```
peterbe: nit: excess second `None` parameter.

PR #1103, opened 2019-03-22
risk_technical and risk_technical_description invariance fixes #1082

### LINKED ISSUE #1082: The "Technical Risks Description" input field can wrongly be saved with empty spaces
**[Affected versions]:**
- Firefox Release 66.0 and above
- Chrome 73.0.3683.75

**[Affected Platforms]:** 	 	
- All Windows
- All Mac 	 	
- All Linux

**[Prerequisites]:**
- You are logged with your LDAP account on the "https://experimenter.stage.mozaws.net/" website.
- Have an experiment created on the Experimenter stage website.

**[Steps to reproduce]:**
2. Open the created experiment from prerequisites.
3. Click the "Edit" button from "Risks" section.
2. Insert an empty space in the input field.
5. Click the "Save Draft and Finish" button and observe the behavior.

**[Expected result]:**
- The "! This field is required." error message is displayed.

**[Actual result]:**
- The experiment is saved.

**[Notes]:**
TechDesc

peterbe:
Fixes #1082

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -689,6 +689,27 @@ class ExperimentRisksForm(ChangeLogMixin, forms.ModelForm):
             "qa_status",
         )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if (
+            "risk_technical" in cleaned_data
+            and "risk_technical_description" in cleaned_data
+        ):
+            # Both checked, now we need to do an invariance check on these
+            # two. This is to match what's done with jQuery in the form:
+            # the 'risk_technical_description' needs to be required
+            # if 'risk_technical' is truthy.
+            if cleaned_data.get("risk_technical") == "True":
+                if not cleaned_data["risk_technical_description"]:
+                    msg = (
+                        f"This is required if "
+                        f"'{Experiment.RISK_TECHNICAL_LABEL}' is true."
+                    )
+                    raise forms.ValidationError(
+                        {"risk_technical_description": msg}
+                    )
+        return cleaned_data
+
 
 class ExperimentReviewForm(
     ExperimentConstants, ChangeLogMixin, forms.ModelForm
--- a/app/experimenter/experiments/tests/test_forms.py
+++ b/app/experimenter/experiments/tests/test_forms.py
@@ -1081,6 +1081,46 @@ class TestExperimentRisksForm(MockRequestMixin, TestCase):
         self.assertEqual(experiment.test_builds, data["test_builds"])
         self.assertEqual(experiment.qa_status, data["qa_status"])
 
+    def test_form_risk_technical_invariance(self):
+        created_experiment = ExperimentFactory.create_with_status(
+            Experiment.STATUS_DRAFT
+        )
+
+        data = {
+            "risk_internal_only": True,
+            "risk_partner_related": True,
+            "risk_brand": True,
+            "risk_fast_shipped": True,
+            "risk_confidential": True,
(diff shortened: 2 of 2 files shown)


app/experimenter/experiments/forms.py
```diff
@@ -689,6 +689,27 @@ class Meta:
...
+        cleaned_data = super().clean()
+        if (
+            "risk_technical" in cleaned_data
+            and "risk_technical_description" in cleaned_data
+        ):
+            # Both checked, now we need to do an invariance check on these
+            # two. This is to match what's done with jQuery in the form:
+            # the 'risk_technical_description' needs to be required
+            # if 'risk_technical' is truthy.
+            if cleaned_data.get("risk_technical") == "True":
```
peterbe: It's weird to me that these are strings like that. Even after supposed cleaning. I've never seen that before. Is there a better way to do this?

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1081,6 +1081,46 @@ def test_form_saves_risks(self):
...
+        self.assertFalse(form.is_valid())
+
+        # It would be okey if 'risk_technical' was falsy.
+        data["risk_technical"] = False
+        form = ExperimentRisksForm(
+            request=self.request, data=data, instance=created_experiment
+        )
+        self.assertTrue(form.is_valid())
+
+        # And definitely OK if both at truthy.
```
jaredkerim: if both *are truthy

app/experimenter/experiments/forms.py
jaredkerim: Yeah that is weird?

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1081,6 +1081,46 @@ def test_form_saves_risks(self):
         self.assertEqual(experiment.test_builds, data["test_builds"])
         self.assertEqual(experiment.qa_status, data["qa_status"])
 
+    def test_form_risk_technical_invariance(self):
```
jaredkerim: I don't think the word 'invariance' is descriptive here, the issue is that if risk_technical is true then risk_technical_description is required, so maybe something like `test_risk_technical_description_optionally_required` would be clearer.

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1081,6 +1081,46 @@ def test_form_saves_risks(self):
...
+        )
+        self.assertTrue(form.is_valid())
+
+        # And definitely OK if both at truthy.
+        data["risk_technical"] = True
+        data["risk_technical_description"] = "Some text"
+        form = ExperimentRisksForm(
+            request=self.request, data=data, instance=created_experiment
+        )
+        self.assertTrue(form.is_valid())
```
jaredkerim: I'd rather have each of these cases tested in separate tests, and also shouldn't we be testing the case where risk_technical is true but risk_technical_description is empty and the form fails to validate?

peterbe:
I'm resigning responsibility to complete this PR. Please feel free to update and do what you want with it.

app/experimenter/experiments/forms.py
robhudson: It looks like this is because the base form field is a `ChoiceField` which coerces all values to strings. Since the underlying model for these fields is a boolean we should use `TypedChoiceField` here and provide a proper `coerce` function to make this nicer.

robhudson:
Continuation of this PR in #1148

PR #1105, opened 2019-03-22
747 intent to ship email

peterbe:
Lots of work remain. It's just a function so far. No unit tests. Also, since the body of the message is non-trivial there needs to be a lot more work within the template code to use things like `{% if experiment.risk_something %}Rist: {{ experiment.risk_something }}{% endif %}`.


jaredkerim:
I think @glasserc has started a new PR for this so I will close this one.

glasserc:
Closing in favor of #1147.

Issue #1075, opened 2019-03-19
ExperimentManager probably won't scale


peterbe:
Consider this queryset:
```python
for e in Experiment.objects.filter(name__contains='Global').only('name'):
    print(e.name)
```
Here's the SQL you'd *expect*:

```
SELECT "experiments_experiment"."id", "experiments_experiment"."name" FROM "experiments_experiment" WHERE "experiments_experiment"."name"::text LIKE '%Global%';
```

Here's the SQL it *actually requires*:

```
SELECT "experiments_experiment"."id", "experiments_experiment"."name", MAX("experiments_experimentchangelog"."changed_on") AS "latest_change" FROM "experiments_experiment" LEFT OUTER JOIN "experiments_experimentchangelog" ON ("experiments_experiment"."id" = "experiments_experimentchangelog"."experiment_id") WHERE "experiments_experiment"."name"::text LIKE '%Global%' GROUP BY "experiments_experiment"."id"

SELECT "experiments_experimentchangelog"."id", "experiments_experimentchangelog"."experiment_id", "experiments_experimentchangelog"."changed_on", "experiments_experimentchangelog"."changed_by_id", "experiments_experimentchangelog"."old_status", "experiments_experimentchangelog"."new_status", "experiments_experimentchangelog"."message" FROM "experiments_experimentchangelog" WHERE "experiments_experimentchangelog"."experiment_id" IN (22) ORDER BY "experiments_experimentchangelog"."changed_on" ASC

SELECT "experiments_experiment"."id", "experiments_experiment"."owner_id" FROM "experiments_experiment" WHERE "experiments_experiment"."id" = 22

SELECT "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined" FROM "auth_user" WHERE "auth_user"."id" IN (15)

SELECT "experiments_experimentcomment"."id", "experiments_experimentcomment"."experiment_id", "experiments_experimentcomment"."created_by_id", "experiments_experimentcomment"."created_on", "experiments_experimentcomment"."section", "experiments_experimentcomment"."text" FROM "experiments_experimentcomment" WHERE "experiments_experimentcomment"."experiment_id" IN (22) ORDER BY "experiments_experimentcomment"."created_on" ASC
```

[...]

PR #1106, opened 2019-03-22
Add text search for experiment details

### LINKED ISSUE #400: Add text search for experiment details
It'll be nice to look up all experiments done by Pocket team with a search string like "Pocket".

peterbe:
Fixes #400 

Screen Shot 2019-03-22 at 5 17 08 PM

Works. But...

* No tests at all. 
* If a freetext search is one of the active filters, it should offer to sort by ranking. 
* There's **weight** parameter sent to the `SearchVector` instance. You'd probably want to prioritize `name` over `short_description` for example. 
* Should search on things like `slug` too. 
* There's no DB index to make the search fast. But it probably doesn't need one for a long time since the database is so tiny and there [much bigger fish to fry](https://github.com/mozilla/experimenter/issues/1075).

### DIFF
--- a/app/experimenter/experiments/views.py
+++ b/app/experimenter/experiments/views.py
@@ -2,6 +2,8 @@ import django_filters as filters
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.contrib.postgres.search import SearchVector
+from django.db.models import Q
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


jaredkerim:
@peterbe Thanks for starting this! Closing it now in favour of https://github.com/mozilla/experimenter/pull/1187

Issue #1104, opened 2019-03-22
Change 'study' to 'experiment' in code/UI


jaredkerim:
Historically, experiments were called 'studies', but now it just seems inconsistent inside the Experimenter code and UI that some places we say 'experiment' and others we say 'study'.  I feel like we should converge on just using the word 'experiment'.  

The data science team has suggested that the term 'study' should be used to describe a family of experiments that test related things, so if they start adopting that then we definitely don't want to confuse the meaning.

jaredkerim:
- Here just remove the word study altogether: https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L14-L15

PR #1120, opened 2019-03-26
switched occurences of 'study' to experiment #1104

tiftran:


tiftran:
removed migrations 39 and 40
found additional "Studies" that was missed
made a single one that includes all the changes

Issue #40, opened 2017-05-02
Add 'app' user to Dockerfile


jaredkerim:
We need to add an 'app' user to the Dockerfile so the process does not run as root.

PR #1130, opened 2019-03-28
Add platform field fixes #1048

### LINKED ISSUE #1048: Add Platforms filtering
Add a new field for **Platforms** to the Experiment Population form.  It should go above the current Population Filtering box.
 
- **Platform(s)** is completely optional (not incomplete if blank).  The options should be "All Platforms, All Windows, All Mac, All Linux".

glasserc:


glasserc:
I did this first since it seems like a dependency for #747.

glasserc:
Screenshot from 2019-03-28 16-49-51
Screenshot from 2019-03-28 16-50-23

glasserc:
To be honest I'm not 100% convinced about having any fields before the population match description, it almost reads like part of the last field (in this case, Platforms).

jaredkerim:
@glasserc Oh you mean on the detail page, there's no header for 'client matching' so it just looks like a continuation of the last field?  You could just add a little <strong> title like `Additional Filtering`?

glasserc:
Yes, I think that's better..

Screenshot from 2019-03-28 17-08-39

jaredkerim:
@glasserc Can you rebase this on master and regenerate the migration so that it's #40

glasserc:
OK, done. I also added factories for the locales/countries. The new migration is numbered 40, which will conflict with #1134, but I'm happy to regenerate the migrations again if that helps.

glasserc:
OK, the locales/countries factories are not fully baked, so I'll make that it into another PR. This one will just have to be platforms.

PR #1135, opened 2019-03-29
Add factories for Country and Locale

glasserc:
Some tests previously assumed that countries/locales would be empty, so fix those too.

### DIFF
--- a/app/experimenter/experiments/tests/factories.py
+++ b/app/experimenter/experiments/tests/factories.py
@@ -8,6 +8,7 @@ from django.utils.text import slugify
 from django.utils import timezone
 from faker import Factory as FakerFactory
 
+from experimenter.base.models import Country, Locale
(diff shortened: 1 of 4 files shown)


app/experimenter/experiments/tests/factories.py
```diff
@@ -233,3 +271,33 @@ class ExperimentCommentFactory(factory.django.DjangoModelFactory):
...
+    name = factory.LazyAttribute(lambda o: faker.country())
+
+    class Meta:
+        model = Country
+
+
+class LocaleFactory(factory.django.DjangoModelFactory):
+    code = factory.Sequence(letter_code_sequence)
+    name = factory.LazyAttribute(
+        lambda o: faker.word()  # not technically correct, but sure
```
jaredkerim: ;D

PR #1143, opened 2019-04-02
change str validation for pref type

tiftran:


app/experimenter/experiments/models.py
```diff
@@ -545,7 +545,7 @@ class ExperimentVariant(models.Model):
     is_control = models.BooleanField(default=False)
     description = models.TextField(default="")
     ratio = models.PositiveIntegerField(default=1)
-    value = JSONField(blank=True, null=True)
+    value = models.CharField(max_length=10240, blank=False, null=True)
```
jaredkerim: @tiftran Ah my bad I should've said TextField here, can you switch it?

app/experimenter/experiments/models.py
```diff
@@ -4,7 +4,7 @@
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.contrib.postgres.fields import JSONField
+
```
jaredkerim: don't need an empty line here

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -993,21 +993,21 @@ def test_form_is_invalid_if_branches_have_duplicate_pref_values(self):
...
         form = ExperimentVariantsPrefForm(request=self.request, data=self.data)
         self.assertFalse(form.is_valid())
         self.assertIn("value", form.variants_formset.errors[0])
         self.assertIn("value", form.variants_formset.errors[1])
         self.assertNotIn("value", form.variants_formset.errors[2])
 
-    def test_form_is_invalid_if_pref_value_has_invalid_json(self):
-        self.data["variants-0-value"] = "abc"  # missing quotes
-
+    def test_form_is_valid_if_pref_type_is_string(self):
```
jaredkerim: Can you add two more similar tests, one for int type and one for bool type just so we have our bases covered?

app/experimenter/experiments/forms.py
```diff
@@ -515,22 +515,28 @@ def clean(self):
...
-            if not form.is_valid():
-                # This probably means that JSONField got invalid JSON.
-                continue
-            # Form field is string, but we check value type of parsed JSON.
-            found_type = type(json.loads(form.cleaned_data["value"]))
-            if found_type != expected_type:
-                form.add_error(
-                    "value", f"Unexpected value type (should be {pref_type})"
-                )
+        if pref_type in (Experiment.PREF_TYPE_BOOL, Experiment.PREF_TYPE_INT):
```
jaredkerim: 👍

app/experimenter/experiments/tests/test_forms.py
tiftran: will do

app/experimenter/experiments/models.py
tiftran: Would the max_length be needed it's a TextField?

app/experimenter/experiments/models.py
```diff
@@ -545,7 +544,7 @@ class ExperimentVariant(models.Model):
     is_control = models.BooleanField(default=False)
     description = models.TextField(default="")
     ratio = models.PositiveIntegerField(default=1)
-    value = JSONField(blank=True, null=True)
+    value = models.TextField(max_length=10240, blank=False, null=True)
```
jaredkerim: Sorry I forgot to mention, you don't need to specify the max_length here, the only thing it does is make the text field in the admin comically large 🐶 just take that out and regenerate the migration and we're good to go.

app/experimenter/experiments/models.py
jaredkerim: Oh sorry you asked here, I didn't see, no for TextField the only thing max_length does is set the size of the input field, so you usually just omit it and specify it on the actual widget (which you've done correctly already)

jaredkerim:
@tiftran Perfect, looks ready to go, just rebase the migration on master and we'll land this today :)

PR #1144, opened 2019-04-03
Remove duplicate code of conduct

jaredkerim:
The Mozilla conduct bot merged a code of conduct in, but it was already present so we accidentally had two.


jaredkerim:
I think we have to preserve the uppercase one becuase I think that's what the bot is looking for.

Issue #158, opened 2017-08-08
Add a worker for polling when experiments are published by Normandy


jaredkerim:
When an experiment is published out to Firefox by shield it appears in this API endpoint:

https://normandy.services.mozilla.com/api/v1/recipe/?format=json

We need an experimenter worker to periodically query this, look for its experiments, and flag them as 'launched' when they're found.

shell1:
We also need them to show when they are ended.  Do we need a separate issue for that?

jaredkerim:
@shell1 No we can do both in this ticket, check for 'accepted by normandy' experiments that have gone live, and check for live experiments that have been disabled.

jaredkerim:
Unblocked now that #582 is landed!

jaredkerim:
We have the relevant DB info stored, but this is blocked on the move to GCP so that we can schedule an exactly once instance of the celery beat schedule container.

jaredkerim:
Unblocked now that we've moved to GCP! 🎉

PR #1146, opened 2019-04-04
Add Normandy recipe modal fixes #498

### LINKED ISSUE #498: Add Normandy experiment information view
We need to add a screen that display the information about an experiment in a way that's laid out to be easily input into the Normandy UI.

jaredkerim:
At long last, here is the functionality for generating Normandy Recipes from Experimenter!  The work flow looks like:

When an experiment is Ready to Ship we have a new button called Transfer to Normandy:

Screenshot_2019-04-17 Up-sized grid-enabled Graphical User Interface

Clicking it brings up a modal with the recipe:

Screenshot_2019-04-17 Up-sized grid-enabled Graphical User Interface(1)
Screenshot_2019-04-17 Up-sized grid-enabled Graphical User Interface(2)

After you enter all the information provided into Delivery Console and save it, enter the recipe id and hit save and now the experiment is 'Accepted by Normandy' and the relevant Normandy information is displayed:

Screenshot_2019-04-04 Mozilla Experimenter(5)

If an invalid recipe id is entered you'll get an error like:

Screenshot_2019-04-17 Up-sized grid-enabled Graphical User Interface(3)

This will then allow us to do https://github.com/mozilla/experimenter/issues/158 so that after recipes are published in Normandy the experiment can automatically be set to live (and complete when the recipe is disabled).

I've been waiting a long time for this, I'm excited!


app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -98,24 +127,182 @@ def test_serializer_outputs_expected_schema(self):
...
+            population_percent=Decimal("12.34")
+        )
+        serializer = FilterObjectBucketSampleSerializer(experiment)
+        self.assertEqual(
+            serializer.data,
+            {
+                "type": "bucketSample",
+                "input": ["normandy.recipe.id", "normandy.userId"],
+                "start": 0,
+                "count": 123,
```
mythmon: I don't think I agree with this. I think that we should do one of the following:

* Increase the sample to be per-10,000 instead of permilli (by making count 10,00), and use 1,234 as the count
* Use 123.4 as the count.
* Not allow less than 1-in-1000 (0.1%) precision in experiments.

We recently had an experiment that wanted a .67% sample, which is fine. We did a 67 out of 10,000 sample. As implemented, we probably would have lost that detail and run a differently sized sample. It's not that bad, but we should probably be faithful in preserving those details.

app/experimenter/experiments/tests/test_api_views.py
```diff
@@ -111,6 +114,29 @@ def test_get_experiment_returns_experiment_info(self):
...
+            **{settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+
+        self.assertEqual(response.status_code, 200)
+
+        json_data = json.loads(response.content)
+
+        serialized_experiment = ExperimentRecipeSerializer(experiment).data
+
+        self.assertEqual(serialized_experiment, json_data)
```
mythmon: Stylistically, this test seems really spread out. ¯\\\_(ツ)\_/¯

app/experimenter/experiments/serializers.py
```diff
@@ -88,3 +88,148 @@ class Meta:
...
+
+    class Meta:
+        model = Experiment
+        fields = ("name",)
+
+
+class ExperimentRecipeSerializer(serializers.ModelSerializer):
+    action_name = serializers.SerializerMethodField()
+    filter_object = serializers.SerializerMethodField()
+    extra_filter_expression = serializers.ReadOnlyField(
```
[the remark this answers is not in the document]
jaredkerim: 👍

app/experimenter/experiments/serializers.py
```diff
@@ -88,3 +88,148 @@ class Meta:
...
+            "slug",
+            "experimentDocumentUrl",
+            "preferenceName",
+            "preferenceType",
+            "branches",
+        )
+
+
+class ExperimentRecipeAddonArgumentsSerializer(serializers.ModelSerializer):
+    name = serializers.ReadOnlyField(source="addon_experiment_id")
```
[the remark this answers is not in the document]
jaredkerim: Yes that just landed so I can add that now!

app/experimenter/experiments/tests/test_models.py
```diff
@@ -262,6 +266,64 @@ def test_generate_normandy_slug_is_shorter_than_max_normandy_len(self):
 
         self.assertEqual(len(normandy_slug), settings.NORMANDY_SLUG_MAX_LEN)
 
+    def test_normandy_recipe_json_serializes_pref_study(self):
+        self.maxDiff = None
```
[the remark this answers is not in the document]
jaredkerim: oops

app/experimenter/experiments/tests/test_serializers.py
jaredkerim: A long time ago I asked ilana what the smallest meaningful population size should be and and she told me 0.01% so Experimenter stores the population percent as a Decimal that allows precision to that depth, so yes we should do 10 000 as you say.

app/experimenter/experiments/tests/test_api_views.py
jaredkerim: Oh you're right, you could park a jumbo jet in there, I'll clean that up XD

shell1:
Since we don't have permissions yet, anyone could click these options.  We should subtly discourage anyone that isn't positive they should be clicking these - so they avoid sending to normandy and freezing their experiment.  It also removes some of the the requester ambiguity of "oh god, what do i need to do now".  Folks feel comfortable asking for help before hitting big-red buttons.

1) Can you change the text on the "Save changes" to be "Send to Normandy"?
2) please make the "Transfer to Normandy" button and the "Send to Normandy" button both Red.  

I'd rather have the double-red warning - and DS folks need to learn they can click the first "transfer to Normandy" to see the Jexl (but still not "send to normandy").  

Too many experiment requesters who want their thing to move forward might assume they should click the "transfer to Normandy" button - or be confused that they still need to do a step.   Once on that screen - clicking "save" is so tempting (it's always better to save).  A field for a number isn't too discouraging - since we have lots of bugzilla numbers.

PR #1154, opened 2019-04-04
Add a dummy data management command fixes #268

### LINKED ISSUE #268: Add a dummy data management command
Let's make [a management command](https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/) to populate the db with dummy data for easy testing.

You can use the test factories to make this very simple, for instance

    from experimenter.experiments.tests.factories import ExperimentFactory
    ExperimentFactory.create_with_status(Experiment.STATUS_DRAFT)

will create a fully formed experiment with all the necessary related data.  So I think for this management command, it should take two optional parameters

    python manage.py create-dummy-experiments <number of experiments> <optional status>

Number of experiments: default to 10

Optional status: If status is specified, create all experiments with that status. If no status is specified, just choose at random for each experiment by doing something like
   
    import random
    random_status = random.choice(Experiment.STATUS_CHOICES)[0]
    ExperimentFactory.create_with_status(random_status)

And then add tests and add an alias to the Makefile (look at the load-locales-countries to see an example).  Let me know if you have any questions!

tiftran:


Makefile
```diff
@@ -65,6 +65,9 @@ createuser: compose_build
 load_locales_countries: compose_build
 	docker-compose run app python manage.py load-locales-countries
 
+load_dummy_experiments: compose_build
+	docker-compose run app python manage.py load-dummy-experiments
+
```
tiftran: ooh should the optional params be available here too?

jaredkerim: @tiftran Yeah I've tried that, I couldn't figure out a nice way to do it, I wouldn't worry about it, if you wanna use the parameters you can do it from within the containers.

Also while you're at it, add a makefile command here like

    refresh: kill migrate load_locales_countries load_dummy_experiments

I feel like I do that all the time so having a shortcut would be hand ;)

PR #1174, opened 2019-04-10
Fix #597: Add statsd calls to celery tasks

### LINKED ISSUE #597: Add statsd for all celery tasks
All celery tasks need to emit statsd messages so that we can monitor they are executing correctly.

So we should add the following to each task:

- task started counter
- task completed successfully counter
- task failed counter
- task time taken to complete timer

robhudson:


.env.test
```diff
@@ -10,6 +10,10 @@ DB_HOST=db
 REDIS_HOST=redis
 REDIS_PORT=6379
 REDIS_DB=0
+STATSD_BACKEND=markus.backends.datadog.DatadogMetrics
```
robhudson: Mostly using this Python dotted path as a placeholder for what we need to add to the env for production.

app/.coveragerc
```diff
@@ -2,6 +2,8 @@
 [run]
 source = experimenter
 #branch = True
+omit =
+    */tests/*
```
robhudson: This is not required for this commit but I noticed coverage was including the tests files themselves, which is weird.

docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
     links:
```
robhudson: All the docker stuff below could be dropped if we agree that the feature of `markus` to log the statsd calls locally is sufficient. It has a logging backend.

app/experimenter/experiments/tasks.py
```diff
@@ -71,9 +80,14 @@ def create_experiment_bug_task(user_id, experiment_id):
         logger.info("Bugzilla ticket notification sent")
         raise e
 
+    metrics.incr("create_experiment_bug.completed")
```
jaredkerim: I think I'd rather have this above in the `try` block so it's only sent if the task completes successfully, otherwise to find the number of successful completes you need to do `completed - failed` right?

.env.test
```diff
@@ -10,6 +10,10 @@ DB_HOST=db
...
+STATSD_HOST=statsd
```
jaredkerim: Is specifying the actual connection params necessary for the tests to pass?  If it's not I'd rather omit it here so it's obvious that the tests don't need to connect to an actual statsd host, and move these to .env.sample

app/experimenter/experiments/tasks.py
```diff
@@ -93,5 +107,8 @@ def add_experiment_comment_task(user_id, experiment_id):
         )
         logger.info("Bugzilla comment notification sent")
     except bugzilla.BugzillaError as e:
+        metrics.incr("add_experiment_comment.failed")
         logger.info("Failed to create bugzilla comment")
         raise e
+
+    metrics.incr("add_experiment_comment.completed")
```
jaredkerim: Same here

docker-compose.yml
jaredkerim: Yeah I agree, just using the logging backend is sufficient we don't need to use the echo server 👍

app/experimenter/experiments/tasks.py
robhudson: That makes sense. It would then follow the other code as well. The way I read it the exception would `raise` so that completed `incr` wouldn't get sent anyway. But it follows more logically if it was in the try block. 👍

.env.test
robhudson: I'll remove it... it's not necessary. If it complains I can just use the logging backend.

robhudson:
Updated based on review feedback.

Issue #745, opened 2019-01-03
Update Sign-offs list


shell1:
Change “Save Reviews” button to “Save Sign-offs” button

See the doc for a visual  - all the links are from the same section in the main mana page.
https://docs.google.com/document/d/1LknFtapAZ2OQtXzufRlgwwHjDrkCrQHucXgBTzDlZ5g/edit#heading=h.sac2t61vrqel

Mana help page: https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments

Filing separate bug for each email (QA requested and Intent to Ship).

**Optional** 
Lightning Advisory Help 
----------------------------------------------------
**Required Sign-offs**
Data Science Peer Review  Help
Engineering allocated  Help
QA Requested  Help & Send Email
Intent to Ship email Sent  Help & Send Email
Bugzilla updated  Help
QA Signoff  Help
Relman Signoff  Help
-------------------------------------------------
**Dependent Sign-offs** Help
Legal review  Help
UX review  Help
Security review  Help
VP approval  Help
Data Steward review Help
Mozilla press/comms review  Help

PR #1175, opened 2019-04-10
dynamic checklist generation for partner related risk fixes #856

### LINKED ISSUE #856: Checklist question 1: Is this study partner related?
"Yes" radial button selection for this question - should make "VP Sign-off" move to "required".  Required means they need to check it before they can click "ready for shipping".  

Help for this question should take them to the risks section.  https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-Risk

The checklist item that should be added is "VP Sign-off".  (we are changing from VP Review to VP Sign-off - because the VPs want to click the box themselves). The help next to the actual checklist item should go to the main Dependent section.  Linking there to avoid future name changes breaking the link.  https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-DependentSign-offs

tiftran:
Screenshot of the reviews checklist with new headings:
Screen Shot 2019-04-15 at 2 10 03 PM

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -788,42 +788,42 @@ class ExperimentReviewForm(
     # Optional
     review_advisory = forms.BooleanField(
         required=False,
-        label="Lightning Advisory (Optional)",
+        label="Lightning Advisory",
         help_text=Experiment.REVIEW_ADVISORY_HELP_TEXT,
     )
     review_legal = forms.BooleanField(
         required=False,
-        label="Legal Review (Optional)",
+        label="Legal Review",
         help_text=Experiment.REVIEW_LEGAL_HELP_TEXT,
     )
     review_ux = forms.BooleanField(
         required=False,
-        label="UX Review (Optional)",
+        label="UX Review",
         help_text=Experiment.REVIEW_UX_HELP_TEXT,
     )
     review_security = forms.BooleanField(
         required=False,
-        label="Security Review (Optional)",
+        label="Security Review",
         help_text=Experiment.REVIEW_SECURITY_HELP_TEXT,
     )
     review_vp = forms.BooleanField(
         required=False,
-        label="VP Review (Optional)",
+        label="VP Sign Off",
         help_text=Experiment.REVIEW_VP_HELP_TEXT,
     )
     review_data_steward = forms.BooleanField(
         required=False,
-        label="Data Steward Review (Optional)",
+        label="Data Steward Review",
         help_text=Experiment.REVIEW_DATA_STEWARD_HELP_TEXT,
     )
     review_comms = forms.BooleanField(
         required=False,
-        label="Mozilla Press/Comms (Optional)",
+        label="Mozilla Press/Comms",
         help_text=Experiment.REVIEW_COMMS_HELP_TEXT,
     )
     review_impacted_teams = forms.BooleanField(
         required=False,
-        label="Impacted Team(s) Signed-Off (Optional)",
+        label="Impacted Team(s) Signed-Off",
         help_text=Experiment.REVIEW_IMPACTED_TEAMS_HELP_TEXT,
     )
 
@@ -850,54 +850,22 @@ class ExperimentReviewForm(
         )
 
     @property
-    def _required_reviews_mapping(self):
-        return {"review_vp": self.instance.risk_partner_related}
-
-    @property
     def required_reviews(self):
-        default_required = [
-            self["review_science"],
-            self["review_advisory"],
-            self["review_engineering"],
-            self["review_qa_requested"],
-            self["review_intent_to_ship"],
-            self["review_bugzilla"],
-            self["review_qa"],
-            self["review_relman"],
-        ]
-        return self.add_conditional_reviews(default_required)
+        required_review_names = self.instance.get_all_required_reviews()
+        required_reviews = []
+        for review_name in required_review_names:
+            required_reviews.append(self[review_name])
 
-    def add_conditional_reviews(self, required_reviews):
-        for review, risk in self._required_reviews_mapping.items():
-            if risk:
-                self._remove_optional_label(review)
-                self._change_help_text_to_risk_help(review)
-                required_reviews.append(self[review])
         return required_reviews
-
-    def _remove_optional_label(self, review):
-        self[review].label = self[review].label.strip("(Optional)")
-
-    def _change_help_text_to_risk_help(self, review):
-        self[review].help_text = Experiment.RISK_GENERAL_HELP_TEXT
 
     @property
     def optional_reviews(self):
-        default_optional = [
-            self["review_legal"],
-            self["review_ux"],
-            self["review_security"],
-            self["review_vp"],
-            self["review_data_steward"],
-            self["review_comms"],
-            self["review_impacted_teams"],
-        ]
-        return self._remove_optional_reviews(default_optional)
+        required_review_names = self.instance.get_all_required_reviews()
+        optional_reviews = []
+        for review_name in self.Meta.fields:
+            if review_name not in required_review_names:
+                optional_reviews.append(self[review_name])
 
-    def _remove_optional_reviews(self, optional_reviews):
-        for review, bool_value in self._required_reviews_mapping.items():
-            if bool_value:
-                optional_reviews.remove(self[review])
         return optional_reviews
 
     @property
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -850,8 +850,12 @@ class ExperimentReviewForm(
         )
 
     @property
+    def _required_reviews_mapping(self):
+        return {"review_vp": self.instance.risk_partner_related}
+
+    @property
     def required_reviews(self):
-        return (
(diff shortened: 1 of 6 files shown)


app/experimenter/experiments/forms.py
```diff
@@ -860,19 +864,41 @@ def required_reviews(self):
...
-        )
+        ]
+        return self.add_conditional_reviews(default_required)
+
+    def add_conditional_reviews(self, required_reviews):
+        for review, risk in self._required_reviews_mapping.items():
+            if risk:
+                self._remove_optional_label(review)
+                self._change_help_text_to_risk_help(review)
+                required_reviews.append(self[review])
```
glasserc: I'm always a little hesitant about mutating arguments to a function. In this particular instance we can get away with it because `required_reviews` is newly created, but I'd feel more comfortable doing a shallow copy of `required_reviews` at the beginning of the function.

app/experimenter/experiments/forms.py
```diff
@@ -860,19 +864,41 @@ def required_reviews(self):
...
+        return required_reviews
+
+    def _remove_optional_label(self, review):
+        self[review].label = self[review].label.strip("(Optional)")
```
glasserc: Do these mutate `self` or the fields in a way that exceeds the lifetime of the form? From a quick glance, it seems like they would, but maybe I'm missing something about how Django forms work.

app/experimenter/experiments/forms.py
```diff
@@ -860,19 +864,41 @@ def required_reviews(self):
...
             self["review_comms"],
             self["review_impacted_teams"],
-        )
+        ]
+        return self._remove_optional_reviews(default_optional)
+
+    def _remove_optional_reviews(self, optional_reviews):
+        for review, bool_value in self._required_reviews_mapping.items():
+            if bool_value:
+                optional_reviews.remove(self[review])
```
glasserc: Same concern about mutating `optional_reviews`.

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1270,7 +1270,34 @@ def test_required_reviews(self):
                 form["review_bugzilla"],
                 form["review_qa"],
                 form["review_relman"],
-            ),
+            ],
+        )
+
+    def test_required_reviews_when_a_risk_option_is_true(self):
```
glasserc: Instead of `a_risk_option`, how about `risk_partner_related`?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -832,6 +832,21 @@ def test_is_ready_to_launch_true_when_reviews_and_sections_complete(self):
...
+            review_engineering=True,
+            review_qa_requested=True,
+            review_intent_to_ship=True,
+            review_bugzilla=True,
+            review_qa=True,
+            review_relman=True,
+            review_vp=True,
+            risk_partner_related=True,
+        )
+        self.assertTrue(experiment.is_ready_to_launch)
```
glasserc: Maybe we should have a test that `is_ready_to_launch` is `False` when `risk_partner_related` is `True` but `review_vp` is `False`?

app/experimenter/experiments/forms.py
```diff
@@ -860,19 +864,41 @@ def required_reviews(self):
...
+                self._remove_optional_label(review)
+                self._change_help_text_to_risk_help(review)
+                required_reviews.append(self[review])
+        return required_reviews
+
+    def _remove_optional_label(self, review):
+        self[review].label = self[review].label.strip("(Optional)")
+
+    def _change_help_text_to_risk_help(self, review):
+        self[review].help_text = Experiment.RISK_GENERAL_HELP_TEXT
```
glasserc: I'm not sure I completely understand this. My reading of the linked issue is that the link you have here is for the `Is this partner related?` question, but that the review should instead link to https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-DependentSign-offs instead.

glasserc: Thinking about this more, especially since the linked issue requires the text of the checklist item to change, it may not be possible to do this in a generic way (the way you have).

tiftran: my bads, I definitely read the issue wrong

app/experimenter/experiments/tests/test_models.py
tiftran: will do

app/experimenter/experiments/forms.py
```diff
@@ -832,14 +832,14 @@ class Meta:
...
             "review_science",
-            "review_advisory",
             "review_engineering",
             "review_qa_requested",
             "review_intent_to_ship",
             "review_bugzilla",
             "review_qa",
             "review_relman",
             # Optional
+            "review_advisory",
```
tiftran: @jaredkerim I moved "review_advisory" to optional and edited tests to reflect that since it was optional in models.py.... wasn't sure if that was intentional... could only find #745 which seems to suggest that it should be optional. Will revert if this was intentional

app/experimenter/experiments/forms.py
```diff
@@ -851,28 +851,22 @@ class Meta:
...
-            self["review_science"],
-            self["review_advisory"],
-            self["review_engineering"],
-            self["review_qa_requested"],
-            self["review_intent_to_ship"],
-            self["review_bugzilla"],
-            self["review_qa"],
-            self["review_relman"],
-        )
+        required_review_names = self.instance.get_all_required_reviews()
```
jaredkerim: I think you could shorten this to just

    return [self[r] for r in self.instance.get_all_required_reviews()]

app/experimenter/experiments/models.py
```diff
@@ -477,20 +477,34 @@ def completed_testing(self):
...
+            "review_engineering",
+            "review_qa_requested",
+            "review_intent_to_ship",
+            "review_bugzilla",
+            "review_qa",
+            "review_relman",
+        ]
+
+    def get_all_required_reviews(self):
+        required_reviews = self._default_required_reviews
```
jaredkerim: I get what you're saying here, but in this case using `@property` makes this a little less clear, so I'd just treat it like a regular method to make it clear you're calling it and getting back a new instance of the review list.

app/experimenter/experiments/models.py
```diff
@@ -477,20 +477,34 @@ def completed_testing(self):
...
+            if risk:
+                required_reviews.append(review)
+        return required_reviews
 
     @property
     def completed_required_reviews(self):
-        return all(self._required_reviews)
+        required_review = self.get_all_required_reviews()
+        return all(
+            list(map(lambda review: getattr(self, review), required_review))
```
jaredkerim: In general I'd avoid using map/lambda and just stick to list comprehensions:

    return all([getattr(self, r) for r in self.get_all_required_reviews()])

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1284,15 +1309,41 @@ def test_optional_reviews(self):
...
+            Experiment.STATUS_REVIEW,
+            review_relman=True,
+            review_science=True,
+            risk_partner_related=True,
+        )
+
+        form = ExperimentReviewForm(
+            request=self.request, data={}, instance=experiment
+        )
+
```
jaredkerim: I would just check that the 'review_vp' moved from the optional list to the required list here:

    self.assertIn(form["review_vp"], form.required_reviews)
    self.assertNotIn(form["review_vp"], form.optional_reviews)

The other test above already checks that the whole optional list looks as expected so you don't need to check the whole thing again.

app/experimenter/experiments/tests/test_models.py
jaredkerim: Yes in general both those tests are good ideas, we don't want to recreate those tests for every single pair of risk/review, so I think testing it once here with both   

    test_ready_to_launch_false_without_conditional_review
    test_ready_to_launch_true_with_conditional_review

should cover things well.

app/experimenter/experiments/forms.py
```diff
@@ -851,28 +851,22 @@ class Meta:
...
-        return (
-            self["review_legal"],
-            self["review_ux"],
-            self["review_security"],
-            self["review_vp"],
-            self["review_data_steward"],
-            self["review_comms"],
-            self["review_impacted_teams"],
-        )
+        required_review_names = self.instance.get_all_required_reviews()
```
jaredkerim: I think you can just write this as

    return list(set(self.Meta.fields) - set(self.instance.get_all_required_reviews()))

app/experimenter/experiments/forms.py
jaredkerim: @tiftran Yeah it was always optional, but there was this idea of the top sign-offs being 'ordered' in a certain sense (loosely) like you should go through them in the order they're listed.  But now that we're making a strong distinction between required and optional I think that idea goes out the window, so yes you're correct to move it to optional the way you have.

PR #1181, opened 2019-04-12
Add date range filter to view fixes #799

### LINKED ISSUE #799: Normandy work list - Add 2 date ranges to Filter area
Add 2 date fields in the main listing page to capture any dates in that range (start, enrollment change date, or end date).

Return all studies that have any dates these between those 2 dates: start, enrollment, or experiment end dates.

It should be combine-able with existing filter options.  The main uses cases:
1) Normandy team:  look at all "Live" studies with start, enrollment change, or end date between ___ and ____.   This shows all things that “need action” in a Normandy change window.
2) Normandy team:  look at all "ready to ship" studies with any dates between ___ and ___.  This shows all studies that should launch for a Normandy change window.

Use cases:
1. Shell will be pulling mid week before to do manual checking.  
2. Normandy may pull week before to do final check on how the experiment lines up with recipes to make sure they are clear.
3. Normandy can pull the week of and look at what is ready for change in the normandy change window.

uhlissuh:
Fixes #799 

With this PR I'm adding a new feature that allows a user to search for experiments by date range (after a certain date but before another) or by a single date (i.e. all experiments that pause after date X). 

Screen Shot 2019-04-18 at 3 52 59 PM

If users fill in one date but not the other, we assume they want all experiments up to or after that date.

date-range-demo

We make sure the title of the page reflects the way in which the user is interacting with the date range filter. Below, the user is using both dates to create a range:
Screen Shot 2019-04-18 at 3 57 00 PM
 

And here, the user wants all experiments after a date and until the end of time. 

Screen Shot 2019-04-18 at 3 57 47 PM

### DIFF
--- a/app/experimenter/experiments/tests/test_views.py
+++ b/app/experimenter/experiments/tests/test_views.py
@@ -298,6 +298,164 @@ class TestExperimentListView(TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(list(context["experiments"]), list(experiments))
 
+    def set_up_date_tests(self):
+
+        self.exp_1 = ExperimentFactory.create_with_status(
+            Experiment.STATUS_DRAFT,
+            name="experiment 1",
+            proposed_start_date=datetime.date(2019, 4, 5),
+            proposed_duration=30,
+            proposed_enrollment=3,
+        )
+
+        self.exp_2 = ExperimentFactory.create_with_status(
+            Experiment.STATUS_DRAFT,
+            name="experiment 2",
+            proposed_start_date=datetime.date(2019, 3, 29),
+            proposed_duration=14,
+            proposed_enrollment=4,
+        )
+
+        self.exp_3 = ExperimentFactory.create_with_status(
+            Experiment.STATUS_DRAFT,
+            name="experiment 3",
+            proposed_start_date=datetime.date(2019, 5, 29),
+            proposed_duration=30,
+            proposed_enrollment=None,
+        )
+
+        self.exp_4 = ExperimentFactory.create_with_status(
+            Experiment.STATUS_DRAFT,
+            name="experiment 4",
+            proposed_start_date=datetime.date(2019, 3, 15),
+            proposed_duration=25,
+            proposed_enrollment=1,
+        )
+
+        self.start_range_date = "2019-04-01"
+        self.end_range_date = "2019-05-01"
+        self.user_email = "user@example.com"
+
+    def test_list_shows_all_experiments_with_start_in_range(self):
+
+        self.set_up_date_tests()
+
+        response = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"),
+                params=urlencode(
+                    {
+                        "experiment_date_field": Experiment.EXPERIMENT_STARTS,
+                        "date_range_after": self.start_range_date,
+                        "date_range_before": self.end_range_date,
+                    }
+                ),
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: self.user_email},
+        )
+
+        context = response.context[0]
+
+        self.assertEqual(set(context["experiments"]), set([self.exp_1]))
+
+    def test_list_shows_all_experiments_with_pause_in_range(self):
+        self.set_up_date_tests()
+
+        response = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"),
+                params=urlencode(
+                    {
+                        "experiment_date_field": Experiment.EXPERIMENT_PAUSES,
+                        "date_range_after": self.start_range_date,
+                        "date_range_before": self.end_range_date,
+                    }
+                ),
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: self.user_email},
+        )
+
+        context = response.context[0]
+
+        self.assertEqual(
+            set(context["experiments"]), set([self.exp_1, self.exp_2])
+        )
+
+    def test_list_shows_all_experiments_with_end_in_range(self):
+        self.set_up_date_tests()
+
+        response = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"),
+                params=urlencode(
+                    {
+                        "experiment_date_field": Experiment.EXPERIMENT_ENDS,
+                        "date_range_after": self.start_range_date,
+                        "date_range_before": self.end_range_date,
+                    }
+                ),
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: self.user_email},
+        )
+
+        context = response.context[0]
+
+        self.assertEqual(
+            set(context["experiments"]), set([self.exp_2, self.exp_4])
+        )
+
+    def test_list_shows_all_experiments_with_start_in_range_start_date_only(
+        self
+    ):
+
+        self.set_up_date_tests()
+
+        response = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"),
+                params=urlencode(
+                    {
+                        "experiment_date_field": Experiment.EXPERIMENT_STARTS,
+                        "date_range_after": self.start_range_date,
+                        "date_range_before": "",
+                    }
+                ),
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: self.user_email},
+        )
+
+        context = response.context[0]
+
+        self.assertEqual(
+            set(context["experiments"]), set([self.exp_1, self.exp_3])
+        )
+
+    def test_list_shows_all_experiments_with_start_in_range_end_date_only(
+        self
+    ):
+        self.set_up_date_tests()
+
+        response = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"),
+                params=urlencode(
+                    {
+                        "experiment_date_field": Experiment.EXPERIMENT_STARTS,
+                        "date_range_after": "",
+                        "date_range_before": self.end_range_date,
+                    }
+                ),
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: self.user_email},
+        )
+
+        context = response.context[0]
+
+        self.assertEqual(
+            set(context["experiments"]),
+            set([self.exp_1, self.exp_2, self.exp_4]),
+        )
+
     def test_list_view_shows_all_including_archived(self):
         user_email = "user@example.com"
 
--- a/app/experimenter/experiments/views.py
+++ b/app/experimenter/experiments/views.py
@@ -12,6 +12,7 @@ from django.contrib.postgres.search import (
     SearchRank,
     SearchVector,
 )
+import django_filters.widgets as widgets
 
 from experimenter.projects.models import Project
 from experimenter.experiments.forms import (
@@ -72,8 +73,32 @@ class ExperimentFiltersetForm(forms.ModelForm):
         if user_id is not None:
             return str(get_user_model().objects.get(id=user_id))
 
+    def get_display_start_date_info(self):
+        experiment_date_field = self.data.get("experiment_date_field")
+        date_after = self.data.get("date_range_after")
+        date_before = self.data.get("date_range_before")
+
+        if date_after and date_before:
+            return (
+                f"{experiment_date_field} between "
+                f"{date_after} and {date_before}"
+            )
+        elif date_after and date_before == "":
+            return f"{experiment_date_field} after {date_after}"
+        elif date_after == "" and date_before:
+            return f"{experiment_date_field} before {date_before}"
+        else:
+            return ""
+
+
+# the default widget has a dash character between the two date fields,
+# and what we want is the word "To", so we are making our own widget here
+class DateRangeWidget(widgets.DateRangeWidget):
+    template_name = "widgets/date_range.html"
+
 
 class ExperimentFilterset(filters.FilterSet):
+
     search = filters.CharFilter(
         method="filter_search",
         widget=forms.widgets.TextInput(
@@ -84,7 +109,6 @@ class ExperimentFilterset(filters.FilterSet):
             }
         ),
     )
-
     type = filters.ChoiceFilter(
         empty_label="All Types",
         choices=Experiment.TYPE_CHOICES,
@@ -118,6 +142,22 @@ class ExperimentFilterset(filters.FilterSet):
     archived = filters.BooleanFilter(
         label="Show archived experiments", widget=forms.CheckboxInput()
     )
+    experiment_date_field = filters.ChoiceFilter(
+        empty_label="No Date Restriction",
+        choices=[
+            (Experiment.EXPERIMENT_STARTS, "Experiment Starts"),
+            (Experiment.EXPERIMENT_PAUSES, "Experiment Pauses"),
+            (Experiment.EXPERIMENT_ENDS, "Experiment Ends"),
+        ],
+        widget=forms.Select(attrs={"class": "form-control"}),
+        method="experiment_date_field_filter",
+    )
+    date_range = filters.DateFromToRangeFilter(
+        method="date_range_filter",
+        widget=DateRangeWidget(
+            attrs={"type": "date", "class": "form-control"}
+        ),
+    )
 
     class Meta:
         model = Experiment
@@ -150,6 +190,38 @@ class ExperimentFilterset(filters.FilterSet):
             .filter(search=value)
             .order_by("-rank")
         )
+
+    def experiment_date_field_filter(self, queryset, name, value):
+        # this custom method isn't doing anything. There has to
+        # be a custom method to be able to display the select
+        # filter that controls which date range we show
+        return queryset
+
+    def date_range_filter(self, queryset, name, value):
+
+        date_type = self.form.cleaned_data["experiment_date_field"]
+
+        experiment_date_field = {
+            Experiment.EXPERIMENT_STARTS: "start_date",
+            Experiment.EXPERIMENT_PAUSES: "enrollment_end_date",
+            Experiment.EXPERIMENT_ENDS: "end_date",
+        }[date_type]
+
+        results = []
+
+        for experiment in queryset.all():
+            date = getattr(experiment, experiment_date_field)
+
+            # enrollment end dates are optional, so there won't always
+            # be a pause date for an experiment
+            if date:
+                if value.start and date < value.start.date():
+                    continue
+                if value.stop and date > value.stop.date():
+                    continue
+                results.append(experiment.id)
+
+        return queryset.filter(pk__in=results)
 
 
 class ExperimentOrderingForm(forms.Form):
--- a/app/experimenter/static/js/experiment-date-filter.js
+++ b/app/experimenter/static/js/experiment-date-filter.js
@@ -1,0 +1,19 @@
+let dateRangeSelector = document.getElementById('id_experiment_date_field');
+let rangeElementInput0 = document.getElementById("id_date_range_0");
+let rangeElementInput1 = document.getElementById("id_date_range_1");
+
+
+function setUpDateRange () {
+  if (dateRangeSelector.value == '') {
+    rangeElementInput0.setAttribute("disabled", "")
+    rangeElementInput1.setAttribute("disabled", "")
+  } else {
+    rangeElementInput0.removeAttribute("disabled");
+    rangeElementInput1.removeAttribute("disabled");
+  }
+}
+
+setUpDateRange();
+
+
+dateRangeSelector.addEventListener("change", setUpDateRange);
--- a/app/experimenter/templates/experiments/list.html
+++ b/app/experimenter/templates/experiments/list.html
@@ -39,6 +39,9 @@
       with text "{{ filter.form.search.value }}"
     {% endif %}
 
+    {{ filter.form.get_display_start_date_info }}
+
+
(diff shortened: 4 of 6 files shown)


app/experimenter/settings.py
```diff
@@ -46,6 +46,7 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "django.forms",
+    "django_filters",
```
glasserc: I'm sort of on the fence about this -- we probably *should* have this here, but we got by without it, and now the only thing it provides is a template that we override, so do we really need it?

app/experimenter/templates/experiments/list.html
```diff
@@ -134,16 +137,46 @@ <h5>{{ experiment.population }}</h5>
...
+            From
+          </label>
+          {{ field }}
+        </div>
+
+      {% elif field.name == "end_date_range"%}
+        <div id="end-both-date-inputs" class="form-group">
+          <label class="date-range-label">
+            From
+          </label>
```
glasserc: I feel like it might have been better to put the `From` label in the template too (although this is less important if we only have one such field).

app/experimenter/experiments/views.py
```diff
@@ -100,11 +143,75 @@ class ExperimentFilterset(filters.FilterSet):
         label="Show archived experiments", widget=forms.CheckboxInput()
     )
 
+    show_experiment_changes = filters.ChoiceFilter(
```
glasserc: Nit: a better name might be `experiment_date_field`

app/experimenter/experiments/tests/test_views.py
```diff
@@ -220,6 +220,136 @@ def test_list_view_lists_experiments_with_default_order_no_archived(self):
...
+                ),
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+
+        context_5 = response_5.context[0]
+
+        self.assertEqual(
+            set(context_5["experiments"]), set([exp_1, exp_2, exp_4])
+        )
```
glasserc: Nit: this might be better as a separate `TestCase`, with all the experiment creation done in the `setUp`/`setUpClass` and each request a separate method. But I see you're following the pattern of other tests so it's not a big deal.

app/experimenter/experiments/views.py
```diff
@@ -25,6 +26,10 @@
 
 
 class ExperimentFiltersetForm(forms.ModelForm):
+    show_experiment_changes = forms.ChoiceField(required=False)
```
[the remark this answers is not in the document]
uhlissuh: So I put this in here because that's what I thought you had to do to be able to get it into the `fields` section in the `Meta` class on the form... which is what appears to be determining the order of the filters in the whole filter area. I learned this from the search text area PR #1187. But now that you point this out.. I'm realizing that I don't care about this because right now all the date stuff lives at the bottom of the search area anyway.. and if I comment all this out, it works just fine. I think I could do away with this stuff, and the corresponding lines in the `fields` section. I have to admit it was always magic to me why this worked for the search text area I added in PR #1187 .

app/experimenter/experiments/views.py
```diff
@@ -100,11 +143,75 @@ class ExperimentFilterset(filters.FilterSet):
...
+        results = []
+        date = None
+
+        for experiment in experiments:
+            if name == "start_date_range":
+                date = experiment.start_date
+            elif name == "end_date_range":
+                date = experiment.end_date
+            elif name == "pause_date_range":
+                date = experiment.enrollment_end_date
```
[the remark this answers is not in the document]
uhlissuh: Yes, there was much talk/effort about doing this in the DB instead of loading the experiments into memory and then using the properties... and I definitely think it's theoretically possible... but it was proving pretty challenging.. I could definitely still hack away at that if we want me to.

uhlissuh:
@glasserc Firstly, thanks! So about the three separate filters thing (for start, pause, and end), the reason I arrived at this was that I don't know how to access filter values from a different filter. So I don't know how to access the value of `show_experiment_changes` filter within the date range filter. Do you have any ideas on this? It would be REALLY awesome to be able to do this.

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
glasserc: I don't know what Jared thinks -- personally, I don't think it's necessary at the scale that Experimenter works at, as long as there's a comment that expresses why we're doing it this way.

However, in thinking about this a little bit more, I think it would be fairly easy (as a minor optimization) do the "start date" comparison using the ORM, and just do the end date stuff in Python. What do you think?

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
glasserc: Hmm. I'm looking through the django-filter code now and I see https://github.com/carltongibson/django-filter/blob/24adad8c48bc9e7c7539b6510ffde4ce4effdc29/django_filters/filterset.py#L332-L334 -- maybe it will give you a warning if you comment it out?

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
glasserc: One other minor tweak here is that maybe it would be good to extract the mapping of filter name -> attribute name outside the loop. Something maybe like this:

```
attributes = {
    "start_date_range": "start_date",
    "end_date_range": "end_date",
    "pause_date_range": "enrollment_end_date"
}
attribute = attribute_names[name]

for experiment in experiments:
    date = getattr(experiment, attribute)
    # ...
```

glasserc:
Ohhh, I see. OK, I had to read up on django-filter a little bit to make sense of what was going on. OK, here are some thoughts.

- According to [[1](https://django-filter.readthedocs.io/en/master/guide/usage.html?highlight=request#request-based-filtering)], you can access the `request` (if present). Looking at [the code](https://github.com/carltongibson/django-filter/blob/master/django_filters/filterset.py) it looks like it's available as `self.request`. From there I guess you could do the usual stuff to get request parameters, validate them, etc.
- We could maybe override [`get_filters`](https://github.com/carltongibson/django-filter/blob/master/django_filters/filterset.py#L314) or [`filter_queryset`](https://github.com/carltongibson/django-filter/blob/master/django_filters/filterset.py#L215), or maybe even subclass [`Filter`](https://github.com/carltongibson/django-filter/blob/24adad8c48bc9e7c7539b6510ffde4ce4effdc29/django_filters/filters.py#L65).
- We could lean into this framework and support filtering by all the different options at once. This seems more powerful although I admit it may make the UI too complicated.
- We could still use the (current) multiple-date-range approach but simplify the HTML by just changing the field `name`s when the `<select>` changes.
- If we do use the multiple-date-range approach, it might be cleaner to metaprogram the `date_range_filter` method -- have a different method for each and just call `date_range_filter` with a `attr` argument, something like this:

```
    start_date_range = filters.DateFromToRangeFilter(
        method="start_date_range_filter",
        widget=DateRangeWidget(
            attrs={"type": "date", "class": "form-control range-field"}
        ),
    )

    # ...
    def start_date_range_filter(self, queryset, name, value):
        return self.date_range_filter(queryset, "start_date", value)
```

I don't have strong feelings about which way is best. I kind of like the possibility of being able to filter according to multiple dates at once (even though it's kind of overkill for this project), but I don't love how complex it is, in particular the templates and hide/show stuff in JS.

uhlissuh:
@glasserc @jaredkerim I went with the first bullet option and used the request. I like it a lot. Gets it down to one date filter and simplifies the JS a little too. (On the 3rd bullet point, I implemented it this way initially and @jaredkerim and @shell1 didn't love the UI).

app/experimenter/experiments/views.py
uhlissuh: Made this change!

app/experimenter/templates/experiments/list.html
uhlissuh: Made this change!

app/experimenter/settings.py
uhlissuh: Made this change!

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
uhlissuh: I actually haven't seen any warnings with this stuff taken out. I am going to leave it out since it apparently doesn't do anything.

app/experimenter/experiments/views.py
```diff
@@ -25,6 +26,10 @@
 
 
 class ExperimentFiltersetForm(forms.ModelForm):
+    show_experiment_changes = forms.ChoiceField(required=False)
+    start_date_range = forms.DateField(required=False)
+    pause_date_range = forms.DateField(required=False)
+    end_date_range = forms.DateField(required=False)
```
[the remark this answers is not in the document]
uhlissuh: This is taken care of with the removal of this code, discussed in another section.

app/experimenter/experiments/views.py
```diff
@@ -72,6 +73,39 @@ def get_owner_display_value(self):
...
+            date_after = self.data.get("date_range_after")
+            date_before = self.data.get("date_range_before")
+            type = "starting"
+        elif experiment_date_field == "pauses_between":
+            date_after = self.data.get("date_range_after")
+            date_before = self.data.get("date_range_before")
+            type = "pausing"
+        elif experiment_date_field == "ends_between":
+            date_after = self.data.get("date_range_after")
+            date_before = self.data.get("date_range_before")
```
glasserc: I think you can extract the `date_after` and `date_before` assignments to above the if statement.

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
glasserc: I still think it would be good to move the logic about which field to look at outside the loop, since it doesn't change on different iterations of the loop.

app/experimenter/static/js/experiment-date-filter.js
```diff
@@ -0,0 +1,27 @@
+let dateRangeSelector = document.getElementById('id_experiment_date_field');
+let rangeElementInput0 = document.getElementById("id_date_range_0");
+let rangeElementInput1 = document.getElementById("id_date_range_1");
+console.log(dateRangeSelector);
```
glasserc: I think we can get rid of this.

app/experimenter/static/js/experiment-date-filter.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+function setUpDateRange () {
+  if (dateRangeSelector.value == 'starts_between') {
+    enableBothInputs();
+  } else if (dateRangeSelector.value == 'pauses_between') {
+    enableBothInputs();
+  } else if (dateRangeSelector.value == 'ends_between') {
+    enableBothInputs();
+  } else {
```
glasserc: It might be easier to check for the default "empty" value and enable the fields otherwise.

app/experimenter/templates/experiments/list.html
```diff
@@ -138,16 +141,37 @@ <h5>{{ experiment.population }}</h5>
...
+
+      {% elif field.name == "pause_date_range" %}
+        <div id="pause-both-date-inputs" class="form-group">
+          {{ field }}
+        </div>
+
+      {% elif field.name == "end_date_range"%}
+        <div id="end-both-date-inputs" class="form-group">
+          {{ field }}
+        </div>
```
glasserc: I think we can get rid of all this now?

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
uhlissuh: I definitely thought it over but it's hard for me to see how this is an improvement.. do you think it's a better performance thing?

app/experimenter/static/js/experiment-date-filter.js
uhlissuh: oops yes.

app/experimenter/templates/experiments/list.html
uhlissuh: totally, thank you.

app/experimenter/experiments/views.py
uhlissuh: yea definitely, super oops.

app/experimenter/experiments/tests/test_views.py
uhlissuh: @glasserc did you have any thoughts on the new tests written? @jaredkerim preferred me not to make a separate class but instead make a custom `set_up_date_tests` method to call in each different method.

glasserc: That's fine with me.

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
glasserc: I think the performance difference is likely to be negligible. Mostly I just think it feels cleaner to put stuff in the loop only when it needs to rerun on each loop iteration. The way it's written right now slightly implies that the loop could change `name`. But I don't feel strongly about it, so I'll just mark this as resolved.

app/experimenter/static/js/experiment-date-filter.js
uhlissuh: yep you're right.

uhlissuh:
@glasserc @jaredkerim I made Ethan's most recent changes, let me know what else I need to improve. 🌴

app/experimenter/experiments/tests/test_views.py
```diff
@@ -298,6 +298,164 @@ def test_list_view_lists_experiments_with_default_order_no_archived(self):
...
+            proposed_start_date=datetime.date(2019, 3, 15),
+            proposed_duration=25,
+            proposed_enrollment=1,
+        )
+
+        self.start_range_date = "2019-04-01"
+        self.end_range_date = "2019-05-01"
+        self.user_email = "user@example.com"
+
+    def test_list_shows_all_experiments_with_start_in_range(self):
```
jaredkerim: Tests look really good!

app/experimenter/experiments/tests/test_views.py
```diff
@@ -298,6 +298,164 @@ def test_list_view_lists_experiments_with_default_order_no_archived(self):
...
+
+        self.set_up_date_tests()
+
+        response = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"),
+                params=urlencode(
+                    {
+                        "experiment_date_field": "starts_between",
```
jaredkerim: After you move these values to constants remember to refer to constants here too

app/experimenter/experiments/views.py
```diff
@@ -72,6 +73,33 @@ def get_owner_display_value(self):
...
+
+        if experiment_date_field == "starts_between":
+            type = "starting"
+        elif experiment_date_field == "pauses_between":
+            type = "pausing"
+        elif experiment_date_field == "ends_between":
+            type = "ending"
+
+        if date_after and date_before:
+            return f"{type} between {date_after} and {date_before}"
```
jaredkerim: Very nice 👍

app/experimenter/experiments/views.py
```diff
@@ -151,6 +194,44 @@ def filter_search(self, queryset, name, value):
             .order_by("-rank")
         )
 
+    def experiment_date_field_filter(self, queryset, name, value):
+        # this custom method isn't doing anything. There has to
+        # be a custom method to be able to display the select
+        # filter that controls which date range we show
+        return queryset
```
jaredkerim: Ah that's annoying, oh well

app/experimenter/static/css/experimenter.css
```diff
@@ -182,3 +182,11 @@ pre.json {
   line-height: 2em;
   border: none !important;
 }
+
+.range-field {
```
jaredkerim: You don't need to add custom CSS rules, bootstrap has these handy margin classes so you can just say

    class="mb-1"

We use them all over the place, docs here:

https://getbootstrap.com/docs/4.0/utilities/spacing/

app/experimenter/experiments/views.py
```diff
@@ -72,6 +73,33 @@ def get_owner_display_value(self):
         if user_id is not None:
             return str(get_user_model().objects.get(id=user_id))
 
+    def get_display_start_date_info(self):
+        experiment_date_field = self.data.get("experiment_date_field")
+        date_after = self.data.get("date_range_after")
+        date_before = self.data.get("date_range_before")
+
+        if experiment_date_field == "starts_between":
+            type = "starting"
```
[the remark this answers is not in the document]
uhlissuh: gotcha!

app/experimenter/experiments/views.py
```diff
@@ -72,6 +73,33 @@ def get_owner_display_value(self):
         if user_id is not None:
             return str(get_user_model().objects.get(id=user_id))
 
+    def get_display_start_date_info(self):
```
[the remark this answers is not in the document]
uhlissuh: It didn't work to move this method to the filterset. I put it here in the form because that's where other methods called in the template are located... what am I missing?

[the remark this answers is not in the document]
jaredkerim: Okay, keep it here, see if you can add

- experiment_date_field
- date_range_before
- date_range_after

to Meta.fields in this form and see if that works or if things blow up

app/experimenter/experiments/views.py
```diff
@@ -72,6 +73,33 @@ def get_owner_display_value(self):
...
+        experiment_date_field = self.data.get("experiment_date_field")
+        date_after = self.data.get("date_range_after")
+        date_before = self.data.get("date_range_before")
+
+        if experiment_date_field == "starts_between":
```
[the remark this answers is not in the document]
uhlissuh: GOOD IDEA!

app/experimenter/experiments/views.py
```diff
@@ -118,6 +145,22 @@ class ExperimentFilterset(filters.FilterSet):
     archived = filters.BooleanFilter(
         label="Show archived experiments", widget=forms.CheckboxInput()
     )
+    experiment_date_field = filters.ChoiceFilter(
+        empty_label="No Date Restriction",
+        choices=[
+            ("starts_between", "Experiment Starts"),
```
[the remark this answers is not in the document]
uhlissuh: I added them to the `experimenter/experiments/constants.py` file, so that I could call them by saying `Experiment.EXPERIMENT_STARTS` and they are in scope in other places like the filter method. It's done like this in the other filters. Do you dislike this?

app/experimenter/experiments/views.py
```diff
@@ -151,6 +194,44 @@ def filter_search(self, queryset, name, value):
...
 
+    def experiment_date_field_filter(self, queryset, name, value):
+        # this custom method isn't doing anything. There has to
+        # be a custom method to be able to display the select
+        # filter that controls which date range we show
+        return queryset
+
+    def date_range_filter(self, queryset, name, value):
+
+        type = self.request.GET["experiment_date_field"]
```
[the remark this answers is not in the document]
uhlissuh: Rad about the form! Learn new Django stuff everyday... And yes, I moved everything away from the `type` variable name. 😄

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
jaredkerim: Yep that seems good!

app/experimenter/experiments/views.py
uhlissuh: I guess I can't be 110% sure that this is true.. what do you think?

jaredkerim: I think... If you take the definition of this field out of the filterset altogether, and just declare it as a choice field on the form, then it'll still show up, show the choices, you'll still be able to look up the value from the filterset method, and then you won't need this method at all?  Give that a shot.

app/experimenter/experiments/views.py
```diff
@@ -151,6 +194,44 @@ def filter_search(self, queryset, name, value):
...
+
+        experiments = queryset.all()
+        results = []
+        date = None
+
+        for experiment in experiments:
```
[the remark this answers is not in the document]
uhlissuh: @glasserc wanted me to do it this way too. 😄 The change is made!

app/experimenter/experiments/tests/test_views.py
uhlissuh: yay!

uhlissuh:
@jaredkerim So, I've updated the pull request with all the changes except for moving the method definition on the `ExperimentFiltersetForm` to `ExperimentFilterset` (the method was just never called when it was inside of Filterset) and attempting to do away with that empty method by moving the date selector to a field on `ExperimentFiltersetForm` (I get errors about anything in `Meta.fields` of the form that doesn't also have an entry in the `ExperimentFilterset`).  I wish I could say more than "it doesn't work" but I'm still just generally confused about why there is this FiltersetForm class. I don't see it done this way in the docs, so I don't have much of a reference point for having this other Form class in addition to the Filterset. I'm sure you had reasons for doing it!! Perhaps if we really want to do it differently we could jump on a call or move it over to a Slack discussion so that I can get into your head more about how the form and filterset comingle.

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
uhlissuh: I made a general PR comment that addressed my questions with this!

app/experimenter/experiments/views.py
uhlissuh: I made a general PR comment that addressed my questions with this!

PR #1187, opened 2019-04-15
Add free text search to filter list fixes #400

### LINKED ISSUE #400: Add text search for experiment details

uhlissuh:
Fixes #400 

This adds a field in the filter list that takes text and returns any experiments that have the text in their name or short description.

Screen Shot 2019-04-15 at 2 56 28 PM

### DIFF
--- a/app/experimenter/experiments/tests/test_views.py
+++ b/app/experimenter/experiments/tests/test_views.py
@@ -177,6 +177,84 @@ class TestExperimentFilterset(TestCase):
             set(Experiment.objects.filter(firefox_channel=include_channel)),
         )
 
+    def test_list_filters_by_search_text(self):
+        user_email = "user@example.com"
+
+        exp_1 = ExperimentFactory.create_with_status(
+            random.choice(Experiment.STATUS_CHOICES)[0],
+            name="Experiment One Cat",
+            short_description="",
+            slug="exp-1",
+            related_work="",
+            addon_experiment_id="1",
+            pref_key="",
+            public_name="",
+            public_description="",
+            objectives="",
+            analysis="",
+            analysis_owner="",
+            engineering_owner="",
+            bugzilla_id="4",
+            normandy_slug="",
+        )
+
+        exp_2 = ExperimentFactory.create_with_status(
+            random.choice(Experiment.STATUS_CHOICES)[0],
+            name="Experiment Two Cat",
+            short_description="",
+            slug="exp-2",
+            related_work="",
+            addon_experiment_id="2",
+            pref_key="",
+            public_name="",
+            public_description="",
+            objectives="",
+            analysis="",
+            analysis_owner="",
+            engineering_owner="",
+            bugzilla_id="5",
+            normandy_slug="",
+        )
+
+        exp_3 = ExperimentFactory.create_with_status(
+            random.choice(Experiment.STATUS_CHOICES)[0],
+            name="Experiment Three Dog",
+            short_description="",
+            slug="exp-3",
+            related_work="",
+            addon_experiment_id="3",
+            pref_key="",
+            public_name="",
+            public_description="",
+            objectives="",
+            analysis="",
+            analysis_owner="",
+            engineering_owner="",
+            bugzilla_id="6",
+            normandy_slug="",
+        )
+
+        first_response_context = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"), params=urlencode({"search": "Cat"})
+            ),
+            **{settings.OPENIDC_EMAIL_HEADER: user_email},
+        ).context[0]
+
+        second_response_context = self.client.get(
+            "{url}?{params}".format(
+                url=reverse("home"), params=urlencode({"search": "Dog"})
+            ),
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/views.py
```diff
@@ -104,6 +119,11 @@ class Meta:
         form = ExperimentFiltersetForm
         fields = ExperimentFiltersetForm.Meta.fields
 
+    def filter_search(self, queryset, name, value):
+        return queryset.annotate(
+            search=SearchVector("name", "short_description")
```
jaredkerim: I think the fields we wanna search should include:

- owner.email
- name
- slug
- short_description
- related_work
- addon_experiment_id
- pref_key
- public_name
- public_description
- objectives
- analysis
- analysis_owner
- engineering_owner
- bugzilla_id
- normandy_slug

I think the test you have is fine, you don't need to test against searching each of these fields, but check by hand to see if they work as expected.

uhlissuh: @jaredkerim, I had a bit of a discussion days ago with @mythmon about this, and he had brought up a few concerns about text searching so many fields that an experiment could be returned even though it really wasn't obvious why it was returned, with nothing in the experiment preview box obviously matching the search text. Does this worry you at all, or would you rather just have the wider search net thrown? One thing we might want to consider if we use the multiple fields is the concept of 'weighting' or 'ranking' the results (I've seen these mentioned in the `SearchVector` docs) so that the better matches float higher. Second thought: I was wondering if since we already have a filter dedicated to searching a field like owner email.. should still include that one? I definitely see why looking for owner email with the text search might be easier than using that owner email select filter.

jaredkerim: @uhlissuh Yeah I'm not worried about 'why' it matches being reflected, just that if someone throws an email in that field, we should return all the experiments that refer to that email, or any kind of slug/identifier, like normandy slug/addon id, etc.  

You don't want the opposite to happen, ie you search for an email/identifier, don't see the thing you're looking for in the search results, find it some other way and clearly right on the detail page is the term you searched for being clearly displayed, but wasn't included in the search fields.

If we bump into performance limitations we can think about that separately.

Also yes adding some kind of weighting would also be super cool!

uhlissuh: OK great, will implement this change to what we're filtering against and the other changes you requested! :)

app/experimenter/experiments/tests/test_views.py
```diff
@@ -177,6 +177,52 @@ def test_filters_by_firefox_channel(self):
...
+
+        first_response_ids = []
+        for experiment in list(first_response_context["experiments"]):
+            first_response_ids.append(experiment.id)
+
+        second_response_ids = []
+        for experiment in list(second_response_context["experiments"]):
+            second_response_ids.append(experiment.id)
+
+        first_response_ids.sort()
```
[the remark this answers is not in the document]
uhlissuh: 😄

app/experimenter/experiments/tests/test_views.py
```diff
@@ -177,6 +177,84 @@ def test_filters_by_firefox_channel(self):
...
+            public_name="",
+            public_description="",
+            objectives="",
+            analysis="",
+            analysis_owner="",
+            engineering_owner="",
+            bugzilla_id="6",
+            normandy_slug="",
+        )
+
```
uhlissuh: So, I had to override all the pertinent fields that Factory Boy creates because it was actually occasionally generating the words "Dog" and "Cat" in some of the fields.. which would then make the tests fail.  😆 If there was a uniqueness constraint issue I put in something... unique.

glasserc:
Looks great and it seems like Jared gave a pretty clear list of fields (which corresponds to the one in the code). I wonder if it might also be worth including fields like `risks`, `risk_technical_description`, `testing`, `test_builds`, `qa_status`. I guess probably there's some reason Jared didn't include them in his list, and either way it's easy to add them later, so I'd be in favor of merging this as-is.

app/experimenter/experiments/views.py
```diff
@@ -104,6 +123,33 @@ class Meta:
         form = ExperimentFiltersetForm
         fields = ExperimentFiltersetForm.Meta.fields
 
+    def filter_search(self, queryset, name, value):
+        vector = SearchVector(
+            "name",
+            "short_description",
+            "owner__email",
```
jaredkerim: 👍

app/experimenter/experiments/views.py
```diff
@@ -104,6 +123,33 @@ class Meta:
...
+            "bugzilla_id",
+            "normandy_slug",
+        )
+
+        query = SearchQuery(value)
+
+        return (
+            queryset.annotate(rank=SearchRank(vector, query), search=vector)
+            .filter(search=value)
+            .order_by("-rank")
```
jaredkerim: Fantastic 🎉

uhlissuh:
I wish I could take more credit for it! 😄

PR #1188, opened 2019-04-17
Don't change slug for existing instances fixes #1129

### LINKED ISSUE #1129: Slug should only be set on initial creation
Presently you can change the title of an experiment and save it and it may update the slug, which is a behaviour we want to avoid because if a user bookmarks or links the previous slug they won't be able to find the experiment.  We should only set the slug at experiment creation time.

jaredkerim:
We auto generate the slug for things like Experiment, ExperimentVariant from its name, but we were doing it on every save so it's possible to change the slug of an experiment after it's been created (which is bad because then it changes the URL you find it at so you could break somebody's bookmarks).  This change prevents changing the slug if it's already been set.

### DIFF
--- a/app/experimenter/projects/forms.py
+++ b/app/experimenter/projects/forms.py
@@ -23,8 +23,11 @@ class NameSlugFormMixin(object):
     def clean(self):
         cleaned_data = super().clean()
 
-        name = cleaned_data.get("name")
-        cleaned_data["slug"] = slugify(name)
+        if self.instance.slug:
+            del cleaned_data["slug"]
+        else:
+            name = cleaned_data.get("name")
+            cleaned_data["slug"] = slugify(name)
 
         return cleaned_data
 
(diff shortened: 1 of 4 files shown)


app/experimenter/projects/forms.py
```diff
@@ -23,8 +23,11 @@ def clean_name(self):
     def clean(self):
         cleaned_data = super().clean()
 
-        name = cleaned_data.get("name")
-        cleaned_data["slug"] = slugify(name)
+        if self.instance.slug:
+            del cleaned_data["slug"]
+        else:
+            name = cleaned_data.get("name")
+            cleaned_data["slug"] = slugify(name)
```
glasserc: This might be outside the scope of this PR, but why are we doing this in the form code and not in the model code?

jaredkerim: @glasserc Putting this automation on the model makes it difficult to edit it by hand from the admin or override for testing in a console or in unit tests.  

In general I try to avoid putting too much data automation in models because for every place it makes your life easier it'll blow up in some annoying way somewhere else, whereas keeping it in the form makes it clear that from the user facing workflow it'll take place, but if you go in under the hood you shouldn't expect any 'magic'.

Issue #461, opened 2018-07-04
Create a data migration to add risk defaults to existing experiments


jaredkerim:
Let's make a simple data migration to add all the required reviews to the existing experiments.

jaredkerim:
Fixed in #857

PR #1193, opened 2019-04-18
changes to risks associated to legal review

tiftran:


app/experimenter/experiments/models.py
```diff
@@ -503,7 +506,10 @@ def completed_testing(self):
 
     @property
     def _conditional_required_reviews_mapping(self):
-        return {"review_vp": self.risk_partner_related}
+        return {
+            "review_vp": self.risk_partner_related,
+            "review_legal": self.risk_partner_related,
```
jaredkerim: I merged the other review_vp pr so we should rebase this.

app/experimenter/experiments/migrations/0046_experiment_risk_data_category.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 2.1.7 on 2019-04-18 22:01
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
jaredkerim: Ah a new risk field, now we gotta start doing those annoying data migrations I was mentioning.  

https://docs.djangoproject.com/en/2.2/topics/migrations/#data-migrations

Basically after we add the field, we have to find all existing experiments that have already passed the 'Sign Off' state (Ready to Ship, Accepted by Normandy, Live, Complete) and set any risks that are `None` to `False`.  Check all the risk fields because there's a couple we've added that we didn't do a data migration for and then we can also close https://github.com/mozilla/experimenter/issues/461

tiftran: will do

app/experimenter/experiments/migrations/0047_auto_20190422_1819.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 2.1.7 on 2019-04-22 18:19
```
jaredkerim: Yep exactly, but you don't have to do it in a separate migration file, just move all this to 46 and then in the operations at the bottom you put them in order

    operations = [
        migrations.AddField(....
        migrations.RunPython(....
    ]

app/experimenter/experiments/migrations/0047_auto_20190422_1819.py
```diff
@@ -0,0 +1,36 @@
...
+    for experiment in Experiment.objects.all():
+        if experiment.status not in (
+            ExperimentConstants.STATUS_DRAFT,
+            ExperimentConstants.STATUS_REVIEW,
+        ):
+            for risk in all_risks():
+                setattr(
+                    experiment, risk, any([getattr(experiment, risk), False])
+                )
+                experiment.save()
```
jaredkerim: Ah this will do a roundtrip to the db for every field, which adds unnecessary i/o.  If you move this out of the for loop it should speed things up.

app/experimenter/experiments/migrations/0047_auto_20190422_1819.py
```diff
@@ -0,0 +1,36 @@
...
+
+def all_risks():
+    return (
+        "risk_internal_only",
+        "risk_partner_related",
+        "risk_brand",
+        "risk_fast_shipped",
+        "risk_confidential",
+        "risk_release_population",
+        "risk_technical",
```
jaredkerim: And don't forget the new one `risk_data_category`

app/experimenter/experiments/migrations/0047_auto_20190422_1819.py
tiftran: let me fix that....

jaredkerim: Actually I just realized you could just do this replacing the loop body with

    Experiment.objects.filter(**{risk: None}).update(**{risk: False})

And that way it hits multiple experiments in a single query and doesn't bother hitting ones that don't need to be updated.

I don't think there's a clean way to do it for all the risk fields in a single query...

tiftran: just want to make sure i understand correctly, this would mean that I would also be getting rid of the outer loop that loops over every experiment as well?

jaredkerim:
Oh @tiftran also gotta update https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/models.py#L485 to include the new risk

app/experimenter/experiments/migrations/0047_auto_20190422_1819.py
jaredkerim: Yep exactly, just loop over the risk fields and execute that query once per risk and it'll hit all the experiments.

tiftran: would this mean that all experiments existing at the time of the migration will have risk values converted to False from None, regardless of state/status?

jaredkerim: Oh yeah you'd want to preserve the status filters, so you could either fold it into the filter part of each query, or use the fact that django querysets are lazy and go

    filtered_experiments = Experiment.objects.filter(status__in=...)
    for risk in risks:
        filtered_experiments.filter(...).update(...)

tiftran: i see i see

jaredkerim: extra reading for fun:

https://docs.djangoproject.com/en/2.2/topics/db/queries/#querysets-are-lazy

PR #1197, opened 2019-04-22
Add filter for experiments in QA fixes #941

### LINKED ISSUE #941: Add State filter for work that is "In QA"
QA needs a way to see what work is coming to them or that should be in their backlog to make sure their work load planning is accurate.

We want to be able to filter for studies where the first box for QA is checked (something about "Ready for QA" or "QA Requested") but the second QA box is NOT checked "QA Sign-off".  

That should be expose in the Filter page - under a new Filter called "States".  Default should be "All States" and the state that filters for the first but not second check box is "In QA".

uhlissuh:
Fixes #941 

This PR adds a filter field for experiments that are "in QA"... or more specifically, if the `review_qa_requested` box has been checked...and the `review_qa` box has NOT been checked. 
Questions that remain are:
* Wording! Initially Shell suggested the select be called "all states" but that is kind of close to "all statuses". I used "all reviews" as a placeholder but I would love if someone weighed in on the best wording. Would be happy doing "states" if we want.
* Another wording question.. As I have it now, the title after filtering becomes something like "2 **in QA** experiments".. this isn't the best, grammatically speaking, but I think to make it fancier (like making it say "2 Experiments **in QA**") I'd have to put a method on the form, which is totally easy to do, but not sure if it's worth the extra code.
* If we'd like more tests than I currently have. I have a unit test but not an integration test. 

Screen Shot 2019-04-22 at 4 23 39 PM
Here's what the title currently looks like:
Screen Shot 2019-04-22 at 4 23 28 PM

### DIFF
--- a/app/experimenter/experiments/tests/test_views.py
+++ b/app/experimenter/experiments/tests/test_views.py
@@ -256,6 +256,23 @@ class TestExperimentFilterset(TestCase):
             set(second_response_context["experiments"]), set([exp_3])
         )
 
+    def test_filters_by_review_in_qa(self):
+        exp_1 = ExperimentFactory.create_with_variants(
+            review_qa_requested=True, review_qa=False
+        )
+        ExperimentFactory.create_with_variants(
+            review_qa_requested=False, review_qa=False
+        )
+        ExperimentFactory.create_with_variants(
+            review_qa_requested=True, review_qa=True
+        )
+
+        filter = ExperimentFilterset(
+            {"in_qa": "on"}, queryset=Experiment.objects.all()
+        )
+
+        self.assertEqual(set(filter.qs), set([exp_1]))
+
 
 class TestExperimentOrderingForm(TestCase):
 
--- a/app/experimenter/experiments/views.py
+++ b/app/experimenter/experiments/views.py
@@ -31,6 +31,7 @@ from experimenter.experiments.models import Experiment
 
 
 class ExperimentFiltersetForm(forms.ModelForm):
+    in_qa = forms.BooleanField(required=False)
 
     search = forms.CharField(required=False)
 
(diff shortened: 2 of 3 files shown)


app/experimenter/experiments/tests/test_views.py
```diff
@@ -256,6 +256,30 @@ def test_list_filters_by_search_text(self):
...
+
+        filter = ExperimentFilterset(
+            {"reviews": "in_qa"}, queryset=Experiment.objects.all()
+        )
+
+        self.assertEqual(
+            set(filter.qs),
+            set(
+                Experiment.objects.filter(
+                    review_qa_requested=True, review_qa=False
```
jaredkerim: I would just assign each of the test experiments to variables and then refer to them here, ie

    exp1 = ExperimentFactory.create(......
    ...
    self.assertEqual(set(filter.qs), set([expN, ....])

that way it's more explicit exactly which experiments are matching/not matching without needing to query the db again.

uhlissuh: gotcha!

uhlissuh:
Okay here's how it looks now with that change made!

Screen Shot 2019-04-24 at 10 29 47 AM

app/experimenter/experiments/tests/test_views.py
```diff
@@ -256,6 +256,23 @@ def test_list_filters_by_search_text(self):
...
 
+    def test_filters_by_review_in_qa(self):
+        exp_1 = ExperimentFactory.create_with_variants(
+            review_qa_requested=True, review_qa=False
+        )
+        ExperimentFactory.create_with_variants(
+            review_qa_requested=False, review_qa=False
+        )
+        ExperimentFactory.create_with_variants(
+            review_qa_requested=True, review_qa=None
```
jaredkerim: When does this case happen?

uhlissuh: Yea this was kind of nonsensical so I got rid of the `None`.

Issue #1178, opened 2019-04-12
Fixes to add-on section


shell1:
Add-on Section needs:

1) remove the "signed testing URL" - not needed & confusing.  They can provide that info in the QA section.

shell1:
shell do full review on true optional...

PR #1198, opened 2019-04-23
added impact teams conditional review fixes #864

### LINKED ISSUE #864: Checklist question 9: Does this experiment impact teams outside of your own?
"Yes" radial button selection for this question - should make "Impacted Team(s) Review" move to "required". Required means they need to check it before they can click "ready for shipping".

Help for this question should take them to the risks section. https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-Risk

The help next to the actual checklist item should go to the main Dependent section. Linking to the section instead of the specific action, avoids future name changes breaking the header link. https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-DependentSign-offs

Need to add in dummy migrations with any new input needed (fields, radials, etc.).  Add default “no” for Live or Complete experiments.  This avoids “incomplete section” errors -  which block that section from displaying in the read-only mode.

tiftran:


jaredkerim:
@tiftran too late I already merged #1178 now you gotta make this migration 48 >_< We gotta find a better solution to this migration situation.

PR #1199, opened 2019-04-23
Move public name/description to overview form fixes #1179

### LINKED ISSUE #1179: Move public_name/public_description to overview form
Right now the public name/description are displayed in the 'Overview' section on the detail page, but appear in the 'Population Form' which means clicking 'edit' doesn't take you to the form that lets you edit them.  We need to move those two fields to the Overview form.

jaredkerim:
Screenshot_2019-04-23 Enterprise-wide stable attitude
Screenshot_2019-04-23 Enterprise-wide stable attitude(1)


app/experimenter/templates/experiments/section_overview.html
```diff
@@ -5,14 +5,18 @@
...
+    <p>{{ experiment.engineering_owner }}</p>
   {% endif %}
-  {{ experiment.short_description|urlize|linebreaks }}
-  <p><strong>Public Name:</strong></p> 
-  {{ experiment.public_name| linebreaks }}
+  <p><strong>Public Name:</strong></p>
+  {{ experiment.public_name|linebreaks }}
   <p><strong>Public Description:</strong> </p>
-  {{ experiment.public_description| linebreaks }}
+  {{ experiment.public_description|linebreaks }}
```
tiftran: Didn't think of it when working on this earlier and not entirely sure about the use case for public description, but would we want "urlize" here too or would public descriptions ever contain a link?

jaredkerim: Yep good catch

jaredkerim: Public description is the text that will go in `about:studies`.

PR #1217, opened 2019-04-26
added new required fields fixes #1216

### LINKED ISSUE #1216: Bugzilla tickets have new required fields
Bugzilla recently added some new required fields, so we have to update our background tasks to include them:

 "A value must be set for the 'Is Regression' field."

tiftran:


tiftran:
@glasserc unfortunately the bugzilla documentation hasn't changed to reflect the new required fields, we only went by what we saw after calling a GET and Shell's comment in the issue and the fact that the current request was returning 400s

glasserc:
Spoke with @tiftran on Slack. The added fields came from looking at an existing Bugzilla ticket via the API. There doesn't appear to be any "Is Regression" / "Is Defect" field. This PR seems to respond to the ticket, so I think we can go ahead and merge this. If there's something left that Shell wanted addressed, she can reopen the issue or create a new one.

PR #1225, opened 2019-04-30
Add permissions for QA and Relman signoff fixes #1171

### LINKED ISSUE #1171: Verify or limit who can check "QA Sign-off" and "Relman Sign-off"
Today anyone can check any of the required check-boxes in the Sign-off section.  

There are 2 in particular that should block if checked by the wrong person:
1.  QA Sign-off
2.  Relman Sign-off

We want those to only be available for the right people or to error if someone outside of the right people check it off.  

@shell1 to get the list of people for each sign-off.  
We'd also want everyone on Experimenter and Normandy team to be able to check either - as an override for exceptions.  Until we have a good method of re-launching worked out - there may be exceptions needed.

uhlissuh:
Fixes #1171 

This PR adds user groups to Experimenter with special permissions to be able to _check or uncheck_ the QA signoff and Release Management signoff boxes. The groups will be created if they don't exist and permissions added upon app startup. The admin page will need to be used to add specific users to each group. The form now checks that the user has permissions to check QA Signoff and Relman Signoff before the DB is updated. 

If a user tries to check a box they don't have permissions for, a message appears on their screen that warns them they can't do it, and the changes are not made to the experiment in the DB. If a user is also checking/unchecking other boxes at the time of form submission in addition to the box they don't have permissions for, the form _will_ submit the permitted changes to the DB but will skip making any changes to the restricted fields. It will warn the user that the action isn't allowed plus inform them of the changes that were successfully made. 

Questions that remain _for me_ are:

- Styling of two NOT permitted messages happening simultaneously. Right now, I throw up two separate messages, figuring the case will be rare enough that it didn't warrant the extra effort of making them the same message... but I can most certainly fix this if someone thinks I should. 

- This change broke other tests that involved submitting general fake form data. In some cases it seemed helpful/pertinent to add the permissions to the user for an existing test so that qa/relman signoff data could be submitted in the fake form. In a few cases it seemed not germane to the test at hand, so I just changed the dummy data to other fields that didn't have permissions issues to keep the test simple. Let me know if I ran afoul here anywhere with test changes. 

Screen Shot 2019-04-29 at 8 02 37 PM

Screen Shot 2019-04-29 at 8 03 00 PM

Screen Shot 2019-04-29 at 8 03 56 PM


app/experimenter/experiments/apps.py
```diff
@@ -8,4 +8,29 @@ class ExperimentsConfig(AppConfig):
...
     def ready(self):
+
+        # create user groups and add permissions to them
+        from django.contrib.auth.models import Group, Permission
+        from django.contrib.contenttypes.models import ContentType
+        from experimenter.experiments.models import Experiment
+
         markus.configure(settings.MARKUS_BACKEND)
+
+        relman_signoff_group, created = Group.objects.get_or_create(
```
glasserc: This `created` gets ignored. Is that intentional? I think it's safe-ish to assume that both will be created or neither will be, but maybe we should make this explicit by marking one of these `created` with `_`.

app/experimenter/templates/base.html
```diff
@@ -99,6 +99,24 @@ <h3 class="pt-2">
       </div>
     {% endif %}
 
+    {% if messages %}
+    <div class="alert-danger">
+      <div class="container">
+        <div class="row">
+            <div class="col pt-3 pb-1">
```
glasserc: Weird spacing here.

app/experimenter/templates/base.html
```diff
@@ -99,6 +99,24 @@ <h3 class="pt-2">
       </div>
     {% endif %}
 
+    {% if messages %}
```
glasserc: Is this really the first use of `messages` in this project? I thought I had seen other messages elsewhere..

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1317,6 +1336,86 @@ def test_optional_reviews_when_a_risk_option_is_true(self):
...
+        permission = Permission.objects.get(
+            content_type=content_type, codename="can_check_relman_signoff"
+        )
+        user_1.user_permissions.add(permission)
+        self.assertTrue(
+            user_1.has_perm("experiments.can_check_relman_signoff")
+        )
+        self.assertFalse(
+            user_2.has_perm("experiments.can_check_relman_signoff")
+        )
```
glasserc: I'm not sure if these assertions are really worth it.

app/experimenter/experiments/apps.py
uhlissuh: @glasserc I actually changed from doing this inside `ready` to doing it in a data migration so this code is gone!

app/experimenter/experiments/tests/test_forms.py
uhlissuh: gotcha, I'll change.

app/experimenter/templates/base.html
uhlissuh: @glasserc Could you elaborate a bit more on what you are thinking here?

glasserc: It seemed odd to me that we never used `messages` here, since we seem to have other kinds of messages already in the project. Indeed, it seems like there's a `Notification` model. Do we need both? Can we stick with the one that already exists in the project?

app/experimenter/experiments/migrations/0050_create_group_add_permissions.py
```diff
@@ -0,0 +1,45 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("experiments", "0049_add_permissions"),
+        ("contenttypes", "0002_remove_content_type_name"),
+        ("experiments", "0048_experiment_risk_external_team_impact"),
+        ("auth", "0009_alter_user_last_name_max_length"),
+    ]
+
+    operations = [migrations.RunPython(add_groups, remove_groups)]
```
mythmon: The filename is "create_group_add_permissions", but it doesn't really explain what's going on here. Maybe something more descriptive like "create_relman_and_qa_groups"?

app/experimenter/experiments/migrations/0050_create_group_add_permissions.py
```diff
@@ -0,0 +1,45 @@
...
+    qa_signoff_group = Group.objects.get(name="QA Signoff")
+    qa_signoff_group.delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("experiments", "0049_add_permissions"),
+        ("contenttypes", "0002_remove_content_type_name"),
+        ("experiments", "0048_experiment_risk_external_team_impact"),
```
mythmon: You probably don't need to depend on both 48 and 49. Just one will do.

app/experimenter/experiments/migrations/0050_create_group_add_permissions.py
```diff
@@ -0,0 +1,45 @@
...
+    relman_signoff_group.delete()
+
+    qa_signoff_group = Group.objects.get(name="QA Signoff")
+    qa_signoff_group.delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("experiments", "0049_add_permissions"),
```
mythmon: Is there a reason that 49 and 50 can't be merged together? It seems like a better pattern to have these two migrations in the same file, if possible.

app/experimenter/templates/base.html
uhlissuh: I didn't use it because I didn't think this message warranted writing to the database since it's just a warning and a lack of actual action. Seems like if you use the `Notification` class you are writing the change made to the database (like X person added the datascience review) so that changes can be kind of kept track of, displayed in the review history section, and also then send that little blue message. Kinda like a changelog for changes made in the review process. I didn't think "user tried to check something they weren't allowed to check and we didn't save anything" rose to that same level of importance.. Maybe my understanding of `Notification` is off.

app/experimenter/experiments/apps.py
```diff
@@ -8,4 +8,5 @@ class ExperimentsConfig(AppConfig):
     name = "experimenter.experiments"
 
     def ready(self):
+
```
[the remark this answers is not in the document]
uhlissuh: oops!

app/experimenter/templates/base.html
uhlissuh: To add on, I actually was intentionally not using it because I didn't want that "tried to check permission but couldnt" going into that review history section, so if I did use Notification I'd have to have some extra logic to fish it out of that section maybe...

mythmon: The module being used here, `django.contrib.messages`, is about one-off messages that happen within a request/response cycle, usually feedback about that specific response. The Notification model seems to be longer term notifications that carry a read/unread state, and is set up to tie in to sending emails (which messages wouldn't be good for).

I think that having both systems is good.

That being said, I think the styling of these messages here isn't very good. Since this is the first time we're integrating messages, I think we can make it a bit less confusing. Right now the messages blend into the navigation header too much, and it doesn't distinguish between the levels of the message (like warning, error, or info).

I'd suggest something outside of the usual flow of the page, like a toast:

image

uhlissuh: So @mythmon, are you saying in "I think having both systems is good" that you mean you agree with the decision to use plain `messages` instead of `Notification` for this use case? 

Also, are you saying you want me to change the styling to this example styling in the PR?

mythmon: Yes, I think using `messages` in this case is the right choice.

Since the styling for messages was introduced in this PR, it's in scope to make it look not-confusing. I don't have strong opinions about how it should look, just that the current UI is confusing because it blends in with site navigation. The screenshot I posted is an idea, but it's a rough idea.

I think the best choice here would be to use Bootstrap 4.2's Toast component, but Experimenter uses Bootstrap 4.1, and I don't know how big of an upgrade that would be.

uhlissuh: OK, I don't mind changing the styling. Perhaps I'll run past @jaredkerim when he gets back and @shell1 to see what hopes and dreams they would have for the styling of these warning messages?

app/experimenter/experiments/migrations/0050_create_group_add_permissions.py
uhlissuh: I think they can me merged now that I'm doing the python methods for each thing.

app/experimenter/templates/base.html
glasserc: OK, thanks for explaining. I agree that having both is the right decision.

app/experimenter/experiments/migrations/0050_create_group_add_permissions.py
uhlissuh: I was able to combine them and I made that change.

PR #1228, opened 2019-04-30
Change label to Jira request sent with a link fixes #1219

### LINKED ISSUE #1219: change the wording of "QA Requested" checklist item and add link
Change the wording of "QA Requested" to "QA Jira Request Sent" and make the "Jira" be a hyperlink to https://moz-pi-test.atlassian.net/servicedesk/customer/portal/9

uhlissuh:
Fixes #1219

Screen Shot 2019-04-30 at 3 30 57 PM

Do we want the blue link text though? Feeling unsure.

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -8,6 +8,7 @@ from django.forms import BaseInlineFormSet
 from django.forms import inlineformset_factory
 from django.forms.models import ModelChoiceIterator
 from django.utils import timezone
+from django.utils.safestring import mark_safe
 
 from experimenter.base.models import Country, Locale
 from experimenter.experiments.constants import ExperimentConstants
@@ -773,7 +774,10 @@ class ExperimentReviewForm(
     )
     review_qa_requested = forms.BooleanField(
         required=False,
-        label="QA Requested",
+        label=mark_safe(
+            f"QA <a href={settings.JIRA_URL} target='_blank'>"
+            "Jira</a> Request Sent"
+        ),
         help_text=Experiment.REVIEW_QA_REQUESTED_HELP_TEXT,
     )
     review_intent_to_ship = forms.BooleanField(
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/forms.py
```diff
@@ -773,7 +774,11 @@ class ExperimentReviewForm(
     )
     review_qa_requested = forms.BooleanField(
         required=False,
-        label="QA Requested",
+        label=mark_safe(
+            'QA <a href="'
+            'https://moz-pi-test.atlassian.net/servicedesk/customer/portal/9" '
```
mythmon: I think this should be a [setting](https://github.com/mozilla/experimenter/blob/22b94d6534ebcca1927cd5b56ad1f98f3031c344/app/experimenter/settings.py#L310-L315) instead. That seems to be the style used in experimenter for URLs, and I imagine it's pretty likely for this to change over time or between environments.

uhlissuh: I made a change in this vein... let me know what you think.

app/experimenter/experiments/forms.py
```diff
@@ -773,7 +774,10 @@ class ExperimentReviewForm(
     )
     review_qa_requested = forms.BooleanField(
         required=False,
-        label="QA Requested",
+        label=mark_safe(
+            f"QA <a href={settings.JIRA_URL} target='_blank'>"
+            "Jira</a> Request Sent"
```
mythmon: I bet you could recombine these back into one line now.

uhlissuh: it was still slightttlllyyyy too long I think.... I'll double check.

uhlissuh: I'm seeing 
`./experimenter/experiments/forms.py:778:80: E501 line too long (83 > 79 characters)` when trying to change it, so I think I'm slightly over the limit without two lines.

mythmon: Ah, my line-length limits are tuned to Normandy, where we use 99 chars instead of 79. Carry on.

PR #1236, opened 2019-05-02
Fix migration order issue regarding migration numbers 49-53

uhlissuh:
After a PR #1225 merged with master, there were database migrations that were out of order. This reorders the migration dependencies and file names.


glasserc:
Thanks for jumping on this!

PR #1241, opened 2019-05-03
Use end date if the experiment has completed else calculate fixes #1166

### LINKED ISSUE #1166: End date should reflect Completed date from Normandy
After an experiment is complete, the change log records the actual dates the experiment started and ended, and so we should update the timeline section with separate 'proposed timeline' 'actual timeline' sections so it's clear to see when and how long the experiment actually ran for.

uhlissuh:
Fixes #1166 

in this PR we're updating the `end_date` property on the experiment model to either:

- calculate the end date based on the proposed duration if the experiment hasn't ended
- use the status change from "live" to "complete" to assign the actual date the experiment ended.

It follows the pattern on the `start_date` property.

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -356,7 +356,9 @@ class Experiment(ExperimentConstants, models.Model):
 
     @property
     def end_date(self):
-        return self._compute_end_date(self.proposed_duration)
+        return self._transition_date(
+            self.STATUS_LIVE, self.STATUS_COMPLETE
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/models.py
```diff
@@ -356,7 +356,9 @@ def _compute_end_date(self, duration):
 
     @property
     def end_date(self):
-        return self._compute_end_date(self.proposed_duration)
+        return self._transition_date(
```
jaredkerim: tenor

PR #1242, opened 2019-05-03
added more fields to bugzilla ticket creation fixes #802

### LINKED ISSUE #802: Bugzilla Experimenter bug creation when “Ready for Sign-offs” button clicked
This bug is a blank holder for QA comments and release management.  Bugzilla allows studies to fit into existing queries and release processes.  Bug not truly used (to deploy Normandy) until after “ready to ship” is clicked.

_Jared - Let me know if parts of these bug editing requests are technically complex.  We can include those in manual instructions for what bug editing is need._
 
**BUG TITLE** = [STUDY]:{type} {Name} 
**SEE ALSO** = {Data Science Bug} _Issue - need only the numbers in bugzilla (not URLs)_
**BLOCKS** = If {Feature Meta Bug} exists, put in BLOCKS _Issue - need only the numbers (not URLs)_
**Set TrackingRelease{release}**=? set this initial assessment- flags for QA and Relman what is targeting a release.  This will be written again when “ready to Ship” is clicked.  Shell will document in the checklist for bugzilla, that they need to manually edit the TrackingRelease field to remove “?” if the study release/channel changed.  Currently we can only set this for the one version we get back from experimenter.  But we should be getting min version and max version soon ... and be able to add all those. That will be a new ticket.
**URL** = {experimenter link} 
**Comment** = “Experimenter is the source of truth for details and delivery.  {experimenter specific study link}.   Changes to Bugzilla are not reflected in Experimenter and will not change delivery configuration.”  
**Whiteboard** = “Ready for Sign-offs”

tiftran:


app/experimenter/experiments/constants.py
```diff
@@ -818,8 +818,10 @@ class ExperimentConstants(object):
 
 {experiment.short_description}
 
+Experimenter is the source of truth for details and delivery. Changes to Bugzilla are not reflected in Experimenter and will not change delivery configuration.
+
 More information: {experiment.experiment_url}
-        """
+        """  # noqa
```
tiftran: assuming the ``` # noqa``` allows flake8 and black to skip over it.
Thought it would be better to have the source of truth sentence as one long line since it looked funny being cut at ~79 characters in bugzilla?

app/experimenter/experiments/bugzilla.py
```diff
@@ -84,6 +92,15 @@ def create_experiment_bug(experiment):
     return response_data["id"]
 
 
+def get_firefox_major_version(version):
+    return version.split(".")[0]
+
+
+def get_bugzilla_id(experiment_url):
+    if experiment_url:
+        return int(experiment_url.split("id=")[1])
```
jaredkerim: Ah this is a little fragile, you can use `urllib.parse.urlparse` to pull out the query string and then `urllib.parse.parse_qs` to get the actual id field out:

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs

app/experimenter/experiments/bugzilla.py
```diff
@@ -84,6 +92,15 @@ def create_experiment_bug(experiment):
     return response_data["id"]
 
 
+def get_firefox_major_version(version):
+    return version.split(".")[0]
+
+
+def get_bugzilla_id(experiment_url):
```
jaredkerim: I think this should be `bug_url`?

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -44,13 +50,23 @@ def test_creating_pref_bugzilla_ticket_returns_ticket_id(self):
                 "cc": settings.BUGZILLA_CC_LIST,
                 "type": "task",
                 "priority": "P3",
+                "see_also": [
+                    get_bugzilla_id(experiment.data_science_bugzilla_url)
```
jaredkerim: Try to hardcode as much as possible in your expected output checks, so here hard code the actual ids.

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -44,13 +50,23 @@ def test_creating_pref_bugzilla_ticket_returns_ticket_id(self):
...
+                ],
+                "blocks": [get_bugzilla_id(experiment.feature_bugzilla_url)],
+                "url": experiment.experiment_url,
+                "whiteboard": experiment.STATUS_REVIEW_LABEL,
+                cf_tracking: "?",
```
jaredkerim: Why is cf_tracking "?" here?

tiftran: I took the issue description literally ```Set TrackingRelease{release}=?``` seeing that cf_status_firefoxXX has permissible values of ---, ?, +,  and -

jaredkerim: Perfect!

app/experimenter/experiments/bugzilla.py
```diff
@@ -84,6 +93,16 @@ def create_experiment_bug(experiment):
...
 
+def get_firefox_major_version(version):
+    return version.split(".")[0]
+
+
+def get_bugzilla_id(bug_url):
+    if bug_url:
+        query = urlparse(bug_url).query
+        return int(parse_qs(query)["id"][0])
+
```
tiftran: Not exactly sure if this is what you meant

PR #1252, opened 2019-05-07
Make grafana url reflect experiment start and end dates fixes #1159

### LINKED ISSUE #1159: Filter grafana links with start/end dates
When we link out to grafana dashboards, we can also pass parameters to specify the start and end timestamps so that the link automatically shows the entire duration of the experiment (instead of now where it defaults to the last 6 hours and you have to click at the top right to change it to start and whenever the experiment started.)

The relevant url parameters are:

    &from=1554046342962
    &to=1554491997919

which look like javascript timestamps in ms, so if we convert the [experiment start and end dates](https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/models.py#L307-L320) to that format we should be easily able to prepopulate them.

If the experiment is live we should only pass in 'from' so that it just shows from the start to now, and if it's complete then we can pass both.

uhlissuh:
Fixes #1159 

Previously, when a user clicked on the 'Live Monitoring Dashboard', the dashboard defaulted to the last 6 months of time irrespective of an experiment's timeline. 

This PR changes the Grafana monitoring URL to include parameters for start and end dates (end dates only if the experiment has completed) so that the dashboard comes preset with a date range that focuses on an experiment's actual timeline. 

Screen Shot 2019-05-06 at 6 07 20 PM

If an experiment is 'Live' and doesn't have an end date, we just use the start date in the parameters. 

Screen Shot 2019-05-06 at 6 04 21 PM

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -1,5 +1,6 @@
 import json
 import datetime
+import time
 from collections import defaultdict
 from urllib.parse import urljoin
 
@@ -267,8 +268,26 @@ class Experiment(ExperimentConstants, models.Model):
 
     @property
     def monitoring_dashboard_url(self):
+
+        def to_timestamp(date):
+            return int(time.mktime(date.timetuple())) * 1000
+
+        start_date = ""
+        end_date = ""
+
         if self.is_begun and self.normandy_slug:
-            return settings.MONITORING_URL.format(slug=self.normandy_slug)
+            start_date = to_timestamp(
+                self.start_date - datetime.timedelta(days=1)
+            )
+
+            if self.status == self.STATUS_COMPLETE:
+                end_date = to_timestamp(
+                    self.end_date + datetime.timedelta(days=2)
+                )
+
+            return settings.MONITORING_URL.format(
+                slug=self.normandy_slug, from_date=start_date, to_date=end_date
+            )
 
     def generate_normandy_slug(self):
         if self.is_addon_experiment:
--- a/app/experimenter/experiments/tests/test_models.py
+++ b/app/experimenter/experiments/tests/test_models.py
@@ -120,9 +120,59 @@ class TestExperimentModel(TestCase):
             status=Experiment.STATUS_LIVE,
             normandy_slug="normandy-slug",
         )
+
+        ExperimentChangeLogFactory.create(
+            experiment=experiment,
+            old_status=Experiment.STATUS_ACCEPTED,
+            new_status=Experiment.STATUS_LIVE,
+            changed_on=datetime.date(2019, 5, 1),
+        )
+
+        changed_on_in_milliseconds = 1556582400000
+
         self.assertEqual(
             experiment.monitoring_dashboard_url,
-            settings.MONITORING_URL.format(slug=experiment.normandy_slug),
+            settings.MONITORING_URL.format(
+                slug=experiment.normandy_slug,
+                from_date=changed_on_in_milliseconds,
+                to_date="",
+            ),
+        )
+
+    def test_monitoring_dashboard_url_returns_url_when_experiment_is_complete(
+        self
+    ):
+        experiment = ExperimentFactory.create(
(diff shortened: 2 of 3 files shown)


app/experimenter/experiments/models.py
```diff
@@ -265,10 +266,35 @@ def bugzilla_url(self):
...
+
     @property
     def monitoring_dashboard_url(self):
-        if self.is_begun and self.normandy_slug:
-            return settings.MONITORING_URL.format(slug=self.normandy_slug)
+        start_date = ""
+        end_date = ""
+
+        if (
+            self.status == "Live" or self.status == "Complete"
```
jaredkerim: Isn't this the same as `is_begun`?  Also you definitely want to avoid hardcoding strings like that, in general refer to constants like `self.STATUS_LIVE` `self.STATUS_COMPLETE`

app/experimenter/experiments/models.py
```diff
@@ -265,10 +266,35 @@ def bugzilla_url(self):
...
+        ) and self.normandy_slug:
+            start_date = self.convert_date_to_milliseconds(self.start_date)
+
+            if self.status == "Complete":
```
jaredkerim: Same here for the hard coded status value

app/experimenter/experiments/models.py
```diff
@@ -265,10 +266,35 @@ def bugzilla_url(self):
...
+                end_date = self.convert_date_to_milliseconds(self.end_date)
+
+            return settings.MONITORING_URL.format(
+                slug=self.normandy_slug, from_date=start_date, to_date=end_date
+            )
+        else:
```
jaredkerim: This is redundant, functions will default return None if you do nothing.

jaredkerim:
If you want you could just pad it with a day on either side and skip all the timezone logic and do something like

        In [10]: time.mktime((experiment.start_date - timedelta(days=1)).timetuple())
        Out[10]: 1550275200.0

app/experimenter/experiments/models.py
uhlissuh: I changed it to `is_started` per our slack convo..

uhlissuh:
@jaredkerim I went with that idea to pad the dates on both sides of start and end. 

For the **end date**, for whatever reason maybe again pertaining to timezones, padding the end date with a timedelta of one day wasn't actually enough extra time to roll over fully to the next day on Grafana's end. I thought that for visual consistency on the Grafana end we should do a timedelta of 2 days so that it appears as actually rolled over to the next day when people on the pacific coast look. 

Let me know if this 2 day time delta is not appealing and you want me to just let Grafana do whatever it's gonna do with a timedelta of one day for both start and end.

app/experimenter/experiments/models.py
```diff
@@ -267,8 +268,39 @@ def bugzilla_url(self):
 
     @property
     def monitoring_dashboard_url(self):
-        if self.is_begun and self.normandy_slug:
-            return settings.MONITORING_URL.format(slug=self.normandy_slug)
+        start_date = ""
+        end_date = ""
+
+        if (
+            self.status == self.STATUS_LIVE
```
jaredkerim: Ah sorry to be a pain about this but I'd really rather you use the aggregate status aliases, `is_begun` in this case.  That way if we add/change statuses in the future and we can update the aliases without needing to update the call sites.

app/experimenter/experiments/models.py
```diff
@@ -267,8 +268,39 @@ def bugzilla_url(self):
...
+            or self.status == self.STATUS_COMPLETE
+        ) and self.normandy_slug:
+            start_date = (
+                int(
```
jaredkerim: Ack!  `black` had a field day with this XD

Try splitting this up a bit so it doesn't get spread out so much. 

You can define a local function inside this function for the conversion:

    def to_timestamp(date):
        return int(time.mktime(date.timetuple())) * 1000

and then just call that here and that should make things a little cleaner.

PR #1257, opened 2019-05-07
Add the ability to subscribe to experiments

### LINKED ISSUE #808: Ability to subscribe/unsubscribe from an experiment
Relman, Data Science, QA, and Product all wanted the ability to  “subscribe” to certain experiments for updates.  I’m thinking this could be where we get all the email lists from for reminders later?  We cc: all subscribed people on all emails relevant to a study?  

This will give folks a way to easily mark what studies they are involved with, filter for them, and later get status updates.

Put a “Subscribe to this experiment” link or "subscribe" button or checkbox -  above the History section.  I'm not fussy on the method.  It needs an undo capability: "unsubscribe from this study", or a check box they check uncheck, or the button changes to "unsubscribe".

If there is a place to put a help link - here is where it should go.  If there isn't, don't worry, we will just train people. https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-Subscribe

jaredkerim:
Fixes #808

We will follow this up with https://github.com/mozilla/experimenter/issues/1259 to filter for subscribed experiments

UI:

Screenshot_2019-05-08 Streamlined impactful open system
Screenshot_2019-05-08 Streamlined impactful open system(1)
Screenshot_2019-05-08 Streamlined impactful open system(3)


app/experimenter/templates/base.html
```diff
@@ -62,6 +54,18 @@ <h3 class="pt-2">
...
           </div>
+          <div class="col text-right">
+            <div>
+              <span class="fas fa-user"></span>
+              {{ request.user }}
+            </div>
+
+            <div>
+              <a class="nocolorstyle" href="{% url "home" %}?owner={{ request.user.id }}&amp;archived=on">
+                {{ request.user.owned_experiments.count }} Owned Experiment{{ request.user.experiment_set.count|pluralize:"s" }}
```
robhudson: You're counting two different things here. One is `owned_experiments.count` and one is `experiment_set.count`. I believe that's why the zero isn't getting pluralized as it should in the screenshot.

app/experimenter/templates/base.html
```diff
@@ -62,6 +54,18 @@ <h3 class="pt-2">
...
+              </a>
+            </div>
+          </div>
```
robhudson: Were you going to add a similar display for subscribed experiments here?

app/experimenter/templates/base.html
robhudson: Also, `|pluralize` and `|pluralize:"s"` are equivalent since the `"s"` is the default. It could be dropped if you prefer.

app/experimenter/templates/base.html
jaredkerim: Yeah but taht requires adding a bunch of extra logic so I moved it to https://github.com/mozilla/experimenter/issues/1259

PR #1260, opened 2019-05-08
checks for fields for request body of bugzilla creation fixes #1258

### LINKED ISSUE #1258: Only set cf_tracking if firefox_version is set
An experiment can be created and moved to the sign-off state before a Firefox version is set, in which case we get:

{'documentation': 'http://www.bugzilla.org/docs/4.2/en/html/api/', 'error': True, 'code': 53, 'message': 'cf_tracking_firefox is not a valid parameter for the Bugzilla::Bug::create function.'}

so we should only set that field if the version is set at ticket creation time.

tiftran:


app/experimenter/experiments/bugzilla.py
```diff
@@ -89,6 +98,24 @@ def create_experiment_bug(experiment):
...
         )
+
+    # Firefox Version given might not be an available
+    # bugzilla tracking parameter, so remove and retry
+    err_code = response_data.get("code", None) == INVALID_PARAMETER_ERROR_CODE
+    if err_code:
+        tracking_msg = "cf_tracking_firefox" in response_data.get(
+            "message", None
+        )
+    if err_code and tracking_msg:
```
jaredkerim: You don't need to check err_code twice in a row, just embed this block inside the preceding if statement.

app/experimenter/experiments/bugzilla.py
```diff
@@ -89,6 +98,24 @@ def create_experiment_bug(experiment):
...
+        bug_data = bug_data.copy()
+        cf_tracking = "cf_tracking_firefox{}".format(
```
jaredkerim: Probably just move this into a property method right on experiment:

    experiment.bugzilla_tracking_key

so that you can call it from the format_bug_data function and here without repeating it.

app/experimenter/experiments/bugzilla.py
```diff
@@ -73,12 +70,24 @@ def create_experiment_bug(experiment):
         "type": "task",
         "priority": "P3",
         "see_also": [get_bugzilla_id(experiment.data_science_bugzilla_url)],
-        "blocks": [get_bugzilla_id(experiment.feature_bugzilla_url)],
         "url": experiment.experiment_url,
         "whiteboard": experiment.STATUS_REVIEW_LABEL,
-        cf_tracking: "?",
     }
+    if experiment.firefox_version:
```
jaredkerim: 👍

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -103,6 +103,41 @@ def test_create_bugzilla_ticket_retries_with_no_assignee(self):
             settings.BUGZILLA_CREATE_URL, expected_call_data
         )
 
+    def test_create_bugzilla_ticket_retries_with_no_cf_tracking(self):
+        experiment = ExperimentFactory.create_with_status(
+            Experiment.STATUS_DRAFT, name="An Experiment"
+        )
+        self.setUpMockBugzillaInvalidFirefoxVersion()
+
+        response_data = create_experiment_bug(experiment)
```
jaredkerim: I would call this variable `bugzilla_id` so it's clear it's the ID and not the entire response body.

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -103,6 +103,41 @@ def test_create_bugzilla_ticket_retries_with_no_assignee(self):
             settings.BUGZILLA_CREATE_URL, expected_call_data
         )
 
+    def test_create_bugzilla_ticket_retries_with_no_cf_tracking(self):
```
jaredkerim: Very nice 👍

Issue #982, opened 2019-02-28
No protection for "bad" responses from Bugzilla


peterbe:
See these lines: https://github.com/mozilla/experimenter/blob/57f776d6df2d2e23797ca0018e15f92354f0189f/app/experimenter/experiments/bugzilla.py#L48-L49

If the response from bugzilla is 400 or 403 (misconfigured API key) that will go unnoticed. I.e. no exception and no Sentry log. 

We should at least use `response.raise_for_status()`.

peterbe:
By the way, we shouldn't do

```python
response = requests.post(url, data)
return json.loads(response.content)
```

Instead, we should do:

```python
response = requests.post(url, data)
return response.json()
```

[The one in `requests`](https://github.com/kennethreitz/requests/blob/a345b776fb2cf67d1a1810349e71fd60a83f53eb/requests/models.py#L873-L897) respects the charset and does validation on the response JSON byte string.

jaredkerim:
This is fixed now.

PR #1263, opened 2019-05-09
moved "ready to ship" comment to userstory fixes #803

### LINKED ISSUE #803: Bugzilla updates when “Ready to Ship” button clicked
**BUG TITLE** =  append to current title Fx{version} {Channel} Or rewrite [STUDY]:{type} {Name} Fx{version} {Channel} 
In the **USER STORY** field (instead of a comment - User Story field is editable and also has notification updates):
**More information / Authoritative source:** = {their experimenter URL} 
**Description** = {Description}
**Pref** = {pref} or **Add-on** = {Add-on}
**Population** = {Population} of Firefox {version} {channel}
**Branches** = {Branches}
**Locales** = {locales}
**Countries** = {countries}
Set **TrackingRelease{release}**=?
**Whiteboard** = overwrite existing with  “Ready to Ship”

NOTE: TrackingRelease{release}=? just added a new "comment" required.  If a comment is required - it should just get the "Start Date:{start date}, End Date:{end date} for {experiment URL}

tiftran:
Ref of what the user story section might look like: https://bugzilla-dev.allizom.org/show_bug.cgi?id=1395334


app/experimenter/experiments/bugzilla.py
```diff
@@ -38,13 +53,52 @@ def format_bug_body(experiment):
...
+    summary = "[Experiment] {experiment_name} Fx {version} {channel}".format(
+        experiment_name=experiment,
+        version=experiment.firefox_version,
+        channel=experiment.firefox_channel,
+    )
+    return {
+        "summary": summary,
+        "cf_user_story": format_bug_body(experiment),
+        "whiteboard": experiment.STATUS_SHIP_LABEL,
+        cf_tracking: "?",
```
tiftran: bugzilla's api ignores invalid fields when doing a PUT, so if cf_tracking version is missing or the version is not a valid option, the rest call will still be made successfully without it. Not sure if this is okay behaviour or should we check and notify user that cf_tracking will not be updated due it either being missing or not an option in bugzilla?

app/experimenter/experiments/bugzilla.py
```diff
@@ -38,13 +53,52 @@ def format_bug_body(experiment):
...
+    body = format_update_body(experiment)
+    make_bugzilla_put_call(
+        settings.BUGZILLA_UPDATE_URL.format(id=experiment.bugzilla_id), body
+    )
+
+
+def make_bugzilla_put_call(url, data):
+    try:
+        response = requests.put(url, data)
+        return json.loads(response.content)
```
tiftran: followed what was done with existing rest calls but just curious the diff between ``` response.json()``` vs ``` json.loads(response.content)``` <- might be a silly question 🙈

app/experimenter/experiments/bugzilla.py
jaredkerim: Ideally we wanna know if it fails and raise some kind of error to go to sentry and notify the user, but if we don't get any signals back about what failed then there's not much we can do.

app/experimenter/experiments/bugzilla.py
jaredkerim: @tiftran Good catch! @peterbe also pointed this out a little while ago and filed this:

https://github.com/mozilla/experimenter/issues/982

so if you want to switch that here we can also close 982, or we could leave it and do it separately, up to you.

app/experimenter/experiments/constants.py
```diff
@@ -897,7 +902,7 @@ class ExperimentConstants(object):
 
     BUGZILLA_ADDON_TEMPLATE = (
         """
-    Experiment Type: Opt-Out experiment
+    Experiment Type: AddOn experiment
```
jaredkerim: I think it's "Add-on":

Screenshot 2019-05-21 17 20 23

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -111,40 +112,62 @@ def test_create_bugzilla_ticket_creation_failure(self):
...
         )
+        cf_tracking = "cf_tracking_firefox{}".format(
+            get_firefox_major_version(experiment.firefox_version)
+        )
 
-        comment_id = add_experiment_comment(experiment)
-
-        self.assertEqual(comment_id, self.bugzilla_id)
+        update_experiment_bug(experiment)
+        summary = "[Experiment] {exp_name} Fx {version} {channel}".format(
```
jaredkerim: Declare these values right on the factory and then hard code this whole string.

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -111,40 +112,62 @@ def test_create_bugzilla_ticket_creation_failure(self):
...
-    def test_add_bugzilla_comment_pref_experiment(self):
+    def test_update_bugzilla_pref_experiment(self):
         experiment = ExperimentFactory.create_with_status(
             Experiment.STATUS_DRAFT,
             name="An Experiment",
             bugzilla_id="123",
             type=Experiment.TYPE_PREF,
         )
+        cf_tracking = "cf_tracking_firefox{}".format(
+            get_firefox_major_version(experiment.firefox_version)
```
jaredkerim: Yeah same here, if you hard code the version on the factory you can hard code this whole string.

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -159,18 +182,38 @@ def test_api_error_logs_message(self):
...
         self.assertEqual(response_data, mock_response_data)
 
     def test_request_error_passes_silently(self):
         self.mock_bugzilla_requests_post.side_effect = RequestException()
 
         with self.assertRaises(BugzillaError):
-            make_bugzilla_call("/url/", {})
+            make_bugzilla_post_call("/url/", {})
 
     def test_json_parse_error_passes_silently(self):
```
jaredkerim: This test isn't named properly (my fault).  I think it should say something like

    test_json_parse_error_raises_bugzilla_error

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -291,85 +291,76 @@ def test_experiment_bug_successfully_created(self):
             ),
         )
 
-    def test_bugzilla_error_doesnt_create_notification(self):
+    def test_bugzilla_error_doesnt_create_notifications(self):
```
jaredkerim: Ah yeah, I think we want the opposite behaviour of this now:

https://github.com/mozilla/experimenter/issues/1282#issuecomment-494567539

jaredkerim: I would say leave this as is in here (ie consistent with the rest of the code) and do those changes in separately in #1282.  Don't wanna cram too much into one PR.

app/experimenter/experiments/bugzilla.py
```diff
@@ -38,13 +53,52 @@ def format_bug_body(experiment):
...
+
+
+def update_experiment_bug(experiment):
+    body = format_update_body(experiment)
+    make_bugzilla_put_call(
+        settings.BUGZILLA_UPDATE_URL.format(id=experiment.bugzilla_id), body
+    )
+
+
+def make_bugzilla_put_call(url, data):
```
jaredkerim: Also if the only difference here is `requests.put` instead of `requests.post` then just collapse it back into a single function and make that an argument like

    def make_bugzilla_call(url, data, method=requests.post):
        try:
            response = method(url, data)
        ....

    make_bugzilla_call(url, data, method=requests.put)

jaredkerim:
@tiftran @shell1 We're gonna pause this PR pending further design discussion.

app/experimenter/experiments/tests/test_tasks.py
tiftran: will leave this be for now, there's probably too much in this PR already

app/experimenter/experiments/bugzilla.py
tiftran: kinda did this^ but left the default out since for tests it needed to have the request method stated instead of defaulting to post or else it would use the requests instead of the mock.
Also had to mock out "put" and "post" since mocking requests made requestExceptions funny

tiftran:
will gut out the no longer necessary fields for bug creation in #1284 or a separate PR...

Issue #902, opened 2019-02-09
 Allow null values in Population Wizard when creating the Draft


shell1:
History: The goal is to be able to start and save the draft before you know all the answers - capturing information you have as you go. Today the validation on each section requires putting in dummy information (if you don't yet know the answers to the most critical fields) - in order to pass section by section validation on Required information.

Ticket work: Please make it so in the "Population Wizard" there is a way to leave any of these fields blank or "null" when saving.  These are all pretty complex and may be worked out in bits and parts.  Should only fail validation at the "ready for sign-offs" point.

Not sure if this should be part of this bug or bug #900: Please make any Null required fields Red in the Wizard mode.

in Bug #900 we want to throw a validation error if required fields are left empty when they click "ready for sign-offs". It would be nice if they could go into edit mode and easily see the "red" fields. User experience works if null fields are only "red" when they fail validation at "ready to ship" - or if they are "red from the first save that they were blank". Putting here as not sure if that is easiest done as part of this issue or the final validation.

jaredkerim:
The goal here is to allow users to be able to partially fill out the form on the Population page and save it, so if you only know the branches but not the population or visa versa, and be able to come back to it later.  But as with the rest of the form, if a section is only partially complete, it should be displayed as such on the detail page and block moving to 'Ready to Ship'

1) Make the following fields optional:
- firefox_channel
- firefox_version
- pref_key
- pref_type
- pref_branch

2) Update https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/models.py#L376 to check that those fields have non default/non dempty values

3) Update tests

I think that should cover it.

jaredkerim:
Okay I was wrong, I think the solution here is to move all the population information into the first form, the 'Overview' form, and make them optional and treat them the same way we treat all the other 'delayed required' fields, ie prevent 'Ready to Ship' until they're completed but make them optional right on the form.  

We need to keep all the pref fields on the same form as the branch form, and make that page strongly validated at save time, ie you can't save partially defined branches.

jaredkerim:
Okay so, new pass.  I think we need to introduce a new page to the form, so

1) Overview: only the required fields to create a placeholder for the experiment
2) Population: all the population and date fields, all optional (this page will then be editable while the experiment is live later on)
3) Firefox: pref/addon/branch information, these fields are all required to save because there's cross validation
4) Objectives & Analysis: all fields optional
5) Risks & Testing: all fields optional

[...]

PR #1264, opened 2019-05-10
Add survey fields to analysis section fixes #507

### LINKED ISSUE #507: Add survey fields
We should add new optional fields under the analysis section

Link "help" for this section to one location: https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-SurveyChecklist

Survey Required? - boolean

**If yes and add-on experiment:** 
Survey URL(s) - open text field
Launch timing - open text field

**If yes and pref-flip experiment**
Survey URL(s) - open text field
Duration until add-on expiration - open text field

If presenting 2 different options for pref-flip versus add-on... talk to @shell1 or @jaredkerim  about next best options.  Also I'd be interested in your opinion on what it would look like to put the specific help guidance from the page in some of the open text boxes (like "Launch Timing" or "Duration until add-on expiration")

uhlissuh:
Fixes #507 

This PR adds survey fields to the bottom of the objectives/analysis section.

 When editing the form, extra, non-required fields URLs and Launch Instructions are shown via JS if a user checks "Yes" for the survey radio button. 

surveyfields

On the overview page, the UI is as follows for the survey area: 

Screen Shot 2019-05-10 at 9 06 16 AM

Screen Shot 2019-05-10 at 9 07 33 AM

I think the UI on the overview page needs a bit more work. I could have a title of 'Surveys' inside the 'Analysis' section, like 'Risk Detail' in the Risks section, but perhaps I should only have that title when there is a survey, because if it were there when the survey is no... there'd be a big title with only line below it... 

See my review for questions about specific pieces of this PR!


app/experimenter/experiments/forms.py
```diff
@@ -552,7 +552,22 @@ def clean(self):
         return cleaned_data
 
 
+class RadioWidget(forms.widgets.RadioSelect):
+    template_name = "experiments/radio_widget.html"
+
+
+class RadioWidgetCloser(forms.widgets.RadioSelect):
+    template_name = "experiments/radio_widget_closer.html"
+
```
uhlissuh: I make a custom widget because the `RadioWidget` created here and used in the Risks section uses col-2 for spacing of the Yes and No buttons. This spacing is too wide for this application, and col-1s are too small for the button and label. I take them out of columns and just use some padding.

app/experimenter/experiments/forms.py
```diff
@@ -570,13 +585,50 @@ class ExperimentObjectivesForm(ChangeLogMixin, forms.ModelForm):
...
+        label="Survey URLs",
+        widget=forms.Textarea(attrs={"class": "form-control", "rows": 1}),
+    )
+    survey_instructions = forms.CharField(
+        required=False,
+        label=Experiment.SURVEY_INSTRUCTIONS_LABEL,
+        help_text=Experiment.SURVEY_LAUNCH_INSTRUCTIONS_HELP_TEXT,
+        widget=forms.Textarea(attrs={"class": "form-control", "rows": 10}),
+    )
+
```
uhlissuh: survey itself is required for the form... because a user can say No/Yes... is this correct? URLs and instructions are not required.

app/experimenter/experiments/forms.py
```diff
@@ -570,13 +585,50 @@ class ExperimentObjectivesForm(ChangeLogMixin, forms.ModelForm):
...
+        # form, but now are removing the survey, that we clear out the
+        # database of the url and instructions field
+        cleaned_data = super().clean()
+        if not cleaned_data.get("survey"):
+            cleaned_data["survey_urls"] = ""
+            cleaned_data["survey_instructions"] = ""
 
-class RadioWidget(forms.widgets.RadioSelect):
-    template_name = "experiments/radio_widget.html"
+        return cleaned_data
```
uhlissuh: Is the clean method the best place for making sure that if the user goes from Yes to No... that we are removing the urls and instructions from the DB?

app/experimenter/static/js/edit-objectives.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+
+
+$('[type="radio"]').change(function (e) {
+  updateSurvey();
+
+});
+
+updateSurvey();
```
uhlissuh: I took the 'hiding of the optional button' from the risk-technical section... where we show and hide an additional text box there (Though it's done incorrectly there and I plan to fix it in a separate PR). It makes total sense in the risk-technical section, because if a user says Yes to the risk-technical question, they are required then to put an explanation in the text box. I am debating whether it makes sense here... because urls and instructions are _not_ required by the form in the formal sense... but if you answer Yes to the survey... you really do need to add the URLs and instructions (I'm assuming). So I'm not sure what to do about that.

app/experimenter/templates/experiments/field_inline_help_link.html
```diff
@@ -0,0 +1,23 @@
...
+    {% include "experiments/field_input_inline.html" with field=field %}
+
+    {% include "experiments/field_errors_inline.html" with field=field %}
+
+    <div id="{{ field.id_for_label }}-help" class="text-muted collapse {% if field.errors %}in{% endif %} pt-2">
+      {{ field.help_text|safe }}
+    </div>
+
+  </div>
+</div>
```
uhlissuh: I added this template because the existing ones (`form-inline`) did not have the orientation where there is a help button that opens a separate window with a link (as opposed to popping up some help text under the textfield box)..

app/experimenter/templates/experiments/radio_field_label_inline.html
```diff
@@ -0,0 +1,17 @@
...
+  </div>
+  <div class="col-9">
+    <div class="row">
+      {{ field }}
+    </div>
+
+  </div>
+
+  {% include "experiments/field_errors_inline.html" with field=field %}
+</div>
```
uhlissuh: Again, needed a custom template to make an inline radio button that had a label on the left (as opposed to above).

app/experimenter/templates/experiments/radio_widget_closer.html
```diff
@@ -0,0 +1,7 @@
+{% for group, options, index in widget.optgroups %}
+  {% for option in options %}
+		<div class="radio ml-4">
+			{% include option.template_name with widget=option %}
+		</div>
+  {% endfor %}
+{% endfor %}
```
uhlissuh: Here is the custom widget that I referenced earlier in this review. I use `ml` to accomplish the right spacing.

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -7,4 +7,25 @@
...
+    </p>
+
+    <strong>Survey Launch Instructions</strong>
+    <p>
+      {{ experiment.survey_instructions }}
+    </p>
+  {% endif %}
+
+
+
```
uhlissuh: Per my discussion in the main part of this PR... in this section overview, we could add a title inside the box for surveys... but we might only want to show this title if there is a survey..? Because if we have this title even if there isn't a survey.. there would be a big title with only one line below it. Maybe that's fine? I dunno, this part of the UI probably needs more.

app/experimenter/static/js/edit-objectives.js
uhlissuh: also I used Jquery because risk-technical stuff used it..

app/experimenter/experiments/forms.py
glasserc: That seems correct. Per discussions in Slack, these fields are not "required" even if a user says "Yes" to a survey, because although we need them, we can ask for them manually later.

app/experimenter/experiments/forms.py
glasserc: I'm not sure about architecture but it seems OK to me.

app/experimenter/experiments/models.py
```diff
@@ -126,6 +126,11 @@ class Experiment(ExperimentConstants, models.Model):
         default=ExperimentConstants.ANALYSIS_DEFAULT, blank=True, null=True
     )
     analysis_owner = models.CharField(max_length=255, blank=True, null=True)
+
+    survey = models.NullBooleanField(default=None, blank=True, null=True)
+    survey_urls = models.TextField(blank=True, null=True)
+    survey_instructions = models.TextField(blank=True, null=True)
```
glasserc: I admit that I don't know 100% how Jared prefers these things, but I would probably have tried to avoid `null` and made them default to empty string. For `survey` I guess we have to leave null rather than default to "no".

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1062,6 +1065,40 @@ def test_form_saves_objectives(self):
...
+            request=self.request, data=data, instance=experiment
+        )
+
+        self.assertTrue(form.is_valid())
+
+        experiment = form.save()
+
+        self.assertFalse(experiment.survey)
+        self.assertFalse(experiment.survey_urls)
+        self.assertFalse(experiment.survey_instructions)
```
glasserc: Should these be false, or empty string? (Are you taking advantage of Python falsiness here? I would probably be explicit.)

app/experimenter/static/js/edit-objectives.js
glasserc: I think using jQuery for consistency is fine.

I'm a bit torn myself about hiding the `Optional` label. According to Shell's remarks on Slack, we *do* need this information, but we *don't* want to block the form submission. In other words, the fields are really optional. I think we should assume the user will read the lack of `Optional` as indicating the fields are required, which would make (some of) them not submit it, which almost makes the form required. I think we shouldn't get too fancy and just leave the `Optional` labels as they are.

I do wonder if it's possible to do this more simply, but I think it isn't. We can't use CSS trickery to hide/show the following elements because the `:checked` radio will be inside some `div`s. Can we use Bootstrap collapse even though it's a `radio` instead of a button?

app/experimenter/templates/experiments/field_inline_help_link.html
```diff
@@ -0,0 +1,23 @@
+<div class="form-group row">
+  <div class="col-3 text-right">
+    <label class="col-form-label" for="{{ field.id_for_label }}">
+      <strong>{{ field.label }}</strong>
+      {% if not field.field.required %}
+        <span class="text-muted optional-marker">Optional</span>
+      {% endif %}
+    </label>
+    <br>
+    <a target="_blank" rel="noreferrer noopener" href="{{ field.help_text }}">Help</a>
```
glasserc: I'm a bit confused here -- is `field.help_text` a URL, or HTML? You seem to be using it as a URL here but HTML down below.

app/experimenter/templates/experiments/field_inline_help_link.html
glasserc: I think this would make a great comment at the top of this template!

app/experimenter/templates/experiments/radio_widget_closer.html
```diff
@@ -0,0 +1,7 @@
+{% for group, options, index in widget.optgroups %}
+  {% for option in options %}
+		<div class="radio ml-4">
```
glasserc: The spacing looks a bit odd here -- you have tabs instead of spaces I think.

app/experimenter/templates/experiments/radio_field_label_inline.html
glasserc: It took me a bit but yes, I see the other radio fields have the labels above the buttons.

app/experimenter/experiments/forms.py
glasserc: I think this class could use a docstring saying something along these lines.

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -7,4 +7,25 @@
...
   {{ experiment.analysis|urlize|linebreaks }}
+
+  {{ experiment.SURVEY_REQUIRED_LABEL }}
+  <strong class="{% if experiment.survey %}text-danger{% endif %}">
+    {{ experiment.survey|yesno:"Yes,No" }}
+  </strong>
+
+  {% if experiment.survey %}
+  <br>
+    <strong>Survey URLs</strong>
```
glasserc: The spacing seems a bit strange here.

app/experimenter/templates/experiments/section_analysis.html
glasserc: I think this is fine, and Shell seemed OK with it too!

app/experimenter/static/js/edit-objectives.js
uhlissuh: It might be possible with bootstrap.

app/experimenter/experiments/constants.py
```diff
@@ -581,6 +583,34 @@ class ExperimentConstants(object):
...
+            The most common survey trigger is upon add-on expiration
+            (built into the add-on).  Add-on experiments don't leave
+            breadcrumbs behind on the user systems, so there is no way
+            to target those users for surveying after the Normandy recipe
+            is ended.  The Normandy end date needs to be scheduled AFTER
+            the add-on has expired (for the last enrolled users), plus a 7
+            day survey response window.
+        </p>
+
+
```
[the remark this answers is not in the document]
uhlissuh: oops

app/experimenter/templates/experiments/radio_widget_closer.html
uhlissuh: this is odd.. not sure what happened.

app/experimenter/templates/experiments/section_analysis.html
uhlissuh: also odd yes. I'm sorry for making you deal with that.

app/experimenter/templates/experiments/field_inline_help_link.html
uhlissuh: Yea, that later piece was not needed. I think I was getting kinda some template fatigue going and I just missed taking that out. Thank you!!

uhlissuh: It's a URL. That was the point of this template... to have a input field with a help button that was a link, instead of popping up helper text underneath the box.

app/experimenter/static/js/edit-objectives.js
uhlissuh: I removed the optional stuff and just wondering if we think I should explore switching this over to bootstrap  or leave as is.

uhlissuh: wondering also what @jaredkerim is feeling.

app/experimenter/experiments/models.py
uhlissuh: Good point, @jaredkerim can weigh in too.

app/experimenter/experiments/forms.py
```diff
@@ -570,13 +589,50 @@ class ExperimentObjectivesForm(ChangeLogMixin, forms.ModelForm):
...
+            "survey_urls",
+            "survey_instructions",
+        )
 
+    def clean(self):
+        # if the survey radio button is false, we make sure to not
+        # save any information for urls or instructions. we also want
+        # to make sure that if they once answered yes, and saved the
+        # form, but now are removing the survey, that we clear out the
+        # database of the url and instructions field
```
jaredkerim: I understand your logic here, but this isn't generally a pattern I think we adopt in other places, ie clearing out fields.  I *think* that could lead to confusing interactions/behaviours.  I would suggest rather than clearing out the fields, we only show the survey information if the survey yes/no field is set to yes.

So then it would look like, surveys required: yes. urls/info: etcetc.  Survey section is shown.  Go back and change survey required: no. info is hidden on detail page.  Oops it is still required, go back and change back to yes, info shows back up again.

So then on the detail page you just check that the survey required field is true before showing the info.

Does that seem sensible @uhlissuh ?

app/experimenter/experiments/models.py
jaredkerim: So there's a lot of cases where we set fields to blank/null=True and let the default be null/None so that we can distinguish between "the user has not yet interacted with this field" and "the user did interact and entered some falsy value".  So for risk questions, we don't want to default to "No" becuase we can't know that's true and we want to force the user to make an explicit declaration.

In this case though, I think it is sensible to have the survey required field default to No, since defaulting to that doesn't introduce risk to the experiment.  Because I think our behaviour is that if an experiment owner doesn't interact with the survey fields at all the experiment should proceed as normal.

Edit: confirmed with @shell1

app/experimenter/static/js/edit-objectives.js
jaredkerim: @uhlissuh @glasserc Yeah so this is an ugly situation we've backed ourselves into.

Originally the design was that some fields were 'required' in the django form sense, and some were 'optional'.  So then on a given form page, you couldn't save the form at all without all the required fields being completed, and then the optional ones may or may not need to be completed before you can move to 'ready to ship'.

But then people complained that they wanted to save some fields and not others at the same time (because they don't know the answers yet) so we started changing some 'required' fields to 'optional' in the django form sense, even though they're 'delayed required' ie block ready to ship. 

So now we're in this stupid situation where fields say 'optional' even though they're more like 'delayed required' and we don't make that clear anywhere.

So we filed this ticket:  https://github.com/mozilla/experimenter/issues/902 to address this more concretely.  The point of that ticket is to basically move every single field to be 'django optional' so nothing is required to be able to save a form page, but fields that are still 'delayed required' continue to behave that way.  

Part of that change will be to take out those 'optional' labels we show.  So I would say just take out that one line of jquery that hides the optional label, leave the rest as is, and we can fix that more coherently in the other ticket. /rant

app/experimenter/templates/experiments/edit_objectives.html
```diff
@@ -8,6 +8,22 @@
   </a> Objectives &amp; Analysis
 {% endblock %}
 
+{% block edit_form %}
+  {% include "experiments/field_inline.html" with field=form.objectives%}
+  {% include "experiments/field_inline.html" with field=form.analysis %}
+  {% include "experiments/field_inline.html" with field=form.analysis_owner %}
+  {% include "experiments/radio_field_label_inline.html" with field=form.survey %}
+
+  <div id="survey-url-and-instructions"  class="collapse">
```
jaredkerim: 👍

app/experimenter/templates/experiments/edit_objectives.html
```diff
@@ -8,6 +8,22 @@
...
+    {% include "experiments/field_inline_help_link.html" with field=form.survey_urls %}
+    {% include "experiments/field_inline.html" with field=form.survey_instructions %}
+  </div>
+
```
jaredkerim: That's a lot of extra new lines?

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -7,4 +7,25 @@
 {% block section_content %}
   <p><strong>Data Science Owner: {{ experiment.analysis_owner }}</strong></p>
   {{ experiment.analysis|urlize|linebreaks }}
+
+  {{ experiment.SURVEY_REQUIRED_LABEL }}
+  <strong class="{% if experiment.survey %}text-danger{% endif %}">
+    {{ experiment.survey|yesno:"Yes,No" }}
+  </strong>
+
+  {% if experiment.survey %}
```
jaredkerim: 👍

app/experimenter/templates/experiments/edit_objectives.html
```diff
@@ -25,3 +41,8 @@ <h4><span class="fas fa-info-circle"></span> Objectives &amp; Analysis</h4>
     Describe both in detail here.
   </p>
 {% endblock %}
+
+{% block extrascripts %}
+    <script src="{% static "js/edit-objectives.js" %}"></script>
+
```
jaredkerim: extra newline?

app/experimenter/templates/experiments/field_inline_help_link.html
```diff
@@ -0,0 +1,22 @@
+<!-- this template exists to provide form-inline html with a
```
jaredkerim: Okay I see what you did here, this is sensible.  There's a lot of fields where the help text is only a link, like:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L619-L706

and if we're going to introduce a new form field template that expects the help text to just be an outgoing link instead of a drop down we should probably switch everything to use it, but that's a lot of changes that don't belong in this PR.  

So I would say either move this template out into a separate issue/pr, and migrate all the fields that need it to it, or land the template as part of this change and file a separate issue to move all the other fields over to using it, whichever feels saner to you.

This is one of those cases where this is kind of a separate change that supports this PR but if we reverted this PR, we wouldn't want to also lose this change, becuase it's just a generally better way to organize this.  So I'll leave it to you to decide whether to keep it here and file the separate issue to change other fields to use it or do all that in a separate PR.  

Either way good call this is a good change 👍

app/experimenter/templates/experiments/radio_widget_closer.html
jaredkerim: Oh yep those are tab characters, definitely switch those to spaces.  Tabs are the devil 👿

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -7,4 +7,25 @@
 {% block section_content %}
   <p><strong>Data Science Owner: {{ experiment.analysis_owner }}</strong></p>
   {{ experiment.analysis|urlize|linebreaks }}
+
+  {{ experiment.SURVEY_REQUIRED_LABEL }}
```
jaredkerim: I think this gets 

    <p><strong>...</strong></p>

around it?

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -7,4 +7,25 @@
...
+  <strong class="{% if experiment.survey %}text-danger{% endif %}">
```
jaredkerim: Don't forget to wrap your <strong>'s in <p>'s so you get the nice line break spacing

app/experimenter/templates/experiments/field_inline_help_link.html
uhlissuh: Hey sooo.... I'm not totally getting what you're saying here. The fields you are highlighting seem kind of different. The review links, like `REVIEW_BUGZILLA_HELP_TEXT` and `REVIEW_ENGINEERING_HELP_TEXT`.. aren't those used to display the word help after the checkbox in the required signoff section?  In that case the formatting is decided by Django and just comes from declaring `help_text` on the form... right? I can't see how what we are doing in that section relates to what we're doing here. And in the case of those other constants, like `RISK_INTERNAL_ONLY_HELP_TEXT` and `RISK_PARTNER_RELATED_HELP_TEXT`.. those are using the radio widget.. final formatting also defined by saying `help_text` on the form... I guess I am not seeing why those need to be changed or exactly how it relates to needing an inline_field with help link instead of help drop down.

app/experimenter/templates/experiments/edit_objectives.html
uhlissuh: i removed some of all the extra new lines this file got 🤦‍♀

app/experimenter/static/js/edit-objectives.js
uhlissuh: I took the optional stuff out of the jQuery.

app/experimenter/experiments/models.py
uhlissuh: I made this change... hopefully you can check to make sure I understood you correctly.

app/experimenter/experiments/forms.py
uhlissuh: Yes totally. I took out the logic that was clearing out the database after a switch from Yes to No.

app/experimenter/experiments/forms.py
```diff
@@ -570,13 +589,50 @@ class ExperimentObjectivesForm(ChangeLogMixin, forms.ModelForm):
         widget=forms.Textarea(attrs={"class": "form-control", "rows": 20}),
     )
 
+    survey = forms.TypedChoiceField(
```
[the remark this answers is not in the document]
uhlissuh: made the change.

app/experimenter/experiments/forms.py
```diff
@@ -552,7 +552,22 @@ def clean(self):
...
+
+
 class ExperimentObjectivesForm(ChangeLogMixin, forms.ModelForm):
+    RADIO_OPTIONS = ((False, "No"), (True, "Yes"))
+
+    def coerce_truthy(value):
+        if value.lower() == "true":
+            return True
+        elif value.lower() == "false":
+            return False
```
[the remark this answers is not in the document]
uhlissuh: I think I have done what you both mean.

uhlissuh:
OK thanks for the review @jaredkerim! So the outstanding things are:

- me not understanding what you mean on the template help link thing
- checking that I did the right thing with the `coerce_truthy` stuff
- checking to make sure I changed the model field to booleanfield correctly....

Can't wait to get this PR off my mental ledger.

app/experimenter/experiments/forms.py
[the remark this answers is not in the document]
jaredkerim: nit: you don't need to have 'else' here, you can just say 'return false'

app/experimenter/experiments/models.py
```diff
@@ -126,6 +126,11 @@ class Experiment(ExperimentConstants, models.Model):
         default=ExperimentConstants.ANALYSIS_DEFAULT, blank=True, null=True
     )
     analysis_owner = models.CharField(max_length=255, blank=True, null=True)
+
+    survey_required = models.BooleanField(default=False)
```
jaredkerim: 👍

app/experimenter/templates/experiments/field_inline_help_link.html
jaredkerim: It's okay nevermind me, I'll have to look at it again and figure out if there's something we can do to make things more consistent, but this is fine.

app/experimenter/templates/experiments/radio_widget_closer.html
```diff
@@ -0,0 +1,10 @@
+<!-- the buttons are moved closer together as compared to RadioWidget -->
+<!-- by using ml-4 instead of putting in bootstrap columns -->
+
+{% for group, options, index in widget.optgroups %}
+  {% for option in options %}
+		<div class="radio ml-4">
```
jaredkerim: I still see tabs here?

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -7,4 +7,27 @@
 {% block section_content %}
   <p><strong>Data Science Owner: {{ experiment.analysis_owner }}</strong></p>
   {{ experiment.analysis|urlize|linebreaks }}
+
+  <p>
+  {{ experiment.SURVEY_REQUIRED_LABEL }}
```
jaredkerim: Formatting here is still a bit wonky:

Screenshot 2019-05-22 11 25 36

"Is a survey required?" should be `<strong>` and I think you can get rid of the `<br>` altogether?

app/experimenter/experiments/forms.py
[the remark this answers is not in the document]
uhlissuh: oops, duh.

app/experimenter/templates/experiments/section_analysis.html
uhlissuh: ok got it

PR #1280, opened 2019-05-13
Add filter for surveys fixes 1265

### LINKED ISSUE #1279: Add filter for experiments with surveys
We need a way to filter out experiments on the list view that have surveys attached to them so that the surveys team can easily check.

uhlissuh:
Fixes #1279 
Depends on #1264 

Here we are adding a filter for experiments that have surveys attached to them. 

Screen Shot 2019-05-13 at 3 35 53 PM

When a user has selected the radio button we filter and add "with surveys" to the title.

Screen Shot 2019-05-13 at 3 54 48 PM

### DIFF
--- a/app/experimenter/experiments/views.py
+++ b/app/experimenter/experiments/views.py
@@ -33,7 +33,7 @@ from experimenter.experiments.models import Experiment
 
 class ExperimentFiltersetForm(forms.ModelForm):
     in_qa = forms.BooleanField(required=False)
-
+    surveys = forms.BooleanField(required=False)
     search = forms.CharField(required=False)
 
     class Meta:
@@ -47,6 +47,7 @@ class ExperimentFiltersetForm(forms.ModelForm):
             "project",
             "owner",
             "in_qa",
+            "surveys",
             "archived",
         )
 
@@ -169,6 +170,12 @@ class ExperimentFilterset(filters.FilterSet):
         method="in_qa_filter",
     )
 
+    surveys = filters.BooleanFilter(
+        label="Show experiments with surveys",
+        widget=forms.CheckboxInput(),
+        method="surveys_filter",
+    )
+
     class Meta:
         model = Experiment
         form = ExperimentFiltersetForm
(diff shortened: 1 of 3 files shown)


app/experimenter/templates/experiments/list.html
```diff
@@ -25,16 +25,20 @@ <h3 class="m-0">
...
     {% if filter.form.project.value %}
       {{ filter.form.get_project_display_value }}
     {% endif %}
 
     Experiment{{ paginator.count|pluralize:"s" }}
 
+    {% if filter.form.in_qa.value %}
+       in QA
+    {% endif %}
+
```
uhlissuh: While I was adding the title language for surveys, I realized that this piece of code right here, introduced by me weeks back, was in a non-ideal location. With it being in it's previous location, above the code that inserts the word "Experiment" into the title at line 32, the title reads " X in QA Experiments" instead of the better title... "X Experiments in QA". That's why I moved this bit of code to below line 32.

app/experimenter/templates/experiments/list.html
```diff
@@ -145,12 +149,7 @@ <h5>{{ experiment.population }}</h5>
     </div>
 
     {% for field in filter.form %}
-      {% if field.name == "archived" %}
-        <label>
-          {{ field }}
-          {{ field.label }}
-        </label>
-      {% elif field.name == "in_qa" %}
+      {% if field.name == "in_qa" or field.name == "surveys" or field.name == "archived" %}
```
uhlissuh: I also realized that we had some redundant pieces of template here that could be combined into one larger `if` statement.

app/experimenter/experiments/views.py
```diff
@@ -238,6 +245,12 @@ def in_qa_filter(self, queryset, name, value):
         else:
             return queryset
 
+    def surveys_filter(self, queryset, name, value):
+        if value:
+            return queryset.filter(survey=True)
+        else:
```
jaredkerim: nit: don't need the else here

app/experimenter/templates/experiments/list.html
jaredkerim: Ah the ol PR switcheroo

app/experimenter/experiments/views.py
uhlissuh: I think you do.. I tried to take it away and I got errors. I think every filter has to return a queryset by default.

jaredkerim: No you literally don't need the `else`:

    if value:
        return queryset.filter(...)
    return queryset

PR #1284, opened 2019-05-16
check for bug ids and change to check 'assigned to' val fixes #1281

### LINKED ISSUE #1281: Handle bug ID does not exist in bugzilla client
When we call the bugzilla API to create a ticket, it's possible the bug numbers entered may not exist on the bugzilla instance, so we should handle that case.

tiftran:


jaredkerim:
@tiftran @shell1 We're gonna pause this PR pending further design discussion.

app/experimenter/experiments/bugzilla.py
```diff
@@ -105,55 +122,33 @@ def format_creation_bug_body(experiment):
...
         "type": "task",
         "priority": "P3",
-        "see_also": [get_bugzilla_id(experiment.data_science_bugzilla_url)],
         "url": experiment.experiment_url,
-        "whiteboard": experiment.STATUS_REVIEW_LABEL,
     }
-    if experiment.bugzilla_tracking_key:
-        bug_data[experiment.bugzilla_tracking_key] = "?"
+
+    if user_exists(experiment.owner.email):
```
jaredkerim: I would do this in `create_experiment_bug` rather than in here.  We may want to use this format function in some other context and then it would be unexpected for it to make outgoing network calls.  For these optional fields, just pass them in as args to this function.

app/experimenter/experiments/bugzilla.py
```diff
@@ -105,55 +122,33 @@ def format_creation_bug_body(experiment):
...
+        bug_data["assigned_to"] = experiment.owner.email
+
+    data_science_bug_id = get_bugzilla_id(experiment.data_science_bugzilla_url)
+    if bug_exists(data_science_bug_id):
```
jaredkerim: Same here

app/experimenter/experiments/bugzilla.py
```diff
@@ -67,24 +67,41 @@ def format_update_body(experiment):
...
         settings.BUGZILLA_UPDATE_URL.format(id=experiment.bugzilla_id),
-        body,
-        method=requests.put,
+        requests.put,
+        data=body,
     )
 
 
-def make_bugzilla_call(url, data, method):
+def user_exists(user):
```
jaredkerim: Very nice 👍

app/experimenter/experiments/bugzilla.py
```diff
@@ -105,55 +122,33 @@ def format_creation_bug_body(experiment):
...
 
 def create_experiment_bug(experiment):
 
     bug_data = format_creation_bug_body(experiment)
     response_data = make_bugzilla_call(
-        settings.BUGZILLA_CREATE_URL, bug_data, method=requests.post
+        settings.BUGZILLA_CREATE_URL, requests.post, data=bug_data
     )
 
-    # The experiment owner might not exist in bugzilla
```
jaredkerim: Bye bye, error checking code! 👋

app/experimenter/experiments/tests/mixins.py
```diff
@@ -58,37 +89,27 @@ def setupMockBugzillaCreationFailure(self):
 
     def setUpMockBugzillaInvalidUser(self):
 
-        def mock_reject_assignee(url, bug_data):
-            if "assigned_to" in bug_data:
-                return self.buildMockFailureResponse()
-            else:
-                return self.buildMockSuccessResponse()
+        self.mock_bugzilla_requests_get.side_effect = [
+            self.buildMockFailureResponse(),
```
jaredkerim: Wow I didn't know you could do that that's so coooool 🐧

app/experimenter/experiments/tests/mixins.py
```diff
@@ -58,37 +89,27 @@ def setupMockBugzillaCreationFailure(self):
...
+            self.buildMockSuccessBugResponse(),
+            self.buildMockSuccessBugResponse(),
+        ]
 
-        self.mock_bugzilla_requests_post.side_effect = mock_reject_assignee
+    def setUpMockBugzillaFirstBadTicket(self):
```
jaredkerim: The body of these helper methods is so short and if they're only used once I'd just move this to the actual test.

app/experimenter/experiments/bugzilla.py
```diff
@@ -164,4 +164,7 @@ def create_experiment_bug(experiment):
 
 def get_bugzilla_id(bug_url):
     query = urlparse(bug_url).query
-    return int(parse_qs(query)["id"][0])
+    try:
+        return int(parse_qs(query)["id"][0])
```
jaredkerim: Do this exact check but in the form at validation time so we don't even let them enter it

jaredkerim: Oh but we also have to do this check here as well for the case that someone already entered it in some invalid form in the production database by teh time we launch this code, but if you can't parse the id out of the url qs then just fail silently and pass as if the url wasn't set at all.

app/experimenter/experiments/bugzilla.py
```diff
@@ -162,3 +159,14 @@ def create_experiment_bug(experiment):
...
+
+
+def set_bugzilla_id_value(bug_url):
+    if "show_bug.cgi?id=" in bug_url:
+        try:
+            data_science_bug_id = get_bugzilla_id(bug_url)
+            if bug_exists(data_science_bug_id):
+                return [data_science_bug_id]
+        except ValueError:
+            return None
```
jaredkerim: Both of these `return None` are redundant since a function will always return None.  Also I think this try/except is also unnecessary, I would change the get_bugzilla_id to `.get("id", [None])[0]` so that it doesn't raise at all and then just check if it's None before you make the bug_exists call.

app/experimenter/experiments/bugzilla.py
```diff
@@ -162,3 +159,14 @@ def create_experiment_bug(experiment):
 def get_bugzilla_id(bug_url):
     query = urlparse(bug_url).query
     return int(parse_qs(query)["id"][0])
+
+
+def set_bugzilla_id_value(bug_url):
+    if "show_bug.cgi?id=" in bug_url:
```
jaredkerim: Also if you parse the qs and use `.get` I don't think you need to do this check either.

app/experimenter/experiments/forms.py
```diff
@@ -47,10 +48,15 @@ def clean(self, value):
...
 
         if cleaned_value:
+            err_str = "Please Provide a Valid URL ex: {}show_bug.cgi?id=1234"
             if settings.BUGZILLA_HOST not in cleaned_value:
-                err_str = (
-                    "Please Provide a Valid URL ex: {}show_bug.cgi?id=1234"
+                raise forms.ValidationError(
+                    err_str.format(settings.BUGZILLA_HOST)
                 )
+            query = urlparse(cleaned_value).query
```
jaredkerim: Since this is basically the same as what we do in `get_bugzilla_id` I'd import that function and use it here instead of reimplementing it, and then just check if it's `None` and raise the ValidationError that way.

app/experimenter/experiments/bugzilla.py
```diff
@@ -161,4 +158,12 @@ def create_experiment_bug(experiment):
...
     query = urlparse(bug_url).query
-    return int(parse_qs(query)["id"][0])
+    bugzilla_id = parse_qs(query).get("id", [""])[0]
+    if bugzilla_id.isnumeric():
+        return int(bugzilla_id)
+
+
+def set_bugzilla_id_value(bug_url):
+    data_science_bug_id = get_bugzilla_id(bug_url)
+    if bug_exists(data_science_bug_id):
```
jaredkerim: `if data_science_bug_id and bug_exists(data_science_bug_id)`

PR #1286, opened 2019-05-17
changes so serializer would return proper bool/int values fixes #1261

### LINKED ISSUE #1261: Recipe serializer should emit proper boolean/int values
For boolean and integer values, the recipe serializer should emit actual booleans/integers instead of strings.  Right now it does:

     "arguments": {
        "preferenceBranchType": "default",
        "slug": "pref-asdfasdf-nightly-55.0-bug-1395380",
        "experimentDocumentUrl": "https://localhost/experiments/asdfasdf/",
        "preferenceName": "browser.newtabpage.activity-stream.enabled",
        "preferenceType": "boolean",
        "branches": [
          {
            "ratio": 50,
            "slug": "on",
            "value": "true"
          },
          {
            "ratio": 50,
            "slug": "off",
            "value": "false"
          }
        ]
      }

tiftran:


app/experimenter/experiments/serializers.py
```diff
@@ -154,11 +154,19 @@ def get_countries(self, obj):
...
 
     class Meta:
         model = ExperimentVariant
         fields = ("ratio", "slug", "value")
 
+    def get_value(self, obj):
+        pref_type = obj.experiment.pref_type
+        if pref_type in (Experiment.PREF_TYPE_BOOL, Experiment.PREF_TYPE_INT):
+            return json.loads(obj.value)
+        else:
```
jaredkerim: nit: don't need `else` here

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -43,9 +44,42 @@ def test_field_returns_none_if_no_datetime_passed_in(self):
...
+        variant = ExperimentVariantFactory.create(experiment=experiment)
+        serializer = ExperimentRecipeVariantSerializer(variant)
+
+        self.assertEqual(type(serializer.data["value"]), bool)
+        self.assertEqual(
+            serializer.data,
+            {
+                "ratio": variant.ratio,
+                "slug": variant.slug,
+                "value": json.loads(variant.value),
```
jaredkerim: I would just hard code the values passed into the factory and hard them here in the check.

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -43,9 +44,42 @@ def test_field_returns_none_if_no_datetime_passed_in(self):
...
+        variant = ExperimentVariantFactory.create(experiment=experiment)
         serializer = ExperimentRecipeVariantSerializer(variant)
+
+        self.assertEqual(type(serializer.data["value"]), int)
+        self.assertEqual(
+            serializer.data,
+            {
+                "ratio": variant.ratio,
+                "slug": variant.slug,
+                "value": json.loads(variant.value),
```
jaredkerim: same hardcode

jaredkerim:
👍

PR #1287, opened 2019-05-17
Add ability to clone experiments fixes #163

### LINKED ISSUE #163: Add the ability to clone experiments
We need the ability to clone existing experiments to relaunch rejected experiments.

uhlissuh:
Fixes #163 

This PR adds the ability to clone an experiment so that a user doesn't have to go through the form fillout process if a new experiment is going to be very similar to an old experiment. 

clone


app/experimenter/experiments/views.py
```diff
@@ -327,6 +336,99 @@ def get_initial(self):
...
+        experiment.status = ExperimentConstants.STATUS_DRAFT
+
+        experiment.owner = get_user_model().objects.get(
+            id=self.request.user.id
+        )
+        experiment.parent = Experiment.objects.filter(
+            slug=self.kwargs["slug"]
+        ).get()
+        experiment.proposed_start_date = datetime.date.today()
+        experiment.archived = False
```
jaredkerim: Yep good idea.  Actually this is kind of fun, you can create an archived template and then clone it a bunch.

app/experimenter/experiments/views.py
```diff
@@ -327,6 +336,99 @@ def get_initial(self):
...
+        experiment.slug = slugify(experiment.name)
+        experiment.status = ExperimentConstants.STATUS_DRAFT
+
+        experiment.owner = get_user_model().objects.get(
+            id=self.request.user.id
+        )
+        experiment.parent = Experiment.objects.filter(
+            slug=self.kwargs["slug"]
+        ).get()
+        experiment.proposed_start_date = datetime.date.today()
```
jaredkerim: Hrmm I think I said this to you but I take it back, is it possible to set this to None?

app/experimenter/experiments/views.py
```diff
@@ -327,6 +336,99 @@ def get_initial(self):
...
+        # duplicate the old experiment by setting id=none, clearing
+        # appropriate fields, then calling save()
+
+        experiment.id = None
+        experiment.name = clone_experiment_name(experiment.name)
+        experiment.slug = slugify(experiment.name)
+        experiment.status = ExperimentConstants.STATUS_DRAFT
+
+        experiment.owner = get_user_model().objects.get(
+            id=self.request.user.id
```
jaredkerim: Very good catch

app/experimenter/experiments/views.py
```diff
@@ -327,6 +336,99 @@ def get_initial(self):
...
+            "review_comms",
+            "review_impacted_teams",
+        ]
+
+        # duplicate the old experiment by setting id=none, clearing
+        # appropriate fields, then calling save()
+
+        experiment.id = None
+        experiment.name = clone_experiment_name(experiment.name)
+        experiment.slug = slugify(experiment.name)
```
jaredkerim: If you move this to a form you should be able to use the name/slug mixins to do this automatically

app/experimenter/experiments/views.py
```diff
@@ -327,6 +336,99 @@ def get_initial(self):
...
+        experiment.archived = False
+
+        for field in set_to_none_fields:
+            setattr(experiment, field, None)
+
+        experiment.save()
+
+        # for the variants on the old experiment, duplicate each
+        # with id=none, set the experiment foreignkey to the new clone
+        for variant in variants:
```
jaredkerim: 👍

app/experimenter/experiments/views.py
```diff
@@ -327,6 +336,99 @@ def get_initial(self):
...
+            variant.id = None
+            variant.experiment = experiment
+            variant.save()
+
+        ExperimentChangeLog.objects.create(
```
jaredkerim: 👍 👍 👍

app/experimenter/templates/base.html
```diff
@@ -104,21 +104,20 @@ <h3 class="pt-2">
...
+                  {{ message }}
+                </p>
+              {% endfor %}
+            </div>
           </div>
         </div>
       </div>
-
-    </div>
     {% endif %}
```
uhlissuh: This is a whitespace change...  from fixing indentation of the block after the `{% if messages %}`, the reason why this has a change at all is because in earlier iterations this PR, this code was altered for clone success message notifications. I've taken those success messages out now that it's obvious something is happening when you go to make the clone. Also, to make those messages work now that we use an ajax request and modal seems tricky and possibly not worth the effort. Django `messages` use the request/response cycle to display messages the user hasn't seen yet, and we wouldn't be able to use that with modal and ajax.

uhlissuh:
The gif I recorded was before I changed the button on the model to be 'Clone' instead of 'Send'.

app/experimenter/experiments/views.py
uhlissuh: @jaredkerim, made this change in the latest version.

app/experimenter/experiments/models.py
```diff
@@ -660,6 +665,64 @@ def firefox_channel_sort():
             output_field=models.IntegerField(),
         )
 
+    def clone(self, name, user):
```
jaredkerim: This method is a work of art

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1043,6 +1043,65 @@ def test_experiment_firefox_channel_sort_does_sorting(self):
             ],
         )
 
+    def test_clone(self):
```
jaredkerim: Yeah this is a bit tricky, in some senses we should check literally every single field here, but the test would be super long, and also super fragile because we'd have to update it for any/every field change we land so, i'm on the fence about how explicit to make this test. @tiftran what do you think?

app/experimenter/experiments/serializers.py
```diff
@@ -238,3 +240,27 @@ def get_arguments(self, obj):
             return ExperimentRecipePrefArgumentsSerializer(obj).data
         elif obj.is_addon_experiment:
             return ExperimentRecipeAddonArgumentsSerializer(obj).data
+
+
+class ExperimentCloneSerializer(serializers.ModelSerializer):
```
jaredkerim: Very nice 💯

app/experimenter/experiments/tests/test_models.py
tiftran: could we utilize something like the ```set_to_none_fields``` and have a loop through the fields to assert that they're ```None```. That would maybe reduce the amount of future updating of this test? Other than that, I'm not too sure, but I agree this test would be super long if we check every field...

uhlissuh: After some thought, I think we should just test some fields, not all, for now.

Issue #1235, opened 2019-05-02
Tooling should catch bad merges


glasserc:
The CI started failing after commit https://github.com/mozilla/experimenter/commit/95345463a6cb5f28b85ceaa2e2c3e2ea60f7a5d9. This commit added a migration 49 even though migrations through 52 already existed. This wasn't caught until the merge happened. It would be good to have tooling that prevented this sort of thing, probably by running the tests on any prospective merge.

One hypothetical way to provide this tooling is for the CI, on every PR, to not just run the tests against the code in the PR, but also against the commit that would be produced by merging the PR. Travis offers this feature. Unfortunately, it seems like CircleCI doesn't yet provide this feature (see https://ideas.circleci.com/ideas/CCI-I-926).

Another approach is to use something like [bors](https://bors.tech/). I personally don't love bors, and even if I did, I'd consider it @jaredkerim's decision whether or not to use it.

jaredkerim:
We enabled the "Branch must be up to date with master" branch protection in GH to solve this and I think it's working so I'm going to close this.

PR #1299, opened 2019-05-23
Fix ordering of migrations for build fail after merging surveys PR

uhlissuh:
The build is currently failing due to ordering of migrations problem. This PR fixes the ordering!


mythmon:
For the record, #1235 would have caught this ahead of time.

PR #1300, opened 2019-05-24
add periodic task to poll normandy to update exp. status fixes #158

### LINKED ISSUE #158: Add a worker for polling when experiments are published by Normandy

tiftran:


app/experimenter/experiments/tests/test_normandy.py
```diff
@@ -0,0 +1,45 @@
...
+from experimenter.experiments.normandy import (
+    NormandyError,
+    make_normandy_call,
+    get_recipe,
+)
+from experimenter.experiments.tests.mixins import MockNormandyMixin
+
+
+class TestMakeNormandyCall(MockNormandyMixin, TestCase):
+
```
tiftran: Not completely sure how to go about testing...

app/experimenter/settings.py
```diff
@@ -278,6 +279,12 @@
 CELERY_BROKER_URL = "redis://{host}:{port}/{db}".format(
     host=REDIS_HOST, port=REDIS_PORT, db=REDIS_DB
 )
+CELERY_BEAT_SCHEDULE = {
+    "debug_task": {
+        "task": "experimenter.experiments.tasks.update_experiment_status",
+        "schedule": 10,
```
tiftran: Was not sure what would a reasonable time interval would be

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+                    "email"
+                ]
+                approver, _ = get_user_model().objects.get_or_create(
+                    email=approver_email
+                )
+
+                experiment.status = status_mapping[experiment.status]
+                experiment.save()
+
+                experiment.changes.create(
```
mythmon: I'm a little fuzzy on this part. @jkerim can you take a look?

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+    accepted_experiments = Experiment.objects.filter(
+        Q(status=Experiment.STATUS_ACCEPTED) | Q(status=Experiment.STATUS_LIVE)
+    )
+    status_mapping = update_mapping()
+    for experiment in accepted_experiments:
+        try:
+            recipe_data = normandy.get_recipe(experiment.normandy_id)
+            if needs_to_be_updated(recipe_data["enabled"], experiment.status):
+
+                approver_email = recipe_data["approval_request"]["approver"][
```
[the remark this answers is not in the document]
tiftran: just to be sure, you're referring to:  
 ``` json      
"enabled_states": [
            {
                "id": 921,
                "created": "2019-05-14T19:44:36.139792Z",
                "creator": {
                    "id": 1,
                    "first_name": "Michael",
                    "last_name": "Cooper",
                    "email": "mcooper@mozilla.com"
                },
                "enabled": true,
                "carryover_from": null
            }
        ]
```
?

File: app/experimenter/experiments/tests/test_normandy.py [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+        response_data = make_normandy_call("/url/")
+
+        self.assertEqual(response_data, mock_response_data)
+
+    def test_make_normandy_call_with_request_exception(self):
+        self.mock_normandy_requests_get.side_effect = RequestException()
+
+        with self.assertRaises(NormandyError):
+            make_normandy_call("/url/")
```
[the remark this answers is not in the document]
tiftran: yup, will do!

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,75 @@ def test_internal_only_does_not_add_comment(self):
         self.mock_bugzilla_requests_post.assert_not_called()
 
         self.assertEqual(Notification.objects.count(), 0)
+
+
+class TestUpdateExperimentStatus(
```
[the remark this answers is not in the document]
tiftran: yup, will do!

docker-compose.yml
```diff
@@ -28,6 +28,17 @@ services:
...
       - public_nw
+  beat:
+    image: app:build
+    env_file: .env
+    links:
+      - db
+      - redis
+    volumes:
+      - ./app:/app
+    command: bash -c "/app/bin/wait-for-it.sh db:5432 -- celery -A experimenter beat -l debug"
```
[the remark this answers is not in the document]
tiftran: i don't think it's set up and installed in prod... would have to double check that...

app/experimenter/experiments/normandy.py
```diff
@@ -0,0 +1,31 @@
...
+    except requests.exceptions.RequestException as e:
+        logging.exception("Error calling Normandy API: {}".format(e))
+        raise NormandyError(*e.args)
+    except ValueError as e:
+        logging.exception("Error parsing JSON Normandy response: {}".format(e))
+        raise NormandyError(*e.args)
+
+
+def get_recipe(recipe_id):
+    recipe_url = settings.NORMANDY_API_RECIPE_URL.format(id=recipe_id)
```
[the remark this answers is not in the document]
tiftran: yup, it's already defined https://github.com/mozilla/experimenter/blob/master/app/experimenter/settings.py#L316

app/experimenter/experiments/tasks.py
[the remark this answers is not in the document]
mythmon: Exactly, yes.

docker-compose.yml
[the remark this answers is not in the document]
mythmon: I'd like to at least figure out the state of this before signing off on this PR, if not making sure it is ready. We can talk about this next week on Slack if you like.

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+@app.task
+@metrics.timer_decorator("update_experiment_status.timing")
+def update_experiment_status():
+    metrics.incr("update_experiment_status.started")
+    accepted_experiments = Experiment.objects.filter(
+        Q(status=Experiment.STATUS_ACCEPTED) | Q(status=Experiment.STATUS_LIVE)
+    )
+    status_mapping = update_mapping()
+    for experiment in accepted_experiments:
+        try:
```
jaredkerim: We should emit a log msg here saying something like 'Updating Experiment Whatever'

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
         metrics.incr("add_experiment_comment.failed")
         logger.info("Failed to create bugzilla comment")
         raise e
+
+
+@app.task
+@metrics.timer_decorator("update_experiment_status.timing")
+def update_experiment_status():
+    metrics.incr("update_experiment_status.started")
```
jaredkerim: Add a log message here that says something like 'Updating experiment statuses'

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+    accepted_experiments = Experiment.objects.filter(
+        Q(status=Experiment.STATUS_ACCEPTED) | Q(status=Experiment.STATUS_LIVE)
+    )
+    status_mapping = update_mapping()
```
jaredkerim: I don't think this needs to be a function call, I think you can just make this a constant at the top of the file?

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+    for experiment in accepted_experiments:
+        try:
+            recipe_data = normandy.get_recipe(experiment.normandy_id)
+            if needs_to_be_updated(recipe_data["enabled"], experiment.status):
```
jaredkerim: I like this check pattern 👍

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+                approver, _ = get_user_model().objects.get_or_create(
+                    email=approver_email
+                )
+
+                experiment.status = status_mapping[experiment.status]
+                experiment.save()
+
+                experiment.changes.create(
+                    changed_by=approver,
+                    old_status=experiment.status,
```
jaredkerim: You update the status on line 137 so I think this will be the new_status?

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+                    new_status=status_mapping[experiment.status],
```
jaredkerim: Same thing here, won't experiment.status be the new status?

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+                )
+                metrics.incr("update_experiment_status.updated")
```
jaredkerim: Log msg here saying 'Finished updating Experiment Whatever'

app/experimenter/experiments/tasks.py
```diff
@@ -110,3 +112,56 @@ def add_experiment_comment_task(user_id, experiment_id):
...
+            if needs_to_be_updated(recipe_data["enabled"], experiment.status):
+
+                approver_email = recipe_data["approval_request"]["approver"][
+                    "email"
+                ]
+                approver, _ = get_user_model().objects.get_or_create(
+                    email=approver_email
+                )
+
+                experiment.status = status_mapping[experiment.status]
```
jaredkerim: I guess we should start getting into the habit of using `transaction.atomic()` in places where we do multiple updates:

https://docs.djangoproject.com/en/2.2/topics/db/transactions/#django.db.transaction.atomic

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,75 @@ def test_internal_only_does_not_add_comment(self):
...
+):
+
+    def setUp(self):
+        super().setUp()
+
+        self.experiment = ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_ACCEPTED, normandy_id=1234
+        )
+
+    def test_successful_experiment_status_update(self):
```
jaredkerim: I think you need a couple extra test cases here, something like

    test_accepted_experiment_becomes_live_if_normandy_enabled
    test_accepted_experiment_stays_accepted_if_normandy_disabled
    test_live_experiment_statys_live_if_normandy_enabled
    test_live_experiment_becomes_complete_if_normandy_disabled

that sets up a mock experiment in each state and the mock normandy response and checks the status at the end.  

Also probably check that the right changelog is created?

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,75 @@ def test_internal_only_does_not_add_comment(self):
...
+            )
+            self.assertTrue(
+                mm.has_record(
+                    markus.INCR,
+                    "experiments.tasks.update_experiment_status.completed",
+                    value=1,
+                )
+            )
+
+    def test_experiement_status_update_with_failure(self):
```
jaredkerim: It would probably be good if you made two mock experiments and force the normandy mocker to fail on one of them but not the other so we make sure if one fails the other will get updated.

You can set this up by making a little function that checks the id passed into the mock normandy call and if it's equal to one of the experiments, return a valid normandy response and if it's the other, throw the exception.

app/experimenter/settings.py
jaredkerim: Yeah I think 5 minutes is probably fine?

docker-compose.yml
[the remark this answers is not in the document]
jaredkerim: @mythmon This is our first scheduled task so there hasn't been a beat process up to this point.  And actually this was blocked on our move to GPC/k8s becuase our AWS setup didn't allow for 'exactly 1' container instances whereas k8s does, and celery beat processes are finicky and expect there to be exactly one instance.  So adding a beat container to the production pods will be part of the deploy of this change, but will happen after this is merged and tagged for deployment, so I'm not sure it makes sense to block approving this PR until that step is done?

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,75 @@ def test_internal_only_does_not_add_comment(self):
...
+
+    def setUp(self):
+        super().setUp()
+
+        self.experiment = ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_ACCEPTED, normandy_id=1234
+        )
+
+    def test_successful_experiment_status_update(self):
+        with MetricsMock() as mm:
```
jaredkerim: I think we should actually take out the statsd assertions from the tests altogether.  In general in the past I've avoided testing them, it's a huge amount of verbosity and boilerplate in the tests for very little value.  I might file a ticket to remove all the statsd test assertions.

app/experimenter/experiments/tests/test_normandy.py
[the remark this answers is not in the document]
tiftran: @mythmon Not completely done with PR changes... but realized that Normandy Errors aren't raised with strings like "Error calling Normandy API" and they're only logged, so I modified tests to check for those log outputs. It isn't exactly what you asked, but was wondering if it's okay..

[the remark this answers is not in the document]
mythmon: The test you wrote will work, and it makes sure the right error case is being handled. On the other hand, it is kind of awkward. The error thrown by a method should tell you what broke, instead of having to rely on logging. Many times you'll only have access to the errors and not the logs (such as in Sentry).

Because of this, I suggest changing the way you handle errors to make them easier to distinguish by the code catching the exception. One way you could do that would be to add the error strings to the exceptions you're throwing. Another way would be to use entirely separate classes for each type of error. This makes it much clearer what is failing. You can see [an example of this pattern in Normandy](https://github.com/mozilla/normandy/blob/028960f99367e069e6bc4ca53fcc13c6393f19d2/normandy/recipes/signing.py#L157-L170).

docker-compose.yml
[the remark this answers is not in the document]
mythmon: I was mostly curious if the necessary packages were installed and if it docker container(s) had all the needed parts. I guess that's probably all set up, if this docker-compose entry works, but I didn't see anything like that in this PR so I wanted to double check.

app/experimenter/experiments/tests/test_tasks.py
mythmon: Out of curiosity, could you think of a better way to test this? I know that making metrics easier to test is one of @willkg's goals with Marcus. I wonder if the library could do a better job helping here?

At least for me, metrics are very important because if the metrics break, we can't tell what's going on in prod. To me that makes the tests pretty valuable, but I agree they're pretty verbose. I wonder if there is some middleground where we could still assert the right metrics are generated, but is less verbose than this?

docker-compose.yml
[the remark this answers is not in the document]
mythmon: I was mostly curious if any needed packages are already installed and set up (as much as is needed) for use in Docker deployments. I didn't see any changes to Python packages or Dockerfiles in this PR.

app/experimenter/experiments/tasks.py
```diff
@@ -126,19 +126,18 @@ def update_experiment_status():
         try:
             recipe_data = normandy.get_recipe(experiment.normandy_id)
             if needs_to_be_updated(recipe_data["enabled"], experiment.status):
-
-                approver_email = recipe_data["approval_request"]["approver"][
+                creator_email = recipe_data["enabled_states"][0]["creator"][
```
mythmon: A small thing, maybe "enabler" instead of "creator"?

app/experimenter/experiments/tests/test_tasks.py
willkg: I test the metrics I have monitoring alerts tied to and the ones that I use to diagnose problems. The tests verify the keys and values.

Maybe you could compare the list of metrics generated against an expected list:

```
expected = [
    (markus.INCR, "experiments.tasks.update_experiment_status.started", 1, []),
    (markus.INCR, "experiments.tasks.update_experiment_status.updated", 1, []),
    (markus.INCR, "experiments.tasks.update_experiment_status.completed", 1, []),
    WHATEVER
]

self.assertEqual(mm.get_records(), expected)
```

Are they generated in the same order? What's the missing one?

docker-compose.yml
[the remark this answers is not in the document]
jaredkerim: Yeah exactly all the necessary packages exist, it just requires another entry point command, so here it shows up as a compose entry and in production it'll be another pod with the same command.  But the due diligence is appreciated :)

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,88 @@ def test_internal_only_does_not_add_comment(self):
...
+
+
+class TestUpdateExperimentStatus(
+    MockRequestMixin, MockNormandyMixin, TestCase
+):
+
+    def setUp(self):
+        super().setUp()
+
+        self.experiment = ExperimentFactory.create_with_status(
```
jaredkerim: I wouldn't even put this here, just build the experiment in each test.

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,88 @@ def test_internal_only_does_not_add_comment(self):
...
+            ).exists()
+        )
+
+    def test_accepted_experiment_stays_accepted_if_normandy_disabled(self):
+
+        self.mock_normandy_requests_get.return_value = (
+            self.buildMockSuccessDisabledResponse()
+        )
+        experiment = Experiment.objects.get(normandy_id=1234)
+        tasks.update_experiment_status()
```
jaredkerim: Swap 406 and 407 around and then also assert that its status remains accepted.

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,88 @@ def test_internal_only_does_not_add_comment(self):
...
+                old_status=Experiment.STATUS_ACCEPTED,
+                new_status=Experiment.STATUS_LIVE,
+            ).exists()
+        )
+
+    def test_live_experiment_stays_live_if_normandy_enabled(self):
+        experiment = Experiment.objects.get(normandy_id=1234)
+        experiment.status = Experiment.STATUS_LIVE
+        tasks.update_experiment_status()
+        self.assertEqual(experiment.status, Experiment.STATUS_LIVE)
```
jaredkerim: For this assertion to work you'd have to look up the experiment from the db, because even if the task updated the status of the experiment, the instance you're checking against is the in memory one you got on line 416 and in memory instances don't automatically update to reflect the db.

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,88 @@ def test_internal_only_does_not_add_comment(self):
...
+    def test_live_experiment_becomes_complete_if_normandy_disabled(self):
+
+        self.mock_normandy_requests_get.return_value = (
+            self.buildMockSuccessDisabledResponse()
+        )
+        experiment = Experiment.objects.get(normandy_id=1234)
+        experiment.status = Experiment.STATUS_LIVE
+        experiment.save()
+        tasks.update_experiment_status()
+        self.assertTrue(
```
jaredkerim: Same thing as above, after you run the task you'll want to look up the experiment and check that its status is the one you expect it to be.

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,88 @@ def test_internal_only_does_not_add_comment(self):
...
+            experiment.changes.filter(
+                old_status=Experiment.STATUS_LIVE,
+                new_status=Experiment.STATUS_COMPLETE,
+            ).exists()
+        )
+
+    def test_one_failure_does_not_affect_other_experiment_status_updates(self):
+        self.setUpMockNormandyFailWhenIdIs1234()
```
jaredkerim: Yeah this would be clearer if you built both experiments in here, set there ids explicitly, and explicitly pass the one you want to fail to the setup mock method.

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,88 @@ def test_internal_only_does_not_add_comment(self):
...
+        self.experiment2 = ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_ACCEPTED, normandy_id=1235
+        )
+
+        tasks.update_experiment_status()
+        self.assertFalse(
```
jaredkerim: And then down here check that the one that was skipped hasn't changed its status, and the other one has.

app/experimenter/experiments/tests/mixins.py
```diff
@@ -5,6 +5,73 @@
...
+        }
+        mock_response = mock.Mock()
+        mock_response.json = mock.Mock()
+        mock_response.json.return_value = mock_response_data
+        mock_response.raise_for_status = mock.Mock()
+        mock_response.raise_for_status.side_effect = None
+        mock_response.status_code = 200
+        return mock_response
+
+    def setUpMockNormandyFailWithSpecifiedID(self, id):
```
jaredkerim: Tiny nit: `id` is a reserved word in python (it returns a unique identifier for an in memory object)

https://docs.python.org/3/library/functions.html#id

tiftran: okie doke, will change to normandy_id?

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,104 @@ def test_internal_only_does_not_add_comment(self):
...
         self.assertEqual(Notification.objects.count(), 0)
+
+
+class TestUpdateExperimentStatus(
+    MockRequestMixin, MockNormandyMixin, TestCase
+):
+
+    def test_experiment_without_normandy_id(self):
+        ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_ACCEPTED
```
jaredkerim: It would be nice if we explicitly set the `normandy_id = None` here

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,104 @@ def test_internal_only_does_not_add_comment(self):
...
+    def test_accepted_experiment_becomes_live_if_normandy_enabled(self):
+        ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_ACCEPTED, normandy_id=1234
+        )
+        tasks.update_experiment_status()
+        experiment = Experiment.objects.get(normandy_id=1234)
+        self.assertEqual(experiment.status, Experiment.STATUS_LIVE)
+        self.assertTrue(
+            experiment.changes.filter(
+                old_status=Experiment.STATUS_ACCEPTED,
```
jaredkerim: We should also check that the `changed_by.email` matches the email in the normandy data

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +374,104 @@ def test_internal_only_does_not_add_comment(self):
...
+                new_status=Experiment.STATUS_LIVE,
+            ).exists()
+        )
+        self.assertTrue(
+            updated_experiment2.changes.filter(
+                old_status=Experiment.STATUS_ACCEPTED,
+                new_status=Experiment.STATUS_LIVE,
+            ).exists()
+        )
```
jaredkerim: Also it occurs to me we should add a test case where there's an unrecognized email in the normandy data and a user is created with that email

app/experimenter/experiments/tests/test_tasks.py
jaredkerim: Still gotta add this check

app/experimenter/experiments/tests/test_tasks.py
jaredkerim: Still gotta add this check

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -373,3 +379,147 @@ def test_internal_only_does_not_add_comment(self):
...
+        self.assertEqual(experiment.status, Experiment.STATUS_LIVE)
+        self.assertTrue(
+            experiment.changes.filter(
+                changed_by=user,
+                old_status=Experiment.STATUS_ACCEPTED,
+                new_status=Experiment.STATUS_LIVE,
+            ).exists()
+        )
+
+    def test_experiment_status_updates_by_new_user(self):
```
jaredkerim: 👍

PR #1302, opened 2019-05-24
changes so form validations checks against settings bugz.host fixes #…

tiftran:
…1301

Screen Shot of current changes...
Screen Shot 2019-05-28 at 1 54 13 PM

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -49,7 +49,7 @@ class BugzillaURLField(forms.URLField):
         if cleaned_value:
             if settings.BUGZILLA_HOST not in cleaned_value:
                 raise forms.ValidationError(
-                    "Please provide a valid Bugzilla URL"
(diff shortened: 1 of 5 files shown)


app/experimenter/experiments/forms.py
```diff
@@ -42,13 +42,12 @@ def clean(self, value):
...
-    BUGZILLA_BASE_URL = "https://bugzilla.mozilla.org/"
 
     def clean(self, value):
         cleaned_value = super().clean(value)
 
         if cleaned_value:
-            if self.BUGZILLA_BASE_URL not in cleaned_value:
+            if settings.BUGZILLA_HOST not in cleaned_value:
                 raise forms.ValidationError(
                     "Please provide a valid Bugzilla URL"
```
jaredkerim: We have to include the host in the error message otherwise people on dev/stage/local won't understand why it's not valid.

jaredkerim:
@uhlissuh Can you also give this a look before we merge it?

PR #1307, opened 2019-06-01
Add filter for experiments to which a user is subscribed fixes #1259

### LINKED ISSUE #1259: Filter for subscribed experiments
After #808 lands we need the ability to filter for subscribed experiments so we should

1) Add 'subscribed' to the view FilterSet
2) Which will expose a checkbox on the filters on the list page and then
3) Add an alias to the header for 'N Subscribed Experiments'
4) Add a bell icon on the list and detail pages next to the title of the experiment to indicate whether an experiment is subscribed to by the current user

uhlissuh:
Fixes #1259 

In this PR we add another filter of the checkbox variety to the filter set. We also add a bell next to each experiment name on the list page, and a bell the the title of each experiment in the detail view, indicating the user is subscribed. 

perhaps the color of this bell will be in dispute, so I have screenshots below in two colors... my preferred, the pink, and maybe a more standard blue (although I think using this blue could be confusing because it's often used for a link... and also, there's already a blue pill for 'Complete' experiments). 

Screen Shot 2019-05-31 at 4 01 29 PM

Screen Shot 2019-05-31 at 4 23 16 PM

Screen Shot 2019-05-31 at 5 10 35 PM

__________________________________________

this is an alternate color.. which doesn't look as good to me....
_________________________________________
Screen Shot 2019-05-31 at 4 23 00 PM

Screen Shot 2019-05-31 at 4 22 38 PM

### DIFF
--- a/app/experimenter/experiments/tests/test_views.py
+++ b/app/experimenter/experiments/tests/test_views.py
@@ -18,7 +18,10 @@ from experimenter.experiments.forms import (
 from experimenter.experiments.forms import NormandyIdForm
 from experimenter.experiments.models import Experiment
 from experimenter.experiments.tests.factories import ExperimentFactory
-from experimenter.experiments.tests.mixins import MockTasksMixin
+from experimenter.experiments.tests.mixins import (
+    MockTasksMixin,
+    MockRequestMixin,
+)
 from experimenter.openidc.tests.factories import UserFactory
 from experimenter.projects.tests.factories import ProjectFactory
 from experimenter.experiments.views import (
@@ -49,7 +52,7 @@ class TestExperimentFiltersetForm(TestCase):
         )
 
 
-class TestExperimentFilterset(TestCase):
+class TestExperimentFilterset(MockRequestMixin, TestCase):
 
     def test_filters_out_archived_by_default(self):
         for i in range(3):
@@ -285,6 +288,21 @@ class TestExperimentFilterset(TestCase):
         )
 
         self.assertEqual(set(filter.qs), set([exp_1, exp_2]))
+
+    def test_filters_for_subscribed_experiments(self):
+        exp_1 = ExperimentFactory.create(name="Experiment", slug="experiment")
+        ExperimentFactory.create()
+        ExperimentFactory.create()
+
+        exp_1.subscribers.add(self.user)
+
+        subscribed_filter = ExperimentFilterset(
+            {"subscribed": "on"},
+            request=self.request,
+            queryset=Experiment.objects.all(),
+        )
+
+        self.assertEqual(list(subscribed_filter.qs), [exp_1])
 
 
 class TestExperimentOrderingForm(TestCase):
--- a/app/experimenter/experiments/views.py
+++ b/app/experimenter/experiments/views.py
@@ -35,6 +35,7 @@ class ExperimentFiltersetForm(forms.ModelForm):
     in_qa = forms.BooleanField(required=False)
     surveys = forms.BooleanField(required=False)
     search = forms.CharField(required=False)
+    subscribed = forms.BooleanField(required=False)
 
     class Meta:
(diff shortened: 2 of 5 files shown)


app/experimenter/experiments/tests/test_views.py
```diff
@@ -286,6 +289,23 @@ def test_filters_experiments_with_surveys(self):
...
         self.assertEqual(set(filter.qs), set([exp_1, exp_2]))
 
+    def test_filters_for_subscribed_experiments(self):
+        exp_1 = ExperimentFactory.create(name="Experiment", slug="experiment")
+        ExperimentFactory.create()
+        ExperimentFactory.create()
+
+        exp_1.subscribers.add(self.user)
+
+        self.assertTrue(self.user in exp_1.subscribers.all())
```
mythmon: I don't think this assertion is useful, since it is mostly testing that the ORM works as documented.

app/experimenter/experiments/tests/test_views.py
```diff
@@ -286,6 +289,23 @@ def test_filters_experiments_with_surveys(self):
...
+
+        filter = ExperimentFilterset(
```
mythmon: It isn't a big deal in this case, but can you not use the name `filter` here, since it shadows a builtin function?

app/experimenter/experiments/tests/test_views.py
```diff
@@ -286,6 +289,23 @@ def test_filters_experiments_with_surveys(self):
...
+            {"subscribed": "on"},
+            request=self.request,
+            queryset=Experiment.objects.all(),
+        )
+
+        self.assertEqual(set(filter.qs), set([exp_1]))
```
mythmon: I don't think you need the set here, since there is only one item. You still need to ask the ORM to actually run the query though, and converting it to a list with `list()` will do that:

```suggestion
        self.assertEqual(list(filter.qs), [exp_1])
```

uhlissuh: Makes sense. I switched it to list like you suggested. I could also call .get() on it and just compare it to exp_1... can't tell if there's something inherently worse in doing that.

app/experimenter/experiments/tests/test_views.py
uhlissuh: done.

app/experimenter/experiments/tests/test_views.py
uhlissuh: removed!

uhlissuh:
@mythmon, Yea I definitely see what you are saying about the color drawing more attention than needed to the bell. I am still feeling like it needs to be _some_ color.  Also, I am in agreement with you that a better placement for the bell likely exists somewhere in space/time. I don't think we are going to pursue moving to an outlined bell on the left and then having that bell be clickable in this PR. I feel like given the scope/desires contained in the issue that this PR addresses, the bell is in a solid-enough location and other team members seem to like the color selection/placement at least for now.  I am feeling like we should move forward with this design and keep a redesign on our minds (or explicitly in an issue). It seems like from your comment you are OK with shipping this as is, knowing it isn't the perfect solution.

app/experimenter/experiments/tests/test_views.py
mythmon: I thought about `.get()` as well, and I didn't suggest it because it introduced a weird asymetry between this case and the case where multiples are returned. Either would work though.

jaredkerim:
@uhlissuh D'oh! I wanted a link in the header that says like 'N Subscribed Experiments' underneath 'N Owned Experiments' but that's okay I filed it as a separate ticket: https://github.com/mozilla/experimenter/issues/1320

But good job on this looks really great 💃

uhlissuh:
@jaredkerim woops!!! I'll do that ticket though. :)

PR #1328, opened 2019-06-07
bugzilla gets updated when experimenter goes to Archive fixes #806

### LINKED ISSUE #806: Bugzilla update when Experimenter goes to Archive
**Bugzilla State** = fixed / won’t fix

tiftran:


app/experimenter/experiments/forms.py
```diff
@@ -1134,6 +1134,13 @@ class Meta:
     def clean_archived(self):
         return not self.instance.archived
 
+    def save(self, *args, **kwargs):
+        experiment = super().save(*args, **kwargs)
+        tasks.update_bug_resolution_task.delay(
```
jaredkerim: We need to update the tests for this to mock out this delay call and make sure it gets called

app/experimenter/experiments/tests/test_bugzilla.py
```diff
@@ -192,6 +193,33 @@ def test_update_bugzilla_addon_experiment(self):
...
+            archived=True,
+        )
+
+        update_bug_resolution(experiment)
+        self.mock_bugzilla_requests_put.assert_called_with(
+            settings.BUGZILLA_UPDATE_URL.format(id=experiment.bugzilla_id),
+            {"status": "RESOLVED", "resolution": "WONTFIX"},
+        )
+
+        experiment.archived = False
```
jaredkerim: This could be split off into its own test I think?

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1584,7 +1584,6 @@ def test_form_flips_archive_bool(self):
             self.request, instance=experiment, data={}
         )
         self.assertTrue(form.is_valid())
```
jaredkerim: Gotta update this test to mock out the delay and check that it's called

PR #1347, opened 2019-06-11
Add celery beat files to gitignore fixes #1346

### LINKED ISSUE #1346: Add celery beat files to gitignore
We need to gitignore the following two files:

        app/celerybeat-schedule
        app/celerybeat.pid

jaredkerim:


jaredkerim:
Actually we should just do this instead https://github.com/mozilla/experimenter/issues/1348

PR #1349, opened 2019-06-11
 move celery beat files into app container fixes #1348 

### LINKED ISSUE #1348: Move celery beat files into the app container
There's no need for those files to be mirrored on the hostfile system, we can just keep them inside the container by specifying paths outside of /app

--pidfile for celerybeat.pid 
-s for the celerybeat-schedule

tiftran:


docker-compose.yml
```diff
@@ -36,7 +36,7 @@ services:
       - redis
     volumes:
       - ./app:/app
-    command: bash -c "/app/bin/wait-for-it.sh db:5432 -- celery -A experimenter beat -l debug"
+    command: bash -c "/app/bin/wait-for-it.sh db:5432 -- celery -A experimenter beat --pidfile ./celerybeat.pid -s ./celerybeat-schedule -l debug"
```
jaredkerim: I think that will put the files in the workdir, which is /app, which will then place them on the host filesystem, which is the opposite of what we want.  Try putting them right at the root path like `/celerybeat.pid` `/celerybeat-schedule`?

tiftran: my bads, did not completely understand... i think my new commit is what is originally asked?

PR #1361, opened 2019-06-13
store changed vals in changelog fixes #464

### LINKED ISSUE #464: Store changed fields on changelog
When someone edits an experiment, we can look at the 'changed_data' field to know which fields were edited and add that to the change log.  We can probably add two JSON fields called `old_values` and `new_values` and store a dict of {field: value, ...} in each.

tiftran:


app/experimenter/experiments/tests/test_forms.py
```diff
@@ -202,6 +205,302 @@ class Meta:
...
+            "review_impacted_teams": True,
+        }
+
+        class TestForm(ChangeLogMixin, forms.ModelForm):
+
+            class Meta:
+                model = Experiment
+                fields = "__all__"
+
+        class TinyTestForm(ChangeLogMixin, forms.ModelForm):
```
jaredkerim: Tiny Teacup Dog approves of this Tiny Test Form

Teacup-Dogs-These-will-melt-your-heart

tiftran:
@jaredkerim feeling like a goof... after spending a while today looking at why short description was picked up as a changed field, i found:
Screen Shot 2019-06-24 at 2 17 00 PM

where the change is "\n" -> "\r\n". Don't think it's too much of a problem since changes made via experiment does not have this problem and this only occurs when our dummy experiments are generated.

app/experimenter/experiments/serializers.py
```diff
@@ -49,6 +49,89 @@ class Meta:
...
 
 
+class ChangeLogSerializer(serializers.ModelSerializer):
+    variants = ExperimentVariantSerializer(many=True, required=False)
+    locales = LocaleSerializer(many=True, required=False)
+    countries = CountrySerializer(many=True, required=False)
+
+    class Meta:
+        model = Experiment
+        fields = (
```
jaredkerim: Experiment: the model with too many fields -_- I should've set this up differently, now it's gonna be a mess when we have to go in and split things up. *jared sticks a TECH DEBT sticker on his forehead*

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -202,6 +205,302 @@ class Meta:
         self.assertEqual(change.old_status, old_status)
         self.assertEqual(change.new_status, new_status)
 
+    def test_mixin_sets_old_values_and_new_values(self):
+        experiment = Experiment()
```
jaredkerim: In this test you test the case of an empty experiment and an all fields form that sets all the fields and checks new values, and then a populated experiment with a single field form that checks the old and new values for that one field.  Could these be separate tests? What about the case of old_values for an all fields form on a fully populated experiment?

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -202,6 +205,302 @@ class Meta:
...
+        self.assertEqual(
+            experiment.changes.latest().old_values,
+            {"review_impacted_teams": True},
+        )
+        self.assertEqual(
+            experiment.changes.latest().new_values,
+            {"review_impacted_teams": False},
+        )
+
+    def test_mixin_sets_old_values_and_new_values_with_m2m(self):
```
jaredkerim: The name suggests this is testing old and new values for m2m fields, but down in the assertions it looks like it's only testing new values for the m2m fields, doesn't seem to match the name?

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -202,6 +205,302 @@ class Meta:
...
+                    "ratio": 50,
+                    "value": "8",
+                    "is_control": True,
+                    "description": " old branch 1 desc",
+                }
+            ],
+            latest_changes.old_values["variants"],
+        )
+
+        self.assertEquals(len(latest_changes.new_values["countries"]), 0)
```
jaredkerim: Ah I see, this tests wiping out the countries/locales, but should we be testing the case where we start with some set countries/locales, and wind up with a modified set of locales/countries (ie adding some, removing some)?

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -98,6 +99,28 @@ def test_serializer_outputs_expected_schema(self):
         )
 
 
+class TestChangeLogSerializer(TestCase):
+
+    def test_serializer_outputs_expected_schema(self):
+        experiment = ExperimentFactory.create()
+        serializer = ChangeLogSerializer(experiment)
+
+        self.assertEqual(serializer.data.get("type"), experiment.type)
```
jaredkerim: Why test only this particular set of fields?

app/experimenter/experiments/forms.py
```diff
@@ -63,29 +64,75 @@ class ChangeLogMixin(object):
...
+        if self.changed_data:
+            if latest_change:
+                old_status = latest_change.new_status
+
+                for field in self.changed_data:
+                    if field in self.old_values:
+                        old_values[field] = self.old_values[field]
+
+            else:
+                prev_values = {field: None for field in self.changed_data}
```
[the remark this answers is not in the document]
tiftran: In the event that there are no "latest changes" this would occur when the experiment is first created. want to have old_values in the changelog have values of None, while the new_values is the value that is initially set. Not entirely sure if this is necessary looking at it again...

app/experimenter/experiments/tests/test_forms.py
tiftran: will do

app/experimenter/experiments/forms.py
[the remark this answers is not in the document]
jaredkerim: It makes more sense now, thnx :)

app/experimenter/experiments/forms.py
```diff
@@ -63,29 +64,68 @@ class ChangeLogMixin(object):
...
         experiment = super().save(*args, **kwargs)
 
+        old_values = {}
+        new_values = {}
         old_status = None
 
+        self.new_serialized_vals = ChangeLogSerializer(self.instance).data
         latest_change = experiment.changes.latest()
+
+        # account for changes in variant values
```
jaredkerim: Awesome I find this so much easier to read, thank you! 🤓

tiftran:
Current changes suggested via slack by Jared

i think we can just scrap the fields="__all__" testing cases altogether

1) we don't have any forms that have all fields
2) the way the change capturing is implemented, it also doesn't care about the difference between all fields or some fields
and 3) writing out the list of all fields in each of those tests is super fragile and we'll end up having to change them anytime we add/remove fields, which feels tedious and unnecessary

so i think we can get away with soemthing like two tests
1) new experiment with no previous changes
2) existing experiment with previous changes
and for each of those just choose 5 random fields for the dummy form, and only modify 3
including variants in both

tiftran:
So... this took longer than expected...
Reduced test to a smaller number of fields. Was unable to reduce to two tests with that each included variants... as variants and experiments are two diff models... and I wasn't able to use only one form?

I've also included an non empty variants value for the serializer test

Currently the change log looks like...
Screen Shot 2019-07-16 at 12 40 35 PM

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -202,6 +205,320 @@ class Meta:
...
+        form = TinyTestForm(
+            request=self.request, data=data, instance=experiment
+        )
+
+        self.assertTrue(form.is_valid())
+        form.save()
+        self.assertEqual(experiment.changes.count(), 1)
+        self.assertEqual(experiment.changes.latest().new_values, data)
+
+    def test_mixin_changelog_values_with_prev_log(self):
```
tiftran: @jaredkerim  just want to make sure i understood correctly, you want me to scrap this and keep 
https://github.com/mozilla/experimenter/pull/1361/files#diff-3d5bec13bbcacb975a33ef9256fa8a4eR302?

PR #1368, opened 2019-06-14
Notify the user for failed bugzilla tasks fixes #1282

### LINKED ISSUE #1282: Ensure all failed bugzilla calls notify the user
I'm still able to create situations locally where a bugzilla call fails but I do not receive a notification, and no stack trace is sent to sentry.  We should figure out how to replicate those cases and ensure that any failed bugzilla call always results in a user notification and a sentry record.

I'll figure out how to recreate it and add instructions here.

jaredkerim:


jaredkerim:
@uhlissuh @tiftran This PR wants to be your friend

PR #1381, opened 2019-06-21
add minimum and maximum firefox versions fixes #880

### LINKED ISSUE #880: Add 'Max Firefox Version'
We should make it so the current version field is a minimum, and add an optional maximum version.

### LINKED ISSUE #1295: Expose the "min-max" version in the listing blurb
Once min and max version are available in population filtering - move that information onto the listing summary.  right now it says "Firefox 67.0".  Ideally it says "Firefox 67.0-68.0".

Use case:  There is a different release manager for each Firefox release.  It is hard to tell if Live experiments are going to keep runing into your coming release.  Release management searching experiments can look more easily at the list to see what will keep running in the specific release they manage.

image

uhlissuh:
fixes #880
fixes #1295 

Here we're changing Firefox Version to two versions... a min and max. 

We've changed the version filter to return any experiment that includes the selected version in its range.

Screen Shot 2019-06-19 at 11 35 53 AM

Screen Shot 2019-06-19 at 11 35 16 AM

Screen Shot 2019-06-19 at 11 56 39 AM


app/experimenter/experiments/constants.py
```diff
@@ -76,6 +75,14 @@ class ExperimentConstants(object):
         ("70.0", "Firefox 70.0"),
```
jaredkerim: I guess we might want to tack a few more versions on there while we're doing this since we're about to hit 70.  Maybe do up to 80 and then we don't need to think about it again for a little while?

app/experimenter/experiments/models.py
```diff
@@ -330,9 +336,16 @@ def generate_normandy_slug(self):
         if not self.bugzilla_id:
             raise ValueError(error_msg.format(field="Bugzilla ID"))
 
+        if self.firefox_max_version:
+            version_string = (
+                f"{self.firefox_min_version}-{self.firefox_max_version}"
+            )
+        else:
+            version_string = self.firefox_min_version
```
jaredkerim: It's a tiny stylistic thing, but I prefer to write things like this as

    thing = default
    if something:
        thing = special_case

The way you've written it is valid Python but it always irks me to introduce a variable inside an if statement, becuase you could sometimes forget a case and then suddenly wind up referring to a varaible that wasn't declared.  But it's just a preferential thing.

app/experimenter/experiments/tests/factories.py
```diff
@@ -75,7 +75,10 @@ class ExperimentFactory(
     pref_branch = factory.LazyAttribute(
         lambda o: random.choice(Experiment.PREF_BRANCH_CHOICES[1:])[0]
     )
-    firefox_version = factory.LazyAttribute(
+    firefox_min_version = factory.LazyAttribute(
+        lambda o: random.choice(Experiment.VERSION_CHOICES[1:])[0]
+    )
+    firefox_max_version = factory.LazyAttribute(
```
jaredkerim: Actually it would be good if we also threw in a case that this is sometimes randomly empty so we stress test the cases where it's not set.

jaredkerim: I think if you just drop the [1:] slice then this should randomly be None sometimes which should cover it.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -285,14 +285,15 @@ def test_generate_normandy_slug_returns_expected_slug(self):
...
+            firefox_min_version="57.0",
+            firefox_max_version="59.0",
             firefox_channel="Nightly",
             bugzilla_id="12345",
         )
 
         self.assertEqual(
             experiment.generate_normandy_slug(),
-            "pref-experiment-slug-nightly-57.0-bug-12345",
+            "pref-experiment-slug-nightly-57.0-59.0-bug-12345",
```
jaredkerim: I guess maybe we should also test the case where no max is set?

app/experimenter/experiments/views.py
```diff
@@ -225,6 +228,13 @@ def experiment_date_field_filter(self, queryset, name, value):
         # filter that controls which date range we show
         return queryset
 
+    def version_filter(self, queryset, name, value):
+
+        return queryset.filter(
+            Q(firefox_min_version__lte=value, firefox_max_version__gte=value)
```
jaredkerim: 👌

app/experimenter/experiments/constants.py
uhlissuh: sounds good!

app/experimenter/experiments/models.py
uhlissuh: ok I'll change that

app/experimenter/experiments/tests/test_models.py
```diff
@@ -285,14 +285,29 @@ def test_generate_normandy_slug_returns_expected_slug(self):
...
+        )
+
+        experiment = ExperimentFactory.create(
```
jaredkerim: If this test fails is it clear exactly what's not working?

uhlissuh: I guess I see what you're saying... that it could be failing due to other parts of the normandy slug formatting... nothing is coming to mind for an improvement though..

PR #1393, opened 2019-06-25
update bugzilla when exp. is live fixes #804

### LINKED ISSUE #804: Bugzilla update when Experimenter goes to “Live” state
~~**Whiteboard**= overwrite with “Live”~~

~~Set **Status-firefox{version} to FIXED** (see below image).  I checked with relman, this would move it along through their channel.~~

**New Comment** 
Start date: {start date}, 
Experiment end date: {calculated experiment end date}. 

Screen Shot 2019-04-26 at 8 45 52 AM

EDIT: Based on discussions with @shell1 scope of bugzilla behaviour has been reduced

tiftran:


app/experimenter/experiments/tasks.py
```diff
@@ -176,9 +184,13 @@ def update_status(experiment):
                 old_status=old_status,
                 new_status=new_status,
             )
-            metrics.incr("update_experiment_status.updated")
+            metrics.incr("update_experiment_info.updated")
             logger.info("Finished updating Experiment: {}".format(experiment))
 
+        if experiment.status == Experiment.STATUS_LIVE:
+            experiment = Experiment.objects.get(id=experiment.id)
```
tiftran: I think this is kinda awkward, but not sure what to do otherwise to retrieve the updated start and end date

tiftran:
Also thought about somehow manipulating dummy data (with valid normandy ids???) so that it's possible to observe the workflow on stage??? @jaredkerim  @uhlissuh  thoughts?

app/experimenter/experiments/tasks.py
jaredkerim: The `start_date` and `end_date` properties are methods that already do a bunch of database lookups when they're invoked, do you need to reload the instance in memory?

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/models.py#L397-L412

tiftran:
@jaredkerim yes, we scrapped the whiteboard thing

app/experimenter/experiments/tasks.py
tiftran: I mean I could always default the start date to "today" and end date to today+ duration (is that awkward too?). Without reloading, the existing instance would still have the pre existing start and end date prior to the status change.

jaredkerim:
@tiftran Can you update the issue with the feedback we got from @shell1 about simplifying it so that QA knows what behaviour to expect when they test it?

PR #1398, opened 2019-06-26
Add filter for longrunning experiments fixes #1294

### LINKED ISSUE #1294: add filter for "show only long running - 3+ versions"
Once min and max version exist in the Population targeting section - want to add a filter for ""show only long running - 3+ versions".  

60, 61, 62 is an example of long-running.  Use case: QA can easily find which experiments span multiple versions.  They can look at those experiments when they are testing a release and evaluate which ones need additional testing (since Firefox changing may break experiments - long-running add-on experiments are particularly risky).

image

uhlissuh:
Fixes #1294 

This checkbox filters out experiments that span 3 or more Firefox versions.

Screen Shot 2019-06-26 at 12 57 36 PM

### DIFF
--- a/app/experimenter/experiments/views.py
+++ b/app/experimenter/experiments/views.py
@@ -2,8 +2,9 @@ import django_filters as filters
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.db.models import Q, F, FloatField
+from django.db.models import Q, F, IntegerField
 from django.db.models.functions import Cast
+from django.db.models.expressions import Func, Value
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.views.generic import CreateView, DetailView, UpdateView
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/views.py
```diff
@@ -279,6 +288,25 @@ def subscribed_filter(self, queryset, name, value):
...
+                    firefox_min_float=Cast(
+                        "firefox_min_version", FloatField()
+                    ),
+                    firefox_max_float=Cast(
+                        "firefox_max_version", FloatField()
+                    ),
+                    version_count=F("firefox_max_float")
+                    - F("firefox_min_float"),
+                )
+                .filter(version_count__gte=3.0)
```
jaredkerim: This is clever and elegant and works well for the way we have the version values setup (they easily parse into floats), but I wonder what would happen if we added some versions that contain non numerical characters.  It hasn't happened yet but in theory it could.  Doing it this way could introduce a little hidden contact that the version value strings must be parseable into floats otherwise this breaks.  What do you think?

uhlissuh: What non-numerical versions are you thinking might happen?

jaredkerim: Like if we add some special version strings in the future that might look like "62.01a" or include some non numerical characters, or if we add the ability for people to specify custom versions in the future.  Doesn't necessarily mean we have to deal with a potentially non existent future case right now, but just to think about if we do in the future, and if this breaks, will it be clear what went wrong or what the baked in assumptions are?

uhlissuh: Hmm okay, well yes I feel like it would be clear that if you made a significant change to the way we represent versions that it would likely break the filters that have to parse and do math to versions. Is there something you were thinking I could add that would make you feel more confident about this approach? I think if I wanted to build a longrunning filter right now that allowed versions with nonnumerical characters, I'd have to do some raw sql stuff instead of using the ORM methods, and nobody seems to like that idea generally.

jaredkerim: http://andilabs.github.io/django/django-1-8/expressions/func/f/2015/08/29/django-1.8-new-db-models-expressions-F-Func.html

jaredkerim: @uhlissuh Oh the link was here!

uhlissuh: @jaredkerim OMGGGG.

PR #1406, opened 2019-06-27
Add experiment launch email fixes #809

### LINKED ISSUE #809: Launch emails
Send this email when experimenter goes to "Live".

the {link} in the email should go to https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-NormandyChangeWindow

In the example below is a place {type}: {name}, {their experimenter URL}. It would be amazing if it could say "Pref-Flip: Foo experiment name" and that hyperlink right to the experimenter link for that study. I wasn't sure how to ask that. If not, listing "pref-flip: Foo experiment name, Experimenter URL" works too.

To: Owner, anyone subscribed
From: {experimenter bot?}
Subject: Experimenter launch verification for {type}: {name}

Hello,

This is an automatic email from Experimenter about {type}: {name} {their experimenter URL}, which went live on {start date}.  Experimenter updates the proposed start date to the actual start date when the state changes to "Live".  The subsequent scheduled change dates (ex: enrollment pause and experiment end) are recalculated based on the actual start date.

Please watch your scheduled change dates as the data starts being gathered.  If you need scheduling changes (ex: delay enrollment pause, delay end), please SLACK: #ask-experimenter.  If everything looks good, no action is needed.

Thank you!
Experimenter and Normandy Teams

uhlissuh:
Fixes #809 

This PR adds an email that gets sent when an experiment goes live in Normandy.


app/experimenter/experiments/email.py
```diff
@@ -60,3 +60,25 @@ def send_intent_to_ship_email(experiment_id):
         cc=[experiment.owner.email],
     )
     email.send(fail_silently=False)
+
+
+def send_experiment_launch_email(experiment):
+
+    content = render_to_string(
+        "experiments/launch_experiment_email.txt", {"experiment": experiment}
```
jaredkerim: I'm so glad we're doing this text file email template thing, it's such a nice little pattern.  Maybe it would be helpful to put them in a subfolder like `experiments/emails/`?

app/experimenter/experiments/tasks.py
```diff
@@ -175,6 +175,8 @@ def update_status(experiment):
         )
         old_status = experiment.status
         new_status = STATUS_UPDATE_MAPPING[old_status]
+        if new_status == Experiment.STATUS_LIVE:
```
jaredkerim: Kinda feels like this is crammed in the middle of another sentence. 'Okay our old status is this, the new one is this- **SEND THE EMAIL!!!!** ...as I was saying, the new status is this, let's save and add a change log.'  This could be broken up a little, but that's just me being a cranky nitpicker.

On a more grounded note, it would probably be prudent to emit logs when we send the email?

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -405,6 +405,15 @@ def test_accepted_experiment_becomes_live_if_normandy_enabled(self):
...
         )
 
+    def test_launch_email_sent_when_experiment_becomes_live(self):
+        ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_ACCEPTED, normandy_id=1234
+        )
+
+        tasks.update_experiment_status()
+
+        self.assertEqual(len(mail.outbox), 1)
```
jaredkerim: Couldn't you just fold this assertion in the test directly above?  It's testing basically the same thing I think?

app/experimenter/experiments/email.py
uhlissuh: good idea.

app/experimenter/experiments/tasks.py
uhlissuh: good idea! I did this with `logger.info()`.

uhlissuh: Also I moved the statement. Lol.

PR #1408, opened 2019-06-28
added prefetch method fixes #1405

### LINKED ISSUE #1405: Change ExperimentManager get_queryset behaviour, and change views.py s.t details and listview still uses a cached version of experiments
As a result of https://github.com/mozilla/experimenter/pull/1393
The caching behaviour for experiment.changes.all() can't be the default

tiftran:


app/experimenter/experiments/models.py
```diff
@@ -22,6 +22,13 @@
...
     def get_queryset(self):
+        return (
+            super()
+            .get_queryset()
+            .annotate(latest_change=Max("changes__changed_on"))
+        )
+
+    def get_prefetched(self):
         return (
             super()
```
jaredkerim: I think if you omit the super() call here, then you'll call the subclassed get_queryset written above, which means you'll get the annotation for free so you can also omit it in get_prefetched?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -56,6 +56,32 @@ def test_queryset_annotated_with_latest_change(self):
             [experiment1, experiment2],
         )
 
+    def test_get_prefetched_changes(self):
```
jaredkerim: Aaaaaahhhhhh I get what you're doing here, but this is one of those rare times when I'd suggest NOT testing this behaviour.  

The behaviour that when you call 'get prefetched' you get back the same queryset instead of a new one is completely built into django and therefore tested by its tests, so you don't really need to test it.

What the method is really trying to accomplish is that it makes fewer database calls when you look up all the related fields, so if this tested anything it would be something like getting an experiment, then looking up a bunch of related stuff, and checking that it doens't make additional hits to the database, but even then I'd say don't write such a test.  

I've seen it done in the past and it winds up being a test that you have to change so often that it provides very little value, so in general I think this falls in the same camp as testing HTML in django templates: just don't do it.

PR #1413, opened 2019-06-28
disabled archive button when exp. is live fixes #1389

### LINKED ISSUE #1389: Disable ability to archive experiment when state is live
When archiving an experiment, the bugzilla ticket associated to it would change to closed/won't fix, which is something we probably don't want when an experiment is live

tiftran:


jaredkerim:
A couple things stand out to me about this:

- What happens if someone had the page open before the experiment is live, the button is enabled, then the experiment goes live, and without refreshing the page they click the button?

- This leads into a broader idea about 'trusting the client' vs 'not trusting the client' (client in this case means the browser).  If we really don't want live experiments to become archived because it'll actually cause bugs/crashes/inconsistent states, should we be doing this check somewhere else, or somewhere in addition to in the client? 

- If we hard code this check of `status == live` in the template and maybe end up checking it somewhere else at some point, and then later we change the definition of when an experiment can be archived, will we have to modify all those checks in every place that they're used?

tiftran:
@jaredkerim  @uhlissuh  hmmm would this sound better? 
allowing the user to click the archive button and have the archive form check for the experiment status and if not live archive, else create a notification...

jaredkerim:
@tiftran Well we also don't want the user to think they can archive if we know they can't when the page is rendered, the UI should guide the user to only correct/valid interactions.  You just don't want to rely solely on the UI to keep the system in a sane state, that's the backend's job.

tiftran:
So took in all the feedback... and here's my new take on this....
The button being disabled is still there.

To address the case where "had the page open before the experiment is live, the button is enabled, then the experiment goes live, and without refreshing the page they click the button" I thought of having a check to see it's archivable before it's saved. 

Also removed the hard coded check for status in the template to a property method in the experiment model as ```is_archivable```

uhlissuh:
good solution!

PR #1428, opened 2019-07-03
Add check for existing name fixes #1410

### LINKED ISSUE #1410: Cloning an experiment wrongly validates the previous name of the cloned experiment
**[Affected versions]:**
- Firefox Release 67.0.4 and above
- Chrome 75 and above.

**[Affected Platforms]:** 	 	
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- You are logged in your LDAP account on the "https://stage.experimenter.nonprod.dataops.mozgcp.net/" website.
- Have an experiment created on Experimenter stage.

**[Steps to reproduce]:**
2. Open the experiment from prerequisites.
3. Click the "Edit" button from the "Overview" section.
4. Change the name of the experiment from the "Name" section.
5. Click the "Save Draft" button.
6. Click the "Clone this Experiment" button and delete the "V2" string from the prompt displayed.
7. Click the "Clone" button and observe the behavior.

**[Expected result]:**
- The "This experiment name already exists." error message is displayed.

**[Actual result]:**
- No error message is displayed and the "Clone" button, remains in a "Cloning" state.

**[Notes]:**
- The "Clone" button remains in a "Cloning" state until the page is refreshed.
- If the previous name of the cloned experiment is inserted in the "Name of New Experiment:" input field and cloned, results in getting the "This experiment name already exists." error message.
Name

uhlissuh:
Fixes #1410 

Instead of just finding experiments in the database that have the same slug (in order to prevent cloning if we find them), I modified the lookup to be a Q statement so that we can check if the slug or the name exists. 

The problem stemmed from the fact that when you modify the name of an experiment, we keep the old slug so people's links don't break. But that means that an experiment can be a state with the name "cat" and the slug "dog". If we then try to clone an experiment in this state, and we enter in the name "cat" to be the clone's name, we can't just say are there any existing experiments with slug "cat" because there won't be the cat slug (the parent cat name experiment has the dog slug). We have to check if the slug exists and if the name exists.

### DIFF
--- a/app/experimenter/experiments/tests/test_serializers.py
+++ b/app/experimenter/experiments/tests/test_serializers.py
@@ -372,11 +372,22 @@ class TestExperimentRecipeSerializer(TestCase):
 
 class TestCloneSerializer(MockRequestMixin, TestCase):
 
-    def test_clone_serializer_rejects_duplicate_name(self):
-        experiment = ExperimentFactory.create(
-            name="great experiment", slug="great-experiment"
+    def test_clone_serializer_rejects_duplicate_slug(self):
+        experiment_1 = ExperimentFactory.create(
+            name="good experiment", slug="great-experiment"
         )
         clone_data = {"name": "great experiment"}
+        serializer = ExperimentCloneSerializer(
+            instance=experiment_1, data=clone_data
+        )
+
+        self.assertFalse(serializer.is_valid())
+
+    def test_clone_serializer_rejects_duplicate_name(self):
+        experiment = ExperimentFactory.create(
+            name="wonderful experiment", slug="amazing-experiment"
+        )
+        clone_data = {"name": "wonderful experiment"}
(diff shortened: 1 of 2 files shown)


jaredkerim:
Teeeesssstttt for this?

giphy

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -373,12 +373,22 @@ def test_serializer_excludes_countries_if_none_set(self):
 class TestCloneSerializer(MockRequestMixin, TestCase):
 
     def test_clone_serializer_rejects_duplicate_name(self):
```
jaredkerim: The test is called 'rejects duplicate name', but it's testing two things in one test, duplicate names and duplicate slugs.  Should those be two separate tests......?

uhlissuh: i guess

PR #1434, opened 2019-07-05
added max-height for list exp. boxes and fade fixes #1212

### LINKED ISSUE #1212: Truncate descriptions on list page
Right now users can enter fairly long descriptions in the 'short description' field which can cause the list page entries to be very large.  Maybe we should truncate them on the list page so that each list item has a maximum height so you can see more on a single screen?

Add a "..." to end of any that are truncated to indicate that it continues.

tiftran:
Screen Shot 2019-07-05 at 10 40 22 AM

### DIFF
--- a/app/experimenter/static/css/experimenter.css
+++ b/app/experimenter/static/css/experimenter.css
@@ -186,3 +186,18 @@ pre.json {
   line-height: 2em;
   border: none !important;
 }
+
+.list-summary:before {
+  content:'';
+  width:100%;
+  height:100%;    
+  position:absolute;
+  left:0;
+  top:0;
+  background:linear-gradient(transparent 12em, white);
+}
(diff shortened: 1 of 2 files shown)


app/experimenter/static/css/experimenter.css
```diff
@@ -186,3 +186,19 @@ pre.json {
...
+  left:0;
+  top:0;
+  background:linear-gradient(transparent 12em, white);
+}
+
+.list-summary{
+  overflow: hidden;
+  max-height: 15em;
+  
+}
```
uhlissuh: this is a total nit but should there not be an extra blank line here?

tiftran: you're right. lemme fix that....

PR #1444, opened 2019-07-08
changes to normandy parsing to check for null vals fixes #1442

### LINKED ISSUE #1442: Celery Beat update task still running into Key Error and Type Errors

tiftran:


app/experimenter/settings.py
```diff
@@ -28,6 +28,8 @@
 
 DEV_USER_EMAIL = "dev@example.com"
 
+NORMANDY_DEFAULT_CHANGELOG_USER = "normandy@admin.com"
```
mythmon: As a drive by comment, this shouldn't be `@admin.com`. Admin.com is a real site that sells a book. We shouldn't be using their email addresses. This should either be something under `mozilla.com` (maybe `unknown-user@normandy.mozilla.com`?) an `@example.com` address. Example.com [is specially reserved](https://tools.ietf.org/html/rfc2606#section-3) for example data

tiftran: will do

jaredkerim: @mythmon Is there some actual email we can use for this that goes to a normandy mailing list or something?

mythmon: For some history: I'm pretty sure that these missing email addresses are due to [a bug we had a while ago](https://github.com/mozilla/normandy/issues/1460) that prevented us from storing the creator of revisions. Instead we stored null. We fixed this in August 2019. I don't expect any future revisions of recipes to have null creators (barring future bugs).

If we need an email that actually goes somewhere, we could use `normandy-notifications@mozilla.com` we use that for some security notifications (we should make sure to transfer that to you, @jaredkerim). However, I'd lean towards something like "unknown-user@", since this is an error case I don't expect to re-occur.

Perhaps we could also explore filling in this data in the Normandy database, since all revisions *should* have creators. Like I mentioned in the issue above, we should be able to figure out who the proper author is and do some database surgery to add it back in.

tiftran: Given the above, I'll leave the email as is, since this seems unlikely to happen in production or in a full workflow. The null value email occurred in stage where most likely it was recycling and used a pre-existing recipe

PR #1450, opened 2019-07-09
Insert launch and ship emails into new email DB table fixes #1437

### LINKED ISSUE #1437: Record launch emails and intent to ship emails
we made a new database table that records what emails have been sent out for a given experiment. we should record launch and intent to ship in that table also.

uhlissuh:
Since we added a table to keep track of ending emails sent, I went back and inserted into the table when other emails were sent also. Now we'll be able to keep track of all the emails we send users!


uhlissuh:
@jaredkerim Yes I can see that one day being useful!

PR #1451, opened 2019-07-10
added exp. paused val and periodic normandy check fixes #1318

### LINKED ISSUE #1318: Read pause value from normandy API
blocked on #158

tiftran:


jaredkerim:
@tiftran I'm pretty sure Shell wants to know the *date* it was paused on?

app/experimenter/experiments/tasks.py
```diff
@@ -198,6 +198,10 @@ def update_status(experiment):
         if experiment.status == Experiment.STATUS_COMPLETE:
             bugzilla.update_bug_resolution(experiment)
 
+    if recipe_data and (is_paused(recipe_data) != experiment.paused):
+        experiment.paused = not experiment.paused
```
jaredkerim: Instead of flipping it here, why not just get whatever value is returned by the API and store it directly?  Why add the complication of making it dependent on the relationship between experimenters state and Normandy's state? What if they become inconsistent?

app/experimenter/experiments/models.py
```diff
@@ -247,6 +247,8 @@ class Experiment(ExperimentConstants, models.Model):
         default=None, blank=True, null=True
     )
 
+    paused = models.BooleanField(default=False)
```
jaredkerim: I think in general it's a good practice to label boolean fields like this as `is_whatever`

tiftran: will do

app/experimenter/experiments/tasks.py
tiftran: did not think the case where they become inconsistent. will change to store whatever API gives

uhlissuh:
Yeah, wondering in relation to @jaredkerim's comment... could we update the enrollment end date in a similar way to how we update the start and end date property? use the proposed enrollment until we know the actual date it occurred? Not sure exactly what confusion that would introduce.

tiftran:
@uhlissuh enrollment end date? did you mean enrollment pause date? if so... talked to Shell and the scope of the ticket has been reduced to just knowing the state of the experiment being paused or not paused, created another ticket for the date and labelled it as p5?

app/experimenter/experiments/tasks.py
```diff
@@ -198,6 +198,11 @@ def update_status(experiment):
         if experiment.status == Experiment.STATUS_COMPLETE:
             bugzilla.update_bug_resolution(experiment)
 
+    if recipe_data and (is_paused(recipe_data) != experiment.is_paused):
+        with transaction.atomic():
```
tiftran: Not exactly sure why this is needed, I'm assuming it has to do with the transaction.atomic() above, but without it I get a django.db.transaction.TransactionManagementError from time to time

tiftran:
@jaredkerim  @uhlissuh  thought just occurred would it make sense if we have Paused as an experiment status instead of its own field?

uhlissuh:
Talked direction to Tif about this on slack, but I think that's a good idea... adding it to the changelog... then we could have it in the history section... maybe?

tiftran:
thought about it some more.... in the event that "pause" is a status and the celery beat process/instance goes down and unnoticed or the patch for it is deployed on the next day. Wouldn't this change the accuracy of the changelog? This would probably extend to the start and end date as well... Not sure if this is too much of a concern....

tiftran:
after discussing with @jaredkerim, leaving is_paused as an extra bool field

app/experimenter/experiments/tasks.py
jaredkerim: Huh interesting, it should complete the transaction when the with block ends...

app/experimenter/experiments/tasks.py
```diff
@@ -198,6 +198,11 @@ def update_status(experiment):
...
+            experiment.is_paused = is_paused(recipe_data)
```
jaredkerim: You don't need to call this twice, just pull it out into a variable.

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -332,6 +332,62 @@ def test_experiment_with_no_creator_in_recipe(self):
             ).exists()
         )
 
+    def test_experiment_with_pause_true(self):
```
jaredkerim: This happens to me all the time, there's some polynomial combination of states and if you test every single one of them you end up with this huge battery of nearly identical tests.  Try to cut this down to two test cases.  Think about it from what cases the code cares about rather than the possible state combinations.

app/experimenter/experiments/tasks.py
jaredkerim: Well at any rate in theory we should be using that decorator more so I guess this is good?

app/experimenter/experiments/tests/test_tasks.py
tiftran: Like Case 1 where state stays the same and Case 2 where it changes?

jaredkerim: BINGO

PR #1458, opened 2019-07-12
Remove release drivers from launch and ending emails fixes #1457

### LINKED ISSUE #1457: Remove release drivers from launch and ending emails
we are sending status emails to release drivers which isn't right AT ALL.

uhlissuh:
instead of sending to release drivers and cc-ing the recipients we're interested in, owner and subscribed, we are just directly sending to owner and subscribed, removing release drivers entirely.


uhlissuh:
I decided to remove the recipients test to the email tests directly instead of having the check on the task test... it seems more sensical.

PR #1470, opened 2019-07-15
Whitelist recipe API fixes #1456

### LINKED ISSUE #1456: Add experiment recipe API to auth whitelist
We need to whitelist the experiment recipe API so that delivery console can read it without authentication.

jaredkerim:


uhlissuh:
as far as I can tell... 👍

PR #1472, opened 2019-07-15
Refactor email code fixes #1471

### LINKED ISSUE #1471: Refactor email code
I need to  clean stuff up across all emails we send.

uhlissuh:
some of the main things going on in this code refactor are...

Extraction of a method `format_and_send_html_email` that can be used by every html email we want to send. 

To do this, I had to change the intent to ship email and launch email to be an html email, so you will see those email bodies file changed. Sending as html will give us the ability to do links, which will be useful. 

Other major changes include changing how the intent to ship email _test_ worked. We are no longer matching the entire body string, because this was really fragile and tedious.


app/experimenter/experiments/email.py
```diff
@@ -14,96 +14,75 @@ def send_intent_to_ship_email(experiment_id):
...
-        [settings.EMAIL_RELEASE_DRIVERS],
-        cc=[experiment.owner.email],
-    )
-    email.send(fail_silently=False)
-
-    ExperimentEmail.objects.create(
-        experiment=experiment, type=Experiment.INTENT_TO_SHIP_EMAIL_LABEL
+        Experiment.INTENT_TO_SHIP_EMAIL_SUBJECT,
+        Experiment.INTENT_TO_SHIP_EMAIL_LABEL,
+        cc_recipients=[settings.EMAIL_RELEASE_DRIVERS],
```
jaredkerim: Oh I see so emails always go to the owner and subscribers, and then intent to ship just also goes to the release drivers.  Yeah that seems sensible.

app/experimenter/experiments/email.py
```diff
@@ -14,96 +14,75 @@ def send_intent_to_ship_email(experiment_id):
...
         {
             "experiment": experiment,
             "change_window_url": ExperimentConstants.NORMANDY_CHANGE_WINDOW,
         },
+        Experiment.ENDING_EMAIL_SUBJECT,
+        Experiment.EXPERIMENT_ENDS,
     )
 
+
+def format_and_send_html_email(
```
jaredkerim: So we don't test this function directly, we just test its uses in the specific emails?

app/experimenter/experiments/email.py
uhlissuh: yea that's it!

app/experimenter/experiments/email.py
uhlissuh: Yes I see what you mean, that would have prevent some small duplication across all tests... maybe for the next refactor 🤓

PR #1476, opened 2019-07-16
Add enrollment pause emails fixes #811

### LINKED ISSUE #811: email to check on enrollment progress
For "Live" experiments only - where the enrollment end date is different than the experiment end date, send email 5 days before the enrollment end-date.   If enrollment end is closer than 5 days (ex: a very short enrollment after launch), it is OK if this is sent with the launch emails.

hyperlink reference in the email should go here: https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-NormandyChangeWindow

In the example below is a place {type}: {name}, {their experimenter URL}.  It would be amazing if it could say "Pref-Flip: Foo experiment name" and that hyperlink right to the experimenter link for that study.  I wasn't sure how to ask that.  If not, listing "pref-flip: Foo experiment name, Experimenter URL" works too.

To: Owner, anyone subscribed
From: Experimenter Bot
Subject: Experimenter enrollment ending verification for {name}

Hello,

This is an automatic email from Experimenter about {type}: {name}, {their experimenter URL}.  Enrollment is scheduled to end on or within 6 days after {enrollment end date}, during the next Normandy change window {hyper link on "change window"}.  Please check how enrollment is progressing.   If you need a specific date sooner or if you need the enrollment extended, please SLACK: #ask-experimenter for changes.  

{name} is scheduled to end on {end date}.  If everything looks good, no action is needed. 

Thank you!
Experimenter and Normandy Teams

uhlissuh:
Adds enrollment ending email! Will send email 5 days (or less) out to owner, subscribers!

### DIFF
--- a/app/experimenter/experiments/tasks.py
+++ b/app/experimenter/experiments/tasks.py
@@ -216,7 +216,7 @@ def send_period_ending_emails(experiment):
         if (experiment.enrollment_end_date - date.today()) <= timedelta(
             days=5
         ):
-            if not Experiment.objects.filter(
+            if not ExperimentEmail.objects.filter(
                 experiment=experiment,
                 type=ExperimentConstants.EXPERIMENT_PAUSES,
             ):
--- a/app/experimenter/experiments/tasks.py
+++ b/app/experimenter/experiments/tasks.py
@@ -10,7 +10,6 @@ from experimenter.experiments import bugzilla, normandy, email
 from experimenter.experiments.constants import ExperimentConstants
 from experimenter.experiments.models import Experiment, ExperimentEmail
 from experimenter.notifications.models import Notification
-from datetime import date, timedelta
 
 
 logger = get_task_logger(__name__)
@@ -201,10 +200,10 @@ def update_status(experiment):
 
 def send_period_ending_emails(experiment):
     # send experiment ending soon emails if end date is 5 days out
-    if (experiment.end_date - date.today()) <= timedelta(days=5):
+    if experiment.ending_soon:
         if not ExperimentEmail.objects.filter(
             experiment=experiment, type=ExperimentConstants.EXPERIMENT_ENDS
-        ):
+        ).exists():
             email.send_experiment_ending_email(experiment)
             logger.info(
                 "Sent ending email for Experiment: {}".format(experiment)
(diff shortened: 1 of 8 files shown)


app/experimenter/experiments/tasks.py
```diff
@@ -210,6 +210,23 @@ def send_period_ending_emails(experiment):
                 "Sent ending email for Experiment: {}".format(experiment)
             )
 
+    # send enrollment ending emails if enrollment end
+    # date is 5 days out
+    if experiment.enrollment_end_date:
+        if (experiment.enrollment_end_date - date.today()) <= timedelta(
+            days=5
+        ):
+            if not Experiment.objects.filter(
```
tiftran: Might be a dumb question... but is this supposed to be ExperimentEmail.objects.filter()?

uhlissuh: WOW.. oh my Gaia this was a great catch Tif!!! Can't believe I missed that and obviously I need to fix the tests as well, this should have been caught by testing!

uhlissuh:
@tiftran I made that change you pointed out... I also just generally strengthened those tests.. the tests that should have been in place to check that. :facepalming I also strengthened tests for email for exp ending... even though this PR doesn't cover those and there wasn't a problem with that code. Not sure if I should have strengthened them in another PR or not.

app/experimenter/experiments/tasks.py
```diff
@@ -210,6 +210,23 @@ def send_period_ending_emails(experiment):
...
+    # send enrollment ending emails if enrollment end
+    # date is 5 days out
+    if experiment.enrollment_end_date:
+        if (experiment.enrollment_end_date - date.today()) <= timedelta(
+            days=5
+        ):
+            if not ExperimentEmail.objects.filter(
+                experiment=experiment,
+                type=ExperimentConstants.EXPERIMENT_PAUSES,
+            ):
```
jaredkerim: This can help make the db query faster/simpler:

https://docs.djangoproject.com/en/2.2/ref/models/querysets/#exists

app/experimenter/experiments/tasks.py
```diff
@@ -210,6 +210,23 @@ def send_period_ending_emails(experiment):
                 "Sent ending email for Experiment: {}".format(experiment)
             )
 
+    # send enrollment ending emails if enrollment end
+    # date is 5 days out
+    if experiment.enrollment_end_date:
+        if (experiment.enrollment_end_date - date.today()) <= timedelta(
```
jaredkerim: These two checks, the date check and the email exists check, both only need the experiment instance and some constants, so in theory it could be an experiment instance method?

uhlissuh: Yes, I see what you mean about that...

uhlissuh:
@jaredkerim I liked those ideas and I made the changes. 😄

PR #1481, opened 2019-07-17
added survey indicators fixes #1265

### LINKED ISSUE #1265: Way to identify and find experiments with a survey
blocked on #507 

We had 3 issues with surveys.  1) we didn't get the info needed.  #507 addresses that 2) surveys often require delicate ending coordination - so I need a way to make sure we triple check experiments with surveys 3) the survey team should have a simple way to see what experiments need surveys.

* can you think of a place to add a filter to identify studies with a survey?  I can't see how the filtering for "in QA" looks anymore in stage (#941) in staging - but would like it to make sense with that filtering model.  It's a similar use case - ability to slice out experiments that need a certain type of attention (in this case "survey" related).  
* is there a way to add a visual notification under the experiment timeline if there is a survey needed bubble checked?   We see that summary screen when we filter for what changes need to be made.  So a visual indicator would be very helpful - i'm fine with the words "includes survey" under the date range.  maybe in bubble format?  I'm totally open to ideas (if it's even feasible)
image

uper helpful for us and the survey team to be able to pull that info out.  Surveys are work for them and they can totally screw up our timing when we end a study.  we've been bit so many times on that.

tiftran:
Overview
Screen Shot 2019-07-17 at 12 39 06 PM

Experiment Detailed...
Screen Shot 2019-07-17 at 1 20 03 PM

### DIFF
--- a/app/experimenter/templates/experiments/detail_base.html
+++ b/app/experimenter/templates/experiments/detail_base.html
@@ -17,6 +17,9 @@
   </h3>
 
   {{ experiment.dates }}
+  {% if experiment.survey_required %}
+    <span class="badge badge-secondary">Includes Survey</span>
+  {% endif %}
 {% endblock %}
 
 {% block header_links %}
--- a/app/experimenter/templates/experiments/list.html
+++ b/app/experimenter/templates/experiments/list.html
@@ -110,6 +110,9 @@
               {% if experiment.enrollment_end_date %}
                 <p>Enrolling until {{ experiment.enrollment_end_date }}</p>
               {% endif %}
+              {% if experiment.survey_required %}
+                <span class="badge badge-secondary mb-2">Includes Survey</span>
+              {% endif %}
             </div>
           </div>
         </div>


app/experimenter/templates/experiments/detail_base.html
```diff
@@ -17,6 +17,9 @@ <h3>
   </h3>
 
   {{ experiment.dates }}
+  {% if experiment.survey_required %}
+    <span class="badge badge-secondary">Includes Survey</span>
+  {% endif %}
 {% endblock %}
```
uhlissuh: Is the screenshot from experimenter channel what this ended up looking like? I am wondering if this pill could use the slightest bit of margin on the left..

uhlissuh:
Hey Tif, would you put screenshots in the PR for easier viewing next time maybe? This one is so visual-heavy. 😄

app/experimenter/templates/experiments/list.html
```diff
@@ -110,6 +110,9 @@ <h5>{{ experiment.population }}</h5>
               {% if experiment.enrollment_end_date %}
                 <p>Enrolling until {{ experiment.enrollment_end_date }}</p>
               {% endif %}
+              {% if experiment.survey_required %}
+                <span class="badge badge-secondary mb-2">Includes Survey</span>
+              {% endif %}
             </div>
```
uhlissuh: what's the margin bottom for? `mb-2`

app/experimenter/templates/experiments/detail_base.html
tiftran: ooh you're right i should've put screen shots....

app/experimenter/templates/experiments/list.html
tiftran: Visually I think it looks better

without
Screen Shot 2019-07-17 at 3 53 30 PM

with
Screen Shot 2019-07-17 at 3 53 44 PM

uhlissuh: oh yea definitely! I wasn't considering the situation where the description column had less height than the includes survey column!

PR #1490, opened 2019-07-19
added enrollment complete indicator fixes #1293

### LINKED ISSUE #1293: Add an “Enrollment Paused” label next to the “Live” label.
Add an “Enrollment Paused” label next to the “Live” label.  
image

NEW REQUEST: Have a label on experiments have "Enrollment Paused"  - that is based off the Normandy API Pause information.

OLD REQUEST (we aren't doing this based on ability to get info from Normandy API):
This should be controlled by a Normandy folks manual click.  There is not a programmatic way for Normandy to indicate if an experiment is paused.  So would like a manual click, so we know when these transitions have been done without looking at code.

Maybe under “Archive Experiment” - but it’s not available unless you report to Benson (or are benson or shell).

tiftran:
Screen Shot 2019-07-19 at 10 45 15 AM

Screen Shot 2019-07-19 at 2 00 13 PM


jaredkerim:
@tiftran screenshots?

tiftran:
🤦‍♀ sorry... I had them... but i didn't include them

uhlissuh:
is there any way we can make that filter label less long so it doesn't go to a new line? 😄 seems hard if we can't use the word "paused" since people outside of experimenter don't know what that word means?

tiftran:
hmmmm @uhlissuh  @jaredkerim  would "Show completed enrollments" in place of "show enrollment complete experiments" make sense?

app/experimenter/experiments/models.py
```diff
@@ -700,6 +700,10 @@ def firefox_channel_sort():
     def is_archivable(self):
         return self.status != self.STATUS_LIVE
 
+    @property
+    def is_enrollment_complete(self):
+        return self.is_paused and self.status == self.STATUS_LIVE
```
jaredkerim: No model tests for this method?

tiftran:
@jaredkerim @uhlissuh i was thinking of a diff color that was distinctly different from all the other status pills and ended up with yellow... secretly hoping yellow was different enough to the sign off orange...

Maybe a dark green might be a better idea? so that greens are reserved for live?
Screen Shot 2019-07-23 at 1 58 03 PM

jaredkerim:
Yeah I like the dark green better it feels less confusing.  So funny a bunch of engineers trying to figure out what colours mean....

app/experimenter/experiments/tests/test_models.py
```diff
@@ -569,6 +569,23 @@ def test_observation_dates_returns_date_string(self):
             "Jan 11, 2019 - Jan 21, 2019 (10 days)",
         )
 
+    def test_enrollment_is_complete(self):
```
jaredkerim: 👍

app/experimenter/experiments/tests/test_models.py
```diff
@@ -569,6 +569,23 @@ def test_observation_dates_returns_date_string(self):
...
+        )
+        self.assertTrue(experiment.is_enrollment_complete)
+
+    def test_enrollment_is_not_complete(self):
+        experiment = ExperimentFactory.create_with_status(
+            target_status=Experiment.STATUS_LIVE, is_paused=False
+        )
+        self.assertFalse(experiment.is_enrollment_complete)
+
+        # wrong status
```
jaredkerim: What's this doing here?

tiftran: 🤔  not sure... will remove...

PR #1493, opened 2019-07-19
Add bugzilla urls to searchable fields fixes #1492

### LINKED ISSUE #1492: Make bugzilla ticket numbers searchable
We should be able to type a bug number into the search box and get all the experiments that match back from feedback in https://github.com/mozilla/experimenter/issues/1482

jaredkerim:


uhlissuh:
good!

uhlissuh:
the tests are a little flimsy that I wrote on this way back.. doesn't test any fields really besides name. Does that concern you?

jaredkerim:
@uhlissuh Yeah I was thinking about that, this is one of those cases where you can think of it either as 'the tests say these exact FIELDS are searchable' or 'the tests say that fields defined in the list of searchable fields are searchable'.  There's no additional code or logic to make new fields searchable, it's just declarative, so as long as the tests say 'whichever fields are defined as searchable will be searchable' then the logic is covered.  I tend to think that's sufficient, what do you think?

PR #1496, opened 2019-07-22
added archive change logs fixes #1395

### LINKED ISSUE #1395: Archive date does not appear in history
When a user archives or unarchives an experiment, it should create a record in the changelog history.  Looks like it doesn't right now.

tiftran:
Screen Shot 2019-07-22 at 12 45 13 PM


app/experimenter/experiments/forms.py
```diff
@@ -96,7 +97,16 @@ def save(self, *args, **kwargs):
             old_values["variants"] = None
             new_values["variants"] = self.new_serialized_vals["variants"]
 
-        if self.changed_data:
+        if experiment.archived or "archived" in self.changed_data:
```
jaredkerim: What happens if someone edits an archived experiment?

tiftran:
@jaredkerim not sure if this was the right way to go about it, but I felt that there was too much clutter in ChangeLogMixin and archive changelogs would only occur from the archiveForm. I moved the archive changelog logic into the form...

app/experimenter/experiments/forms.py
jaredkerim: Ugh sorry I was totally out to lunch when I reviewed this the first time, I was too caught in the details.

PR #1501, opened 2019-07-25
Support multiple normandy recipe ids fixes #1407

### LINKED ISSUE #1407: "Multiple Recipe" handling - gathering additional recipe(s) number(s) and adding a label
We need a way to mark an experiment that has multiple recipes.  The reasoning is these are more complex to manage - and we want to make sure when we are adjusting that we manage the additional recipes as needed.

Could we add an optional field in the Normandy Recipe view for "Additional Recipe ID(s)"?   This will not connect to any automation - it is purely for ease of manual tracking.  Ideally these will also be shown in the experiment view (in Normandy Info section) as "Additional Recipe ID(s)" (below "Normandy Recipe ID").  If there is any information in that "Additional Recipe ID(s)" field - it would be nice to label the experiment as "Multiple Recipes" (similar to "Live" or "Complete" bubbles).  If that is a problem - we can potentially live without it.

image

uhlissuh:
Fixes #1407 

In this PR we add support for other normandy recipe ids that may exist when a recipe needs to be split into two. 

Here's the look of the normandy modal now. I had to fit in an extra field, and was having a lot of trouble with getting bootstrap's rows and columns inside a form inside the `modal-footer` class... so I just did away with it and put stuff in the `modal-body` and tried to make it look like the footer's styling. 

Screen Shot 2019-07-24 at 8 31 49 PM

Here is where I put all the errors...

Screen Shot 2019-07-24 at 8 32 21 PM

And here's what the normandy info section now looks like, with a redesign to support the idea of there being more than one ID... 

Screen Shot 2019-07-24 at 2 45 12 PM

and if there is just one id....

Screen Shot 2019-07-24 at 2 45 38 PM

I'll make a review to explain some of the decisions I made.


app/experimenter/experiments/models.py
```diff
@@ -377,17 +381,38 @@ def has_normandy_info(self):
...
+                            "id": id,
+                            "normandy_url": normandy_recipe_url.format(id=id),
+                            "DC_url": delivery_console_url.format(id=id),
+                        }
+                    )
+
+            return urls
+
+        return None
+
```
uhlissuh: at first glance it might be like... what's up with this property that makes this list of dicts.... but essentially I arrived at this because I needed a method that I could call in the template that could turn all of the `other_normandy_ids` into DC urls and normandy urls. Can't call a function and pass things into it in a django template very easily... so I made a property on the model that would essentially return a dict for each id, the main and extra ids if there are any... with both of the necessary urls... and then it made it pretty simple on the template side to loop over this list to display the ids and urls in the normandy info section....  perhaps someone has a better idea...

app/experimenter/experiments/models.py
```diff
@@ -377,17 +381,38 @@ def has_normandy_info(self):
...
 
     @property
-    def normandy_api_recipe_url(self):
-        if self.normandy_id:
-            return settings.NORMANDY_API_RECIPE_URL.format(id=self.normandy_id)
+    def format_dc_normandy_urls(self):
+        # returns a list of dictionaries containing D.C. and Normandy
+        # urls for the main normandy id and other normandy ids if they exist
+        normandy_recipe_url = settings.NORMANDY_API_RECIPE_URL
+        delivery_console_url = settings.DELIVERY_CONSOLE_RECIPE_URL
```
uhlissuh: Had to make these variables because the Black didn't like a lot of the long lines in this method here...

app/experimenter/experiments/models.py
```diff
@@ -377,17 +381,38 @@ def has_normandy_info(self):
...
-                id=self.normandy_id
+            urls = []
+            urls.append(
+                {
+                    "id": self.normandy_id,
+                    "normandy_url": normandy_recipe_url.format(
+                        id=self.normandy_id
+                    ),
+                    "DC_url": delivery_console_url.format(id=self.normandy_id),
+                }
```
uhlissuh: again these keys have pretty short names... because of line length problems..

app/experimenter/experiments/tests/test_forms.py
```diff
@@ -1954,3 +1955,39 @@ def test_form_removes_subscribers(self):
...
+    def test_form_valid_when_other_ids_formatted_correctly(self):
+        experiment = ExperimentFactory.create()
+
+        form = NormandyIdForm(
+            self.request,
+            instance=experiment,
+            data={"normandy_id": "4343", "other_normandy_ids": "4323, 5671"},
+        )
+
+        self.assertTrue(form.is_valid())
```
uhlissuh: there were no form tests for the NormandyIdForm... so I made some since I added an elaborate clean method... didn't exhaustively test all possible wonky things that could be passed in to the form... but I did try to break this in lots of ways and I think I caught a good portion of the wonky stuff that could be passed in...

app/experimenter/experiments/tests/test_models.py
```diff
@@ -376,34 +376,66 @@ def test_has_normandy_info_true_with_normandy_id(self):
...
+                "http://normandy.example.com/recipe/32/",
+            )
+            self.assertEqual(
+                experiment.format_dc_normandy_urls[1]["normandy_url"],
+                "http://normandy.example.com/recipe/43/",
+            )
+            self.assertEqual(
+                experiment.format_dc_normandy_urls[2]["normandy_url"],
+                "http://normandy.example.com/recipe/56/",
             )
```
uhlissuh: here i am testing that model property that makes the list of dicts...

app/experimenter/templates/experiments/detail_ship.html
```diff
@@ -41,33 +41,35 @@ <h5 class="modal-title" id="normandyModalLabel">Normandy Recipe</h5>
         <div class="modal-body">
           <p>Enter the following information into Delivery Console exactly as shown:</p>
           <pre class="bg-light json prettyprint">{{ experiment.normandy_recipe_json }}</pre>
-          <p>When the recipe is saved in Delivery Console, copy the Recipe ID below and hit save.</p>
-        </div>
 
-        <div class="modal-footer">
```
uhlissuh: here is where I bidded adieu the modal-footer... which really wasn't loving the embedded form... and me trying to make rows and columns in the form...

app/experimenter/templates/experiments/section_normandy.html
```diff
@@ -12,21 +12,23 @@
...
+            Delivery Console Recipe
+          </a> |
+          <a href="{{ entry.normandy_recipe_url }}" target="blank" rel="noreferrer noopener">
+            Normandy API Recipe
+          </a>
+          {% if experiment.normandy_id == entry.id %}
+            <span class="badge badge-secondary">Primary</span>
+          {% endif %}
+        </p>
+      {% endfor %}
```
uhlissuh: here's where i use my list of dicts to display the ids and urls...

app/experimenter/experiments/forms.py
```diff
@@ -1271,10 +1272,42 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+            }
+        ),
+        required=False,
+    )
+
+    def clean_other_normandy_ids(self):
+        other_ids_list = []
+
+        if self.cleaned_data["other_normandy_ids"]:
+            ids = re.split(",", self.cleaned_data["other_normandy_ids"])
```
jaredkerim: You don't need `re` for this, strings have `split` built in.

app/experimenter/experiments/forms.py
```diff
@@ -1271,10 +1272,42 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+            for id in ids:
```
jaredkerim: Technically `id` is a reserved word in python.  Python doesn't stop you from reusing reserved words but it's generally a good idea to avoid doing it.

app/experimenter/experiments/forms.py
```diff
@@ -1271,10 +1272,42 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                if re.match(r"^[0-9]+$", id.strip()):
```
jaredkerim: You could also use [`.isdigit()`](https://docs.python.org/3/library/stdtypes.html#str.isdigit) here instead of needing a regex

app/experimenter/experiments/forms.py
```diff
@@ -1271,10 +1272,42 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                    if "normandy_id" in self.cleaned_data:
```
jaredkerim: This will have to be here because the field is required and that will get checked before this validation happens.

app/experimenter/experiments/forms.py
```diff
@@ -1271,10 +1272,42 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                        if int(id) == self.cleaned_data["normandy_id"]:
```
jaredkerim: In Django, if you want to validate multiple fields together you have to do it in the [clean method](https://docs.djangoproject.com/en/2.2/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other)

jaredkerim: but 👍 for doing this check :D

app/experimenter/experiments/forms.py
```diff
@@ -1271,10 +1272,42 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                            raise forms.ValidationError(
+                                "Duplicate IDs are not accepted."
+                            )
+                        else:
+                            other_ids_list.append(int(id))
+                else:
+                    raise forms.ValidationError(
```
jaredkerim: Instead of doing this with a bunch of `else` statements that get deeper and deeper, you could flatten it more like

    if not some_check:
        raise forms.ValidationError('🐧')
    if not some_other_check:
        raise forms.ValidationError('🐻')

app/experimenter/experiments/models.py
```diff
@@ -377,17 +381,38 @@ def has_normandy_info(self):
...
+                        {
+                            "id": id,
+                            "normandy_url": normandy_recipe_url.format(id=id),
+                            "DC_url": delivery_console_url.format(id=id),
+                        }
+                    )
+
+            return urls
+
+        return None
```
jaredkerim: Functions always return `None` by default

app/experimenter/experiments/models.py
```diff
@@ -377,17 +381,38 @@ def has_normandy_info(self):
...
+        # urls for the main normandy id and other normandy ids if they exist
+        normandy_recipe_url = settings.NORMANDY_API_RECIPE_URL
+        delivery_console_url = settings.DELIVERY_CONSOLE_RECIPE_URL
 
-    @property
-    def delivery_console_recipe_url(self):
         if self.normandy_id:
-            return settings.DELIVERY_CONSOLE_RECIPE_URL.format(
-                id=self.normandy_id
+            urls = []
```
jaredkerim: You can collapse these two lines together...

    urls = [{
        ....
    }]

jaredkerim: Alternatively you could change this whole function signature to always return a list

    urls = []
    if self.normandy_id:
        ....
    if self.other_normandy_ids:
        ....
    return urls

then the return type is always the same and you can just say

    {% for url in experiment.format.... %}
    {% %}

instead of needing an `if` check?

app/experimenter/experiments/models.py
jaredkerim: Yeah that's fine, linters are your friend, do what they say. 👬

app/experimenter/experiments/models.py
jaredkerim: Yeah I mean I do this kind of thing all the time, your instincts are correct here 👍

jaredkerim: Or at least if it's wrong, then we're both wrong together XD

app/experimenter/experiments/tests/test_forms.py
jaredkerim: Yep fantastic, I was just saying this to Jenny earlier today, test the most obvious failure cases, test the happy case, that's good enough until QA finds arcane eldritch incantations to break it in ways that cause our fragile human minds to reel and roil and the piecing together of dissociated knowledge will open up such terrifying vistas of reality, and of our frightful position therein, that we shall either go mad from the revelation or flee from the light into the peace and safety of a new dark age. Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -376,34 +376,66 @@ def test_has_normandy_info_true_with_normandy_id(self):
...
-                experiment.normandy_api_recipe_url,
-                "http://normandy.example.com/recipe/123/",
+                experiment.format_dc_normandy_urls[0]["DC_url"],
+                "http://delivery-console.example.com/recipe/445/",
             )
 
-    def test_delivery_console_recipe_url_returns_none_if_no_normandy_id(self):
-        experiment = ExperimentFactory.create(normandy_id=None)
-        self.assertEqual(experiment.delivery_console_recipe_url, None)
+        with override_settings(
```
jaredkerim: excellent excellent excellent 👏

app/experimenter/experiments/tests/test_models.py
jaredkerim: Yep looks good, verbose but only necessarily so 🤷‍♀

app/experimenter/experiments/views.py
```diff
@@ -509,14 +509,11 @@ def form_valid(self, form):
...
-            "{url}?normandy_id={value}".format(
-                url=reverse(
-                    "experiments-detail", kwargs={"slug": self.kwargs["slug"]}
-                ),
-                value=form.data["normandy_id"],
-            )
+        url = reverse(
+            "experiments-detail", kwargs={"slug": self.kwargs["slug"]}
         )
+        query_parameters = form.data.urlencode()
```
jaredkerim: 🎉 🎉 🎉 🎉 🎉 🎉 🎉

app/experimenter/templates/experiments/detail_ship.html
jaredkerim: BYEEEEEEEEE
giphy

app/experimenter/templates/experiments/detail_ship.html
```diff
@@ -41,33 +41,35 @@ <h5 class="modal-title" id="normandyModalLabel">Normandy Recipe</h5>
...
 
-          <form class="form-inline" method="post" action="{% url "experiments-normandy-update" slug=experiment.slug %}">
+          <p>When the recipe is saved in Delivery Console, copy the Recipe ID below and hit save.</p>
```
jaredkerim: I like this layout, but I will say that errors really should go as close to the field they're talking about as possible.  I think you could put them either just above or just below their respective fields?

app/experimenter/templates/experiments/section_normandy.html
```diff
@@ -12,21 +12,23 @@
...
-    </p>
+  {% if experiment.normandy_id  %}
+    <p><strong>Normandy Recipe IDs:</strong></p>
 
-    <p>
-      <a href="{{ experiment.normandy_api_recipe_url }}" target="blank" rel="noreferrer noopener">
-        Normandy API Recipe
-      </a>
-    </p>
+      {% for entry in experiment.format_dc_normandy_urls %}
```
jaredkerim: Ah yeah, this is silently failing on trying to loop over `None` but Django templates try their best to hide errors so it's just being quiet about it.

app/experimenter/experiments/models.py
jaredkerim: Ah I forgot that templates will just silently fail on that, but still, if this ever gets used somewhere else it might be nice if the return type is always a list..... 🤔

app/experimenter/experiments/forms.py
uhlissuh: hmm.. so you are saying this check should be moved to the clean method?

uhlissuh:
so you mean the errors right under the fields like this? even though it pushes the buttons down?

Screen Shot 2019-07-29 at 3 22 20 PM

Screen Shot 2019-07-29 at 3 22 48 PM

app/experimenter/experiments/forms.py
jaredkerim: Actually you can't check the cleaned value of other fields from a specific fields clean method, you have to do that in a `clean` method for the whole form because fields can be validated in any order.

app/experimenter/experiments/models.py
uhlissuh: Yes I updated it to return an empty list either way. I considered that initially... not sure why I returned none instead...

app/experimenter/templates/experiments/detail_ship.html
uhlissuh: ok I fixed the layout.

app/experimenter/experiments/forms.py
uhlissuh: did some flattening..

app/experimenter/experiments/forms.py
uhlissuh: moved to clean method solved a few probs...

uhlissuh:
@jaredkerim , I have now fixed the issue of removing the csrf token from the url parameters by using `.copy()` to get a mutable querydict then poping off that key.

app/experimenter/experiments/forms.py
```diff
@@ -1277,10 +1277,47 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                "placeholder": "Other Recipe IDs (Optional)",
+            }
+        ),
+        required=False,
+    )
+
+    def clean(self):
+        if "other_normandy_ids" in self.cleaned_data:
+
+            for number in self.cleaned_data["other_normandy_ids"]:
```
jaredkerim: Oh actually you don't need to loop and compare, you can just do `if x in y`

app/experimenter/experiments/forms.py
```diff
@@ -1277,10 +1277,47 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                "class": "form-control",
+                "placeholder": "Other Recipe IDs (Optional)",
+            }
+        ),
+        required=False,
+    )
+
+    def clean(self):
+        if "other_normandy_ids" in self.cleaned_data:
+
```
jaredkerim: empty_line--

app/experimenter/experiments/forms.py
```diff
@@ -1277,10 +1277,47 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+        if self.cleaned_data["other_normandy_ids"]:
+            ids = self.cleaned_data["other_normandy_ids"].split(",")
+            for item in ids:
+                if not item.strip().isdigit():
+                    raise forms.ValidationError(
+                        "IDs must be numbers separated by commas."
+                    )
+
+                other_ids_list.append(int(item))
```
jaredkerim: So, in THEORY, something could pass `item.strip().isdigit()` but still fail `int(item)` and then this would crash.  If you're going to do `int(item)` anyways (because that's how the db expects it) then you could just let that be your actual test and then catch the `ValueError` and let that be the reason you throw the `ValidationError` and this would be a little shorter and also prevent the case that it passes the first check but throws when you cast to int?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -376,34 +376,68 @@ def test_has_normandy_info_true_with_normandy_id(self):
...
+        self.assertEqual(len(experiment.format_dc_normandy_urls), 0)
 
-    def test_normandy_api_recipe_url_returns_url_with_normandy_id(self):
-        experiment = ExperimentFactory.create(normandy_id="123")
+    def test_format_dc_normandy_urls_with_only_main(self):
+        experiment = ExperimentFactory.create(normandy_id="445")
         with override_settings(
-            NORMANDY_API_RECIPE_URL="http://normandy.example.com/recipe/{id}/"
+            DELIVERY_CONSOLE_RECIPE_URL=(
+                "http://delivery-console.example.com/recipe/{id}/"
```
jaredkerim: Nice nice nice 👍

app/experimenter/experiments/tests/test_models.py
```diff
@@ -376,34 +376,68 @@ def test_has_normandy_info_true_with_normandy_id(self):
...
+                experiment.format_dc_normandy_urls[1]["DC_url"],
+                "http://delivery-console.example.com/recipe/43/",
+            )
+            self.assertEqual(
+                experiment.format_dc_normandy_urls[2]["DC_url"],
+                "http://delivery-console.example.com/recipe/56/",
+            )
+
+        with override_settings(
+            NORMANDY_API_RECIPE_URL="http://normandy.example.com/recipe/{id}/"
```
jaredkerim: Why not just do both overrides in one statement and assert everything in one go?

app/experimenter/experiments/views.py
```diff
@@ -525,14 +525,14 @@ def form_valid(self, form):
...
-                url=reverse(
-                    "experiments-detail", kwargs={"slug": self.kwargs["slug"]}
-                ),
-                value=form.data["normandy_id"],
-            )
+        url = reverse(
+            "experiments-detail", kwargs={"slug": self.kwargs["slug"]}
         )
+        query_parameters = form.data.copy()
+        if query_parameters.__contains__("csrfmiddlewaretoken"):
```
jaredkerim: :O You know there's a more idiomatic way to know if something is `in` something else, right?  There's a keyword I have `in` mind, perhaps it could provide some `in`sight into what you're doing here, perhaps I'm `in`dicating something......

app/experimenter/experiments/views.py
```diff
@@ -525,14 +525,14 @@ def form_valid(self, form):
...
-            "{url}?normandy_id={value}".format(
-                url=reverse(
-                    "experiments-detail", kwargs={"slug": self.kwargs["slug"]}
-                ),
-                value=form.data["normandy_id"],
-            )
+        url = reverse(
+            "experiments-detail", kwargs={"slug": self.kwargs["slug"]}
         )
+        query_parameters = form.data.copy()
```
jaredkerim: Why do you have to copy? You can't pop from the original?

app/experimenter/experiments/views.py
```diff
@@ -525,14 +525,14 @@ def form_valid(self, form):
...
+        if query_parameters.__contains__("csrfmiddlewaretoken"):
+            query_parameters.pop("csrfmiddlewaretoken")
```
jaredkerim: Also is there a way to pass a fallback value to `pop` so you don't need to do the `if` check first?

app/experimenter/experiments/views.py
uhlissuh: @jaredkerim  the original is immutable...

app/experimenter/experiments/views.py
uhlissuh: oh yea duh.

app/experimenter/experiments/forms.py
```diff
@@ -1279,10 +1279,51 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+        return self.cleaned_data
+
+    def clean_other_normandy_ids(self):
+        other_ids_list = []
+
+        if self.cleaned_data["other_normandy_ids"]:
+            ids = self.cleaned_data["other_normandy_ids"].split(",")
+            for item in ids:
+                try:
+                    other_ids_list.append(int(item.strip()))
```
jaredkerim: List comprehensions make everything better:

    try:
        return [int(i.strip()) for i in self.cleaned_data["other_normandy_ids"].split(",")]
    catch ValueError:
        raise forms.ValidationError(.....)

app/experimenter/experiments/forms.py
```diff
@@ -1279,10 +1279,51 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                raise forms.ValidationError(
+                    {"other_normandy_ids": "Duplicate IDs are not accepted."}
+                )
+
+        return self.cleaned_data
+
+    def clean_other_normandy_ids(self):
+        other_ids_list = []
+
+        if self.cleaned_data["other_normandy_ids"]:
```
jaredkerim: I don't think you need to do this check, by the time we get to this method the value should always be there:

https://github.com/django/django/blob/master/django/forms/forms.py#L395-L398

app/experimenter/experiments/forms.py
```diff
@@ -1279,10 +1279,51 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+        ),
+        required=False,
+    )
+
+    def clean(self):
+        # we need to check that both keys exist because if the user enters
+        # a noninteger normandy_id, the key is taken out of cleaned_data
+        # and this method will still be run causing a key error
+
+        if {"normandy_id", "other_normandy_ids"} <= set(self.cleaned_data):
```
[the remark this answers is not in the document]
uhlissuh: yea that was the internet right there with that fanciness.

app/experimenter/experiments/forms.py
uhlissuh: if this check is not here, the form won't let the other_normandy_ids field be optional.

app/experimenter/experiments/forms.py
```diff
@@ -1279,10 +1279,47 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+            "other_normandy_ids", []
+        ):
+            raise forms.ValidationError(
+                {"other_normandy_ids": "Duplicate IDs are not accepted."}
+            )
+
+        return cleaned_data
+
+    def clean_other_normandy_ids(self):
+        if not self.cleaned_data["other_normandy_ids"].strip():
```
jaredkerim: Good idea!

app/experimenter/experiments/forms.py
```diff
@@ -1279,10 +1279,47 @@ class NormandyIdForm(ChangeLogMixin, forms.ModelForm):
...
+                "placeholder": "Other Recipe IDs (Optional)",
+            }
+        ),
+        required=False,
+    )
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if cleaned_data.get("normandy_id") in cleaned_data.get(
```
jaredkerim: 🙆

PR #1504, opened 2019-07-26
added enrollment info into changelog fixes #1452

### LINKED ISSUE #1452: Find a way to obtain date of enrollment pauses
currently #1318 gives the state of a experiment whether it's paused or not. It would be a nice to have to know the date as well

tiftran:
Screen Shot 2019-07-29 at 8 38 47 AM

### DIFF
--- a/app/experimenter/experiments/tasks.py
+++ b/app/experimenter/experiments/tasks.py
@@ -170,7 +170,7 @@ def update_status(experiment):
                 enabler_email = creator.get("email")
 
         enabler, _ = get_user_model().objects.get_or_create(
-            email=enabler_email
+            email=enabler_email, username=enabler_email
         )
 
         old_status = experiment.status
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tasks.py
```diff
@@ -170,7 +170,7 @@ def update_status(experiment):
                 enabler_email = creator.get("email")
 
         enabler, _ = get_user_model().objects.get_or_create(
-            email=enabler_email
+            email=enabler_email, username=enabler_email
```
tiftran: had to add username, to ensure that there was an email in the changelog

mythmon:
Are you often seeing the case that produces `unknown-user@normandy.mozilla.com` from a Normandy API? Or is just an artifact of test data?

tiftran:
@mythmon, unknown-user@normandy.mozilla.com is the default user when there isn't a user found in normandy I think you mentioned that it may happen but rare due to **https://github.com/mozilla/normandy/issues/1460** but at the same time, is there a way to know who changed the enrollment paused value?

mythmon:
You shouldn't be seeing mozilla/normandy#1460 on any recipes made this year. If you are, please flag that since it would be a bug.

As for finding who changed the enrollment paused value, yes that is possible. You'll have to look at the history of the recipe. For example, take a look at [the history of recipe 819](https://normandy.cdn.mozilla.net/api/v3/recipe/819/history/?format=json).

You can see that in the newest revision (with id 2662), `arguments.isEnrollmentPaused=true`, and in the previous revision (id 2660) `arguments.isEnrollmentPaused=false`. So that means that revsion 2662 is when it was paused, and the author of that revision is in the `creator` field. In this case it was Josh who paused it.

It's possible for a revision to change after it is paused, so to be accurate you'd have to scan through the revisions to find the point that paused changed from false to true.

tiftran:
@mythmon should've more detailed previously unknown-user@normandy.mozilla.com is used just in case of when testing on stage a random recipe id is given. Previously discussed in https://github.com/mozilla/experimenter/pull/1444#discussion_r301715293

But seeing as we can find who changed isEnrollmentPaused, will look into that

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -532,20 +603,23 @@ def test_send_experiment_ending_email(self):
             proposed_start_date=date.today(),
             proposed_enrollment=0,
             proposed_duration=5,
+            is_paused=True,
```
tiftran: Had to add these so, the pause enrollment workflow isn't tested in these tests for now...will fix when we break down the update_experiment_info task?

tiftran:
removing commit that digs through recipe history, as @jaredkerim mentioned it was a bit much

PR #1514, opened 2019-07-30
Add versions filter object to normandy recipe serializer fixes #1447

### LINKED ISSUE #1447: Add version filter_object to recipe serializer
Normandy supports a version filter object so we should add that to the recipe serializer.

uhlissuh:
fixes #1447 

adds filter object with list of integers for all versions the recipe affects. 

Screen Shot 2019-07-30 at 1 49 06 PM

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -669,6 +669,28 @@ class Experiment(ExperimentConstants, models.Model):
             return self.firefox_min_version
 
     @property
+    def versions_integer_list(self):
+        min_integer = int(
+            ExperimentConstants.VERSION_REGEX.match(
+                self.firefox_min_version
+            ).group(0)
+        )
+
+        integers_list = [min_integer]
+
+        if self.firefox_max_version:
+            max_integer = int(
+                ExperimentConstants.VERSION_REGEX.match(
+                    self.firefox_max_version
+                ).group(0)
+            )
+
+            for x in range(min_integer + 1, max_integer + 1):
+                integers_list.append(x)
+
+        return integers_list
+
+    @property
     def population(self):
         return "{percent:g}% of {channel} Firefox {firefox_version}".format(
             percent=float(self.population_percent),
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -343,15 +343,12 @@ class Experiment(ExperimentConstants, models.Model):
         if not self.bugzilla_id:
             raise ValueError(error_msg.format(field="Bugzilla ID"))
 
-        slug_min_version = ExperimentConstants.VERSION_REGEX.match(
-            self.firefox_min_version
-        ).group(0)
-        version_string = slug_min_version
+        version_string = self.firefox_min_version_integer
         if self.firefox_max_version:
-            slug_max_version = ExperimentConstants.VERSION_REGEX.match(
-                self.firefox_max_version
-            ).group(0)
-            version_string = f"{slug_min_version}-{slug_max_version}"
+            version_string = (
+                f"{self.firefox_min_version_integer}-"
+                f"{self.firefox_max_version_integer}"
+            )
 
         slug_prefix = f"{self.type}-"
         slug_postfix = (
@@ -669,26 +666,29 @@ class Experiment(ExperimentConstants, models.Model):
             return self.firefox_min_version
 
     @property
-    def versions_integer_list(self):
-        min_integer = int(
-            ExperimentConstants.VERSION_REGEX.match(
-                self.firefox_min_version
-            ).group(0)
-        )
-
-        integers_list = [min_integer]
-
+    def firefox_max_version_integer(self):
         if self.firefox_max_version:
-            max_integer = int(
+            return int(
                 ExperimentConstants.VERSION_REGEX.match(
                     self.firefox_max_version
                 ).group(0)
             )
 
-            for x in range(min_integer + 1, max_integer + 1):
-                integers_list.append(x)
+    @property
+    def firefox_min_version_integer(self):
+        return int(
+            ExperimentConstants.VERSION_REGEX.match(
+                self.firefox_min_version
+            ).group(0)
+        )
 
-        return integers_list
+    @property
+    def versions_integer_list(self):
+        max = (
+            self.firefox_max_version_integer
+            or self.firefox_min_version_integer
+        )
+        return list(range(self.firefox_min_version_integer, max + 1))
 
     @property
     def population(self):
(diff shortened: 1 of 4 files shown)


app/experimenter/experiments/models.py
```diff
@@ -668,6 +668,28 @@ def format_firefox_versions(self):
...
+
+        if self.firefox_max_version:
+            max_integer = int(
+                ExperimentConstants.VERSION_REGEX.match(
+                    self.firefox_max_version
+                ).group(0)
+            )
+
+            for x in range(min_integer + 1, max_integer + 1):
+                integers_list.append(x)
```
tiftran: would something like  `return list(range(min_integer, max_integer + 1)` be better than a for loop? not sure?

app/experimenter/experiments/models.py
```diff
@@ -668,6 +668,28 @@ def format_firefox_versions(self):
         else:
             return self.firefox_min_version
 
+    @property
+    def versions_integer_list(self):
+        min_integer = int(
```
jaredkerim: I feel like we do these regex extractions in a bunch of places, would it make sense to move this to a separate model method?  This method here shouldn't be as long as it is....

app/experimenter/experiments/models.py
jaredkerim: Weeeeelllllllll, think about what ways it might be better?  Is the code shorter?  Is it more resilient to refactoring (ie if stuff really gets moved around and variables change)?  Does mutating variables give you an uncomfortable feeling in your gut?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1085,6 +1085,17 @@ def test_format_firefox_versions_returns_correct_string(self):
         self.assertEqual(experiment_1.format_firefox_versions, "57.0")
         self.assertEqual(experiment_2.format_firefox_versions, "57.0 to 59.0")
 
+    def test_versions_integer_list_returns_correct_list(self):
```
jaredkerim: If a test fails, it should be super clear from the fact that the whole test failed exactly what is wrong.  If a single test tests multiple things or cases and then fails, then you have to go digging around to figure out WHICH case inside the test was failing, which adds a step to fixing it.  Is this test testing a single case or multiple cases?

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -315,6 +316,27 @@ def test_serializer_outputs_expected_schema(self):
         )
 
 
+class TestFilterObjectVersionsSerializer(TestCase):
+
+    def test_serializer_outputs_version_string_with_only_min(self):
```
jaredkerim: These tests are so clean I could eat off them 🍽 A++++

app/experimenter/experiments/models.py
uhlissuh: @jaredkerim yes I was debating making them their own model method. So I will!

app/experimenter/experiments/models.py
uhlissuh: yea it seems @tiftran 's is better... because it's shorter.

app/experimenter/experiments/models.py
```diff
@@ -668,6 +665,31 @@ def format_firefox_versions(self):
         else:
             return self.firefox_min_version
 
+    @property
+    def firefox_max_version_integer(self):
```
jaredkerim: Nice

app/experimenter/experiments/models.py
```diff
@@ -668,6 +665,31 @@ def format_firefox_versions(self):
...
+        if self.firefox_max_version:
+            return int(
+                ExperimentConstants.VERSION_REGEX.match(
+                    self.firefox_max_version
+                ).group(0)
+            )
+
+    @property
+    def firefox_min_version_integer(self):
```
jaredkerim: Very nice

app/experimenter/experiments/models.py
```diff
@@ -668,6 +665,31 @@ def format_firefox_versions(self):
...
+        return int(
+            ExperimentConstants.VERSION_REGEX.match(
+                self.firefox_min_version
+            ).group(0)
+        )
+
+    @property
+    def versions_integer_list(self):
```
jaredkerim: Super nice

app/experimenter/experiments/serializers.py
```diff
@@ -222,6 +222,21 @@ def get_channels(self, obj):
...
+
+    class Meta:
+        model = Experiment
+        fields = ("type", "versions")
+
+    def get_type(self, obj):
+        return "version"
+
+    def get_versions(self, obj):
+        return obj.versions_integer_list
```
jaredkerim: Damn right

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1085,6 +1085,34 @@ def test_format_firefox_versions_returns_correct_string(self):
         self.assertEqual(experiment_1.format_firefox_versions, "57.0")
         self.assertEqual(experiment_2.format_firefox_versions, "57.0 to 59.0")
 
+    def test_versions_integer_list_with_only_min_returns_correct_list(self):
+        experiment = ExperimentFactory(
+            firefox_min_version="57.0", firefox_max_version=""
+        )
+
+        self.assertEqual(experiment.versions_integer_list, [57])
```
jaredkerim: Awesome

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1085,6 +1085,34 @@ def test_format_firefox_versions_returns_correct_string(self):
...
+
+    def test_versions_integer_list_with_min_max_returns_correct_list(self):
+        experiment = ExperimentFactory(
+            firefox_min_version="57.0", firefox_max_version="59.0"
+        )
+
+        self.assertEqual(experiment.versions_integer_list, [57, 58, 59])
```
jaredkerim: Super awesome

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1085,6 +1085,34 @@ def test_format_firefox_versions_returns_correct_string(self):
...
+
+    def test_firefox_max_version_integer_returns_correct_integer(self):
+        experiment = ExperimentFactory(
+            firefox_min_version="57.0", firefox_max_version="59.0"
+        )
+
+        self.assertEqual(experiment.firefox_max_version_integer, 59)
```
jaredkerim: I love it

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1085,6 +1085,34 @@ def test_format_firefox_versions_returns_correct_string(self):
...
+
+    def test_firefox_min_version_integer_returns_correct_integer(self):
+        experiment = ExperimentFactory(
+            firefox_min_version="57.0", firefox_max_version="59.0"
+        )
+
+        self.assertEqual(experiment.firefox_min_version_integer, 57)
```
jaredkerim: Fantastic

PR #1527, opened 2019-08-02
Add API endpoint for changelog dates, fixes #1524

### LINKED ISSUE #1524: Add API endpoint for changelog dates
Avocado would like to access status change dates (from ExperimentChangeLog) when accessing the Experimenter API. This could be put under a key called `changes`, with value being an array of changes

zyjenny:
Screen Shot 2019-08-02 at 5 32 33 PM


app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -118,6 +120,14 @@ def test_serializer_outputs_expected_schema(self):
         )
 
 
+class TestStatusSerializer(TestCase):
```
tiftran: This should probably be `TestExperimentChangeLogSerializer` to be consistent

zyjenny: Thanks for pointing that out!!! :D

PR #1559, opened 2019-08-14
Refactoring scheduled celery task

tiftran:


app/experimenter/experiments/tasks.py
```diff
@@ -122,100 +121,124 @@ def update_experiment_bug_task(user_id, experiment_id):
...
+    update_live_experiments()
+    metrics.incr("update_experiment_info.completed")
+
 
-    for experiment in launch_experiments:
+def update_accepted_experiments():
+    accepted_experiments = Experiment.objects.filter(
+        status=Experiment.STATUS_ACCEPTED
+    )
+    for experiment in accepted_experiments:
```
jaredkerim: Fold these two loops back together

app/experimenter/experiments/tasks.py
```diff
@@ -122,100 +121,124 @@ def update_experiment_bug_task(user_id, experiment_id):
...
         try:
             logger.info("Updating Experiment: {}".format(experiment))
             if experiment.normandy_id:
-                update_status(experiment)
-                if experiment.status == Experiment.STATUS_LIVE:
-                    send_period_ending_emails(experiment)
+                recipe_data = normandy.get_recipe(experiment.normandy_id)
+                if needs_to_be_updated(recipe_data, experiment.status):
+                    update_status(experiment, recipe_data)
```
jaredkerim: This structure is really good and clear, just do the accepted/live check before this

app/experimenter/experiments/tasks.py
```diff
@@ -122,100 +121,124 @@ def update_experiment_bug_task(user_id, experiment_id):
 def update_experiment_info():
     metrics.incr("update_experiment_info.started")
     logger.info("Updating experiment info")
-    launch_experiments = Experiment.objects.filter(
-        Q(status=Experiment.STATUS_ACCEPTED) | Q(status=Experiment.STATUS_LIVE)
-    )
+    update_accepted_experiments()
```
jaredkerim: I think if you fold these two back together you can just flatten them back into this function

app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -369,29 +361,20 @@ def test_experiment_with_re_enabled_enrollment(self):
...
+        experiment = ExperimentFactory.create_with_status(
             target_status=Experiment.STATUS_LIVE,
             normandy_id=1234,
             is_paused=False,
         )
 
-        self.mock_normandy_requests_get.return_value = (
-            self.buildMockSucessWithNoPauseEnrollment()
-        )
-        tasks.update_experiment_info()
```
jaredkerim: Ah I realized something after we talked the other day, I called this 'cheating' but here's a concrete example:

If you deleted `set_is_paused_value(...)` from `update_experiment_info`, then this test would still pass but it when the task actually ran it would be broken, so it's generally good to test from the entry point that will get used in when the thing actually runs.

PR #1575, opened 2019-08-21
changelog bugs fixes #1574

### LINKED ISSUE #1574: Changelog Bugs
- Current implementation doesn't handle case where there are no changes in variants compare to the previous log. Current behvaiour: shows logs with prev variants value as none.
Screen Shot 2019-08-20 at 2 29 43 PM

- Reviews are being tracked in the change log, so pretty status for reviews aren't displayed
Screen Shot 2019-08-20 at 2 32 37 PM

-Old status in changelogs aren't consistently stored, which results in weird behaviour where "created experiment" is seen rather than edited experiment
Screen Shot 2019-08-20 at 2 30 24 PM

tiftran:


app/experimenter/experiments/forms.py
```diff
@@ -86,17 +86,15 @@ def save(self, *args, **kwargs):
 
         # account for changes in variant values
         if latest_change:
-            if self.variants_has_changed(
-                latest_change.new_values, self.new_serialized_vals
+            old_status = latest_change.new_status
```
jaredkerim: I love the changes in this file, much simpler and cleaner 🙆 🙆 🙆

PR #1583, opened 2019-08-22
Remove accept/reject API fixes #1430

### LINKED ISSUE #1430: Remove accept/reject API
The original workflow for experimenter involved delivery console calling into the Experimenter API to accept/reject experiments, but that is no longer the case so we should remove those API endpoints:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/api_urls.py#L15-L24

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/api_views.py#L32-L74

jaredkerim:


jaredkerim:
I ALWAYS forget the README.  Thanks @tiftran !

PR #1603, opened 2019-08-27
changelog refactoring and display names

tiftran:
Screen Shot 2019-08-26 at 8 47 59 PM

I believe the data migration should work, but put the removal of the new_values and old_values, if we maybe want to do a two stage roll out of the data migration. In the case that the data migration goes funny?


app/experimenter/experiments/models.py
```diff
@@ -922,27 +922,29 @@ def current_datetime():
     )
     message = models.TextField(blank=True, null=True)
 
-    old_values = JSONField(encoder=DjangoJSONEncoder, blank=True, null=True)
-    new_values = JSONField(encoder=DjangoJSONEncoder, blank=True, null=True)
+    changed_vals = JSONField(encoder=DjangoJSONEncoder, blank=True, null=True)
```
jaredkerim: I would call this `changed_values`....

PR #1610, opened 2019-08-29
Add mypy?

jaredkerim:
I'm exploring whether we should add mypy and static typing to Experimenter, I'm leaving this PR open for people interested in tracking it, I'm just doing this in my free time for now as a way to learn the tool.


jrbenny35:
I like mypy so I am ```+1```

jaredkerim:
Here's an example from @tiftran that would've been caught by `mypy`:

https://github.com/mozilla/experimenter/pull/1633/files#diff-3c07b241c0540355205fb823789bd402R215

jaredkerim:
Actually this https://github.com/mozilla/experimenter/pull/1632/files#diff-3c07b241c0540355205fb823789bd402R162 would've been caught by mypy as well because we would've had to explicitly declare a type for the argument that is serializable.

jaredkerim:
I think I'm gonna close this for now, it's still here in case I want to refer back to it but it doesn't look like we're gonna add mypy to this project anytime soon.

PR #1619, opened 2019-09-03
Add 3 branch tests.

jrbenny35:
```test_add_branch```: Test adding a new branch.
```test_remove_branch```: Test removing a branch.
```test_duplicate_branch_name```: Test adding a branch with the same name as the control branch.

### CODE REVIEW

File: tests/integration/conftest.py [hunk 1]
```diff
@@ -13,17 +13,18 @@ def capabilities(capabilities):
 @pytest.fixture
 def firefox_options(firefox_options):
     """Set Firefox Options."""
-    firefox_options.headless = True
+    # firefox_options.headless = True
```
jaredkerim: Why is this commented out?

tests/integration/conftest.py
```diff
@@ -13,17 +13,18 @@ def capabilities(capabilities):
...
     return firefox_options
 
 
-@pytest.fixture(scope='session', autouse=True)
+@pytest.fixture(scope="session", autouse=True)
```
jaredkerim: Oh interesting, I'm curious about this.  What does this do?

tests/integration/conftest.py
jrbenny35: Good catch

tests/integration/conftest.py
jrbenny35: This is so this fixture only runs once per test session. And ```autouse``` means it is used automatically. So this fixture runs automatically once per test session.

jrbenny35: Unless you are talking about the quotes, then that is just a python black package change.

PR #1620, opened 2019-09-03
Move normandy client to normandy app fixes #1597

### LINKED ISSUE #1597: Move Normandy specific code to a Normandy app
Let's make a new app called Normandy and move the relevant client code and tests there.

jaredkerim:
In an effort to split things up into more modules, this is the first step to moving normandy stuff into its own app.  Later we'll make some models in there for recipes and stuff.


app/pytest.ini
```diff
@@ -4,8 +4,3 @@ DJANGO_SETTINGS_MODULE = experimenter.settings
 # -- recommended but optional:
 addopts = -p no:warnings
 python_files = tests.py test_*.py *_tests.py
-testpaths =
```
jaredkerim: I don't like having to update this file everytime I move tests around.

PR #1622, opened 2019-09-03
added ds bugzilla updates fixes #1482

### LINKED ISSUE #1482: Update Data Science bug with experiment statuses
We write to the experimenter bug at certain points - can we send information to the data science bug as well.  

Emily brought this idea up - need to thing through avoiding making an overly chatty systems - but also some things would be nice. 

Example scenario - they have a DS bug.  that bug launches v1 of an experiment.  the experiment fails for whatever reason and they need v2.  There isn't a great correlation between DS bug and experimenter ticket to be able to track down v2 (unless you know it happened).

also no search for bug numbers (any associated with the experimenter) in experimenter search - would be a huge help if that is simple.

tiftran:


app/experimenter/experiments/tests/test_tasks.py
```diff
@@ -131,6 +131,79 @@ def test_bugzilla_error_creates_error_notification(self):
...
+            Experiment.STATUS_SHIP
+        )
+        self.experiment.save()
+
+    def test_ds_bugs_successfully_updates_with_comment(self):
+        self.assertEqual(Notification.objects.count(), 0)
+
+        tasks.update_ds_bug_task(self.experiment.id)
+
+        self.mock_bugzilla_requests_post.assert_called()
```
jaredkerim: Do you wanna check the parameters here using `assert_called_with`?  Looks like we do that some places and not others?  We don't have to, just curious.

tiftran: @jaredkerim would probably be best `assert_called()` were replaced with `assert_called_with` created https://github.com/mozilla/experimenter/issues/1628 to cover that as `assert_called()` has been used outside of this PR

PR #1644, opened 2019-09-06
Use separate docker stack for integration tests fixes #1641

### LINKED ISSUE #1641: Integration tests failing locally
Right now on multiple peoples machines the integration tests are failing, but passing in circle.  My guess is it's related to the Experimenter server instance being persistent across multiple runs of the integration tests, so we need to make sure it builds a completely fresh copy each time.

jaredkerim:
Rather than using the same docker stack both for local dev and for the integration tests (which makes it hard to share config and lead to tests failing on master), we should use a separately configured stack that doesn't collide with the local dev one.


.circleci/config.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
   integration_test:
     machine:
-      image: ubuntu-1604:201903-01
+      docker_layer_caching: true
```
jrbenny35: What image will this use?

.circleci/config.yml
```diff
@@ -59,9 +59,8 @@ jobs:
       - run:
           name: Run integration tests
           command: |
-            cp .env.sample .env
-            sudo chown -R $USER:$USER .
-            make integration_test_ci
+            sudo chown -R ubuntu tests/
```
jrbenny35: Any reason why you removed ```$USER:$USER```?

.env.integration
```diff
@@ -0,0 +1,36 @@
+BUGZILLA_API_KEY=
```
jrbenny35: Do we need another whole file? What was wrong with the ```cp``` command?

docker-compose.integration-test.yml
```diff
@@ -1,13 +1,103 @@
 version: "3"
 
 services:
+
```
jrbenny35: What is the idea behind this new yml file? It seems to copy the production yaml pretty closely.

jrbenny35:
Have you seen problems with using the local dev stack with the integration tests?

.circleci/config.yml
jaredkerim: It uses the default circle machine image which is what we use on our other tests, all we expect it to provide is docker.  Is there something special we need from that particular image?

.circleci/config.yml
jaredkerim: Yeah it looked to me like that set it to `circleci` user, but when the container tries to write out to the integration test folder on the host machine its already owned by the `ubuntu` user?  You previously had the container execute some chowns to `seluser` which is `1001` which maps to `ubuntu` on the host, so this seemed to accomplish the same thing?

.env.integration
jaredkerim: Yeah so the specific problem this is trying to solve is that in the tests, one of the fields that gets filled is the data science bugzilla url, and because we strongly validate that the URL provided matches the host name in the settings, in the test it's written as `http://bugzilla.com/whatever` so we need `http://bugzilla.com/` to appear as the BUGZILLA_HOST in the settings, so this way we can maintain a separate environment file in case we end up having to hardcode any other specific settings for it (like email domains or something if we ever start to include those in the tests).

docker-compose.integration-test.yml
jaredkerim: The idea of this PR is to create a separate full local experimenter stack specifically just for the integration tests rather than use the dev environment one.  The problem we ran into was on someones local environment, their dev environment might have some arbitrary settings (including the BUGZILLA_HOST) that don't match what the tests expect, so even on a 'clean' master checkout, the tests were failing.  So I changed it to make its own separate isolated experimenter environment which mirrors the way prod is setup, except without a couple things it doesn't need specifically for development (like shared volumes) and in its own isolated namespace `experimenter_integration` so that it doesn't collide with the dev environment.  This way the state of your dev environment shouldn't affect whether the integration tests pass.

.circleci/config.yml
jrbenny35: What I had was probably unnecessary too

.circleci/config.yml
jrbenny35: The default image is really old, but we can watch and see if it causes any problems.

PR #1645, opened 2019-09-06
reverted char reduced limit and truncated exp name for bugzilla

tiftran:


app/experimenter/experiments/bugzilla.py
```diff
@@ -155,6 +158,18 @@ def format_creation_bug_body(experiment, extra_fields):
     return bug_data
 
 
+def format_summary(experiment):
+    experiment_string = str(experiment)
+
+    if len(experiment_string) > EXPERIMENT_NAME_MAX_LEN:
```
jaredkerim: Do you need this check, can't you just always truncate and if it's less then it's basically a no-op?

app/experimenter/experiments/bugzilla.py
```diff
@@ -155,6 +158,18 @@ def format_creation_bug_body(experiment, extra_fields):
...
+        truncated_name = experiment_string[0:EXPERIMENT_NAME_MAX_LEN]
+
+        return "[Experiment]: {experiment}...".format(
```
jaredkerim: So this will always be 167 characters for any name longer than 150 characters, and bugzilla max is 255, does that gap matter?

app/experimenter/experiments/bugzilla.py
```diff
@@ -155,6 +158,18 @@ def format_creation_bug_body(experiment, extra_fields):
     return bug_data
 
 
+def format_summary(experiment):
+    experiment_string = str(experiment)
```
jaredkerim: Should you use the name field explicitly?

app/experimenter/experiments/bugzilla.py
tiftran: well after an experiment goes to shipped FX version and channel are appended to the bugzilla summary. After accounting for that we could allow the cap to be ~200

jaredkerim: Oh right I see those things get tacked onto the end so you're leaving space.  Awesome sounds good 👍

PR #1676, opened 2019-09-13
added docker-compose to include norm.and delivery console fixes #1459

### LINKED ISSUE #1459: Add a docker compose file that includes delivery console, normandy
Now that experimenter strongly interacts with delivery console and normandy, it's becoming increasingly necessary to setup an end to end stack for local testing.  We should be able to add a docker compose file that includes delivery console and normandy so that something like `make up-all` also starts and configures them all to talk to each other.

tiftran:
dc: dockerfile PR https://github.com/mozilla/delivery-console/pull/1049


docker-compose-full.yml
```diff
@@ -0,0 +1,129 @@
...
+
+services:
+  app:
+    image: app:build
+    env_file: .env
+    stdin_open: true
+    tty: true
+    links:
+      - db
+      - redis
```
jaredkerim: How is experimenter gonna query normandy's API?

docker-compose-full.yml
```diff
@@ -0,0 +1,129 @@
+version: "3"
+
+services:
+  app:
```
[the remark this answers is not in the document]
jaredkerim: Nevermind me that would be too nuts of a change

docker-compose-full.yml
jaredkerim: Nevermind me I'm dumb the worker calls it 👍

PR #1692, opened 2019-09-22
changelog stops creating logs when no change fixes #1670

### LINKED ISSUE #1670: Multiple 'Created Experiment' changelogs in history
In this experiment we see many 'Created Experiment' entries in the history, that doesn't seem right:

https://experimenter.services.mozilla.com/experiments/activity-stream-pkt-new-tab-release69-layout-holdback-1573312/

tiftran:
Issue of multiple "Created Experiment" logs was due to `old_status` not being set. 
Screen Shot 2019-09-22 at 10 53 01 AM
This should have been fixed by https://github.com/mozilla/experimenter/pull/1575/files#diff-b75d2c1707450c0ad27a64a54719a0bbR89. Which was pushed into production ~Sept 10th, which would mean prior experiments would have the weird/duplicate "Created Experiments" changelogs

Additionally change logs are being created while no change was made to the experiment. Current PR changes are made to address that.

In addition would we like to have a data migration of some sort to prune the duplicate "Created Experiment" changelogs?


app/experimenter/experiments/forms.py
```diff
@@ -198,6 +198,15 @@ def _get_display_name(self, field):
             return self.fields[field].label
         return field.replace("_", " ").title()
 
+    def _has_changed(self, old_status, changed_values, experiment):
+        return any(
+            [
+                changed_values != {},
```
jaredkerim: `{}` evaluates to False

app/experimenter/experiments/forms.py
```diff
@@ -198,6 +198,15 @@ def _get_display_name(self, field):
...
+                self.get_changelog_message() != "",
```
jaredkerim: `""` evaluates to False

app/experimenter/experiments/forms.py
```diff
@@ -198,6 +198,15 @@ def _get_display_name(self, field):
             return self.fields[field].label
         return field.replace("_", " ").title()
 
+    def _has_changed(self, old_status, changed_values, experiment):
+        return any(
```
jaredkerim: Between things evaluating to False and using `or` instead of `any([])` I think this should fit on one line?

mythmon:
Migration tests are pretty rare in Django-land in my experience. Because of that, I have found it difficult to find docs or examples on how to write them. To kickstart the process here, I'd like to point you towards [some migrations tests in Normandy](https://github.com/mozilla/normandy/blob/master/normandy/recipes/tests/test_migrations.py) that have served use well.

PR #1708, opened 2019-09-30
Move values below headers on detail view fixes #1669

### LINKED ISSUE #1669: Move detail page values below headers
Rather than beside the header, I think most other fields go directly below the header, so for ones that still go beside the header we should move it to below.  Example of one to be changed:

Screenshot 2019-09-12 12 54 56

uhlissuh:
Screen Shot 2019-09-30 at 2 01 34 PM

 there were only two fields, both inside normandy section.


uhlissuh:
@tiftran yes you were right! Thank you!!!

PR #1722, opened 2019-10-03
Reactify design page fixes #1586

### LINKED ISSUE #1586: Port branch page to react
In preparation for new intake types, we should rewrite the branch page UI logic from jQuery based to React based so that it will be easier to add dynamic form behaviours and unit test them.

uhlissuh:
OK. Here's a work in progress. There might be some ugly unfactored code in here... 😰 

What I think mostly works:
- all types (generic, pref flip, addon) api endpoints, serializers.
- the react stuff for these forms, showing the error messages... visually looking like the old form

The known-knowns of what's not working yet:
- i have been assuming throughout working on this that we are switching to a new type of approach for what fields are "optional" or "required"... so, I've been assuming that we want to allow the form to be saved with technically _no_ fields filled out. I've got a problem with the serializer for `ratio` on branch/variant/whatever where I am using a `IntegerField` and I can't figure out how to let it be optional... to allow the user to leave the ratio (also referred to as branch size) empty. You can see I passed in about every potential property into that thing to get it to be, essentially, `required=false`. It's not figured out yet. 
- the react is probably wacky in places???
- you will see the form flash for a second, as it figures out what type of form to show for what type of experiment this is. Rehan told me how to fix this and I haven't done it yet, but I plan to remove that temporary flash as it loads. 
- i am missing some font awesome symbols here and there... the spacing might not be perfectly identical yet. 
- there are no tests yet, I'm working on it. 
- the help links do nothing
- other various and assorted bugs
___________________________________________________________________
I would put more screenshots but it kinda just looks like the old form 😄 
BUT I GOT GIFS. 

Screen Shot 2019-10-02 at 10 53 18 PM

prefflipreactdemo

prefflipreactdemo1

prefflipreactdemo2


app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+        description: "",
+        value: "",
+        is_control: false,
+      }],
+      errors: {}
+    };
+  }
+
+  componentDidMount() {
+    const that = this;
```
jaredkerim: I'm not a JS expert but I think if you use arrow functions `(arg) => { body }` then it uses `this` from the parent scope, and then you don't need this alias? @rehandalal ?

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+
+    let url;
+    if (this.state.type == "pref") {
+      url = `/api/v1/experiments/${this.slug}/design-pref`
```
jaredkerim: These URLs should have trailing slashes.

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+        } else {
+          return that.setState({
+            type: json.type,
+            pref_key: json.pref_key,
+            pref_type: json.pref_type,
+            pref_branch: json.pref_branch,
+            addon_experiment_id: json.addon_experiment_id,
+            addon_release_url:json.addon_release_url,
+            design: json.design,
+            variants: json.variants
```
jaredkerim: If the only difference between these two cases is this one field then can you just create the object once and optionally add this field before you pass it to `setState`?

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+
+  updateValue(e) {
+    var stateCopy = { ...this.state.variants };
+    stateCopy[e.target.dataset.index].value = e.target.value;
+
+    this.setState(stateCopy);
+  }
+
+  handleInputChange(e) {
+    if (e.target.name == "pref-name") {
```
jaredkerim: Alternatively could you create an object to relate these fields to their paired state key names and then just to the look up and set once?

app/experimenter/static/js/design-page.js
jaredkerim: Also we could probably just do this if statement to find the correct URL once and store that on the component at init time and then just refer to that everywhere?

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+    const res = await fetch(url, {
+      method: "PUT",
+      body: JSON.stringify(this.state),
+      headers: {
+        "Content-Type": "application/json"
+      }
+    });
+
+    if (res.status == "200") {
+      location.replace(`/experiments/${this.slug}`)
```
jaredkerim: Needs a trailing slash

jaredkerim:
This is so cool @uhlissuh I am so stoked for this! 🎉 🎉 🎉 🎉 🎉

app/experimenter/static/js/design-page.js
rehandalal: good catch, yes arrow functions do not have their own binding to `this` so it ends up using `this` from parent scope. (see: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions))

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+  Container,
+  Row,
+  Col,
+  FormControl,
+  FormLabel
+} from "react-bootstrap";
+
+const branchesDiv = document.getElementById("react-branches-form");
+
+function PrefForm(props) {
```
rehandalal: I also forgot to mention all these form components should go in their own JS files.

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+                  className="mb-4"
+                  onClick={this.addBranch}
+                >
+                  + Add Branch
+                </Button>
+              </Col>
+            </Row>
+            <Row>
+              <Col className="text-right">
+                <a className="mr-1">X Cancel Editing</a>
```
rehandalal: nit: use `&times;` instead of X

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+                    )}
+                  </Col>
+                  <Col md={5} className="text-right">
+                    {index != 0 ? (
+                      <Button
+                        variant="danger"
+                        data-index={index}
+                        onClick={this.removeBranch}
+                      >
+                        X Remove Branch
```
rehandalal: nit: use `&times;` instead of X

app/experimenter/static/js/design-page.js
rehandalal: spread operator is your friend

app/experimenter/static/js/design-page.js
rehandalal: > Also we could probably just do this if statement to find the correct URL once and store that on the component at init time and then just refer to that everywhere?

given that this is related to state and state is not a constant we should not do that. instead maybe refactoring this into a helper function like:
```js
getApiUrl() {
  if (["pref", "addon"].includes(this.state.type)) {
    return `/api/v1/experiments/${this.slug}/design-${this.state.type}/`;
  }
  return `/api/v1/experiments/${this.slug}/design-generic/`;
}
```

app/experimenter/static/js/design-page.js
uhlissuh: I can just get rid of all of this repetition by setting the state to the json just like... 

`  return that.setState(json)`

this one line is a huge improvement on like 30 repeating ones.

uhlissuh: (and fixing the that issue also).

uhlissuh:
@rehandalal and @jaredkerim .... I am rerequesting review even though the integration tests are failing... it's going to take me a minute to fix those and this is a pretty big PR to review so... getting the ball re-rolling...

app/experimenter/static/js/design-page.js
uhlissuh: huhmmm?

app/experimenter/static/js/design-page.js
uhlissuh: uh, I just removed the `this = that` thing and all was fine so... 🤷‍♀

app/experimenter/static/js/design-page.js
```diff
@@ -0,0 +1,494 @@
...
+            pref_type: json.pref_type,
+            pref_branch: json.pref_branch,
+            addon_experiment_id: json.addon_experiment_id,
+            addon_release_url:json.addon_release_url,
+            design: json.design,
+            variants: json.variants
+          });
+        }
+      });
+  }
```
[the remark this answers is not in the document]
uhlissuh: I couldn't see how pulling out a method here helped that much.

uhlissuh:
note to @jaredkerim , i know there is duplication in the serializer tests... some stuff that could be moved in to helper functions that get called across all the serializer testcases... i am probably going to be doing that this afternoon. 

I am not removing the django forms and formset associated with the old implementation... was kind of thinking I could leave them in place and remove them in another pr, since this one is so big already.

uhlissuh:
@rehandalal and @jaredkerim, I think I addressed the concerns you brought up.

uhlissuh:
Screen Shot 2019-10-22 at 4 06 33 PM

 problem 1 is solved.

uhlissuh:
Screen Shot 2019-10-22 at 4 10 55 PM
@jaredkerim slugify prob solved.

uhlissuh:
I've addressed the feedback I could reproduce, but I can't reproduce the control branch ordering problem. Over here it's always ordered correctly.

Issue #1723, opened 2019-10-03
Indicate completed fields in all forms


jaredlockhart:
Now that we're switching to all fields being 'optional' at form save time but 'required' before an experiment can launch, I've marked fields as 'required' or 'optional', but it would probably also be helpful to indicate on each field whether its been completed or not.  I was going to do it as part of #1686 but then I realized that @uhlissuh is doing the react form refactors and it's probably better to wait until that lands before we change too many more behaviours of the forms so I'm filling this separately.  Basically I'm picturing beside every field something like

Field not completed:

Required ❌ 

Field completed:

Required ✅

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-98)

jaredlockhart:
Highlight the field in yellow (warning) colour

jaredlockhart:
I have a guess:

Maybe in the input inline we can say

if mark_required && value=None && experimernt.id: highlight

PR #1724, opened 2019-10-03
Make all form fields optional (but required to ship)

jaredkerim:
So this change finally makes all fields optional at form save time, but some fields required to ship an experiment.  Fields that are required are marked in red as `Required`.  I filed https://github.com/mozilla/experimenter/issues/1723 as a follow up after @uhlissuh lands her react stuff.  The required fields on the Overview page are actually required before the initial form can be saved but after that any fields can be filled out in any order.  Here's how it all looks:

Screenshot_2019-10-03 Mozilla Experimenter
Screenshot_2019-10-03 Fields are different
Screenshot_2019-10-03 Fields are different (1)
Screenshot_2019-10-03 Fields are different (2)
Screenshot_2019-10-03 Fields are different (3)
Screenshot_2019-10-03 Fields are different (4)
Screenshot_2019-10-03 Fields are different (5)


shell1:
- "active experiment name" in add-ons design should not be required (since that was a very confusing field and with new add-on capabilities we don't need that anymore.)  
- public name and description are required now (both pref flip and add-on)

jaredkerim:
Oh right and the other caveat is on the design page the branches are still required to save because they have to be validated together.

uhlissuh:
@jaredkerim oh that's an important piece of info for my current work..

uhlissuh:
@jaredkerim it makes sense, but I don't think I had considered that fully.

jaredkerim:
It occurs to me I should leave this for 2.42 because it introduces a lot of changes and I'd rather let the react stuff land and settle first.

PR #1725, opened 2019-10-03
Change analysis owner to foreign key fixes #1652

### LINKED ISSUE #1652: Change data science owner field to User foreign key
We should switch the Data Science Owner field to be an actual foreign key to User and figure out some intermediate way of backfilling the existing entries, using fuzzy string matching or something?

jaredkerim:
So this was a bit tricky to get working but I think I got it.

I tried to use `User.objects.filter(email__search=...)` to find the nearest user, but it complained I couldn't do that on an email field 🙄 so I had to use the built in Python string matcher:

https://docs.python.org/3.7/library/difflib.html#difflib.SequenceMatcher

I had to split the migration into three parts

1) add the new field without dropping the old one
2) populate it by looking up from the old field
3) drop the old field
4) rename the new one to the old one

It complained if the migration wasn't reversible so I added that.

Updated the forms and filters, so now you can filter experiments on the list page by data science owner (which was the point of this but I almost forgot to do it 😆 )

Edit: Oh and also the previous migration test by @tiftran was written using factories instead of `apps.get_model` so it completely broke as soon as there was a new migration, and rather than rewrite all of it I just opted to delete it since its usefulness was short lived but if it really seems to people that we should keep it around I could go back and update it.

Screenshot_2019-10-03 Mozilla Experimenter(1)
Screenshot_2019-10-03 Mozilla Experimenter(2)
Screenshot_2019-10-03 Mozilla Experimenter(3)


jaredkerim:
Okay I updated the integration tests, I changed a couple things around.  I set it up so now you can run `make integration_up` in one window, then `make integration_up_shell` in another, which gives you a shell in the firefox container but it's connected to all the running containers in the other `integration_up` window and you can run `tox -c tests/integration` in the shell to quickly re run all the tests and watch the `integration_up` shell to see them running.  

Also added `public_nw` to nginx in the integration up compose file so you can navigate to `https://localhost` and play around with the instance that the tests run against so you can inspect dom elements, etc. I had to add `localhost` to `ALLOWED_HOSTS` to be able to navigate to it though because the tests expect the hostname to be `nginx` and I couldn't figure out how to pass two hostnames in through environment variables as `compose run` arguments, maybe we can clean that up later.  

Still haven't figured out the VNC part yet. 

Also changed the create test user to use `manage.py createsuperuser` instead of just doing it through a python shell.

jrbenny35:
For VNC try: https://github.com/SeleniumHQ/docker-selenium#debugging

app/experimenter/experiments/migrations/0073_experiment_new_analysis_owner.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def forward_analysis_owner(apps, schema_editor):
+    Experiment = apps.get_model("experiments", "Experiment")
+    User = apps.get_model(*settings.AUTH_USER_MODEL.split("."))
+
+    for experiment in Experiment.objects.all():
+        experiment.new_analysis_owner = closest_user(
+            User.objects.all(), experiment.analysis_owner
+        )
```
tiftran: Sequencematcher is really cool!
Was playing around with it, it's case sensitive, would we want to lower analysis owner for better ratios?
```
>>> SM(None, "ttran@mozilla.com", "tif tran").ratio()
0.4
>>> SM(None, "ttran@mozilla.com", "Tif Tran").ratio()
0.24

```

jaredkerim: Oh good idea 👍

PR #1727, opened 2019-10-04
added experimenter slug to recipe serializer fixes #1705

### LINKED ISSUE #1705: Add experimenter slug to Normandy Recipe Serializer
When we serialize an experiment to Normandy we have to include the experimenter slug to be stored in Normandy.

tiftran:


tiftran:
@jaredkerim 🤦‍♀ my bads

app/experimenter/experiments/serializers.py
```diff
@@ -324,7 +324,7 @@ class ExperimentRecipeSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Experiment
-        fields = ("action_name", "name", "filter_object", "comment", "arguments")
+        fields = ("action_name", "name", "filter_object", "comment", "arguments", "slug")
```
jaredkerim: In the serializer we want it to be labeled as `experimenter_slug` but the field on the model is `slug` so you gotta do it like this:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/serializers.py#L291

PR #1734, opened 2019-10-07
Initial commit adding 5 tests for timeline and population.

jrbenny35:
Adds some tests for the first 5 fields on the timeline and population page.

### DIFF
--- a/tests/integration/pages/experiment_timeline_and_population.py
+++ b/tests/integration/pages/experiment_timeline_and_population.py
@@ -21,7 +21,8 @@ class TimelineAndPopulationPage(Base):
         self.wait.until(
             lambda _: "Experiment Timeline"
             in self.find_element(
-                By.CSS_SELECTOR, "div.form-group:nth-child(2) > h4:nth-child(1)"
+                By.CSS_SELECTOR,
+                "body.page-edit-timeline-and-population"
             ).text
         )
         return self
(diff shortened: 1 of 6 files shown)


tests/integration/pages/experiment_timeline_and_population.py
```diff
@@ -0,0 +1,87 @@
...
+    _proposed_duration_locator = (By.CSS_SELECTOR, "#id_proposed_duration")
+    _proposed_enrollment_locator = (By.CSS_SELECTOR, "#id_proposed_enrollment")
+    _proposed_start_date_locator = (By.CSS_SELECTOR, "#id_proposed_start_date")
+    _population_precentage_locator = (By.CSS_SELECTOR, "#id_population_percent")
+
+    def wait_for_page_to_load(self):
+        self.wait.until(
+            lambda _: "Experiment Timeline"
+            in self.find_element(
+                By.CSS_SELECTOR, "div.form-group:nth-child(2) > h4:nth-child(1)"
```
jaredkerim: You can add a block like this:

https://github.com/mozilla/experimenter/blob/0ad80eee4bffd5448f4fd3f37e0d46986de164c3/app/experimenter/templates/experiments/edit_base.html#L66

at the bottom here:

https://github.com/mozilla/experimenter/blob/0ad80eee4bffd5448f4fd3f37e0d46986de164c3/app/experimenter/templates/experiments/edit_timeline_population.html

and then just do the same thing we did here:

https://github.com/mozilla/experimenter/blob/master/tests/integration/pages/experiment_design.py#L33-L36

jrbenny35: This is a good idea, I'll update and then merge if everything passes

PR #1738, opened 2019-10-08
added branched addon serializer fixes #1729

### LINKED ISSUE #1729: Add Normandy Branched Addon Recipe
We can implement the Recipe Serializer for Branched Addon recipes without needing the UI for it because we can now use it for even the single addon case. So let's implement the new schema as a separate set of serializers, and then we'll follow a simple rule to determine which serializer to use:

Min Version >= 68: use BranchedAddonRecipeSerializer
Else: use SingleAddonRecipeSerializer

Multi addon recipe schema: https://normandy.cdn.mozilla.net/api/v3/action/8/

tiftran:


app/experimenter/experiments/models.py
```diff
@@ -630,6 +630,10 @@ def firefox_min_version_integer(self):
             ExperimentConstants.VERSION_REGEX.match(self.firefox_min_version).group(0)
         )
 
+    @property
+    def firefox_min_addon_serializer_version(self):
```
jaredkerim: Shouldn't this be a constant in constants.py?

app/experimenter/experiments/serializers.py
```diff
@@ -307,6 +314,22 @@ class Meta:
...
+    )
+    userFacingDescription = serializers.ReadOnlyField(source="public_description")
+    branches = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Experiment
+        fields = ("slug", "userFacingName", "userFacingDescription", "branches")
+
+    def get_branches(self, obj):
+        variants = obj.variants
```
jaredkerim: Couldn't you just do that in the next line? 🤔

app/experimenter/experiments/serializers.py
```diff
@@ -356,10 +379,16 @@ def get_filter_object(self, obj):
         return filter_objects
 
     def get_arguments(self, obj):
+        min_version = obj.firefox_min_version_integer
         if obj.is_pref_experiment:
             return ExperimentRecipePrefArgumentsSerializer(obj).data
-        elif obj.is_addon_experiment:
+        elif (
+            obj.is_addon_experiment
+            and min_version <= obj.firefox_min_addon_serializer_version
```
jaredkerim: How about `obj.is_branched_addon` method on the model.....? 🤔

jaredkerim: I imagine we'll need to do this check more than once....

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -437,6 +438,7 @@ def test_serializer_outputs_expected_schema_for_pref_experiment(self):
     def test_serializer_outputs_expected_schema_for_addon_experiment(self):
         experiment = ExperimentFactory.create_with_status(
             Experiment.STATUS_SHIP,
+            firefox_min_version="63.0",
```
jaredkerim: Very nice 👍

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -462,6 +464,17 @@ def test_serializer_outputs_expected_schema_for_addon_experiment(self):
...
 
+    def test_serializer_outputs_expect_schema_for_branched_addon(self):
+        experiment = ExperimentFactory.create(
+            firefox_min_version="70.0", type=Experiment.TYPE_ADDON
+        )
+        serializer = ExperimentRecipeSerializer(experiment)
+        self.assertEqual(serializer.data["action_name"], "opt-out-study")
+        self.assertEqual(
+            serializer.data["arguments"],
+            ExperimentRecipeBranchedArgumentsSerializer(experiment).data,
```
jaredkerim: If we don't test the actual individual fields anywhere we could just remove all the fields from the serializer and all the tests would still pass but it would break in prod?

app/experimenter/experiments/serializers.py
tiftran: oooh you're right, sounds good will do

app/experimenter/experiments/models.py
```diff
@@ -630,6 +630,13 @@ def firefox_min_version_integer(self):
             ExperimentConstants.VERSION_REGEX.match(self.firefox_min_version).group(0)
         )
 
+    @property
+    def is_branched_addon(self):
```
jaredkerim: Shouldn't this also check if it's an addon experiment? 🤔

jaredkerim: Also maybe add a couple little tests for this in `test_models`

jaredkerim: Fantastiche 🙆

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1058,6 +1058,22 @@ def test_firefox_min_version_integer_returns_correct_integer(self):
 
         self.assertEqual(experiment.firefox_min_version_integer, 57)
 
+    def test_is_branched_addon_returns_true_for_addon_and_greater_version(self):
+        experiment = ExperimentFactory(
+            type=Experiment.TYPE_ADDON, firefox_min_version="70.0"
+        )
+        self.assertTrue(experiment.is_branched_addon)
+
+    def test_is_branched_addon_when_exp_is_addon_but_lower_version(self):
```
jaredkerim: I usually like to put both the expected behaviour/output and the test conditions in the name, but it's also hard to fit in a single line.  So here I would just include something about expecting it to be false when the version is lower.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1058,6 +1058,22 @@ def test_firefox_min_version_integer_returns_correct_integer(self):
 
         self.assertEqual(experiment.firefox_min_version_integer, 57)
 
+    def test_is_branched_addon_returns_true_for_addon_and_greater_version(self):
```
jaredkerim: Parfait 💯

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1058,6 +1058,22 @@ def test_firefox_min_version_integer_returns_correct_integer(self):
...
+        )
+        self.assertTrue(experiment.is_branched_addon)
+
+    def test_is_branched_addon_when_exp_is_addon_but_lower_version(self):
+        experiment = ExperimentFactory(
+            type=Experiment.TYPE_ADDON, firefox_min_version="66.0"
+        )
+        self.assertFalse(experiment.is_branched_addon)
+
+    def test_is_branched_addon_returns_false_for_pref_type(self):
```
jaredkerim: Excellente 🐶

app/experimenter/experiments/tests/test_serializers.py
jaredkerim: I'd still like to see the fields for this serializer unpacked in this test.

tiftran:
@jaredkerim realized that action name should probably be different for branched addons... based on https://normandy.cdn.mozilla.net/api/v3/action/8/

jaredkerim:
As per @mythmon and @tiftran on slack:

every branch we want `extensionApiId: null`

tiftran:
@jaredkerim  extensionApiId field added

mythmon:
> As per @mythmon and @tiftran on slack:
> 
> every branch we want `extensionApiId: null`

For some background (in case this comes back to bite us). This is technically not valid with the schema of the action, becaues `extensionApiId` needs to be an integer not null. Having this here will make it easy to enter the correct values in the JSON while we are hand-updating the recipe to launch.

PR #1765, opened 2019-10-15
added results filter and badge

tiftran:
Screen Shot 2019-10-15 at 3 14 24 PM
Screen Shot 2019-10-15 at 3 15 35 PM


app/experimenter/templates/experiments/list.html
```diff
@@ -165,7 +168,7 @@ <h5>{{ experiment.population }}</h5>
     </div>
 
     {% for field in filter.form %}
-      {% if field.name == "in_qa" or field.name == "surveys" or field.name == "archived" or field.name == "subscribed" or field.name == "longrunning" or field.name == "is_paused" %}
+      {% if field.name == "in_qa" or field.name == "surveys" or field.name == "archived" or field.name == "subscribed" or field.name == "longrunning" or field.name == "is_paused" or field.name == "completed_results" %}
```
jaredkerim: This line is getting starting to get a little out of hand we should do something about that at some point.

tiftran: agreed

PR #1766, opened 2019-10-16
Migration 073 fails for equal string matches fixes #1757

### LINKED ISSUE #1757: Migration 073 breaks for Users with equal scores
From:

https://sentry.prod.mozaws.net/operations/experimenter-dev/issues/6440581/?query=server_name:%22experimenter-dev-experimenter-app-1-6fc7897b94-gc99p%22

If two users have the same string match ratio then it falls back to comparing the user objects, which can't be compared, so I gotta add a fallback.  Good thing we have that test.

jaredkerim:
If two users string match identically then `sorted` moves on to compare the next element, which in the original version was a `User` object which doesn't implement `__cmp__` (which I find silly it should just fall back to comparing db ids, or maybe there's a meta field for that?), so if I just stick `user.id` into the tuple it'll fall back to comparing those.

### DIFF
--- a/app/experimenter/experiments/migrations/0073_experiment_new_analysis_owner.py
+++ b/app/experimenter/experiments/migrations/0073_experiment_new_analysis_owner.py
@@ -11,11 +11,11 @@ def closest_user(users, target_name):
     if target_name:
         return sorted(
             [
-                (SM(None, user.email.lower(), target_name.lower()).ratio(), user)
+                (SM(None, user.email.lower(), target_name.lower()).ratio(), user.id, user)
                 for user in users
             ],
             reverse=True,
-        )[0][1]
+        )[0][2]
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/migrations/0073_experiment_new_analysis_owner.py
```diff
@@ -11,11 +11,11 @@ def closest_user(users, target_name):
     if target_name:
         return sorted(
             [
-                (SM(None, user.email.lower(), target_name.lower()).ratio(), user)
+                (SM(None, user.email.lower(), target_name.lower()).ratio(), user.id, user)
                 for user in users
```
tiftran: clever, wish I thought of that 🙂

PR #1772, opened 2019-10-18
include platform info in comment in recipe serializer fixes #1770

### LINKED ISSUE #1770: OS filters don't get exported to Delivery Console
OS filters (like "Windows only") don't get exported in the Normandy recipe that Delivery Console imports. Normandy doesn't support a filter object for OS, so this would need to go into the "notes" section.

tiftran:


app/experimenter/experiments/serializers.py
```diff
@@ -361,6 +361,11 @@ def get_arguments(self, obj):
         elif obj.is_addon_experiment:
             return ExperimentRecipeAddonArgumentsSerializer(obj).data
 
+    def get_comment(self, obj):
```
jaredkerim: Can you try to do this with an `f"{...}"` style string?

PR #1773, opened 2019-10-18
added multipref serializer fixes #1717

### LINKED ISSUE #1717: Add Normandy Multi Pref Recipe
We can implement the Recipe Serializer for Multi-Pref recipes without needing the UI for it because we can now use it for even the single recipe case.  So let's implement the new schema as a separate set of serializers, and then we'll follow a simple rule to determine which serializer to use:

Min Version >= 68: use MultiPrefRecipeSerializer
Else: use SinglePrefRecipeSerializer

Multi pref recipe schema: https://normandy.cdn.mozilla.net/api/v3/action/9/

tiftran:


jaredkerim:
I just got this when I tried it:

Screenshot 2019-11-01 16 49 05

PR #1815, opened 2019-10-25
added dummy variants to designserializers fixes #1787

### LINKED ISSUE #1787: Move default state to backend serializers

tiftran:


app/experimenter/experiments/serializers.py
```diff
@@ -35,6 +35,22 @@ def to_representation(self, obj):
...
 
+class VariantsListSerializer(serializers.ListSerializer):
+
+    def to_representation(self, data):
+        data = super().to_representation(data)
+        if data == []:
+            blank_variant = {}
+            for field in self.child.fields:
+                blank_variant[field] = None
+                control_blank_variant = blank_variant.copy()
```
jaredkerim: Heh, I see what you did here and that's clever but it also wasn't obvious at first glance, just create an empty one for each branch and explicitly set the field on each one ;)

app/experimenter/experiments/serializers.py
```diff
@@ -35,6 +35,22 @@ def to_representation(self, obj):
...
+                blank_variant["is_control"] = False
```
jaredkerim: If this is inside the loop doesn't it do the same thing on every iteration?

tiftran: oooh some bad indenting.... 🙁

PR #1819, opened 2019-10-27
save and continues redirects to obj and analysis fixes #1816

### LINKED ISSUE #1816: Design save and continue redirects to detail page
Save and Continue on the design page should redirect to the next form page, 4. Objectives & Analysis

tiftran:
not sure if this was the right/best way to go about this....

### DIFF
--- a/app/experimenter/static/js/components/DesignForm.js
+++ b/app/experimenter/static/js/components/DesignForm.js
@@ -125,16 +125,6 @@ export default class DesignForm extends React.Component {
 
   }
 
-  @boundMethod
-  setContinue(e){
-    this.setState({continue:true})
-  }
-
-  @boundMethod
-  setNotToContinue(e){
-    this.setState({continue:false})
-  }
-
   render() {
     const dataExists = this.state.values.loaded;
     if (!dataExists) {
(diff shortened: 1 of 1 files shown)


app/experimenter/static/js/designForm.js
```diff
@@ -107,13 +108,29 @@ export default class DesignForm extends React.Component {
...
     if (res.status == "200") {
-      location.replace(`/experiments/${this.props.slug}/`);
+      if (this.state.continue){
+        location.replace(`/experiments/${this.props.slug}/edit-objectives/`);
+      }
+      else{
+        location.replace(`/experiments/${this.props.slug}/`);
+      }
+      
     }
```
tiftran: Just thinking.... would it better for me to reset continue back to false here?

jaredkerim: Well, try to always remember my personal golden rule of react: if you can avoid putting something in state, don't put it in state.  When the user clicks either `Save` or `Save and Continue` and from that point on it's very unidirectional, we call a method, make a fetch, and if it's valid then we redirect.  So do we need to put it in state at all?  Could we just have two separate methods for submitting?

PR #1853, opened 2019-10-30
Add JS linting and code formatting fixes #1794

### LINKED ISSUE #1794: Add linting, auto formatting

uhlissuh:
JS linting and JS formatting each have their own make command now. added the code formatting to `make code_format`.

I used Jenny's eslintrc.js file for the most part, so if we want different rules we can talk about that. 

I haven't added the linting check to `make check` yet, because there are like 200 linting errors... and if we add it to make check then this pr would fail without fixing all the linting errors in the same PR... right?


app/experimenter/static/js/components/BranchManager.js
```diff
@@ -30,20 +30,21 @@ export default class BranchManager extends React.Component {
...
-    this.setState(this.state);
+    this.setState(stateCopy);
   }
 
   render() {
     return (
       <div>
         {Object.keys(this.state.variants).map((variant_id, index) => (
-          <div>
+          <div key={index}>
```
uhlissuh: I needed to add this key prop here for the linter. 🤷‍♀

uhlissuh:
@rehandalal @jaredkerim, Rehan requested that I only submit the parts of code that added the linting, but not actually fix the linting errors in this PR. You had requested differently previously, but.. Rehan is PR overlord this week while you are out, so I am going with this approach of only adding the linting but not  fixing the errors or enforcing it in existing `make code_format` and `make check`. This further makes sense because all the linting fixes that were in this PR were shaken up quite a bit by landing Rehan's restructuring PR. 

Rehan, other things we talked about for fixes here were unpinning the version types, which I believe I have successfully done. The yarn lock file was regenerated anew by removing the file and then running yarn install (inside the container). We also talked about disabling the escaping rule, which is apparently called `no-useless-escape`... so I added the removal of that rule to the eslintrc file.

PR #1859, opened 2019-10-31
added jest/react-testing-library setup fixes #1789

### LINKED ISSUE #1789: Add react tests

tiftran:
setting up jest/enzyme took longer than anticipated... thought it would be best to have a PR up before completing testing a component so that it doesn't block anyone while I try to figure out enzyme


app/experimenter/static/js/tests/Error.test.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import {render} from "@testing-library/react";
+import '@testing-library/jest-dom/extend-expect';
+import Error from "experimenter/components/Error";
+
+
+describe("<Errors>",()=>{
```
rehandalal: this is generally the right approach, just a couple little tweaks:
```js
describe("The `Error` component", () => {
    const error = "an error occured!";
    it("renders the error message",()=>{
        const { getByText } = render(<Error error={error} />);
        expect(getByText(error)).toBeInTheDocument();
    });
})
```

Just a small note on the changes:
- `"<Errors>"` doesn't really make it clear what is being tested since `Errors` is not an actual component. Also on that note typically the string for the `it` method should be something that completes the sentence from the phrase in `describe`.
- I like to be explicit about the props that are being passed to the component so I went with `<Error error={error} />` instead of using a spread operator expand error props.
- I re-used the `error` variable in the `expect` method to keep things more DRY.

Makefile
```diff
@@ -23,6 +23,8 @@ test: test_build
 test-watch: compose_build
 	docker-compose -f docker-compose-test.yml run app sh -c "/app/bin/wait-for-it.sh db:5432 -- ptw -- --testmon --show-capture=no --disable-warnings"
 
+test-react:
```
rehandalal: nit: i think this should be `test_js` or at the very least `test_react`. I'm not sure why `test-watch` isn't also an underscore but it seems to be underscores everywhere else

app/package.json
```diff
@@ -24,19 +42,26 @@
     "parcel-bundler": "1.12.3",
     "popper.js": "1.15.0",
     "prettier": "1.18.2",
-    "react": "16.8.0",
-    "react-bootstrap": "1.0.0-beta.11",
-    "react-dom": "16.8.0"
+    "react": ">=16.8.0",
+    "react-bootstrap": "^1.0.0-beta.11",
+    "react-dom": ">=16.8.0",
+    "testing-library": "^0.0.2"
```
rehandalal: do we need this in package.json? seems like a sub-dependency.

app/package.json
```diff
@@ -24,19 +42,26 @@
...
   "devDependencies": {
     "@babel/core": "7.0.0",
-    "@babel/preset-react": "7.6.3"
+    "@babel/preset-react": "7.6.3",
+    "@testing-library/jest-dom": "^4.2.3",
+    "@testing-library/react": "^9.3.1",
+    "babel-jest": "^24.9.0",
+    "jest": "^24.9.0",
+    "jest-dom": "^4.0.0",
+    "react-testing-library": "^8.0.1"
```
rehandalal: this also seems like a package that is not necessary in `package.json`.

app/package.json
```diff
@@ -24,19 +42,26 @@
...
+    "react": ">=16.8.0",
+    "react-bootstrap": "^1.0.0-beta.11",
+    "react-dom": ">=16.8.0",
+    "testing-library": "^0.0.2"
   },
   "devDependencies": {
     "@babel/core": "7.0.0",
-    "@babel/preset-react": "7.6.3"
+    "@babel/preset-react": "7.6.3",
+    "@testing-library/jest-dom": "^4.2.3",
```
rehandalal: please make sure to pin these versions. i'm also guessing you're manually adding these to the file. you should ideally be using:
```
$ yarn add <package_name> -E
```

app/package.json
tiftran: eek! lemme fix all this and lint...

tiftran:
@rehandalal, @jaredkerim linted and package.json funniness should be fixed up now...
Have added some more tests... i know i probably did something funny/weird with them 🙈

app/experimenter/static/js/tests/Branch.test.js
```diff
@@ -0,0 +1,53 @@
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import AddonForm from "experimenter/components/AddonForm";
+
+describe("The `AddonForm` component", () => {
+  const setup = () => {
+    const data = {
```
rehandalal: i'd love to see these fixtures replaced by factories. here's how i did that in delivery console:
- Generic factory implementation: https://github.com/mozilla/delivery-console/blob/master/src/tests/factory.js
- Shaped object factory implementation: https://github.com/mozilla/delivery-console/blob/master/src/tests/state/actions/index.js#L10
- Factory usage: https://github.com/mozilla/delivery-console/blob/master/src/tests/state/actions/selectors.test.js#L8

tiftran:
still kind of rough.... wasn't able to fully get lazy loading? for the factories

app/experimenter/static/js/tests/DesignForm.test.js
```diff
@@ -0,0 +1,115 @@
+import React from "react";
+import { render, cleanup, fireEvent } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
+import DesignForm from "experimenter/components/DesignForm";
+import * as Api from "experimenter/utils/api";
+import { makeApiRequest } from "../utils/api";
```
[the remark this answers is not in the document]
tiftran: gah.... wasn't able to figure out how to do it without the imports to mock... was wondering if I could get guidance towards how to go about that?

app/experimenter/static/js/tests/DesignForm.test.js
```diff
@@ -0,0 +1,267 @@
+import React from "react";
```
jaredkerim: Okay so I just had a long conversation with @rehandalal about this, and I think I understand what's going on here.

What @rehandalal is suggesting is that, bare minimum, we start by testing the actual UI that users interact with.  So in that context that means specifically AddonForm, PrefForm, GenericForm, RolloutForm.

So I think what we want is four test files, one for each of those, that only sets up and instantiates DesignForm in one of those modes, and then tests all the interactions with it that we care about, just like you've done here.

So rename this file to AddonForm.test.js, move the render test you had in `Branch.test.js` into here, and then we can land this as the AddonForm tests and do the other 3 as follow ups.

app/experimenter/static/js/components/AddonForm.js
```diff
@@ -21,7 +21,7 @@ export default class AddonForm extends React.PureComponent {
   renderSingleAddonFields() {
     if (!this.props.data.get("is_branched_addon")) {
       return (
-        <React.Fragment>
+        <div data-testid="addonUrl">
```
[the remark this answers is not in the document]
jaredlockhart: Looks like using test ids is the most idiomatic way to do things with this testing library so let's keep it this way.

PR #1868, opened 2019-10-31
Reject ratios of 0 or above 100 fixes #1863

### LINKED ISSUE #1863: Branch size must be between 1-100
You can currently save two branches with sizes 0 and 100 and it'll validate, branch size should be between 1-100.

uhlissuh:
this check happens before the ratio adding to 100 check, because it happens on the base serializer

Screen Shot 2019-10-31 at 11 26 20 AM
Screen Shot 2019-10-31 at 11 26 32 AM
Screen Shot 2019-10-31 at 11 26 57 AM

### DIFF
--- a/app/experimenter/experiments/tests/test_serializers.py
+++ b/app/experimenter/experiments/tests/test_serializers.py
@@ -745,6 +745,22 @@ class TestExperimentDesignBaseSerializer(TestCase):
         self.assertFalse(serializer.is_valid())
         self.assertIn("variants", serializer.errors)
 
+    def test_serializer_rejects_ratios_0_or_100(self):
+        experiment = ExperimentFactory.create(type=ExperimentConstants.TYPE_ADDON)
+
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -745,6 +745,22 @@ def test_serializer_rejects_ratio_not_100(self):
         self.assertFalse(serializer.is_valid())
         self.assertIn("variants", serializer.errors)
 
+    def test_serializer_rejects_ratios_0_or_above_100(self):
+        experiment = ExperimentFactory.create(type=ExperimentConstants.TYPE_ADDON)
+
+        self.control_variant_data["ratio"] = 0
+        self.treatment_variant_data["ratio"] = 100
```
jaredkerim: I think the branch size on 752 doesn't match the name of the test?

PR #1884, opened 2019-11-03
changed validation to not allow special characters fixes #1879

### LINKED ISSUE #1879: Valid branch name allows a separate invalid branch name
From romartin on slack:

hey, the special characters in name bug seems to be fixed if you have only Control Branch in Generic Experiment. If you add another branch and name that one with only special characters, the FE error is displayed for both, but as long as you have the branch name with a valid name, the control branch name can be special characters only

tiftran:
since `slugify()` removes anything that isn't alphanumerics, underscores, hyphens, or whitespace, added validation to check if name only contain alphanumeric characters and spaces since the error message says no special characters. Not sure if this is too restrictive and we should be allowing underscores, hyphens and whitespaces as well.

### DIFF
--- a/app/experimenter/experiments/serializers.py
+++ b/app/experimenter/experiments/serializers.py
@@ -1,6 +1,5 @@
 import time
 import json
-import re
 from rest_framework import serializers
 from django.utils.text import slugify
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/serializers.py
```diff
@@ -508,6 +507,16 @@ def validate(self, data):
 
         return data
 
+    def is_variant_valid(self, variants):
+        unique_names = len(
+            set([slugify(variant["name"]) for variant in variants])
+        ) == len(variants)
+        all_contains_alphanumeric_and_spaces = all(
+            [re.match("^[A-Za-z0-9 ]*$", variant["name"]) for variant in variants]
+        )
```
mythmon: There are a couple very small improvements you could make here. First, I'd remove the square brackets `[]`. With those in place, it will create a intermediate list and then process the list. Without the square brackets you have a generator expression that will stream the data to the `all` method without creating an intermediate result.

Second, I'd pull the regex creation out of the loop, so that you aren't re-compiling the regex for each variant. Something like

```
alpha_and_spaces_re = re.compile(r"^[a-z0-9 ]+$", re.IGNORE_CASE)
all_contains_alpha_numeric_and_spaces = all(
  alpha_and_spaces_re.match(variant["name"]) for variant in variants
)
```

It may also be worth pulling that regex out into a class or module variable so it can be computed once instead of once per function call.

PR #1885, opened 2019-11-03
handle empty string results filtering case fixes #1831

### LINKED ISSUE #1831: when filtering for "has results", some experiments without results are returned
On production...

Noticed this occuring. When I am filtering for experiments with results, two experiments without results (but that are completed) are returned. Why?!? If you go to these experiment's detail pages, the results section is not filled out at all. 

To duplicate, on production, go use the "how experiments with results completed" filter...

Screen Shot 2019-10-28 at 3 54 13 PM

tiftran:
naively assumed that if Results Form values were not None, it was filled out. Neglected to consider the case where a user looks at the result form and saves it with nothing that the values would be save with empty strings.


app/experimenter/experiments/filtersets.py
```diff
@@ -266,7 +266,7 @@ def completed_results_filter(self, queryset, name, value):
         if value:
             return queryset.exclude(
                 results_url=None, results_initial=None, results_lessons_learned=None
-            )
+            ).exclude(results_url="", results_initial="", results_lessons_learned="")
```
mythmon: This would only exclude something if all the fields were `None`, or all the fields were empty strings. It doesn't deal with mixed results, like

```json
{
  "results_url": null,
  "results_initial: "",
}
```

Is that ok? I think you may want to do something more that looks at each field individually and treats both `None` and the empty string as missing values. Maybe something with [`Q` objects](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#q-objects)?

tiftran: @mythmon  so i want something along the lines of... `(results_url=="" OR results_url == null) AND (results_initial=="" OR result_initial==null) AND(results_lesson=="" OR result_lesson==null)`?

mythmon: I think that will work, yeah.

app/experimenter/experiments/filtersets.py
```diff
@@ -265,7 +265,9 @@ def is_paused_filter(self, queryset, name, value):
     def completed_results_filter(self, queryset, name, value):
         if value:
             return queryset.exclude(
-                results_url=None, results_initial=None, results_lessons_learned=None
+                (Q(results_url="") | Q(results_url=None)),
+                (Q(results_initial="") | Q(results_initial=None)),
+                (Q(results_lessons_learned="") | Q(results_lessons_learned=None)),
```
mythmon: Minor nit: You don't need the parens around this, since they are separate parameters.

PR #1889, opened 2019-11-04
Fix types in tests fixes #1873

### LINKED ISSUE #1873: Experiment Design Serializer tests have mismatched types
Some tests have the wrong experiment type set, examples:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/tests/test_serializers.py#L1102

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/tests/test_serializers.py#L1018

We should do a pass through these tests and make sure the tests belonging to a given test class are testing the right serializer.

uhlissuh:
in some of the tests, we had the wrong `type` set. for instance, we were testing the pref serializer but by accident, in the data, we would set the `type` to addon. woops!

for clarity and consistency, i decided that all the tests on `TestExperimentDesignBaseSerializer` should just use the `generic` experiment type. so youll see me change some of the not messy tests to use these experiment types just for... again... consistency..

in the process of looking at this problem, i had to fix some of these types of errors in the tests on each experiment type called `test_serializer_outputs_dummy_variants_when_no_variants`. i realized we could move this test to to `TestExperimentDesignBaseSerializer` and NOT have this test on every experiment type. This behavior is implemented on the base, so it makes sense to me to test it only on `TestExperimentDesignBaseSerializer`.

### DIFF
--- a/app/experimenter/experiments/tests/test_serializers.py
+++ b/app/experimenter/experiments/tests/test_serializers.py
@@ -832,13 +832,13 @@ class TestExperimentDesignBaseSerializer(TestCase):
         )
 
     def test_serializer_rejects_ratio_not_100(self):
-        experiment = ExperimentFactory.create(type=ExperimentConstants.TYPE_ADDON)
+        experiment = ExperimentFactory.create(type=ExperimentConstants.TYPE_GENERIC)
 
         self.control_variant_data["ratio"] = 50
         self.treatment_variant_data["ratio"] = 40
 
         data = {
-            "type": ExperimentConstants.TYPE_PREF,
+            "type": ExperimentConstants.TYPE_GENERIC,
             "variants": [self.control_variant_data, self.treatment_variant_data],
         }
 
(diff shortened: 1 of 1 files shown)


app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -877,6 +877,34 @@ def test_serializer_rejects_duplicate_branch_names(self):
...
+                        "name": None,
+                        "ratio": None,
+                        "value": None,
+                    },
+                    {
+                        "description": None,
+                        "is_control": False,
+                        "name": None,
+                        "ratio": None,
+                        "value": None,
```
tiftran: oh yikes, this test shouldn't be passing if value is present.... seems like I've done something weirder than inputting the wrong types 👀

tiftran: I think you might need to break this test and compare one field/value at a time. Apparently i misused CountEqual somehow here...
Screen Shot 2019-11-05 at 7 47 23 PM

PR #1890, opened 2019-11-05
Add maxlength to character fields in serializer fixes #1862

### LINKED ISSUE #1862: Design form serializers require max length properties
I guess the serializers do not by default require max length parameters to all their fields so we can overflow the database sizes:

https://sentry.prod.mozaws.net/operations/experimenter-stage/issues/6576312/?query=is:unresolved

We should add a max length property to anything that takes text.

uhlissuh:
if we don't set max length, then a person will see a 500 if entering a long character string... because... the database  has max lengths set.

### DIFF
--- a/app/experimenter/experiments/tests/test_serializers.py
+++ b/app/experimenter/experiments/tests/test_serializers.py
@@ -36,6 +36,7 @@ from experimenter.experiments.serializers import (
     ExperimentDesignGenericSerializer,
     ExperimentDesignBaseSerializer,
     ExperimentRecipeMultiPrefVariantSerializer,
+    ExperimentDesignBranchBaseSerializer,
 )
 
 from experimenter.experiments.constants import ExperimentConstants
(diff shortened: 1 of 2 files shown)


uhlissuh:
@tiftran , added some tests. not sure I feel like it's worth doing a constant that we reference in these 5ish spots right now, so unless you or someone else has a strong feeling, I left that out. We can always do it in another PR I guess.

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -1116,9 +1131,59 @@ def test_serializer_rejects_inconsistent_pref_type_json(self):
...
+            "type": ExperimentConstants.TYPE_PREF,
+            "pref_type": "json string" * 100,
+            "pref_key": "name",
+            "pref_branch": "default",
+            "variants": [self.variant_1, self.variant_2],
+        }
+        serializer = ExperimentDesignPrefSerializer(data=data)
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("pref_type", serializer.errors)
+
```
uhlissuh: `pref_type` comes in from a select box so the likelihood it would ever be a long string is low. however... there is a model limit on the number of characters, so I included it in the serializer and made a test for it. Same with `pref_branch`.

PR #1892, opened 2019-11-05
Disable button on submit to prevent dup fixes #1872

### LINKED ISSUE #1872: Prevent multiple clicks of submit button on design page
We need to disable the save buttons while a submission is being made, it's not very noticeable when the connection is fast but for slow connections we can create race conditions.

uhlissuh:
using refs to disable and enable the save buttons... to prevent duplicate calls... database probs..


uhlissuh:
moving to a new PR for this work... it's pr #1900

uhlissuh:
I am abandoning this pr... it required a lot of rebasing for relatively little new code, and I just wanted to start fresh.

PR #1900, opened 2019-11-07
Make submit buttons disabled while saving fixes #1872

### LINKED ISSUE #1872: Prevent multiple clicks of submit button on design page

uhlissuh:
This is a new PR/branch... there was too much rebasing required for that other PR (#1892), especially considering how little new code I was actually trying to introduce. I just started over on with a new solution on a clean, up to date branch.


app/experimenter/static/js/components/DesignForm.js
```diff
@@ -86,6 +89,7 @@ class DesignForm extends React.PureComponent {
       .catch(err => {
         this.setState({
           errors: err.data,
+          saving: false,
```
rehandalal: i think this works in practice because we redirect the user on success but if we didn't this would not re-enable the button after successful saves. a better approach may be:
```js
requestSave
  .then(...)
  .catch(...)
  .finally(() => {
    this.setState({
      saving: false,
    });
  })
```

`finally` is always called whether the promise resolves successfully or not.

Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

uhlissuh: ok @rehandalal, I think I successfully made this change.

PR #1902, opened 2019-11-07
Add eslinting commands to existing make check fixes #1901

### LINKED ISSUE #1901: Add linting to existing make check and fix linting errors
Per the decision to split adding linting and actually fixing the linting into two tickets... here's the ticket for adding eslinting to make check and make code format and then fixing the linting errors.

uhlissuh:
Per discussion with @rehandalal, I am going to with this PR do the following:

1) Push 1 commit with just adding the new make commands to the old make commands, and then see CI fail. 

2) Push another commit that fixes the linting errors, and see the CI pass.


uhlissuh:
So, @rehandalal requested that we use trailing commas in our code formatting. In the last commit I added the prettier rule to the package.json and then formatted so trailing commas were added in.

PR #1911, opened 2019-11-13
Remove errors when removing branch fixes #1909

### LINKED ISSUE #1909: Removing branch shifts errors to wrong branch
Because the errors are tracked in an array, if we remove a branch, then all the errors shift up to the next branch, so we have to remove the errors when we remove a branch, and insert an empty error record if we create a new branch.

Example:

branches_errors 2019-11-13 13_03_27

uhlissuh:
branch-errors


uhlissuh:
@jaredkerim, I cannot figure out why this one CI test is failing.... locally this test does not fail. I've retriggered the CI and run the tests locally several times.... remove_branch test does not fail for me locally ever.... but continuously fails on CI. Ideas?

uhlissuh:
I don't see how to do this without initializing errors in state, and then adding and removing from that with addbranch and removebranch.

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -26,14 +26,22 @@ class DesignForm extends React.PureComponent {
...
     this.setState({
       loaded: true,
       data,
+      errors,
     });
   }
 
   addBranch() {
     const { data } = this.state;
+    const { errors } = this.state;
```
jaredkerim: I think you can do 42 and 43 on the same line?

uhlissuh: this got changed around.

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -46,15 +52,25 @@ class DesignForm extends React.PureComponent {
       value: "",
       is_control: false,
     });
+    const emptyErrorsBranch = fromJS({
```
jaredkerim: Sorry I don't think the multiline comment worked right? What I meant was `emptyBranch` and `emptyErrorsBranch` are almost exactly the same, do they need to be declared separately?

uhlissuh: What I am realizing is that we can just add an empty object to errors, it doesn't need to have any keys.

PR #1913, opened 2019-11-14
Move assets inside containers to fix yarn watch fixes #1869

### LINKED ISSUE #1869: Store assets only inside the app container
I *think* based on my experience messing around with the yarn watcher the other day, and what I just realized in #1865 , that I can do the same thing for assets, ie leave it only inside the built container, but not in a persistent volume.

jaredkerim:
Okay so I think the reason yarn watch wasn't working was because
 
- there was a bug inside `react-bootstrap` 
- that was preventing it from building in `yarn watch` 
- which is weird because it would build in `yarn build` but not in `yarn watch`
- `yarn watch` uses fancy stdout so `docker compose` was hiding its text output so i didn't see it until
- i tried `docker-compose -f docker-compose.yml run yarn bash` and then `yarn watch` and i saw the error
- so i did `yarn upgrade` to update all the packages
- now suddenly `yarn watch` works properly
- so i tried going and editing a js file to add an alert, reload the page, alert shows up, remove it, reload, alert goes away

So I think that was the core of the problem with `yarn watch`.  

After that I think another part of the problem was having the `/static/assets/` folder be on the host instead of only inside the container, because then you can get this case where the assets on the host get out of sync with what's in the container and you get this confusing state where you can switch branches but it's using another branches version of assets, so I moved `/static/assets/` into the containers entirely, so everyone should

`rm -Rf app/experimenter/static/assets/` after they pull in this change

~~Finally I made a small fix to `make integration_test` to kill the non integration stack first, then run tests, then kill the integration stack, so you don't have to remember to do it before or after you try to run tests.~~ I filed this as a followup instead https://github.com/mozilla/experimenter/issues/1914

*Edit:  I tried to get the hot module reloading working, but the problem is it would need to listen on `0.0.0.0` inside the container, but try to connect to `localhost` outside the container, and I don't think parcel watch expects those two addresses to be different.  I tried running the yarn container on the `host` network so localhost inside the container is the same as localhost outside the container, but that didn't work either, so I don't know what else to try.  As it stands now, you can edit a js file, wait till yarn says 'built', then reload the page manually and that works, tiny hassle but at least watch works now so 🤷‍♀

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,5 @@ integration_up_detached: integration_build
 integration_up: integration_build
 	docker-compose -p experimenter_integration -f docker-compose.integration-test.yml up
 
-integration_test: kill integration_build
+integration_test: integration_build
 	docker-compose -p experimenter_integration -f docker-compose.integration-test.yml run firefox tox -c tests/integration
-	docker-compose -p experimenter_integration -f docker-compose.integration-test.yml kill 
-
--- a/Makefile
+++ b/Makefile
@@ -121,5 +121,7 @@ integration_up_detached: integration_build
 integration_up: integration_build
 	docker-compose -p experimenter_integration -f docker-compose.integration-test.yml up
 
-integration_test: integration_build
+integration_test: kill integration_build
 	docker-compose -p experimenter_integration -f docker-compose.integration-test.yml run firefox tox -c tests/integration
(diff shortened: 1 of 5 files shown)


jaredkerim:
Also this forgets to update docker-compose-full.yml but I think I'll just do this intead: https://github.com/mozilla/experimenter/issues/1915

PR #1918, opened 2019-11-14
Reduce duplication in compose files fixes #1915

### LINKED ISSUE #1915: Reduce duplication in docker compose files
To reduce duplication in our docker compose files, we can use the techniques listed here:

https://docs.docker.com/compose/extends/

so that we don't end up having to repeat things like #1869 in multiple places.

jaredkerim:
Turns out you can smush one compose file on top of another if you just say `-f composeA -f composeB` so you only need to specify the things you want to change in `composeB`.  That's pretty handy.


jaredkerim:
I think this is broken until I can land #1869

PR #1922, opened 2019-11-15
removed id for dummy variants fixes #1921

### LINKED ISSUE #1921: Can't Save Branches for New Experiments
GET request returns variants with null ids and that is passed back at PUT, experimenter doesn't take in null ids

tiftran:


app/experimenter/experiments/serializers.py
```diff
@@ -45,8 +45,11 @@ def to_representation(self, data):
             for field in self.child.fields:
```
jaredkerim: How about up here doing something like

    initial_fields = set(self.child.fields) - set(["id"])

?

tiftran: 👍 yes set differences.... why didn't i think of that 🤦‍♀

jaredkerim: 😉

PR #1940, opened 2019-11-19
Added Changelogs to back to Design Page fixes #1844

### LINKED ISSUE #1844: Add changelogs to design form serializers
We need to add changelogs back to experiment design serializers.

tiftran:
I think I somewhat successfully removed the changes comparision out of forms.py...
We do lose the labels of the design form as display names as it currently is

Did not add tests to see whether changelog existed to the existing Variant/Design Forms Tests, since I would assume these may be taken out?


app/experimenter/experiments/changelog_utils.py
```diff
@@ -0,0 +1,89 @@
+# blank lines...
```
tiftran: i wasn't able to get flake8 and black happy without this...

jaredkerim: Lol silly formatters 😆

PR #1953, opened 2019-11-19
Always put control branch first fixes #1926

### LINKED ISSUE #1926: Control Branch isnt being displayed as control
Screen Shot 2019-11-15 at 1 45 47 PM

jaredkerim:
Basically the serializer needs to always put the control branch first, and then we can expect that ordering once it gets to the client.  

If we don't do that and only separate them when we render, then you can wind up in this situation where the top level state has `[branch1, control]` but it's rendered out as `[control, branch1]` so when you send the update info back up to the top it's in the wrong order.  

So I changed the serializer to always put control first, then the client will just blindly expect that.  

Also I changed the serializer to always order by db id, which will mimic the order they're created in, so that way when you edit they should be sticky with the order they were originally created in, so you don't get branches jumping around which is confusing.

### DIFF
--- a/app/experimenter/static/js/components/BranchManager.js
+++ b/app/experimenter/static/js/components/BranchManager.js
@@ -40,25 +40,15 @@ class BranchManager extends React.PureComponent {
     );
   }
 
-  renderBranches() {
-    const { branches } = this.props;
-
-    return branches
-      .withMutations(mutable => {
-        // Make sure the control branch is the first branch
-        mutable
-          .filter(b => b.get("is_control"))
-          .concat(branches.filter(b => !b.get("is_control")));
-      })
-      .map(this.renderBranch);
-  }
-
   render() {
-    const { onAddBranch } = this.props;
+    const { onAddBranch, branches } = this.props;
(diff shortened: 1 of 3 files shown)


app/experimenter/static/js/components/BranchManager.js
```diff
@@ -40,25 +40,15 @@ class BranchManager extends React.PureComponent {
...
-    const { onAddBranch } = this.props;
+    const { onAddBranch, branches } = this.props;
+    const controlBranch = branches.get(0);
+    const treatmentBranches = branches.slice(1);
 
     return (
       <React.Fragment>
-        {this.renderBranches()}
+        {this.renderBranch(controlBranch, 0)}
+        {treatmentBranches.map((b, i) => this.renderBranch(b, i + 1))}
```
tiftran: probably a dumb question, but is it cause of ordering that we separate the control branch from the rest?

jaredkerim: Oh you're right I don't even need that! I had an in between state where I was putting control first on the client and the server, but now we don't even need this, good catch.

PR #1963, opened 2019-11-21
Move branch methods into branch manager fixes #1920

### LINKED ISSUE #1920: Move addBranch/removeBranch to BranchManager
I think with this new setup of storing the state in the top level Form component, we can still move the addBranch/removeBranch methods down to the BranchManager component and that can call handleDataChange on the parent passed in as a prop, and that way the logic is localized to BranchManager.  Might wanna do this sooner than later so we don't build ourselves into a corner.

jaredkerim:
So my reasoning here was that there's gonna be many types of intake that have no branches (rollouts to start but probably others coming), so we should move the branch methods down into branch manager.  Also since we're going to be reactifying the timeline/population page and we'll probably split DesignForm into something more reusable, it's a good step towards making it more generic.

### DIFF
--- a/app/experimenter/static/js/components/Branch.js
+++ b/app/experimenter/static/js/components/Branch.js
@@ -17,28 +17,27 @@ class Branch extends React.PureComponent {
 
   handleChange(key, value) {
     const { onChange, branch } = this.props;
-    const updated = branch
-      .set("is_control", this.props.index === 0)
-      .set(key, value);
-    onChange(updated);
+    onChange(branch.set(key, value));
   }
 
   renderTitle() {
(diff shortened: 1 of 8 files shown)


app/experimenter/static/js/components/Branch.js
```diff
@@ -17,28 +17,27 @@ class Branch extends React.PureComponent {
 
   handleChange(key, value) {
     const { onChange, branch } = this.props;
-    const updated = branch
-      .set("is_control", this.props.index === 0)
```
jaredkerim: Noticed I forgot this in here after I did the control stuff, so got rid of that.

PR #1991, opened 2019-11-27
multi pref intake fixes #1421

### LINKED ISSUE #1421: Pref studies need the Branching section updated to support multiple prefs
Remove the existing top 3 fields that are currently under "Firefox Preference" heading - (they will be going under branches).  We want them to behave the same in the branching section as these fields did.

Inside the Branch details: 
- Add "Branch" to the existing "Name" field title - so it is now clearly a field for "Branch Name".
- Add a "Pref Name" field before the existing "Pref Value" field.  It should be the same type and same validation properties of the original "Pref Name" field. 
- Add "Pref Branch" here - same drop down options and validation as the original pref branch.
- New: "Add Pref" button INSIDE each branch section.  Clicking should pop up 3 new boxes for "Pref Name", "Pref Value", "Pref Branch".

Sending info to Normandy about multiple prefs:
- Need new schema generation - Now we’ll need to send over:  branchname - pref name - pref value - pref branch:default/user.  Existing schema https://searchfox.org/mozilla-central/source/toolkit/components/normandy/actions/schemas/index.js#230  
- Add a new serializer for new recipe format.  

image

tiftran:
Screen Shot 2019-11-28 at 5 55 50 PM
Screen Shot 2019-11-28 at 4 08 44 PM


app/experimenter/experiments/models.py
```diff
@@ -92,6 +93,8 @@ class Experiment(ExperimentConstants, models.Model):
         validators=[MaxValueValidator(ExperimentConstants.MAX_DURATION)],
     )
 
+    is_multi_pref_format = models.BooleanField(default=False)
```
jaredkerim: I think in the name of consistency this could just be `is_multi_pref` like `is_branched_addon`?

app/experimenter/experiments/models.py
```diff
@@ -804,6 +807,60 @@ def json_dumps_value(self):
...
+        null=True,
+    )
+    pref_branch = models.CharField(
+        max_length=255,
+        choices=ExperimentConstants.PREF_BRANCH_CHOICES,
+        blank=True,
+        null=True,
+    )
+
+    pref_value = models.CharField(max_length=255, blank=False, null=False)
```
jaredkerim: How come type and branch have blank/null=True but name/value don't?

app/experimenter/experiments/models.py
```diff
@@ -804,6 +807,60 @@ def json_dumps_value(self):
...
+
+            except (json.JSONDecodeError, ValueError):
+                raise ValidationError(
+                    "value", f"Unexpected value type (should be {self.pref_type})"
+                )
+
+    def save(self, *args, **kwargs):
+        self.clean()
+
+        super().save()
```
jaredkerim: This is the same as a regular variant, can't we pull this out into a mixin?

app/experimenter/experiments/serializers.py
```diff
@@ -413,7 +432,7 @@ class Meta:
         )
 
     def get_action_name(self, obj):
-        if obj.is_multi_pref:
+        if obj.is_multi_pref_experiment or obj.is_multi_pref_format:
```
jaredkerim: Can't we just do the `is_multi_pref_format` check inside of `is_multi_pref_experiment`?

app/experimenter/experiments/models.py
jaredkerim: Actually do we need to do this on the model at all?

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,11 +1,12 @@
+"""
```
jaredkerim: Ah yeah in branched we just delete this altogether

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -106,13 +106,15 @@ class DesignForm extends React.PureComponent {
...
       );
     }
-
     // Select the appropriate form component to use based on experiment type
     let Form;
     switch (this.state.data.get("type")) {
       case "pref":
         Form = PrefForm;
         break;
+      case "multi-pref":
```
jaredkerim: Is multi-pref a separate experiment type?

app/experimenter/static/js/index.js
```diff
@@ -17,7 +17,8 @@ const branchesDiv = document.getElementById("react-branches-form");
 
 if (branchesDiv) {
   const slug = branchesDiv.dataset.experimentSlug;
-  const experimentType = branchesDiv.dataset.experimentType;
+  //const experimentType = branchesDiv.dataset.experimentType;
+  const experimentType = "multi-pref";
```
jaredkerim: Ah yep don't do it this way, just pass `is_multi_pref1` as a separate prop, look at teh way I did it for branched.

app/experimenter/experiments/models.py
tiftran: ooohh so, this should be `is_multi_pref` and `is_multi_pref_experiment` should be `use_multi_pref_serializer`?

app/experimenter/experiments/models.py
tiftran: lemme fix this...

app/experimenter/experiments/models.py
tiftran: I think the rationale is that VariantPreferences are saved separately from variants, and this was put in place just in case validation from the serializer fails?

app/experimenter/experiments/serializers.py
tiftran: kk will do

app/experimenter/experiments/models.py
jaredkerim: Note: we chatted about this on zoom and agreed to do validation in the serializers and not on the model.

app/experimenter/static/js/components/DesignForm.js
jaredkerim: Note: we chatted and talked about passing this in as a prop to DesignForm through the template the way we do for `is_branched_addon`

app/experimenter/experiments/serializers.py
```diff
@@ -616,36 +714,65 @@ def update(self, instance, validated_data):
...
+        existing_pref_ids = self.get_existing_preference_ids(instance)
+        submitted_pref_ids = []
+        for variant_data, pref in variant_preferences:
+
+            variant = ExperimentVariant.objects.get(**variant_data)
+            for preference in pref:
+                preference["variant_id"] = variant.id
+                VariantPreferences(**preference).save()
+
+                if preference.get("id"):
```
jaredkerim: Omg it's time for the walrus operator:

    if preference_id := preference.get("id"):
        submitted_pref_ids.append(preference_id)

:O :O :O

app/experimenter/experiments/serializers.py
```diff
@@ -544,6 +588,60 @@ class Meta(ExperimentDesignVariantBaseSerializer.Meta):
...
+        ) == len(preferences)
+
+        all_contains_alphanumeric_and_spaces = all(
+            Experiment.EXPERIMENT_VARIANT_NAME_REGEX.match(pref["pref_name"])
+            for pref in preferences
+        )
+
+        return unique_names and all_contains_alphanumeric_and_spaces
+
+    def is_value_type_match(self, preferences):
```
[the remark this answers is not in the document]
jaredkerim: Also usually methods like `is_thingy`/`has_whatever` return a bool.

app/experimenter/experiments/serializers.py
```diff
@@ -616,36 +714,65 @@ def update(self, instance, validated_data):
...
+            VariantPreferences.objects.filter(id__in=removed_ids).delete()
+
+        return instance
+
+    def get_existing_preference_ids(self, instance):
+        pref_ids = []
+
+        for variant in instance.variants.all():
+            pref_ids.extend([p.id for p in variant.preferences.all()])
+        return pref_ids
```
jaredkerim: I think you can just do

    experiment.variants.all().values_list("preferences__id", flat=True)

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -28,16 +29,21 @@ class DesignForm extends React.PureComponent {
     };
   }
 
+  isMultiPref() {
+    return this.state.data.get("is_multi_pref", this.props.isMultiPref);
```
jaredkerim: Does this do the right thing.....?  Maybe it does....

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -28,16 +29,21 @@ class DesignForm extends React.PureComponent {
...
+  }
+
   isBranchedAddon() {
-    return (
-      this.state.data.get("is_branched_addon") || this.props.isBranchedAddon
-    );
+    return this.state.data.get("is_branched_addon", this.props.isBranchedAddon);
```
jaredkerim: Sneaking things in I see ;)

jaredkerim: Also these both suffer from the bug I'm fixing in my pr, but don't worry I'll fix it.

app/experimenter/static/js/components/Pref.js
```diff
@@ -0,0 +1,143 @@
...
+class Pref extends React.PureComponent {
+  static propTypes = {
+    preference: PropTypes.instanceOf(Map),
+    errors: PropTypes.instanceOf(Map),
+    index: PropTypes.number,
+    onChange: PropTypes.func,
+    remove: PropTypes.func,
+  };
+
+  handleChange(key, value) {
```
jaredkerim: Remember the pattern I was talking about, `handleChange` should include WHAT's changing ie `handleWhateverChange`

app/experimenter/static/js/components/PrefForm.js
```diff
@@ -15,72 +17,102 @@ import {
...
+            value={this.props.data.get("pref_type")}
+            error={this.props.errors.get("pref_type", "")}
+            as="select"
+            helpContent={PREF_TYPE_HELP}
+          >
+            <option>Firefox Pref Type</option>
+            <option>boolean</option>
+            <option>integer</option>
+            <option>string</option>
+            <option>json string</option>
```
jaredkerim: Don't do it here but in mine I moved all these to constants.

app/experimenter/static/js/components/PrefForm.js
```diff
@@ -15,72 +17,102 @@ import {
...
+            }}
+            value={this.props.data.get("pref_branch")}
+            error={this.props.errors.get("pref_branch", "")}
+            as="select"
+            helpContent={PREF_BRANCH_HELP}
+          >
+            <option>Firefox Pref Branch</option>
+            <option>default</option>
+            <option>user</option>
+          </DesignInput>
```
jaredkerim: We've done this a couple times in the react so far, where we omit a value from `<option>` and the label just *happens* to be the actual correct value, but we shouldn't keep that, I cleaned up a bunch of them in my PR but.... you know we should avoid it.

app/experimenter/static/js/components/PrefForm.js
```diff
@@ -15,72 +17,102 @@ import {
...
+
   render() {
     return (
       <div>
         <Row className="mb-3">
           <Col md={{ span: 4, offset: 3 }}>
-            <h4>Firefox Pref</h4>
+            <h3>Firefox Pref</h3>
           </Col>
         </Row>
```
jaredkerim: I think you can get rid of this.

app/experimenter/templates/experiments/section_branches.html
```diff
@@ -5,19 +5,31 @@
...
   {% for variant in experiment.variants.all %}
     <div class="border-left-grey pl-3 mb-4">
       <h5>{{ variant.ratio }}% {{ variant.name }}</h5>
-      {% if experiment.is_pref_experiment %}
+      {% if experiment.is_multi_pref %}
+        {% for preference in variant.preferences.all %}
+        
+          <div class="border-left-grey pl-3 ml-2"> 
+            <p><strong> Pref Type = <span class="text-info">{{ preference.pref_type }}</span></strong></p>
+            <p><strong> Pref Branch = <span class="text-info">{{ preference.pref_branch }}</span></strong></p>
```
jaredkerim: I wouldn't do these two like this....  but I think just leave this and we'll land it and we can sit and play with styles together.

app/experimenter/templates/experiments/section_branches.html
```diff
@@ -5,19 +5,31 @@
...
+            <p><strong> {{ preference.pref_name }} = <span class="text-info">{{ preference.pref_value }}</span></strong></p>
+          </div>
+        {% endfor %}
+      {% elif experiment.is_pref_experiment %}
         {% if experiment.pref_type == "json string" %}
           <p><strong>{{ experiment.pref_key }} = <pre class="text-info">{{ variant.json_dumps_value}}</pre></strong></p>
         {% else %}
+        <h1>HI</h1>
```
jaredkerim: HI!!! 😄

app/experimenter/static/js/components/DesignForm.js
tiftran: i think it does?

app/experimenter/experiments/serializers.py
```diff
@@ -616,36 +714,65 @@ def update(self, instance, validated_data):
...
-                json.loads(pref_value)
-            except ValueError:
-                return {value: "The pref value must be valid JSON."}
-        return {}
+        instance = super().update(instance, validated_data)
+        existing_pref_ids = self.get_existing_preference_ids(instance)
+        submitted_pref_ids = []
+        for variant_data, pref in variant_preferences:
+
+            variant = ExperimentVariant.objects.get(**variant_data)
```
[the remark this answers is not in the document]
tiftran: not exactly sure what do you mean?

app/experimenter/experiments/serializers.py
tiftran: walrus notation isn't support by black yet, so it fails linting, or at least it was failing locally?

app/experimenter/experiments/serializers.py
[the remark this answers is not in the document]
jaredkerim: variant_data = {"id": 1, name="wahtever", .....}

.get(**variant_data) -> .get(id=1, name="wahtever', this=that)

PR #1998, opened 2019-11-28
Add rollout intake fixes #1964

### LINKED ISSUE #1964: Add Rollout Intake
I'm creating this issue to track the rollout type.  @shell1 filed separate tickets for various aspects of its design but I think from an implementation perspective it'll make more sense to do it all in one change.

jaredkerim:
Okay this is a bit of a doozy, probably shouldn't have let it get so big but whatever.  Also I fixed a bunch of random bugs in here which I also probably should not have done but also whatever.

So there's a lot of changes I had to make because a bunch of experiment stuff doesn't apply here, like there's no variants, so there was a bunch of changes for that.  Not all fields apply so changes for that.  Some risks change....  I honestly can't remember everything that had to change but I think it's all in there. 

Here's a bajillion screenshots showing the whole end to end for both addons and prefs:

Screenshot_2019-12-09 Mozilla Experimenter
Screenshot_2019-12-09 Addon Rollout1
Screenshot_2019-12-09 Addon Rollout1(1)
Screenshot_2019-12-09 Addon Rollout1(3)
Screenshot_2019-12-09 Addon Rollout1(4)
Screenshot_2019-12-09 Addon Rollout1(5)
Screenshot_2019-12-09 Addon Rollout1(6)
Screenshot_2019-12-09 https localhost
Screenshot_2019-12-09 Mozilla Experimenter(1)
Screenshot_2019-12-09 Pref Rollout 1
Screenshot_2019-12-09 Pref Rollout 1(1)
Screenshot_2019-12-09 Pref Rollout 1(4)
Screenshot_2019-12-09 Pref Rollout 1(5)
Screenshot_2019-12-09 Pref Rollout 1(6)
Screenshot_2019-12-09 Pref Rollout 1(7)
Screenshot_2019-12-09 https localhost(1)
Screenshot_2019-12-09 Mozilla Experimenter(2)


Makefile
```diff
@@ -59,10 +59,10 @@ check_migrations: test_build
 	$(COMPOSE_TEST) run app sh -c "$(WAIT_FOR_DB) $(PYTHON_CHECK_MIGRATIONS)"
 
 check: test_build
-	$(COMPOSE_TEST) run app sh -c "$(WAIT_FOR_DB) $(PYTHON_CHECK_MIGRATIONS)&&$(PYTHON_CHECK_MIGRATIONS)&&$(BLACK_CHECK)&&$(FLAKE8)&&$(ESLINT)&&$(PYTHON_TEST)"
+	$(COMPOSE_TEST) run app sh -c "$(WAIT_FOR_DB) $(PYTHON_CHECK_MIGRATIONS)&&$(BLACK_CHECK)&&$(FLAKE8)&&$(ESLINT)&&$(PYTHON_TEST)"
```
jaredkerim: I noticed I put this in twice oops

app/experimenter/experiments/api_urls.py
```diff
@@ -1,16 +1,17 @@
 from django.conf.urls import url
 
 from experimenter.experiments.api_views import (
+    ExperimentCloneView,
```
jaredkerim: For no good reason I resorted a bunch of imports alphabetically 🤷‍♀

app/experimenter/experiments/models.py
```diff
@@ -806,10 +888,6 @@ def type(self):
         else:
             return "Treatment"
 
-    @property
-    def json_dumps_value(self):
-        return json.dumps(json.loads(self.value), indent=2)
```
jaredkerim: Moved this to a template filter because it's used in 3 place snow

app/experimenter/experiments/serializers.py
```diff
@@ -381,6 +381,7 @@ class Meta:
 class ExperimentRecipeBranchedAddonArgumentsSerializer(
     ExperimentRecipeBranchedArgumentsSerializer
 ):
+    slug = serializers.ReadOnlyField(source="normandy_slug")
```
jaredkerim: Slug is required for branched addons, fixed that here....

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -867,6 +916,97 @@ def test_serializer_outputs_expected_schema_for_multipref(self):
...
+            name="Experimenter Name",
+            normandy_slug="normandy-slug",
+            platform=Experiment.PLATFORM_WINDOWS,
+            population_percent=30.0,
+            rollout_type=Experiment.TYPE_ADDON,
+            slug="experimenter-slug",
+            type=Experiment.TYPE_ROLLOUT,
+        )
+        serializer = ExperimentRecipeSerializer(experiment)
+        self.assertDictEqual(
```
jaredkerim: Oh related to what we were saying about comparing dicts, I think we should be using this method everywhere, but I stopped myself from going and changing every test to use it.  Maybe we can file that as a follow up.

app/experimenter/experiments/views.py
```diff
@@ -102,18 +99,10 @@ class ExperimentTimelinePopulationUpdateView(ExperimentFormMixin, UpdateView):
...
     template_name = "experiments/edit_design.html"
 
-    def get_form_class(self):
-        if self.object.is_addon_experiment:
-            return ExperimentDesignAddonForm
-        elif self.object.is_pref_experiment:
-            return ExperimentDesignPrefForm
-        elif self.object.is_generic_experiment:
-            return ExperimentDesignGenericForm
-
```
jaredkerim: Noticed this view was still doing form lookups so I got rid of that.

app/experimenter/static/css/experimenter.css
```diff
@@ -216,3 +216,11 @@ pre.json {
...
   border-color: #ced4da;
 }
+
+.required-label.required .text-muted {
+  display: none;
+}
+
+.required-label.optional .text-danger {
+  display: none;
+}
```
jaredkerim: I added this little thing so we can flip 'optional' and 'required' fields in static forms easily.

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -30,11 +37,17 @@ class DesignForm extends React.PureComponent {
   }
 
   isMultiPref() {
-    return this.state.data.get("is_multi_pref", this.props.isMultiPref);
+    return (
+      this.props.experimentType === TYPE_PREF &&
```
jaredkerim: This was an edge case I noticed where if you change the experiment type, this line would still catch and fetch the wrong serialized data so I fixed it for both multi pref and branched cases.

app/experimenter/static/js/components/DesignForm.js
```diff
@@ -124,14 +137,18 @@ class DesignForm extends React.PureComponent {
     // Select the appropriate form component to use based on experiment type
     let Form;
     switch (this.props.experimentType) {
-      case "pref":
+      case TYPE_PREF:
```
jaredkerim: Constants are our friends.

app/experimenter/static/js/scripts/edit-overview.js
```diff
@@ -0,0 +1,35 @@
+jQuery(function ($) {
```
jaredkerim: This whole thing is crappy but it was the only way I could find to reasonably hide/show required/optional fields while switching experiment types 🤷‍♀

app/experimenter/experiments/forms.py
```diff
@@ -1015,7 +1037,7 @@ class ExperimentReviewForm(ExperimentConstants, ChangeLogMixin, forms.ModelForm)
     # Optional
     review_advisory = forms.BooleanField(
         required=False,
-        label="Lightning Advisory (Optional)",
+        label="Lightning Advisory",
```
tiftran: Would this be a little confusing? 
From what I understand from https://github.com/mozilla/experimenter/pull/1998/files#diff-b75d2c1707450c0ad27a64a54719a0bbR1106, rollouts requires lightning advisory as a hard requirement while the other types it's still technically optional but it's in the required list?

app/experimenter/experiments/serializers.py
```diff
@@ -424,6 +426,30 @@ class Meta:
...
+class ExperimentRecipeAddonRolloutArgumentsSerializer(serializers.ModelSerializer):
+    slug = serializers.ReadOnlyField(source="normandy_slug")
+    extensionApiId = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Experiment
+        fields = ("slug", "extensionApiId")
+
+    def get_extensionApiId(self, obj):
+        return f"TODO: {obj.addon_release_url}"
```
tiftran: I like this, should this be a branched addon thing too?

jaredkerim: Could do, ya.

app/experimenter/experiments/forms.py
jaredkerim: I feel like it's confusing either way, if it's in the required list but says optional, and now it might be required it might not, so I was banking on letting the process sort it out and we might find a new way to represent it rather than having custom logic to shift the label back and forth.

app/experimenter/experiments/migrations/0078_add_rollouts.py
```diff
@@ -0,0 +1,82 @@
+# Generated by Django 2.1.11 on 2019-12-10 04:30
+
```
uhlissuh: this will have to be renumbered to 79, i think, we already have a 78.

jaredkerim:
@mythmon Oh yes good catch I need to remove that field as well!

jaredkerim:
Okay I took out population percent for rollouts:

Screenshot_2019-12-13 Rollout
Screenshot_2019-12-13 Mozilla Experimenter

PR #2000, opened 2019-11-28
pref validation mixin fixes #1996

### LINKED ISSUE #1996: Pull PrefType, PrefValue Validation into Mixin

tiftran:


app/experimenter/experiments/serializers.py
```diff
@@ -616,7 +616,36 @@ def update(self, instance, validated_data):
         return instance
 
 
-class ExperimentDesignPrefSerializer(ExperimentDesignBaseSerializer):
+class PrefValidationMixin(serializers.ModelSerializer):
```
jaredkerim: Oops, mixins, should inherit from (object)

app/experimenter/experiments/serializers.py
```diff
@@ -616,7 +616,36 @@ def update(self, instance, validated_data):
...
+
+    def validate_pref(self, pref_type, pref_value, error_list):
+        if pref_type == "integer":
+            try:
+                int(pref_value)
+            except ValueError:
+                error_list.append({"value": "The pref value must be an integer."})
```
jaredkerim: Isn't `"value"` here specific to the field on the given model? This should probably be a parameter.

app/experimenter/experiments/serializers.py
```diff
@@ -616,7 +616,36 @@ def update(self, instance, validated_data):
...
+
+        if pref_type == "json string":
+            try:
+                json.loads(pref_value)
+            except ValueError:
+                error_list.append({"value": "The pref value must be valid JSON."})
+            else:
+                error_list.append({})
+        if pref_type == "string":
+            error_list.append({})
```
jaredkerim: Do we need this at all?

app/experimenter/experiments/serializers.py
```diff
@@ -653,33 +682,12 @@ def validate(self, data):
...
-                    json.loads(variant["value"])
-                except ValueError:
-                    error_list.append({"value": ["The pref value must be valid JSON."]})
-                else:
-                    error_list.append({"value": ""})
-
-        for entry in error_list:
-            if entry["value"] != "":
-                raise serializers.ValidationError({"variants": error_list})
+            self.validate_pref(pref_type, variant["value"], error_list)
```
jaredkerim: Very nice 👍

app/experimenter/experiments/serializers.py
jaredkerim: OH I see.  You're passing in the list, no this is too specific to `variants`, just build a dictionary in here and return it and then append to `error_list` outside.

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -72,6 +73,27 @@ def test_json_field(self):
         )
 
 
+class TestPrefValidationMixin(TestCase):
```
jaredkerim: We really should test all 6 possibilities, int/bool/json valid/invalid.

PR #2036, opened 2019-12-06
Display one empty pref per variant for multipref fixes #2029

### LINKED ISSUE #2029: If user selects multi pref, but doesn't add any prefs, server returns 400 and we don't catch it.
bug

we aren't catching a 400 returned by the server when a user has selected multi pref and hasn't put in any prefs on the branches

uhlissuh:
this pr creates an empty preferences branch on each variant. it doesn't have its remove button, because each variant has to have at least one pref for multi pref. 

multidemo
Screen Shot 2019-12-05 at 10 47 08 PM


app/experimenter/static/js/components/DesignForm.js
```diff
@@ -56,12 +56,31 @@ class DesignForm extends React.PureComponent {
...
-  handleDataChange(key, value) {
+  async handleDataChange(key, value) {
     this.setState(({ data }) => ({
       data: data.set(key, value),
+      loading: true,
     }));
   }
 
+  async handleMultiPrefChange(key, value) {
+    this.setState(
```
tiftran: ohh we're going forward with click button -> rest call already :)

app/experimenter/static/js/components/Pref.js
```diff
@@ -33,17 +33,19 @@ class Pref extends React.PureComponent {
...
-    );
+    if (index != 0) {
+      return (
+        <Button
+          variant="outline-danger"
+          onClick={() => {
+            this.props.remove(index);
+          }}
+          id="remove-branch-button"
+        >
```
tiftran: could you rename the id for me, seems like i forgot :(

app/experimenter/static/js/components/DesignForm.js
uhlissuh: yes, @jaredkerim thought I should go with this approach here. Or else, I would have had to add the blank preferences on the client side, which is kind of out of pattern at this point. :)

uhlissuh: It didn't turn out too bad, me thinks.

app/experimenter/static/js/components/Pref.js
uhlissuh: sure!

uhlissuh: done.

uhlissuh:
I also realized that I had some unwanted code from previous solutions I was thinking about.. woops!... Oviously I don't need to make `handleDataChange` async or change loading to true... so I removed those oversights from this PR on my latest commit.

app/experimenter/static/js/components/DesignForm.js
jaredkerim: Oh and it occurred to me we should stick to the pattern of always calling methods that modify state on components with `handle` at the start, and also I think `reload` is probably a better word than `refresh` here, and we should probably mention that this calls the API, so I think `handleReloadAPIState` is probably what we want.  Thanks for humouring me @uhlissuh ! Sorry to be such a pain!  This change is really succinct, it looks great!

giphy

PR #2048, opened 2019-12-09
allow hyphens for variant name

tiftran:


tiftran:
based on slack convo made changes so that special characters are allowed

app/experimenter/experiments/tests/test_serializers.py
```diff
@@ -1433,17 +1433,16 @@ def test_serializer_rejects_duplicate_branch_names(self):
...
-        self.control_variant_data["name"] = "&re@t br@nche$!"
+        self.control_variant_data["name"] = "&re@t -br@nche$!"
 
         data = {"variants": [self.control_variant_data, self.treatment_variant_data]}
 
         serializer = ExperimentDesignBaseSerializer(instance=experiment, data=data)
 
-        self.assertFalse(serializer.is_valid())
-        self.assertIn("variants", serializer.errors)
+        self.assertTrue(serializer.is_valid())
```
jaredkerim: I would also save and check that the 

- name is preserved
- the slug is the expected value

app/experimenter/experiments/constants.py
```diff
@@ -194,7 +194,7 @@ class ExperimentConstants(object):
     FX_MIN_MULTI_BRANCHED_VERSION = 68
 
     # ExperimentVariantName REGEX
-    EXPERIMENT_VARIANT_NAME_REGEX = re.compile(r"^[a-z0-9 ]+$", re.IGNORECASE)
+    EXPERIMENT_VARIANT_NAME_REGEX = re.compile(r"^[a-z0-9 \-]+$", re.IGNORECASE)
```
jaredkerim: Wait do we still need this?

tiftran: no, planning to remove it in https://github.com/mozilla/experimenter/pull/2051 after this was merged

PR #2052, opened 2019-12-09
Add higher risk Q and UI red pill fixes #1992

### LINKED ISSUE #1992: Add Risk Question that will put a "higher risk" label on Experiment
Add a new question:  
"I have been advised that this delivery design creates a higher risk of errors due to complexity or timing requirements."  Yes = riskier 

if they say Yes - add a Red label to the experiment that says "Higher Risk"

Help should link here https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#Pref-FlipandAdd-OnExperiments-Risk

uhlissuh:
Screen Shot 2019-12-09 at 2 06 57 PM
Screen Shot 2019-12-09 at 2 06 35 PM
Screen Shot 2019-12-09 at 2 06 16 PM
Screen Shot 2019-12-09 at 1 42 13 PM


jaredkerim:
> I think it might be worth having a filter to filter only high risk deliveries? thoughts?

Sounds like a good idea but we could also do it as a follow up, @tiftran can you file a bug and we can ask shell about it at next triage?

app/experimenter/experiments/forms.py
```diff
@@ -966,6 +975,7 @@ class Meta:
             "risk_security",
             "risk_revision",
             "risk_technical",
+            "risk_higher_risk",
```
jaredkerim: Love the name of this field 🤣

PR #2054, opened 2019-12-10
Wait for form tag on design page fixes #2053

### LINKED ISSUE #2053: Integration test fails on design page
I think this is a timing issue, ie the selenium driver is trying to access a field on the design page before the react has finished rendering so we need to add like a delay or something at the start of all the design page integration tests.

Example:

https://circleci.com/gh/mozilla/experimenter/6171#artifacts/containers/0

uhlissuh:
Switching the page wait locator for design page to the actual form tag, which won't load until we've gotten back data from the server.

previously, we were waiting on the body to be present, which in the case of the design form, happens before we've fetched from server and set up the form.

### DIFF
--- a/tests/integration/pages/experiment_design.py
+++ b/tests/integration/pages/experiment_design.py
@@ -32,7 +32,7 @@ class DesignPage(Base):
     )
     _page_wait_locator = (
         By.CSS_SELECTOR,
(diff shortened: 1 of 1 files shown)


jaredkerim:
@uhlissuh So does waiting on the pref key field work?

PR #2093, opened 2019-12-19
Add js tests for single pref form fixes #2082

### LINKED ISSUE #2082: JS Tests for Single Pref Design Page

uhlissuh:
I don't do anything with adding branches or removing them... as in the addonform tests... I think that stuff should be moved to a designform test because branch addition/removal happens for pref, addon, and generic tests alike. Each experiment type test should be shorter and more about making sure we are displaying what comes back from the server properly, that we can change those fields, and then that we are sending the updated fields back to the server as I have done here. 
~~left to do:~~
~~do a bit of refactoring in datafactory so postgeneration method isn't just a c/p~~


uhlissuh:
update: per a group discussion, added tests for add branch and remove branch and displaying errors to the single pref tests.

uhlissuh:
update two: I added the radio button switch to Multipref to these tests. Also checking that a blank pref branch form shows. Interestingly, these blank pref form tests are currently failing on my computer because I still have the problem of, inexplicably, the blank pref branches not showing on my copy of Experimenter. But they pass here.

app/experimenter/static/js/tests/PrefForm.test.js
```diff
@@ -0,0 +1,200 @@
...
+import DesignForm from "experimenter/components/DesignForm";
+import * as Api from "experimenter/utils/api";
+import { PrefDataFactory } from "./DataFactory";
+import {
+  addBranch,
+  removeBranch,
+  waitForFormToLoad,
+} from "experimenter/tests/helpers.js";
+
+describe("The `DesignForm` component for Prefs", () => {
```
jaredlockhart: I would say 'Pref Experiment' just to be explicit

uhlissuh:
the final commit fixes two bugs, despite only mentioning one. First: a serializer bug where we weren't showing blank pref branches if there was already design-related data saved. Second: makes unique ids for the pref branch fields.

PR #2137, opened 2020-01-07
Show blank pref forms when data exists fixes #2136

### LINKED ISSUE #2136: If experiment already has design info, and user switches to multipref, no blank pref form
I realized that this is why I kept on asking teammates if they were seeing blank pref forms, and they were saying yes, and I wasn't seeing them.

If a user has already filled out design form info and saved it, or if a factory has already filled out that info, and then the user tries to switch to multi pref, blank forms do not show. When looking at the serializer code, it's clear why this is happening... we prepopulating with blank pref branches _only inside_ of an if statement that reads: `if data == []:`... so in the case where there was design form data already, data was not empty... so we weren't running the code that prepopulates with blank. 

this is on dev:
Screen Shot 2020-01-07 at 9 24 53 AM

uhlissuh:


uhlissuh:
@tiftran, I think that actually, my single pref tests PR will address this, because I have a test that checks that if you click on the radio button, a blank for shows. That PR is currently red, failing in this exact test, and I'm not sure why yet.

uhlissuh:
@jaredlockhart this is what we were talking about in our meeting, I'm going to move this code into my single pref tests and then close this PR. :)

uhlissuh:
This bug fix is going to be rolled into another PR.

PR #2139, opened 2020-01-08
added error handling for swapping variants

tiftran:
Made it so that the error was somewhat generalized...
Screen Shot 2020-01-07 at 3 16 42 PM


app/experimenter/experiments/serializers/design.py
```diff
@@ -198,32 +199,40 @@ def is_variant_valid(self, variants):
 
         return unique_names and non_empty
 
+    @transaction.atomic
```
jaredlockhart: Do you have to use the decorator and the with statement? I think you only need one?

app/experimenter/experiments/serializers/design.py
```diff
@@ -198,32 +199,40 @@ def is_variant_valid(self, variants):
...
+                    ExperimentVariant.objects.filter(id__in=removed_ids).delete()
 
-            if removed_ids:
-                ExperimentVariant.objects.filter(id__in=removed_ids).delete()
+            self.update_changelog(instance, validated_data)
 
-        self.update_changelog(instance, validated_data)
+            return instance
+        except IntegrityError as e:
+            error_string = "Experimenter Error Occured: {}".format(e)
```
jaredlockhart: Eh I would just make some pretty text like 'unable to save this change, please contact an experimenter administrator' or something.  Users don't know what db integrity errors mean.

PR #2141, opened 2020-01-09
 convert recipe serializers to declarative fixes #2120 

### LINKED ISSUE #2120: Convert all recipe serializers to purely declarative
Rather than using method overrides, which can cause inconsistencies and cuase us to forget to use the write pref type/value serializers, we should switch to using only field declarations for all recipe serializers.

tiftran:


app/experimenter/experiments/serializers/recipe.py
```diff
@@ -11,6 +11,23 @@
 from experimenter.experiments.serializers.entities import PrefTypeField
 
 
+class PrefValueField(serializers.Field):
+
+    def to_representation(self, obj):
+        model = type(obj).__name__
+        if model == "ExperimentVariant":
+            pref_type = obj.experiment.pref_type
```
jaredlockhart: I realized this morning this was gonna be an issue.  This solution works but is a bit inflexible.  You could just pass the field name lookup as a string and use getattr? And that way you can specify the look up path at the field instantiation point and if we ever need to change it we change it there?

tiftran: Created a separate issue https://github.com/mozilla/experimenter/issues/2149 for this. Will be merging it, that way it doesn't get blocked to be pushed up to prod due to me tinkering with it.

PR #2144, opened 2020-01-09
Add Pref Rollout js tests fixes #2085

### LINKED ISSUE #2085: JS Tests for Pref Rollout Design Page

jaredlockhart:


app/experimenter/static/js/tests/PrefRolloutForm.test.js
```diff
@@ -0,0 +1,174 @@
...
+    const design = "it's my design description value";
+    const prefType = "boolean";
+    const prefName = "browser.enabled";
+    const prefValue = "true";
+
+    const designInput = getByTestId("design");
+    const prefTypeInput = getByTestId("prefType");
+    const prefNameInput = getByTestId("prefName");
+    const prefValueInput = getByTestId("prefValue");
+
```
uhlissuh: You could use getByLabelText instead of getByTestId right? If you need to, due to other text like "required" being in the label, you can pass an expression, then you wouldn't have to add that prop to your DesignInput. Like this:

`const prefNameInput = getByLabelText(/Pref Name/)`

jaredlockhart: Oh good idea!!  Yeah I don't like adding these test specific ids, feels unnecessary.

PR #2169, opened 2020-01-14
added swagger ui api docs fixes #1625

### LINKED ISSUE #1625: Add swagger API auto documentation
Becuase I forget to update the API docs *every single time* I make a change, let's add swagger auto API documentation.

https://www.django-rest-framework.org/topics/documenting-your-api/

tiftran:
Made another attempt at swagger, since CoreApi seems to be deprecated.
Screen Shot 2020-01-13 at 7 52 52 PM

Went with [ dynamic](https://www.django-rest-framework.org/api-guide/schemas/#generating-a-dynamic-schema-with-schemaview) schema generation rather than [ static](https://www.django-rest-framework.org/api-guide/schemas/#generating-a-static-schema-with-the-generateschema-management-command) (refs links just in case, i'm confusing). Understanding is with dynamic, every time the template is called/viewed, schemas are generated at that point?


app/experimenter/templates/swagger-ui.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
```
tiftran: Straight up c+p from https://www.django-rest-framework.org/topics/documenting-your-api/#a-minimal-example-with-swagger-ui

jaredlockhart:
@tiftran Isn't there some way to actually generate the docs as a bunch of static files we can dump into a `/docs` folder instead of hosting it from within the django app?

app/experimenter/base/management/commands/generate-docs.py
```diff
@@ -0,0 +1,51 @@
...
+            output.seek(0)
+            api_json = output.read()
+
+            doc_rendered = render_to_string(
+                "swagger-ui-template.html", context={"spec": api_json}
+            )
+
+            with open("/app/experimenter/docs/openapi-schema.json", "w+") as f:
+                f.write(api_json)
+                f.close()
```
jaredlockhart: See if the context manager will close for you (I think it might)

app/experimenter/base/management/commands/generate-docs.py
```diff
@@ -0,0 +1,51 @@
...
+    def handle(self, *args, **options):
+        self.generate_docs(options)
+
+    @staticmethod
+    def generate_docs(options):
+        if options["check"]:
+            output = io.StringIO()
+            call_command("generateschema", "--format=openapi-json", stdout=output)
+            output.seek(0)
+            api_json = output.read()
```
jaredlockhart: These four lines are duplicated verbatim, could pull this out into a function.

app/experimenter/base/management/commands/generate-docs.py
jaredlockhart: See if the context manager will close for you (I think it might)

app/experimenter/base/management/commands/generate-docs.py
```diff
@@ -0,0 +1,51 @@
...
+
+            with open("/app/experimenter/docs/openapi-schema.json", "r") as f:
```
jaredlockhart: These paths should be relative to project root and probably stored in settings.py

app/experimenter/base/management/commands/generate-docs.py
```diff
@@ -0,0 +1,51 @@
...
+            output = io.StringIO()
+            call_command("generateschema", "--format=openapi-json", stdout=output)
+            output.seek(0)
+            api_json = output.read()
+
+            doc_rendered = render_to_string(
+                "swagger-ui-template.html", context={"spec": api_json}
+            )
+
+            with open("/app/experimenter/docs/openapi-schema.json", "w+") as f:
```
jaredlockhart: These paths should be relative to project root and probably stored in settings.py

PR #2193, opened 2020-01-22
added/edited js tests for 100% line coverage

tiftran:
Line coverage is at 100% and coverage is turned on to expect 100% line coverage, wasn't sure if we wanted to consider other types of coverage.

Current coverage summary for this PR:
Screen Shot 2020-01-22 at 3 19 57 PM

I made very minor tweaks to existing tests so that they are more uniform. We also have variations of using data test id or the for attribute for tests which i wasn't entirely sure what to do with... but thought this PR could be a start


app/experimenter/static/js/tests/DataFactory.js
```diff
@@ -18,7 +18,7 @@ export class VariantsFactory extends Factory {
 export class GenericDataFactory extends Factory {
   getFields() {
     return {
-      designs: new Field(faker.lorem.paragraph),
+      design: new Field(faker.lorem.paragraph),
```
jaredlockhart: LOL how did this ever pass

tiftran:
@jaredlockhart @uhlissuh I had to make tweaks to the integration tests to get this PR to pass integration checks.

I had to clear the percentage field before it sets a new value, or else the result is:

Screen Shot 2020-01-23 at 2 02 07 PM

Couldn't narrow down, why or when did this start happening as there hasn't been any commits that changed the tests. Also couldn't get it to fail locally either...

app/experimenter/static/js/tests/DataFactory.js
tiftran: 🙈

PR #2194, opened 2020-01-23
Add progress bar fixes #2171

### LINKED ISSUE #2171: Experiment state progress bar
Let's add a nice big progress bar at the top of the experiment detail page that shows all the states an experiment can be in and which one it's in now, so that the whole workflow is more clearly laid out for users.  I'm picturing something like this:

but with buttons on teh far left and far right for moving to the previous and next state, ie the buttons that we currently have at the top right.

So for an experiment currently in the Sign Off state, I think it would be something like

```
                
<<Return to Draft<<           [Draft][Sign Off][Ready to Ship][Accepted][Live][Complete]       >>Ready to Ship>>      
                            *******+++++++++++-------------------------------------------
```

Note: this isn't a literal mockup, i'm just using * + - to denote which ones already completed, which one we're currently in, and which ones haven't been done yet.

where the Draft state is filled in that it's done, Sign Off is highlighted as the current state, and all the ones after are greyed out that we haven't hit them yet.

In theory we could also put dates next to each state that are done, but don't worry about that for mvp.

Also I think the links we have across teh top should move to the right sidebar (bugzilla, dashboard, etc).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-124)

uhlissuh:
Screen Shot 2020-01-22 at 4 28 47 PM
Screen Shot 2020-01-22 at 4 28 19 PM
Screen Shot 2020-01-22 at 4 28 05 PM


app/experimenter/templates/experiments/detail_base.html
```diff
@@ -31,38 +31,30 @@ <h3>
...
-        </a>
-      {% endif %}
+{% block progress_bar %}
+<div class="container">
+  <div class="row my-4">
+    <div class="col-9 h-100">
+      <div class="row whole-bar">
+        {% for status in experiment.progress_bar_highlighted %}
+          {% if status.1 %}
+            <div class="col-2 bar-piece-highlighted">
```
jaredlockhart: You can actually put the if statement right inside the class attribute since the rest of the body of this is the same like

`class="col-2 bar-piece{% if status.1 %}-highlighted{% endif %}"`

app/experimenter/templates/experiments/detail_base.html
```diff
@@ -107,11 +99,61 @@ <h3>
 
 {% endblock %}
 
+
 {% block main_sidebar %}
   <h4>&nbsp;</h4>
 
-  {% block main_sidebar_buttons %}
-  {% endblock %}
+  {% if experiment.has_external_urls %}
```
jaredlockhart: Gotta stick some bottom margin on this to match the other sidebar things.

uhlissuh:
@jaredlockhart Hey Jared, I think I mostly get all that you are asking for here, except for this: 



are you meaning that have the bars just colored at one edge?

jaredlockhart:
@uhlissuh Sorry what I mean is:

Each progress block should have the colour of its status since I tried to make it so everywhere that a status is referred to, the right colour is used (I'm not a designer but I thought using colours to indicate status was maybe a designy thing?)

So you could fill in each progress block with the corresponding status colour instead of the blue.  If that looks okay then great.

But if that is too visually intense, you could instead just add the colour as a highlight along a border as in the example above.  

It's up to you, play around with it and see what looks nice, the outcome I'm looking for is that each progress bar block has the colour of the status it represents 😸

app/experimenter/templates/experiments/detail_base.html
```diff
@@ -31,38 +31,30 @@ <h3>
...
-      {% if experiment.monitoring_dashboard_url %}
-        <a class="alert-link" target="_blank" rel="noreferrer noopener" href="{{ experiment.monitoring_dashboard_url }}">
-          <span class="fas fa-chart-bar"></span>
-          Live Monitoring Dashboard
-        </a>
-      {% endif %}
+{% block progress_bar %}
+<div class="container">
+  <div class="row my-4">
+    <div class="col-9 h-100">
```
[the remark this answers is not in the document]
uhlissuh: removed

uhlissuh:
@jaredlockhart I'm curious as to why you don't like the previous button to where I moved it... It utilizes the space created by the "Overview" title... and to me, it's placement under the progress bar indicates a "going back-ness" really well... it's also a bigger hassle, spacing wise, to move it under the other button, because it will create dead space between the bar and the overview section. You could put the buttons in the main_sidebar and then make that be a column that goes all the way to the top, which would eliminate the spacing weirdness I just mentioned, but when there aren't buttons, the links would then be up way higher than the overview section.....

jaredlockhart:
@uhlissuh Yeah I see what you mean I just confused when I was actually playing with it because I expected the button to be at the top right so for a moment i was like THERE IS NO BACK BUTTON HALP.  But the more I looked at it and played with it and talked to shell about it, the more I realized really the problem we need to solve is that the 'continue to X' button is what's confusing to people, because they think it's a status indicator and not actually a button, so I thought if we changed the layout to make it clearer where they are and which things move them through it that would stand out more. 

I still think it's worth trying to make the 

    <<BACK        STATUS|STATUS|STATUS|DISBLED|DISABLED|DISABLED    NEXT>>

layout try to work, what do you think?

Also it occurred to me we could highlight the current status a bit more if we add a CSS glow on it so it's super clear which status you're currently on?

uhlissuh:
@jaredlockhart took over this progress bar work.

Issue #2248, opened 2020-02-04
Add some wording to increase accurate answers to this risk question...


shell1:
Based on feedback from jdavis on the pocket team, it was requested we add some wording to one risk questions to help prompt people to fully consider possible revenue impacts.

Old: Does this delivery have possible negative impact on revenue? 

New: Does this delivery have possible negative impact on revenue (ex: search, pocket revenue, New Tab page experience)?

PR #2251, opened 2020-02-05
risk question edits simplify duplication fixes #2246

### LINKED ISSUE #2246: simplifying duplicate risk questions
combine:
 Is this delivery sensitive and/or internal only? Help 
 Is this delivery confidential to Mozilla? Help 

to:

Is this delivery confidential to Mozilla, sensitive, or internal only?  Help

________________
Keep the behavior of the question the same as "Is this delivery confidential to Mozilla? Help"

tiftran:
fixes #2246 and #2248 

not sure if we should consider a data migration that ORs risk_internal_only and risk_confidential


app/experimenter/experiments/migrations/0081_remove_experiment_risk_internal_only.py
```diff
@@ -0,0 +1,22 @@
...
+from django.db import migrations
+from django.db.models import Q
+
+
+def update_risk_confidential(apps, schema_editor):
+    Experiment = apps.get_model("experiments", "Experiment")
+    experiments = Experiment.objects.filter(
+        Q(risk_confidential=True) | Q(risk_internal_only=True)
+    )
+    experiments.update(**{"risk_confidential": True})
```
jaredlockhart: Isn't this superfluous?  Can't you just say `.update(risk_confidential=True)`?

jaredlockhart: And you can tack the `.update` onto the end of the `.filter` and don't need the variable at all?

PR #2292, opened 2020-02-18
Explicitly set postgres password and version fixes #2291

### LINKED ISSUE #2291: Explicitly set POSTGRES_PASSWORD=postgres for local/ci
The latest postgres docker image `9.16.17` removes the default for `POSTGRES_HOST_AUTH_METHOD=trust` which means we either need to explicitly set that or the password, so we should just set the password.

jaredlockhart:
Looks like the postgres docker image default security behaviour has changed: https://github.com/docker-library/postgres/pull/658 so we gotta explicitly set the password now, and we should pin to a specific version so that random changes like this don't sneak in and start breaking local dev or ci.

### DIFF
--- a/docker-compose-full.yml
+++ b/docker-compose-full.yml
@@ -43,10 +43,12 @@ services:
       - public_nw
 
   normandy_db:
-    image: postgres:9.6
-    environment:
-      POSTGRES_DB: normandy
+    restart: always
+    image: postgres:9.6.17
     ports:
       - "5431:5432"
     networks:
       - private_nw
+    environment:
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: normandy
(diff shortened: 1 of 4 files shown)


docker-compose-full.yml
```diff
@@ -43,10 +43,10 @@ services:
...
   normandy_db:
-    image: postgres:9.6
-    environment:
-      POSTGRES_DB: normandy
+    restart: always
+    image: postgres:9.6.17
     ports:
       - "5431:5432"
-    networks:
-      - private_nw
```
jaredlockhart: oops that shouldn't be removed...

docker-compose-test.yml
```diff
@@ -7,7 +7,6 @@ services:
     environment:
       - DEBUG=False
     volumes:
-      - ./app:/app
```
jaredlockhart: I realized we don't need to volume map this in during the test run since it always copies the freshest version at build time.

PR #2312, opened 2020-02-25
Convert timeline/population to react fixes #2119

### LINKED ISSUE #2119: Convert timeline/pop page to react

uhlissuh:
not much to show here in the way of screenshots, as the goal was to make it look the same. Main UI difference is the multiselects. 
Screen Shot 2020-02-27 at 11 53 49 AM
multiselect


tests/integration/pages/experiment_timeline_and_population.py
```diff
@@ -126,30 +125,6 @@ def firefox_max_version(self, version=None):
...
-    def locale(self, locale=None):
-        element = self.find_element(*self._locale_btn_locator)
-        self.selenium.execute_script("arguments[0].click();", element)
-        locales = self.selenium.find_elements(By.CSS_SELECTOR, "div.show:nth-child(3) a")
-        for item in locales:
-            if locale in item.find_element(By.CSS_SELECTOR, "span.text").text:
-                self.selenium.execute_script("arguments[0].click();", item)
-                break
-        self.selenium.execute_script("arguments[0].click();", element)
-
```
uhlissuh: In this PR I've eventually landed on taking out the locales test. I couldn't figure out how to write a selenium test to grab the right elements. I think it's made much more complicated by this the element being made by a third party lib, "react-select".. there are many divs this library creates for the element with weird autogenerated class names... essentially there's a lot of divs that I can't really control. I can't add helper classes to things to make the grabbing of elements easier. I'm wondering... do we need to test the locales field box, especially when we aren't testing every field on this page with integration tests? I test the react-select element function in two other places in the front-end tests. If we really want to integration test the selects on this page, perhaps Ben can help me on this, since he's the expert here.

app/experimenter/experiments/tests/test_api_views.py
```diff
@@ -552,3 +555,46 @@ def test_put_to_view_design_is_optional(self):
...
+        )
+
+        self.assertEqual(response.status_code, 200)
+
+        json_data = json.loads(response.content)
+
+        serialized_experiment = ExperimentTimelinePopSerializer(experiment).data
+        self.assertEqual(serialized_experiment, json_data)
+
+    def test_put_to_view_saves_design_info(self):
```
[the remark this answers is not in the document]
uhlissuh: oops yes c/p

app/experimenter/experiments/serializers/timeline_population.py
```diff
@@ -0,0 +1,138 @@
...
+        required=False, allow_null=True, default=None
+    )
+    proposed_enrollment = serializers.IntegerField(
+        required=False, allow_null=True, default=None
+    )
+    population_percent = serializers.DecimalField(
+        required=False,
+        max_digits=7,
+        decimal_places=4,
+        max_value=100.0000,
```
jaredlockhart: I think you also need a min_value here?

app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -0,0 +1,122 @@
...
+        serializer = ExperimentTimelinePopSerializer(self.experiment)
+
+        self.assertEqual(
+            serializer.data["proposed_start_date"],
+            self.experiment.proposed_start_date.strftime("%Y-%m-%d"),
+        )
+        self.assertEqual(
+            serializer.data["firefox_channel"], self.experiment.firefox_channel
+        )
+        self.assertEqual(serializer.data["platform"], self.experiment.platform)
```
jaredlockhart: Why does this test only check these three fields?

app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -0,0 +1,122 @@
...
+        locale = LocaleFactory()
+        countries = [{"value": country.id, "label": country.name}]
+        locales = [{"value": locale.id, "label": locale.name}]
+
+        data = {
+            "firefox_min_version": "67.0",
+            "firefox_max_version": "68.0",
+            "countries": countries,
+            "locales": locales,
+        }
```
jaredlockhart: Why does it only save these four fields as opposed to all of them?

app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -0,0 +1,122 @@
...
+        data = {
+            "firefox_min_version": "69.0",
+            "firefox_max_version": "68.0",
+            "countries": [],
+            "locales": [],
+        }
+
+        serializer = ExperimentTimelinePopSerializer(instance=self.experiment, data=data)
+
+        self.assertFalse(serializer.is_valid())
```
jaredlockhart: Maybe just check that the min/max fields are in the errors list here?

app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def test_serializer_rejects_date_in_past(self):
+        data = {"proposed_start_date": "2019-2-01", "countries": [], "locales": []}
```
jaredlockhart: Don't change this, that'll always pass (unless the server falls into a time warp), but in general for testing dates like this I usually do something like `datetime.date.today() - datetime.timedelta(wahtever)`

app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -0,0 +1,122 @@
...
+
+        serializer = ExperimentTimelinePopSerializer(instance=self.experiment, data=data)
+
+        self.assertFalse(serializer.is_valid())
```
jaredlockhart: Maybe check that the start date field is in errors here?

app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -0,0 +1,122 @@
...
+        data = {
+            "countries": [],
+            "locales": [],
+            "proposed_enrollment": 40,
+            "proposed_duration": 30,
+        }
+
+        serializer = ExperimentTimelinePopSerializer(instance=self.experiment, data=data)
+
+        self.assertFalse(serializer.is_valid())
```
jaredlockhart: Same here, check for fields?

app/experimenter/static/js/components/LabeledMultiSelect.js
```diff
@@ -0,0 +1,115 @@
...
+    label: PropTypes.string,
+    name: PropTypes.string,
+    note: PropTypes.string,
+    onChange: PropTypes.func,
+    rows: PropTypes.string,
+    value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+    type: PropTypes.string,
+    optional: PropTypes.bool,
+    options: PropTypes.array,
+    placeholder: PropTypes.string,
```
jaredlockhart: Looks like we only pass some of these into the Select down below, does it get cranky if you don't specify all of these?  Can you specify only the ones that we pass into the Select?

app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -0,0 +1,372 @@
...
+    if (value == "") {
+      value = null;
+    }
+    this.setState(({ data }) => ({
+      data: data.set(key, value),
+    }));
+  }
+
+  handleMultiSelectDataChange(selectedOptions, name) {
+    if (selectedOptions == null) {
```
jaredlockhart: Should be a `===` check here?  I'm surprised the linter didn't complain about this, usually in JS you're supposed to always use `===` I think?

app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  handleSubmitSave(event) {
+    this.handleSubmit(event, `/experiments/${this.props.slug}/`);
+  }
+
+  handleSubmitContinue(event) {
+    this.handleSubmit(
+      event,
+      `/experiments/${this.props.slug}/edit-objectives/`,
```
jaredlockhart: Should be design here

app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -0,0 +1,372 @@
...
+              onChange={value => {
+                this.handleDataChange("firefox_channel", value);
+              }}
+              value={this.state.data.get("firefox_channel")}
+              error={this.state.errors.get("firefox_channel", "")}
+              helpContent={CHANNEL_HELP}
+              labelColumnWidth={2}
+              helpIsExternalLink={true}
+            >
+              <option>Firefox Channel</option>
```
jaredlockhart: Is there a way to mark this option as disabled?  I realized recently we had a similar issue on the design page where you can set these back to the placeholder label and it'll save.

app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -0,0 +1,372 @@
...
+    if (selectedOptions == null) {
+      selectedOptions = [];
+    }
+    this.setState(({ data }) => ({
+      data: data.set(name, fromJS(selectedOptions)),
+    }));
+  }
+
+  displayPopulationPercent() {
+    if (this.props.shouldHavePopPercent == "True") {
```
jaredlockhart: 1) use `===` comparisons
2) Is there a way to do this that uses the actual boolean value and not a string representation of a Python bool?

tests/integration/pages/experiment_timeline_and_population.py
jaredlockhart: Yeah I think it's okay to drop these, @jrbenny35 is adding a bunch of new tests that will cover filling out all the forms and walking through the rest of the flow, and as those land I think we'll be able to drop all these field specific tests, so ya this is fine.

app/experimenter/experiments/serializers/timeline_population.py
uhlissuh: yes.

app/experimenter/experiments/tests/serializers/test_timeline_population.py
uhlissuh: I'll make it check all of them.

app/experimenter/static/js/components/LabeledMultiSelect.js
uhlissuh: I took out the ones that weren't actually applicable here. I probably started with a big list copied from another component's prop types and didn't take out all the inapplicable ones.

app/experimenter/static/js/components/TimelinePopForm.js
uhlissuh: I looked into it more, and if you use `==` with null, it will also check if it's undefined. Which in this case I probably want. But the others should be changed to `===`.

app/experimenter/static/js/components/TimelinePopForm.js
uhlissuh: oopsie.

app/experimenter/static/js/components/TimelinePopForm.js
jaredlockhart: Interesting, good to know!  I'm always learning about the weird dark magic of equality comparisons in JS

app/experimenter/static/js/components/TimelinePopForm.js
uhlissuh: Disabling that option could work, but isn't it OK that it saves when someone sets it back to Firefox Channel? Don't we want someone to be able to decide later that they want to leave that blank for a while? We just want to make sure that this part of the form isn't considered done if the value of channel is "Firefox Channel", right? Maybe the fix is to make sure that the model method that checks for the population form doneness checks that the value isn't "Firefox Channel"?

I can make it so that "Firefox Channel" isn't clickable, but again the end result of that will be once someone chooses a channel, they won't ever be able to set it back to what is essentailly supposed to be "blank" again.

app/experimenter/static/js/components/TimelinePopForm.js
uhlissuh: I addressed 1) elsewhere.

2) yes there was a way to do it.. in `edit_timeline_population.html`, I elected not to pass a value to the div if should_have_population_percentage was false, thus allowing the JS in TimelinePopForm to read:

`if (this.props.shouldHavePopPercent) {...}`

app/experimenter/static/js/components/TimelinePopForm.js
jaredlockhart: Yeah I see your point.  I don't think there's a case where we'd want someone to be able to set it back to blank.  Like they have to choose one, so they could go back and set it to another channel.

But it's also different on this page than design, because on design you have to fill everything out before you can save whereas here you can't, so if we'd have to make it so that the option is disabled (ie not selectable after they choose a value) but can still save if they havne't chosen anything yet, and those two things might conflict?

In which case the solution is probably to leave it selectable and do the thing we talked about elsewhere which is just highlight it as an incomplete field if they come back to the page.

In either case don't worry about it here, just leave it as is and we can find a better solution in a separate PR.  Thanks for thinking it through with me :)

app/experimenter/static/js/components/TimelinePopForm.js
jaredlockhart: Yeah but isn't this case different for the equality comparison than the null case?  You do want the `===` strictness here?

Also that seems good to me, why did you switch it to this?

uhlissuh: yes yes, you are right, the `===` still needs to be used if i left it as a string.

I switched it to the above way because it seemed better than having the `this.props.shouldHavePopPercent == "True"` stuff.

app/experimenter/static/js/components/TimelinePopForm.js
uhlissuh: ok sounds good.

uhlissuh:
Hey @jaredlockhart, I think I have addressed every change you requested.

PR #2313, opened 2020-02-25
added multipref for rollout fixes #2278

### LINKED ISSUE #2278: Add Multipref option for pref rollouts
We didn't realize when it was designed that people would want to rollout multipref.  Since that is true - if it's a pref rollouts - they need the ability to chose single or multipref so that populates over to Normandy (same as experiments).

add-ons branching is not a thing in rollouts - so we don't need any changes for add-ons.

tiftran:
Screen Shot 2020-02-25 at 10 35 07 AM


tiftran:
Gah, these tiny things were so silly 🙈 

Made pref branch as default only

Added a Pref Type check for pref rollouts and multipref. Added a Pref Branch check for multipref, so it will complain if nothing is selected

The weird saving values and then disappearing while trying to edit them was due to a type in the logic for the set difference between existing ids and submitted ids. [here](https://github.com/mozilla/experimenter/pull/2313/commits/23cee96276dd5b0a95c159830697696948e9b1a6#diff-a6a941d0695873dea4026bcf7b7a185cL313)

The weirdness for the detail page showing None=None, was because I forgot to change it... 

In addition to the changes above, I took out pref_branch out of the RolloutPreference model

app/experimenter/experiments/serializers/design.py
```diff
@@ -237,52 +267,80 @@ def update(self, instance, validated_data):
...
+        preferences_data = validated_data.pop("preferences", [])
+        instance = super().update(instance, validated_data)
+
+        if preferences_data:
+            existing_preference_ids = set(
+                instance.preferences.all().values_list("id", flat=True)
+            )
+            for preference_data in preferences_data:
+                preference_data["experiment"] = instance
+                RolloutPreference(**preference_data).save()
```
jaredlockhart: **Opinion:**

This works, you can also do `RolloutPreference.objects.create(**preference_data)`.  I'm not sure it's all that different than this, I guess this just creates an instance and then immediately refcounts it to 0.

app/experimenter/experiments/tests/serializers/test_recipe.py
```diff
@@ -200,57 +203,122 @@ def test_serializer_outputs_expected_schema(self):
...
+
+    def test_serializer_outputs_expected_schema_for_multi_pref(self):
+
+        experiment = ExperimentFactory.create(
+            type=Experiment.TYPE_ROLLOUT,
+            normandy_slug="normandy-slug",
+            rollout_type=Experiment.TYPE_PREF,
+        )
+
+        preference = RolloutPreference(
```
jaredlockhart: **Opinion:**

I woulda called these preference1 and preference2 but that's just a style thing

app/experimenter/experiments/serializers/design.py
```diff
@@ -33,7 +34,23 @@ def update_changelog(self, instance, validated_data):
...
+    def validate_multi_preference(self, pref):
+        pref_type = pref["pref_type"]
+        pref_value = pref["pref_value"]
+        field_name = "pref_value"
+        errors = {}
+        if pref_type == "Firefox Pref Type":
+            errors["pref_type"] = "Please select a pref type"
+
+        errors.update(self.validate_pref_value(pref_type, pref_value, field_name))
+        return errors
```
jaredlockhart: **Opinion:**

Just a super personal annoying style thing, for this type of pattern I usually do something like

        def something():
            thingtoReturn = {}  # initialize default return at the top of the function
            # buncha logic
            if whatever:
                thingToReturn[whatever] = whatever
            # buncha logic
            return thingToReturn  # last line returns the variable defined at the top


This makes it easier to go back and move things around without modifying that outer wrapping of setting up the default case and always returning it

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1527,6 +1534,57 @@ def test_is_json_string_type_returns_true_for_json_string_type(self):
...
+        pv2.save()
+
+        self.assertTrue(self.experiment.preferences.count, 2)
+
+    def test_valid_pref_is_able_to_update_itself(self):
+        self.pref.value = 6
+
+        self.pref.save()
+
+        self.assertTrue(self.experiment.preferences.count, 1)
```
jaredlockhart: This is all stuff built into Django and the Django test suite has tests that cover this, I don't think this is necessary.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1527,6 +1534,57 @@ def test_is_json_string_type_returns_true_for_json_string_type(self):
...
+class TestRolloutPreference(TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.experiment = ExperimentFactory.create()
+        self.pref = RolloutPreference()
+        self.pref.experiment = self.experiment
+        self.pref.pref_type = Experiment.PREF_TYPE_BOOL
+        self.pref.pref_value = "true"
+        self.pref.pref_name = "pref_name"
```
jaredlockhart: Should this be a RolloutPreferenceFactory?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1527,6 +1534,57 @@ def test_is_json_string_type_returns_true_for_json_string_type(self):
...
+
+    def test_valid_pref_is_able_to_update_itself(self):
+        self.pref.value = 6
+
+        self.pref.save()
+
+        self.assertTrue(self.experiment.preferences.count, 1)
+
+    def test_is_json_string_type_returns_false(self):
+        self.assertFalse(self.pref.is_json_string_type)
```
jaredlockhart: The test should setup the data it needs to pass inside the test otherwise we could change some defaults and this would suddenly fail.

app/experimenter/experiments/serializers/design.py
tiftran: yup yup, I agree. I originally wanted to do this, but realized that the singular pref also relies on this and thought it would've been messy to change it. Turns out it wasn't so bad at all 🤷‍♀

app/experimenter/experiments/tests/test_models.py
tiftran: removed

app/experimenter/experiments/serializers/design.py
```diff
@@ -33,23 +34,47 @@ def update_changelog(self, instance, validated_data):
...
+                return "The pref value must be a boolean."
 
         if pref_type == "json string":
             try:
                 json.loads(pref_value)
             except ValueError:
-                return {field_name: "The pref value must be valid JSON."}
-        return {}
+                return "The pref value must be valid JSON."
+
```
jaredlockhart: Have this always return an empty dict or the {field_name: error} dict like the other methods do

app/experimenter/experiments/serializers/design.py
```diff
@@ -142,20 +181,12 @@ def validate_preferences(self, data):
...
-        return unique_names
-
     def validate_value_type_match(self, preferences):
         error_list = []
         for pref in preferences:
-            pref_type = pref.get("pref_type", "")
-            pref_value = pref["pref_value"]
-            field_name = "pref_value"
-            error_list.append(self.validate_pref(pref_type, pref_value, field_name))
+            errors = self.validate_pref_branch(pref["pref_branch"])
```
jaredlockhart: Build the empty errors dict just above this `errors = {}` and then update it with both responses so it doesnt' look like they're order dependent.

PR #2316, opened 2020-02-26
removed data science bugzilla stuff with jira

tiftran:


tests/integration/conftest.py
```diff
@@ -40,5 +40,5 @@ def fill_overview(selenium, base_url):
     experiment.short_description = "Testing in here"
     experiment.public_name = "Public Name"
     experiment.public_description = "Public Description"
-    experiment.bugzilla_url = "http://bugzilla.com/show_bug.cgi?id=1234"
+    experiment.bugzilla_url = "https://jira.mozilla.com/browse/DS-123"
```
jaredlockhart: Hrmmmm I'll need to update my integration test PR for this

tests/integration/test_create_experiment.py
```diff
@@ -14,7 +14,7 @@ def test_add_branch(base_url, selenium):
     experiment.short_description = "Testing in here"
     experiment.public_name = "Public Name"
     experiment.public_description = "Public Description"
-    experiment.bugzilla_url = "http://bugzilla.com/show_bug.cgi?id=1234"
+    experiment.bugzilla_url = "https://jira.mozilla.com/browse/DS-123"
```
jaredlockhart: Leave these like this and we'll land this first and I'll fix it in my PR

.env.sample
```diff
@@ -7,6 +7,7 @@ DB_PASS=postgres
 DB_USER=postgres
 DEBUG=True
 DELIVERY_CONSOLE_HOST=
+DS_ISSUE_HOST=
```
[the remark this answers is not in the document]
tiftran: From lurking in the jira-support channel, I don't think there's a dev instance, there's https://jira.allizom.org which they call staging, but it seems to be more like a space for people to set up jira projects before they get migrated to production

app/experimenter/experiments/forms.py
```diff
@@ -42,6 +44,30 @@ def clean(self, value):
...
+                DS_root not in cleaned_value and DO_root not in cleaned_value
+            ) or self.get_ds_issue_id(cleaned_value) is None:
+
+                raise forms.ValidationError(err_str.format(ds_url=settings.DS_ISSUE_HOST))
+        return cleaned_value
+
+    def get_ds_issue_id(self, bug_url):
+        ds = re.match(re.escape(settings.DS_ISSUE_HOST) + r"(DS|DO)-(\w+.*)", bug_url)
+
+        return ds.group(1)
```
jaredlockhart: Wait what am I thinking, couldn't you do this whole thing with just the regex match looking for the ID and if match is None then we know it's invalid?

.env.sample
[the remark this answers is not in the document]
jaredlockhart: Okay then just stick `https://jira.example.com/` in there, we'll need something in there for integration tests to pass.

app/experimenter/experiments/forms.py
tiftran: being the regex newbie, what do you mean by regex match looking for the ID, like if the url ends in numbers?

jaredlockhart: I mean like if `ds is None`, then the regex failed to find the ID at the end of the URL because it didn't match the regex, so then we know whatever URL was provided isn't a 'valid' jira URL, which accomplishes the same thing as the two `not in` checks above it, so you should be able to remove them and just do the regex check?

.env.sample
```diff
@@ -7,6 +7,7 @@ DB_PASS=postgres
 DB_USER=postgres
 DEBUG=True
 DELIVERY_CONSOLE_HOST=
+DS_ISSUE_HOST=https://jira.mozilla.com
```
jaredlockhart: Nit: `https://jira.example.com/browse/`

app/experimenter/experiments/forms.py
```diff
@@ -42,6 +43,25 @@ def clean(self, value):
...
+
+            if self.validate_ds_issue_url(cleaned_value) is None:
+
+                raise forms.ValidationError(err_str.format(ds_url=settings.DS_ISSUE_HOST))
+        return cleaned_value
+
+    def validate_ds_issue_url(self, bug_url):
+        ds = re.match(re.escape(settings.DS_ISSUE_HOST) + r"(DS|DO)-(\w+.*)", bug_url)
+
+        return ds
```
jaredlockhart: Just inline this above, doesn't need to be a method

PR #2327, opened 2020-03-03
edits to make reviews list in specific order fixes #2325

### LINKED ISSUE #2325: "Engineering allocated" moved to the bottom of the required checklist
The required checklist order got messed up recently.  It should be:

Data Science Sign-Off Help
Lightning Advisory Help
Engineering Allocated Help 
QA Jira Request Sent Help
Intent to Ship Email Sent Help
Bugzilla Updated Help
QA Sign-Off Help
Release Management Sign-Off Help

tiftran:


app/experimenter/experiments/models.py
```diff
@@ -674,16 +674,22 @@ def _conditional_required_reviews_mapping(self):
         }
 
     def _default_required_reviews(self):
+        # Ref: #2325, request for specific order of reviews
```
jaredlockhart: Eh I don't think you need to leave comments like this hanging around

jaredlockhart: In general the only time I think I see things like this are if there's something that needs changing and you leave a TODO: #bugnumber, but in this case the git history will record the commit where that line was changed which will refer to the bug number so I don't think you need to leave a reference right in the code?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1260,6 +1260,33 @@ def test_is_ready_to_launch_is_false_without_conditional_review(self):
 
         self.assertFalse(experiment.is_ready_to_launch)
 
+    def test_review_order_is_correct_for_experiment(self):
+        experiment = ExperimentFactory.create()
```
jaredlockhart: I'd be explicit about the experiment type here.

jaredlockhart: One thing we see a bunch is tests where we rely on the defaults of the factory to be valid for the test then we change those defaults and a test fails that shouldn't, so we should try to always be explicit about the required values of a factory for a given test.  

I think if we go ahead with that change to make the factories more random that'll expose a lot more of these cases.

app/experimenter/experiments/models.py
tiftran: right right completely forgot about git history

Issue #2317, opened 2020-02-26
Add docs to run integration-tests


jrbenny35:
There are currently no docs on how to run the integration-test suite. We should fix that by adding docs on how they can be run locally.

jrbenny35:
Here is what I have so far:
```
# Experimenter Integration Tests

### Prerequisites
- Docker
- Docker-Compose
- make
- [geckodriver]
- Python 3
- [tox]

### Setup using docker and make

To run the tests via docker-compose use the following make commands.

- Start and run the tests
```
make integration_test_run
```
You can view the tests running via VNC and port 5900. The password to connect to the container is `secret`.
To trigger the tests within the container:
```
docker-compose -p experimenter_integration -f docker-compose.yml -f docker-compose.integration-test.yml exec firefox sh -c "tox -c tests/integration"
```
With the VNC window open, you should see Firefox open and Selenium executing commands.

### Setup using local Firefox

Install [geckodriver] in your path. You should be able to do `geckodriver --version` from your terminal if your path is correct. Firefox must also be available from
your path. To verify this use the following command: `firefox --version`. If there is no output check your configuration.

Now navigate to the directory that you have cloned experimenter into and follow these steps:

1. Within the experimenter root directory:
```
mkdir python_env
python3 -m venv python_env
source python_env/bin/activate
```
If successful your shell should have an `(python_env)` prefix.
2. Install Tox
```
pip install tox
```
3. Start the tests
```
tox -c tests/integration/tox.ini
```
Firefox should open and the tests should execute.

### Extras

- To run a single test with tox you must pass the arguments through tox using `--` and then use the _keyword_ argument within pytest.
Ex: `tox -c tests/integration -- -k name_of_test`
- To deactivate the virtual env: `source deactivate`
- Setup and start containers but do not run tests: `make integration_up_detached`
- Setup and start containers but do not run tests and show all container messages: `make integration_up`
- Kill all containers: `make kill`


[geckodriver]: https://github.com/mozilla/geckodriver/releases
[tox]: https://tox.readthedocs.io/en/latest/
```

uhlissuh:
Hey I edited your comment because it had an errant "```" that was messing up the formatting... hope you don't mind. 

[...]

PR #2354, opened 2020-03-04
Simplify integration test workflow fixes #1914

### LINKED ISSUE #1914: make integration_test should *just work*
Right now you have to remember to `make kill` before you start `make integration_test`, and then after the integration tests are done their containers hang around, so we should make that *just work*.

jaredlockhart:
Okay! So I made some changes to the integration test workflow to hopefully make things simpler to use.

Changes:

  - Reuse the standard local dev for running the integration tests instead of a separate copy
  - Change `make kill` to REALLY kill everything it can
  - Setup the vnc server to disable headless mode so you can watch the tests running
  - Had to take out one of the tests because I couldn't think of a good way to get it passing while reusing the local dev instance (@jrbenny35 if you have an idea of how to keep that test around let me know and I'll put it back otherwise I'd rather lose a test to get this workflow working smoothly)

How to test this:
  - Manually kill/remove all your containers with
        `docker ps -aq | xargs docker kill`
        `docker ps -aq | xargs docker rm`
        `docker volume ls -q | xargs docker volume rm`
  - Refresh to create a new db
        `make refresh`
  - Start the regular dev server with either of
        `make up`
        `make up_detached`
  - Check that the local dev is running
        [https://localhost/](https://localhost)
  - Run the headless integration tests in a separate shell
        `make integration_test`
  - Start the vnc server with either of
        `make integration_vnc_up`
        `make integration_vnc_up_detached`
  - Connect to vnc server with password `secret`
        `vnc://localhost:5900`
  - Right click on the linux desktop and go to Applications > Shell > Bash and enter
        `tox -c tests/integration/`
  - This should run all the tests and  you should be able to see the browser running them
  - Close vnc
  - Check your local dev is still running and full of dummy integration test experiments nowo
        [https://localhost/](https://localhost)
  - You should also be able to use @jrbenny35 's instructions to run the tests locally against the same dev instance: https://github.com/mozilla/experimenter/issues/2317

So this way you can keep your local dev running the whole time and then hit it with headless/headful(?) integration tests in any order/combination.  I think this should make using/maintaining/modifying the integration tests a lot simpler hopefully 🤞 

Let me know if it breaks!


jaredlockhart:
Blocked on conflicts with https://github.com/mozilla/experimenter/pull/2316/files#

We should land that one first and then I'll update this

Makefile
```diff
@@ -128,43 +135,32 @@ bash: compose_build
...
-
-integration_up: integration_build
-	$(COMPOSE_INTEGRATION) up
+integration_vnc_up: integration_build
+	$(COMPOSE_INTEGRATION) up vnc
 
-integration_test_run: integration_build
-	$(COMPOSE_INTEGRATION) run firefox tox -c tests/integration
+integration_vnc_up_detached: integration_build
+	$(COMPOSE_INTEGRATION) up vnc -d
```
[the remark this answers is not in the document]
jaredlockhart: Good catch!!  Thanks for looking @jrbenny35 !

jaredlockhart:
Also it just occurred to me I should add those exact instructions to the README or something.

tests/integration/conftest.py
```diff
@@ -15,7 +22,7 @@ def capabilities(capabilities):
 @pytest.fixture
 def firefox_options(firefox_options):
     """Set Firefox Options."""
-    firefox_options.headless = True
+    firefox_options.headless = not bool(os.environ.get("DISABLE_HEADLESS"))
```
jrbenny35: We can get rid of this line and just use the env variable ```MOZ_HEADLESS```.

docker-compose.integration-test.yml
```diff
@@ -1,23 +1,24 @@
...
+      - .:/code
+    links:
+      - nginx
+    expose:
+      - "4444"
+    shm_size: 2g
+    networks:
+      - private_nw
+      - public_nw
+  vnc:
```
jrbenny35: I think instead of having 2 services, we could just use the ```MOZ_HEADLESS``` env variable. If it is set then the tests would not be viewable through the vnc server.

tests/integration/test_overview.py
```diff
@@ -17,18 +17,6 @@ def test_overview_type_changes_correctly(base_url, selenium):
     assert exp_type in experiment.experiment_type
 
 
-@pytest.mark.nondestructive
-def test_overview_owner_changes_correctly(base_url, selenium):
```
jrbenny35: Instead of removing this can we just skip it? ```@pytest.mark.skip``` should do the trick.

jaredlockhart: Good idea!

tests/integration/conftest.py
jaredlockhart: I couldn't get it to work quickly so I filed it as a followup:

https://github.com/mozilla/experimenter/issues/2389

PR #2384, opened 2020-03-15
added Projects fixes #2298

### LINKED ISSUE #2298: Revenge of the Projects
So when Experimenter started, I had the idea that each experiment belonged to a long running project, examples would be like Activity Stream, Pocket, FXA, Search, Platform, Trailhead, Onboarding, etc etc.  I had some simple UI to assign each experiment to a project and then you could filter them on the list page.

That idea kind of lost steam and I ended up pulling all the relevant code out, but now interest in the idea has resurfaced so ironically we need to go and put it back in, but I can think of a simpler way to do it.

If we add a simple project model,  ie just a name/slug, and manage them purely from the admin, ie not allow users to create/modify them, then we can just add a multiselect to the delivery overview form for users to check off which project(s) the delivery belongs to.  I realized at some point that it makes more sense to allow deliveries to belong to potentially multiple projects.

Then we can add the filter back to the list page, which I think is the core requirement being requested.

tiftran:


app/experimenter/experiments/forms.py
```diff
@@ -168,6 +169,15 @@ class ExperimentOverviewForm(ChangeLogMixin, forms.ModelForm):
         help_text="Is this related to a previously run delivery?",
         queryset=Experiment.objects.all(),
     )
+    projects = forms.ModelMultipleChoiceField(
+        required=False,
+        label="Related Projects",
+        help_text=(
+            """Is this delivery related to a specific project?
+         Ask #ask_experimenter if you to add a new project"""
```
jaredlockhart: Missing word: `Ask #ask_experimenter if you *need to add a new project`

app/experimenter/experiments/tests/factories.py
```diff
@@ -333,3 +350,12 @@ class LocaleFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = Locale
         django_get_or_create = ("code",)
+
+
+class ProjectFactory(factory.django.DjangoModelFactory):
+    name = factory.LazyAttribute(lambda o: faker.word())
```
jaredlockhart: Use the same faker we use for experiment name so it looks more name like

app/experimenter/projects/models.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+
+class Project(models.Model):
+    name = models.CharField(max_length=255, unique=True, blank=False, null=False)
+    slug = models.SlugField(max_length=255, unique=True, blank=False, null=False)
```
jaredlockhart: I think `blank=False, null=False` are the defaults so it's redundant?  If I'm wrong just leave it.

app/experimenter/templates/experiments/section_overview.html
```diff
@@ -43,4 +43,13 @@
...
   {% endif %}
 
+  {% if experiment.projects.all %}
+    <p><strong> Projects</strong></p>
+    <p>
+      {% for project in experiment.projects.all %}
+      {{project}}{% if not forloop.last %},{% endif %}
+    {% endfor %}
+    </p>
+  {% endif %}
```
jaredlockhart: We might wanna make these links, but that's a small thing we can do as a follow up

app/experimenter/experiments/filtersets.py
```diff
@@ -279,6 +286,12 @@ def get_type_display_value(self):
             ]
         )
 
+    def get_project_display_value(self):
+        project_id = self.data.get("projects")
+
+        if project_id is not None:
+            return Project.objects.get(id=self.data.get("projects"))
```
jaredlockhart: Nit: you already looked up `project_id`

app/experimenter/experiments/serializers/entities.py
```diff
@@ -40,6 +41,13 @@ class Meta:
         fields = ("pref_name", "pref_type", "pref_branch", "pref_value")
 
 
+class ProjectSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = Project
+        fields = ("name",)
```
jaredlockhart: Should this be slug?

app/experimenter/projects/models.py
```diff
@@ -0,0 +1,17 @@
...
+    class Meta:
+        ordering = ("name",)
+        verbose_name = "Project"
+        verbose_name_plural = "Projects"
+
+    def __str__(self):
+        return self.name
+
+    def __repr__(self):
+        return self.name
```
jaredlockhart: Is this redundant?  Usually I only ever override __str__?

tiftran: gah leftover from tinkering

app/experimenter/experiments/filtersets.py
tiftran: yup yup you're right

app/experimenter/experiments/serializers/entities.py
tiftran: will do

PR #2405, opened 2020-03-19
allow stage pref rollouts to be updated fixes #2404

### LINKED ISSUE #2404: Unable to update RolloutPrefs

tiftran:


app/experimenter/experiments/serializers/design.py
```diff
@@ -323,7 +323,7 @@ def update(self, instance, validated_data):
             )
             for preference_data in preferences_data:
                 preference_data["experiment"] = instance
-                RolloutPreference.objects.create(**preference_data)
+                RolloutPreference(**preference_data).save()
```
jaredlockhart: I think we should try to stick to using the [queryset manager methods](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#update-or-create) where possible.  This technically works but it's kind of a hidden behaviour.

app/experimenter/experiments/serializers/design.py
```diff
@@ -321,14 +321,18 @@ def update(self, instance, validated_data):
...
+
+            submitted_preference_ids = []
             for preference_data in preferences_data:
                 preference_data["experiment"] = instance
-                RolloutPreference.objects.create(**preference_data)
+                pref_id = preference_data.pop("id", None)
+                if pref_id:
+                    submitted_preference_ids.append(pref_id)
+                RolloutPreference.objects.update_or_create(
+                    id=pref_id, defaults=preference_data
```
jaredlockhart: Oh okay good so this worked, we should try to use this pattern everywhere we can I think.

PR #2406, opened 2020-03-19
Fix rollout timeline/pop page fixes #2403

### LINKED ISSUE #2403: The “Timeline” section wrongly appears incomplete after all fields have been filled for Rollouts
**[Affected Versions]:**
- Firefox Nightly 76.0a1, Build ID 20200318213346
- Firefox Beta 75.0b5, Build ID 20200317211402
- Firefox Release 74.0, Build ID 20200309095159

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Perquisites]:**
- You are logged in with your LDAP account into the https://stage.experimenter.nonprod.dataops.mozgcp.net/ website.

**[Steps to reproduce]:**
1. Click the “Create Delivery” button.
2. Choose the “Staged Rollout” Type from “Overview”.
3. Fill in all of the “Required” fields with valid data for all the sections.
4. Click “Save Draft and Finish”.
5. Observe the “Timeline” section.

**[Expected results]:**
- The “Timeline” section is complete.

**[Actual results]:**
- The “Timeline” section displays the “You must complete the timeline form before you can launch your delivery.” message and the delivery can’t be set to “Confirm Ready to Ship”.

**[Notes]:**
- While editing the sections, the “2. Timeline & Population” and “3. Design” sections do not have a  checkmark displayed next to their descriptions.
- The issue is reproducible for both Pref and Addon rollouts.
- The issue is **not** reproducible with Pref or Addon experiments.
experimenter timeline

uhlissuh:
Screen Shot 2020-03-19 at 11 12 41 AM


app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -94,6 +96,49 @@ class TimelinePopForm extends React.PureComponent {
     }
   }
 
+  displayRolloutPlaybookOrEnrollment() {
```
jaredlockhart: I'd rather split this into two things like

    displayRolloutPlaybook
    displayEnrollmentDuration


and then inside each one just start with the bool check about whether it should show

uhlissuh: will do.

app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -94,6 +96,53 @@ class TimelinePopForm extends React.PureComponent {
...
+          error={this.state.errors.get("rollout_playbook", "")}
+          labelColumnWidth={2}
+          helpContent={ROLLOUT_PLAYBOOK_HELP}
+          helpIsExternalLink={true}
+        >
+          <option>Rollout Playbook</option>
+          <option>Low Risk Schedule</option>
+          <option>High Risk Schedule</option>
+          <option>Marketing Launch Schedule</option>
+          <option>Custom Schedule</option>
```
jaredlockhart: Oh shouldn't these be 

1) in the js constants file
2) those are the display strings but don't the options also need the string values (the lower case ones) here? https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L25-L35

uhlissuh: you'd like to see if more how I did the firefox versions?

uhlissuh: Because I just put the value into every option and that works, however I can refactor it to look more like firefox versions, where there's a function that format's each option withwhat's in the JS constants file.

jaredlockhart: Oooohhh I see what's happening here.

The fields in the database are charfields, so they'll accept anything, so anything you put in there is valid and will just get saved and displayed.

Usually in Django world you define two things, a db value and a display value, and then define the  pair as `CHOICES`, example:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L10-L20

Then later when you're dealing with the field in code somewhere, you're comparing  against the db value not the display value.

We didn't notice because a bunch of these things just only stay as text and we never refer to teh values in code, we just take wahtever it is and display it on  the detail page, and the db values and display values are the same, like for platform:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L152-L157

So yes exactly it should be like the way versions are done, and actually the real fix here is to change the serializer to use a ChoiceField instead of a CharField so it actually checks that the values are from the list we expect, but I think we can do that as a followup ticket so we don't blow this one up.

jaredlockhart: Followup: https://github.com/mozilla/experimenter/issues/2407

uhlissuh: ok I've made the change to where the playbook options are in constants and then they are displayed with the function `displayOptions`, which is a generalized version of a function that used to just be formatting the versions options... but is now formatting versions and playbook options.

uhlissuh:
Screen Shot 2020-03-19 at 12 05 39 PM

now you can see that high-risk is being saved correctly in DB and displaying the right graph.

PR #2409, opened 2020-03-19
Make timeline/pop and design serializers use choiceFields fixes #2407

### LINKED ISSUE #2407: Use ChoiceField instead of CharField in serializers
In our react serializers, for every field where we have a list of choices defined as constants, we should be using a [choice field](https://www.django-rest-framework.org/api-guide/fields/#choicefield) instead of a char field so the serializer validates that we only send in values we expect so that later when we're comparing values in code it's only the correct values.

uhlissuh:


app/experimenter/experiments/tests/serializers/test_timeline_population.py
```diff
@@ -189,14 +189,14 @@ def test_serializer_outputs_expected_schema_rollout(self):
     def test_serializer_saves_rollout_playbook_field(self):
         experiment = ExperimentFactory.create(
             type=ExperimentConstants.TYPE_ROLLOUT,
-            rollout_playbook="Low Risk Schedule",
+            rollout_playbook="low_risk",
```
jaredlockhart: Use the value constant here and in the other two:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L25

PR #2411, opened 2020-03-20
sign offs for exp after ship state fixes #2378

### LINKED ISSUE #2378: Add signoffs to factory experiments after Ready to Ship state
When we create a factory experiment that is in Ready to Ship or after, we should check off all the signoffs automatically.

tiftran:


app/experimenter/experiments/tests/factories.py
```diff
@@ -171,6 +171,18 @@ def create_with_status(cls, target_status, *args, **kwargs):
             old_status = status_value
             now += datetime.timedelta(days=random.randint(5, 20))
 
+        # set signoffs to true
+        if experiment.status in (
+            Experiment.STATUS_SHIP,
+            Experiment.STATUS_ACCEPTED,
+            Experiment.STATUS_LIVE,
+            Experiment.STATUS_COMPLETE,
+        ):
```
jaredlockhart: Can you move this into an experiments model method called `is_shipped`

tiftran: will do

PR #2442, opened 2020-03-25
Add results questions fixes #2283

### LINKED ISSUE #2283: Add outcome questions to results section
In service of monitoring the overall health/performance/quality of the delivery platform over time, I think we should track some basic delivery outcomes in experimenter for each delivery so that we can visualize them over time.  

Tentative proposal of outcome questions to add to results section (as a checklist) (subject to review)

- [ ] Did this delivery fail to launch at the expected time?
- [ ] Did this delivery require any restarts after it launched?
- [ ] Did this delivery fail to enroll the expected number of users?
- [ ] Did this delivery end before it was expected to?
- [ ] Did this delivery fail to generate usable data?
- [ ] Did this delivery complete successfully but fail to meet its intended outcome (edit: Did this delivery end prematurely)?

Edit: Maybe we could also have a notes field for any additional information

uhlissuh:
Screen Shot 2020-03-25 at 12 26 54 PM

results_qs


app/experimenter/templates/experiments/results_radio_field.html
```diff
@@ -0,0 +1,6 @@
+<p>
+  {{ label }}
+  <strong class="{% if field %}text-danger{% elif field is None %}text-muted{% endif %} ">
```
jaredlockhart: wicked

app/experimenter/experiments/models.py
```diff
@@ -614,7 +632,28 @@ def completed_objectives(self):
...
+            "results_changes_to_firefox",
+            "results_data_for_hypothesis",
+            "results_confidence",
+            "results_measure_impact",
+            "results_impact_notes",
+        )
+        for field in results_fields:
+            if getattr(self, field):
+                return True
+
```
tiftran: would something along the lines of this work and be cleaner? (idk if i'm really sure this would work)
```python
results_fields = (self. results_url, self.results_intial, <and the rest of the> ...)
return any(results_fields)
```

app/experimenter/templates/experiments/section_results.html
```diff
@@ -24,12 +31,37 @@
...
+  {% include "experiments/results_radio_field.html" with label=experiment.RESULTS_RECIPE_ERRORS_LABEL field=experiment.results_recipe_errors %}
+  {% include "experiments/results_radio_field.html" with label=experiment.RESULTS_RESTARTS_LABEL field=experiment.results_restarts %}
+  {% include "experiments/results_radio_field.html" with label=experiment.RESULTS_LOW_ENROLLMENT_LABEL field=experiment.results_low_enrollment %}
+  {% include "experiments/results_radio_field.html" with label=experiment.RESULTS_EARLY_END_LABEL field=experiment.results_early_end %}
+  {% include "experiments/results_radio_field.html" with label=experiment.RESULTS_NO_USABLE_DATA_LABEL field=experiment.results_no_usable_data %}
+
+  {% if experiment.results_failures_notes %}
+    <p><strong>Notes: </strong></p>
+    <p>{{ experiment.results_failures_notes|urlize|linebreaks }}</p>
+  {% endif %}
```
tiftran: Probs shameless plug, but markdown? 😄

app/experimenter/experiments/models.py
jaredlockhart: Yeah I think following the pattern of the risks we could just do something like

        any([getattr(field, self) for field in results_fields])

app/experimenter/experiments/forms.py
```diff
@@ -531,6 +531,16 @@ class Meta:
 
 
 class ExperimentResultsForm(ChangeLogMixin, forms.ModelForm):
+    RADIO_OPTIONS = ((False, "No"), (True, "Yes"))
+
+    def coerce_truthy(value):
+        if type(value) == bool:
+            return value
+        if value.lower() == "true":
```
uhlissuh: @jaredlockhart The initital two "no"s you were seeing were coming from the factory, where I had prepopulated two results questions fields. The reason I was doing that was because I was not reaching 100% test coverage due to the `coerce_truthy` method for the radio buttons. I wasn't able to hit line 537-538 here. In the risks tests, this line was hit because the factory prepopulates the risk questions as False. But I just realized that when I create the experiment with the factory in the test, if I set a value for a question to False when creating the experiment with the factory, the same thing will be achieved and line 537-538 will be covered. I'm not sure I have explained this well.

uhlissuh: So you shouldn't see the nos anymore, I took that out of the factory.

uhlissuh:
Screen Shot 2020-03-25 at 3 13 46 PM

app/experimenter/experiments/models.py
uhlissuh: thanks all.

PR #2453, opened 2020-03-29
added new results fields to changelog fixes #2446

### LINKED ISSUE #2446: Capture results survey in changelogs
When a user enters things into our new results survey we should also capture that in the changelog

tiftran:


app/experimenter/experiments/serializers/entities.py
```diff
@@ -49,6 +49,7 @@ class Meta:
 
 
 class ExperimentVariantSerializer(serializers.ModelSerializer):
+
```
jaredlockhart: nit: I generally go with a new line before the first method, but class properties can go directly under the class name line without a newline.  I just looked in the Google Python Style guide and didn't see something specifically about new line between class line and first property, but the rest of our classes follow that so I don't think it's worth making these ones different?

tiftran: gah will fix, I guess i over fixed the whitespacing that vs code removed

jaredlockhart:
Also small request, it helps me if you add screenshots for anything with visible changes so I can see how it looks without me needing to check out the PR locally and actually spin it up.

Issue #2323, opened 2020-02-27
Integration Test Audit 1 - Feb 2020


jrbenny35:
I'll be going through the current automation plan for the integration tests, as well as the CI configuration and a few other things. Please add any thoughts here.

Works on #2221

jrbenny35:
Docs are being added - #2317

jrbenny35:
Changes:

- The tests can be run in parallel on CI
- Headless Firefox option should only be set for CI.
-

jrbenny35:
@jaredlockhart Anything else here?

jaredlockhart:
- Can we move the tests dir into the experimenter dir?
- Should this use fill_overview fixture?
  - https://github.com/mozilla/experimenter/blob/master/tests/integration/test_create_experiment.py#L8-L17
  - https://github.com/mozilla/experimenter/blob/master/tests/integration/test_create_experiment.py#L30-L37

jrbenny35:
> All CSS locators should be constants? 
Some of these are chained off of other elements. Would you still want these to be added as a constant?

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_timeline_and_population.py#L71

jaredlockhart:
Yeah I don't see why not?

jaredlockhart:
Some pages define a page wait locator and don't use it:

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_detail.py#L13

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_objective_and_analysis.py#L18

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/home.py#L9

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/owned_experiments.py#L11

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/subscribed_experiments.py#L11

Some pages define a page wait locator and use the wrong locator:

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_design.py#L21-L27


Some pages don't define a page wait locator at all:

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_overview.py


Should be a class var:

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_timeline_and_population.py#L32

jrbenny35:
> Should be a class var:
> 
[...]

PR #2454, opened 2020-03-30
More integration test audit changes. Issue #2323

jrbenny35:
This should be the last PR. This has fixups for locators and page waits.


app/tests/integration/pages/experiment_design.py
```diff
@@ -18,7 +18,7 @@ class DesignPage(Base):
     _firefox_pref_type_locator = (By.CSS_SELECTOR, "#id_pref_type")
     _firefox_pref_branch_locator = (By.CSS_SELECTOR, "#id_pref_branch")
     _new_branch_locator = (By.CSS_SELECTOR, "#add-branch-button")
-    _page_wait_locator = (By.CSS_SELECTOR, ".page-edit-design")
+    _page_wait_locator = (By.CSS_SELECTOR, "body #design-form .form-label h5")
```
jaredlockhart: This is the kind of structural query I'd like to avoid, is there some element on the page you can add a class to and just look that up?

jaredlockhart: Example: I added this here for this same reason: https://github.com/mozilla/experimenter/blob/master/app/experimenter/static/js/components/TimelinePopForm.js#L204

jrbenny35: Gotcha, I'll do that

jrbenny35: So I did this because the usual page wait doesn't work. ```is_displayed()``` is checking for a CSS ```display``` property, which this page doesn't have.

jrbenny35: I don't think its a big deal if we can't use the same page wait. As long as selenium waits correctly. Unless you want to create a react element that changes when the page finishes loading.

jaredlockhart: The react component top level div has a class name:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/static/js/components/DesignForm.js#L192

That is not defined in the template:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/templates/experiments/edit_design.html

So that class name should only become present after react has finished loading, right?  That's what  I did on the timeline page?

Am I missing something?

PR #2493, opened 2020-04-02
added results to api fixes #2456

### LINKED ISSUE #2456: Expose outcomes in API
Now that we added these outcomes to the results section, let's expose them in the API so that we can visualize them in avocado.

tiftran:
Screen Shot 2020-04-02 at 2 17 18 PM


app/experimenter/experiments/serializers/entities.py
```diff
@@ -182,6 +205,7 @@ class ExperimentSerializer(serializers.ModelSerializer):
     countries = CountrySerializer(many=True)
     pref_type = PrefTypeField()
     changes = ExperimentChangeLogSerializer(many=True)
+    results = serializers.SerializerMethodField()
```
jaredlockhart, line 208: I wish there was a way to do this as `ResultsSerializer(source='self')` so you don't need the method

PR #2500, opened 2020-04-03
Add enrolled clients fields fixes #2359

### LINKED ISSUE #2359: Store estimated population size in number of profiles
We currently capture the definition of a delivery's target population by requesting a percentage of the total population, ie 25% of Firefox Beta.  However this same percentage can have wildly different outcomes in terms of the total number of profiles enrolled depending on the additional filters that are defined for the delivery.  This can make it hard to know whether the real enrollment numbers reflect the expected target population size while monitoring a live delivery.

Perhaps a simple solution to this is to additionally capture an estimate of the number of enrolled profiles as a reference point.  I don't think we need to add any fancy live integration with telemetry for this to update  it live while it's running, just put it somewhere on the detail page.

It's worth noting that there are significant differences between the 'active client ids per 24 hours' and  'cumulative enrolled clients' so we should take that into account and be clear about which we mean, my guess would be the latter.

uhlissuh:
Screen Shot 2020-04-03 at 10 20 16 AM
Screen Shot 2020-04-03 at 9 59 57 AM

Screen Shot 2020-04-03 at 10 19 57 AM


app/experimenter/templates/experiments/edit_objectives.html
```diff
@@ -12,6 +12,10 @@
   {% include "experiments/field_inline.html" with field=form.objectives marked_required=True %}
   {% include "experiments/field_inline.html" with field=form.analysis marked_required=True %}
 
+  {% if not object.is_rollout or object.is_generic_experiment %}
```
jaredlockhart, line 15: Put this on a model method like `should_have_total_enrolled` and check the types in there

app/experimenter/experiments/forms.py
```diff
@@ -289,6 +289,13 @@ def coerce_truthy(value):
         widget=forms.Textarea(attrs={"class": "form-control", "rows": 20}),
     )
 
+    total_enrolled_clients = forms.IntegerField(
+        required=False,
+        min_value=1,
+        label="Total Enrolled Clients",
```
jaredlockhart, line 295: Label it as `Estimated Total Enrolled Clients`

app/experimenter/templates/experiments/section_analysis.html
```diff
@@ -13,6 +13,12 @@
       </span>
     </strong>
   </p>
+  {% if experiment.total_enrolled_clients %}
+    <p>
+      <strong>Total Enrolled Clients:</strong> {{ experiment.total_enrolled_clients }}
```
jaredlockhart, line 18: Label it as `Estimated Total Enrolled Clients`

uhlissuh:
Screen Shot 2020-04-03 at 3 13 53 PM
Screen Shot 2020-04-03 at 3 13 28 PM
@jaredlockhart here's with adding Estimated to label.

PR #2504, opened 2020-04-04
Run pytest in parallel fixes #2001

### LINKED ISSUE #2001: Improve make check performance
I've been running `make check` a whole lot lately, and 3m is way too long.  I just messed around with a bunch of things (turns out running the db in non durability mode in a ramdisk was pointless, I guess SSDs really are fast), and I found out that django has a parallel test runner built in (how did I miss this!) using

`./manage.py test --parallel`

between that and running all the lint tasks in parallel in bash jobs, I got the whole thing down to 30s.  

Also a nice little treat is adding a `.dockerignore` to get the build context down to <20m.  I saw it was a few hundred mb on tif's machine.

I also wanna see if I can parallelize any of the build steps, like `pip install` to get build time down.

More fasterer is more betterer.

jaredlockhart:
I got bored and gave this another shot.  Got it working! With parallel coverage and everything!

On my machine,  2018 15" MBP 6(12) Cores/32gb ram:

Serial test suite: 2m32s
Parallel test suite @ 12 proc: 100s
Parallel test suite @ 4 proc: 75s

So I set it to 4.


jaredlockhart:
Make some optimizations to the factories, moved countries into a fixture, and now I've  got:

`make refresh` 50s -> 20s
`make check` 2m30s  -> 60s

jaredlockhart:
Okay so the major changes I  made here

- Run all checks in parallel using `GNU Parallel`
- Run  pytest in parallel using `pytest-xdist`
- Profile the tests and make the following performance optimizations
  - Reduce unnecessary database saves
    - Save experiment once in `create_with_status`
    - Cut down on locales/countries in experiment factory
  - Reduce the number of random numbers used
  - Refactor the radio widgets to use `"0" "1"` and `'int' instead of `"False" "True"` and `coerce_truthyness` because it was calling `.lower()` a lot and apparently  that was taking a lot of time

PR #2522, opened 2020-04-07
Make platform a multiselect fixes #1210

### LINKED ISSUE #1210: Fix "Platforms" support to allow multi-select (same as Locales and Countries) and add a few more specific options
**[Notes]:**
- I filled this issue because in the past we had experiments that were requested to be tested only on two different platforms (eg: Mac and Windows) or other combinations of only two platforms.

**[Affected versions]:**
- Firefox Release 66.0.3 and above
- Chrome 74.0. and up

**[Affected Platforms]:** 	 	
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have an experiment in the "Ready for Sign-Off state created on the "https://stage.experimenter.nonprod.dataops.mozgcp.net/" website.
- You are logged in your LDAP account on stage website.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites and navigate to "https://stage.experimenter.nonprod.dataops.mozgcp.net/".
2. Open the experiment from prerequisites.
3. Click the "Edit" button from the "Population" section.
4. Click the "Platform" input field and select the "All Windows" option.
5. Click again on the "Platform" input field and select the "All Mac" option.
6. Observe the "Platform" input field.
 
**[Expected result]:**
- Both options are displayed in the "Platform" input field. 

**[Actual result]:**
- Only the "All Mac" option is displayed in the "Platform" input field.

**[Additional Notes]:**
- This issue is reproducible regardless of which combined selection of two is chosen, out of the three available platforms ("All Windows", "All Mac", "All Linux"). 
One on a time

uhlissuh:
Screen Shot 2020-04-06 at 5 10 07 PM
Screen Shot 2020-04-06 at 5 09 50 PM
Screen Shot 2020-04-06 at 5 09 28 PM

I removed the platform integration tests for the same reason as when I was working on the countries and locales multiselect.. it's very hard to grab the elements for the integration tests because they have weird classes generated by the component.. I couldn't figure out how to do it. In that case I deleted the integration tests that were trying to cover countries and locales and @jaredlockhart was OK with it, so I did it again here with platforms.


app/experimenter/experiments/migrations/0090_platforms_to_array.py
```diff
@@ -0,0 +1,46 @@
...
+import django.contrib.postgres.fields
+from django.db import migrations, models
+import experimenter.experiments.models
+
+
+def forward_platform_string_to_array(apps, schema_editor):
+    Experiment = apps.get_model("experiments", "Experiment")
+
+    for experiment in Experiment.objects.all():
+        experiment.platforms = [experiment.platform]
```
jaredlockhart, line 12: What about the `All Platforms` case?  Shouldn't this add the list of all three?

app/experimenter/experiments/models.py
```diff
@@ -23,6 +23,14 @@
 from experimenter.experiments.constants import ExperimentConstants
 
 
+def default_all_platforms():
+    return [
+        ExperimentConstants.PLATFORM_MAC,
+        ExperimentConstants.PLATFORM_WINDOWS,
+        ExperimentConstants.PLATFORM_LINUX,
+    ]
```
jaredlockhart, line 31: Should this go in constants?

app/experimenter/experiments/models.py
```diff
@@ -147,11 +155,11 @@ class Experiment(ExperimentConstants, models.Model):
...
     projects = models.ManyToManyField(Project, blank=True)
-    platform = models.CharField(
-        max_length=255,
-        choices=ExperimentConstants.PLATFORM_CHOICES,
-        default=ExperimentConstants.PLATFORM_ALL,
+    platforms = ArrayField(
+        models.CharField(max_length=200),
+        blank=True,
         null=True,
+        default=default_all_platforms,
```
jaredlockhart, line 162: Does this have to be a callable?  You can't just put the array literal?

app/experimenter/experiments/models.py
```diff
@@ -574,6 +582,17 @@ def should_have_population_percent(self):
...
         return self.type not in (self.TYPE_GENERIC, self.TYPE_ROLLOUT)
 
+    @property
+    def display_platforms(self):
+        if {
+            ExperimentConstants.PLATFORM_MAC,
+            ExperimentConstants.PLATFORM_LINUX,
+            ExperimentConstants.PLATFORM_WINDOWS,
+        } == set(self.platforms):
+            return "All"
```
jaredlockhart, line 592: nit: `All Platforms`

app/experimenter/experiments/serializers/recipe.py
```diff
@@ -67,6 +67,21 @@ def get_channels(self, obj):
         return [obj.firefox_channel.lower()]
 
 
+class FilterObjectPlatformSerializer(serializers.ModelSerializer):
+    type = serializers.SerializerMethodField()
+    platforms = serializers.SerializerMethodField()
```
jaredlockhart, line 72: I think you can omit this field definition and the method altogether?  If you're just returning the  field verbatim won't it just serialize it automatically?

app/experimenter/experiments/serializers/recipe.py
```diff
@@ -336,6 +351,9 @@ def get_filter_object(self, obj):
         if obj.countries.count():
             filter_objects.append(FilterObjectCountrySerializer(obj).data)
 
+        if len(obj.platforms) <= 2:
```
jaredlockhart, line 354: If you move the list of all platforms on to constants, then you can do `len(PLATFORMS)`  here instead of hardcoding `2`.  Likelihood of adding platforms is small, but still if we ever change that list then this will  update accordingly.

app/tests/integration/pages/experiment_timeline_and_population.py
```diff
@@ -121,19 +119,3 @@ def firefox_max_version(self, version=None):
...
-            if item.get_attribute("selected"):
-                return item.get_attribute("value")
-
-    @platform.setter
-    def platform(self, platform=None):
-        element = self.find_element(*self._platform_locator)
-        platforms = element.find_elements(*self._platform_option_locator)
-        for item in platforms:
-            if item.get_attribute("value") == platform:
-                item.click()
```
jaredlockhart, line 139: Yeah I see your point about having to deal with the weird markup from the select box.  I guess the sanest thing to do is take these out, since they don't apply anymore, and @jrbenny35 can write new ones for the new select input as he adds the new end to end tests.

app/tests/integration/test_timeline_and_population.py
```diff
@@ -78,13 +78,3 @@ def test_firefox_max_version_updates_correctly(selenium, base_url, fill_overview
...
-
-
-@pytest.mark.nondestructive
-def test_platform_selection_updates_correctly(selenium, base_url, fill_overview):
-    """Test platform selection updates."""
-    timeline_pop_form = fill_overview.save_and_continue_btn()
-    assert timeline_pop_form.platform == "All Platforms"
-    channel = "All Linux"
-    timeline_pop_form.platform = channel
-    assert timeline_pop_form.platform == channel
```
jaredlockhart, line 90: @jrbenny35 showed me a better way to do this, just mark it as

`@pytest.mark.skip`

like so:

https://github.com/mozilla/experimenter/blob/f563df6936c06c1cf68096a57905e1b6d2e5a0f7/app/tests/integration/test_overview.py#L30

app/experimenter/experiments/models.py
uhlissuh: Yes this has to be a callable. Django constantly warns you about this if you don't make it a callable. Also in the docs, there is this: "If you give the field a default, ensure it’s a callable such as list (for an empty default) or a callable that returns a list (such as a function). Incorrectly using default=[] creates a mutable default that is shared between all instances of ArrayField."

app/tests/integration/test_timeline_and_population.py
uhlissuh: ok so you want me to leave the test but just skip it?

app/experimenter/experiments/models.py
jaredlockhart: Oh right right, becuase python mutability, yep got it.  Okay then yeah leave it as a callable but just move the list of the defaults onto constantst and then you can just write a lambda that returns the constant?

app/tests/integration/test_timeline_and_population.py
jaredlockhart: Yep exactly and then we can at least refer to it later for reference.

app/tests/integration/pages/experiment_timeline_and_population.py
jrbenny35: I _could_ write new ones, but the agreement was, if you broke the tests you fix them. I'd be happy to help fix these :)

jaredlockhart: Okay then let's leave these intact, they won't get hit if the test is skipped, and @jrbenny35 can refactor them to support the new multiselect, and that way we can reuse that logic for any future tests that need to interact with them.

app/experimenter/experiments/models.py
```diff
@@ -574,6 +582,17 @@ def should_have_population_percent(self):
     def should_have_total_enrolled(self):
         return self.type not in (self.TYPE_GENERIC, self.TYPE_ROLLOUT)
 
+    @property
+    def display_platforms(self):
+        if {
+            ExperimentConstants.PLATFORM_MAC,
+            ExperimentConstants.PLATFORM_LINUX,
+            ExperimentConstants.PLATFORM_WINDOWS,
```
jaredlockhart, line 590: Also if you move the list to a constant then you can just refer to the constant here too

app/tests/integration/pages/experiment_timeline_and_population.py
jaredlockhart: Also sorry @jrbenny35 what I meant was when we have the new end to end integration tests it will cover the case of tests that only check specific fields, not that you'd rewrite these same tests again.

uhlissuh: This logic is more fine-grained than needs to be tested via integration tests, in my opinion. I have front end tests covering this component.

app/experimenter/experiments/serializers/recipe.py
uhlissuh: yes, woops.

app/experimenter/experiments/models.py
uhlissuh: Yes, I tried that actually, however if you try to do something in the model field like:

`default=lambda: ExperimentConstants.PLATFORMS_LIST` 

you get another error when trying to generate the migration about not being able to serialize the lambda function. I believe the way I have done it is the only way to make it work.

app/experimenter/experiments/models.py
uhlissuh: I suppose so?

app/experimenter/experiments/migrations/0090_platforms_to_array.py
uhlissuh: Yes, oops, this of course needs to be changed to address the "All Platforms" case.

PR #2537, opened 2020-04-09
Add support for NDT export button injection

rehandalal:
This just adds a div that NDT can use to inject an export button into. It also provides easy access to the slug without needing to pull it from `window.location`.

Resulting injected button:
Screenshot from 2020-04-09 01-17-27

r?


jaredlockhart:
Also is there a timing of when this should land depending on when NDT is ready to handle taking over for DC?

rehandalal:
> But what about the existing Send to Normandy button? Shouldn't that be removed?

Yes, probably. For now NDT hides it if you have NDT installed but we should keep the button there till we're ready to replace DC with NDT (the linked PR above outlines what NDT is doing in case you're curious)

> Also is there a timing of when this should land depending on when NDT is ready to handle taking over for DC?

I think this can land right away since we're not getting rid of the existing button

PR #2542, opened 2020-04-12
Better support for NDT injection #2561

### LINKED ISSUE #2561: Add support for NDT button/link injection
We need to be able to inject an export button and recipe details page links from NDT to experimenters UI. 

Experimenter should have hidden UI that NDT can un-hide and inject appropriate hrefs into.

rehandalal:
Fixes #2561 

After some reflection, I think this is a more scalable and sensible way to handle injection.

NDT doesn't need to be opinionated about the elements being injected --- so we don't need to understand Experimenter's markup/styling. 

The visual presentation of buttons/links remains an Experimenter concern. NDT simply handles showing/hiding elements as necessary and injecting the appropriate href.

r?


rehandalal:
A little issue I realized we're not dealing with is how to actually get the Normandy Recipe ID back into Experimenter... I think this is worth having a short conversation about.

jaredlockhart:
@rehandalal I can think of a few options.

If it's an API call, then NDT would have to have an authorization header for Experimenter.

But we could put a hidden form on the page,  basically the same as what's in the modal now, and NDT could fill it in and submit it?

jaredlockhart:
Oh also @rehandalal in general we always file an issue and reference its number from the commit.  Thinking about it more though I realize the big point of that was to associate the issue with the milestone as a changelog, and now we don't have milestones, so maybe it's superfluous now?

PR #2577, opened 2020-04-21
Add windows versions multiselect fixes #2265

### LINKED ISSUE #2265: Add OS version input
We need to capture the OS Version if someone chooses to filter by OS, so we should add an optional text field that appears if someone selects anything other than `All Platforms`

uhlissuh:
Screen Shot 2020-04-21 at 5 26 05 PM

Screen Shot 2020-04-21 at 5 27 50 PM

versions_demo_2


app/experimenter/experiments/serializers/timeline_population.py
```diff
@@ -25,14 +25,22 @@ class Meta:
...
         return data["value"]
 
 
+class PlatformsMultiSelectSerializer(serializers.Serializer):
+    def to_representation(self, data):
+        return {"value": data, "label": data[4:]}
+
+    def to_internal_value(self, data):
+        return data["value"]
+
```
uhlissuh, line 42: Has to have it's own serializer because of removing "All " from the label.

app/experimenter/static/js/components/TimelinePopForm.js
```diff
@@ -152,6 +159,11 @@ class TimelinePopForm extends React.PureComponent {
     return choicesJSX;
   }
 
+  isWindowsVersionEnabled() {
+    const platforms = this.state.data.get("platforms").toJS();
+    return platforms.length === 1 && platforms[0].value === "All Windows";
```
jaredlockhart, line 164: Yep this looks good I would just move the actual `"All Windows"` into a string constant in `constants.js`

app/experimenter/static/js/components/constants.js
```diff
@@ -241,9 +241,16 @@ export const PLAYBOOK_CHOICES = [
...
+  { value: "All Mac", label: "Mac" },
+  { value: "All Linux", label: "Linux" },
+  { value: "All Windows", label: "Windows" },
+];
+
+export const WINDOWS_VERSIONS_CHOICES = [
+  { value: "Windows 7", label: "Windows 7" },
+  { value: "Windows 8", label: "Windows 8" },
+  { value: "Windows 8.1", label: "Windows 8.1" },
+  { value: "Windows 10+", label: "Windows 10+" },
```
jaredlockhart, line 253: I think it'd be better to move all the actual strings into individual constants and then refer to them in these arrays/objs, and then if we need to refer to an individual one (like above) you also have the individual value.  I think I found a way to serialize the `constants.py` into a json object and load it into the frontend so I think we'll be able to get rid of this file altogether soon.

app/experimenter/static/js/tests/DataFactory.js
```diff
@@ -220,8 +220,8 @@ export class TimelinePopDataFactory extends Factory {
       locales: [{ value: "NP", label: "Nepali" }],
       countries: [{ value: "US", label: "United States" }],
       platforms: [
-        { value: "All Windows", label: "All Windows" },
-        { value: "All Mac", label: "All Mac" },
+        { value: "All Windows", label: "Windows" },
+        { value: "All Mac", label: "Mac" },
```
jaredlockhart, line 224: Here we can refer to constants

app/experimenter/static/js/tests/TimelinePopForm.test.js
```diff
@@ -107,7 +119,7 @@ describe("The TimelinePopForm component for experiments", () => {
     expect(data.firefox_min_version).toBe("60.0");
     expect(data.firefox_max_version).toBe("64.0");
     expect(data.platforms).toEqual([
-      { value: "All Windows", label: "All Windows" },
+      { value: "All Windows", label: "Windows" },
```
jaredlockhart, line 122: +constants

PR #2578, opened 2020-04-21
Remove platform filter object from recipe serializer fixes #2568

### LINKED ISSUE #2568: DC doesn't support platform filter objects
Will need to revert to platforms back to comments.

Screen Shot 2020-04-20 at 10 31 41 PM

uhlissuh:
I commented out the filter object. If I leave the platforms FO, but dont use it, we won't pass our test coverage. I was always testing the platforms fo as part of the whole recipe output. 

I left the comment in the form of a list of strings. i didn't think it was worth putting effort into formatting a comment to look slightly nicer when we're about to throw this comment code away in a few weeks anyway.

### DIFF
--- a/app/experimenter/experiments/serializers/recipe.py
+++ b/app/experimenter/experiments/serializers/recipe.py
@@ -67,17 +67,6 @@ class FilterObjectChannelSerializer(serializers.ModelSerializer):
         return [obj.firefox_channel.lower()]
 
 
-# class FilterObjectPlatformSerializer(serializers.ModelSerializer):
-#     type = serializers.SerializerMethodField()
-#
-#     class Meta:
-#         model = Experiment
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/serializers/recipe.py
```diff
@@ -68,15 +67,15 @@ def get_channels(self, obj):
...
-        return "platform"
+# class FilterObjectPlatformSerializer(serializers.ModelSerializer):
+#     type = serializers.SerializerMethodField()
+#
+#     class Meta:
+#         model = Experiment
+#         fields = ("type", "platforms")
+#
+#     def get_type(self, obj):
+#         return "platform"
```
jaredlockhart, line 78: Just delete this, and when we wanna add it back we can just revert this commit.

PR #2584, opened 2020-04-23
Add fixtures to fill individual form pages

jrbenny35:
* Refactor tests to use new fixtures
* Refactor page objects to accept an experiments url
* Added logic so that fixtures can be used in any test


jrbenny35:
This looks like a lot but I moved everything from the fixture in ```test_multi_pref_e2e``` to individual fixtures in conftest. I also added an option to pass in a url to the page objects, so they can be loaded after ```fill_overview``` and they load the right experiment within the browser.

Also some of the tests relied on ```fill_overview``` returning the next page. I have changed that, so each test for the specific page should now go right to that page via the page object.

PR #2589, opened 2020-04-27
Add profile age intake fixes #2581

### LINKED ISSUE #2581: Filter on profile creation date

uhlissuh:
Screen Shot 2020-04-27 at 4 27 07 PM
Screen Shot 2020-04-27 at 4 26 05 PM


app/experimenter/experiments/serializers/recipe.py
```diff
@@ -355,5 +355,7 @@ def get_arguments(self, obj):
     def get_comment(self, obj):
         return (
             f"Platform: {obj.platforms}\n"
-            f"Windows Versions: {obj.windows_versions}\n{obj.client_matching}"
+            f"Windows Versions: {obj.windows_versions}\n"
+            f"Profile Age: {obj.profile_age}\n"
```
jaredlockhart, line 359: We should only add `profile_age` here if it's not `PROFILES_ALL`

app/experimenter/static/js/components/RadioButtonInlineLabel.js
```diff
@@ -0,0 +1,120 @@
+import { boundClass } from "autobind-decorator";
+import PropTypes from "prop-types";
+import React from "react";
+import { Form, Row, Col } from "react-bootstrap";
+
+import Error from "experimenter/components/Error";
+import HelpBox from "experimenter/components/HelpBox";
+
+@boundClass
+class RadioButtonInlineLabel extends React.PureComponent {
```
jaredlockhart, line 10: Oh good actually I was thinking all our radios should use something like this, maybe we can remove the other radio and just use this for everything.

PR #2623, opened 2020-05-04
automate saving recipe id from normandy fixes #2562

### LINKED ISSUE #2562: Read normandy recipe IDs for shipped deliveries
Rather than the current workflow of manually entering Recipe IDs in the modal, let's update the background worker to look in the Normandy  API  for Experimenter slugs that don't have recipe IDs in our  database and save them and update  the  state to accepted, so then the workflow would  be

1) Click `Send to Normandy` in Experimenter
2) Fill out and save in Delivery Console
3) - Experimenter automatically detects that a `Ready to Ship` delivery that does not have a Normandy ID now appears in the Normandy V3 Recipe API and saves that recipe ID to the database.  
    -  If it finds a second  recipe with the same slug, it adds that recipe ID to the 'other recipe ids' field.  
    - Experimenter updates the status to `Accepted by Normandy`

tiftran:


app/experimenter/experiments/tasks.py
```diff
@@ -123,8 +125,61 @@ def update_experiment_bug_task(user_id, experiment_id):
 @app.task
 @metrics.timer_decorator("update_experiment_info.timing")
 def update_experiment_info():
-    metrics.incr("update_experiment_info.started")
-    logger.info("Updating experiment info")
+    update_ready_to_ship_experiments()
+    update_launched_experiments()
```
jaredlockhart, line 129: Don't you have to call `.delay()` here to actually dispatch it as a separate task?

app/experimenter/experiments/tasks.py
```diff
@@ -123,8 +125,61 @@ def update_experiment_bug_task(user_id, experiment_id):
...
+                    email=normandy_user, username=normandy_user
+                )
+
+                generate_change_log(
+                    old_serialized_exp,
+                    new_serialized_exp,
+                    experiment,
+                    changed_data,
+                    default_user,
+                )
```
jaredlockhart, line 171: Can we move all this into a model method?  I feel like we have to do this pattern a whole bunch and it's a bit verbose to have to do all these steps here explicitly.

app/experimenter/normandy/client.py
```diff
@@ -44,6 +44,12 @@ def get_recipe(recipe_id):
     return recipe_data["approved_revision"]
 
 
+def get_recipe_list(params={}):
```
jaredlockhart, line 47: We know this will allways only be called to get the recipes for an experiment slug, so rather than exposing the whole params dict, this function could just take in a slug and build the dict internally.  If you expose the whole raw params dict to callers outside of the normandy module, then they have to know something about how the normandy API works, could inject parameters wrong, etc

app/experimenter/templates/experiments/detail_ship.html
```diff
@@ -39,74 +39,6 @@
...
-
-            <div class="form-group">
-              <button type="submit" class="btn btn-success">Save Recipe ID</button>
-            </div>
-          </form>
-        </div>
-
-        <div class="modal-footer">
-          <div class="flex-fill">
-            <a href="{{ experiment.api_recipe_url }}" target="_blank">Preview Recipe JSON</a>
```
jaredlockhart, line 100: Oh actually, can you also add this link in the links section (where the bugzilla links are etc) to the experiment recipe json that only shows up for [SHIP, ACCEPTED, LIVE, COMPLETE] because we'll still need to be able to get to that JSON for other stuff.

I've been using a pattern lately of having a method on the model whenever stuff needs to show up selectively like

```py
def should_have_recipe_link(self):
    return whatever
```

and calling that from the template

jaredlockhart: You could also try writing a parameterized test if you're feeling brave ⚔️

app/experimenter/experiments/tasks.py
tiftran: i moved this into changelog utils instead, since moving it to models.py creates a circular import problem

app/experimenter/templates/experiments/detail_ship.html
tiftran: I lucked out and the method already exists 😁

app/experimenter/experiments/changelog_utils.py
```diff
@@ -1,4 +1,24 @@
-from experimenter.experiments.models import ExperimentChangeLog
+from django.contrib.auth import get_user_model
+from django.conf import settings
+
+from experimenter.experiments.models import Experiment, ExperimentChangeLog
+from experimenter.experiments.serializers.entities import ChangeLogSerializer
+
+
+def update_experiment_with_change_log(experiment, changed_data):
```
jaredlockhart, line 8: Ah I would call this `old_experiment` rather than reuse and reassign to `experiment`, could introduce wonkiness in the future to reuse variable names

app/experimenter/experiments/changelog_utils.py
```diff
@@ -1,4 +1,24 @@
...
+
+    old_serialized_exp = ChangeLogSerializer(experiment).data
+    Experiment.objects.filter(id=experiment.id).update(**changed_data)
+    experiment = Experiment.objects.get(slug=experiment.slug)
```
jaredlockhart, line 12: And `new_experiment` here

tiftran:
Oooh would there be a case where an experiment will have many recipes, one recipe go live and then have other recipes created afterwards?

jaredlockhart:
Well it's not just go live, even if they're inactive recipes, if it goes

1) Export from Experimenter
2) Save to Normandy
3) Experimenter task kicks in, finds one recipe with experimenter_slug, saves it, moves to Accepted by Normandy
4) Create a second recipe in Normandy immediately afterwards

None of them were published as active, but the job kicked in right between creating the first and second recipes, experimenter wouldn't pick up the second one with the way this is written now, right?

tiftran:
I totally agree that I need to change my implementation for the edge case you mentioned. My question is that should we still be looking for recipe IDs after the primary recipe goes live and therefore the experiment status is live? Or should I just be looking at experiments that have status shipped and accepted by normandy?

jaredlockhart:
Oh I see what you mean, I think it's only the first two, ready to ship and accepted by normandy.  If the third case comes up we can add it but I think we need the first two cases out of the box.

tiftran:
Okie doke will do

tiftran:
yup yup

PR #2625, opened 2020-05-04
add env var for worker refresh rate fixes #1739

### LINKED ISSUE #1739: Add env var for worker refresh rate
We have the workers hardcoded to update from normandy every 5 minutes (300s) but now that we have the capability to test the entire workflow locally it's annoying to wait 5 minutes, it should be an env var so we can set it to something like 30s if we want to.

tiftran:
I don't think this would require changes to PROD's env vars right?


.env.sample
```diff
@@ -1,6 +1,7 @@
 BUGZILLA_API_KEY=
 BUGZILLA_CC_LIST=
 BUGZILLA_HOST=https://bugzilla.allizom.org/
+CELERY_BEAT_SCHEDULE=
```
jaredlockhart, line 4: Should be `CELERY_SCHEDULE_INTERVAL`?

PR #2647, opened 2020-05-11
Add multipref e2e test.

### LINKED ISSUE #2224: End to End Test: Pref Experiment (Multipref)

jrbenny35:
Closes #2224

Adds multipref branch e2e tests.

### DIFF
--- a/app/tests/integration/e2e_test_variables.json
+++ b/app/tests/integration/e2e_test_variables.json
@@ -48,8 +48,8 @@
                     {
                         "firefox_pref_branch": "default",
                         "firefox_pref_name": "e2e-testing-branch-1-1",
-                        "firefox_pref_type": "boolean",
-                        "firefox_pref_value": "false"
+                        "firefox_pref_type": "integer",
+                        "firefox_pref_value": "10"
                     }
                 ]
             },
@@ -60,14 +60,14 @@
                     {
                         "firefox_pref_branch": "default",
                         "firefox_pref_name": "e2e-testing-branch-2",
-                        "firefox_pref_type": "boolean",
-                        "firefox_pref_value": "false"
+                        "firefox_pref_type": "string",
+                        "firefox_pref_value": "pref-string"
                     },
                     {
                         "firefox_pref_branch": "default",
                         "firefox_pref_name": "e2e-testing-branch-2-1",
-                        "firefox_pref_type": "boolean",
-                        "firefox_pref_value": "true"
+                        "firefox_pref_type": "string",
+                        "firefox_pref_value": "{\"object\": \"name\"}"
                     }
                 ]
             }
--- a/app/tests/integration/e2e_test_variables.json
+++ b/app/tests/integration/e2e_test_variables.json
@@ -24,5 +24,33 @@
                 "firefox_pref_branch": "default"
             }
         ]
+    },
+    "multi-pref-experiment": {
+        "action_name": "multi-preference-experiment",
+        "name": "E2E experimenter multipref-test",
+        "type": "channel",
+        "channels": "Nightly",
+        "min_version": "80.0",
+        "max_version": "100.0",
+        "userFacingName": "e2e testing name",
+        "userFacingDescription": "e2e testing description",
+        "branches": [
+            {
+                "branch_description": "multipref-branch-description",
+                "branch_name": "e2e-multiepref-branch",
+                "firefox_pref_branch": "default",
+                "firefox_pref_name": "e2e-testing",
+                "firefox_pref_type": "boolean",
+                "firefox_pref_value": "true"
+            },
+            {
+                "branch_description": "multipref-branch-description-2",
+                "branch_name": "e2e-multiepref-branch-2",
+                "firefox_pref_branch": "default",
+                "firefox_pref_name": "e2e-testing",
+                "firefox_pref_type": "boolean",
+                "firefox_pref_value": "false"
+            }
+        ]
     }
 }
--- a/app/tests/integration/e2e_test_variables.json
+++ b/app/tests/integration/e2e_test_variables.json
@@ -66,7 +66,7 @@
                     {
                         "firefox_pref_branch": "default",
                         "firefox_pref_name": "e2e-testing-branch-2-1",
-                        "firefox_pref_type": "string",
+                        "firefox_pref_type": "json string",
                         "firefox_pref_value": "{\"object\": \"name\"}"
                     }
                 ]
--- a/app/tests/integration/e2e_test_variables.json
+++ b/app/tests/integration/e2e_test_variables.json
@@ -37,19 +37,39 @@
         "branches": [
             {
                 "branch_description": "multipref-branch-description",
-                "branch_name": "e2e-multiepref-branch",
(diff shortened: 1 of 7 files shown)


jrbenny35:
I'll edit the commit message before merging

app/tests/integration/e2e_test_variables.json
```diff
@@ -24,5 +24,33 @@
...
+                "firefox_pref_type": "boolean",
+                "firefox_pref_value": "true"
+            },
+            {
+                "branch_description": "multipref-branch-description-2",
+                "branch_name": "e2e-multiepref-branch-2",
+                "firefox_pref_branch": "default",
+                "firefox_pref_name": "e2e-testing",
+                "firefox_pref_type": "boolean",
+                "firefox_pref_value": "false"
```
jaredlockhart, line 52: Can you put 2-3 prefs in each branch?  I'd like to stress test that they all get in.

jrbenny35: We could but that may be out of scope. Maybe a multi-pref multipref e2e test?

jrbenny35: I could add that right after this

jaredlockhart: I don't understand what you mean.  How is testing multiple prefs per branch out of scope for the multipref e2e test?  Isn't that exactly what this should be testing?

app/tests/integration/e2e_test_variables.json
```diff
@@ -24,5 +24,33 @@
...
+        "type": "channel",
+        "channels": "Nightly",
+        "min_version": "80.0",
+        "max_version": "100.0",
+        "userFacingName": "e2e testing name",
+        "userFacingDescription": "e2e testing description",
+        "branches": [
+            {
+                "branch_description": "multipref-branch-description",
+                "branch_name": "e2e-multiepref-branch",
```
jaredlockhart, line 40: Spelling: `e2e-multipref-branch`

app/tests/integration/e2e_test_variables.json
```diff
@@ -24,5 +24,33 @@
...
+                "firefox_pref_branch": "default",
+                "firefox_pref_name": "e2e-testing",
+                "firefox_pref_type": "boolean",
+                "firefox_pref_value": "true"
+            },
+            {
+                "branch_description": "multipref-branch-description-2",
+                "branch_name": "e2e-multiepref-branch-2",
```
jaredlockhart, line 48: Spelling: `e2e-multipref-branch`

app/tests/integration/e2e_test_variables.json
jrbenny35: Right now we are, but adding multiple prefs would be testing something else as well. This is the minimum case for a multipref experiment. Adding more prefs should be its own test imo

jaredlockhart: I disagree, of the three cases you're describing:

1) Same pref for all branches (already tested)
2) Different prefs per branch with 1 pref per branch
3) Different prefs per branch with multiple prefs per branch


1 and 3 are the important cases, 2 is the edge case, so I think 1 and 3 are the tests we should have now before we start considering new test cases.  Although I didn't provide details in the bug, the specific thing I was looking for in this test is case 3.

jrbenny35: But 2 leads to 3. So why not start with 2? I was trying to create minimum test cases for each of these e2e tests, and then expand them to be specific later. But having a minimum test case tested is where we should start.

jrbenny35: Is 1 pref per branch not something that usually happens?

jaredlockhart: Not with 1 unique pref per branch, no.  Either it's 1 pref, but it's the same pref for both branches (the single pref case), or it's multiple prefs and there's multiple copies of the same pref in each branch (case 3).  Case 2 doesn't generally happen.

And I get your logic about progressively increasing cases, but each test has a cost, in terms of development time, and in terms of run time, so I think it's prudent to be conservative with adding cases and right now we have a concrete list of the core cases I think are the minimum surface to ensure everything's behaving correctly, and those are cases 1 and 3.

jrbenny35: Alrighty, I will add 2 prefs ber branch. Is that more in line with what would work best?

jaredlockhart: @jrbenny35 Yes exactly, thank you!

app/tests/integration/e2e_test_variables.json
```diff
@@ -24,5 +24,53 @@
...
+                        "firefox_pref_name": "e2e-testing-branch-2",
+                        "firefox_pref_type": "boolean",
+                        "firefox_pref_value": "false"
+                    },
+                    {
+                        "firefox_pref_branch": "default",
+                        "firefox_pref_name": "e2e-testing-branch-2-1",
+                        "firefox_pref_type": "boolean",
+                        "firefox_pref_value": "true"
+                    }
```
jaredlockhart, line 71: Thank you for adding multiple preferences!  One last thing, can you make them each one of each of the four types we support:

1) boolean
2) int
3) str
4) json str

Just so we can hit all four cases in one test?

app/tests/integration/e2e_test_variables.json
```diff
@@ -24,5 +24,53 @@
...
+                    {
+                        "firefox_pref_branch": "default",
+                        "firefox_pref_name": "e2e-testing-branch-2",
+                        "firefox_pref_type": "string",
+                        "firefox_pref_value": "pref-string"
+                    },
+                    {
+                        "firefox_pref_branch": "default",
+                        "firefox_pref_name": "e2e-testing-branch-2-1",
+                        "firefox_pref_type": "string",
```
jaredlockhart, line 69: This should be: `"firefox_pref_type": "json string",`

Defined here:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/experiments/constants.py#L226

jrbenny35: This is what I mentioned in Slack. Setting the test variable as ```json string``` results in a failure because the ```/recipe``` api changes the json string selection to just ```string```.

jaredlockhart: @jrbenny35 Yeah it has to be `"json string"` going into the input and `"string"` coming out in the recipe, but we need to do both.

app/tests/integration/tox.ini
```diff
@@ -6,7 +6,7 @@ passenv = *
 recreate = true
 commands =
     pip install -r ../requirements.txt
-    pytest --variables e2e_test_variables.json --verify-base-url --base-url https://nginx/ --driver Firefox . {posargs} -vvv
+    pytest --variables e2e_test_variables.json --verify-base-url --base-url https://nginx/ --driver Firefox . -k test_multi_pref {posargs} -vvv
```
jaredlockhart, line 9: Is this `-k test_multipref` supposed to be committed?

jrbenny35: ahh frick

PR #2650, opened 2020-05-11
Add detail view to whitelist fixes #2638

### LINKED ISSUE #2638: Add v1 api detail view to public read only API
The V1 read only API should be public, and the list view is, but the detail one is still behind auth0 right now.  That should be public as well.

uhlissuh:
Wasn't sure exactly how to test this.

I see we have a middleware test that tests that the things in the whitelist are allowed without auth.


jaredlockhart:
Could've added a test to the api view test but, yeah like you said, the middleware is tested.

PR #2659, opened 2020-05-12
Add validation for swapped pref names fixes #2640

### LINKED ISSUE #2640: Sentry failure for multipref
https://sentry.prod.mozaws.net/operations/experimenter-prod/issues/8356500/?query=is%3Aunresolved

uhlissuh:
we were getting errors when people tried to swap the names of the prefs. Solution could be deleting all of them and recreating on any updates to the prefs 😆 😆 , or this... 

prefswap

### DIFF
--- a/app/experimenter/experiments/serializers/design.py
+++ b/app/experimenter/experiments/serializers/design.py
@@ -171,16 +171,17 @@ class ExperimentDesignBranchMultiPrefSerializer(
 
     def validate_preferences(self, data):
         variant = None
-        error_list = [{}] * len(data)
-        for i, pref in enumerate(data):
-            id = pref.get('id')
+        error_list = [{"pref_name": "You cannot swap pref names."}] * len(data)
+        for pref in data:
+            id = pref.get("id")
             if id and not variant:
                 variant = VariantPreferences.objects.get(id=id).variant
-            if (variant and variant.preferences
-                    .filter(pref_name=pref["pref_name"])
-                    .exclude(id=pref.get('id'))
-                    .exists()):
-                error_list[i] = {"pref_name": "You cannot swap pref names."}
+            if (
+                variant
+                and variant.preferences.filter(pref_name=pref["pref_name"])
+                .exclude(id=pref.get("id"))
+                .exists()
+            ):
                 raise serializers.ValidationError(error_list)
 
         if not self.is_pref_valid(data):
--- a/app/experimenter/experiments/serializers/design.py
+++ b/app/experimenter/experiments/serializers/design.py
@@ -173,16 +173,14 @@ class ExperimentDesignBranchMultiPrefSerializer(
         variant = None
         error_list = [{}] * len(data)
         for i, pref in enumerate(data):
-            id = pref.get("id")
+            id = pref.get('id')
             if id and not variant:
                 variant = VariantPreferences.objects.get(id=id).variant
-            if (
-                variant
-                and variant.preferences.filter(pref_name=pref["pref_name"])
-                .exclude(id=pref.get("id"))
-                .exists()
-            ):
-                error_list[i] = {"pref_name": "This pref name already exists."}
+            if (variant and variant.preferences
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/serializers/design.py
```diff
@@ -170,6 +170,21 @@ class Meta(ExperimentDesignVariantBaseSerializer.Meta):
...
+            if id and not variant:
+                variant = VariantPreferences.objects.get(id=id).variant
+            if (
+                variant
+                and variant.preferences.filter(pref_name=pref["pref_name"])
+                .exclude(id=pref.get("id"))
+                .exists()
+            ):
+                error_list[i] = {"pref_name": "This pref name already exists."}
+                raise serializers.ValidationError(error_list)
```
jaredlockhart, line 186: I think we can simplify this two ways

1) Just do what the other check below this is doing and raise the same error for all the prefs instead of just specific ones

2) You might be able to do the whole thing with a single list comprehension?

```py
swapped_names = [variant.preferences.filter(...).exclude(...).exists() for p in data]
if any(swapped_names):
   raise for all fields
```

app/experimenter/experiments/serializers/design.py
```diff
@@ -170,6 +170,21 @@ class Meta(ExperimentDesignVariantBaseSerializer.Meta):
         model = ExperimentVariant
 
     def validate_preferences(self, data):
+        variant = None
+        error_list = [{}] * len(data)
+        for i, pref in enumerate(data):
+            id = pref.get("id")
+            if id and not variant:
```
tiftran, line 177: variant at this point is always `None` here right ?

app/experimenter/experiments/serializers/design.py
uhlissuh: Would you replace this whole for loop with this ☝️ list comprehension? Right now, `variant` is assigned inside of the for loop, using the id of the first existing preference. Would you do that in a separate step?

uhlissuh:
error message on all fields.

Screen Shot 2020-05-12 at 3 24 47 PM

app/experimenter/experiments/serializers/design.py
```diff
@@ -170,6 +170,20 @@ class Meta(ExperimentDesignVariantBaseSerializer.Meta):
         model = ExperimentVariant
 
     def validate_preferences(self, data):
+        variant = None
```
jaredlockhart, line 173: `variant = self.model.objects.get(id=self.data.get('id'))`

app/experimenter/experiments/serializers/design.py
```diff
@@ -170,6 +170,20 @@ class Meta(ExperimentDesignVariantBaseSerializer.Meta):
...
+        error_list = [{"pref_name": "You cannot swap pref names."}] * len(data)
+        for pref in data:
+            id = pref.get("id")
+            if id and not variant:
+                variant = VariantPreferences.objects.get(id=id).variant
```
jaredlockhart, line 178: Might be able to get rid of this?

app/experimenter/experiments/serializers/design.py
```diff
@@ -170,6 +170,20 @@ class Meta(ExperimentDesignVariantBaseSerializer.Meta):
...
+            if (
+                variant
```
jaredlockhart, line 180: If you look up the variant at the top, you can move

```py
if variant:
```

to the top before the loop

app/experimenter/experiments/serializers/design.py
uhlissuh: I tried this, but self doesn't have a model attribute, or so the error says. Another odd this is that self.data was coming in with `none` and blank fields, which was confusing. I think we should go with what I have.

jaredlockhart: Ah it should be `self.Meta.model`, but yes, it works as is so land it 👍

PR #2671, opened 2020-05-18
Split fixtures for e2e tests.

jrbenny35:
Was adding some addon e2e tests and noticed this fixtures are getting very ugly and confusing. I split up the design page fixtures to be easier to read and use.

### DIFF
--- a/app/tests/integration/conftest.py
+++ b/app/tests/integration/conftest.py
@@ -51,6 +51,7 @@ def _verify_url(request, base_url):
 
 @pytest.fixture
 def fill_overview(selenium, base_url, ds_issue_host, request, variables):
+    """Fills overview page."""
     selenium.get(base_url)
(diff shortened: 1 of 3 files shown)


jrbenny35:
I might still improve on this but this is a good step.

app/tests/integration/conftest.py
```diff
@@ -103,89 +105,92 @@ def fill_timeline_page(selenium, base_url, request, variables, fill_overview):
 
 @pytest.fixture
 def fill_design_page(selenium, base_url, request, variables, fill_overview):
```
jaredlockhart, line 107: Is this for Generic?   Can you specify in the name which type this is for?

jrbenny35: Yeah Generic, I'll add that to it

PR #2672, opened 2020-05-19
swap urlize and markdown template filter fixes #2637

### LINKED ISSUE #2637: Symbols are escaped in markdown code snippets
Not a big deal but I noticed as I was using code blocks in markdown, I noticed some pretty aggressive escaping going on: 

Expected: `"branch_1"`
Actual: image

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-172)

tiftran:
Screen Shot 2020-05-18 at 11 14 04 PM

A semi fix.... multi-line code blocks unfortunately doesn't seem to work though....

### DIFF
--- a/app/experimenter/templates/experiments/section_analysis.html
+++ b/app/experimenter/templates/experiments/section_analysis.html
@@ -6,7 +6,7 @@ Analysis
 
 {% block section_content %}
   <div id="analysis-content">
-    {{ experiment.analysis|urlize|markdown }}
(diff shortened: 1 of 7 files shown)


jaredlockhart:
I don't think this is gonna work:

Screenshot 2020-05-19 15 28 01

We might just be stuck with the way it works now, the thing we have to prioritize is that people pasting plaintext urls get picked up as links and accept that some edge casey markdown formatting will break.

PR #2682, opened 2020-05-25
Export experiments to csv fixes #2634

### LINKED ISSUE #2634: Using "related projects" to group experiments for reporting to managers/leadership
ckarlof relayed this scenario - that Selena wanted.  dcamp also shared this as a want a while back for his areas owners to be able to present cohesive information to him.  Experimenter is good source for lots of info on one experiment - but not to get an overview across many experiments/learnings happening in one area.

**Scenario**:  As an area owner needing to present to my director or VP - I want the ability to: 

- use Experimenter filter to easily discover all experiments meeting my current presentation need.  The filters work well, but I need the ability to customize the info I'm getting back for a particular need.  Example filter could be by for my area (related projects) that started in the last quarter.  

- I'd love an "export" option to send that info to a CSV so I can easily manipulate that information.  Example use is presenting a view into all the ways we are looking to move the needle in my area.  That covers information beyond what is in experimenter: different types of research, experiments, live changes, etc.  I may also want to combine with planning update - ex: next steps to taking action on based on learnings, ETA for in product.  In the first comment are visuals from spreadsheet examples Baine consulting shared with our executives on how they handle test, learn, adapt communication with leadership.

**Experimenter question**:  Can we easily add an export button to the filter page that dumps into a CSV?
We should dump more info than needed and so the customer can choose to hide/delete unwanted columns.  Looking through - this is my initial pass of order and info that should move over.

* public name, 
* name, 
* experimenter URL, 
* related projects, 
* status, 
* proposed start date, 
* proposed duration, 
* firefox channel, 
* short description (400 character limit), 
* public description, 
* experiment owner, 
* data scientist, 
* engineering owner, 
* objective, 
* results URL, 
* results initial, 
* related work (255 character limit), 
* data science URL, 
* feature bugzilla URL, 
* related to, 
* parent, 
* experiment type, 
* normandy slug. 
- not all deliveries have the exact same fields - so if the fields don't match a specific experimenter type - just leave blank.

**Experimenter question**:  Can Shell manually add existing experiments to projects?  I added projects for privacy and security and search - but I can't see a field to add that detail to individual  experiments in experimenter admin tool.

jaredlockhart:
This introduces the ability to export experiments to CSV for use in status reports, etc.

The workflow is

1) Land on list page
2) Use the filters on the right to filter for whatever you want
3) Click the 'Export to CSV' link at the top left
4) Receive CSV!

Screenshot 2020-05-25 18 16 35

### DIFF
--- a/app/experimenter/docs/openapi-schema.json
+++ b/app/experimenter/docs/openapi-schema.json
@@ -5,6 +5,158 @@
     "version": ""
   },
   "paths": {
+    "/api/v1/experiments/csv/": {
+      "get": {
+        "operationId": "listExperiments",
+        "description": "",
+        "parameters": [],
+        "responses": {
+          "200": {
+            "content": {
+              "text/csv": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "properties": {
+                      "name": {
+                        "type": "string",
+                        "maxLength": 255
+                      },
+                      "type": {
+                        "enum": [
+                          "pref",
+                          "addon",
+                          "generic",
+                          "rollout",
+                          "message"
+                        ]
+                      },
+                      "status": {
+                        "enum": [
+                          "Draft",
+                          "Review",
+                          "Ship",
+                          "Accepted",
+                          "Live",
+                          "Complete"
+                        ]
+                      },
+                      "experiment_url": {
+                        "type": "string",
+                        "readOnly": true
+                      },
+                      "public_name": {
+                        "type": "string",
+                        "nullable": true,
+                        "maxLength": 255
+                      },
+                      "public_description": {
+                        "type": "string",
+                        "nullable": true
+                      },
+                      "owner": {
+                        "type": "string",
+                        "readOnly": true
+                      },
+                      "analysis_owner": {
+                        "type": "string",
+                        "readOnly": true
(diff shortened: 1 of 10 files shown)


tiftran:
It doesn't seem to be able to export when it's filtering by subscribed experiments for me?

ncloudioj:
While I am still wrapping my head around the codebase, this patch looks good to me.

 r=nanj, plus the issue that @tiftran has pointed out.

Noticed that the JSON API schema defined in openapi.json and swagger.html they look quite similar, if they are exactly the same, maybe we could have a single copy and then include it to those two files? As an example, we're doing this in the [Mozilla-pipeline-schemas](https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/master/templates/include)  repo.

jaredlockhart:
@ncloudioj Thanks for the review!  The JSON is the same in both files, the HTML one is a template and we just inject the JSON into it:

https://github.com/mozilla/experimenter/blob/master/app/experimenter/docs/swagger-ui-template.html#L13

ncloudioj:
> The JSON is the same in both files, the HTML one is a template and we just inject the JSON into it.

Ah nice, good to know 👍

PR #2684, opened 2020-05-26
Updates to conftest and fixtures for e2e integration test.

jrbenny35:
Just a fixup from working with @hackebrot

### DIFF
--- a/app/tests/integration/conftest.py
+++ b/app/tests/integration/conftest.py
@@ -49,30 +49,46 @@ def _verify_url(request, base_url):
         session.get(base_url, verify=False)
 
 
+@pytest.fixture(name="experiment_type")
+def fixture_experiment_type():
+    """Empty fixture for overriding"""
+    return
+
+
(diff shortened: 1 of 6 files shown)


app/tests/integration/conftest.py
```diff
@@ -49,30 +49,46 @@ def _verify_url(request, base_url):
         session.get(base_url, verify=False)
 
 
+@pytest.fixture(name="experiment_type")
+def fixture_experiment_type():
```
hackebrot, line 53: Hi @jrbenny35! 👋 

If `experiment_type` is defined in modules containing tests that depend on this to be set, you should be able to delete this here. Otherwise you might be seeing `KeyError` exceptions in `experiment_overview_info` instead of a pytest error.

jrbenny35: So I left it here because the HTML validation tests don't need an experiment type. I guess I could just split the ```fixture_experiment_info``` into a default and a ```variables``` selected one

PR #2692, opened 2020-05-26
Add rapid experiment base template fixes #2685

### LINKED ISSUE #2685: Create empty Rapid Message template
Let's create an empty template where the react skeleton will live, so this should include

- A button on the landing page to create a new rapid message experiment
- A view/template with just the header where the rest of the react flow can live

jaredlockhart:
This introduces a new view and base template for the new rapid experimentation type.  I hid it behind the old `FEATURE_MESSAGE_TYPE` flag because that's already set on `stage=True` and `prod=False`.  Next we'll add the React skeleton to this.

### DIFF
--- a/app/experimenter/templates/experiments/rapid.html
+++ b/app/experimenter/templates/experiments/rapid.html
@@ -1,0 +1,8 @@
+{% extends "base.html" %}
+
+{% block main %}
+Rapid Experiments
+{% endblock %}
+
+{% block extrascripts %}
+{% endblock %}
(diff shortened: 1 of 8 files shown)


app/experimenter/templates/experiments/rapid.html
```diff
@@ -0,0 +1,8 @@
+{% extends "base.html" %}
+
+{% block main %}
+Rapid Experiments
+{% endblock %}
+
+{% block extrascripts %}
+{% endblock %}
```
rehandalal, line 8: nittiest of nits: EOF new line

jaredlockhart: Yeah we don't have any way to autoformat the templates, I've been looking....

PR #2694, opened 2020-05-27
 Refactor API file layout fixes #2693 

### LINKED ISSUE #2693: Reorganize API files
Rather than the structure we have now in the `experiments/` app, I'd like to have a structure more like

```
experiments/api/v1/serializers.py|views.py|urls.py
experiments/api/v2/serializers.py|views.py|urls.py
experiments/api/v3/serializers.py|views.py|urls.py
```

So that it's easier to work on stuff in isolation

jaredlockhart:
So this is a LOT but it's basically only reorganizing everything, so nothing should functionally change.

I moved serializers to the same place where their models live (so Country/Locale move to base, project moves to project, recipe serializers move to normandy), same with their respective factories.

Inside experiments I made an `api` folder with `v1` and `v2` and then each of their urls/views/serializers all live in there, and then reorganized the tests to match that.

I'm about to add a new API for the new rapid stuff so I wanted a clearer structure for that.

So it looks like a lot but theoretically it's only shuffling things around.  The thing I was checking was that we started with 522 tests and ended with 522 tests, and everything's passing.  So if our test infrastructure and static checks are doing their job I haven't broken anything.  Also, once again, this is another perfect reason for the integration tests to make sure I *REALLY* didn't break anything, so once again, props to @jrbenny35 !


.gitignore
```diff
@@ -25,10 +25,6 @@ var/
 .installed.cfg
 *.egg
 
-#Symlink
-normandy
-delivery-console
-
```
jaredlockhart, line 31: Oh and I had to get rid of this otherwise I coudln't commit new files to the `normandy` app, so we'll have to figure out a better solution to this.  Maybe using different names for the symlinks that are unique and explicit like `normandy-app` or something?

PR #2695, opened 2020-05-27
Tweak OIDC middleware to support non-docker workflow

rehandalal:
cc: @tiftran - since you've asked me before about setting up without docker this might be of interest

r?


.gitignore
```diff
@@ -70,6 +70,7 @@ target/
 
 # Environment file
 .env
+.envrc
```
rehandalal, line 73: added to support [direnv](https://direnv.net/)

app/experimenter/openidc/middleware.py
```diff
@@ -30,7 +30,8 @@ def __call__(self, request):
         except Resolver404:
             pass
 
-        openidc_email = request.META.get(settings.OPENIDC_EMAIL_HEADER, None)
+        default_email = settings.DEV_USER_EMAIL if settings.DEBUG else None
```
jaredlockhart, line 33: Coulda done the debug check in settings.py and set DEV_USER_EMAIL to None in there and then just refer to here and use the None check on the following line to catch both.

jaredlockhart: But that's just a thought, this is fine as is.

rehandalal:
> Doesn't this need the settings.py change as well?

It did when I had added a new setting but then I realized we already had one with an email and this just made the behavior consistent with Docker where you cannot choose the dev email (afaict)

PR #2721, opened 2020-06-03
Move shared dev dependencies to root package.json fixes #2719

jaredlockhart:
The lint commands weren't working for me, it says it can't find eslint, I read online that with yarn workspaces it's okay to put shared dev dependencies in the root package.json and that seemed to fix it, not sure if that's the correct pattern?

PR #2722, opened 2020-06-03
Add Rapid Experimenter API fixes #2674

### LINKED ISSUE #2674: Initial Rapid Experiment Intake API
We need a new API for writing experiments from the new rapid message intake UI, I think we should probably prefix this one as the beginning of `/api/v3/`.

Here is the mockup of the form:

Screenshot 2020-06-01 16 42 03

The first pass of this should include

- Public Name
- Hypothesis
- Firefox Version (optional)

Owner can be automatically populated by the backend based on the requesting user

jaredlockhart:
Here's the initial API, only covers `name` and `hypothesis` (`objectives` in the API/DB) for
- create
- edit
- get detail

Depends on https://github.com/mozilla/experimenter/pull/2721


app/experimenter/static/core/package.json
```diff
@@ -44,28 +44,7 @@
     "react-select": "3.1.0"
   },
   "devDependencies": {
```
pdehaan, line 46: Random thought, but this feels like a possibly odd place for a package.json file since it gets served at https://experimenter.services.mozilla.com/static/core/package.json

rehandalal: in general i dont think any of these should have lived in the `static` folder but stuff was already there so i was trying to minimize how much i reorganized files. i think the typical pattern is to create a source directory for all the front end code that django never sees and add the output folder for the webpack/parcel build as a static directory in django. i can go do that work if anyone wants.

app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+            "slug",
+            "objectives",
+        )
+
+    def create(self, validated_data):
+        experiment = super().create(validated_data)
+        experiment.slug = slugify(experiment.name)
+        experiment.owner = self.context["request"].user
+        experiment.save()
+        return experiment
```
rehandalal, line 29: this is a definite nit so feel free to disregard.

This seems clearer to me:
```py
def create(self, validated_data):
  validated_data.update({
    "slug": slugify(experiment.name),
    "owner": self.context["request"].user,
  })
  return super().create(validated_data)
```

PR #2794, opened 2020-06-09
changed date calc to reflect actual when dates exists fixes #2202

### LINKED ISSUE #2202: Date Math issues for various phases
Screen Shot 2020-01-24 at 11 39 45 AM

I was looking at some past experiments and noticed the durations for Total Duration and Observation were both off by quite a bit (this one was https://experimenter.services.mozilla.com/experiments/activity-stream-contextual-onboarding-experiment-for-dynamic-triplets-on-newtab-in-firefox-72/)

tiftran:


app/experimenter/experiments/models.py
```diff
@@ -444,6 +444,12 @@ def _transition_date(self, old_status, new_status):
             if change.old_status == old_status and change.new_status == new_status:
                 return change.changed_on.date()
 
+    @property
+    def enrollment_complete_date(self):
+        change = self.changes.filter(message="Enrollment Complete")
```
jaredlockhart, line 449: We gotta find a better way to deal with this case...

Filed a followup: https://github.com/mozilla/experimenter/issues/2798

app/experimenter/experiments/models.py
```diff
@@ -444,6 +444,12 @@ def _transition_date(self, old_status, new_status):
...
+        if change:
+            return change[0].changed_on.date()
```
jaredlockhart, line 451: Nit: if this is storing a list I'd pluralize the variable: `changes`

jaredlockhart: Variable names: the true Python type system XD

app/experimenter/experiments/models.py
jaredlockhart: Oh oh, no make this behave the same as the others, if there's a real enrollment complete date return that otherwise return the computed one

jaredlockhart: Oh you already have that below:

https://github.com/mozilla/experimenter/pull/2794/files#diff-562e6046de58c57f7e5e7f6ecdb15efbR486-R488

Just use that everywhere?

app/experimenter/experiments/models.py
```diff
@@ -473,10 +483,18 @@ def ending_soon(self):
...
+    def enrollment_duration(self):
+        return (self.enrollment_end_date - self.start_date).days
 
     @property
     def observation_duration(self):
+        if self.enrollment_complete_date:
+            return (self.end_date - self.enrollment_complete_date).days
         if self.proposed_enrollment:
             return self.proposed_duration - self.proposed_enrollment
         return 0
```
jaredlockhart, line 500: And then fold these two cases together so it's just the top one and they'll always either be the real values or the computed ones?

app/experimenter/experiments/models.py
tiftran: relooking at what's already there, i think it makes sense to fold it into `enrollment_end_date`
I made changes around this, i think the time math still makes sense...

app/experimenter/experiments/models.py
```diff
@@ -473,12 +477,22 @@ def ending_soon(self):
...
+    def enrollment_duration(self):
+        return (self.enrollment_end_date - self.start_date).days
 
     @property
     def observation_duration(self):
-        if self.proposed_enrollment:
-            return self.proposed_duration - self.proposed_enrollment
+        if self.enrollment_end_date:
+            duration = (self.end_date - self.enrollment_end_date).days
+            if duration > 0:
```
jaredlockhart, line 494: Is this check necessary?

PR #2800, opened 2020-06-10
added changelogs for rapid experiments fixes #2717

### LINKED ISSUE #2717: Create changelogs for Rapid Experiment changes
Whenever a rapid experiment is created or modified we should also create a corresponding changelog.

tiftran:


app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -26,4 +27,11 @@ def create(self, validated_data):
         experiment.slug = slugify(experiment.name)
         experiment.owner = self.context["request"].user
         experiment.save()
+
+        self.update_changelog(experiment, validated_data)
```
jaredlockhart, line 31: Ah then actually rather than setting `slug` and `owner` on the instance, we should stick them in `validated_data` so that they also get reflected in the changelog

jaredlockhart: @rehandalal was right all along, as always ;D

tiftran: experimenter slug isn't picked up since it's not a field in ChangeLogSerializer, not sure if you wanted that changed?

PR #2805, opened 2020-06-12
resorted out mixins and tasks fixes #2796

### LINKED ISSUE #2796: Split tasks into per app tasks.py
We should take the giant tasks.py we have in experiments and split it up so the tasks pertaining to each app go in that app:

* bugzilla
* normandy

and then when I add remote settings I'll add a tasks file just for that new app.

tiftran:


tiftran:
Agreed that it should be tested on stage before it reaches to prod.... will leave PR unmerged for now

PR #2774, opened 2020-06-04
ESLint all the things and set React version

pdehaan:
Fixes #2772 
Fixes #2773

PR #2833, opened 2020-06-16
Use ext for eslint

k88hudson:
I'd recommend using the `ext` syntax, which also removes the need for `--no-error-on-unmatched-pattern`, i.e.`--ext=.js,.jsx,.ts,.tsx ` v.s. `./*.ts ./*.tsx ./**/*.ts ./**/*.tsx ./**/*.js` 

There are some inconsistencies between ts and js linting at the moment (@pdehaan mentioned on in other patches) that can be resolved with changes to `.eslintrc.js` but i can file those as a follow-up

Review steps:
1. Pull `k88hudson:eslint-rapid-errors`
2. Try linting, add your own errors if you want. 
3. Ensure you see the lint task fail for warnings, errors


k88hudson:
Oops, according to the docs the max-warnings syntax is fine like it is, so I'll switch this to just ext

pdehaan:
Ref https://github.com/mozilla/experimenter/pull/2774#discussion_r436060822

I think there were some JavaScript config files in the root dir That @rehandalal didnt want linted, so we might need to explicitly ignore in .eslintignore file.

PR #2847, opened 2020-06-18
set public description for rapid a/a fixes #2846

### LINKED ISSUE #2846: Set public description for rapid A/A
The list page requires a public_description to be set, so even though we don't expose it in the form we have to set a value.  We should use that constant we're using for the bugzilla.

tiftran:


app/experimenter/experiments/constants.py
```diff
@@ -1231,9 +1231,5 @@ def FEATURE_TYPE_CHOICES(cls):  # pragma: no cover
...
 {experiment.experiment_url}
         """
-    BUGZILLA_RAPID_EXPERIMENT_TEMPLATE = """ This is an empty CFR A/A experiment.
-     The A/A experiment is being run to test the automation, effectiveness,
-     and accuracy of the rapid experiments platform.
-     The experiment is an internal test,
-     and Firefox users will not see any noticeable change
-     and there will be no user impact."""
+    BUGZILLA_RAPID_EXPERIMENT_TEMPLATE = """ This is an empty CFR A/A experiment.The A/A experiment is being run to test the automation, effectiveness, and accuracy of the rapid experiments platform.
+    The experiment is an internal test, and Firefox users will not see any noticeable change and there will be no user impact."""  # noqa
```
tiftran, line 1235: It looked clunky...

app/experimenter/experiments/constants.py
```diff
@@ -1231,9 +1231,5 @@ def FEATURE_TYPE_CHOICES(cls):  # pragma: no cover
 
 {experiment.experiment_url}
         """
-    BUGZILLA_RAPID_EXPERIMENT_TEMPLATE = """ This is an empty CFR A/A experiment.
-     The A/A experiment is being run to test the automation, effectiveness,
-     and accuracy of the rapid experiments platform.
-     The experiment is an internal test,
-     and Firefox users will not see any noticeable change
-     and there will be no user impact."""
+    BUGZILLA_RAPID_EXPERIMENT_TEMPLATE = """ This is an empty CFR A/A experiment.The A/A experiment is being run to test the automation, effectiveness, and accuracy of the rapid experiments platform.
```
pdehaan, line 1234: nit: add space between the "experiment." and "The A/A"

app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -11,6 +11,7 @@ class ExperimentRapidSerializer(ChangelogSerializerMixin, serializers.ModelSeria
     rapid_type = serializers.HiddenField(default=Experiment.RAPID_AA_CFR)
     owner = serializers.ReadOnlyField(source="owner.email")
     slug = serializers.ReadOnlyField()
+    public_description = serializers.ReadOnlyField()
```
jaredlockhart, line 14: Actually I think this can be a hidden field since we only wanna set it but we don't need it in the frontend....

app/experimenter/experiments/constants.py
tiftran: eek, lemme fix that, thanks @pdehaan

app/experimenter/experiments/api/v3/serializers.py
tiftran: kk will do

app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -11,6 +11,9 @@ class ExperimentRapidSerializer(ChangelogSerializerMixin, serializers.ModelSeria
     rapid_type = serializers.HiddenField(default=Experiment.RAPID_AA_CFR)
     owner = serializers.ReadOnlyField(source="owner.email")
     slug = serializers.ReadOnlyField()
+    public_description = serializers.HiddenField(
+        default=Experiment.BUGZILLA_RAPID_EXPERIMENT_TEMPLATE
```
jaredlockhart, line 15: Oh snap I didn't even think of that.  BOOYAH 🎉

PR #2877, opened 2020-06-22
Add Generic experiment end-to-end test. #2222

### LINKED ISSUE #2222: End to End Test: Generic Experiment

jrbenny35:
Closes #2222


app/tests/integration/conftest.py
```diff
@@ -183,6 +183,23 @@ def fill_design_page(selenium, base_url, request, fill_overview):
     return design
 
 
+@pytest.fixture
+def fill_design_page_generic_experiment(
+    selenium, base_url, request, default_data, experiment_type, fill_overview
+):
+    """Fills design page according to single pref requirements."""
+    design = DesignPage(selenium, base_url, experiment_url=f"{fill_overview.url}").open()
+    current_branchs = design.current_branches
```
pdehaan, line 192: typo: `current_branchs` 👉  `current_branches` (es).

app/tests/integration/test_generic_experiment_e2e.py
```diff
@@ -0,0 +1,84 @@
...
+    fill_timeline_page,
+    fill_design_page_generic_experiment,
+    fill_analysis_page,
+    fill_risks_page,
+    signoff_and_ship,
+    variables,
+    default_data,
+):
+    url = urlparse(selenium.current_url)
+    experiment_url = f"{url.scheme}://{url.netloc}/api/v1{url.path}recipe/"
```
pdehaan, line 77: Will everything always be `/v1/`, or does that need to be a variable too?

app/tests/integration/test_generic_experiment_e2e.py
```diff
@@ -0,0 +1,84 @@
...
+    experiment_json = requests.get(f"{experiment_url}", verify=False).json()
```
pdehaan, line 78: Apologies, ignorant Python question alert!
Why `f"{experiment_url}"` and not just `experiment_url` since you aren't interpolating anything else into the string?

app/tests/integration/test_generic_experiment_e2e.py
```diff
@@ -0,0 +1,84 @@
...
+    api_json = APIDataClass(**experiment_json)
+    assert default_data.action_name == api_json.action_name
+    assert default_data.experiment_type == api_json.filter_object[1].type
+    assert default_data.channels.lower() in api_json.filter_object[1].channels
+    assert default_data.min_version == api_json.filter_object[2].versions[0]
+    assert default_data.max_version == api_json.filter_object[2].versions[-1]
```
pdehaan, line 84: Were you saying https://pydantic-docs.helpmanual.io/ would help avoid a big list of assertions? Or does that not work here because we're calling `.lower()` and `versions[-1]`, etc?

app/tests/integration/test_generic_experiment_e2e.py
jrbenny35: @jaredlockhart ^

app/tests/integration/test_generic_experiment_e2e.py
jrbenny35: Oh that's a good one

app/tests/integration/test_generic_experiment_e2e.py
jrbenny35: Yeah so with this pydantic doesn't work like in my demo because the api is in a weird structure compared to the base dataclass. And it doesn't make sense trying to match it, so we cant just do ```assert obj == obj```. There are a bunch of little issues on why we cant do it how I showed it.

jrbenny35: We will probably iterate on it though, so if you have an idea please open an issue @pdehaan

app/tests/integration/test_generic_experiment_e2e.py
jaredlockhart: V1 is the public api so for now yes, it'll always be V1.

pdehaan: Sorry, my eyes just spotted this in my Gmail notification...
I also don't know the exact shape of the data, but if `{url.path}` doesn't always have a leading and trailing `/`, I imagine this will have potential issues.

jaredlockhart: Oh yeah, the better thing to do is use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

but since we do this pattern in a bunch of places that can probably be a fix in another issue

pdehaan: > V1 is the public api so for now yes, it'll always be V1.

Ah, OK. Sorry, I think I was just confused with the V3 stuff in https://github.com/mozilla/experimenter/tree/master/app/experimenter/experiments/api/v3

jaredlockhart: Yeah that's an internal only api so it's only covered by its internal unit tests.  When we have integration tests for the new ui, it'll be indirectly tested, but the integration tests shouldn't hit it directly.

app/tests/integration/test_generic_experiment_e2e.py
jaredlockhart: Yeah @jrbenny35 and I spent a bunch of time trying to figure out how to get them to match but there wasn't an obvious good way so I think this is still totally fine.

PR #2886, opened 2020-06-24
Add pref rollout e2e test. #2228

### LINKED ISSUE #2228: End to End Test: Rollout (Pref)

jrbenny35:
Closes #2228


app/tests/integration/conftest.py
```diff
@@ -133,7 +134,11 @@ def fill_overview(selenium, base_url, ds_issue_host, default_data):
     experiment.public_name = default_data.user_facing_name
     experiment.public_description = default_data.user_facing_description
     experiment.internal_description = "Testing in here"
-    experiment.ds_issue_url = f"{ds_issue_host}DS-12345"
+    # ds issue url is not always needed
+    try:
+        experiment.ds_issue_url = f"{ds_issue_host}DS-12345"
+    except ElementNotInteractableException:
+        pass
```
jaredlockhart, line 141: Oh interesting, so there's a little js snippet that hides that field for the rollout type, so then when this thing tries to set it it throws that little error.  Do you wanna do that check here in the fixture, or here in the actual field setter:

https://github.com/mozilla/experimenter/blob/master/app/tests/integration/pages/experiment_overview.py#L127-L132

Doesn't seem like it makes a big difference now since we're only setting it in the one place, but if we ever need to set it somewhere else it might make sense to put that logic in the field setter itself.

Also the way that field hiding logic works, it listens to a change on the type dropdown then sets it, so I wonder if we might run into a race condition where this thing tries to set it before that's finished....  But in this example it would set the field, field isn't required, I don't think that would throw any errors or break the test.  But something to keep in mind if we start seeing some weird behaviour later.

jrbenny35: Logic in the field setter would be better I should have done that!

pdehaan:
> Great job, thanks @jrbenny35 ! 👏

I will +💯  on @jrbenny35 being super awesome! ❤️

PR #2892, opened 2020-06-25
Docker ignore .cache folders fixes #2891

### LINKED ISSUE #2891: Rapid assets should be stored in a volume
Rapid assets are bleeding out to the host filesystem, they should stay in a volume.

jaredlockhart:
Aha I think I got it, this should prevent rebuilding static assets on docker build if the asset sources haven't changed!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,8 @@ target/
 
 version.json
 
+.vscode/
+
 .psql_history
 
 # Hosted assets
@@ -86,3 +88,4 @@ app/experimenter/static/assets/
 .tmontmp
 app/coverage_html_report/
 node_modules
+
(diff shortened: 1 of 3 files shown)


jaredlockhart:
@rehandalal so the `.dockerignore` file is used to ignore the files that get considered as part of a docker build.  If docker doesn't see any files have changed, it uses its build cache instead of rebuilding.  So what was happening was the watchers were spitting `app/experimenter/static/rapid/.cache/wahtever.json` files back out to the host filesystem and becuase I only had `.cache` in the `.dockerignore`, it wasn't ignoring `.cache` folders at arbitrary depth, so adding `**/.cache/` says to ignore those folders at any depth.  That seems to fix it.

PR #2896, opened 2020-06-25
Added feature/audience choices

k88hudson:
Still in progress, need to change the constants defined for features/audiences


jaredlockhart:
This is a good start but let's just split this up into attaching the existing dummy values and then replacing those with the real values after they land in Nimbus Shared.

PR #2958, opened 2020-07-02
add check for valid slug fixes #2946

### LINKED ISSUE #2946: Rapid Experiment with no/invalid slug should return a validation error
If we enter something for name that will slugify to an empty slug, we have to throw a validation error.

Example:

name = `","`
slug = `""`

tiftran:


app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -30,6 +30,14 @@ class Meta:
             "audience",
         )
 
+    def validate_name(self, name):
+        slug = slugify(name)
+        if slug:
+            return name
+        raise serializers.ValidationError(
+            ["Name needs to contains alphanumeric characters"]
```
pdehaan, line 38: nit: Should this be "contain" singular? (here and then in the assert below)

app/experimenter/experiments/tests/api/v3/test_serializers.py
```diff
@@ -156,3 +156,21 @@ def test_serializer_creates_changelog_for_updates(self):
                 changed_values=changed_values,
             ).exists()
         )
+
+    def test_serializer_returns_errors_for_non_alpha_numeric_name(self):
+
+        data = {
+            "name": "!!!!!!!!!!!!!!!",
```
pdehaan, line 163: I should probably just verify on Experimenter Stage using un-rapid experiments, but... Do we have somewhere that verifies that a slug isn't already used? Or do name collisions on slugs not matter?

tiftran: oooh you're right... should check for slug uniqueness as well

app/experimenter/experiments/api/v3/serializers.py
tiftran: gah, bad grammar 😅

app/experimenter/experiments/tests/api/v3/test_serializers.py
jaredlockhart: Good catch @pdehaan let's do that as a followup: https://github.com/mozilla/experimenter/issues/2959

PR #2960, opened 2020-07-02
slug uniqueness for rapid fixes #2959

### LINKED ISSUE #2959: Name/slug must be unique for rapid experiments
We need to slugify the name and check that the slug is unique when validating the name of a rapid experiment.

tiftran:


app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -32,12 +32,17 @@ class Meta:
...
-        raise serializers.ValidationError(
-            ["Name needs to contain alphanumeric characters"]
-        )
+        existing_slugs = Experiment.objects.values_list("slug", flat=True)
+
+        if not slug:
+            raise serializers.ValidationError(
+                ["Name needs to contain alphanumeric characters"]
+            )
+        if slug in existing_slugs:
```
jaredlockhart, line 41: I think you can just `if Experiment.objects.filter(slug=slug).exists()` here instead of checking membership of the whole list?

app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -32,12 +32,17 @@ class Meta:
...
+            raise serializers.ValidationError(
+                ["Name maps to a pre-existing slug, please choose another name"]
```
jaredlockhart, line 43: We can just reuse the same error it already gives: `Experiment with this name already exists.`

jaredlockhart: I wonder if there's a way to trigger that same error programmatically, but I don't see anything from a quick scan....

jaredlockhart:
@tiftran beep!

Issue #2941, opened 2020-06-29
Inconsistent select field UI in rapid experiment form


pdehaan:
The first two select fields (Features and Audience) use a custom `<select>` field, whereas the last two select fields (Trigger and Firefox Version) use the native `<select>` UI. Inconsistent, but probably not a deal breaker (since this is an internal only tool).

> Should we also convert the `<select />` fields below on L138 and L154?
>
> _Originally posted by @pdehaan in https://github.com/mozilla/experimenter/pull/2888_

![Mozilla_Experimenter](https://user-images.githubusercontent.com/557895/86060671-1441e400-ba1a-11ea-9bfb-b5fdd48a7fdc.png)

jaredlockhart:
Some of them are single choice and some are multi choice and I think only the multiple choice need to use the new one?

jaredlockhart:
Trigger is going away altogether, so I think only Features needs to use XSelect, and Audience and Firefox Version can use a regular select (single choice)

pdehaan:
We're losing Trigger for MVP? Or is that post MVP? Do we need to update the Figma spec?
/cc @AnaMedinac

jaredlockhart:
Yeah trigger is gone for MVP.  Yeah we can use this component for all selects, @tiftran informs me that you can use it for single and multi select cases so it makes more sense to maintain a single select component with consistent labeling/error handling/etc.

jaredlockhart:
Also I don't like `XSelect` as the name, especially if we're using it for single and multi case.

AnaMedinac:
Btw, I've updated the Figma to reflect the removal of trigger for the MVP.

pdehaan:
This is probably invalid now that we've lost the Trigger dropdown, and it looks like a recent PR converted that last "Firefox Version" dropdown to use the XSelect for consistency greatness.

tiftran:
Closing as #2993 has addressed this

PR #2993, opened 2020-07-06
remove trigger form fixes #2964

### LINKED ISSUE #2964: Remove `trigger` from rapid intake UI
`trigger` has been moved out of MVP1 so we should take it out of the UI.

tiftran:
Decided to smush #2941 into this PR as well


pdehaan:
Does this need a migration to remove the field?
https://github.com/mozilla/experimenter/blob/5e5c11e93085aa99f427daf9e377a4dc366d1fa5/app/experimenter/experiments/migrations/0094_message_experiments.py#L47-L51

tiftran:
> Does this need a migration to remove the field?
> 
> https://github.com/mozilla/experimenter/blob/5e5c11e93085aa99f427daf9e377a4dc366d1fa5/app/experimenter/experiments/migrations/0094_message_experiments.py#L47-L51

Based on the github issue description, it seems that trigger would isn't included in MVP1, but is still a field that we would like to have later? idk if that's correct. 
If that's the case, the trigger field isn't required, so it should not block workflows. To me, I don't see harm in leaving it there as we will need it later?, unless I'm missing something @pdehaan @jaredlockhart ?

jaredlockhart:
> > Does this need a migration to remove the field?
> > https://github.com/mozilla/experimenter/blob/5e5c11e93085aa99f427daf9e377a4dc366d1fa5/app/experimenter/experiments/migrations/0094_message_experiments.py#L47-L51
> 
> Based on the github issue description, it seems that trigger would isn't included in MVP1, but is still a field that we would like to have later? idk if that's correct.
> If that's the case, the trigger field isn't required, so it should not block workflows. To me, I don't see harm in leaving it there as we will need it later?, unless I'm missing something @pdehaan @jaredlockhart ?

Yep that's exactly right @tiftran 👍

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -163,10 +146,11 @@ const ExperimentForm: React.FC = () => {
           Is there a minimum Firefox Release version this experiment should be
           run on?
         </p>
-        <select
-          className="form-control w-100"
+        <XSelect
+          className="w-100"
           id="field-version"
           name="version"
+          options={[]}
```
jaredlockhart, line 153: Aha! Snuck that in there ;)

PR #2994, opened 2020-07-07
Rapid exp title changes dependent on new or edit fixes #2901

### LINKED ISSUE #2901: Editing rapid experiment should say 'Edit <name>' in the title instead of Create.....
Screenshot 2020-06-26 18 01 33

tiftran:
Screen Shot 2020-07-07 at 1 48 37 PM

Not entirely sure if it's desirable but when changing the public name, the title will change to match it....

### DIFF
--- a/app/experimenter/static/rapid/components/experiments/ExperimentEditHeader.tsx
+++ b/app/experimenter/static/rapid/components/experiments/ExperimentEditHeader.tsx
@@ -1,13 +1,10 @@
 import React from "react";
-import { useHistory } from "react-router-dom";
-
 import { useExperimentState } from "experimenter-rapid/contexts/experiment/hooks";
 
 const ExperimentFormPage: React.FC = () => {
-  const { location } = useHistory();
-  let pageHeading = "Create a New A/A Experiment";
-  if (location.pathname.includes("edit")) {
-    const experimentData = useExperimentState();
+  const experimentData = useExperimentState();
+  let pageHeading = `Create a New A/A Experiment: ${experimentData.name}`;
+  if (experimentData.slug) {
     pageHeading = `Edit Experiment: ${experimentData.name}`;
   }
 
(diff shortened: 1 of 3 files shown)


app/experimenter/static/rapid/components/experiments/ExperimentEditHeader.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { useHistory } from "react-router-dom";
+
+import { useExperimentState } from "experimenter-rapid/contexts/experiment/hooks";
+
+const ExperimentFormPage: React.FC = () => {
+  const { location } = useHistory();
+  let pageHeading = "Create a New A/A Experiment";
+  if (location.pathname.includes("edit")) {
```
jaredlockhart, line 9: Can't we look at the experiment data to see if it has a name instead of looking at the url?

app/experimenter/static/rapid/components/experiments/ExperimentEditHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const experimentData = useExperimentState();
+    pageHeading = `Edit Experiment: ${experimentData.name}`;
+  }
+
+  return (
+    <div className="mb-4">
```
pdehaan, line 15: Below this was wrapped in a `<div className="col pt-3">` (see deleted L9 in app/experimenter/static/rapid/components/pages/ExperimentFormPage.tsx).
Not sure if that is still needed, but 🤷 .

tiftran: I intially had it, but it gave me a weird indent 🤷‍♀️. Not exactly the best with front end stuff 😅

app/experimenter/static/rapid/components/experiments/ExperimentEditHeader.tsx
tiftran: so... i took a stab at only using experiment data. Since the name data value changes when a user changes the name in the text field, I couldn't think of a way how would that indicate whether the experiment was new, or an edit. I decided using the presence of a slug in the experiment data instead?

Issue #3012, opened 2020-07-09
Can't request review until bugzilla ticket is created


jaredlockhart:
A rapid experiment can't be serialized until it has a bugzilla ticket, and that happens asynchronously after it's created.  So in theory someone could create an experiment and very quickly hit the request approval button, it would try to serialize it and fail, and then just get stuck in a broken state.  

I was thinking we could disable the button in the UI until we know it has a bugzilla id, but then I just realized we could also handle this transparently by only attempting to serialize and push to kinto if it has a bugzilla id, and if it doesn't, just quietly pass over it and then try again when the task wakes back up, which I think is a better experience for the user.

jaredlockhart:
Add a method like

`is_ready_for_rapid_launch`

that's false if it doens't have a bugzilla_id

Issue #3009, opened 2020-07-09
Rapid detail page state for Live experiment


jaredlockhart:
When the experiment goes live, we need to reflect that on the detail page. Some ideas:

- Status pill goes to Live
- `Request Review` button is gone?
- Links to dashboards?

AnaMedinac:
Here is a screen shot of the `Live` state:

<img width="658" alt="image" src="https://user-images.githubusercontent.com/22782298/87334699-12dde480-c50d-11ea-9aab-6f02694552ec.png">


I've updated the Figma to reflect the states, here: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/WIP-Platform-Exp?node-id=214583%3A0

Also as reference here is the user flow with the states: https://miro.com/app/board/o9J_kpOgUTE=/

Issue #3003, opened 2020-07-08
Rapid detail page state for rejected experiments


jaredlockhart:
If a reviewer rejects an experiment in remote settings, a rejection reason will be entered in the remote settings admin.  We can then mark the experiment in Experimenter as rejected and show the rejection reason, and in theory the owner should be able to modify their experiment accordingly and then resubmit for review.  We'll need workflow and UX/UI for this.

AnaMedinac:
Here is a screen shot of the Rejected state:

<img width="644" alt="image" src="https://user-images.githubusercontent.com/22782298/87333048-8df1cb80-c50a-11ea-8140-eaab216870e3.png">

I've updated the Figma to reflect the states, here: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/WIP-Platform-Exp?node-id=214583%3A0

Also as reference here is the user flow with the states: https://miro.com/app/board/o9J_kpOgUTE=/

jaredlockhart:
So
- Add a `rejected` status to experiment statuses
- Add it to the state transitions for rapid experiments
- Update the `check_kinto_queue` task becuase i think the way this will have to work is
  - experiment gets pushed to remote settings
  - reviewer rejects
  - now there is nothing in the queue but if we push immediately we lose the rejection message so you'll have to FIRST check if there's a rejection message, create a changelog that sets status to rejected and put the rejection message in the changelog, then proceed to pushing hte next one into the queue
- Add `rejection_reason` to V3 API that comes out of the REVIEW -> REJECTED changelog
- Update detail page to show rejection reason and allows for editing which then sets it back to DRAFT

jaredlockhart:
After I land https://github.com/mozilla/experimenter/issues/3146 we should also omit rejected experiments from V4 API

pdehaan:
If the experiment is rejected (meaning it never went live, presumably), it probably doesn't make sense to display the "Results" block of heading+text since there will never be results.

The "Back" and "Request Approval" navigation probably only makes sense if users can go back to edit the experiment and correct the issue that caused the experiment to be initially rejected and resubmit.

jaredlockhart:
@pdehaan exactly

pdehaan:
And since this is the UI topic, I can probably ask here...

Can a rejected experiment be fixed and resubmitted? Or once it's rejected is it a dead and and a user would have to redo the experiment from scratch (and now get a name/slug collision)?
[...]

Issue #2995, opened 2020-07-07
Rapid detail page state for requested review


jaredlockhart:
After the review is requested, we need a state on the detail page to show that the review has already been requested, and that the experiment can no longer be edited.

jaredlockhart:
Probably want to do a couple things:

- Update 'Draft' to 'Pending Review' or something
- Grey out the blue 'Request Review' button, maybe change its text?
- Disable the 'Back' button that takes you back to the edit page (I would think the disabled state should be grey, so maybe it shouldn't be grey when it's not disabled like it is now?)

AnaMedinac:
Here is a screen shot of the `Requested Review` state:

<img width="644" alt="image" src="https://user-images.githubusercontent.com/22782298/87332918-5aaf3c80-c50a-11ea-92ce-215d294f31a7.png">



I've updated the Figma to reflect the states, here: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/WIP-Platform-Exp?node-id=214583%3A0

Also as reference here is the user flow with the states: https://miro.com/app/board/o9J_kpOgUTE=/

PR #3013, opened 2020-07-09
Connect request review button to kinto publish task fixes #2878

### LINKED ISSUE #2878: Connect publish button in rapid detail page to publish task
Once the rapid detail page lands, we need to connect its publish button to the publish to remote settings task.  We need to queue it in the db somehow, either with the Experiments model or maybe some new model, I'm not sure.

jaredlockhart:
Okay there's a lot in here!!

There's a bunch of hard coded assumptions

- Design is hard coded (ie branches) https://github.com/mozilla/experimenter/issues/2890
- Firefox version isn't hooked up to the form yet https://github.com/mozilla/experimenter/issues/2948
- Firefox channel is missing from audience https://github.com/mozilla/nimbus-shared/issues/60
- Bucketing isn't hooked up yet https://github.com/mozilla/experimenter/issues/3014
- UI doesn't update to reflect review has been requested https://github.com/mozilla/experimenter/issues/2995
- UI doesn't update to reflect experiment is Live when accepted https://github.com/mozilla/experimenter/issues/3009
- UI doesn't update to reflect rejected review https://github.com/mozilla/experimenter/issues/3003
- Ordering dependency on bugzilla ticket number https://github.com/mozilla/experimenter/issues/3012

So we'll have to update a bunch of things after this lands.

But this takes care of the basics:


app/experimenter/docs/openapi-schema.json
```diff
@@ -7613,39 +7613,49 @@
...
+                    "status": {
                       "enum": [
-                        "AUDIENCE 1",
-                        "AUDIENCE 2"
+                        "Draft",
+                        "Review",
+                        "Ship",
+                        "Accepted",
+                        "Live",
+                        "Complete"
```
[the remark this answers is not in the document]
pdehaan: The states listed on the current [Figma spec](https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/WIP-Platform-Exp?node-id=214583%3A0) are:
1. Draft
1. Requested Approval
1. Rejected
1. Live
1. Complete

(I reordered slightly to match what I imagine the actual flow orders would be, but yeah.)

I'm fine if we need to modify the valid experiment statuses, but we should also file an issue to update the spec if we need to deviate from it so we can still maintain a source of truth in Figma.

app/experimenter/docs/openapi-schema.json
```diff
@@ -7613,39 +7613,49 @@
               "application/json": {
                 "schema": {
                   "properties": {
-                    "owner": {
-                      "type": "string",
-                      "readOnly": true
+                    "audience": {
+                      "enum": [
+                        "AUDIENCE 1",
+                        "AUDIENCE 2"
```
pdehaan, line 7619: Guess I haven't thought of this... since this is a JSON file, how will we be able to grab these enum values from the nimbus-shared library? Unless we have some other .js file which generates the openapi-schema.json files and adds a bit of scripting flair.

app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -66,14 +81,46 @@ def create(self, validated_data):
             {
                 "slug": slugify(validated_data["name"]),
                 "owner": self.context["request"].user,
+                "firefox_min_version": Experiment.VERSION_CHOICES[0][0],
```
pdehaan, line 84: Wouldnt this currently just be hardcoding everything to Firefox `"55.0"`, per app/experimenter/experiments/constants.py#L144?

app/experimenter/experiments/models.py
```diff
@@ -1078,6 +1078,12 @@ def clone(self, name, user):
 
         return cloned
 
+    def prepare_rapid_experiment_for_publish(self):
+        self.status = Experiment.STATUS_ACCEPTED
+        self.proposed_start_date = datetime.date.today()
```
pdehaan, line 1083: Ah, this reminds me... I wanted to make sure I confirmed that we have oodles of date fields in the Experimenter side. Mainly just for date that the submitter hit Save/Submit button, and the date the experiment was approved in RemoteSettings/Experimenter, and the date the experiment was set to "Live" (which I assume is the same date as approved date above).

Or more accurately, I want to know if somebody can run some queries against production to tell us what the average/median/mean times are for an experiment to actually be approved and what our outliers are (ie: an experiment that sat there for 2 weeks, or average time is 4 days, etc). I know the average will most likely be over 24 hours due to weekends and timezones, but just want to know if we can determine averages and stats.

app/experimenter/normandy/tasks.py
```diff
@@ -38,7 +38,8 @@ def update_recipe_ids_to_experiments():
 
     ready_to_ship_experiments = Experiment.objects.filter(
         status__in=[Experiment.STATUS_SHIP, Experiment.STATUS_ACCEPTED]
-    )
+    ).exclude(type=Experiment.TYPE_RAPID)
```
tiftran, line 41: should rapid experiments be excluded [here](https://github.com/mozilla/experimenter/pull/3013/files#diff-0f18d819e50bfd722ece30241e99a6f9R77) too?

jaredlockhart: That link doesnt' seem to work from me, which line were you referencing?

tiftran: gah, line 77 `update_launched_experiments`

jaredlockhart: Oh good call!  I think it would do nothing since it wouldn't find anything in normandy for those slugs anyways, but it's still making unnecessary calls 👍

app/experimenter/docs/openapi-schema.json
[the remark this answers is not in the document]
jaredlockhart: The spec is correct, these are all the possible states any experiment in Experimenter can be in, and the set of states specific to a rapid experiment are here:

https://github.com/mozilla/experimenter/pull/3013/files#diff-00de22e6c754b3d410f1d3e631e1825eR128-R131

app/experimenter/docs/openapi-schema.json
```diff
@@ -7613,39 +7613,49 @@
...
+                      "readOnly": true
+                    },
+                    "status": {
                       "enum": [
-                        "AUDIENCE 1",
-                        "AUDIENCE 2"
+                        "Draft",
+                        "Review",
+                        "Ship",
+                        "Accepted",
```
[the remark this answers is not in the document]
jaredlockhart: Yep you're exactly right, Review -> Live or Review -> Rejected.

app/experimenter/docs/openapi-schema.json
jaredlockhart: The `openapi-schema.json` is auto generated from the python code that implements the APIs, it's not something we manipulate by hand.  The dummy constants appear here right now:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants.py#L68-L70

and these are what will be replaced by the values in `nimbus-shared`.  The API JSON will update itself automatically to reflect it.

app/experimenter/experiments/api/v3/serializers.py
jaredlockhart: Exactly, it's a hard coded dummy value until we get Firefox versions hooked up, as I called out in the list of todos at the top of the PR 👍

app/experimenter/experiments/models.py
jaredlockhart: Oh yeah the way we track those is through ChangeLogs.  You just reminded me I need to add one for this publish step, thanks!

PR #3037, opened 2020-07-13
added bugzilla url to serializer fixes #3016

### LINKED ISSUE #3016: Attach bugzilla link on detail page to real bugzilla link
Let's add the bugzilla link to the serializer and attach it on the detail page

tiftran:
kinda clunky since you would need to refresh the page.... not sure if we would like that fixed in this issue.

### DIFF
--- a/app/experimenter/docs/openapi-schema.json
+++ b/app/experimenter/docs/openapi-schema.json
@@ -7639,6 +7639,10 @@
                         "AUDIENCE 1",
                         "AUDIENCE 2"
                       ]
+                    },
+                    "bugzilla_url": {
+                      "type": "string",
+                      "readOnly": true
                     }
                   },
                   "required": [
@@ -7800,6 +7804,10 @@
                         "AUDIENCE 1",
                         "AUDIENCE 2"
                       ]
+                    },
+                    "bugzilla_url": {
+                      "type": "string",
+                      "readOnly": true
                     }
                   },
                   "required": [
@@ -7943,6 +7951,10 @@
                         "AUDIENCE 1",
                         "AUDIENCE 2"
                       ]
+                    },
+                    "bugzilla_url": {
+                      "type": "string",
+                      "readOnly": true
                     }
                   },
                   "required": [
@@ -8096,6 +8108,10 @@
                         "AUDIENCE 1",
                         "AUDIENCE 2"
                       ]
+                    },
+                    "bugzilla_url": {
+                      "type": "string",
+                      "readOnly": true
                     }
                   },
                   "required": [
--- a/app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
+++ b/app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
@@ -36,7 +36,7 @@ describe("<ExperimentDetails />", () => {
     expect(getByDisplayValue("Test objectives")).toBeInTheDocument();
   });
 
-  it("renders without bugzilla info when data provided", async () => {
+  it("renders with bugzilla info when data provided", async () => {
     const { getByText } = renderWithRouter(
       wrapInExperimentProvider(<ExperimentDetails />, {
         initialState: {
@@ -54,6 +54,27 @@ describe("<ExperimentDetails />", () => {
     await waitFor(() => {
       return expect(getByText(/Bugzilla ticket/)).toBeInTheDocument();
     });
+  });
+
+  it("renders without bugzilla info when data missing", async () => {
+    const { getByDisplayValue, queryByText } = renderWithRouter(
+      wrapInExperimentProvider(<ExperimentDetails />, {
+        initialState: {
+          slug: "test-slug",
+          name: "Test Name",
+          objectives: "Test objectives",
+          owner: "test@owner.com",
+          features: ["FEATURE 1", "FEATURE 2"],
+          audience: "AUDIENCE 1",
+        },
+      }),
+    );
+
+    await waitFor(() => {
+      return expect(getByDisplayValue("test@owner.com")).toBeInTheDocument();
+    });
+
+    expect(queryByText(/Bugzilla ticket/)).toBe(null);
   });
 
   it("sends you to the edit page when the 'Back' button is clicked", async () => {
--- a/app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
+++ b/app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
@@ -36,6 +36,26 @@ describe("<ExperimentDetails />", () => {
     expect(getByDisplayValue("Test objectives")).toBeInTheDocument();
   });
 
+  it("renders without bugzilla info when data provided", async () => {
+    const { getByText } = renderWithRouter(
+      wrapInExperimentProvider(<ExperimentDetails />, {
+        initialState: {
+          slug: "test-slug",
+          name: "Test Name",
+          objectives: "Test objectives",
+          owner: "test@owner.com",
+          features: ["FEATURE 1", "FEATURE 2"],
+          audience: "AUDIENCE 1",
+          bugzilla_url: "https://example.com",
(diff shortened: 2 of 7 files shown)


app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -53,7 +53,8 @@ const ExperimentDetails: React.FC = () => {
       <LabelledRow label="Experiment Owner" value={data.owner} />
       <LabelledRow label="Public Name" value={data.name}>
         <div className="my-2">
-          Bugzilla ticket can be found <a href="#">here</a>.
+          Bugzilla ticket can be found{" "}
+          <a href={data.bugzilla_url ? data.bugzilla_url : "#"}>here</a>.
```
jaredlockhart, line 57: Oh I was gonna say just hide the text and link altogether if the url isn't present

tiftran: will do

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -48,13 +48,20 @@ const ExperimentDetails: React.FC = () => {
     // No-op
   };
 
+  let bugzilla_url;
+  if (data.bugzilla_url) {
```
pdehaan, line 52: Sorry, I'm not 100% on serializers and the backend of this... but what are the circumstances around there _not_ being a `bugzilla_url`?
Is it just a timing issue where I create a new rapido experiment and then hit "Next" while in draft and the bug hasn't been created in bugzilla yet due to timing issues, but would always be there on a subsequent page refresh?

tiftran: ^ this would be a case where there isn't a bugzilla_url. The clunkiness of needing to refresh will be addressed in diff issue/PR

another reason why there wouldn't be a bugzilla url would be if the bugzilla api for some reason fails?

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -48,13 +48,20 @@ const ExperimentDetails: React.FC = () => {
...
+    bugzilla_url = (
+      <div className="my-2">
+        Bugzilla ticket can be found <a href={data.bugzilla_url}>here</a>.
```
pdehaan, line 55: Not sure if you want a `target="_blank" rel="noopener noreferrer"` bit on this to spawn the bug in a new window to avoid losing the current Experimenter window. I'm mildly indifferent, but thought it was worth calling out in case anybody has strong feelings one way or the other (or if Experimenter has prior art on these links).

app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
```diff
@@ -36,6 +36,26 @@ describe("<ExperimentDetails />", () => {
...
+          features: ["FEATURE 1", "FEATURE 2"],
+          audience: "AUDIENCE 1",
+          bugzilla_url: "https://example.com",
+        },
+      }),
+    );
+
+    await waitFor(() => {
+      return expect(getByText(/Bugzilla ticket/)).toBeInTheDocument();
+    });
```
pdehaan, line 56: Do we have/need a test where the experiment data does _not_ have a `bugzilla_url` (and we want to assert that `Bugzilla ticket` text is not in the DOM?

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
tiftran: oooh yes

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
pdehaan: Ah, thanks. I think I was confusing it with a test above where were concatenating a `bug_id` onto `bug_host` and was thinking it'd concatenate a null/undefined `bug_id` and we'd have a URL like `https://bugzilla.whatever/show_bug.cgi?id=` (which would never be falsy) but now I see you're storing the entire URL and not just the `bugzilla_id` (or maybe you're storing both).

app/experimenter/docs/openapi-schema.json
```diff
@@ -7639,6 +7639,10 @@
                         "AUDIENCE 1",
                         "AUDIENCE 2"
                       ]
+                    },
+                    "bugzilla_url": {
+                      "type": "string",
+                      "readOnly": true
```
pdehaan, line 7645: Probably irrelevant but I found this elsewhere in the schema, and it looked interesting/relevant. Not sure if the `bugzilla_url` should also have a `"format": "uri"` or `pattern` defined.

```js
                      "feature_bugzilla_url": {
                        "type": "string",
                        "format": "uri",
                        "nullable": true,
                        "maxLength": 200,
                        "pattern": "^(?:[a-z0-9\\.\\-\\+]*)://(?:[^\\s:@/]+(?::[^\\s:@/]*)?@)?(?:(?:25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)(?:\\.(?:25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)){3}|\\[[0-9a-f:\\.]+\\]|([a-z\u00a1-\uffff0-9](?:[a-z\u00a1-\uffff0-9-]{0,61}[a-z\u00a1-\uffff0-9])?(?:\\.(?!-)[a-z\u00a1-\uffff0-9-]{1,63}(?<!-))*\\.(?!-)(?:[a-z\u00a1-\uffff-]{2,63}|xn--[a-z0-9]{1,59})(?<!-)\\.?|localhost))(?::\\d{2,5})?(?:[/?#][^\\s]*)?\\Z"
                      },
```

jaredlockhart: @pdehaan `openapi-schema.json` is automatically generated so in general I'd recommend treating it more like a build artifact than source code.

pdehaan: Good point, although now I guess I'm more confused as to why it wouldn't be more consistenter. But don't take my general curiosity and ignorance as a merge blocker.

jaredlockhart: For the sake of curiosity, the field on teh serializer you're looking at is defined here:

https://github.com/mozilla/experimenter/blob/d654aba1d65a36772fd27387375b922f9af19880/app/experimenter/experiments/api/v1/serializers.py#L174

which is just referencing the model field without explicitly declaring any serializer field interface, so DRF will just try to automatically build a serializer field out of whatever the modelfield says, which is:

https://github.com/mozilla/experimenter/blob/151f9612d802414c736a47f0c639e19a071ba55d/app/experimenter/experiments/models.py#L238

a Django URLField, and so that will include the regex validation.

But there's a weird nuance in the way we treat different bugzilla tickets, so for the feature bugzilla ticket, we only care about the URL since we just show it as a link.   But for the experiment bugzilla ticket, we need to store the actual ID:

https://github.com/mozilla/experimenter/blob/151f9612d802414c736a47f0c639e19a071ba55d/app/experimenter/experiments/models.py#L232

since we use that to make API calls to bugzilla to manipulate the ticket, and then we reconstruct the URL manually:

https://github.com/mozilla/experimenter/blob/151f9612d802414c736a47f0c639e19a071ba55d/app/experimenter/experiments/models.py#L337-L340

so we never store the experiment bugzilla ticket as an actual url, which is why, when we serialize it, Django/DRF don't know to treat it as a url, and so there is no fancy regex ;)

jaredlockhart: Also the regex would be used to validate incoming values, but we're treating it as a ready only field here, so you don't need the regex anyways ;)

PR #3041, opened 2020-07-14
2957

tiftran:


app/experimenter/static/rapid/contexts/experiment/actions.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const saveExperiment = async (
+  experimentSlug: string,
+  formData: ExperimentData,
+): Promise<Response> => {
+  const url = experimentSlug
+    ? `/api/v3/experiments/${experimentSlug}/`
+    : "/api/v3/experiments/";
+  return await fetch(url, {
+    method: experimentSlug ? "PUT" : "POST",
```
tiftran, line 33: i tried to return just the response here, but i realized that we need the slug from the response in the experiment form

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -40,38 +43,18 @@ const ExperimentForm: React.FC = () => {
...
-      : "/api/v3/experiments/";
-    const response = await fetch(url, {
-      method: experimentSlug ? "PUT" : "POST",
-      headers: {
-        "Content-Type": "application/json",
-      },
-      body: JSON.stringify(formData),
-    });
+    const response = await saveExperiment(experimentSlug, formData);
+
```
jaredlockhart, line 57: So you opted to just return the Response and unpack the errors externally?  It's fine we're gonna move that into its own reducer and we can just do it when we do that.

jaredlockhart:
Oh also just make sure the commit name uses the first commit name when you squash and merge

PR #3048, opened 2020-07-16
v4 public api fixes #3010

### LINKED ISSUE #3010: Public API for Rapid Experiments
Let's add the public API for Rapid Experiments using the Rapid Serializer here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v1/serializers.py#L252

Update:

Let's make this a new API version, so we'd have

`/api/v4/experiments/`
`/api/v4/experiments/<slug>/`

using the rapid recipe serializer for experiments of type Rapid in the following states:

Accepted
Live
Complete

and use the `get_prefetched` method on the model manager so we dont' make a ton of db calls, example:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v1/views.py#L19

tiftran:


app/experimenter/experiments/api/v4/urls.py
```diff
@@ -0,0 +1,23 @@
...
+from experimenter.experiments.api.v4.views import (
+    ExperimentListView,
+    ExperimentRapidDetailsView,
+    ExperimentRapidRecipeView,
+)
+
+
+urlpatterns = [
+    url(
+        r"^experiments$", ExperimentListView.as_view(), name="experiments-rapid-api-list",
```
jaredlockhart, line 11: Missing trailing slash here

app/experimenter/experiments/api/v4/urls.py
```diff
@@ -0,0 +1,23 @@
...
+    ),
+    url(
+        r"^experiments/(?P<slug>[\w-]+)$",
```
jaredlockhart, line 14: Missing trailing slash here

app/experimenter/experiments/api/v4/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+class ExperimentListView(ListAPIView):
+    filter_fields = ("status",)
+    queryset = Experiment.objects.get_prefetched().filter(type=Experiment.TYPE_RAPID)
+    serializer_class = ExperimentRapidSerializer
+
+
+class ExperimentRapidDetailsView(RetrieveAPIView):
+    lookup_field = "slug"
+    queryset = Experiment.objects.get_prefetched()
```
jaredlockhart, line 21: I'd filter this on type too, that way if someone puts in the wrong slug they'll get 404

jaredlockhart:
Oh actually you could simplify this even more if you use the pattern @rehandalal showed me that I used on V3:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v3/views.py#L12-L20

Just use a single viewset and mixin the list/retrieve mixins then you only have to declare a single url pattern and viewset.

app/experimenter/experiments/api/v4/urls.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework.routers import SimpleRouter
+from experimenter.experiments.api.v4.views import ExperimentRapidDetailsView
+
+
+router = SimpleRouter()
+router.register(r"experiments", ExperimentRapidDetailsView, "experiment-rapid-recipe")
```
jaredlockhart, line 6: Nit: `ExperimentRapidViewSet`

app/experimenter/experiments/api/v4/views.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import viewsets, mixins
+
+from experimenter.experiments.models import Experiment
+from experimenter.experiments.api.v4.serializers import ExperimentRapidRecipeSerializer
+
+
+class ExperimentRapidDetailsView(
```
jaredlockhart, line 7: Nit: ExperimentRapidViewSet

app/experimenter/experiments/tests/api/v4/test_serializers.py
```diff
@@ -0,0 +1,83 @@
...
+        ExperimentVariantFactory.create(experiment=experiment, slug="variant-2")
+
+        serializer = ExperimentRapidRecipeSerializer(experiment)
+        data = serializer.data
+
+        fn = os.path.join(os.path.dirname(__file__), "experimentRecipe.json")
+
+        with open(fn, "r") as f:
+            json_schema = json.load(f)
+        self.assertIsNone(validate(instance=data, schema=json_schema))
```
jaredlockhart, line 44: Oh right we had to do this becuase we couldn't validate with the shared repo last time.  Let's leave this like this and then we can file a separate ticket to switch to use the shared repo validator: https://github.com/mozilla/experimenter/issues/3058

PR #3054, opened 2020-07-16
Use real experiment status in rapid UI fixes #3053

### LINKED ISSUE #3053: Add status to V3 API and use real status on rapid detail page
Draft is hard coded right now, let's use the real status.

jaredlockhart:
There was a little hiccup because `Draft` was hard coded in the header, but that's outside the `ExperimentDetails` component, and way contexts/providers/etc work, you have to wrap the outer component in them to access the hook state (react makes you write weird sentences).  So I moved everything right into the detail component.  I think I'll go back and reorganize the way these components are laid out, the Provider should be given once right at the top so you don't run into this.  Plus I find the way the pages and components are laid out confusing.


jaredlockhart:
Also note: I *LOVE* working in TypeScript.  The feedback loop with the editor statically checking everything and the jest tests automatically running as you edit and highlighting everything is just fantastic.  Python feels so ancient and sluggish by comparison.

PR #3062, opened 2020-07-19
Use typescript-eslint parser for rapid .eslintrcjs

### LINKED ISSUE #2834: Use typescript-eslint parser for rapid .eslintrcjs
Since we are using typescript-eslint rules, it would be a good idea to use the `@typescript-eslint/parser` for the `parser` setting in `.eslintrc`.

See https://github.com/typescript-eslint/typescript-eslint#what-about-babel-and-babel-eslint

> The key trade-off can be summarized as babel-eslint supports additional syntax which TypeScript itself does not, but typescript-eslint supports creating rules based on type information, which is not available to babel because there is no type-checker.

> Because they are separate projects powered by different underlying tooling, they are currently not intended to be used together.

bhushan-borole:
Fixes #2834 
Used typescript eslint parse instead of babel eslint parse.

### DIFF
--- a/app/experimenter/static/rapid/.eslintrc.js
+++ b/app/experimenter/static/rapid/.eslintrc.js
@@ -4,7 +4,7 @@ const baseSettings = {
     browser: true,
     es6: true,
   },
-  parser: "babel-eslint",
+  parser: "@typescript-eslint/parser",
   plugins: ["import"],
   extends: [
     "eslint:recommended",


jaredlockhart:
Awesome thank you again @bhushan-borole !  Also thank you for picking up a ticket in https://github.com/mozilla/experimenter/projects/27 which is the highest priority work we have right now.  Feel free to keep picking up those P5 tickets at the bottom of the engineering column and if there's anything that's not clear in them please ask and I'll add more detailed instructions.

Also tiny request, please put `fixes #2834` at the end of the commit title rather than in the PR body, it just makes it easier to jump to issues from the commit list 👍 

Thanks for contributing!! 🎉 🎉 🎉

bhushan-borole:
Sure, I will remember that tip.
Thank you @jaredlockhart.
Looking forward for even more contributions.

PR #3064, opened 2020-07-20
Mypy is fun

jaredlockhart:
I still want to try to make this work.  It's gotten better since the last time I tried this, in terms of django integration, but the major roadblock i'm hitting is dealing with dicts in general, what we need to do is find some way to declare TS style interfaces for all of them, maybe using pydantic, maybe using TypedDict, but does that just mean creating duplicate definitions on every serializer?  I'm not sure.  This is still just idle musing right now, but if we could get it going I think it would make editors come to life more.


jaredlockhart:
This is what we'd need:

https://github.com/typeddjango/djangorestframework-stubs/issues/36

jaredlockhart:
I would like to move towards this in the future but for now there is too much uncertainty about the future direction of the codebase overall so I'm not sure it makes sense to invest heavily in this until that becomes clearer.  I'm going to close this for now and if we collectively decide it's a direction we'd like to move in we can take another stab at it.

Issue #3058, opened 2020-07-17
Use nimbus_shared to validate rapid experiment schema


jaredlockhart:
Now that the experiment schema is landed in the shared repo and the validator is there, let's use it to validate the schema in our tests.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-199)

jaredlockhart:
Note: we're missing proposedDuration, but that'll just be fixed when we validate against the proper schema.

mythmon:
@jaredlockhart can you add some details about what is blocking this?

Issue #2890, opened 2020-06-25
Create A/A Experiment with Design Template Values


jaredlockhart:
Templated design:

https://github.com/mozilla/nimbus-shared/blob/main/data/ExperimentDesignPresets.ts#L29-L34

We have to automatically create 
- branches
- proposed duration
- proposed enrollment duration

Replace this:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v3/serializers.py#L95-L104

With using those constants.

There might be a clever way to pass the JSON data from the preset directly into the serializer rather than writing a bunch of imperative logic like this https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v3/serializers.py#L86-L93

Bucketing will have to be assigned when we publish.

jaredlockhart:
And remove any unnecessary constants we added.

jaredlockhart:
Blocked on: https://github.com/mozilla/nimbus-shared/issues/68

jaredlockhart:
Blocked on https://github.com/mozilla/nimbus-shared/milestone/3

jaredlockhart:
AND in here we can use the `filter_expression` and `targeting` from the DesignPreset in the rapid experiment serializer:

https://github.com/mozilla/nimbus-shared/blob/main/data/ExperimentDesignPresets.ts#L25-L27

We'll have to `&&` the audience `targeting` together with the design preset `targeting`

PR #3091, opened 2020-07-22
hardcoded count for bucket config fixes #3014

### LINKED ISSUE #3014: Compute bucket size at launch time
When we have real data for audience and features we'll be able to compute the number of buckets an experiment needs.  Awaiting details from @tdsmith .

tiftran:


tiftran:
ooh i guess https://github.com/mozilla/experimenter/issues/2890 would cover using the design template?

jaredlockhart:
Oh and the namespace is the normandy slug

app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -53,6 +70,9 @@ class Meta:
         )
 
     def get_bucketConfig(self, obj):
+
+        if hasattr(obj, "bucket"):
```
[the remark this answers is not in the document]
jaredlockhart: Oh you're right, god that's so dumb it should just be None.  I wonder if we can add something for the OneToOne field to set this to None if it doesn't exist, we shoudln't have to use `hasattr`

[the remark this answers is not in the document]
jaredlockhart: Yeah I think the only way to do it is to do it hte other way around 

```py
if ExperimentBucketRange.objects.filter(experiment=obj).exists():
```

which is much wordier, but less fickle than checking the attribute directly.

Yeah let's just leave it like this for now.

Another way to approach this is to just not use OneToOne field, use a regular foreign key, use a `unique` Meta check to ensure there's only one bucket per experiment and then you can just say

```py
experiment.buckets.exists()
experiment.buckets.get()
```

maybe we should just do that later.  Anyways this is fine for now.

app/experimenter/experiments/tests/api/v4/test_serializers.py
```diff
@@ -88,3 +92,87 @@ def test_serializer_outputs_expected_schema(self):
...
+            experiment=experiment, slug="control", is_control=True
+        )
+        ExperimentVariantFactory.create(experiment=experiment, slug="variant-2")
+
+        ExperimentBucketNamespace.request_namespace_buckets(
+            experiment.normandy_slug,
+            experiment,
+            NIMBUS_DATA["ExperimentDesignPresets"]["empty_aa"]["preset"]["arguments"][
+                "bucketConfig"
+            ]["count"],
```
jaredlockhart, line 122: This doesn't need to refer to the presets in the test, you can just hard code a value here.

jaredlockhart, line 122: Actually this doesn't even need to call this here, all teh inputs to this test can just be hard coded, the place where we'd want to test this is on the task, below.

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -32,6 +33,10 @@ def test_push_experiment_to_kinto_sends_experiment_data(self):
             if_not_exists=True,
         )
 
+        self.assertTrue(
+            ExperimentBucketRange.objects.filter(experiment=self.experiment).exists()
+        )
+
```
jaredlockhart, line 39: I think here is where we'd want to check that the buckets are generated using hte preset values.

app/experimenter/kinto/tasks.py
```diff
@@ -4,24 +4,38 @@
...
 from experimenter.experiments.changelog_utils import update_experiment_with_change_log
-from experimenter.experiments.models import Experiment
+from experimenter.experiments.models import Experiment, ExperimentBucketNamespace
 from experimenter.kinto import client
 
 
 logger = get_task_logger(__name__)
 metrics = markus.get_metrics("kinto.tasks")
 
+BUCKET_COUNT = 100
```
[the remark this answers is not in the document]
tiftran: removed

app/experimenter/experiments/api/v4/serializers.py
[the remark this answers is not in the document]
tiftran: will do as a follow up

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -25,6 +30,18 @@ def test_push_experiment_to_kinto_sends_experiment_data(self):
...
+        bucketConfig = data["arguments"]["bucketConfig"]["count"]
+
+        designPreset = NIMBUS_DATA["ExperimentDesignPresets"]["empty_aa"]["preset"][
+            "arguments"
+        ]["bucketConfig"]["count"]
+
+        self.assertEqual(bucketConfig, designPreset)
```
tiftran, line 43: this seems kinda weird to be only checking count.... but total, namespace, and start aren't part of the designPresets. and the randomization differs where it's "userId" in the designPreset but experimenter defaults it to "normandy_id". Not sure whether this is intentional

app/experimenter/experiments/tests/api/v4/test_serializers.py
```diff
@@ -5,13 +5,17 @@
...
+from mozilla_nimbus_shared import get_data
+
+from experimenter.experiments.models import Experiment, ExperimentBucketNamespace
 from experimenter.experiments.tests.factories import (
     ExperimentFactory,
     ExperimentVariantFactory,
 )
 from experimenter.experiments.api.v4.serializers import ExperimentRapidRecipeSerializer
 
+NIMBUS_DATA = get_data()
```
jaredlockhart, line 17: No longer used in this file?

app/experimenter/kinto/tests/test_tasks.py
jaredlockhart: OH snap! yes it should be what's in the design preset!  

https://github.com/mozilla/nimbus-shared/commit/336d51620c1ac6d44417f7c4a3dff83d34f6e706

So we gotta update the serializer to use that, and check it here in the test.  Good catch.

tiftran:
So i changed the randomizationUnit that made the schema validation angry, so i switched over to the nimbus _shared validation, which resulted in adding some missing fields too

jaredlockhart:
Yeah right, it's more important to get this right than to fight with the schema validation.  Don't add the new validation, and take the old validation out, and proceed with the changes I suggested.  We can sort out the schema validation after.

app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -71,6 +96,16 @@ def get_startDate(self, obj):
...
 
+    def get_proposedEnrollment(self, obj):
+        return NIMBUS_DATA["ExperimentDesignPresets"]["empty_aa"]["preset"]["arguments"][
+            "proposedEnrollment"
+        ]
+
+    def get_proposedDuration(self, obj):
+        return NIMBUS_DATA["ExperimentDesignPresets"]["empty_aa"]["preset"]["arguments"][
+            "proposedDuration"
+        ]
```
jaredlockhart, line 107: No no we already did this when we create it:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v3/serializers.py#L104-L105

you don't need to do taht here just serialize what's in the db

app/experimenter/experiments/tests/api/v4/experimentRecipe.json
```diff
@@ -58,7 +58,7 @@
           "properties": {
             "randomizationUnit": {
               "type": "string",
-              "enum": ["client_id", "normandy_id"],
+              "enum": ["client_id", "normandy_id", "user_id"],
               "description": "The randomization unit. Note that client_id is not yet implemented.",
               "default": "normandy_id"
             },
```
jaredlockhart, line 64: Just delete this static json file it's already way out of date and remove the checks that use it in the tests

jaredlockhart:
Yeah don't blow this up anymore than it needs to we have a separate ticket filed to make the shared schema validator happy:  https://github.com/mozilla/experimenter/issues/3058

Just take the old schema validator out and focus on the bucket stuff.

app/experimenter/experiments/tests/api/v4/test_serializers.py
```diff
@@ -88,3 +79,78 @@ def test_serializer_outputs_expected_schema(self):
...
+                    "total": bucket.namespace.total,
+                },
+            },
+        )
+        converted_branches = [dict(branch) for branch in branches]
+        self.assertEqual(
+            converted_branches,
+            [
+                {"ratio": 33, "slug": "variant-2", "value": None},
+                {"ratio": 33, "slug": "control", "value": None},
```
pdehaan, line 154: Random Q: If there are only two branches, why are the ratios 33/33 and not 50/50?

jaredlockhart: It's just a dummy test value, doesn't really matter what the value is:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/factories.py#L241

PR #3094, opened 2020-07-23
Update detail page on request approval fixes #3004

### LINKED ISSUE #3004: Update detail page after Review Request button is clicked
After a user clicks the Review Request button, we need to update the experiment state and then update the UI to reflect the current state:

- Review Request button is disabled
- Back button is disabled (it occurs to me that we have it in the grey state now even when it's enabled so we should probably change that to be more consistent: blue enabled, grey disabled)
- `Draft` now shows `In Review`

jaredlockhart:
I tried to figure out how to use jest-fetch-mock to mock out the second call to get the experimentData and assert that it's called but I couldn't figure it out and I'd rather get the UI landed this week and deal with it later.

### DIFF
--- a/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
+++ b/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
@@ -70,6 +70,11 @@ const ExperimentDetails: React.FC = () => {
   }
 
   const buttonsDisabled = experimentData.status !== ExperimentStatus.DRAFT;
+  let buttonsClass = "btn btn-primary";
+  if (buttonsDisabled) {
+    buttonsClass = "btn btn-secondary";
+  }
+
   const buttonsShown = ![
     ExperimentStatus.LIVE,
     ExperimentStatus.COMPLETE,
@@ -77,46 +82,29 @@ const ExperimentDetails: React.FC = () => {
 
   let changeStatusButtons;
   if (buttonsShown) {
-    if (buttonsDisabled) {
-      changeStatusButtons = (
-        <div className="d-flex mt-4">
-          <span>
-            <button disabled className="btn btn-secondary">
-              Back
-            </button>
-          </span>
+    changeStatusButtons = (
+      <div className="d-flex mt-4">
+        <span>
+          <Link
+            className={buttonsClass}
+            to={buttonsDisabled ? "#" : `/${experimentData.slug}/edit/`}
+          >
+            Back
+          </Link>
+        </span>
 
-          <span className="flex-grow-1 text-right">
-            <button disabled className="btn btn-secondary" type="button">
-              Request Approval
-            </button>
-          </span>
-        </div>
-      );
-    } else {
-      changeStatusButtons = (
-        <div className="d-flex mt-4">
-          <span>
-            <Link
-              className="btn btn-primary"
-              to={`/${experimentData.slug}/edit/`}
-            >
-              Back
-            </Link>
-          </span>
-
-          <span className="flex-grow-1 text-right">
-            <button
-              className="btn btn-primary"
-              type="button"
-              onClick={handleClickRequestApproval}
-            >
-              Request Approval
-            </button>
-          </span>
-        </div>
-      );
-    }
+        <span className="flex-grow-1 text-right">
+          <button
+            className={buttonsClass}
+            disabled={buttonsDisabled}
+            type="button"
+            onClick={handleClickRequestApproval}
+          >
+            Request Approval
+          </button>
+        </span>
+      </div>
+    );
   }
 
   return (
--- a/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
+++ b/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
@@ -6,7 +6,11 @@ import {
   audienceOptions,
   firefoxVersionOptions,
 } from "experimenter-rapid/components/forms/ExperimentFormOptions";
-import { useExperimentState } from "experimenter-rapid/contexts/experiment/hooks";
+import { requestReview } from "experimenter-rapid/contexts/experiment/actions";
+import {
+  useExperimentState,
+  useExperimentDispatch,
+} from "experimenter-rapid/contexts/experiment/hooks";
 import { ExperimentStatus } from "experimenter-rapid/types/experiment";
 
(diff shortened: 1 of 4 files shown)


app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -68,31 +69,54 @@ const ExperimentDetails: React.FC = () => {
...
-        <span>
-          <Link
-            className="btn btn-secondary"
-            to={`/${experimentData.slug}/edit/`}
-          >
-            Back
-          </Link>
-        </span>
+  if (buttonsShown) {
+    if (buttonsDisabled) {
```
tiftran, line 80: could we reduce this is else for button disabled to:
```
      <div className="d-flex mt-4">
        <span>
          <Link to={`/${experimentData.slug}/edit/`}>
            <button className="btn btn-primary" disabled={buttonsDisabled}>
              Back
            </button>
          </Link>
        </span>

        <span className="flex-grow-1 text-right">
          <button
            disabled={buttonsDisabled}
            className="btn btn-primary"
            type="button"
            onClick={handleClickRequestApproval}
          >
            Request Approval
          </button>
        </span>
      </div>
```
?

The disable buttons would be blue though?

pdehaan: We could use the [**classnames**](https://www.npmjs.com/package/classnames) npm module to conditionally set the `className` values to toggle between "btn-primary" and "btn-secondary".

app/experimenter/static/rapid/contexts/experiment/actions.ts
```diff
@@ -50,3 +50,15 @@ export const updateExperiment = (name: string, value: string | string[]) => (
...
 };
+
+export const requestReview = () => async (
+  experimentData: ExperimentData,
+  dispatch: React.Dispatch<ExperimentReducerAction>,
+): Promise<void> => {
+  await fetch(`/api/v3/experiments/${experimentData.slug}/request_review/`, {
+    method: "POST",
+  });
+  if (experimentData.slug) {
```
pdehaan, line 61: Under what circumstances would an experiment _not_ have a slug?
And if there was no slug, what'd be the side effects of the `fetch()` call above on L58-59 which uses `${experimentData.slug}/request_review/` to construct a beautiful request review URL?

jaredlockhart: Slug is `null` before the experiment is created (ie form is filled out but isn't saved yet), so it's optional on the data type:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/static/rapid/types/experiment.ts#L18

so TS enforces checking for it

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: Lol you'd think!  That only works in `react-bootstrap`, the Link component we're using doesn't have it:

```
Type '{ children: string; disabled: boolean; className: string; to: string; }' is not assignable to type 'IntrinsicAttributes & LinkProps<PoorMansUnknown> & RefAttributes<HTMLAnchorElement>'.
  Property 'disabled' does not exist on type 'IntrinsicAttributes & LinkProps<PoorMansUnknown> & RefAttributes<HTMLAnchorElement>'.ts(2322)
```

but actually I just found another hack:

```js
            to={buttonsDisabled ? "#" : `/${experimentData.slug}/edit/`}
```

it still shows the hand cursor as if it's clickable, stackoverflow says to disable the pointer using custom CSS but meh.  We should just add react bootstrap that has the correct attributes.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -68,23 +69,34 @@ const ExperimentDetails: React.FC = () => {
...
+            to={buttonsDisabled ? "#" : `/${experimentData.slug}/edit/`}
           >
             Back
           </Link>
         </span>
 
         <span className="flex-grow-1 text-right">
           <button
-            className="btn btn-primary"
+            className={buttonsClass}
```
pdehaan, line 98: Ah, the interesting side effect here is that I _believe_ that the `buttonsClass` is shared for both the "Back" and "Request Approval" buttons (whereas previously the Back was btn-secondary and the "Request Approval" is btn-primary).

Not a deal/merge blocker, but might need a bit of fine tuning in the future.

jaredlockhart: Yeah I talked to Ana about this about making it consistent, ie `blue` is enabled and `grey` is disabled, but I don't think it made its way back into the wireframes yet.

@AnaMedinac

app/experimenter/static/rapid/contexts/experiment/actions.ts
pdehaan: But if `slug` is `null`, wouldn't that make the `await fetch()` call try and set the request_review URL to `/api/v3/experiments//request_review/` (depending on if `null` gets stringified to an empty string or the literal string "null" -- which now I'll have to try verifying out of pure, childlike curioisity).

jaredlockhart: Oh good point we should wrap the whole thing!

jaredlockhart: Or actually, by the time we land on the detail page we expect there to be a slug, so maybe we should just remove the check altogether and let it error out and report since otherwise it's just a totally broken behaviour ie clicking the button and nothing happenign and no error being reported

jaredlockhart: Plus they wouldn't be able to land on the page without a slug

jaredlockhart: Oh no I'm wrong we do have to check it, okay i'll just wrap the whole thing, TS doesn't typecheck the string format it seems.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
pdehaan: Sure, sure, sure... naming issues and UX-crimes aside, that'd work.
Let's merge and unblock this and we can leave it to our future selves to noodle with this.

PR #3096, opened 2020-07-23
Use rapid preset data fixes #2890

### LINKED ISSUE #2890: Create A/A Experiment with Design Template Values

jaredlockhart:
`empty_aa` is hard coded for now, in MVP2 when there's more than one preset there will be UI for picking one.

### DIFF
--- a/app/experimenter/experiments/api/v3/serializers.py
+++ b/app/experimenter/experiments/api/v3/serializers.py
@@ -8,6 +8,9 @@ from experimenter.experiments.models import Experiment, ExperimentVariant
 from experimenter.experiments.changelog_utils import ChangelogSerializerMixin
 
 
+NIMBUS_DATA = get_data()
+
+
 class ExperimentRapidChangelogSerializerMixin(ChangelogSerializerMixin):
     def create(self, validated_data):
         experiment = super().create(validated_data)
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -87,25 +89,30 @@ def validate(self, data):
...
-            ratio=1,
-            is_control=True,
-        )
-        ExperimentVariant.objects.create(
-            experiment=experiment, name="treatment", slug="treatment", ratio=1,
-        )
+        for branch_data in preset_data["branches"]:
+            ExperimentVariant.objects.create(
+                experiment=experiment,
+                name=branch_data["slug"],
```
pdehaan, line 111: Q: should this be `branch_data["slug"]` or `"name"`?
I believe this is correct, per https://github.com/mozilla/nimbus-shared/blob/ecf090a942e29d76ce73a0d84d542cc91edf2398/data/ExperimentDesignPresets.ts#L31-L34 (if I'm looking at the correct lines); but just wanted to confirm.

jaredlockhart: Yep I filed: https://github.com/mozilla/nimbus-shared/issues/95

PR #3098, opened 2020-07-23
Add monitoring link to rapid detail page fixes #3090

### LINKED ISSUE #3090: Add Grafana live monitoring link to live rapid experiment
Becuase the monitoring and analysis dashboards are separate for the MVP we need to add a grafana monitoring link to teh rapid detail page.  We should just be able to reuse the same method we use to generate it now, so I think it's as simple as adding it to the API and adding a link.

@AnaMedinac FYI since it's not reflected in the wireframes

jaredlockhart:


app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -68,6 +68,26 @@ const ExperimentDetails: React.FC = () => {
...
+      <>
+        <h3 className="my-4">Monitoring</h3>
+        <p>
+          The live monitoring dashboard can be found here:
+          <a
+            href={experimentData.monitoring_dashboard_url}
+            rel="noopener noreferrer"
+            target="_blank"
+          >
+            {" "}
```
pdehaan, line 83: nit: do we want the spacer _outside_ of the linked text? Like, maybe at end of L77 above?

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -68,6 +68,26 @@ const ExperimentDetails: React.FC = () => {
...
+            (link here)
```
pdehaan, line 84: Are we replacing the `(link here)` with the actual link URL eventually?
This currently reads a bit awkward:

> ### Monitoring
> The live monitoring dashboard can be found here: (link here)

Or if we change it to 

> The live monitoring dashboard can be found &lt;here&gt;"

Or even better is possibly just

> Visit &lt;live monitoring dashboard&gt;.

I'll check the Figma, but wasn't sure if the "link here" was a placeholder for the actual URL in some of the specs.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -139,6 +161,7 @@ const ExperimentDetails: React.FC = () => {
         <p>
           The results can be found here: <a href="#">(link here)</a>
```
pdehaan, line 162: Tangentially related to the above "(link here)" debate, but we might want to make this link text/format consistent (either in this PR or I can file a new issue once this gets merged).

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: Eslint fix adds this automatically, if I take it out it'll just put it back.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: I'm not sure, there is no monitoring link in the figma at all so I think we'll have to get @AnaMedinac to update the mocks and then we can PR the relevant changes in.

PR #3105, opened 2020-07-24
task to update experiments to live

tiftran:


.env.sample
```diff
@@ -21,6 +21,7 @@ EMAIL_SHIP=
 EXPERIMENTS_PAGINATE_BY=50
 HOSTNAME=localhost
 KINTO_BUCKET=main-workspace
+KINTO_BUCKET_MAIN=main
```
jaredlockhart, line 24: Oh yeah so this really was how to do it eh, cool interesting, g/j making it work

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -102,3 +102,48 @@ def test_check_with_rapid_review_and_no_kinto_pending_pushes_experiment(self):
         self.assertEqual(experiment.firefox_min_version, Experiment.VERSION_CHOICES[0][0])
         self.assertEqual(experiment.firefox_channel, Experiment.CHANNEL_RELEASE)
         self.assertEqual(experiment.normandy_slug, "bug-12345-rapid-test-release-55")
+
+
+class TestCheckExperimentIsLive(MockKintoClientMixin, TestCase):
+    def test_experiment_updates_when_recipe_is_in_main(self):
```
jaredlockhart, line 108: We coulda stuck a draft in here just to make sure it DOESN'T get included in the check, but that's just a nice to have, not a blocker

app/experimenter/kinto/tasks.py
```diff
@@ -67,3 +67,35 @@ def check_kinto_push_queue():
...
+
+    for experiment in accepted_experiments:
+        if experiment.normandy_slug in record_ids:
+            logger.info(
+                "{experiment} status is being updated to live".format(
+                    experiment=experiment
+                )
+            )
+            experiment.status = Experiment.STATUS_LIVE
+            experiment.save()
```
jaredlockhart, line 92: Oh you don't have to do this here at all, the `update_experiment_with_change_log` also updates the instance:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/changelog_utils.py#L145

app/experimenter/kinto/tests/test_tasks.py
jaredlockhart: Alright let's just add it

PR #3107, opened 2020-07-25
Expose end_date in V4 API fixes #3103

### LINKED ISSUE #3103: End Date: null in rapid recipe
Just saw this on stage, end date should be computed since we have all that information now:

```json
{
  "enabled": true,
  "arguments": {
    "slug": "bug-1629019-rapid-lets-do-this-release-80",
    "active": true,
    "endDate": null,
    "branches": [
      {
        "slug": "treatment",
        "ratio": 1,
        "value": null
      },
      {
        "slug": "control",
        "ratio": 1,
        "value": null
      }
    ],
    "features": [
      "picture_in_picture",
      "pinned_tabs"
    ],
    "startDate": "2020-07-24",
    "bucketConfig": {
      "count": 0,
      "start": 0,
      "total": 10000,
      "namespace": "",
      "randomizationUnit": "normandy_id"
    },
    "userFacingName": "Let's do this",
    "referenceBranch": "control",
    "isEnrollmentPaused": false,
    "proposedEnrollment": 7,
    "userFacingDescription": " This is an empty CFR A/A experiment. The A/A experiment is being run to test the automation, effectiveness, and accuracy of the rapid experiments platform.\n    The experiment is an internal test, and Firefox users will not see any noticeable change and there will be no user impact."
  },
  "targeting": "localeLanguageCode == 'en' && (isFirstStartup || currentExperiment.slug in activeExperiments) && browserSettings.update.channel == 'release'",
  "filter_expression": "env.version|versionCompare('80.0') >= 0",
  "id": "bug-1629019-rapid-lets-do-this-release-80",
  "last_modified": 1595551593204
}
```

jaredlockhart:
This was much finickier than it should have been >_<

Also I based this on https://github.com/mozilla/experimenter/pull/3091 so we gotta land that first

### DIFF
--- a/app/experimenter/experiments/api/v4/serializers.py
+++ b/app/experimenter/experiments/api/v4/serializers.py
@@ -5,6 +5,7 @@ from mozilla_nimbus_shared import get_data
 from experimenter.experiments.models import (
     Experiment,
     ExperimentVariant,
+    ExperimentBucketRange,
 )
 
 NIMBUS_DATA = get_data()
@@ -20,6 +21,22 @@ class ExperimentRapidBranchesSerializer(serializers.ModelSerializer):
     def get_value(self, obj):
         # placeholder value
         return None
+
+
+class ExperimentBucketRangeSerializer(serializers.ModelSerializer):
+    namespace = serializers.SlugRelatedField(read_only=True, slug_field="name")
+    randomizationUnit = serializers.SerializerMethodField()
+    total = serializers.SerializerMethodField()
+
+    class Meta:
+        model = ExperimentBucketRange
+        fields = ("randomizationUnit", "namespace", "start", "count", "total")
+
+    def get_randomizationUnit(self, obj):
+        return obj.namespace.randomization_unit
+
+    def get_total(self, obj):
+        return obj.namespace.total
 
 
 class ExperimentRapidArgumentSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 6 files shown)


app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -53,24 +70,14 @@ class Meta:
         )
 
     def get_bucketConfig(self, obj):
-        return {
-            "randomizationUnit": "normandy_id",
-            "namespace": "",
-            "start": 0,
-            "count": 0,
-            "total": 10000,
-        }
```
jaredlockhart, line 62: Also I got rid of this because we shouldn't really emit non sensical invalid bucket range definition before the real one is created.

app/experimenter/experiments/tests/api/v4/test_serializers.py
```diff
@@ -36,10 +41,81 @@ def test_serializer_outputs_expected_schema(self):
         ExperimentVariantFactory.create(experiment=experiment, slug="variant-2")
 
         serializer = ExperimentRapidRecipeSerializer(experiment)
-        data = serializer.data
+        data = serializer.data.copy()
 
-        fn = os.path.join(os.path.dirname(__file__), "experimentRecipe.json")
```
jaredlockhart, line 41: And I got rid of the schema validation in the DRAFT test because a bunch of fields aren't present before the experiment launches and that doesn't match the schema, so either we:

- update the schema to allow for those fields to be `null`able before launch
- don't emit a rapid recipe at all for draft experiments
- emit a rapid recipe for draft experiments with those fields set to `null` and then just say it doesn't need to validate against the schema, which is the option I went for here, and the fully completed recipe for the review state can be validated against the schema

Since I think we'll be moving all of that into the Experimenter repo and out of the shared repo, we can re evaluate how we handle this after that happens.

jaredlockhart:
Actually crap I forgot we decided to change how this works altogether, closing this in favour of:

https://github.com/mozilla/experimenter/issues/3108

PR #3136, opened 2020-07-28
Set APPEND_SLASH to False fixes #3125

### LINKED ISSUE #3125: Set APPEND_SLASH to False
We keep running into situations where clients hit endpoints without trailing slashes and get redirected through the auth boundary, so we should just set this to False in settings.py

https://docs.djangoproject.com/en/3.0/ref/settings/#append-slash

`settings.py`:
```py
APPEND_SLASH = False
```

bhushan-borole:
Fixed #3125


bhushan-borole:
Do I have to append slashed manually in the `urls.py` file?

jaredlockhart:
@bhushan-borole Sorry I didn't get a chance to look at this yet, I ran it locally and got the same failures.  Either we can update the tests or this just might not work, I'll look more closely tomorrow!

bhushan-borole:
No worries @jaredlockhart.

jaredlockhart:
@bhushan-borole I'll be away next week so we'll have to pick this up when I get back 🌞

bhushan-borole:
It's fine @jaredlockhart.
We'll work on it after you are back!

jaredlockhart:
@bhushan-borole Thank you for giving this a go! I don't think this will be workable for now so we'll have to close this and possibly revisit in the future.

bhushan-borole:
No problem @jaredlockhart

PR #3144, opened 2020-07-28
Add firefox channel drop down to rapid form fixes #3138

### LINKED ISSUE #3138: Need to be able to choose channel
After doing some end-to-end testing, I realized it's pretty hard to test (and verify anything in the system) without the ability to run A/As on different channels (i.e. release, nightly, beta).

I think in order to get this working, we could either:
1. Add some UI to choose the channel
2. Just remove it from targeting right now
3. Infer the channel based on the selected minimum version

I'm inclined to go with 2 for now since it's the simplest, and we can improve from there?

jaredlockhart:
- Update to nimbus shared 0.0.9
- Add serializer/form field for `firefox_channel`
- Update recipe to format `firefox_channel` into `targeting`

Screenshot_2020-07-28 Mozilla Experimenter
Screenshot_2020-07-28 Mozilla Experimenter(1)

```json
{
  "id": "bug-1526514-rapid-asdf-release-77",
  "arguments": {
    "slug": "bug-1526514-rapid-asdf-release-77",
    "userFacingName": "asdf",
    "userFacingDescription": " This is an empty CFR A/A experiment. The A/A experiment is being run to test the automation, effectiveness, and accuracy of the rapid experiments platform.\n    The experiment is an internal test, and Firefox users will not see any noticeable change and there will be no user impact.",
    "active": true,
    "isEnrollmentPaused": false,
    "features": [
      "picture_in_picture"
    ],
    "proposedEnrollment": 7,
    "bucketConfig": {
      "randomizationUnit": "userId",
      "namespace": "bug-1526514-rapid-asdf-release-77",
      "start": 0,
      "count": 100,
      "total": 10000
    },
    "startDate": "2020-07-28",
    "endDate": null,
    "branches": [
      {
        "slug": "treatment",
        "ratio": 1,
        "value": null
      },
      {
        "slug": "control",
        "ratio": 1,
        "value": null
      }
    ],
    "referenceBranch": "control"
  },
  "filter_expression": "env.version|versionCompare('77.0') >= 0",
  "enabled": true,
  "targeting": "[userId, \"bug-1526514-rapid-asdf-release-77\"]|bucketSample(0, 100, 10000) && localeLanguageCode == 'en' && (isFirstStartup || 'bug-1526514-rapid-asdf-release-77' in activeExperiments) && browserSettings.update.channel == 'release'",
  "last_modified": 1595968370079
}
```


pdehaan:
What happens if I choose "Firefox Minimum Version: Firefox 99.0" and "Firefox Channel: Firefox Beta"?
I'm semi-afraid that we now have an easy way that this can get out of sync with some invalid data.

What if I say "Firefox 80.0" and "Release", would it delay the experiment until 80 is the release train and then start enrollment? Would it be an invalid combo because neither of those conditions would be true within the 7 day enrollment period and therefore no data would be collected and the experiment would be a giant noop.

Curious if the Version and Channel should be combined into a single XSelect (similar to the values from https://whattrainisitnow.com/), or if that's trying to control too many variables with a single value.

pdehaan:
Re: `"userFacingDescription": " This is an empty CFR A/A experiment. ..."`

Curious that the description has a leading space, and is this a "CFR A/A" or should that just say "A/A" (minus "CFR" reference)

jaredlockhart:
> What happens if I choose "Firefox Minimum Version: Firefox 99.0" and "Firefox Channel: Firefox Beta"?
> I'm semi-afraid that we now have an easy way that this can get out of sync with some invalid data.
> 
> What if I say "Firefox 80.0" and "Release", would it delay the experiment until 80 is the release train and then start enrollment? Would it be an invalid combo because neither of those conditions would be true within the 7 day enrollment period and therefore no data would be collected and the experiment would be a giant noop.
> 
> Curious if the Version and Channel should be combined into a single XSelect (similar to the values from https://whattrainisitnow.com/), or if that's trying to control too many variables with a single value.

@pdehaan Yeah so far we've avoided integrating any strong knowledge of the release schedule.  All those combinations would be 'valid' in so much as they're valid recipes they just wouldn't target on anything so after you launched you'd get some kind of indicator that you weren't enrolling anybody.  In general the pattern we see is someone creates an experiment with some version, then a long time goes by of that experiment in draft/review, and by the time it launches the version is too low.  I think because rapid experiments won't have that long draft/review time this will not be an issue, but we'll just have to wait and see when people start using it.

app/experimenter/static/rapid/types/experiment.ts
```diff
@@ -7,11 +7,18 @@ export enum ExperimentStatus {
   REJECTED = "Rejected",
 }
 
+export enum FirefoxChannel {
+  NIGHTLY = "Nightly",
+  BETA = "Beta",
+  RELEASE = "Release",
```
pdehaan, line 13: Not sure if I'm understanding all the data flows and transforms correctly or if these are even related, but I vaguely recalled from https://github.com/mozilla/nimbus-shared/pull/99/files that the `browserSettings.update.channel == 'release'` and `normalized_channel = 'release'` channel names were both lowercased.
Not sure if that's in any way relevant or important, but figured worth calling out and highlighting my vast ignorance.

jaredlockhart: Yeah, so the frontend contract is with the Experimenter V3 API, which defines the choice options as

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants.py#L194-L204

The contract between Experimenter and Firefox is encoded in the V4 API, which lower cases them here: 

https://github.com/mozilla/experimenter/pull/3144/files#diff-1146634a352f06a65bcaff252215820cR128

But an astute observation!

app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -123,7 +123,9 @@ def get_targeting(self, obj):
             randomization_unit = bucket_config["randomizationUnit"]
             bucket_count = bucket_config["count"]
             bucket_total = bucket_config["total"]
-            audience_targeting = NIMBUS_DATA["Audiences"][obj.audience]["targeting"]
+            audience_targeting = NIMBUS_DATA["Audiences"][obj.audience][
+                "targeting"
+            ].format(slug=obj.normandy_slug, firefox_channel=obj.firefox_channel.lower(),)
```
k88hudson, line 128: :+1:

PR #3154, opened 2020-07-30
Added Rejected workflow fixes #3003

tiftran:
Screen Shot 2020-07-29 at 11 15 55 PM

PR #3155, opened 2020-07-30
Added Rejected workflow fixes #3003

### LINKED ISSUE #3003: Rapid detail page state for rejected experiments

jaredlockhart:
Taking over for @tiftran from https://github.com/mozilla/experimenter/pull/3154


app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -131,34 +131,35 @@ const ExperimentDetails: React.FC = () => {
...
     changeStatusButtons = (
       <div className="d-flex mt-4">
         <span>
-          <Link
-            className={buttonsClass}
-            to={buttonsDisabled ? "#" : `/${experimentData.slug}/edit/`}
-          >
-            Back
+          <Link to={`/${experimentData.slug}/edit/`}>
+            <button className="btn btn-primary" disabled={backButtonDisabled}>
```
k88hudson, line 153: Why did you change this to a button?

jaredlockhart: Oh I didn't notice she did that, that's clever!  This is so it gets properly disabled rather than 'fake' disabled which was driving @pdehaan nuts

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -37,7 +38,10 @@ const ErrorList: React.FC<ErrorListProperties> = ({ errors }) => {
 };
 
 const ExperimentForm: React.FC = () => {
-  const formData = useExperimentState();
+  // eslint-disable-next-line
```
k88hudson, line 41: What are you disabling here?

jaredlockhart: Oh interesting, i'll check that out

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: Oh it doesn't actually solve the problem, it still shows the hand cursor.

jaredlockhart: Sigh we probably have to write some css for that.

k88hudson: i'll take a look when i run it

k88hudson: Ok, this is what I'd do:

```ts
<Link
  className={
    "btn btn-primary" + (backButtonDisabled ? " disabled" : "")
  }
  to={`/${experimentData.slug}/edit/`}
  onClick={(e) => backButtonDisabled && e.preventDefault()}
  >
  Back
</Link>
```

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
k88hudson: Ah, I see what's going on. I'd recommend changing this to

```ts
const formData = {...useExperimentState(), status: ExperimentStatus.DRAFT}
```

pdehaan: +1. Also, it's usually more helpful to just ignore/disable a specific ESLint rule on the next line, if possible (vs shotgun disabling everything). But of course now that I've said that, I'll have to do some <kbd>git grep</kbd> to check our current eslint disable usages or the curiosity will drive me crazy.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -169,6 +170,21 @@ const ExperimentDetails: React.FC = () => {
...
 
+  let rejectFeedback;
+
+  if (experimentData.reject_feedback) {
+    const messageDate = new Date(experimentData.reject_feedback.changed_on);
+    rejectFeedback = (
+      <>
+        <h3 className="my-4">Review Feedback</h3>
+        <div className="alert alert-secondary" role="alert">
+          <p className="font-weight-bold"> {messageDate.toDateString()}</p>
```
pdehaan, line 181: Why the space before the date string?

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -169,6 +170,21 @@ const ExperimentDetails: React.FC = () => {
...
+          <p>Reject reason: {experimentData.reject_feedback.message}</p>
```
pdehaan, line 182: Is the expectation that we'd always have a single line of feedback from Remote Settings?
Also, a bit extreme, but do we need to make sure this is sanitized at all since it's coming from Remote Settings and we aren't going to XSS ourselves or something? (Although I guess if somebody already has access to our Remote Settings bucket, having them return a reject reason with potential `<script>` tags is probably the least of our worries)

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
jaredlockhart: @k88hudson that doesn't work because then you're including `state` twice in the object and TS complains, once in the first and then explicitly in the second.

jaredlockhart: I can't see an easy way to do this so I'm just gonna leave it how @tiftran wrote it.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: Both fine points but seems to support multiple lines and properly escape characters:

Screenshot_2020-07-30 Mozilla Experimenter(6)

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: 🤷‍♀️

app/experimenter/kinto/tasks.py
```diff
@@ -25,13 +29,14 @@ def push_experiment_to_kinto(experiment_id):
...
 
     experiment = Experiment.objects.get(id=experiment_id)
-    ExperimentBucketNamespace.request_namespace_buckets(
-        experiment.normandy_slug,
-        experiment,
-        NIMBUS_DATA["ExperimentDesignPresets"]["empty_aa"]["preset"]["arguments"][
-            "bucketConfig"
-        ]["count"],
-    )
+    if not ExperimentBucketRange.objects.filter(experiment=experiment).exists():
```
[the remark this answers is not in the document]
jaredlockhart: Ya

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
jaredlockhart: Oh it's htis:

```js
  const formData = {
    ...useExperimentState(),
    ...{ status: ExperimentStatus.DRAFT },
  };
```

k88hudson: Really? `const formData = {...useExperimentState(), status: ExperimentStatus.DRAFT}` works for me (they should be equivalent?)

PR #3195, opened 2020-08-07
Make Nimbus version comparisons work on all channels fixes #3194

### LINKED ISSUE #3194: Nimbus version comparisons in filter expressions don't work for non-Release
During the initial launch test, we learned that the `versionCompare` filter treats versions like "80.0a1" (as used in Nightly) as less than "80.0" (as used in Release). Because of this we couldn't effectively target Nightly versions. We found a workaround, but it's something we should fix. When targeting version X and a non-Release channel, Experimenter should generate a filter expression that works.

mythmon:


jaredlockhart:
This makes sense to me but I'd like to get @k88hudson 's feedback before we land it

jaredlockhart:
Also in theory @mythmon couldn't this go in here:

https://github.com/mozilla/nimbus-shared/blob/main/data/ExperimentDesignPresets.ts#L25

so it says 

```
      filter_expression: "env.version|versionCompare('{minFirefoxVersion}.!') >= 0",
```

and just have experimenter inject the major version e.g. `80`?

app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -106,9 +106,31 @@ class Meta:
...
+        # it (like 80.0).
+        #
+        # To account for this, we replace the ".0" in "80.0" with "!", which is
+        # the character that has the smallest ASCII value of all printable,
+        # non-whitespace characters. This will ensure that it is less than all
+        # 80-series version numbers.
+        #
+        # It isn't used now, but if a max version comparison is needed in the
+        # future the equivalent modification would be "80.*". This is because
+        # "*" is treated specially as the largest possible version in it's
```
k88hudson, line 121: its

app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -106,9 +106,31 @@ class Meta:
...
+        # slot. That would allow anything that is in the 80-series, and nothing
+        # in the 81-series.
+        #
+        # To summarize, 80.! < 80.0a1 < 80.0 < 80.1 < 80.* < 81.0
+
+        firefox_min_version = obj.firefox_min_version
+        assert firefox_min_version.endswith(".0")
+        firefox_min_version = firefox_min_version[:-1] + "!"
```
jaredlockhart, line 129: Oh you can just use this actually:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models.py#L913-L916

To get the major version and then just format it into the template

Also we haven't been in the habit of throwing asserts in the middle of application logic in general, I would avoid that pattern here.

app/experimenter/experiments/api/v4/serializers.py
```diff
@@ -106,9 +106,31 @@ class Meta:
...
+
         return NIMBUS_DATA["ExperimentDesignPresets"]["empty_aa"]["preset"][
             "filter_expression"
-        ].format(minFirefoxVersion=obj.firefox_min_version)
+        ].format(minFirefoxVersion=firefox_min_version)
```
jaredlockhart, line 133:
```py
         ].format(minFirefoxVersion=f"{obj.firefox_min_version_integer}.!")
```

mythmon:
> I'd like to see a couple more tests

I can't think of any more tests that could be added in this repo, without using very heavy integration tests. If you have something specific in mind I'd like to hear it.

k88hudson:
> > I'd like to see a couple more tests
> 
> I can't think of any more tests that could be added in this repo, without using very heavy integration tests. If you have something specific in mind I'd like to hear it.

Oh yeah I guess the comparison code is all in Firefox – that makes sense. Let's stick with some verification by @pdehaan then

mythmon:
@jaredlockhart why did you merge master into this PR? Was there something on master that we needed here?

jaredlockhart:
@mythmon I have the 'require prs to be up to date with master' thing enabled in github for all prs

PR #3202, opened 2020-08-10
v4 api to use normandy slug/ available after status accepted fixes #3131

### LINKED ISSUE #3131: V4 API uses Experimenter slug instead of Normandy Slug
Oops! We're using the Experimenter slug in the V4 API detail URL, but then we only expose the `normandy_slug` as the `id`.  To me this suggests we shouldn't expose Draft experiments to the V4 API whatsoever since it won't have the `normandy_slug` until it's pushed into Kinto.

tiftran:


app/experimenter/experiments/api/v4/views.py
```diff
@@ -7,10 +7,10 @@
 class ExperimentRapidViewSet(
     mixins.RetrieveModelMixin, mixins.ListModelMixin, viewsets.GenericViewSet,
 ):
-    lookup_field = "slug"
+    lookup_field = "normandy_slug"
     queryset = (
         Experiment.objects.get_prefetched()
         .filter(type=Experiment.TYPE_RAPID)
-        .exclude(status=Experiment.STATUS_DRAFT)
+        .exclude(status__in=[Experiment.STATUS_DRAFT, Experiment.STATUS_REVIEW])
```
jaredlockhart, line 14: wait what's this for

jaredlockhart: oh cuz it doesn't hav ea normandy slug

jaredlockhart: @tiftran Don't you have to update this? https://github.com/mozilla/experimenter/blob/bee394a241175880dc8611a506871e4b5e0f0caa/app/experimenter/experiments/tests/api/v4/test_views.py#L25-L26

tiftran: ooh right right

jaredlockhart: Oh! the tests pass because we're using the ExperimentFactory that just always generates a normandy slug!  D'oh!  God we need that rapid factory XD

tiftran: we do? maybe I'm missing something but seems like https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/factories.py#L158 is the only line that has a normandy slug generated, but we're calling create_with_variants, instead of create_with_status? was assuming that since the normandy slug is None, we're comparing sets of None?

jaredlockhart: OH! XD the test was never valid!  Goooooood eye @tiftran thnx for fixing ;)

PR #3206, opened 2020-08-10
normandy_slug -> recipe_slug fixes #3000

### LINKED ISSUE #3000: Rename `normandy_slug` to `recipe_slug`
Since we'll be using the `normandy_slug` field now for both normandy experiments and rapid experiments, we should use a more generic name, I think `recipe_slug` is the closest fit?

tiftran:


jaredlockhart:
@tiftran I just ran through a pref flip on my local and when i get to the recipe I see:

```json
{"action_name":"multi-preference-experiment","name":"asdf","filter_object":[{"type":"bucketSample","input":["normandy.userId"],"start":0,"count":1000,"total":10000},{"type":"channel","channels":["nightly"]},{"type":"version","versions":[55,56]}],"comment":"Prefs:\n\nExperiments:\n\nAny additional filters:\n","arguments":{"slug":null,"userFacingName":"asdf","userFacingDescription":"asdf","branches":[{"preferences":{"pref.1":{"preferenceBranchType":"default","preferenceType":"string","preferenceValue":"pls"},"pref.2":{"preferenceBranchType":"user","preferenceType":"string","preferenceValue":"{\"a\": 1}"}},"ratio":50,"slug":"treatment"},{"preferences":{"pref.1":{"preferenceBranchType":"default","preferenceType":"boolean","preferenceValue":true},"pref.2":{"preferenceBranchType":"user","preferenceType":"integer","preferenceValue":1}},"ratio":50,"slug":"control"}],"experimentDocumentUrl":"https://localhost/experiments/asdf/"},"experimenter_slug":"asdf"}
```

slug is `null`?

tiftran:
@jaredlockhart  tried to reproduce a couple times, was only able to do it, when i remove the bugzilla host from my .env file

https://github.com/mozilla/experimenter/pull/3206/files#diff-562e6046de58c57f7e5e7f6ecdb15efbR365-R385, seems to suggest that a slug won't be generated when there isn't a bugzilla id?

PR #3208, opened 2020-08-11
added react-bootstrap fixes #2942

### LINKED ISSUE #2942: Use React-Bootstrap for Rapid UI
We've used React-Bootstrap successfully for our old UI and it does a lot of things for us in terms of consistent design and layout, I'd like to use them for the Rapid UI as well.

tiftran:
Decided to leave the form/inputs alone since, it seems like we might be changing that....

### DIFF
--- a/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
+++ b/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
@@ -1,6 +1,6 @@
 import React from "react";
-import { Link } from "react-router-dom";
 import { Button } from "react-bootstrap";
+import { Link } from "react-router-dom";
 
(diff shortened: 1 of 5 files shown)


jaredlockhart:
@tiftran could you pls also do all the `col`/`row` divs into `Col`/`Row` 🙏

tiftran:
@jaredlockhart yup, for sure. Missed that.

jaredlockhart:
I guess this is blocked on that TS dependency issue with popper.  I'd like to be able to do this but I'm not sure what the solution is.  @rehandalal If you get some spare cycles could you see if you can find a quick solution?  If not then I guess we table this and reconsider it later.

jaredlockhart:
Let's revisit this in a few weeks when we can have a broader discussion about how we want to structure the frontend.

PR #3214, opened 2020-08-12
Added TabRoutes component (fixes #3211)

### LINKED ISSUE #3211: Implement sidebar tab UI with sub-routes for create view
For now, let's just do a simple tab UI. Clicking on `Branches` should trigger a different route (`/branches`)
image

k88hudson:
This patch adds a `TabRoutes` component that adds nested sub-routes for each tab. I added it to the experiment form page with some placeholder content for the branches tab for now (this should be replaced with UI for editing branches eventually).

#### Example usage:

```ts
<TabRoutes
  tabs={[
    {
      label: "Settings",
      path: "",
      component: SettingsForm,
    },
    {
      label: "Branches",
      path: "branches",
      component: BranchesForm,
    },
  ]}
/>
```

#### QA steps
1. Go to `experiments/rapid/new`. You should see the Settings tab by default.
2. Click the Branches tab. You should now see the Branches tab.
3. Load `experiments/rapid/new/branches` directly. You should see the Branches tab. Try creating an experiment.
4. Also check `experiments/rapid/{some experiment}/edit` and `experiments/rapid/{some experiment}/edit/branches`. Try editing an experiment and make sure your changes get saved.

image

### DIFF
--- a/app/experimenter/static/rapid/components/forms/TabRoutes.tsx
+++ b/app/experimenter/static/rapid/components/forms/TabRoutes.tsx
@@ -1,0 +1,60 @@
+import React from "react";
+import {
+  NavLink,
+  Route,
+  RouteProps,
+  Switch,
+  useRouteMatch,
+} from "react-router-dom";
+
+function pathJoin(...paths: Array<string>): string {
+  return paths.join("/").replace(/\/{1,}/g, "/");
+}
+
+interface TabProps {
+  tabs: Array<{
+    className?: string;
+    label: string;
+    path: string;
+    render?: RouteProps["render"];
+    component?: RouteProps["component"];
+  }>;
+}
+
+/**
+ * Creates horizontal tabs that will also trigger a route change.
+ * Paths are relative to the page on which TabRoutes is rendered, so
+ * the default route should be path=""
+ */
+export const TabRoutes: React.FC<TabProps> = ({ tabs }) => {
+  const { url } = useRouteMatch();
+  return (
+    <div>
+      <ul className="nav nav-tabs mb-4">
+        {tabs.map((tab) => (
+          <li key={tab.path} className="nav-item">
+            <NavLink
+              activeClassName="active"
+              className="nav-link"
+              exact={true}
+              to={pathJoin(url, tab.path)}
+            >
+              {tab.label}
+            </NavLink>
+          </li>
+        ))}
+      </ul>
+      <Switch>
+        {tabs.map((tab) => (
+          <Route
+            key={tab.path}
+            component={tab.component}
+            exact={true}
+            path={pathJoin(url, tab.path)}
+            render={tab.render}
+          />
+        ))}
+      </Switch>
+    </div>
+  );
+};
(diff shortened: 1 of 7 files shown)


jaredlockhart:
@k88hudson looks really good! I'm curious why we're removing the trailing slashes from all the URLs?  Django can be finicky about that and cause auth gate redirects if it detects a missing trailing slash and tries to automatically postpend and redirect it.

k88hudson:
> @k88hudson looks really good! I'm curious why we're removing the trailing slashes from all the URLs? Django can be finicky about that and cause auth gate redirects if it detects a missing trailing slash and tries to automatically postpend and redirect it.

Oh sorry about that, I was trying a couple of different strategies to get the sub-routes to work and I think I might have left some changes in, I'll put them back

jaredlockhart:
@k88hudson I don't see the tabs on the form anymore?  The pr introduces the branch component but doesn't add it to the UI? Is that intentional?

Screenshot_2020-08-13 Mozilla Experimenter

pdehaan:
Random Q, but what's the reasoning for moving this from a 1 page intake form to a "2 page" form?

k88hudson:
@pdehaan with the move to supporting a wider range of experiments other than A/A, it's helpful to break up into other sections. Influenced by Optimizely's UI (which has a 2-stage approach) and some suggestions from Ryan's [design ideation work](https://www.figma.com/file/lAP4IeohBNMPawhsUaB4nq/Nimbus?node-id=0%3A1)

jaredlockhart:
@k88hudson Looks like something got messed up with the way the state is managed, steps to reproduce:

1) Create Rapid Experiment
2) Fill out form and hit save
3) Hit back
4) Hit save

Results:

Screenshot_2020-08-13 Mozilla Experimenter(1)

k88hudson:
@jaredlockhart oops, good catch! Updated with a test

app/experimenter/static/rapid/components/forms/TabRoutes.tsx
```diff
@@ -0,0 +1,64 @@
...
+} from "react-router-dom";
+
+/**
+ * Helper function to join paths without duplicating slashes.
+ * Adds a trailing slash to be inline with Experimenter's style of URL
+ */
+export function pathJoinWithTrailingSlash(...paths: Array<string>): string {
+  return `${paths.join("/")}/`.replace(/\/{1,}/g, "/");
+}
+
```
jaredlockhart, line 17: It's too bad the browser doesn't expose a url joiner, I only find this:

https://www.npmjs.com/package/url-join

On the one hand adding a new dependency isn't great, but on the other there are a lot of nuances in URL paths that we might accidentally miss, so maybe adding a dependency isn't the worst?

k88hudson: I think I'd rather leave it as-is for now unless we run into issues / need to expand the scope, and which point replacing this function (or the whole thing) might be worth doing

PR #3218, opened 2020-08-13
added changelog to capture recipes when pushed to kinto fixes #2894

### LINKED ISSUE #2894: Store rapid experiment serialized recipe on a textfield
@mythmon made a good point that we should store the actual recipe that we sent to remote settings, because if we want to refer back to it much later and the serializers have changed it won't be clear what exact recipe was sent at the time the recipe happened.

tiftran:


app/experimenter/kinto/tasks.py
```diff
@@ -45,6 +47,19 @@ def push_experiment_to_kinto(experiment_id):
...
+        experimenter_kinto_user, _ = get_user_model().objects.get_or_create(
+            email=settings.KINTO_DEFAULT_CHANGELOG_USER,
+            username=settings.KINTO_DEFAULT_CHANGELOG_USER,
+        )
+        ExperimentChangeLog.objects.create(
+            experiment=experiment,
+            old_status=experiment.status,
+            new_status=experiment.status,
+            message="Recipe Sent to Kinto",
+            changed_values=data,
```
jaredlockhart, line 59: Yeah I think we can just wrap this in an outer object that makes it mirror the structure we use in the other ones so that later when we visualize it it's all consistent, something like

```py
{'recipe': {
  'new_value': data,
  'old_value': None,
  'display_name': 'Recipe',
}
```

PR #3225, opened 2020-08-13
Set startDate/endDate to None in recipe fixes #3108

### LINKED ISSUE #3108: Don't set `proposed_start_date` at rapid publish time
I forgot about this whole thing we discussed earlier.  If we set the `proposed_start_date` at publish time, then the dates that go into the recipe and into remote settings might not match the dates the Experimenter API publishes, so what we should do isntead is not set `proposed_start_date` at all during publish, startDate/endDate will both be `null` as the recipe goes into remote settings, and then after it gets marked as live in the db, those dates will become available, and so they will exist and be correct in API calls to Experimenter after the experiment is live.

jaredlockhart:
So basically to prevent the recipe dates from being wrong, we'll just set them to `None`, not use `proposed_start_date` at all, and then when the changelog for going live appears, the real startDate/endDate will be set and available in the V4 API.

### DIFF
--- a/app/experimenter/experiments/api/v4/serializers.py
+++ b/app/experimenter/experiments/api/v4/serializers.py
@@ -50,7 +50,7 @@ class ExperimentRapidArgumentSerializer(serializers.ModelSerializer):
     proposedEnrollment = serializers.ReadOnlyField(source="proposed_enrollment")
     bucketConfig = serializers.SerializerMethodField()
     startDate = serializers.SerializerMethodField()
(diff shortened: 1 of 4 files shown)


tiftran:
just to confirm, start date and end date will be an experimenter only field?

jaredlockhart:
@tiftran Well in the recipe they'll always be `None` and then in the V4 API they'll be `None` during `ACCEPTED` and then be set for `[LIVE, COMPLETE]`.  Is that what you mean?

PR #3226, opened 2020-08-14
Rename RAPID_AA_CFR to RAPID_AA 

### LINKED ISSUE #2952: Rename RAPID_AA_CFR to RAPID_AA
Originally the AA type was going to be a type of CFR experiment, but now I think that is not true so we should remove CFR from the constant label.

bhushan-borole:
Fixes #2952


bhushan-borole:
My pleasure @jaredlockhart !!

PR #3227, opened 2020-08-16
Objectives -> hypothesis plus help text changes fixes #3175

### LINKED ISSUE #3175: Rename “Objective” to "Hypothesis"
**Rename** "Objective" to "Hypothesis"

**new help text:**  Please fill in the hypothesis following the formula in the box.  There may be multiple hypothesis in one experiment.  
For roll-outs - please fill in follow the same formula - but substitute the second sentence with "we believe a rollout is needed to mitigate `describe the technical or perception risk`".

**new in box guidance (should not disappear when they start typing):**
If we `"do this/build this/create this" change in the experiment` for `these users`, then we will see `this outcome`.

We believe this because we have observed `this` via `data source, UR, survey`.

Optional - We believe this outcome will `describe impact` on `core metric`.

**note** - i wasn't sure how to indicate the text should be different, without messing up the ticket because my angle brackets were interpreted as mark-up.  I like how this looks in github - but please use whatever will look good in experimenter.  It may be sharp angle brackets around the info - if the inbox text doesn't like the mark-up.

**update** help text on "Internal Description"
**new help text**
Describe the details / purpose of your delivery in for internal consumers.  This assists supporting teams to understand what you are doing and helps avoid delays.  It is also a useful reference to understand the experiments that have been run.

Please add any bugs and/or issues related to this delivery work. Link to any PRDs, Invision, or related documents. Please include a description of for each link. 

**Hide** "Related Work URLs" 
With the mandatory addition of "Public Description" - "internal description" can now capture everything - including what we used to put in "related work URLs".  Removing this duplicate / possibly confusing field.

tiftran:
Screen Shot 2020-08-16 at 4 30 07 PM


app/experimenter/templates/experiments/section_overview.html
```diff
@@ -24,9 +24,6 @@
     <p>{{ experiment.engineering_owner }}</p>
   {% endif %}
 
-  <p><strong>Related Work</strong></p>
-  {{ experiment.related_work|urlize|linebreaks }}
-
   {% if experiment.related_to.all %}
```
jaredlockhart, line 27: Do we need to remove this from the form as well?

tiftran: right right

tiftran:
@jaredlockhart the link above doesn't seem to show me which line or file you're referring to?

jaredlockhart:
@tiftran Oh sorry stupid thing, it's here:

https://github.com/mozilla/experimenter/blob/bf16f053f42a96b3d0cf390cb3ce8beaac4a6ecc/app/experimenter/experiments/forms.py#L191

just the `"related_work"` field in the list of form fields

PR #3249, opened 2020-08-18
 updates to csv fields and their names fixes #3178 

### LINKED ISSUE #3178: Adding a new order/content of what is exported to CSV - called Stakeholder Report
blocked on https://github.com/mozilla/experimenter/issues/3175 and https://github.com/mozilla/experimenter/issues/3176.

Bonus - people are using the CSV export!!  Now our customers are ready to customize the export.

The original ticket had a pretty large list of options https://github.com/mozilla/experimenter/issues/2634.  That was a first SWAG at what should go into the report.  Here is the adapted list after feedback.  I'm happy to see this added as another export link option in addition to the existing one - if that is possible.  

I'm not sure if we are able to have the column title be different than the experimenter field name - but I'm adding what I'd like the column name to be if possible - after the field name - when it would be nice to rename.  Not required.  

Status
Public name - Name (note: here we'd love if the name could be a link to the experimenter URL - if that is possible - then we don't need a column for experimenter URL later)
Public description - Description
Hypothesis -  https://github.com/mozilla/experimenter/issues/3175
Analysis Plan - Leading indicators - https://github.com/mozilla/experimenter/issues/3176
experimenter URL - only include separate column if can't be just linked to the name
proposed start date - Start date
experiment type - Type
proposed duration - Length
firefox channel - Channel
Experiment owner - Owner
Data Scientist
Estimated Total Enrolled Clients - Enrolled target
Results URL
Related Projects - Project - requesting rename here https://github.com/mozilla/experimenter/issues/3177
Locales - believe nothing here would mean all - which is fine
Country - believe nothing here would mean all - which is fine

This is taken from the [demo sheet.](https://docs.google.com/spreadsheets/d/1apgrYEti0O3UVQMv4wp3IQDOj2US1Y3UFMopLlDodlY/edit#gid=2063326940)  That sheet includes "desired" fields we don't have well defined or a way to collect yet - which is why this list doesn't match exactly.

tiftran:
Spreadsheet example
https://docs.google.com/spreadsheets/d/1-jhmb1SQYQIvFKt75qd6_L_qBUioITIDDofSap_GMc8/edit?usp=sharing

### DIFF
--- a/app/experimenter/experiments/api/v2/views.py
+++ b/app/experimenter/experiments/api/v2/views.py
@@ -109,10 +109,15 @@ class ExperimentTimelinePopulationView(RetrieveUpdateAPIView):
     serializer_class = ExperimentTimelinePopSerializer
 
 
+class ExperimentCSVRenderer(CSVRenderer):
+    header = ExperimentCSVSerializer.Meta.fields
+    labels = dict(((field, field.replace("_", " ").title()) for field in header))
+
+
 class ExperimentCSVListView(ListAPIView):
     queryset = Experiment.objects.get_prefetched().order_by("status", "name")
     serializer_class = ExperimentCSVSerializer
-    renderer_classes = (CSVRenderer,)
+    renderer_classes = (ExperimentCSVRenderer,)
 
     def get_queryset(self):
         return ExperimentFilterset(
--- a/app/experimenter/docs/openapi-schema.json
+++ b/app/experimenter/docs/openapi-schema.json
@@ -1006,6 +1006,23 @@
                         "type": "string",
                         "maxLength": 255
                       },
+                      "description": {
+                        "type": "string"
+                      },
+                      "hypothesis": {
+                        "type": "string"
+                      },
+                      "leading_indicators": {
+                        "type": "string"
+                      },
+                      "experiment_url": {
+                        "type": "string",
+                        "readOnly": true
+                      },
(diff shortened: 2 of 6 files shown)


app/experimenter/experiments/api/v2/views.py
```diff
@@ -109,10 +109,15 @@ class ExperimentTimelinePopulationView(RetrieveUpdateAPIView):
     serializer_class = ExperimentTimelinePopSerializer
 
 
+class ExperimentCSVRenderer(CSVRenderer):
+    header = ExperimentCSVSerializer.Meta.fields
+    labels = dict(((field, field.replace("_", " ").title()) for field in header))
+
```
tiftran, line 115: To format the column headings

jaredlockhart:
@tiftran I can't view your example spreadsheet, maybe the 'Sharing -> Copy Link (visible all mozilla)' or something?

tiftran:
@jaredlockhart  how about https://docs.google.com/spreadsheets/d/1-jhmb1SQYQIvFKt75qd6_L_qBUioITIDDofSap_GMc8/edit?usp=sharing?

app/experimenter/experiments/api/v2/views.py
jaredlockhart: Oh very nice 👍

PR #3251, opened 2020-08-18
Adds v3 serializer for branches

k88hudson:
This adds a serializer for branches (will file a ticket for adding an actual update function) – I didn't have time to finish fixing/updating tests, so could use some help there

### DIFF
--- a/app/experimenter/experiments/api/v3/serializers.py
+++ b/app/experimenter/experiments/api/v3/serializers.py
@@ -33,6 +33,29 @@ class ExperimentRapidChangelogSerializerMixin(ChangelogSerializerMixin):
         return updated_instance
 
 
+class ExperimentRapidVariantSerializer(serializers.ModelSerializer):
+    id = serializers.ReadOnlyField()
+    slug = serializers.ReadOnlyField()
+
+    def validate_ratio(self, value):
+        if 1 <= value <= 100:
+            return value
+
+        raise serializers.ValidationError(["Branch sizes must be between 1 and 100."])
+
+    class Meta:
+        fields = [
+            "id",
+            "slug",
+            "name",
+            "description",
+            "is_control",
+            "ratio",
+            "value"
+        ]
+        model = ExperimentVariant
+
+
 class ExperimentRapidSerializer(
     ExperimentRapidChangelogSerializerMixin, serializers.ModelSerializer
 ):
@@ -63,6 +86,8 @@ class ExperimentRapidSerializer(
     monitoring_dashboard_url = serializers.ReadOnlyField()
     reject_feedback = serializers.SerializerMethodField()
 
+    variants = ExperimentRapidVariantSerializer(many=True)
+
     class Meta:
         model = Experiment
         fields = (
@@ -82,6 +107,7 @@ class ExperimentRapidSerializer(
             "type",
             "reject_feedback",
             "recipe_slug",
+            "variants"
         )
 
     def get_reject_feedback(self, obj):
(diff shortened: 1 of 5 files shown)


app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -109,13 +126,36 @@ def validate(self, data):
                         ]
                     }
                 )
+
+        variants = data.get("variants")
+
+        if variants:
+            if not self.is_variant_valid(variants):
```
jaredlockhart, line 133: I would name this more descriptively like `all_variant_slugs_unique` or something

app/experimenter/experiments/tests/api/v3/test_views.py
```diff
@@ -12,6 +12,24 @@
 from experimenter.bugzilla.tests.mixins import MockBugzillaTasksMixin
 
 
+FAKE_VARIANTS = [
```
jaredlockhart, line 15: If this is only used in one test case you can just stick this right on the test class, either as a class var like

```py
class TestWhatever
    FAKE_VARIANTS = { .... }
```

or

```
class TestWhatever
    def setUp(self):
        self.variants_data = { .... }
```

and then you can refer to it in the test as `self.FAKE_VARIANTS` or `self.variants_data`

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -197,6 +197,8 @@ export const SettingsForm: React.FC = () => {
 };
 
 export const BranchesForm: React.FC = () => {
+  const formData = useExperimentState();
+  const variants = formData.variants || [];
```
jaredlockhart, line 201: Does the serializer spit out `null` or `[]` if it doesnt' have any variants?

app/experimenter/experiments/tests/api/v3/test_serializers.py
```diff
@@ -8,20 +9,39 @@
...
+        "is_control": True,
+    },
+    {
+        "slug": "variant",
+        "name": "variant",
+        "ratio": 50,
+        "description": "a variant",
+        "is_control": False,
+    },
+]
```
jaredlockhart, line 38: I set a bad precedent here with putting these at the top of the file, we should try to keep test data that's only relevant to a single testcase inside the class, but we can deal with that later.

app/experimenter/experiments/tests/api/v3/test_serializers.py
```diff
@@ -239,6 +363,7 @@ def test_serializer_creates_experiment_and_sets_slug_and_changelog(self):
             "features": ["picture_in_picture", "pinned_tabs"],
             "firefox_min_version": FIREFOX_VERSION,
             "firefox_channel": Experiment.CHANNEL_RELEASE,
+            "variants": FAKE_VARIANTS,
```
jaredlockhart, line 366: @k88hudson @tiftran Don't we need to add checks that the variants were created correctly in this test?

k88hudson:
> I don't think adding the frontend changes were necessary in this PR

I did the bare minimum so it's easier to manually test

app/experimenter/experiments/tests/api/v3/test_serializers.py
k88hudson: I can add to the class instead if that's preferable

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
k88hudson: good question, I was hoping it should be `[]` – what's the right way to do that?

app/experimenter/experiments/tests/api/v3/test_serializers.py
```diff
@@ -262,6 +390,10 @@ def test_serializer_creates_experiment_and_sets_slug_and_changelog(self):
             experiment.public_description, Experiment.BUGZILLA_RAPID_EXPERIMENT_TEMPLATE
         )
 
+        self.assertEqual(experiment.variants.count(), 2)
+        self.assertTrue(experiment.variants.filter(**self.variants_data[0]).exists())
+        self.assertTrue(experiment.variants.filter(**self.variants_data[1]).exists())
```
jaredlockhart, line 395: Parfait 💯

k88hudson:
Thanks @tiftran! What do you think @jaredlockhart?

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
jaredlockhart: Oh I wasn't even thinking.  Yeah so basically the pattern we've gone with after having to go through this a few times now is for the serializer to spit out an 'empty' state that's valid for the frontend, so we'd make sure that when you pass no instance into the serializer what comes out is valid.  So in this case it should be `variants: []`, which I think will happen automatically with the list serializer, but it's good to assert on.

Which reminds me that we didn't add a testcase for just the ExperimentRapidVariantSerializer by itself, which we should do.

Sorry @tiftran ! Can you pls add a test class for the ExperimentRapidVariantSerializer that tests

- its expected output
- move the ratio validation test in there

jaredlockhart:
> > I don't think adding the frontend changes were necessary in this PR
> 
> I did the bare minimum so it's easier to manually test

Oh I see, yeah I still wouldn't have included that in here and put it in the frontend pr.  It's frontend changes that aren't strongly related to the backend code working or tests passing, so I don't think it's strictly necessary for it to be in here.

PR #3252, opened 2020-08-18
UI for branches (WIP)

k88hudson:
Depends on  #3251 (serializer)

Adds some basic read-only UI for branches 

I removed the editing/updating functions for now until the update serializer changes land.


app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -200,31 +200,118 @@ export const BranchesForm: React.FC = () => {
...
+        // Sort control branch to the top
+        .sort((a, b) => (a.is_control < b.is_control ? 1 : -1))
+        .map((variant, i) => {
+          return (
+            <div key={i} className="card mb-4">
+              <ul className="list-group list-group-flush">
+                <li className="list-group-item">
+                  <div className="container">
+                    <div className="row">
+                      <div className="col">
```
jaredlockhart, line 218: This is why I wanted to add `react-bootstrap` so we could just use the components instead of having to do it this way, then you get to use the props to pass config to them too.

k88hudson: Yeah I can see how that would be nice. Especially for the closing tags (a million nested divs is annoying!)

PR #3259, opened 2020-08-18
 ADR for Experimenter Architecture fixes #3256

### LINKED ISSUE #3256: ADR for Experimenter Architecture
Let's analyze the current structure of Experimenter and determine how best to proceed to meet the needs of current and feature experimentation efforts.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-67)

jaredlockhart:
Here is an ADR to discuss the future architecture of Experimenter for Project Nimbus.

Based on ADR templates from FXA:

https://github.com/mozilla/fxa/tree/main/docs/adr

https://github.blog/2020-08-13-why-write-adrs/


clouserw:
Thanks @jaredlockhart .  This really helps add some context.  I'll ping some people for any comments on this draft.  Anyone feel free to add anyone else please.  @emtwo @k88hudson @tiftran @lmorchard @bbangert @jodyheavener @LZoog

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+  - Changes that must apply to both Normandy and Nimbus experiments must be made in two places while they're both being actively maintained
+  - Divergence in tooling/processes over time may create confusion when context switching between the two
+  - Possibly still somewhat constricted by past core design decisions
+  - [Python/Django may not align with FXA tools/processes](https://github.com/mozilla/fxa/blob/main/docs/adr/0020-application-architecture.md)
+
+### Start A New Application in TypeScript/Node
+
+Creating a new web application in TypeScript/Node would allow us to completely restart our thinking about the entire structure of the application which would result in both code and workflow/UI that are purpose built from the ground up to solely address the concerns of Nimbus experiments.
+
+- Pros
```
clouserw, line 91: The assisting FxA team has a strong background in TypeScript/Node as well as the supporting libraries, frameworks, and processes around that.  I think that's a definite `Pro` here*, particularly since we'd like to move quickly on this.

* I'm wondering if I would call it a `Con` for the others....  We have Python/Django experience, but it's years old.  Since rapid development and `Fitting cohesively into the FXA engineering pipeline/culture` are decision drivers listed above, I guess I would.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
+# Create A New Web Application To Manage Nimbus Experiments
+
+* Status: proposed
+* Deciders: Kate Hudson, Jared Lockhart, TBD
+* Date: 2020-08-18
+
+## Context and Problem Statement
+
+Experimenter is a web application developed solely for the purpose of managing product experiments at Mozilla.  It was originally designed to manage the Shield experimentation workflow, and has expanded to include a comprehensive workflow for managing Normandy experiments.  Project Nimbus is being developed with very different core assumptions than the previous experimentation platforms, and it is unclear whether building on top of the existing Experimenter infrastructure or creating a new purpose specific application will best suit the project's needs.
```
lmorchard, line 9: Reading that "Project Nimbus is being developed with very different core assumptions" makes me want to see what the diff in assumptions are in order to know how that would affect the decision here.

Not very familiar with Experimenter, Shield, Normandy, or Nimbus. Some links to context might help - docs, talks, whatever might be available.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+  - Reuse existing components
+    - Database schemas are largely reusable
+    - Public APIs work seamlessly for Normandy and Nimbus experiments
+    - External service integrations (bugzilla/remote settings/email/auth0) work seamlessly
+  - Comprehensive development/testing/deployment infrastructure in place
+    - 100% unit testing coverage across Python/JS/TS
+    - Automated Selenium Integration tests for frontend
+    - Continuous Deployment
+- Cons
+  - Existing workflow/UI not applicable
```
lmorchard, line 59: I haven't seen any substantial context on Nimbus. I'm also not familiar with how Experimenter works for Shield and Normandy. Would be useful to see what the current UI is and how/why it is not applicable to Nimbus.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Context and Problem Statement
+
+Experimenter is a web application developed solely for the purpose of managing product experiments at Mozilla.  It was originally designed to manage the Shield experimentation workflow, and has expanded to include a comprehensive workflow for managing Normandy experiments.  Project Nimbus is being developed with very different core assumptions than the previous experimentation platforms, and it is unclear whether building on top of the existing Experimenter infrastructure or creating a new purpose specific application will best suit the project's needs.
+
+## Decision Drivers <!-- optional -->
+
+* Meeting Project Nimbus' needs in a rapid, continuous fashion with minimal downtime
+* Contributability by a large, distributed group of people with different backgrounds (web engineers, data engineers, data scientists)
+* Fitting cohesively into the FXA engineering pipeline/culture
```
lmorchard, line 15: Looks like the consideration of python vs node/typescript is tied to this driver.

We do have some Python experience on the team (me included). Like @clouserw says, it's been a few years. But, it would be interesting if we could balance learning curve vs code reuse vs greenfield opportunity. (e.g. I'm onboard to dust off my Python if it means we could simultaneously reuse some existing stuff but also have a relatively clean start.)

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+  - Logic/types in TS can be easily shared across multiple projects potentially including Desktop/Mobile clients
+  - TS may integrate more smoothly with Rust than Python does to more tightly couple the server and the new Experiment Client
+- Cons
+  - Lose single source of truth
+    - Possibly require people to know which application to refer to depending on experiment type
+    - Possibly create an API proxy in front of both to have a single endpoint for dependent callers
+    - More difficult to correlate experiments/learnings across the two applications
+  - Changes that must apply to both Normandy and Nimbus experiments must be made in two places while they're both being actively maintained, but with the added complexity that one is in Python/Django and the other is in TS/Node which increases the context switching complexity
+  - Node application frameworks are less mature than Django
+  - No Django Admin Panel for easy data access
```
lmorchard, line 107: Yeah, that is a thing I miss from living in Django a few years ago. A lot of batteries included in the admin panel & ORM stuff. I have to imagine there's something similar in node-land, but also I haven't been on a project using it if there is one.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+    - Possibly create an API proxy in front of both to have a single endpoint for dependent callers
+    - More difficult to correlate experiments/learnings across the two applications
+  - Changes that must apply to both Normandy and Nimbus experiments must be made in two places while they're both being actively maintained
+  - Divergence in tooling/processes over time may create confusion when context switching between the two
+  - Possibly still somewhat constricted by past core design decisions
+  - [Python/Django may not align with FXA tools/processes](https://github.com/mozilla/fxa/blob/main/docs/adr/0020-application-architecture.md)
+
+### Start A New Application in TypeScript/Node
+
+Creating a new web application in TypeScript/Node would allow us to completely restart our thinking about the entire structure of the application which would result in both code and workflow/UI that are purpose built from the ground up to solely address the concerns of Nimbus experiments.
```
lmorchard, line 89: I'd say this is a big fulcrum for the decision: Just what exactly is the diff between Normandy & Nimbus? Is there a sufficient volume of cruft / crud / debt / blue sky that would make greenfield really attractive in implementing that diff?

I think that's even more important than picking between python vs node, IMO

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Pros
+  - Totally greenfield approach allows fresh perspectives to flourish
+  - Stronger sense of collective ownership from all new contributors
+  - Every decision and line of code built specifically to address the needs of Nimbus
+  - TypeScript provides excellent feedback and guarantees which improves development experience
+  - Native async/websocket support allows richer UI experience when coupled with SPA frameworks (like React)
+  - [Fits more cohesively into the FXA team's tooling/processes](https://github.com/mozilla/fxa/blob/main/docs/adr/0020-application-architecture.md)
+  - Logic/types in TS can be easily shared across multiple projects potentially including Desktop/Mobile clients
+  - TS may integrate more smoothly with Rust than Python does to more tightly couple the server and the new Experiment Client
```
rfk, line 99: At first read, I don't buy this "they both have strong typesystems and hence might couple more smoothly" suggestion; do we have a concrete example of what sort of benefit we might derive from this in practice? The systems are going to be intermediated by an webapp API on the server (http+json? graphql?) and by a foreign-language-bindings layer on the client which might negate any coupling benefits here.

(I can imagine a benefit of having *similarly-structured* datatypes on the server and client, but this bullet seems to imply a tighter level of coupling than that)

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+    - More difficult to correlate experiments/learnings across the two applications
+  - Changes that must apply to both Normandy and Nimbus experiments must be made in two places while they're both being actively maintained, but with the added complexity that one is in Python/Django and the other is in TS/Node which increases the context switching complexity
+  - Node application frameworks are less mature than Django
+  - No Django Admin Panel for easy data access
+  - Forced to rewrite all the components that do still apply to both Normandy and Nimbus Experiments
+    - Core db state/logic
+    - Public and Private APIs
+    - Dependent service integration (bugzilla/remote settings/email/auth0)
+    - History/Changelog management
+    - User roles/permissions
```
rfk, line 113: IMHO, we know that engineers tend to have a bias towards clean rewrites, and should mentally weight the complexity of these things-to-be-re-implemented higher in order to try to counter it. Would the re-write still be a good idea if these took twice as long to re-implement as we think they will?

(Conversely, how important are these to the MVP success of Nimbus?)

rfk (commented):
Thanks for writing this up! I don't have a lot of personal skin in the game here but I tried to ask a few questions to help discussion along.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
k88hudson: I think the most relevant thing to note here is that prior to project Nimbus, Experimenter played a supporting role in the *planning* and *reporting* phases of some types of experiments (mainly for Firefox Desktop), but the configuration, launch, and review of experiments happens across many different tools (Normandy, Github/Remote Settings, or third-party tools like Leanplum, among others), most of which require expert operators and coordinators to be involved. 

In Nimbus, the central assumption is that the Experiment Console should play a much more central role as the single source of truth for all experiments. It should provide a single web interface for the configuration/management/review/launch of experiments across platforms, rather than having to be translated manually by operators and coordinators into different systems for each experiment type. It should enforce a single API and experiment format across platforms (mobile, desktop, and eventually web).

This [Vision Document](https://docs.google.com/document/d/10XFq-nWS76UBUsytHnwoaie4Kz9-zO8JrtMZzcaJ5h0/edit#) goes into more detail about why, this [system diagram of experiments in Firefox](https://docs.google.com/document/d/1ypWh-aug9NFdArQfYCuLYpUpK_IKdgELVmx21CaWI6w/edit#heading=h.s770289vwy3r) describes the system as it exists prior to nimbus, at least the part relevant to Desktop.

bbangert: From the docs I've seen on Nimbus, the architecture has some fundamental differences (typically simplifications) that:
- Change what service/UX is responsible for a given function.
- Consolidate or remove portions to streamline interaction and/or process.

I assume this is the 'very different core assumption' being referenced here?

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+Creating a new web application in TypeScript/Node would allow us to completely restart our thinking about the entire structure of the application which would result in both code and workflow/UI that are purpose built from the ground up to solely address the concerns of Nimbus experiments.
+
+- Pros
+  - Totally greenfield approach allows fresh perspectives to flourish
+  - Stronger sense of collective ownership from all new contributors
+  - Every decision and line of code built specifically to address the needs of Nimbus
+  - TypeScript provides excellent feedback and guarantees which improves development experience
+  - Native async/websocket support allows richer UI experience when coupled with SPA frameworks (like React)
+  - [Fits more cohesively into the FXA team's tooling/processes](https://github.com/mozilla/fxa/blob/main/docs/adr/0020-application-architecture.md)
+  - Logic/types in TS can be easily shared across multiple projects potentially including Desktop/Mobile clients
```
k88hudson, line 98: Because we already have [types written](https://github.com/mozilla/nimbus-shared/tree/main/types) for experiments that are in use (at least, for Desktop) this would be an immediate benefit

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+  - TS may integrate more smoothly with Rust than Python does to more tightly couple the server and the new Experiment Client
+- Cons
+  - Lose single source of truth
```
k88hudson, line 101: It's worth noting that while we will pause feature development, we will still need to maintain and triage bug reports for Experimenter until we no longer support Normandy experiments (or could provide an acceptable alternative for intake), and the people likely to have to do that overlap with this team.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+    - Possibly require people to know which application to refer to depending on experiment type
+    - Possibly create an API proxy in front of both to have a single endpoint for dependent callers
+    - More difficult to correlate experiments/learnings across the two applications
+  - Changes that must apply to both Normandy and Nimbus experiments must be made in two places while they're both being actively maintained, but with the added complexity that one is in Python/Django and the other is in TS/Node which increases the context switching complexity
+  - Node application frameworks are less mature than Django
+  - No Django Admin Panel for easy data access
+  - Forced to rewrite all the components that do still apply to both Normandy and Nimbus Experiments
```
k88hudson, line 108: In addition to application logic, I would imagine there is a significant cost to having to implement (or at least make decisions about) stuff that's included in Django/Django REST (e.g. an ORM, roles, admin panel as mentioned above) v.s. a node framework, although that might be negated if this team has a node stack that is already proven and is much more familiar with using it it.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+    - Core db state/logic
+    - Public and Private APIs
+    - Dependent service integration (bugzilla/remote settings/email/auth0)
+    - History/Changelog management
+    - User roles/permissions
+
```
k88hudson, line 114: There is a cost to supporting teams, including cloud ops/IT/QA, to setting up and running an additional application. I don't think that should be a blocker but in the current climate it's at least worth considering.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+    - Database schemas are largely reusable
+    - Public APIs work seamlessly for Normandy and Nimbus experiments
+    - External service integrations (bugzilla/remote settings/email/auth0) work seamlessly
+  - Comprehensive development/testing/deployment infrastructure in place
+    - 100% unit testing coverage across Python/JS/TS
+    - Automated Selenium Integration tests for frontend
+    - Continuous Deployment
+- Cons
+  - Existing workflow/UI not applicable
+  - Some database schemas require refactoring to isolate Normandy and Nimbus concerns
```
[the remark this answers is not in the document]
k88hudson: ...however, this could be mitigated by creating new tables/models.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+## Context and Problem Statement
+
+Experimenter is a web application developed solely for the purpose of managing product experiments at Mozilla.  It was originally designed to manage the Shield experimentation workflow, and has expanded to include a comprehensive workflow for managing Normandy experiments.  Project Nimbus is being developed with very different core assumptions than the previous experimentation platforms, and it is unclear whether building on top of the existing Experimenter infrastructure or creating a new purpose specific application will best suit the project's needs.
+
+## Decision Drivers <!-- optional -->
+
+* Meeting Project Nimbus' needs in a rapid, continuous fashion with minimal downtime
+* Contributability by a large, distributed group of people with different backgrounds (web engineers, data engineers, data scientists)
+* Fitting cohesively into the FXA engineering pipeline/culture
+* Continuing to support the existing experimentation (Normandy) workflow
```
emtwo, line 16: I think something we need to add as a decision driver is related to who will be maintaining the tooling long term.

If there is any chance that the data org/data tools team would be part of the long-term maintenance, it's worth considering what type of tooling we work with in our pipelines. Generally we work with Python/Svelte.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+  - Get a big head start by not needing to reimplement the reusable parts
+  - Isolate Nimbus concerns without fear of breaking Normandy experiments
+  - Consistent development and deployment workflow for both old and new applications
+  - Opportunity to rethink core code/architectural/tooling decisions where appropriate
+- Cons
+  - Lose single source of truth
+    - Possibly require people to know which application to refer to depending on experiment type
+    - Possibly create an API proxy in front of both to have a single endpoint for dependent callers
+    - More difficult to correlate experiments/learnings across the two applications
+  - Changes that must apply to both Normandy and Nimbus experiments must be made in two places while they're both being actively maintained
```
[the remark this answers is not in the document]
emtwo: I have the same question. If the plan is to deprecate Normandy experiments in the future, that gives a stronger reason to start fresh here.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
emtwo: I'm really tempted to refer back to the question of whether Normandy will eventually be deprecated. If so, a rewrite seems much more worthwhile.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: Yep that's exactly what I meant with that point 👍

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: Yeah I think the broadest way I can describe it are as:

The old (Normandy focused) workflow is about capturing an experiment _proposal_ in high level natural language, and then many parties (like data science, firefox engineering, release management, qa) converting that into a concrete experiment implementation, analyzing and capturing risks, and making sure those are all signed off on.  It's a very involved workflow that touches many parties, and Experimenter's role is to capture and orchestrate all of it.

The new (Nimbus) workflow is about exposing a configuration panel that in theory a single experiment owner can use to define and launch an entire experiment without needing any of the steps I listed above.  The differences that allow this are greatly restricting and simplifying the experiment configuration space, Experimenter generating the experiment configuration (recipe) in its entirety, and sending that directly to Firefox through Remote Settings.

So the overall effect is that the majority of the UI that was built for the old workflow doesn't apply, but the core mechanics and logic in the backend internals still do.  The way we've approach that thus far is to create a new set of isolated APIs and React frontend for the new workflow that sits next to the old one, but sharing the same db models and logic.  It's worked well so far.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: I commented above with some more detailed description of those differences, I could move it into the ADR itself if it would be helpful?

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: Yeah I wouldn't characterize it as debt more just which parts still apply (mostly backend, listed above) and don't apply (frontend, strategy described above)

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
jaredlockhart: Yeah if we continue to build on top of Experimenter I would expect that when Normandy experiments become deprecated we'd remove all the relevant parts from Experimenter without disturbing any of the Nimbus specific parts.  If we start from scratch we'd do the opposite, re implement the shared components (if we're starting in Node) and then just deprecate the existing Experimenter codebase entirely.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: Well I was thinking of the case where we have the Rust experiments client, and we want to strongly enforce the contract between that client and the server by including the client in the server's CI test flow.  You can [expose Rust functions to Python](https://github.com/PyO3/pyo3), and you can [expose Rust functions to TypeScript](https://github.com/rustwasm/wasm-bindgen), but with TypeScript it'll also [generate the TypeScript stubs from the Rust types](https://rustwasm.github.io/wasm-bindgen/reference/cli.html#--typescript) so I think that'll allow tighter integration and a smoother developer workflow, but I'm purely speculating.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: Yeah that's a good point @rfk, if we restart from scratch I think we'll have to allow for some downtime possibly at the scale of months while the core mechanics of Experimenter that are needed for Nimbus are reimplemented.  It might be possible to reorder the things I listed above to hit a working MVP of a Node implementation for Nimbus sooner, but I think all of those things will be necessary at some point.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+
+* Status: proposed
+* Deciders: Kate Hudson, Jared Lockhart, TBD
+* Date: 2020-08-18
+
+## Context and Problem Statement
+
+Experimenter is a web application developed solely for the purpose of managing product experiments at Mozilla.  It was originally designed to manage the Shield experimentation workflow, and has expanded to include a comprehensive workflow for managing Normandy experiments.  Project Nimbus is being developed with very different core assumptions than the previous experimentation platforms, and it is unclear whether building on top of the existing Experimenter infrastructure or creating a new purpose specific application will best suit the project's needs.
+
+## Decision Drivers <!-- optional -->
```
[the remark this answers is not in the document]
jaredlockhart: Yeah I think that's the right way to think about it.  I think broadly building on top of what we have would give faster short term deliverables, but thinking to 1+ year from now, if the core maintaining team has focused on Node as a platform for all services, then taking the hit to deliverability now could be worth it?

Conversely we might wind up in a situation where restarting now would halt Nimbus progress, and as development continues on Experimenter for its existing experiment types, the gap between the two could widen sufficiently that it becomes more and more difficult for the Node implementation to catch up.  An example would be the work that's going on now to build high level reporting on experiment outcomes for leadership, which would apply equally to Normandy and Nimbus experiments.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+* …
+
+## Pros and Cons of the Options
+
+### Build On Top Of Experimenter
+
+Much of the existing Experimenter infrastructure can be reused, which gives Project Nimbus a well established framework to build on top of.
+
+- Pros
+  - Single source of truth for all past/future experiments
```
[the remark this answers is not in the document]
jaredlockhart: I would also get @shell1 's feedback on this since she has been working closely with leadership on this

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
jaredlockhart: Yeah that's what I meant, right now we're just naively using the same model for both, but it's very possible to split it up into pieces such that the nimbus focused code only interacts with db models that are specific to its concerns.  Using this approach it would be possible to have some shared db models for things that apply to both, allowing things like public apis and high level reporting to treat all experiments in a consistent fashion while isolating the Normandy and Nimbus specific models to their respective flows.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: Good point

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
rfk: > we have the Rust experiments client, and we want to strongly enforce the contract between that client
> and the server by including the client in the server's CI test flow

Oh, interesting point, thanks!

Amusing semi-aside: the currently plan for exposing the Rust SDK to other languages is to use a [cross-language bindings generator](https://github.com/mozilla/uniffi-rs), which coincidentally happens to [already support python](https://github.com/mozilla/uniffi-rs/tree/main/uniffi_bindgen/src/bindings/python) but for which we haven't tackled JS/TS yet.

tdsmith:
With respect to accessibility to partner teams, I wouldn't expect substantive code contributions from data scientists per se to either a Django or a Typescript codebase. I expect most data scientists would interact with any application in the same ways (i.e. through a web UI or by editing some kind of data file). The Django admin UI is a nice advantage if we make all of the nimbus-shared content Experimenter's responsibility but that's a judgment for y'all.

I also don't think it matters that the experiment analysis platform, [Jetstream](https://github.com/mozilla/jetstream), depends on the scientific Python stack. It consumes JSON and emits BigQuery tables and those interfaces don't depend on the technology choice here. It runs as a batch job invoked from the nightly data platform ETL run so I think it's not super obvious that we'd want to integrate it here.

As an Experimenter user, I'd be sad if we tore it down and we lost our records of historical experiments and all of our links broke; I do pull up old PHDs every now and again. Maybe there's something clever we can do about that.

If we'd actually started to use Experimenter as a results repository, I'd be sad to lose that history, but we haven't -- I think this is an interesting area to develop but I don't think it weighs on the decision here except that I think we'd like to have a good story moving forward.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
shell1: Taking a near-term delivery timeline hit now to refactor optimize for FxA current practices- assumes longer term ownership of the code area.  We've seen things in this area being more fluid than most.  What if experimenter/nimbus experimenter ends up under Data in a year?  Will the delayed delivery have been worth it?  Will we be in a less ideal place with 2 code bases for very similar functionalities?

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
shell1: I suspect that supporting Normandy existing workflow is 12-18 months at least, with fewer use cases going through Normandy as we transition more to the Nimbus flow.  We all hope it will be quicker - but just based on past experience - removing humans can mean lots of supporting work to ensure that is smooth.

It's not just normandy pref-flip experiments that experimenter supports.  It's also rollouts and add-ons.  It's also the pref-flips used to experiment on other systems without built in experimentation (ex: pocket, snippets, surveys). 

Nimbus will start taking over feature flags and xman direct integration - but there are more integration efforts on the path to Nimbus replacement for experimentation.  

The deprecation or re-homing investigation effort will continue in parallel (for capabilities we don't plan to move to Nimbus).

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
shell1: It is maximally critical to get be able to utilize a single source of truth for experimentation.  Not just between Normandy and Nimbus, but with the various web platforms, and mobile.  

Stakeholders are leadership, anyone planning experiments, and anyone needing to know what is going on in a product area (ex: PM).  The goal is to easily be able to learn, inform current plans, and present our learnings as part of the direction narrative for a product area.  For that we need to be able to answer:

"what hypothesis have we tested in the past, are we testing currently, or will we test in the future for ___ area".  "how did we measure those hypothesis?" 

One big weakness is different teams experimenting in the same area - in different ways (ex: Product and Marketing running experiments in same area - but results not cross-shared.  

We want to increase our ability to weave a complete narrative.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
shell1: We should add a pro of "dynamic risk mitigation".  I embrace designing for the simplest case - and agree Nimbus is removing a lot of the risk mitigation steps needed.  But in the design -there have been several cases where risk will need to set on different path for mitigation (partner involved, new metrics needed, new targeting needed).  The goal is 100% to move to lighter weight/no review whenever possible - but kick the ones that need more hands-on into that flow.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
```diff
@@ -0,0 +1,119 @@
...
+Creating a new web application in Python/Django would allow us to reuse much of the existing code/infrastructure where applicable while clearly isolating the concerns of supporting existing experiment types and building out new Nimbus functionality.
+
+- Pros
+  - Start by cloning the repo and deleting all the unnecessary UI/DB/APIs
+  - Get a big head start by not needing to reimplement the reusable parts
+  - Isolate Nimbus concerns without fear of breaking Normandy experiments
+  - Consistent development and deployment workflow for both old and new applications
+  - Opportunity to rethink core code/architectural/tooling decisions where appropriate
+- Cons
+  - Lose single source of truth
```
LZoog, line 78: I mention this in another comment, but to reiterate/expand here: how much weight does this con have? If we want to check various prior experiments often even after they aren't supported any longer, it seems having access to a history of Normandy experiments would be handy to query in Nimbus?

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
tdsmith: Shield is an obsolete name for the Firefox experimentation program; "Shield studies" were/are shipped using Normandy.

Experimenter helps manage some (but not all) experimentation for our mobile products, which do not use Normandy because the client is not available. This is an example of an experiment deployed on Fenix using Mako (which is part of Android Components and which consumes content from Kinto), which had some metadata tracked in Experimenter: https://experimenter.services.mozilla.com/experiments/search-widget-discoverability/

rfk: > This is an example of an experiment deployed on Fenix using Mako

Adding for completeness: replacing Mako is an explicit goal of the nimbus mobile SDK

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
tdsmith: I was hoping that integrating experiment reporting into Experimenter would allow us to use Experimenter as a results repository. Stakeholders do seem to want one-stop shopping, and Jetstream and the reporting interface will support both Nimbus- and Normandy-flavored experiments. If we end up with a legacy Normandy-flavored product and a new Nimbus-flavored product and we end up supporting both for 6-12 months, that story gets a little harder to write.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
shell1: Normandy experiments (pref-flip and add-on experiments)
Message system (xman) [CFR](https://experimenter.services.mozilla.com/experiments/high-velocity-message-testing-multi-account-containers-retention-study/) + about:welcome experiments -[ex](https://experimenter.services.mozilla.com/experiments/multi-stage-aboutwelcome-v3-80-release/) 
Mobile experiments on Mako ([ex](https://experimenter.services.mozilla.com/experiments/search-widget-discoverability/))
Rollouts (add-on & [pref-flip](https://experimenter.services.mozilla.com/experiments/activity-stream-pocket-new-tab-spoc-personalization-v2-rollout/)) 
Using pref flips to add experimentation to other systems - Pocket ([ex](https://experimenter.services.mozilla.com/experiments/activity-stream-pocket-new-tab-story-spocs-in-uk-experiment/)) + snippets ([ex](https://experimenter.services.mozilla.com/experiments/pref-flip-experiment-high-velocity-message-testing-f100-snippets/))
Generic experiments for visibility or for someone needing a tool to help coordinate but not having one that fits. ([ex](https://experimenter.services.mozilla.com/experiments/youtube-regrets-reporter/)).  We use generic intake as we address new request for intake assistance as an easy way to test out scenario to see if custom work is needed.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
jaredlockhart: I think @shell1 addresses that well here:

https://github.com/mozilla/experimenter/pull/3259#discussion_r475719606

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
rafrombrc: Yes, this is a very important point. Long term ownership is very much not resolved, and if I were to guess there's at least a 50% chance that long term ownership will fall to the Data org, especially given my understanding that Nimbus ultimately is meant to serve all products that use the Glean telemetry system, i.e. more than just Firefox. The current architecture is more closely aligned with Data's Python-centric tooling skills, so in that case we'd be sacrificing velocity now only to cause more pain later.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
rafrombrc: Unfortunately AFAIK we don't yet know who will have long term ownership.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
rafrombrc: I'm repeating myself here, but `Fitting cohesively into the FXA engineering pipeline/culture` is more relevant if the FxA team has long term maintenance / ownership of this app. In the equally likely case that the Data team has long term ownership, then this decision drive should be replaced (or at least tempered) with `Supporting smooth transition into the Data Tools engineering pipeline/culture`, which is strongly Python focused.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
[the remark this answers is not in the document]
clouserw: The good news about the frontend, (ie. "build high level reporting on experiment outcomes for leadership") is that the react components should still be usable, regardless of the decision about the backend

Regarding the concern about ownership: that is a known unknown at this point.  We don't know the long term support plan for the site so we're just guessing.  What we can do is mitigate it with good development practices, full test coverage, and well written documentation (which we should be doing, regardless).

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
mdboom: Personal opinion of course, but I think I'd prefer to (continue to) have a well-defined JSON-based protocol rather than relying on any language magic to connect the client and the server.  If we want to run integration tests, we can make a cli client based on the Rust library and puppet it as its own process.

app/experimenter/docs/adrs/0001-create-new-nimbus-application.md
sciurus: In addition to the long terms costs of an additional application, there's the initial spike of spinup effort. Given the cloudops team's workload and the likely complexity of this new application, that could contribute to delays launching an initial version.

k88hudson:
First of all, huge thanks to everyone who contributed to the discussion on this topic and wrote supporting documentation, especially @jaredlockhart for writing up this ADR and @bbangert / @LZoog for proposals and estimates.

Because this ADR has cross-system dependencies and stakeholders, we’ve supplemented it with a [Decision Brief](https://docs.google.com/document/d/1AwtxZZAqP_2adyfDN5juTHW0TiuU8VV8VzDDo695kZQ/edit#) that clarifies the decision makers and ranks criteria.

After careful consideration of both options, we have decided to adopt the proposal to build on top of Experimenter and invest in improvements. This includes work to create new models, improve and document a development workflow that can be used by all members of the team, and write a new front end in React[1]. While we carefully considered the proposal to start a new application in Node/Typescript, we determined that the risk around the time to bring a new application up to parity and the operations overhead of maintaining parallel systems was too significant to adopt the proposal.

We already have a [list of improvements we’d like to make](https://docs.google.com/document/d/1QMROYab-S5ZBL3oU_hoJxn5V2ocdPG5hlH_kYRMjO_4/edit#heading=h.7fphirx3wcg) – over the next few days we will work on clarifying any details, breaking things down, and getting everyone up to speed on running the existing code base. 

@jaredlockhart can can you update this PR with the text of the decision + a link to the brief so we can get this merged?

[1] Broadly following the front-end rewrite proposal, but within the Experimenter code base

PR #3261, opened 2020-08-18
used actual times when available for csv + status field fixes #2711

### LINKED ISSUE #2711: Use computed dates in CSV rather than proposed
We should be using the computed start/end/duration/enrollment duration values in the CSV reports otherwise all the dates will be wrong.

tiftran:
miss the status field from #3178, slipped it in here...


app/experimenter/experiments/api/v2/serializers.py
```diff
@@ -749,6 +749,7 @@ class Meta:
         model = Experiment
         fields = (
             "name",
+            "status",
```
jaredlockhart, line 752: What's this for?

jaredlockhart: Oh heh kk ;)

PR #3262, opened 2020-08-19
allow projects to be multi select for filtering fixes #2669

### LINKED ISSUE #2669: make "projects" multi-select for tagging and filtering - and expose "no project" selected in filter
using this to slice in multiple ways (ex: projects and KPI projects):

- look at all experiments related to increasing revenue KPI across several teams
- looks at all experiments in pocket related to increasing revenue. 

also want a way to see deliveries that aren't tagged - so "no selection".  This is so shell doesn't go blind trying to see if she missed tagging any - but also in the future to see if there are just experiments that should have been tagged that were not.

tiftran:


app/experimenter/static/scripts/experiment-list-select.js
```diff
@@ -1,3 +1 @@
-$("select[multiple]").selectpicker({
-  noneSelectedText: "All Types"
-});
+$("select[multiple]").selectpicker({});
```
tiftran, line 1: I wasn't able to get rid of this without making the filtering dropdowns disappear?

PR #3267, opened 2020-08-21
Added Platform

tiftran:
Screen Shot 2020-08-20 at 8 17 43 PM


jaredlockhart:
Since we will be restarting our efforts on the frontend and this introduces a data migration, which has long term implications, let's close this for now and revisit it when the frontend becomes more stable.

PR #3294, opened 2020-08-27
added ability to add and delete branches fixes #3292

### LINKED ISSUE #3291: Handle Errors for branches in rapid experiments
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-61)
┆Issue Number: EXP-61

### LINKED ISSUE #3292: Allow ability to add and remove branches for rapid experiments
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-60)

tiftran:
Screen Shot 2020-08-26 at 3 13 50 PM


app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -219,6 +219,28 @@ export const BranchesForm: React.FC = () => {
     dispatch(updateExperiment({ variants: updatedVariants }));
   };
 
+  const NEW_BRANCH = {
```
k88hudson, line 222: Can you add a type?

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -219,6 +219,28 @@ export const BranchesForm: React.FC = () => {
...
+    name: "",
```
k88hudson, line 223: I would be nice to give a default name to each branch of `variant-${i}` (assuming it's not already taken)

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -219,6 +219,28 @@ export const BranchesForm: React.FC = () => {
...
+    is_control: false,
+    description: "An empty branch",
+    value: "",
+    ratio: 50,
```
k88hudson, line 227: Although this produces the results we actually want (which is for new branches to be the same ratio by default ) the UI doesn't make sense in terms of its representation as a percentage

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -285,6 +307,18 @@ export const BranchesForm: React.FC = () => {
...
                       </div>
+                      {!variant.is_control && (
+                        <div>
+                          <button
+                            aria-label="Close"
+                            className="close"
+                            type="button"
+                            onClick={() => handleRemoveBranch(i)}
+                          >
+                            <span style={{ color: "red" }}>&times;</span>
```
k88hudson, line 318: Using `red` emphasizes this in the UI, which I don't think is what we want here. Just keep it the default colour

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -219,6 +219,28 @@ export const BranchesForm: React.FC = () => {
...
+    ratio: 50,
+  };
+
+  const handleAddBranch = () => {
+    const updatedVariants = [...variants, NEW_BRANCH];
+
+    dispatch(updateExperiment({ variants: updatedVariants }));
+  };
+
+  const handleRemoveBranch = (i) => {
```
k88hudson, line 236: `i: number` will help catch improper usages

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
tiftran: would 0 be a better default value then?

k88hudson: Ok, as discussed lets do the following for now:
- make this field disabled / not editable
- always set the value of ratio to `1` (including for new branches)
- in the UI, render it as a percentage to 2 decimal points
- remove the validator in the serializer requiring this to add up to 100%
- file a ticket to make ratio editable, which will involve doing client-side validation of adding to 100% and converting back to ratios from percentages (or maybe a different ui for ratios?)

app/experimenter/static/rapid/components/forms/ExperimentForm.tsx
```diff
@@ -198,10 +198,14 @@ export const SettingsForm: React.FC = () => {
 
 export const BranchesForm: React.FC = () => {
   const formData = useExperimentState();
+
   const variants =
     [...formData.variants].sort((a, b) =>
       a.is_control < b.is_control ? 1 : -1,
     ) || [];
+  const [variantRatio, setVariantRatio] = React.useState<string>(
```
k88hudson, line 206: You don't need to use any state here. Just create a function that you can pass to the `value` prop of the input. Something like:
```tsx
  function ratioToPercentage(ratio: number) {
    const ratioTotal: number = variants.reduce((a, b) => a + b.ratio, 0);
    return Math.round((ratio / ratioTotal) * 10000) / 100;
  }
```

Also, I'd probably go with `Math.round` here since it will only display decimals *if necessary*
Which you can then call later:
```tsx
<input value={ratioToPercentage(variant.ratio)} />
```

app/experimenter/experiments/api/v3/serializers.py
```diff
@@ -102,8 +102,16 @@ def get_reject_feedback(self, obj):
...
     def validate(self, data):
 
-        validated_data = super().validate(data)
-        if validated_data.get("slug") is None:
+        variants = data.get("variants")
+
+        if not self.is_variant_valid(variants):
+            error_list = []
+            for variant in variants:
+                error_list.append({"name": [("All branches must have a unique name")]})
```
k88hudson, line 110: :+1:

PR #3338, opened 2020-08-31
added errors for branches fixes #3291

### LINKED ISSUE #3291: Handle Errors for branches in rapid experiments

tiftran:
pulled out the error state from SettingsForm, since the BranchForm needed that as well, unsure of whether I overdid/over complicated it...
Screen Shot 2020-08-30 at 5 56 20 PM

### DIFF
--- a/app/experimenter/static/rapid/contexts/experiment/provider.tsx
+++ b/app/experimenter/static/rapid/contexts/experiment/provider.tsx
@@ -5,18 +5,35 @@ import { fetchExperiment } from "experimenter-rapid/contexts/experiment/actions"
 import context, {
   INITIAL_CONTEXT,
 } from "experimenter-rapid/contexts/experiment/context";
-import reducer from "experimenter-rapid/contexts/experiment/reducer";
-import { Action, ExperimentData } from "experimenter-types/experiment";
+import reducer, {
+  errorReducer,
+} from "experimenter-rapid/contexts/experiment/reducer";
+import {
+  Action,
+  ErrorAction,
+  ExperimentData,
+  ExperimentErrors,
+} from "experimenter-types/experiment";
 
-const ExperimentProvider: React.FC<{ initialState?: ExperimentData }> = ({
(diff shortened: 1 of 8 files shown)


app/experimenter/static/rapid/contexts/experiment/provider.tsx
```diff
@@ -25,7 +42,11 @@ const ExperimentProvider: React.FC<{ initialState?: ExperimentData }> = ({
     dispatch(fetchExperiment(experimentSlug));
   }, [experimentSlug]);
 
-  return <Provider value={{ state, dispatch }}>{children}</Provider>;
+  return (
+    <Provider value={{ state, errors, dispatch, errorDispatch }}>
```
jaredlockhart, line 46: OOohhhhhh okay I see what you were getting at.  I've been curious how we're gonna handle this case.

So the theory with reducers is you can define many of them in isolation to each manage other own state, and then 'combine' them into a single state with a single dispatch, like a tree, and then you send a single action into the reducer tree and it propagates down and because the action names are unique constants, only the right reducer responds to the right message, or multiple reducers can respond to teh same message if necessary, otherwise we'd have to keep adding dispatches to the provider like you did here everytime we add a new one.

Redux has a 'combineReducers' funciton built in, it looks like react hooks don't, but it's a pretty small thing to add, this stackoverflow describes some manual solutions, you can see they're quite small:

https://stackoverflow.com/questions/59200785/react-usereducer-how-to-combine-multiple-reducers/61439698#61439698

At the bottom there it suggests a library:

https://github.com/the-road-to-learn-react/use-combined-reducers

See if you can get that library to work, then it should be pretty straight forward.  If it gets too messy we can try something else.

jaredlockhart:
Since we'll be restarting our efforts on the frontend and changing the way we do state management, let's close this for now rather than putting more effort in.

PR #3372, opened 2020-09-02
Add ExperimentRapidFactory fixes #3050

### LINKED ISSUE #3050: Create ExperimentRapidFactory
Since the rapid experiments populate some Experiment fields differently and omit a ton of others, we should make a purpose specific factory just for rapid experiments so we can more effectively test teh APIs and workers and etc with experiments in the states we actually expect them to be.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-198)

jaredlockhart:
So this will introduce a new test factory that only populates the fields that rapid experiments care about, and drops out all the logic that pertains to the legacy experiments.  This makes the rapid tests more succinct and hopefully should remedy some of the intermittent test failures we've seen.  When we split the models we can update this factory to point to the rapid experiment model.


app/experimenter/experiments/tests/api/v3/test_serializers.py
```diff
@@ -262,7 +220,7 @@ def test_serializer_bad_audience_value(self):
             "objectives": "gotta go fast",
             "audience": "WRONG AUDIENCE CHOICE",
             "features": ["picture_in_picture", "pinned_tabs"],
-            "firefox_min_version": FIREFOX_VERSION,
+            "firefox_min_version": "80.0",
```
bbangert, line 223: Why not use the constant or `experiment.firefox_min_version` here?

app/experimenter/experiments/tests/api/v3/test_views.py
```diff
@@ -158,7 +157,7 @@ def test_post_list_creates_rapid_experiment(self):
 
     def test_request_review_updates_status_creates_changelog(self):
         user_email = "user@example.com"
-        experiment = ExperimentFactory.create_with_status(
+        experiment = ExperimentRapidFactory.create_with_status(
             Experiment.STATUS_DRAFT,
             type=Experiment.TYPE_RAPID,
```
bbangert, line 162: Type shouldn't be needed anymore?

app/experimenter/experiments/tests/factories.py
```diff
@@ -216,6 +220,69 @@ def projects(self, create, extracted, **kwargs):
         self.projects.add(*extracted)
 
 
+class ExperimentRapidFactory(ExperimentConstants, factory.django.DjangoModelFactory):
+    type = Experiment.TYPE_RAPID
```
bbangert, line 224: Why is there a type here?

app/experimenter/experiments/tests/factories.py
```diff
@@ -216,6 +220,69 @@ def projects(self, create, extracted, **kwargs):
...
+    features = factory.LazyAttribute(lambda o: list(NIMBUS_DATA["features"].keys()))
+    firefox_min_version = factory.LazyAttribute(
+        lambda o: random.choice(Experiment.VERSION_CHOICES[1:])[0]
+    )
+    firefox_channel = factory.LazyAttribute(
+        lambda o: random.choice(Experiment.CHANNEL_CHOICES[1:])[0]
+    )
+
+    class Meta:
+        model = Experiment
```
bbangert, line 244: Shouldn't this be `ExperimentRapid`?

app/experimenter/experiments/tests/factories.py
```diff
@@ -216,6 +220,69 @@ def projects(self, create, extracted, **kwargs):
...
+            update_experiment_with_change_log(
+                experiment,
+                {"status": status},
+                experiment.owner,
+            )
+
+            if status == target_status:
+                break
+
+        return Experiment.objects.get(id=experiment.id)
```
bbangert, line 283: `ExperimentRapid`?

bbangert (commented):
Looks generally fine, but I'm a bit curious how come the factories don't have the meta set and such.

app/experimenter/experiments/tests/factories.py
bbangert: Ah, its not here yet, I get it.

app/experimenter/experiments/tests/factories.py
jaredlockhart: This is built on the current model definitions, which requires the type be set.  We can take it out in the split pr.

bbangert: Nevermind. :)

app/experimenter/experiments/tests/factories.py
jaredlockhart: That model won't exist when this pr lands.

app/experimenter/experiments/tests/factories.py
jaredlockhart: That model won't exist when this pr lands.

app/experimenter/experiments/tests/api/v3/test_views.py
jaredlockhart: Good catch 👍

app/experimenter/experiments/tests/api/v3/test_serializers.py
jaredlockhart: Ya good idea 👍

jaredlockhart: Oh no that's data going in, there is no `experiment` in that context, it's all just literals.

jaredlockhart: I had that constant up at the top of the file for some reason at some point but it's not generally a pattern we do and I didn't want to encourage it since test constants should be scoped to the test method or test class, so I wanted to get rid of it.  What do you think?  Mind you we're getting rid of these tests when we add graphql soon.

bbangert (approved):

tiftran (approved):

PR #3392, opened 2020-09-08
ADR for Nimbus front-end stack

LZoog:
This ADR regards the "new front-end" mentioned in [ADR 1](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0001-build-nimbus-console-in-experimenter.md). This outlines an app tree directory change and the front-end stack.

### DIFF
--- a/app/experimenter/docs/adrs/0002-nimbus-front-end-directory-tree-and-stack.md
+++ b/app/experimenter/docs/adrs/0002-nimbus-front-end-directory-tree-and-stack.md
@@ -1,0 +1,129 @@
+# Nimbus Front-End: Directory Tree Changes and Use React/CRA, Bootstrap, and GQL
+
+* Status: accepted
+* Deciders: Lauren Zugai, Jody Heavener, Kate Hudson, Jared Lockhart
+* Date: 2020-09-09
+
+## Context and Problem Statement
+
+The front-end for Experimenter is currently composed of two applications. The application currently being used in production for legacy experimentation, survey, and delivery channels including Normandy pref-flips, roll-outs, and heartbeat surveys, lives at `app/experimenter/static/core`. It is composed of a mix of Django templates, vanilla JS, jQuery, and React components, and uses a combination of Bootstrap CSS, `react-bootstrap`, and custom CSS. A newer React application for Nimbus experiments (previously named Rapid experiments) can be found at `app/experimenter/static/rapid` and uses only Bootstrap CSS.
+
+With the decision made to build the Nimbus console as part of the Experimenter codebase ([see ADR #0001](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0001-build-nimbus-console-in-experimenter.md)), it was determined we likely want a "new front-end." After that ADR was merged, through discussions it was agreed upon that [graphene-django](https://github.com/graphql-python/graphene-django) would be used in the back-end, allowing the use of GQL in the Nimbus UI.
+
+This ADR, while slightly atypical as it does not serve to document a single architectural change, outlines a proposed directory tree change regarding the front-end, a proposed new front-end stack, how it compares to the current `rapid` stack, and how we will benefit from these changes.
+
+## Decision Drivers
+
+Drivers for directory tree changes:
+* A more obvious separation of concerns for:
+  * the front-end and back-end
+  * legacy front-end code and Nimbus front-end code
+
+Drivers for the proposed new stack:
+* Development speed
+* Better development tooling which is especially important since we intend to move a lot of application logic to the front-end 
+* A more cohesive fit into the FxA team's stack and processes
+
+Drivers for starting a new application:
+* Time needed and ease of refactoring the current `rapid` React application vs time needed to start with a new application
+
+## Proposed Changes
+
+### App Directory Tree
+
+Currently, a `static/` directory contains the front-end applications with some shared assets, and a `templates/` directory lives adjacent to it.  To provide a more obvious separation of concerns in the directory tree:
+
+* Rename `static/` to `legacy-ui/` since `static/rapid/` will become obsolete and `static/core/` will be the legacy Normandy platform
+* Move `templates/` into `legacy-ui/` and add a README with historical information
+* Use `create-react-app` to create a new React application in `app/experimenter/nimbus-ui/` with a README for development documentation
+
+### Front-End Stack
+
+* Continue using React, but:
+  * Use the `create-react-app` toolchain which gives us Jest, eslint, HMR, and more out of the box
+  * Use Typescript over Proptypes for more strict typing
+  * Use GraphQL for API calls and `apollo-client-cache` for global state management
+  * Use `reach-router` instead of `react-router`
+  * Use directories for components instead of filenames
+    * The Storybook and test file will live adjacent to the component `index.tsx` file which will allow all associated component files to live in the same directory and makes imports shorter.
+    * Keep images inside of the component directory where used until they need to be shared across multiple components
+    * Maintain a flat `src/components/` directory structure for easy imports and file finding, preferring a `NounDescriptor` naming approach to keep similar files together in the tree, such as `PageExperimentForm/` and `PageExperimentDetails/`
+* Continue using Bootstrap (Bootstrap CSS core and `react-bootstrap`), but:
+  * Add stylelint for SCSS linting overrides and document that any custom overrides should be kept very minimal
+  * Use breakpoints established in FxA
+* Utilize Storybook for browsing the UI in various states without running the stack
+* Add Sentry to catch client-side errors
+
+## Considered Alternatives
+
+* Further front-end and back-end directory tree separation with `app/client/` and `app/server/` directories
+* Refactor existing `rapid` project rather than use CRA for `nimbus-ui`
+* Use existing API calls instead of switching to GQL
+* Use Tailwind instead of Bootstrap
+
+### Positive Consequences
+
+Building the front-end application inside of `app/experimenter/nimbus-ui` and moving the `core` application and templates into `app/experimenter/legacy-ui` allows for better separation of concerns (front-end/back-end, legacy/Nimbus front-end code) while continuing to serve the front-end inside of the Django application. Continuing to use React and Bootstrap while starting a new app with `create-react-app` and GraphQL/Apollo allows us to benefit from stack and tooling improvements while avoiding an entire rewrite of all front-end components.
+
+Some members of the FxA team are rolling off of a React/GQL project, `fxa-settings`, with a similar front-end stack and [good documentation](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings). This project can be referenced for using CRA, Storybook examples, how to set up and use the apollo cache, testing helpers, and testing examples using GQL queries and mutations.
+
+### Negative Consequences
+
+This approach for `nimbus-ui` will require time to reach parity to the current `rapid` project. 
+
+## Pros and Cons of the Considered Alternatives
+
+### Further FE and BE separation
+
+The team considered further isolating the front-end and back-end with a more top-level separation, taking `app/experimenter/nimbus-ui/` proposed in this ADR and placing it instead in `app/client/`, with the back-end moving from `app/experimenter/` to `app/server/experimenter/`. Since the React application would then live outside the top-level Django, we would need to serve the front-end with nginx and several changes would need to be made for local dev and production.
+
+At the time of writing, this has been decided against due to additional set up time required, but could potentially be done at a later time. [See the discussion held here](https://github.com/mozilla/experimenter/pull/3392#discussion_r485063856) for more details.
+
+### Refactor existing project
+
+* Pros
+  * Less time spent on new project scaffolding
+* Cons
+  * Because the `rapid` project is small, this may cause points of frustration and more time than expected shuffling files around, refactoring, modifying tests, and reconnecting the pieces with a new global state management approach compared to starting anew
+  * We desire to use the build pipeline offered and managed by `create-react-app` regardless (see [FxA's React Toolchain ADR](https://github.com/mozilla/fxa/blob/main/docs/adr/0013-react-toolchain-for-settings-redesign.md)) and it may be awkward porting an existing application over
+  * Refactoring instead of "starting fresh" provides a weaker sense of collective ownership from new contributors and excitement for fresh perspectives
+
+### Don't use GQL
+
+* Pros
+  * Less additional work on the front-end and back-end initially
+  * Non-FxA engineers would likely experience a learning curve switching to GQL
+* Cons
+  * As the application grows, we will likely need to set up Redux or grow the current ad-hoc global state management
+  * We won't reap the benefits of a GQL application (see [FxA's GQL ADR](https://github.com/mozilla/fxa/blob/main/docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md)) such as reduced network requests and data transferred, clarity around expected data and server responses, client-side caching and global state management via `apollo-client-cache`, helpful developer tooling, potentially [GQL subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/), and speedier development in the long run
+
+### Use Tailwind instead of Bootstrap
+
+* Bootstrap
+  * Pros
+    * Offers a complete UI kit/library to aid in accelerated component styling, leading to less time spent on a custom UI and less time needed on style polishing
+    * Is already being used in the `rapid` project, leading to less component refactoring. With a new `nimbus-ui` application, bits and pieces can likely be copied over
+  * Cons
+    * Isn’t being used in FxA and prevents any potential React component sharing down the line with FxA if we make `fxa-react` more widely available
+    * Is another CSS framework and style pattern for newcomers to pick up on and has a steeper learning curve than other CSS solutions because it offers more
+    * Is a little bulky - we won’t use everything it offers, but this is a much less important consideration for an internal tool like Nimbus
+    * Lacks customization options without forking our own copy or adding style overrides
+* Tailwind
+  * Pros
+    * Has already been setup in `fxa-admin-panel`, `fxa-settings`, and `fxa-react`, meaning 1) it should be straightforward to setup in Nimbus and 2) gives us an option to potentially share React components with FxA (if we make `fxa-react` more widely available)
+    * Developers using it in FxA have given positive feedback on its ease of use
+    * `fxa-settings` has a design guide available for reference and docs written for FxA styling component conventions
+    * Allows us more flexibility without overriding styles on what the overall visual design of the UI will end up as
+  * Cons
+    * Non-FxA engineers may need to ramp up since the `rapid` project is currently using Bootstrap
+    * Will require significant refactoring from what already exists in the `rapid` project
+    * Is a much less complete solution than Bootstrap/`react-bootstrap`, leading to much more time spent on custom UI classes and our own form components
+
+Ultimately, Bootstrap was chosen to stick around because Bootstrap CSS and `react-bootstrap` offer a lot of form (and other) components and styles out of the box and the "development speed" decision driver holds significant weight for this project.
+
+## Links
+
+[ADR #0001](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0001-build-nimbus-console-in-experimenter.md) - documented decision to build Nimbus in this repository with the existing Django back-end\
+[graphene-django](https://github.com/graphql-python/graphene-django) - package to allow GraphQL in Python/Django\
+[FxA Settings README](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings) - well-documented project with a similar front-end stack that can be referenced for examples with other relevant ADRs (GQL, React Toolchain) linked at the top
+


app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,112 @@
+# Use React, Bootstrap, CRA, and GQL for the Nimbus Front-End Stack
+
+* Status: proposed
+* Deciders: Lauren Zugai, Jody Heavener, [TBD]
+* Date: 2020-09-08
+
+## Context and Problem Statement
+
+The front-end for Experimenter is currently composed of two applications. A more more mature application for Normandy experiments, found at `app/experimenter/static/core`, is composed of a mix of Django templates, vanilla JS, jQuery, and React components. A newer React application for Nimbus experiments (previously named Rapid experiments) can be found at `app/experimenter/static/rapid`.
```
chelsea-lewis, line 9: Remove duplicate "more"

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+The front-end for Experimenter is currently composed of two applications. A more mature application for Normandy experiments, found at `app/experimenter/static/core`, is composed of a mix of Django templates, vanilla JS, jQuery, and React components. A newer React application for Nimbus experiments (previously named Rapid experiments) can be found at `app/experimenter/static/rapid`.
+
+With the decision made to build the Nimbus console as part of the Experimenter codebase ([see ADR #0001](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0001-build-nimbus-console-in-experimenter.md)), it was determined we likely want a "new front-end." This ADR, while slightly atypical as it does not serve to document a single architectural change, outlines a proposed directory tree change regarding the front-end, a proposed new front-end stack, how it compares to the current `rapid` stack, and how we will benefit from these changes.
+
+## Decision Drivers
+
+Drivers for directory tree changes:
+* A more obvious separation of concerns for:
+  * the front-end and back-end
+  * legacy front-end code and Nimbus front-end code
```
jodyheavener, line 18: Perhaps a naive question, but what are our intentions with legacy front-end Nimbus code? Assuming it is no longer needed with the new UI, does it need to stay in the repo or can it be purged [once the new UI is fully built, tested, deployed etc]?

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+
+### App Directory Tree
+
+Currently, a `static/` directory contains the front-end applications with some shared assets, and a `templates/` directory lives adjacent to it.  To provide a more obvious separation of concerns in the directory tree:
+
+* Rename `static/` to `legacy-ui/` since `static/rapid/` will become obsolete and `static/core/` will be the legacy Normandy platform
+* Move `templates/` into `legacy-ui/`
+* Use `create-react-app` to create a new React application in `app/experimenter/nimbus-ui/`
+
+
```
jodyheavener, line 38: nit: extra newline

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+* Time needed and ease of refactoring the current `rapid` React application vs time needed to start with a new application
+
+## Proposed Changes
+
+### App Directory Tree
+
+Currently, a `static/` directory contains the front-end applications with some shared assets, and a `templates/` directory lives adjacent to it.  To provide a more obvious separation of concerns in the directory tree:
+
+* Rename `static/` to `legacy-ui/` since `static/rapid/` will become obsolete and `static/core/` will be the legacy Normandy platform
+* Move `templates/` into `legacy-ui/`
```
jodyheavener, line 35: I might suggest throwing a README inside `legacy-ui` as well to document what that directory is, why it's there, etc. Maybe this ADR suffices, I just know there's a lot happening in `app/experimenter/`, so any clarification is nice to have.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Considered Alternatives
+
+* Refactor existing `rapid` project rather than use CRA for `nimbus-ui`
+* Use existing API calls instead of switching to GQL
+* Use Tailwind instead of Bootstrap
+
+### Positive Consequences
+
+Continuing to use React and Bootstrap while starting a new app with `create-react-app` and GraphQL/Apollo allows us to benefit from stack and tooling improvements while avoiding an entire rewrite of all front-end components.
```
jodyheavener, line 62: Worth noting that [some](https://github.com/mozilla/experimenter/blob/main/app/experimenter/static/core/js/components/MessageForm.js#L29:L29) of the React is written with `react-bootstrap` while [some](https://github.com/mozilla/experimenter/blob/main/app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx#L198:L198) is written with Bootstrap CSS. This may impact component re-usability depending on desired code consistency.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+* Bootstrap
+  * Pros
+    * Offers a complete UI kit/library to aid in accelerated component styling, leading to less time spent on a custom UI and less time needed on style polishing
+    * Is already being used in the `rapid` project, leading to less component refactoring. With a new `nimbus-ui` application, bits and pieces can likely be copied over
+  * Cons
+    * Isn’t being used in FxA and prevents any potential React component sharing down the line with FxA if we make `fxa-react` more widely available
+    * Is another CSS framework and style pattern for newcomers to pick up on and has a steeper learning curve than other CSS solutions because it offers more
+    * Is a little bulky - we won’t use everything it offers, but this is a much less important consideration for an internal tool like Nimbus
+    * Lacks customization options without forking our own copy or adding style overrides
+  * Tailwind
```
jodyheavener, line 101: Is this unintentionally nested under Bootstrap?

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+  * Less additional work on the front-end and back-end initially
+  * Non-FxA engineers would likely experience a learning curve switching to GQL
+* Cons
+  * As the application grows, we will likely need to set up Redux or grow the current ad-hoc global state management
+  * We won't reap the benefits of a GQL application (see (FxA's GQL ADR)[https://github.com/mozilla/fxa/blob/main/docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md]) such as reduced network requests and data transferred, clarity around expected data and server responses, client-side caching and global state management via `apollo-client-cache`, helpful developer tooling, and speedier development in the long run
+
+### Use Tailwind instead of Bootstrap
+
+* Bootstrap
+  * Pros
```
jodyheavener, line 93: \+ Bootstrap's [utility classes](https://getbootstrap.com/docs/4.5/utilities/), if we should choose to use them, are somewhat similar to Tailwind classes (just a little more verbose). Reduced learning curve. There are also a multitude of IDE extensions (like [VS Code](https://marketplace.visualstudio.com/items?itemName=Zaczero.bootstrap-v4-snippets)).

jodyheavener (commented):
Looking great. I'll probably do another pass in a bit, but left a few comments in the meantime. Overall I'm agreement with the proposed decision, though I acknowledge I'm biased towards this stack, so I'd love to hear what others think.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+### Front-End Stack
+
+* Continue using React, but:
+  * Use the `create-react-app` toolchain
+  * Use Typescript over Proptypes for more strict typing
+  * Use GraphQL for API calls and `apollo-client-cache` for global state management
+  * Use `reach-router` instead of `react-router`
+  * Use directories for components instead of filenames
+    * The Storybook and test file will live adjacent to the component `index.tsx` file. This allows all associated component files to live in the same directory and makes imports shorter.
+* Continue using Bootstrap (`react-bootstrap`), but:
```
k88hudson, line 48: What's the strategy for including custom css when necessary?

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Proposed Changes
+
+### App Directory Tree
+
+Currently, a `static/` directory contains the front-end applications with some shared assets, and a `templates/` directory lives adjacent to it.  To provide a more obvious separation of concerns in the directory tree:
+
+* Rename `static/` to `legacy-ui/` since `static/rapid/` will become obsolete and `static/core/` will be the legacy Normandy platform
+* Move `templates/` into `legacy-ui/`
+* Use `create-react-app` to create a new React application in `app/experimenter/nimbus-ui/`
```
jodyheavener, line 36: Is there any chance of moving this up under `/app` so it's not so heavily nested? This could also help make the distinction between front-end and server-side concerns.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jodyheavener: Oh yes +1 to this. With Tailwind we determined that there is almost no reason to use truly custom CSS because a utility class existed for every need (unless the custom selector is being used to stack utility classes). @LZoog because we'll undoubtedly need utility classes, I would say a con might be not being able to hook into PurgeCSS.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+  * Use GraphQL for API calls and `apollo-client-cache` for global state management
+  * Use `reach-router` instead of `react-router`
+  * Use directories for components instead of filenames
+    * The Storybook and test file will live adjacent to the component `index.tsx` file. This allows all associated component files to live in the same directory and makes imports shorter.
+* Continue using Bootstrap (`react-bootstrap`), but:
+  * Add stylelint for SCSS linting overrides
+  * Use breakpoints established in FxA
+* Utilize Storybook for browsing the UI in various states without running the stack
+* Add Sentry to catch client-side errors
+
```
jodyheavener, line 53: I see the existing `rapid` setup is using `experiments`, `forms` and `pages` to organize components by type. This type of organization was recently discussed for an FxA React project, so I'm curious if we'll be employing the same structure with the new app. My personal preference is a flat directory of well-named components.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: I'm thinking it'll be deleted since it won't be used.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: 👍 Worth doing. We should also keep a README in `nimbus-ui` up to date and easy to work with like Settings IMO.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: @bbangert @jaredlockhart Since the Django app is `app/experimenter/`, would moving a React application into `app/nimbus-ui/` (and building into `app/nimbus-ui/dist/`) make serving it more difficult or would it be easy to point it here?

I like the idea of more FE and BE separation so I'd be fine with that if it doesn't complicate much. Though, one could argue the UI sits on top of `experimenter` and since the legacy UI will be in `/experimenter`, maybe it makes sense to leave it there. 🤷

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: I mostly included this piece because 1) under the `core` Normandy UI, we have files like `adminextra.css` where overrides are added and 2) `react-bootstrap` offers some customization options in SCSS like so: https://react-bootstrap.github.io/getting-started/introduction/#customize-bootstrap

Any custom SCSS should be kept to a minimum for maintenance purposes. 

>  I would say a con might be not being able to hook into PurgeCSS.

Ah yeah, I'll add that as an extension of this con I have for Bootstrap: "* Is a little bulky - we won’t use everything it offers, but this is a much less important consideration for an internal tool like Nimbus"

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jodyheavener: Yeah, totally fair! Y’all have far more insight into what would make sense here 🙂

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: While I'm more indifferent to this, it was clear the FxA team likes the flat structure, so I can include that as an additional point under "Use directories for components instead of filenames" above

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: Huh, so actually the `core` project uses `react-bootstrap` while the `rapid` project uses Bootstrap CSS. I should better differentiate the two so good call out.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jaredlockhart: @LZoog Yeah so it depends on how we intend to serve it.  Right now all the React parts in the legacy and nimbus UIs are hosted using the Django static assets system and the [whitenoise module](http://whitenoise.evans.io/en/stable/), so they have to live inside the Django tree `app/experimenter/` so Django can find them.  If we want to host them separately, like directly with nginx, they could live outside the tree, but that would involve changing the way the production deployment is configured.  Either is doable, just needs some figuring out.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: Nope, just a spacing mishap :P

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: I'm curious if @tiftran or @jaredlockhart have a strong preference here. Check out our [component directory structure](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings/src/components) for `fxa-settings`, we're generally using a `nounDescriptor` approach to keep similar files like Inputs and Pages grouped together in the tree but keep all of them one directory deep under `components`. Makes it easy to scan what's available while keeping imports simple.

FWIW we've separated "Pages" out in our Storybook which [you can see here](https://mozilla-fxa.github.io/storybooks/commits/5f2ad51fadd80ea3afef726ea7d66b2eecdc8987/fxa-settings/index.html?path=/story/pages-settings--cold-start).

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
k88hudson: Once nimbus can replace all experimentation that goes through the old UI, it can be removed. We will need to make the old data available somehow either through migrating it or having some kind of statically accessible archive.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
k88hudson: We had an issue for starting to use `react-boostrap` (https://github.com/mozilla/experimenter/issues/2942) but it ended up getting blocked on a dependency issue on `react-bootstrap`'s side (@tiftran or @jaredlockhart do you happen to have an issue reference on react-bootstrap for this?)

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
+# Use React, Bootstrap, CRA, and GQL for the Nimbus Front-End Stack
+
+* Status: proposed
+* Deciders: Lauren Zugai, Jody Heavener, [TBD]
+* Date: 2020-09-08
+
+## Context and Problem Statement
+
+The front-end for Experimenter is currently composed of two applications. A more mature application for Normandy experiments, found at `app/experimenter/static/core`, is composed of a mix of Django templates, vanilla JS, jQuery, and React components. A newer React application for Nimbus experiments (previously named Rapid experiments) can be found at `app/experimenter/static/rapid`.
```
k88hudson, line 9: Experimenter does handle non-normandy experiments as well as non-experiment tasks (roll-outs), so I'd probably describe this as `an application currently being used in production for legacy experimentation, survey, and delivery channels including Normandy pref-flips, roll-outs, and heartbeat surveys`

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jaredlockhart: Yeah teh problem was with their dependency on popper and some TS bug it had:

https://github.com/mozilla/experimenter/pull/3208#issuecomment-673050527

I was waiting for them to fix it on their side, not sure if it's been resolved yet I don't have a link handy to the issue on their side.

jaredlockhart: Might be this:

https://github.com/popperjs/react-popper/issues/373

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+
+Drivers for directory tree changes:
+* A more obvious separation of concerns for:
+  * the front-end and back-end
+  * legacy front-end code and Nimbus front-end code
+
+Drivers for the proposed new stack:
+* Better development tooling
+* A more cohesive fit into the FxA team's stack and processes
+* Development speed
```
k88hudson, line 23: One thing that might be worth noting is that we intend on moving a lot of the application logic to the front-end v.s. the old back-end-heavy architecture

k88hudson (approved):
Thanks for writing this up! Looks good to get started, just a few small questions inline.

I didn't see testing/linting mentioned, can you add a brief point about Jest/eslint + I'm assuming prettier? Will you use react testing-library like the [existing tests](https://github.com/mozilla/experimenter/tree/main/app/experimenter/static/rapid/__tests__) or something else?

jodyheavener:
@k88hudson 

> I didn't see testing/linting mentioned, can you add a brief point about Jest/eslint + I'm assuming prettier? Will you use react testing-library like the existing tests or something else?

CRA comes with Jest and react-test-library baked in, so I'm fully in favor of sticking with that (though I might advocate for placing tests adjacent to component file). It also includes ESLint and extends the react-scripts preset. 

@LZoog with the CRA route, do you think we'll need to be able to import components from outside our `src` dir? If so, or if we plan to heavily modify Jest or any Webpack-level config in CRA we'll need to consider Rescripts, like we have with FxA.

k88hudson:
> CRA comes with Jest and react-test-library baked in, so I'm fully in favor of sticking with that (though I might advocate for placing tests adjacent to component file). It also includes ESLint and extends the react-scripts preset.

Perfect, thanks. +1 to having tests adjacent to components unless it makes things really difficult to set up.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+* Move `templates/` into `legacy-ui/`
+* Use `create-react-app` to create a new React application in `app/experimenter/nimbus-ui/`
+
+
+### Front-End Stack
+
+* Continue using React, but:
+  * Use the `create-react-app` toolchain
+  * Use Typescript over Proptypes for more strict typing
+  * Use GraphQL for API calls and `apollo-client-cache` for global state management
```
lmorchard, line 44: I might be missing context, but "Use GraphQL for API calls" seems to go beyond front-end stack concerns for this ADR. 

Given that [the Decision Outcome](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0001-build-nimbus-console-in-experimenter.md#decision-outcome) over in ADR 0001 was to build on top of Experimenter, these GQL APIs would need be changes or new APIs in the Django app. Could also be supported by a node/TS GQL proxy server, but that's probably unwarranted complexity.

We have FxA-team experience with GQL in node/TS - have we looked into the same under Python & Django?

lmorchard (commented):
Not sure if this rises to a 'request changes' level. Noticed the bit about GQL but wondering about how we plan to implement GQL in Python / Django

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
k88hudson: Might be worth clarifying in the ADR – but we did agree to go with [graphene](https://github.com/graphql-python/graphene-django) for the back end. I think Jared and Ben are working on filing some of that this week

LZoog: This is a great call-out as I didn't realize the first ADR doesn't even mention GQL. I agree we should expound upon this a bit here.

@bbangert, I know you'd mentioned looking into [graphene-django](https://github.com/graphql-python/graphene-django). Does it seem like it's a relatively low level of effort to implement?

lmorchard: > Might be worth clarifying in the ADR – but we did agree to go with graphene for the back end.

Yeah, I don't see mentions of graphene in this ADR or 0001, so that sounds like an Architectural Decision that wasn't Recorded. :) Otherwise, I'm scratching my head reading this as trying to further leverage FxA experience with GQL in node/TS that's not exactly applicable here.

LZoog:
> I didn't see testing/linting mentioned, can you add a brief point about Jest/eslint + I'm assuming prettier? Will you use react testing-library like the existing tests or something else?

Yeah, Jody beat me to it. I can mention that CRA offers us Jest CRA comes react-test-library out of the box. Both `core` and `rapid` have an eslint file so I didn't mention that being a "change," although it does look like we could use prettier (I'll add something about that).

> ... though I might advocate for placing tests adjacent to component file

I've got that covered with this here 😁:
> * Use directories for components instead of filenames
    * The Storybook and test file will live adjacent to the component `index.tsx` file. This allows all associated component files to live in the same directory and makes imports shorter.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+  * Refactoring instead of "starting fresh" provides a weaker sense of collective ownership from new contributors and excitement for fresh perspectives
+
+### Don't use GQL
+
+* Pros
+  * Less additional work on the front-end and back-end initially
+  * Non-FxA engineers would likely experience a learning curve switching to GQL
+* Cons
+  * As the application grows, we will likely need to set up Redux or grow the current ad-hoc global state management
+  * We won't reap the benefits of a GQL application (see (FxA's GQL ADR)[https://github.com/mozilla/fxa/blob/main/docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md]) such as reduced network requests and data transferred, clarity around expected data and server responses, client-side caching and global state management via `apollo-client-cache`, helpful developer tooling, and speedier development in the long run
```
jaredlockhart, line 88: +1 for using graphql

jaredlockhart (commented):
Thanks for writing this up @LZoog !

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: Sweet, yeah I think `directory/index.tsx` especially makes sense with the introduction of Storybook files. Keeping the test file and Storybook file adjacent to the component file has been super handy, and we've also been adding images to these directories when they're only used in that component (and then we move them into a [shared `images/` directory](https://github.com/mozilla/fxa/tree/main/packages/fxa-react/images) when needed). The [Settings Nav](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings/src/components/Nav) is a good example.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: If we didn't use `react-bootstrap` wouldn't we miss out on a lot of benefits like what's in https://react-bootstrap.github.io/components/forms/, or maybe we could just use those as references to build our own? I don't think using the last-working stable version then bumping is too horrible considering this issue seems fairly new, unless this is something you've ran into multiple times.

jaredlockhart: No I mean you can just use bootstrap in the old fashioned way and say stuff like

```html
<div className="form">
  <div className="form-group">
```

rather than the react-bootstrap way of

```jsx
<Form>
  <FormGroup>
```

you lose the nice auto complete and properties and stuff, but it's basically 1:1 to convert them from one to the other when they fix the ts bug in popper.

jaredlockhart: Or maybe it's already fixed 🤞

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
```diff
@@ -0,0 +1,113 @@
...
+* Development speed
+
+Drivers for starting a new application:
+* Time needed and ease of refactoring the current `rapid` React application vs time needed to start with a new application
+
+## Proposed Changes
+
+### App Directory Tree
+
+Currently, a `static/` directory contains the front-end applications with some shared assets, and a `templates/` directory lives adjacent to it.  To provide a more obvious separation of concerns in the directory tree:
```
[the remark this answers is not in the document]
LZoog: Yeah, TBH I like both of those suggestions (probably the latter the best), though I hesitate due to needing to serve with nginx just for project structure. Regarding "client," I actually called it "client" at first, but I think @k88hudson said "client" sounds like the SDK. 🤔

I think `app/client/` and `app/server/` is pretty slick, but are there any long-term downsides to using nginx to serve the application besides an initial setup and is "this is slightly better for separation of concern" a good enough reason to justify it (and the sweeping filename changes from `experimenter` to `server`)?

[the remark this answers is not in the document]
jaredlockhart: Well for the server I think we'd just move the entire `app/experimenter/` dir to `app/server/experimenter/` whole sale so we woudln't need to rename anything.  We'd only have to rename things if we did it the other way around `app/experimenter/server/` if we wanted to put client inside the Django top level.  If we put the client outside the Django top level to serve it with nginx we don't need to rename anything.

There's nothing stopping us from putting the entire frontend outside of Django and only letting nginx know about it, but we'd have to change a couple steps for local dev and prod

1) Expose the built assets (`dist`) in a shared volume to the nginx container
2) Add a rule to nginx to host that on some path
3) Hard code that path in the templates to initialize it like 
```html
  <script src="/<nginx-nimbus-asset-path>/index.js"></script>
```

 rather than using the Django static asset path resolver `{% static "assets/rapid/main.js" %}` see [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/templates/experiments/rapid.html#L10)

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: For someone that just got started with the project, I honestly was a little confused by (well firstly where to find the front-end, ha) `core` and `rapid`. I think if we stuck with those that could be mitigated by a quick explanation in the root README, but:

>  I'd suggest we also rename any other references in the code that currently use 'core' and 'rapid',

I'd be fine with us doing this personally. I was going to suggest that a happy medium could potentially be changing `rapid` to `nimbus` and leaving `core` as-is, but as "nimbus" is to be the shiny new centralized location for things, maybe "core" would be confusing later once Nimbus is built out?

jaredlockhart: Yeah and `core` gets extra confusing now that we're trying to split up the models and in theory you have something like `ExperimentBase` as the base model and `ExperimentCore` `ExperimentRapid` as the children models, `core` and `base` are ambiguous 😆 

Whoever said that naming was the hardest problem in software 🧐

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: This conversation is continued here: https://github.com/mozilla/experimenter/pull/3392/files#r485158559

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jaredlockhart: I was trying to avoid the name nimbus in the code becuase it's the code name, but it could also be just as confusing to see 'rapid' everywehre when we know its' called 'nimbus' so yeah maybe renaming everything to 'Legacy' and 'Nimbus' is the right call.  I'll have to `awk/sed` this whole model pr, or maybe we can do that as a separate pr.....

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jodyheavener: Worth mentioning this in this ADR, or possibly in the pending [README](https://github.com/mozilla/experimenter/pull/3392#discussion_r485043805)?

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jodyheavener: > I was waiting for them to fix it on their side, not sure if it's been resolved yet I don't have a link handy to the issue on their side.

That PR confuses me. I commented in it to see if it's fixed or not. Though it looks like at time of open the latest version was 1.3.7, but they're on 2.2.3 now.

jodyheavener: Oh wait... they [dropped popper](https://github.com/react-bootstrap/react-bootstrap/commit/5bcfff2e1a6c8658f49be9c439e8dc447bef96b5).

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jodyheavener: Oh actually, it might be a bit of work, but [here](https://lukelowrey.com/automatically-removeunused-css-from-bootstrap-or-other-frameworks/) are instructions on how to set up PurgeCSS with Bootstrap. Looks like we'd need access to Webpack though.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: Ah, that makes sense on the "nimbus" naming hesitation since it's a code name. That didn't even cross my mind honestly since I've only ever seen it called that. 😅

I do personally like "nimbus" and "legacy" but I'm still open to other suggestions. 👍

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
LZoog: Eh yeah, I'm not sure it'd be worth it since it's an internal tool - we'd save like what, 50-100KB max? For a user-facing app I'd say let's do it for sure but I'm less inclined here to spend the effort setting it up. Worth filing as a "nice to have" though.

jodyheavener: Rescripts would be helpful. But I also agree that the time to implement may not be worth it initially. Seems like a good candidate for some downtime or as a refinement.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
jaredlockhart: Looking at this model refactor PR more, I don't think it'll be feasible (or imminently practical) to rename the entire Python code base, but I also agree that 'core' and 'rapid' are confusing and we don't necessarily need to be locked into them.  I'd say disregard what I said earlier about maintaining consistency and use the names you prefer 👍

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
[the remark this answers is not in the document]
LZoog: @jaredlockhart I'm thinking at present, it may not be worth serving the FE with nginx for this degree of FE/BE separation due to the additional work needed. I personally am OK with `app/experimenter/nimbus-ui`, as in my mind it solves my initial "I want more separation between FE/BE + nimbus/legacy" problem. We could also potentially do this later if we wanted to. I'm writing up a section in this ADR about considering this, thank you for the breakdown. Feel free to disagree (@jodyheavener  might argue it's worth it). 🙂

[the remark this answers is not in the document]
jaredlockhart: That would be the simplest approach to start 👍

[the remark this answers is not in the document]
jodyheavener: Yeah, I generally think it would be good to maintain a clear separation, and I'm sure there are merits to doing it out of the gate versus refactoring later, but I agree that this might be more effort than it's currently worth if we're trying to reach MVP sooner rather than later. I'm cool with `app/experimenter/nimbus-ui`.

app/experimenter/docs/adrs/0002-use-react-bootstrap-cra-gql-for-nimbus.md
lmorchard: I think this has been resolved in updates to the ADR

jaredlockhart (approved):
Looks great @LZoog !  Thanks for putting this together 🎉

jodyheavener (approved):
I love this! Thanks for all the discussion.

pdehaan:
For anybody else curious what the current directory structure is:

```sh
tree app -d -I 'node_modules|tests|__tests__'

app
├── bin
├── experimenter
│   ├── base
│   │   ├── fixtures
│   │   ├── management
│   │   │   └── commands
│   │   └── migrations
│   ├── bugzilla
│   ├── docs
│   │   └── adrs
│   ├── experiments
│   │   ├── api
│   │   │   └── {v1-v4}
│   │   ├── migrations
│   │   └── templatetags
│   ├── kinto
│   ├── normandy
│   ├── notifications
│   │   └── migrations
│   ├── openidc
│   ├── projects
│   │   └── migrations
│   ├── static
│   │   ├── assets
│   │   │   ├── css
│   │   │   └── js
│   │   ├── core
│   │   │   ├── css
│   │   │   ├── imgs
│   │   │   └── js
│   │   │       ├── components
│   │   │       └── utils
│   │   ├── rapid
│   │   │   ├── components
│   │   │   │   ├── experiments
│   │   │   │   ├── forms
│   │   │   │   └── pages
│   │   │   ├── contexts
│   │   │   │   └── experiment
│   │   │   └── types
│   │   └── scripts
│   └── templates
│       ├── admin
│       ├── experiments
│       │   └── emails
│       ├── projects
│       └── widgets
└── requirements

51 directories
```

LZoog:
@pdehaan Thanks for that comment! Helps visualize it a bit better. The only thing to note is that `legacy-ui` will still contain `core` and `rapid`. When `nimbus-ui` reaches parity with `rapid`, we should be able to delete `rapid`. And then, once Nimbus can replace all experimentation moving forward and we either make legacy experiment data available in Nimbus or archive it elsewhere ([see thread](https://github.com/mozilla/experimenter/pull/3392#discussion_r485035976)), we can then delete `legacy-ui`.

lmorchard:
🎉🎉🎉

PR #3395, opened 2020-09-09
Fix v3 intermittent test failure fixes #3394

### LINKED ISSUE #3394: Intermittent test failure: V3 API
```
=================================== FAILURES ===================================
_ TestExperimentRapidSerializer.test_serializer_updates_experiment_and_creates_changelog _
[gw2] linux -- Python 3.8.2 /usr/local/bin/python

self = <experimenter.experiments.tests.api.v3.test_serializers.TestExperimentRapidSerializer testMethod=test_serializer_updates_experiment_and_creates_changelog>

    def test_serializer_updates_experiment_and_creates_changelog(self):
        owner = UserFactory(email="owner@example.com")
        experiment = ExperimentRapidFactory.create_with_status(
            Experiment.STATUS_DRAFT,
            owner=owner,
            name="rapid experiment",
            slug="rapid-experiment",
            objectives="gotta go fast",
            audience="us_only",
            features=["picture_in_picture", "pinned_tabs"],
            firefox_channel=Experiment.CHANNEL_RELEASE,
        )
        experiment.variants.all().delete()
        variant = ExperimentVariantFactory.create(experiment=experiment)
    
        self.assertEqual(experiment.changes.count(), 1)
        data = {
            "name": "changing the name",
            "objectives": "changing objectives",
            "audience": "all_english",
            "features": ["pinned_tabs"],
            "firefox_channel": Experiment.CHANNEL_NIGHTLY,
            "firefox_min_version": Experiment.VERSION_CHOICES[1][0],
            "variants": [
                {
                    "id": variant.id,
                    "name": "something else",
                    "description": "something",
                    "is_control": True,
                    "ratio": 50,
                    "value": "something",
                },
                {
                    "name": "variant1",
                    "description": "variant1 description",
                    "is_control": False,
                    "ratio": 50,
                    "value": "variant value",
                },
            ],
        }
    
        serializer = ExperimentRapidSerializer(
            instance=experiment, data=data, context={"request": self.request}
        )
    
        experiment = serializer.save()
        self.assertEqual(experiment.changes.count(), 2)
    
        changelog = experiment.changes.latest()
        self.assertEqual(changelog.old_status, Experiment.STATUS_DRAFT)
        self.assertEqual(changelog.new_status, Experiment.STATUS_DRAFT)
>       self.assertEqual(
            set(changelog.changed_values.keys()),
            set(
                [
                    "audience",
                    "features",
                    "firefox_channel",
                    "firefox_min_version",
                    "name",
                    "objectives",
                    "public_description",
                    "variants",
                ]
            ),
        )
E       AssertionError: Items in the second set but not the first:
E       'firefox_min_version'

experimenter/experiments/tests/api/v3/test_serializers.py:433: AssertionError
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-254)

jaredlockhart:
Oops! Always declare all the expected inputs to your test factory!

### DIFF
--- a/app/experimenter/experiments/tests/api/v3/test_serializers.py
+++ b/app/experimenter/experiments/tests/api/v3/test_serializers.py
@@ -388,6 +388,7 @@ class TestExperimentRapidSerializer(MockRequestMixin, MockBugzillaTasksMixin, Te
             audience="us_only",
             features=["picture_in_picture", "pinned_tabs"],
             firefox_channel=Experiment.CHANNEL_RELEASE,
+            firefox_min_version="79.0",
         )
         experiment.variants.all().delete()
         variant = ExperimentVariantFactory.create(experiment=experiment)
@@ -399,7 +400,7 @@ class TestExperimentRapidSerializer(MockRequestMixin, MockBugzillaTasksMixin, Te
             "audience": "all_english",
             "features": ["pinned_tabs"],
             "firefox_channel": Experiment.CHANNEL_NIGHTLY,
-            "firefox_min_version": Experiment.VERSION_CHOICES[1][0],
+            "firefox_min_version": "80.0",
             "variants": [
                 {
                     "id": variant.id,


bbangert (approved):
This looks fine, I don't entirely understand why these would pass intermittently before though.

jaredlockhart:
@bbangert The factories will generate random data for every field, so if you don't explicitly specify a field it'll still get set just with a random value, and then depending on what the test expects that might randomly be a value that makes the test fail.

In this case we're checking whether a changelog gets created, which only happens if the field value changes, so if we don't explicitly specify the value to the factory, it might randomly choose a value that's the same as the one we're updating it to and then the changelog will skip it, so the fix is to explicitly specify different values for the factory and the updated POST value that way it'll always be caught by the changelog.

The pattern in general is just to remember to always specify all the fields in the factory that you're asserting on later in the test.  If we had one of those fancy AST linters we might be able to create a rule for it XD

Issue #123, opened 2017-06-29
Add invalid state to experiments


jaredkerim:
If an experiment never started and we decided not to do it, it's rejected.  If it started and we found a bug and close it, it's invalid.

Issue #3399, opened 2020-09-09
Change dependabot config to update yarn.lock


jaredlockhart:
When I added the new dependabot config I told it to look in each individual workspace for package.json, but I just learned that that means it doesn't update yarn.lock.  I'm confused about how builds were still proceeding with yarn.lock being out of date? But in theory if we change the dependabot config to point to just the top level it should find the workspaces.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-257)

PR #3419, opened 2020-09-10
Add a visualization package wrapped in a feature flag.

emtwo:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-341)


app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
```diff
@@ -285,6 +285,32 @@ describe("<ExperimentDetails />", () => {
     });
   });
 
+  it("Initializes the svelte code as expected", async () => {
```
jaredlockhart, line 288: Is this just checking that it doens't throw any errors?  Is there any way to assert that the svelte parts are actually rendered?

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -20,6 +20,12 @@ import { ExperimentStatus } from "experimenter-rapid/types/experiment";
 
 export const POLL_TIMEOUT = 30000;
 
+declare global {
+  interface Window {
+    initSvelte(svelteRoot: HTMLElement): void;
+  }
+}
+
```
jaredlockhart, line 28: awesome 👍

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -58,6 +64,15 @@ const ExperimentDetails: React.FC = () => {
...
   const handleClickRequestApproval = async () => dispatch(requestReview());
 
+  React.useEffect(() => {
+    const visualizationRoot = document.getElementById("visualization-root");
+    if (!visualizationRoot || !window.initSvelte) {
+      return;
+    }
+
+    window.initSvelte(visualizationRoot);
+  });
```
jaredlockhart, line 74: Oh there's a thing with useEffect.  By default it runs on every render:

`
The default behavior for effects is to fire the effect after every completed render. That way an effect is always recreated if one of its dependencies changes.
`

But you can pass it a second arg which is the value it checks for changes:

https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

You have it setup now to just return if it's already initialized, but the pattern to make it only run once at startup is to pass an empty array:

`
If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. 
`

so I think if you do that you don't need the conditional check and it'll only run once?

app/experimenter/templates/experiments/rapid.html
```diff
@@ -2,7 +2,15 @@
...
 
+{% load static %}
+
 {% block main %}
+  {% if FEATURE_ANALYSIS %}
+    <script src="{% static "assets/visualization/bundle.js" %}"></script>
+    <link rel="stylesheet" href="{% static "assets/visualization/bundle.css" %}" />
+  {% endif %}
+
+  <link rel="stylesheet" href="{% static "assets/css/index.css" %}">
```
jaredlockhart, line 13: Is this css for the visualization?

jaredlockhart (changes requested):
Looks great @emtwo !  I think if you use the second arg to `useEffect` you can simplify it and prevent re-renders, other than that looks perfect 🎉

app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
jaredlockhart: I guess right now it's not dumping out any content so there's no content to look for, since it's behind a feature flag maybe you can just put a little piece of text or something and assert that that's present?

emtwo: nothing gets rendered in this PR, but yea probably in future PRs we can do that.

Right now, this is just to get code coverage to 100%

app/experimenter/templates/experiments/rapid.html
emtwo: yes

app/experimenter/static/rapid/__tests__/experimentDetails.test.tsx
jaredlockhart: Oh yeah true coverage helps here

app/experimenter/templates/experiments/rapid.html
jaredlockhart: That path/name might collide, is there any way to to something like `assets/css/visualization/index.css`?

jaredlockhart: Or `assets/visualization/index.css`?

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
LZoog: I'm not sure if this matters right now since we're replacing `rapid` soon, but we shouldn't use `document.getElementById` in a React component, using a [ref](https://reactjs.org/docs/refs-and-the-dom.html) is the way to reference elements👍

jaredlockhart: Oh neat!

app/experimenter/templates/experiments/rapid.html
emtwo: sure

app/experimenter/static/visualization/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@tsconfig/svelte": "^1.0.10",
+    "@typescript-eslint/eslint-plugin": "^4.1.0",
+    "@typescript-eslint/parser": "^4.1.0",
+    "eslint": "^7.8.1",
+    "eslint-plugin-import": "^2.20.2",
+    "eslint-plugin-prettier": "^3.1.3",
+    "eslint-plugin-svelte3": "^2.7.3",
+    "prettier": "^2.0.5",
+    "prismjs": "^1.20.0",
+    "rimraf": "3.0.2",
```
pdehaan, line 29: Seems to be the only devDep without a semver range operator (`^`).
I know we're planning a refactor of the nested app/experimenter/ directories, but we should also make sure that dependabot also picks up on this package.json file.

app/experimenter/static/visualization/rollup.config.js
```diff
@@ -0,0 +1,66 @@
+import commonjs from "@rollup/plugin-commonjs";
+import resolve from "@rollup/plugin-node-resolve";
+import node from "@rollup/plugin-node-resolve";
```
pdehaan, line 3: Looks like we're importing the same dependency twice with different aliases?

app/experimenter/templates/experiments/rapid.html
```diff
@@ -2,7 +2,15 @@
 {% load static%}
 
 
+{% load static %}
```
pdehaan, line 5: This is called above on L2. Does it need to be called twice? Sorry, I'm not totally certain what these template shortcodes are doin.

jaredlockhart: Oh good catch! Yeah you shouldn't need to do that twice.

app/experimenter/static/visualization/package.json
jaredlockhart: Yeah good call @pdehaan , I think doing this change will make dependabot pick up on all the frontend dependencies with a single config: https://github.com/mozilla/experimenter/issues/3399 so we can probably skip adding a specific dependabot config here.

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
emtwo: So, I've updated this to use a ref. However, using the `[]` as a second param for `useEffect()` causes it to only be called once before rendering I think, because the ref is not set at that point.

app/experimenter/templates/experiments/rapid.html
emtwo: Actually realized this wasn't needed at all

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -54,10 +60,19 @@ const displaySelectOptionLabels = (options, values) => {
...
 
   const handleClickRequestApproval = async () => dispatch(requestReview());
 
+  React.useEffect(() => {
+    if (!visualizationRef.current || !window.initSvelte) {
+      return;
+    }
+
+    window.initSvelte(visualizationRef.current);
+  });
```
jaredlockhart, line 74: Oh actually I think if you pass as the second arg `[visualizationRef]` then that will achieve the desired effect since it'll fire the effect whenever that changes, so then i think it looks something like

```js
  React.useEffect(() => {
    if (visualizationRef) {
    window.initSvelte(visualizationRef.current);
    }
  }, [visualizationRef]);
```

emtwo: Actually, I tried it with both `[visualizationRef]` and `[visualizationRef.current]` as the second arg and neither seem to work. They both result in only one call to `useEffect()` in which `visualizationRef.current` is still `undefined`

emtwo: Update: `[visualizationRef.current]` works but it takes like 30 seconds or more until `useEffect()` is called a second time.

jaredlockhart: Oh yikes what the heck

jaredlockhart: Oh it's because 30s is the timeout to refetch from the API which is triggering a re render.  Yeah I tried locally and got exactly what you're describing, okay the way you have it written totally works I vote just keep it like that unless @LZoog knows something better?

pdehaan: Is this because we're using React _and_ Svelte? Or would we still see this 30s penalty box if we were just using React?

jaredlockhart:
Also it would be good to add an issue for this and refer to it by # in the commit it like `fix #123: Add a visualization package wrapped in a feature flag`

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
jaredlockhart: @pdehaan it's what I just said above ^

LZoog: Apologies for the late response here, I missed the @ mention😩

```
  React.useEffect(() => {
    if (visualizationRef.current) {
      window.initSvelte(visualizationRef.current);
    }
  }, [visualizationRef]);
```
This is what I would have suggested, if this doesn't work as expected then I think this is fine, we'll review this when we get to it in `nimbus-ui` anyway 🙂

jaredlockhart: Yeah so I did some more digging @LZoog  

It's not `visualizationRef` since that's an object that gets returned from the `useRef` call, what changes is the `visualizationRef.current` value inside, and if you set that as the useEffect trigger, it "works" in the sense that 

- On the first pass rendering the component, `visualizationRef.current` is `null`, the if statement fails, and it passes over, then it finishes its first render, which adds the `div` and updates the ref
- Now it's done rendering, and nothing triggers it to render again so it just sets there
- After 30s, the API fetch timeout happens, which updates state, which triggers a re render, and now the next pass through when it hits `useEffect` the value has changed, so it calls again, passes the if check, and inits svelte

So I think useEffect doesn't actively 'listen' for changes to the value, it only compares them when something else triggers a render, so for the way you wrote it to work, we'd have to trigger a second render after the first render places the div, which I wasn't sure how to do.

Anyways yes it works the way @emtwo wrote it so, no reason not to keep it that way 👍

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
```diff
@@ -20,6 +20,12 @@ import { ExperimentStatus } from "experimenter-rapid/types/experiment";
 
 export const POLL_TIMEOUT = 30000;
 
+declare global {
+  interface Window {
+    initSvelte(svelteRoot: HTMLElement | null): void;
```
jaredlockhart, line 25: Oh I just noticed if you look at the implementation of this below:

https://github.com/mozilla/experimenter/pull/3419/files#diff-78d5c7f0dd2f373f32fc8310fd33a0b4R6

I'm not sure the root element should be allowed to be `null` and this is also missing the `data` arg, but I think it's not being used yet.  Not worth blocking on, just mentioning.

jaredlockhart (approved):
Awesome looks good! 👍 👍 👍

app/experimenter/static/rapid/components/experiments/ExperimentDetails.tsx
emtwo: Well, `visualizationRef` gets initialized as `null` and `visualizationRef.current` is still `null` on the first call of `useEffect()` so typescript complains if I don't add the `null` there. I think it's fine though because in `useEffect()`, `initSvelte()` is only called when `visualizationRef.current` is not null.

In terms of `data`, yes that arg is just not being used at the moment

PR #3421, opened 2020-09-10
feat: create new nimbus model

### LINKED ISSUE #3396: Create NimbusExperiment model
Rather than refactoring the models to split them in half, let's just add a new model called NimbusExperiment.  That'll be much simpler.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-255)

bbangert:
Because:

* We want to utilize a new Nimbus Experiment model rather than
  increasing complexity with multi-table inheritance.

This commit:

* Add's a new NimbusExperiment model class.

Fixes #3396


app/experimenter/experiments/models/nimbus.py
```diff
@@ -0,0 +1,99 @@
...
+        choices=ExperimentConstants.CHANNEL_CHOICES,
+        blank=True,
+        null=True,
+    )
+    projects = models.ManyToManyField(Project, blank=True)
+    objectives = models.TextField(
+        default=ExperimentConstants.OBJECTIVES_DEFAULT, blank=True, null=True
+    )
+    bugzilla_id = models.CharField(max_length=255, blank=True, null=True)
+    recipe_slug = models.CharField(max_length=255, blank=True, null=True)
```
jaredlockhart, line 70: Don't need this, see https://mozilla.slack.com/archives/C011MC5EB0V/p1599767736076000

app/experimenter/experiments/models/nimbus.py
```diff
@@ -0,0 +1,99 @@
...
+
+    rapid_type = models.CharField(
+        max_length=255,
+        blank=True,
+        null=True,
+        choices=ExperimentConstants.RAPID_TYPE_CHOICES,
+    )
```
jaredlockhart, line 77: Don't need this, we might want something like this later but for now I don't think it applies

app/experimenter/experiments/models/nimbus.py
```diff
@@ -0,0 +1,99 @@
...
+        null=True,
+        choices=ExperimentConstants.RAPID_AUDIENCE_CHOICES,
+    )
+
+    class Meta:
+        verbose_name = "NimbusExperiment"
+        verbose_name_plural = "NimbusExperiments"
+
+    def __str__(self):
+        return self.full_name
```
jaredlockhart, line 99: This would be `self.name` since `full_name` is not an attribute on this model

jaredlockhart (commented):
Looking good so far!

jaredlockhart:
Oh and for these I would just use a 

` # noqa` at the end of both of those lines

It's generally possible to make flake8 happy, but this is a case where we just have to tell it to ignore these lines.

app/experimenter/experiments/models/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    ExperimentConstants,
+    ExperimentEmail,
+    ExperimentManager,
+    ExperimentVariant,
+    Locale,
+    NimbusExperiment,
+    Preference,
+    RolloutPreference,
+    VariantPreferences,
+]
```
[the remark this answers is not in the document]
jaredlockhart: Actually get rid of `Country` and `Locale` from this list, since those are defined here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/base/models.py

and this is what I was saying, accidental re-export 😆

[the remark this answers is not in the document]
jaredlockhart: Oh I think I know why you put those here, becuase other modules are accidentally importing them from here instead of properly from `experimenter.base.models` okay if taking them out of here breaks tests then leave them in adn we can file a follow up to clean that up

app/experimenter/experiments/models/nimbus.py
```diff
@@ -0,0 +1,82 @@
+from django.contrib.auth import get_user_model
+from django.contrib.postgres.fields import ArrayField
+from django.core.validators import MaxValueValidator
+from django.db import models
+
+from experimenter.projects.models import Project
+from experimenter.experiments.constants import ExperimentConstants
```
jaredlockhart, line 7: Oh and I was gonna say we should start a new constants file for Nimbus but we can also do that as a followup

app/experimenter/experiments/tests/test_models.py
```diff
@@ -65,6 +66,12 @@ def test_queryset_annotated_with_latest_change(self):
         )
 
 
+class TestNimbusExperimentModel(TestCase):
+    def test_str(self):
+        experiment = NimbusExperimentFactory.create(slug="experiment-slug")
+        self.assertEqual(str(experiment), experiment.name)
```
jaredlockhart, line 72: Very nice, and I'd say we can move these tests into `experiments/models/tests/` and split them up by old and new, but we can do that as a followup as well.

jaredlockhart (approved):
Looks good to me, thanks @bbangert ! 🎉

app/experimenter/experiments/models/__init__.py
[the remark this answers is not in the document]
bbangert: Yup, that's exactly why I had to import them.

PR #3430, opened 2020-09-11
fix #3410 added dependabot config for integration test dependencies

### LINKED ISSUE #3410: Add dependabot config for integration tests
The integration tests aren't being caught by dependabot updates, so we should add a rule to dependabot to find and update them.

┆Issue is synchronized with this [Jira Story](https://jira.mozilla.com/browse/EXP-268)

tiftran:


.github/dependabot.yml
```diff
@@ -15,3 +15,8 @@ updates:
   schedule:
     interval: weekly
   target-branch: dependencies
+- package-ecosystem: pip
+  directory: "/app/tests"
+  schedule:
+    interval: weekly
+  target-branch: dependencies
```
pdehaan, line 22: Nice catch. I think the only one that might not have be included is the root app/package.json (which currently doesn't have any dependencies or devdependencies, so hard meh):

```sh
git ls-files | grep -E "requirements|package"

app/experimenter/static/core/package.json
app/experimenter/static/rapid/package.json
app/package.json
app/requirements/constraints.txt
app/requirements/default.txt
app/tests/requirements.txt
```

jaredlockhart (approved):
Very nice ty 👍

PR #3447, opened 2020-09-13
fix #3423 added isort

### LINKED ISSUE #3423: Use `isort` to format imports
https://pypi.org/project/isort/

This thingy will automatically group and sort imports, so we should add it to `code_format`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-277)

tiftran:


jaredlockhart (approved):
Wow this is so fancy thank you @tiftran ! I think if we put `isort` before `black` then black will fixup any inconsistencies between the way they format things.

PR #3475, opened 2020-09-15
fix #3473: __all__ should contain only strs

### LINKED ISSUE #3473: Experiment models/__init__.py: must be str, not ModelBase
Oops, `__all__` has to be strings only:

```
if a package’s __init__.py code defines a list named __all__, it is taken to be the list of module names that should be imported when from package import * is encountered. It is up to the package author to keep this list up-to-date when a new version of the package is released. Package authors may also decide not to support it, if they don’t see a use for importing * from their package. For example, the file sound/effects/__init__.py could contain the following code:

__all__ = ["echo", "surround", "reverse"]
```

Noticed when trying to import in my shell:

```
In [1]: from experimenter.experiments.models import Experiment

In [2]: from experimenter.experiments.models import *
TypeError                                 Traceback (most recent call last)
<ipython-input-2-6c362cba7d9f> in <module>
----> 1 from experimenter.experiments.models import *

/usr/local/lib/python3.8/importlib/_bootstrap.py in _handle_fromlist(module, fromlist, import_, recursive)

/usr/local/lib/python3.8/importlib/_bootstrap.py in _handle_fromlist(module, fromlist, import_, recursive)

TypeError: Item in experimenter.experiments.models.__all__ must be str, not ModelBase
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-303)

jaredlockhart:
* Change `app/experimenter/experiments/models/__init__.py:__all__` to only include `str`s
* Allows `from experimenter.experiments.models import *` on shell for local dev
* `*` import should still be avoided for code


bbangert (approved):
Oh right, those are strings..... amazing how much Python knowledge can fade so quick. 😆

PR #3481, opened 2020-09-15
Add API endpoint for exposing visualization data.

emtwo:
This PR adds an API endpoint `/visualization/api/v1/<slug>` which looks up the analysis data from Jetstream, stored in a GCS bucket.

Since the data accessed is not coming from the django db but rather from a gcs bucket, it's a REST call rather than graphQL.


app/experimenter/urls.py
```diff
@@ -10,6 +10,9 @@
     re_path(r"^api/v2/experiments/", include("experimenter.experiments.api.v2.urls")),
     re_path(r"^api/v3/", include("experimenter.experiments.api.v3.urls")),
     re_path(r"^api/v4/", include("experimenter.experiments.api.v4.urls")),
+    re_path(
+        r"^api/visualizations/v1/", include("experimenter.visualization.api.v1.urls")
```
jaredlockhart, line 14: I think this breaks the API versioning pattern we've been using.  We've been using the V1 API as the public read only API but since this is sensitive data I'd expect it to be behind an auth gate.  The V3 API is auth gated and currently hosts stuff used by the nimbus frontend, so you could stick it in there, or make a new one and call it V5?

Also in general the pattern we've been using is

`api/Vx/<resource>/`

so I'd expect it to something like that?

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
...
+            else []
+        )
+
+    def get(self, request, *args, **kwargs):
+        experiment_json = {}
+
+        slug = self.kwargs["slug"]
+        current_expermient_arr = Experiment.objects.get_prefetched().filter(slug=slug)
+        if len(current_expermient_arr) < 1:
+            return HttpResponse(json.dumps(experiment_json))
```
jaredlockhart, line 24: I would expect this to raise a 404?

jaredlockhart: There's actually a handy dealy for this:

https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#get-object-or-404

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
...
+from experimenter.experiments.models import Experiment
+
+
+class VisualizationView(View):
+    def load_data_from_gcs(self, filename):
+        return (
+            json.loads(default_storage.open(filename).read().decode("utf8"))
+            if default_storage.exists(filename)
+            else []
+        )
```
jaredlockhart, line 16: I see below you're just using a truthy test to check for the case that nothing gets read, but you could also just return `None` here so there's no confusion between the read being successful but empty, vs not being able to find the file?

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def get(self, request, *args, **kwargs):
+        experiment_json = {}
+
+        slug = self.kwargs["slug"]
+        current_expermient_arr = Experiment.objects.get_prefetched().filter(slug=slug)
```
jaredlockhart, line 22: `get_prefetched` is a convenience method to automatically join on all the related tables, but if you only need `recipe_slug` then I don't think this is necessary?

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -0,0 +1,69 @@
...
+from experimenter.experiments.tests.factories import ExperimentFactory
+
+
+class TestVisualizationView(TestCase):
+    @parameterized.expand(
+        [
+            ExperimentConstants.STATUS_ACCEPTED,
+            ExperimentConstants.STATUS_COMPLETE,
+        ]
+    )
```
jaredlockhart, line 18: Very nice 👍

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -0,0 +1,69 @@
...
+    @patch("django.core.files.storage.default_storage.open")
+    @patch("django.core.files.storage.default_storage.exists")
+    def test_visualization_view_data(self, status, mock_exists, mock_open):
+        DATA_ROW = {"point": 12, "upper": 13, "lower": 10}
+        FULL_DATA = {"daily": DATA_ROW, "weekly": DATA_ROW}
+
+        class file(object):
```
jaredlockhart, line 45: [Class names should usually be uppercase](https://www.python.org/dev/peps/pep-0008/#class-names)

jaredlockhart: [Also subclassing `object` isn't necessary anymore, it's implicit in Python 3](https://docs.python.org/3/tutorial/classes.html#multiple-inheritance)

app/experimenter/settings.py
```diff
@@ -136,6 +136,7 @@
 
 OPENIDC_EMAIL_HEADER = config("OPENIDC_HEADER")
 OPENIDC_AUTH_WHITELIST = (
+    "visualization",
```
jaredlockhart, line 139: Sort alphabetically pls 🙏

app/experimenter/settings.py
```diff
@@ -379,3 +380,9 @@
 KINTO_BUCKET = config("KINTO_BUCKET")
 KINTO_BUCKET_MAIN = config("KINTO_BUCKET_MAIN")
 KINTO_COLLECTION = config("KINTO_COLLECTION")
+
+
+# Jetstream GCS Bucket data
+DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
+GS_PROJECT_ID = "experiments-analysis"
+GS_BUCKET_NAME = "mozanalysis"
```
jaredlockhart, line 388: Interesting! Fancy! I guess we don't use any filestorage reading anywhere so setting this as the default doesn't break anything.  If we ever need to read from the filesystem we'll have to update this, but for now this is totally fine, it's so nice that this is so clean and JustWorks™️

app/experimenter/visualization/api/v1/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from experimenter.visualization.api.v1.views import VisualizationView
+
+urlpatterns = [
+    url(
+        r"^(?P<slug>[\w-]+)/$",
+        VisualizationView.as_view(),
+        name="visualization",
```
jaredlockhart, line 9: Usually we give views names like `app-name` so since you already called this app `visualization` maybe you can find a word to use to describe what this specific view is doing?

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
+import json
+
+from django.core.files.storage import default_storage
+from django.http import HttpResponse
+from django.views import View
+
+from experimenter.experiments.models import Experiment
+
+
+class VisualizationView(View):
```
jaredlockhart, line 10: Similarly something a little more descriptive here?

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
+import json
+
+from django.core.files.storage import default_storage
+from django.http import HttpResponse
+from django.views import View
```
jaredlockhart, line 5: Oh!  Since this is an API you could use the [Django Rest Framework APIView](https://www.django-rest-framework.org/api-guide/views/) as your base here, that way it handles JSON automatically and any global configuration we add applies to it as well.  And it might even get automatically documented somehow?

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+from experimenter.experiments.models import Experiment
+
+
+class VisualizationView(View):
+    def load_data_from_gcs(self, filename):
+        return (
+            json.loads(default_storage.open(filename).read().decode("utf8"))
```
jaredlockhart, line 13: Since this is doing a decode, in theory it could fail, maybe add a test case for data that isn't valid utf8 and make sure it gets caught nicely?

jaredlockhart (changes requested):
Looking great @emtwo !  Very clean and straight forward 😸

app/experimenter/visualization/api/v1/views.py
jaredlockhart: Sorry scratch this, using `[RetrieveAPIView](https://www.django-rest-framework.org/api-guide/generic-views/#retrieveapiview)` will do all this automatically

app/experimenter/visualization/api/v1/views.py
jaredlockhart: Oh actually using using `[RetrieveAPIView](https://www.django-rest-framework.org/api-guide/generic-views/#retrieveapiview)` should do exactly what you want

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
+import json
+
+from django.core.files.storage import default_storage
+from django.http import HttpResponse
+from django.views import View
+
+from experimenter.experiments.models import Experiment
```
jaredlockhart, line 7: Oh and this should be `NimbusExperiment` not `Experiment`

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def get(self, request, *args, **kwargs):
+        experiment_json = {}
+
+        slug = self.kwargs["slug"]
+        current_expermient_arr = Experiment.objects.get_prefetched().filter(slug=slug)
+        if len(current_expermient_arr) < 1:
+            return HttpResponse(json.dumps(experiment_json))
+
+        recipe_slug = current_expermient_arr[0].recipe_slug.replace("-", "_")
```
jaredlockhart, line 26: Oh I just remembered on `NimbusExperiment` this is just `slug`, not `recipe_slug`

app/experimenter/visualization/api/v1/views.py
jaredlockhart: If using a class based view doesn't work well you can also keep it in the form it's in now and just decorate it to be a [DRF function based view](https://www.django-rest-framework.org/api-guide/views/#function-based-views)

app/experimenter/visualization/api/v1/views.py
jaredlockhart: You can also use [`get_object_or_404`](https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#get-object-or-404) here to properly raise 404 if the slug isn't found

jaredlockhart:
Also it would be good to add an issue for this and refer to it by # in the commit it like `fix #123: Add API endpoint for exposing visualization data`

emtwo:
Addressed review comments. I put the review addressing in a separate commit for readability. I can squash it with the previous commit when it's ready to be merged

app/experimenter/visualization/api/v1/views.py
emtwo: I'm not sure why I had the decode there, but I actually don't need it so I removed it.

app/experimenter/urls.py
emtwo: I put it under v3

app/experimenter/settings.py
```diff
@@ -136,6 +136,7 @@
 
 OPENIDC_EMAIL_HEADER = config("OPENIDC_HEADER")
 OPENIDC_AUTH_WHITELIST = (
+    "analysis-results-data",
```
jaredlockhart, line 139: Wait why are we adding this here?  This would mean unauthenticated clients can all this API, but we should only need to call this from teh frontend, which is authenticated, right?  If we expose this here than non LDAP could call it and get potentially sensitive data?

app/experimenter/visualization/api/v1/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from experimenter.visualization.api.v1.views import analysis_results_view
+
+urlpatterns = [
+    url(
+        r"^(?P<slug>[\w-]+)/$",
+        analysis_results_view,
+        name="analysis-results-data",
```
jaredlockhart, line 9: 👍 for adding a descriptive name, but in general the first part of the name should be the name of the "app" it's contained within (ie the name of the folder inside `experimenter/`) which in this case is `visualization`?  Just makes it easier to know where to find the view that implements it.

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,38 @@
+import json
+
+from django.core.files.storage import default_storage
+from django.http import HttpResponse
```
jaredlockhart, line 4: Oh Django Rest Framework has its own `Response` object:

[`Handler methods may return REST framework's Response, instead of Django's HttpResponse. The view will manage content negotiation and setting the correct renderer on the response.`](https://www.django-rest-framework.org/api-guide/views/#api_view)

app/experimenter/visualization/api/v1/views.py
```diff
@@ -0,0 +1,38 @@
...
+        json.loads(default_storage.open(filename).read())
+        if default_storage.exists(filename)
+        else None
+    )
+
+
+@api_view()
+def analysis_results_view(request, slug):
+    experiment = get_object_or_404(NimbusExperiment.objects.filter(slug=slug))
+    experiment_json = {}
```
jaredlockhart, line 22: Nit: this is getting supernitpicky but I usually I use the suffixes `_data` if it's a dict that came from some json and `_json` if it's the actual json string.  But that's just me.

app/experimenter/urls.py
```diff
@@ -9,6 +9,9 @@
     re_path(r"^api/v1/experiments/", include("experimenter.experiments.api.v1.urls")),
     re_path(r"^api/v2/experiments/", include("experimenter.experiments.api.v2.urls")),
     re_path(r"^api/v3/", include("experimenter.experiments.api.v3.urls")),
+    re_path(
+        r"^api/v3/visualizations/", include("experimenter.visualization.api.v1.urls")
```
jaredlockhart, line 13: Ah good thank you for reordering that, but I also think we could:

1) Exclude the `visualizations/` subpath here so it looks the same as the regex matcher above it (Django will merge these two together properly
2) It's confusing to put 'v3' in the url but 'v1' in the import path, lumping it together with v3 seems reasonable to me since that's already waht' sused as the private API for the frontend, but maybe name the folder `visualization/api/v3` so that it matches its public interface?

app/experimenter/visualization/api/v1/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from experimenter.visualization.api.v1.views import analysis_results_view
+
+urlpatterns = [
+    url(
+        r"^(?P<slug>[\w-]+)/$",
```
jaredlockhart, line 7: Move the `visualizations/` prefix out of the project wide urls.py and into this line

jaredlockhart (changes requested):
Looking great @emtwo ! Just a couple more things and this should be ready to go 👍 👍 👍

jaredlockhart (approved):
AWESOME @emtwo perfect everything looks great!!  Thank you for addressing all the feedback 🎉 🎉 🎉

emtwo:
@robotblake heads up on this PR. I'm wondering if I just need to file a bugzilla ticket to request setting the environment `GOOGLE_APPLICATION_CREDENTIALS` environment variable in prod and stage?

PR #3483, opened 2020-09-15
Dependencies

jodyheavener:
Let's see how this goes...


jaredlockhart:
That's odd, why is poetry lock conflicting?

jodyheavener:
Think it just needed an update from main

jodyheavener:
Neat

jaredlockhart:
Ah yeah it's react-testing-library tightening the reigns on us.

```
    TestingLibraryElementError: Found a label with the text of: Features, however the element associated with this label (<div />) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a <div />, you can use aria-label or aria-labelledby instead.

    > 94 |     const featuresField = getByLabelText("Features");
```

So we either have to
a) Change the html so we're not trying to label a div
b) Change the test so we look up by text rather than label
c) Revert the `react-testing-library` version bump and deal with it all later

jodyheavener:
@jaredlockhart Ideally we wouldn't revert the react-test-library version, but I'm a little stumped here.

> Change the html so we're not trying to label a div

It looks like this is occurring at the react-select level, and not something we're doing? I'm not entirely familiar with it.

> Change the test so we look up by text rather than label

I've tried moving it to getByTestId as well as getByText, the first straight up can't find the element for some reason, and the second can find the element but doesn't seem to be able to act on the select. It appears to be using react-select-event which I'm again not familiar with.

jaredlockhart:
Oh you might have to revert this one as well: https://github.com/mozilla/experimenter/pull/3483/commits/3a443cc15dd14a27a6d142556b76bd73948a2c8a

jodyheavener:
😒 I'm getting this same failure locally even after rollback.

jodyheavener:
Well, it passes on main, so something in here is doing it.

jodyheavener:
@jaredlockhart I can't seem to get that one test passing, regardless of any recent dependabot changes. I even dropped all the commits locally. I'm wondering if this breaking upgrade happened a while back, but because the yarn lock file wasn't being updated it was never caught?

jaredlockhart:
Hrmmm let me try it locally

jaredlockhart:
Okay! Yes you're right it was yarn.lock, we really need to update dependabot to keep that up to date with everything as well, I'll update dependabot right now.

So what I did was

1) Snapshot yarn.lock on `main`
2) Checkout `dependencies`
3) Revert https://github.com/mozilla/experimenter/pull/3483/commits/98119c362c7bd916bcb22440a016cf6fb24a4966
4) Overwrite yarn.lock with the one from `main`
5) Update yarn.lock with `yarn`

And that seems to work.  Sorry for all the confusion!

jaredlockhart:
And I'll update the instructions with some more helpful detail based on your feedback @jodyheavener 😄

jodyheavener:
@jaredlockhart sorry to bug you on this again, I'm seeing a failure that looks related to a dependency in the Django app, which is not my strong suit 😅 If you get time would you be able to take a look, or guide me in the direction to resolving it myself?

jaredlockhart:
Yeah of course!  It's actually in the integration tests, which are written in python, but are separate from the main app so they have their own requirements file and build process.

It looks to me like it's complaining about one of those cross package version range checks, so usually you just gotta go in and bump another package so it's happy, the python package manager isn't smart enough to do this automatically.  I'll take a look at it in a little bit.

jaredlockhart (approved):

PR #3486, opened 2020-09-15
chore(docs): Add clarification footnote to ADR 0002

LZoog:
It's atypical to edit an existing ADR but because 0002 was merged so recently and I've been asked about this twice now, I thought we could make an exception for this date-stamped note.


jaredlockhart (approved):
Yep good note 👍

PR #3487, opened 2020-09-16
 fix #3578 added is_high_population to v1 api

tiftran:


app/experimenter/normandy/tests/test_tasks.py
```diff
@@ -338,6 +338,32 @@ def test_live_rollout_updates_population_percent(self):
...
+        }
+        mock_response = mock.Mock()
+        mock_response.json = mock.Mock()
+        mock_response.json.return_value = mock_response_data
+        mock_response.raise_for_status = mock.Mock()
+        mock_response.raise_for_status.side_effect = None
+        mock_response.status_code = 200
+
+        self.mock_normandy_requests_get.return_value = mock_response
+
```
jaredlockhart, line 362: Maybe check that it's in the default state before the update is called?

jaredlockhart (approved):
Maybe add the field to `admin.py` in case we need to check it or change it?  Looks great @tiftran !

app/experimenter/normandy/tasks.py
```diff
@@ -163,6 +165,17 @@ def update_population_percent(experiment, recipe_data):
             experiment.save()
 
 
+def update_is_high_population(experiment, recipe_data):
+    recipe_args = recipe_data.get("arguments", {})
+    if recipe_data and ("isHighPopulation" in recipe_args):
+        recipe_is_high_population = recipe_args["isHighPopulation"]
```
jaredlockhart, line 171: Oh you know you could do

```py
if recipe_data and recipe_is_high_population := recipe_args.get("isHighPopulation")
```

with the fancy walrus thing

walrus-scratch

app/experimenter/normandy/tasks.py
```diff
@@ -163,6 +165,17 @@ def update_population_percent(experiment, recipe_data):
...
+        if (experiment.is_high_population is None) or (
+            experiment.is_high_population == recipe_is_high_population
+        ):
+            experiment.is_high_population = recipe_args["isHighPopulation"]
+            experiment.save()
```
jaredlockhart, line 176: Just occurred to me we're not creating a changelog here, I think it's probably less relevant here than it is for something like updating the rollout percentage, since it'll get set once at the start of the recipe and theoretically not change, so probably not worth adding here, but if it comes up later that people aren't sure when/where it comes from, could be worth thinking about later?

tiftran: i did consider a changelog here, but considering nothing in the frontend was changing i thought it might be a little confusing. But that being said, I think i added a logging statement here instead

app/experimenter/settings.py
```diff
@@ -302,18 +302,6 @@
...
-        "schedule": config("CELERY_SCHEDULE_INTERVAL", default=300, cast=int),
-    },
-    "check_experiment_is_live": {
-        "task": "experimenter.kinto.tasks.check_experiment_is_live",
-        "schedule": config("CELERY_SCHEDULE_INTERVAL", default=300, cast=int),
-    },
-    "check_experiment_is_complete": {
-        "task": "experimenter.kinto.tasks.check_experiment_is_complete",
-        "schedule": config("CELERY_SCHEDULE_INTERVAL", default=300, cast=int),
-    },
```
jaredlockhart, line 316: Why is this coming out?

app/experimenter/normandy/tasks.py
```diff
@@ -163,6 +165,21 @@ def update_population_percent(experiment, recipe_data):
             experiment.save()
 
 
+def update_is_high_population(experiment, recipe_data):
+    recipe_args = recipe_data.get("arguments", {})
+    if recipe_data and recipe_is_high_population := recipe_args.get("isHighPopulation"):
```
jaredlockhart, line 170: `if recipe_args and recipe_is_high_population := recipe_args.get("isHighPopulation"):`

`recipe_args` and not `recipe_data`?

jaredlockhart: Actually, it's kind of a superfluous check since if it defaults to `{}` and then you walrus assign from that it'll be `None` which will fail the check anyways so you should be able to just do 

`if recipe_is_high_population := recipe_args.get("isHighPopulation"):`?

app/experimenter/settings.py
tiftran: oops

PR #3508, opened 2020-09-17
    fix #3496: Add bucket logic for Nimbus

### LINKED ISSUE #3496: Move BucketRange to Nimbus models
We currently have the [BucketRange models](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/legacy.py#L1303-L1366) in the legacy models file, but this only applies to Nimbus so we should move it over there.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-317)

jaredlockhart:
* Add NimbusBucketNamespace and NimbusBucketRange to Nimbus models
    * Add related test factories
    * Add related tests to Nimbus model tests
    * Leave the old RapidExperiment versions in place to preserve the functioning of the existing Rapid workflow until Nimbus reaches parity


app/experimenter/experiments/models/nimbus.py
```diff
@@ -80,3 +80,71 @@ class Meta:
...
+        verbose_name = "Bucket Namespace"
+        verbose_name_plural = "Bucket Namespaces"
+        unique_together = ("name", "instance")
+        ordering = ("name", "instance")
+
+    def __str__(self):  # pragma: no cover
+        return f"{self.name}-{self.instance}"
+
+    @classmethod
+    def request_namespace_buckets(cls, name, experiment, count):
```
k88hudson, line 103: We discussed whether it's necessary to actually store a namespace for orthogonal experiments, but we figured it would be useful for the case of wanting to exclude another experiment that was added later.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -80,3 +80,71 @@ class Meta:
 
     def __str__(self):
         return self.name
+
+
+class NimbusBucketNamespace(models.Model):
```
k88hudson, line 85: can we call this `NimbusIsolationGroup`?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -80,3 +80,71 @@ class Meta:
...
+    @classmethod
+    def request_namespace_buckets(cls, name, experiment, count):
+        if cls.objects.filter(name=name).exists():
+            namespace = cls.objects.filter(name=name).order_by("-instance").first()
+        else:
+            namespace = cls.objects.create(name=name)
+
+        return namespace.request_buckets(experiment, count)
+
+    def request_buckets(self, experiment, count):
```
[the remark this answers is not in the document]
k88hudson: Because this would requrie a bunch of changes including to the client, let's land the current implementation and revisit.

app/experimenter/experiments/tests/test_models/test_nimbus_models.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+
+class TestNimbusExperimentModel(TestCase):
+    def test_str(self):
+        experiment = NimbusExperimentFactory.create(slug="experiment-slug")
+        self.assertEqual(str(experiment), experiment.name)
+
+
+class TestNimbusBucketNamespace(TestCase):
```
k88hudson, line 20: Can you add a comment that describes which case this handles

e.g. An experiment is created that is orthogonal to all other experiments

k88hudson (approved):
R+ with the rename and comments

Issue #3524, opened 2020-09-22
Use black profile for isort


jaredlockhart:
isort and black were confusing each other just now and then I discovered you can tell isort to use your black profile:

`python -m isort . --profile black`

and that seems to resolve it so we should use that

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-343)
┆Issue Number: EXP-343

PR #3525, opened 2020-09-22
fix #3510: chore: Split legacy and nimbus constants

### LINKED ISSUE #3510: Split ExperimentConstants into NimbusConstants
We currently have the ExperimentConstants class in constants.py:

[https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants.py](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants.py)

but we should split this out to a separate constants file for NimbusConstants and mix that into NimbusExperiment

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-327)
┆Issue Number: EXP-327

jaredlockhart:
Because

* We will be setting constants specifically for nimbus we should put them in their own file

This commit:

* Splits constants.py into a module with nimbus and legacy specific constants file
* Updates nimbus models/tests to use the nimbus constants


jaredlockhart:
Note: blocked on https://github.com/mozilla/experimenter/issues/3524

app/experimenter/experiments/constants/__init__.py
```diff
@@ -0,0 +1,2 @@
+from experimenter.experiments.constants.legacy import ExperimentConstants  # noqa: F401
+from experimenter.experiments.constants.nimbus import NimbusConstants  # noqa: F401
```
bbangert, line 2: Didn't want to use a `__all__`?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -0,0 +1,106 @@
...
+        (STATUS_ACCEPTED, STATUS_ACCEPTED_LABEL),
+        (STATUS_LIVE, STATUS_LIVE_LABEL),
+        (STATUS_COMPLETE, STATUS_COMPLETE_LABEL),
+        (STATUS_REJECTED, STATUS_REJECTED_LABEL),
+    )
+
+    MAX_DURATION = 1000
+
+    # Version stuff
+    VERSION_CHOICES = (
```
bbangert, line 32: I haven't seen exactly how we use these, but would we get clearer code elsewhere when using this if this was a list of namedtuple's instead of a tuple of tuples where our access is going to be by index only?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -0,0 +1,106 @@
...
+
+    STATUS_DRAFT_LABEL = "Draft"
+    STATUS_REVIEW_LABEL = "Ready for Sign-Off"
+    STATUS_SHIP_LABEL = "Ready to Ship"
+    STATUS_ACCEPTED_LABEL = "Accepted by Normandy"
+    STATUS_LIVE_LABEL = "Live"
+    STATUS_COMPLETE_LABEL = "Complete"
+    STATUS_REJECTED_LABEL = "Rejected"
+
+    STATUS_CHOICES = (
```
bbangert, line 19: Same as below, would we use this more clearly using namedtuple's?

bbangert (approved):
Looks good, I'm curious if namedtuple's would help readability. We should also probably consider whether we'd benefit from a state machine library like https://pypi.org/project/Automat/ as we start to build out the valid state transitions and handling.

app/experimenter/experiments/constants/__init__.py
jaredlockhart: Oh I discovered that `__all__` is only used when you `import *`, not when you import individually by name, and since we don't use `import *` anywhere it doesn't seem to provide any value?

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Oh that's a good point this might be a good opportunity to use enums!

bbangert: An enum would also help with readability.

app/experimenter/experiments/constants/__init__.py
bbangert: I meant so that the `noqa` wouldn't be needed.

jaredlockhart: Yeah I know but the other thing with `__all__` is it's lazily evaluated so we had a situation last week where we made changes to the module but forgot to make them in `__all__` and it passed through all the checks and made it into `main` and wasn't noticed until it broke when I was poking around at stuff, so it seems to

- not provide value
- introduce possible risk/complexity

so just disabling that one lint check seems less risky? What do you think?

bbangert: I'm good with disabling the check for these.

app/experimenter/experiments/constants/nimbus.py
bbangert: The version choice might be better as a namedtuple.

jaredlockhart: Oh django has a dealy for this now https://docs.djangoproject.com/en/3.1/ref/models/fields/#enumeration-types

jaredlockhart:
@bbangert Oh there's also this that integrates with django models: https://github.com/viewflow/django-fsm

bbangert (approved):
Nice!

PR #3526, opened 2020-09-22
fix #3524 isort to use black profile

### LINKED ISSUE #3524: Use black profile for isort

tiftran:


jaredlockhart (approved):
Very nice 🙆‍♀️

PR #3531, opened 2020-09-23
fix #3461 multi-pref design page changes

### LINKED ISSUE #3461: Multipref missing "remove pref" button for top pref
https://experimenter.services.mozilla.com/experiments/shirley-omnibus-experiment/ while trying to work around a true defect of missing the "remove pref" button for the top preference, Kimmy hit what I consider an acceptable behavior of "you cannot swap pref names".  

I replicated the set-up on stage here https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/testing-deletion-v2/edit-design/ for a dev to see.  

If you go to try and remove the top preference - **there is no "remove pref" button for pref 1**.  The "remove pref" button applies to the pref AFTER the button.  Since the top one has no button - there is no way to remove it.  
image

To work around it, Kimmy copied and pasted the names from following prefs up a row - then deleted the last one... and tried to save.  That made her hit the "you cannot swap pref names".  That one has a work around of adding a character, saving, and deleting those characters.  I'm not worried about this "swap pref names" behavior - as it's not likely unless people are working around the missing "remove pref" button on pref 1.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-294)

tiftran:
Below is the simplest solution, I could come up with, where the first pref is removable, but the UI still enforces that there is at least one pref per branch. The remove button disappears when there is only one pref.

I also added dividers between prefs to make it more clear which button corresponds to which pref.

In addition, while not in the original issue, I moved the fields around so that there is more space to input the pref name and pref value.

thoughts? @shell1 @jaredlockhart 

Multiple Prefs
Screen Shot 2020-09-22 at 7 49 47 PM
Singular Pref
Screen Shot 2020-09-22 at 7 49 40 PM


jaredlockhart (approved):
Very nice, elegant solution 👍 Just poked around at it, everything seems to behave correctly, all the validation is working, removing the first one works exactly as I'd expect, everything looks correct great job 🎉

PR #3533, opened 2020-09-23
fix #3492: feat: Add NimbusBranch model

### LINKED ISSUE #3492: Create NimbusBranch model
Let's add a model for NimbusBranch that foreignkeys onto NimbusExperiment and has the following fields:

* name: charfield
* slug: charfield
* description: textfield (optional)
* ratio: positiveinteger (default 1)
* value: textfield
* enabled: booleanfield

and add a test factory.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-314)
┆Issue Number: EXP-314

jaredlockhart:
Because

* We need to add a branch model for NimbusExperiment

This commit

* Adds NimbusBranch model
* Adds factories for control and treatment
* Exposes NimbusBranch to the admin via an inline


app/experimenter/experiments/models/nimbus.py
```diff
@@ -82,6 +82,31 @@ def __str__(self):
...
+        null=False,
+        related_name="branches",
+        on_delete=models.CASCADE,
+    )
+    is_control = models.BooleanField(default=False)
+    enabled = models.BooleanField(default=True)
+    name = models.CharField(max_length=255, blank=False, null=False)
+    slug = models.SlugField(max_length=255, blank=False, null=False)
+    description = models.TextField(default="")
+    ratio = models.PositiveIntegerField(default=1)
```
bbangert, line 98: Ratio's never include decimal places? Are there docs somewhere on what a Nimbus branch is?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -82,6 +82,31 @@ def __str__(self):
...
 
+class NimbusBranch(models.Model):
+    experiment = models.ForeignKey(
+        NimbusExperiment,
+        blank=False,
+        null=False,
+        related_name="branches",
+        on_delete=models.CASCADE,
+    )
+    is_control = models.BooleanField(default=False)
```
bbangert, line 93: Would it be worthwhile to put a table constraint on to ensure a given Nimbus experiment can have no more than one enabled control branch?

bbangert (approved):

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Yeah I've thought about it but I don't know how to express it?  You can't use a uniqueness constraint to do it cuz you can have multiple falses?

bbangert: Unique indexes can have where statements, and maybe a check statement can do it. e.g.
```
CREATE UNIQUE INDEX only_one_true ON nimbus_branch (experiment_pk, is_control)
WHERE is_control IS true
```

Or something like that? I found this answer which discusses some approaches:
https://dba.stackexchange.com/questions/73125/restrict-two-specific-column-values-from-existing-at-the-same-time/73133#73133

bbangert: BTW, Postgres doesn't allow that format, but it does allow a unique index which is enforced the same:
https://stackoverflow.com/questions/16236365/postgresql-conditionally-unique-constraint

app/experimenter/experiments/models/nimbus.py
k88hudson: This is in terms of the DTO but describes the requirements for ratio : https://docs.google.com/document/d/1jZQ--ozuQ0_Su-lh7-dX-3yqmWsIan2QPYyTt6lv4ow/edit#heading=h.5bv7czvcn6dm

bbangert (approved):

Issue #3429, opened 2020-09-11
Allow consuming server settings values on client, serve app at `/nimbus`


data-sync-user:
Certain server settings should be exposed from Django to the client, including at least a Sentry DSN. In FxA, we expose server config data [through a meta tag](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/public/index.html#L13). We need to do this similarly, but set up through a Django template (see https://mozilla.slack.com/archives/CANCHH6R1/p1600797437034700), and the application should be served at `/nimbus`.

Closing this issue should set us up to grab the Sentry DSN URL in EXP-282 and will serve the app to `/nimbus`. It wouldn't hurt to update the README with a quick blurb about this.

(Context history: see this Slack thread|https://mozilla.slack.com/archives/C011MC5EB0V/p1599764101070400] and a [meeting note doc](https://docs.google.com/document/d/1bM5ZDXsl2bcWRWW9W8tk47Ik9eTQJqKObHBDj5w1sao/edit) around this)

Notes from Jared:
1) add a template to host the frontend that basically looks like https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/templates/experiments/rapid.html
2) a view to host it + tests
3) a url path to it
4) the logic around using the yarn dev server on dev and static assets management on prod
The view should look like: https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/views.py#L234-L235, and here's the test: https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/test_views.py#L807-L814
Example URLs: https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/urls.py#L22-L28, but in that case we were mirroring the route definitions in both django and react:
https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/rapid/components/App.tsx#L17-L33

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-283)
┆Issue Number: EXP-283

data-sync-user:
➤ Lauren Zugai commented:

Note: I've modified this to include serving from `/nimbus` but left it at a 3. If it seems more like a 5 when someone starts on it, bump it up. 👍

data-sync-user:
➤ Jody Heavener commented:

[...]

PR #3541, opened 2020-09-24
fix #3493 feat: Add SENTRY_DSN to context processors

### LINKED ISSUE #3493: Add SENTRY_DSN to template context processor
We have a [template context processor]([https://github.com/mozilla/experimenter/blob/main/app/experimenter/base/context_processors.py)](https://github.com/mozilla/experimenter/blob/main/app/experimenter/base/context_processors.py)) for exposing things like settings to templates, we should add SENTRY_DSN there so that when the new frontend infrastructure is in place we can pass it along.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-315)
┆Issue Number: EXP-315

jaredlockhart:
Because

* We need to be able to configure sentry for the react frontend

This commit

* Adds SENTRY_DSN to the global context processor


app/experimenter/base/context_processors.py
```diff
@@ -11,4 +11,5 @@ def features(request):
     return {
         "FEATURE_MESSAGE_TYPE": settings.FEATURE_MESSAGE_TYPE,
         "FEATURE_ANALYSIS": settings.FEATURE_ANALYSIS,
+        "SENTRY_DSN": settings.SENTRY_DSN,
```
bbangert, line 14: I don't really understand how this ropes into anything. Is it used elsewhere to turn on Sentry?

jaredlockhart: Will be consumed here:

https://github.com/mozilla/experimenter/issues/3429

bbangert (approved):

PR #3542, opened 2020-09-24
fix #3532 feat: Limit slugs to 80 characters

### LINKED ISSUE #3532: Experiment and branch slugs should be a maxmimum of *80 characters
It looks like [Glean](https://mozilla.github.io/glean/book/user/experiments-api.html) is expecting a maximum of 100 characters for experiment and branch slugs, so I think it makes sense to strongly require that these can only be 100 characters in the DB.

Note from [~jkerim@mozilla.com]: Becuase we will also be emitting non-glean telemetry and the max there is 80 characters, we should limit to 80.  Confirmed with [~tismith@mozilla.com]

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-347)
┆Issue Number: EXP-347

jaredlockhart:
Because

* Telemetry systems including Firefox Desktop Telemetry v4 and Glean have limits on the length of their unique identifiers,
we should limit the size of our slugs to the smallest limit, which is 80 for Firefox Desktop Telemetry v4.

This commit

* Limits the length of NimbusExperiment and NimbusBranch slugs to 80 characters


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -47,6 +47,7 @@ class Version(models.TextChoices):
         FIREFOX_99 = "99.0"
         FIREFOX_100 = "100.0"
 
+    MAX_SLUG_LEN = 80
```
k88hudson, line 50: Can you add a comment that says what you said in the ticket, i.e. 
`Telemetry systems including Firefox Desktop Telemetry v4 and Glean have limits on the length of their unique identifiers,
we should limit the size of our slugs to the smallest limit, which is 80 for Firefox Desktop Telemetry v4.`

bbangert: Why not put the `MAX_SLUG_LEN` under the comment?

jaredlockhart: Is there a preferred ordering between constants and comments I don't know about?

bbangert: In general I'm used to comments describing things immediately under/after them, this is the first time I've seen it in reverse. 🤷‍♂️

bbangert (approved):

PR #3519, opened 2020-09-21
feat(nimbus-ui): Init nimbus-ui with CRA Typescript, remove cruft, add tests, linting, workspace, README

LZoog:
I'm not sure where we want to serve this since the Rapid project serves at `/experiments/rapid/`, maybe we can call it `/nimbus` until we replace what's in `core` and then we can change to `/`? **Edit**: This will be done in another issue but Kate confirmed it'll be served at `/nimbus`.

I had to downgrade `react-scripts` to 3.4.0 because otherwise at least the `App` test appears to fail with the `--coverage` flag, see [#9322](https://github.com/facebook/create-react-app/issues/9322).

~To do (and why this is opened as a draft):~
* ~For some reason I'm getting a TS complaint on using 'screen', will fix~
* ~`nimbus-ui` is producing a `yarn.lock` file so I've got something incorrectly setup in yarn workspaces~
* ~Prettier~
* ~Remove gitignore~

fixes #3416

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-339)

PR #3547, opened 2020-09-24
Add nimbus-ui dev server to docker-compose file

### LINKED ISSUE #3551: Add nimbus-ui dev server to docker-compose file
We want to be able to spin up the nimbus-ui CRA dev server with everything else via `make up`. At the time of writing there's a WIP PR/branch here with more details: [https://github.com/mozilla/experimenter/pull/3547](https://github.com/mozilla/experimenter/pull/3547)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-354)
┆Issue Number: EXP-354

LZoog:
This should have been included in #3519.

@jaredlockhart I thought this would be a quick fix (famous last words) but I spent a little more time here than expected ~so I'm thinking I'll go ahead and file that issue~ (done, it's #3551) for this and circle back to it unless there's something I'm missing.

`app:build` is what runs with `make up` so I've stuck the command (`yarn start`) for the dev server here, but it 1) clears the console output above it (and there's not an option to stop this but there's probably a workaround we can use, see https://github.com/facebook/create-react-app/issues/2495) and 2) I have to pass in `CHOKIDAR_USEPOLLING=true` to get HMR working which appears to use more CPU resources. I'd like to at least look into a solution for the first and see if there's a better solution for the second.

I don't _think_ we need the `assets_volume` or `static_volume` since it runs the dev server. (Actually, I think I can remove `/app/experimenter/nimbus-ui/node_modules/` as well.)

fixes #3551


lmorchard:
FWIW, I ran into that console-clearing issue with CRA back when we first introduced it to FxA. No clean workarounds, though [these at least seem implementable](https://github.com/facebook/create-react-app/issues/2495#issuecomment-488505385) - e.g. you could try this as a hack to force `stdout.isTTY===false`:

```
FORCE_COLOR=true yarn workspace @experimenter/nimbus-ui start | cat
```

We mostly got around it on FxA by all the logs running through pm2

docker-compose.yml
```diff
@@ -24,6 +24,20 @@ services:
...
+    env_file: .env
+    environment:
+      - CHOKIDAR_USEPOLLING=true
+    tty: true
+    ports:
+      - "3000:3000"
+    volumes:
+      - ./app:/app
+      - /app/node_modules/
+      - /app/experimenter/nimbus-ui/node_modules/
```
jaredlockhart, line 38: No I think you wanna keep this cuz you wanna use the node modules that are built inside the container, not the outside ones

docker-compose.yml
```diff
@@ -24,6 +24,20 @@ services:
       - static_volume:/app/experimenter/served/
     command: bash -c "/app/bin/wait-for-it.sh kinto:8888 -- python bin/setup_kinto.py;/app/bin/wait-for-it.sh db:5432 -- python /app/manage.py collectstatic --noinput;python /app/manage.py runserver 0:7001"
 
+  yarn-nimbus-ui:
+    image: app:build
+    env_file: .env
+    environment:
+      - CHOKIDAR_USEPOLLING=true
```
jaredlockhart, line 31: Yeah this is waht tripped me up last time I tried this, normal HMR uses one port for the web content (3000 in this case) and then a separate port for the websocket, which I think will be randomly generated everytime you start the server normally, or you can pass a flag to lock it to something, so in theory you lock it to a port and then expose that in compose and it should work but I couldn't get it working last time I tried, not sure if it's worth another try here?  This semed to work pretty quickly for me

jaredlockhart (commented):
@LZoog seems to work properly for me 👍

jaredlockhart:
I see waht you mean about that console clearing thing that is annoying XD if we can find some way to shut that off, is there some arg you can pass it that just saw treat it as an append only log?

LZoog:
> Nope, the CRA devs have asserted that console clearing is the best experience and not worth making configurable in that github issue.

Yeah and I'm surprised they decided against it, there's _several_ feature requests and closed issues for it (you even commented on that issue I see!). I'll see what I can do Monday about https://github.com/mozilla/experimenter/pull/3547#issuecomment-698639876 and get this open as a non-draft PR 👍

docker-compose.yml
```diff
@@ -24,6 +24,20 @@ services:
...
+    tty: true
+    ports:
+      - "3000:3000"
+    volumes:
+      - ./app:/app
+      - /app/node_modules/
+      - /app/experimenter/nimbus-ui/node_modules/
+    command: bash -c "yarn workspace @experimenter/nimbus-ui start"
```
lmorchard, line 39: FWIW, I just tried this locally and it seems to hack around the console clearing nonsense:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 9cd20549..1ba50734 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
       - ./app:/app
       - /app/node_modules/
       - /app/experimenter/nimbus-ui/node_modules/
-    command: bash -c "yarn workspace @experimenter/nimbus-ui start"
+    command: bash -c "FORCE_COLOR=true yarn workspace @experimenter/nimbus-ui start | cat"
 
   yarn-core:
     image: app:build
```

LZoog:
After trying to find an alternative to `CHOKIDAR_USEPOLLING`, I found [this SO question](https://stackoverflow.com/questions/44643045/running-development-server-with-create-react-app-inside-of-a-docker-container) and per the second answer, I removed the option and restarted Docker and it magically seems to be working now. 🤨

Just to be sure I started/stopped `make up` twice, restarted Docker again and then ran `make up`, and HMR is working for me after refreshing `localhost:3000` once. I encourage others to check into the branch to make sure it works as expected for you as well.

Thanks to Les's comment too the console no longer clears, so I think we're g2g here. 🙌

jodyheavener (approved):
Works for me

jaredlockhart (approved):

PR #3550, opened 2020-09-25
fixes #3417 feat: Add AppLayout and Storybook to nimbus-ui

### LINKED ISSUE #3417: Stub AppLayout, add Storybook
The `AppLayout` will be a wrapper for main content, ~keeping the header and footer across page changes.~ (This is no longer the case since the Experiment view won't have a header/footer) For now, it’ll just be a placeholder to get Storybook setup with a component and can return a div and its children from props.

Create this in `src/components/AppLayout/index.tsx` to start us on the “Use directories for components instead of filenames” point in ADR #2. Add Storybook to the project with a `.stories` file in the AppLayout directory and ensure it loads as expected.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-275)
┆Issue Number: EXP-275

LZoog:
fixes #3417 

Added a TODO that we'll remove in epic 2 when we actually create the meat of those components. Also code coverage will be displayed by default when running `yarn test`.

The default Storybook port is `52911`, I chose `3001` since the React dev server runs on `3000`.


app/experimenter/nimbus-ui/src/AppLayout/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+type AppLayoutProps = {
+  children: React.ReactNode;
+};
+
+// TODO: Depending on the main page ("directory view") design, we may want two
+// AppLayouts - one with a header/footer, and another with a nav sidebar.
+
+export const AppLayout = ({ children }: AppLayoutProps) => {
+  return (
+    <main id="main" data-testid="main">
```
jodyheavener, line 16: Unless you're thinking we might add a Skip button I don't think we need this ID. #downwithids

app/experimenter/nimbus-ui/README.md
```diff
@@ -4,7 +4,7 @@
```
jodyheavener, line 4: Let's add some docs (even a one-liner, or just copy directly from fxa-settings) about what Storybook is and how to start it.

jodyheavener (changes requested):
Turns on for me! Docs and that little ID thing and this is g2g.

app/experimenter/nimbus-ui/src/AppLayout/index.tsx
jodyheavener: Omg especially if there might be multiple `AppLayout`s can you IMAGINE the scandal here??

app/experimenter/nimbus-ui/src/AppLayout/index.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
```
jodyheavener, line 1: Oh no I missed that this doesn't have an MPL header! It's okay, I don't want to do another CI round. We can add it later.

jodyheavener (approved):
Thank you!

app/experimenter/nimbus-ui/src/AppLayout/index.stories.tsx
LZoog: Dang it, I missed it too. I swear that needs to be automated. With whatever I'm doing next I'll add it here unless you beat me to it. :P

PR #3553, opened 2020-09-25
feat(nimbus-ui): add bootstrap setup, stylelint, document how to write styles

jodyheavener:
## This Pull Request

- Adds Bootstrap SCSS and `react-bootstrap`
  - Sets up FxA breakpoints
  - Fully supported in Storybook
- Adds Stylelint, `lint:styles`
- Documents how to write styles
- Housekeeping: put components in `src/components/`


app/experimenter/nimbus-ui/package.json
```diff
@@ -7,16 +7,20 @@
     "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
     "build": "SKIP_PREFLIGHT_CHECK=true react-scripts build",
     "test": "SKIP_PREFLIGHT_CHECK=true react-scripts test --coverage",
+    "lint": "yarn lint:eslint && yarn lint:tsc && yarn lint:styles",
```
jaredlockhart, line 10: If this is our lint now then we should update the alias in the Makefile as well:

https://github.com/mozilla/experimenter/blob/main/Makefile#L34

app/experimenter/nimbus-ui/src/components/App/index.tsx
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from "react";
-import AppLayout from "./AppLayout";
+import AppLayout from "../AppLayout";
```
jaredlockhart, line 6: Oh you don't do import path aliases so it always looks like `"/nimbus-ui/Applayout"` using absolute imports instead of relative ones?

jaredlockhart (changes requested):
Thnx @jodyheavener! Thanks for all the comprehensive docs, very nice 🎉  Looks good to me just gotta update that Makefile alias 👍

app/experimenter/nimbus-ui/src/components/App/index.tsx
LZoog: @jaredlockhart I did notice when going through `rapid` that you do absolute imports, on FxA we've done relative - I like relative since it's shorter (and in tests we can `import MyComponent from "."` - definitely up for hearing reasoning for preferring absolute :)

app/experimenter/nimbus-ui/.stylelintrc
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "stylelint-config-prettier",
+  "rules": {
+    "declaration-empty-line-before": "never",
+    "selector-max-id": [
+      0,
+      {
+        "message": "Using ID selectors is discouraged",
```
LZoog, line 8: 🎉

app/experimenter/nimbus-ui/README.md
```diff
@@ -28,6 +28,51 @@ This project uses [Storybook](https://storybook.js.org/) to visually show each c
...
+    </Container>
+    ```
+  - A [UI component](https://react-bootstrap.github.io/components/alerts/) encapsulates a category of UI, such as a button or alert banner, and tailors how it looks through the props. Example:
+    ```tsx
+    <Alert variant="danger">
+      This is a Danger alert!
+    </Alert>
+    ```
+
+**What about the JavaScript plugins?** While Bootstrap does provide optional JavaScript plugins, they are written in jQuery, which we are decidedly not using in this codebase. You are free to build your own equivelenet in vanilla JavaScript or as a React component, or introduce a new package (within reason).
```
LZoog, line 53: ~equivelenet~ equivalent

Also do we want to "allow" vanilla JS? I think that should be avoided and don't think we'll need it for this project, I think everything we need to do can just be done inside the React components.

app/experimenter/nimbus-ui/src/styles/index.scss
```diff
@@ -0,0 +1,17 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Looking to customize Bootstrap styles? Place SCSS variables
+// modifications *above* the Bootstrap import below. Learn more:
+// https://getbootstrap.com/docs/4.4/getting-started/theming/
+
+$grid-breakpoints: (
+  xs: 0,
+  sm: 480px,
+  md: 768px,
```
LZoog, line 12: I'm guessing there's no way to add the `min-height` of `481px` here to match FxA right? I thought that might be complex, if so I think it's totally fine to leave it out since mobile styles won't get much love on this project anyway.

LZoog (commented):
Looking good to me, thank you for the docs!

BTW, didn't check into the branch so just double checking classes and styles look as expected in Storybook?

app/experimenter/nimbus-ui/src/components/App/index.tsx
jodyheavener: I'm cool with either way, so long as we're consistent across the whole of nimbus-ui. I can make this change here if we'd like.

app/experimenter/nimbus-ui/README.md
jodyheavener: I'm not sure what you mean by allow -- by vanilla JavaScript I just meant any JS without a backing library. Did you have something else in mind?

app/experimenter/nimbus-ui/src/styles/index.scss
jodyheavener: Correct - this variable is used as a sizing value beyond breakpoints, so a min-height wouldn't be applicable here. However, this doesn't preclude us from doing something custom if we want.

> mobile styles won't get much love on this project anyway.

This makes me sad. Would like to discuss this more at some point.

jodyheavener:
> classes and styles look as expected in Storybook

🤦 I waited for your PR to merge and didn't even add this. Updating now.

app/experimenter/nimbus-ui/src/components/App/index.tsx
jaredlockhart: Oh yeah if it's not a pattern you typically prefer no need to block on it.  I just like absolute because it makes it easier to search for stuff when you're looking for every reference to something, especially if you move stuff around, but TS is also better about keep track of that so maybe it's a moot point.  Anyways don't change it on my account, I was just curious :)

jaredlockhart (approved):
LGTM 👍

LZoog (approved):
LGTM :D

PR #3557, opened 2020-09-28
fixes #3399 chore: Change dependabot to update yarn.lock

### LINKED ISSUE #3399: Change dependabot config to update yarn.lock

jaredlockhart:
Becuase

* Dependabot was configured to look at each npm dir individually it was missing the global yarn.lock

This commit

* Changes dependabot to look at root and it should pick up each workspace and update yarn.lock

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,17 +1,7 @@
 version: 2
 updates:
 - package-ecosystem: npm
-  directory: "/app/experimenter/nimbus-ui"
-  schedule:
-    interval: weekly
-  target-branch: dependencies
-- package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


jodyheavener (approved):
Dunno what the changes mean, but if it works it's cool with me!

LZoog (approved):
Sweeeeet

PR #3559, opened 2020-09-28
fixes #3503 chore: Run circle on dependency prs

### LINKED ISSUE #3503: Modify dependabot workflow to run CI on each pr before merge to dependencies
So maybe we can find a middle ground between the large number of CI runs we had when we CI'ed every dependabot PR onto main, and the single CI run with the current workflow.   Looking at the last couple weeks, there's been multiple cases where individual dependency changes break some part of the build/tests and then chasing them down is very time consuming.  

So here's the workflow I propose

1) Dependabot is configured to create PRs against `dependencies` branch and automatically merge them if they're successful
2) Circle is configured to do a full build/integration run against each individual dependabot PR as they are created against the `dependencies` branch
3) There are no branch protections against `dependencies` branch, such as `PRs must be up to date` like we have on `main`, which means each individual PR can be merged without needing to rebase after another one merges
4) Dependabot PRs that break build/tests will get stuck as a PR and can be inspected individually and either closed or modified to pass individually
5) When all dependabot PRs (minus ones that get stuck and closed) are merged into `dependencies` a single PR can be manually created against `main`
6) The rest of the current flow remains the same

This will result in more CI runs than the current flow but less than the original flow, and should hopefully be smoother than the workflow we have now.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-322)
┆Issue Number: EXP-322

jaredlockhart:
Because

* Running circle jobs against each individual dependency PR makes it easier to know
  if a single dependency change breaks the build rather than having to only run tests
  against all dependencies combined

This commit

* Enables circle jobs on dependency PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,16 +72,12 @@ workflows:
     jobs:
       - build:
           name: build
-          filters:
-            branches:
-              ignore: /dependabot.*/
       - integration:
           name: integration
           filters:
             branches:
               ignore:
                 - main
-                - /dependabot.*/
       - deploy:
           requires:
             - build
--- a/contributing.md
+++ b/contributing.md
@@ -69,14 +69,19 @@ Dependencies are automatically updated by [Dependabot](https://dependabot.com/) 
 directly into GitHub.  Each week Dependabot will create a large number of individual PRs that update
 each dependency.  To merge those into main, use the following process:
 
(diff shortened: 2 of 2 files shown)


jodyheavener:
Does this need a doc update?

jaredlockhart:
@jodyheavener I did update the docs

jodyheavener (approved):
sorry @jaredlockhart not sure how I missed the doc change.

PR #3574, opened 2020-09-28
feat: add graphql setup

### LINKED ISSUE #3490: Add Graphene for GQL API
Let's add [Graphene]([https://graphene-python.org/)](https://graphene-python.org/)) to expose a GQL API to support the new frontend.  I think in keeping with our API versioning scheme this could could go under

`/api/v5/graphql`

I think we only need to expose the NimbusExperiment model, and then as we add more related models for branches, etc we can add them to this API.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-312)
┆Issue Number: EXP-312

bbangert:
Because:

* We want to use graphql for nimbus.

This commit:

* Add's basic django-graphene integration with most of the models
  represented.

Closes #3490

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -1,0 +1,39 @@
+from graphene_django.types import DjangoObjectType
+
+from experimenter.experiments.models.nimbus import (
+    NimbusBranch,
+    NimbusBucketRange,
+    NimbusExperiment,
+    NimbusIsolationGroup,
+)
+from experimenter.projects.models import Project
+
+
+class NimbusBranchType(DjangoObjectType):
+    class Meta:
+        model = NimbusBranch
+        exclude = ("id", "experiment", "nimbusexperiment")
+
+
+class NimbusExperimentType(DjangoObjectType):
+    class Meta:
+        model = NimbusExperiment
+        exclude = ("id",)
+
+
+class ProjectType(DjangoObjectType):
+    class Meta:
+        model = Project
+        exclude = ("id",)
+
+
+class NimbusIsolationGroupType(DjangoObjectType):
+    class Meta:
+        model = NimbusIsolationGroup
+        exclude = ("id",)
+
+
+class NimbusBucketRangeType(DjangoObjectType):
+    class Meta:
+        model = NimbusBucketRange
+        exclude = ("id", "experiment")
--- a/app/experimenter/experiments/api/v5/urls.py
+++ b/app/experimenter/experiments/api/v5/urls.py
@@ -1,0 +1,11 @@
+from django.conf.urls import url
+from django.views.decorators.csrf import csrf_exempt
+from graphene_django.views import GraphQLView
+
+urlpatterns = [
+    url(
+        r"graphql",
+        csrf_exempt(GraphQLView.as_view(graphiql=True)),
+        name="nimbus-api-graphql",
+    ),
+]
--- a/app/pyproject.toml
+++ b/app/pyproject.toml
@@ -53,8 +53,10 @@ mozilla-nimbus-shared = "0.0.9"
 isort = "5.5.2"
 google-cloud-storage = "1.31.2"
 django-storages = "1.7.1"
+graphene-django = "^2.13.0"
 
 [tool.poetry.dev-dependencies]
+rope = "^0.17.0"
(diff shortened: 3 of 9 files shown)


app/experimenter/experiments/api/v5/types.py
```diff
@@ -0,0 +1,39 @@
...
+    NimbusExperiment,
+    NimbusIsolationGroup,
+)
+from experimenter.projects.models import Project
+
+
+class NimbusBranchType(DjangoObjectType):
+    class Meta:
+        model = NimbusBranch
+        exclude = ("id", "experiment", "nimbusexperiment")
```
jaredlockhart, line 15: I usually do it the other way around and explicitly list fields rather than using exclude, but then I was thinking about it and I realized that makes sense for a REST API where you only care about some given subset, but for GQL you want the other way around and just expose the whole kit'n'kaboodle.

app/experimenter/experiments/api/v5/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.conf.urls import url
+from django.views.decorators.csrf import csrf_exempt
+from graphene_django.views import GraphQLView
+
+urlpatterns = [
+    url(r"graphql", csrf_exempt(GraphQLView.as_view(graphiql=True)), name="graphql"),
```
jaredlockhart, line 6: Ty 🙏

jaredlockhart: Oh name should be qualified with the app as well since it's a single global namespace, so something like

`nimbus-api-graphql`

or wahtever

app/pyproject.toml
```diff
@@ -53,8 +53,10 @@ mozilla-nimbus-shared = "0.0.9"
 isort = "5.5.2"
 google-cloud-storage = "1.10.0"
 django-storages = "1.7.1"
+graphene-django = "^2.13.0"
 
 [tool.poetry.dev-dependencies]
+rope = "^0.17.0"
```
jaredlockhart, line 59: Sneaking rope in? ;D  Does this integrate with vscode?  Or you just call it from the command line?

jaredlockhart (approved):
Awesome looks great @bbangert :D

app/pyproject.toml
bbangert: I honestly have no idea what put that in. It's a mystery to me as well.

app/experimenter/experiments/api/v5/types.py
bbangert: Right, the client-side is going to request whatever subset it cares about. Since this is an internal tool it feels a bit faster on turn-around to exclude up front.

app/pyproject.toml
bbangert: Ok, I found out what happened. I guess I tried to use a VS Code refactor, and it said, "Mind if I install a Python refactor tool?", and I was like, "yea, whatever", and so VS Code installed it and made a record of it for poetry.

pdehaan: Did we mean to revert the `rope` dev-dependency before merging this?

bbangert: I was not intending to, as its a useful dependency for VS Code. Due to the container setup its a major pain to always be re-installing this manually.

pdehaan:
Sorry, I'm not super fluent in GraphQL, but now that this has [landed on stage](https://stage.experimenter.nonprod.dataops.mozgcp.net/api/v5/graphql#query=query%20%7B%0A%20%20allExperiments%20%7B%0A%20%20%20%20name%0A%20%20%20%20slug%0A%20%20%20%20publicDescription%0A%20%20%7D%0A%7D%0A) :tada:, I just wanted to confirm this behavior in the GraphiQL UI:



### INPUT

```gql
query {
  allExperiments {
    name
    slug
    publicDescription
  }
}
```

### OUTPUT

```json
{
  "data": {
    "allExperiments": []
  }
}
```

I wasn't expecting an empty array of `allExperiments`, but wasn't sure if I possibly copy/pasted the query language from the wrong place in the commits above, or if there possibly be needs to be a bit more plumbing.

bbangert:
@pdehaan yup, if there are no results, then an empty result set will be returned. Note that this is only querying Nimbus experiments.

PR #3578, opened 2020-09-28
fix #3577 chore: Use pip installer to speed up build times

### LINKED ISSUE #3577: Improve build speed with poetry export
`poetry install` is noticeably slower than `pip install`, on my local machine for Experimenter it comes out to:

`poetry install`: 3m
`pip install`: 50s

So I as just reading through https://github.com/python-poetry/poetry/issues/338 and someone suggested doing

`poetry export -f requirements.txt --dev | pip install -r /dev/stdin`

See: https://github.com/python-poetry/poetry/issues/338#issuecomment-546813466

So let's do that until their new parallel installer:

https://github.com/python-poetry/poetry/pull/2595

is released.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-361)
┆Issue Number: EXP-361

jaredlockhart:
Because

* poetry install takes significantly longer to install python packages

This commit

* Uses poetry to generate a requirements file and then uses pip to actually install everything
* I made a node that we should revert this when the new parallel poetry installer comes out

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -35,7 +35,8 @@ RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-
 ENV PATH "/root/.poetry/bin:${PATH}"
 RUN poetry config virtualenvs.create false
 COPY poetry.lock pyproject.toml ./
-RUN poetry install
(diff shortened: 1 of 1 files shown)


bbangert (approved):
Mmmm, faster installs....

PR #3629, opened 2020-10-01
fix #3628 chore: Use poetry install instead of export

### LINKED ISSUE #3628: Use poetry parallel installer
Poetry 1.1.0 just came out and includes the new parallel installer, which runs in 42s on my machine which beats the 50s from the export to pip method.  Ironically the export method also broke so we have no choice but to revert back to regular `poetry install`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-430)
┆Issue Number: EXP-430

jaredlockhart:
Because

* Poetry released their new version with the fast parallel installer
* And broke the export function we were using

This commit

* Reverts back to using poetry install
* Updates the poetry lockfile

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -35,8 +35,7 @@ RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-
 ENV PATH "/root/.poetry/bin:${PATH}"
 RUN poetry config virtualenvs.create false
 COPY poetry.lock pyproject.toml ./
-# Temporary speedup hack until Poetry releases their parallel installer, ref: https://github.com/python-poetry/poetry/issues/338#issuecomment-546813466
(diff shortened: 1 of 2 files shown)


bbangert (approved):

app/Dockerfile
```diff
@@ -36,7 +36,7 @@ ENV PATH "/root/.poetry/bin:${PATH}"
 RUN poetry config virtualenvs.create false
 COPY poetry.lock pyproject.toml ./
 # Temporary speedup hack until Poetry releases their parallel installer, ref: https://github.com/python-poetry/poetry/issues/338#issuecomment-546813466
```
pdehaan, line 38: Is this comment still valid, or do we need to ✂️ ✂️ &nbsp; ?

jaredlockhart: Lol @pdehaan what would I do without you

PR #3637, opened 2020-10-02
fix #3425 chore(project): Add linux non root instructions for Docker …

### LINKED ISSUE #3425: Add non-root user setup for Linux users
Docker on Linux may setup the docker daemon as root, so we should add instructions to the docs:

https://docs.docker.com/engine/security/rootless/

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-278)
┆Issue Number: EXP-278

jaredlockhart:
…to README

Because

* Docker will by default use root permissions on linux

This commit

* Adds instructions on how to run docker as non root on linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,11 @@ Experimenter is a platform for managing experiments in [Mozilla Firefox](https:/
 
 ## Installation
 
-Node ^14.0.0 is required
-
 ### General Setup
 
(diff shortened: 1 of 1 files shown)


bbangert (approved):
Oooo, useful! I'm going to do this right now. 👍

PR #3642, opened 2020-10-02
fix #3634 chore(project): Use separate build and deploy Docker images

### LINKED ISSUE #3634: Use separate build and deploy Docker images
To speed up and improve the consistency of our deploys, we should separate our Docker image into two stages, one for build and one for deploy.  The current image is 1gb, and 821mb of that is `node_modules` so if we just nuke that in the deploy image we should be able to bring it way down.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-435)
┆Issue Number: EXP-435

jaredlockhart:
Because

* We want to shrhink the size of the deployable Docker image down from its current 2.71gb(1gb compressed)

This commit

* Adds a second stage to our Dockerfile for a deploy image which is a much more reasonable 480mb(133mb compressed)
* Use deploy image for Django container in local dev compose

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
           command: |
             ./scripts/build.sh
             docker login -u $DOCKER_USER -p $DOCKER_PASS
-            docker tag app:build ${DOCKERHUB_REPO}:latest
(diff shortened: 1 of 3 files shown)


robotblake (approved):
\o/

PR #3645, opened 2020-10-03
fix #3633 feat: add create experiment mutation

### LINKED ISSUE #3633: Experiment creation mutation
Mutation that creates an experiment, which only requires the public name, hypothesis, and the targeted application.

Per screenshot 2 of [https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216037%3A1694](https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216037%3A1694)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-434)
┆Issue Number: EXP-434

bbangert:
Because:

* We want to create experiments via GraphQL.

This commit:

* Add's a graphql mutation for createExperiment.

Closes #3633

### DIFF
--- a/app/experimenter/experiments/api/v5/forms.py
+++ b/app/experimenter/experiments/api/v5/forms.py
@@ -1,0 +1,9 @@
+from django import forms
+
+from experimenter.experiments.models.nimbus import NimbusExperiment
+
+
+class CreateExperimentForm(forms.ModelForm):
+    class Meta:
+        model = NimbusExperiment
+        fields = ("name", "slug", "hypothesis", "public_description")
--- a/app/experimenter/experiments/tests/api/v5/test_mutation.py
+++ b/app/experimenter/experiments/tests/api/v5/test_mutation.py
@@ -1,0 +1,55 @@
+import json
+
+from django.conf import settings
+from django.urls import reverse
+from graphene_django.utils.testing import GraphQLTestCase
+
+from experimenter.experiments.models.nimbus import NimbusExperiment
+
+
+class TestCreateExperiment(GraphQLTestCase):
+    GRAPHQL_URL = reverse("nimbus-api-graphql")
+
+    def test_create_experiment(self):
+        user_email = "user@example.com"
+        response = self.query(
+            """
+            mutation($input: CreateExperimentMutationInput!) {
+                createExperiment(input: $input) {
+                    nimbusExperiment {
+                        status
+                        name
+                        slug
+                    }
+                    errors {
+                        field
+                        messages
+                    }
+                    clientMutationId
+                }
+            }
+            """,
+            variables={
+                "input": {
+                    "name": "test1234",
+                    "slug": "slug1234",
+                    "clientMutationId": "randomid",
+                }
+            },
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        result = content["data"]["createExperiment"]
+        experiment = result["nimbusExperiment"]
+        self.assertEqual(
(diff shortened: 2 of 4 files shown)


app/experimenter/experiments/tests/api/v5/test_mutation.py
```diff
@@ -0,0 +1,50 @@
...
+        result = content["data"]["createExperiment"]
+        experiment = result["nimbusExperiment"]
+        self.assertEqual(
+            experiment, {"status": "DRAFT", "name": "test1234", "slug": "slug1234"}
+        )
+
+        self.assertEqual(result["clientMutationId"], "randomid")
+
+        errors = result["errors"]
+        self.assertEqual(errors, [])
```
jaredlockhart, line 50: We would also expect that the model is created here right?  If so then we should also look up the model instance and check that the fields are set.

Also in general will the GQL interface use slugs as primary identifiers for everything since we hid the db ids?  How will that work for creating related models, like branches?

app/experimenter/experiments/api/v5/forms.py
```diff
@@ -0,0 +1,9 @@
+from django import forms
+
+from experimenter.experiments.models.nimbus import NimbusExperiment
+
+
+class CreateExperimentForm(forms.ModelForm):
+    class Meta:
+        model = NimbusExperiment
+        fields = ("name", "slug", "hypothesis", "public_description")
```
jaredlockhart, line 9: OOOOOOOHHHHH it uses FORMS?? Oooooohhh.  Well that's nice, that's a very nice pattern to do validation!  Heck we even have mixins for doing changelogs automatically in forms so we can reuse that.  That's so interesting.

jaredlockhart (changes requested):
Wow this is so nice and elegant, I love that it uses forms.  So do we expect that we'll do server side validation at the form level?

app/experimenter/experiments/tests/api/v5/test_mutation.py
bbangert: Returning the model requires it to save correctly, so I figured if the result came back it must be in the database or there would've been an error. I can add a bit to test that it went into the db as expected, at that point we are testing that library code from graphene-django works as advertised though.

Yes, its assumed we're using slugs as primary identifiers. I'm not sure what you mean about how it works for branches.

bbangert:
Yes, we can validate requirements beyond GraphQL types (aka, a specific type of string vs. String in general) using django form restrictions.

app/experimenter/experiments/tests/api/v5/test_mutation.py
jaredlockhart: We can worry about the branch stuff later.  And yeah I agree it's testing the internals but in general for this kind of behaviour, ie hit the API with some data, expect a change to the db, I would always still explicitly assert the db changes to test the integration.  Ie if the form were changed from a ModelForm to a Form, it could still pass this test but not do any db inserts.

bbangert: Okey, added the db lookup to verify it was added.

app/experimenter/experiments/tests/api/v5/test_mutation.py
```diff
@@ -0,0 +1,55 @@
...
+
+        exp = NimbusExperiment.objects.first()
+        self.assertEqual(exp.slug, "slug1234")
```
jaredlockhart, line 55: Ty 👍

jaredlockhart (approved):
👍 👍 👍

PR #3668, opened 2020-10-05
fixes #3630 feat(backend): add status experiment filter

### LINKED ISSUE #3630: Create experiment query with status filter
Nimbus experiments need to be able to be located by status code. A query needs to be added that supports filtering by status code.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-431)
┆Issue Number: EXP-431

bbangert:
Because:

* We want to filter experiments by their status for display.

This commit:

* Adds a new experimentsByStatus query.

Closes #3630

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_query.py
+++ b/app/experimenter/experiments/tests/api/v5/test_query.py
@@ -79,3 +79,31 @@ class TestExperimentListView(GraphQLTestCase):
         content = json.loads(response.content)
         experiment = content["data"]["experimentBySlug"]
         self.assertIsNone(experiment)
+
+    def test_experiments_by_status(self):
+        user_email = "user@example.com"
+        draft_exp = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.DRAFT
+        )
+        NimbusExperimentFactory.create_with_status(NimbusExperiment.Status.ACCEPTED)
+
+        response = self.query(
+            """
+            query {
+                experimentsByStatus(status: DRAFT) {
+                    name
+                    slug
+                    publicDescription
+                }
+            }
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -45,11 +45,12 @@ class Meta:
         model = NimbusExperiment
 
     @classmethod
-    def create_with_status(cls, target_status, **kwargs):
+    def create_with_status(cls, target_status=NimbusExperiment.Status.DRAFT, **kwargs):
```
jaredlockhart, line 48: I wouldn't use a default argument here, we've always tried to be explicit about the status a factory experiment is created with right in the test for readability.

app/experimenter/experiments/tests/api/v5/test_query.py
```diff
@@ -79,3 +79,54 @@ def test_experiment_by_slug_not_found(self):
...
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        experiments = content["data"]["experimentsByStatus"]
+        self.assertEqual(len(experiments), 1)
+        for key in experiments[0]:
+            self.assertEqual(
+                experiments[0][key], str(getattr(accepted_exp, to_snake_case(key)))
+            )
```
jaredlockhart, line 132: I don't think this is really necessary here?  Should be sufficient to just test against a single status, shouldn't it?

jaredlockhart (approved):
I love how clean these GQL interfaces are 💯 💯 💯

PR #3670, opened 2020-10-05
fixes #3429 feat(nimbus): serve nimbus app with rest of setup, consume server config

### LINKED ISSUE #3429: Allow consuming server settings values on client, serve app at `/nimbus`

jodyheavener:
Because:

- We need to be able to visit the Nimbus React app when the rest of the Docker setup is running
- We need the Nimbus React app to be able to consume configuration values defined on the server side and by env vars

This PR:

- Sets up Nimbus React app to be accessible at `https://localhost/experiments/nimbus/`, as is consistent with the Rapid setup
- Sets up that Nimbus view to use DEBUG and APP_CONFIG context variables.
- Uses those new variables to conditionally render development/prod assets.
- Encodes and renders the app configuration in the HTML template
- Decodes and parses the app configuration in the React app, making it available across the whole app
- Adds Rescripts - disables chunking/splitting so we can reliably find the JS files in the Django template

Closes #3429


app/experimenter/experiments/views.py
```diff
@@ -233,3 +233,15 @@ def form_invalid(self, form):
...
+class ExperimentNimbusUIView(TemplateView):
+    template_name = "experiments/nimbus-ui.html"
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context["nimbus_dev"] = settings.NIMBUS_DEV
+        context["app_config"] = json.dumps({
+            "sentry_dsn": settings.SENTRY_DSN,
+        })
+        return context
```
jaredlockhart, line 247: I already did this for you 😃 

https://github.com/mozilla/experimenter/blob/main/app/experimenter/base/context_processors.py#L14

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
```diff
@@ -0,0 +1,38 @@
+{% load static %}
+{% load experiment_extras %}
+
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8" />
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  {% if nimbus_dev %}
```
jaredlockhart, line 10: Could we just use `DEBUG` for this rather than creating a second setting?

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
```diff
@@ -0,0 +1,38 @@
...
+    <link rel="icon" href="http://localhost:3000/favicon.ico" />
```
jaredlockhart, line 11: If this is the same global `favicon.ico` that the rest of the site uses, this should always be available in `static` shouldn't it?

jodyheavener: HEY. DIS A DRAFT.

jaredlockhart: EarlyHealthyEsok-small

app/experimenter/experiments/views.py
jodyheavener: @jaredlockhart can you elaborate on how this is used? My understanding is this makes each key available in the template (e.g. `SENTRY_DSN` or `FEATURE_MESSAGE_TYPE`), but I'm looking for the whole object to be readily available at render time so I can encode the whole thing. Maybe I'm misunderstanding.

jodyheavener: Unless we want each unique setting to be its own data-attribute? I can get behind that

jaredlockhart: Yeah exactly, this exposes it as a template var, so you could pass it in as a data attribute, or if you wanna pass it in as a whole object you could either do what you were doing with dumping it into JS here the same way you were doing in the view, or you can build the object in the template like

```js
<script>
  const config = {
    SENTRY_DSN: "{{ SENTRY_DSN }}",
  };
  App.init(config);
</script>
```

Or if you wanna only do it in the view you can just keep it how you have it in the view and we can rip it out of here, there's just no need to have it in one place.

jodyheavener: Okay cool. I think I'll avoid setting up an object and initializing from the page because we already init from the script automatically, and it would involve exposing an object that the page could call init on. But I did update to move the whole config under the context processor.

app/experimenter/experiments/views.py
```diff
@@ -233,3 +233,12 @@ def form_invalid(self, form):
...
     template_name = "experiments/rapid.html"
+
+
+class ExperimentNimbusUIView(TemplateView):
+    template_name = "experiments/nimbus-ui.html"
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context["debug"] = settings.NIMBUS_DEV
+        return context
```
jodyheavener, line 244: @jaredlockhart tests keep yelling at me that this isn't getting tested, but afaict it is [here](https://github.com/mozilla/experimenter/pull/3670/files#diff-d62dba79b6a4e33ab663cf5e648aee8fR826). What would I be missing?

jaredlockhart: Oh if you look up you'll see that the test is failing becuase `experiments-nimbus` isn't a valid url name, looks like you forgot to rename it here:

https://github.com/mozilla/experimenter/pull/3670/files#diff-23077f4ae423069e9d89856456c09167R23-R25


The coverage is complaining those lines aren't covered becuase the test failed, so those lines never get hit.  Usually you wanna look at test failures first, then if they're all passing, then look at coverage.

jodyheavener: Ahhhh thank you. My bad!

jodyheavener:
Aiiiight this took longer than I'd hoped, but have at it.

app/experimenter/nimbus-ui/src/index.tsx
```diff
@@ -6,10 +6,18 @@ import React from "react";
...
+try {
+  const root = document.getElementById("root")!;
+  readConfig(root);
+
+  ReactDOM.render(
+    <React.StrictMode>
+      <App />
+    </React.StrictMode>,
+    root
+  );
```
jaredlockhart, line 20: I'm confused about this pattern, how does the `App` read the config?  Is it a hidden shared state inside the config module?

app/experimenter/nimbus-ui/src/lib/config.ts
```diff
@@ -0,0 +1,73 @@
...
+      delete (config as any)[key];
+    }
+  });
+}
+
+export function update(newData: { [key: string]: any }) {
+  deepMerge(config, newData);
+}
+
+const config: Config = getDefault();
```
jaredlockhart, line 72: So this treats `config` as a mutable global singleton?  Is there not a way to just pass it in to the react init as a paramter or a prop or something?

jaredlockhart (commented):
Looks generally good, but it seems like a lot of logic for this use case.  The config object has a single key right now, is using `deepMerge` necessary?

app/experimenter/experiments/templatetags/experiment_extras.py
```diff
@@ -48,3 +49,8 @@ def pagination_url(context, page, **kwargs):
 @register.filter
 def as_json(value):
     return json.dumps(json.loads(value), indent=2)
+
+
+@register.filter
+def encode_uri(data):
+    return urllib.parse.quote(data)
```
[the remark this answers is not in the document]
jodyheavener: I can dig it

.env.sample
```diff
@@ -2,6 +2,7 @@ BUGZILLA_API_KEY=
 BUGZILLA_CC_LIST=
 BUGZILLA_HOST=https://bugzilla.allizom.org/
 CELERY_SCHEDULE_INTERVAL=300
+COMPOSE_HTTP_TIMEOUT=120
```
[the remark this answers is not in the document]
jodyheavener: Oh I can totally remove it, I just realized I have to use it every time and figured it was a handy default. Will remove.

app/experimenter/experiments/views.py
```diff
@@ -233,3 +233,12 @@ def form_invalid(self, form):
...
     template_name = "experiments/rapid.html"
+
+
+class ExperimentNimbusUIView(TemplateView):
+    template_name = "experiments/nimbus-ui.html"
+
+    def get_context_data(self):
+        context = super().get_context_data()
+        context["debug"] = settings.DEBUG
+        return context
```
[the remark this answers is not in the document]
jodyheavener: Huh, I did try that but it doesn't appear to be coming through for me. It renders nothing. Referencing the linked docs, my local DEBUG is True and I'm accessing from our standard localhost address.

app/experimenter/nimbus-ui/src/index.tsx
jodyheavener: When `readConfig` is called above the [config lib](https://github.com/mozilla/experimenter/pull/3670/files#diff-c98cd03bc4fd3949ecfd9580d0d2bc68R19-R24) accesses the encoded data attr value on the mounting element, parsing it and adding it to the config object that can then be imported anywhere in the app.

app/experimenter/nimbus-ui/src/lib/config.ts
jodyheavener: It could certainly be passed in to the React app, but then you need to continue passing it down to each component that uses or has children that use it. We could also store it in Context if we wanted to avoid the whole component tree. Open to suggestions. Tagging @LZoog as well.

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
```diff
@@ -0,0 +1,32 @@
...
+<body>
+  <noscript>Mozilla Experimenter requires JavaScript to run.</noscript>
+  <div id="root" data-config="{{ APP_CONFIG|encode_uri }}"></div>
+
+  {% if debug %}
+    <script src="http://localhost:3000/static/js/bundle.js"></script>
+    <script src="http://localhost:3000/static/js/0.chunk.js"></script>
+    <script src="http://localhost:3000/static/js/main.chunk.js"></script>
+  {% else %}
+    {# TODO: this will not work as CRA generates multiple JS files #}
```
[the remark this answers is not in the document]
jodyheavener: Ack, yeah. When the build command is run react-scripts compiles everything into what looks like 3, but might vary, fingerprinted scripts that we'll need to render. Generally the CRA build process takes care of modifying the HTML file, but things are a little different for us. I suppose we'll need to traverse the build folder for `*.js` files and print them in to this template.

Screen Shot 2020-10-06 at 8 25 46 PM

jodyheavener:
> The config object has a single key right now, is using deepMerge necessary?

At the moment, no. This comes from FXA, where we have historically passed in objects as well. If we think this will never be the case I'm totally cool to remove it :)

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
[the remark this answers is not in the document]
jodyheavener: Naively noodling around this morning, could we glob it and render each returned file (would need to also modify each result to replace local path with URL I suppose):

```python
current = pathlib.Path().absolute()
listing = glob.glob(os.path.join(current, '../nimbus-ui/build/static/js/*.js'))
```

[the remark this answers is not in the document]
jodyheavener: Alternatively, it looks like [there](https://github.com/facebook/create-react-app/issues/3365#issuecomment-376546407) are [ways](https://stackoverflow.com/questions/49123097/generate-single-physical-javascript-file-using-create-react-app) to override or augment the create-react-app Webpack config if we really want to generate a single output file. We've done this in FxA, and it's a little tricky in some respects, but it sounds do-able.

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
jaredlockhart: Hrmm k let me poke at it

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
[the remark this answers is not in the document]
jaredlockhart: Yeah I'd think finding some way to force it to a single output file would be preferred long term solution instead of having to try to discover the randomly generated ones every time.

app/experimenter/nimbus-ui/src/lib/config.ts
jaredlockhart: Right yeah, the ol prop hot potato pattern.  I've attached things to `window` to solve this in the past, like `window.nimbus_config` or some such, that way it's easy to poke at while debugging and it's obvious where it's being stored, but this works too.

jaredlockhart:
@jodyheavener Yeah looking at our other react frontends we never passed more than a handful of keys across, so we can add something like deepmerge later if we need it :)

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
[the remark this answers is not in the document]
jodyheavener: Will try it out and report back :)

jaredlockhart (commented):
Yeah I'd say if we don't need `deepMerge` now it's not worth landing, and let me see if I can figure out what's going on with `debug`

Thnx @jodyheavener this is looking really good!

jaredlockhart:
Also @jodyheavener you can use `make code_format` to reformat everything in a way that should let `make check` pass (at least for formatting concerns)

app/experimenter/experiments/views.py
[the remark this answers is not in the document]
jaredlockhart: Okay I figured it out, I remember this now, so annoying.

It's cuz there's a setting called `INTERNAL_IPS` where you're supposed to hard code a list of pre-approved IPs for certain debug info as a security precaution, but it always screws everything up.  People always hack around this, like setting it to an object that just always returns `True`:

```py
if DEBUG:
    # `debug` is only True in templates if the vistor IP is in INTERNAL_IPS.
    INTERNAL_IPS = type(str('c'), (), {'__contains__': lambda *a: True})()
```

But that's a janky mess, so I would say just add

```py
    "DEBUG": settings.DEBUG
```

into `experimenter.base.context_processors` and be done with it XD

jaredlockhart:
Commit formatting nit:

`fix #bugnum type(scope): title`

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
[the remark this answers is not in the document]
jodyheavener: Think I've got something working.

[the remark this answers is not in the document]
jaredlockhart: Oh sick 🎉

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
```diff
@@ -0,0 +1,30 @@
+{% load static %}
+
+<!DOCTYPE html>
```
LZoog, line 3: Would it be possible for this to live somewhere outside of `legacy-ui` since we want to delete this directory down the line? I'm not sure of the best placement, I mean even `app/experimenter/nimbus-ui.html` could make sense as it'd be adjacent to `nimbus-ui/`.

app/experimenter/nimbus-ui/README.md
```diff
@@ -6,6 +6,27 @@
...
+
+Nimbus uses two approaches for retrieving configuration:
+
+#### Template-rendered configuration
+
+This is used to ensure critical information about the app is retrieved and made available immediately. These details are encoded and rendered in the HTML when the main [Django template](../legacy-ui/templates/experiments/nimbus-ui.html) for loading `nimbus-ui` is requested, and are subsequently parsed by the React app when it is initialized.
+
+The values for this configuration are defined in the [`APP_CONFIG`](../base/context_processors.py) custom context processor.
+
+Accessing these configuration values in the React app is simple: just import the `config` object and access its properties. Example:
```
LZoog, line 19: Yaaas documentation 🎉

app/experimenter/nimbus-ui/public/index.html
```diff
@@ -1,20 +1,5 @@
...
-  <body>
-    <noscript>Mozilla Experimenter requires JavaScript to run.</noscript>
-    <div id="root"></div>
-  </body>
-</html>
+<!--
+  NOTICE: This HTML file is no longer used. To modify the HTML file used
+  for this React app please make changes to the following file:
+  app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
+-->
```
LZoog, line 5: I like the idea of leaving this around with a comment👍

app/experimenter/nimbus-ui/src/lib/config.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    expect(config.version).toBeDefined();
+    expect(config.version).toStrictEqual(data.version);
+  });
+});
+
+describe('decode', () => {
+  it('converts a JSON-stringified, URI-encoded string to a proper object', () => {
+    const data = {
+      decimated: 'dreams',
```
LZoog, line 34: 😆

app/experimenter/nimbus-ui/webpack.config.js
```diff
@@ -0,0 +1,21 @@
+const path = require("path");
+const glob = require("glob");
+const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
+
+// NOTICE: This is a suplementary Webpack config used for production
+// building our JavaScript, and is not the config create-react-app
+// uses for running and building the React app.
```
LZoog, line 7: This file is missing the license.

We could name this `webpack.config.prod.js` which is conventional when there's a dev/prod version of webpack but I think it's also fine to leave with the comment. It could also be useful to add _why_ we need this like "we conditionally require JS files in `nimbus-ui.html` which enables us to pass server configuration to the client and reduce them to one here" for extra breadcrumbs but not necessary.

This sorta makes me wonder if we need to regenerate and also expose source maps though? I think Sentry uses those for error tracebacks. 😓

I guess bonus points would be removing all those files since there's now extra cruft in that `js` folder inside the build that won't be used, but meh. Something like this could be a good gulp/grunt task, but we already have webpack as a dependency, so hey less the better.

app/experimenter/nimbus-ui/src/lib/config.test.ts
jodyheavener: oh shit I forgot I left MCR lyrics in here

app/experimenter/nimbus-ui/src/lib/config.ts
LZoog: I'd like to avoid attaching global properties to `window` if possible but I also know sometimes that can be useful. We're doing something similar for `fxa-settings` so I think leaving it is fine🤔

app/experimenter/legacy-ui/templates/experiments/nimbus-ui.html
jaredlockhart: Oh yeah we can put it in `nimbus-ui/templates` and just add it here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L94-L97

app/experimenter/nimbus-ui/src/lib/config.test.ts
lmorchard: Hey, I leave 8675309 all over my code, live a little

app/experimenter/nimbus-ui/public/index.html
jodyheavener: Also CRA blows up if it's not present :)

app/experimenter/nimbus-ui/webpack.config.js
jodyheavener: File name updated, note added, license added, leftover build files removed after joining. Unsure about source maps right now, but there's one issue I'm stuck on:

I can't get this to actually work. I am joining the CSS/JS files in the same order CRA places them if they were individually rendered in the HTML, they're not being additionally minified, the JS file loads on the page, and there are no console errors, but in testing the React app doesn't start (blank page). My theory at the moment is that Webpack is adding some additional bundling mechanics to the mix that is making the React script not work. But I'm officially looking for input if anyone has any (cc @jaredlockhart and @lmorchard).

(Context for others: in production I'm using [Webpack to bundle](https://github.com/mozilla/experimenter/pull/3670/files#diff-ff3e1bc0781aa8c83a318f2e5b9a3315) all of CRA's output files in to a single `bundle.min.js` so it can be rendered in [the template](https://github.com/mozilla/experimenter/pull/3670/files#diff-23698a628d30539e82056573789ef0eaR27).)

jodyheavener: FWIW this all works as expected in development, so if we want to merge this to unblock other stuff I can follow up.

jaredlockhart: This is beyond my limited knowledge 😞

jodyheavener: Cancel this. Taking a different approach.

jodyheavener:
Let's try this again @LZoog (and @jaredlockhart if you'd like!) I added a bit more documentation for Rescripts as well.

app/experimenter/nimbus-ui/templates/nimbus-ui.html
```diff
@@ -0,0 +1,28 @@
+{% load static %}
```
jaredlockhart, line 1: Oh good yes, the only thing I'd maybe suggest is instead of having it in the template root, putting it in a subpath so referring to it later in views is less ambiguous, like 

`templates/nimbus/index.html`  

so the view would say

`template_name = "nimbus/index.html"`

But that's just a suggestion 👍

jaredlockhart (approved):
Looks great @jodyheavener thnx for addressing all the feedback! :D

app/experimenter/nimbus-ui/templates/nimbus-ui.html
jodyheavener: I can dig it :)

app/experimenter/nimbus-ui/templates/nimbus/index.html
```diff
@@ -0,0 +1,28 @@
...
+  <meta name="viewport"
+    content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes" />
+  <title>Mozilla Experimenter</title>
+</head>
+
+<body>
+  <noscript>Mozilla Experimenter requires JavaScript to run.</noscript>
+  <div id="root" data-config="{{ APP_CONFIG }}"></div>
+
+  {% if not DEBUG %}
```
jaredlockhart, line 22: @3670 sorry wait, shouldn't this be the other way around? if `DEBUG` we DO want to use localhost?

jaredlockhart (changes requested):

app/experimenter/nimbus-ui/templates/nimbus/index.html
jodyheavener: Damn, yeah, sorry that's a relic of testing the prod build. Fixed.

jaredlockhart (approved):
👏 

Also if you want to test with `DEBUG=False` you can change that in your `.env` file instead of needing to modify code 👍

jodyheavener:
I *restart* the app? You think I have a death wish??

PR #3676, opened 2020-10-05
fix #3428 feat(nimbus-ui): add AppErrorBoundary and Sentry for error logging

### LINKED ISSUE #3428: Add AppErrorBoundary and Sentry for error logging
See [https://github.com/mozilla/fxa/blob/main/packages/fxa-react/components/AppErrorBoundary/index.tsx](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/components/AppErrorBoundary/index.tsx) for the component built for `fxa-payments` that's also used in `fxa-settings`. It should be imported and wrapped around the application in `index.tsx`.

We should be able to reuse a lot of the Sentry TS code made for `fxa-settings`, see this PR: [https://github.com/mozilla/fxa/pull/5203](https://github.com/mozilla/fxa/pull/5203)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-282)
┆Issue Number: EXP-282

lmorchard:
Because we want to catch errors within the nimbus-ui React app, this
commit adds a Logger service wrapper for console.log and a Sentry
service to report errors back to us.

I copied over the React components and service code from FxA. Needed to convert tests from Mocha to Jest along the way and added some coverage to get to 100%. Also made an executive decision to use a `services` directory instead of the more generic `lib`

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppErrorDialog/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppErrorDialog/index.tsx
@@ -1,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+
+const AppErrorDialog = ({ error }: { error: Error }) => {
+  return (
+    <div className="bg-grey-20 flex items-center flex-col justify-center h-screen">
+      <div className="text-center max-w-lg">
+        <h2
+          className="text-grey-600 font-header text-lg font-bold mb-3"
+          data-testid="error-loading-app"
+        >
+          General application error
+        </h2>
+        <p className="text-grey-400">
+          Something went wrong. Please try again later.
+        </p>
+      </div>
+    </div>
+  );
+};
+
+export default AppErrorDialog;
--- a/app/experimenter/nimbus-ui/src/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/index.test.tsx
@@ -5,6 +5,7 @@
 import React from "react";
 import ReactDOM from "react-dom";
 import App from "./components/App";
+import AppErrorBoundary from "./components/AppErrorBoundary";
 
 jest.mock("react-dom", () => ({ render: jest.fn() }));
 
@@ -16,7 +17,9 @@ describe("index.tsx", () => {
     require("./index");
     expect(ReactDOM.render).toHaveBeenCalledWith(
       <React.StrictMode>
-        <App />
+        <AppErrorBoundary>
+          <App />
+        </AppErrorBoundary>
       </React.StrictMode>,
       div,
     );
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -16,9 +16,12 @@
     "eject": "react-scripts eject"
   },
   "jest": {
-    "collectCoverageFrom": ["!src/**/*.stories.*"]
+    "collectCoverageFrom": [
+      "!src/**/*.stories.*"
+    ]
   },
   "dependencies": {
+    "@sentry/browser": "^5.24.2",
     "bootstrap": "^4.5.2",
     "react": "^16.13.1",
     "react-bootstrap": "^1.3.0",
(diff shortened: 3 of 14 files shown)


app/experimenter/nimbus-ui/src/components/AppErrorDialog/index.tsx
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+
+const AppErrorDialog = ({ error }: { error: Error }) => {
+  return (
+    <div className="bg-grey-20 flex items-center flex-col justify-center h-screen">
```
LZoog, line 9: Huh, I was going to go check this out in Storybook build from CI but it doesn't seem to have triggered with this PR? I'm guessing these classes aren't actually working, but I'm sure we can easily grab some from Bootstrap to make this look decent.

I actually didn't think about styling this state, might've broken it up into another issue as a est of 1-2. We could still do that (I could file it in the second "after setup" epic) if you'd like since the meat of the issue is the Sentry setup.

Also, is `Dialog` still a good name here if it's not a modal? Maybe it is since we're communicating to the user and it's technically a blocking message since they can't do anything else. 😛

app/experimenter/nimbus-ui/src/index.test.tsx
```diff
@@ -16,7 +17,9 @@ describe("index.tsx", () => {
     require("./index");
     expect(ReactDOM.render).toHaveBeenCalledWith(
       <React.StrictMode>
-        <App />
+        <AppErrorBoundary>
```
LZoog, line 20: Bleh, I took `.toHaveBeenCalledWith` out of this test in [another PR](https://github.com/mozilla/experimenter/pull/3679/files#diff-824146ab1d2c016a050b808c1082ac96R21), seemed fragile to match _exactly_ what we're outputting here since the App has to be wrapped with Apollo Client - `MockedProvider`  which is the recommendation for mocking it doesn't work there and, passing in an identical Apollo Client to the one ran in the index in the test file still makes this test fail.

I'm still checking we're using React Strict mode in that PR and I think it's worth ensuring the index file renders the `AppErrorBoundary`. It looks like [Enzyme might have something](https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/contains.html) we can check the children with which if that works for React.Strict too would probably be better than what I'm doing in my PR or `.toHaveBeenCalledWith`, but we haven't needed Enzyme in `fxa-settings` or `fxa-payments`. 🤔

LZoog (commented):
Looking awesome so far! I also like `services`. I wondered where we should put test helpers and utils, I feel those could still go under the generic `lib`.

I think we'll also want to/be able to initialize with the Sentry DSN once #3670 is in.

jaredlockhart:
Commit formatting nit:

List of scopes: https://github.com/mozilla/experimenter/blob/main/contributing.md#scope

lmorchard:
Oh, I submitted this with my own fork, which will skip the storybook build. I'll close & reopen from an upstream repo branch to get Storybook building.

app/experimenter/nimbus-ui/src/components/AppErrorDialog/index.tsx
lmorchard: Storybook build skipped because I submitted this from my own fork - will close this PR and re-open from upstream repo. For styling, I think I started this before we [got bootstrap merged in](https://github.com/mozilla/experimenter/commit/47dfb6a0e770deffeed37b76391b37caa431c63e) so I kind of punted on it. Also could just rename "Dialog" to "Message" but not sure it matters overmuch

app/experimenter/nimbus-ui/src/index.test.tsx
lmorchard: Seems like maybe these should be separate tests? One for strict mode with the mocked render and another more like the old test? Switching to Enzyme might not be worth it, since I think we don't really want to mix react-testing-library and Enzyme

lmorchard:
Closed in favor of a new PR from upstream repo branch, which should get covered by a Storybook build

https://github.com/mozilla/experimenter/pull/3690

app/experimenter/nimbus-ui/src/index.test.tsx
LZoog: Ah ok, I didn't realize using Enzyme meant "switching to" because of the different test environments.

I'm not sure we can do the "old test" with `.toHaveBeenCalledWith` and Apollo Client or at least, I spent a decent amount of time trying to get that to pass and it refused. I even copied and pasted the "expected" and "received" output in a diff checker to see what character was different and it came back exactly identical, yet still didn't pass. I could try again with fresh eyes🤷‍♀️

PR #3677, opened 2020-10-06
refactor(backend): fixup query options

bbangert:
Because:

* We want to be able to paginate and re-order experiment listings.

This commit:

* Add's limit/offset/order optional parameters to a new single
  experiments gql query.

### DIFF
--- a/app/experimenter/experiments/api/v5/query.py
+++ b/app/experimenter/experiments/api/v5/query.py
@@ -1,31 +1,34 @@
 import graphene
 
 from experimenter.experiments.api.v5.types import NimbusExperimentType
+from experimenter.experiments.constants.nimbus import NimbusConstants
 from experimenter.experiments.models.nimbus import NimbusExperiment
 
 
 class Query(graphene.ObjectType):
-    experiments_by_status = graphene.Field(
-        graphene.List(NimbusExperimentType),
-        status=graphene.Argument(
-            NimbusExperimentType._meta.fields["status"]._type, required=False
-        ),
-    )
-    all_experiments = graphene.List(
+    experiments = graphene.List(
         NimbusExperimentType,
+        description="List Nimbus Experiments.",
+        offset=graphene.Int(),
+        limit=graphene.Int(),
+        status=graphene.Enum(
+            "NimbusExperimentOptionalStatus", NimbusConstants.Status.choices
+        )(),
     )
     experiment_by_slug = graphene.Field(
-        NimbusExperimentType, slug=graphene.String(required=True)
+        NimbusExperimentType,
+        description="Retrieve a Nimbus experiment by its slug.",
+        slug=graphene.String(required=True),
     )
 
-    def resolve_all_experiments(root, info):
-        return NimbusExperiment.objects.all()
+    def resolve_experiments(root, info, offset=0, limit=20, status=None):
+        q = NimbusExperiment.objects.all()
+        if status:
+            q = q.filter(status=status)
+        return q[offset:limit]
 
     def resolve_experiment_by_slug(root, info, slug):
         try:
             return NimbusExperiment.objects.get(slug=slug)
         except NimbusExperiment.DoesNotExist:
             return None
-
-    def resolve_experiments_by_status(root, info, status):
-        return NimbusExperiment.objects.filter(status=status).all()
--- a/app/experimenter/experiments/api/v5/mutation.py
+++ b/app/experimenter/experiments/api/v5/mutation.py
@@ -14,4 +14,6 @@ class CreateExperimentMutation(DjangoModelFormMutation):
 
 
 class Mutation(graphene.ObjectType):
-    create_experiment = CreateExperimentMutation.Field()
+    create_experiment = CreateExperimentMutation.Field(
+        description="Create a new Nimbus Experiment."
+    )
--- a/app/experimenter/experiments/tests/api/v5/test_query.py
+++ b/app/experimenter/experiments/tests/api/v5/test_query.py
@@ -12,14 +12,14 @@ from experimenter.experiments.tests.factories import NimbusExperimentFactory
 class TestExperimentListView(GraphQLTestCase):
     GRAPHQL_URL = reverse("nimbus-api-graphql")
 
-    def test_all_experiments(self):
+    def test_experiments(self):
         user_email = "user@example.com"
         exp = NimbusExperimentFactory.create_with_status(NimbusExperiment.Status.DRAFT)
 
         response = self.query(
             """
             query {
-                allExperiments {
+                experiments {
                     name
                     slug
                     publicDescription
@@ -30,10 +30,38 @@ class TestExperimentListView(GraphQLTestCase):
         )
         self.assertEqual(response.status_code, 200)
         content = json.loads(response.content)
-        experiments = content["data"]["allExperiments"]
+        experiments = content["data"]["experiments"]
         self.assertEqual(len(experiments), 1)
         for key in experiments[0]:
             self.assertEqual(experiments[0][key], str(getattr(exp, to_snake_case(key))))
+
+    def test_experiments_by_status(self):
+        user_email = "user@example.com"
+        draft_exp = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.DRAFT
+        )
+        NimbusExperimentFactory.create_with_status(NimbusExperiment.Status.ACCEPTED)
+
+        response = self.query(
+            """
+            query {
+                experiments(status: Draft) {
+                    name
+                    slug
+                    publicDescription
+                }
+            }
+            """,
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        experiments = content["data"]["experiments"]
+        self.assertEqual(len(experiments), 1)
+        for key in experiments[0]:
+            self.assertEqual(
+                experiments[0][key], str(getattr(draft_exp, to_snake_case(key)))
+            )
 
     def test_experiment_by_slug(self):
         user_email = "user@example.com"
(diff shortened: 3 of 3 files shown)


app/experimenter/experiments/api/v5/types.py
```diff
@@ -37,3 +38,26 @@ class NimbusBucketRangeType(DjangoObjectType):
...
+
+    """
+
+    DRAFT = "Draft"
+    REVIEW = "Review"
+    SHIP = "Ship"
+    ACCEPTED = "Accepted"
+    LIVE = "Live"
+    COMPLETE = "Complete"
+    REJECTED = "Rejected"
```
jaredlockhart, line 63: Can't we just refer to the constants here? https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L5-L12

app/experimenter/experiments/api/v5/query.py
```diff
@@ -1,31 +1,39 @@
...
-        ),
-    )
-    all_experiments = graphene.List(
+    experiments = graphene.List(
         NimbusExperimentType,
+        description="List Nimbus Experiments.",
+        offset=graphene.Int(),
+        limit=graphene.Int(),
+        order=NimbusExperimentIdOrder(),
+        status=NimbusExperimentOptionalStatus(),
```
jaredlockhart, line 18: So to filter on a field on the model you have to explicitly declare it here?

app/experimenter/experiments/api/v5/query.py
```diff
@@ -1,31 +1,39 @@
...
-    def resolve_all_experiments(root, info):
-        return NimbusExperiment.objects.all()
+    def resolve_experiments(
+        root, info, offset=0, limit=20, order="ascending", status=None
+    ):
+        q = NimbusExperiment.objects
+        q = q.order_by("-id") if order == "descending" else q.order_by("id")
+        if status:
+            q = q.filter(status=status)
+        return q.all()[offset:limit]
```
jaredlockhart, line 33: Where is sorting by ID called for?  In general we've found other things like ordering by dates, names, projects etc are what users do most frequently.

app/experimenter/experiments/api/v5/query.py
jaredlockhart: Alternatively would be more effective for GQL just send all experiments to the frontend and the filtering/sorting be done on the client side?

jaredlockhart (commented):
I see where you're going with this, we've used this thing for sorting and filtering in other places:

https://django-filter.readthedocs.io/en/stable/

I wonder if we could use it here?

app/experimenter/experiments/api/v5/types.py
bbangert: I was not having any luck attempting to rope those in, I tried about a dozen various invocations and failed before giving up and just putting this together so it would work.

app/experimenter/experiments/api/v5/query.py
bbangert: It's possible to reach into the `NimbusExperimentType` object, but because those types are generated off the model, the status becomes a required field. In this case we want the status to be an optional argument so using the generated required graphene type doesn't work. I spent quite a bit of time trying to grab the object and manipulate it back to optional for use here, but that didn't work out.

app/experimenter/experiments/api/v5/query.py
bbangert: Sorting isn't called for at all right now, so I just with with the id as its a stable ordering identifier.

jaredlockhart: If we want a stable ordering I'd just define that right on the model here instead of adding it here:

https://docs.djangoproject.com/en/3.1/ref/models/options/#ordering

bbangert:
We could use `django-filter` if we use GraphQL Relay, which is a slightly different standard for GraphQL. e.g. https://docs.graphene-python.org/projects/django/en/latest/filtering/ documents how `django-filter` is used, but that is only available if we use https://relay.dev/

We haven't used Relay before, it requires a different client so we'll want @jodyheavener and @LZoog to take a look at that first before deciding to go that route.

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Looking at:

https://docs.graphene-python.org/en/latest/types/enums/#definition

Seems like this works:

```py
NimbusExperimentOptionalStatus = Enum(
    "NimbusExperimentOptionalStatus", NimbusConstants.Status.choices
)
```

app/experimenter/experiments/api/v5/query.py
jaredlockhart: Ah ya I see 👍

app/experimenter/experiments/api/v5/types.py
bbangert: I know, it really seems like it should work, right? Unfortunately it doesn't.

jaredlockhart: I just tried it and the tests passed? Where doesn't it work?

jaredlockhart:
```
diff --git a/app/experimenter/experiments/api/v5/types.py b/app/experimenter/experiments/api/v5/types.py
index a0093659..f008b139 100644
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -8,6 +8,7 @@ from experimenter.experiments.models.nimbus import (
     NimbusIsolationGroup,
 )
 from experimenter.projects.models import Project
 
 
 class NimbusBranchType(DjangoObjectType):
@@ -45,19 +46,6 @@ class NimbusExperimentIdOrder(Enum):
     DESCENDING = "descending"
 
 
-class NimbusExperimentOptionalStatus(Enum):
-    """Nimbus Experiment Optional Status
-
-    The NimbusExperimentStatus type on the NimbusExperimentType is a required
-    field preventing it from being used as an optional argument. This
-    duplicate is used to allow for optional usage as a Query argument.
-
-    """
-
-    DRAFT = "Draft"
-    REVIEW = "Review"
-    SHIP = "Ship"
-    ACCEPTED = "Accepted"
-    LIVE = "Live"
-    COMPLETE = "Complete"
-    REJECTED = "Rejected"
+NimbusExperimentOptionalStatus = Enum(
+    "NimbusExperimentOptionalStatus", NimbusConstants.Status.choices
+)
diff --git a/app/experimenter/experiments/tests/api/v5/test_query.py b/app/experimenter/experiments/tests/api/v5/test_query.py
index b16ce90d..7a538391 100644
--- a/app/experimenter/experiments/tests/api/v5/test_query.py
+++ b/app/experimenter/experiments/tests/api/v5/test_query.py
@@ -45,7 +45,7 @@ class TestExperimentListView(GraphQLTestCase):
         response = self.query(
             """
             query {
-                experiments(status: DRAFT) {
+                experiments(status: Draft) {
                     name
                     slug
                     publicDescription
```

bbangert: `
    experiments = graphene.List(
        NimbusExperimentType,
        description="List Nimbus Experiments.",
        offset=graphene.Int(),
        limit=graphene.Int(),
        order=NimbusExperimentIdOrder(),
        status=graphene.Enum(
            "NimbusExperimentOptionalStatus", NimbusConstants.Status.choices
        ),
    )
`

It's not working here.

jaredlockhart:
Ah that's too bad about `django-filter`

app/experimenter/experiments/api/v5/types.py
bbangert: Ugh, the traceback is annoyingly devoid of helpful advice. I was missing an additional instantiation.

LZoog:
@bbangert Hrm, good timing since I just opened a PR using Apollo Client (#3679), but we already have nice test helpers and examples from Settings with Apollo Client and I've made tickets based on working with the Apollo cache so it'd be a bummer to switch. I haven't checked it out at all though, I'll try to make time to poke through their docs later today and circle back.

app/experimenter/experiments/api/v5/query.py
bbangert: How would I reverse that if the order is changed in a query?

jaredlockhart: Oh yeah that was my point, if we don't need to explicitly reorder ascending/descending by id, should we even bother doing it at all?  Just trying to avoid building things that aren't called for in specs/ux, but not a blocker.

app/experimenter/experiments/api/v5/query.py
```diff
@@ -1,31 +1,41 @@
...
+        slug=graphene.String(required=True),
     )
 
-    def resolve_all_experiments(root, info):
-        return NimbusExperiment.objects.all()
+    def resolve_experiments(
+        root, info, offset=0, limit=20, order="ascending", status=None
+    ):
+        q = NimbusExperiment.objects
+        q = q.order_by("-id") if order == "descending" else q.order_by("id")
```
jaredlockhart, line 32: We can just refer to these string constants by reference from here, right?

https://github.com/mozilla/experimenter/pull/3677/files#diff-21992c79713e9ebfac085a293b207bbdR44-R45

In general I like to avoid hard coding duplicate string constants.

jaredlockhart: Caveat: Except in tests where hard coding expectations is generally fine, I think.

jaredlockhart (changes requested):
Looks good @bbangert !

app/experimenter/experiments/api/v5/query.py
bbangert: Ok, I'm just gonna remove it since we'll prolly have to sort differently later anyways.

app/experimenter/experiments/api/v5/query.py
```diff
@@ -1,31 +1,34 @@
...
+        slug=graphene.String(required=True),
     )
 
-    def resolve_all_experiments(root, info):
-        return NimbusExperiment.objects.all()
+    def resolve_experiments(root, info, offset=0, limit=20, status=None):
+        q = NimbusExperiment.objects
+        if status:
+            q = q.filter(status=status)
+        return q.all()[offset:limit]
```
jaredlockhart, line 28: Oh oh, `objects` is a `ModelManager` instance, not a `QuerySet` instance, they have some of the same methods but not all, so the way this is written technically works but could break if we try to do some other operations that querysets support that managers don't.  This should probably be written:

```py
        q = NimbusExperiment.objects.all()
        if status:
            q = q.filter(status=status)
        return q[offset:limit]
```

jaredlockhart (changes requested):

app/experimenter/experiments/api/v5/query.py
bbangert: Ahhhh

jaredlockhart (approved):
Yep looks great this is much nicer thank you @bbangert ! 🎉

PR #3679, opened 2020-10-06
fix #3442 feat(nimbus-ui): Add Apollo client/cache

### LINKED ISSUE #3442: Set up apollo client/cache, add test helper
Apollo client cache can store app-global date for us, including any local global data. We need to set up apollo client with apollo client cache and wrap our App instance with it.

Cache setup example: [https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/cache.ts#L79-L106](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/cache.ts#L79-L106)

Apollo client setup example (config URL should just be `/graphql`, and we shouldn't need `authLink`): [https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/gql.ts](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/gql.ts)

Wrapping the App instance: [https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/index.tsx#L31](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/index.tsx#L31)

We also have several useful test helpers in fxa-settings that we can copy over. For this issue, check out `MockedProvider`: [https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/models/_mocks.tsx#L114](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/models/_mocks.tsx#L114)

We don't need the `account` or `session` benefits here just yet (we may want something like this later though [check out the docs](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings#components-that-use-useaccount-usesession-or-need-a-gql-mock)), we just want to be able to use `<MockedProvider>` in tests as a convenient way to provide a default cache for tests.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-285)
┆Issue Number: EXP-285

LZoog:
My original ask in #3442 was to also add a test helper (`MockedCache` from `fxa-settings`) - I started doing this, and then realized it's going to make a whole lot more sense to add that when we write and then read something from the cache. That test helper 1) makes it easy to override data inside the cache for tests, in which case we need to know what high-level pieces of data we're going to want to read/write, and 2) writes an initial state to the cache, which we won't have at least as part of epic 2/3 since the "directory view" is a bonus.

Instead, I worked a tad on the index file tests since `expect(ReactDOM.render).toHaveBeenCalledWith(...` is fragile. Now it just checks mounting on the root el and ensures we're using Strict mode at the top level. This reaches 100% test coverage - we'll be testing for `ApolloClient` with `MockedProvider` once we start talking to the server. We test that `App` can be rendered in the App test.

---

Because:
* We want to communicate with the server through Apollo Client and will need a client cache.

This commit:
* Adds Apollo client/cache
* Adjusts the index bootstrap test

fixes #3442, minus test helper (I'll add a note in the first query ticket to add this)

### DIFF
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -21,8 +21,10 @@
     ]
   },
   "dependencies": {
+    "@apollo/client": "^3.2.2",
     "@sentry/browser": "^5.24.2",
     "bootstrap": "^4.5.2",
+    "graphql": "^15.3.0",
     "react": "^16.13.1",
     "react-bootstrap": "^1.3.0",
     "react-dom": "^16.13.1",
@@ -40,7 +42,7 @@
     "@testing-library/react": "^11.0.4",
     "@testing-library/user-event": "^7.1.2",
     "@types/jest": "^24.0.0",
-    "@types/node": "^12.0.0",
+    "@types/node": "^14.11.5",
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
     "@typescript-eslint/eslint-plugin": "2.x",
--- a/app/experimenter/nimbus-ui/src/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/index.test.tsx
@@ -21,6 +21,13 @@ jest.mock("./components/AppErrorBoundary", () => ({
   ),
 }));
 
+jest.mock("@apollo/client", () => ({
+  ...jest.requireActual("@apollo/client"),
+  ApolloProvider: ({ children }: { children: ReactNode }) => (
+    <div data-testid="ApolloProvider">{children}</div>
+  ),
+}));
+
 jest.mock("./components/App", () => ({
   __esModule: true,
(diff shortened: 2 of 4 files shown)


jaredlockhart:
Commit formatting nit:

List of scopes: https://github.com/mozilla/experimenter/blob/main/contributing.md#scope

LZoog:
I'd like to [figure this out](https://github.com/mozilla/experimenter/pull/3676#discussion_r500656552) before rebasing/merging. Would be ideal if Les's PR got in sometime tomorrow so I could take a look towards the end of the day.

jodyheavener (approved):

jodyheavener:
oh no sorry @jaredlockhart I missed your commit format nit before hitting the button 🙈

jaredlockhart:
tenor

PR #3682, opened 2020-10-06
fix #3402 feat(nimbus): add application and channels to NimbusExperiment

### LINKED ISSUE #3402: Add application and channel to experiment model
Applications should include reference-browser, `fenix,` and `firefox-desktop`

Please change firefox_channel to be application_channels

Channels should be specific to each application – note that for mobile applications, each channel has an app id which needs to be used in the targeting config.

Also, channels should be many-to-many (so store as an array)

Please also update the serializers.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-260)
┆Issue Number: EXP-260

jaredlockhart:
Because

* We need to target different applications and channels for each experiment

This commit

* Adds application and channels to NimbusExperiment with a fixed list of valid options for each


jaredlockhart:
Blocked on getting list of valid channels for mobile here: https://github.com/mozilla/experimenter/issues/3402#issuecomment-704482307

bbangert (approved):

PR #3688, opened 2020-10-06
fix #3686 fix(project): Use built app path in deploy image

### LINKED ISSUE #3686: Use built experimenter path in deploy image
Oops.  I got overzealous on dev environment optimization and accidentally occluded the built static assets from the deploy container.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-455)
┆Issue Number: EXP-455

jaredlockhart:
Because

* The built app path contains the built static assets

This commit

* Uses the built app path from the build image in the deploy image

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -83,4 +83,4 @@ COPY --from=build /app/bin/ /app/bin/
 COPY --from=build /app/manage.py /app/manage.py
 COPY --from=build /usr/local/bin/ /usr/local/bin/
 COPY --from=build /usr/local/lib/python3.8/site-packages/ /usr/local/lib/python3.8/site-packages/
-COPY ./experimenter/ /app/experimenter/
+COPY --from=build /app/experimenter/ /app/experimenter/


tiftran (approved):

jaredlockhart:
Expanded explanation:

We split the Dockerfile into two stages: `build` and `deploy`.  The `build` image installs all the necessary dev packages for Python and Node to build the project and finally copies in the app source:

https://github.com/mozilla/experimenter/blob/64d889cacf77b474fc2550caa0e66669ac682818/app/Dockerfile#L2-L65

The import part is here:

https://github.com/mozilla/experimenter/blob/64d889cacf77b474fc2550caa0e66669ac682818/app/Dockerfile#L56-L61

Which builds all the static frontend assets and dumpts them into their respect distributable folders.

Then later when we build the deploy image, we copy only select parts from the `build` image:

https://github.com/mozilla/experimenter/blob/64d889cacf77b474fc2550caa0e66669ac682818/app/Dockerfile#L82-L86

There's a subtle thing here, I originally had it as

` COPY --from=build /app/experimenter/ /app/experimenter/`

which copies the entire `experimenter` subtree over to the deploy image from the build image, inside there are all the built assets that were built in the build stage.

I changed it to

` COPY ./experimenter/ /app/experimenter/`

which copies from the host filesystem instead of the `build` image becuase it copies a lot faster, but I forgot that that would omit the built assets, so this PR just changes that back.

I think the important takeaway here is we need to test against the `deploy` image instead of the `build` image to make sure we don't break it, so I filed https://jira.mozilla.com/browse/EXP-456 as a followup.

PR #3693, opened 2020-10-07
fix #3692 chore(project): reorganize static asset paths

### LINKED ISSUE #3692: Reorganize static asset paths
To help distingiush between legacy and nimbus static assets we can reorganize our static paths and folders to help isolate and organize them.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-459)
┆Issue Number: EXP-459

jaredlockhart:
Because

* We now have 4 different frontend builds we should better organize where they are hosted and stored to disambiguate import paths

This commit

* Uses /static/ as the root path for all static assets for legacy and nimbus
* Explicitly copies only the built assets of each into /served/ rather than all source
* Updates import paths of affected assets in legacy
* Fixes a bug in the volumes_rm command
* Removes legacy unused served and asset volumes from docker compose
* Removes legacy unused images
* Removes legacy unused project templates


app/experimenter/legacy-ui/rapid/package.json
```diff
@@ -58,4 +58,4 @@
     "test": "jest --env=jest-environment-jsdom-sixteen",
     "watch": "webpack --watch --mode development --progress"
   }
-}
+}
```
jodyheavener, line 61: Whatchu got against EOF newlines punk

app/experimenter/settings.py
```diff
@@ -168,11 +168,16 @@
...
 
-STATIC_ROOT = os.path.join(os.path.join(BASE_DIR, "served"), "legacy-ui")
+STATIC_ROOT = os.path.join(os.path.join(BASE_DIR, "served"), "static")
 
-STATICFILES_DIRS = [os.path.join(BASE_DIR, "legacy-ui")]
+STATICFILES_DIRS = [
+    ("assets", os.path.join(BASE_DIR, "legacy-ui", "assets")),
+    ("scripts", os.path.join(BASE_DIR, "legacy-ui", "scripts")),
+    ("imgs", os.path.join(BASE_DIR, "legacy-ui", "imgs")),
+    ("nimbus", os.path.join(BASE_DIR, "nimbus-ui", "build")),
```
jodyheavener, line 179: Huzzah. Love it.

jaredlockhart:
@jodyheavener so then you'll be able to do 

`{% static "nimbus/whatever" %}` 

and it'll get hosted at 

`/static/nimbus/whatever`

jaredlockhart:
I tried to do the same for legacy but then we would'v ehad to update every path in every html/js/css and i didn't want to disturb all that stuff, so all that stuff is under `assets` and nimbus stuff is under `nimbus`

jodyheavener (approved):
Makes sense to me. Once the Circle Gods let you through you have my blessing.

One thought: this is something that I think might make sense to throw into documentation somewhere ("Here's a breakdown of how and where our various apps access their static assets"). However, the root README isn't super set up to toss in little tidbits like this (lest ye want to expand under the Frontend section), so it's not critical in this PR. Just something to noodle on as we expand and more folks dig through the code.

app/experimenter/legacy-ui/rapid/package.json
jaredlockhart: My editor is vicious 😭

PR #3694, opened 2020-10-07
fix #3480 chore(project): use docker caching for circle deploy

### LINKED ISSUE #3480: Circle Deploy task not using docker cache
I just noticed that the deploy task is building the image from scratch, it should be using the cache.  Probably just need to add a line to the circle config.

https://app.circleci.com/pipelines/github/mozilla/experimenter/6085/workflows/047d0688-a930-4ced-9d42-3489def42a4a/jobs/15239

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-308)
┆Issue Number: EXP-308

jaredlockhart:
Because

* The circle deploy task runs after every PR lands and it's not being cached it takes ~10m to rebuild everything from scratch every time

This commit

* Enables docker layer caching for the deploy task which means in the common case it should take <1m
* Bump to the latest circle docker image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     machine:
       docker_layer_caching: true
-      image: ubuntu-1604:201903-01 # recommended linux image - includes Ubuntu 16.04, docker 18.09.3, docker-compose 1.23.1
+      image: ubuntu-1604:202007-01 # Ubuntu 16.04, Docker v19.03.12, Docker Compose v1.26.1
     resource_class: large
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -55,9 +55,10 @@ jobs:
           path: app/tests/integration/integration-test.html
 
   deploy:
+    machine:
+      docker_layer_caching: true
```
bbangert, line 59: Doh!

bbangert (approved):

PR #3709, opened 2020-10-08
fix #3707 chore(project): isolate assets to containers

### LINKED ISSUE #3707: Frontend builds invalidating docker cache
Something is changing in the frontend builds that is invalidating the docker cache which causes it to rebuild frontend assets everytime the containers start, which is affecting local dev and CI.  The simplest approach I've learned to manage this is disable all gitignores, re run all the docker commands, and git will expose what paths have changed.  It's probably a cache path getting dumped back to the host filesystem we just have to add a callout for in the compose files.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-472)
┆Issue Number: EXP-472

jaredlockhart:
Because

* Recently changed the way assets are grouped I accidentally broke the way we isolate built assets to the containers which was causing the docker cache to be invalidated in circle and increasing build/test times

This commit

* Catches all built asset paths for legacy and nimbus uis and isolates them to shared volumes in the containerized dev environment


.env.sample
```diff
@@ -2,6 +2,7 @@ BUGZILLA_API_KEY=
 BUGZILLA_CC_LIST=
 BUGZILLA_HOST=https://bugzilla.allizom.org/
 CELERY_SCHEDULE_INTERVAL=300
+COMPOSE_HTTP_TIMEOUT=120
```
jaredlockhart, line 5: @jodyheavener Yeah you were right we need this, integration on circle has been complaining about it XD

jodyheavener: Aha!

jodyheavener (approved):
I know less about this, but a passing CI gives me confidence 💪

PR #3710, opened 2020-10-08
fix #3408 feat(nimbus): add FeatureConfig model

### LINKED ISSUE #3408: Add FeatureConfig Model
Let's create a FeatureConfig model.

Feature Configs need to have:**

* **slug**: unique identifier for the feature**
* **applicationId**: identifier for application

* **name**: human-readable name for the feature

* **description**: human-readable description for the feature

* **owner**: text field - email address of the owner of the feature**

* **schema** (optional):  If the feature configuration should include custom parameters, this should be a JSON Schema ([draft-07](http://json-schema.org/draft/2019-09/json-schema-validation.html)) describing those parameters. If the feature is a simple flag, this should be null.

Experiments should optionally define a relationship with a single feature id.

The value in Branch should be associated with a JSON schema for that feature.

See Technical documentation: [https://docs.google.com/document/d/1AqRetDJRMIEWGTBdNkwH9zQ2WP-v2mIklY4shJN7QWk/edit](https://docs.google.com/document/d/1AqRetDJRMIEWGTBdNkwH9zQ2WP-v2mIklY4shJN7QWk/edit#heading=h.5ekpmdxm3mm2)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-266)
┆Issue Number: EXP-266

jaredlockhart:
Because

* FeatureConfigs need to be stored in the database and editable in the admin

This commit

* Adds a FeatureConfig model and factory
* Adds FeatureConfig to the admin
* Updates the visualization logic


jaredlockhart:
Ack okay note: the visualization code was using what was called 'Features', but what is now called 'ProbeSets'.  'FeatureConfigs' were added but are not what the visualization code wants.  I'll switch it over when I do the next ticket: https://jira.mozilla.com/browse/EXP-353

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -2,14 +2,15 @@
...
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from django.urls import reverse
 from parameterized import parameterized
 
 from experimenter.experiments.constants import ExperimentConstants
 from experimenter.experiments.tests.factories import NimbusExperimentFactory
 
 
+@override_settings(FEATURE_ANALYSIS=False)
```
jaredlockhart, line 13: The visualization API code is referring to a setting being read from `.env`, so without overriding these tests break depending on what is set in your local dev, so this just isolates them.

jaredlockhart: Also @emtwo see above I added some `__init__.py` to the visualization subtree that were missing.  These are so annoying, Python code will import them correctly but some things, like test discovery, will ignore them if they're missing.

app/experimenter/experiments/api/v5/types.py
```diff
@@ -37,3 +38,9 @@ class NimbusBucketRangeType(DjangoObjectType):
     class Meta:
         model = NimbusBucketRange
         exclude = ("id", "experiment")
+
+
+class NimbusFeatureConfigType(DjangoObjectType):
+    class Meta:
+        model = NimbusFeatureConfig
+        exclude = ("id",)
```
bbangert, line 46: In my current PR, I'm showing ID's for anything that can be updated/deleted. I assume this will need to allow that?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -74,6 +73,9 @@ class NimbusExperiment(NimbusConstants, models.Model):
         ),
         default=list,
     )
+    feature_config = models.ForeignKey(
+        "NimbusFeatureConfig", blank=True, null=True, on_delete=models.CASCADE
```
bbangert, line 77: I thought a single nimbus experiment could have multiple feature configs, maybe I misunderstood?

bbangert (approved):

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Oh yeah that's what I was wondering, k I'll take that out 👍

PR #3713, opened 2020-10-09
fix #3712 fix(legacy): stop changelogs with no version change

### LINKED ISSUE #3712: [Legacy]Normandy Version Update task creates changelog regardless of whether there's a change...
Screen Shot 2020-10-09 at 12 26 33 AM

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-474)
┆Issue Number: EXP-474

tiftran:
Because

* Changelogs are being created for no reason

This commit

* Converts normandy versions from recipe to a string before comparing with experimenter versions

### DIFF
--- a/app/experimenter/normandy/tasks.py
+++ b/app/experimenter/normandy/tasks.py
@@ -172,8 +172,9 @@ def update_firefox_versions(experiment, recipe_data, filter_objects):
     changed_data = {}
 
     if versions := filter_objects.get("version"):
-        min_version = str(min(versions["versions"]) * 1.0)
-        max_version = str(max(versions["versions"]) * 1.0)
+
+        min_version = str(float(min(versions["versions"])))
+        max_version = str(float(max(versions["versions"])))
         if experiment.firefox_min_version != min_version:
             changed_data["firefox_min_version"] = min_version
         if experiment.firefox_max_version != max_version:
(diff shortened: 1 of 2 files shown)


app/experimenter/normandy/tasks.py
```diff
@@ -172,8 +172,8 @@ def update_firefox_versions(experiment, recipe_data, filter_objects):
     changed_data = {}
 
     if versions := filter_objects.get("version"):
-        min_version = min(versions["versions"]) * 1.0
-        max_version = max(versions["versions"]) * 1.0
+        min_version = str(min(versions["versions"]) * 1.0)
+        max_version = str(max(versions["versions"]) * 1.0)
```
jaredlockhart, line 176: Oooohh multiplying by `1.0` to convert it to a float.

More idiomatic way is to just use `float` like

`str(float(min(versions["versions"])))`

Alternatively you could just use a format string like

`f"{min(versions['versions'])}.0"`

but then you'd either have to be sure it's an `int` or wrap it in an `int()` call maybe like

`f"{int(min(versions['versions']))}.0"`

Iunno there's a bunch of ways to do it, but I'd just use casts instead of multiplication to force it to the type you want 👍

jaredlockhart (approved):
Good catch @tiftran! 🦆

PR #3718, opened 2020-10-09
fix #3631 feat(backend): add update experiment mutation

### LINKED ISSUE #3631: Mutation to update experiments basic fields
Per slide 4 of [https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216037%3A1694](https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216037%3A1694)

And the Overview side.

There should be a mutation to update any of:

* Public name
* Slug
* Hypothesis
* Application
* Public Description

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-432)
┆Issue Number: EXP-432

bbangert:
Because:

* We want to allow updating the mutation.
* Additional magic from graphene-django is starting to get in the way.

This commit:

* Add's a NimbusExperiment update mutation.
* Refactors the create experiment mutation to use the more verbose but
  configurable method used for update.

Closes #3631

### DIFF
--- a/app/experimenter/experiments/api/v5/inputs.py
+++ b/app/experimenter/experiments/api/v5/inputs.py
@@ -1,0 +1,19 @@
+import graphene
+
+
+class ExperimentInput(graphene.InputObjectType):
+    clientMutationId = graphene.String()
+    name = graphene.String()
+    slug = graphene.String()
+    application = graphene.String()
+    public_description = graphene.String()
+    hypothesis = graphene.String()
+
+
+class CreateExperimentInput(ExperimentInput):
+    name = graphene.String(required=True)
+    slug = graphene.String(required=True)
+
+
+class UpdateExperimentInput(ExperimentInput):
+    id = graphene.ID(required=True)
--- a/app/experimenter/experiments/api/v5/mutation.py
+++ b/app/experimenter/experiments/api/v5/mutation.py
@@ -1,19 +1,77 @@
 import graphene
-from django.db.models import Field
-from graphene_django.forms.mutation import DjangoModelFormMutation
 
-from experimenter.experiments.api.v5.forms import CreateExperimentForm
+from experimenter.experiments.api.v5.inputs import (
+    CreateExperimentInput,
+    UpdateExperimentInput,
+)
+from experimenter.experiments.api.v5.serializers import NimbusExperimentSerializer
 from experimenter.experiments.api.v5.types import NimbusExperimentType
+from experimenter.experiments.models import NimbusExperiment
 
 
-class CreateExperimentMutation(DjangoModelFormMutation):
-    experiment = Field(NimbusExperimentType)
+class ObjectField(graphene.Scalar):
+    """Utilized to serialize out Serializer errors"""
 
-    class Meta:
-        form_class = CreateExperimentForm
+    @staticmethod
+    def serialize(dt):
+        return dt
+
+
+class CreateExperiment(graphene.Mutation):
+    clientMutationId = graphene.String()
+    nimbus_experiment = graphene.Field(NimbusExperimentType)
+    message = ObjectField()
+    status = graphene.Int()
+
+    class Arguments:
+        input = CreateExperimentInput(required=True)
+
+    @classmethod
+    def mutate(cls, root, info, input: CreateExperimentInput):
+        serializer = NimbusExperimentSerializer(data=input)
+        if serializer.is_valid():
+            obj = serializer.save()
+            msg = "success"
+        else:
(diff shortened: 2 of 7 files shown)


app/experimenter/experiments/api/v5/inputs.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class UpdateExperimentInput(graphene.InputObjectType):
+    id = graphene.ID(required=True)
+    clientMutationId = graphene.String()
+    name = graphene.String(required=True)
+    slug = graphene.String(required=True)
+    application = graphene.String()
+    public_description = graphene.String()
+    hypothesis = graphene.String()
```
jaredlockhart, line 20: Can you subclass the create one and just add id?

app/experimenter/experiments/api/v5/mutation.py
```diff
@@ -1,19 +1,83 @@
...
+    message = ObjectField()
+    status = graphene.Int()
+
+    class Arguments:
+        input = CreateExperimentInput(required=True)
+
+    @classmethod
+    def mutate(cls, root, info, input: CreateExperimentInput):
+        clientMutationId = input.clientMutationId
+        del input.clientMutationId
```
jaredlockhart, line 32: What does mutation id do?

jaredlockhart: Also what happens if you don't delete it out of the input and just pass it to the serializer?

app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -0,0 +1,28 @@
...
+from experimenter.experiments.api.v5.serializers import NimbusExperimentSerializer
+from experimenter.experiments.models import NimbusExperiment
+from experimenter.experiments.tests.factories import NimbusExperimentFactory
+
+
+class TestNimbusExperimentSerializer(TestCase):
+    maxDiff = None
+
+    def test_serializer_outputs_expected_schema_for_accepted(self):
+        experiment = NimbusExperimentFactory.create_with_status(
```
jaredlockhart, line 12: You can chop `accepted` off the title, and use `create` instead of `create_with_status`, since there isn't any per-status specific logic in this serializer (yet!).

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -55,7 +55,7 @@ class Meta:
         model = NimbusExperiment
 
     @classmethod
-    def create_with_status(cls, target_status, **kwargs):
+    def create_with_status(cls, target_status, **kwargs) -> NimbusExperiment:
```
jaredlockhart, line 58: Adding type annotations?

jaredlockhart (changes requested):
Awesome, very straight forward.  Just a couple small things, but looking very nice so far 👍

app/experimenter/experiments/api/v5/inputs.py
bbangert: Probably, though I just noticed this is actually wrong, the update does *not* require the same fields as the creation.

app/experimenter/experiments/tests/factories/nimbus.py
bbangert: Habit, I can leave it off, but its nice having VS Code use proper command completion which sorta works when these are added.

app/experimenter/experiments/api/v5/inputs.py
jaredlockhart: Well in either case if one is a superset of the other you can probably just subclass?

app/experimenter/experiments/tests/factories/nimbus.py
jaredlockhart: Yeah pylance will pick them up. I've been trying to figure out how to add type annotations in a nice way and I've been wondering whether we can get away with adding type annotations for just the nimbus stuff, but there's enough overlap with legacy stuff I don't think we can do it for just Nimbus and actually enforce it.  I'd prefer if we do it to go all in on it instead of making it haphazard, but that's just my preference.

app/experimenter/experiments/api/v5/mutation.py
bbangert: Ah, it doesn't apparently care, so I'll stop doing that.

jaredlockhart (approved):
Gorgeous 👏 👏 👏 
source

PR #3734, opened 2020-10-12
fix #3463 feat(nimbus): add reach-router and initial stub pages

### LINKED ISSUE #3463: Add reach-router and new experiment page
1) Add reach-router
2) Add a "Create new" link on the `/nimbus` page that will link to a new route, `nimbus/new`
3) Add a dummy `PageExperimentNew/index.tsx` component to display on `nimbus/new`
4) Add a “Pages” section in Storybook (see [https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/PageChangePassword/index.stories.tsx](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/PageChangePassword/index.stories.tsx) for an example) to show the "Home" page and "New Experiment" page

Other pages will be done in follow ups.

---

Additional context: 

Currently, the old (core) application shows all of the experiments on the “home” page and new experiments for the core application at `experiments/new/`.

The `rapid` application is found at `experiments/rapid/` (no home page populating experiments yet) which has a "create" link taking the user to a new experiment page `experiments/rapid/new` showing an active "Settings" tab, along with a "branches" tab that navigates to `experiments/rapid/new/branches/`. When the user clicks "Save," they're brought to the "page summary" page at `experiments/rapid/:slug:` (which is the `ExperimentDetailsPage` component).

In the new `nimbus-ui` application, users will still click a "create" link that takes them to a new experiment page, but instead be filling out the "Summary" and "Branches" form, they'll fill out a much smaller "Overview" form at `nimbus/new` that, on submission (equivalent to `rapid` "Save"), submits the experiment in draft status, and then takes the user to an "edit" view for that experiment at `nimbus/:slug:/edit/` or `nimbus/:slug:/edit/overview` (shows the same page). In this view, they'll access the "branches" page at `nimbus/:slug:/edit/branches`, and the summary page will be linked in a "Review & Launch" link in the edit view and will still be at `nimbus/:slug:`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-293)
┆Issue Number: EXP-293

lmorchard:


app/experimenter/nimbus-ui/src/pages/ExperimentNew/index.test.tsx
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
LZoog, line 1: When working on `fxa-settings`, I had proposed to the team we place our "pages" in `components` since `pages` are still React components, and ended up creating `components/_pages/` so it'd be at the top of the tree, which I didn't end up liking. 

[See this thread ](https://mozilla.slack.com/archives/G0177C5FTCZ/p1598983640019000?thread_ts=1598979686.013700&cid=G0177C5FTCZ) where we talk about a `pages/` dir too - we ended up agreeing on a flat directory structure inside components, in which case this would be named `PageExperimentNew`.

We also talked about a flat directory structure [in the FE ADR](https://github.com/mozilla/experimenter/pull/3392#discussion_r485067560) where I reference `fxa-settings`.

_However_, I have fiddled a bit with NextJS on the side since these discussions and the convention there is to use `pages/` like you are here. Do you have a strong preference for `pages/ExperimentNew/` rather than `PageExperimentNew`? If so I'm fine with it and betting the rest of the team is too but we might want to double check with Jody and Jared just in case they strongly prefer the flat structure. 🙂

lmorchard: > we place our "pages" in components since pages are still React components

Personally, I'd rather have `pages/` separate from `components/` - while pages *are* technically components, they're different enough in purpose & usage & structure to be distinct, IMO.

Maybe arbitrary, but I think "components" should be mostly for presentation whereas pages / routes should handle composition & orchestration of other resources that feed into presentation components. That's how we organized things on payments-server and kind of to match content-server.

jaredlockhart: I prefer flat since I can never guess which folder a component I need is in, but that's just my vote 🖐️

jodyheavener: While I'm personally more partial to a flat structure, I'm happy to fall back here if the consensus is having distinct `pages` components 😄 I guess I'd want to know what belongs in a page component and how strict we should be -- should _anything_ that's not specific to layout be allowed to live in a page component? My guess is no.

jodyheavener: Maybe it's overkill, but this kinda feels worth a super-tiny ADR?

LZoog: > Maybe it's overkill, but this kinda feels worth a super-tiny ADR?

Or just a quick edit of this line if we go the `pages/` route:
https://github.com/mozilla/experimenter/pull/3392/files#diff-3a526648f7d71c5ec63e63771f901fd86a3b6735a5cb16928900b19469ab902cR50

I made an after-merged edit here: https://github.com/mozilla/experimenter/pull/3486/files

lmorchard: I mostly missed the original FE ADR discussion, so I didn't catch the mentions of a single component directory. I can just move things in this PR. Not my preference - but ADRs exist for a reason that supersedes preference, so I don't want to randomly bikeshed it in a PR.

> Or just a quick edit of this line if we go the pages/ route:

I wouldn't make an edit to an already-merged ADR unless it was to clarify the original decision. This sounds like it would be a change in the decision, which should be a new ADR describing if/why we changed our minds.

lmorchard: > I guess I'd want to know what belongs in a page component and how strict we should be -- should anything that's not specific to layout be allowed to live in a page component?

Just to clarify this bit - it's not so much about layout. 

[A page / route / container component](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) is where complex things happen. Maybe where Redux comes in or API calls, along with process orchestration. Business & interaction logic. Tests & stories are more complex and require more mocking because these are where the meat of the app lives.

Other components should largely just be "functions" - i.e. data goes in, DOM comes out, all logic is presentational. Tests & stories are simple because these are mainly just formatting components.

Having said all the above, though, even the author of that post I linked says this isn't a pattern to be dogmatically followed. So, 🤷‍♂️

lmorchard:
Rebased against main. Already got some discussion going, so I'll just open for review proper.

app/experimenter/nimbus-ui/src/index.test.tsx
```diff
@@ -102,3 +73,30 @@ describe("index", () => {
...
+  ...jest.requireActual("@apollo/client"),
+  ApolloProvider: mockComponent("ApolloProvider"),
+}));
+
+jest.mock("./components/App", () => ({
+  __esModule: true,
+  default: mockComponent("App"),
+}));
+
+function mockComponent(testid: string) {
```
lmorchard, line 98: Oh, and this is an annoying thing I discovered: I really wanted to DRY up the little functional components in those `jest.mock()` calls, maybe move that `mockComponent` function into a shared module.

As it turns out, though, Jest uses a babel plugin to reorder the code at compile time. All the `jest.mock` calls are moved to the top of the syntax tree before all import statements in order to ensure the mocks happen before loading the mocked modules. That means these mocks can't use imported code. So, things are only partly DRY-er

jaredlockhart (approved):
👍 lgtm 🐨

app/experimenter/nimbus-ui/src/index.tsx
```diff
@@ -28,7 +28,7 @@ try {
     <React.StrictMode>
       <AppErrorBoundary>
         <ApolloProvider {...{ client }}>
-          <App />
+          <App basepath="/experiments/nimbus" />
```
LZoog, line 31: Ahh I was going to say this should be `/nimbus` (URL confirmed with Kate a few weeks ago) but it looks like that's how we're setup to serve this now. Since the links are relative `./new` if we do change from `experiments/nimbus` to `/nimbus` in a follow up hopefully it won't touch many files.

LZoog (approved):
I checked into your branch and poked around, this LGTM!!

Issue #2538, opened 2020-04-09
Update versions from Normandy API


jaredlockhart:
It's possible that the versions of a delivery in Normandy can be updated after the delivery is published, in which case Experimenter does not reflect the currently targeted versions.  We should update the versions in Experimenter when we read the Normandy API to update other things (like the status, etc).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-160)
┆Issue Number: EXP-160

PR #3739, opened 2020-10-13
fix #3716 fix(legacy): Prune all "add version" changelogs

### LINKED ISSUE #3716: 502 or very slow load for Live experiment editing in admin tool
Seems that most of the live experiments either takes forever to load or 502.  Tif thinks the admin tool is possibly having trouble displaying the flood of changelogs based on this change https://github.com/mozilla/experimenter/issues/3712 

She'll verify this is working again with those changes.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-476)
┆Issue Number: EXP-476

tiftran:
Because:

* change from #2538 flooded prod instance with empty changelogs

This commit:

* has a migration that should filter  "Added Version(s)" changelogs and delete them


jaredlockhart (approved):
Nuke am all 👍

PR #3740, opened 2020-10-14
chore(deps): bump django from 3.0.7 to 3.1.2 in /app

lmorchard:
Bumps [django](https://github.com/django/django) from 3.0.7 to 3.1.2.
- [Release notes](https://github.com/django/django/releases)
- [Commits](https://github.com/django/django/compare/3.0.7...3.1.2)

Signed-off-by: dependabot[bot] <support@github.com>


lmorchard:
This was PR #3728, [which I closed to just punt on the upgrade for now](https://github.com/mozilla/experimenter/pull/3728#issuecomment-708034723).

But, that bugged me, so I poked around some more and am curious now if maybe I did find a fix? Opening as a draft to see if the tests pass in CI.

lmorchard:
> Hrmm, why would moving it up to the manager fix it....

As far as I can tell, it's because like [the validation for sort order](https://github.com/django/django/commit/513948735b799239f3ef8c89397592445e1a0cd5#diff-fd2300283d1546e36141373b0621f142ed871e3e8856e07efe5a22ecc38ad620R1916) happens [in the superclass' `get_queryset()`](https://github.com/mozilla/experimenter/pull/3740/files#diff-067426c4b37d1512505cb576ee0ce4dcddab1b8666940d06dcdb9f35f2e31b4aL50) before our code has a chance [to annotate it with `firefox_channel_sort`](https://github.com/mozilla/experimenter/pull/3740/files#diff-067426c4b37d1512505cb576ee0ce4dcddab1b8666940d06dcdb9f35f2e31b4aL51) - so the new validation error gets thrown.

But yeah, if we want to remove the feature altogether, I can just close this and we can take the next dependabot update after #3748 is addressed

PR #3752, opened 2020-10-14
fix #3512 feat(nimbus): add TargetingConfig

### LINKED ISSUE #3512: Add TargetingConfig constants
TargetingConfig should migrate this data:
https://github.com/mozilla/nimbus-shared/blob/main/data/Audiences.ts

NimbusExperiment must have exactly one targeting config
Get rid of Audience

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-329)
┆Issue Number: EXP-329

jaredlockhart:
Because

* We need a way to store the targeting configuration for a NimbusExperiment

This commit

* Migrates the targeting configuration data from the NimbusShared repo into constants in Experimenter


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -1,6 +1,60 @@
...
+@dataclass
+class NimbusTargetingConfig:
+    name: str
+    slug: str
+    description: str
+    targeting: str
+    desktop_telemetry: str
+
+
+TARGETING_ALL_ENGLISH = NimbusTargetingConfig(
```
bbangert, line 15: If these configurations originate elsewhere, like in the Nimbus Shared repo, it would be helpful if they included links/references back to the source material and perhaps a 'last update' or something so we know when they were last synced up. There's a pretty abundant use of jargon in the Python code so far, and practically no doc strings (I'm not sure I've seen any at all actually) explaining things at all.

jaredlockhart: Ah no @bbangert this file will be the only source and definition of targeting configs for Nimbus experiments.  This PR is meant to migrate those definitions away from shared.

jaredlockhart: And then they will be removed from the shared repo entirely.

bbangert: Ah, ok. For cross reference, where are the docs on what any of this means? I'm having a tough time reviewing this in as much as 'it looks like valid Python' as I have no reference for this.

jaredlockhart: I think the best documentation for that is here:

https://docs.google.com/document/d/16ekd-cXkJF8_UTgkZ3ufzJFusYz9htd5EE2NpIFFJp4/edit?skip_itp2_check=true&pli=1

bbangert (approved):
This looks fine,

PR #3762, opened 2020-10-15
fix #3548 feat(nimbus): add v6 REST API for DTO

### LINKED ISSUE #3548: Add v6 API DTO Serializer
Implement a serializer for the NimbusExperiment DTO (Data Transfer Object) that matches the schema in https://github.com/mozilla/nimbus-shared/blob/main/types/experiments.ts

When that schema is published to pypi we can use it to validate the output of the serializer (either in this ticket or a followup)

This will be used to publish the experiment to RS as well as be consumed by jetstream for automated analysis.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-352)
┆Issue Number: EXP-352

jaredlockhart:
Because
    
* We need a way to serialize a Nimbus experiment for both RS and automated data analysis
    
This commit
    
* Adds adds a JSON serializer for a NimbusExperiment and wraps it in a simple REST API
* Matches the schema found in https://github.com/mozilla/nimbus-shared/blob/main/types/experiments.ts


app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -0,0 +1,105 @@
...
+class NimbusBranchSerializer(serializers.ModelSerializer):
+    feature = serializers.SerializerMethodField()
+
+    class Meta:
+        model = NimbusBranch
+        fields = ("slug", "ratio", "feature")
+
+    def get_feature(self, obj):
+        return {
+            "featureId": obj.experiment.feature_config.slug,
```
bbangert, line 39: We're linking to the feature Id based on its slug, and not its id?

app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -0,0 +1,105 @@
...
+            "enabled": obj.enabled,
+            "value": obj.value,
+        }
+
+
+class NimbusExperimentSerializer(serializers.ModelSerializer):
+    id = serializers.ReadOnlyField(source="slug")
```
bbangert, line 46: These aren't stable because the UX lets the user change them at will. Is that still an appropriate output identifier?

app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -0,0 +1,105 @@
...
+            "endDate",
+            "proposedDuration",
+            "proposedEnrollment",
+            "referenceBranch",
+        )
+
+    def get_probeSets(self, obj):
+        return list(obj.probe_sets.all().order_by("slug").values_list("slug", flat=True))
+
+    def get_referenceBranch(self, obj):
```
bbangert, line 83: Why not just call this the controlBranch?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -45,8 +39,8 @@ class NimbusTargetingConfig:
         "locale using the release channel."
     ),
     targeting=(
-        "localeLanguageCode == 'en' && (isFirstStartup || '{slug}' in "
-        "activeExperiments) && browserSettings.update.channel == '{firefox_channel}'"
+        "localeLanguageCode == 'en' "
+        "&& (isFirstStartup || experiment.slug in activeExperiments)"
```
bbangert, line 43: Heh, that didn't take long for an update on how the targeting works. 😉

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -16,21 +16,15 @@ class NimbusTargetingConfig:
...
 
 TARGETING_US_ONLY = NimbusTargetingConfig(
     name="US users (en)",
     slug="us_only",
     description="All users in the US with an en-* locale using the release channel.",
-    targeting=(
-        "localeLanguageCode == 'en' && region == 'US' && "
-        "browserSettings.update.channel == '{firefox_channel}'"
-    ),
+    targeting="localeLanguageCode == 'en' && region == 'US'",
```
bbangert, line 27: Definitely worth considering whether we want to know historically whether an experiment in the past had a different targeting config for the same slug. Maybe @k88hudson would have some insight on if we want to track evolution of targeting config.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -99,14 +99,14 @@ class NimbusBranch(models.Model):
...
-    enabled = models.BooleanField(default=True)
     name = models.CharField(max_length=255, blank=False, null=False)
     slug = models.SlugField(
         max_length=NimbusConstants.MAX_SLUG_LEN, blank=False, null=False
     )
     description = models.TextField(default="")
     ratio = models.PositiveIntegerField(default=1)
-    value = models.TextField(blank=True, null=True)
+    feature_enabled = models.BooleanField(default=True)
+    feature_value = models.TextField(blank=True, null=True)
```
bbangert, line 109: 💯

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -37,10 +38,10 @@ class NimbusExperimentFactory(factory.django.DjangoModelFactory):
         lambda o: decimal.Decimal(random.randint(1, 10) * 10)
     )
     firefox_min_version = factory.LazyAttribute(
-        lambda o: random.choice(NimbusExperiment.Version.choices)
+        lambda o: random.choice(NimbusExperiment.Version.choices)[0]
```
bbangert, line 41: These aren't named tuples? Might be useful to use named tuples instead so we can have more meaningful access to the attribute.

bbangert (changes requested):

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: Yep in the client all identifiers are slugs

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: Yep, the experiment locks and is no longer editable when it is shipped to remote settings, so it becomes stable at that point.

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: The schema is defined here:

https://github.com/mozilla/nimbus-shared/blob/main/types/experiments.ts

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: ;)

app/experimenter/experiments/tests/factories/nimbus.py
jaredlockhart: Ah in theory, but using the Django TextChoices enum converts them to unnamed tuples when you access its choices list

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: And in the data science tooling as well

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Yep totally worth discussing, but also probably outside the scope of this PR

app/experimenter/experiments/tests/factories/nimbus.py
jaredlockhart: Oh! Actually you can treat the enum as an iterable and then you get the nice accessors, let's use those 👍

app/experimenter/experiments/api/v6/serializers.py
bbangert: Unless its rejected, in which case it can be edited again.

app/experimenter/experiments/api/v6/serializers.py
bbangert: Ok, perhaps obvious follow-up question, why not call this Reference Branch everywhere? (In our UX, in the model, etc.)

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: Yeah so the way we handled this in rapid was to omit experiments in `DRAFT` and `REVIEW` states from the API altogether, so that it only appears in the API when its in its locked state.  We could do that here too?

app/experimenter/experiments/api/v6/serializers.py
bbangert: Okey, so this is immutable?

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: Could be, but thinking about how things have gone in the past, the external facing names can change frequently and by that time it could wind up being major refactors to make the code reflect the external facing names, which we have done, but becomes more and more difficult over time.  This is how we wound up with `variants` vs `branches` for instance.  So we could take the opportunity to rename it in the code now while Nimbus is still young, but that doesn't guarantee it won't change again in the future.

app/experimenter/experiments/api/v6/serializers.py
bbangert: Well, there's the question of whether we should allow the user to edit the slug at all. If we're going to use this as an identifier (which it appears we are), I'd prefer the user have nothing to do with it. e.g. we slugify it on the server and that's what it is from then on.

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: We'll have to find a way to enforce that in the workflow since it'll have to match against expected identifiers in the client.  @k88hudson and I have talked about clients publishing a list of known identifiers, but I think we'll have to work that out with the client team later.

app/experimenter/experiments/constants/nimbus.py
bbangert: Yup.

bbangert (approved):
My remaining questions don't require code changes at this point and are mainly for reference on things I need to change in the GraphQL API.

app/experimenter/experiments/api/v6/serializers.py
jaredlockhart: Yeah that's what we did with legacy, slugs were never exposed in the UI and only autogenerated once at creation time and not editable after that.  But we found that in some cases users wanted to have explicit control over them so the theory in Nimbus was to expose them in the UI and then just follow the pattern of locking the experiment once it moves out of Experimenter.

PR #3766, opened 2020-10-16
fix #3748 fix(legacy): removed firefox channel sorting

### LINKED ISSUE #3748: Remove Firefox Channel Sorting
A long long time ago, in a galaxy far far away, we added an ability to sort by Firefox Channel that preserves the 'intuitive' ordering of Firefox Channels: Nightly < Beta < Release.  I don't think anyone has used this feature in 3 years and now a Django update is hiccuping on it so we should just remove it.

Let's get rid of this:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/legacy.py#L954-L972

and anything that depends on it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-489)
┆Issue Number: EXP-489

tiftran:
Because

* It's not being used by anyone
* it's cause some django update problems

This commit:

* Removes firefox channel sorting and everything related to it

### DIFF
--- a/app/experimenter/experiments/forms.py
+++ b/app/experimenter/experiments/forms.py
@@ -1008,8 +1008,6 @@ class ExperimentOrderingForm(forms.Form):
         ("latest_change", "Least Recently Updated"),
         ("firefox_min_version", "Firefox Min Version Ascending"),
         ("-firefox_min_version", "Firefox Min Version Descending"),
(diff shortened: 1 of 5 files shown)


jaredlockhart (approved):
Tytytyty 🙏 🙏 🙏

PR #3786, opened 2020-10-20
fixes #3466 feat(nimbus): apollo client + cache improvements, gql type gen, and useExitWarning hook

### LINKED ISSUE #3466: Add `experimentData` local state to Apollo Client cache
The [https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/rapid/contexts/experiment/hooks.ts](https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/rapid/contexts/experiment/hooks.ts) file contains `useExperimentState` and `useExperimentDispatch`. We want to setup the GQL implementation of this - that is, store the experiment data in apollo cache instead of context, and allow for a way to read and write to this data. This will, to reach parity, allow users to navigate various pages with their experiment editing work persisted, but will have additions later.

Check out the apollo client local state docs: [https://www.apollographql.com/docs/react/v2.4/essentials/local-state/](https://www.apollographql.com/docs/react/v2.4/essentials/local-state/)

See the types in the following file, we'll want to copy over/reference a large chunk of what's here: [https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/rapid/types/experiment.ts](https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/rapid/types/experiment.ts)

Also see what `typePolicies` look like in the `fxa-settings` project, which at the time of writing has an `Account` policy and a `Session` policy, giving us a "top level" data object in the cache: https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/cache.ts#L79

1) We want to add an `ExperimentData` policy in our cache, or perhaps we just want to call it `Experiment` or even `CurrentExperiment` - it's only going to be read/written to when an experiment is in the "Experiment Edit" view. Managing the cache for multiple experiments is out of scope for this issue.
2) For allowing updates to this piece of data, we probably want to set up a client resolver for writing/updating, which may also enable us to use [GQL polling](https://www.apollographql.com/docs/react/data/queries/#polling) for this data in the future. Make the convention for writing to the cache clear (maybe it will be just by writing tests for it).
3) Set up a `useExperimentData` hook to use to easily read this data (see the [Session file](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/models/Session.ts) in fxa-settings which can be used like `const session = useSession()` or `const { verified } = useSession();` )
4) Create a `MockedCache` test helper similar to [what's in fxa-settings](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/models/_mocks.tsx#L114) with `experiment` or `experimentData` instead of fxa-setting's `account`/`session` to make future testing with this data much easier. Document this and how to use it in the README.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-297)
┆Issue Number: EXP-297

jodyheavener:
This PR is a bit of a hodgepodge of things. 😅 It closes #3466, which originally intended to set up a system for pending Experiment changes to be carried from page-to-page in the cache until they were saved by the user. After [further discussion](https://mozilla.slack.com/archives/CANCHH6R1/p1603120693338100) we changed how this flow will work, and as a result we no longer need the ability to store changes, but we can still make use of some of the code I already wrote (even if it won't be immediately used).

With that in mind, this PR...

- Sets up a more elaborate Apollo client and cache. Additionally introduces the foundation and instructions for `MockedCache`, with which we'll be able to mock out cached GraphQL mutations once we start writing them down the road.
- Adds and documents the `yarn types` command that uses the [Apollo CLI](https://www.npmjs.com/package/apollo#apollo-clientcodegen-output)'s `codegen` command to generate Typescript types for GraphQL queries.
- Adds a new hook, `useExitWarning`, to allow you to prompt a user to confirm that they want to exit the page. This will be useful for when the Experiment form pages are implemented and a user makes changes that are not saved before they exit the page.

### DIFF
--- a/app/experimenter/nimbus-ui/README.md
+++ b/app/experimenter/nimbus-ui/README.md
@@ -35,7 +35,47 @@ Except, we know what we're doing. And because we _do_ want all the benefits of C
 
 You can read up over on their page if you want to know all the details, but if you're looking to make changes to CRA's Webpack configuration, you can do so inside [`.rescriptsrc.js`](.rescriptsrc.js). Rescripts will consume each exported item from this file, supplying it the full config as a function argument, and expecting the modified config back in return.
 
-## Testing
+## Accessing GraphQL Type Definitions
+
+All of our GraphQL resolvers produce types that are compatible with TypeScript, so there is no need to write new types for the queries you perform with Apollo.
+
+To generate types, just run `yarn types`. This will generate type declarations, inside a `types/` directory, relative to the path where the query was performed. Once generated you can import and use as you see fit.
+
+For example, a query that looks like this:
+
+```ts
+export const GET_EXPERIMENT_OVERVIEWS = gql`
+  query GetExperimentOverviews {
+    experiments {
+      name
+      slug
+      hypothesis
+    }
+  }
+`;
+```
+
+Would be analyzed and the corresponding types generated would look like this:
+
+```ts
+export interface GetExperimentOverviews_experiments {
+  __typename: "NimbusExperimentType";
+  name: string;
+  slug: string;
+  hypothesis: string | null;
+}
+
+export interface GetExperimentOverviews {
+  /**
+   * List Nimbus Experiments.
+   */
+  experiments: (GetExperimentOverviews_experiments | null)[] | null;
+}
+```
+
+**Note**: this command performs introspection on the GraphQL endpoint, so your local server must be running for it to work.
+
+## Testing and Mocking
 
 This package uses [Jest](https://jestjs.io/) to test its code. By default `yarn test` will test all JS/TS files under `src/`.
 
@@ -50,6 +90,52 @@ yarn test -t="renders as expected"
 ```
 
 Refer to Jest's [CLI documentation](https://jestjs.io/docs/en/cli) for more advanced test configuration.
+
+### Components that need a GQL Mock
+
+⚠️ TODO: This section needs to be expanded with mock-able properties once we need to read and write from the cache.
+
+[MockedCache](./src/services/mocks.tsx) is a convenient way to test components that make use of GraphQL mutations. Use it in place of [MockedProvider](https://www.apollographql.com/docs/react/api/react/testing/#mockedprovider) without prop overrides to use the default mocked cache. A `mocks` prop can also be passed in when a query or mutation needs success or failure mocks.
+
+Example:
+
+```jsx
+const mocks = [];
+<MockedCache {...{ mocks }}>
+  <ExperimentsDirectory />
+</MockedCache>;
+```
+
+### Mocking mutation errors
+
+Testing for GQL and network errors is also pretty straightforward. In your tests, wrap your component in a `MockedCache` and provide it with mock mutations that produce either an array of `GraphQLError`s, or a standard `Error`. Example with both:
+
+```tsx
+const mocks = [
+  {
+    request: {
+      query: GET_EXPERIMENT_OVERVIEW,
+      variables: { slug: 'foo' },
+    },
+    result: {
+      errors: [new GraphQLError('invalid slug')],
+    },
+  },
+  {
+    request: {
+      query: GET_EXPERIMENT_OVERVIEW,
+      variables: { slug: 'foo' },
+    },
+    error: new Error('network error'),
+  },
+];
+
+renderWithRouter(
+  <MockedCache {...{ mocks }}>
+    <ExperimentsDirectory {...{ onDismiss, onError }} />
+  </MockedCache>
+);
+```
 
 ## Storybook
 
--- a/app/experimenter/nimbus-ui/src/services/apollo.test.ts
+++ b/app/experimenter/nimbus-ui/src/services/apollo.test.ts
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { createApolloClient, cache } from "./apollo";
+import { ApolloClient, InMemoryCache } from '@apollo/client/core';
+
+describe("services/apollo", () => {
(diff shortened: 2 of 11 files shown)


app/experimenter/nimbus-ui/README.md
```diff
@@ -51,6 +91,52 @@ yarn test -t="renders as expected"
 
 Refer to Jest's [CLI documentation](https://jestjs.io/docs/en/cli) for more advanced test configuration.
 
+### Components that need a GQL Mock
+
+⚠️ TODO: This section needs to be expanded with mock-able properties once Experimenter queries are added.
```
LZoog, line 96: I feel the real value of this test helper can't really be seen until we work on the directory view and need to set the cache (which right now is empty). I think we can do the same thing with `<MockedProvider {...{ mocks }}>` as we do here. Should this say "once we need to read and write from the cache" instead of "once Experimenter queries are added"?

jodyheavener: Yeah that's cool with me! I'll note that since `MockedCache` takes `mocks`, you could still use it over `MockedProvider` (if for no reason other than to be consistent across tests).

app/experimenter/nimbus-ui/src/services/apollo.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    afterEach(() => {
+      (console.error as jest.Mock).mockRestore()
+    })
+
+    it("logs the errors (for now)", () => {
+      errorHandler(errorResponse);
+      expect(console.error).toHaveBeenCalledWith(
+        errorResponse.graphQLErrors,
+        errorResponse.networkError
```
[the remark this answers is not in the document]
jodyheavener: Ooohh you know what, you are correct. The response has an array of `errors`:

Screen Shot 2020-10-20 at 11 24 24 AM

Will update (maybe remove for EXP-473)

app/experimenter/nimbus-ui/README.md
```diff
@@ -51,6 +91,52 @@ yarn test -t="renders as expected"
...
+  {
+    request: {
+      query: GET_EXPERIMENT_OVERVIEW,
+      variables: { slug: 'foo' },
+    },
+    error: new Error('network error'),
+  },
+];
+
+renderWithRouter(
```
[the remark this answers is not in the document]
jodyheavener: Oops, that was a copy-paste. 😅

app/experimenter/nimbus-ui/src/services/hooks.ts
```diff
@@ -0,0 +1,48 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { useEffect, useState } from "react";
+
+/**
+ * Hook to trigger a browser dialog when a user attempts to exit the page.
+ *
+ * Example:
+ *
+ * const shouldWarnOnExit = useExitWarning();
```
LZoog, line 12: What's the use case for `shouldWarnOnExit(false)`, wouldn't we just not use this hook? If we set `useExitWarning` to default to true, we don't even need `const shouldWarnOnExit = useExitWarning();`, we can simply use 

```
          <button
            onClick={() => {
              useExitWarning();
            }}
          >
```

jodyheavener: Good question. We need to keep a reference to the hook so when the user clicks the "Next" button (or whatever the save and continue button is) we can turn off this exit warning.

LZoog: Since we'll have a "Save" button and a "Next" button which is essentially "Save and navigate," I think we'll only need this in the navigation right? We could have something like, a `hasMadeChanges` var that sets to `true` onChange, and if that's true, use the hook on navigation attempt. At least, this is what I thought "Next" would be. Maybe it doesn't also "Save"?

jodyheavener: Yeah, that def could work. Isn't that more or less the same level of effort than setting the hook up at the top and setting it to true or false?

app/experimenter/nimbus-ui/src/services/hooks.test.tsx
```diff
@@ -0,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
lmorchard, line 1: Maybe we should rename `services/` to `lib/`?

FWIW, I had something like [this](https://dev.to/farazamiruddin/an-opinionated-guide-to-react-folder-structure-file-naming-1l7i) or [this](https://www.robinwieruch.de/react-folder-structure) in mind when I started the `services/` directory. So, I would put this file in `hooks/index.tsx` and keep `services/` just for external network or DOM API code.

But, this is a road leading straight to a bikeshed and IMO not worth the energy to draft an ADR. So it's probably worth simplifying.

app/experimenter/nimbus-ui/src/services/hooks.ts
LZoog: I'm going to go ahead and r+ since this shouldn't block the PR merge, if we want to simplify this later when we actually implement it we can.

app/experimenter/nimbus-ui/src/services/hooks.test.tsx
jodyheavener: That's a good point. I'm happy to adopt the structure from that first link, or move it to `lib/`. What do you think @LZoog?

Either way I would want to do this in a followup so I can ignore the ref locally

LZoog (approved):

app/experimenter/nimbus-ui/src/services/hooks.test.tsx
LZoog: Yeah when I reviewed this I almost suggested moving it into `lib/` or another directory since I didn't really consider hooks "services," but was OK with it if you preferred them in `services/`. I'm down for moving it out.

PR #3787, opened 2020-10-20
fix #3715 feat(visualization) New overview design.

### LINKED ISSUE #3715: Update the visualization overview design
The overview design will look a little different as per mockups.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-475)
┆Issue Number: EXP-475

emtwo:
Screenshot for reference:

Screen Shot 2020-10-20 at 3 56 48 PM

### DIFF
--- a/app/experimenter/legacy-ui/rapid/components/experiments/utils.ts
+++ b/app/experimenter/legacy-ui/rapid/components/experiments/utils.ts
@@ -1,0 +1,28 @@
+import { XSelectOption } from "experimenter-rapid/components/forms/XSelect";
+
+export const displaySelectOptionList = (
+  options: XSelectOption[],
+  values: string | string[],
+): string[] => {
+  let selectedValue: string | string[] = values;
+  if (!Array.isArray(values)) {
+    selectedValue = [values];
+  }
+
+  const selectedOption = options.reduce((filtered: string[], element) => {
+    if (selectedValue.includes(element.value)) {
+      filtered.push(element.label);
+    }
+
+    return filtered;
+  }, []);
+  return selectedOption;
+};
+
+export const displaySelectOptionLabels = (
+  options: XSelectOption[],
+  values: string | string[],
+): string => {
+  const selectedOption = displaySelectOptionList(options, values);
+  return selectedOption.join(", ");
+};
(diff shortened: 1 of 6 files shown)


emtwo:
@jodyheavener @LZoog I added both of you to the review. Wondering what your general preferences are, if any, for review? Whoever gets to it first? Whoever has been reviewing this area of the code already (Jody)? Taking turns? No preference?

jaredlockhart:
@jodyheavener Yeah I think we can totally do that if we want to, maybe we should bring it up at the retro?

app/experimenter/legacy-ui/rapid/components/experiments/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+    selectedValue = [values];
+  }
+
+  const selectedOption = options.reduce((filtered: string[], element) => {
+    if (selectedValue.includes(element.value)) {
+      filtered.push(element.label);
+    }
+
+    return filtered;
+  }, []);
```
jodyheavener, line 18: Is this just filtering an array? Maybe I'm misreading it. If it is, you could just do

```js
const selectedOption = options.filter(o => selectedValue.includes(o.value))
```

jodyheavener: Oh I see you're also returning the label string. In that case:

```js
const selectedOption = options.filter(o => selectedValue.includes(o.value)).map(o => o.label);
```

But that starts to get into the size of just using reduce, so I'm fine leaving it as you have it.

app/experimenter/legacy-ui/rapid/components/visualization/Overview.tsx
```diff
@@ -0,0 +1,69 @@
...
+              {displaySelectOptionList(
+                featureOptions,
+                experimentData.features,
+              ).map((feature) => {
+                return (
+                  <p key={feature} className="h6">
+                    {feature}
+                  </p>
+                );
+              })}
```
jodyheavener, line 55: Since you're not performing any JS operations before render you could just return the JSX shorthand style (same with `targetingValues`):

```tsx
).map((feature) => (
  <p key={feature} className="h6">
    {feature}
  </p>
))} 
```

Not a blocker though; using explicit return is totally valid.

jodyheavener (approved):
This looks great! No blockers from me, just some ramblings that you can probably ignore 🤪

PR #3788, opened 2020-10-20
fix #3639 feat(nimbus): add changelogs to Nimbus experiment

### LINKED ISSUE #3639: Add Changelogs to NimbusExperiment
We need to attach Changelogs to NimbusExperiment to track changes, dates, users, etc.

We'll need to add a second foreign key on Changelog to NimbusExperiment, and make sure all the utility functions accept NimbusExperiment correctly.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-438)
┆Issue Number: EXP-438

jaredlockhart:
Because

* We have used a pattern of storing all changes to an experiment in a changelog model
* It stores the old status, new status, who changed it, when, and what was changed on the model
* This lets us track the dates that all changes are made, especially when it launched or ended by looking
  for the changelog with the relevant pair of statuses

This commit

* Adds a NimbusChangeLog model very similar to the old ExperimentChangeLog model
* Adds a serializer to dump out the entire experiment to store on the changelog
* Adds a convenience method to generate changelogs that can be used anywhere we want to change a nimbus
  experiment, such as in GraphQL mutations, celery tasks, etc
* Updates existing GraphQL mutations to create a changelog
* Adds all relevant tests


app/experimenter/experiments/tests/test_changelog_utils/test_changelog_utils_nimbus.py
```diff
@@ -0,0 +1,162 @@
...
+        project = ProjectFactory.create()
+
+        experiment = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.COMPLETE,
+            application=application,
+            probe_sets=[probe_set],
+            feature_config=feature_config,
+            projects=[project],
+        )
+        data = dict(NimbusExperimentChangeLogSerializer(experiment).data)
```
tiftran, line 61: feel like i'm missing something, why do we need to cast it to a dict?

jaredlockhart: It comes out of the serializer as an OrderedDict, so for proper dict comparisons gotta cast it to a dict explicitly

jaredlockhart: In some of our older tests we did it the other way around and compared everything as OrderedDicts but when I was playing with it I found it simpler to compare everything as dicts

tiftran: ohhhhh right, the pain of OrderedDicts. Makes sense

tiftran (commented):
Changelog logic seems good to me?

jaredlockhart:
Thnx for looking @tiftran !

bbangert (approved):

PR #3765, opened 2020-10-16
fix #3465 feat(nimbus): add new experiment form and page

lmorchard:
Because
* We want to accept submissions for new experiments

This commit:
* Adds a form to the new experiment page to accept experiment details

PR #3791, opened 2020-10-21
fix #3468 feat(nimbus): wire up GQL mutation for create form submit

### LINKED ISSUE #3468: `PageNew` form submission (`createExperiment` GQL mutation)
EXP-296 adds a dummy “Create experiment” button on the `nimbus/new` page. This task handles submission.

1) If a field is invalid when a form is attempted to be submitted, make sure "This field may not be blank.” displays similarly to how it does currently (see https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/rapid/new/ and click “save”).

2) Execute the `createExperiment` GQL mutation, capturing the user's input in the variables, to submit the form. You can search through the FxA codebase for `useMutation` examples, [like at the top of this file](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx). Assume the mutation will be successful, we'll deal with any server errors in an immediate follow up ticket.

Navigating to a new page will be done in a follow up as well, feel free to throw in a console log or alert message when the mutation goes through assuming that doesn't add testing pain, it'll be removed in EXP-295.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-299)
┆Issue Number: EXP-299

lmorchard:
Because:

* we want to support creating an experiment from user input

This commit:

* sets the base URL for the GQL API from Django-supplied configuration

* tweaks server-side createExperiment mutation to auto-slugify
  experiment name and better match fields from UX design

* adds support for submit in-progress state and error messages in
  FormExperimentOverviewPartial

* wires up the client-side createExperiment GQL mutation in PageNew to
  create a new experiment on form submission

* adds SimulatedMockLink helper to simulate GQL responses in stories
  with arbitrary user input

### DIFF
--- a/app/experimenter/experiments/api/v5/inputs.py
+++ b/app/experimenter/experiments/api/v5/inputs.py
@@ -18,7 +18,8 @@ class ExperimentInput(graphene.InputObjectType):
 
 class CreateExperimentInput(ExperimentInput):
     name = graphene.String(required=True)
-    slug = graphene.String(required=True)
+    hypothesis = graphene.String(required=True)
+    application = graphene.String(required=True)
 
 
 class UpdateExperimentInput(ExperimentInput):
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -14,7 +14,7 @@
     "storybook": "start-storybook -p 3001",
     "build-storybook": "build-storybook",
     "eject": "react-scripts eject",
-    "types": "NODE_TLS_REJECT_UNAUTHORIZED=0 apollo codegen:generate --target typescript --endpoint https://localhost/api/v5/graphql types"
+    "types": "NODE_TLS_REJECT_UNAUTHORIZED=0 apollo codegen:generate --target typescript --endpoint https://localhost/api/v5/graphql --outputFlat src/types"
   },
   "jest": {
     "collectCoverageFrom": [
--- a/app/experimenter/nimbus-ui/src/components/FormExperimentOverviewPartial/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormExperimentOverviewPartial/index.tsx
@@ -3,24 +3,37 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React, { useCallback } from "react";
-import { useForm, SubmitHandler } from "react-hook-form";
+import { useForm } from "react-hook-form";
 import Form from "react-bootstrap/Form";
+import Alert from "react-bootstrap/Alert";
 
 type FormExperimentOverviewPartialProps = {
-  onSubmit: SubmitHandler<Record<string, any>>;
+  isLoading: boolean;
+  submitErrors: Record<string, string[]>;
+  onSubmit: (data: Record<string, any>, reset: Function) => void;
   onCancel: (ev: React.FormEvent) => void;
   applications: string[];
 };
 
 const FormExperimentOverviewPartial = ({
+  isLoading,
+  submitErrors,
   onSubmit,
   onCancel,
   applications,
 }: FormExperimentOverviewPartialProps) => {
-  const { handleSubmit, register, errors, formState } = useForm({
+  const { handleSubmit, register, reset, errors, formState } = useForm({
     mode: "onTouched",
   });
   const { isSubmitted, isValid, isDirty, touched } = formState;
+
+  const handleSubmitAfterValidation = useCallback(
+    (data: Record<string, any>) => {
+      if (isLoading) return;
+      onSubmit(data, reset);
+    },
+    [isLoading, onSubmit, reset],
+  );
 
   const handleCancel = useCallback(
     (ev: React.FormEvent) => {
@@ -33,31 +46,52 @@ const FormExperimentOverviewPartial = ({
   const nameValidated = (
     name: string,
     // This could be ValidationRules from react-hook-form/validator, but
-    // register() has several signatures and only this one produces a ref 
-    validateRule: Parameters<typeof register>[1] = { required: "Required" },
+    // register() has several signatures and only this one produces a ref
+    validateRule: Parameters<typeof register>[1] = {
+      required: "This field may not be blank.",
+    },
   ) => ({
     name,
-    isInvalid: touched[name] && errors[name],
-    isValid: touched[name] && !errors[name],
+    isInvalid: !!submitErrors[name] || (touched[name] && errors[name]),
+    isValid: !submitErrors[name] && touched[name] && !errors[name],
     ref: register(validateRule),
   });
+
+  const FormErrors = ({ name }: { name: string }) => (
+    <>
+      {errors[name] && (
+        <Form.Control.Feedback type="invalid" data-for={name}>
+          {errors[name].message}
+        </Form.Control.Feedback>
+      )}
+      {submitErrors[name] && (
+        <Form.Control.Feedback type="invalid" data-for={name}>
+          {submitErrors[name]}
+        </Form.Control.Feedback>
+      )}
+    </>
+  );
 
   return (
     <Form
       noValidate
-      onSubmit={handleSubmit(onSubmit)}
+      onSubmit={handleSubmit(handleSubmitAfterValidation)}
       validated={isSubmitted && isValid}
       data-testid="FormExperimentOverviewPartial"
     >
+      {submitErrors["*"] && (
+        <Alert data-testid="submit-error" variant="warning">
+          {submitErrors["*"]}
+        </Alert>
+      )}
+
       <Form.Group controlId="name">
         <Form.Label>Public name</Form.Label>
         <Form.Control {...nameValidated("name")} type="text" autoFocus />
         <Form.Text className="text-muted">
           This name will be public to users in about:studies.
         </Form.Text>
-        <Form.Control.Feedback type="invalid">
-          This field may not be blank.
-        </Form.Control.Feedback>
+        <FormErrors name="name" />
       </Form.Group>
 
       <Form.Group controlId="hypothesis">
@@ -66,9 +100,7 @@ const FormExperimentOverviewPartial = ({
         <Form.Text className="text-muted">
           You can add any supporting documents here.
         </Form.Text>
-        <Form.Control.Feedback type="invalid">
-          This field may not be blank.
-        </Form.Control.Feedback>
+        <FormErrors name="hypothesis" />
       </Form.Group>
 
       <Form.Group controlId="application">
@@ -82,19 +114,23 @@ const FormExperimentOverviewPartial = ({
         <Form.Text className="text-muted">
           Experiments can only target one Application at a time.
         </Form.Text>
-        <Form.Control.Feedback type="invalid">
-          This field may not be blank.
-        </Form.Control.Feedback>
+        <FormErrors name="application" />
       </Form.Group>
 
       <div className="d-flex flex-row-reverse bd-highlight">
         <div className="p-2">
           <button
+            data-testid="submit-button"
             type="submit"
+            onClick={handleSubmit(handleSubmitAfterValidation)}
             className="btn btn-primary"
-            disabled={!isDirty || !isValid}
+            disabled={isLoading || !isDirty || !isValid}
           >
-            Create experiment
+            {isLoading ? (
+              <span>Submitting</span>
+            ) : (
+              <span>Create experiment</span>
+            )}
           </button>
         </div>
         <div className="p-2">
--- a/app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
@@ -3,15 +3,26 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from "react";
-import { render, screen, fireEvent, cleanup } from "@testing-library/react";
+import {
+  render,
+  screen,
+  fireEvent,
+  cleanup,
+  act,
+} from "@testing-library/react";
+import { MockedCache } from "../../lib/mocks";
 import PageNew from ".";
+import { CREATE_EXPERIMENT_MUTATION } from "../../gql/experiments";
 
 describe("PageNew", () => {
-  let origConsoleLog: typeof global.console.log;
-
   beforeEach(() => {
     origConsoleLog = global.console.log;
     global.console.log = jest.fn();
+    mockSubmit = {
+      name: "Foo bar baz",
+      hypothesis: "Some thing",
+      application: "firefox-desktop",
+    };
   });
 
   afterEach(() => {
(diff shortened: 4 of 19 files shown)


lmorchard:
Opening as draft to see if CI passes and for any early comments. Still need to update stories to handle the form changes and mock GQL mutations.

lmorchard:
Rebased against main, reworked some tests, finally got past the linters, opening for review

jodyheavener (commented):
Couple initial thoughts, doing another pass shortly.

app/experimenter/experiments/api/v5/inputs.py
```diff
@@ -12,7 +12,8 @@ class ExperimentInput(graphene.InputObjectType):
 
 class CreateExperimentInput(ExperimentInput):
     name = graphene.String(required=True)
-    slug = graphene.String(required=True)
+    hypothesis = graphene.String(required=True)
+    application = graphene.String(required=True)
```
bbangert, line 16: This should be an `enum`, but I can fix this in a later PR as it will require some additional tweaks in my current issue I'm almost done with.

app/experimenter/nimbus-ui/src/components/FormExperimentOverviewPartial/index.tsx
```diff
@@ -33,31 +46,56 @@ const FormExperimentOverviewPartial = ({
   const nameValidated = (
     name: string,
     // This could be ValidationRules from react-hook-form/validator, but
-    // register() has several signatures and only this one produces a ref 
-    validateRule: Parameters<typeof register>[1] = { required: "Required" },
+    // register() has several signatures and only this one produces a ref
+    validateRule: Parameters<typeof register>[1] = {
+      required: "This field may not be blank.",
+    },
```
jodyheavener, line 52: UX thought: currently if you click outside of the focused field even briefly it turns red and yells at you (it also shifts the page). Would it be possible to move field validation to form submit instead of field blur?

bbangert (commented):
The backend stuff looks good to me!

jodyheavener:
I don't know if there's a good solution to this, but wanted to point out that this looks a little funky

Screen Shot 2020-10-22 at 11 40 25 AM

app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
```diff
@@ -40,33 +96,67 @@ describe("PageNew", () => {
...
+  } = {}) => [
+    {
+      request: {
+        query: CREATE_EXPERIMENT_MUTATION,
+        variables: {
+          input: mockSubmit,
+        },
+      },
+      result: {
+        errors: undefined as undefined | any[],
```
jodyheavener, line 112: What's `undefined as undefined` doing?

app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
```diff
@@ -40,33 +96,67 @@ describe("PageNew", () => {
...
+        data: {
+          createExperiment: {
+            clientMutationId: "8675309",
+            message,
+            status,
+            nimbusExperiment,
+          } as any | undefined,
```
jodyheavener, line 119: Once types are automatically generated it should be possible to swap out `any` I think?

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
```diff
@@ -2,29 +2,70 @@
...
+
 const TRAINING_DOC_URL =
   "https://mana.mozilla.org/wiki/display/FJT/Project+Nimbus";
 
 type PageNewProps = {} & RouteComponentProps;
 
 const PageNew = (props: PageNewProps) => {
-  // TODO: Get this from constants / config loaded at app start?
+  // TODO: EXP-462 Get this from constants / config loaded at app start?
   const applications = ["firefox-desktop", "fenix", "reference-browser"];
```
jodyheavener, line 25: No need to use a constant or load it at app start - since we can auto-generate types we can just grab this enum as demonstrated here: https://github.com/mozilla/experimenter/pull/3798/files#diff-978e704d3641f4527d638f00c3609cbbb5bd612bdd3b4ceb21eeeb271ea5c766R13-R17

(unless we feel like this needs to be super fresh and come directly from the server)

app/experimenter/nimbus-ui/src/models/Experiment.ts
```diff
@@ -0,0 +1,38 @@
...
+  error: string;
+  message: string | Record<string, string[]>;
+  clientMutationId: string | null;
+  nimbusExperiment: {
+    name: string;
+    slug: string;
+    hypothesis: string;
+    application: string;
+  };
+};
```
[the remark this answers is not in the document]
lmorchard: Yeah, I couldn't get `yarn types` to work... looking at it again, I think it's because I'm using a semi-dockerized setup, so my Django is running at `http://localhost:7000` and not the `--endpoint https://localhost` in package.json.

Are those types going to get checked in? If so, we might want to work some of that into CI. That is, fail if types haven't been generated for a commit. We have something like that for API docs where the job fails if checked-in docs differ from what a new generation run would produce

For `models` vs `gql` folder names, I was trying to follow what looks like the convention [from fxa-settings](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings/src/models). But I can change that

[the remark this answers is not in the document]
jodyheavener: > I think it's because I'm using a semi-dockerized setup

Hmm, I'd love to use staging or prod so it doesn't require your local server to be up, but it is tucked under LDAP.

> Are those types going to get checked in?

I think they should be, unless we want to shove the command in front of the `yarn start` command so they are reliably generated each time.

app/experimenter/nimbus-ui/src/components/FormExperimentOverviewPartial/index.tsx
```diff
@@ -33,31 +46,56 @@ const FormExperimentOverviewPartial = ({
...
+    isInvalid: !!submitErrors[name] || (touched[name] && errors[name]),
+    isValid: !submitErrors[name] && touched[name] && !errors[name],
     ref: register(validateRule),
   });
 
+  const FormErrors = ({ name }: { name: string }) => (
+    <>
+      {!errors[name] ? (
+        ""
+      ) : (
```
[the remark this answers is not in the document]
lmorchard: Ah, yup, that would be simpler. I see what I did... at one point this only had one of the two error messages, and React didn't like that `{condition && </>}` returned undefined for the overall component. But now that there's two things here, it's at least an empty component rather than undefined.

lmorchard:
For the select element styling, that seem to be what Bootstrap is doing out of the box. [lzoog saw it in my PR for EXP-296 and I think we're aiming at EXP-462 for style tweaks](https://github.com/mozilla/experimenter/pull/3765#discussion_r506532205)

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
lmorchard: Yeah, that's a concern for EXP-462 - been assuming that it'll need to come from the server, not sure how variable the application list will be

app/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -60,3 +70,55 @@ export class MockedCache extends React.Component<MockedProps, MockedState> {
...
+
+type ResultFunction = (operation: Operation) => FetchResult;
+
+type SimulatedMockedResponses = ReadonlyArray<{
+  request: MockedResponse["request"];
+  result: ResultFunction;
+  delay?: number;
+}>;
+
+export class SimulatedMockLink extends ApolloLink {
```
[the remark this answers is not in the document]
lmorchard: I kind of hacked this together as an alternative to MockedProvider for stories. Looked like the out-of-box mock required all expected inputs to be defined up front as a part of tests. But, I wanted a more flexible way to simulate requests with arbitrary input for manually-driven stories. I'll throw something in the README about it

app/experimenter/nimbus-ui/src/models/Experiment.ts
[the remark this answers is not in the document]
lmorchard: Okay, looks like I got the type generation to work locally. Seeing some wrinkles in the process, though.
```
npx apollo codegen:generate --target typescript --endpoint http://localhost:7001/api/v5/graphql types
```
There seem to be two types directories produced, one of which is outside of the `src/` directory:
```
types/globalTypes.ts
src/models/types/createExperiment.ts
```
Appears to work, but I seem to remember that CRA balked at code outside of `src/`.

Also wrestling a little bit with the generated types: `createExperiment_createExperiment_nimbusExperiment` has `slug` as a required field, but as far as I can tell it's not required on the Django side. 

Also the types all have `__typename` as a required constant, which doesn't seem correct? Not sure because I'm kind of still wrapping my head around this stuff.

And last thing I'm seeing now: the types are un-tested, so we probably want to exclude those files from code coverage (they come up as 0%).

app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
lmorchard: Not just `undefined as undefined` - the `| any[]` also means that the result type for the function allows undefined or an array for errors. Did this inline rather than define a separate return type for `mkGqlMocks`

app/experimenter/nimbus-ui/src/models/Experiment.ts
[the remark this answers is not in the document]
jodyheavener: > For models vs gql folder names, I was trying to follow what looks like the convention from fxa-settings. But I can change that

Yeah totally, I guess I figured since it _only_ contained GQL stuff, and other utilities for the same record would be spread in to, for example, `hooks/`, it didn't really feel model-worthy. Definitely not a hill I'd die on though :)

app/experimenter/nimbus-ui/src/components/FormExperimentOverviewPartial/index.tsx
lmorchard: Can we push that to a follow-up issue? Will be a point or two of reworking tests for that and this PR's gotten pretty big as-is.

jodyheavener: Oh of course, it might also be worth asking the design folks, they might disagree with me.

app/experimenter/experiments/api/v5/inputs.py
bbangert: @lmorchard if you rebase, my update with `graphene.Enum`'s are in now, so this can be updated to use a `NimbusExperimentApplicationType` enum in a similar manner from my PR. You'd need to make that type like so:
https://github.com/mozilla/experimenter/pull/3803/files#diff-0f23f9b75c80649a9611b784fb9a9c1b2bd38b3bed19cf3bea723fe59763c5a9R17

Add it under the NimbusExperiment:
https://github.com/mozilla/experimenter/pull/3803/files#diff-0f23f9b75c80649a9611b784fb9a9c1b2bd38b3bed19cf3bea723fe59763c5a9R47

Then use it in the Input:
https://github.com/mozilla/experimenter/pull/3803/files#diff-1bc92d6c96b25929bbe8ae9fbccc529d28cdd5a3a276f487b9a99bca09485db3R77

app/experimenter/nimbus-ui/src/models/Experiment.ts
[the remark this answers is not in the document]
jodyheavener: > I seem to remember that CRA balked at code outside of src/.

Hm, I know the error you're talking about, and I think that might be specific to component imports? Maybe not. However, I'm not seeing this error when I yarn build in my branch.

> Also the types all have __typename as a required constant

If the addition of __typename is causing issues we can def remove them with `--no-addTypename` ([docs](https://www.npmjs.com/package/apollo#apollo-clientcodegen-output))

app/experimenter/nimbus-ui/package.json
```diff
@@ -14,7 +14,7 @@
     "storybook": "start-storybook -p 3001",
     "build-storybook": "build-storybook",
     "eject": "react-scripts eject",
-    "types": "NODE_TLS_REJECT_UNAUTHORIZED=0 apollo codegen:generate --target typescript --endpoint https://localhost/api/v5/graphql types"
+    "types": "NODE_TLS_REJECT_UNAUTHORIZED=0 apollo codegen:generate --target typescript --endpoint https://localhost/api/v5/graphql --outputFlat src/types"
```
lmorchard, line 17: Wrestled with the type generation a bit and had to make this change. Some types were generated outside of the source tree, which broke Storybook at least.

app/experimenter/nimbus-ui/src/models/Experiment.ts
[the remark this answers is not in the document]
lmorchard: Moved the GQL to `gql/` and rolled in the generated types. Discovered that `createExperiment_createExperiment_nimbusExperiment` was the wrong type, using `CreateExperimentInput` from global types worked here.

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
lmorchard: We can still revisit in EXP-462, but I switched to the generated enum here

app/experimenter/nimbus-ui/package.json
jodyheavener: Ooh good catch for Storybook, hadn't tested that.

app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
lmorchard: Oh yeah, and the reason for this typing here is that it enables [this other line](https://github.com/mozilla/experimenter/pull/3791/files#diff-b489135824d87fb1bef4f7cb4c9258c8bac80d980a2321f80a3edf2c8c5929bcR84):
```
gqlMocks[0].result.errors = [new Error(expectedError)]
```

app/experimenter/nimbus-ui/src/components/PageNew/index.test.tsx
lmorchard: This was to accommodate an intentional breakage in a test to cause an error. Reworked it a bit to make that clearer closer to where it happens.

app/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -60,3 +70,55 @@ export class MockedCache extends React.Component<MockedProps, MockedState> {
...
+          );
+          observer.complete();
+        },
+        delay ? delay : 0,
+      );
+
+      return () => clearTimeout(timer);
+    });
+  }
+}
```
jodyheavener, line 124: I'm not actually sure what the rule is for the Nimbus app, but it seemed like at least for the rest of the codebase we're aiming for 100% coverage. These new mock utils would need to be covered, but also... they only exist for the purpose of testing, so 🤷

lmorchard: Was tempted to find a different directory for this - it's mainly for stories, and we don't test stories (kind of circular & redundant). It could be used in tests, at which point it would be used. Not sure we can / should count test mocks as part of coverage (also kind of circular & redundant)

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
jodyheavener: Oh hrm, maybe these don't work here like I expected them to? I just tried to create an experiment and got `"FENIX" is not a valid choice.`

jodyheavener: Maybe they just need to be downcased. But I dont want this to block you; if we need to use the old const for now hthat's fine with me.

app/experimenter/experiments/api/v5/inputs.py
lmorchard: I think we should defer this to EXP-462 https://jira.mozilla.com/browse/EXP-462

lmorchard:
Rebased, reworked to use generated GQL types. Some general polish left to do in this area, but I think that's mainly laid out in EXP-462.

jodyheavener (approved):
I'm happy with this, thank you for updating to used the generated types (and sorry for the headache with them).

r+ with one comment left about the application string being passed to the mutation that should be fixed so you can successfully create the Experiment, and a little note in the README about your mocking update would be good.

lmorchard:
> r+ with one comment left about the application string being passed to the mutation that should be fixed so you can successfully create the Experiment

Oh, hmm, it *was* working before I rebased. Looks like the list of Applications in the generated type do not match what the server considers valid choices. I'm just going to revert that tweak for now and punt to FXA-462 for polish - using this list allows successful creation `["firefox-desktop", "fenix", "reference-browser"]`

lmorchard:
Added a bit in the README about SimulatedMockLink, aiming to merge if/when CI goes green. Then, I'll probably pick up EXP-462 to move onto sanding off all the rough edges I can

jodyheavener:
Cool with me!

PR #3792, opened 2020-10-21
chore(nimbus): restructure js into technical folder types

jodyheavener:
@lmorchard had suggested [here](https://github.com/mozilla/experimenter/pull/3786#discussion_r508699126) that we could improve the folder structure of our JS, and I agree. While I didn't go as aggressive as the linked articles proposed (like creating a folder-index structure for each lib/service/hook, or moving page components into Pages), I did break things out into individual technical type folders. Totally open to feedback.

New structure:

```
- src/
--- components/
----- ComponentName/
------- index.tsx
------- index.test.tsx
--- hooks/
----- hookName.ts
----- hookName.test.ts
--- services/
----- serviceName.ts
----- serviceName.test.ts
--- lib/
----- libName.ts
----- libName.test.ts
```


app/experimenter/nimbus-ui/src/lib/config.test.ts
```diff
@@ -0,0 +1,125 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import config, {
```
lmorchard, line 5: I'd say `config` could be considered a service because it interfaces with browser DOM API to derive app config and needs mocking in tests. But, I wouldn't block over it

lmorchard (approved):

LZoog (approved):
I sort of wonder about separating every hook out into its own file (it's nice to import all the hooks you need from one file) but I don't necessarily think the solution is to stick them all into one `hooks.tsx` file is ideal either so this WFM.

app/experimenter/nimbus-ui/src/lib/config.test.ts
jodyheavener: I can dig it!

lmorchard:
> I sort of wonder about separating every hook out into its own file (it's nice to import all the hooks you need from one file) but I don't necessarily think the ideal solution is to stick them all into one hooks.tsx file so this WFM.

Yeah, one-hook-per-file might be overkill... but then again it keeps them nice and isolated and simple.

For import convenience, we could add an `index.tsx` to the hooks directory that's basically just a re-exporting table of contents:
```
export { useExitWarning } from 'useExitWarning';
```
Downside is that that would be manually maintained.

PR #3798, opened 2020-10-21
fixes #3471 feat(nimbus): query experimentBySlug for edits page

### LINKED ISSUE #3471: `experimentBySlug` GQL query on a`:slug:` route
When the user hits any `/:slug:/` page, we need to fetch the latest experiment data from the database to display it in the forms.

**See bottom of issue for decided approach, just keeping this around for context / in case the engineer taking this wants to know why this approach was chosen.**

10/19 edit: One more reason to go with this approach is we shouldn't be touching the apollo client cache now in the create/edit UI view based on a Slack/stand up convo. We need to send up all of the experiment data when updating any experiment fields.

---

This is done in `rapid` with approach 1 and an `ExperimentProvider` that wraps the summary page and the "experiment edit" page.

There's three ways I see we can do this. Note that "saving" functionality mentioned below is out of scope for this issue and "the data may become stale" refers to if another user (user 2) edited and saved the same experiment before user 1 saved their changes, if a user has two browser tabs with the same experiment open (probably without realizing it) and edits/saves the experiment in one and makes edits on the other, or if the user goes to the "edit" view of an experiment and either leaves the tab open for a period of time before making edits or even navigates elsewhere and comes back to "edit" - in these scenarios we may not be up-to-date with the experiment draft data in the server and stale data will be shown in the UI. 

1) Fetch the data once and use it as an initial `experimentData` stored in Apollo Client cache, and update this object when edits on the form are made. Then, send this object up when the form is saved.

* Pros: This is how the `rapid` project does it, might be easiest to reach parity
* Cons: The data could become stale, the object is mutable (code smell) and we lose reference to the original data after making updates, and if wanted later, we'll have no way of conditionally enabling "Save" comparing the data vs edits

2) Fetch the data once and store it in a different cache key like `experiment`, and all edits go in an `experimentEdits` key

* Pros: We can duplicate the data in `experimentData` as a base and still use it for enabling "Save" or highlighting what's been edited, less network requests than option 2b/2c or 3
* Cons: the data from the server may become even more stale than option 1 because if the data exists in the cache we wouldn't refetch it even if the user navigates to a different experiment and comes back

2b) #2, but poll the server for newest data

* Pros: Best way to guarantee freshest data from the server while keeping the edited data and data saved in the server separated
* Cons: Requires the most work to set up and test, hits the server with additional requests when this "stale data" issue is edgecasey

2c) #2, but add a "pull latest from server" button

* Pros: Similar to 2b with less guarantee of freshness
* Cons: Similar to 2b but less network requests, data may still become stale and its use might be confusing to users without a detailed explanation

3) Refetch the data per page navigation, don't store it in the cache, and all edits go in an `experimentEdits` key

* Pros: Easy to set up since we don't have to touch the cache, probably the best way to guarantee freshness without polling
* Cons: Less than 2b), but we'll hit the server with additional requests than 1) or 2)

---

At the time of writing, option 3 seems like the best bet. We can tweak this later if we think storing it in the cache and referencing that instead of requerying would be beneficial. Assuming we go in this direction, we literally just need to:

1) Execute the `experimentBySlug` GQL query on `PageEditOverview`, `PageEditBranches`, and `PageRequestReview` - we'll use the data in other issues

I'll leave this estimated at a 5 to account for all the reading above :wink:, in case there's further discussion, and to account for tests.

Note: Don't include the call to `/api/v3/visualization/` shown in `rapid` `fetchExperiment`, that'll be handled in post-parity and is blocked by EXP-286.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-302)
┆Issue Number: EXP-302

jodyheavener:
Closes #3471

This PR adds a hook, `useExperiment`, that takes a slug and performs a GraphQL query for Experiment data. The hook is then placed, as somewhat of a placeholder, in the `PageEditOverview` and `PageEditBranches` pages for future issues.


jodyheavener:
Oh! Also, you'll notice I added a new `gql` folder. I was thinking we could place GraphQL queries in here as they tend to run a little longer, and we don't have a model-specific place to put these. Open to feedback.

app/experimenter/nimbus-ui/src/gql/types/GetExperiment.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface GetExperiment {
+  /**
+   * Retrieve a Nimbus experiment by its slug.
+   */
+  experimentBySlug: GetExperiment_experimentBySlug | null;
+}
+
+export interface GetExperimentVariables {
+  slug: string;
+}
```
jodyheavener, line 54: @bbangert this is what the resulting Apollo codegen output looks like, in case you were curious 🙂

bbangert: Ah, interesting.

app/experimenter/nimbus-ui/src/hooks/useExperiment.tsx
```diff
@@ -0,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { useQuery } from "@apollo/client";
+import { GET_EXPERIMENT } from "../gql/experiments";
+import { GetExperiment } from "../gql/types/GetExperiment";
+
+/**
+ * Hook to retrieve all Experiment data to fill in fields on the Edit pages.
```
LZoog, line 10: Maybe rephrase this to...

> Hook to retrieve Experiment data by slug.

We'll be using this on the request review page as well as at least the visualization "Results" page.

app/experimenter/nimbus-ui/package.json
```diff
@@ -18,7 +18,9 @@
   },
   "jest": {
     "collectCoverageFrom": [
-      "!src/**/*.stories.*"
+      "!src/**/*.stories.*",
+      "!src/lib/mocks.tsx",
```
LZoog, line 22: Ooh, I filed a ticket where I mention this here: https://github.com/mozilla/experimenter/issues/3808

We can discuss later today maybe on if we want to exclude these from coverage?

app/experimenter/nimbus-ui/src/hooks/useExperiment.tsx
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Example:
+ *
+ * const experiment = useExperiment('the-slug');
+ *
+ * if (experiment === undefined) {
+ *   return <Loading />;
+ * }
+ *
+ * if (experiment === null) {
```
LZoog, line 24: Do these examples need to be updated since it looks like we can check `if (notFound)`?

jodyheavener: Oh, yes, thank you!

LZoog (approved):
LGTM! 👍

PR #3801, opened 2020-10-21
fix(nimbus): generate and add css to nimbus template in prod

jodyheavener:
This should fix the unstyled app on staging/production.


LZoog (approved):
Sweet, glad this doesn't _look_ like it was too much trouble at least, thank you for the fix!

PR #3805, opened 2020-10-22
fix #3783 chore(nimbus): rename control_branch to reference_branch

### LINKED ISSUE #3783: Rename control_branch to reference_branch
We have used the term `control` to refer to the special branch that automated analysis uses to compare branches, however in the data science toolsets (jetstream, mozanalysis) this is referred to as the `reference` branch, so we should update our code to use the word `reference` so that it's consistent with external APIs and dependencies.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-508)
┆Issue Number: EXP-508

jaredlockhart:
Because

* The data science tooling and shared schema refer to the special branch as "reference branch"

This commit

* Renames control_branch to reference_branch so that naming can be consistent across tools and domains


bbangert (approved):
Nice!

PR #3806, opened 2020-10-22
fix #3795 feat(nimbus): add status update mutation

### LINKED ISSUE #3795: GraphQL mutation to mark for review and restrict prior mutations
When an experiment is ready for review, a mutation to mark update the experiment by changing its status should exist. Prior mutations should also be updated to only allow the mutation to succeed if the experiment is in Draft status.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-514)
┆Issue Number: EXP-514

bbangert:
Because:

* We want to update the status to indicate it is ready for review.
* We want to restrict mutations to an experiment that is no longer
  in draft status.

This commit:

* Add's a mutation that allows a DRAFT experiment to be marked for
  review.
* Add's a serializer mixin to mutations to prevent updates if the
  experiment is not in draft status.

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -11,6 +11,24 @@ from experimenter.experiments.models.nimbus import (
 )
 
 
+class NimbusStatusRestrictionMixin:
+    required_status = NimbusExperiment.Status.DRAFT
+
+    def validate(self, data):
+        data = super().validate(data)
+        if self.instance and self.instance.status != self.required_status:
+            status = self.instance.status
+            raise serializers.ValidationError(
+                {
+                    "experiment": [
+                        f"Nimbus Experiment has status '{status}', but can only "
+                        f"be changed when set to '{self.required_status}'."
+                    ]
+                }
+            )
+        return data
+
+
 class NimbusChangeLogMixin:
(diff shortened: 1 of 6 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -11,6 +11,24 @@
...
+            status = self.instance.status
+            raise serializers.ValidationError(
+                {
+                    "experiment": [
+                        f"Nimbus Experiment has status '{status}', but can only "
+                        f"be changed when set to '{self.required_status}'."
+                    ]
+                }
+            )
+        return data
```
jaredlockhart, line 29: Yeah I was gonna suggest we do this at the queryset lookup level rather than at the validation level, but this also gives a very nice clear error message so that's probably better than 'unable to mutate: no such experiment', so ya looks good 👍

jaredlockhart (approved):
Looks great @bbangert !!  Very nic eputting it as a serializer mixin 👍 

In the old one, we had a specific state transition mapping:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/legacy.py#L123-L130

So we could do something like that here too, but I don't think we need to dump it in here we can add it later as a small extension to this.

A+ 10/10 would r+ again 🎉

PR #3807, opened 2020-10-23
fix #3790 feat(nimbus): add feature config mutation validation

### LINKED ISSUE #3790: Validate experiment branch updates
When updating an experiments branches, the following validation rules should be enforced:

* If a feature config is used, then the branch feature values must validate against the feature config schema
* The branch name can be run through slugify.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-510)
┆Issue Number: EXP-510

bbangert:
Because:

* We want to verify branch updates that include feature config
  information validate against the feature config schema.
* Branch names must be able to slugify.

This commit:

* Add's additional validation logic to ensure feature config's are
  present when needed and validate the enabled branch feature values.
* Ensures branch names result in a slugify value.

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1,3 +1,6 @@
+import json
+
+import jsonschema
 from django.db import transaction
 from django.utils.text import slugify
 from rest_framework import serializers
@@ -37,6 +40,34 @@ class NimbusChangeLogMixin:
 
 
 class NimbusBranchSerializer(NimbusChangeLogMixin, serializers.ModelSerializer):
+    def validate_name(self, value):
+        slug_name = slugify(value)
+        if not slug_name:
+            raise serializers.ValidationError(
+                "Name needs to contain alphanumeric characters."
+            )
+        return value
+
+    def validate(self, data):
+        data = super().validate(data)
+        if data.get("feature_enabled", False) and "feature_value" not in data:
+            raise serializers.ValidationError(
+                {
+                    "feature_enabled": (
+                        "feature_value must be specified if feature_enabled is True."
+                    )
+                }
+            )
+        if data.get("feature_value") and "feature_enabled" not in data:
+            raise serializers.ValidationError(
+                {
+                    "feature_value": (
+                        "feature_enabled must be specificed to include a feature_value."
+                    )
+                }
+            )
+        return data
+
     class Meta:
         model = NimbusBranch
         fields = (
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -114,6 +145,57 @@ class Meta:
...
+
+    def validate(self, data):
+        data = super().validate(data)
+        # Determine if we require a feature_config
+        feature_config_required = any(
+            [data["reference_branch"].get("feature_enabled", False)]
+            + list(
+                map(lambda b: b.get("feature_enabled", False), data["treatment_branches"])
+            )
+        )
```
jaredlockhart, line 166: For things where there's multiple complex nested expressions like this, I usually just break it into separate lines since it usually winds up being the same number of lines and is a little less fragile, something like:

```py
feature_config_required = data["reference_branch"].get("feature_enabled", False)
for branch in data["treatment_branches"]:
    branch_required = branch.get("feature_enabled", False)
    feature_config_requried = feature_config_required or branch_required
```

but that's just a style suggestion

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -114,6 +145,57 @@ class Meta:
...
+        for branch_data in data["treatment_branches"]:
+            if branch_data.get("feature_enabled", False):
+                errors = self._validate_feature_value_against_schema(
+                    schema, branch_data["feature_value"]
+                )
+                if errors:
+                    error_result.setdefault("treatment_branches", []).extend(errors)
+        if error_result:
+            raise serializers.ValidationError(error_result)
+        return data
```
jaredlockhart, line 197: Yep all looks great 👍 👍 👍 👍

app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -126,7 +143,69 @@ def test_saves_existing_experiment_with_changelog(self):
         self.assertEqual(experiment.changes.count(), 2)
 
 
-class TestUpdateNimbusExperimentOverviewSerializer(TestCase):
+class TestNimbusBranchSerializer(TestCase):
+    def test_branch(self):
```
jaredlockhart, line 147: If this is testing that a valid branch passes validation I might just say something about that in the test name

jaredlockhart (approved):
Ya yep everything looks right to me, couple naming/style suggestions but great test coverage and very idiomatic validation A+ 👏 👏 👏 👏

PR #3822, opened 2020-10-26
fix #3687 chore(project): run integration tests against deploy image

### LINKED ISSUE #3687: Run integration tests against deploy image
To make sure our deploy image is correctly we should use it in some tests, I think integration tests are the best target.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-456)
┆Issue Number: EXP-456

jaredlockhart:
Because

* We split the docker image into two images, one for local dev and one for deploying to prod so that the prod one would be small ~100mb
* We should find some way to test that the deploy image is working correctly
* A good candidate is to run the integration tests against it

This commit

* Adds a new make up_prod command that sets up an environment that more closely mirrors prod ie
  - Use the prod image
  - Disable all the yarn watchers
* Reorganizes the dockerfile to move frontend asset building out of the dev container to increase docker cache hits for local dev


Makefile
```diff
@@ -61,27 +65,44 @@ compose_build: build ssl
...
 	docker volume prune -f
 
+static_rm:
+	rm -Rf app/node_modules
+	rm -Rf app/experimenter/legacy-ui/core/node_modules
+	rm -Rf app/experimenter/legacy-ui/rapid/node_modules
+	rm -Rf app/experimenter/legacy-ui/visualization/node_modules
+	rm -Rf app/experimenter/nimbus-ui/node_modules
+	rm -Rf app/experimenter/legacy-ui/assets/
+	rm -Rf app/experimenter/nimbus-ui/build/
```
bbangert, line 85: Should this remove the Python cache directories too?

jaredlockhart: Oh we could add one for that, I was just thinking of this as a way to reset all the static paths, I didn't hit a need to remove the python cache dirs since they're ignored by the dockerignore

bbangert: The node_modules are also in the dockerignore, that's why I was wondering what the criteria used here was.

bbangert (approved):

Makefile
jaredlockhart: No this was for the workflow of resetting the things that run outside the container, a couple times recently people hit a case where those things got out of sync and resetting it all fixed it so I thought it would just be handy to have a reference for it, I've never hit a case where it was necessary to reset py cache dirs

PR #3823, opened 2020-10-26
fixes #3711 feat(nimbus): standardize and document error handling

### LINKED ISSUE #3711: Set GQL mutation/query error handling convention
We need to set a convention for error handling from the server in the `nimbus-ui` application when performing a GQL query or mutation.

It could be worth thumbing through the [`fxa-settings` docs](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings#gql-error-handling) or Apollo docs on error handling for the `onError` syntax, but **keep in mind** we're using Apollo Server in fxa-settings and the Apollo docs assume use of Apollo Server, yet we're using graphene in Nimbus.

The errors returned by the server are `field` or `messages`; reference the GQL playground to see these in action. Figure out what we need to send off to Sentry and set up a convention for how we'll handle `onError` in GQL calls. In fxa-settings, we have a [custom useMutation hook](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/hooks.tsx#L68) that differentiates between error types and reports to sentry if it's a certain type, but there are no error types returned from graphene. Maybe we want to just report them all, but we probably don't want to report the same errors the server is reporting and we'll at least want to report if the server can't be reached. We'll need a small hook so we don't have to import Sentry in every component.

Assuming EXP-299 is completed prior to starting this issue, add the error handling there. Add a section to the `nimbus-ui` README about GQL error handling and what was done here.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-473)
┆Issue Number: EXP-473

jodyheavener:
This is mostly just documentation on how we should approach errors in the Nimbus app. I originally started with the idea that we should create custom `useMutation`/`useQuery`/etc functions that handled all possible error paths for us, but eventually found this approach to be messy and possibly unreliable, so instead I returned to the idea of "you're responsible for implementation, but here are some guidelines".

As part of the standardizing I did a small update to the exist New Experiment form to display generic errors instead of technical ones.

Additionally, while I was in hook-land, I added an `index.ts` to that directory so you don't need to import hooks from each individual hook file.

### DIFF
--- a/app/experimenter/nimbus-ui/README.md
+++ b/app/experimenter/nimbus-ui/README.md
@@ -75,6 +75,106 @@ export interface GetExperimentOverviews {
 
 **Note**: this command performs introspection on the GraphQL endpoint, so your local server must be running for it to work.
 
+## Error handling
+
+This app has an [`AppErrorBoundary`](./src/components/AppErrorBoundary/index.tsx) that will capture any uncaught error that occurs and report them to Sentry. This acts as our last line of defense, but ideally we are able to handle errors before they get to this stage. As well, errors that we wish to display to the user should generally be handled in a consistent fashion.
+
+There are two types of errors to account for:
+
+### Validation errors
+
+These occur when you try to create or modify a record and the server tells us that one or more fields provided incorrect data. For these you should use React Bootstrap's [`Form.Control.Feedback`](https://react-bootstrap.github.io/components/forms/#form-control-feedback-props) component in combination with other form components.
+
+A basic example might look like this:
+
+```tsx
+<Form.Group controlId="description">
+  <Form.Label>Description</Form.Label>
+  <Form.Control type="text" autoFocus />
+  <Form.Text>This is the description.</Form.Text>
+  {errors["description"] && (
+    <Form.Control.Feedback type="invalid">
+      {errors["description"].message}
+    </Form.Control.Feedback>
+  )}
+</Form.Group>
+```
+
+### General errors
+
+General errors can occur when an operation fails altogether. This could be for any reason, such as when a network request fails, or a bad code path raises an exception. For these errors you'll use React Bootstrap's [`Alert`](https://react-bootstrap.github.io/components/alerts/) components.
+
+For example, if error occurs while saving a record:
+
+```tsx
+<Alert variant="warning">Sorry, there was a problem.</Alert>
+```
+
+_Note:_ displaying technical error messages to the user may not be very helpful, so it's generally advised to either provide a [generic message](./src/lib/constants.ts) or instructions on what to do.
+
+## Handling GraphQL operation errors
+
+As GQL queries and mutations are how we get data in and out of the app this is where a majority of our errors are going to occur. These should be handled in a consistent fashion as well.
+
+Both operations can result in a general error, such as a `NetworkError`, or GraphQL errors, which in most cases will be validation errors. Importantly, these two types of errors occur in different ways; general errors will throw or trigger the operations `onError` function, and GraphQL errors will be present in the result of the operation.
+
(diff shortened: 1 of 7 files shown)


app/experimenter/nimbus-ui/src/lib/constants.ts
```diff
@@ -3,3 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 export const BASE_PATH = "/experiments/nimbus";
+
+export const UNKNOWN_ERROR =
+  "Sorry, an error occurred but we don't know why. We're looking into it.";
+export const SUBMIT_ERROR =
+  "Sorry, on error occurred while submitting. Please try again.";
```
LZoog, line 10: ~on~ an

app/experimenter/nimbus-ui/README.md
```diff
@@ -75,6 +75,106 @@ export interface GetExperimentOverviews {
...
+
+General errors can occur when an operation fails altogether. This could be for any reason, such as when a network request fails, or a bad code path raises an exception. For these errors you'll use React Bootstrap's [`Alert`](https://react-bootstrap.github.io/components/alerts/) components.
+
+For example, if error occurs while saving a record:
+
+```tsx
+<Alert variant="warning">Sorry, there was a problem.</Alert>
+```
+
+_Note:_ displaying technical error messages to the user may not be very helpful, so it's generally advised to either provide a [generic message](./src/lib/constants.ts) or instructions on what to do.
```
LZoog, line 113: We do return `<Alert>` in `AppErrorBoundary`; is that the only place we might need this? I'm not sure when else a "general" error might come up.

app/experimenter/nimbus-ui/README.md
```diff
@@ -75,6 +75,106 @@ export interface GetExperimentOverviews {
...
+
+An example of handling a `useQuery` operation:
+
+```tsx
+// Set up the query, and because it executes immediately, handle it
+// immediately by passing in an `onError` option. This is the same as
+// wrapping it in a try/catch, just a little more grouped together.
+const { data } = useQuery(GET_RECORD_QUERY, {
+  onError(error) {
+    console.log("A general error occurred", error.message);
```
LZoog, line 163: Did we want a component or something here to display errors with? Or maybe we'll just display at the top of the page somewhere.

LZoog (approved):
LGTM! Pointed out one typo and have a couple clarification questions but don't think those will cause much/any change so r+, thanks for the nice docs.

app/experimenter/nimbus-ui/README.md
jodyheavener: In an effort not to overkill this section I'm just demonstrating that "this is where you need to do something", and the `<Alert>` component is demonstrated in the section above, which is what you would need to use for something like this, but would involve setting up state and some other stuff to demonstrate here.

app/experimenter/nimbus-ui/README.md
jodyheavener: A general error could occur during a network request, such as `NetworkError`. Noted a little further up:

> This could be for any reason, such as when a network request fails, or a bad code path raises an exception.

LZoog: I suppose I'm wondering where else/when we'd use `<Alert>` in the code. Doesn't it bubble up to `AppErrorBoundary`?

app/experimenter/nimbus-ui/README.md
LZoog: Ohh I see, maybe ignore my last comment then, _this_ is where we'd use `<Alert>`

PR #3829, opened 2020-10-27
fix #3640 feat(nimbus): add remote settings tasks for nimbus

### LINKED ISSUE #3640: Add RS publishing tasks for NimbusExperiment
We need to add RS celery tasks for publishing NimbusExperiments, we can just duplicate the rapid publishing tasks we have now and change to the different model and use the relevant collections.

The existing tasks are here:

[https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py](https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py)

These tasks are responsible for:

* Checking for experiments in the review state that are ready to push to RS
* Checking if there is already an experiment waiting to be reviewed in RS
* Checking if there is already an experiment that's been rejected in RS, and if so, updating the experiment with the relevant rejection info
* Pushing an experiment to RS if there's nothing in review
* Checking if any experiments are live in RS and updating them in the db
* Checking if an experiment has been manualy marked for completion and remove from collection
* Checking if any experiments are complete (live in the db but don't appear in RS at all) and updating them in the db

We can use these tasks as written for Nimbus, we just need to change the references from Experiment to NimbusExperiment (and update any specific fields/methods as necessary).  Rather than update the existing ones, we should create a duplicate set that are specifically for the NimbusExperiment, that way we can preserve the functionality of the existing RapidExperiment prototype until the new NimbusExperiment workflow is fully functional.

We'll also need to duplicate the relevant tests.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-439)
┆Issue Number: EXP-439

jaredlockhart:
Because

* We need to send and receive nimbus experiments from a specific remote settings collection

This commit

* Adds tasks to push and read nimbus experiments from the nimbus specific collection
  - Push to collection and mark for review with a changelog
  - Read rejected experiments and mark them as draft with a changelog
  - Check for reviewed experiments that have been accepted and mark them live with a changelog
  - Check for removed experiments and mark them completed with a changelog

### DIFF
--- a/app/bin/setup_kinto.py
+++ b/app/bin/setup_kinto.py
@@ -8,13 +8,17 @@ ADMIN_USER = ADMIN_PASS = "admin"
 REVIEW_USER = REVIEW_PASS = "review"
 EXPERIMENTER_USER = os.environ["KINTO_USER"]
 EXPERIMENTER_PASS = os.environ["KINTO_PASS"]
+KINTO_HOST = os.environ["KINTO_HOST"]
+KINTO_BUCKET = os.environ["KINTO_BUCKET"]
+KINTO_COLLECTION_LEGACY = (os.environ["KINTO_COLLECTION"],)
+KINTO_COLLECTION_NIMBUS = os.environ["KINTO_COLLECTION_NIMBUS"]
 
 
 def create_user(user, passw):
     print(f">>>> Creating kinto user: {user}:{passw}")
     print(
         requests.put(
-            urllib.parse.urljoin(os.environ["KINTO_HOST"], f"/accounts/{user}"),
+            urllib.parse.urljoin(KINTO_HOST, f"/accounts/{user}"),
             json={"data": {"password": passw}},
         ).content,
     )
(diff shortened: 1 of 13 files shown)


app/bin/setup_kinto.py
```diff
@@ -31,51 +32,52 @@ def create_user(user, passw):
...
 
 
-print(">>>> Creating kinto group: editors")
-print(
-    client.create_group(
-        id=f"{os.environ['KINTO_COLLECTION']}-editors",
-        bucket=os.environ["KINTO_BUCKET"],
-        data={"members": [f"account:{os.environ['KINTO_USER']}"]},
-        if_not_exists=True,
+for collection in os.environ["KINTO_COLLECTION"], os.environ["KINTO_COLLECTION_NIMBUS"]:
```
bbangert, line 42: How come these aren't pulled out at the top? Usually when I write CLI type code like this in Python I use the `click` package, so that there's useful help for the script as well as argument checking up front.

app/bin/setup_kinto.py
```diff
@@ -31,51 +32,52 @@ def create_user(user, passw):
...
+    print(f">>>> Creating kinto collection: {collection}")
+    print(
+        client.create_collection(
+            id=collection,
+            bucket=KINTO_BUCKET,
+            permissions={
+                "read": ["system.Everyone"],
+                "write": [
+                    (
+                        f"/buckets/{os.environ['KINTO_BUCKET']}/groups/"
```
bbangert, line 72: It's odd to go back/forth between using the `KINTO_BUCKET` var and the `os.environ` access to it.

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+def get_kinto_user():
+    user, _ = get_user_model().objects.get_or_create(
+        email=settings.KINTO_DEFAULT_CHANGELOG_USER,
+        username=settings.KINTO_DEFAULT_CHANGELOG_USER,
+    )
+    return user
+
+
+@app.task
+@metrics.timer_decorator("push_experiment_to_kinto.timing")
```
bbangert, line 29: It would probably be useful to differentiate Nimbus vs. legacy metrics.

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+def nimbus_push_experiment_to_kinto(experiment_id):
+    kinto_client = KintoClient(settings.KINTO_COLLECTION_NIMBUS)
+
+    metrics.incr("push_experiment_to_kinto.started")
```
bbangert, line 33: Or if this metrics lib supports it, tags could be used to differentiate legacy vs. nimbus experiment stats.

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+
+    experiment = NimbusExperiment.objects.get(id=experiment_id)
```
bbangert, line 35: Does it make sense to double-check here that the status is suitable for being pushed to kinto? Or are we sufficiently confident that could never happen?

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+    if not NimbusBucketRange.objects.filter(experiment=experiment).exists():
+        NimbusIsolationGroup.request_isolation_group_buckets(
+            experiment.slug,
+            experiment,
+            NimbusExperiment.BUCKET_COUNT,
+        )
+
+    data = NimbusExperimentSerializer(experiment).data
```
bbangert, line 43: This could never fail to serialize? Or is the default error handling sufficient for this case?

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+        experiment.status = NimbusExperiment.Status.DRAFT
+        experiment.save()
+
+        generate_nimbus_changelog(
+            experiment,
+            get_kinto_user(),
+            message=f'Rejected: {rejected_collection_data["last_reviewer_comment"]}',
+        )
+
+        kinto_client.delete_rejected_record(rejected_slug)
```
bbangert, line 87: It's only possible during rejection for a single record to be rejected at a time? There can't be 3 rejected nimbus experiments?

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+@app.task
+@metrics.timer_decorator("check_kinto_push_queue")
+def nimbus_check_kinto_push_queue():
+    kinto_client = KintoClient(settings.KINTO_COLLECTION_NIMBUS)
+
+    metrics.incr("check_kinto_push_queue.started")
+
+    queued_experiments = NimbusExperiment.objects.filter(
+        status=NimbusExperiment.Status.REVIEW
+    )
```
bbangert, line 72: This portion of the code makes it appear that this function is going to be working with multiple queued experiments. However the rejected code and the portion below where it actually pushes an experiment, only works with a single experiment at a time. This function would greatly benefit from a detailed doc string that indicates the desired method of operation of this task and reasoning behind it.

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+    metrics.incr("check_experiment_is_live.started")
+
+    accepted_experiments = NimbusExperiment.objects.filter(
+        status=NimbusExperiment.Status.ACCEPTED
+    )
+
+    records = kinto_client.get_main_records()
+    record_ids = [r.get("id") for r in records]
+
+    for experiment in accepted_experiments:
```
bbangert, line 117: The task above publishes individually, this one says its to `check_experiment` yet its checking all accepted experiments. I would expect the task to be something like `nimbus_check_experiments_are_live`.

app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -0,0 +1,164 @@
...
+    metrics.incr("check_experiment_is_complete.started")
+
+    live_experiments = NimbusExperiment.objects.filter(
+        status=NimbusExperiment.Status.LIVE
+    )
+
+    records = kinto_client.get_main_records()
+    record_ids = [r.get("id") for r in records]
+
+    for experiment in live_experiments:
```
bbangert, line 149: Same as above regarding naming. In the absence of module docs, function doc strings, etc. the desired functioning of this process is not documented anywhere and must be inferred from the code and naming, which isn't clear here either.

bbangert (changes requested):

app/bin/setup_kinto.py
jaredlockhart: Oh good catch I missed that one 👍

app/experimenter/kinto/tasks/nimbus.py
jaredlockhart: yes good catch!

jaredlockhart: Can just namespace it here: `metrics = markus.get_metrics("kinto.nimbus_tasks")`

app/experimenter/kinto/tasks/nimbus.py
jaredlockhart: Yeah that logic lives in the other task, this one should just push whatever it's given.

app/experimenter/kinto/tasks/nimbus.py
jaredlockhart: Yeah if we wind up in a situation where an experiment makes it here and completely fails to serialize then we should just blow up, abort, and report the error.

app/experimenter/kinto/tasks/nimbus.py
jaredlockhart: Yeah exactly, a limitation on remote settings is that only a single record can be in review at a time, and there's only a single instance of the task that communicates with remote settings, so it'll always hit this first, unpack the rejected experiment before it pushes another one for review.

bbangert: Ok, so to make sure I understand this right.... Kinto only allows a single record to be in review, so that's why this code only pushes a single record at a time? Does that mean that once a person requests a review, no more records can be pushed until its been reviewed?

jaredlockhart: Yes exactly

bbangert: Great, excellent doc string.

bbangert (approved):
Nice!

PR #3839, opened 2020-10-28
fix #3835 feat(nimbus): set start and end dates in V6 API

### LINKED ISSUE #3835: Add start/end date to V6 API
Now that we have changelogs and background tasks we can determine the start and end dates by looking up the relevant changelogs.

Start: ACCEPTED -> LIVE

End: LIVE -> COMPLETE

Let's add convenience methods to NimbusExperiment to look up those dates and then expose them through the V6 API.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-525)
┆Issue Number: EXP-525

jaredlockhart:
Because

* V6 API consumers need to know the start and end date of experiments for automated analysis

This commit

* Adds model methods for start and end date based on the relevant status transition changelogs
* Exposes the start and end date in the V6 API as ISO8601 strings


app/experimenter/docs/swagger-ui.html
```diff
@@ -3551,11 +3551,11 @@
           },
           "startDate": {
             "type": "string",
-            "readOnly": true
+            "format": "date-time"
```
bbangert, line 3554: Are these no longer readOnly then?

jaredlockhart: The whole V6 API is read only, it only responds to GET requests:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v6/views.py#L11-L12

bbangert (approved):

Issue #3696, opened 2020-10-08
`PageNew` modifications - back links, add "Application" field, finish styles


data-sync-user:
Figma: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216066%3A11123

On `PageExperimentNew`: 
1) Add the “X” button and cancel buttons, both of which should be links taking the user from `/nimbus/new/` back to `/nimbus/ (the directory view)
2) Add the “Application” drop-down field which roughly maps to [https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L48](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L48). This is not currently in the Experimenter platform (`rapid` or `core`) in the front-end but was implemented in the BE in https://github.com/mozilla/experimenter/commit/cdf90494a7861d72ef400121e7281e2b8b2d4479.
3) Make any other text or styling adjustments to closely resemble the Figma comp

Note: much of this was already completed in another ticket so it's been reduced from a 5 to a 3.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-462)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/14367/screenshot-1.png">screenshot-1.png</a> | <a href="https://jira.mozilla.com/secure/attachment/14420/Screen+Shot+2020-10-29+at+5.55.59+PM.png">Screen Shot 2020-10-29 at 5.55.59 PM.png</a>
┆Issue Number: EXP-462

data-sync-user:
➤ Jared Kerim commented:

We need to make sure that field specific errors are reflected on the form.  For instance here, `application` has an invalid value, but the application field is green/valid:

data-sync-user:
➤ Lauren Zugai commented:

We also need to make sure the drop-down is populated with label values that should come from the `useConfig` hook. Attached a screenshot of what this looks like in the `rapid` project.  !screenshot-1.png|thumbnail!

jodyheavener:
> We need to make sure that field specific errors are reflected on the form.  For instance here, application has an invalid value, but the application field is green/valid

[...]

PR #3843, opened 2020-10-29
fixes #3697 feat(nimbus): send overview mutation when form submitted

### LINKED ISSUE #3697: `PageEditOverview` "Save" experiment edits (GQL mutation) + "Next"
1) Add a "Save" and "Next" button at the bottom of this page.
2) On "Save" click, combine any changed fields with data from EXP-302 to send the GQL mutation up to the server to save the experiment.
3) Keep track of if the user makes any changes on a field to:
3a) Activate the "Save" button (default state should be disabled)
3b) disable the "Next" button (default state should be enabled)
3c) Use the available hook to alert the user if they attempt to navigate away from the page via the sidebar with unsaved changes
4) "Next" should take the user to the `/branches` page

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-463)
┆Issue Number: EXP-463

jodyheavener:
This PR:

- Adds the mutation to update an experiment
  - This updated generated types
- Updates the shared FormOverview to enable Save and Next based on the status of the fields
  - Save will only be enabled if there are changes
  - Next will only be enabled if there are no changes (initial state and after save)
- Updates the shared FormOverview to use config-based Application dropdown (eliminating it from #3696)
- Because why not, I can remove this if desired: enables the Cancel button in the New Experiment form (eliminating it from #3696)
- Warns you on the New and Edit pages if you make changes and attempt to leave the page
- Moves some test mutation mocking over to the shared mocks file
- Adds missing Storybook for FormOverview with experiment mock data
- Updates tests

Some of these tests were a little clunky to get to 100%, so feel free to shame me if I've done anything super unclear.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
@@ -3,47 +3,34 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from "react";
-import {
-  render,
-  screen,
-  act,
-  fireEvent,
-  cleanup,
-} from "@testing-library/react";
-import FormOverview from ".";
+import { render, screen, act, fireEvent } from "@testing-library/react";
 import { mockExperimentQuery } from "../../lib/mocks";
-import { getExperiment } from "../../types/getExperiment";
+import { Subject } from "./mocks";
 
 describe("FormOverview", () => {
-  afterEach(() => {
-    cleanup();
-  });
-
   it("renders as expected", async () => {
     render(<Subject />);
-    await act(async () => {
-      expect(screen.getByTestId("FormOverview")).toBeInTheDocument();
-    });
+    await act(async () =>
+      expect(screen.getByTestId("FormOverview")).toBeInTheDocument(),
+    );
   });
 
   it("calls onCancel when cancel clicked", async () => {
     const onCancel = jest.fn();
     render(<Subject {...{ onCancel }} />);
-    await act(async () => {
-      const cancelButton = screen.getByText("Cancel");
-      await fireEvent.click(cancelButton);
-      expect(onCancel).toHaveBeenCalled();
-    });
+
+    const cancelButton = screen.getByText("Cancel");
+    await act(async () => void fireEvent.click(cancelButton));
+    expect(onCancel).toHaveBeenCalled();
   });
 
   it("calls onNext when next clicked", async () => {
     const onNext = jest.fn();
     render(<Subject {...{ onNext }} />);
-    await act(async () => {
-      const nextButton = screen.getByText("Next");
-      await fireEvent.click(nextButton);
-      expect(onNext).toHaveBeenCalled();
-    });
+
+    const nextButton = screen.getByText("Next");
+    await act(async () => void fireEvent.click(nextButton));
+    expect(onNext).toHaveBeenCalled();
   });
 
   const fillOutNewForm = async (expected: Record<string, string>) => {
@@ -55,12 +42,18 @@ describe("FormOverview", () => {
       const fieldName = screen.getByLabelText(labelText);
       expect(fieldName).not.toHaveClass("is-invalid");
       expect(fieldName).not.toHaveClass("is-valid");
-      await fireEvent.click(fieldName);
-      await fireEvent.blur(fieldName);
+
+      await act(async () => {
+        fireEvent.click(fieldName);
+        fireEvent.blur(fieldName);
+      });
       expect(fieldName).toHaveClass("is-invalid");
       expect(fieldName).not.toHaveClass("is-valid");
-      await fireEvent.change(fieldName, { target: { value: fieldValue } });
-      await fireEvent.blur(fieldName);
+
+      await act(async () => {
+        fireEvent.change(fieldName, { target: { value: fieldValue } });
+        fireEvent.blur(fieldName);
+      });
       expect(fieldName).not.toHaveClass("is-invalid");
       expect(fieldName).toHaveClass("is-valid");
     }
@@ -71,7 +64,7 @@ describe("FormOverview", () => {
       ["Public name", expected.name],
       ["Hypothesis", expected.hypothesis],
       ["Application", expected.application],
-      ["Public Description", expected.publicDescription],
+      ["Public description", expected.publicDescription],
     ]) {
       const fieldName = screen.getByLabelText(labelText) as HTMLInputElement;
       expect(fieldName.value).toEqual(fieldValue);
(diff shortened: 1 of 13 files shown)


app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
```diff
@@ -206,26 +210,27 @@ describe("FormOverview", () => {
...
     await act(async () => {
-      const errorFeedback = screen.getByText(submitErrors["name"]);
       expect(errorFeedback).toHaveClass("invalid-feedback");
       expect(errorFeedback).toHaveAttribute("data-for", "name");
     });
   });
 });
 
+// TODO: EXP-462 Get this from constants / config loaded at app start?
 const APPLICATIONS = ["firefox-desktop", "fenix", "reference-browser"];
```
lmorchard, line 234: Not going to get the runtime app config in a test and constants won't do - this should be mocked here

jodyheavener: Good point, with #3851 now merged the mocked cache now has a mock config loaded that should be useful.

jodyheavener:
@lmorchard While this is mostly still reviewable I'm going to hold until #3849 is merged and then update to reflect the per-compoment mock structure you have set up.

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -31,6 +31,12 @@ const FormOverview = ({
     mode: "onTouched",
   });
   const { isSubmitted, isValid, isDirty, touched } = formState;
+  const shouldWarnOnExit = useExitWarning();
```
LZoog, line 34: Hmm, I'm not sure this work as expected. I can make an edit on the overview page and click on "Branches" to navigate away after I've made an edit without an alert showing.

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
```diff
@@ -3,35 +3,74 @@
...
+          throw new Error("Save failed, no error available");
+        }
+
+        const { message } = result.data.updateExperimentOverview;
+
+        if (message !== "success" && typeof message === "object") {
+          return void setSubmitErrors(message);
+        }
+
+        resetForm();
```
LZoog, line 61: I think we want to remove L61 or rework it somehow. When I make an edit and click "Save," the old value is repopulated in the field. If we just remove it though, the "Save" button doesn't disable with "Next" re-enabled.

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
```diff
@@ -3,35 +3,74 @@
...
-    {},
+  const { experiment, notFound, loading: experimentLoading } = useExperiment(
+    slug,
   );
+  const [submitErrors, setSubmitErrors] = useState<Record<string, any>>({});
 
-  const onFormSubmit = useCallback(() => {
-    console.log("SUBMIT TBD");
-  }, []);
+  const onFormSubmit = useCallback(
```
LZoog, line 33: We need a mock in Storybook for this not to error, though it would be also be nice to have a state besides "basic" to show an error state.

image

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
jodyheavener: That's fair, this appears to be an issue with how it works with React Router (maybe it doesn't send the `beforeunload` event). It works reliably with true browser navigation for me, though. We'll probably want to break this out into a bug issue.

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
jodyheavener: Sure I can add that. FWIW I'm on the fence about this, because `FormOverview` does have its own stories (including with existing experiment data and error states), so we're really looking to demonstrate the page layout and not necessarily the form. But I can see how for completeness you should be able to successfully hit Save. Maybe we don't need all the error states though?

LZoog: Yeah, I feel like things under `Pages` should work somewhat well like clicking "Save" either sends a successful mutation or simply links to `PageEditOverview` (the latter probably being easier). It'd be nice for one of the pages to show what the error state looks like, anything extra is probably bonus points 🤷

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
LZoog: WFM to do in a follow up. We can ask Ana how imperative it is that they can't navigate through the sidebar, we may want to add it to the "create/edit flow" epic and if what you've got is mostly sufficient we can file an issue in the "UX nice-to-haves".

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
jodyheavener: Fixed

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
jodyheavener: Issue filed: https://jira.mozilla.com/browse/EXP-545

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
jodyheavener: To recap our Slack discussion, it doesn't appear that there's a great approach here without pulling things apart a bit.

> clicking "Save" either sends a successful mutation

While it's possible to create a mock for this, the mutation input needs to match the form input, otherwise it won't get picked up. So a mock would immediately become invalidated once a user starts typing in any of the fields.

> like on the PageNew page, simply links to PageEditOverview

The Save button, by way of `FormOverview`, actually already links to this component! But I think Storybook intercepts this as already being on that story, so it does nothing and still results in the error being shown.

There might be other approaches that I'm not thinking of at the moment, but it sounds like we might want to follow up on this to avoid blocking? Is that correct?

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
```diff
@@ -2,43 +2,85 @@
...
-            applications,
-            experiment,
-            submitErrors,
-            onSubmit: onFormSubmit,
-            onNext: onFormNext,
-          }}
-        />
-      )}
+      {({ experiment }) => {
+        currentExperiment.current = experiment;
```
lmorchard, line 70: Hmm, feeling a little leery of this bit of ref-based state management. But, I can't think of any clear problems this would cause.

lmorchard (approved):
Looks like a couple of follow-ups to pursue, but pretty much LGTM unless lzoog has further comments

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
jodyheavener: I only did this so I could define `onFormSubmit` outside of the render; but happy remove this and move it back in?

app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
```diff
@@ -82,111 +75,107 @@ describe("FormOverview", () => {
...
+    await act(async () => fillOutNewForm(expected));
+    await act(async () => void fireEvent.click(submitButton));
 
     expect(onSubmit).toHaveBeenCalled();
     expect(onSubmit.mock.calls[0][0]).toEqual(expected);
   });
 
-  it("with missing public description, does not allow submit", async () => {
-    const { data } = mockExperimentQuery("boo", { publicDescription: "" });
+  it("with existing experiment data, asserts field values before allowing submit and next", async () => {
```
LZoog, line 95: Ohh, you know I think we actually want to allow saving pieces of an experiment without filling out required fields. I think a user should be able to leave "Public Description" blank if they wanted to only update say, the hypothesis, and save that, to update the public description later.

However, we could double check this with Ana and maybe do this in a follow up as I'm not _entirely_ certain of this.

LZoog (approved):
LGTM as well, might need to double check with Ana on intended UX for one point but nothing blocking👍

app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
```diff
@@ -55,12 +53,18 @@ describe("FormOverview", () => {
       const fieldName = screen.getByLabelText(labelText);
       expect(fieldName).not.toHaveClass("is-invalid");
       expect(fieldName).not.toHaveClass("is-valid");
-      await fireEvent.click(fieldName);
-      await fireEvent.blur(fieldName);
+
+      await act(async () => {
+        fireEvent.click(fieldName);
+        fireEvent.blur(fieldName);
+      });
```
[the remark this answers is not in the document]
jodyheavener: @LZoog pointed out to me that VS Code will helpfully place a lil dot underneath `await`s that do not need it!

app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
jodyheavener: Updated to make optional

app/experimenter/nimbus-ui/src/components/PageEditOverview/index.tsx
lmorchard: Oof, speaking of the PageNew story: I just noticed that it looks like this commit broke it.

* Before: https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/265a366028dce040318162dd5fa0459c45e41e69/nimbus-ui/index.html?path=/story/pages-new--basic

* After: https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/5797b0f1eaeb53adb68951e40cf0f2715cc0fe8f/nimbus-ui/index.html?path=/story/pages-new--basic

PR #3848, opened 2020-10-29
fix #3844 chore(nimbus): add eslint/prettier

### LINKED ISSUE #3844: Add lint fix to nimbus UI npm scripts
We should have a canonical way to format the code in the nimbus UI code that uses eslint --fix and expose it in the makefile so it's triggered by `make code_format` and validated using `make check`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-527)
┆Issue Number: EXP-527

jaredlockhart:
Because

* We want the code to be auto formatted in a consistent way to prevent noisy PRs from individual editor formatting settings

This commit

* Adds eslint/prettier and attaches it to the make code_format and make check commands


app/experimenter/nimbus-ui/src/components/AppErrorBoundary/index.tsx
```diff
@@ -26,11 +26,11 @@ class AppErrorBoundary extends React.Component {
 
   render() {
     const { error } = this.state;
-    return error ? <AppErrorAlert {...{ error }} /> : this.props.children;
+    return error ? <AppErrorAlert /> : this.props.children;
   }
 }
 
-export const AppErrorAlert = ({ error }: { error: Error }) => {
```
lmorchard, line 33: Hmm... looks like this got snipped out because we're not actually displaying anything related to the specific error. Do we want this removed - or do we actually want to do something in the alert with the error?

jaredlockhart: Exactly, `no-unused-vars`

jaredlockhart: I dunno, we could display the error?

lmorchard, line 33: Hmm... looks like this got snipped out because we're not actually displaying anything related to the specific error. Do we want this removed - or do we actually want to do something in the alert with the error?

app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.test.tsx
```diff
@@ -27,7 +27,7 @@ describe("PageNew", () => {
   });
 
   describe("navigation links", () => {
-    const SidebarRoot = (props: RouteComponentProps) => (
+    const SidebarRoot: React.FunctionComponent<RouteComponentProps> = () => (
```
lmorchard, line 30: Probably just a bikesheddy nit, but I haven't actually seen React.FunctionComponent type used before. Seems more verbose than what it's replacing and inconsistent with other functional components.

Guess it gets used when we don't actually use the `props` parameter directly? Seems like that's the case in this mock component and in most page components that use hooks for context.

lmorchard (approved):
Scratched my head at this and could only come up with fussy nits probably not worth much thought

app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.test.tsx
jaredlockhart: Yeah exactly, `props` isn't being used, so `no-unused-vars` hits it, but if you take it out entirely, typescript can't infer the type when `path` gets passed in at the call site, so typescript needs some way to know that that's a valid parameter, this is how you annotate the component with its props even if they aren't being used.

app/experimenter/nimbus-ui/src/components/AppErrorBoundary/index.tsx
jaredlockhart: Oh why did github throw out my comment.  Sorry anyways, yes the error isn't used so it gets caught by `no-unused-vars`.  We could display it? I tried not to change the behaviour of any components.

app/experimenter/nimbus-ui/package.json
```diff
@@ -91,4 +95,4 @@
       "last 1 safari version"
     ]
   }
-}
+}
```
jodyheavener, line 98: newline newline newline!

app/experimenter/nimbus-ui/.eslintrc.js
```diff
@@ -0,0 +1,40 @@
...
+        '@typescript-eslint/no-empty-function': 'off',
+        'react/display-name': 'off',
+        'react/prop-types': 'off',
+    },
+    settings: {
+        react: {
+            version: "detect"
+        }
+    }
+};
```
jodyheavener, line 40: Will React also use this config during dev compilation and prod build?

jodyheavener (approved):
Just one question, but lgtm.

app/experimenter/nimbus-ui/.eslintrc.js
jodyheavener: Oh also, does this pull in the  `app/.prettierrc`?

jaredlockhart: > Oh also, does this pull in the `app/.prettierrc`?

Oh we can probably get rid of that that's probably very old.  Formatting configs for js should live in their relevant folders, this probably isn't being used anymore.

app/experimenter/nimbus-ui/package.json
jaredlockhart: > newline newline newline!

Aye aye!

app/experimenter/nimbus-ui/.eslintrc.js
jodyheavener: 😆 This is where I've been pointing my VS Code at. I would love it if we had a single ESLint and Prettier config instead of unique ones per-project.

jaredlockhart: That would be nice but probably way overkill considering rapid will be deleted as soon as we hit parity and legacy is in maintenance mode.  I don't think you need to worry about anything other than `nimbus-ui`

jaredlockhart: Oh no I was wrong, it does use .prettierrc for everything so we can just leave it there and treat it as a global formatting config, as you said 👍

jaredlockhart: I don't think we can have a single eslint though since everything uses different versions of react and not all of them are typescript.

jaredlockhart: > Will React also use this config during dev compilation and prod build?

Looks like everything is building and hosting correctly to me

Issue #3469, opened 2020-09-15
Add `PageEditBranches` dummy form


data-sync-user:
Copy over the `BranchesForm` and tests in `legacy-ui/rapid/components/forms/ExperimentForm` into the `PageEditBranches` page. Don't worry about state updates (anything in `dispatch`), those will be handled in a follow up. Anything to match Figma is a bonus here, we just need to reach parity.

Note: you can likely do EXP-301 at the same time, when these tickets were filed, 301 also involved reading from the cache.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-300)
┆Issue Number: EXP-300

PR #3851, opened 2020-10-30
fixes #3855 feat(nimbus): read config from graphql server

### LINKED ISSUE #3855: Consume GraphQL Configuration
It's possible this is part of another issue, but I couldn't find one specifically for setting up GQL config usage ([this](https://jira.mozilla.com/browse/EXP-283) exists but is not the same).

Basically, query the `nimbusConfig` on app start, cache it, and make it available throughout the app.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-534)
┆Issue Number: EXP-534

jodyheavener:
I took a look but didn't see an issue this relates to, but I think we're going to need this pretty soon, so here it is.

This PR adds loading in the GraphQL Nimbus Config on app start and makes it available throughout the app via the `useConfig` hook.

### DIFF
--- a/app/experimenter/nimbus-ui/src/hooks/useConfig.test.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useConfig.test.tsx
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { useConfig } from "./useConfig";
+import { MockedCache, MOCK_CONFIG } from "../lib/mocks";
+import { render, waitFor } from "@testing-library/react";
+import serverConfig from "../services/config";
+
+describe("hooks/useConfig", () => {
+  describe("useConfig", () => {
+    let hook: ReturnType<typeof useConfig>;
+    const config = { ...MOCK_CONFIG, ...serverConfig };
+
+    const TestConfig = () => {
+      hook = useConfig();
+      return <p>Bière forte</p>;
+    };
+
+    it("returns the config", async () => {
+      render(
+        <MockedCache>
+          <TestConfig />
+        </MockedCache>,
+      );
+
+      await waitFor(() => expect(hook).toEqual(config));
+    });
+  });
+});
--- a/app/experimenter/nimbus-ui/src/hooks/useConfig.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useConfig.tsx
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { useApolloClient } from "@apollo/client";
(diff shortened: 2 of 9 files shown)


app/experimenter/nimbus-ui/src/hooks/useConfig.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import serverConfig from "../services/config";
+
+describe("hooks/useConfig", () => {
+  describe("useConfig", () => {
+    let hook: ReturnType<typeof useConfig>;
+    const config = { ...MOCK_CONFIG, ...serverConfig };
+
+    const TestConfig = () => {
+      hook = useConfig();
+      return <p>Bière forte</p>;
```
jaredlockhart, line 18: Magnifique! 🍻

app/experimenter/nimbus-ui/src/types/getConfig.ts
```diff
@@ -0,0 +1,83 @@
+/* tslint:disable */
```
jaredlockhart, line 1: So neat that it auto generates these types for us, I wanted to do this with the rest API and its auto generated openjson schema.

app/experimenter/nimbus-ui/src/types/globalTypes.ts
```diff
@@ -7,13 +7,10 @@
 // START Enums and Input Objects
```
jaredlockhart, line 7: Why are these constants defined here?  These should only be fetched from the graphql API?

jaredlockhart:
How did we miss filing a ticket for this on the client?  I would file one and add it to the sprint and add its github number here so it's tracked.

app/experimenter/nimbus-ui/src/types/globalTypes.ts
lmorchard: These are a part of the auto-generated types, they're enums pulled from the GQL API schema by the code generator

app/experimenter/nimbus-ui/src/hooks/useConfig.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * NOTE: This hook is dependent on the initial query to retrieve
+ * GraphQL config being performed inside the <App> component. Do
+ * not use before this component is rendered.
+ *
+ * Example:
+ *
+ * const { channels, probeSets } = useConfig();
+ */
+
+export function useConfig() {
```
lmorchard, line 22: Maybe a non-thing, but: Could we rename this to something like `useServerConfig` or `useGQLConfig`? We already have [config from the meta header](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/services/config.ts), so there's a bit of a name collision there.

app/experimenter/nimbus-ui/src/types/globalTypes.ts
jodyheavener: https://jira.mozilla.com/browse/EXP-536

lmorchard (approved):
I'd say this looks good and will come in handy for me in PR #3469 

Should probably file a follow-up to suss out including type generation checks in CI

app/experimenter/nimbus-ui/src/types/globalTypes.ts
jaredlockhart: Thanks @jodyheavener !

PR #3860, opened 2020-10-31
feat(nimbus): move common experiment edit elements to PageEditContainer component

jodyheavener:
All this does is DRY up some of the Experiment edit pages by moving the Experiment slug query, loading state, and not found state to its own component.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
@@ -4,16 +4,14 @@
 
 import React from "react";
 import { RouteComponentProps } from "@reach/router";
-import AppLayoutWithSidebar from "../AppLayoutWithSidebar";
+import PageEditContainer from "../PageEditContainer";
 
 type PageRequestReviewProps = {} & RouteComponentProps;
 
 const PageRequestReview: React.FunctionComponent<PageRequestReviewProps> = () => (
-  <AppLayoutWithSidebar>
-    <section data-testid="PageRequestReview">
-      <h1>PageRequestReview</h1>
-    </section>
-  </AppLayoutWithSidebar>
+  <PageEditContainer title="Review & Launch" testId="PageRequestReview">
+    {({ experiment }) => <p>{experiment.name}</p>}
+  </PageEditContainer>
 );
 
 export default PageRequestReview;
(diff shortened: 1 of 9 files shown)


app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -4,16 +4,14 @@
...
 
 type PageRequestReviewProps = {} & RouteComponentProps;
 
 const PageRequestReview: React.FunctionComponent<PageRequestReviewProps> = () => (
-  <AppLayoutWithSidebar>
-    <section data-testid="PageRequestReview">
-      <h1>PageRequestReview</h1>
-    </section>
-  </AppLayoutWithSidebar>
+  <PageEditContainer title="Review & Launch" testId="PageRequestReview">
```
LZoog, line 12: I like taking out the `isLoading` and `notFound` if check redundancies and this PR overall, but I don't believe we will want to use `PageEditContainer` on the "Review & Launch" page.

Since that page is a "review" page and not a page in which you can edit an experiment, I don't believe we want `HeaderEditExperiment` on there, plus we'll need some kind of polling. I like this component for `PageEditOverview`, `PageEditBranches`, and later `PageEditAudience` and `PageEditMetrics` though.

jodyheavener: Oh? I was just going off of what the Figma doc showed (it has the header bit)

LZoog: @jodyheavener oh nope you're right, it is present with the most recent comps. Sorry, Figma is ever-evolving, I still pictured one day this turning into an "edit" form in which case I didn't think we'd want to allow that on the "review" page. I bet we can finesse polling in there with some option, otherwise whomever works on that can take the request review page out of this container so seems good to me.

LZoog (approved):
r+'ing, but still might be good to get Les's feedback too. 🙂

lmorchard:
I kind of agree that [function a child component is an anti-pattern](https://americanexpress.io/faccs-are-an-antipattern/). But, not strongly enough to block an r+ - the DRYing here is nice and I think other alternatives would require context hooks or other gymnastics that wouldn't necessarily be any clearer than this PR as-is

LZoog:
Sweet, seems it could be improved but doesn't block the review/merge, and since I can use this in the branch I'm working on now and you're on PTO I'm going to go ahead and merge it @jodyheavener, thanks! 👍

PR #3873, opened 2020-11-02
fix #3865 feat(nimbus): split probe_sets into primary/secondary

bbangert:
Because:

* We want to assign probe sets to a Nimbus experiment as either
  primary or secondary.

This commit:

* Updates the Django models so that the ProbeSet is linked through
  a m2m table that has its own model to store the type of link.


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -219,6 +219,17 @@ class Meta:
         model = NimbusExperiment
         fields = ("probe_sets",)
 
+    def update(self, experiment, data):
+        probe_sets = data.pop("probe_sets")
+        experiment = super().update(experiment, data)
+        with transaction.atomic():
```
jaredlockhart, line 225: Could move the experiment update call into the transaction so it all fails as one?

app/experimenter/experiments/migrations/0132_nimbus_add_primary_secondary_probe_sets.py
```diff
@@ -0,0 +1,54 @@
+# Generated by Django 3.0.7 on 2020-10-30 21:14
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
```
jaredlockhart, line 6: Oh, fold these two migrations together, just delete them both and then run `makemigrations` again.  Generally there's no reason to have more than one migration per pr.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -297,6 +299,18 @@ def __str__(self):  # pragma: no cover
         return self.name
 
 
+class NimbusExperimentProbeSets(models.Model):
+    experiment = models.ForeignKey(
+        NimbusExperiment,
+        on_delete=models.CASCADE,
+    )
+    probe_set = models.ForeignKey(NimbusProbeSet, on_delete=models.CASCADE)
+    is_primary = models.BooleanField(null=False)
```
jaredlockhart, line 308: Shouldn't this have `default=False`?

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -66,17 +67,23 @@ class Meta:
 
     @factory.post_generation
     def probe_sets(self, create, extracted, **kwargs):
+        is_primary = kwargs.pop("is_primary", True)
         if not create:
             # Simple build, do nothing.
             return
 
-        if extracted:
+        if isinstance(extracted, Iterable):
```
jaredlockhart, line 75: Yeah this works, I was also thinking it could be 

`extracted=None`
`if extracted is not None:`

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -219,6 +219,17 @@ class Meta:
...
+            for probe_set in probe_sets:
+                experiment.probe_sets.add(
+                    probe_set, through_defaults={"is_primary": True}
```
jaredlockhart, line 228: So we'll setup the primary/secondary distinction in the other pr?

jaredlockhart (changes requested):
Looks good just gotta fold those migrations together and I think `is_primary` should probably have `default=False`?

app/experimenter/experiments/models/nimbus.py
bbangert: I considered that, but didn't want errors to occur if someone forgot to specify what it should be. I can't think of any good reason off-hand why it should default to one or the other.

app/experimenter/experiments/tests/factories/nimbus.py
bbangert: Since the code here relies on having an iterable with the for loop, this seemed like the best option. It would error if this branch executed with a `True` for example (though I can't imagine why one would pass that in).

app/experimenter/experiments/api/v5/serializers.py
bbangert: Yup.

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Seems reasonable.

jaredlockhart (approved):
LGTM 👍 👍 👍

PR #3875, opened 2020-11-02
fix #3403 feature(nimbus): support mobile remote settings collection

### LINKED ISSUE #3403: Experimenter should publish to correct remote settings collection based on application field
We should only publish experiments to the appropriate remote settings collection for that platform – desktop experiments to the desktop collection, mobile to mobile.

I'm not sure if we want to do something different for the reference browser v.s. fenix or put it in the same collection – i think it's probably fine to put both in the same collection, but then the client sdk would need some logic to filter on the correct platform as well.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-261)
┆Issue Number: EXP-261

jaredlockhart:
Because

* We need to be able to send experiments to the relevant remote settings bucket for desktop and mobile experiments

This commit

* Moves the kinto bucket/collection configuration out of environment and into settings because it's consistent across all deployments
* Adds the mobile collection constant
* Modifies the kinto tasks to push to the relevant collection and monitor all collections for live/complete changes


app/experimenter/settings.py
```diff
@@ -406,10 +406,11 @@
...
 KINTO_PASS = config("KINTO_PASS")
-KINTO_BUCKET = config("KINTO_BUCKET")
-KINTO_BUCKET_MAIN = config("KINTO_BUCKET_MAIN")
-KINTO_COLLECTION = config("KINTO_COLLECTION")
-KINTO_COLLECTION_NIMBUS = config("KINTO_COLLECTION_NIMBUS")
+KINTO_BUCKET = "main-workspace"
+KINTO_BUCKET_MAIN = "main"
+KINTO_COLLECTION = "messaging-experiments"
+KINTO_COLLECTION_NIMBUS_DESKTOP = "nimbus-desktop-experiments"
+KINTO_COLLECTION_NIMBUS_MOBILE = "nimbus-mobile-experiments"
```
bbangert, line 413: 👍

bbangert (changes requested):
Looks good! Needs re-base for migration before merge.

jaredlockhart:
HA! good idea requesting change to keep the migrations in sync ;)

bbangert (approved):
Verrrry nice! 👍

PR #3877, opened 2020-11-03
fix #3858 feat(nimbus): add primary/secondary probe set gql

### LINKED ISSUE #3858: Update GQL query for primary/secondary probe sets
The GQL query needs to return primary/secondary probe sets, as well as allow mutations of both.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-537)
┆Issue Number: EXP-537

bbangert:
Because:

* We want to allow primary/secondary probe sets to be updated and
  displayed via GQL.

This commit:

* Updates the GQL types and mutations for primary/secondary probe sets.

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -211,21 +211,59 @@ class NimbusBranchUpdateSerializer(
 class NimbusProbeSetUpdateSerializer(
     NimbusChangeLogMixin, NimbusStatusRestrictionMixin, serializers.ModelSerializer
 ):
-    probe_sets = serializers.PrimaryKeyRelatedField(
+    primary_probe_sets = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=NimbusProbeSet.objects.all()
+    )
+
+    secondary_probe_sets = serializers.PrimaryKeyRelatedField(
         many=True, queryset=NimbusProbeSet.objects.all()
     )
 
     class Meta:
         model = NimbusExperiment
-        fields = ("probe_sets",)
+        fields = ("primary_probe_sets", "secondary_probe_sets")
+
+    def validate_primary_probe_sets(self, value):
+        if len(value) > 2:
+            raise serializers.ValidationError(
+                "Exceeded maximum primary probe set limit of 2."
+            )
+        return value
+
+    def validate(self, data):
+        """Validate the probe sets don't overlap and have no more than 2 primary.
+
+        Note that the default DRF validation ensures all the probe id's are valid and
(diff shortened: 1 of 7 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -211,22 +211,54 @@ def update(self, experiment, data):
...
+        Note that while the default validation can ensure all the probes are valid
+        and by default it does not save probe sets specified in both fields, it does
+        not throw a validation error. We manually check for overlap here to deliver a
+        useful validation error.
+
+        """
+        data = super().validate(data)
+        primary_probe_sets = set(data["primary_probe_sets"])
+        secondary_probe_sets = set(data["secondary_probe_sets"])
+        if primary_probe_sets.intersection(secondary_probe_sets):
```
jaredlockhart, line 238: ❤️ sets

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -211,22 +211,54 @@ def update(self, experiment, data):
...
-            for probe_set in probe_sets:
+            experiment.probe_sets.clear()
+            for probe_set in primary_probe_sets:
                 experiment.probe_sets.add(
                     probe_set, through_defaults={"is_primary": True}
                 )
+            for probe_set in secondary_probe_sets:
+                experiment.probe_sets.add(
+                    probe_set, through_defaults={"is_primary": False}
+                )
```
jaredlockhart, line 261: Yep this is a very nice design 👍

jaredlockhart (changes requested):
Is perfect 👍 Let's limit primary probe sets to max 2 per @tdsmith @emtwo

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -211,22 +211,57 @@ def update(self, experiment, data):
...
+                {
+                    "primary_probe_sets": (
+                        "Primary probe sets cannot overlap with secondary probe sets."
+                    )
+                }
+            )
+        if len(primary_probe_sets) > 2:
+            raise serializers.ValidationError(
+                {"primary_probe_sets": ("Exceeded maximum primary probe set limit of 2.")}
+            )
```
jaredlockhart, line 248: Usually per field validation can go in a separate method, only cross field validation needs to go in `validate` so this could go in `def validate_primary_probe_sets`

jaredlockhart (approved):
👍 👍 👍 thnx @bbangert !

app/experimenter/experiments/api/v5/serializers.py
bbangert: Ah right.

PR #3879, opened 2020-11-03
fix #3876 fix(nimbuS): add owner GQL type

### LINKED ISSUE #3876: Nimbus experiment should show owner in GQL
The GQL response for NimbusExperiment should include the owner information.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-540)
┆Issue Number: EXP-540

bbangert:
Because:

* We want owner information for Nimbus Experiments to be shown.

This commit:

* Includes the needed GQL type for the Nimbus Experiment owner.

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -1,4 +1,5 @@
 import graphene
+from django.contrib.auth import get_user_model
 from graphene_django.types import DjangoObjectType
 
 from experimenter.experiments.constants.nimbus import NimbusConstants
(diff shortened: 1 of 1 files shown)


app/experimenter/experiments/api/v5/types.py
```diff
@@ -81,6 +82,12 @@ class NimbusLabelValueType(graphene.ObjectType):
     value = graphene.String()
 
 
+class NimbusExperimentOwner(DjangoObjectType):
+    class Meta:
+        model = get_user_model()
+        fields = ("id", "username", "first_name", "last_name", "email")
+
+
 class NimbusExperimentType(DjangoObjectType):
```
jaredlockhart, line 91: You don't explicitly have to refer to it here?

bbangert: Nope, nor do we in the Nimbus models. I wanted to ensure they're the same model so I used the same function.

jaredlockhart (approved):
Nice 👍

PR #3880, opened 2020-11-03
fix #3874 feat(nimbus): compute bucket sample from population percent

### LINKED ISSUE #3874: Compute bucket count from population size
Rather than using a hard coded bucket count of 100 we should compute the number of buckets using the population_percentage field.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-539)
┆Issue Number: EXP-539

jaredlockhart:
Because

* We now expose a population percentage field in the nimbus ui to determine sizing

This commit

* Uses the user entered population_percent field to compute the number of buckets rather than using a hard coded default

### DIFF
--- a/app/experimenter/kinto/tasks/nimbus.py
+++ b/app/experimenter/kinto/tasks/nimbus.py
@@ -1,3 +1,5 @@
+from decimal import Decimal
+
 import markus
 from celery.utils.log import get_task_logger
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


app/experimenter/kinto/tasks/nimbus.py
```diff
@@ -49,7 +51,11 @@ def nimbus_push_experiment_to_kinto(experiment_id):
             NimbusIsolationGroup.request_isolation_group_buckets(
                 experiment.slug,
                 experiment,
-                NimbusExperiment.BUCKET_COUNT,
+                int(
+                    experiment.population_percent
+                    / Decimal("100.0")
+                    * NimbusExperiment.BUCKET_TOTAL
+                ),
```
bbangert, line 58: Is this going to round up/down by default? If we want it to round, might be better to explicitly call it out with `math.ceil` or `floor`.

jaredlockhart: Interesting point, I'd rather wait until we hit a case where that matters and have a clear signal for how to handle it before we explicitly address it.

bbangert (approved):

PR #3886, opened 2020-11-04
fix #3743 fix #3744 feat(nimbus): feature branch editing UI logic

### LINKED ISSUE #3743: `PageEditBranches` ratio mods + "split evenly" checkbox
1) Update the ratio functionality to match Figma (taking in an integer instead of percentage)
2) Add the "users should be split evenly between all branches" checkbox that, on check, should change all "ratio" fields to "1". It should be clicked by default.
3) Move "Add Branch" button to its proper place in Figma
4) Make sure the copy on this page matches Figma.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-483)
┆Issue Number: EXP-483

### LINKED ISSUE #3744: `PageEditBranches` feature configuration
Rework the feature configuration on the Branches page to reflect what's in Figma. See https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216266%3A109

It now toggles to open (with the "+Feature configuration" button and the closing X which should close all of them) and always maps to the same FeatureConfig, see notes in Figma. Clicking the X deletes all of them.

**Note: This section has undergone changes - if the above or Figma seems confusing, ask Ana and/or Jared to clarify expected functionality.**

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-484)
┆Issue Number: EXP-484

lmorchard:
Because

* we want to manage editing feature branches for an experiment

This commit:

* wires up buttons and form elements in FormBranch and FormBranches

* implements support for adding / removing / changing a feature for configuration (issue #3744)

* implements equal ratio checkbox (issue #3743)

* implements a local reducer for managing editing state

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.test.tsx
@@ -1,0 +1,158 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
+
+import { MOCK_CONFIG } from "../../lib/mocks";
+import {
+  SubjectBranch,
+  MOCK_ANNOTATED_BRANCH,
+  MOCK_FEATURE_CONFIG,
+  MOCK_FEATURE_CONFIG_WITH_SCHEMA,
+} from "./mocks";
+
+describe("FormBranch", () => {
+  it("renders as expected", () => {
+    render(<SubjectBranch />);
+    expect(screen.getByTestId("FormBranch")).toBeInTheDocument();
+    expect(screen.queryByTestId("control-pill")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("equal-ratio")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("feature-config-edit")).toBeInTheDocument();
+    expect(screen.queryByTestId("feature-config-add")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("feature-value-edit")).not.toBeInTheDocument();
+  });
+
+  it("includes a control label when reference branch", () => {
+    render(<SubjectBranch isReference />);
+    expect(screen.getByTestId("control-pill")).toBeInTheDocument();
+  });
+
+  it("indicates equal ratio when enabled", () => {
+    render(<SubjectBranch equalRatio />);
+    expect(screen.getByTestId("equal-ratio")).toBeInTheDocument();
+  });
+
+  it("reflects when feature is disabled", () => {
+    const { container } = render(
+      <SubjectBranch
+        branch={{ ...MOCK_ANNOTATED_BRANCH, featureEnabled: false }}
+      />,
+    );
+    const featureSwitchLabel = container.querySelector(
+      "label[for=demo-featureEnabled]",
+    );
+    expect(featureSwitchLabel).toHaveTextContent("Off");
+  });
+
+  it("hides feature configuration edit when feature not selected", () => {
+    render(
+      <SubjectBranch
+        branch={MOCK_ANNOTATED_BRANCH}
+        experimentFeatureConfig={null}
+      />,
+    );
+    expect(screen.queryByTestId("feature-config-edit")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("feature-config-add")).toBeInTheDocument();
+  });
+
+  it("hides feature value edit when schema is null", () => {
+    render(
+      <SubjectBranch
+        branch={{
+          ...MOCK_ANNOTATED_BRANCH,
+        }}
+        experimentFeatureConfig={MOCK_FEATURE_CONFIG}
+      />,
+    );
+    expect(screen.queryByTestId("feature-value-edit")).not.toBeInTheDocument();
+  });
+
+  it("hides feature value edit when feature disabled", () => {
+    render(
+      <SubjectBranch
+        branch={{ ...MOCK_ANNOTATED_BRANCH, featureEnabled: false }}
+        experimentFeatureConfig={MOCK_FEATURE_CONFIG_WITH_SCHEMA}
+      />,
+    );
+    expect(screen.queryByTestId("feature-value-edit")).not.toBeInTheDocument();
+  });
+
+  it("displays feature value edit when value is non-null", () => {
+    render(
+      <SubjectBranch
+        branch={{
+          ...MOCK_ANNOTATED_BRANCH,
+          featureValue: "this is a default value",
+          featureEnabled: true,
+        }}
+        experimentFeatureConfig={MOCK_FEATURE_CONFIG_WITH_SCHEMA}
+      />,
+    );
+    expect(screen.queryByTestId("feature-value-edit")).toBeInTheDocument();
+  });
+
+  it("calls onFeatureConfigChange with selected feature", async () => {
+    const featureIdx = 1;
+    const onFeatureConfigChange = jest.fn();
+    render(<SubjectBranch {...{ onFeatureConfigChange }} />);
+    fireEvent.change(screen.getByTestId("feature-config-select"), {
+      target: { value: featureIdx },
+    });
+    expect(onFeatureConfigChange).toHaveBeenCalledWith(
+      MOCK_CONFIG!.featureConfig![featureIdx],
+    );
+  });
+
+  it("calls onFeatureConfigChange with null when blank feature selected", async () => {
+    const onFeatureConfigChange = jest.fn();
+    render(<SubjectBranch {...{ onFeatureConfigChange }} />);
+    fireEvent.change(screen.getByTestId("feature-config-select"), {
+      target: { value: "" },
+    });
+    expect(onFeatureConfigChange).toHaveBeenCalledWith(null);
+  });
+
+  it("calls onRemove when the branch remove button is clicked", async () => {
+    const onRemove = jest.fn();
+    render(<SubjectBranch {...{ onRemove }} />);
+    fireEvent.click(screen.getByTestId("remove-branch"));
+    expect(onRemove).toHaveBeenCalled();
+  });
+
+  it("calls onChange when branch fields are edited", async () => {
+    const id = "testing123";
+    const onChange = jest.fn();
+    const branch = {
+      ...MOCK_ANNOTATED_BRANCH,
+      featureValue: "this is a default value",
+      featureEnabled: true,
+    };
+
+    const { container } = render(
+      <SubjectBranch
+        {...{ id, onChange, branch }}
+        experimentFeatureConfig={MOCK_FEATURE_CONFIG_WITH_SCHEMA}
+      />,
+    );
+
+    const expectedData = {
+      name: "example name",
+      description: "example description",
+      ratio: "42",
+      featureValue: "example value",
+    };
+
+    for (const [name, value] of Object.entries(expectedData)) {
+      onChange.mockClear();
+      const field = container.querySelector(`#${id}-${name}`);
+      expect(field).not.toBeNull();
+      fireEvent.change(field!, { target: { value } });
+      expect(onChange).toHaveBeenCalledWith({ ...branch, [name]: value });
+    }
+
+    fireEvent.click(container.querySelector(`#${id}-featureEnabled`)!);
+    expect(onChange).toHaveBeenCalledWith({ ...branch, featureEnabled: false });
+  });
+});
--- a/app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
@@ -1,0 +1,224 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import Form from "react-bootstrap/Form";
+import Col from "react-bootstrap/Col";
+import Button from "react-bootstrap/Button";
+import Badge from "react-bootstrap/Badge";
+import { ReactComponent as DeleteIcon } from "../../images/x.svg";
+
+import { getExperiment_experimentBySlug } from "../../types/getExperiment";
+
+import {
+  getConfig_nimbusConfig,
+  getConfig_nimbusConfig_featureConfig,
+} from "../../types/getConfig";
+
+import { AnnotatedBranch } from "./reducer";
+
+export const FormBranch = ({
+  id,
+  branch,
+  equalRatio,
+  isReference,
+  experimentFeatureConfig,
+  featureConfig,
+  onRemove,
+  onChange,
+  onAddFeatureConfig,
+  onRemoveFeatureConfig,
+  onFeatureConfigChange,
+}: {
+  id: string;
+  branch: AnnotatedBranch;
+  equalRatio?: boolean;
+  isReference?: boolean;
+  experimentFeatureConfig: getExperiment_experimentBySlug["featureConfig"];
+  featureConfig: getConfig_nimbusConfig["featureConfig"];
+  onRemove?: () => void;
+  onChange: (branch: AnnotatedBranch) => void;
+  onAddFeatureConfig: () => void;
+  onRemoveFeatureConfig: () => void;
+  onFeatureConfigChange: (
+    featureConfig: getConfig_nimbusConfig_featureConfig | null,
+  ) => void;
+}) => {
+  const { name, description, ratio, featureValue, featureEnabled } = branch;
+
+  const handleRemoveClick = () => {
+    onRemove && onRemove();
+  };
+
+  const handleChange = (
+    propertyName: string,
+    targetAttr: keyof React.ChangeEvent<HTMLInputElement>["target"] = "value",
+  ) => (ev: React.ChangeEvent<HTMLInputElement>) => {
+    onChange({
+      ...branch,
+      [propertyName]: ev.target[targetAttr],
+    });
+  };
+
+  const handleFeatureConfigChange = (
+    ev: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    const selectedIdx = ev.target.value;
+    if (selectedIdx === "") {
+      return onFeatureConfigChange(null);
+    }
+    const feature = featureConfig![parseInt(selectedIdx, 10)];
+    return onFeatureConfigChange(feature);
+  };
+
+  return (
+    <Form
+      className="mb-3 border border-secondary rounded"
+      data-testid="FormBranch"
+    >
+      <Form.Group className="p-1 mx-3 mt-2 mb-0">
+        <Form.Row>
+          <Form.Group as={Col} controlId={`${id}-name`} sm={4} md={3}>
+            <Form.Label>
+              Branch{" "}
+              {isReference && (
+                <Badge pill variant="primary" data-testid="control-pill">
+                  control
+                </Badge>
+              )}
+            </Form.Label>
+            <Form.Control
+              type="text"
+              defaultValue={name}
+              onChange={handleChange("name")}
+            />
+          </Form.Group>
+          <Form.Group as={Col} controlId={`${id}-description`}>
+            <Form.Label>Description</Form.Label>
+            <Form.Control
+              type="text"
+              defaultValue={description}
+              onChange={handleChange("description")}
+            />
+          </Form.Group>
+          <Form.Group as={Col} controlId={`${id}-ratio`} sm={2} md={2}>
+            <Form.Label>Ratio</Form.Label>
+            {equalRatio ? (
+              <p data-testid="equal-ratio" className="p-0 m-0">
+                Equal
+              </p>
+            ) : (
+              <Form.Control
+                type="text"
+                defaultValue={ratio}
+                onChange={handleChange("ratio")}
+              />
+            )}
+          </Form.Group>
+          <Form.Group as={Col} sm={1} className="align-top text-right">
+            {!isReference && onRemove && (
+              <Button
+                data-testid="remove-branch"
+                variant="light"
+                className="bg-transparent border-0 p-0 m-0"
+                title="Remove branch"
+                onClick={handleRemoveClick}
+              >
+                <DeleteIcon width="18" height="18" />
+              </Button>
+            )}
+          </Form.Group>
+        </Form.Row>
+      </Form.Group>
+
+      {experimentFeatureConfig === null ? (
+        <Form.Group className="px-2 mx-3">
+          <Form.Row>
+            <Button
+              variant="outline-primary"
+              size="sm"
+              data-testid="feature-config-add"
+              onClick={onAddFeatureConfig}
+            >
+              + Feature configuration
+            </Button>
+          </Form.Row>
+        </Form.Group>
+      ) : (
+        <Form.Group
+          className="px-3 py-2 border-top"
+          data-testid="feature-config-edit"
+        >
+          <Form.Row>
+            <Col className="px-2">Feature configuration</Col>
+            <Form.Group as={Col} sm={1} className="align-top text-right">
+              <Button
+                variant="light"
+                className="bg-transparent border-0 p-0 m-0"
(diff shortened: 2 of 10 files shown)


lmorchard:
Oof, and I was hoping to at least generate storybooks with this draft - but I forgot that I wired it up to fail on test failure, so no storybooks until either I change that or increase test coverage

lmorchard:
Oh yeah, and if only as a note to self after catching up on Slack: Need to filter available features in branches by experiment application. Though we could possibly chase that as a follow up issue or as part of EXP-505 before implementing save

lmorchard:
Looks like I finally got tests & linting happy. Going to open for review.

The main attractions are here in these stories:

* [EditBranches](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/34272700f52847861ada379207bd385a06b00a53/nimbus-ui/index.html?path=/story/pages-editbranches--basic)
* [FormBranches](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/34272700f52847861ada379207bd385a06b00a53/nimbus-ui/index.html?path=/story/components-formbranches--with-feature-values)
* [FormBranch](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/34272700f52847861ada379207bd385a06b00a53/nimbus-ui/index.html?path=/story/components-formbranches-formbranch--reference-branch)

LZoog (commented):
Went ahead and did a first pass, this is looking great overall and what you've added to Storybook seems so complete. 😍

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
```diff
@@ -2,47 +2,142 @@
...
+    featureConfig: experiment.featureConfig,
+    referenceBranch: experiment.referenceBranch,
+    treatmentBranches: experiment.treatmentBranches,
+    equalRatio: initialEqualRatio,
+  });
+
+  const handleAddBranch = () => dispatch({ type: "addBranch" });
+
+  const handleRemoveBranch = (idx: number) => () =>
+    dispatch({ type: "removeBranch", idx });
```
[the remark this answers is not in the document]
jodyheavener: I actually can't Add a Branch on an existing experiment (generated on Docker startup). Button click borks the page with this:

Screen Shot 2020-11-05 at 2 51 42 PM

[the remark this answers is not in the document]
jodyheavener: Oh I see, this happens when I land on a page with a reference and treatment branch and click Add a branch. If I first delete the existing treatment branch this error does not occur when clicking Add a branch.

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
```diff
@@ -2,47 +2,142 @@
...
+
   return (
     <section data-testid="FormBranches" className="border-top my-3">
       <Form className="p-2">
         <Form.Row className="my-3">
           <Form.Group controlId="evenRatio">
             <Form.Check
+              data-testid="equal-ratio-checkbox"
+              onChange={handleEqualRatioChange}
+              checked={equalRatio}
```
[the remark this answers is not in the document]
lmorchard: This will be totally dependent on whether the server-side experiment branches are all already in equal ratio when the page loads up. There is no other indicator or flag available in the data model to decide whether to check this box by default.

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { MOCK_CONFIG } from "../../lib/mocks";
+import {
+  SubjectBranch,
+  MOCK_BRANCH,
+  MOCK_FEATURE_CONFIG,
+  MOCK_FEATURE_CONFIG_WITH_SCHEMA,
+} from "./mocks";
+
+describe("FormBranch", () => {
+  it("renders as expected", async () => {
```
jodyheavener, line 17: nit: this doesn't need to be async

lmorchard:
Rebased against main, made a few small changes from comments, going to look into the "add branch" issues

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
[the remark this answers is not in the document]
lmorchard: Hmm, I wonder if it has something to do with the TODO and blank defaults I'm scratching my head at here... The branches really only have a slug to go by as unique identifier, but I'm currently adding a new branch with a blank slug.

https://github.com/mozilla/experimenter/pull/3886/files#diff-a3ac0a11b006a28c5e1601638ebf1a57b5317c3884090b1b75520f32a2898be0R47-R57

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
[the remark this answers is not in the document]
lmorchard: Also, having a control & treatment by default might be a thing we need to rig up on the server side? Not sure what their names & slugs would be, etc

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+    render(<SubjectBranch isReference />);
+    expect(screen.getByTestId("control-pill")).toBeInTheDocument();
+  });
+
+  it("indicates equal ratio when enabled", () => {
+    render(<SubjectBranch equalRatio />);
+    expect(screen.getByTestId("equal-ratio")).toBeInTheDocument();
+  });
+
+  xit("reflects when feature is disabled", () => {
```
jodyheavener, line 37: is this coming back or should it be removed?

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
```diff
@@ -0,0 +1,228 @@
...
+            <Form.Group as={Col} controlId={`${id}-feature`}>
+              <Form.Control
+                as="select"
+                data-testid="feature-config-select"
+                onChange={handleFeatureConfigChange}
+                value={featureConfig!.findIndex(
+                  (feature) =>
+                    experimentFeatureConfig !== null &&
+                    feature !== null &&
+                    feature.slug === experimentFeatureConfig.slug,
```
jodyheavener, line 178: Does `feature?.slug === experimentFeatureConfig?.slug` work?

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
```diff
@@ -0,0 +1,228 @@
...
+                )}
+              >
+                <option value="">Select...</option>
+                {featureConfig?.map(
+                  (feature, idx) =>
+                    feature !== null && (
```
jodyheavener, line 184: Could this be `feature && (`?

app/experimenter/nimbus-ui/src/components/PageEditBranches/index.test.tsx
```diff
@@ -3,28 +3,62 @@
...
+    });
+  });
+
+  it("handles onNext from FormBranches", async () => {
+    const { mock } = mockExperimentQuery("demo-slug");
+    render(<Subject mocks={[mock]} />);
+    await waitFor(() => {
+      expect(screen.getByTestId("PageEditBranches")).toBeInTheDocument();
+    });
+    fireEvent.click(await screen.getByTestId("next-button"));
```
jodyheavener, line 59: nit: don't need to await this

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+type AddBranchAction = {
+  type: "addBranch";
+};
+
+function addBranch(newState: FormBranchesState) {
+  // TODO: figure out some default values to stick here
+  const newBranch = {
+    name: "",
+    // TODO: maybe need to slugify client-side?
```
jodyheavener, line 51: Do we need issues for this and the above TODO?

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
```diff
@@ -0,0 +1,162 @@
...
+}
+
+export const REFERENCE_BRANCH_IDX = -1;
+
+export type FormBranchesState = {
+  equalRatio: boolean;
+  featureConfig: getExperiment_experimentBySlug["featureConfig"];
+  referenceBranch: getExperiment_experimentBySlug["referenceBranch"];
+  treatmentBranches: getExperiment_experimentBySlug["treatmentBranches"];
+};
```
jodyheavener, line 33: This could get slightly less verbose if you used `Pick<getExperiment_experimentBySlug, "featureConfig" | "referenceBranch" | "treatmentBranches"> & { equalRatio: boolean; };`, but this is perfectly fine (this is mostly me just excited to know about `Pick`)

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
```diff
@@ -0,0 +1,228 @@
...
+            <Col sm={1} md={1} className="px-2 text-center">
+              is
+            </Col>
+            <Form.Group as={Col} controlId={`${id}-featureEnabled`}>
+              <Form.Check
+                type="switch"
+                label={featureEnabled ? "On" : "Off"}
+                defaultChecked={featureEnabled}
+                onChange={handleChange("featureEnabled", "checked")}
+              />
```
jodyheavener, line 204: Is this toggle also supposed to stay in sync with the other toggles? They seem to operate independently for me.

jodyheavener (commented):
Couple thoughts and I saw that same Add branch issue as Lauren but overall this works how I expected it to and looks great.

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
lmorchard: I want to say I *did* have `feature &&` here and Typescript didn't like it

app/experimenter/nimbus-ui/src/components/PageEditBranches/index.test.tsx
lmorchard: Ah, yeah, I think in my flailings I added the await here before I added the `waitFor` before it

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.test.tsx
lmorchard: Ah yeah, that shouldn't be disabled.

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
lmorchard: No, this toggle is per-branch, since whether the feature is enabled or not can be part of the experiment branch

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
lmorchard: I *think* so? Though this would be true if both are null, but also that shouldn't be real case encountered

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
lmorchard: Planning to either resolve this as part of this PR if it's behind the add-branch issues - or, it can be a part of EXP-505, because I'm expecting some of this will need refinement when joined up with a mutation

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <Form.Control
+              type="text"
+              defaultValue={description}
+              onChange={handleChange("description")}
+            />
+          </Form.Group>
+          <Form.Group as={Col} controlId={`${id}-ratio`} sm={2} md={2}>
+            <Form.Label>Ratio</Form.Label>
+            {equalRatio ? (
+              <p data-testid="equal-ratio" className="p-0 m-0">
```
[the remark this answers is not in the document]
lmorchard: Yeah, I wrestled with that a bit but kind of gave up to get the rest of the functionality done. Might be better for a polish bug

lmorchard:
Okay, some significant changes here after head-scratching at the add/remove branch problems:

* Generating a default name like "New treatment 1" on adding a branch so that there's at least *something* and not blank content

* Converting branch name into an invisible slug on every update.

* Switched to annotating branches with a locally-generated `__key` property, since there is no good unique ID on any of these branches.
  * When I tried using the slug, that would result in weirdness where every character typed in "name" field caused the key / ID of all form fields to change and bad things happened.

lmorchard:
Several force-pushes later and I've learned: While `yarn lint-fix` does fix some linting problems, it doesn't run `tsc`. So, always run `yarn lint` also. 😞

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
```diff
@@ -6,43 +6,114 @@ import React from "react";
...
               type="checkbox"
               label="Users should be split evenly between all branches"
             />
           </Form.Group>
           <Form.Group as={Col} className="align-top text-right">
-            <Button variant="outline-primary" size="sm">
+            <Button
```
jodyheavener, line 111: @lmorchard It looks like, if you start with control+treatment branch, and then either 1) remove the treatment and click add branch, or 2) just click add branch twice, it adds duplicate branches 🤔

app/experimenter/nimbus-ui/src/lib/slugify.ts
```diff
@@ -0,0 +1,16 @@
...
+    .toString()
+    .toLowerCase()
+    .replace(/\s+/g, "-") // Replace spaces with -
+    .replace(/[^\w-]+/g, "") // Remove all non-word chars
+    .replace(/--+/g, "-") // Replace multiple - with single -
+    .replace(/^-+/, "") // Trim - from start of text
+    .replace(/-+$/, ""); // Trim - from end of text
+}
+
+export default slugify;
```
jodyheavener, line 16: I'm surprised this isn't being counted against test coverage?

jodyheavener:
I suspect we'll want to prevent accidental reloads/navigations with the [`useExitWarning`](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/hooks/useExitWarning.tsx) hook

app/experimenter/nimbus-ui/src/lib/slugify.ts
lmorchard: It's covered because other tested code is using it

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
lmorchard: I'm not seeing this happening - how are you reproducing it exactly?

lmorchard:
> I suspect we'll want to prevent accidental reloads/navigations with the useExitWarning hook

I'd defer that to EXP-505 / #3871 where we try to hook up the mutation

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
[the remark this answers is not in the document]
jaredlockhart: Oh yeah this is a discussion we've had countless times.  The pattern we generally settled on in the past was to have the server spit out the default empty configuration, ie an empty control branch and an empty treatment branch, and that way the UI can just take what the server gives it and spit it out verbatim without needing to do any custom preparation logic.  We can do that here if it's a blocker?

app/experimenter/nimbus-ui/src/components/FormBranches/index.tsx
lmorchard: Cannot reproduce this in Storybook. Did a full re-install of Experimenter and can finally see this happening when served up from the full app context with data from Django.

Not sure exactly what the problem is yet, but it seems like somehow the reducer `dispatch` function is getting called twice for adding a branch. But, only from the real server, not Storybook.

lmorchard: Looks like I've hit [this behavior in `useReducer`](https://github.com/facebook/react/issues/16295) and have some unintended mutations in the reducer.

lmorchard:
Okay. Rebased against main and reworked the form reducer to iron out unintended mutations. I *think* that resolves the double-add issue

jaredlockhart (commented):
Oh yeah @lmorchard it's already setup to the server only accepts `name` and automatically slugifies it:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/schema.graphql#L232-L246

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L198-L207

So I think you shouldn't even be able to set slug through the mutation.

lmorchard:
> So I think you shouldn't even be able to set slug through the mutation.

Okay, will strip out slugification code then

jaredlockhart:
Oh I see this is only dealing with up to the reducer level and doesn't actually introduce the gql mutation to persist to the server?

lmorchard:
> Oh I see this is only dealing with up to the reducer level and doesn't actually introduce the gql mutation to persist to the server?

Correct - this doesn't get wired up to the mutation until issue #3781

lmorchard:
Another force-push to strip out client-side slugification of branch name

jaredlockhart:
Interesting so I'm curious which TS type this is using that allowed the `slug` field to be set on a branch at all?  Without the mutation, apollo won't autogenerate the TS types.  So in theory if you had landed this expecting to be able to set the slug on the client side, when you added the mutation and tried to pass the data into it it would've complained that you can't set the slug and at that point it would've been obvious that you can't set the slug on the client side?

jaredlockhart:
How do you enter the feature value?

Screenshot_2020-11-09 Mozilla Experimenter(1)

lmorchard:
> Interesting so I'm curious which TS type this is using that allowed the slug field to be set on a branch at all?

[`AnnotatedBranch`](https://github.com/mozilla/experimenter/pull/3886/files#diff-a3ac0a11b006a28c5e1601638ebf1a57b5317c3884090b1b75520f32a2898be0R223-R225), which adds a `__key` for local use by React to [`getExperiment_experimentBySlug_treatmentBranches`](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/types/getExperiment.ts#L27-L35)

> How do you enter the feature value?

[Choose a feature that has a non-null schema and enable the feature](http://localhost:3001/?path=/story/components-formbranches--with-feature-values) - but it looks like none of the dummy experiment features has a non-null schema.

jaredlockhart:
> > Interesting so I'm curious which TS type this is using that allowed the slug field to be set on a branch at all?
> 
> [`AnnotatedBranch`](https://github.com/mozilla/experimenter/pull/3886/files#diff-a3ac0a11b006a28c5e1601638ebf1a57b5317c3884090b1b75520f32a2898be0R223-R225), which adds a `__key` for local use by React to [`getExperiment_experimentBySlug_treatmentBranches`](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/types/getExperiment.ts#L27-L35). There's no read-only restriction there
> 

Oh I see, in this case the [output](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/schema.graphql#L37-L44) and [input](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/schema.graphql#L232-L246) schemas are different, something keep in mind as we hook more of the frontend and backend up.

> > How do you enter the feature value?
> 
> [Choose a feature that has a non-null schema and enable the feature](http://localhost:3001/?path=/story/components-formbranches--with-feature-values) - but it looks like none of the dummy experiment features has a non-null schema.

Ah makes sense!

lmorchard:
> Oh I see, in this case the output and input schemas are different, something keep in mind as we hook more of the frontend and backend up.

Yeah, I'm hoping this [`extractSaveState` helper](https://github.com/mozilla/experimenter/pull/3886/files#diff-a3ac0a11b006a28c5e1601638ebf1a57b5317c3884090b1b75520f32a2898be0R63-R72) in the reducer is where things will need to be massaged for issue #3781 - that's where the `__key` annotation gets stripped and ideally any other transformations can happen just before submitting the mutation

app/experimenter/nimbus-ui/src/components/FormBranches/FormBranch.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+      const field = container.querySelector(`#${id}-${name}`);
+      expect(field).not.toBeNull();
+      fireEvent.change(field!, { target: { value } });
+      expect(onChange).toHaveBeenCalledWith({ ...branch, [name]: value });
+    }
+
+    fireEvent.click(container.querySelector(`#${id}-featureEnabled`)!);
+    expect(onChange).toHaveBeenCalledWith({ ...branch, featureEnabled: false });
+  });
+});
```
jodyheavener, line 158: Non-blocking nit: none of these need to be async

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
```diff
@@ -0,0 +1,266 @@
...
+  | SetEqualRatioAction
+  | SetFeatureConfigAction;
+
+type AddBranchAction = {
+  type: "addBranch";
+};
+
+function addBranch(state: FormBranchesState) {
+  let { lastId, referenceBranch, treatmentBranches } = state;
+  lastId++;
```
jodyheavener, line 105: This is more of a nitpick and probably isn't worth the effort, but it would be cool if the count decremented when you removed a branch.

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
```diff
@@ -0,0 +1,266 @@
...
+) {
+  let { referenceBranch, treatmentBranches } = state;
+  if (equalRatio) {
+    if (referenceBranch !== null) {
+      referenceBranch = { ...referenceBranch, ratio: 1 };
+    }
+    if (Array.isArray(treatmentBranches)) {
+      treatmentBranches = treatmentBranches.map(
+        (treatmentBranch) =>
+          treatmentBranch && {
```
jodyheavener, line 208: What scenario might this be falsy? Does this map need a filter to ensure only branch objects make it through?

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
```diff
@@ -0,0 +1,266 @@
...
+}
+
+export function stripOneAnnotatedBranch(branch: null | AnnotatedBranch) {
+  if (!branch) return null;
+  const {
+    __key, // eslint-disable-line @typescript-eslint/no-unused-vars
+    ...strippedBranch
+  } = branch;
+  return strippedBranch;
+}
```
jodyheavener, line 244: ~This function has me scratching my head a little. What's the destructuring doing?~ Ooooh it's spreading the rest of the properties to a new object. Clever!

jodyheavener (approved):
The duplication issue I was experiencing is no longer present, and everything works great. Left you a couple small comments but I wouldn't consider them blockers. Good on my end. 🏆

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
lmorchard: It's not a count, it's an attempt at a unique ID. If the index goes down, React may re-use existing nodes in the DOM with the wrong data. That's the problem I was seeing when I brought in the ID code

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
lmorchard: No idea when it might be falsy - I don't think it will in practical use. The generated type has a null case, so this tries to cover that.

app/experimenter/nimbus-ui/src/components/FormBranches/reducer.tsx
lmorchard: Yeah, basically a way to use destructuring to omit certain properties from a new object.

lmorchard:
Going to merge - some of these things may further change for #3781, so I'd like to unblock that

PR #3888, opened 2020-11-04
fixes #3887 fix(legacy): ensure pop percent are fixed before comparing

### LINKED ISSUE #3887: [Legacy] population percent extra changelogs with no changes
Extra changelogs are listed
https://experimenter.services.mozilla.com/experiments/measure-set-to-default-adoption-impact-of-changing-the-prompt-to-set-firefox-as-default-v2-win7/

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-546)
┆Issue Number: EXP-546

tiftran:
Because:

* division and precision

This commit:

* formats normandy percent decimal before comparing with population percent
* migration to prune empty population percent changelogs

### DIFF
--- a/app/experimenter/normandy/tasks.py
+++ b/app/experimenter/normandy/tasks.py
@@ -166,11 +166,15 @@ def update_population_percent(experiment, recipe_data, filter_objects):
         "namespaceSample"
     ):
 
-        percent_decimal = bucket_sample["count"] / bucket_sample["total"] * 100
-        rounded_decimal = format(percent_decimal, ".4f")
+        decimal.getcontext().prec = 5
+        percent_decimal = (
+            decimal.Decimal(bucket_sample["count"])
+            / decimal.Decimal(bucket_sample["total"])
+            * decimal.Decimal(100)
+        )
 
-        if str(experiment.population_percent) != rounded_decimal:
-            changed_data = {"population_percent": rounded_decimal}
+        if experiment.population_percent != percent_decimal:
+            changed_data = {"population_percent": percent_decimal}
 
     if changed_data:
         user_email = (
--- a/app/experimenter/normandy/tasks.py
+++ b/app/experimenter/normandy/tasks.py
@@ -165,11 +165,12 @@ def update_population_percent(experiment, recipe_data, filter_objects):
     if bucket_sample := filter_objects.get("bucketSample") or filter_objects.get(
         "namespaceSample"
     ):
-        percent_decimal = decimal.Decimal(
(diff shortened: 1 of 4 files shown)


app/experimenter/normandy/tasks.py
```diff
@@ -165,11 +165,12 @@ def update_population_percent(experiment, recipe_data, filter_objects):
...
-            bucket_sample["count"] / bucket_sample["total"] * 100
-        )
-        if experiment.population_percent != percent_decimal:
-            changed_data = {"population_percent": percent_decimal}
+
+        percent_decimal = bucket_sample["count"] / bucket_sample["total"] * 100
+        rounded_decimal = format(percent_decimal, ".4f")
+
+        if str(experiment.population_percent) != rounded_decimal:
+            changed_data = {"population_percent": rounded_decimal}
```
jaredlockhart, line 173: Oh you're comparing them as strings, in theory that works, I was thinking of explicitly setting the Decimal precision and comparing them that way:

https://gist.github.com/jackiekazil/6201722#3-round-decimal-by-setting-precision

I can't think off the top of my head whether that makes a real difference, but if this goes sideways then we can try it that way.

app/experimenter/normandy/tests/test_tasks.py
```diff
@@ -475,6 +475,43 @@ def test_firefox_version_no_updates_when_no_change(self):
 
         self.assertFalse(experiment.changes.filter(message="Added Version(s)").exists())
 
+    def test_live_experiment_rounds_population_percent_correctly(self):
```
jaredlockhart, line 478: I think we should have a test that replicates the case where unnecessary duplicates are created and asserts they aren't?

jaredlockhart (changes requested):
Thnx @tiftran !

app/experimenter/normandy/tests/test_tasks.py
tiftran: https://github.com/mozilla/experimenter/pull/3888/files#diff-2a8ff87f0439ee59f7dd865e60a59a2e0d52045a48fbcf807b7e4d532e0865f3R511
?

jaredlockhart: Oh oh! lol I'm dumb I thought that was an asserTrue yes good yep 👍

jaredlockhart (approved):
Yep is good thnx @tiftran ! 🎉

jaredlockhart:
"In a minute there is time for division and precision which a PR can reverse" - The Bug Ticket of T Tran Prufrock by T.S. Eliot

Issue #3833, opened 2020-10-27
Summary table edits and additions


data-sync-user:
Check out Figma on the "Summary" page.

Alter the `TableSummary` component to match the table under "Summary".

Also, on `PageSummary`, add the "Timeline" with total duration / enrollment. The blue bar with start/end date will be handled later. Also add the link for JSON representation (such as https://experimenter.services.mozilla.com/api/v6/experiments/homepage-remediation-exposure-nightly-test/, only available when in review or live status)

We may also want to consider renaming this table from `TableSummary` to `TableOverview` since it will no longer be an experiment summary table but a summary of the overview section; other tickets address the "Audience" and "Branch" tables. `ExperimentSummary` could be a good container name to hold the overview, branches, and audience tables since we want to display these 3 components on the summary page and request review page.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-524)
┆Issue Number: EXP-524

data-sync-user:
➤ Chelsea Lewis commented:

Draft in review https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=217258%3A1 ; aiming to finalize by end of day

PR #3889, opened 2020-11-04
fixes #3474 feat(nimbus): Add summary table to request review page

### LINKED ISSUE #3474: `PageRequestReview` copy component from `rapid`, display saved data
In EXP-295 the `PageRequestReview` component was created. 

Copy what's needed from [`rapid`'s `ExperimentDetails`](https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/rapid/components/experiments/ExperimentDetails.tsx#L210) into this page to display the table ([see in staging](https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/rapid/testing-test-test))`. Plopping this in directly from `rapid` will work, but also check out the table for [Figma Review & Launch](https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=216763%3A13530). (Ana has said she likes [Bootstrap's striped rows ](https://getbootstrap.com/docs/4.0/content/tables/#striped-rows).

Note that anything involving visualization related and will be handled in another issue, we just need to display the experiment data (from EXP-302) in the table. If it makes sense to not reference `rapid` and build it from scratch that works too.

`useInterval` for `fetchExperiment` and the `requestReview` (as well as the request review button and checkboxes) will be handled in follow ups.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-304)
┆Issue Number: EXP-304

LZoog:
This commit:
* Adds a TableSummary component to display experiment data on the request review page
* Adds owner and proposedDuration to getExperimentBySlug query
* Updates mocks as needed
* Renames PageEditContainer to ContainerEditPage

Because:
* We want to see a summary of the experiment on this page as well as on the Results page later

--

Got us a head start on EXP-524 / #3833 too.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableSummary/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableSummary/index.stories.tsx
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { mockExperimentQuery } from "../../lib/mocks";
+import AppLayout from "../AppLayout";
+import TableSummary from ".";
+import { RouterSlugProvider } from "../../lib/test-utils";
+
+storiesOf("components/TableSummary", module)
+  .add("filled out", () => {
+    const { data } = mockExperimentQuery("demo-slug");
+    return (
+      <Subject>
+        <TableSummary experiment={data!} />
+      </Subject>
+    );
+  })
+  .add("partially filled out", () => {
+    const { data } = mockExperimentQuery("demo-slug", {
+      secondaryProbeSets: [],
+      channels: [],
+      proposedDuration: 0,
+    });
+    return (
+      <Subject>
+        <TableSummary experiment={data!} />
+      </Subject>
+    );
+  })
+  .add("missing all fields", () => {
+    const { data } = mockExperimentQuery("demo-slug", {
+      owner: null,
+      hypothesis: null,
+      primaryProbeSets: [],
+      secondaryProbeSets: [],
+      channels: [],
+      firefoxMinVersion: null,
+      populationPercent: 0,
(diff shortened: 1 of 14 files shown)


app/experimenter/nimbus-ui/src/components/TableSummary/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+// type displayConfigOptionsProps =
+//   | Pick<getConfig_nimbusConfig, "firefoxMinVersion">
+//   | Pick<getConfig_nimbusConfig, "channels">
+//   | Pick<getConfig_nimbusConfig, "targetingConfigSlug">;
+
+const displayConfigLabelOrNotSet = (
+  description: string,
+  value: string | null,
+  options: any,
```
LZoog, line 130: @jodyheavener this and the object in `options.find` is where I know I need some TypeScript love. I kind of wonder if setting this to a generic array of objects is the best bet here (there's probably a prettier solution than what I have commented out regardless)... Open to any suggestions here. :)

app/experimenter/nimbus-ui/src/components/TableSummary/index.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      channels: [],
+      proposedDuration: 0,
+    });
+    return (
+      <Subject>
+        <TableSummary experiment={data!} />
+      </Subject>
+    );
+  })
+  .add("missing all fields", () => {
```
LZoog, line 33: FWIW this should never happen... the "Results" link will be disabled later if any _required fields_ are missing, but checking for a null value for each piece and displaying that "not set" text seemed like the right move to guard against it as a final sanity check on the request page (or if a user chooses to hit it directly and we don't have a redirect set up for it).

jodyheavener (commented):
Honestly I've been looking through and playing with this, and I can't really find anything I would change myself. I'll follow up once the owner/probeset/config options stuff is in, and probably r+

jodyheavener (approved):
Works how I expect it to, in app and Storybook. Thanks for renaming `ContainerEditPage`! :shipit:

PR #3892, opened 2020-11-04
fix #3857 chore(nimbus): check graphql types during make check

### LINKED ISSUE #3857: Ensure generated types are updated when GQL backend  changes
From [~jkerim@mozilla.com]:

"We should figure out a way to add a check in `make check` that ensures if we make changes on the backend and forget to re run `yarn types` it'll fail.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-536)
┆Issue Number: EXP-536

jaredlockhart:
Because

* It would be nice to know if the backend changed that the frontend would break

This commit

* Adds make commands to check and fix that the graphql types match the nimbus-ui types
* This will cause CI to break if a backend change would break the frontend

Example failure: https://gist.github.com/jaredlockhart/b218231de9990806bdcdb7b29281dfb9


bbangert (approved):
LGTM!

jodyheavener (approved):

PR #3904, opened 2020-11-05
fix #3832 feat(nimbus): add review ready gql query

### LINKED ISSUE #3832: Add read only GQL query for total experiment validation
Before an experiment can be marked for review, all fields must be filled out.  To simplify this logic for the frontend, we should add a read only query on the backend that checks that all the fields on NimbusExperiment, including m2m relationships (channels, probesets, branches) are complete and valid.

The query should check that all the fields are in a non empty/default state, and return a mapping of field: bool that is false if the field is incomplete and true if it is complete.

When all fields are marked as true, the review button in the UI can be unlocked.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-522)
┆Issue Number: EXP-522

bbangert:
Because:

* The UX needs to know when the experiment is review-ready.

This commit:

* Add's a GQL query that verifies the Nimbus experiment with the
  provided slug is ready or not.

### DIFF
--- a/app/experimenter/experiments/api/v5/query.py
+++ b/app/experimenter/experiments/api/v5/query.py
@@ -1,5 +1,7 @@
 import graphene
 
+from experimenter.experiments.api.v5.mutation import ObjectField
+from experimenter.experiments.api.v5.serializers import NimbusReadyForReviewSerializer
 from experimenter.experiments.api.v5.types import (
     NimbusExperimentType,
     NimbusFeatureConfigType,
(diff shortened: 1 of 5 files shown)


jaredlockhart (approved):
This is so perfect @bbangert thank you.  We'll probably have to extend this as the workflow becomes more nuanced, but we can do that as followups as the ux evolves.  Amazing thank you!

jaredlockhart:
@bbangert Oh and you have to rebase this on main to get circle to pass.

PR #3908, opened 2020-11-05
fix #3903 feat(nimbus): include app/channel map in GQL config

### LINKED ISSUE #3903: Add channel/application mapping to config gql query
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-551)
┆Issue Number: EXP-551

bbangert:
Because:

* We want to display the appropriate channels for each application.

This commit:

* Includes a label mapping of application label to the channel labels
  in the GQL config query.

### DIFF
--- a/app/experimenter/experiments/api/v5/query.py
+++ b/app/experimenter/experiments/api/v5/query.py
@@ -16,9 +16,15 @@ from experimenter.experiments.models.nimbus import (
 )
 
 
+class ApplicationChannel(graphene.ObjectType):
+    label = graphene.String()
+    channels = graphene.List(graphene.String)
+
+
 class NimbusConfigurationType(graphene.ObjectType):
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/api/v5/query.py
```diff
@@ -16,9 +16,15 @@
 )
 
 
+class ApplicationChannel(graphene.ObjectType):
+    label = graphene.String()
```
jaredlockhart, line 20: `label` here could be `application` to be a little more readable?

jaredlockhart (approved):
Looks great thnx @bbangert !

app/experimenter/experiments/api/v5/query.py
bbangert: I called it label to match that its the label for the application in the list.

PR #3910, opened 2020-11-05
fix #3905 feat(nimbus): validate channels for experiment mutation

### LINKED ISSUE #3905: GQL mutation for channels should validate against application
The GQL mutation to add channels should validate its valid against the experiment application.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-553)
┆Issue Number: EXP-553

bbangert:
Because:

* We want to avoid setting invalid channels on an experiment.

This commit:

* Adds channel validation to the audience update mutation to ensure
  submitted channels are valid for the application.

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_serializers.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers.py
@@ -552,6 +552,7 @@ class TestNimbusAudienceUpdateSerializer(TestCase):
         user = UserFactory()
         experiment = NimbusExperimentFactory(
             channels=[],
+            application=NimbusExperiment.Application.DESKTOP,
             firefox_min_version=None,
             population_percent=None,
             proposed_duration=None,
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -593,6 +594,40 @@ def test_serializer_updates_audience_on_experiment(self):
...
+        user = UserFactory()
+        experiment = NimbusExperimentFactory(
+            channels=[],
+            application=NimbusExperiment.Application.FENIX,
+            firefox_min_version=None,
+            population_percent=None,
+            proposed_duration=None,
+            proposed_enrollment=None,
+            targeting_config_slug=None,
+            total_enrolled_clients=0,
```
jaredlockhart, line 607: Are these all necessary for this test?

jaredlockhart (approved):
Parfait, thanks @bbangert ! 🎉

app/experimenter/experiments/tests/api/v5/test_serializers.py
bbangert: I wanted to retain them for consistency so it was clear what change causes the error.

jaredlockhart: Ah yep 👍

PR #3915, opened 2020-11-06
fix #3907 chore(nimbus): make types_check should fail with graphql changes

### LINKED ISSUE #3907: make types_check not detecting changes
We want make types_check to fail if there's changes in the backend that aren't updated into the frontend.  The first pass on it broke if the backend and frontend didn't match for currently defined fields, ie renaming a field on the backend and not the frontend.  But it doesn't seem to be failing if you make changes in the backend that don't explicitly break the frontend, so I should see if I can get it to break there too so we don't miss changes.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-555)
┆Issue Number: EXP-555

jaredlockhart:
Because

* We'd like to make sure that the apollo client generated code stays in sync with the server

This commit

* Uses the schema.graphql to check whether the backend is in sync with the frontend codebase
* Regenerates the schema.graphql and typescript stubs together so they stay in sync

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -46,8 +46,14 @@ RUN poetry check
 COPY ./package.json /app/package.json
 COPY ./yarn.lock /app/yarn.lock
 COPY ./experimenter/legacy-ui/core/package.json /app/experimenter/legacy-ui/core/package.json
+RUN yarn install --frozen-lockfile
+
 COPY ./experimenter/legacy-ui/rapid/package.json /app/experimenter/legacy-ui/rapid/package.json
+RUN yarn install --frozen-lockfile
+
 COPY ./experimenter/legacy-ui/visualization/package.json /app/experimenter/legacy-ui/visualization/package.json
+RUN yarn install --frozen-lockfile
+
 COPY ./experimenter/nimbus-ui/package.json /app/experimenter/nimbus-ui/package.json
 RUN yarn install --frozen-lockfile
 
--- a/app/experimenter/nimbus-ui/README.md
+++ b/app/experimenter/nimbus-ui/README.md
@@ -51,7 +51,7 @@ You can read up over on their page if you want to know all the details, but if y
 
 All of our GraphQL resolvers produce types that are compatible with TypeScript, so there is no need to write new types for the queries you perform with Apollo.
 
-To generate types, just run `yarn types`. This will generate type declarations inside a `types/` directory at the root of this app. Once generated you can import and use as you see fit.
+To generate types, just run `yarn generate-types`. This will generate type declarations inside a `types/` directory at the root of this app. Once generated you can import and use as you see fit.  If there are changes to the graphql API on the server, run `make generate_types` to export the server schema and update the typescript stubs.
 
 For example, a query that looks like this:
 
@@ -84,8 +84,6 @@ export interface GetExperimentOverviews {
   experiments: (GetExperimentOverviews_experiments | null)[] | null;
 }
 ```
-
-**Note**: this command performs introspection on the GraphQL endpoint, so your local server must be running for it to work.
 
 ## Error handling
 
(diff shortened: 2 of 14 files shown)


app/Dockerfile
```diff
@@ -46,8 +46,14 @@ RUN poetry check
 COPY ./package.json /app/package.json
 COPY ./yarn.lock /app/yarn.lock
 COPY ./experimenter/legacy-ui/core/package.json /app/experimenter/legacy-ui/core/package.json
+RUN yarn install --frozen-lockfile
+
 COPY ./experimenter/legacy-ui/rapid/package.json /app/experimenter/legacy-ui/rapid/package.json
+RUN yarn install --frozen-lockfile
+
 COPY ./experimenter/legacy-ui/visualization/package.json /app/experimenter/legacy-ui/visualization/package.json
+RUN yarn install --frozen-lockfile
```
jodyheavener, line 55: Will a `yarn install` at the app root not work (over all the workspaces, or does that not exist in this context)?

app/experimenter/nimbus-ui/apollo.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  client: {
+      service: {
+      name: 'experimenter',
+      localSchemaFile: './schema.graphql'
+      }
```
jodyheavener, line 6: nit: indentation, but not a blocker

jodyheavener (approved):
Neat, makes sense to me. Two non blocking comments.

app/Dockerfile
jaredlockhart: Yeah so, the way it was setup was copy all the package.jsons in and then install all packages, but if you change something in just nimbus-ui/package.json it'll invalidate that whole cache layer and reinstall everything, so i broke it up to cache each of those layers separately, so if we make any changes to just the nimbus one it won't invalidate the others, which should make things a little quick for local dev, and also speed up dependabot prs.  Just a small performance improvement.

jodyheavener (changes requested):
This is awesome. Just gotta update the Nimbus README to reflect these changes and I'll r+

app/experimenter/nimbus-ui/README.md
```diff
@@ -51,7 +51,7 @@ You can read up over on their page if you want to know all the details, but if y
 
 All of our GraphQL resolvers produce types that are compatible with TypeScript, so there is no need to write new types for the queries you perform with Apollo.
 
-To generate types, just run `yarn types`. This will generate type declarations inside a `types/` directory at the root of this app. Once generated you can import and use as you see fit.
+To generate types, just run `yarn generate-types`. This will generate type declarations inside a `types/` directory at the root of this app. Once generated you can import and use as you see fit.  If there are changes to the graphql API on the server, run `make generate_types` to export the server schema and update the typescript stubs.
```
jodyheavener, line 54: Oh hah, I was actually referring to the bit a little further down about your local server needing to be running: 

> this command performs introspection on the GraphQL endpoint, so your local server must be running for it to work.

I don't want you to have to rerun CI just for that tho. Let's just get it in a later PR

jodyheavener (approved):

app/experimenter/nimbus-ui/README.md
jaredlockhart: Oh no I gotta repush anyways, updated 👍

jodyheavener: Soooo happy about this change ahah 🙂

jodyheavener: Not needing to introspect a live server is a world of improvement for me.

jaredlockhart: Yeah so glad I realized I could just dump the schema out from the server instead XD  That's how we do it for the old API and I realized wait I'll bet graphene does something similar.  Of course it does ;D

Issue #3759, opened 2020-10-15
Conditionally display "Design" + "Results" links based on experiment status


data-sync-user:
Status will be available from EXP-302 and visualization data from EXP-286. EXP-497 has similar conditionals.

1) If the experiment state is "accepted," "live," or "complete," all of the `nimbus/:slug:/edit` links in the nav should be hidden in favor of displaying a different set of links, `nimbus/:slug:/design` and `nimbus/:slug:/results`, pages/routes which were created in EXP-495.
 2) Add the icons for both links, make sure the "active" and other states set up in EXP-295 and EXP-468 work as expected

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-498)
┆Issue Number: EXP-498

data-sync-user:
➤ Jody Heavener commented:

Removed the third point and split it to https://jira.mozilla.com/browse/EXP-642 since it's a little more involved

Issue #3932, opened 2020-11-09
Add "Summary" page with header and summary table


data-sync-user:
We need a summary page (`PageSummary`?) to display the experiment header (showing name, slug, status) and the `TableSummary` which should show at `nimbus/:slug:/`. Add the page to Storybook of course.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-563)
┆Issue Number: EXP-563

PR #3935, opened 2020-11-09
fixes #3756 feat(nimbus): Add Results and Design pages and routes

### LINKED ISSUE #3756: Add "Design" and "Results" routes + placeholder pages
When the experiment is in “Accepted” status, the sidebar will reflect links to a “Design” page and disabled “Results” link which will be enabled when the results are ready.

We won't conditionally display these links in the issue, but we need these pages to exist so moving forward, Visualization work can be built into its intended location instead of `PageDetails` in the `rapid` project.

1) Add a dummy `PageExperimentDesign/index.tsx` component to display on `nimbus/:slug:/design`, wrapped in `AppLayoutWithSidebar`. This doesn't actually have to display any content, just a page component that’ll be ready for buildout.
2) Add a dummy `PageExperimentResults/index.tsx` component to display on `nimbus/:slug:/results`, wrapped in `AppLayoutWithSidebar`. This doesn't actually have to display any content, just a page component that’ll be ready for buildout.
3) A GQL query added in EXP-302 should be present on the Branches and Overview page. Add this onto the Results page as well as we'll need the data in EXP-461.
4) Don’t forget to the new pages to Storybook.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-495)
┆Issue Number: EXP-495

LZoog:
Because:
* We need these pages for when the experiment is no longer in draft status.

This commit:
* Adds placeholder pages with routes for "Results" and "Design"
* Renames ContainerEditPage since it's used on more than just "Edit" pages

---

Open to suggestions on the new name for `ContainerEditPage`. I know `AppLayoutWithSidebarAndData` is very wordy, but it seemed to fit our naming pattern of other components and it's clear IMO what you're getting when using it. 🤷 

Also FWIW, displaying the links in the sidebar will be done in #3759.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/App/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/App/index.tsx
@@ -9,6 +9,8 @@ import { GET_CONFIG_QUERY } from "../../gql/config";
 import PageLoading from "../PageLoading";
 import PageHome from "../PageHome";
 import PageNew from "../PageNew";
+import PageDesign from "../PageDesign";
+import PageResults from "../PageResults";
 import PageEditOverview from "../PageEditOverview";
 import PageEditBranches from "../PageEditBranches";
 import PageRequestReview from "../PageRequestReview";
@@ -36,6 +38,8 @@ const App = ({ basepath }: { basepath: string }) => {
         <PageEditBranches path="branches" />
       </Root>
       <PageRequestReview path=":slug/request-review" />
+      <PageDesign path=":slug/design" />
+      <PageResults path=":slug/results" />
     </Router>
   );
 };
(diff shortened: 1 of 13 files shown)


jodyheavener (approved):
Whew, `AppLayoutWithSidebarAndData` is a mouthful. I'm not sure if I would recognize that `Data` meant "Experiment data", though, since data it's a pretty broad term. But I also don't think it should be `AppLayoutWithSidebarAndExperiment`, and we could spend all day finding the perfect name for this, so this is fine with me.

G2G after addressing those linting issues

LZoog:
> Whew, `AppLayoutWithSidebarAndData` is a mouthful. I'm not sure if I would recognize that Data meant "Experiment data", though, since data it's a pretty broad term. But I also don't think it should be `AppLayoutWithSidebarAndExperiment` and we could spend all day finding the perfect name for this, so this is fine with me.

I agree. We chatted some in Slack about this but it can probably be improved in #3932 since we'll need this component on the summary page. Prolly something like: `<AppLayoutWithExperiment sidebar>` (or `noSidebar`) to conditionally render `AppLayout` or `AppLayoutWithSidebar`.

PR #3942, opened 2020-11-09
    fix #3853 fix(legacy): Added subscriber emails on comments

### LINKED ISSUE #3853: Send a subscriber email on comments
Let's add a new email that notifies subscribers when someone comments on an experiment.

    Subject: <user email> commented on <experiment name>
    Body:
    <User Email>:
    <Comment body>
    <link to comment>

We can add a function in [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/email/legacy.py) called `send_comment_email`

and then trigger it [right here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/views.py#L215)

no need to use a celery task.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-533)
┆Issue Number: EXP-533

tiftran:
Because

* Users would like to be notified when a comment is made

This commit

* sends an email to subscribers when a comment is made in legacy experimenter


app/experimenter/experiments/tests/test_email/test_email_legacy.py
```diff
@@ -254,3 +258,18 @@ def test_send_experiment_pausing_email(self):
             ],
         )
         self.assertIn("May 6, 2019", sent_email.body)
+
+    def test_send_experiment_comment_email(self):
+        experiment = ExperimentFactory.create(name="experiment1")
```
jaredlockhart, line 263: make sure to set type to pref flip in here otherwise this will intermittently fail

app/experimenter/experiments/constants/legacy.py
```diff
@@ -1117,6 +1119,8 @@ def FEATURE_TYPE_CHOICES(cls):  # pragma: no cover
         "Experimenter enrollment ending verification " "for: {name} {version} {channel}"
     )
 
+    COMMENT_EMAIL_SUBJECT = "[Experimenter]: {email} commented on {name}"
```
jaredlockhart, line 1122: You could just pass the experiment in here and do `{experiment.owner.email} commented on {experiment.name}` so you just have to pass the experiment in

app/experimenter/experiments/email/legacy.py
```diff
@@ -73,6 +73,27 @@ def send_enrollment_pause_email(experiment):
     )
 
 
+def send_experiment_comment_email(comment):
+    experiment = comment.experiment
+    comment_url = "{experiment_url}/#comment{comment}".format(
+        experiment_url=experiment.experiment_url, comment=comment.id
+    )
```
jaredlockhart, line 80: Oooo very nice, you could make this a `get_absolute_url` on the comment model

jaredlockhart (approved):
Very nice 👏 💯 

Looks like there might be an intermittent test failure hiding in there, but otherwise looks great 👍

PR #3943, opened 2020-11-09
chore(project): run make check on main

jaredlockhart:
Because

* We disabled the `up to date with main` branch protection, it's possible for changes that work as a pr to land on main and break main

This commit

* Gates deploying on running `make check` on main so if a pr lands on main that breaks we'll catch it before it gets deployed


lmorchard (approved):
LGTM!

PR #3945, opened 2020-11-09
chore (project): run build on main

jaredlockhart:
Because

* We want to run check on main and check requires build

This commit

* Enables build on main.  Derp.


lmorchard (approved):
Also lgtm, but my eyes seem substandard today

jaredlockhart:
@lmorchard mine too XD

PR #3946, opened 2020-11-10
fix #3914 fix(legacy): Change DC links and buttons to NDT web

### LINKED ISSUE #3914: [Legacy] Change Delivery Console Links and buttons to NDT web
Now that NDT web is here, we should change delivery console links and buttons to redirect to NDT web.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-557)
┆Issue Number: EXP-557

tiftran:
Because

* NDT is the preferred way to view recipes

This commits

* Swaps DC related things to NDT things
* Introduces new environment variables: EXPERIMENTER_INSTANCE and NORMANDY_DEVTOOLS_HOST


.env.sample
```diff
@@ -20,6 +20,7 @@ EMAIL_REVIEW=
 EMAIL_SENDER=
 EMAIL_SHIP=
 EXPERIMENTS_PAGINATE_BY=50
+EXPERIMENTER_INSTANCE=local
```
jaredlockhart, line 23: Can't this just use `HOSTNAME` [https://github.com/mozilla/experimenter/blob/main/.env.sample#L26](here)?

.env.sample
```diff
@@ -30,6 +31,7 @@ KINTO_USER=experimenter
 LOGGING_CONSOLE_LEVEL=INFO
 LOGGING_USE_JSON=False
 NORMANDY_API_HOST=
+NORMANDY_DEVTOOLS_HOST=https://normandy-devtools.services.mozilla.com/
```
jaredlockhart, line 34: Will this be different for different deployments?  Or will it always point to the prod one?

.env.sample
tiftran: oh! is hostname for stage: "stage" and prod: "prod"?

jaredlockhart: no it's the full url, do you need something that's just 'local' 'stage' 'prod'? how does ndt use that?

.env.sample
tiftran: I don't think it should be different. stage and prod could point to the prod one. One could navigate to stage recipes with the prod devtools instance.

.env.sample
tiftran: yup, devtools urls follow this pattern ext+normandy://<local, dev, stage, prod>/recipes

jaredlockhart: does ndt use that to figure out which experimenter url to use?

.env.sample
jaredlockhart: If it's always the same then you can just move it right into settings.py and not make it an env var at all?

.env.sample
tiftran: errrm do you mean does experimenter uses it to figure out which ndt url to use?

jaredlockhart: No like why is it passed to NDT? How does it use it?

app/experimenter/settings.py
```diff
@@ -375,16 +375,19 @@
...
-DELIVERY_CONSOLE_RECIPE_URL = urljoin(DELIVERY_CONSOLE_HOST, "/recipe/{id}/")
-DELIVERY_CONSOLE_EXPERIMENT_IMPORT_URL = urljoin(
-    DELIVERY_CONSOLE_HOST, "/recipe/import/{slug}/"
-)
 NORMANDY_API_HOST = config("NORMANDY_API_HOST")
 NORMANDY_API_RECIPE_URL = urljoin(NORMANDY_API_HOST, "/api/v3/recipe/{id}/")
 NORMANDY_API_RECIPES_LIST_URL = urljoin(NORMANDY_API_HOST, "/api/v3/recipe/")
 
+NORMANDY_DEVTOOLS_HOST = config("NORMANDY_DEVTOOLS_HOST")
+NORMANDY_DEVTOOLS_RECIPE_URL = "{root}{recipe_url}".format(
```
tiftran, line 383: using urljoin makes the "#<instance>" disappear...

jaredlockhart: Here?  Or you mean the other thing about comments?

jaredlockhart (commented):
Looks good 👍 

Did you already file a bugzilla to add the env vars for stage/prod?  Otherwise they won't start after this merges

tiftran:
> Looks good 👍
> 
> Did you already file a bugzilla to add the env vars for stage/prod? Otherwise they won't start after this merges

@jaredlockhart  no, just wanted to ensure this was good to go, before creating a bugzilla ticket, considering that we changed from two env vars to one

jaredlockhart:
@tiftran Yep sounds good, can file bugzilla now 👍

tiftran:
Bugzilla ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=1677921

jaredlockhart (approved):
@tiftran We can merge this now! 🎉 Thnx @robotblake :D

PR #3949, opened 2020-11-10
fixes #3443 feat(nimbus): add ability to retrieve analysis results, display them on results page

### LINKED ISSUE #3443: Allow Visualization API endpoint to be consumed, add to `PageResults`
In the `rapid` project, the `fetchExperiment` function fetches the experiment data at `/api/v3/experiments/${experimentSlug}/` as well as the visualization data at `/api/v3/visualization/${experimentSlug}/`. The experiment data was fetched in EXP-302; this ticket concerns fetching the visualization data.

This must be a REST API call because it exposes Jetstream data and should be the only REST API this application needs. We don't need to store this data in the Apollo Client cache because we want the data to be fresh every time a page using that data is hit.

EXP-495 adds a `ResultsPage` at `nimbus/:slug:/results` and at the time of writing is the only page we need to use this for.

1) Copy the `legacy-ui/visualization/` package into `nimbus-ui/visualization/`, which at present is a tiny package that's only used for using Svelte embedded in React that will be used at a later time but can be copied over now as it's related and a trivial task.
2) Consume the API endpoint on `PageResults`. [This React hook](https://github.com/slorber/react-async-hook) might come in handy). We don't need to do anything with the data yet but we will in an immediate follow up, EXP-461, just make sure the endpoint is successfully hit at `nimbus/:slug:/results`.

See this PR which added this endpoint for us in the `rapid` project: [https://github.com/mozilla/experimenter/pull/3481/files](https://github.com/mozilla/experimenter/pull/3481/files)
 Also see the visualization ADR: [https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0003-analysis-visualization-architecture-approach.md](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/adrs/0003-analysis-visualization-architecture-approach.md)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-286)
┆Issue Number: EXP-286

jodyheavener:
This PR...

- Adds the `useAnalysis` React hook to retrieve an experiment's analysis data from the API
  - Also sets up and exports the TS interface of the API results
  - Also sets up the hook's basic usage in the `PageResults` component
- ~Copies Rapid's `visualization` package over to Nimbus (that's it, doesn't set it up or anything)~

@emtwo tagging you for visibility into how this works :)


app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -160,7 +160,7 @@ const FormOverview = ({
             name="publicDescription"
             as="textarea"
             rows={3}
-            defaultValue={experiment?.publicDescription || ""}
+            defaultValue={experiment?.publicDescription as string}
```
jodyheavener, line 163: This is not related but satisfies that one annoying uncovered test line

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -3,13 +3,39 @@
...
 
 import React from "react";
-import { RouteComponentProps } from "@reach/router";
+import { RouteComponentProps, useParams } from "@reach/router";
 import AppLayoutWithSidebarAndData from "../AppLayoutWithSidebarAndData";
+import { useAnalysis } from "../../hooks";
 
 const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
+  const { slug } = useParams();
+  const { loading, error, result } = useAnalysis(slug);
```
LZoog, line 12: I think we may want to check for some truthiness _above_ the return statement - this page is useful when the visualization data is available and in #3759 we'll be disabling the link to it if it's not available (I imagine we'll use this hook elsewhere for that check if the experiment status means the visualization data could be available).

I think we could at least check if it's loading and show the loading page if so. I'm less certain of an error state, it might make sense to go ahead and let the `useExperiment` hook fire a request up to the server to display the experiment data with an error message like you have here for analysis data. 🤔

@emtwo is the Results page going to still be useful if the visualization data errors? If not, we might as well save a request here and not query for experiment data if `error` is truthy here.

app/experimenter/nimbus-ui/src/components/PageResults/index.test.tsx
```diff
@@ -7,18 +7,72 @@ import { screen, render, waitFor } from "@testing-library/react";
...
-      </RouterSlugProvider>,
+    render(<Subject />);
+    await waitFor(() => {
+      expect(screen.queryByTestId("PageResults")).toBeInTheDocument();
+    });
+  });
+
+  it("displays analysis data", async () => {
+    fetchMock.mockResponseOnce(
+      JSON.stringify({ burrito: "Nacho Cheese Doritos® Locos Tacos Supreme®" }),
```
LZoog, line 43: I see you wrote these tests at lunch time😂

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
emtwo: Experiment data might still be useful in the results page even if there is an error. We can still show things like branch names, probe sets, firefox versions, etc.

app/experimenter/nimbus-ui/visualization/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@experimenter/visualization",
```
emtwo, line 2: So, there is a good chance we won't end up using this visualization package so I don't think we need to copy it over. I think we will end up using another graphing library instead of doing all the svelte shenanigans

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
LZoog: @jodyheavener I asked Ana about this and she'd like the error message to include something along the lines of, "Please contact data science in #cirrus about this."

Sounds like yeah, let's let the experiment data be queried even if the visualization data errors. I'll also take back my loading page comment, it looks like the visualization API request and GQL query happen in parallel and the GQL query isn't requeried when the `result` becomes available so; LGTM here.

Though, I still think we can probably show the loading page, otherwise `useExperiment` is hit twice, once for `loading` from the visualization hook and another when `result` is ready because of the rerender.

jodyheavener: @LZoog cool, I'll update the error message (we may want to tweak when actual visualization UI work begins since I meant for this to be placeholder until then :D)

app/experimenter/nimbus-ui/visualization/package.json
LZoog: Sweet! I think after one more ticket we'll be safe to completely delete the `rapid` project under `nimbus-ui` - looks like this package currently lives at `legacy-ui/visualization` and not inside `rapid` so it can be kept around where it is now.

LZoog (approved):
LGTM! You may want to reword your commit message though since it no longer moves the visualization package over. :)

app/experimenter/nimbus-ui/visualization/package.json
emtwo: Ok this is my bad because I didn't add other tickets for 2 more PRs that closed under legacy for visualization. I will add them

PR #3950, opened 2020-11-10
fix(visualization): Adjust location of monitoring.

emtwo:
Screenshots below with where monitoring is shown in both cases (visualization on and off) below:

Screen Shot 2020-11-10 at 4 22 34 PM
Screen Shot 2020-11-10 at 4 22 45 PM


jaredlockhart (approved):
👍 👍 👍

PR #3957, opened 2020-11-12
fixes #3956 feat(nimbus): add targeting from Chrome attribution

### LINKED ISSUE #3956: Add chrome attribution targeting for first run
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-572)
┆Issue Number: EXP-572

k88hudson:
Because

* We want to support experiments that targeting Chrome users for
  aboutwelcome experiments.

This commit

* Add TARGETING_FIRST_RUN_CHROME_ATTRIBUTION for first run users that
  download the Browser from chrome


jaredlockhart (approved):
LGTM excepting actually testing the strings in a browser.  It'll be nice when we get those tests going.  Separately, I realize storing the targeting expressions this way is a bad design because every time we change them it'll introduce another db migration.  It's not the worst but, also not the best.  Something to think about.

PR #3959, opened 2020-11-12
fixes #3958 bug(nimbus): add opt-out pref to targeting

### LINKED ISSUE #3958: Add opt-out preference to targeting for Desktop <84
We fixed this issue in Desktop 84 https://bugzilla.mozilla.org/show_bug.cgi?id=1675544, but for clients in 83 and below we should include the `app.shield.optoutstudies.enabled` pref in targeting to exclude them.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-573)
┆Issue Number: EXP-573

k88hudson:
Because

  * Firefox Desktop <84 does not have global opt-out preference disabling

  This commit

  * Add opt-out preference app.shield.optoutstudies.enabled to targeting


jaredlockhart (approved):
👍

PR #3964, opened 2020-11-12
fix #3696 feat(nimbus): update overview form to match figma specs

### LINKED ISSUE #3696: `PageNew` modifications - back links, add "Application" field, finish styles

jaredlockhart:
Because

* We need the following changes to the create and edit overview pages
  - Add an X cancel button to create in the corner on create
  - Add the overview guidance text and validation for it
  - Make application readonly on edit overview

This commit

* Adds a cancel X at the top right of create
* Adds guidance text to the V5 API and frontend
* Adds validation for the guidance text on the backend
* Updates the overview form to treat server and client validation the same way
* Removes the invalid checks on button disables because you need to be able to save in an invalid state
* Removes the dirty checks on button disables because it requires focusing away from the input before it's clickable

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -113,6 +113,13 @@ class NimbusExperimentOverviewSerializer(
 
         return name
 
+    def validate_hypothesis(self, hypothesis):
+        if hypothesis.strip() == NimbusExperiment.HYPOTHESIS_DEFAULT.strip():
+            raise serializers.ValidationError(
+                "Please describe the hypothesis of your experiment."
+            )
+        return hypothesis
+
     def create(self, validated_data):
         validated_data.update(
             {
--- a/app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
@@ -12,6 +12,7 @@ import { useConfig } from "../../hooks/useConfig";
 
 type FormOverviewProps = {
   isLoading: boolean;
+  isServerValid: boolean;
   submitErrors: Record<string, string[]>;
   experiment?: getExperiment["experimentBySlug"];
   onSubmit: (data: Record<string, any>, reset: Function) => void;
@@ -21,6 +22,7 @@ type FormOverviewProps = {
 
 const FormOverview = ({
   isLoading,
+  isServerValid,
   submitErrors,
   experiment,
   onSubmit,
@@ -30,13 +32,15 @@ const FormOverview = ({
   const { handleSubmit, register, reset, errors, formState } = useForm({
     mode: "onTouched",
   });
-  const { isSubmitted, isValid, isDirty, touched } = formState;
-  const shouldWarnOnExit = useExitWarning();
-  const { application } = useConfig();
+  const { isSubmitted, isDirty, touched } = formState;
+  const { application, hypothesisDefault } = useConfig();
+  const isValid = isServerValid && formState.isValid;
+  const isDirtyUnsaved = isDirty && !(isValid && isSubmitted);
 
+  const shouldWarnOnExit = useExitWarning();
   useEffect(() => {
-    shouldWarnOnExit(isDirty);
-  }, [shouldWarnOnExit, isDirty]);
+    shouldWarnOnExit(isDirtyUnsaved);
+  }, [shouldWarnOnExit, isDirtyUnsaved]);
 
   const handleSubmitAfterValidation = useCallback(
     (data: Record<string, any>) => {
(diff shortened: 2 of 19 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -113,6 +113,13 @@ def validate_name(self, name):
 
         return name
 
+    def validate_hypothesis(self, hypothesis):
+        if hypothesis.strip() == NimbusExperiment.HYPOTHESIS_DEFAULT.strip():
+            raise serializers.ValidationError(
+                "Please describe the hypothesis of your experiment."
```
jodyheavener, line 119: This description is probably fine, and I'm sure some basic training in experimenter would make this point moot, but worth noting that I wasn't entirely sure, with the default text in the field, that this error message actually meant "you must change these values or write your own" (because on initial glance it looked like there _was_ a hypothesis value). But like I said, this might be nothing.

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -123,34 +127,37 @@ const FormOverview = ({
         <Form.Control
           {...nameValidated("hypothesis")}
           as="textarea"
-          rows={3}
-          defaultValue={experiment?.hypothesis || ""}
+          rows={5}
+          defaultValue={experiment?.hypothesis || hypothesisDefault || ""}
```
jodyheavener, line 131: I suppose this could just be `experiment?.hypothesis as string || hypothesisDefault`, no need for the empty string fallback

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -123,34 +127,37 @@ const FormOverview = ({
...
-        </Form.Control>
-        <Form.Text className="text-muted">
-          Experiments can only target one Application at a time.
-        </Form.Text>
-        <FormErrors name="application" />
-      </Form.Group>
+      {experiment ? (
+        <React.Fragment>
+          <p>Application</p>
+          <p>{experiment.application}</p>
```
jodyheavener, line 142: You mentioned this could be a followup; if you wanted to throw it in here I think it's as simple as `{application!.find((a) => a?.value === experiment.application)?.label}`

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -123,34 +127,37 @@ const FormOverview = ({
...
+                {app!.label}
+              </option>
+            ))}
+          </Form.Control>
+          <Form.Text className="text-muted">
+            Experiments can only target one Application at a time.
+          </Form.Text>
+          <FormErrors name="application" />
+        </Form.Group>
+      )}
```
jodyheavener, line 160: Actually this whole experiment check could probably be moved further in so we can keep the visual consistency of the form label and grouping all intact. Something like this:

```tsx
<Form.Group controlId="application">
  <Form.Label>Application</Form.Label>
  {experiment ? (
    <p>
      <b>
        {
          application!.find((a) => a?.value === experiment.application)
            ?.label
        }
      </b>
    </p>
  ) : (
    <>
      <Form.Control {...nameValidated("application")} as="select">
        <option value="">Select...</option>
        {application!.map((app, idx) => (
          <option key={`application-${idx}`} value={app!.value as string}>
            {app!.label}
          </option>
        ))}
      </Form.Control>
      <Form.Text className="text-muted">
        Experiments can only target one Application at a time.
      </Form.Text>
      <FormErrors name="application" />
    </>
  )}
</Form.Group>
```

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
```diff
@@ -7,6 +7,8 @@ import { RouteComponentProps } from "@reach/router";
 import AppLayout from "../AppLayout";
 import LinkExternal from "../LinkExternal";
 import FormOverview from "../FormOverview";
+import Row from "react-bootstrap/Row";
+import Col from "react-bootstrap/Col";
```
jodyheavener, line 11: Nit: all bootstrap components can come from one import

```tsx
import { Row, Col } from "react-bootstrap";
```

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
```diff
@@ -47,20 +50,33 @@ const PageNew: React.FunctionComponent<PageNewProps> = () => {
...
 
   return (
     <AppLayout testid="PageNew">
-      <h1 className="h2">Create a new Experiment</h1>
+      <Row>
+        <Col>
+          <h1 className="h2">Create a new Experiment</h1>
+        </Col>
+        <Col className="text-right">
+          <button className="btn pr-0" onClick={onFormCancel}>
```
jodyheavener, line 75: This button could use a title since there is no descriptive text content

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
```diff
@@ -47,20 +50,33 @@ const PageNew: React.FunctionComponent<PageNewProps> = () => {
...
+            &#10005;
```
jodyheavener, line 76: Can we use the SVG glyph here? Looks like we already have it in the repo, so just

```tsx
import { ReactComponent as DeleteIcon } from "../../images/x.svg";

// and then
<DeleteIcon width="18" height="18" />
```

jodyheavener (commented):
Couple thoughts here and there, but overall works well and makes sense. Ping me again when tests cooperate!

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Yeah it's definitely a training thing, they have to know how to fill that in correctly and it's more than we can put in a help text.

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
jaredlockhart: 🔥 🔥 🔥

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
jaredlockhart: Good call 👍

app/experimenter/nimbus-ui/src/components/PageNew/index.tsx
jaredlockhart: [You should import individual components like: react-bootstrap/Button rather than the entire library.](https://react-bootstrap.github.io/getting-started/introduction/#importing-components)

jodyheavener: TIL!

jodyheavener (approved):
Weeooo!

PR #3968, opened 2020-11-13
fixes #3967 bug(nimbus): Fix the nested expressions for first run

### LINKED ISSUE #3967: Fix nested expressions in TARGETING_FIRST_RUN targeting
There's some gnarly nesting in the first run expressions that needs to get fixed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-579)
┆Issue Number: EXP-579

k88hudson:
Because

  * The nested expressions in first run targeting was incorrect

This commit

  * Fixes TARGETING_FIRST_RUN and TARGETING_FIRST_RUN_CHROME_ATTRIBUTION


jaredlockhart (approved):
Oh that's fun, just referring to the other targeting expressions and expanding on them.  I think we'll be able to nuke that idea if we switch to multiselect for targeting in the future.  LGTM 👍

PR #3973, opened 2020-11-13
fixes #3695 feat(nimbus): Display viz analysis in PageResults, add HighlightsTable, stub other tables

### LINKED ISSUE #3695: Visualization: PageResults stub out + copy HighlightsTable & needed components from rapid
The visualization data will be available on the `PageResults` table after EXP-286 is completed.

In the `rapid` project, visualization data is displayed in a `ResultsTable` and a `HighlightsTable` on the `ExperimentDetails` page. We need to transfer both of these into `PageResults` and should essentially be able to copy-and-paste the work into this page.

The PR where the `ResultsTable` was added in `rapid`: [https://github.com/mozilla/experimenter/pull/3543](https://github.com/mozilla/experimenter/pull/3543)
 The PR where the `HighlightsTable` was added in `rapid`: [https://github.com/mozilla/experimenter/pull/3714](https://github.com/mozilla/experimenter/pull/3714)

The PR where the `Overview` was added in `rapid`: [https://github.com/mozilla/experimenter/pull/3787](https://github.com/mozilla/experimenter/pull/3787)

The PR where `PrimaryMetricTable` was added in `rapid`  [https://github.com/mozilla/experimenter/pull/3830](https://github.com/mozilla/experimenter/pull/3830)

The PR where `SecondaryMetricTable` was added in `rapid` [https://github.com/mozilla/experimenter/pull/3890](https://github.com/mozilla/experimenter/pull/3890)

NOTE: when this ticket is started there might be more here than what was merged. Maybe double check with Marina to make sure all the work was transferred over. Marina should be able to continue visualization work on the Results page after this ticket is merged.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-461)
┆Issue Number: EXP-461

LZoog:
This commit:
* Transfers legacy UI "page details" into PageResults with analysis available/unavailable conditions
* Adds highlights table
* Stubs out placeholders for other needed tables
* Adds tooltips and icons
* Adds better typing

Because:
* We want to display table visualizations on the "results" page

---

This PR is a little weird since so much is commented out, but it was necessary to show 100% line test coverage.

As other tables are built out, everything commented out will become uncommented and test coverage should remain at 100%. Also, I thought the highlights table used `ConfidenceInterval` since `TableVisualizationRow` returns it in `conversionChangeField` so I even started adding stories to Storybook for it and then realized it doesn't hit the switch case that calls that function, heh, but we'll use it in another table later so I didn't remove it since I added slightly better typing to that file.


app/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -260,7 +260,16 @@ export const mockExperimentQuery = (
               featureValue: '{"environmental-fact": "really-citizen"}',
               featureEnabled: true,
             },
-            featureConfig: null,
+            featureConfig: {
```
LZoog, line 263: @lmorchard, a heads up here that I added a default `featureConfig` to the mock. If this causes headache for you and your branch, I could revert this back to `null` and set a const for a featureConfig mock and pass that in where a non-null feature config is needed.

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { RouterSlugProvider } from "../../lib/test-utils";
+import { withLinks } from "@storybook/addon-links";
+import { mockExperimentQuery } from "../../lib/mocks";
+import TableVisualizationRow from ".";
+import { mockAnalysis } from "../../lib/visualization/mocks";
+import { DISPLAY_TYPE } from "../../lib/visualization/constants";
+
+const { mock } = mockExperimentQuery("demo-slug");
+
+// TODO: this can use a lot more stories as other tables are built out
```
LZoog, line 16: @emtwo as we use this row to display different states, we should add them here. I'm not even sure this one looks that great, but it gets us started.

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+  return (
+    <>
+      <p {...{ className }} data-testid={className}>
+        {significanceIcon}
+        &nbsp;
+        {tableLabel === TABLE_LABEL.HIGHLIGHTS
+          ? `${name} ${changeText} ${intervalText}`
+          : `${interval}`}
+      </p>
+      <ReactTooltip />
```
LZoog, line 81: @emtwo I replaced the tooltips for this table to better reflect Figma:

image

app/experimenter/nimbus-ui/src/lib/mocks.tsx
lmorchard: Yeah, I think this will break tests and stories

emtwo (commented):
This is awesome! I am just about to try it out and will come back and comment here.

app/experimenter/nimbus-ui/src/components/TableResults/index.test.tsx
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import TableResults from ".";
+
+describe("TableMetricSecondary", () => {
```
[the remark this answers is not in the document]
LZoog: Ah yep, copy pasta!

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/significance-neutral.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M20 12c0 .6-.4 1-1 1H5c-.6 0-1-.4-1-1s.4-1 1-1h14c.6 0 1 .4 1 1z" fill="#212529"/><mask id="a" maskUnits="userSpaceOnUse" x="4" y="11" width="16" height="2"><path d="M20 12c0 .6-.4 1-1 1H5c-.6 0-1-.4-1-1s.4-1 1-1h14c.6 0 1 .4 1 1z" fill="#fff"/></mask></svg>
```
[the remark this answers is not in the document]
LZoog: Ooo good question, I know `rapid` was using font awesome which we haven't used in `nimbus`. I used SVGs since they were a copy-and-paste from the icons shown in Figma, to be fair I didn't think to look at Bootstrap icons, but it looks like they're a different package/dependency...

https://getbootstrap.com/docs/4.5/extend/icons/

> While you won’t find a built-in icon library in Bootstrap, our separate Bootstrap Icons projects is a growing set of open source SVGs you can use. While they’re designed first and foremost to work with our components and documentation, you can use them in any project.

I suppose we could install the package, but it gives us another dependency, we won't use most of them, and I'm guessing there's icons in the navigation that aren't in this library (I could be wrong). The other option is to download them separately as SVGs in which case we might as well use what's in Figma? 🤔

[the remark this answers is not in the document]
emtwo: Oh cool. Using svg's makes sense in this case it seems then. Thanks for looking into it!

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -51,4 +51,71 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
...
+    <h3 className="text-uppercase h6">Hypothesis</h3>
+    <p>{experiment.hypothesis}</p>
+    <TableHighlights
+      featureConfig={experiment.featureConfig!}
+      results={analysis?.overall!}
+    />
+    <TableOverview />
+
+    <h2>Results</h2>
+    <h2 className="font-weight-bold">Results</h2>
```
emtwo, line 81: Results twice. Only need one!

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -7,41 +7,41 @@ import { RouteComponentProps, useParams } from "@reach/router";
...
-  const { loading, error, result } = useAnalysis(slug);
+  const { loading, error, result: analysis } = useAnalysis(slug);
 
   return (
-    <AppLayoutWithExperiment
-      title="Results"
-      testId="PageResults"
-      sidebar={false}
-    >
+    <AppLayoutWithExperiment title="Results" testId="PageResults">
```
emtwo, line 25: There's already the `Results` subtitle below. Perhaps we can called this `Analysis` instead? I know this ruins the component labeling a bit since it's called `PageResults`.

emtwo (approved):
I tried this out locally and here is my screenshot:
Screen Shot 2020-11-17 at 1 20 55 PM

The highlights table looks spot on! I'm wondering about some of the font sizing and spacing though. For example:
* Some of the subtitles are larger than the main `Testing Nimbus` title in the screenshot
* Some spacing between `Overview` and `Hypothesis`

This is kind of nitpicky though and we can take care of it in later PRs.

Overall lgtm

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  results: AnalysisData["overall"];
+};
+
+const TableHighlights = ({
+  featureConfig,
+  results = {},
+}: TableHighlightsProps) => {
+  const highlightMetricsList = [
+    {
+      value: `${featureConfig.slug.replace(/-/g, "_")}_ever_used`,
```
emtwo, line 31: I just realized something else here. We should be looking at primary probe sets not `featureConfig` in this case. This will need to be changed

emtwo (changes requested):
Just pending the switch to primary probe sets instead of feature config

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
emtwo: Also, It looks like this only adds 1 featureConfig. When this is switched to primary probe sets, there could be multiple primary probe sets and it needs to add all of them

jodyheavener (commented):
Looking good! I left you a few nitpicks and suggestions here and there.

Regarding the commented out code just to get tests passing, I think you should be able to do something like `/* istanbul ignore file */` at the top of each file, or `/* istanbul ignore next */` before a function, that you want temporarily excluded from tests

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+              </td>
+            </tr>
+          );
+        })}
+      </tbody>
+    </table>
+  );
+};
+
+export default TableHighlights;
```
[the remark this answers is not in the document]
emtwo: This works for me. I wouldn't use those badges though for `control` and `treatment` because we use badges elsewhere in the design to show `primary metric` `guardrail metric` etc.

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
LZoog: @emtwo Ohh I see, in `rapid` the highlights table uses `experimentData.features`, which I must have mistakenly thought was `featureConfig` in nimbus, of which there can only be one type of. I thought the primary/secondary probe sets would be displayed in the primary/secondary metric tables. I'll go take a look 🤔

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -7,41 +7,41 @@ import { RouteComponentProps, useParams } from "@reach/router";
...
-                  about this.
-                </Alert>
-              )}
-              {result && (
-                <p data-testid="analysis-data">
-                  Data: {JSON.stringify(result)}
-                </p>
+              <h3 className="h5">Monitoring</h3>
+              <p>
+                {/* TODO: replace '#' with dashboard URL after EXP-571 is complete */}
```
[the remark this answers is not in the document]
LZoog: Made a ticket for this, will update.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
LZoog: This makes me wonder if we should change the page component to be named `PageAnalysis` and make the URL `/analysis`! I'll change just the page title to "Analysis" for now.

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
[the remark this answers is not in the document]
LZoog: Oh man @jodyheavener coming in here with FE nitpicks 🤣 FWIW I did think we'd address these later in some FE polish ticket after design review, but I don't mind fixing them here. Thanks for the detailed review BTW!

> A little more space above the table would give it some breathing room (maybe my-4)
> Wrapping "All Users" in a span and giving it align-middle will help it vertically line up with the SVG

👍

> There appears to be a bottom margin on the last paragraph in each column, resulting in some funky extra space. If we remove that it looks a little more vertically-aligned

Huh I actually meant to remove that padding-bottom with `p-1 pb-lg-0 p-lg-3` but it doesn't work, this is why I don't like `!important`, `padding-bottom` should be a more specific property 🤦‍♀️ Changed to `pt-1 pl-1 pr-1 pt-lg-3 pr-lg-3 pl-lg-3` which looks nuts but looks nice. If you have a better class name set let me know, heh.

> It would be nice to have a border-bottom on the last row

I left this out since the Overview table is right below it and figured it would have a top/bottom border.

Per Marina's comment I'll leave out the pills.

> This could also benefit from a light background to give it some depth - I used `bg-light`

I actually think this looks nicer with the white background, otherwise it looks like it should have left/right borders. Maybe someone else will disagree. Actually, according to Figma, it looks like the entire page should have a light grey BG. Maybe users will like all the tables to be light grey though, I am not sure, let's table that for now? (see what I did there)

app/experimenter/nimbus-ui/src/images/info.svg
```diff
@@ -0,0 +1 @@
...8.037.037.075.037z" fill="#212529"/><mask id="a" maskUnits="userSpaceOnUse" x="0" y="0" width="18" height="18"><path fill-rule="evenodd" clip-rule="evenodd" d="M9.035 17.25c4.592 0 8.281-3.675 8.281-8.25S13.627.75 9.035.75C4.442.75.753 4.425.753 9s3.689 8.25 8.282 8.25zm.15-6.825c.301-.075 2.786-.975 2.786-2.925 0-1.275-.828-2.4-2.033-2.775a2.933 2.933 0 00-3.84 1.8c-.075.375.076.825.452.825.377.15.828-.075.979-.45.301-.75 1.204-1.2 1.957-.9.603.225.98.75.98 1.425 0 .825-1.28 1.425-1.733 1.575-.376.15-.602.6-.451.975.075.3.376.525.677.525.076 0 .15 0 .226-.075zM15.811 9c0-3.75-3.012-6.75-6.776-6.75-3.765 0-6.776 3-6.776 6.75s3.011 6.75 6.776 6.75c3.764 0 6.776-3 6.776-6.75zm-6.4 3.15c.075 0 .15.075.15.075.076.075.151.15.076.225.075.075.075.225.075.3a.68.68 0 01-.226.525.685.685 0 01-.527.225.685.685 0 01-.527-.225.68.68 0 01-.226-.525c0-.088.026-.15.048-.201.015-.037.028-.068.028-.099 0-.075.075-.15.15-.225.076-.075.15-.15.226-.15C8.808 12 8.96 12 9.11 12c.043.043.062.061.084.07.016.005.034.005.066.005.038 0 .057.019.076.038.019.018.037.037.075.037z" fill="#fff"/></mask><g mask="url(#a)"><path fill-rule="evenodd" clip-rule="evenodd" d="M0 0h18.07v18H0V0z" fill="#212529"/></g></svg>
```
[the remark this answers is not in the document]
LZoog: Yeah, Figma isn't being great for exports. I'll remove them.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -51,4 +51,71 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
...
+
+const AnalysisAvailable = ({
+  experiment,
+  analysis,
+}: {
+  experiment: getExperiment_experimentBySlug;
+  analysis: AnalysisData | undefined;
+}) => (
+  <>
+    <h3 className="text-uppercase h6">Hypothesis</h3>
```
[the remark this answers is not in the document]
LZoog: Heh, I'll take it out for now because I agree

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+                <p>{branch}</p>
+                <p className="h6">
+                  {userCountMetric[BRANCH_COMPARISON.ABSOLUTE]["point"]}{" "}
+                  participants ({userCountMetric["percent"]}%)
+                </p>
+              </th>
+              <th className="align-middle p-1 p-lg-3" scope="row">
+                All Users&nbsp;
+                <Info data-tip={SEGMENT_TIPS.ALL_USERS} />
+              </th>
```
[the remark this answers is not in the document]
LZoog: Eh yeah it was copypasta from rapid, I'll update, thanks for catching.

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <table data-testid="table-highlights">
+      <tbody>
+        {Object.keys(results).map((branch) => {
+          const userCountMetric =
+            results[branch]["branch_data"][METRIC.USER_COUNT];
+          return (
+            <tr key={branch} className="border-top">
+              <th className="align-middle p-1 p-lg-3" scope="row">
```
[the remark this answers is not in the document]
LZoog: Changed to `<td>`.

app/experimenter/nimbus-ui/src/components/ConfidenceInterval/index.tsx
```diff
@@ -0,0 +1,120 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* istanbul ignore file */
```
LZoog, line 5: @jodyheavener Thanks for this, it totally slipped my mind I could add this. I left the `TableVisualizationRow` file as-is though, it ended up being more effort than that one was worth adding `/* istanbul ignore next */` everywhere and I don't want us to miss removing them later.

app/experimenter/nimbus-ui/src/components/TableHighlights/index.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return (
+      <RouterSlugProvider mocks={[mock]}>
+        <TableHighlights
+          primaryProbeSets={data!.primaryProbeSets!}
+          results={mockAnalysis().overall}
+        />
+      </RouterSlugProvider>
+    );
+  })
+  .add("with multiple primary probe sets", () => {
```
LZoog, line 26: @emtwo 😎
image

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
LZoog: Updated!

emtwo (approved):
awesome!

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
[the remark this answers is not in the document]
jodyheavener: Thanks for updating this. I would still say that the page could benefit from some distinction between the table and the rest of the page, and unless we are also removing the margin between this table and the Overview table below it should prob still have its own bottom border. But both of these can be addressed later.

app/experimenter/nimbus-ui/src/components/ConfidenceInterval/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    case SIGNIFICANCE.NEUTRAL:
+      return (
+        <>
+          {padding}
+          {numbers}
+          {padding}
+        </>
+      );
+  }
+};
```
jodyheavener, line 55: It seems to me that these could just be strings and not JSX?

app/experimenter/nimbus-ui/src/components/TableOverview/index.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { RouterSlugProvider } from "../../lib/test-utils";
+import { withLinks } from "@storybook/addon-links";
+import { mockExperimentQuery } from "../../lib/mocks";
+import TableOverview from ".";
+// import { mockAnalysis } from "../../lib/visualization/mocks";
```
jodyheavener, line 11: keep/remove?

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+    <>
+      {tableLabel === TABLE_LABEL.HIGHLIGHTS ? (
+        <div key={metricKey}>{field}</div>
+      ) : (
+        <td key={metricKey} className="align-middle">
+          <div>{field}</div>
+        </td>
+      )}
+    </>
+  );
```
jodyheavener, line 203: No need to wrap this in a fragment:

```tsx
  return tableLabel === TABLE_LABEL.HIGHLIGHTS ? (
    <div key={metricKey}>{field}</div>
  ) : (
    <td key={metricKey} className="align-middle">
      <div>{field}</div>
    </td>
  );
```

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/significance-positive.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M18.7 10.7c-.2.2-.4.3-.7.3-.3 0-.5-.1-.7-.3L13 6.4V20c0 .6-.4 1-1 1s-1-.4-1-1V6.4l-4.3 4.3c-.4.4-1 .4-1.4 0-.4-.4-.4-1 0-1.4l6-6c.1-.1.2-.2.3-.2.2-.1.5-.1.8 0 .1.1.2.1.3.2l6 6c.4.4.4 1 0 1.4z" fill="#28A745"/></svg>
```
jodyheavener, line 1: This is of _little significance_ (😉), but I generally try to set SVG fills to `currentColor`, and let CSS style it (either from the `fill` property or inheriting from text `color`). That being said, I doubt we will need to change the colors of these negative/positive/neutral, just something to keep in mind. Maybe you already considered it.

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const getTableDisplayType = (
+  metricKey: string,
+  tableLabel: string,
+  isControl: boolean,
+): DISPLAY_TYPE => {
+  let displayType;
+  switch (metricKey) {
+    // case METRIC.USER_COUNT:
+    //   displayType = DISPLAY_TYPE.POPULATION;
+    //   break;
```
jodyheavener, line 16: keep/remove?

jodyheavener (approved):
Noted that you left `TableVisualizationRow` alone WRT removing commented stuff out. Few small suggestions/questions, but I'm also happy leaving them for a followup.

app/experimenter/nimbus-ui/src/components/TableOverview/index.stories.tsx
LZoog: Keeping, we'll need it when we build the TableOverview out. https://github.com/mozilla/experimenter/pull/3973#discussion_r524815256

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/significance-positive.svg
LZoog: Yeah, I really briefly thought about this but figured we wouldn't be changing the color of these. If we need to down the line, we can adjust them then.

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
LZoog: Keeping, we will need this in another table.

app/experimenter/nimbus-ui/src/components/ConfidenceInterval/index.tsx
LZoog: `numbers` is a JSX element and isn't a string.

jodyheavener: nvvvvm, `numbers` is JSX, my bad

jodyheavener: 4mvtwy

PR #3979, opened 2020-11-13
fix #3978 bug(nimbus): use browserSettings.update.channel for channel

### LINKED ISSUE #3978: Use browserSettings.update.channel for channel targeting
Due to weirdness that exists with how beta changes over to RC, we should use `browserSettings.update.channel` not `channel` for targeting.

In a future release, we could fix `channel` to point to the correct value.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-587)
┆Issue Number: EXP-587

k88hudson:
Because:
* There is a bug that causes channel to return "release" during the week
  of RC in beta https://bugzilla.mozilla.org/show_bug.cgi?id=1677226

This commit
* Change channel targeting expression to use browserSettings.update.channel


jaredlockhart (approved):
Oh that's handy 👍

PR #3993, opened 2020-11-16
fixes #3992 fix(legacy) corrected email subscription url

### LINKED ISSUE #3992: Broken "view it on Experimenter"-link in e-mail notification
STR:
- Receive an e-mail notification for a comment

ER:
- The link "view it on Experimenter" should lead to the comment on the experimenter page

AR:
- The link leads to a 404 page.

I noticed this on a notification I received today. The link from the e-mail is `https://experimenter.services.mozilla.com/experiments/reduce-fingerprintability-by-restricting-access-to-user-installed-fonts//#comment1225`

Note the double-slash towards the end, which leads to the problem. It should be a single slash instead.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-589)
┆Issue Number: EXP-589

tiftran:
Because

* There's an extra slash in the comment url

This commit:

* changes get_absolute_url for ExperimentComment to use urljoin

### DIFF
--- a/app/experimenter/experiments/models/legacy.py
+++ b/app/experimenter/experiments/models/legacy.py
@@ -1285,9 +1285,8 @@ class ExperimentComment(ExperimentConstants, models.Model):
 
     objects = ExperimentCommentManager()
 
-    @property
     def get_absolute_url(self):
-        return urljoin(self.experiment.experiment_url, "#/comment{id}".format(id=self.id))
+        return f"{self.experiment.experiment_url}#comment{self.id}"
 
     class Meta:
         verbose_name = "Experiment Comment"
--- a/app/experimenter/experiments/models/legacy.py
+++ b/app/experimenter/experiments/models/legacy.py
@@ -1287,9 +1287,7 @@ class ExperimentComment(ExperimentConstants, models.Model):
 
     @property
     def get_absolute_url(self):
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/models/legacy.py
```diff
@@ -1287,9 +1287,7 @@ class ExperimentComment(ExperimentConstants, models.Model):
 
     @property
     def get_absolute_url(self):
-        return "{comment.experiment.experiment_url}/#comment{comment.id}".format(
-            comment=self
-        )
+        return urljoin(self.experiment.experiment_url, "#/comment{id}".format(id=self.id))
```
jaredlockhart, line 1290: is `#/` right?

The url should look like `experimenter/experiments/experiment-slug/#comment123`?

Maybe it'd be clearer if there was a test on this method?

tiftran: you're right, will fix

jaredlockhart: If urljoin strips out the # then don't use urljoin at all just format string it `f"{self.experiment.experiment_url}#comment{id}`

jaredlockhart (approved):
Parfait 🎉

PR #3998, opened 2020-11-16
fix #3995 feat(nimbus): add max primary probe sets constant to config graphql

### LINKED ISSUE #3995: Store max primary probe sets as a constant and send to frontend
We're constraining the maximum number of primary probe sets to an arbitrary constant, right now it's 2, so we should add that to nimbus constants, use that constant in the backend and add it to the frontend configuration gql endpoint.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-591)
┆Issue Number: EXP-591

jaredlockhart:
Because

* We can enforce the maximum number of primary probe sets right on the frontend

This commit

* Exposes it as a constant in the graphql config endpoint


bbangert (approved):
LGTM!

PR #4000, opened 2020-11-16
fix #3997 chore(project): remove legacy rapid kinto tasks

### LINKED ISSUE #3997: Remove messaging-experiments collection
The messaging-experiments collection has gone to a farm upstate, so we gotta take references to it.  We should just remove all the legacy kinto tasks along with it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-593)
┆Issue Number: EXP-593

jaredlockhart:
Because

* The collection that the legacy rapid kinto tasks has been removed from remote settings

This commit

* Removes the reference to the collection
* Removes the relevant tasks
* removes the relevant settings


bbangert (approved):
Nice to drop the outdated stuff, I assume v4 and such are going to be removed as well?

PR #4025, opened 2020-11-18
fixes #3854 feat(legacy): send a subscriber email for changes

### LINKED ISSUE #3854: Send a subscriber email on changelogs
Let's add a new email that notifies subscribers when someone edits an experiment.

     Subject: <user email> edited <experiment name>
     Body:
    <User Email>:
    <change table formatted same way as changelog modal>
    <link to experiment_url>

We can add a function in [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/email/legacy.py) called `send_changelog_email`

And trigger it [here](https://github.com/mozilla/experimenter/blob/b20fc194b84086e114c9984e0edaa517c11c92c8/app/experimenter/experiments/changelog_utils/legacy.py#L245)

and that should cover all edits through the API and the forms.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-532)
┆Issue Number: EXP-532

tiftran:
Because:

* QA and other stakeholders would like to be notified of changes

This commit:

* has an email template for changes
* makes a call to send an email after a changelog is created


app/experimenter/experiments/email/legacy.py
```diff
@@ -90,6 +90,20 @@ def send_experiment_comment_email(comment):
     )
 
 
+def send_experiment_change_email(change):
+
+    # Launch has its own separate email
+    if str(change) not in ("Launched Experiment"):
```
jaredlockhart, line 96: If you wanna do tuple membership check, you need a comma here:

`("Launched Experiment")` evaluates as an expression, not a tuple: `"LaunchedExperiment"` so it's checking that `str(change)` string is a substring of `"LaunchedExperiment"`, which might also work but I don't think is exactly what you meant by the way it's written.

If you wanna do membership you gotta do

`if str(change) not in ("Launched Experiment",):`

It's kinda dumb in Python that `(a)` is an expression and the parens just go away vs `(a,)` which is a tuple of length 1.

jaredlockhart (approved):
Looks good @tiftran !  Might wanna look at that launched check, but otherwise is good 👍

app/experimenter/experiments/email/legacy.py
tiftran: oh no, this was not what I wanted 🤦‍♀️ results of tinkering... will fix

PR #4026, opened 2020-11-18
fix #3996 feat(nimbus): move ready for review to experiment by slug query

### LINKED ISSUE #3996: Merge experimentReadyForReviewBySlug values into experimentBySlug query
Let's move the `ready` and `message` values from the `experimentReadyForReviewBySlug` GQL query into the `experimentBySlug` query.

Why? We perform both queries on all Edit page loads anyway, so doing this will:

* Remove a network request
* Reduce some jankiness in the sidebar when the experiment query finishes before the review query
* Remove the need to double mock each test/storybook that uses both queries

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-592)
┆Issue Number: EXP-592

jaredlockhart:
Because

* It will save extra queries to fold the ready for review check into the experiment by slug query

This commit

* Folds the ready for review check into the experiment by slug query

### DIFF
--- a/app/experimenter/experiments/api/v5/mutation.py
+++ b/app/experimenter/experiments/api/v5/mutation.py
@@ -28,7 +28,7 @@ from experimenter.experiments.api.v5.serializers import (
     NimbusProbeSetUpdateSerializer,
     NimbusStatusUpdateSerializer,
 )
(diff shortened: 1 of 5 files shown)


bbangert (approved):
Nice! I like the way you folded that into the overall experiment type.

PR #4030, opened 2020-11-18
fixes #4029 feat(nimbus): add targeting for google homepage users

### LINKED ISSUE #4029: Add new targeting for google homepage users
We need this to support the homepage remediation CFR experiment.

The targeting JEXL: `!homePageSettings.isDefault && homePageSettings.isCustomUrl && homePageSettings.urls[.host == 'google.com']|length > 0`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-618)
┆Issue Number: EXP-618

ncloudioj:


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -65,6 +65,18 @@ class NimbusTargetingConfig:
     ),
 )
 
+TARGETING_HOMEPAGE_GOOGLE = NimbusTargetingConfig(
+    name="Homepage set to google.com",
+    slug="homepage_google_dot_com",
+    description="Users with their Homepage set to google.com",
+    targeting=(
```
k88hudson, line 72: I think you probably want to add `TARGETING_US_ONLY.targeting` in here

ncloudioj: Done

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -65,6 +65,20 @@ class NimbusTargetingConfig:
     ),
 )
 
+TARGETING_HOMEPAGE_GOOGLE = NimbusTargetingConfig(
+    name="Homepage set to google.com",
+    slug="homepage_google_dot_com",
+    description="US users (en) with their Homepage set to google.com",
+    targeting=(
+        "localeLanguageCode == 'en' && "
+        "region == 'US' && "
```
jaredlockhart, line 74: Coulda embedded an existing targeting expression like here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L40

but this is also totally fine 👍

k88hudson (approved):

app/experimenter/experiments/constants/nimbus.py
ncloudioj: Let's leave it as it is then 😜. Will revisit/refactor those entries once the targeting combination gets implemented.

jaredlockhart: Yeah it's just a thought, I'm still not sure whether that's a good pattern or not.  On the one hand it could help keep things consistent and reduce duplication, on the other hand we could wind up making this big tangled mess of targeting configs that depend on each other and changing one could break everything.  So we'll just have to see how it looks the more we use it.

PR #4031, opened 2020-11-18
fix #3976 chore(visualization): Copy overview table from `rapid` into TableOverview component.

### LINKED ISSUE #3976: (visualization) Copy overview table from `rapid` into TableOverview component
EXP-461 creates a placeholder for the `TableOverview` component which is displayed on the Results page. Copy over this table from the `rapid` project into this component and add it and its various states into Storybook. Note that the experiment data structure differs slightly between the projects.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-585)
┆Issue Number: EXP-585

emtwo:
Building off of https://github.com/mozilla/experimenter/pull/3973

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableOverview/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableOverview/index.stories.tsx
@@ -8,14 +8,41 @@ import { RouterSlugProvider } from "../../lib/test-utils";
 import { withLinks } from "@storybook/addon-links";
 import { mockExperimentQuery } from "../../lib/mocks";
 import TableOverview from ".";
-// import { mockAnalysis } from "../../lib/visualization/mocks";
-
-const { mock } = mockExperimentQuery("demo-slug");
+import { mockAnalysis } from "../../lib/visualization/mocks";
 
 storiesOf("visualization/TableOverview", module)
   .addDecorator(withLinks)
-  .add("basic", () => (
-    <RouterSlugProvider mocks={[mock]}>
-      <TableOverview />
-    </RouterSlugProvider>
-  ));
+  .add("basic, with one primary probe set", () => {
+    const { mock, data } = mockExperimentQuery("demo-slug");
+    return (
+      <RouterSlugProvider mocks={[mock]}>
+        <TableOverview experiment={data!} results={mockAnalysis().overall} />
+      </RouterSlugProvider>
+    );
+  })
+  .add("with multiple primary probe sets", () => {
+    const { mock, data } = mockExperimentQuery("demo-slug", {
+      primaryProbeSets: [
+        {
+          __typename: "NimbusProbeSetType",
+          slug: "picture-in-picture",
+          name: "Picture-in-Picture",
+        },
+        {
+          __typename: "NimbusProbeSetType",
+          slug: "feature-b",
+          name: "Feature B",
+        },
+        {
+          __typename: "NimbusProbeSetType",
+          slug: "feature-c",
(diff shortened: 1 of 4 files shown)


app/experimenter/nimbus-ui/src/components/TableOverview/index.tsx
```diff
@@ -3,9 +3,81 @@
...
+  | getConfig_nimbusConfig["firefoxMinVersion"]
+  | getConfig_nimbusConfig["channels"]
+  | getConfig_nimbusConfig["targetingConfigSlug"];
+
+const TableOverview = ({ experiment }: TableOverviewProps) => {
+  const { firefoxMinVersion, channels, targetingConfigSlug } = useConfig();
+
+  return (
+    <div className="mb-5" data-testid="table-overview">
+      <table className="table text-left h5 m-0">
```
LZoog, line 26: If you take `h5` out here, you can take `h6` off of all the `<p>` tags.

app/experimenter/nimbus-ui/src/components/TableOverview/index.tsx
```diff
@@ -3,9 +3,81 @@
...
+        <tbody>
+          <tr>
+            <th scope="col">
```
LZoog, line 29: I think these should be `<td>` since every cell shouldn't be a table header. Then, "Targeting", "Probe Sets", and "Owner" can be in `<h3 className="h6">` I believe.

app/experimenter/nimbus-ui/src/components/TableOverview/index.tsx
```diff
@@ -3,9 +3,81 @@
...
+            </th>
+          </tr>
+        </tbody>
+      </table>
+      <hr className="mt-0 mb-5" />
+    </div>
+  );
+};
+
+const displayConfigLabel = (
```
LZoog, line 76: Since we're using `displayConfigLabelOrNotSet` in `TableSummary`, it could be nice to extract this out to a file under `lib/` since we'll use it elsewhere, but it's not a 1:1 exactly with that function so it might make more sense to wait until we're using this in another table.

app/experimenter/nimbus-ui/src/components/TableOverview/index.stories.tsx
```diff
@@ -8,14 +8,40 @@ import { RouterSlugProvider } from "../../lib/test-utils";
 import { withLinks } from "@storybook/addon-links";
 import { mockExperimentQuery } from "../../lib/mocks";
 import TableOverview from ".";
-// import { mockAnalysis } from "../../lib/visualization/mocks";
+import { mockAnalysis } from "../../lib/visualization/mocks";
 
-const { mock } = mockExperimentQuery("demo-slug");
+const { mock, data } = mockExperimentQuery("demo-slug");
```
LZoog, line 13: For better scoping, this should probably be moved into the first story.

LZoog (commented):
Just a couple minor things, looking great! :)

app/experimenter/nimbus-ui/src/components/TableOverview/index.tsx
emtwo: Yes, I did get the idea from there but it's not exactly the same which is why I didn't pull it out. We could do further refactoring later.

LZoog: Sounds good to me!

LZoog (approved):
LGTM!

PR #4043, opened 2020-11-18
chore(project): disable deprecated tests

jaredlockhart:
Because

* We are now migrating all visualization work out of rapid and into nimbus
* All of the legacy/core ui is in maintenance mode
* Tests take up a lot of circle credits

This commit

* Disables all but the end to end integration tests for the legacy workflow
* Disables all of the rapid/visualization js tests/linting
* Disables the rapid/visualization dev yarn processes
* Decreases the size of the make check circle instance
* This should make the local dev lighter, make our tests faster, and reduce circle usage


jrbenny35 (changes requested):
f

jrbenny35 (approved):

PR #4045, opened 2020-11-19
fixes #3977 feat(nimbus): Add Results table to Results page

### LINKED ISSUE #3977: (visualization) Copy results table from `rapid` into TableResults component
EXP-461 creates a placeholder for the `TableResults` component which is displayed on the Results page. Copy over this table from the `rapid` project into this component and add it and its various states into Storybook. Note that the experiment data structure differs slightly between the projects.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-586)
┆Issue Number: EXP-586

LZoog:
This commit:
* Adds the Results table, copied from the 'rapid' legacy project, into the Results page

Because:
* We want to be able to see this table when the analysis data is available.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
@@ -40,7 +40,7 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
                 </LinkExternal>{" "}
                 to view the live monitoring dashboard.
               </p>
-              <h3 className="h5 my-3">Overview</h3>
+              <h3 className="h5 mb-3 mt-4">Overview</h3>
 
               {error && <AnalysisFetchError />}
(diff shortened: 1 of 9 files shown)


emtwo (commented):
Looks great! I think we may need to tweak spacing still once we look at the full page altogether but I'm sure we will need a polish.

The only thing that might be good to fix here is font size. I noticed while working on the Primary Metrics PR that all the fonts are too large compared to the rest of the page and that is the case here too. I'll attach a screenshot of what I'm looking at

LZoog:
@emtwo sure thing, let me know what you think should be reduced in size... It looks good to me (I made some more tweaks on spacing and such here) but I'm probably missing it. :) 

image

emtwo (approved):
Ignore my spacing/font size comment. We must have made changes at some point to fix it and I missed them.

Looks great!

PR #4057, opened 2020-11-19
fix #4009 feat(nimbus): send default branches for new experiments in graphql

### LINKED ISSUE #4009: Server should send an empty control and feature branch by default
We want the branches form to display a control and feature branch by default.

See https://github.com/mozilla/experimenter/pull/3886#discussion_r520084787

"The pattern we generally settled on in the past was to have the server spit out the default empty configuration, ie an empty control branch and an empty treatment branch, and that way the UI can just take what the server gives it and spit it out verbatim without needing to do any custom preparation logic."

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-603)
┆Issue Number: EXP-603

jaredlockhart:
Becasue

* We want the branch form to be prepopulated with an empty control and treatment branch

This commit

* Adds an empty control and treatment branch to the graphql serializer so the form has some empty defaults

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -120,6 +120,16 @@ class NimbusExperimentType(DjangoObjectType):
         model = NimbusExperiment
         exclude = ("branches", "probe_sets")
 
+    def resolve_reference_branch(self, info):
+        if self.reference_branch:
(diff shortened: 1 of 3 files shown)


bbangert (approved):
All those little things.... 😀

PR #4059, opened 2020-11-20
fixes #3737 feat(legacy): Added QA Launch Changelog for QA Launches

### LINKED ISSUE #3737: [Legacy] Representation for QA recipe Launches
Experimenter needs a nice way to represent experiments that are launched for QA before their actual launch.

We should not set the experiment to "Live" when it's launched for QA but instead, we should leave it as "Accepted to Normandy" and then have a changelog for the day it was launched for QA

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-477)
┆Issue Number: EXP-477

tiftran:
Because:

* QA launches aren't actual launches which causes confusion when
experiment goes "Live" for QA Launches

This Commit:

* Introduces new changelog that gets create when the "qaOnly" filter object is
present for the recipe

### DIFF
--- a/app/experimenter/normandy/tasks.py
+++ b/app/experimenter/normandy/tasks.py
@@ -27,6 +27,8 @@ STATUS_UPDATE_MAPPING = {
 logger = get_task_logger(__name__)
 metrics = markus.get_metrics("experiments.tasks")
 
+QA_LAUNCH_MESSAGE = "Launched for QA"
+
 
 @app.task
 @metrics.timer_decorator("update_recipe_ids_to_experiments.timing")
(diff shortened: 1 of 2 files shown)


app/experimenter/normandy/tasks.py
```diff
@@ -108,20 +110,43 @@ def update_launched_experiments():
     metrics.incr("update_launched_experiments.completed")
 
 
+def is_qaOnly(recipe_data):
+    if "filter_object" in recipe_data:
+        filter_objects = {f["type"]: f for f in recipe_data["filter_object"]}
```
jaredlockhart, line 115: You know if you take `: f` out of this, it'll be a set comprehension instead of a dict comprehension and do the same thing as what you're doing here, checking membership of the key.  Not a requirement to change, just a tip ;)

app/experimenter/normandy/tasks.py
```diff
@@ -108,20 +110,43 @@ def update_launched_experiments():
     metrics.incr("update_launched_experiments.completed")
 
 
+def is_qaOnly(recipe_data):
```
jaredlockhart, line 113: wx4a325srgz51

jaredlockhart (approved):
Looks good 👍  What's with the mixed casing?

app/experimenter/normandy/tasks.py
tiftran: 😂 was trying to be consistent with the normandy recipe value

PR #4060, opened 2020-11-20
fixes #3795 feat(nimbus): Add secondary metric table to Results page + other vis table tweaks

### LINKED ISSUE #3795: GraphQL mutation to mark for review and restrict prior mutations

LZoog:
This commit:
* Copies the secondary metric table from the legacy UI into nimbus UI results page
* Tweaks all visualization tables for consistency in classes
* Adds slightly better typing in related files
* Removes unneeded GQL mocks in some storybook stories

Because:
* We want to see the secondary probe sets table on the Results page and want the tables to appear consistently styled

---

@emtwo I did a little extra here with the styles, feel free to push back on anything. :)

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
@@ -29,7 +29,7 @@ const getHighlightMetrics = (
   const highlightMetricsList = [...HIGHLIGHTS_METRICS_LIST];
   probeSets.forEach((probeSet) => {
     highlightMetricsList.unshift({
-      value: `${probeSet!.slug.replace(/-/g, "_")}_ever_used`,
+      value: `${probeSet!.slug}_ever_used`,
       name: `${probeSet!.name} conversion`,
       tooltip: METRICS_TIPS.CONVERSION,
     });
@@ -45,7 +45,7 @@ const TableHighlights = ({
   const highlightMetricsList = getHighlightMetrics(primaryProbeSets);
 
   return (
-    <table data-testid="table-highlights" className="mt-4">
+    <table data-testid="table-highlights" className="table mt-4 mb-0">
       <tbody>
         {Object.keys(results).map((branch) => {
           const userCountMetric =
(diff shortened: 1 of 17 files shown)


app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -29,7 +29,7 @@ const getHighlightMetrics = (
   const highlightMetricsList = [...HIGHLIGHTS_METRICS_LIST];
   probeSets.forEach((probeSet) => {
     highlightMetricsList.unshift({
-      value: `${probeSet!.slug.replace(/-/g, "_")}_ever_used`,
+      value: `${probeSet!.slug}_ever_used`,
```
LZoog, line 32: @emtwo I'll double check this in the morning, but looks like the slugs for probe sets are already snake cased. Otherwise, I'll change this in all 4 tables.

LZoog:
I'm showing 100% line coverage here but we should really increase branch line coverage across all of the visualization tables. I'll file a ticket for this and add it to our "nice-to-haves" epic.

image

emtwo (commented):
This looks good to me! Thanks for the extra polish in the other tables. 
Let me know when you've looked into the slug `-` vs. `_` and what you find. I'll come back and r+ after that in case you need to make more changes.

emtwo:
> @emtwo I did a little extra here with the styles, feel free to push back on anything. :)

I think it looks pretty! :D

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
LZoog: @jaredlockhart is working on the Metrics page - it seems like I could add `.replace(/-/g, "_")` in the four places we need `_ever_used` as a safeguard either way, but I want to make sure our mocked data reflects what we expect to see in production so I'll wait for his input. 👀  So Jared, `🐍 || 🍡`?

LZoog: @emtwo Jared got back to me, says we should file a ticket for enforcing snake-case on probe set slugs since Jetstream uses snake case. So, this should be g2g.

emtwo: awesome! thanks

emtwo (approved):

PR #4065, opened 2020-11-20
fixes #4046 fix(legacy): switched to format_ndt_normandy_urls in temp…

### LINKED ISSUE #4046: [Nimbus UI] Cannot read property 'name' of undefined
[https://sentry.prod.mozaws.net/operations/experimenter-stage/issues/10074552/](https://sentry.prod.mozaws.net/operations/experimenter-stage/issues/10074552/)

We should also try to set up source maps in the JS

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-626)
┆Issue Number: EXP-626

tiftran:
…late

Because:

* it was format_dc_normandy_urls before which is wrong

This Commit:

* switches to format_ndt_normandy_urls in normandy template


jaredlockhart (approved):
;)

PR #4095, opened 2020-12-02
fix(nimbus): allow apollo codegen to pass through custom scalars

jodyheavener:
Our GraphQL schema has a few types that don't directly translate to TypeScript, and as a result they end up as type `any`, which isn't the end of the world but also isn't great.

With this change we're now allowing these scalar types to pass through and bridging them with our own declarations.


jaredlockhart (approved):
Oh neat!  Seems good to me 👍

PR #4096, opened 2020-12-02
fixes #4083 fix(legacy): Ensure prefs from multipref are in changelog

### LINKED ISSUE #4083: Pref changes are no longer recorded in History if the same user edited them twice in a row
**[Notes]**
- The issue is reproducible if any of the pref "Branch", "Name", "Type" or "Value" is changed twice in a row by the same user.
- This behavior is encountered quite often because the user which creates an Experimenter page, usually edits the pref branches at least two times until the Experimenter page is moved to "Ready for Sign off" state. 
- Since these changes are not recorded, a person (in particular) who checks the Experimenter page can miss that the pref branches have been changed. QA has almost missed these types of changes a few times.

**[Affected versions]:**
- Firefox 83.0 Release
- Chrome 87.0.4280.66 

**[Affected Platforms]:**
- All Windows;
- All macOS;
- All Linux;

**[Prerequisites]:**
- You are logged in your LDAP account.
- Have an experiment created on the "https://stage.experimenter.nonprod.dataops.mozgcp.net/" website.
- The Branches section is already completed during the Experiment creation.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites and navigate to "https://stage.experimenter.nonprod.dataops.mozgcp.net/".
2. Open the experiment from prerequisites and click the "Edit" button from "Branches" section.
3. Edit the pref Name, Type or Value and click the "Save Draft" button.
4. Repeat step 2 and step 3.
5. Observe the "History" section.

**[Expected result]:**
- The pref changes are recorded in the History section for both changes.

**[Actual result]:**
- No new pref changes are recorded in the History section.

**[Additional Notes]:**
- The issue is NOT reproducible if multiple users edit the pref "Branch", "Name", "Type" or "Value" in turn.
- The issue is NOT reproducible if there is One Pref for all branches.
- The Pref changes are recorded in the History section only if the branch "Size", "Name", or "Description" is edited.
- Here is a link with the screen recording of the issue: [Link](https://drive.google.com/file/d/1KdJBPxwPlMXlbvXEt3YgUhQqEW9DCfrH/view?usp=sharing)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-645)
┆Issue Number: EXP-645

tiftran:
Because:

* Changelogs need to capture multipref changes

This commit:

* Moves the update_changelog method out of the DesignBaseSerializer
* Implements the update method for all design serializers as needed

### DIFF
--- a/app/experimenter/experiments/api/v2/serializers.py
+++ b/app/experimenter/experiments/api/v2/serializers.py
@@ -229,7 +229,7 @@ class ExperimentDesignBaseSerializer(
 
         return unique_names and non_empty
 
-    def update(self, instance, validated_data):
+    def update_instance(self, instance, validated_data):
         try:
             with transaction.atomic():
                 variants_data = validated_data.pop("variants", [])
@@ -254,8 +254,6 @@ class ExperimentDesignBaseSerializer(
                 if removed_ids:
                     ExperimentVariant.objects.filter(id__in=removed_ids).delete()
 
-            self.update_changelog(instance, validated_data)
-
             return instance
         except IntegrityError:
             error_string = (
@@ -264,6 +262,11 @@ class ExperimentDesignBaseSerializer(
             error = [{"name": error_string}] * len(variants_data)
 
             raise serializers.ValidationError({"variants": error})
+
+    def update(self, instance, validated_data):
+        instance = self.update_instance(instance, validated_data)
+        self.update_changelog(instance, validated_data)
+        return instance
 
 
 class ExperimentDesignRolloutPreferenceSerializer(
@@ -309,8 +312,10 @@ class ExperimentDesignPrefRolloutSerializer(
         return data
 
     def update(self, instance, validated_data):
+        validated_data_copy = validated_data.copy()
         preferences_data = validated_data.pop("preferences", [])
-        instance = super().update(instance, validated_data)
+
+        instance = super().update_instance(instance, validated_data)
 
         if preferences_data:
             existing_preference_ids = set(
@@ -332,6 +337,7 @@ class ExperimentDesignPrefRolloutSerializer(
             if removed_ids:
                 RolloutPreference.objects.filter(id__in=removed_ids).delete()
 
+        self.update_changelog(instance, validated_data_copy)
         return instance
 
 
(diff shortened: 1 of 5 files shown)


app/experimenter/experiments/api/v2/serializers.py
```diff
@@ -254,8 +254,6 @@ def update(self, instance, validated_data):
                 if removed_ids:
                     ExperimentVariant.objects.filter(id__in=removed_ids).delete()
 
-            self.update_changelog(instance, validated_data)
```
jaredlockhart, line 257: This is a little bit wonky, taking it out of base and then adding it back into every child?  We should be able to keep it in here and not have to add it to every subclass?

app/experimenter/experiments/api/v2/serializers.py
```diff
@@ -309,6 +307,7 @@ def validate(self, data):
         return data
 
     def update(self, instance, validated_data):
+        validated_data_copy = validated_data.copy()
```
jaredlockhart, line 310: Since the pop is destructive, depending on the ordering of parent/child methods they could wind up with `validated_data` with or without `variants`.  Maybe it'd help if instead of popping from the passed in reference, you pop from the copy, so any parent/child invocations always get the unmodified total `variants_data`?

jaredlockhart (changes requested):
There should be a way to do this without having to implement it in every child class?

tiftran:
@jaredlockhart did some more poking and I think I understand the underlying problem better, but unsure of how to address it.

For the Changelog to behave correctly, a complete set of changed_values needs to be present. In addition to that, the changelog assumes when it's called the experiment is done updating itself, since it serializes the experiment at that point for comparison. 

Since `self.update_changelog` is called in the base serializer, for the cases of multipref and rollouts, the preferences aren't saved at that point.  Preferences are popped off of the validated data, then calls `super().update()` and then saves the preferences?

While I agree that my current method of removing the self.update_changelog call out of the base serializer and calling it for every serializer seems awful, I'm not sure how to go about this?

app/experimenter/experiments/api/v2/serializers.py
```diff
@@ -265,6 +263,11 @@ def update(self, instance, validated_data):
 
             raise serializers.ValidationError({"variants": error})
 
+    def update(self, instance, validated_data):
+        instance = self.update_instance(instance, validated_data)
```
jaredlockhart, line 267: You're calling `super.update()` inside of here which is a bit wonky and indirect, can you move the super call into here and just call `update_instance` after?

jaredlockhart (changes requested):
This is looking much better, I think if you just bump that super call out it should be good to go!

jaredlockhart (approved):
Alright nevermind this is great, good job getting it into the base class 👍 Thanks @tiftran !

PR #4070, opened 2020-11-20
fixes #4014 feat(nimbus): Add monitoring dashboard URL to results page

LZoog:

Issue #4058, opened 2020-11-19
Add analysis data polling to Results page


data-sync-user:
If the analysis data is unavailable (`analysis.show_analysis` equaling `false` in the results page component), we want to poll the endpoint for this data every 30 seconds.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-637)
┆Issue Number: EXP-637

PR #4108, opened 2020-12-03
fixes #4067 feat(nimbus): move analysis lookup to AppLayoutWithExperiment level

### LINKED ISSUE #4067: Move analysis data lookup
In EXP-498 we had:

> If the visualization data is not yet available (`analysis.show_analysis` will be `false`), disable the "Results" link and show the greyed-out state, replace the icon with the :no_entry_sign: icon, and add "Experiment results not yet ready" text

In order to achieve this we need to know the analysis status, which we don't have access to in the sidebar, because at the moment the analysis data lookup is embedded in the [PageResults](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx) component, but it needs to be moved further up, likely into **AppLayoutWithExperiment**, and passed down to **PageResults** and **AppLayoutWithSidebar**. Only then will we be able disable that sidebar item correctly.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-642)
┆Issue Number: EXP-642

jodyheavener:
Closes #4067

Because:

We currently fetch experiment analysis data on the the experiment results page, at the `PageResults` component level, but we need need to be able to update the contents of `AppLayoutWithSidebar` based on analysis status.

This PR:

Moves the analysis lookup request up a few levels and into `AppLayoutWithExperiment`, where it can then pass the data down to both `PageResults` _and_ `AppLayoutWithSidebar`.

---

This change felt a little clunky as it apparently touches a number of areas, you'll see a good chunk of the diff is just shuffling things around; please do let me know if you see anything that feels like it could be refactored.

Also, while I'm following a pattern of passing props down, I'm starting to think using Context might be beneficial for things like analysis and review. If others agree I'll make an issue and throw it into the Maintenance tag.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
@@ -2,16 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { RouteComponentProps, useParams } from "@reach/router";
 import AppLayoutWithSidebar from "../AppLayoutWithSidebar";
 import HeaderExperiment from "../HeaderExperiment";
 import PageLoading from "../PageLoading";
 import PageExperimentNotFound from "../PageExperimentNotFound";
-import { useExperiment } from "../../hooks";
+import { useAnalysis, useExperiment } from "../../hooks";
 import { getExperiment_experimentBySlug } from "../../types/getExperiment";
 import AppLayout from "../AppLayout";
 import { NimbusExperimentStatus } from "../../types/globalTypes";
+import { AnalysisData } from "../../lib/visualization/types";
 
 type AppLayoutWithExperimentChildrenProps = {
   experiment: getExperiment_experimentBySlug;
@@ -19,6 +20,7 @@ type AppLayoutWithExperimentChildrenProps = {
     isMissingField: (fieldName: string) => boolean;
     refetch: () => void;
   };
+  analysis?: AnalysisData;
 };
 
 type AppLayoutWithExperimentProps = {
@@ -33,6 +35,15 @@ type AppLayoutWithExperimentProps = {
 
 export const POLL_INTERVAL = 30000;
 
+const experimentActive = (status: NimbusExperimentStatus | null): boolean => {
+  if (!status) return false;
+
+  return [
+    NimbusExperimentStatus.LIVE,
+    NimbusExperimentStatus.COMPLETE,
+  ].includes(status);
+};
+
 const AppLayoutWithExperiment = ({
   children,
   testId,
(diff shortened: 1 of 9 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -44,11 +56,40 @@ const AppLayoutWithExperiment = ({
...
+    execute: fetchAnalysis,
+    result: analysis,
+    loading: analysisLoading,
+  } = useAnalysis();
+
+  useEffect(() => {
+    if (
+      !analysisFetched &&
+      !experimentLoading &&
+      experimentLocked(experiment?.status)
```
LZoog, line 81: It looks like we have a nearly identical `experimentLocked` in `AppLayoutWithSidebar`.

Was going to suggest DRYing that up, but I think in _this_ instance, we only want to check if the experiment is live or complete, not "accepted" since results will never be available in that state, in which case the name `experimentLocked` may also not make sense (I like it for accepted/live/complete statuses though).

image

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -17,43 +16,32 @@ import TableMetricPrimary from "../TableMetricPrimary";
...
-    </AppLayoutWithExperiment>
-  );
-};
+const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
+  <AppLayoutWithExperiment title="Analysis" testId="PageResults">
+    {({ experiment, analysis }) => (
+      <>
+        <h3 className="h5 mb-3">Monitoring</h3>
+        <p>
+          {/* TODO: start/end dates with EXP-607 */}
```
LZoog, line 25: Hmm might need to rebase, looks like [these changes](https://github.com/mozilla/experimenter/pull/4070/files) would be overwritten?

LZoog (commented):
I need to go through this more thoroughly but did a first quick pass.

One thing to keep in mind is we're going to need to (well, it's a nice-to-have) add polling when the user is on the Results page for that analysis data. I think we can just check what URL the user is on in the sidebar component and refetch based on that, but I don't think we have a ticket made for it yet.

> Also, while I'm following a pattern of passing props down, I'm starting to think using Context might be beneficial for things like analysis and review. If others agree I'll make an issue and throw it into the Maintenance tag.

I'm thinking the same for `experiment` in the branch I'm working on. I don't want to over-complicate that though with our numerous "Save" buttons... I think it's worth creating an issue for though, something like evaluating if Context is the right fit in a couple places and implement if so. (Actually, would you mind creating a ticket for polling analysis data in `Maintenance` fix version as well while you're at it?)

Edit: the polling ticket for that page was created a couple weeks ago, #4058

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
jodyheavener: Related: https://jira.mozilla.com/browse/EXP-653

I've removed the Accepted criteria!

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
jodyheavener: Whoops. Changed it back.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -137,7 +125,7 @@ const AnalysisUnavailable = ({
   error,
 }: {
   experiment: getExperiment_experimentBySlug;
-  error: Error | undefined;
+  error: boolean;
```
jodyheavener, line 128: @LZoog with this change we no longer pass an actual error down to PageResults, but in practice the absence of analysis data (remember: even if the analysis isn't ready, the data os still returned) is indicative of an error.

LZoog: Couldn't it just be unavailable but not error? That's why I've got `{error ? <AlertError /> : <AlertUnavailable />}`. 🤔 When we do the route protection stuff it shouldn't matter either way so meh either way.

LZoog (approved):
LGTM!

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
jodyheavener: If it's unavailable the request would still complete and the returned object would contain `show_analysis: false`. But if `analysis` is straight up `undefined` that's an error.

PR #4119, opened 2020-12-04
fix #4111: fix(visualization) Send correct retention data and handle missing data in frontend.

### LINKED ISSUE #4111: General Application Error on Nimbus Results page
This is likely due to the missing `retained` metric. Made the assumption it would be available from Jetstream but alas. It is not. Will need to fix

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-661)
┆Issue Number: EXP-661

emtwo:


emtwo:
Couple of screenshots of recently completely experiments for reference:

Screen Shot 2020-12-04 at 3 11 25 PM
Screen Shot 2020-12-04 at 3 11 42 PM

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
```diff
@@ -162,43 +162,56 @@ const TableVisualizationRow: React.FC<{
...
-    case DISPLAY_TYPE.CONVERSION_COUNT:
-      field = conversionCountField(count, userCountMetric);
-      break;
-    case DISPLAY_TYPE.CONVERSION_CHANGE:
-      field = conversionChangeField(lower, upper, significance);
-      break;
+
+  if (!metricData) {
+    field = `${metricName} is not available`;
+    className = "text-danger";
```
jaredlockhart, line 171: You could just set these as the defaults like

```js
let field = `${metricName} is not available`;
let className = "text-danger";
if(metricData) {
  ...
}
```

app/experimenter/visualization/api/v3/views.py
```diff
@@ -110,9 +110,30 @@ def append_conversion_count(results, primary_metrics_set):
...
-    append_population_percentages(data)
-    results, primary_metrics_set = generate_results_object(data, experiment)
+def get_week_x_retention(week_index, weekly_data):
+    return list(
+        filter(
+            lambda row: row["window_index"] == str(week_index)
+            and row["metric"] == "retained",
+            weekly_data or [],
+        )
+    )
```
jaredlockhart, line 120: Generally list comprehensions are preferred over filter/lambda but it's not a huge deal

jaredlockhart (approved):
LGTM good catch 👍

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
LZoog: I like Jared's suggestion but in `if (metricData)`, set `className` to an empty string. When making any sort of default for a `className` make sure it's a string or we get `undefined` rendered on the element. 

image

LZoog: I think you can also use the `classnames` package we have installed (search for `className={classNames`, I _think_ that takes care of an undefined class name but that might be wrong.

app/experimenter/nimbus-ui/src/lib/visualization/mocks.tsx
```diff
@@ -203,3 +203,180 @@ export const mockAnalysis = (modifications = {}) =>
     },
     modifications,
   );
+
+export const mockIncompleteAnalysis = (modifications = {}) =>
```
LZoog, line 207: Since the only diff between `mockAnalysis` and `mockIncompleteAnalysis` is `retained`, you could just call `mockAnalysis` inside this function and send in `retained` as modifications? If that's too messy maybe a comment noting what makes an analysis "incomplete" will do.

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.stories.tsx
```diff
@@ -76,4 +80,13 @@ storiesOf("visualization/TableVisualizationRow", module)
       tableLabel={TABLE_LABEL.PRIMARY_METRICS}
       metricKey="feature_c_ever_used"
     />
+  ))
+  .add("Count field missing values", () => (
```
LZoog, line 84: 😁 yay for Storybook

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
pdehaan:
```css
        .undefined::before {
            content: "¯\\_(ツ)_/¯";
            padding-right: 1em;
            color: red;
        }
```

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
```diff
@@ -163,42 +163,54 @@ const TableVisualizationRow: React.FC<{
...
-      break;
-    case DISPLAY_TYPE.CONVERSION_COUNT:
-      field = conversionCountField(count, userCountMetric);
-      break;
-    case DISPLAY_TYPE.CONVERSION_CHANGE:
-      field = conversionChangeField(lower, upper, significance);
-      break;
+
+  let field = `${metricName} is not available`;
+  let className = "text-danger";
```
jodyheavener, line 169: Don't let this block you but if you do another push these could be `const`

app/experimenter/nimbus-ui/src/lib/visualization/mocks.tsx
emtwo: It was a bit messy using the modifications since the object is quite nested. I added a comment instead.

LZoog (approved):
LGTM :)

PR #4124, opened 2020-12-04
fixes #4113 fix(nimbus): refactor mockExperimentQuery to use generic for return type, rename data to experiment

### LINKED ISSUE #4113: Use TS generic for mockExperimentQuery return type
The `mockExperimentQuery` function can return experiment data or null depending on how you've set it up, and that makes the return type tricky to work with since you have to ! the experiment data, and 90% of the time you're working with experiment data.

We should set it up so you can specify the return type in a generic, defaulting to `getExperiment["experimentBySlug"]`, and if you want to test a null response just set it up as `mockExperimentQuery<null>(...)`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-662)
┆Issue Number: EXP-662

jodyheavener:
Closes #4113

This PR:

- Refactors `mockExperimentQuery` to use generic for return type, rename data to experiment
	- Now we no longer need to non-null assert on the experiment data every single time
	- The majority of our mock components and functions accept `experiment` as an arg/prop, so destructuring it as `experiment` allows us to shorthand it everywhere

### DIFF
--- a/app/experimenter/nimbus-ui/src/lib/mocks.tsx
+++ b/app/experimenter/nimbus-ui/src/lib/mocks.tsx
@@ -17,7 +17,10 @@ import { equal } from "@wry/equality";
 import { DocumentNode, print } from "graphql";
 import { cacheConfig } from "../services/apollo";
 import { GET_EXPERIMENT_QUERY } from "../gql/experiments";
-import { getExperiment } from "../types/getExperiment";
+import {
+  getExperiment,
+  getExperiment_experimentBySlug,
+} from "../types/getExperiment";
 import { getConfig_nimbusConfig } from "../types/getConfig";
 import { GET_CONFIG_QUERY } from "../gql/config";
 import { NimbusExperimentStatus } from "../types/globalTypes";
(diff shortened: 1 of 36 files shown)


app/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -231,89 +234,91 @@ export class SimulatedMockLink extends ApolloLink {
...
-              __typename: "NimbusReadyForReviewType",
-            },
-            startDate: new Date().toISOString(),
-            endDate: new Date(Date.now() + 12096e5).toISOString(),
-          },
-          modifications,
-        );
+  if (modifications === null) {
+    experiment = null;
+  }
```
LZoog, line 321: With this being more obvious, you can probably remove the comment on L317-318

LZoog (approved):
These kind of changes are why we need maintenance sprints. 👍

I checked into your branch and tried to find any you may have missed but I think you got 'em all.

PR #4125, opened 2020-12-05
fixes #4017 #4018 feat(nimbus): support newlines and links in hypothesis text

### LINKED ISSUE #4017: Hypothesis text strips newlines in the Summary view
https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/nimbus/lorem-ipsum-hypothesis-test/edit/overview looks like this:

Mozilla_Experimenter

But by the time I get to the https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/nimbus/lorem-ipsum-hypothesis-test/request-review page, the newlines are gone and its One Great Paragraph:

Mozilla_Experimenter

Pretty sure the Hypothesis is internal only, so this isn't user facing, but the lack of formatting makes it difficult to review.

I also almost expected it to auto-link the hyperlink in the overview, but that's probably outside the scope of this newline discussion so I can file that shortly.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-609)
┆Issue Number: EXP-609

### LINKED ISSUE #4018: Auto link URLs in Hypothesis text in summary
If you have a detailed hypothesis with various links to UI specs or tech docs or whimsical GIFs, the text appears as plain text (with linebreaks removed, see #4017), meaning you need to scan the text, manually copy links (like a chump) and then open a new tab, paste the text link to see the content.

There should be several npm modules which auto-link text ([**linkify-urls**](https://www.npmjs.com/package/linkify-urls) or [**linkify-it**](https://www.npmjs.com/package/linkify-it) or probably dozens of others), and I'm sure PyPI has some too, depending on where the logic would need to be injected (but probably best to do it at render time vs preprocessing).

Mozilla_Experimenter
Ref: https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/nimbus/lorem-ipsum-hypothesis-test/request-review

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-610)
┆Issue Number: EXP-610

jodyheavener:
Closes #4017
Closes #4018

This PR:

- Adds a new component that takes a string of returns markup with newline characters converted to line break tags and URLs converted to anchor tags
- Uses this new component when rendering the hypothesis

In both issues @lmorchard noted that this is something that could be done on frontend or backend. @jaredlockhart do you have thoughts or strong opinions on that? This wasn't a lot of work if we do want to move it to backend

Storybook: [`RichText`](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/30328930e7fcb5d455f77ae201069ab1ff1ce174/nimbus-ui/index.html?path=/story/components-richtext--basic)

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import RichText from ".";
+
+it("renders text with newlines converted to <br> tags", () => {
+  const text = `Aliens
+
+  Galactic Federation
+
+  We are not ready`;
+
+  const result = render(<RichText {...{ text }} />);
+
+  // Each return should be replaced with a <br />
+  const lineBreaks = result.container.querySelectorAll("br");
+  expect(lineBreaks).toHaveLength(4);
+});
+
+it("renders text with strings that look like links converted to <a> tags", () => {
+  const url1 = "https://www.mozilla.org/about/manifesto";
+  const url2 = "http://foundation.mozilla.org/";
+  const url3 = "www.mozilla.org/contribute";
+  const url4 = "mozilla.org/about/leadership/";
+  const text = `Mozilla Manifesto: ${url1}
+  Mozilla Foundation: ${url2}
+  Get Involved: ${url3}
+  Leadership: ${url4}`;
+
+  const result = render(<RichText {...{ text }} />);
+
+  const anchors = result.container.querySelectorAll("a");
+  expect(anchors).toHaveLength(3);
+
+  expect(anchors[0]).toHaveProperty("href", url1);
+  expect(anchors[0]).toHaveTextContent(url1);
+
+  expect(anchors[1]).toHaveProperty("href", url2);
+  expect(anchors[1]).toHaveTextContent(url2);
+
+  // Non protocol links should be converted, with the text
+  // unchanged but the href prepended with https://
+  expect(anchors[2]).toHaveProperty("href", `https://${url3}`);
+  expect(anchors[2]).toHaveTextContent(url3);
+});
--- a/app/experimenter/nimbus-ui/src/components/RichText/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/RichText/index.tsx
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import LinkExternal from "../LinkExternal";
+
+// Match http://, https://, www., a combo, but not neither
+const LINK_MATCHER = /((?<!\+)(?:https?:\/\/(?:www\.)?|www\.)(?:[a-zA-Z\d-_.]+(?:(?:\.|@)[a-zA-Z\d]{2,}))(?:(?:[-a-zA-Z\d:%_+.~#*$!?&//=@]*)(?:[,](?![\s]))*)*)/gi;
+const NEWLINE_MATCHER = /(\r\n|\r|\n)/gi;
+
+const RichText = ({ text }: { text: string }) => (
+  <>
+    {text
+      .split(
+        new RegExp(`${LINK_MATCHER.source}|${NEWLINE_MATCHER.source}`, "gi"),
+      )
+      .map((word, idx) => {
+        const linkMatch = word?.match(LINK_MATCHER);
+        const newlineMatch = word?.match(NEWLINE_MATCHER);
+
+        if (linkMatch) {
+          let href = (text = linkMatch[0]);
+          if (!href.startsWith("http")) {
+            href = `https://${href}`;
+          }
+          return <LinkExternal {...{ href, key: idx }}>{text}</LinkExternal>;
+        }
+
+        if (newlineMatch) {
+          return (
+            <React.Fragment key={idx}>
+              {word}
+              <br />
+            </React.Fragment>
+          );
+        }
+
+        return word;
+      })}
+  </>
+);
+
+export default RichText;
--- a/app/experimenter/nimbus-ui/src/components/TableSummary/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableSummary/index.tsx
@@ -7,6 +7,7 @@ import { getExperiment_experimentBySlug } from "../../types/getExperiment";
 import { Table } from "react-bootstrap";
 import { useConfig } from "../../hooks";
 import { displayConfigLabelOrNotSet, NotSet } from "../Summary";
+import RichText from "../RichText";
 
 type TableSummaryProps = {
   experiment: getExperiment_experimentBySlug;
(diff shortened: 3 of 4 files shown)


app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import RichText from ".";
+
+const url = "https://www.mozilla.org/en-CA/about/manifesto/";
```
LZoog, line 9: We could also test for `http` and not just `https`. Do we want to support links provided without this like `www.mozilla.org` as well or to require it? I imagine something missing `http` / `www` can be rendered as plaintext (like `mozilla.org`), but if we want to require `http`/`https`, maybe we want to note that to the user in the form itself?

LZoog (commented):
Just one comment, holding on r+ for Jared to take a looksee in case he has a preference for this living in the BE, but I lean towards it living in the FE so we don't have to parse HTML coming from the BE.

pdehaan:
Not sure if it'd be easier to use something like Sindre's [**sindresorhus/linkify-urls**](https://github.com/sindresorhus/linkify-urls) (where we'd get future bug fixes for "free").
Although I think the RegExp's look roughly similar [and I think we're already handling the external links `target="_blank"` business via `<LinkExternal>` component].

I'm guessing treating the `hypothesis` like it's a Markdown and letting **markdown-it** convert from Markdown->HTML might be a bigger can of worms than we want to chew off (although we'd be able to inline images and better support for bulleted/numbered lists, etc).

app/experimenter/nimbus-ui/src/components/TableSummary/index.tsx
```diff
@@ -38,7 +39,9 @@ const TableSummary = ({ experiment }: TableSummaryProps) => {
         </tr>
         <tr>
           <th>Hypothesis</th>
-          <td data-testid="experiment-hypothesis">{experiment.hypothesis}</td>
+          <td data-testid="experiment-hypothesis">
+            <RichText text={experiment.hypothesis || ""} />
+          </td>
         </tr>
         <tr>
           <th>Public description</th>
```
pdehaan, line 47: Does the public description field also need to be `<RichText>` wrapper? I imagine it could potentially include a linebreak and/or hyperlinks.

jodyheavener: cc @k88hudson @AnaMedinac

jodyheavener:
> Not sure if it'd be easier to use something like Sindre's sindresorhus/linkify-urls

I wouldn't mind using a library that performed both newline and link transformation, but I would also want it to return JSX because React really doesn't like you manually inserting your own HTML (which is what we'd be doing with Sindre's lib AFAICT)

> treating the hypothesis like it's a Markdown

I would _love_ it if we rendered Markdown, there are many React components that'll do it for us, and support for rendering more than plain text would be wonderful. The problem with that, though, is that we then store MD syntax, and it would be up to each client that renders the hypothesis to also parse the MD lest we want to see asterisks and underscores throughout (versus what we have now -- if a link isn't clickable it's still a link). It's probably worth a discussion though!

app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
jodyheavener: [Updated](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/8cdd72eaf1c28a9edd50229b795414e9004f05b0/nimbus-ui/index.html?path=/story/components-richtext--basic) to support optional protocol, but note that since both `www` and `http/s` are optional, it means text with _neither_ (e.g. `foo.com`) will also get turned into a link. I think that should be fine, but it also might be a little eager and start converting things erroneously. Might need to keep an eye on it.

app/experimenter/nimbus-ui/src/components/RichText/index.tsx
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
jodyheavener: As in, is there a component that we can use instead of rolling our own? Probably. There's almost certainly one to convert links, but both links and newlines? Not sure.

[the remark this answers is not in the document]
jodyheavener: @jaredlockhart while your here, did you see the comment above about backend?

app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
LZoog: I meant it seemed sufficient for `http`, `https` or `www` to render as a link and if it's missing one of those, don't render the link, but I don't mind just a `.org` getting one and maybe that mitigates needing a tooltip or some other informative text about what will be linked and what won't. Might check with @AnaMedinac / @k88hudson on this?

app/experimenter/nimbus-ui/src/components/TableSummary/index.tsx
k88hudson: No, public descriptions have to be rendered in UI for each platform (see about:studies [here](https://searchfox.org/mozilla-central/source/toolkit/components/normandy/content/about-studies/about-studies.js#302)). Keeping it plain text will make it a lot easier to coordinate this across platforms. 

Since this is different from hypothesis, we could add some help text about this being a single paragraph with no formatting or something in help text

app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
jodyheavener: Bump @AnaMedinac @k88hudson

jaredlockhart:
Oh oh sorry @jodyheavener I missed your request.  Yeah we could totally do it on the backend but we'd have to use https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml to inject the content from the backend and then use https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#urlize to detect and convert the links.  Then we wouldn't need any parsing libraries on the frontend?

jaredlockhart (approved):
But if you already got it all working on the frontend we could just land it like this and change it later if we need to.  I'm generally favour of not vendoring in external dependencies where we don't have to, less code to test and maintain, but don't need to block this so looks good and we can revisit later if we need to 👍

jodyheavener:
@jaredlockhart yeah, I'm not super enthused about setting the HTML like that, but I suppose it's not that bad because it's not _any_ HTML and the user has no control over it, it's only specific.

jaredlockhart:
@jodyheavener Yeah I mean this is how we already do it:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/legacy-ui/templates/experiments/section_overview.html#L9

app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
LZoog: TLDR on above:

* Should we render all links including for URLs that just end in `.com`/`.org` etc. as links or should it be only URLs beginning with `https`/`http` and/or `www`? At the moment it's the former, [see Storybook](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/8cdd72eaf1c28a9edd50229b795414e9004f05b0/nimbus-ui/index.html?path=/story/components-richtext--basic). Is there any case where we wouldn't want a URL to link-out?
* Probably only applicable if we change the behavior from linked Storybook - do we want a note (or tooltip) somewhere to inform the user their URLs will link-out if a link has X factor (like starting with `http`/`https`)?

k88hudson: Hmm, sometimes app ids are app.whatver.org so i'd probably go with just http/https/www personally, but we could see if anyone runs into it

LZoog (approved):
LGTM

app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
jodyheavener: Updated to [match](https://regex101.com/r/3lWJNS/2) protocol and/or www, but not neither ([Storybook](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/906e6114f4e331feb3811b826578151d997c9799/nimbus-ui/index.html?path=/story/components-richtext--basic))

PR #4128, opened 2020-12-06
fixes #4098 feat(nimbus): use a common method of checking an experiment's status

### LINKED ISSUE #4098: Standardize on checking experiment statuses
If you search for `NimbusExperimentStatus` across the `nimbus-ui` code you'll see we have a [number](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.tsx#L34-L40) of [inline checks](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx#L76-L103) and [functions](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/SummaryTimeline/index.tsx#L19-L27) to determine the various states of an experiment.

We should come up with a utility function that can accept an experiment and allow checks individual experiment statuses (e.g.: `status.isDraft`) as well as common combinations of statuses (e.g.: `status.isLocked`).

Bonus points if you want to create a general `Experiment` class that implements the `getExperiment["experimentBySlug"]` interface, so that it can provide the status checks but also other experiment-related properties and functions (such as manipulating start/end dates).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-653)
┆Issue Number: EXP-653

jodyheavener:
Closes #4098

This PR:

- Adds a `getStatus` function that accepts an experiment and returns an object of status check. It includes checks for all the available `NimbusExperimentStatus` types, as well as some other common ones:
  - "released" if the experiment is or was out in the wild, that is live or complete
  - "locked" if the experiment's fields are locked, as in it is not a draft or in review
- Replaces all the various ways we've previously checked on an experiment's status with this new function

### DIFF
--- a/app/experimenter/nimbus-ui/src/lib/experiment.ts
+++ b/app/experimenter/nimbus-ui/src/lib/experiment.ts
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getExperiment_experimentBySlug } from "../types/getExperiment";
+import { NimbusExperimentStatus } from "../types/globalTypes";
+
+export type StatusCheck = {
+  draft: boolean;
+  review: boolean;
+  accepted: boolean;
+  live: boolean;
+  complete: boolean;
+  locked: boolean;
+  released: boolean;
+};
+
+export function getStatus(
+  experiment?: getExperiment_experimentBySlug,
+): StatusCheck {
+  const status = experiment?.status;
+
+  const released = status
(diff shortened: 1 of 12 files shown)


app/experimenter/nimbus-ui/src/lib/experiment.ts
```diff
@@ -0,0 +1,38 @@
...
+  draft: boolean;
+  review: boolean;
+  accepted: boolean;
+  live: boolean;
+  complete: boolean;
+  locked: boolean;
+  released: boolean;
+};
+
+export function getStatus(
```
LZoog, line 18: Might be good to have a comment above this function so one can quickly see what `released` and `locked` means.

LZoog (approved):
Just have one suggestion on adding a comment - this is a great refactor! :)

PR #4144, opened 2020-12-07
fixes #4143 fix(nimbus): Check for owner email truthiness to prevent page erroring

### LINKED ISSUE #4143: Pages displaying owner email error
The experiment owner hasn't been set on experiments created yet: EXP-611

This is causing pages showing the owner email to error since we assume it's present on experiment data.

Once EXP-611 is fixed, on experiments moving forward we won't have to check for truthiness on `experiment.owner.email` but for previously created experiments we will (or delete all experiments on staging/prod up until this point).

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-671)
┆Issue Number: EXP-671

LZoog:
Because:
* We've had a bug hanging around that's prevented owner email from being added to experiments created so all experiments created in Nimbus UI until that bug is fixed will not have an owner experiment set. This prevents those experiment pages from erroring.

This commit:
* Displays "Not set" if experiment owner is not set in the Summary Table

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableSummary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableSummary/index.test.tsx
@@ -27,6 +27,14 @@ describe("TableSummary", () => {
     );
   });
 
+  it("renders Not Set if experiment owner is not set", () => {
+    const { data } = mockExperimentQuery("demo-slug", {
+      owner: null,
+    });
+    render(<Subject experiment={data!} />);
+    expect(screen.getByTestId("experiment-owner")).toHaveTextContent("Not set");
+  });
+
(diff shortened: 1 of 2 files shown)


jodyheavener (approved):
LGTM. Curious, though, once #4142 is merged will this even be possible (excluding already-created ones)?

LZoog:
@jodyheavener Correct, it shouldn't be possible. I'd originally told @pdehaan I wasn't going to fix this since we won't need the patch on experiments after #4142 is fixed, but previously-created experiments will still error unless we choose to delete them all in staging/prod or add a random email to them... Since this was a super fast/minimal patch, seemed best to just make it it and get it in. We can remove this if we delete all or update all Nimbus experiments made through the UI up until #4142 is merged.

PR #4170, opened 2020-12-08
fixes #3746 feat(nimbus): Add Tooltips to Metrics page

### LINKED ISSUE #3746: Add Tooltips
A tooltip is shown on a couple of pages in Figma, on at least the "Metrics' page.

Add the tooltip shown on each page, copy which may need to come from Ana directly if not in Figma. Can we use https://react-bootstrap.github.io/components/overlays/#tooltip-props?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-486)
┆Issue Number: EXP-486

LZoog:
Because:
* We want to display extra info to the user about what probe sets are.

This commit:
* Adds primary probe sets and secondary probe sets tooltips.

---

Opening as a draft, Ana aims to get me these by EOD.

These had to be placed in the form label itself, placing them just after it caused other test failures that looked hairy to fix.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/FormMetrics/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormMetrics/index.test.tsx
@@ -7,6 +7,7 @@ import { render, screen, act, fireEvent } from "@testing-library/react";
 import { mockExperimentQuery } from "../../lib/mocks";
 import { Subject } from "./mocks";
 import { getConfig_nimbusConfig_probeSets } from "../../types/getConfig";
+import { PRIMARY_PROBE_SETS_TOOLTIP, SECONDARY_PROBE_SETS_TOOLTIP } from ".";
 
 describe("FormMetrics", () => {
   const probeSets: getConfig_nimbusConfig_probeSets[] = [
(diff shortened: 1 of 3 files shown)


AnaMedinac:
Primary Metrics tooltip: Specific metrics you’d like to impact in your experiment, which will be part of the analysis.
Secondary Metrics tooltip: Specific metrics that you are interested in observing in your experiment but they don't affect the results of your experiment.

app/experimenter/nimbus-ui/src/components/FormMetrics/index.test.tsx
```diff
@@ -38,6 +38,15 @@ describe("FormMetrics", () => {
...
       expect(screen.getByTestId("FormMetrics")).toBeInTheDocument(),
     );
+
+    expect(screen.getByTestId("tooltip-primary-probe-sets")).toHaveAttribute(
+      "data-tip",
+      "Specific metrics you’d like to impact in your experiment, which will be part of the analysis.",
+    );
+    expect(screen.getByTestId("tooltip-secondary-probe-sets")).toHaveAttribute(
+      "data-tip",
+      "Specific metrics that you are interested in observing in your experiment but they don't affect the results of your experiment.",
```
jodyheavener, line 48: given the length/specificity of these strings in two places it might make sense to constantize them, but i'm on the fence since it mean doing it for all tooltips

jodyheavener (approved):

app/experimenter/nimbus-ui/src/components/FormMetrics/index.test.tsx
LZoog: That's a better idea, I'll go ahead and do that for these at least

PR #4173, opened 2020-12-08
fix #4172 chore(nimbus): use string for populationPercent

### LINKED ISSUE #4172: make check not detecting graphql schema changes
I just ran `make generate_types` on main and the following change popped out:

```
-  populationPercent?: number | null;
+  populationPercent?: Decimal | null;
```

so the checker should've caught that and prevented it from landing on main.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-697)
┆Issue Number: EXP-697

jaredlockhart:
Because

* We were inconsistently using number/float/Decimal for populationPercent in the V5 API

This commit

* Uses string everywhere in the gql/frontend for populationPercent to preserve the exact value input by users
* The V5 serializer will enforce parsing and validating the string into a Decimal with the correct number of decimal places


lmorchard (approved):
Seems to work

PR #4174, opened 2020-12-08
fix 4157: allow audience populationPercent to be a decimal number

lmorchard:


lmorchard:
I haven't played with `type="number"` inputs a bunch, but this seems to fix the issue?

app/experimenter/nimbus-ui/src/components/FormAudience/index.tsx
```diff
@@ -234,6 +234,7 @@ export const FormAudience = ({
                 type="number"
                 min="0"
                 max="100"
+                step="0.01"
```
jaredlockhart, line 237: populationPercent allows 4 decimal places:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L47-L49

jaredlockhart (changes requested):

app/experimenter/nimbus-ui/src/components/FormAudience/index.tsx
lmorchard: I don't think this restricts the number of decimal places, just indicates that it's not an integer and provides a step value for the up/down arrows

jaredlockhart (approved):
👍 🐱

PR #4207, opened 2020-12-10
fix #4149: tweak branch JSON validation errors to identify field

### LINKED ISSUE #4149: V5 invalid JSON in branch feature value should raise per field error
Right now the server side is spitting out a total form error if a single branch has invalid json in a single field, it should be per field

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-676)
┆Issue Number: EXP-676

lmorchard:
Because:

- we want to display per-branch feature value validation errors at the
  field level

This commit:

- tweaks the data structure describing per-field JSON validation errors
  in branch mutation

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -184,14 +184,22 @@ class NimbusBranchUpdateSerializer(
                 schema, data["reference_branch"]["feature_value"]
             )
             if errors:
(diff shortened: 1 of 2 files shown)


lmorchard:
I think this does the trick?

Capture

jaredlockhart (approved):
Oh yes good catch this looks good thank you! 👍

Issue #4154, opened 2020-12-07
Improve user feedback / always enable "Save" button on Branches page


data-sync-user:
There’s no user feedback when a feature config is “on” but branch name/description is missing - the “Save” button remains disabled and user may not know why.

The user can also delete the feature config and the “Save” button is disabled.

From our 12/7 meeting on validation/error messaging, it seems like the easiest solution is to just always enable and never disable the "Save" button on the Branches page. It'd probably be best to play around with the form to make sure it makes sense to always be enabled.

Estimating this at a 2 to always leave the button enabled - if we need to leave it disabled in certain scenarios, it'll need to be bumped up to accommodate better user feedback.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-679)
┆Issue Number: EXP-679

data-sync-user:
➤ Les Orchard commented:

Bumping this up a bit: Walking through the way branch page validation works again, I think I need to shuffle around what happens to the data on submit vs change.

PR #4216, opened 2020-12-10
issue #4154: progress on reworking branches form validation and submission

lmorchard:
Opening as a draft in case anyone wants to see what I've got so far. Still need to tighten a few bolts and update tests.

Main change here is to stop updating the Branch page reducer on every form field change. Instead, the relationship is flipped so that the forms manage themselves up until the submit button is clicked, at which point the parent component walks through to force validation and gather the form data. This allows a forced validation with feedback on submission, as well as reducing a bunch of glitchiness while typing


lmorchard:
Actually, going to close this. Just read through react-hook-form docs and understand better how to do a nested form. Going to rework my rework and see if I come up with something more maintainable.

PR #4217, opened 2020-12-10
fix(nimbus): update todos with jira tickets

jodyheavener:
I just wanted to make sure all our TODOs had appropriate Jira tickets, otherwise they may never get TODONE


LZoog (approved):
TY! I have two draft PRs pending product that remove a few of these TODOs but good to have 'em all

PR #4225, opened 2020-12-10
fix #4187 feat(nimbus): Change channel to single select

### LINKED ISSUE #4187: Change Channel to single select
After realizing that we won't be able to analyze an experiment targeting multiple channels effectively, we need to change channel to a single select rather than a multi select.  We should be able to get away with doing this at the API layer and changing the multiselect to a dropdown in the UI.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-708)
┆Issue Number: EXP-708

jaredlockhart:
Because

* As it turns out, deploying to multiple channels makes it difficult to reason about population sizing/branch allocation/automated analysis, it's best to target single channels for the time being

This commit

* Changes channel from a multi select to a single select
* Update database
* Update serializers
* Update UI
* Update tests
* Update all the things


k88hudson (changes requested):
Just manually testing now, but  I think this looks good (suggestion on the JEXL change).

We should ask @pdehaan to verify this since it's a targeting change.

LZoog:
This LGTM except for missing a few `channels` in `hooks/useExperiment.tsx` and in a few stories. The "missing details" bit isn't getting rendered here: https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/b57261af3434d4f8712efcf01dec54303b5051e1/nimbus-ui/index.html?path=/story/pages-editaudience--missing-fields

Search `.tsx` files for `channels` and make needed changes? Will r+ after :D

jaredlockhart:
@LZoog Oh good catch!  Interesting TS didn't catch those....

LZoog (approved):
👍

PR #4226, opened 2020-12-10
fixes #3971 feat(nimbus): promote Nimbus UI to /nimbus, removing /experiments from path

### LINKED ISSUE #3971: Serve application at `/nimbus` rather than `/experiments/nimbus`
We don't need "experiments" in the URL, the app can be served at simply "/nimbus" instead. Be sure to update the `BASE_PATH` in `nimbus-ui` constants too.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-581)
┆Issue Number: EXP-581

jodyheavener:
Close #3971

I think I did this correctly?

This PR moves the Nimbus UI app from `/experiments/nimbus` to `/nimbus`. I also updated the code to remove the concept that Nimbus is an "experiment" (although the view is still living in `/experiments/views.py`, how do we feel about that? Where would it live otherwise?).

### DIFF
--- a/app/experimenter/urls.py
+++ b/app/experimenter/urls.py
@@ -3,7 +3,7 @@ from django.conf.urls import include, re_path
 from django.conf.urls.static import static
 from django.contrib import admin
 
-from experimenter.experiments.views import ExperimentListView
+from experimenter.experiments.views import ExperimentListView, NimbusUIView
 
 urlpatterns = [
     re_path(r"^api/v1/experiments/", include("experimenter.experiments.api.v1.urls")),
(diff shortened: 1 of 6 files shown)


app/experimenter/urls.py
```diff
@@ -14,6 +14,8 @@
     re_path(r"^admin/", admin.site.urls),
     re_path(r"^experiments/", include("experimenter.experiments.urls")),
     re_path(r"^$", ExperimentListView.as_view(), name="home"),
```
jaredlockhart, line 16: This should be at the bottom

app/experimenter/urls.py
```diff
@@ -14,6 +14,8 @@
...
+    re_path(r"^nimbus/", NimbusUIView.as_view(), name="nimbus"),
```
jaredlockhart, line 17: I'd call this `name="nimbus-list"`

app/experimenter/urls.py
```diff
@@ -14,6 +14,8 @@
...
+    re_path(r"^nimbus/(?P<slug>[\w-]+)/", NimbusUIView.as_view(), name="nimbus"),
```
jaredlockhart, line 18: And this `name="nimbus-detail"`

jaredlockhart (changes requested):
Oh ty this will be nicer 🙏

jaredlockhart (approved):
Looks right to me!

PR #4227, opened 2020-12-10
fixes #4106 fix(Legacy): Set initial value rollout schedule to blank

### LINKED ISSUE #4106: For Rollouts - remove the default value in "rollout schedule"
small UX change to help with user confusion.  the default value of "rollout playbook" makes the field looked filled in - so people frequently miss answering and can't figure out why the section says it's not complete.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-659)
┆Issue Number: EXP-659

tiftran:
Because:

 * users were confused and often miss that field

This commit:

 * sets intial rollout schedule to be blank


jaredlockhart (approved):
Oh yep that's it that works that looks right perfect thank you 🙏 

Can you just also change it here so it's consistent and isn't confusing later pls

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/legacy.py#L67

PR #4231, opened 2020-12-11
chore(project): fix conflicting migrations

jaredlockhart:
Because

* We landed conflicting migrations to main

This commit

* Adds a merge migration


emtwo (approved):
Thank you!

PR #4233, opened 2020-12-11
fixes #4164: fix(visualization) show loader until data loads.

### LINKED ISSUE #4164: Visualization page loads summary table first
This is a bug. The summary table is shown briefly until the visualizations have loaded.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-689)
┆Issue Number: EXP-689

emtwo:
Because
* We don't want content flashing then changing

This commit
* Waits for data to load before showing content

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
@@ -49,30 +49,22 @@ const AppLayoutWithExperiment = ({
   const {
     experiment,
     notFound,
-    loading: experimentLoading,
+    loading,
     startPolling,
     stopPolling,
     review,
   } = useExperiment(slug);
-  // We won't know if an analysis lookup is required until the initial experiment query
-  // is complete and we inspect its status. To prevent content from flashing on the screen
-  // between the experiment and analysis requests, assume we need to wait for analysis
-  // until we can prove otherwise.
   const [analysisFetched, setAnalysisFetched] = useState<boolean>(false);
   const status = getStatus(experiment);
 
-  const {
-    execute: fetchAnalysis,
-    result: analysis,
-    loading: analysisLoading,
-  } = useAnalysis();
+  const { execute: fetchAnalysis, result: analysis } = useAnalysis();
 
   useEffect(() => {
-    if (!analysisFetched && !experimentLoading && status.released) {
+    if (!analysisFetched && !loading && status.released) {
       fetchAnalysis([experiment?.slug]);
       setAnalysisFetched(true);
     }
-  }, [fetchAnalysis, experimentLoading, experiment, analysisFetched, status]);
+  }, [fetchAnalysis, loading, experiment, analysisFetched, status]);
 
   useEffect(() => {
     if (polling && experiment) {
@@ -83,7 +75,7 @@ const AppLayoutWithExperiment = ({
     };
   }, [startPolling, stopPolling, experiment, polling]);
 
-  if (experimentLoading || (analysisRequired && analysisLoading)) {
+  if (loading || (analysisRequired && !analysis)) {
     return <PageLoading />;
   }
 
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
@@ -32,6 +32,7 @@ type AppLayoutWithExperimentProps = {
   title?: string;
   polling?: boolean;
   sidebar?: boolean;
+  analysisRequired?: boolean;
 } & RouteComponentProps;
 
 export const POLL_INTERVAL = 30000;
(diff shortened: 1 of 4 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -88,7 +81,7 @@ const AppLayoutWithExperiment = ({
     };
   }, [startPolling, stopPolling, experiment, polling]);
 
-  if (experimentLoading || analysisRequired) {
```
emtwo, line 91: The issue, I noticed, was that `experimentLoading` would become false but `analysisRequired` would also become false due to the `useEffect()` that I removed. There, the experiment would be done loading but `analysisLoading` would still be false because it would not have started loading yet. Then `analysisRequired` is set to false too soon, causing the summary table to display first.

I found that `analysisRequired` didn't seem necessary since we can just check `analysisLoading` instead.

emtwo:
@jodyheavener I think you made some of these changes in `AppLayoutWithExperiment` so I wanted to check with you what you think of this?

emtwo:
For reference, here's a link where the bug happens: https://experimenter.services.mozilla.com/nimbus/attribution-targeting-release-aa-test/results

jodyheavener:
Darn I thought I had solved that. @emtwo since this component is used on every Experiment page, can you confirm that this change still only performs the analysis request on pages that require it? For example, none of the Edit pages should need to wait for analysis request to complete.

emtwo:
> Darn I thought I had solved that. @emtwo since this component is used on every Experiment page, can you confirm that this change still only performs the analysis request on pages that require it? For example, none of the Edit pages should need to wait for analysis request to complete.

Oh! That makes sense why you were checking `analysisRequired` then. It was whether it's required for the current page. I think this change doesn't address this.

Back to the drawing board!

jodyheavener:
Hmmm. `AppLayoutWithExperiment` is getting pretty bloated as it is. I wonder if we should split that component into two components; one targeted at Design/Results pages (which should always perform the analysis request) and one targeted at the Edit/Request Review pages (which should never perform that request)? cc @LZoog

emtwo:
What if we use the same component but pass in an optional prop `analysisRequired`. It would be false by default. But `PageResults` would set it to true. I don't think that's *too* much additional bloating

jodyheavener:
Oh hey that seems like a very reasonable solution 🤦

jodyheavener:
Note that `PageDesign` also needs it, since the sidebar that is displayed with that page needs to know if the analysis is ready or not.

emtwo:
@jodyheavener I updated it with what I mentioned. Works now without loading analysis for edit pages

app/experimenter/nimbus-ui/src/components/PageDesign/index.tsx
```diff
@@ -8,7 +8,11 @@ import AppLayoutWithExperiment from "../AppLayoutWithExperiment";
 
 const PageDesign: React.FunctionComponent<RouteComponentProps> = () => {
   return (
-    <AppLayoutWithExperiment title="Design" testId="PageDesign">
+    <AppLayoutWithExperiment
+      title="Design"
+      testId="PageDesign"
+      analysisRequired={true}
```
jodyheavener, line 14: Minor, not a blocker: I think you can just specify the prop (not need to set `={true}`)

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -56,7 +58,6 @@ const AppLayoutWithExperiment = ({
   // is complete and we inspect its status. To prevent content from flashing on the screen
   // between the experiment and analysis requests, assume we need to wait for analysis
   // until we can prove otherwise.
```
jodyheavener, line 60: This comment should be able to be removed (or reworded) since it's not longer relevant.

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -88,7 +83,7 @@ const AppLayoutWithExperiment = ({
     };
   }, [startPolling, stopPolling, experiment, polling]);
 
-  if (experimentLoading || analysisRequired) {
+  if (experimentLoading || (analysisRequired && analysisLoading)) {
```
jodyheavener, line 86: Hmmmm, I think you will still get a flash of content here. `experimentLoading` becomes false ever so slightly before `analysisLoading` becomes true. You can see this locally by throttling the Network tab to something like 3G, but I assume this would be more visible once deployed. 

It's almost like we need a way for `analysisLoading` to _start_ true. Or we need a way to do `(experimentLoading || (analysisRequired && analysisRequestComplete))` to check if the request was finished (which would be irrelevent in renders where analysisRequired is false). 🤔

jodyheavener (commented):
@emtwo I love how simple this change is, but I'm unsure if it solves the problem you're encountering (see last comment). Happy to discuss more!

jodyheavener:
One thing to consider with what I just pushed: I think this would show an indefinite loading screen if there is a legit error performing the analysis request.

jodyheavener (approved):
I'm biased but I think this looks _GREAT_

emtwo:
@LZoog adding you here since Jody and I both made changes and we need an unbiased opinion

LZoog (approved):
LGTM!

PR #4251, opened 2020-12-14
fixes #4165: fix(visualization) Add retention tooltip.

### LINKED ISSUE #4165: When retention is not displayed due to a short duration, this should be shown on hover
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-688)
┆Issue Number: EXP-688

emtwo:
Because
* We want users to understand why retention is missing

This commit
* Adds a tooltip for when retention is missing

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
@@ -10,7 +10,10 @@ import {
   VARIANT_TYPE,
   DISPLAY_TYPE,
 } from "../../lib/visualization/constants";
-import { SIGNIFICANCE_TIPS } from "../../lib/visualization/constants";
+import {
+  SIGNIFICANCE_TIPS,
+  GENERAL_TIPS,
+} from "../../lib/visualization/constants";
 import { BranchDescription } from "../../lib/visualization/types";
 import { ReactComponent as SignificanceNegative } from "./significance-negative.svg";
 import { ReactComponent as SignificancePositive } from "./significance-positive.svg";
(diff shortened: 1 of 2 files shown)


emtwo:
Screen Shot 2020-12-14 at 11 24 52 AM

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
```diff
@@ -168,9 +171,11 @@ const TableVisualizationRow: React.FC<{
   const metricData = branch_data[metricKey];
 
   let field = <>{metricName} is not available</>;
+  let tooltipText =
+    metricKey === METRIC.RETENTION ? GENERAL_TIPS.MISSING_RETENTION : "";
   let className = "text-danger";
   if (metricData) {
-    className = "";
+    className = tooltipText = "";
```
jaredlockhart, line 178: Does the little `= ""` need to be at the end still?

app/experimenter/nimbus-ui/src/lib/visualization/constants.ts
```diff
@@ -30,6 +30,11 @@ export const BADGE_TIPS = {
   GUARDRAIL_METRIC: "Metric that should not regress",
 };
 
+export const GENERAL_TIPS = {
+  MISSING_RETENTION:
+    "Retention is only available after at least 1 week of data post enrollment",
```
jaredlockhart, line 35: 👍 👍 👍 👍

jaredlockhart (approved):
No test for this?  Not sure if it needs one, could have one though if you want.  Looks good!

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
emtwo: yes. we need to reset it

pdehaan: Yeah, it's a double assignment where we're setting *both* `className` and `tooltipText` to an empty string.
I think ESLint has a rule for disallowing multi-assignments and forces you to do two separate `className = ""; tooltipText = "";` assignments instead.

PR #4261, opened 2020-12-14
Add filter expression tests

### LINKED ISSUE #3243: Validate JEXL targeting property in Firefox in CI
Let's see if we can setup a way for a Firefox client (maybe of some specified version(s)?) to execute the `targeting` JEXL expressions that Experimenter generates for rapid Nimbus experiments in circle as part of our CI run.

Instructions from @k88hudson:

Ok, this works with a couple of important caveats:

* randomizationUnit has to be `userId`, not `normandy_id`. This would be easy enough to express if this were in code but it might simplest here just to write out `userId` in the template string since we only support a single randomization unit option right now anyway.
* this will fail if either `audienceTargeting` or `audienceFilterExpression` are undefined.

I tested with the following steps.

1. Open Firefox Nightly (version 80)
1. Add the following to the Browser Toolbox console:

{{Cu.import("resource://gre/modules/components-utils/FilterExpressions.jsm");
Cu.import("resource://gre/modules/components-utils/ClientEnvironment.jsm");
Cu.import("resource://messaging-system/lib/RemoteSettingsExperimentLoader.jsm");

// Filter expression
var minFirefoxVersion = "78";
var audienceFilterExpression = undefined;

// Targeting
var randomizationUnit = "userId";
var bucketNamespace = "foo-1232ad3";
var bucketStart = 0;
var bucketCount = 50;
var bucketTotal = 100;
var audienceTargeting = "localeLanguageCode == 'en'";

var recipe = {
  // These are your preset expressions, converted to JS
  filter_expression: `env.version|versionCompare('${minFirefoxVersion}') >= 0 && ${audienceFilterExpression}`,
  targeting: `[${randomizationUnit}, "${bucketNamespace}"]|bucketSample(${bucketStart}, ${bucketCount}, ${bucketTotal}) && ${audienceTargeting}`
};

// 1. Test filter_expression by simulating RS filtering https://searchfox.org/mozilla-central/source/services/settings/remote-settings.js#63
await FilterExpressions.eval(recipe.filter_expression, {env: ClientEnvironmentBase});

// 2. Test targeting by calling RemoteSettingsExperimentLoader
await RemoteSettingsExperimentLoader.checkTargeting(recipe);}}

1. Try different user Ids to see if you get a different result from the `targeting`. These are the ones I tried:

{{app.normandy.user_id=dd438e6c-03cf-d144-8e71-ab94b3d6c570 => true
app.normandy.user_id=asdasdas => false}}

1. Try changing the `minFirefoxVersion` to something in the future (e.g. 88) to see if it resolves to `false`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-71)
┆Issue Number: EXP-71

jrbenny35:
Closes #3243

### DIFF
--- a/app/tests/integration/utils/filter_expression.js
+++ b/app/tests/integration/utils/filter_expression.js
@@ -12,7 +12,6 @@ async function remoteSettings(arguments) {
     const ENVIRONMENT = {version: arguments[0], localeLanguageCode: "en-US"};
     let targeting = new TargetingContext(ENVIRONMENT);
     const EXPRESSION = `version|versionCompare('${arguments[1]}.!') >= 0 && localeLanguageCode == 'en'`;
-    console.log(EXPRESSION);
     const actual = await targeting.evalWithDefault(EXPRESSION, ENVIRONMENT);
     return actual
 }
--- a/app/tests/integration/utils/filter_expression.js
+++ b/app/tests/integration/utils/filter_expression.js
@@ -2,38 +2,21 @@ Components.utils.import("resource://gre/modules/components-utils/FilterExpressio
 Components.utils.import("resource://gre/modules/components-utils/ClientEnvironment.jsm");
 Components.utils.import("resource://messaging-system/lib/RemoteSettingsExperimentLoader.jsm");
 
-// Filter expression
-var minFirefoxVersion = arguments[1];
-var audienceFilterExpression = undefined;
-
 // Targeting
-var randomizationUnit = arguments[2];
-var bucketNamespace = "foo-1232ad3";
-var bucketStart = 0;
-var bucketCount = 50;
-var bucketTotal = 100;
-var audienceTargeting = "localeLanguageCode == 'en'";
 
-var recipe = {
-// These are your preset expressions, converted to JS
-filter_expression: `env.version|versionCompare('${minFirefoxVersion}') >= 0 && ${audienceFilterExpression}`,
-targeting: `[${randomizationUnit}, "${bucketNamespace}"]|bucketSample(${bucketStart}, ${bucketCount}, ${bucketTotal}) && ${audienceTargeting}`
-};
-async function myFunction(recipe) {
-    let res = await FilterExpressions.eval(recipe.filter_expression, {env: ClientEnvironmentBase});
-    if (res == undefined) {
-        return "undefined"
-    }
-    return res
-};
-async function remoteSettings(recipe) {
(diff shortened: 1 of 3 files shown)


app/tests/integration/utils/filter_expression.js
```diff
@@ -0,0 +1,22 @@
...
+// Targeting
+
+async function remoteSettings(arguments) {
+
+    const { TargetingContext } = ChromeUtils.import(
+        "resource://messaging-system/targeting/Targeting.jsm"
+    );
+    const ENVIRONMENT = {version: arguments[0], localeLanguageCode: "en-US"};
+    let targeting = new TargetingContext(ENVIRONMENT);
+    const EXPRESSION = `version|versionCompare('${arguments[1]}.!') >= 0 && localeLanguageCode == 'en'`;
```
pdehaan, line 14: Should we add a couple tests for testing locales?

app/tests/integration/utils/filter_expression.js
```diff
@@ -0,0 +1,22 @@
...
+    console.log(EXPRESSION);
```
pdehaan, line 15: Probably can ✂️ this debuggery.

pdehaan (commented):
Lookin' good!
Are these tests running in CI somewhere yet?

jaredlockhart (approved):
Thank you @jrbenny35 !  This is a great first step I'm really excited for this.  As the UI solidifies more we can start to expand this to include walking through the form steps and get to the point where we can generate DTOs (née recipes) for a set of different targeting configurations and assert that they work in some set of channels/versions.  Let's get this landed and then we can start expanding from there :D  Also thank you @piatra and @k88hudson for helping @jrbenny35 get this working! 🎉

app/tests/integration/utils/filter_expression.js
jrbenny35: I tried but locales don't work the way it would seem, at least for how we are doing it.

jaredlockhart: Yeah I think the important part will be generating a DTO from the UI and testing the targeting expression it spits out against the browser.

app/tests/integration/utils/filter_expression.js
jrbenny35: > Lookin' good!
> Are these tests running in CI somewhere yet?

They're running in the integration job here. Take a look at the "checks" section on github.

PR #4272, opened 2020-12-15
fix #4271 fix(nimbus): extend isolation group unique constraint to include application

### LINKED ISSUE #4271: Isolation group/bucket allocator not detecting existing isolation group
The isolation group/bucket allocator should reuse an existing isolation group of it exists, but it looks like it's trying to recreate one with an existing slug/application pair, so that's not right.

https://sentry.prod.mozaws.net/operations/experimenter-prod/issues/10268083/?referrer=slack

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-759)
┆Issue Number: EXP-759

jaredlockhart:
Because

* We recently changed isolation groups to be bound by applications but neglected to update its uniqueness constraint to include application

This commit

* Adds application to the isolation group unique constraint


emtwo (approved):
lgtm

PR #4286, opened 2020-12-16
fix #4285 fix(nimbus): handle empty feature_value

### LINKED ISSUE #4285: V6 Serializer fails if feature_value is None
The serializer should pass over empty feature_value values.

https://sentry.prod.mozaws.net/operations/experimenter-prod/issues/10269638/?query=is%3Aunresolved

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-767)
┆Issue Number: EXP-767

jaredlockhart:
Because

* Even if an experiment sets a feature_config, not all branches are required to have a feature_value

This commit

* Checks whether a feature_value is defined on a branch before attempting to serialize it
* Adds the schema check to all serializer tests just to ensure we always match the schema

### DIFF
--- a/app/tests/integration/test_filter_expression_validation.py
+++ b/app/tests/integration/test_filter_expression_validation.py
@@ -1,6 +1,4 @@
-def test_filter_expressions_with_matching_firefox_versions(
-    base_url, selenium
-):
+def test_filter_expressions_with_matching_firefox_versions(base_url, selenium):
     selenium.get("about:blank")
     with open("utils/filter_expression.js") as js:
         with selenium.context(selenium.CONTEXT_CHROME):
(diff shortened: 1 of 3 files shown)


app/tests/integration/test_filter_expression_validation.py
```diff
@@ -1,16 +1,12 @@
...
     with open("utils/filter_expression.js") as js:
         with selenium.context(selenium.CONTEXT_CHROME):
             script = selenium.execute_script(js.read(), 80.1, 80)
     assert script is True
 
 
-def test_filter_expressions_with_mismatching_firefox_versions(
-    base_url, selenium
-):
+def test_filter_expressions_with_mismatching_firefox_versions(base_url, selenium):
```
emtwo, line 9: Is this just linting? Surprised it didn't get caught sooner.

emtwo (approved):
lgtm

app/tests/integration/test_filter_expression_validation.py
jaredlockhart: Yeah I think what's happening is black is set to format that path but not check that path otherwise it shoudln't have landed in main.  I'm gonna fix that now.

PR #4290, opened 2020-12-16
fix #4185 feat(nimbus): validate population percent 0 < x <= 100

### LINKED ISSUE #4185: Enforce 0 < populationPercent <=100 in BE
#4156 covers client-side validation for this but we should enforce this at the server level as well.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-706)
┆Issue Number: EXP-706

jaredlockhart:
Because

* Population percent has to be greater than 0 and less than or equal to 100 to launch

This commit

* Ensures the field is set with a default of 0.0 which is valid during edit time but not to launch
* Bounds checks population percent 0 <= x <= 100 during edit time
* Requires non zero to launch


lmorchard (approved):
LGTM

PR #4302, opened 2020-12-18
fixes #4230 fix(nimbus): Handle error responses from /v3 aPI

### LINKED ISSUE #4230: General application error on /results
I get a general application error when going to https://experimenter.services.mozilla.com/nimbus/first-aboutnewtab-nightly-aa/results

image

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-742)
┆Issue Number: EXP-742

k88hudson:
I think this should fix this issue with application errors on PageResults  – any suggestions about a test for this @emtwo ?

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
@@ -65,9 +65,14 @@ const AppLayoutWithExperiment = ({
     stopPolling,
     review,
   } = useExperiment(slug);
-  const { execute: fetchAnalysis, result: analysis } = useAnalysis();
+  const {
+    execute: fetchAnalysis,
+    status: analysisStatus,
+    result: analysis,
+  } = useAnalysis();
   const [analysisFetched, setAnalysisFetched] = useState<boolean>(false);
-  const status = getStatus(experiment);
+
+  const experimentStatus = getStatus(experiment);
 
   // If the redirect prop function is supplied let's call it with
   // experiment status, review, and analysis details. If it returns
@@ -76,7 +81,7 @@ const AppLayoutWithExperiment = ({
   if (
     !loading &&
     redirect &&
-    (getRedirect = redirect!({ status, review, analysis }))
+    (getRedirect = redirect!({ status: experimentStatus, review, analysis }))
   ) {
     redirectPath = `${BASE_PATH}/${slug}/${getRedirect}`;
   }
@@ -88,11 +93,11 @@ const AppLayoutWithExperiment = ({
   }, [redirectPath]);
 
   useEffect(() => {
-    if (!analysisFetched && !loading && status.locked) {
+    if (!analysisFetched && !loading && experimentStatus.locked) {
       fetchAnalysis([experiment?.slug]);
       setAnalysisFetched(true);
     }
-  }, [fetchAnalysis, loading, experiment, analysisFetched, status]);
+  }, [fetchAnalysis, loading, experiment, analysisFetched, experimentStatus]);
 
   useEffect(() => {
     if (polling && experiment) {
@@ -103,7 +108,7 @@ const AppLayoutWithExperiment = ({
     };
   }, [startPolling, stopPolling, experiment, polling]);
 
-  if (loading || (analysisRequired && !analysis)) {
+  if (loading || (analysisRequired && analysisStatus === "loading")) {
     return <PageLoading />;
   }
 
@@ -114,7 +119,9 @@ const AppLayoutWithExperiment = ({
   const { name } = experiment;
 
   return (
-    <Layout {...{ sidebar, children, review, analysis, status }}>
+    <Layout
+      {...{ sidebar, children, review, analysis, status: experimentStatus }}
+    >
       <section data-testid={testId}>
         <Head
           title={title ? `${experiment.name} – ${title}` : experiment.name}
@@ -124,7 +131,7 @@ const AppLayoutWithExperiment = ({
           {...{
             slug,
             name,
-            status,
+            status: experimentStatus,
           }}
         />
         {title && (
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -50,11 +47,12 @@ const AnalysisAvailable = ({
   analysis,
 }: {
   experiment: getExperiment_experimentBySlug;
-  analysis: AnalysisData | undefined;
+  analysis: AnalysisData;
```
k88hudson, line 50: This should never be rendered if analysis is falsey, so I thought it would be better to type it that way

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -50,11 +47,12 @@ const AnalysisAvailable = ({
...
 }) => {
   const slugUnderscored = experiment.slug.replace(/-/g, "_");
   return (
     <>
+      <MonitoringLink {...experiment} />
```
k88hudson, line 55: Now that the summary page has its own monitoring link this was getting duplicated

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -103,7 +108,7 @@ const AppLayoutWithExperiment = ({
     };
   }, [startPolling, stopPolling, experiment, polling]);
 
-  if (loading || (analysisRequired && !analysis)) {
+  if (loading || (analysisRequired && analysisStatus === "loading")) {
```
jodyheavener, line 111: FWIW the useAsync method does have a `loading` property. But I'm curious: One thing that is hard to test locally but I know @emtwo had done some work to try and avoid was flashing an error/unavailable state on the page if analysis results were still loading, and because the experiment request finishes loading at a separate time than the analysis request, I have a feeling this behavior may be re-introduced with this change. [This PR](https://github.com/mozilla/experimenter/pull/4233/files) has more details of those changes if you're curious.

jodyheavener (commented):
Hmm, from looking at stage it seems like the error is coming from trying to access the object keys of `analysis?.other_metrics!`, when the network request returned:

```
{"show_analysis":true,"daily":null,"weekly":null,"overall":null}
```

So it seems like the loader in `AppLayoutWithExperiment` was working correctly, we just need to check for the existence of `other_metrics` (which apparently does not exist in that object at all?).

(As an aside, I don't know why `show_analysis` would be true, but all the data properties would be `null`.)

@k88hudson what do you think?

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
k88hudson: I noticed when testing this locally that (in addition to the issue with trying access the `undefined` object) if the server returns a `500` response and `analysis` is `undefined`,  `PageLoading` will show up forever – I'll take a look at the PR and see if I can reproduce though

LZoog:
> So it seems like the loader in `AppLayoutWithExperiment` was working correctly, we just need to check for the existence of `other_metrics` (which apparently does not exist in that object at all?).

This is what I was thinking as well, I'd left a comment here: https://github.com/mozilla/experimenter/issues/4230#issuecomment-743261664

Seems like we just need to check that another property is available?

> (As an aside, I don't know why show_analysis would be true, but all the data properties would be null.)

@emtwo left a comment here: https://github.com/mozilla/experimenter/issues/4230#issuecomment-743287022

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
k88hudson: Oh yeah hm, the flashing does seem to be an issue. Lemme look again

jodyheavener: Yeah so my thinking with the 500 error issue causing indefinite loading is that there shouldn’t be a non-addressable scenario where that is occurring - outlined in more detail here: https://github.com/mozilla/experimenter/pull/4233#issuecomment-743397223

but I also recognize “it shouldn’t happen” doesn’t equal “we shouldn’t try to handle it”, so if there’s a better solution I’m super game

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -32,16 +32,13 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
...
-        <MonitoringLink {...experiment} />
-        {analysis?.show_analysis ? (
-          <AnalysisAvailable {...{ experiment, analysis }} />
-        ) : (
-          <AnalysisUnavailable {...{ experiment, error: analysis == null }} />
-        )}
-      </>
-    )}
+    {({ experiment, analysis }) =>
+      analysis?.show_analysis ? (
```
LZoog, line 36: This is the line @jodyheavener and I think needs a tweak. Looking at our mocked data here, is the fix as simple as changing this line to `analysis?.show_analysis && analysis?.overall ? (` or even just changing our check to `.overall` since @emtwo pointed out this should now always be true?

emtwo: yes, I was just going to say this! I just tested it out locally and adding that extra check here for `analysis?.overall` fixes the issue.

emtwo: For context, to test it locally, I created one of the experiments that has this issue in prod: `first-aboutnewtab-nightly-aa` using the admin. I made sure its status is set to complete. The viz API returns an object with `overall: null` so that check here makes it fall back to analysis unavailable.

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
k88hudson: Ok, I *think* the last commit I pushed works for this  by handling the loading and never-attempted-to-fetch

emtwo: I'm not seeing the updates for some reason? Did the push go through?

LZoog:
Closing this in favor of #4350

PR #4307, opened 2020-12-18
fixes mozilla/jetstream#385: GS_LOCATION statistics prefix

scholtzan:
Depends on https://github.com/mozilla/jetstream/pull/386 (so they should be merged together)

### DIFF
--- a/app/experimenter/settings.py
+++ b/app/experimenter/settings.py
@@ -405,5 +405,6 @@ KINTO_COLLECTION_NIMBUS_MOBILE = "nimbus-mobile-experiments"
 DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
 GS_PROJECT_ID = "experiments-analysis"
 GS_BUCKET_NAME = "mozanalysis"
+GS_LOCATION = "statistics"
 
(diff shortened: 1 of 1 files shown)


app/experimenter/settings.py
```diff
@@ -405,5 +405,6 @@
 DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
 GS_PROJECT_ID = "experiments-analysis"
 GS_BUCKET_NAME = "mozanalysis"
+GS_LOCATION = "statistics"
```
scholtzan, line 408: I'm not very familiar with this code base, but this should be the only changed required based on the docs: https://django-storages.readthedocs.io/en/latest/backends/gcloud.html

emtwo (approved):
This looks good to me, but I want to be somewhat cautious as it's a change that could potentially break the working dashboards if there is an issue and it's Friday before a bunch of people will be out for a couple of weeks for holidays.

E.g. is the data all migrated already to the new stats folder so it can be looked up from there? We need to do that first so all the same data is still available.

scholtzan:
The data needs to be copied over into the `statistics/` sub directory, but that's an easy thing to do. 

I agree, it is best to wait until next year with merging these changes.

scholtzan:
@emtwo Can we go ahead with merging this?

PR #4311, opened 2020-12-18
Create codeql-analysis.yml

clouserw:
Now that we have Actions enabled, this built-in code scanning might be interesting.


jodyheavener (approved):
Woo! Guessing this needs to be merged to take effect so fire away

PR #4288, opened 2020-12-16
chore(project): add colour output to make check

jaredlockhart:
Because

* GNU Parallel suppresses the colourized output of all the tests/lints
* I keep missing failed python tests because of the way pytest formats the end of its output it's hard to visually distinguish between passes and failures

This commit

* Adds a bright green PASSED and bright red FAILED to the end of make check

PR #4325, opened 2020-12-21
fix(all): fix check; address various broken tests, outdated docs, failing lints

jodyheavener:
When we added the color output to test results a week ago it stopped producing the failure exit code CircleCI needed to fail the `check` job. As a result our tests have been unknowingly failing for various reasons.

This PR:

- Adds the exit code to the failure portion of the color output commands
- Updates docs with `make generate_docs`
- Fixes a handful of test and linting errors in Nimbus
- Temporarily disables a Django test (TODO [EXP-786](https://jira.mozilla.com/browse/EXP-786))


jodyheavener:
LIES!!!

Screen Shot 2020-12-21 at 4 12 51 PM

LZoog (approved):
Wheeee thanks for fixing these, checked into your branch and LGTM. I wonder why they weren't caught by CI to begin with 🙈

jodyheavener:
I don't know if I should merge this. I cleaned up the JS linting/test issues but I get the above Django failure locally still.

jodyheavener:
gonna see if I can nail down why it's failing

jodyheavener:
Oh actually the failure is visible in CI as well, it's just not being recognized https://app.circleci.com/pipelines/github/mozilla/experimenter/7827/workflows/02259b3b-8e59-4456-83f4-f5073809e067/jobs/20197

LZoog (approved):
```
-------------------------------------------------
All Checks Passed
-------------------------------------------------
```

Looked again since you pushed changes and perhaps I missed something with my last r+, but things definitely pass now. Thanks for figuring these fixes out and for filing EXP-786. I'm sure this is fine to merge, but it'd probably be good to have @jaredlockhart take a looksee after we get back from break. I'm still not sure how CI has been passing on PRs.

jodyheavener:
> I'm still not sure how CI has been passing on PRs.

Basically our tests produce a code when they run. You can see in my most recent run that `CircleCI received exit code 0`, which means things are great! But if you go back to a previous run, where there was a test failure it `received exit code 2`, which CircleCI knows to interpret as a job failure.

The reason the CircleCI job was passing even though the tests were failing (to my understanding; may be slightly off) was because when we added the color output it [wrapped](https://github.com/mozilla/experimenter/pull/4288/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R66) the commands in parens and performed an operation after the commands finished, which swallowed up the exit code. This meant that every `check` job had a [good] [exit code of `0`](https://app.circleci.com/pipelines/github/mozilla/experimenter/7757/workflows/d6627355-8737-46d7-99eb-2876526a79da/jobs/19940). The fix was to manually use a non-zero exit code in the failure message.

pdehaan:
And now all we need are some failing tests with a non-zero exit code…

PR #4223, opened 2020-12-10
fixes #3758 #3757 feat(nimbus): redirect pages based on status, review, and analysis

jodyheavener:
Closes #3758
Closes #3757

This PR:

- Adds a new `redirect` function prop to `AppLayoutWithExperiment` where you can inspect the status, review state, and analysis data of an experiment to determine if the user should be redirected away from the page
- Updates all the page components that render `AppLayoutWithExperiment` to perform their own checks to enforce these redirects.

PR #4350, opened 2021-01-05
fixes #4230 fix(nimbus): Results page analysis unavailable error

### LINKED ISSUE #4230: General application error on /results

LZoog:
Because:
* A general application error is shown on the Results page when results are unavailable instead of redirecting the user or displaying the error state.

This commit:
* Adds analysisAvailable helper checking for 'overall' as well as 'analysis_available' in fetched data
* Checks for error from visualization API call and displays this state in the sidebar instead of on page since the user should be redirected to the Design page

---

This PR fixes #4230 by checking for `overall !== null` (as well as `show_analysis === true`) when checking if analysis data is available, per [this comment](https://github.com/mozilla/experimenter/pull/4302/files#r546082662) / https://github.com/mozilla/experimenter/issues/4230#issuecomment-743287022.

Additionally, I explored addressing [this problem](https://github.com/mozilla/experimenter/pull/4302/files#r546079822), which [you can visually see hangs on "loading" Storybook](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/d58dbc127d95f54e142d233507a69321b02d6d81/nimbus-ui/index.html?path=/story/pages-results--analysis-fetch-errors) when the visualization fetch fails which lead me to question if (looking at Storybook) the "analysis unavailable" and "analysis fetch errors" state should even be handled on the Results page. Since we have a redirect in place to the Design page if the analysis is unavailable, these states should not be reachable. I've moved handling of this into the sidebar.

--

So in bulleted point form, (and to make it easier for @AnaMedinac to check this out), **besides the bug fix** -

Before this PR:
* a visualization data fetch fail was supposed to [show this screen](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/db8fcebc619ad03a61493455e99d62ed16813af6/nimbus-ui/index.html?path=/story/pages-results--analysis-fetch-errors) with the "contact #cirrus" text, but regression from a different bug caused [this screen to show](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/d58dbc127d95f54e142d233507a69321b02d6d81/nimbus-ui/index.html?path=/story/pages-results--analysis-fetch-errors) with the loader indefinitely
* if the analysis was unavailable and the Results page was hit, [this screen](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/d58dbc127d95f54e142d233507a69321b02d6d81/nimbus-ui/index.html?path=/story/pages-results--analysis-unavailable) was supposed to be shown, but as of #4223, this state isn't accessible because the user is redirected to the Design page

After this PR:
* a visualization data fetch fail ([see Storybook](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/504af50f4705f75966814dd50301bf3376444607/nimbus-ui/index.html?path=/story/components-applayoutwithsidebar--analysis-results-error)) redirects the user to the Design page with the error handled in the sidebar component with a "contact #cirrus" message
* The "analysis unavailable" state is removed from the Results page since the user is redirected, the user can only access the Results page if the fetch was successful and results are available

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, { useEffect, useState } from "react";
+import React, { useEffect } from "react";
 import { navigate, RouteComponentProps, useParams } from "@reach/router";
 import AppLayoutWithSidebar from "../AppLayoutWithSidebar";
 import HeaderExperiment from "../HeaderExperiment";
@@ -35,6 +35,7 @@ type AppLayoutWithExperimentProps = {
     status,
     review,
     analysis,
+    analysisError,
   }: {
     status: StatusCheck;
     review?: {
@@ -42,6 +43,7 @@ type AppLayoutWithExperimentProps = {
       invalidPages: string[];
     };
     analysis?: AnalysisData;
+    analysisError?: Error;
   }) => string | void;
 } & RouteComponentProps;
 
@@ -65,9 +67,14 @@ const AppLayoutWithExperiment = ({
     stopPolling,
     review,
   } = useExperiment(slug);
-  const { execute: fetchAnalysis, result: analysis } = useAnalysis();
-  const [analysisFetched, setAnalysisFetched] = useState<boolean>(false);
+  const {
+    execute: fetchAnalysis,
+    result: analysis,
+    error: analysisError,
+    status: analysisFetchStatus,
+  } = useAnalysis();
   const status = getStatus(experiment);
+  const analysisFetched = analysisFetchStatus !== "not-requested";
 
   // If the redirect prop function is supplied let's call it with
   // experiment status, review, and analysis details. If it returns
@@ -90,7 +97,6 @@ const AppLayoutWithExperiment = ({
   useEffect(() => {
     if (!analysisFetched && !loading && status.locked) {
       fetchAnalysis([experiment?.slug]);
-      setAnalysisFetched(true);
     }
   }, [fetchAnalysis, loading, experiment, analysisFetched, status]);
 
@@ -103,7 +109,7 @@ const AppLayoutWithExperiment = ({
     };
   }, [startPolling, stopPolling, experiment, polling]);
 
-  if (loading || (analysisRequired && !analysis)) {
+  if (loading || (analysisRequired && !analysisFetched)) {
     return <PageLoading />;
   }
 
@@ -114,7 +120,7 @@ const AppLayoutWithExperiment = ({
   const { name } = experiment;
 
   return (
-    <Layout {...{ sidebar, children, review, analysis, status }}>
+    <Layout {...{ sidebar, children, review, analysis, analysisError, status }}>
       <section data-testid={testId}>
         <Head
           title={title ? `${experiment.name} – ${title}` : experiment.name}
@@ -144,11 +150,19 @@ type LayoutProps = {
   status: StatusCheck;
   review: ExperimentReview;
   analysis?: AnalysisData;
+  analysisError?: Error;
 };
 
-const Layout = ({ sidebar, children, review, status, analysis }: LayoutProps) =>
+const Layout = ({
+  sidebar,
+  children,
+  review,
+  status,
+  analysis,
+  analysisError,
+}: LayoutProps) =>
   sidebar ? (
-    <AppLayoutWithSidebar {...{ status, review, analysis }}>
+    <AppLayoutWithSidebar {...{ status, review, analysis, analysisError }}>
       {children}
     </AppLayoutWithSidebar>
   ) : (
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.test.tsx
@@ -21,6 +21,7 @@ const { mock } = mockExperimentQuery("my-special-slug");
 const Subject = ({
   status = NimbusExperimentStatus.DRAFT,
   withAnalysis = false,
+  analysisError,
   review,
 }: RouteComponentProps & {
   status?: NimbusExperimentStatus;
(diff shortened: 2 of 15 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.tsx
```diff
@@ -106,9 +110,20 @@ export const AppLayoutWithSidebar = ({
                     </LinkNav>
                   ) : (
                     <DisabledItem name="Results" testId="show-no-results">
-                      {status?.accepted
-                        ? "Waiting for experiment to launch"
-                        : "Experiment results not yet ready"}
+                      {status?.accepted ? (
+                        "Waiting for experiment to launch"
+                      ) : analysisError ? (
```
LZoog, line 115: I am normally not a fan of nested ternaries, but the logic seemed clear enough here.

LZoog: @pdehaan Thanks for your comment!

I dug a bit deeper into what `analysisError` _is_ - unfortunately it's a generic "the thing didn't work" catch-all. It's the same `SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data` message if I force a `500` internal server error error as well as if the API call `404`s because we attempt to JSON parse the response from the server no matter what.

The package we're using to fetch the data in a React hook (and receive `analysisError` from), `react-async-hook`, [doesn't actually hook (pun intended) into HTTP here](https://github.com/slorber/react-async-hook/issues/20#issuecomment-574543580), so I can't parse the error code/status from `analysisError`.

If we want to display the message from the server or even display the error code to help see what's going on, we'll need a little refactoring. I actually spent half an hour attempting it, as I thought I could throw a `try`/`catch` into our `fetchData` function (supplied to the aforementioned package) to use _that_ caught error instead which should contain the error code and message, but couldn't get it working like I wanted.

I'm hoping that with the bug fix included, this scenario won't occur very often at all. When it does, anything going awry on the server-side should already be sent off to Sentry as well as anything that creeps up into our client-side `AppErrorBoundary`.

Since the bug fix here is highest priority and this other stuff is extra anyway, let's make a follow up ticket to see if we can capture the actual error from the HTTP request with code & message?

With that said, I'm unsure if data science is who should be contacted if this errors. Could be a bug on their side, but also on ours. 🤷‍♀️

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -27,125 +24,79 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
...
-      />
-      <TableHighlightsOverview
-        {...{ experiment }}
-        results={analysis?.overall!}
-      />
+    {({ experiment, analysis }) => {
+      // For testing - users will be redirected if the analysis is unavailable
+      // before reaching this return, but tests reach this return and
+      // analysis.overall is expected to be an object
+      if (!analysisAvailable(analysis)) return <></>;
```
LZoog, line 36: This could also be `analysis?.overall`, or stored in a var to only run the function once (it's super lightweight though) but finding a work around to remove these lines would be much nicer. 🙃

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -76,7 +82,7 @@ const AppLayoutWithExperiment = ({
   if (
     !loading &&
     redirect &&
-    (getRedirect = redirect!({ status, review, analysis }))
+    (getRedirect = redirect!({ status, review, analysis, analysisError }))
```
jodyheavener, line 85: I don't think `analysisError` is being used in any of the `redirect` prop calls

jodyheavener (approved):
Couple minor comments, but I wouldn't consider them blockers.

Optional: I also found a way we could eliminate a state check in the code and rely entirely on the status of the `useAsync` hook to determine whether or not we should show the Loading screen while looking up the analysis. Details: https://gist.github.com/jodyheavener/ae2ed26446a54f0050fae0c3798375a1

Aaaand, as we discussed, it feels odd to have that early return in `PageResults`, but I haven't been able to come up with a better solution (my attempt at blocking rendering while we performed redirect operations failed) and I know you've spent a good chunk of time on it. So let's use it. Good opportunity for a maintenance/inno sprint.

AnaMedinac:
@LZoog thanks for the bulleted points, made it much easier to follow along. 

I just have a suggestion to shorten the copy to: "Could not get experiment data. Please contact DS in #cirrus". Otherwise, this works for me.

app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.test.tsx
```diff
@@ -199,6 +202,17 @@ describe("AppLayoutWithSidebar", () => {
...
     });
 
+    it("when complete and analysis results fetch errors", async () => {
+      render(
+        <Subject status={NimbusExperimentStatus.COMPLETE} analysisError />,
+      );
+
+      expect(screen.queryByTestId("show-no-results")).toBeInTheDocument();
+      expect(screen.queryByTestId("show-no-results")).toHaveTextContent(
+        "Could not fetch experiment analysis data. Please contact data science",
```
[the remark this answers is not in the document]
LZoog: I only took part of the string because of the external link, the test receives "Open in new window" as well as the SVG so I didn't match it precisely.

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
LZoog: Oh yep, good catch

app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.tsx
LZoog: Ana confirmed the text here (it's a shortened version of what it was) - @pdehaan if you think it would be valuable to display the error message from the server instead, we can file a follow up, but otherwise I'll leave it as-is and we can see how often this actually occurs when it's in the wild.

Issue #4298, opened 2020-12-17
Population percent of 0.0001 not considered valid


data-sync-user:
It looks like if you set your Audience population percent to 0.0001 this is considered invalid even though the server says you can use it. Under the hood the GQL readyForReview check is saying "Ensure this value is greater than or **equal to** 0.0001". 0.0002 does work, so my guess it's actually checking greater than and not equal to.

As an aside this should also be showing that message as an error on the field, but that's a separate issue (EXP-773).

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-774)
┆Issue Number: EXP-774

data-sync-user:
➤ Jody Heavener commented:

Just poked at this and sure enough if I change the `population_percent` serializer from `min_value=0.0001` to `min_value=0.00009` then 0.0001 is accepted. I don't know how Django serializers work, but documentation says what we have should be working.

Issue #4291, opened 2020-12-17
Display server-generated messages when determining review-readiness


data-sync-user:
We are currently using static messages when displaying the icon/messaging for when a field is blocking the experiment from going into review. The server provides us with these messages when querying the experiment, under the readyForReview property, and because they can vary we should be using them.

Update (and possibly rename) the [isMissingField](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/hooks/useExperiment.tsx#L57:L57) function provided by the useExperiment hook to pass along the messages for the given field name, and update all instances where it should print these messages.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-773)
┆Issue Number: EXP-773

Issue #4289, opened 2020-12-16
Update UI of review-readiness message for individual branch fields


lmorchard:
In EXP-773 we updated the [isMissingField](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/hooks/useExperiment.tsx#L57:L57) function provided by the useExperiment hook to start passing along server-generated review messages. Currently individual review-readiness messages for branch fields are [displayed](https://github.com/mozilla/experimenter/pull/4283#discussion_r544423148) in the top right of a branch’s container.

Update the individual branch fields to display review-readiness messages using the new consolidated inline-icon and visible feedback style we’re using.

There is one review message for when the entire control branch is missing/null - this can be left alone here as it will be addressed in EXP-978.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-768)
┆Issue Number: EXP-768

PR #4358, opened 2021-01-05
chore(nimbus): refactor graphql into a single mutation

jaredlockhart:
Because

* While implementing changes to the graphql API to make all mutations partial, I discovered it would be easier if we had a single mutation rather than many and have to update many inputs, serializers, queries, etc

This commit

* Flattens all mutations into a single partial mutation on the experiment object which gets us most of the way to having fully partial mutations per page, I think we might need to do a bit extra on top of this as an immediate followup but this should get most of the way there.
* Plus I cleaned a few things up like
  - Making sure we use the same key for experiment id everywhere (instead of id sometimes and nimbus_experiment_id others)
  - Making sure nimbus experiment id is always a number instead of sometimes a number and sometimes a string
  - I noticed ids for other objects are sometimes strings but I can do that as a followup it's filed as a separate ticket

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -14,32 +14,7 @@ from experimenter.experiments.models.nimbus import (
 )
 
 
-class NimbusStatusRestrictionMixin:
-    required_status = NimbusExperiment.Status.DRAFT
-
-    def validate(self, data):
-        data = super().validate(data)
-        if self.instance and self.instance.status != self.required_status:
-            status = self.instance.status
-            raise serializers.ValidationError(
-                {
-                    "experiment": [
-                        f"Nimbus Experiment has status '{status}', but can only "
-                        f"be changed when set to '{self.required_status}'."
-                    ]
-                }
-            )
-        return data
-
-
-class NimbusChangeLogMixin:
-    def save(self, *args, **kwargs):
-        experiment = super().save(*args, **kwargs)
-        generate_nimbus_changelog(experiment, self.context["user"])
-        return experiment
-
-
-class NimbusBranchSerializer(NimbusChangeLogMixin, serializers.ModelSerializer):
+class NimbusBranchSerializer(serializers.ModelSerializer):
     def validate_name(self, value):
         slug_name = slugify(value)
         if not slug_name:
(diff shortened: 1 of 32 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -319,9 +267,42 @@ def validate_channel(self, value):
         return value
 
 
-class NimbusStatusUpdateSerializer(
-    NimbusChangeLogMixin, NimbusStatusRestrictionMixin, serializers.ModelSerializer
+class NimbusExperimentUpdateSerializer(
+    NimbusChangeLogMixin,
```
lmorchard, line 271: Huh, I really need to read up / get refreshed on how multiple inheritance here works with REST Framework. Really nice that all of that can just be converted into mixins with `validate` methods!

app/experimenter/nimbus-ui/src/gql/experiments.ts
```diff
@@ -21,8 +21,8 @@ export const CREATE_EXPERIMENT_MUTATION = gql`
 `;
 
 export const UPDATE_EXPERIMENT_OVERVIEW_MUTATION = gql`
-  mutation updateExperimentOverview($input: UpdateExperimentInput!) {
-    updateExperimentOverview(input: $input) {
+  mutation updateExperimentOverview($input: ExperimentInput!) {
```
lmorchard, line 24: Might be nice to collapse all these mutations into one on the client-side at some point, but this seems fine enough to minimize impact of the changes for now

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -319,9 +267,42 @@ def validate_channel(self, value):
...
+    primary_probe_set_ids = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=NimbusProbeSet.objects.all(), required=False
+    )
+
+    secondary_probe_set_ids = serializers.PrimaryKeyRelatedField(
+        many=True, queryset=NimbusProbeSet.objects.all(), required=False
+    )
+    population_percent = serializers.DecimalField(
+        7, 4, min_value=0.0, max_value=100.0, required=False
+    )
```
lmorchard, line 304: This *might* be a pre-existing bug that I just noticed:

When filling everything out, using `0.00002` for `population_percent` on the Audience page results in a missing details warning preventing review - but the page isn't named in the sidebar.

I see it properly coming up in the `readyForReview` property of the `getExperiment` GraphQL response.  (i.e. too many decimal places). That what makes me think it's not this PR doing it. That, and I can't find any change here that would cause it.

A fix is probably to add `population_percent` to the `fieldPageMap.audience` list here: 

https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/hooks/useExperiment.tsx#L12-L18

lmorchard: Another probably unrelated-to-this-PR issue:

Just noticed that I can submit `0.00007` for `population_percent` - client-side seems to report the save succeeded, but there are submit errors reported in the GQL response. I think there's probably a snake_case / camelCase mismatch going on with our new `useCommonForm` hook - i.e. `population_percent` vs `populationPercent`

lmorchard (approved):
Gave this a good hard stare and failed to find any obvious gotchas. 

Found a couple of problems that I think are actually unrelated to this PR worth follow-up issues.

Really expected this to end up being more complicated than a bunch of renaming of things. Guess our mutations were just that redundant and GraphQL that magical.

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Huzzah!

app/experimenter/nimbus-ui/src/gql/experiments.ts
jaredlockhart: Yeah I didn't wanna mess with too much structure, but yes that would be interesting.

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Oh interesting.  Are there issues filed for those?

lmorchard: Hmm, now that I dig around, I think arguably these problems fall under issue #4289 and #4291, maybe also adjacent to #4298

jaredlockhart: Oh good yes, so if we finish those off and still find weirdness we can file additional followups.  Thanks for checking @lmorchard !

PR #4366, opened 2021-01-06
fix #4365 bug(nimbus): update branch deletion logic

### LINKED ISSUE #4365: Deleting branches is broken
I was working on the partial submission stuff and I realized our branch deleting logic is all wonky.  The branch foreign keys on to experiment, which means deleting a branch will also delete the experiment, which will then delete all its other related entities.  We didn't notice up to this point because the way the logic is specifically laid out results in the experiment being deleted, all its related entities being deleted, and then immediately resaved to the same ID.  The most telling symptom would have been all the change logs vanishing but they're not exposed through the UI yet.

So this needs some rethinking before we can proceed with the partial saving stuff.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-799)
┆Issue Number: EXP-799

jaredlockhart:
Because

* Experiment fks onto experiment so if you delete an experiments branches, it also deletes the experiment and all its related entities

This commit

* Adds a convenience method to unset the reference_branch fk before deleting branches
* Updates tests
* I noticed we were using input in our tests which is a reserved word in python so I updated that to data

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -180,25 +180,28 @@ class NimbusExperimentBranchMixin:
     def update(self, experiment, data):
         control_branch_data = data.pop("reference_branch", {})
         treatment_branches_data = data.pop("treatment_branches", {})
-        instance = super().update(experiment, data)
 
-        if control_branch_data or treatment_branches_data:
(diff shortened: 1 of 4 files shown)


lmorchard (approved):
:+1: yup, that looks like it's deleting things the right way round now

PR #4380, opened 2021-01-08
fix #4376 feat(nimbus): Add No Channel, No Version, No Targeting options

### LINKED ISSUE #4376: Add empty defaults for channel, version, targeting
Because each of channel, version, targeting, may not apply to a given experiment depending on its platform, we should add an empty default for each of these fields and not block launching on them being set.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-809)
┆Issue Number: EXP-809

jaredlockhart:
Because

- Not all experiments will require a channel, version, or targeting

This commit

- Adds No Channel, No Version, No Targeting options to each of those fields
- Changes the db defaults to sane defaults
- Move the default duration/enrollment values into the database rather than the frontend


app/experimenter/experiments/migrations/0149_auto_20210108_1427.py
```diff
@@ -0,0 +1,139 @@
...
+                ],
+                default="",
+                max_length=255,
+            ),
+        ),
+        migrations.AlterField(
+            model_name="nimbusexperiment",
+            name="hypothesis",
+            field=models.TextField(
+                default="If we <do this/build this/create this change in the experiment> for <these users>, then we will see <this outcome>.\nWe believe this because we have observed <this> via <data source, UR, survey>.\n\nOptional - We believe this outcome will <describe impact> on <core metric>\n    "
```
jodyheavener, line 92: I added trimming in the client to deal with it, but is there a particular reason we have the extra newline and spaces at the end of this?

jodyheavener (approved):
nice!

app/experimenter/experiments/migrations/0149_auto_20210108_1427.py
jaredlockhart: Oh good question probably not 🤷‍♀️

PR #4381, opened 2021-01-08
fix #4378 bug(nimbus): don't delete branches if they're not being updated

### LINKED ISSUE #4378: Saving any pages deletes branches
Ah.  Now that we use a single serializer for all pages, and it's designed to delete all branches on save and replace them with newly created ones, if you save on ANOTHER page it also deletes all branches.  Sploops.  So this means we need to distinguish calls that intend to update branches from calls that do not.  Might be able to check for the presence of the key in the submitted data dict, but that might also not work.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-810)
┆Issue Number: EXP-810

jaredlockhart:
Because

- We recently folded all the experiment serializers into one
- But neglected to only delete branches if they're also going to be updated

This commit

- Only deletes branchces if they're also going to be updated

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -178,12 +178,14 @@ class NimbusExperimentBranchMixin:
         return data
 
     def update(self, experiment, data):
-        control_branch_data = data.pop("reference_branch", {})
-        treatment_branches_data = data.pop("treatment_branches", {})
-
         with transaction.atomic():
+            if set(data.keys()).intersection({"reference_branch", "treatment_branches"}):
+                experiment.delete_branches()
+
+            control_branch_data = data.pop("reference_branch", {})
+            treatment_branches_data = data.pop("treatment_branches", [])
+
             experiment = super().update(experiment, data)
-            experiment.delete_branches()
 
             if control_branch_data:
                 experiment.reference_branch = NimbusBranch.objects.create(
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -178,12 +178,14 @@ def validate(self, data):
         return data
 
     def update(self, experiment, data):
-        control_branch_data = data.pop("reference_branch", {})
-        treatment_branches_data = data.pop("treatment_branches", {})
-
         with transaction.atomic():
+            if set(data.keys()).intersection({"reference_branch", "treatment_branches"}):
```
lmorchard, line 182: Hmm, does this transaction need to be started at all here, if these keys are absent? Or could `update` just bail out early with a `return experiment` if these are missing?

lmorchard (approved):
LGTM, with one small head-scratch

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Gotta call super

lmorchard: Yeah, but I mean something like:

```python
control_branch_data = data.pop("reference_branch", None)
treatment_branches_data = data.pop("treatment_branches", None)

experiment = super().update(experiment, data)
if control_branch_data is None and treatment_branches_data is None:
    return experiment

with transaction.atomic():
    experiment.delete_branches()
    # etc...
```

PR #4384, opened 2021-01-08
fixes #4379 fix(nimbus): Don't redirect from Results page when analysis is available

### LINKED ISSUE #4379: Results Page Doesn't Load
E.g. `Attribution Targeting Release AA Test` is in the "completed" tab. When I click `results` on the right for it, it redirects to the design page.

I think this is a regression starting in https://github.com/mozilla/experimenter/commit/c756d99b35cc0ddf2cb3d390a08329c4eeb0867e as it seems to work fine locally when I tested in an earlier commit.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-812)
┆Issue Number: EXP-812

LZoog:
Fixes #4379

Because:
* The Results page should display as expected when results are available

This commit:
* Fixes the bug
* Shows the loader on Results page until experiment data and results have both been fetched
* Shows 'results loading' text in the sidebar on the Design page so the page content can be displayed before results are fetched

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
@@ -12,7 +12,7 @@ import TableHighlightsOverview from "../TableHighlightsOverview";
 import TableMetricPrimary from "../TableMetricPrimary";
 import TableMetricSecondary from "../TableMetricSecondary";
 import MonitoringLink from "../MonitoringLink";
-import { analysisAvailable } from "../../lib/visualization/utils";
+import { analysisUnavailable } from "../../lib/visualization/utils";
 
 const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
   <AppLayoutWithExperiment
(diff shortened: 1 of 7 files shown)


app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -10,6 +10,9 @@ import { AnalysisData } from "./types";
 export const analysisAvailable = (analysis: AnalysisData | undefined) =>
   analysis?.show_analysis === true && analysis?.overall !== null;
 
+export const analysisUnavailable = (analysis: AnalysisData | undefined) =>
```
LZoog, line 13: To be clear, this is what fixes the bug. We have to check for `analysis` truthiness.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -24,7 +24,7 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
         return "edit/overview";
       }
 
-      if (!analysisAvailable(analysis)) {
+      if (analysisUnavailable(analysis)) {
```
LZoog, line 27: To be clear, this is what fixes the redirection bug, we have to check for `analysis` truthiness.

LZoog:
FWIW, made #4398 after poking at this since `AppLayoutWithExperiment` is getting a little bulky.

emtwo (approved):
This looks good to me. Tested it locally and the bug is fixed! Yay

PR #4389, opened 2021-01-08
fixes #4305 feat(nimbus): Add Summary component to Design page

### LINKED ISSUE #4305: Make the design page reflect experiment summary
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-783)
┆Issue Number: EXP-783

LZoog:
Because:
* We want the Design page to display all tables summarizing the experiment

This commit:
* Displays the Summary component inside PageDesign

---
fixes #4305
Quick end-of-day Friday task. 😎

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageDesign/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageDesign/index.test.tsx
@@ -27,9 +27,10 @@ describe("PageDesign", () => {
         <PageDesign />
       </RouterSlugProvider>,
     );
-    await waitFor(() => {
+    waitFor(() => {
       expect(screen.queryByTestId("PageDesign")).toBeInTheDocument();
     });
+    expect(screen.queryByTestId("summary")).toBeInTheDocument();
   });
 
   it("redirects to the edit overview page if the experiment status is draft", async () => {
--- a/app/experimenter/nimbus-ui/src/components/PageDesign/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageDesign/index.tsx
@@ -5,6 +5,7 @@
 import React from "react";
 import { RouteComponentProps } from "@reach/router";
 import AppLayoutWithExperiment from "../AppLayoutWithExperiment";
+import Summary from "../Summary";
 
 const PageDesign: React.FunctionComponent<RouteComponentProps> = () => {
   return (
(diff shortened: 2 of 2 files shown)


app/experimenter/nimbus-ui/src/components/PageDesign/index.test.tsx
```diff
@@ -30,6 +30,7 @@ describe("PageDesign", () => {
     await waitFor(() => {
       expect(screen.queryByTestId("PageDesign")).toBeInTheDocument();
     });
+    expect(screen.queryByTestId("summary")).toBeInTheDocument();
```
pdehaan, line 33: Silly Qs:

1. Why is the other TestId on L31 above "PageDesign" but this is all-lowercase "summary"? Or does it even matter (consistency wise, I didn't see a lot of other TestIds in these files to see if one is a more established standard or not.

2. Why does the example above on L30-32 wrap the `expect()` in an `await waitFor(() => {...})` but this `expect()` isn't. The code looks like it does the same thing (waiting for that TestId to be in the document).

jodyheavener: 1. I think this is just an arbitrary pattern we set up at some point, where in some cases we're looking for the outermost container representing the ComponentName and in others we're just looking for a child element. In other projects this has typically been all lowercase, but meh.
2. The first expect is wrapped in a waitFor because some async activities happen (I think in this case we're waiting for a GQL query to resolve) that result in a re-render. If we didn't have it we'd only be able to inspect the loading state. The second expect doesn't need the waitFor because by the time we reach it, the thing we were waiting for has already occurred. And on that note @LZoog I don't think waitFor needs `await`.

jodyheavener (approved):
One comment for you in my response to pdehaan, but it's a non blocker. LGTM.

app/experimenter/nimbus-ui/src/components/PageDesign/index.test.tsx
LZoog: Yeah, we _could_ go and update a bunch for consistency but we mostly went with "match component name"... I suppose `summary` could be `Summary` but we're a bit inconsistent across the board in our test ID names 😅

Yep @jodyheavener we don't need the `await`, will push the change before merging.

pdehaan: Awesome, thanks for the notes.
Yeah, I don't think I'm invested enough in *.test.tsx files to think that we bother refactoring for consistency. If this was the only inconsistent name maybe, but not really worth the effort here (IMO).

PR #4390, opened 2021-01-09
fix #4370: add documentation_links to DB model and GQL query / mutation

### LINKED ISSUE #4370: Add additional links to DB model and experiment query/mutation
We need to store "additional links" titles + urls, add the field to the `experimentBySlug` GQL query so it can be consumed, and add to the `updateExperiment` mutation so it can be updated.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-802)
┆Issue Number: EXP-802

lmorchard:
Because:

- We need to store "additional links" titles + urls

This commit:

- adds a new documentation_links model to the DB, related to experiments

- adds handling of documentation_links updating to serializers

- adds documentation_links to Django admin

- bugfix for Django admin where experiments using feature config with no
  schema could not be saved

- adds documentationLinks to GQL experiment queries

- adds documentationLinks to GQL experiment mutation


lmorchard:
Hitting all the branches on the linting tree on the way down, again.

lmorchard:
Hmm, more things seem broken - switching to draft mode and will return to the drawing board on monday

lmorchard:
Was too done with the day on Friday to see it, but it looks like git decided to delete a few unexpected things out in auto-resolution of merge conflicts when I rebased against main. Opening for review, now that it looks like everything finally passed

lmorchard:
@jodyheavener - you might be interested in this, since I think it blocks a couple issues you've picked up

app/experimenter/experiments/admin/nimbus.py
```diff
@@ -24,11 +25,14 @@ def clean_feature_value(self):
...
-                self.cleaned_data["experiment"].feature_config.schema
-            )
+        feature_schema = None
+        if feature_config and feature_config.schema:
+            try:
+                feature_schema = json.loads(feature_config.schema)
+            except json.JSONDecodeError:
+                pass
 
+        if feature_value and feature_schema:
```
jaredlockhart, line 35: What's this? Is this related?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -272,10 +273,36 @@ def validate_channel(self, value):
...
+            "link",
+        )
+
+
+class NimbusExperimentDocumentationLinkMixin:
+    def update(self, experiment, data):
+        documentation_links_data = data.pop("documentation_links", None)
+        experiment = super().update(experiment, data)
+        if documentation_links_data is not None:
+            with transaction.atomic():
```
jaredlockhart, line 290: Oh yeah this is what I meant the other day sorry I meant to respond in the other PR, I figured `update` should be inside `transaction.atomic` so the whole thing was atomic?  What do you think?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -195,6 +195,24 @@ def __str__(self):
...
+
+    class Meta:
+        verbose_name = "Nimbus Documentation Link"
+        verbose_name_plural = "Nimbus Documentation Links"
+        ordering = ("id",)
+
+    def __str__(self):
+        return "%s (%s)" % (self.title, self.link)
+
+
```
jaredlockhart, line 215: Oh I see, you did it as a 1:M, we did this in the old flow with a text field that people could paste arbitrary links into and then just `urlize` when we display it so they're all clickable.  Do we need to store them independently in the database?

jaredlockhart (commented):
Looks pretty straight forward so far, but if we don't need to store each link independently it'd be a lot simpler to treat it as a text blob?

app/experimenter/experiments/admin/nimbus.py
lmorchard: Related insofar as I couldn't exercise the admin panel without it

jodyheavener:
> I think it blocks a couple issues you've picked up

I was just going to fake it until this was done, but hooray! I spun it up and tried it out and all works as I expected it to.

app/experimenter/experiments/models/nimbus.py
lmorchard: The links are independent title / link fields in the figma mocks

https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/?node-id=217143%3A15592

jodyheavener: > a text field that people could paste arbitrary links into

Not sure that would work since the mocks have us using separate title and link fields.

Screen Shot 2021-01-11 at 12 48 48 PM

But that's just a UI thing; we could still store and parse a single JSON string if you want it in one DB column.

app/experimenter/experiments/api/v5/serializers.py
lmorchard: I wouldn't think the entire experiment update needs to be in the transaction - the delete-and-add is the main atomic concern here I think? And that doesn't need doing at all if the property isn't in the update data

app/experimenter/experiments/models/nimbus.py
lmorchard: Could rework this PR to just store a JSON blob of links on the backend, shift more of the data management to the frontend to still conform to the mocked UX. Or we could rethink the UX? This seemed like the most straightforward option though.

app/experimenter/experiments/admin/nimbus.py
jaredlockhart: Oh yep yes, we should rip that whole thing out that's not needed anymore, I'll file a followup for that.

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Yeah that's what I was thinking about, you definitely don't want the delete succeeding and the related insertions failing or visa versa, but does it make sense for that whole section to fail but the other update fields to succeed?  Or should the whole update fail?  I would think the whole update should fail, otherwise it's a partial update and some fields are saved and others aren't?  Like if you're not doing the related entity logic, it's just a single transaction to update the fields on the parent model, and if one of those fields fails that whole update is thrown out, I would think preserving that behaviour would be desired?

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Oh sorry yes I didn't look at the figma, I should've looked at that first 😅 Yeah I think if we don't need to do any rich logic on these individual links we could simplify this a lot with just a single text field of arbitrary text and highlight links in the UI.  Maybe we can ping @AnaMedinac and review the ui together?

jodyheavener: I'm partial to itemized title/link items personally. At least from a UX perspective it makes it clear what's expected. With a textbox you might not know if you can include regular text or what happens to it if you do use it. Additionally, with a text box you likely lose the ability to assign titles to links.

LZoog: Hmm I like document title/link as well FWIW, they'll be displayed in the Summary component with the title: 
image

But I'm sure Ana would be flexible.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -195,6 +195,24 @@ def __str__(self):
...
+class NimbusDocumentationLink(models.Model):
+    experiment = models.ForeignKey(
+        NimbusExperiment,
+        related_name="documentation_links",
+        on_delete=models.CASCADE,
+    )
+    title = models.CharField(max_length=255, null=False)
+    link = models.URLField(max_length=255, null=False)
+
+    class Meta:
```
jaredlockhart, line 207: We could put a unique_together here on `('title', 'experiment')` to prevent duplicated links on a single experiment.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -195,6 +195,24 @@ def __str__(self):
...
+        verbose_name = "Nimbus Documentation Link"
+        verbose_name_plural = "Nimbus Documentation Links"
+        ordering = ("id",)
```
jaredlockhart, line 210: Given that we delete/recreate every link on every save, id won't be a stable ordering, we should probably use `title`?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -195,6 +195,24 @@ def __str__(self):
...
+
+    def __str__(self):
+        return "%s (%s)" % (self.title, self.link)
```
jaredlockhart, line 213: In general [fstrings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) are preferred for formatting now.

app/experimenter/experiments/models/nimbus.py
jaredlockhart: We discussed this in slack and captured the results [here](https://jira.mozilla.com/browse/EXP-802) tldr we'll use the 1:M approach and potentially change `title` to an enum drop down so we can richly validate/identify given links.

app/experimenter/experiments/tests/api/v5/test_mutation.py
```diff
@@ -268,6 +289,52 @@ def test_update_experiment_error(self):
...
+        content = json.loads(response.content)
+        result = content["data"]["updateExperiment"]
+        self.assertEqual(
+            result["nimbusExperiment"],
+            {
+                "id": experiment.id,
+                "status": experiment.status.upper(),
+                "name": experiment.name,
+                "slug": experiment.slug,
+                "documentationLinks": documentation_links,
```
jaredlockhart, line 325: Will this ordering be stable in this test?  Assertions around lists can randomly reorder depending on what phase the moon is in so it's usually better to do a set assertion.

app/experimenter/experiments/tests/api/v5/test_mutation.py
```diff
@@ -268,6 +289,52 @@ def test_update_experiment_error(self):
...
+        )
+
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        experiment_links = experiment.documentation_links.all()
+        self.assertEqual(len(experiment_links), len(documentation_links))
+        for idx in range(0, len(documentation_links)):
+            link_data = documentation_links[idx]
+            link = experiment_links[idx]
+            for key, val in link_data.items():
+                self.assertEqual(getattr(link, key), val)
```
jaredlockhart, line 336: Once you know the length is equal you can just check membership of each dict rather than looking them up by index.

app/experimenter/experiments/tests/api/v5/test_query.py
```diff
@@ -115,6 +115,39 @@ def test_experiments_with_branches_returns_branch_data(self):
...
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        experiment_data = content["data"]["experiments"][0]
+        self.assertEqual(
+            {b["title"] for b in experiment_data["documentationLinks"]},
+            {b.title for b in documentation_links},
+        )
+        self.assertEqual(
+            {b["link"] for b in experiment_data["documentationLinks"]},
+            {b.link for b in documentation_links},
```
jaredlockhart, line 148: Set assertions are best assertions 💯

app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -233,6 +233,59 @@ def test_branch_name_cant_slugify(self):
...
+            ],
+        }
+        serializer = NimbusExperimentUpdateSerializer(
+            experiment, data=data, partial=True, context={"user": self.user}
+        )
+        self.assertTrue(serializer.is_valid())
+        serializer.save()
+        self.assert_documentation_links(experiment.id, data["documentation_links"])
+
+    def test_serializer_preserves_links_when_absent_in_data(self):
```
jaredlockhart, line 259: Very nice 👍

app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -233,6 +233,59 @@ def test_branch_name_cant_slugify(self):
         )
 
 
+class TestNimbusExperimentDocumentationLinkMixin(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.user = UserFactory()
+
+    def test_serializer_update_links(self):
```
jaredlockhart, line 241: I would also check in here that the previously stored links are deleted.

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -114,6 +115,22 @@ def branches(self, create, extracted, **kwargs):
...
+
+        if isinstance(extracted, Iterable):
+            # A list of links were passed in, use them
+            for link in extracted:
+                self.documentation_links.add(link)
+        else:
+            for i in range(3):
+                self.documentation_links.add(
+                    NimbusDocumentationLinkFactory.create(experiment=self)
+                )
```
jaredlockhart, line 132: Very nice 👍

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -189,6 +190,12 @@ def test_str(self):
         self.assertEqual(str(branch), branch.name)
 
 
+class TestNimbusDocumentLink(TestCase):
+    def test_str(self):
+        doco_link = NimbusDocumentationLinkFactory.create()
+        self.assertEqual(str(doco_link), "%s (%s)" % (doco_link.title, doco_link.link))
```
jaredlockhart, line 196: Yeah you can test that but it's also okay to just mark the `__str__` method as `# pragma: nocover`

jaredlockhart (changes requested):
👏 👏 👏  Very nice @lmorchard thank you!  Covers everything nicely.  Couple small things to look at, but overall looks great!

app/experimenter/experiments/models/nimbus.py
lmorchard: Well, they should be re-created with the order in which they came from the form, which I think should result in sequential IDs? Not the originals but still in order

jaredlockhart: In my experience you can't depend on database id ordering being stable, so I typically use something in the data itself.

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
lmorchard: Eh, we test the name for branches, figured I might as well follow pattern

app/experimenter/experiments/tests/api/v5/test_serializers.py
lmorchard: Adding a `self.assertEqual(len(links_data), len(documentation_links))` which I think should do the trick, since the contents are also asserted

lmorchard: Actually, never mind that, just going to switch to a set assertion here and also over in `test_mutation`

app/experimenter/experiments/tests/api/v5/test_mutation.py
lmorchard: Switching to a set assertion here, which seems to rope everything in more concisely

lmorchard:
> We discussed this in slack and captured the results here tldr we'll use the 1:M approach and potentially change title to an enum drop down so we can richly validate/identify given links.

FWIW, the follow-up for title-as-enum is over here:

https://jira.mozilla.com/browse/EXP-818

app/experimenter/experiments/tests/api/v5/test_mutation.py
lmorchard: Explicitly ordering on ID, so that order should at least be stable

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -272,10 +273,36 @@ def validate_channel(self, value):
...
+class NimbusDocumentationLinkSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = NimbusDocumentationLink
+        fields = (
+            "title",
+            "link",
+        )
+
+
+class NimbusExperimentDocumentationLinkMixin:
```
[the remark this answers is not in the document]
lmorchard: Adding `unique_together` but I'd punt further validation to when we define & add the title enum feature https://jira.mozilla.com/browse/EXP-818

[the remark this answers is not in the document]
jaredlockhart: Yep sounds reasonable.  Django serializers just won't validate uniqueness at validation time, so without doing it explicitly it'll just throw a 500 db constraint violation, but yep we can do it in the follow up 👍

lmorchard:
Force push with some review changes, though I didn't add ordering on title. That could be something to tweak in EXP-818, since we'll be introducing canned titles via enums

lmorchard:
Added a migration, realized I could just replace the single migration, reformatted... hopefully this is the last force-push for a bit

jaredlockhart:
Yeah exactly, prs should only introduce 1 migration each.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -195,6 +195,25 @@ def __str__(self):
...
+    link = models.URLField(max_length=255, null=False)
+
+    class Meta:
+        verbose_name = "Nimbus Documentation Link"
+        verbose_name_plural = "Nimbus Documentation Links"
+        unique_together = (("title", "experiment"),)
+        ordering = ("id",)
+
+    def __str__(self):
+        return f"{self.title} ({self.link})"
```
jaredlockhart, line 214: 👍

app/experimenter/experiments/tests/api/v5/test_mutation.py
```diff
@@ -268,6 +289,54 @@ def test_update_experiment_error(self):
...
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        experiment_links = experiment.documentation_links.all()
+        for key in (
+            "title",
+            "link",
+        ):
+            self.assertEqual(
+                {b[key] for b in documentation_links},
+                {getattr(b, key) for b in experiment_links},
+            )
```
jaredlockhart, line 338: 👍 👍 👍

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
jaredlockhart: Also format strings instead of `%` formatting here ^

jaredlockhart (changes requested):

lmorchard:
One more push to change string formatting in the test

app/experimenter/experiments/models/nimbus.py
jaredlockhart: We should keep our eye out if the ordering in the UI becomes wonky, or if tests intermittently fail.

jaredlockhart (approved):
Awesome!

I filed https://github.com/mozilla/experimenter/issues/4405 as a followup.

Thnx @lmorchard! 🐧

PR #4404, opened 2021-01-11
    fix #4293 chore(nimbus): allow empty values in v5 serializers

### LINKED ISSUE #4293: Allow partial submission for all mutations except create
Each of the four mutations:

Overview

Branches

Metrics

Audience

should allow for partial submission, ie any subset of fields.  Fields should only be required in the readyForReview check, not at form save time.  This change is scoped only to the backend.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-770)
┆Issue Number: EXP-770

jaredlockhart:
Because
    
    - We want to allow empty values for all fields in the v5 API
    - We recently merged all the v5 serializers/mutations
    - We did a first pass on reorganizing them but still left them in a bit of a disorganized state
    
    This commit
    
    - Merges all the fields/validators of the NimbusExperiment serializer into one place
    - Updates all fields to allow empty values
    - Adds additional tests for empty values
    - Renames the mutation/query files to plural to be consistent with the rest of the package
    - Reorganizes the probeset saving behaviour to ensure probesets are not deleted when other fields are updated


jaredlockhart:
Oh most of the changes I listed above were already landed in break out prs, this is just the reorganization part.  I'll update the commit details.

app/experimenter/experiments/api/v5/__init__.py
```diff
@@ -1,6 +1,6 @@
 import graphene
 
-from experimenter.experiments.api.v5.mutation import Mutation
-from experimenter.experiments.api.v5.query import Query
+from experimenter.experiments.api.v5.mutations import Mutation
+from experimenter.experiments.api.v5.queries import Query
```
lmorchard, line 4: Do these need renaming? I guess they don't *not* need renaming.

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -239,38 +177,47 @@ def validate(self, data):
...
+            if set(data.keys()).intersection(
+                {"primary_probe_set_ids", "secondary_probe_set_ids"}
+            ):
                 experiment.probe_sets.clear()
+
+            primary_probe_set_ids = data.pop("primary_probe_set_ids", [])
+            secondary_probe_set_ids = data.pop("secondary_probe_set_ids", [])
+            experiment = super().update(experiment, data)
+
+            if primary_probe_set_ids:
```
lmorchard, line 190: Does this all need to be in a transaction? Looks like we have 2 transactions that start & end for every update, even if branches & probe sets aren't included in the update data

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -298,26 +245,27 @@ def update(self, experiment, data):
         return experiment
 
 
-class NimbusExperimentUpdateSerializer(
-    NimbusChangeLogMixin,
-    NimbusStatusRestrictionMixin,
-    NimbusExperimentOverviewMixin,
-    NimbusExperimentDocumentationLinkMixin,
+class NimbusExperimentSerializer(
```
lmorchard, line 248: Does this need renaming? We've got a serializer for in-progress updates and one for the final review state - I'd think that the ready-for-review serializer would fit the name better but 🤷

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -55,68 +55,6 @@ class Meta:
         )
 
 
-class NimbusChangeLogMixin:
```
lmorchard, line 58: I kind of liked these functional bits split out into mixins. But I guess if they're not doing much, it saves some boilerplate & overhead to merge them in and just keep the more complex pieces split out 🤷‍♂️

app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -649,10 +511,220 @@ def test_serializer_rejects_too_many_primary_probe_sets(self):
             serializer.errors["primary_probe_set_ids"][0],
         )
 
+    def test_does_not_delete_probesets_when_other_fields_specified(self):
```
lmorchard, line 514: Always get a little nervous when a refactor also comes with big test diffs, but as far as I can tell they're just additions and moves without big changes. 🤷‍♂️

lmorchard (approved):
LGTM, works & tests pass. Noted some thinking-aloud shrugs that aren't even really nits

app/experimenter/experiments/api/v5/__init__.py
jaredlockhart: Well it was more about consistency since a lot of the other files are pluralized so everytime i was typing out test paths it was inconsistent

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Well I just renamed it because it used to be used only for updating and there was a separate one for creating, but now this one is used for creating and updating so I thought it didn't make sense to keep calling it update?

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Yeah no I feel you, in this case it was because when we had separate serializers, the changelog logic was shared between all of them but now it's only used in the single one and it's only two lines.  Technically none of them need to be mixed in but the heavier ones like the branches and probe sets it's nice to have the logic pulled out so it's easier to read, but this one, so smol.

app/experimenter/experiments/tests/api/v5/test_serializers.py
jaredlockhart: Yeah I was trying to pad out the tests to explicitly validate that we're only updating the fields we care about in this one serializer to try to catch cases like the one we saw last week where updating unrelated fields was clearing out feature_config, for example.  Because there was actually a lot of similar cases where updating one page would clear out values in another, so I wanted to make sure the tests were as exhaustive for those cases as I could.

PR #4407, opened 2021-01-12
fix #4406 fix(legacy): handle no sampling filter recipe pop percent u…

### LINKED ISSUE #4406: [Legacy] Experimenter does not update when sample filter is not present in recipe
When a sample filter is not present in the Normandy recipe, it doesn't update the population percent.
The correct behaviour would be that experimenter infers that the population precent is 100%

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-823)
┆Issue Number: EXP-823

tiftran:
…pdates

Because:

* This is an edge that isn't handled currently, experimenter doesn't update

This commit:

* Checks if there isn't a population filter and updates experimenter population to 100%

### DIFF
--- a/app/experimenter/normandy/tasks.py
+++ b/app/experimenter/normandy/tasks.py
@@ -201,6 +201,10 @@ def update_population_percent(experiment, recipe_data, filter_objects):
         if experiment.population_percent != percent_decimal:
             changed_data = {"population_percent": percent_decimal}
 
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Yup mmhm yep very nice 👍 

d17514e2c03ec6b0e67ba7f18439a011

PR #4408, opened 2021-01-12
fix #4399 chore(nimbus): remove admin validation

### LINKED ISSUE #4399: Remove Nimbus admin validation
We added a bunch of rich validation to the NimbusExperiment admin when we were temporarily using it to launch experiments.  Now that the full UI is ready for use, we don't need this logic anymore and we should rip it out.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-817)
┆Issue Number: EXP-817

jaredlockhart:
Because

- We are no longer using the admin to launch nimbus experiments

This commit

- Removes the additional rich validation we added to support that workflow

### DIFF
--- a/app/experimenter/experiments/admin/nimbus.py
+++ b/app/experimenter/experiments/admin/nimbus.py
@@ -1,6 +1,3 @@
-import json
-
-import jsonschema
 from django import forms
 from django.contrib import admin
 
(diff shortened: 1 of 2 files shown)


lmorchard:
Do we not want JSON schema validation in admin when possible? Seems like that would be a good handrail to not break the JSON if/when we *do* find a need to tinker with it from admin

jaredlockhart:
@lmorchard Yeah we could keep it, generally I treat the admin as just the 'you can dump anything in here you need to' panel for emergencies so typically you want it to be as basic as possible.  But on the other hand, I can't think of a case where we'd WANT to dump json that doesn't match the schema of the feature, so maybe we should just keep it and if it becomes a problem we can either update it or remove it in the future?  What do you think?

lmorchard (approved):
I can really go either way... Kind of like to have some data integrity checks all the way to admin panel to keep guns pointed away from feet. But could help in cases where otherwise we'd have to drop down to something like direct DB queries. :shrug:

jaredlockhart:
@lmorchard yeah exactly, if it comes up that we ended up needing it we can revert easily.

PR #4409, opened 2021-01-12
fix #4362 chore(nimbus): remove clientMutationId

### LINKED ISSUE #4362: Remove client_mutation_id
We added a `client_mutation_id` field in the backend of the graphql API because it's a common GQL pattern, but it looks like the current versions of the Apollo client framework don't use it at all, so there's no need for it, so I think we can safely remove it.

Related reference: https://github.com/wp-graphql/wp-graphql/issues/1123

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-798)
┆Issue Number: EXP-798

jaredlockhart:
Because

- Turns out apollo doesn't utilize clientMutationId

This commit

- Removes clientMutationId

### DIFF
--- a/app/experimenter/experiments/api/v5/inputs.py
+++ b/app/experimenter/experiments/api/v5/inputs.py
@@ -36,7 +36,6 @@ class TreatmentBranchType(BranchType):
 
 
 class ExperimentInput(graphene.InputObjectType):
-    client_mutation_id = graphene.String()
     id = graphene.Int()
     status = NimbusExperimentStatus()
     name = graphene.String()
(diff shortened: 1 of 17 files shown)


lmorchard (approved):
Hmm, I thought this was used in case there were multiple mutations in flight simultaneously. But maybe not? And we don't really do that anyway? 🤷‍♂️

jaredlockhart:
@lmorchard yeah it seems like it was used for relay 1 but isn't used anymore by apollo?

PR #4413, opened 2021-01-12
fix #4368 feat(nimbus): Add risk mitigation checklist link to DB and GQL

### LINKED ISSUE #4368: Add risk mitigation checklist link to DB model and experiment query/mutation
We need to store the risk mitigation checklist link, add the field to the `experimentBySlug` GQL query so it can be consumed, and add to the `updateExperiment` mutation so it can be updated.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-801)
┆Issue Number: EXP-801

lmorchard:
Because:

- We need to store the risk mitigation checklist link, add the field to
  the experimentBySlug GQL query so it can be consumed, and add to the
  updateExperiment mutation so it can be updated.

This commit:

- adds a new field `risk_mitigation_link` to the DB model and GQL API


lmorchard:
Opening as draft in anticipation of failed checks.

Also, adding this as its own field, since that seems straightforward for what will eventually be required in review. Could possibly be a special type of documentation link, but that seems like it would take some gymnastics on the front-end that we probably don't need.

lmorchard:
@jodyheavener hopefully this just kind of slots into your front-end draft PR

lmorchard:
Tweaked the commit message, because I got my parens inside-out

jodyheavener (approved):
LGTM, and yes I think this should slot right into my Draft PR 🙏

PR #4414, opened 2021-01-12
fix #3359: feat(sidebar) Add scrolling sidebar for visualizations

### LINKED ISSUE #3359: Add left side pane for analysis view
This pane is probably something we can/should consider across the whole app so we should think about how that will fit in.

emtwo:
This is still a draft - I have to add the other metric types to the sidebar, clean up the code a bit, and add tests.

I am putting it up early to get your high-level thoughts on the refactoring @LZoog


emtwo:
Screencast for context
https://user-images.githubusercontent.com/784781/104383522-f6435900-54fd-11eb-9837-9b5289c30827.mov

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  return { keys, labels };
+};
+
+type AppLayoutSidebarLockedProps = {
+  testid?: string;
+  children: React.ReactNode;
+  status?: StatusCheck;
+  analysis?: AnalysisData;
+  analysisLoadingInSidebar?: boolean;
```
LZoog, line 37: I believe we can remove the `?` on these since they're required. I think `analysisError` returns `undefined` if there's not an error so that one can probably stay as "Error or undefined".

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+                      "monitoring",
+                      "overview",
+                      "results_summary",
+                      "default",
+                    ].concat(keys)}
+                    className="text-dark"
+                    currentClassName="font-weight-bold text-primary"
+                  >
+                    <div>
+                      <a href="monitoring" style={{ color: "inherit" }}>
```
LZoog, line 104: Can we use a Bootstrap class or a custom class rather than an inline style? I don't think we're using inline styles anywhere else.

app/experimenter/nimbus-ui/src/components/MonitoringLink/index.tsx
```diff
@@ -11,7 +11,9 @@ export const MonitoringLink: React.FC<Pick<
   "monitoringDashboardUrl"
 >> = ({ monitoringDashboardUrl }) => (
   <>
-    <h3 className="h5 mb-3">Monitoring</h3>
+    <h3 className="h5 mb-3" id="monitoring">
+      Monitoring
+    </h3>
```
LZoog, line 16: While you're in this file, this could be named `LinkMonitoring` to match our other naming conventions 🤔

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+export const AppLayoutSidebarLocked = ({
+  children,
+  testid = "AppLayoutSidebarLocked",
+  status,
+  analysis,
+  analysisLoadingInSidebar = false,
+  analysisError,
+}: AppLayoutSidebarLockedProps) => {
+  const { slug } = useParams();
+  const { keys, labels } = getSidebarItems(analysis.other_metrics);
```
LZoog, line 50: I'm getting a `General application error`...

```
Uncaught TypeError: can't access property "other_metrics", analysis is undefined
```

Maybe this component is trying to render before the analysis call is finished? Didn't look too deeply since this is just a quick first pass 🤔

LZoog (commented):
Looking good so far! That package seems slick.

I do see a Typescript error:
```
Could not find a declaration file for module 'react-scrollspy'
```

emtwo:
@LZoog I was having trouble getting the scrollspy package to work in storybook. I think it's fine now so I'll be uploading some storybook updates too soon.

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <RouterSlugProvider>
+      <AppLayoutSidebarLocked
+        status={mockGetStatus(NimbusExperimentStatus.COMPLETE)}
+        analysisError={undefined}
+        analysis={{
+          show_analysis: true,
+          daily: [],
+          weekly: [],
+          overall: {},
+          other_metrics: mockAnalysis().other_metrics,
```
pdehaan, line 53: Te other props above are camelCase, is there a reason `show_analysis` and `other_metrics` are underscored here?

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+  return newMap;
+};
+
+type AppLayoutSidebarLockedProps = {
+  testid?: string;
+  children: React.ReactNode;
+  status: StatusCheck;
+  analysis?: AnalysisData;
+  analysisLoadingInSidebar?: boolean;
+  analysisError: Error | undefined;
```
LZoog, line 65: This can be `analysisError?: Error` to match the others since it's the same as `Error | undefined`, I'd asked before if some of these props had to be optional because I thought maybe we'd always be passing a defined value for some of them (like `analysis: AnalysisData` instead of `analysis?`) but I don't think that's true.

app/experimenter/nimbus-ui/src/components/LinkNav/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  children,
+  disabled = false,
+  storiesOf,
+  testid = "nav-home",
+  className,
+  textColor,
+}: LinkNavProps) => {
+  const to = route ? `${BASE_PATH}/${route}` : BASE_PATH;
+  // an alternative to reach-router's `isCurrent` with identical
+  // functionality; explicitely setting it here allows us to test.
```
pdehaan, line 28: typo: "explicitly" (no second "e")

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+                    <BarChart className="sidebar-icon" />
+                    Results
+                  </LinkNav>
+                  <Scrollspy
+                    items={sidebarKeys}
+                    className="text-dark"
+                    currentClassName="font-weight-bold text-primary"
+                  >
+                    <div>
+                      <a href="#monitoring" className="inherit-color">
```
LZoog, line 146: I think some of these could use more consistent spacing, but fine to do in a follow up.

image

LZoog (commented):
My comments are mostly all nits - will r+ when the text overflow issue is fixed. 🙂

I'll try again tomorrow morning but I'm not sure why, I keep getting a `500` `'NoneType' object has no attribute 'slug'` error following our documentation, but it looks good in Storybook!

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.stories.tsx
emtwo: They are coming from the json response from the server. Since it's in python, it's all snake_case ¯\_(ツ)_/¯

app/experimenter/nimbus-ui/package.json
```diff
@@ -47,9 +47,11 @@
     "react-helmet": "^6.1.0",
     "react-hook-form": "^6.12.2",
     "react-scripts": "3.4.0",
+    "react-scrollspy": "^3.4.3",
     "react-select": "^3.1.1",
     "react-select-event": "^5.1.0",
     "react-tooltip": "^4.2.10",
+    "storybook": "^6.1.14",
```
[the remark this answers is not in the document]
emtwo: I tend to install storybook locally to test. This might be a different testing workflow than others use. @LZoog how do you typically run storybook locally without having this dependency?

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
emtwo: I kind of added the extra spacing on purpose between different sections so they're easier to distinguish. Does it seem ugly? I can just remove it :)

app/experimenter/nimbus-ui/package.json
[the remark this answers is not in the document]
LZoog: Good catch @pdehaan! @emtwo with the `yarn storybook` command in `nimbus-ui`... the package is `@storybook/react` and not `storybook`, you should be able to remove this dependency.

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
LZoog: Meh, we can tweak it later if anyone thinks the same. :P

LZoog (approved):
😎👍

PR #4416, opened 2021-01-13
fix #4368 fix(nimbus): allow risk_management_link to be blank

### LINKED ISSUE #4368: Add risk mitigation checklist link to DB model and experiment query/mutation

lmorchard:


jodyheavener (approved):
🤠

Issue #3898, opened 2020-11-05
Review English-friendly targetingConfigSlug config values from server


data-sync-user:
The verbiage we're using in the `rapid` project for at least these values differ from what we're using in Nimbus. It seems like these should be revisited.

(rapid vs nimbus):
"US users (en)" vs "Us Only"
"All English users" vs "All English"
"First start-up users (en)" vs "First Run"

See this Storybook link for "Us Only" in action: https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/94a2112d1956d5cf8c09f0c0bfe5fb8352e58e7d/nimbus-ui/index.html?path=/story/components-tablesummary--filled-out

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-550)
┆Issue Number: EXP-550

data-sync-user:
➤ Ana Medinac commented:

This has been completed in another ticket.

PR #4419, opened 2021-01-14
fix #4114 fix #4115 feat(nimbus): Add "Full targeting expression" to Audience table

### LINKED ISSUE #4114: Add "Full targeting expression" to Audience table
We need to show the full targeting expression on the last row of the Audience table. If it's not filled out, don't show the row.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-663)
┆Issue Number: EXP-663

### LINKED ISSUE #4115: Add "Full targeting expression" to experiment query
We need to be able to display the targeting expression in the Audience table in the FE. Add this field to the GQL experiment query to allow the client to consume it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-664)
┆Issue Number: EXP-664

lmorchard:
Because:

- We need to show the full targeting expression on the last row of the Audience table.

This commit:

- adds targetingConfigTargeting field to experimentBySlug query as a
  lookup of targetingConfigSlug against constants

- adds a row to TableAudience to display targetingConfigTargeting when
  targetingConfigSlug is set

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableAudience/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableAudience/index.tsx
@@ -65,6 +65,17 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
             </td>
           </tr>
         )}
+        {experiment.targetingConfigTargeting !== "" && (
+          <tr>
+            <th>Full targeting expression</th>
+            <td
+              data-testid="experiment-target-expression"
+              className="text-monospace"
+            >
+              {experiment.targetingConfigTargeting}
+            </td>
+          </tr>
+        )}
       </tbody>
     </Table>
   );
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -120,6 +120,7 @@ class NimbusExperimentType(DjangoObjectType):
     documentation_links = DjangoListField(NimbusDocumentationLinkType)
     treatment_branches = graphene.List(NimbusBranchType)
     targeting_config_slug = NimbusExperimentTargetingConfigSlug()
+    targeting_config_targeting = graphene.String()
     primary_probe_sets = graphene.List(NimbusProbeSetType)
     secondary_probe_sets = graphene.List(NimbusProbeSetType)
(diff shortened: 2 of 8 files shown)


lmorchard:
I swear, the "check" job is my nemesis

app/experimenter/nimbus-ui/src/components/TableAudience/index.tsx
```diff
@@ -55,15 +55,26 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
...
+            <tr>
+              <th>Full targeting expression</th>
+              <td
+                data-testid="experiment-target-expression"
+                className="text-monospace"
+              >
+                {experiment.targetingConfigTargeting}
+              </td>
+            </tr>
+          </>
```
jodyheavener, line 77: Is a targeting config slug of "No Targeting" the same as not having a targeting config? If so, maybe we should hide the row since it yields no expression?

Screen Shot 2021-01-14 at 10 23 01 AM

jodyheavener (approved):
Frontend LGTM with one suggestion, backend makes sense from what I can tell!

jodyheavener:
Was just looking at the Figma doc - are we anticipating custom HTML or some other rich text format for Custom audience?

Screen Shot 2021-01-14 at 12 05 46 PM

app/experimenter/nimbus-ui/src/components/TableAudience/index.tsx
lmorchard: Yeah, no targeting == no expression. Probably good idea to hide the row :+1: - will toss that tweak in there before merging

lmorchard:
>  are we anticipating custom HTML or some other rich text format for Custom audience?

Hmm, not sure? Might be worth a follow up issue - though could possibly fall under #3898 / [EXP-550](https://jira.mozilla.com/browse/EXP-550). As far as I can tell in current implementation, [custom audiences are just constant structs with simple text descriptions](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L7-L80).

PR #4438, opened 2021-01-15
chore(nimbus): move child components into parent component directories

lmorchard:
Because:

- the flat component hierarchy has gotten a bit too long

This commit:

- moves many components into the directories of parent components that
  are their only use cases


lmorchard:
Just opening this as a kind of draft proposal because it was quicker to do it than write it up as an ADR :)

In a nutshell: while looking through components, I realized that many in the flat hierarchy had only one other component using them. Ran down the list and consolidated a bunch into child/parent relationship and came up with this.

app/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.test.tsx
```diff
@@ -4,10 +4,10 @@
 
 import React from "react";
 import { render, screen } from "@testing-library/react";
-import { MockedCache, mockExperimentQuery } from "../../lib/mocks";
-import TableSummary from ".";
-import { getExperiment_experimentBySlug } from "../../types/getExperiment";
-import { NimbusExperimentChannel } from "../../types/globalTypes";
+import { MockedCache, mockExperimentQuery } from "../../../lib/mocks";
+import TableAudience from ".";
```
lmorchard, line 8: Oh yeah, and while nearly all the changes here were made automatically by VSCode, this is one by-hand change I made: Realized the test passes because it's internally consistent, but `TableSummary` should technically be `TableAudience` here

lmorchard:
Actually, I'm just going to open this up to review. We can decide this isn't a thing we want to do or r+ it and do it

jodyheavener:
- Should this pattern also apply to components that are already nested? For example, `RichText` is only used in `TableSummary`, which is already nested in `Summary`. So would `RichText` be nested under those two components?
- Not a question but I can see a developer and reviewer both forgetting to check if this pattern is maintained, either when a component is added or deleted, or when an existing component wants to use an already-nested component and doesn't move it to the top level. I wonder if there's a way we can enforce this 🤔

lmorchard:
> Should this pattern also apply to components that are already nested? For example, RichText is only used in TableSummary, which is already nested in Summary. So would RichText be nested under those two components?

Hmm, I might have just missed that one. But also RichText sounds like a component we'll probably want to use elsewhere eventually?

> Not a question but I can see a developer and reviewer both forgetting to check if this pattern is maintained, either when a component is added or deleted, or when an existing component wants to use an already-nested component and doesn't move it to the top level. I wonder if there's a way we can enforce this 🤔

I wouldn't bother putting much work into enforcing this. More a thing to keep in mind while reviewing I guess and also possibly a fuzzy line. Doesn't hurt anything if we miss it, and we can always tighten it up later during maintenance sprints.

lmorchard:
Rebased against main

jodyheavener:
> I'd also say we don't need to go absolutely nuts with nesting. More just a rough means of fencing off complexity that doesn't need thinking about when working on different areas of the project.

This seems reasonable, and the new structure LGTM. Maybe @LZoog will have more thoughts but I'm happy to merge this. Seems like it'll evolve over time

jodyheavener:
Oh - should the Storybook structure also reflect the new pattern?

lmorchard:
> Oh - should the Storybook structure also reflect the new pattern?

Sure... the Storybook structure is a little weird because it diverges from the literal folder structure, but I *think* I managed to make equivalent changes in a new commit

LZoog:
Hmm so... I'm not sure if I like _all_ of the changes here, or maybe it's just not clear to me when we should nest something or not nest.

From our retro conversation, I think we only talked about a `pages` dir, `visualization` dir, and maybe a `layouts` dir. I don't think we talked about nesting beyond that.

Moving `FormAudience` into `EditAudience` (as well as the other forms) - OK, I could see this, I can't picture a scenario where we'd ever need that form outside of that page, and this looks good to me in Storybook too (might take some confusion of "where can I see this form state? Oh, the form with states is in components" out). But `DirectoryTable` into `PageHome`... I'm not sure we'll ever use that outside of the Home page, but it feels like it's sorta blurring the line on when to nest vs not nest. I liked that prior to this, there was no question on if something should be nested or not, the answer was "no" besides I think one weird exception of `FormBranch` inside `FormBranches`.

Moving the visualization stuff into Results - I'm OK with this too, though I also liked seeing the separate "visualization" section in Storybook. 

I guess my overall opinion is that I feel a little torn on this because I do like benefits of a flat `components` structure but agree it's gotten a little unwieldy. I think I like the idea of only having those top 3 directories to nest in (which I know this PR doesn't do), and could go either way on the forms being inside their requisite pages, but if the team likes this partial nesting, I'm good with it.

@jodyheavener 
> Should this pattern also apply to components that are already nested? For example, `RichText` is only used in `TableSummary`, which is already nested in `Summary`. So would `RichText` be nested under those two components?

Yeah, this is exactly what I would like to avoid. I don't want really deeply nested components. 😅

LZoog:
BTW, would probably be good to get @jaredlockhart's take here before merging, I know he was a fan of the flat structure but has been out sick.

lmorchard:
> Hmm so... I'm not sure if I like all of the changes here, or maybe it's just not clear to me when we should nest something or not nest.

FWIW, the rationale I used here was - if a component is only used by one other component, nest it as a child. If it's shared, leave it out at the top.

> But DirectoryTable into PageHome... I'm not sure we'll ever use that outside of the Home page, but it feels like it's sorta blurring the line on when to nest vs not nest. I liked that prior to this, there was no question on if something should be nested or not, the answer was "no" besides I think one weird exception of FormBranch inside FormBranches.

As far as I can tell, there is no other component besides PageHome that uses DirectoryTable. It does seem like something *could* share DirectoryTable, but nothing does yet.

Personally, I prefer the child component nesting because it bundles up complexity into a neighborhood. Like, PageResults has 6-7 associated components that only it uses. They're kind of noise at the top level otherwise

lmorchard:
> From our retro conversation, I think we only talked about a pages dir, visualization dir, and maybe a layouts dir. I don't think we talked about nesting beyond that.

And yeah, we didn't talk about this specifically, which is why I'm kind of floating it as a proposal

lmorchard:
> I liked that prior to this, there was no question on if something should be nested or not, the answer was "no" besides I think one weird exception of FormBranch inside FormBranches.

There's also that bit where this reorg *could* go further - which is to move a lot of these directory-based components into just files within the parent component. Like, PageResults could just have TableHighlights.tsx and the tests & stories merged in. But that's probably too much bikeshedding for no good reason and piles up complexity in other spots.

LZoog:
> As far as I can tell, there is no other component besides PageHome that uses DirectoryTable. It does seem like something could share DirectoryTable, but nothing does yet. Is that where it seems fuzzy?

I think so, maybe it's more so around the "...does _seem_ like..." meaning we sorta judge it on a case-by-case basis / that makes it a little subjective, but that's OK.

> Personally, I prefer the child component nesting because it bundles up complexity into a neighborhood. Like, PageResults has 6-7 associated components that only it uses. They're kind of noise at the top level otherwise

Yeah, I get this rationale. I think the "visualization" section/its own dir made sense to me for this reason, but lumping it in `PageResults` may make more sense. I asked @emtwo to check this PR out as well to make sure she agrees with the consensus since it affects components she's worked heavily on.

> We could limit the nesting to just one level maybe?

I can get behind this. 😎

Yeah so, with you preferring this and Jody not opposing it, I'm cool with it, especially if we have that "only nest one level" rule. 😄 Though again I do want Jared to take a looksee to have a chance to voice his opinion.

emtwo:
> Yeah, I get this rationale. I think the "visualization" section/its own dir made sense to me for this reason, but lumping it in `PageResults` may make more sense. I asked @emtwo to check this PR out as well to make sure she agrees with the consensus since it affects components she's worked heavily on.

I think having it under `PageResults` works for me. I also like the idea of having only 1 level of nesting for now.

lmorchard:
Rebased against main and moved ConfidenceInterval out of PageResults/TableVisualizationRow down to just PageResults

lmorchard:
Rebased against main to stay ahead of the bitrot

lmorchard:
Hmm, another apparent transient test failure, seems unrelated to this PR:

* https://app.circleci.com/pipelines/github/mozilla/experimenter/8125/workflows/79220b06-0a11-4db6-9cfd-c052175cbcdf/jobs/21371
* https://app.circleci.com/pipelines/github/mozilla/experimenter/8125/workflows/25809b30-3675-4649-abe1-872e1342142b/jobs/21375

jaredlockhart:
> > Should this pattern also apply to components that are already nested? For example, RichText is only used in TableSummary, which is already nested in Summary. So would RichText be nested under those two components?
> 
> Hmm, I might have just missed that one. But also RichText sounds like a component we'll probably want to use elsewhere eventually?
> 

Yeah so this is sort of my question.  Two questions.  One is, what happens when a component starts by being used in only one place, and then later gets used in two? Do we bump it up a dir and move it around?

Question the second.  I noticed there were some reducers embedded inside components, I've usually seen reducers all placed in their own top level next to components, and I think we have some else where, do we want reducers/other state management kind of scattered?

And my big question is, if you know the name of a component but you don't know where it is, is the only way to find it by 
1) looking for where it's used
2) looking at the import path and then following that?

Right now all you need to know is the name of the component and you can just find it.

What's the downside of leaving it flat?

jaredlockhart:
> Hmm, another apparent transient test failure, seems unrelated to this PR:
> 
>     * https://app.circleci.com/pipelines/github/mozilla/experimenter/8125/workflows/79220b06-0a11-4db6-9cfd-c052175cbcdf/jobs/21371
> 
>     * https://app.circleci.com/pipelines/github/mozilla/experimenter/8125/workflows/25809b30-3675-4649-abe1-872e1342142b/jobs/21375

@lmorchard Whenever I see an intermittent test failure I snapshot the failure and file a ticket for it immediately.  Worst case it gets fixed somewhere else and you can just close it, but if not it's good to leave a ticket reference so when you have downtime you can go clean up some intermittent failures.

lmorchard:
> One is, what happens when a component starts by being used in only one place, and then later gets used in two? Do we bump it up a dir and move it around?

Yup, just move it up a level when it starts getting shared. Otherwise it's just unrelated clutter, IMO

> Question the second. I noticed there were some reducers embedded inside components, I've usually seen reducers all placed in their own top level next to components, and I think we have some else where, do we want reducers/other state management kind of scattered?

We've got one reducer right now in nimbus-ui, AFAIK. It's only used locally within FormBranches. If we were using a global state reducer like Redux, I'd say that goes in a more top-level location. But, the reducer used by FormBranches isn't a thing you necessarily want to have in mind when you're working over in FormRequestReview

> And my big question is, if you know the name of a component but you don't know where it is, is the only way to find it by

Well, the components I've moved are only used in the context of the parent component into which I moved them. So, you won't need to find them unless you're already in that directory. Then, they're right there.

> What's the downside of leaving it flat?

The list is getting very long and noisy. Small utility components mixed with big page components. Personally, I'm starting to find it hard to keep it straight

lmorchard:
> Whenever I see an intermittent test failure I snapshot the failure and file a ticket for it immediately.

Yeah, I filed one over here earlier, not sure if this was related but mostly jotting a note about one of the failures of this PR that cleared

https://github.com/mozilla/experimenter/issues/4453

jaredlockhart (approved):
> > What's the downside of leaving it flat?
> 
> The list is getting very long and noisy. Small utility components mixed with big page components. Personally, I'm starting to find it hard to keep it straight. This came up as a topic last Friday during retro - we were discussing other ways to organize things by topic (e.g. pages, layouts, etc). I figured I'd see how hard parent/child was as a reorg

Yep makes sense to me, I'm all for anti clutter 👍

lmorchard:
I think we've got rough consensus here and at least one r+ so I'm gonna hit the button

PR #4457, opened 2021-01-21
feat(nimbus): automatically sort nimbus ts/tsx imports

jodyheavener:
_I did this on a whim, so feel free to push back or suggest alternatives. 🙏_

This PR adds the [`prettier-plugin-organize-imports`](https://github.com/simonhaenisch/prettier-plugin-organize-imports) dependency to automatically sort the imports in our TS/X files. If you use VS Code or another IDE that supports Prettifying on save any new imports will automatically be sorted on save/paste/however you have it set up. And since it's a Prettier plugin, and our Prettier setup hooks into ESLint this allows sorting to happen automatically with our `yarn lint-fix` command.

From what I can tell this sorts imports first by external dependencies alphabetically, and then by relative path level descending then alphabetically. It also sorts multiple imports from the same statement alphabetically.

Alternatives:

- The [sort-imports](https://eslint.org/docs/rules/sort-imports) ESLint rule. I tried this but it appears it doesn't differentiate between dependencies and local modules, and I was encountering weird conflicts between it's automatic sort and Prettier auto-formatting on save. If you know of a way to fix these two things I would prefer to use this rule than another dependency.
- The TypeScript `organizeImports` feature. This is actually being used under the hood by this plugin. We could definitely set it up [at the VS Code level](https://stackoverflow.com/a/54690944/717633), but this would be specific to VS Code. We want this to be available to everyone, especially if CI is going to fail when the order is incorrect.

### DIFF
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -89,6 +89,7 @@
     "mutationobserver-shim": "^0.3.7",
     "node-sass": "^4.14.1",
     "prettier": "^2.1.2",
+    "prettier-plugin-organize-imports": "^1.1.1",
     "stylelint": "^13.7.1",
     "stylelint-config-prettier": "^8.0.2"
   },
--- a/app/experimenter/nimbus-ui/src/components/App/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/App/index.test.tsx
@@ -2,12 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, { ReactNode } from "react";
-import { screen, waitFor } from "@testing-library/react";
-import { renderWithRouter } from "../../lib/test-utils";
-import { MockedCache, mockExperimentQuery } from "../../lib/mocks";
 import * as apollo from "@apollo/client";
+import { screen, waitFor } from "@testing-library/react";
+import React, { ReactNode } from "react";
 import App from ".";
+import { MockedCache, mockExperimentQuery } from "../../lib/mocks";
+import { renderWithRouter } from "../../lib/test-utils";
 
 const { mock } = mockExperimentQuery("my-special-slug");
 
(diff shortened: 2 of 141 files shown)


jodyheavener:
@jaredlockhart Sweet! I took a quick look but didn't see the issue, if you come across it feel free to close once this is merged. I'm going to hold off merging until #4438 is in.

LZoog:
Oh nice! 👍 We have a few open/ready PRs, maybe we should wait to merge until those are in, then this can be rebased and ran in those files? This might cause conflicts in a lot of other PRs otherwise.

lmorchard:
Yeah, sorry I just blew this up by merging #4438. But hopefully you can just re-run the plugin on a fresh update from main? If it's more complicated than that, I might furrow my brow vaguely at this PR. But otherwise, I'm in the 👍 camp too

lmorchard:
Personally, I kind of like conceptually sorting imports in order of, like... framework, then global project utils, then sibling component dependencies, then local dependencies. But that's a weird fuzzy thing and alpha-order is mechanically enforceable and makes probably just as much sense.

lmorchard:
One other thing to be aware of, but probably only peripherally because I don't think it affects us here: Need to take care in sorting components that import CSS resources, because that *can* affect the order of CSS rules in the eventual output. Ran into that as a gotcha on another project.

But, I think we're immune to it here because we're using Bootstrap and utility styles rather than per-component (S)CSS resources

jodyheavener:
> But, I can't recall a tool in JavaScript that automates that sorting preference, so it's always a papercut-level chore. Maybe flake8-import-order would do it for Python - but I don't think such a thing exists in JS-world

@lmorchard Yeah other than the two two alternatives I listed above I don't know of a better way to sort them. The [sort-imports](https://eslint.org/docs/rules/sort-imports) ESLint rule does allow you to specify some preferences for ordering, but AFAICT it doesn't differentiate between system/third-party/local modules (it just sorts the whole lot). Definitely happy to re-visit down the road.

> But, I think we're immune to it here because we're using Bootstrap and utility styles rather than per-component (S)CSS resources

Additionally, since we store component-specific styles at the same level as the TSX file it would come last in the sorting order I think, which kind of aligns with how we'd want the cascade to be ordered.

PR #4458, opened 2021-01-21
fix #4456: feature(visualization) Process weekly data server-side.

### LINKED ISSUE #4456: Send formatted weekly experiment data from the visualization API
similar to the `overall` data format that can be directly used when generating tables and graphs, the `weekly` data also needs to be formatted.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-856)
┆Issue Number: EXP-856

emtwo:
Because weekly data needs to be processed to be graphed more easily

This commit processes it similarly to overall data

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
@@ -50,13 +50,15 @@ const TableHighlights = ({
         {Object.keys(results).map((branch) => {
           const userCountMetric =
             results[branch]["branch_data"][METRIC.USER_COUNT];
+          const participantCount =
+            userCountMetric[BRANCH_COMPARISON.ABSOLUTE]["first"]["point"];
           return (
             <tr key={branch} className="border-top">
               <th className="align-middle p-1 p-lg-3" scope="row">
                 <p>{branch}</p>
                 <p className="h6">
-                  {userCountMetric[BRANCH_COMPARISON.ABSOLUTE]["point"]}{" "}
-                  participants ({userCountMetric["percent"]}%)
+                  {participantCount} participants ({userCountMetric["percent"]}
+                  %)
                 </p>
               </th>
               <td className="p-1 p-lg-3 align-middle">
--- a/app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
@@ -178,12 +178,17 @@ const TableVisualizationRow: React.FC<{
     className = tooltipText = "";
     const percent = branch_data[METRIC.USER_COUNT]["percent"];
     const userCountMetric =
-      branch_data[METRIC.USER_COUNT][BRANCH_COMPARISON.ABSOLUTE]["point"];
+      branch_data[METRIC.USER_COUNT][BRANCH_COMPARISON.ABSOLUTE]["first"][
+        "point"
+      ];
 
     const branchType = is_control ? VARIANT_TYPE.CONTROL : VARIANT_TYPE.VARIANT;
     branchComparison =
       branchComparison || dataTypeMapping[tableLabel][branchType];
-    const { lower, upper, point, count } = metricData[branchComparison];
+
+    const { lower, upper, point, count } = metricData[branchComparison][
+      "first"
+    ];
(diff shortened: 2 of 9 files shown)


app/experimenter/visualization/api/v3/views.py
```diff
@@ -139,7 +141,7 @@ def process_data_for_consumption(overall_data, weekly_data, experiment):
     return results, other_metrics
 
 
-def generate_results_object(data, experiment):
+def generate_results_object(data, experiment, window="overall"):
```
jaredlockhart, line 144: looks like `"overall"` could be a constant somewhere? but i see a bunch of constants littered around so maybe we can clean all that up in a later pass 🤷‍♀️

jaredlockhart (approved):
A lot going on in here, but better to get something working sooner than later.  We should think about breaking all this apart into something more declarative and structural with pydantic or similar, sooner than later. r+

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -50,13 +50,17 @@ const TableHighlights = ({
...
+              ? userCountMetric[BRANCH_COMPARISON.ABSOLUTE][0]["point"]
+              : null;
           return (
             <tr key={branch} className="border-top">
               <th className="align-middle p-1 p-lg-3" scope="row">
                 <p>{branch}</p>
                 <p className="h6">
-                  {userCountMetric[BRANCH_COMPARISON.ABSOLUTE]["point"]}{" "}
-                  participants ({userCountMetric["percent"]}%)
+                  {participantCount} participants ({userCountMetric["percent"]}
```
pdehaan, line 62: Q: Above on L56 we set `participantCount` to `null` if the `userCountMetric[BRANCH_COMPARISON.ABSOLUTE]` array is empty.

How/what would this render if the `participantCount` is null?

app/experimenter/nimbus-ui/src/components/TableVisualizationRow/index.tsx
```diff
@@ -178,12 +178,19 @@ const TableVisualizationRow: React.FC<{
     className = tooltipText = "";
     const percent = branch_data[METRIC.USER_COUNT]["percent"];
     const userCountMetric =
-      branch_data[METRIC.USER_COUNT][BRANCH_COMPARISON.ABSOLUTE]["point"];
+      branch_data[METRIC.USER_COUNT][BRANCH_COMPARISON.ABSOLUTE].length > 0
+        ? branch_data[METRIC.USER_COUNT][BRANCH_COMPARISON.ABSOLUTE][0]["point"]
```
pdehaan, line 182: uber nit: Some of this is getting tricky to read. Not sure on the collective feelings on using a temp variable to reduce code duplication.

```js
const bData = branch_data[METRIC.USER_COUNT][BRANCH_COMPARISON.ABSOLUTE];
const userCountMetric = bData.length ? bData[0].point : null;
```

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
```diff
@@ -50,13 +50,17 @@ const TableHighlights = ({
         {Object.keys(results).map((branch) => {
           const userCountMetric =
             results[branch]["branch_data"][METRIC.USER_COUNT];
+          const participantCount =
+            userCountMetric[BRANCH_COMPARISON.ABSOLUTE].length > 0
+              ? userCountMetric[BRANCH_COMPARISON.ABSOLUTE][0]["point"]
+              : null;
```
pdehaan, line 56: Same comment as below, but not sure if using a temp variable for the `userCountMetric[BRANCH_COMPARISON.ABSOLUTE]` lookup might DRY the code out a bit and shorten the code significantly.

```js
const metric = userCountMetric[BRANCH_COMPARISON.ABSOLUTE];
const participantCount = metric.length ? metric[0].point : null;
```

Not sure if there is a cleverer way that avoids the inline if, maybe `const participantCount = metric[0]?.point;` but maybe that's too much magic happening and I'm just wildly guessing without actually testing that code.

app/experimenter/nimbus-ui/src/components/TableHighlights/index.tsx
emtwo: I've updated the server-side generated data so this is not the case anymore

PR #4476, opened 2021-01-25
fix #4473: fix(visualization) correctly compute truncation for countField

### LINKED ISSUE #4473: TableMetricSecondary has incorrect truncation for `countField`
The calculation for truncation in `countField` in `TableVisualizationRow` has an extra 0 and produces incorrect results.

emtwo:
Because:
* There is a bug in the results page for countField computations

This commit:
* Fixes the bug by removing an extra 0.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.test.tsx
@@ -10,7 +10,7 @@ import { RouterSlugProvider } from "../../../lib/test-utils";
 import { mockAnalysis } from "../../../lib/visualization/mocks";
 
 describe("TableMetricSecondary", () => {
-  it("has the correct headings", async () => {
+  it("has the correct headings", () => {
     const EXPECTED_HEADINGS = ["Count", "Relative Improvement"];
     const { mock, experiment } = mockExperimentQuery("demo-slug");
 
@@ -30,7 +30,7 @@ describe("TableMetricSecondary", () => {
     });
   });
 
-  it("has correctly labelled result significance", async () => {
+  it("has correctly labelled result significance", () => {
     const { mock, experiment } = mockExperimentQuery("demo-slug");
     render(
       <RouterSlugProvider mocks={[mock]}>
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.test.tsx
```diff
@@ -88,4 +88,20 @@ describe("TableMetricSecondary", () => {
     expect(positiveBlock).not.toBeInTheDocument();
     expect(neutralBlock).not.toBeInTheDocument();
   });
+
+  it("shows expected count values", async () => {
```
LZoog, line 92: Does this need to be `async`?

jaredlockhart (approved):
👍 * 10^10

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
```diff
@@ -130,8 +130,8 @@ const countField = (
   metricName: string,
   tableLabel: string,
 ) => {
-  const interval = `${Math.round(lower * 1000) / 100} to ${
-    Math.round(upper * 1000) / 100
+  const interval = `${Math.round(lower * 100) / 100} to ${
+    Math.round(upper * 100) / 100
```
LZoog, line 134: Do we need `*100` and `/100`? Is this line equal to this?
```
const interval = `${Math.round(lower)} to ${Math.round(upper)}`
```

emtwo: Yes, the reason this is done is to keep 2 decimal places.

Eg. 
```
let a = 1324.324334343;
a = a * 100; // a = 132432.4334343
a = Math.round(a); // a = 132432
a = a / 100; // a = 1324.32
```

app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.test.tsx
emtwo: I believe it can be removed. I'll update

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
LZoog: Ahh makes sense, LGTM then. :)

LZoog (approved):

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
jodyheavener: What about `a.toFixed(2)`?

PR #4479, opened 2021-01-26
    chore(nimbus): temporarily disable intermittent visualization test

jaredlockhart:
Because
    
    * There is a frequently intermittently failing test in the visualization test suite
    * Requires some more rigorous digging in to fix
    
    This commit
    
    * Temporarily disables the test so PRs can at least land while it's being investigated


emtwo (approved):
thank you

lmorchard:
FWIW, this PR didn't get linked or mentioned on the original bug - can we make sure the commit message mentions the issue? Just spent awhile also troubleshooting this one and didn't know we already had a commit to disable it

Issue #4299, opened 2020-12-17
Don't require form fields on submission after create (client-side)


data-sync-user:
Each of the four forms after create:

Overview (no changes should be needed)
Branches
Metrics (no changes should be needed)
Audience

should be saveable with any subset of fields filled out.  Required validation can happen at review time.  The save button on all four fields should always be enabled except when `isLoading` is true.

This is blocked on https://jira.mozilla.com/browse/EXP-770

EXP-781 should also likely be done first. Now that our forms besides Branches use `useCommonForm`, the above should be fairly simple to do (pending looking into useCommonForm on Branches) because of this: https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/hooks/useCommonForm.tsx#L46-L47

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-771)
┆Issue Number: EXP-771

data-sync-user:
➤ Lauren Zugai commented:

Dropping this from a 5 to a 3 and may drop to a 2, this has become much easier with some form refactoring we've done, but can't complete until EXP-770 is in.

PR #4480, opened 2021-01-26
fixes #4371 feat(nimbus): add documentation add/remove functionality 

### LINKED ISSUE #4371: Additional links add/remove functionality
1) Create the “+ Add Link” button and onclick, add an additional document title/link input set.
2) Excluding the first input set, add an “X” (as an SVG to match `FormBranches`) button next to the input sets with a title of “Remove link” that onclick, removes the input set.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-805)
┆Issue Number: EXP-805

jodyheavener:
Closes #4371

This PR adds the functionality for adding and removing experiment documentation links.

- Add a new documentation link set using the "+ Add Link" button
- Remove a link set by clicking its adjacent "X" button
- Removing the last set should display a reset/empty set
- Client and server validation errors should show up correctly
- Click "Save" with a partially complete set should strip it and not save it

Additionally, I noticed `SummaryTable` was incorrectly rendering saved documentation links, so I snuck that in here.

PS - You're going to want to enable GitHub's "Hide whitespace changes" when looking at this diff as I had to shift indentation in `FormOverview`.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/FormOverview/FormDocumentationLink.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormOverview/FormDocumentationLink.tsx
@@ -1,0 +1,99 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import Button from "react-bootstrap/Button";
+import Col from "react-bootstrap/Col";
+import Form from "react-bootstrap/Form";
+import { ArrayField, FieldError } from "react-hook-form";
+import { useCommonNestedForm, useConfig } from "../../hooks";
+import { ReactComponent as DeleteIcon } from "../../images/x.svg";
+import { AnnotatedDocumentationLink } from "./documentationLink";
+
+export const documentationLinkFieldNames = ["title", "link"] as const;
+
+type DocumentationLinkFieldName = typeof documentationLinkFieldNames[number];
+
+export type FormDocumentationLinkProps = {
+  fieldNamePrefix: string;
+  documentationLink: Partial<ArrayField<AnnotatedDocumentationLink, "id">>;
+  errors: Record<string, FieldError>;
+  onRemove: () => void;
+  canRemove: boolean;
+  submitErrors: Record<string, string[]>;
+  setSubmitErrors: React.Dispatch<React.SetStateAction<Record<string, any>>>;
+  touched: Record<string, boolean>;
+};
+
+export const FormDocumentationLink = ({
+  fieldNamePrefix,
+  documentationLink,
+  errors,
+  onRemove,
+  canRemove,
+  submitErrors,
+  setSubmitErrors,
+  touched,
+}: FormDocumentationLinkProps) => {
+  const { documentationLink: documentationLinkOptions } = useConfig();
+
+  const { FormErrors, formControlAttrs } = useCommonNestedForm<
+    DocumentationLinkFieldName
+  >(
+    {
+      title: documentationLink.title,
+      link: documentationLink.link,
+    },
+    setSubmitErrors,
+    fieldNamePrefix,
+    submitErrors,
+    errors,
+    touched,
+  );
+
+  return (
+    <Form.Group className="mb-0" data-testid="DocumentationLink">
+      <Form.Row>
+        <Form.Group as={Col} sm={4} md={3}>
+          <Form.Control as="select" {...formControlAttrs("title")}>
+            <option value="">Select document type...</option>
+            {documentationLinkOptions &&
+              documentationLinkOptions.map((linkType, optIdx) => (
+                <option
+                  key={`${documentationLink.id}-opt-${optIdx}`}
+                  value={linkType!.value as string}
+                >
+                  {linkType!.label}
+                </option>
+              ))}
+          </Form.Control>
+          <FormErrors name="title" />
+        </Form.Group>
+        <Form.Group as={Col}>
+          <Form.Control
+            placeholder="Link"
+            type="url"
+            {...formControlAttrs("link")}
+          />
+          <FormErrors name="link" />
+        </Form.Group>
+        {canRemove && (
+          <Form.Group as={Col} className="flex-grow-0">
+            <Button
+              data-testid={`${fieldNamePrefix}.remove`}
+              variant="light"
+              className="bg-transparent border-0 p-0 m-0 mt-px"
+              title="Remove link"
+              onClick={onRemove}
+            >
+              <DeleteIcon width="18" height="18" />
+            </Button>
+          </Form.Group>
+        )}
+      </Form.Row>
+    </Form.Group>
+  );
+};
+
+export default FormDocumentationLink;
--- a/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
@@ -2,7 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { act, fireEvent, render, screen } from "@testing-library/react";
+import {
+  act,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 import React from "react";
 import { DOCUMENTATION_LINKS_TOOLTIP } from ".";
 import { mockExperimentQuery } from "../../lib/mocks";
@@ -36,7 +42,7 @@ describe("FormOverview", () => {
     expect(onNext).toHaveBeenCalled();
   });
 
-  it("renders initial documentation links", () => {
+  it("renders initial documentation links", async () => {
     const { experiment } = mockExperimentQuery("boo", {
       documentationLinks: [
         {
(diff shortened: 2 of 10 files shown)


jodyheavener:
@lmorchard If you have some time I'm running into this somewhat specific but Quite Bad issue were removing the documentation links in a specific order causes infinite re-rendering. @LZoog and I have both come up empty on it, unfortunately, so I wanted to see if you had any ideas.

It's reproducible by:

- Starting with a new experiment
- Clicking "+ Add Link" to add a new set
- Now click the "X" on the first/top-most set to remove it
- With one set remaining, clicking it should trigger the infinite re-render

You can also do it with an experiment that has saved documentation links, just do it in the same order.

lmorchard:
Taking a peek... I suspect there's some bad interaction between React's rendering algo and the `useMemo` hook that establishes `lastIndex`. I'm seeing an error "Encountered two children with the same key, `doc-link-1`" and notice that `lastIndex` resets to 0 at some point. Haven't worked out why that is yet

lmorchard:
Tinkered a bit more and tried reworking the state management. I noticed that things weren't purely idempotent in the `setDocumentationLinks` calls with the `lastIndex++` expression. I've run into weird effects in this case before, such as [React calling setState updaters twice](https://github.com/facebook/react/issues/12856).

I didn't nail it down to an exact sequence of events, but instead just went ahead and tried reworking the state management. This blows up the tests somewhat with new async updating behavior via `useEffect`, but I combined both `lastIndex` and `documentationLinks` into a single state object and that seems to resolve the infinite loop:

https://github.com/mozilla/experimenter/compare/jh/4371/add-remove-doc-links...lmorchard:jh/4371/add-remove-doc-links

pdehaan:
My first gut instinct was that we were using an array index in the `key=` value, which is generally discouraged. But I don't know if that's what's happenin' here'. Or in places where we _DO_ use array indexes, they are only used as part of the key, and not `key={idx}` which is :-1:

Ref: `https://github.com/facebook/react/issues/12873`.

But for what it's worth, I don't think we have the [`react/no-array-index-key`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) rule enabled in our nimbus-ui ESLint config. Or at least when I enable it locally on the **main** branch I get about 14 new errors. Trying to dig into each result to see if the `key` values are suitably unique looking.

jodyheavener:
@lmorchard This is a fantastic update! I appreciate you taking the time to rework it. In my local testing, however, it seems this didn't fix what I was seeing (now I'm curious what _you_ were seeing?!). Here's a video of the steps I take to repro, starting from a fresh experiment:

https://user-images.githubusercontent.com/6392049/106015910-f0687e80-608c-11eb-9dd7-11a70cd11f24.mov

I'll keep at it using your changes as a starting point -- perhaps they made the issue more obvious! If you have any thoughts lemme know 🙂

jodyheavener:
Well, one clue: If I remove these fields' usage of useCommonNestedForm everything works fine. So I'm guessing this is a weird interaction between the field mechanics and it.

LZoog:
@pdehaan FWIW I filed an issue for that: EXP-880

jodyheavener:
@lmorchard @LZoog Thanks for both of your inputs on this. I disabled the ability to remove the "last" set, so the noted issue is suppressed for now (I did also file https://jira.mozilla.com/browse/EXP-882). This PR is ready for eyes now.

LZoog:
> I disabled the ability to remove the "last" set, so the noted issue is suppressed for now

👍 Yeah I think that's the right call to get this functionality in for sure with all the time sunk in already. We should probably call it out in sprint planning tomorrow in case we want to prioritize this in the near future (I'm going to guess not, but still good to let the team know it's expected to not have the "X" option on one link).

LZoog:
😭 @jodyheavener I hate to be the bearer of bad news but the band aid doesn't totally hide the problem.

You can see the infinite render if you:
1. Click "+ Add link" twice
2. Remove the first link you just added (first available "X")
3. Try to remove the second link

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -120,233 +130,217 @@ const FormOverview = ({
...
+                      //@ts-ignore react-hook-form types seem broken for nested fields
+                      errors: (errors?.documentationLink?.[index] ||
+                        {}) as FormDocumentationLinkProps["errors"],
+                      //@ts-ignore react-hook-form types seem broken for nested fields
+                      touched: (touched?.documentationLink?.[index] ||
+                        {}) as FormDocumentationLinkProps["touched"],
+                      onRemove: () => {
+                        removeDocumentationLink(documentationLink);
+                      },
+                      canRemove: index !== 0,
```
LZoog, line 285: You `canRemove` this now right? (see what I did there)

app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
```diff
@@ -184,6 +230,90 @@ describe("FormOverview", () => {
...
+      experiment.documentationLinks!.push({
+        __typename: "NimbusDocumentationLinkType",
+        title: NimbusDocumentationLinkTitle.DESIGN_DOC,
+        link: "https://boingo.oingo",
+      });
+      fillDocumentationLinkFields(experiment.documentationLinks![1], 1);
+    });
+
+    // Add a new set and PARTIALLY populate it
+    // This set should be filtered out and therefor will
```
LZoog, line 277: the tiniest nit in the world, ~therefor~ therefore

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
```diff
@@ -120,233 +130,217 @@ const FormOverview = ({
...
+                to make a copy and add the link above
+              </Form.Text>
+              <FormErrors name="riskMitigationLink" />
+            </Form.Group>
+
+            <Form.Group controlId="documentationLinks">
+              <Form.Label>
+                Additional Links
+                <Info
+                  data-tip={DOCUMENTATION_LINKS_TOOLTIP}
```
LZoog, line 254: This is definitely not blocking, but I feel like we don't need this tooltip anymore since changing the link type to a drop-down. Would of course be an @AnaMedinac question though.

LZoog (commented):
Looks like the last field isn't getting form errors displayed...

image

Then `onChange` of any field, it clears all of them instead of that individual field. I wonder if displaying this last field could be fixed in your PR (didn't play around with it), but the onChange stuff can and should be addressed in #4299. If displaying the last link set form errors isn't quick, I think we can punt that to #4299 as well.

I'm also thinking we could assert a URL-matching client-side validation rule like the [number rule here](https://github.com/mozilla/experimenter/pull/4499/files#diff-52ccc6c28c664b4fad9c3a9af82889cc7a44bff31bfad6e4b94d1aa47c2825daR38) so users don't have to submit before getting feedback, but that can also be done in a follow up. On a related note, it looks like this is not getting an `is-valid` class though when they are valid (no green border/check):

image

I think it's OK for the submit error stuff to be handled in that mentioned issue and for the client-side validation stuff to be handled in a follow up. 🙂

Awesome work here, thank you for sticking with it and working through that infinite rendering problem. I will r+ after the `canRemove` comment is addressed.

BTW, if my PR gets merged first in the AM, you'll can change your `{...formControlAttrs("whatever_name", {})}` references to `{...formControlAttrs("whatever_name")}`, and if this is merged first, I can do that in my PR.

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
jodyheavener: It could be removed but I think I'll leave it in as the [Figma designs](https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=218083%3A23) indicate it should be this way. We could discuss removing it though, for sure, as it could be potentially confusing UX.

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
jodyheavener: Slacked her

app/experimenter/nimbus-ui/src/components/FormOverview/FormDocumentationLink.tsx
```diff
@@ -0,0 +1,99 @@
...
+    submitErrors,
+    errors,
+    touched,
+  );
+
+  return (
+    <Form.Group className="mb-0" data-testid="DocumentationLink">
+      <Form.Row>
+        <Form.Group as={Col} sm={4} md={3}>
+          <Form.Control as="select" {...formControlAttrs("title", {})}>
```
LZoog, line 59: The 2nd argument along with the 2nd argument on L77 should be removable now if you pull latest.

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
LZoog: Ahh I see. I didn't realize that was per design. I suppose we could slap a `margin-right` on the first input to better match.

LZoog (approved):
OK so, after some discussion, this LGTM. :) We've definitely got a few follow ups though, this comment can be edited (by me or you) after things are filed so the issues can be linked:

1) Submit errors are not clearing `onChange`: will be done as part of #4299 (assigned to me, there's also an issue here in Branches)
2) Edgecase issue if a link isn't filled out and users add more and submit/receive submit errors, the submit errors are not displayed correctly/off by the number of blank issues: [insert # here]
3) Client-side validation appears to be a bit flaky and not always display the green border/check, + let's create a `URL_FIELD` const to validate URLs on the client-side: [insert # here]
4) Let's make sure to get an answer from Ana on the tooltip, though not sure we need an issue for that yet

app/experimenter/nimbus-ui/src/components/FormOverview/FormDocumentationLink.tsx
jodyheavener: Updated

app/experimenter/nimbus-ui/src/components/FormOverview/index.tsx
AnaMedinac: Hey! Agreed - I don't think we need the tooltip anymore.

Issue #3373, opened 2020-09-02
Add weekly graphs


emtwo:
These would roughly be implemented in the same way for every metric so it may make sense to do them all together.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-250)
┆Issue Number: EXP-250

PR #4481, opened 2021-01-26
fix(#3373): feature(visualization) - Add weekly graphs

emtwo:
So far this only adds weekly graphs and does not yet have tests/stories. Screen cap for reference

https://user-images.githubusercontent.com/784781/105914215-0417d480-5ffc-11eb-831e-abc63f5bdfcd.mov

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
@@ -1,0 +1,132 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import React, { useState } from "react";
+import Collapse from "react-bootstrap/Collapse";
+import vegaEmbed, { VisualizationSpec } from "vega-embed";
+import {
+  BRANCH_COMPARISON,
+  BRANCH_COMPARISON_TITLE,
+} from "../../../lib/visualization/constants";
+import { lineGraphConfig } from "../../../lib/visualization/graphConfig";
+import {
+  AnalysisDataWeekly,
+  BranchDescription,
+  FormattedAnalysisPoint,
+} from "../../../lib/visualization/types";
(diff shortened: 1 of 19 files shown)


app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { ReactComponent as ExpandPlus } from "./plus.svg";
+
+/**
+ * We should only graph if there are at least 2 weeks of data
+ * because otherwise the graph is just a single point and useless.
+ *
+ * So we return empty mergedBranchData to avoid graphing when there
+ * isn't enough data
+ */
+const getMergedBranchData = (
```
jaredlockhart, line 27: Is this the kind of thing that can be prepared on the backend?

app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const embedGraphs = (
+  probeSetSlug: string,
+  probeSetName: string,
+  mergedBranchData: { [graphID: string]: FormattedAnalysisPoint[] },
+) => {
+  let branchComparisonKey: keyof typeof BRANCH_COMPARISON;
+  for (branchComparisonKey in BRANCH_COMPARISON) {
+    const branchComparison = BRANCH_COMPARISON[branchComparisonKey];
+    const graphID = `${probeSetSlug}_${branchComparison}_graph`;
```
jaredlockhart, line 64: This gets constructed a couple times, I wonder if it's possible to make it a little string template and then substitute in the parameters, just so if we ever need to change it it's in one place?

jaredlockhart (approved):
So neat testing this from storybooks 😄   Looks good to me, stoked to have more graphs! 🎉

app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
emtwo: Yea, some of this could be factored out into the backend. I'll incorporate that in the backend refactor. In a couple of sprints likely

PR #4499, opened 2021-01-28
fixes #4299 feat(nimbus): Make almost all form fields optional, client-side

### LINKED ISSUE #4299: Don't require form fields on submission after create (client-side)

LZoog:
Because:
* We want users to be able to fill in various pieces of experiments rather than requiring every field

This commit:
* Flips the default in useCommonForm to make fields optional
* Makes Overview form fields optional except for those filled out on the "New" page
* Makes Branch fields optional with the exception of a branch name
* Audience/Metrics fields are optional with the useCommonForm change
* Adds improved client-side validation on some Audience fields, requiring a number

fixes #4299

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -11,7 +11,7 @@ import {
 } from "@testing-library/react";
 import React from "react";
 import { snakeToCamelCase } from "../../../lib/caseConversions";
-import { EXTERNAL_URLS } from "../../../lib/constants";
+import { EXTERNAL_URLS, FIELD_MESSAGES } from "../../../lib/constants";
 import { MOCK_CONFIG } from "../../../lib/mocks";
 import {
   NimbusExperimentChannel,
@@ -71,43 +71,7 @@ describe("FormAudience", () => {
   });
 
   it("renders without error with default values", async () => {
-    render(
-      <Subject
-        config={{
-          ...MOCK_CONFIG,
-          targetingConfigSlug: [
-            {
-              __typename: "NimbusLabelValueType",
-              label: NimbusExperimentTargetingConfigSlug.NO_TARGETING,
-              value: NimbusExperimentTargetingConfigSlug.NO_TARGETING,
-            },
-          ],
-          firefoxMinVersion: [
-            {
-              __typename: "NimbusLabelValueType",
-              label: NimbusExperimentFirefoxMinVersion.NO_VERSION,
-              value: NimbusExperimentFirefoxMinVersion.NO_VERSION,
-            },
-          ],
-          channel: [
-            {
-              __typename: "NimbusLabelValueType",
-              label: NimbusExperimentChannel.NO_CHANNEL,
-              value: NimbusExperimentChannel.NO_CHANNEL,
-            },
-          ],
-        }}
-        experiment={{
-          ...MOCK_EXPERIMENT,
-          firefoxMinVersion: NimbusExperimentFirefoxMinVersion.NO_VERSION,
-          channel: NimbusExperimentChannel.NO_CHANNEL,
-          populationPercent: "0.0",
-          proposedDuration: 0,
-          proposedEnrollment: 0,
-          targetingConfigSlug: NimbusExperimentTargetingConfigSlug.NO_TARGETING,
-        }}
-      />,
-    );
+    renderSubjectWithDefaultValues();
     await waitFor(() => {
       expect(screen.queryByTestId("FormAudience")).toBeInTheDocument();
     });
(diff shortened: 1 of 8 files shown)


app/experimenter/nimbus-ui/src/hooks/useCommonForm/useCommonFormMethods.tsx
```diff
@@ -16,6 +17,14 @@ import { camelToSnakeCase } from "../../lib/caseConversions";
 
 export type SelectOption = { label: string; value: string };
 
+export const REQUIRED_FIELD = {
```
LZoog, line 20: Actually, I think I'll stick these in the constants file as well

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -198,33 +198,29 @@ describe("FormAudience", () => {
     expect(onSubmit).not.toHaveBeenCalled();
   });
 
-  it("marks input fields as invalid when blank", async () => {
+  it("displays expected message on number-based inputs when invalid", async () => {
```
LZoog, line 201: writing one more test here for completeness, check back soon

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -270,3 +242,44 @@ describe("FormAudience", () => {
     expect(screen.queryByTestId("missing-duration")).toBeInTheDocument();
   });
 });
+
+const renderSubjectWithDefaultValues = (onSubmit = () => {}) => {
+  render(
+    <Subject
```
lmorchard, line 248: Hmm, if these are default values, then they should probably be defaults within the `Subject` component.

But, they weren't before this PR, so I wouldn't make it a snag for merging

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -9,7 +9,7 @@ import Form from "react-bootstrap/Form";
 import InputGroup from "react-bootstrap/InputGroup";
 import { useCommonForm } from "../../../hooks";
 import { useConfig } from "../../../hooks/useConfig";
-import { EXTERNAL_URLS } from "../../../lib/constants";
+import { EXTERNAL_URLS, NUMBER_FIELD } from "../../../lib/constants";
```
lmorchard, line 12: Thought maybe putting this in constants would be overkill, but a) they're used a lot and b) can't think of a better place so 👍

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.test.tsx
```diff
@@ -136,13 +137,36 @@ describe("FormBranch", () => {
     const branch = {
       ...MOCK_ANNOTATED_BRANCH,
     };
-    const { container } = render(<SubjectBranch branch={branch} />);
```
lmorchard, line 139: Funny that this ends up not saving many characters now, but it would if more props get added

lmorchard (approved):
LGTM, tests pass and things seem functional. Only comments are just kind of thinking out loud and not even nits to fix

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
LZoog: Yeah, I originally had them living in `useCommonFormMethods` and didn't think `constants` was the _most_ ideal place for them, but after pushing and looking at my PR, I thought it was a better spot than where I had them originally due to how much they're used 🤔

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
LZoog: Hmm yeah, they're defaults with a fresh experiment. That would make the most sense, though the Subject mock is pulling from the values we have in `mockExperimentQuery`. I think we'd need some test refactoring :/ I think we've just stuck whatever made the most sense for tests at the time into `mockExperimentQuery`. I'm going to leave this as-is, we _could_ make a ticket for updating what `mockExperimentQuery` to reflect what's default, but I have a feeling we'd never get to it because it'd require a lot of test alterations

PR #4508, opened 2021-01-29
fix #4463 feat(nimbus): add featureIds to V6 API

### LINKED ISSUE #4463: Update the DTO to include a top-level featureIds property
We should add a top-level featureIds property that is an array of all features referenced in branches. Branch schema should stay the same

Github PR: https://github.com/mozilla/nimbus-shared/pull/154/files

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-863)
┆Issue Number: EXP-863

jaredlockhart:
Because

* The 1.3.0 nimbus shared experiment schema allows for a top level featureIds property which stores the list of feature ids targeted by the experiment

This commit

* Adds a featureIds property to the V6 API which includes an array of 1 element for now (until we later support multiple feature configs per experiment)

### DIFF
--- a/app/experimenter/docs/openapi-schema.json
+++ b/app/experimenter/docs/openapi-schema.json
@@ -3014,6 +3014,10 @@
           "referenceBranch": {
             "type": "string",
             "readOnly": true
+          },
+          "featureIds": {
+            "type": "string",
+            "readOnly": true
           }
         },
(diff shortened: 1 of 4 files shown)


k88hudson (approved):
Wow, that was easy!

PR #4533, opened 2021-02-03
fixes #4355 fix(nimbus): Validate branch names for uniqueness server-side

### LINKED ISSUE #4355: Duplicate branch names signal a database error
If you create two branches with the same name, you get a general database error rather than a nice message telling you the names are duplicated:

Screen Shot 2021-01-05 at 2 58 06 PM

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-792)
┆Issue Number: EXP-792

LZoog:
Because:
* We want a human-friendly error message when duplicate branch names are present

This commit:
* Adds a validation function checking for branch name duplicates in the serializer and returns an error message if branch names are not all unique

Co-authored-by: Jared Lockhart <119884+jaredlockhart@users.noreply.github.com>

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -6,6 +6,7 @@ from django.utils.text import slugify
 from rest_framework import serializers
 
 from experimenter.experiments.changelog_utils import generate_nimbus_changelog
+from experimenter.experiments.constants.nimbus import NimbusConstants
 from experimenter.experiments.models import NimbusExperiment
 from experimenter.experiments.models.nimbus import (
     NimbusBranch,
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -68,6 +69,34 @@ def _validate_feature_value_against_schema(self, schema, value):
...
+                branch["name"] for branch in data["treatment_branches"]
+            ]
+            all_names = [ref_branch_name, *treatment_branch_names]
+            unique_names = set(all_names)
+
+            if len(all_names) != len(unique_names):
+                raise serializers.ValidationError(
+                    {
+                        "reference_branch": {
+                            "name": NimbusConstants.ERROR_DUPLICATE_BRANCH_NAME
```
lmorchard, line 89: Hmm, might be better if this were a general global error rather than throwing this on all branches and not just the duplicate branches. But, I can't remember off the top of my head if that's actually possible within this serializer.

lmorchard (approved):
One small thing but overall LGTM :+1:

k88hudson:
Oh nice! I saw this in demos today

app/experimenter/experiments/api/v5/serializers.py
LZoog: Hrmm... Could go either way for me on the preference there. I'll leave it for now, we could explore in a follow up if we'd like.

PR #4534, opened 2021-02-03
Replace `make test` with `make check` in README setup

mythmon:
`make test` doesn't exist. It seems to me like `make check` is the right command.


lmorchard (approved):
Just ran into this again yesterday and of course forgot to do anything about it. :+1:

PR #4537, opened 2021-02-03
fixes #4536 fix(nimbus): serializer fix for mobile experiment targeting

### LINKED ISSUE #4536: Experimenter incorrectly outputs default desktop targeting for mobile experiments
Actual: Experimenter outputs `"browserSettings.update.channel == \"nightly\" && 'app.shield.optoutstudies.enabled'|preferenceValue"` for targeting on mobile experiments, which is Desktop-only

Expected: Experimenter should return `""` for mobile experiments with no targeting.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-909)
┆Issue Number: EXP-909

lmorchard:
Because:

- Experimenter should return "" for mobile experiments with no targeting.

This commit:

- tweaks the composition of the serialized targeting expression so that
  desktop-only parts are only included when the experiment specifies
  the desktop application

### DIFF
--- a/app/experimenter/experiments/api/v6/serializers.py
+++ b/app/experimenter/experiments/api/v6/serializers.py
@@ -122,33 +122,28 @@ class NimbusExperimentSerializer(serializers.ModelSerializer):
             return obj.reference_branch.slug
 
     def get_targeting(self, obj):
-        version_expr = ""
-        if obj.firefox_min_version:
-            version_expr = "{version_check} && ".format(
-                version_check=NimbusExperiment.TARGETING_VERSION.format(
-                    version=obj.firefox_min_version
-                )
-            )
+        exprs = []
 
-        channel_expr = ""
-        if obj.channel:
-            channel_expr = "{channel_check} && ".format(
-                channel_check=NimbusExperiment.TARGETING_CHANNEL.format(
-                    channel=obj.channel
+        if obj.is_desktop_experiment:
+            if obj.channel:
+                exprs.append(
+                    NimbusExperiment.TARGETING_CHANNEL.format(channel=obj.channel)
                 )
-            )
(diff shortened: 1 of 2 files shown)


lmorchard:
So, the bug is that we're not serializing to "" for no targeting on a mobile application experiment.

What I guessed on here is whether we *should* serialize to *something* when there *is* a target.

app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -122,33 +122,28 @@ def get_referenceBranch(self, obj):
...
-                )
-            )
+        exprs = []
 
-        channel_expr = ""
-        if obj.channel:
-            channel_expr = "{channel_check} && ".format(
-                channel_check=NimbusExperiment.TARGETING_CHANNEL.format(
-                    channel=obj.channel
+        if obj.is_desktop_experiment:
```
lmorchard, line 127: There's a redundant `if obj.is_desktop_experiment` here. But, I didn't want to touch all the tests to reorder the expressions. That way, the original tests were mostly intact & untouched & still passing after this refactor. Also, it looks like the second `if obj.is_desktop_experiment` is part of a TODO to remove, so 🤷‍♂️

app/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -235,13 +241,40 @@ def test_serializer_outputs_targeting_for_experiment_without_channels(self):
         )
         check_schema("experiments/NimbusExperiment", serializer.data)
 
+    def test_serializer_outputs_expected_targeting_for_mobile(self):
```
k88hudson, line 244: Yep this looks good!

k88hudson (approved):

PR #4541, opened 2021-02-04
fixes #4423 fixes #4424 feat(nimbus): end experiment tasks and GQL mutation

### LINKED ISSUE #4423: Automated end task
Create a new task that when invoked deletes an experiments record from the remote settings collection and marks the collection for review.  We already have a task to detect when an experiments collection is removed and update it in the db and generate the changelog.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-830)
┆Issue Number: EXP-830

### LINKED ISSUE #4424: Add GQL mutation to invoke the end experiment task
We need to add a simple mutation that will invoke the end experiment celery task.  The only parameter we need is the experiment id, and I think the only validation we need is to check that its status is currently live, and ending hasn't been already requested (by checking whichever way we choose to store that state, see below).  We'll need to track some way to indicate that the end experiment task has been fired and is waiting for review.  I can think of a couple ways to do it

* Add an 'Ending' status after live before complete (probably the most costly to implement since it'll involve a whole new state and have impacts acros the backend and frontend)
* Add an 'end_requested' bool (probably simplest to implement with smallest change surface)
* Create a changelog entry that the end request was received and check for its existence (requires no schema changes but a bit tricky to figure out exactly where in the changelog to store the state)

Calling the end task multiple times is idempotent so it doesn't matter if it's invoked multiple times, but you'll want some way to indicate to the user after they click the 'End Experiment' button that the end request was received and is being processed/awaiting review, so pick whichever way to store that state is most economical.

We'll also want to expose whichever way we store that the end request was logged in the read experiment query, so if it's just a field right on the model (state or bool) we can just expose that field, if it's stored in the changelog probably a method to wrap looking it up will be a helpful abstraction.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-831)
┆Issue Number: EXP-831

lmorchard:
Because:

- we want to end an experiment by deleting it from kinto and tracking
  the pending deletion attempt through kinto approval

This commit:

- adds an is_end_requested flag to the NimbusExperiment model

- adds an EndExperiment GQL mutation to set is_end_requested to true

- adds a delete_from_kinto method to the kinto client wrapper

- adds a nimbus_end_experiment_in_kinto task to nimbus celery tasks that
  deletes an experiment from kinto, sets is_ended = True, and generates a
  changelog entry

- tweaks the nimbus_check_kinto_push_queue periodic task to check for
  experiments with Live status and is_end_requested - if no review is
  pending in kinto, the first of these experiments is deleted

- tweaks the nimbus_check_kinto_push_queue periodoc task to check for
  a rejection from kinto to end an experiment and reverts the
  is_end_requested flag when it happens

- updates tests and GQL types as necessary


app/experimenter/experiments/models/nimbus.py
```diff
@@ -30,6 +30,7 @@ class NimbusExperiment(NimbusConstants, models.Model):
     public_description = models.TextField(default="")
     risk_mitigation_link = models.URLField(max_length=255, blank=True)
     is_paused = models.BooleanField(default=False)
+    is_ended = models.BooleanField(default=False)
```
jaredlockhart, line 33: I think we want this to be the other way around.  When an experiment is complete it moves to `status=COMPLETE`, so I don't think we need a bool to track when it's actually ended, we want something to track when it's REQUESTED to be ended, something like `is_end_requested`.

app/experimenter/kinto/client.py
```diff
@@ -66,3 +66,15 @@ def get_main_records(self):
...
+        self.kinto_http_client.delete_record(
+            id=id,
+            collection=self.collection,
+            bucket=settings.KINTO_BUCKET,
+        )
+        self.kinto_http_client.patch_collection(
+            id=self.collection,
+            data={"status": KINTO_REVIEW_STATUS},
+            bucket=settings.KINTO_BUCKET,
+        )
```
jaredlockhart, line 80: Yep this looks right 👍

app/experimenter/kinto/tasks.py
```diff
@@ -213,3 +213,36 @@ def nimbus_check_experiments_are_complete():
...
+
+    try:
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        logger.info(f"Deleting {experiment} from Kinto")
+
+        kinto_client = KintoClient(
+            NimbusExperiment.KINTO_APPLICATION_COLLECTION[experiment.application]
+        )
+
+        kinto_client.delete_from_kinto(experiment_id)
```
jaredlockhart, line 236: Kinto uses `experiment.slug` as the record ID, not the database ID

app/experimenter/kinto/tasks.py
```diff
@@ -213,3 +213,36 @@ def nimbus_check_experiments_are_complete():
                 logger.info(f"{experiment} status is set to Complete")
 
     metrics.incr("check_experiments_are_complete.completed")
+
+
+@app.task
+@metrics.timer_decorator("end_experiment_in_kinto")
+def nimbus_end_experiment_in_kinto(experiment_id):
```
jaredlockhart, line 220: So if we set the task up like this, as a single invocation, and it fails (the common case of which will be that the collection already has a pending change), then it'll never retry.  So I think a more hardened approach is to

1) mutation sets `is_end_requested` bool to True
2) a scheduled task wakes up and checks for experiments that have end requested but are still present in the collection
3) if the collection is not currently pending a review, then we can call `delete_from_kinto`

So since we already have a task that does most of that here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L80

We can probably slot the removal in right here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L126

instead of making this a separate single invocation task.

jaredlockhart (changes requested):
Thanks for picking this up @lmorchard !  Great first pass on this, all the right things in all the right places, but we might need to reorder the workflow here, I wrote my thoughts below, let me know what you think.

lmorchard:
Pulling back to draft to rework & resolve a few more things

app/experimenter/kinto/tasks.py
```diff
@@ -99,7 +99,14 @@ def nimbus_check_kinto_push_queue():
...
             experiment = NimbusExperiment.objects.get(slug=rejected_slug)
-            experiment.status = NimbusExperiment.Status.DRAFT
+            if (
+                experiment.status == NimbusExperiment.Status.LIVE
+                and experiment.is_end_requested
+            ):
+                experiment.is_end_requested = False
+            else:
+                experiment.status = NimbusExperiment.Status.DRAFT
+
```
lmorchard, line 109: Probably rare that we'll have an experiment deletion rejected on the kinto side. But, when I tried it, I realized that this code threw the status back into Draft status in Experimenter while the experiment was still Live over in remote settings. This leaves it live but reverts the request to end the experiment. Would be nice if we had some indication of the rejection message, but I don't think we surface rejection messages anywhere yet.

app/experimenter/experiments/api/v5/mutations.py
```diff
@@ -68,8 +68,43 @@ def mutate(cls, root, info, input: ExperimentInput):
...
+            experiment.save()
+            obj = experiment
+            msg = "success"
+        else:
+            obj = None
+            status = experiment.status
+            msg = (
+                f"Nimbus Experiment has status '{status}', but can only "
+                "be ended when set to 'Live'."
+            )
```
lmorchard, line 94: Also added a little bit of status validation here because I was able to set a Draft status experiment to end and it proceeded from Live to Completed after I pushed it. Probably not the end of the world, but not useful

lmorchard:
Reworked the flow so that the actual task to end an experiment is called in the periodic task `nimbus_check_kinto_push_queue` based on `is_end_requested` set via GQL mutation in the database.

app/experimenter/kinto/tasks.py
k88hudson: The plan is to move the review step into experimenter in https://jira.mozilla.com/browse/EXP-865, so probably makes sense to wait until then to do that

jodyheavener:
I have no problem with this approach, mostly just curious: could an argument be made for adding `ending` as a type of experiment status so we could just use the `updateExperiment` mutation and look at the `status` field instead of a new mutation and field?

app/experimenter/experiments/api/v5/mutations.py
```diff
@@ -68,8 +68,43 @@ def mutate(cls, root, info, input: ExperimentInput):
...
+    class Arguments:
+        input = ExperimentIdInput(required=True)
+
+    @classmethod
+    def mutate(cls, root, info, input: ExperimentIdInput):
+        experiment = NimbusExperiment.objects.get(id=input.id)
+
+        if experiment.status == NimbusExperiment.Status.LIVE:
+            experiment.is_end_requested = True
+            experiment.save()
```
jaredlockhart, line 85: We probably want to add a changelog here to capture the user that requested the end.

app/experimenter/experiments/api/v5/mutations.py
```diff
@@ -68,8 +68,43 @@ def mutate(cls, root, info, input: ExperimentInput):
         return handle_with_serializer(cls, serializer)
 
 
+class EndExperiment(graphene.Mutation):
```
jaredlockhart, line 71: If we expose the `is_end_requested` field in TheOne™️ mutation, we can validate field/status pair in the serializer, and then we don't need a custom mutation at all.  Is that better than having a separate mutation?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -671,3 +685,51 @@ def test_update_experiment_status_error(self):
...
+        self.assertEqual(response.status_code, 200, response.content)
+        content = json.loads(response.content)
+        result = content["data"]["endExperiment"]
+        self.assertEqual(
+            result["nimbusExperiment"],
+            {
+                "id": experiment.id,
+                "isEndRequested": True,
+            },
+        )
```
jaredlockhart, line 712: I would assert that the db is updated here too.

app/experimenter/kinto/tasks.py
jaredlockhart: Oh good catch 👍

app/experimenter/experiments/api/v5/mutations.py
```diff
@@ -68,8 +68,43 @@ def mutate(cls, root, info, input: ExperimentInput):
...
+    status = graphene.Int()
+
+    class Arguments:
+        input = ExperimentIdInput(required=True)
+
+    @classmethod
+    def mutate(cls, root, info, input: ExperimentIdInput):
+        experiment = NimbusExperiment.objects.get(id=input.id)
+
+        if experiment.status == NimbusExperiment.Status.LIVE:
```
jaredlockhart, line 83: It's not a huge deal but I've learned that these kinds of status checks are better written as model properties because at some point we might add another status that's also part of what we consider LIVE and then there's only one place to update it rather than everywhere we check the status, but we could also leave that till later.

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -671,3 +685,51 @@ def test_update_experiment_status_error(self):
...
+
+    def test_end_experiment_in_kinto_fails_with_nonlive_status(self):
```
jaredlockhart, line 714: 👍

jaredlockhart (approved):
Awesome work @lmorchard and thnx for the fast turnaround on the refactor.  Everything looks right to me, tests are very clear and well laid out, couldn't have written it better myself!  Couple thoughts on tweaks but looks great! 👏 👏 👏

app/experimenter/experiments/api/v5/mutations.py
lmorchard: Yeah, I tried that at one point. Problem is that we have [NimbusStatusRestrictionMixin](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L234-L249) constraining *any* edits via that mutation to just DRAFT status. I started playing with reworking that validator to accept a subset of changes based on the current status, but decided that just adding a new specific mutation was cleaner, at least for now.

jaredlockhart: Ah yep good point 👍

app/experimenter/kinto/tasks.py
lmorchard: > The plan is to move the review step into experimenter in https://jira.mozilla.com/browse/EXP-865, so probably makes sense to wait until then to do that

Yeah, we're going to change the review process in the future - but more immediately, this status change could cause a problem unless we want to delay the end experiment feature

app/experimenter/experiments/api/v5/mutations.py
lmorchard: We might want a follow up issue for that? Doesn't look like NimbusExperiment has model properties for status right now, and most spots I could see were just using constant comparisons like this

jaredlockhart: Yeah we can file a followup 👍

lmorchard:
> I have no problem with this approach, mostly just curious: could an argument be made for adding ending as a type of experiment status so we could just use the updateExperiment mutation and look at the status field instead of a new mutation and field?

Two things with that:

* The updateExperiment mutation rejects making changes while the experiment is Live, by way of [NimbusStatusRestrictionMixin](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L234-L249). Could change that, but reworking the serializer to conditionally accept edits based on status looked like a bigger project than just adding a single-purpose mutation
* Adding a new status type is a bit more complex than just adding a boolean field. And technically, all this status would mean is that a change is pending - for all other intents and purposes the experiment is still Live until the change resolves and it moves to Completed.

If we switch over to using model properties to define current experiment status, we could build in some more complex logic based on other model fields. So, we *could* change this to a new status without needing to change fields in the DB.

app/experimenter/experiments/api/v5/mutations.py
lmorchard: FWIW, we *do* have [a `getStatus` utility function](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/lib/experiment.ts#L9-L39) on the client side, but doesn't looks like we have the equivalent on the server side

jaredlockhart: Oh interesting!

lmorchard:
Rebased against main and added some assertions against the DB for the end experiment mutation

lmorchard:
Alright, added a few more test assertions and now I think I'll merge this to unblock the client-side stuff

PR #4549, opened 2021-02-05
Fix #4548 - fix(visualization) Make page loading faster.

### LINKED ISSUE #4548: Speed up loading of visualizations
its taking 10+ seconds to load the viz page. Need to investigate what is happening and improve this load time. Way too slow!

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-917)
┆Issue Number: EXP-917

emtwo:
Because:
* The visualization page loads really slowly

This commit:
* Moves code out of a loop that doesn't need to be there. Couple other code improvements. Speed increased by 75%+!

I tested this on this page: https://experimenter.services.mozilla.com/nimbus/onoff-aboutwelcome-experiment/results

It takes about 10.5 seconds to load there on prod. With this PR, this goes down to about 2.5 seconds.

### DIFF
--- a/app/experimenter/visualization/api/v3/views.py
+++ b/app/experimenter/visualization/api/v3/views.py
@@ -147,8 +147,11 @@ def generate_results_object(data, experiment, window="overall"):
     # These are metrics sent from Jetstream that are not explicitly chosen
     # by users to be either primary or secondary
     other_metrics = {}
-
     results = {}
+
+    result_metrics, primary_metrics_set = get_results_metrics_map(
+        experiment.primary_probe_sets, experiment.secondary_probe_sets
+    )
     for row in data:
         branch = row.get("branch")
         metric = row.get("metric")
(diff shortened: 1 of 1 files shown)


LZoog:
Woah nice find!! I'll let Jared r+ since he's more familiar here, but that's definitely a crazy speed boost.

pdehaan:
I can definitely reproduce the ~10s delay on production (without this PR merged).
Looks like it's blocking on https://experimenter.services.mozilla.com/api/v3/visualization/onoff-aboutwelcome-experiment/ and we're waiting 8.96s for TTFB (the current response payload is ~14 KB). All the other requests on that experiment page above seem to take between 0.38ms-1.06ms, so definitely is blocking on that single large request.

Q: Are we caching the /v3/visualization/* results on a server, or recalculating dynamically on each page view? My gut (but mostly devtools) is telling me that we calculate each time, since the page load times aren't improving after each refresh.

emtwo:
> Q: Are we caching the /v3/visualization/* results on a server, or recalculating dynamically on each page view? My gut (but mostly devtools) is telling me that we calculate each time, since the page load times aren't improving after each refresh.

Correct, we're not caching the /v3/visualization/* results, but that is totally an optimization that can further improve load time here

pdehaan:
> but that is totally an optimization that can further improve load time

Rad. Filed enhancement request as #4554

jaredlockhart (approved):
🎉 🎉 🎉  great catch @emtwo looks good to me 👍 👍 👍

PR #4556, opened 2021-02-05
fixes #4547 feat(Legacy) added projects to v1 api

### LINKED ISSUE #4547: Expose legacy experiment projects in v1 API
For the [High-level Experiments Metrics project](https://docs.google.com/document/d/1kDjBp46p6TtL31us-olWLayCIk963KKcQQIEKgGkszM/edit#), we need to be able to access the project that an experiment belongs to from the API. It would be great if that could be added to the v1 serializer for experiments.

Alternatively, if there is a better way to get this data, such as another API endpoint, I'd be happy to use that instead.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-916)
┆Issue Number: EXP-916

tiftran:
Because:

* this will allow sources to consume project info for experiments

This Commit:

* adds projects to the v1 experiments api

### DIFF
--- a/app/experimenter/experiments/tests/api/v1/test_serializers.py
+++ b/app/experimenter/experiments/tests/api/v1/test_serializers.py
@@ -153,6 +153,7 @@ class TestExperimentSerializer(TestCase):
                 "results_measure_impact": None,
                 "results_impact_notes": None,
             },
+            "projects": list(experiment.projects.values_list("name", flat=True)),
         }
 
         self.assertEqual(set(serializer.data.keys()), set(expected_data.keys()))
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tests/api/v1/test_serializers.py
```diff
@@ -153,6 +153,7 @@ def test_serializer_outputs_expected_schema(self):
                 "results_measure_impact": None,
                 "results_impact_notes": None,
             },
+            "projects": list(experiment.projects.values_list("name", flat=True)),
```
jaredlockhart, line 156: Might wanna add an explicit ordering here so this doesn't cause an intermittent?  Or do a set comparison?  Either or.

jaredlockhart (approved):
Looks good @tiftran 👍 might just wanna intermittent proof that ordered comparison.

PR #4557, opened 2021-02-06
fixes #4424 feat(nimbus): add GQL mutation to end an experiment

### LINKED ISSUE #4424: Add GQL mutation to invoke the end experiment task

lmorchard:
PR #4541 should get reviewed & merged before this one, but I figured I'd try getting a head start and build on that.

Because:

- we need to add a simple mutation that will invoke the end experiment celery task.

This commit:

- adds an ExperimentIdInput GQL type

- adds an EndExperiment GQL mutation that takes ExperimentIdInput

- updates to mocks & tests as needed

- updates to generated types


app/experimenter/experiments/api/v5/inputs.py
```diff
@@ -35,6 +35,10 @@ class Meta:
         )
 
 
+class ExperimentIdInput(graphene.InputObjectType):
+    id = graphene.Int()
```
jaredlockhart, line 39: Does this need to be its own input or can you use a subset of the ExperimentInput that includes ID?

app/experimenter/experiments/api/v5/mutations.py
```diff
@@ -68,8 +69,28 @@ def mutate(cls, root, info, input: ExperimentInput):
...
+    message = ObjectField()
+    status = graphene.Int()
+
+    class Arguments:
+        input = ExperimentIdInput(required=True)
+
+    @classmethod
+    def mutate(cls, root, info, input: ExperimentIdInput):
+        exp = NimbusExperiment.objects.get(id=input.id)
+        kinto.tasks.nimbus_end_experiment_in_kinto.delay(exp.id)
```
jaredlockhart, line 83: Based on [feedback in the other pr](https://github.com/mozilla/experimenter/pull/4541#discussion_r572259641) we might want to set a db flag here instead of invoking the task

jaredlockhart (commented):
Yep this all looks right, just might need to update it based on changes to the task pr 👍

lmorchard:
Going to close this PR and just roll it into #4423

app/experimenter/experiments/api/v5/inputs.py
lmorchard: Figured this was better as its own input, if only because this is explicit about not needing or wanting any of the other fields

PR #4177, opened 2020-12-08
fixes #4159 fix(nimbus): Display JSON link if status is not draft/review

LZoog:
Because:
* The data from the endpoint is only available when the status is at least 'accepted'.

This commit:
* Updates the conditional for displaying the link in the Summary component.

PR #4573, opened 2021-02-08
fixes #4551 fix(nimbus): Allow the JSON link to show up during REVIEW

### LINKED ISSUE #4551: Add "See full JSON representation" button to Review page
This would make it way easier for nimbus team members to test using staging experimenter.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-919)
┆Issue Number: EXP-919

k88hudson:
Because

* It will make it easier to get access to the JSON for testing via staging

This commit

* Add JSON link to summary view for REVIEW status

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/Summary/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/index.tsx
@@ -34,7 +34,7 @@ const Summary = ({ experiment }: SummaryProps) => {
 
       <div className="d-flex flex-row justify-content-between">
         <h2 className="h5 mb-3">Summary</h2>
-        {!status.draft && !status.review && (
+        {!status.draft && (
           <span>
             <LinkExternal
               href={`/api/v6/experiments/${experiment.slug}/`}
--- a/app/experimenter/experiments/api/v6/views.py
+++ b/app/experimenter/experiments/api/v6/views.py
@@ -14,7 +14,7 @@ class NimbusExperimentViewSet(
 ):
     lookup_field = "slug"
     queryset = NimbusExperiment.objects.all().exclude(
-        status__in=[NimbusExperiment.Status.DRAFT, NimbusExperiment.Status.REVIEW]
+        status__in=[NimbusExperiment.Status.DRAFT]
     )
     serializer_class = NimbusExperimentSerializer
 
--- a/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
@@ -40,31 +40,32 @@ describe("Summary", () => {
   });
 
   describe("JSON representation link", () => {
-    it("renders when status is not 'draft' or 'review'", () => {
+    function renderWithStatus(status: NimbusExperimentStatus) {
       const { experiment } = mockExperimentQuery("demo-slug", {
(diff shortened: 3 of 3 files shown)


LZoog:
Is this link actually available when the status is "review"? I could have sworn it said it wasn't available yet when I did https://github.com/mozilla/experimenter/pull/4177, my commit comment says:

> The data from the endpoint is only available when the status is at least 'accepted'.

k88hudson:
Ah sorry, you're right – review is excluded

k88hudson:
Alright so it's easy enough to allow the API to send experiments in review, but right now buckets are only assigned in the kinto task so– I can't really think of why this has to be the case v.s. when the experiment is put up for review, but let me investigate

jaredlockhart:
The answer is we only assign buckets RIGHT before we push the experiment to kinto to prevent buckets being assigned and unused or two experiments attempting to assign buckets simultaneously (since the kinto queue acts as an exactly once queue preventing collisions).  We COULD assign buckets at review time so the full JSON is available after users click the ready for review button, and in theory it's not likely we'll hit the collision case, and experiments that assign buckets but don't end up using them is fine when all experiments have their own namespace, but it becomes a problem when experiments start sharing namespaces (ie exclusion groups) and potentially get assigned buckets but never launch.  But we can deal with those issues later if we wanna push bucket assignment up to review time.

k88hudson:
> The answer is we only assign buckets RIGHT before we push the experiment to kinto to prevent buckets being assigned and unused or two experiments attempting to assign buckets simultaneously (since the kinto queue acts as an exactly once queue preventing collisions).

Seems like we'll need to change this when we change the collision model, so maybe we could do it then. I would like to get this unblocked for testing through, how would you feel about returning a default value or something? How is this going to work for `PREVIEW` mode?

app/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -34,7 +34,7 @@ const Summary = ({ experiment }: SummaryProps) => {
 
       <div className="d-flex flex-row justify-content-between">
         <h2 className="h5 mb-3">Summary</h2>
-        {!status.draft && !status.review && (
+        {!status.draft && (
```
jaredlockhart, line 37: Oh I didn't realize we had this check mirrored here, it would be nice if we did this in one place on the backend and just sent up a property to the frontend.

jaredlockhart (commented):
Yeah so if we want the bucketing to be available here we'll have to move this

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L50-L59

to here

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L390-L394

when status goes to review

app/experimenter/nimbus-ui/src/components/Summary/index.tsx
k88hudson: Like an `apiLink` or something?

jaredlockhart: Yeah a string that's generated on the backend and the frontend can just show it if it's not null?

jaredlockhart (approved):
LGTM thnx @k88hudson 👍 

Filed https://github.com/mozilla/experimenter/issues/4580 as a followup

Issue #4498, opened 2021-01-28
Update experiment GQL query to reflect having been pushed into the QA / preview collection


data-sync-user:
In order to reflect an experiment having been pushed into the QA / preview collection (EXP-887), we need to add an indication of that in the GQL query fetching an experiment.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-888)
┆Issue Number: EXP-888

data-sync-user:
➤ Les Orchard commented:

Knowing more about how Kinto tasks work, this issue will probably just be naturally resolved as part of EXP-567 and/or EXP-887 by adding a new Preview state that gets automagically included in GQL queries

PR #4579, opened 2021-02-08
fixes #4490 feat(nimbus): redirect away from edit pages when experiment is in preview state

### LINKED ISSUE #4490: Ensure that an experiment in QA / preview state cannot be edited
From https://github.com/mozilla/experimenter/pull/4482#issuecomment-768329521:

bq. the PREVIEW state needs to prevent editing of the experiment like the REVIEW state does (since changes wouldn't get automatically published if an edit was made)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-879)
┆Issue Number: EXP-879

jodyheavener:
Closes #4490 

This PR implements redirecting away from experiment Edit pages while an experiment is in a "preview" state.

It also includes some @ts-ignores until #4498 is complete, similar to how we roughed out the UI.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
@@ -97,6 +97,22 @@ describe("PageEditAudience", () => {
     });
   });
 
+  it("redirects to the review page if the experiment status is preview", async () => {
+    const { mock, experiment } = mockExperimentQuery("demo-slug", {
+      // @ts-ignore EXP-879 mock value until backend API & types are updated
+      status: NimbusExperimentStatus.PREVIEW,
+    });
+    render(<Subject mocks={[mock]} />);
+    await waitFor(() => {
+      expect(navigate).toHaveBeenCalledWith(
+        `${BASE_PATH}/${experiment.slug}/request-review`,
+        {
+          replace: true,
+        },
+      );
+    });
+  });
+
   it("redirects to the design page if the experiment status is complete", async () => {
     const { mock, experiment } = mockExperimentQuery("demo-slug", {
       status: NimbusExperimentStatus.COMPLETE,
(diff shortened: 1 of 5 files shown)


app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -97,6 +97,22 @@ describe("PageEditAudience", () => {
     });
   });
 
+  it("redirects to the review page if the experiment status is review", async () => {
```
pdehaan, line 100: nit: I'm a bit confused by "... if the experiment status is review".

1. Should that be something like "in the 'review' state"?
2. Probably just a copy pasta, but should that say "is preview"?
    L70 above already has a test `it("redirects to the review page if the experiment status is review", async () => {` which could be confusing since they share a test name, but set different statuses.

I guess same comment applies to the 3 similar copy-pasta-pasta-pasta's below too.

jodyheavener: heh, yes good catch. will update.

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -97,6 +97,22 @@ describe("PageEditAudience", () => {
     });
   });
 
+  it("redirects to the review page if the experiment status is preview", async () => {
```
lmorchard, line 100: Huh, seems weird that we have to test this for each and every page - but I see we're also doing it for Complete status. 

Might be worth a maintenance sprint task to pull all of these into a single test for `editCommonRedirects` in `app/experimenter/nimbus-ui/src/lib/experiment.ts`. If we test anything from page-to-page, it should just be that `editCommonRedirects` is called

lmorchard (approved):
:+1: one nit, but not worth solving here I don't think

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
jodyheavener: Deal! https://jira.mozilla.com/browse/EXP-937

PR #4583, opened 2021-02-09
fix #4580 feat(nimbus): allocate buckets at review rather than accepted time

### LINKED ISSUE #4580: Generate bucketing at review time rather than accepted time
To allow users to review an experiments DTO as early as possible, we should generate the bucket allocations for an experiment when it moves into `REVIEW` rather than `ACCEPTED`.  Detailed instructions:

https://github.com/mozilla/experimenter/pull/4573#pullrequestreview-585919933

Additionally, we should make the DTO URL a property on NimbusExperiment and expose it through the GQL API so that it doesn't need to be conditionally generated on the frontend.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-933)
┆Issue Number: EXP-933

jaredlockhart:
Because

* We'd like users to be able to review the full DTO after they mark an experiment for review including bucket range

This commit

* Adds a model method to check when we should allocate buckets so we can extend it to include other statuses later
* Moves bucket allocation to a model method
* Removes bucket allocation from the kinto deploy task
* Adds bucket allocation to the NimbusExperiment V5 serializer
* Updates NimbusExperimentFactory to use the new methods
* Disables an intermittently failing test with reference to ticket

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -1,5 +1,6 @@
 import datetime
 import time
+from decimal import Decimal
 from urllib.parse import urljoin
 
 from django.conf import settings
@@ -177,6 +178,22 @@ class NimbusExperiment(NimbusConstants, models.Model):
         self.reference_branch = None
         self.save()
         self.branches.all().delete()
+
+    @property
+    def should_allocate_bucket_range(self):
+        return self.status in [NimbusExperiment.Status.REVIEW]
+
+    def allocate_bucket_range(self):
+        if not NimbusBucketRange.objects.filter(experiment=self).exists():
+            NimbusIsolationGroup.request_isolation_group_buckets(
+                self.slug,
+                self,
+                int(
+                    self.population_percent
+                    / Decimal("100.0")
+                    * NimbusExperiment.BUCKET_TOTAL
+                ),
+            )
 
 
 class NimbusBranch(models.Model):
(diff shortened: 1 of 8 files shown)


app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -183,6 +185,27 @@ def test_clear_branches_deletes_branches_without_deleting_experiment(self):
         self.assertEqual(experiment.branches.count(), 0)
         self.assertEqual(experiment.changes.count(), 1)
 
+    @parameterized.expand(
```
lmorchard, line 188: Hmm, this is probably a thing we should slap on more status-dependent tests for the sake of completeness

app/experimenter/experiments/models/nimbus.py
```diff
@@ -177,6 +178,22 @@ def delete_branches(self):
         self.save()
         self.branches.all().delete()
 
+    @property
```
lmorchard, line 181: Yeah, we should do more of this for various bits of logic

lmorchard (approved):
LGTM :+1:

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
jaredlockhart: Yeah I've been thinking about that, prolly the kind of thing we can do on a downtime maintenance week 👍

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Ya totes, another thing we can go back through and review when we get some breathing time

PR #4592, opened 2021-02-10
feat(docs): add small readme blurb about accessing rs locally

jodyheavener:


README.md
```diff
@@ -221,6 +221,17 @@ Run the integration test suite inside a containerized instance of Firefox. You m
 
 Start a linux VM container with VNC available over `vnc://localhost:5900` with password `secret`. Right click on the desktop and select `Applications > Shell > Bash` and enter `tox -c tests/integration/` to run the integration tests and watch them run in a Firefox instance you can watch and interact with.
 
+## Accessing Remote Settings locally
+
+In development you may wish to approve or reject changes to experiments as if they were on Remote Settings. You can do so here: `http://localhost:8888/v1/admin/`
```
LZoog, line 226: Will you add this to the list of important links? 😁 (https://github.com/mozilla/experimenter#important-links)

jodyheavener: @LZoog sure, can you tell me the address for prod/stage?

k88hudson: Beat me to it!

prod: https://settings-writer.prod.mozaws.net/v1/admin/#/
stage: https://settings-writer.stage.mozaws.net/v1/admin/#/

jodyheavener: Done!

README.md
```diff
@@ -221,6 +224,17 @@ Run the integration test suite inside a containerized instance of Firefox. You m
...
+
+Default credentials:
+
+- As an admin: `admin` / `admin`
+- As a reviewer: `reviewer` / `review`
```
lmorchard, line 234: Username `reviewer` or `review`? (I *think* it's `review`)

README.md
```diff
@@ -221,6 +224,17 @@ Run the integration test suite inside a containerized instance of Firefox. You m
...
+
+These credentials can be found or updated in your `.env` file.
```
lmorchard, line 236: I don't think this is actually true - or at least I don't see them in [the `.env.sample` file](https://github.com/mozilla/experimenter/blob/main/.env.sample#L28-L29). 

There's a `KINTO_USER` and `KINTO_PASS` setting - but I don't think that *defines* the account, rather it supplies the credentials for the experimenter Django app to use against the local Kinto instance. I was looking around to find a list where the Kinto accounts are defined but have yet to be able to find them

lmorchard: Oh wait, after stumbling around anew a bit more, I think I found where the accounts are defined:

https://github.com/mozilla/experimenter/blob/main/app/bin/setup_kinto.py#L7-L10

lmorchard: So, actually, *yes* `KINTO_USER` and `KINTO_PASS` from `.env` are used to define one of the accounts - `experimenter` / `experimenter` by default from `.env.sample`. The admin & review accounts are additional hardcoded accounts, though.

lmorchard: Might also be interesting to note that these users each [belong to different groups](https://github.com/mozilla/experimenter/blob/main/app/bin/setup_kinto.py#L48-L66) - though [it looks like both groups can write](https://github.com/mozilla/experimenter/blob/main/app/bin/setup_kinto.py#L74-L78), so I'm not sure if that's a distinction with a difference

lmorchard (changes requested):
Okay, an actual review: I think this would be good to merge with tweaks to mention all three accounts - `admin`, `review`, and `experimenter` - as well as a mention of where & how they're defined. Would be handy to also mention why there are 3 accounts and what is different about each, but I confess I don't know that myself.

jodyheavener:
>  Would be handy to also mention why there are 3 accounts and what is different about each

@jaredlockhart @k88hudson can you provide some insight into the differences between the accounts?

README.md
jodyheavener: Thanks for looking into this. I'll update to outline the differences as well as where they are hard-coded or in an env var.

jaredlockhart (approved):
Beautiful thank you @jodyheavener 🙏

jodyheavener:
@jaredlockhart thanks! If you have a sec it would be good to know a little about the 3 separate accounts for RS. e.g. when should one be used over the other?

jaredlockhart:
admin/admin: This account has permission to view and edit all of the collections.
experimenter/experimenter: This account is used by Experimenter to push its changes to Remote Settings and mark them for review.
review/review: This account should generally be used by developers testing the workflow, it can be used to approve/reject changes pushed from Experimenter.

Any change in remote settings requires two accounts: 
- One to make changes and request a review
- One to review and approve/reject those changes

Any of the accounts above can be used for any of those two roles, but your local Experimenter will be configured to make its changes through the experimenter/experimenter account so that account can't also be used to approve/reject those changes, hence the existence of the review/review account.

lmorchard (approved):

lmorchard:
chefkiss

PR #4595, opened 2021-02-10
feat(nimbus): remove limit on home experiment query

jodyheavener:
We're not going to hit any performance issues with the relatively small number of experiments we have, and we'll eventually be updating this page to sort and paginate them.


jaredlockhart (approved):
image

app/experimenter/nimbus-ui/src/gql/experiments.ts
```diff
@@ -194,7 +194,7 @@ export const GET_EXPERIMENT_QUERY = gql`
 // TODO: Only the first 50 until we add pagination.
```
pdehaan, line 194: nit: nuke the comment

jodyheavener:
#4598 will be the ticket to track this change.

PR #4596, opened 2021-02-10
fixes #4598 feat(nimbus): remove limit, offset, and status filter from experiments query entirely

### LINKED ISSUE #4598: Remove home page experiment query limit
This is a stop-gap until proper pagination is introduced. We currently limit homepage experiments to 50, and we've already bumped up against that, so removing this limit will allow users to see all experiments.

This includes a frontend and backend change.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-945)
┆Issue Number: EXP-945

jodyheavener:
Closes #4598

We [recently](https://github.com/mozilla/experimenter/pull/4595) removed the limit from the client, but now the default limit of 20 is kicking in. Will eventually be doing all sorting and paginating in the client anyway.

### DIFF
--- a/app/experimenter/experiments/api/v5/queries.py
+++ b/app/experimenter/experiments/api/v5/queries.py
@@ -6,7 +6,6 @@ from experimenter.experiments.api.v5.types import (
     NimbusLabelValueType,
     NimbusProbeSetType,
 )
-from experimenter.experiments.constants.nimbus import NimbusConstants
 from experimenter.experiments.models.nimbus import (
     NimbusExperiment,
     NimbusFeatureConfig,
@@ -71,11 +70,6 @@ class Query(graphene.ObjectType):
     experiments = graphene.List(
         NimbusExperimentType,
         description="List Nimbus Experiments.",
-        offset=graphene.Int(),
-        limit=graphene.Int(),
-        status=graphene.Enum(
-            "NimbusExperimentOptionalStatus", NimbusConstants.Status.choices
-        )(),
     )
     experiment_by_slug = graphene.Field(
         NimbusExperimentType,
@@ -88,11 +82,8 @@ class Query(graphene.ObjectType):
         description="Nimbus Configuration Data for front-end usage.",
     )
 
-    def resolve_experiments(root, info, offset=0, limit=20, status=None):
-        q = NimbusExperiment.objects.all()
-        if status:
-            q = q.filter(status=status)
-        return q[offset:limit]
+    def resolve_experiments(root, info):
+        return NimbusExperiment.objects.all()
 
     def resolve_experiment_by_slug(root, info, slug):
         try:
--- a/app/experimenter/nimbus-ui/schema.graphql
+++ b/app/experimenter/nimbus-ui/schema.graphql
@@ -137,14 +137,6 @@ enum NimbusExperimentFirefoxMinVersion {
   FIREFOX_100
 }
 
-enum NimbusExperimentOptionalStatus {
-  Draft
-  Review
-  Accepted
-  Live
-  Complete
-}
-
 type NimbusExperimentOwner {
   id: ID!
   username: String!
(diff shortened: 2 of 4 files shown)


app/experimenter/experiments/api/v5/queries.py
```diff
@@ -88,11 +86,11 @@ class Query(graphene.ObjectType):
         description="Nimbus Configuration Data for front-end usage.",
     )
 
-    def resolve_experiments(root, info, offset=0, limit=20, status=None):
+    def resolve_experiments(root, info, status=None):
         q = NimbusExperiment.objects.all()
         if status:
             q = q.filter(status=status)
-        return q[offset:limit]
+        return q
```
jaredlockhart, line 93: Looking at https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/gql/experiments.ts it looks like we also don't filter on status?

jodyheavener: Heh, you're right, client [does it](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx#L36:L36). Want me to 🪓 here?

jaredlockhart: rfixXhr

jaredlockhart: You know it would also be nice if we went through those GQL queries and nuked all the fields that are queried that aren't actually being used on the frontend, and potentially nuked them out of the backend if they're not being used at all.  LIke we're querying 'status' on almost every query but I'm not sure they're all used.  Doing an audit pass on that would be nice.

jodyheavener: That sounds like a very therepautic maintenance issue!

jaredlockhart: I *think* we might even be able to use the type system for it.  If we take the field out of the query, it'll be removed from the generated typestubs, which if they're being access will be cause by tsc:lint, I think?

jaredlockhart: Oh yeah it totally works.

pdehaan:
Semi related, but we might want to remove this line now that we've removed pagination support in https://github.com/mozilla/experimenter/pull/4595/files#r574095777:
https://github.com/mozilla/experimenter/blob/ab5252192f6a68fa3feae2444f23b97aae427005/app/experimenter/nimbus-ui/src/gql/experiments.ts#L194

app/experimenter/experiments/api/v5/queries.py
k88hudson: just wanted to say i love the gif

app/experimenter/experiments/api/v5/queries.py
```diff
@@ -88,11 +86,8 @@ class Query(graphene.ObjectType):
         description="Nimbus Configuration Data for front-end usage.",
     )
 
-    def resolve_experiments(root, info, offset=0, limit=20, status=None):
-        q = NimbusExperiment.objects.all()
-        if status:
-            q = q.filter(status=status)
-        return q[offset:limit]
+    def resolve_experiments(root, info, status=None):
```
pdehaan, line 89: What is `status` for here? It looks unused now.

app/experimenter/nimbus-ui/schema.graphql
```diff
@@ -278,7 +278,7 @@ type ProjectType {
 }
 
 type Query {
-  experiments(offset: Int, limit: Int, status: NimbusExperimentOptionalStatus): [NimbusExperimentType]
+  experiments(status: NimbusExperimentOptionalStatus): [NimbusExperimentType]
```
pdehaan, line 281: Possibly related to previous comment above, but not sure that we're using the `status` attribute anymore.

jaredlockhart (approved):
Cupcake Leaderboard 🧁 🏆 :
@jodyheavener +3 −49

PR #4599, opened 2021-02-11
fix #4342: feature(visualization) - Incorporate metric metadata from Jetstream.

### LINKED ISSUE #4342: Incorporate metadata for metrics into tooltips
Jetstream now exports metadata such as friendly name and description into buckets. These should be pulled in and incorporated into the dashboard tooltips.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-788)
┆Issue Number: EXP-788

emtwo:
Because:
* We want to have human-friendly metric names and detailed descriptinos

This commit:
* Pulls in that data from Jetstream and uses it for labels and tooltips.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
@@ -8,7 +8,7 @@ import Scrollspy from "react-scrollspy";
 import { ReactComponent as ChevronLeft } from "../../images/chevron-left.svg";
 import { ReactComponent as Clipboard } from "../../images/clipboard.svg";
 import { StatusCheck } from "../../lib/experiment";
-import { AnalysisData } from "../../lib/visualization/types";
+import { AnalysisData, MetadataPoint } from "../../lib/visualization/types";
 import { analysisAvailable } from "../../lib/visualization/utils";
 import {
   getExperiment_experimentBySlug_primaryProbeSets,
@@ -56,6 +56,19 @@ const probesetToMapping = (
   return newMap;
 };
 
+const otherMetricsToFriendlyName = (
+  otherMetrics: { [metric: string]: string },
+  metricsMetaData: { [metric: string]: MetadataPoint },
+) => {
+  const newMap: { [key: string]: string } = {};
+  Object.keys(otherMetrics).map(
+    (metric) =>
+      (newMap[metric] =
+        metricsMetaData[metric]!.friendly_name || otherMetrics[metric]),
+  );
+  return newMap;
+};
+
 type AppLayoutSidebarLockedProps = {
   testid?: string;
   children: React.ReactNode;
@@ -84,6 +97,11 @@ export const AppLayoutSidebarLocked = ({
   const { slug } = useParams();
   const primaryMetrics = probesetToMapping(primaryProbeSets || []);
   const secondaryMetrics = probesetToMapping(secondaryProbeSets || []);
+  const otherMetrics = otherMetricsToFriendlyName(
+    analysis?.other_metrics || {},
+    analysis?.metadata?.metrics || {},
+  );
+
   const sidebarKeys = [
     "monitoring",
     "overview",
@@ -94,7 +112,7 @@ export const AppLayoutSidebarLocked = ({
     .concat("secondary-metrics")
     .concat(Object.keys(secondaryMetrics || []))
     .concat("default-metrics")
-    .concat(Object.keys(analysis?.other_metrics || []));
+    .concat(Object.keys(otherMetrics || []));
 
   return (
     <Container fluid className="h-100vh" data-testid={testid}>
@@ -160,11 +178,8 @@ export const AppLayoutSidebarLocked = ({
                       getSidebarItems(primaryMetrics, "Primary Metrics")}
                     {Object.keys(secondaryMetrics).length &&
                       getSidebarItems(secondaryMetrics, "Secondary Metrics")}
-                    {analysis?.other_metrics &&
-                      getSidebarItems(
-                        analysis?.other_metrics,
-                        "Default Metrics",
-                      )}
+                    {otherMetrics &&
+                      getSidebarItems(otherMetrics, "Default Metrics")}
                   </Scrollspy>
                 </>
               ) : (
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.tsx
@@ -3,6 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from "react";
+import ReactMarkdown from "react-markdown";
+import ReactTooltip from "react-tooltip";
+import { ReactComponent as Info } from "../../../images/info.svg";
 import {
   DISPLAY_TYPE,
   METRIC_TYPE,
@@ -23,7 +26,7 @@ type SecondaryMetricStatistic = {
 type TableMetricSecondaryProps = {
   results: AnalysisData;
   probeSetSlug: string;
-  probeSetName: string;
+  probeSetDefaultName: string;
   isDefault?: boolean;
 };
 
@@ -44,10 +47,11 @@ const TableMetricSecondary = ({
     daily: [],
     weekly: {},
     overall: {},
+    metadata: { metrics: {}, probesets: {} },
     show_analysis: false,
   },
   probeSetSlug,
-  probeSetName,
+  probeSetDefaultName,
   isDefault = true,
 }: TableMetricSecondaryProps) => {
   const secondaryMetricStatistics = getStatistics(probeSetSlug);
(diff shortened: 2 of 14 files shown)


app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.tsx
```diff
@@ -56,11 +61,33 @@ const TableMetricSecondary = ({
...
 
   return (
     <div data-testid="table-metric-secondary" className="mb-5">
       <h2 className="h5 mb-3" id={probeSetSlug}>
-        <div>{probeSetName}</div>
+        <div>
+          <div className="d-inline-block">
+            {probeSetName}{" "}
+            {probeSetDescription && (
+              <OverlayTrigger overlay={popover} trigger="click" rootClose>
```
LZoog, line 85: I might be missing something since I'm doing a quick end-of-day review, but I checked out the docs for this to get an idea of what I should be seeing, and I'm wondering why we wouldn't use the `react-tooltip` for consistency but with `event="click"` if we specifically want it to show onclick (if I'm [reading their docs](https://www.npmjs.com/package/react-tooltip#options) correctly)? Maybe we have to or want to use `ReactMarkdown` and it's difficult to get that working with a tooltip?

Also, I think Storybook might need some love, I think we should be seeing that pop up here but when I hover/click the info icon I'm not seeing anything.

image

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -56,6 +56,19 @@ const probesetToMapping = (
   return newMap;
 };
 
+const otherMetricsToMapping = (
+  otherMetrics: { [metric: string]: string },
+  metricsMetdata: { [metric: string]: MetadataPoint },
+) => {
```
LZoog, line 62: Nit: can we name this something like otherMetricsToFriendlyName to make it more clear what it's mapping? I realize you're following the pattern already set with probesetToMapping but I think that could be more clear too (outside scope here). 🤔

Also, I think `metricsMetdata` should be `metricsMetaData`.

LZoog (commented):
LGTM from a quick pass but I'm wondering about the overlay bit. :)

Also, I feel like we could use a doc update somewhere regarding this if it affects accessing visualization data locally? Maybe it doesn't? I don't think we mention Jetstream at all. I'll let you make the call and think that's fine in a quick follow up ticket too if you think we need more than a sentence or two.

emtwo:
> LGTM from a quick pass but I'm wondering about the overlay bit. :) (will r+ after... but I should probably quickly check into your branch and run this locally in the morning to double check, heh)

It's true the overlay shows up on the top, I'm adding a screenshot here:

Screen Shot 2021-02-11 at 10 56 10 PM

I believe it doesn't show up on storybook because of the fact that it's up top. I can probably update it to adjust the popup location dynamically so that it shows up in storybook properly.

> Also, I feel like we could use a doc update somewhere since I don't think we mention/explain the interaction with Jetstream at all, but I'll let you make that call and think that's fine in a quick follow up ticket too if you think we need more than a sentence or two.

I agree this could use some docs. I'll file a follow-up since I do think it would be nice to have a bit more detail about what the relationship with Jetstream looks like.

> I'm wondering why we wouldn't use the react-tooltip for consistency but with event="click" if we specifically want it to show onclick (if I'm reading their docs correctly)? Maybe we have to or want to use ReactMarkdown and it's difficult to get that working with a tooltip?

You're right, I decided to use the popover because the text sent from Jetstream is in markdown and it has clickable links so it can't disappear when you hover away. I had assumed tooltips are not clickable but now that I look at the documentation it seems like it might be possible.

I'm PTO tomorrow but I can look into those things when I'm back.

app/experimenter/visualization/api/v3/views.py
```diff
@@ -216,7 +218,15 @@ def generate_results_object(data, experiment, window="overall"):
...
 def get_data(slug, window):
     filename = f"""statistics_{slug}_{window}.json"""
-    data = load_data_from_gcs(filename)
+    path = os.path.join(STATISTICS_FOLDER, filename)
+    data = load_data_from_gcs(path)
+    return data
+
+
+def get_metadata(slug):
+    filename = f"""metadata_{slug}.json"""
```
jaredlockhart, line 227: Does this need to be triple quoted?

jaredlockhart: Oh also noticed the other one above is too

app/experimenter/visualization/api/v3/views.py
```diff
@@ -216,7 +218,15 @@ def generate_results_object(data, experiment, window="overall"):
 
 def get_data(slug, window):
     filename = f"""statistics_{slug}_{window}.json"""
-    data = load_data_from_gcs(filename)
+    path = os.path.join(STATISTICS_FOLDER, filename)
+    data = load_data_from_gcs(path)
+    return data
```
jaredlockhart, line 223: This could be one line

app/experimenter/visualization/api/v3/views.py
```diff
@@ -216,7 +218,15 @@ def generate_results_object(data, experiment, window="overall"):
...
+
+
+def get_metadata(slug):
+    filename = f"""metadata_{slug}.json"""
+    path = os.path.join(METADATA_FOLDER, filename)
+    data = load_data_from_gcs(path)
     return data
```
jaredlockhart, line 230: This could be one line

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -125,14 +126,23 @@ def test_analysis_results_view_data(self, status, mock_exists, mock_open):
...
             "show_analysis": False,
         }
 
         class File:
+            def __init__(self, filename):
+                self.name = filename
+
             def read(self):
+                if "metadata" in self.name:
+                    return "{}"
```
jaredlockhart, line 139: Hrmm yeah, in theory that'll work for tests just gotta make sure we don't accidentally set that off with a misnamed non metadata file.

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -125,14 +126,23 @@ def test_analysis_results_view_data(self, status, mock_exists, mock_open):
...
                 return json.dumps(DATA_WITHOUT_POPULATION_PERCENTAGE)
 
-        mock_open.return_value = File()
+        def open_file(filename):
+            return File(filename)
```
jaredlockhart, line 143: Oh you know if we're implementing this interface ourself here we can explicitly pass in a parameter to our `File` class that indicates if it's a metadata file and return `{}` isntead of checking the filepath?

jaredlockhart (approved):
Backend looks good to me, couple ideas for some improvements but nothing critical, I'll let others review the frontend in more detail.  G/j @emtwo ! 🎉

app/experimenter/visualization/tests/api/test_views.py
emtwo: I'm not sure what you mean. I think in order to pass in a param to `File` to indicate if it's a metadata file, we would need to know if it's a metadata file. And the only way to know is by checking `if "metadata" in filename` which would just move the check from `read()` to `open_file()`.

app/experimenter/visualization/api/v3/views.py
emtwo: no it doesn't lol good point

emtwo:
@LZoog I managed to get the markdown working with `ReactTooltip`. Storybook looks good in the `basic` view for the results page. You can look at the tooltips there.

Unfortunately, I wasn't able to address the storybook issue that you screenshotted about. I think something about the way storybook renders the component, it doesn't recognize that it's hovering out of view. I couldn't find a way to make this work unfortunately.

Perhaps we can just let it be since it works fine in the `basic` view of storybook? What do you think?

app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.tsx
```diff
@@ -56,11 +60,39 @@ const TableMetricSecondary = ({
...
+              <>
+                <Info
+                  data-tip
+                  data-for={probeSetSlug}
+                  className="align-baseline"
+                />
+                <ReactTooltip
+                  id={probeSetSlug}
+                  multiline={true}
+                  clickable={true}
```
jodyheavener, line 85: This doesn't appear to actually be clickable from what I can [see](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/78ee0b46aac30477c43555b21a73a47ac5d07ad4/nimbus-ui/index.html?path=/story/pages-results--basic)? That's fine with me, I prefer it to be consistent with the other tooltips, but it looks like we're trying to make this clickable and it isn't.

jodyheavener: Also props that are just `true` should be able to simply be just the prop, so `multiline clickable`

jodyheavener (commented):
Left a comment on the clickability and prop declaration, and my only other feedback would be that there appears to be extra space at the bottom of the markdown tooltip:

Screen Shot 2021-02-18 at 10 14 34 AM

This appears to be because it renders <p> tags, where the regular tooltip doesn't. I think we could solve this by adding something like this to the global styles:

```css
.__react_component_tooltip > p:last-child {
  margin-bottom: 0;
}
```

app/experimenter/nimbus-ui/src/components/PageResults/TableMetricSecondary/index.tsx
emtwo: I've updated it to remove `{true}` and I added a link so it's more clear that it's clickable.

emtwo:
> ```css
> .__react_component_tooltip > p:last-child {
>   margin-bottom: 0;
> }
> ```

Thanks! I added this it looks nicer!

emtwo:
@jodyheavener let me know what you think with the changes!

jodyheavener (approved):
We might want to style links in the tooltip to be lighter/white with underline since that blue over grey might be less visible, but I totally happy pushing that to a followup. Looks great!

emtwo:
> We might want to style links in the tooltip to be lighter/white with underline since that blue over grey might be less visible, but I totally happy pushing that to a followup. Looks great!

Thanks! That's a good point. I think if we want all tooltips to look consistent they might all need to be the same colour so that would need to be part of the follow-up to make sure it works across the board. I'll file a ticket

Issue #4485, opened 2021-01-27
Nested form submit errors not clearing onchange


data-sync-user:
Noticed while working on EXP-771 / #4299 that this was happening. On `Branches` (and potentially around Additional Links when that work is merged), submit errors don't clear on change.

I have a solution WIP, but it's not as trivial as I was hoping due to the indexing/multiple treatment branches, so I think it should be done in a separate issue.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-876)
┆Issue Number: EXP-876

LZoog:
Had a comment on "potentially" affecting Additional Links but it definitely does, per https://github.com/mozilla/experimenter/pull/4480#pullrequestreview-578825888 / https://github.com/mozilla/experimenter/pull/4480#pullrequestreview-579365330. Bumping this up from a 2 to a 3, I think the solution for Branches might look slightly different than Additional Links

data-sync-user:
➤ Lauren Zugai commented:

I'd like to finish this now, but I'm bumping this out of the sprint as well as out of Palm due to much higher priority tasks taking precedence. Leaving myself some notes...

I still have a WIP branch, but I'm running into a rendering issue in Branches/Branch after resetting submit errors where the submit error is clearing, but because `errors` are passed into each `FormBranch`, `FormBranch` is rerendering and `formControlAttrs` in `useCommonNestedForm`/`useCommonForm` is getting hit a bunch of times and it resets the `defaultValue` which means any time it clears, the default value is displayed in the input. I've tried a couple things without luck so far (mostly trying to `useEffect` or `useLayoutEffect` in one of the hooks) but haven't gotten it working + my resetting error approach could use a little cleanup.

Also, I'm not seeing submit errors getting displayed under Additional Links. We don't appear to be using the same URL-check server-side as we are on risk mitigation link and when I take out the client-side validation I can't seem to coerce submit errors to show, but it's making me scratch my head since I left a comment in https://github.com/mozilla/experimenter/pull/4480#pullrequestreview-578825888 which appeared to show them. Unless I'm mistaken might need to stick some server-side validation there (might need a separate issue filed), but the solution here should fix any onChange clearing.

PR #4601, opened 2021-02-11
fixes #4460 chore(docs): Add useCommonForm docs, improve validation section

### LINKED ISSUE #4460: Add some common form hook docs
With #4306 / #4420, there's a lot going on with these hooks. We should have some docs on how to use them.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-859)
┆Issue Number: EXP-859

LZoog:
Because:
* We were missing docs around our forms.

This commit:
* Adds useCommonForm docs and provides more details on validation errors
* Moves the GQL error handling section under Error Handling
* Adds an alernative note in the Development section

fixes #4460

P.S. Here's a handy preview link: https://github.com/mozilla/experimenter/blob/issues/4460/app/experimenter/nimbus-ui/README.md

### DIFF
--- a/app/experimenter/nimbus-ui/README.md
+++ b/app/experimenter/nimbus-ui/README.md
@@ -2,7 +2,7 @@
 
 ## Development
 
-- `yarn start` to start the app in development mode on `http://localhost:3000`
+- `yarn start` to start the app in development mode on `http://localhost:3000` (alternatively, run `make up` at the project root and navigate to `localhost/nimbus`)
 - `yarn build` to create a production build
 - `yarn test` to run unit tests, coverage details are displayed by default
 
@@ -87,6 +87,8 @@ export interface GetExperimentOverviews {
 
 ## Error handling
 
+Note: see the "Forms and Validation" section for form-specific error handling.
+
 This app has an [`AppErrorBoundary`](./src/components/AppErrorBoundary/index.tsx) that will capture any uncaught error that occurs and report them to Sentry. This acts as our last line of defense, but ideally we are able to handle errors before they get to this stage. As well, errors that we wish to display to the user should generally be handled in a consistent fashion.
 
 ### General errors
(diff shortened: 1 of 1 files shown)


app/experimenter/nimbus-ui/README.md
```diff
@@ -197,6 +166,69 @@ if (message !== "success") {
...
+  <Form.Label>Public name</Form.Label>
+  <Form.Control
+    {...formControlAttrs("name")}
+    type="text"
+  />
+  <FormErrors name="name" />
+</Form.Group>
+```
+
+Form fields are optional by default - that is, users can save pieces of the experiment without filling in every input on the page. To enforce a required field or client-side validation on the field, pass a rule into the method:
```
jodyheavener, line 200: The predefined validators will likely see the most use, but I wonder if we should point out that this argument is actually a [`RegisterOptions`](https://react-hook-form.com/ts#ValidationRulesRef) object, which can take a `validate` function, a `pattern` for regex matching, and a few other handy properties like min, max, etc. In case they want to make a new predefined one, or they need to do something fancy in a specific field.

jodyheavener (approved):
This is well written and generally matches how I would approach adding form controls. Nicely done. Left you one thought, but I wouldn't block on it.

app/experimenter/nimbus-ui/README.md
```diff
@@ -197,6 +166,69 @@ if (message !== "success") {
...
+
+**See the form validation section of [Storybook](https://storage.googleapis.com/mozilla-storybooks-experimenter/index.html) for examples and a visual guide to the UX interaction.**
+
+There are three kinds of validation that need to be applied to forms: client-side, server-side, and "required for launch".
+
+### `useCommonForm` and `useCommonNestedForm`
+
+All forms use [`react-bootstrap`](https://react-bootstrap.github.io/components/forms/) as well as the custom convenience hooks `useCommonForm` or `useCommonNestedForm` for ease of use when developing new forms, to promote consistent behavior amongst our forms, and to limit any future changes affecting all forms in limited places. These hooks utilize the [`react-hook-form`](https://react-hook-form.com/api) package.
+
+If there are no nested forms, pass the needed parameters into `useCommonForm`. Otherwise, pull form methods out of _our_ `useForm` hook and pass them into `FormContext`. Then, in the nested form(s), pull out the methods and pass them into `useCommonNestedForm`.
```
k88hudson, line 179: A link to the source code would be nice from `needed parameters`

app/experimenter/nimbus-ui/README.md
```diff
@@ -197,6 +166,69 @@ if (message !== "success") {
...
+      {...formSelectAttrs("multi")}
+      options={arrayOfStringOptions}
+    />
+  <FormErrors name="multi" />
+</Form.Group>
+```
+
+### Required for Launch Validation
+
+Some fields like `publicDescription` are allowed to be `null` while someone is editing a draft, but must be set before the experiment can move to the `review` status. You should NOT mark these fields `required`, but rather add help text that indicates they must be filled out before launch.
```
k88hudson, line 230: :+1:

k88hudson (approved):
I like it! Could use an inline link to the code maybe, but other that that LGTM

app/experimenter/nimbus-ui/README.md
LZoog: I'll mention something similar and link to the docs, good call out 👍

app/experimenter/nimbus-ui/README.md
LZoog: I'm a little hesitant since I think it's very possible that it'll become out of date later, we have a TODO a the top of the file and we'll also need some edits in that file when I have time to address #4485 😬

PR #4606, opened 2021-02-12
fix #4422 featu(nimbus): automatically pause experiments after enrollment end date

### LINKED ISSUE #4422: Automated Pause Tasks
We need to add two new celery tasks, and some model methods.

On NimbusExperiment add methods for

* proposed_enrollment_end (start_date + proposed_enrollment)
* should_pause (today >= proposed_enrollment_end)

Then add two celery tasks

The first one will check for experiments that are

* status is Live
* is_paused is False
* should_pause is True

Re serialize the experiment with is_paused as True, but don't update it in the db yet.  Push the new serialization to remote settings and mark the collection for review.

A second celery task that will check for experiments that are currently live, read their record out of the remote settings collection, and if is_paused in the collection and in the db are different, update the db to reflect the value in the collection and generate a change log.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-828)
┆Issue Number: EXP-828

jaredlockhart:
Because

* We want experimenter to automatically set the isEnrollmentPaused flag in remote settings automatically after the enrollment end date passes

This commit

* Adds logic to the nimbus model to compute when an experiment should be paused
* Adds a celery task to update the record in remote settings after the pause date has passed
* Adds a celery task to update the experiment in the db when remote settings has accepted the pause change and generate a changelog

### DIFF
--- a/app/bin/setup_kinto.py
+++ b/app/bin/setup_kinto.py
@@ -9,7 +9,7 @@ REVIEW_USER = REVIEW_PASS = "review"
 EXPERIMENTER_USER = os.environ["KINTO_USER"]
 EXPERIMENTER_PASS = os.environ["KINTO_PASS"]
 KINTO_HOST = os.environ["KINTO_HOST"]
-KINTO_BUCKET = "main-workspace"
+KINTO_BUCKET_WORKSPACE = "main-workspace"
 KINTO_BUCKET_MAIN = "main"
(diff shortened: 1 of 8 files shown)


lmorchard (approved):
Makes sense to me! The pre-existing logic with `start_date` being derived from changelog confused me for a minute, but that makes sense too now that I've seen it

PR #4616, opened 2021-02-12
fix #4612 chore(nimbus): remove unused gql fields

### LINKED ISSUE #4612: Delete unused gql query parameters 🧁
Many of the parameters in our GQL queries are not used by the consumer, which means it's hard to tell by looking at each query which fields are important.  We should delete any unused parameters.  The good news is TS will enforce whether they're being used or not.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-954)
┆Issue Number: EXP-954

jaredlockhart:
Because

* We designed the GQL API with capabilities that are not currently being employed including
  * An unused status field indicating success which mirrors the "success" value in message
  * NimbusExperiment fields queried on mutations which are not used by the client
  * Multiple update experiment mutations which, without unused fields, are identical
  * Database IDs for probesets which are now unused after switching to using slugs

This commit

* Removes unused fields from client queries
* Removes status from all API calls
* Folds all mutations into a single query
* Updates relevant tests and types


app/experimenter/nimbus-ui/src/gql/experiments.ts
```diff
@@ -8,82 +8,17 @@ export const CREATE_EXPERIMENT_MUTATION = gql`
...
-  mutation updateExperimentProbeSets($input: ExperimentInput!) {
-    updateExperiment(input: $input) {
-      nimbusExperiment {
-        id
-        primaryProbeSets {
-          slug
-        }
-        secondaryProbeSets {
-          slug
-        }
```
jodyheavener, line 83: Hrmm, I see why you would combine these, but I would argue the general idea with GQL is you only request what you want back in a given query. I don't think having every field results in additional DB work, but the response will always be the largest possible response. In an app like this it's probably negligible, but probably a good practice to keep.

jodyheavener: Wait we aren't using any of the returned fields? Nvm ignore me

jodyheavener: Oh yeah we do a full refresh after each save anyway.

jaredlockhart: Yep

jodyheavener: I'm trying to think of any reason why we should keep these separate, but coming up empty. Are they just retrofitted from when we had individual mutations?

jaredlockhart: I think so.

jodyheavener (approved):
🧁

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -16,143 +16,27 @@
...
     }
 }
 """
 
 
-UPDATE_EXPERIMENT_OVERVIEW_MUTATION = """\
+UPDATE_EXPERIMENT_MUTATION = """\
 mutation($input: ExperimentInput!) {
     updateExperiment(input: $input) {
-        nimbusExperiment {
```
lmorchard, line 33: Don't we use the `nimbusExperiment` returned in mutation response to update the local cache with experiment changes on the client side?

jaredlockhart: Didn't seem to in my local testing, can you link to the code that uses it?

jodyheavener: FXA does this, but IIRC we decided early on that we wanted the freshest experiment data and opted not to use the client cache at all. I'd be down for revisiting that, though.

jodyheavener: Yeah, [no cache](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/hooks/useExperiment.tsx#L49:L49)

jaredlockhart: Yeah looks to me like it just refetches the whole experiment by slug on each page load and after each save:

Screen Shot 2021-02-16 at 1 44 02 PM

lmorchard: This is mainly what my fuzzy recollection was pointing me to: https://www.apollographql.com/docs/react/data/mutations/#updating-a-single-existing-entity

But if we've got that disabled, then I guess yeah we're not doing client-side caching like that

jodyheavener: FWIW I would love to do use that and cut back on requests. I can't remember what the exact reason was for not using any cache, maybe just because something could change it externally, but I doubt it would cause problems.

app/experimenter/nimbus-ui/src/types/updateExperimentStatus.ts
```diff
@@ -3,22 +3,15 @@
...
-export interface updateExperimentStatus_updateExperiment_nimbusExperiment {
-  __typename: "NimbusExperimentType";
-  status: NimbusExperimentStatus | null;
-}
-
 export interface updateExperimentStatus_updateExperiment {
   __typename: "UpdateExperiment";
   message: ObjectField | null;
-  status: number | null;
-  nimbusExperiment: updateExperimentStatus_updateExperiment_nimbusExperiment | null;
```
LZoog, line 21: I like the combination into a single mutation, but these type files seem to basically do... nothing for us now? They're all:

```
 __typename: "UpdateExperiment";
  message: ObjectField | null;
```

I liked being able to see what I was "allowed" to change per-page, but guess we can always go look at the fields in `UPDATE_EXPERIMENT_MUTATION` and infer it. I know they're auto-generated but it seems like in this case, we don't even need `updateExperimentStatus`, `updateExperimentBranches`, `updateExperimentAudience` etc.?

app/experimenter/experiments/tests/api/v5/test_mutations.py
LZoog: Yeah, we opted not to cache anything at all and to always refresh it per page in order to make sure the freshest experiment data was reflected. Not the most ideal, but I wrote a sorta mini-ADR in #3471 which is related, though I haven't read through this since then.

app/experimenter/nimbus-ui/src/types/updateExperimentStatus.ts
jodyheavener: Oh yeah you could just delete those and change their imports to `updateExperiment_updateExperiment`. `generate-types` doesn't seem to regenerate them.

jaredlockhart: Yeah I think it's a combination of not using hte return values and combining everything into a single mutation.  You're 'allowed' to update anything.  I think a lot of this was built around a set of expectations/designs that didn't actually end up being used, so right now it's a lot of unused code.  This doesn't stop us from changing the way it works, but I think that might be easier if it's clear which parts are being used instead of working around code that's built but doesn't actually do anything?

LZoog: Yeah, I guess by "allowed" I meant it was nice to refer to those types as a sort of guide on what I _should_ be updating on that page.

I'm cool with it though, but maybe as @jodyheavener suggested we can remove even more code and just use `updateExperiment_updateExperiment`? 🧁

jaredlockhart: Yeah it's kind of misleading, cuz the explicit fields are fields being READ not UPDATED, so it's not exactly a good reference for what the page is updating.

jodyheavener: I'm actually a little surprised `generate-types` didn't remove them for us 🤷

jaredlockhart: What we could do is make type declarations for the set of fields each page updates and then type the inputs in the actual components with those?

jaredlockhart: Like we kinda do that here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx#L52-L60

but we could pull that out into a type declaration somewhere else to say that's the input type for that page form so we have it explicitly declared somewhere that those are the fields updated by that form?

jodyheavener: That's certainly doable, but isn't that effectively re-adding what we're taking away here, and then adding the onus of us maintaining them in two spots? With the GQL query the types were automatic 😆

jaredlockhart: No that's what I mean, the auto generated ones this is removing show fields being READ, not WRITTEN.  So you could add or remove arbitrary fields to/from that list without affecting the actual behaviour, it's misleading to refer to that as the list of fields a page updates.  Right?

jodyheavener: Oh sorry I was actually referring to the input types, but it looks like those were yanked even before this PR and are all `ExperimentInput` now. Yeah I see what you're saying, might call it a followup though.

LZoog: @jaredlockhart Yeah I like that idea. It's cool in a follow up to me though (would prefer it be a high priority maintenance item). PR seems good to me. 😎

jaredlockhart: Followup here

https://github.com/mozilla/experimenter/issues/4631

jaredlockhart: Okay I removed those extra result types and updated the references, good catch @jodyheavener

PR #4617, opened 2021-02-12
chore(nimbus): eliminate all instance of __typename

jodyheavener:
We don't use `__typename` anywhere in our code, and we're kind of just carrying it around everywhere. We can just drop it from the generated types and our `MockedProvider` can add them automatically anyway.

🧁

### DIFF
--- a/app/experimenter/nimbus-ui/README.md
+++ b/app/experimenter/nimbus-ui/README.md
@@ -71,7 +71,6 @@ Would be analyzed and the corresponding types generated would look like this:
 
 ```ts
 export interface GetExperimentOverviews_experiments {
-  __typename: "NimbusExperimentType";
   name: string;
   slug: string;
   hypothesis: string | null;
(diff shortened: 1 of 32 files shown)


app/experimenter/nimbus-ui/README.md
```diff
@@ -362,7 +361,7 @@ const mkSimulatedQueries = ({
 const Subject = ({ simulatedQueries = mkSimulatedQueries() }) => {
   const mockLink = new SimulatedMockLink(simulatedQueries, false);
   return (
-    <MockedProvider link={mockLink} addTypename={false}>
```
LZoog, line 365: Oh I thought we needed this in order to drop `__typename`! TIL

jodyheavener: The default is `__typename` gets added by `MockedProvider`, but since we were manually defining them everywhere we added `addTypename={false}` to prevent doubling up.

jaredlockhart (commented):
Oh sick I always wondered what that was for, I like this 👍

jaredlockhart (approved):
Bye bye `__typename`, we knew you too little 👋

PR #4626, opened 2021-02-16
fixes #4022 feat(nimbus): make next button save before continuing

### LINKED ISSUE #4022: Clicking "Next" in nimbus wizard doesn't save changes
1. Go to https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/nimbus/new and create a new experiment.
1. On the `/${slug}/edit/overview/` page, change the Public name and possibly fill in the hypothesis with a short novella.
1. Click the "Next" button.

Note that the changes you just made on the overview page were not saved when clicking "Next". In the case of an experiment name, this is mildly annoying. In the case of me filling in a lengthy hypothesis only to have my changes lost, it'd be pretty frustrating data loss.

Making a user click Save _then_ Next each time feels counterintuitive.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-613)
┆Issue Number: EXP-613

jodyheavener:
Closes #4022

This PR modifies to "Next" button on all four experiment pages:

- Updates their label to "Save and Continue"
- Saves the experiment data before continuing to next page
  - Save must be valid (no client or server/validation errors

Couple notes:

- Although the button was changed from "Next" to "Save and Continue" I left most of the internal references as "next", since it does take you next and renaming everything to the much longer "save-and-continue-button" didn't seem worth it.
- The Audience page needs to prevent you from continuing to the Request Review page until all valid fields are filled out, and so the Save and Continue button is disabled until the server allows you to continue. But since that button also saves, you could in theory set an empty value, hit Save and Continue, the value would save, and you would be redirected to the Review page -- _however_, by the time you hit that page it'll know you have an invalid experiment and will just redirect you back to the Audience page.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
@@ -33,16 +33,6 @@ describe("FormOverview", () => {
     expect(onCancel).toHaveBeenCalled();
   });
 
-  it("calls onNext when next clicked", async () => {
-    const { experiment } = mockExperimentQuery("boo");
-    const onNext = jest.fn();
-    render(<Subject {...{ onNext, experiment }} />);
-
-    const nextButton = screen.getByText("Next");
-    await act(async () => void fireEvent.click(nextButton));
-    expect(onNext).toHaveBeenCalled();
-  });
-
   it("renders initial documentation links", async () => {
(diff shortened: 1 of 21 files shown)


lmorchard (approved):
One nit about `useCallback` and `useMemo` for event handler definition but not an absolute blocker to me

PR #4632, opened 2021-02-16
fixes #4628 feat(nimbus): Return targeting: "true" for experiments match all clients

### LINKED ISSUE #4628: Targeting should be "true" for experiments that should enroll any client
If we want to run an experiment that targets all clients, **targeting: "true"** is a lot more explicit than null or undefined.

This was a recommendation that came out of our targeting post mortem: https://docs.google.com/document/d/1eD5BWlm_1UqTG0c5RvnFue3fTb5J4MDWDRjb-4Gasn8/edit#

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-959)
┆Issue Number: EXP-959

k88hudson:
Because

* 'true' is more explicit than null or undefined  targeting

This commit

* Change the Nimbus model to return targeting: "true" for experiments match all clients

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -122,8 +122,9 @@ class NimbusExperiment(NimbusConstants, models.Model):
         if self.is_desktop_experiment:
             expressions.append("'app.shield.optoutstudies.enabled'|preferenceValue")
 
+        #  If there is no targeting defined all clients should match, so we return "true"
         if len(expressions) == 0:
-            return None
+            return "true"
 
         return " && ".join(expressions)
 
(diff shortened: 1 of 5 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -122,8 +122,9 @@ def targeting(self):
         if self.is_desktop_experiment:
             expressions.append("'app.shield.optoutstudies.enabled'|preferenceValue")
 
+        #  If there is no targeting defined all clients should match, so we return "true"
         if len(expressions) == 0:
-            return None
+            return "true"
```
pdehaan, line 127: Will this ultimately display the word "true" in the UI?

Re: https://experimenter.services.mozilla.com/nimbus/onoff-aboutwelcome-experiment/,
It looks like for each branch's `Enabled` field, we display "True" or "False" (initial uppercase character).

k88hudson: IMO it should show the exact JEXL, which should look like 
```
true
```

jaredlockhart (approved):
Very nice 👍

PR #4639, opened 2021-02-17
fix #4539 feat(nimbus): add enrollment data to v5 api

### LINKED ISSUE #4539: Add isEnrollmentPaused and enrollmentCompletedDate to experiment
_Epic description notes a new property on the experiment query called `isEnrollmentPaused`, but I don't see a ticket for it, so this is it!_

Experiment query needs to return a new property called `isEnrollmentPaused` so we can indicate this in the UI.

In addition to `isEnrollmentPaused` we also need a state for when enrollment ended -- this could be something like `enrollmentComplete`, but preferably `enrollmentCompletedDate` since we plan to use this date in the UI later.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-911)
┆Issue Number: EXP-911

jaredlockhart:
Because

* We need to be able to display the pause state and date in the UI

This commit

* Adds enrollment paused and paused date to V5 API

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -132,6 +132,8 @@ class NimbusExperimentType(DjangoObjectType):
     monitoring_dashboard_url = graphene.String()
     start_date = graphene.DateTime()
     end_date = graphene.DateTime()
+    is_enrollment_paused = graphene.Boolean()
+    enrollment_end_date = graphene.DateTime()
 
     class Meta:
         model = NimbusExperiment
@@ -157,3 +159,9 @@ class NimbusExperimentType(DjangoObjectType):
 
     def resolve_jexl_targeting_expression(self, info):
         return self.targeting
+
+    def resolve_is_enrollment_paused(self, info):
+        return self.is_paused
+
+    def resolve_enrollment_end_date(self, info):
+        return self.proposed_enrollment_end_date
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/api/v5/types.py
```diff
@@ -157,3 +159,9 @@ def resolve_ready_for_review(self, info):
 
     def resolve_jexl_targeting_expression(self, info):
         return self.targeting
+
+    def resolve_is_enrollment_paused(self, info):
+        return self.is_paused
+
+    def resolve_enrollment_end_date(self, info):
+        return self.proposed_enrollment_end_date
```
jodyheavener, line 167: Oh interesting, do we not want the column to distinguish when enrollment actually ended (versus when it was expected to end), or is it impossible enrollment could end early? If they're always one and the same we could even do the start_date + proposed_enrollment math in the client (though this field is still handy).

jodyheavener (approved):
Awesome. One question but carry on if it's moot.

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Yep very good point, I thought about that too.

We could look up the date it really ended by introspecting into the changelog to look for the date that `is_paused` flipped to True, but I think we should hold off on building out that logic until we have a strong case that the frequency of experiment enrollment end drifting from the proposed date is sufficient to warrant it.

Also yes in theory we could do that logic in the UI but I'd generally warn against doing that kind of logic on the frontend, particularly when there is an implementation of it on the backend.  We don't want to wind up in a situation where we have that logic duplicated, change it in one place, and then forget to do it somewhere else.

PR #4643, opened 2021-02-17
fix #3947 feat(nimbus): synchronize preview experiments with preview kinto collection

### LINKED ISSUE #3947: Add QA collection and status and celery tasks
To facilitate QAing Nimbus experiments, let's add a special remote settings collection called `nimbus-qa` that requires no dual sign off, an experiment status called `QA`, and a celery task that checks for experiments in QA status and pushes them to the collection, and experiments in the collection that aren't in QA status and removes them.

Note: One thing I realized is in theory you can 'Launch to Preview' and 'Go back to draft' back and forth instantly and endlessly so we'll have to make sure we account for that in the RS sync mechanism so we don't accidentally get the backend stuck in some weird state

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-567)
┆Issue Number: EXP-567

jaredlockhart:
Because

* We want any experiments in the preview state to be immediately published to the preview kinto collection
* We want any experiments not in the preview state to be immediately unpublished from the preview kinto collection

This commit

* Adds a local nimbus-preview collection to the local dev remote settings and configures it to have reviews disabled
* Adds additional logic to our kinto client to support collections with reviewing disabled
* Adds a celery task that ensures only experiments in the preview state are published in the preview collection
* Adds that celery task to our celery schedule
* Explicitly invokes that celery task when an experiment is updated to the PREVIEW status

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -14,6 +14,7 @@ from experimenter.experiments.models.nimbus import (
     NimbusFeatureConfig,
     NimbusProbeSet,
 )
+from experimenter.kinto.tasks import nimbus_synchronize_preview_experiments_in_kinto
 
 
 class NimbusBranchSerializer(serializers.ModelSerializer):
@@ -382,6 +383,12 @@ class NimbusExperimentSerializer(
             "total_enrolled_clients",
         ]
 
+    def __init__(self, instance=None, data=None, **kwargs):
+        self.should_call_preview_task = (
+            data.get("status") == NimbusExperiment.Status.PREVIEW
+        )
+        super().__init__(instance=instance, data=data, **kwargs)
+
     def validate_name(self, name):
         if not (self.instance or name):
             raise serializers.ValidationError("Name is required to create an experiment")
@@ -427,6 +434,9 @@ class NimbusExperimentSerializer(
 
             if experiment.should_allocate_bucket_range:
                 experiment.allocate_bucket_range()
+
+            if self.should_call_preview_task:
+                nimbus_synchronize_preview_experiments_in_kinto.delay()
 
             generate_nimbus_changelog(experiment, self.context["user"])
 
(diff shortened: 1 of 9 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -382,6 +383,12 @@ class Meta:
             "total_enrolled_clients",
         ]
 
+    def __init__(self, instance=None, data=None, **kwargs):
+        self.should_call_preview_task = (
```
lmorchard, line 387: Should this maybe also include a check of the current status to ensure this is a *transition*?

e.g. `instance.status != data.get("status") and data.get("status") == NimbusExperiment.Status.PREVIEW`

Probably not practically possible for this to happen, but seems like this should be made idempotent.

app/experimenter/kinto/client.py
```diff
@@ -4,28 +4,40 @@
...
+                bucket=settings.KINTO_BUCKET_WORKSPACE,
+            )
+        else:
+            self.kinto_http_client.patch_collection(
+                id=self.collection,
+                data={"status": KINTO_SIGN_STATUS},
+                bucket=settings.KINTO_BUCKET_WORKSPACE,
+            )
+
+    def create_record(self, data):
```
lmorchard, line 33: Yeah, this seems like a tidy little refactor

lmorchard (approved):
:+1: One suggestion on the `should_call_preview_task` logic, but likely a non-issue

app/experimenter/kinto/tasks.py
```diff
@@ -35,26 +35,26 @@ def nimbus_push_experiment_to_kinto(experiment_id):
 
     try:
         experiment = NimbusExperiment.objects.get(id=experiment_id)
-        logger.info(f"Pushing {experiment} to Kinto")
+        logger.info(f"Pushing {experiment.slug} to Kinto")
```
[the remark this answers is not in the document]
jaredlockhart: Oh yeah I just realized since remote settings uses the slug as its identifier it's easier to look at the logs with slugs and compare them to what the api is saying instead of the name.

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Yeah it's not a bad point but I think in general `status` will only be provided in `data` when it's being set, so if the status is something other than preview and `status` is in `data`, that's the case we want, and if for some reason we're setting it to `preview` and it's already `preview`, that's such an edge case it's not worth building around.  Additionally the sync task itself is idempotent so calling it again doesn't hurt anything.

Issue #4587, opened 2021-02-09
Review and Preview states sound awfully close


pdehaan:
I know we kicked around some ideas for the new state name, **Preview**, and this has come up before; but I couldn't find an existing issue for it.

I think the current flow is that you start in:

1. **Draft:** Filling in all the bits and pieces.
2. **Preview:** Send it off to a QA collection for verification and testing.
3. **Review:** Passed QA testing, send it off to RS to be approved/rejected by admin.
4. **Live:** Send our lil' experiment off to the real world.
5. **Complete:** Experiment has ended and we have results.

I can't think of a better name for "Preview" or "Review" to help disambiguate.
For **Preview** I think we considered "QA" or "Staging", but didn't feel like either of those were a 100% fit.

Not sure if we change **Review** to **Pending** or **Approval** or **Ready** (although the latter is a bit ambiguous as to what it's ready for... QA, Sign-off, Shipping?)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-938)
┆Issue Number: EXP-938

Issue #4577, opened 2021-02-08
Combine Summary and Design components, routes


data-sync-user:
For each experiment Nimbus currently has a Summary page (`/:slug`) as well as a Design page (`/:slug/design`). The only visual distinction between the two is that one page [uses a sidebar](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageDesign/index.tsx) and one [does not](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageSummary/index.tsx). In fact, they are so similar that they both just render the same common `<Summary />` component from their page container components. Let's combine these into a single page.

Specifically here's what needs to be done:

* Merge [<PageSummary />](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageSummary/index.tsx) and [<PageDesign />](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageDesign/index.tsx) into a single component (`PageSummary` is fine), and make it available at the root `/:slug`. The route and contents at `/:slug/design` can be removed entirely. In combining them the resulting component should use the sidebar layout.
* Update the sidebar to link to this page in both `<AppLayoutWithSidebar />` and `<AppLayoutSidebarLocked />` components. The link should be called "Summary" and use the paper airplane icon. Reference the [Figma designs here|#].

The combination of the two page components should also include:

* In the Design page polling was disabled, but in the Summary page it was enabled. Keep this enabled in the new page component, but ensure it remains disabled in Storybook.
* The Design page would redirect you back to `/:slug/edit/overview` if the experiment was was not locked (still editing). This should no longer redirect in the new page component as the Summary page should be accessible at any point in the experiment lifecycle.
* Update [editCommonRedirects](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/lib/experiment.ts#L49:L49) to redirect to the new Summary route instead of `/design`. Also double check that other redirects are working correctly.
* In [this Pull Request](https://github.com/mozilla/experimenter/pull/4530) we added a handy set of links to the summary page in the header. This entire PR is moot once these two components are merged; `<HeaderExperiment />` no longer needs the `summaryView` prop or any of the links introduced by the PR. The whole thing can be reverted.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-931)
┆Issue Number: EXP-931

PR #4658, opened 2021-02-18
fix #4496 fix #4495: polish and integration for Launch to Preview UI

### LINKED ISSUE #4495: Ensure transition from preview back to draft status operates correctly
From a comment on https://github.com/mozilla/experimenter/pull/4482:

{quote}One thing I realized is in theory you can 'Launch to Preview' and 'Go back to draft' back and forth instantly and endlessly so we'll have to make sure we account for that in the RS sync mechanism so we don't accidentally get the backend stuck in some weird state{quote}

Not sure if there's specific work to be done here, but filing this follow up issue to capture the concern.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-883)
┆Issue Number: EXP-883

### LINKED ISSUE #4496: Finish & polish roughed-out Launch to Preview UI components
EXP-866 introduced roughed-out UI components for the Launch to Preview feature to allow for review & iteration. Those are disabled behind a feature flag and make mocked-out API calls.

Polish up those components with required changes, flesh out test coverage, and get them making real GQL queries

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-885)
┆Issue Number: EXP-885

lmorchard:
Because:

- The UI for Launch to Preview needs polish and integration work

This commit:

- adds the preview status experiments to the home page directory

- wires up the status change forms in PageRequestReview to use real GQL
  mutations to update experiment status

- re-fetches the experiment on status change success to update state

- adds test coverage for the new status change forms added in EXP-866

- cleans up /* istanbul ignore */ and // @ts-ignore comments so that
  test and type coverage is now enforced for new status change forms

- cleans up feature flag and mock data usage for status change forms

- tweaks the GQL serializer to immediately invoke the
  nimbus_synchronize_preview_experiments_in_kinto task on save for both
  DRAFT -> PREVIEW and PREVIEW -> DRAFT status transitions

- ensures the nimbus_synchronize_preview_experiments_in_kinto method is
  annotated as a Celery task


lmorchard:
Okay, I *think* I've got this functional now.

Going to defer to #4587 for wordsmithing Preview / Review, though.

app/experimenter/kinto/tasks.py
```diff
@@ -308,6 +308,7 @@ def nimbus_update_paused_experiments_in_kinto():
     metrics.incr("nimbus_update_paused_experiments_in_kinto.completed")
 
 
+@app.task
```
lmorchard, line 311: Oh, and this was a weird thing that seems maybe a tiny gotcha in tests:

* The code using this function calls a mock
* The code testing this function calls it directly
* But nothing seems to ensure it's actually been registered as a Celery task

So, when I got down to trying to use this code, it took me a tiny bit to figure this out. Not sure we could wrap this sort of thing in tests, but it's a potential crack down in which things can fall.

app/experimenter/nimbus-ui/src/lib/experiment.ts
```diff
@@ -10,30 +10,37 @@ export function getStatus(
...
-  const released = status
-    ? [NimbusExperimentStatus.LIVE, NimbusExperimentStatus.COMPLETE].includes(
-        status,
-      )
-    : false;
+
+  // The experiment is in review, preview, or accepted state
+  const preparation = [
+    NimbusExperimentStatus.REVIEW,
+    NimbusExperimentStatus.ACCEPTED,
```
pdehaan, line 19: Does this need a check for `NimbusExperimentStatus.PREVIEW` (comment on L16 above implies "review", "preview", or "accepted").

lmorchard: Oh, no I added then removed a check and forgot to update the comment

app/experimenter/nimbus-ui/src/components/PageRequestReview/FormLaunchDraftToReview.tsx
```diff
@@ -36,7 +35,11 @@ const FormLaunchDraftToReview = ({
           <span className="text-danger">
             <AlertCircle /> We recommend previewing before launch.
           </span>{" "}
-          <a href="#" onClick={handleLaunchToPreviewClick}>
+          <a
+            href="#"
```
jodyheavener, line 39: ESLint [during compilation, not yarn lint:eslint (these two should really be one and the same grumble..)] is complaining that this needs to be a valid href: `The href attribute requires a valid value to be accessible.`. I think that means this could simply be `#launch` or similar. Or we could make it a button.

app/experimenter/nimbus-ui/src/components/HeaderExperiment/index.tsx
```diff
@@ -80,11 +80,9 @@ const HeaderExperiment = ({
 
         <p className="header-experiment-status position-relative mt-2 d-inline-block">
           <StatusPill label="Draft" active={status.draft} />
-          {
-            /* istanbul ignore next until EXP-866 final */ status.preview && (
-              <StatusPill label="Preview" active={status.preview} />
-            )
-          }
+          {status.preview && (
+            <StatusPill label="Preview" active={status.preview} />
```
jodyheavener, line 84: I think `active={status.preview}` could just be `active` since it's already wrapped in that same condition.

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.test.tsx
```diff
@@ -44,16 +42,13 @@ describe("PageRequestReview", () => {
   it("renders as expected", async () => {
     const { mock } = mockExperimentQuery("demo-slug");
     render(<Subject mocks={[mock]} />);
-    let submitButton: HTMLButtonElement;
     await waitFor(() => {
       expect(screen.getByTestId("PageRequestReview")).toBeInTheDocument();
-      submitButton = screen.getByText("Launch") as HTMLButtonElement;
+      expect(
+        screen.getByTestId("start-launch-draft-to-review") as HTMLButtonElement,
+      ).toBeInTheDocument();
```
jodyheavener, line 49: Optional: I recently discovered that we never need to use `expect(...).toBeInTheDocument()` because testing-library has an async method for this, so:

```tsx
// this
expect(screen.getByTestId("start-launch-draft-to-review") as HTMLButtonElement).toBeInTheDocument();

// becomes this
await screen.findByTestId('start-launch-draft-to-review');
```

Slightly less bulky, and if you do it for both we can eliminate the `waitFor` entirely here. But again optional - I plan to do a therapeutic sweep for these at some point.

(Also, you shouldn't need to cast to `HTMLButtonElement` if we're just asserting the element's presence.)

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.test.tsx
```diff
@@ -129,22 +124,90 @@ describe("PageRequestReview", () => {
...
+      NimbusExperimentStatus.PREVIEW,
+    );
+    render(<Subject mocks={[mock, mutationMock]} />);
+    let launchButton: HTMLButtonElement;
+    await waitFor(
+      () =>
+        (launchButton = screen.getByTestId(
+          "launch-draft-to-preview",
+        ) as HTMLButtonElement),
+    );
```
jodyheavener, line 152: That async method would be handy for instances of this:

```diff
-    let launchButton: HTMLButtonElement;
-    await waitFor(
-      () =>
-        (launchButton = screen.getByTestId(
-          "launch-draft-to-preview",
-        ) as HTMLButtonElement),
-    );
+    const launchButton = (await screen.findByTestId(
+      "launch-draft-to-preview",
+    )) as HTMLButtonElement;
```

lmorchard:
Rebased against main

jodyheavener (changes requested):
Couple thoughts and suggestions. I think the only thing I'd consider important is the anchor href, if nothing else but to silence the linter.

It continues to look great! I tried out the flows and had these notes:

I'm currently unable to Request Launch because it's claiming I can't go from Preview to Review. That's not true, though, right? A Review _is_ what you're requesting? (These two things visible at the same time also now have me wondering if the yellow container is not the best for a non-warning flow like this, but out of scope)

Screen Shot 2021-02-19 at 11 30 01 AM

Visual nitpick: if we were previously in a wonky state but got out of it perhaps the error message should go away. Edge case and not all that important, so not a blocker.

Screen Shot 2021-02-19 at 11 31 26 AM

Flow nitpick: if I click the "Request Launch without Preview" button, then the "Launch to Preview Now" link, then "Go Back to Draft" it _does_ go back to Draft, but it also takes me back to the Launch without Preview window instead of the initial state where I expected it to. This is also pretty minor and not a blocker.

jaredlockhart:
Clicking an In Preview experiment on the list page takes it to the summary page where there are no controls to move it from preview back to draft:

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -384,8 +384,15 @@ class Meta:
...
+        self.should_call_preview_task = instance and (
+            (
+                instance.status == NimbusExperiment.Status.DRAFT
+                and data.get("status") == NimbusExperiment.Status.PREVIEW
+            )
+            or (
+                instance.status == NimbusExperiment.Status.PREVIEW
+                and data.get("status") == NimbusExperiment.Status.DRAFT
+            )
         )
```
jaredlockhart, line 396: I don't think we need this level of rigor here for two reasons

1) The task itself is idempotent so calling it an extra time doesn't hurt anything
2) I think the only time we want to fast forward this call is when it's moving into preview since the user will want to test without waiting (maybe), but moving back to draft and unpublishing it doesn't block anyones workflow so that can wait until the next scheduled call

Regardless, if we want to add these additional checks here I would expect the tests to be expanded to test each of these cases.

app/experimenter/kinto/tasks.py
jaredlockhart: Yeah totally, in general I've found that kind of thing is best left to integration tests, which we'll have forthcoming when we get more support from qa engineering, otherwise you're really just testing configuration and not logic, which doesn't have a ton of benefit in my experience.

jaredlockhart (changes requested):
Great job @lmorchard !  Looks and works great for me.  I did find that the celery task and database would race each other on my local so the fast forwarded task invocation on save didn't always pick up the changes, but I would expect on production that to be drawn out a bit more.  Plus the fast forwarded call is just a nice to have, the scheduled invocation will pick it up on its next call if the fast forwarded one misses it.

I think the list page not linking back to the page with the controls is a significant blocker since an experiment could get stuck in preview and not be recoverable back to draft by an end user without either knowing to hack the urls or ask someone to fix it in the admin.

And if we're going to expand the logic around when to fast forward the task invocation we should expand its tests to capture all those cases.

lmorchard:
Alright, pushing some tweaks:

* Added a "Go to review" link on the experiment view page when in preview.
  * This doesn't add the forms to transition from Preview to Draft or Review, but links to the page where that can be done.
  * Issue #4577 might be a better solution than duplicating / abstracting those forms to work on the summary page.
* Forgot to ensure PREVIEW -> REVIEW transition was valid - added that and parameterized tests to ensure all transition pairs are valid
* Parameterized test for when the kinto task gets called
* Changed "Launch to Preview Now" from link to button
* Jest test tweaks to use `await findByTestId` instead of `await waitFor`
* Clearing both submit errors and showLaunchDraftToReview state when submitting status update mutation, which should clear up some of the weird sticky state issues

jodyheavener (approved):
LGTM! Love how this turned out. Much better than the check boxes at the bottom of the page.

jaredlockhart (approved):
Thank you for updating the tests and adding the link!  Also I figured out we can add a small delay to invoking the task that gets around the race condition on my local so hopefully that should help it perform reliably in production, so I pushed up a small addendum for that.  Looks and works great, great job @lmorchard 🎉 🎉 🎉

lmorchard:
Hittin the merge button!

giphy

Issue #4659, opened 2021-02-19
Highlights table should show "baseline" for control variant


emtwo:
Based on user feedback, showing the control values here is both redundant with the Results table, and confusing.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-983)
┆Issue Number: EXP-983

PR #4663, opened 2021-02-19
fix #4505, #4659: feat(visualization) Add tooltips for guardrail metrics and update control content for highlights table

### LINKED ISSUE #4505: As a user I want to know how the guardrail metrics in the visualization are calculated
In the visualizations, there is no definition how the guardrail metrics are calculated, and this might help them make a decision. 

Maybe a possible solution could be that when the user hovers over the guardrail metric name there is a tooltip with the definition of the metric.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-870)
┆Issue Number: EXP-870

emtwo:
Because
* The content for "control" in the highlights table was confusing
* The guardrail computations were not clear

This commit:
* Shows "baseline" for control in the highlights table
* Adds tooltips using metadata about the guardrail metrics

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
@@ -12,14 +12,14 @@ import {
   SEGMENT_TIPS,
   TABLE_LABEL,
 } from "../../../lib/visualization/constants";
-import { AnalysisDataOverall } from "../../../lib/visualization/types";
+import { AnalysisData } from "../../../lib/visualization/types";
 import { getTableDisplayType } from "../../../lib/visualization/utils";
 import { getExperiment_experimentBySlug_primaryProbeSets } from "../../../types/getExperiment";
 import TableVisualizationRow from "../TableVisualizationRow";
 
 type TableHighlightsProps = {
   primaryProbeSets: (getExperiment_experimentBySlug_primaryProbeSets | null)[];
-  results: AnalysisDataOverall;
+  results: AnalysisData;
 };
 
 const getHighlightMetrics = (
@@ -40,16 +40,23 @@ const getHighlightMetrics = (
 
 const TableHighlights = ({
   primaryProbeSets,
-  results = {},
+  results = {
+    daily: [],
+    weekly: {},
+    overall: {},
+    metadata: { metrics: {}, probesets: {} },
+    show_analysis: false,
+  },
 }: TableHighlightsProps) => {
   const highlightMetricsList = getHighlightMetrics(primaryProbeSets);
+  const overallResults = results?.overall!;
 
   return (
     <table data-testid="table-highlights" className="table mt-4 mb-0">
       <tbody>
-        {Object.keys(results).map((branch) => {
+        {Object.keys(overallResults).map((branch) => {
           const userCountMetric =
-            results[branch]["branch_data"][METRIC.USER_COUNT];
+            overallResults[branch]["branch_data"][METRIC.USER_COUNT];
           const participantCount =
             userCountMetric[BRANCH_COMPARISON.ABSOLUTE]["first"]["point"];
           return (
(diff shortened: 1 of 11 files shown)


emtwo:
For context:

There are 3 changes in this PR that can be seen in storybook:

1. "baseline" is listed for control in the highlights table as shown in the first screenshot below
2. Guardrail metrics in the highlights table show a tooltip when hovering over them (e.g. search in 1st screenshot)
3. Guardrail metrics in the results table show a tooltip when hovering over them (e.g. search in 2nd screenshot)

Screen Shot 2021-02-19 at 3 31 59 PM
Screen Shot 2021-02-19 at 3 32 14 PM

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -54,10 +54,12 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
           </p>
           <h3 className="h6">Hypothesis</h3>
           <p>{experiment.hypothesis}</p>
-          <TableHighlights
-            primaryProbeSets={experiment.primaryProbeSets!}
-            results={analysis?.overall!}
-          />
+          {analysis && (
```
LZoog, line 57: Hrmm, do we need this check and as well as on L71 since the user should be redirected if the analysis is unavailable? Or does [this comment](https://github.com/mozilla/experimenter/blob/emtwo/more_metadata/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx#L35)/something around it need any tweaking?

app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
```diff
@@ -65,25 +72,34 @@ const TableHighlights = ({
...
+                <td className="pt-3 px-3">
+                  {highlightMetricsList.map((metric) => {
+                    const metricKey = metric.value;
+                    const displayType = getTableDisplayType(
+                      metricKey,
+                      TABLE_LABEL.HIGHLIGHTS,
+                      overallResults[branch]["is_control"],
+                    );
+                    const tooltip =
+                      results.metadata?.metrics[metricKey]?.description ||
```
LZoog, line 89: Huh, I'll look more closely at this tomorrow morning but I'm not seeing this come through on Storybook. 🤔

app/experimenter/nimbus-ui/src/lib/visualization/mocks.tsx
```diff
@@ -27,6 +27,12 @@ export const MOCK_METADATA = {
         "This is a metric description. It's made by a data scientist at the creation time of the metric [I'm a link](https://www.example.com)",
       friendly_name: "Feature D Friendly Name",
     },
+    search_count: {
+      bigger_is_better: true,
+      description:
+        "This is a metric description for search. It's made by a data scientist at the creation time of the metric [Link](https://www.example.com)",
```
LZoog, line 33: What will this be linking to out of curiosity? This may be out of line but should this be like, `This is a [metric description for search](https://www.example.com)]. It's made by a data scientist at the creation time of the metric.`

LZoog (commented):
This is looking really good so far, just have a couple questions and I might need to look harder tomorrow AM at Storybook :)

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
emtwo: Oh, no we don't need it, you're right. I was just making typescript happy, but I think `analysis!` is what we need here

app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
emtwo: Where were you checking on storybook? The screenshot I shared is from storybook so I believe it should show up there.

app/experimenter/nimbus-ui/src/lib/visualization/mocks.tsx
emtwo: Oh, well the descriptions are coming in from Jetstream, they links would be more details for search, for example. I suppose I could update it. I just added the link to show it's possible to have a link and how it would work.

app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
LZoog: I totally see it now! I was looking at search increased, not decreased.

image

BTW, I noticed that some of the tooltips are not at total opacity and others are. It seems like especially the one with the blue link should be on the darker side? This is a tiny nit though (could file a follow up as a `quick-win` to update the opacity for consistency if you'd prefer to get this in and avoid another CI run). I also wonder, do users know that they can see all this nice extra info by hovering over certain pieces?

image

LZoog (approved):
👍

app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
emtwo: I'll file a follow up about the opacity. In user interviews we've seen people try to hover over everything for more info. A little "?" icon might be a bit better but I think we can wait to see if we get that as direct feedback from upcoming interviews

emtwo:
Filed follow up: https://github.com/mozilla/experimenter/issues/4678

app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.tsx
jodyheavener: Interesting. Is the difference in opacity expected? Or is it possible that multiple tooltips are being rendered/layering to create an opaque effect?

PR #4665, opened 2021-02-22
Move test w/ coverage flag to its own command, move mock experiment to its own helper

jodyheavener:

This PR:

1. Updates `yarn test` to _not_ generate coverage output by default, adds a new `yarn test:cov` that does this (and updates the Makefile to use it), and documents how you could run coverage on a subset of tests
  - Why change `test`? During development I often want to have the tests running on file change, and this is obscured by the long output of coverage results on each run.

2. Adds a new helper, `mockExperiment` that just generates a mocked experiment (and updates `mockedExperimentQuery` to use it). We have a number of tests that need a mocked experiment that either make no use of or have nothing to do with GraphQL, so it doesn't make sense that it should need to also mock the query.

### DIFF
--- a/app/experimenter/nimbus-ui/src/lib/mocks.tsx
+++ b/app/experimenter/nimbus-ui/src/lib/mocks.tsx
@@ -232,23 +232,17 @@ export class SimulatedMockLink extends ApolloLink {
   }
 }
 
-export function mockExperimentQuery<
+export function mockExperiment<
   T extends getExperiment["experimentBySlug"] = getExperiment_experimentBySlug
->(
-  slug: string,
-  modifications: Partial<getExperiment["experimentBySlug"]> = {},
-): {
-  mock: MockedResponse<Record<string, any>>;
-  experiment: T;
-} {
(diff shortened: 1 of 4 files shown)


lmorchard (approved):
:+1: Good tweaks!

app/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -312,11 +306,20 @@ export function mockExperimentQuery<
...
+  ) as T;
+}
 
-  if (modifications === null) {
-    experiment = null;
-  }
+export function mockExperimentQuery<
+  T extends getExperiment["experimentBySlug"] = getExperiment_experimentBySlug
+>(
+  slug = "foo",
```
jodyheavener, line 315: Forgot to note this, but I'm also slowly moving us away from needing to specify a slug when creating a new mockedExperimentQuery. In most cases a default slug is fine, and only when we're testing routes and some other minor areas do we need to control the slug.

PR #4666, opened 2021-02-22
feat(nimbus): refactor useAnalysis and useExperiment hooks to use RTL react-hooks

jodyheavener:

As it turns out, there is a testing-library [library](https://react-hooks-testing-library.com/) made specifically for React Hooks!

This PR adds that dependency and refactors the tests for two hooks, `useExperiment` and `useAnalysis`, to use RHTL. 

What does this get us? It takes care of most of the setup for rendering a hook, since it's required to be in the context of a component. Here's an example:

```tsx
// This...
let hook: ReturnType<typeof useAnalysis>;
const TestHook = () => {
  hook = useAnalysis();
  return <p>Foo</p>;
};
render(<TestHook />); 
expect(hook.loading).toBeFalsy();

// Becomes this...
const { result } = renderHook(() => useAnalysis());
expect(result.current.loading).toBeFalsy();
```

In the case of `useExperiment`, where a mocked cache is required, it also allows us to supply a wrapping component.

If we like this I'll work on updating the rest at some point.

### DIFF
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -67,6 +67,7 @@
     "@testing-library/dom": "^7.24.3",
     "@testing-library/jest-dom": "^5.11.8",
     "@testing-library/react": "^11.0.4",
+    "@testing-library/react-hooks": "^5.0.3",
     "@testing-library/user-event": "^7.1.2",
     "@types/jest": "^24.0.0",
     "@types/node": "^14.11.5",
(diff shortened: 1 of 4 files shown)


lmorchard (approved):
Yeah, this is some good simplification! I don't think I'd actually seen `@testing-library/react-hooks` and instead was going off [this kind of advice with a mock component](https://kentcdodds.com/blog/how-to-test-custom-react-hooks) for a long while.

jodyheavener:
> this kind of advice with a mock component for a long while.

@lmorchard I mean to be fairrrr, the article you linked to does mention that RHTL could be reserved for more complicated hooks, which I wouldn't consider to be true for either of these really. So I wouldn't be upset if we wanted to keep using hand-rolled mock components for these. I think I just like RHTL because it's consistent and reduces the likelihood of setup issues.

jaredlockhart:
Oh very nice!

lmorchard:
>  to be fairrrr

(to be fairrr)

But eh, simplifying these simple tests is :ok: too

Issue #4213, opened 2020-12-10
Mocked query test order requires "loading" test at the end


data-sync-user:
From: [https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.test.tsx#L118:L118](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.test.tsx#L118:L118)

This isn't hurting anything but is kind of annoying: when testing mocked GQL queries the test where there are no mocks and the component is intentionally "loading", this test must be placed at the end of the grouping, because it fails all mocked query tests after it.

Closing this ticket should mean finding out why this occurs and updating the tests so they can be executed independent and without impacting one another.

Alternatively, since this is not an issue unless you move tests around, we could close this with the only change being to remove the above noted TODO line. However it's possible - even likely - that someone without this knowledge will copy-paste an existing set of tests for a new component and encounter this issue without knowing why, so it's better we figure it out.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-733)
┆Issue Number: EXP-733

data-sync-user:
➤ Jody Heavener commented:

This is no longer an issue.

PR #4667, opened 2021-02-22
feat(nimbus): update AppLayoutWithExperiment tests to use RTL best practices

jodyheavener:

This PR is a refactor of the tests in `AppLayoutWithExperiment` for a few reasons:

- Organize and tighten them up; remove extra/unneeded code, make descriptions more concise; was able to reduce overall lines
- Address the issue that is part of #4213 - the "loading" state test can now be freely moved anywhere
- Incorporate some React Testing Library best practices. For this PR this includes:
  - Using async `findBy*` methods to determine if an element is in the DOM; only using `toBeInTheDocument()` to assert _absence_ of an element. ([Source](https://testing-library.com/docs/dom-testing-library/api-async/#findby-queries))
  - Where appropriate, using `*ByRole` and `*ByText` methods to assert the presence and inspect properties of DOM elements instead of `*ByTestId`. These methods behave more closely to what the user can actually see and interact with on the page. As a result of the above, was able to remove some `data-testid` attributes from our components. ([Source](https://testing-library.com/docs/queries/about/#priority))

There are more RTL best practices that I plan to fold in over time, and I will probably try to come up with a thorough blog post or even just internal documentation on the subject. Feel free to push back on any of this.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.test.tsx
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { MockedResponse } from "@apollo/client/testing";
 import { navigate } from "@reach/router";
 import { act, render, screen, waitFor } from "@testing-library/react";
 import React from "react";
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.test.tsx
```diff
@@ -146,34 +112,34 @@ describe("AppLayoutWithExperiment", () => {
...
+  ...jest.requireActual("@reach/router"),
+  navigate: jest.fn(),
+}));
+
 const Subject = ({
   mocks = [],
-  polling = false,
-  sidebar = true,
-  withTitle = true,
-  redirect = () => {},
```
lmorchard, line 162: Scratched my head a little at these defaults going away, but looks like this all better matches the actual type of the tested component so :+1:

lmorchard (approved):
👍 

Fun fact: I was *slightly* thrown by "RTL best practices" in the title - because I usually see "RTL" as right-to-left in internationalization contexts and I didn't think we had any translations 😅

PR #4681, opened 2021-02-23
fix #4679 bug(nimbus): check for pending requests in remote settings before updating paused experiments

### LINKED ISSUE #4679: Check for pending reviews in remote settings before making any updates in pause task
In our task to update the enrollment paused state, we should check whether the collection has any pending reviews before attempting to update the record.  Otherwise the task will update the collection every time it is invoked, and spam the whole company with duplicate notification emails all night until someone gets paged and has to frantically disable it in the middle of eating dinner.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-995)
┆Issue Number: EXP-995

jaredlockhart:
Because

* The update paused experiments task would mark the collection for review every time it was invoked regardless of whether there was already a pending change for review
* This accidentally spammed a bajillion notification emails

This commit

* Changes the update paused task to check whether there's a pending review before updating a record

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -290,6 +290,13 @@ def nimbus_update_paused_experiments_in_kinto():
 
     for application, collection in NimbusExperiment.KINTO_APPLICATION_COLLECTION.items():
         kinto_client = KintoClient(collection)
+
+        if kinto_client.has_pending_review():
+            metrics.incr(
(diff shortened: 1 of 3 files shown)


jodyheavener (approved):
This is a little outside my wheelhouse, but based on your description and how the updated test looks I think I understand what's going on and it makes sense. I appreciate our descriptive method names.

Issue #4615, opened 2021-02-12
Add client-side validation with error message to disallow negative numbers for all Audience fields


jaredlockhart:
Expected Number of Clients will show a helpful error message when it is invalid for the case where the value is not a number:

<img width="480" alt="Screen Shot 2021-02-12 at 4 23 22 PM" src="https://user-images.githubusercontent.com/119884/107823963-a818b400-6d4e-11eb-94f4-185abcfbbfae.png">

but not for the case where the value is negative:

<img width="477" alt="Screen Shot 2021-02-12 at 4 22 16 PM" src="https://user-images.githubusercontent.com/119884/107823987-b070ef00-6d4e-11eb-9541-3a494bfc0362.png">

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-956)
┆Issue Number: EXP-956

LZoog:
This is a dupe of #4276, I'll add a comment over there to add helpful messaging

LZoog:
Hmmm so #4276 is related but the actual issue reported there is that you could _save_ the negative number and that's different than adding client-side validation.

I'm reopening this and closing the other. Though, **we should also enforce this rule on other Audience fields, not _just_ number of clients.**

lmorchard:
> we should also enforce this rule on other Audience fields, not just number of clients.

Unless I'm missing something, I think this is already covered on the server-side. I can't reproduce being able to set other negative values in the GQL mutation.

The `population_percent` field was given a positive value constraint here: https://github.com/mozilla/experimenter/commit/69e3bafe9c8db49d4ff4f4f673b81bad2db59ddd

The rest of the fields use `PositiveIntegerField`: https://github.com/mozilla/experimenter/blame/main/app/experimenter/experiments/models/nimbus.py#L35-L44

PR #4701, opened 2021-02-24
fixes #4275 fix(nimbus): accept commas in number of clients field

### LINKED ISSUE #4275: Disallow commas in Audience "Expected number of clients" field
https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/pdehaan-2020-12-15-001/edit/audience

Found by Alex Davis in this morning's launch party.
If you copy the expected number of clients from an external document and it has a comma in it (ie: "360,000"), it looks like Firefox doesn't strip the comma on pasting (but prevents you from typing in a comma). Unable to reproduce in Chrome. Basically Firefox gives you a very confusing "This field may not be blank." error; which is odd since the field is not blank.

Not sure if we need some client side sanitizer to remove non-numeric values onBlur or something.

---

pdehaan_2020-12-15_001_–_Audience___Experimenter

Removing the comma fixes the validation error. Typing the comma reintroduces the error.

pdehaan_2020-12-15_001_–_Audience___Experimenter

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-761)
┆Issue Number: EXP-761

lmorchard:
Because:

- Firefox appears to allow commas on paste into number fields

This commit:

- adds a validator that accepts non-numeric characters but strips them
  before parsing and validating the input as a number

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -119,7 +119,7 @@ describe("FormAudience", () => {
       firefoxMinVersion: MOCK_EXPERIMENT.firefoxMinVersion,
       targetingConfigSlug: MOCK_EXPERIMENT.targetingConfigSlug,
       populationPercent: "" + MOCK_EXPERIMENT.populationPercent,
-      totalEnrolledClients: "" + MOCK_EXPERIMENT.totalEnrolledClients,
+      totalEnrolledClients: MOCK_EXPERIMENT.totalEnrolledClients,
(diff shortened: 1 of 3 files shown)


lmorchard:
This PR does essentially the opposite of the literal ask in the issue: Rather than reject commas, this validator accepts them but strips before handling the input as a number. That way, we should be able to accommodate direct paste-in of numbers from external docs.

lmorchard:
Whoops, missed that linting had failed. Let's try that again

lmorchard:
And there it hits the intermittent failure, rebased against main to try to pick up that CI fix

LZoog (approved):
Nice, LGTM! You could also get #4615 done here or in a quick follow up with this new validation. 😁

lmorchard:
Would rather go for a follow up than expand things here

PR #4737, opened 2021-03-01
chore(project): disable intermittently failing integration tests

jaredlockhart:
Because

* We added integration tests to load and execute targeting expressions
* We use the latest Firefox Nightly build to run our integration tests
* Something may have changed that prevents these tests from running properly

This commit

* Disables the tests until we can find a better fix


jodyheavener (approved):
LGTM provided that same flakiness isn't seen here/you disabled the right tests

PR #4743, opened 2021-03-02
Fix #4507: feat(visualization) Add treatment and control definitions to the top of the page.

### LINKED ISSUE #4507: Add treatment and control definitions to the top of the results page in visualizations
During user testing, we noticed that the user assumed he knew what the definitions were for the control and treatment branch. 

Potential solution would be to add the control and treatment names + descriptions to the top of the results page.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-869)
┆Issue Number: EXP-869

emtwo:
Because:
* Users want to see descriptions of branches when they are looking at results
* Users are confused by "All Users" in the highlights when there is nothing to contrast it

This commit:
* Adds branch descriptions at the top of the results page
* Removes "All Users" from highlights - we can add this back in when there are more segments to work with (e.g. first-run users).

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.test.tsx
@@ -12,25 +12,37 @@ import { mockAnalysis } from "../../../lib/visualization/mocks";
 const { mock, experiment } = mockExperimentQuery("demo-slug");
 
 describe("TableHighlights", () => {
-  it("has participants for all users shown for each variant", () => {
-    const EXPECTED_LABELS = ["participants", "All Users"];
-
+  it("has participants shown for each variant", () => {
     render(
       <RouterSlugProvider mocks={[mock]}>
         <TableHighlights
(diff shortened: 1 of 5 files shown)


emtwo:
For context: 
Screen Shot 2021-03-02 at 3 13 24 PM

app/experimenter/nimbus-ui/src/components/PageResults/TableHighlights/index.test.tsx
```diff
@@ -12,25 +12,41 @@ import { mockAnalysis } from "../../../lib/visualization/mocks";
...
-    EXPECTED_LABELS.forEach((label) => {
-      expect(
-        screen.getAllByText(label, {
-          exact: false,
-        }),
-      ).toHaveLength(2);
-    });
+    expect(
+      screen.getByText(branchDescription, {
+        exact: false,
```
jodyheavener, line 47: Nit: since you're matching against the whole description I think this _can_ be exact

jodyheavener (approved):
One nit but not a blocker. LGTM

PR #4746, opened 2021-03-02
fix #4569 feat(nimbus): load outcomes from jetstream config

### LINKED ISSUE #4569: (back-end) Pull in outcome data + read from fs
Per the design brief (link in epic):
To add a new outcome, a data scientist would write a configuration snippet, commit it to a path in the jetstream-config repository, open a pull request, and merge the pull request once it tests green.

This logic should be loaded in at build time. We need to load in the outcomes file from GH and read it out of memory.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-924)
┆Issue Number: EXP-924

jaredlockhart:
Because

* Rather than storing probesets in the experimenter db we will now consume outcome definitions from the jetstream config repo

This commit

* Adds the jetstream config repo as a git submodule
* Updates the makefile to update the jetstream config submodule on every docker container build
* Adds an outcomes python module that loads the TOML files into memory and provides some handy dandy lookup methods
* Tests for the outcomes module
* Module docs explaining outcomes and their relationship to experiments and use

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,13 +50,17 @@ nginx/cert.pem: nginx/key.pem
 secretkey:
 	openssl rand -hex 24
 
-build_dev:
+jetstream_config:
+	git submodule init
+	git submodule update
+
+build_dev: jetstream_config
 	docker build --target dev -f app/Dockerfile -t app:dev app/
 
-build_test:
+build_test: jetstream_config
 	docker build --target test -f app/Dockerfile -t app:test app/
 
-build_prod:
+build_prod: jetstream_config
 	docker build --target deploy -f app/Dockerfile -t app:deploy app/
 
 test_build: build_test
--- a/app/experimenter/outcomes/README.md
+++ b/app/experimenter/outcomes/README.md
@@ -1,0 +1,36 @@
+# Outcomes
+
+## Overview
+
+[Outcomes](https://github.com/mozilla/jetstream-config/#outcome-snippets) define automated analyses that can be applied to Nimbus experiments. They are stored in the [Jetstream Config repo](https://github.com/mozilla/jetstream-config/tree/main/outcomes) as [TOML](https://github.com/toml-lang/toml) files and are consumed by Experimenter as a [Git Submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules).
+
+## Building and Deploying
+
+Every time the Experimenter Docker container is built, it will synchronize the Jetstream Config repo into the local Experimenter source tree, which will then be copied with the rest of Experimenter's source into the Docker container. The Outcomes module will load the outcome definition TOMLs into memory where they can be accessed.
+
+## Using Outcomes
+
+Outcomes can be accessed by using the Outcome and Outcomes classes. An Outcome has the following properties:
+
+- `application: str` An application ID defined as one of `("fenix", "firefox-desktop")`
+- `app_name: str` An application name defined as one of `("fenix", "firefox_desktop")`
+- `description: str` A helpful description to display in Experimenter's UI when choosing outcomes
+- `friendly_name: str` A name to display in Experimenter's UI when choosing outcomes
+- `slug: str` A unique identifier to include in the experiment DTO when deploying experiments
+
+```py
+from experimenter.experiments.constants import NimbusConstants
+from experimenter.outcomes import Outcomes
+
+# All outcomes for all applications
+for outcome in Outcomes.all():
+    print(outcome.slug)
+
+# All outcomes for a given application
+for outcome in Outcomes.by_app_id(NimbusConstants.Application.FENIX):
+    print(outcome.slug)
+```
+
+## Consistency
+
+A Nimbus experiment stores the slugs of the outcomes that are selected by an experiment's owner when it is created. After that point, the set of outcomes stored in the Jetstream Config repo may change in any way at any time, and so by the time the experiment is complete and analyzed by Jetstream, the set of outcome slugs stored on the experiment and the set of outcome slugs stored in the Jetstream Config repo may differ, in which case Jetstream will find the set of slugs that match and analyze those, disregarding any that were selected by the experiment owner that no longer match any definition in Jetstream Config.
--- a/app/experimenter/outcomes/__init__.py
+++ b/app/experimenter/outcomes/__init__.py
@@ -1,0 +1,76 @@
+import os
+from dataclasses import dataclass
+
+import toml
+from django.conf import settings
+from django.core.checks import Error, register
+
+from experimenter.experiments.constants import NimbusConstants
+
+
+@dataclass
+class Outcome:
+    application: str
+    app_name: str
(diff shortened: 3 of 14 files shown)


app/experimenter/outcomes/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Building and Deploying
+
+Every time the Experimenter Docker container is built, it will synchronize the Jetstream Config repo into the local Experimenter source tree, which will then be copied with the rest of Experimenter's source into the Docker container. The Outcomes module will load the outcome definition TOMLs into memory where they can be accessed.
+
+## Using Outcomes
+
+Outcomes can be accessed by using the Outcome and Outcomes classes. An Outcome has the following properties:
+
+- `app_id: str` An application ID defined as one of `("fenix", "firefox-desktop")`
```
tdsmith, line 15: This name collides with the appId in [the DTO](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=130920248), which has a many-to-one relationship to appName; I suspect this nomenclature is deeply entrenched in Experimenter, but since this is an interface, can we just clarify here that this isn't the same as the DTO appId?

app/experimenter/outcomes/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+            for outcome_name in os.listdir(app_path):
+                outcome_path = os.path.join(app_path, outcome_name)
+
+                with open(outcome_path, "r") as outcome_file:
+                    outcome_toml = outcome_file.read()
+                    outcome_data = toml.loads(outcome_toml)
+
+                    outcomes.append(
+                        Outcome(
+                            app_id=NimbusConstants.APP_NAME_APPLICATION[app_name],
```
tdsmith, line 38: If we support a new app in jetstream-config this will break the Experimenter build -- what about ignoring paths for unrecognized app_names?

tdsmith (approved):
its_happening.gif

app/experimenter/outcomes/__init__.py
jaredlockhart: We could ignore it or we could break on it and be forced to add it to our list?  Will there be jetstream outcome apps that experimenter will never know about?

app/experimenter/outcomes/README.md
jaredlockhart: Wait it's not the same?  Can you provide an example?

app/experimenter/outcomes/__init__.py
tdsmith: > Will there be jetstream outcome apps that experimenter will never know about?

I agree not "never" but maybe transiently.

Hm, I don't think tests here run against the live data so I think the failure mode is actually that the container would build fine but the app would crash on load.

app/experimenter/outcomes/README.md
tdsmith: Yeah, the DTO doc has examples:

> Should match app_id in https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml. Possible values include:
>    "org.mozilla.fenix", for Fenix nightly
>    "org.mozilla.firefox_beta", for Fenix beta
>    "org.mozilla.firefox", for Fenix release
>    "firefox-desktop", for Firefox Desktop

Makefile
```diff
@@ -50,13 +50,17 @@ nginx/cert.pem: nginx/key.pem
 secretkey:
 	openssl rand -hex 24
 
-build_dev:
+jetstream_config:
```
lmorchard, line 53: Is this only ever going to be used for a jetstream git submodule? If not, it could be handy to name this just for purpose (e.g. `init_git_submodules`). Maybe someday we want to roll in other git submodules if only for further shared-configuration integration testing?

app/experimenter/outcomes/README.md
jaredlockhart: Oh yep perfect ty this should not be `app_id` 👍

app/experimenter/outcomes/__init__.py
jaredlockhart: Yep, which would get caught at CI time, the container would fail to start during the integration tests.  I'm fine with doing either, ignoring apps that aren't recognized or breaking entirely, which do you prefer?

app/experimenter/outcomes/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        outcomes = []
+
+        for app_name in os.listdir(settings.JETSTREAM_CONFIG_OUTCOMES_PATH):
+            app_path = os.path.join(settings.JETSTREAM_CONFIG_OUTCOMES_PATH, app_name)
+
+            for outcome_name in os.listdir(app_path):
+                outcome_path = os.path.join(app_path, outcome_name)
+
+                with open(outcome_path, "r") as outcome_file:
+                    outcome_toml = outcome_file.read()
```
lmorchard, line 33: It looks like the jetstream-config repo has validation for these files, but would it be handy to have some kind of last-ditch exception handler here to log errors just in case parsing fails during experimenter startup? 

Probably small chance of it happening and we'll see a toml-related error in CI, so maybe not worth much fuss

app/experimenter/outcomes/tests/test_outcomes.py
```diff
@@ -0,0 +1,52 @@
...
+import pathlib
+
+from django.test import TestCase, override_settings
+
+from experimenter.experiments.constants import NimbusConstants
+from experimenter.outcomes import Outcome, Outcomes
+
+
+@override_settings(
+    JETSTREAM_CONFIG_OUTCOMES_PATH=os.path.join(
```
lmorchard, line 11: Since we have all the available TOML in the submodule, would it be handy to have a test that iterates through all the current files and parses them for at least a sanity check? More of an integration test than a unit test, but could catch problems early

lmorchard (approved):
LGTM, the only things I could think of were double-checking that the TOML in the submodule works as expected in our parsing code. But that's probably an edge-case concern

Makefile
jaredlockhart: Yeah we might, I liked having it explicitly declare that building depends on jetstream config, we could rename it if that comes up good point 👍

app/experimenter/outcomes/__init__.py
jaredlockhart: Yeah I think if they don't adhere to the schema this will just straight up break and complain about a key error which is all the info I think we'll need to debug.  And it'll break in CI.

app/experimenter/outcomes/tests/test_outcomes.py
jaredlockhart: We could do it as a [django system check](https://docs.djangoproject.com/en/3.1/topics/checks/) at startup, that's also a good idea

jaredlockhart: Added a startup check!

PR #4754, opened 2021-03-03
chore(project): allow nested .vscode folders

jaredlockhart:
Because

* It can be necessary to have a .vscode directory under the root or under ./app

This commit

* Ignores .vscode directories at any depth

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,7 +74,7 @@ target/
 .psql_history
 
 # VS Code
-.vscode/*
+**/.vscode/*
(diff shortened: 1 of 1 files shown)


jodyheavener (approved):
Anything in those nested vscode folders you want to add to the repo?

jaredlockhart:
No that's what I posted in the previous PR, but I have both root and `/app` added as workspace folders becuase vscode will only find python config files if they're at the 'root' of the workspace 🤷‍♀️

PR #4756, opened 2021-03-04
fix #4570 feat(nimbus): add primary/secondary outcomes to nimbus expe…

### LINKED ISSUE #4570: (back-end) Update probe set references to outcome
Once the data is available in memory from reading the outcomes data file in EXP-924, we need to change the foreign keys to slug fields and likely add convenience model methods for reading primary/secondary outcomes.

We also need to update the read references to read from this file instead of the DB.

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/queries.py#L55
https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/types.py#L129-L130
https://github.com/mozilla/experimenter/blob/main/app/experimenter/visualization/api/v3/views.py#L167

We can update tests to inject file system fixtures.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-925)
┆Issue Number: EXP-925

jaredlockhart:
…riment

Because

* We want to track primary and secondary outcomes for Nimbus experiments

This commit

* Adds primary_outcomes and secondary_outcomes fields to NimbusExperiment
* Exposes those fields on the V5 API
* Validate against the list of available of outcomes and their respective application


app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -357,6 +363,72 @@ def test_update_experiment_probe_sets_error(self):
...
+        experiment = NimbusExperimentFactory.create(
+            status=NimbusExperiment.Status.DRAFT,
+            application=NimbusExperiment.Application.DESKTOP,
+            primary_outcomes=[],
+            secondary_outcomes=[],
+        )
+        outcomes = [
+            o.slug for o in Outcomes.by_application(NimbusExperiment.Application.DESKTOP)
+        ]
+        primary_outcomes = outcomes[: NimbusExperiment.MAX_PRIMARY_PROBE_SETS]
```
lmorchard, line 377: Ah, this is the reason for `ignore = E203` in flake8? Not used to seeing a space in this notation (probably thanks to flake8 defaults), but it does make it clearer with that long constant name

lmorchard (approved):
Looks straightforward to me, most interesting thing I could come up with as a comment was the tweak to flake8 rules

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: Yeah I think black is doing it and since you can't configure black it's just one of those things where black is like 'this is how I do it, deal with it' so I had to add the exception 🤷‍♀️

Issue #4752, opened 2021-03-03
Filter outcomes by application on the frontend


data-sync-user:
Now that outcomes are bound to applications (unlike probesets), we must filter the set of available outcomes for selection on the frontend.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1038)
┆Issue Number: EXP-1038

data-sync-user:
➤ Les Orchard commented:

Adding EXP-926: can't filter outcomes by application in the front-end until outcomes are implemented in the front-end to begin with

data-sync-user:
➤ Jody Heavener commented:

Thanks, I was wondering about that. I'm looking at Lauren's WIP branch for 926 to see what needs to be done.

PR #4759, opened 2021-03-05
fixes #4572 feat(nimbus): Update probe set references to outcome

### LINKED ISSUE #4572: (mostly front-end) Update probe set references to outcome
Note that this task needs to be done last in the console work for this epic (excluding deprecating the DB tables). The outcomes data should be available after EXP-925 is complete.

This task should be straightforward but will touch a lot of files.

We need to update the BE query to return outcomes instead of probe sets.

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/queries.py#L27
https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/queries.py#L54
https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/gql/config.ts#L31-L44

In addition, we need changing any references to `probeSets` in the front-end, mostly around the Metrics page and Results page (and tables inside the Results page), but also the Summary table and any other references.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-926)
┆Issue Number: EXP-926

LZoog:
This still needs some work but I think I got a lot of the find-and-replacing out of the way at least. **Please feel free to push to this branch** since I'm out on Mon/Tues if we need to get this in. Otherwise I'll get back to this on Wed.

It might be nice to move the type of `Outcome`/`Outcomes` out of `FormMetric` so we can use it across the nimbus UI.


app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -40,17 +40,18 @@ const getSidebarItems = (
   return sidebarItems;
 };
 
+// TODO: fix this
 const probesetToMapping = (
   probesets:
-    | (getExperiment_experimentBySlug_primaryProbeSets | null)[]
-    | (getExperiment_experimentBySlug_secondaryProbeSets | null)[],
+    | (getExperiment_experimentBySlug_primaryOutcomes | null)[]
```
LZoog, line 46: This is from find-and-replace leftover pasta, `getExperiment_experimentBySlug_primaryOutcomes` isn't generated

jodyheavener:
From 121 failures down to 32 😅📉

jodyheavener:
Down to 7 failures! 🤠

app/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return slugs
+      .map((slug) => outcomes!.find((outcome) => outcome!.slug === slug))
+      .filter((outcome) => outcome != null);
+  };
+
+  return {
+    primaryOutcomes: pairOutcomes(experiment.primaryOutcomes),
+    secondaryOutcomes: pairOutcomes(experiment.secondaryOutcomes),
+  };
+}
```
jodyheavener, line 27: Since this hook already knows about the experiment and available config outcomes I think in #4752 we should extend it to also return an `available` property with the outcomes applicable to this experiment application.

jodyheavener:
I'm going to be looking through this myself some more, but it's more or less ready for others' eyes.

app/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
jaredlockhart: Yeah I was thinking about that but to do that I think you'd have to pass in the experiment you're talking about to the ExperimentConfig call which we're not currently doing?  And that would mostly work for now since we enforce setting application at creation time and then not changing it, but in theory either of those things could change which would affect the semantics of making ExperimentConfig dependent on an experiment id?

jodyheavener: We might be talking about two separate things, I was talking about returning `available: Outcome[]` from the `useOutcomes` hook. Something like this:

```
const available = outcomes?.filter(
  (outcome) => outcome?.application === experiment.application,
) || [];
```

This is possible because this hook already knows about the experiment (and its application), and the list of outcomes provided by the initial config query.

Make sense?

jaredlockhart: Oh oh you mean entirely on the frontend oh yes totally 👍

jaredlockhart (approved):
Yep looks good to me, tried locally and works perfectly.  I even replicated the case where the experiment has defined outcome slugs which no longer appear in the list of available outcomes, and the frontend just quietly ignores them which I think is correct.  Great job! 🎉

app/experimenter/nimbus-ui/src/react-app-env.d.ts
```diff
@@ -6,3 +6,7 @@
 
 type DateTime = string;
 type ObjectField = Record<string, any> | string;
+
+type OutcomeSlugs = (string | null)[] | null;
+type Outcome = getConfig_nimbusConfig_outcomes | null | undefined;
```
lmorchard, line 11: I noticed that this type is effectively `any[]` because we need an import in this module:

```javascript
import { getConfig_nimbusConfig_outcomes } from "./types/getConfig"
```

Not sure why typescript isn't complaining about that.

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -44,20 +42,11 @@ const getSidebarItems = (
...
-        | (getExperiment_experimentBySlug_primaryProbeSets | null)
-        | (getExperiment_experimentBySlug_secondaryProbeSets | null),
-    ) => (newMap[probeset!.slug] = probeset!.name),
-  );
-  return newMap;
+const outcomeToMapping = (outcomes: OutcomesList) => {
+  return outcomes.reduce((acc: { [key: string]: string }, outcome) => {
+    acc[outcome?.slug as string] = outcome?.friendlyName!;
+    return acc;
+  }, {});
```
lmorchard, line 49: When I tried adding the type import locally to `react-app-env.d.ts`, I saw some type errors hereabouts. Did some code golfing and came up with this:
```javascript
import { getConfig_nimbusConfig_outcomes } from "../../types/getConfig";

const outcomeToMapping = (outcomes: OutcomesList) =>
  outcomes
    .filter(
      (outcome): outcome is getConfig_nimbusConfig_outcomes =>
        !!outcome && !!outcome.slug,
    )
    .reduce(
      (acc, { slug, friendlyName }) => ({
        ...acc,
        [slug!]: friendlyName,
      }),
      {},
    );
```
Not sure if all those gymnastics to guard against nulls & undefineds will prevent future issues, but typescript seemed to like this better?

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -14,96 +15,105 @@ import TableMetricPrimary from "./TableMetricPrimary";
...
+          // Return to the experiment root/summary page
+          return "";
+        }
+      }}
+    >
+      {({ experiment, analysis }) => {
+        // For testing - users will be redirected if the analysis is unavailable
+        // before reaching this return, but tests reach this return and
+        // analysis.overall is expected to be an object (EXP-800)
+        if (analysisUnavailable(analysis)) return <></>;
```
lmorchard, line 41: Looks like this was moved from existing code, but this should maybe just be a `return` or a `return null`? Or a `return ""` like earlier in the file

lmorchard (approved):
LGTM, except for one type tangle I noticed. Probably worth a quick fix along with squashing commits

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
jodyheavener: Oh yeah this broke all the types being used in the declaration file. I'll move the outcome types to an importable file.

Screen Shot 2021-03-10 at 10 41 31 AM

LZoog:
@jodyheavener I can't r+ my own PR, but all changes LGTM, thanks for finishing this out!

Just waiting for @emtwo to r+, then we can squash & merge

emtwo (approved):
Looks great to me!

PR #4779, opened 2021-03-08
fix #4757 feat(nimbus): add outcomes to V6 API

### LINKED ISSUE #4757: Add outcomes to V6 API
We need to expose the new outcomes in the V6 API for jetstream to consume after [https://github.com/mozilla/nimbus-shared/pull/159](https://github.com/mozilla/nimbus-shared/pull/159) lands

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1041)
┆Issue Number: EXP-1041

lmorchard:
Because:

- We need to expose the new outcomes in the V6 API for jetstream to consume

This commit:

- adds the outcome field tp V6 API output to match nimbus-shared schema

- fixes a small bug where primary_outcomes was listed twice in model
  factories instead of defining secondary_outcomes

- updates to docs via make generate_docs

### DIFF
--- a/app/experimenter/experiments/api/v6/serializers.py
+++ b/app/experimenter/experiments/api/v6/serializers.py
@@ -64,6 +64,7 @@ class NimbusExperimentSerializer(serializers.ModelSerializer):
     isEnrollmentPaused = serializers.ReadOnlyField(source="is_paused")
     bucketConfig = NimbusBucketRangeSerializer(source="bucket_range")
     probeSets = serializers.SerializerMethodField()
+    outcomes = serializers.SerializerMethodField()
     branches = NimbusBranchSerializer(many=True)
     startDate = serializers.DateTimeField(source="start_date")
     endDate = serializers.DateTimeField(source="end_date")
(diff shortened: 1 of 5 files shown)


app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -114,6 +116,17 @@ def get_appId(self, obj):
...
+    def get_outcomes(self, obj):
+        prioritized_outcomes = (
+            ("primary", obj.primary_outcomes),
+            ("secondary", obj.secondary_outcomes),
+        )
+        return [
+            {"slug": slug, "priority": priority}
+            for (priority, outcomes) in prioritized_outcomes
+            for slug in outcomes
+        ]
```
jaredlockhart, line 128: Ha! Clever! I love it.

jaredlockhart (changes requested):
@tdsmith Added outcomes to the schema in [1.4.0](https://github.com/mozilla/nimbus-shared/blob/v1.4.0/types/experiments.ts#L64-L65) so we should also bump `mozilla-nimbus-shared` in here.  Technically this still passes because right now the validation layer is set to accept arbitrary superfluous keys, @k88hudson is gonna add a 'strict' validation mode at some point, but in the mean time we just have to remember to bump the version when we add keys like this here.  Otherwise looks perfect thank you @lmorchard !

lmorchard:
Rebased against main

jaredlockhart (approved):
Oh! Dependabot caught it XD Yep perfect ty 👍

PR #4783, opened 2021-03-09
Fix #4669 - feat(visualization) Add weekly results prior to the end of an experiment.

### LINKED ISSUE #4669: Users should be able to see weekly results in Experimenter prior to experiment completion
Right now, the results for an experiment are only shown at the end. We should be able to show results as soon as 1 week into the experiment since Jetstream produces results that soon.

This will require a bunch of changes in the UI to make it clear that these are partial, in-flight results.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-988)
┆Issue Number: EXP-988

emtwo:
Because:
* Users would like to look at their experiment results once it's available.

This commit:
* Makes weekly results visible 1 week after enrollment is complete.


emtwo:
Note: currently still in draft as I need to add in a warning about the visible data

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
```diff
@@ -170,6 +170,17 @@ const percentField = (
...
 
+const formattedAnalysisPointComparator = (
+  a: FormattedAnalysisPoint,
+  b: FormattedAnalysisPoint,
+) => {
+  if (!a.window_index || !b.window_index) {
+    return 0;
+  }
+  if (a.window_index < b.window_index) return -1;
+  if (a.window_index > b.window_index) return 1;
```
pdehaan, line 181: Do we need a catch-all `return 0;` so it doesn't return null/undefined?
Or something like `return b.window_index - a.window_index;` which might cover both `if..if..else` statements.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -55,23 +57,40 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => (
...
 
-          <h2 className="h5 mb-3" id="results-summary">
-            Results Summary
-          </h2>
-          <TableResults
-            primaryProbeSets={experiment.primaryProbeSets!}
-            results={analysis!}
-          />
+          <div id="results-summary">
+            <h2 className="h5 mb-3">Results Summary</h2>
```
pdehaan, line 77: Will this display the "Results Summary" heading if we don't have either `overall` or `weekly` results, or will we always have at least one of those in this view?

emtwo: We would always have at least one of those in this view since there is a check above for `analysisUnavailable()`. At this point we know analysis is available which means at least one of `overall` or `weekly` results is available.

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
emtwo: good catch, thank you, updated

jodyheavener (approved):
Couple small non-blocking comments but generally LGTM!

PR #4787, opened 2021-03-10
fix #4571 chore(nimbus): remove probesets

### LINKED ISSUE #4571: Deprecate orphan DB tables
EXP-925 changes reading probe set data from the DB to outcomes data from fs.

We no longer need these tables after EXP-925 is complete and they should be deprecated.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-927)
┆Issue Number: EXP-927

jaredlockhart:
Because

* Probe sets are no more

This commit

* Removes all models, logic, and tests around probesets
* Update V3 API to read outcomes instead of probe sets
* Preserve probeSets as an empty list in V6 API


emtwo (approved):
r+ for the visualization portions

lmorchard (approved):
Yup, LGTM too

PR #4792, opened 2021-03-10
fixes #4684 feat(project): modify KintoClient to cache collection response on get_collection calls

### LINKED ISSUE #4684: Make one call to collection instead of multiple in nimbus_check_kinto_push_queue
I just noticed we're calling the collection status twice per collection in nimbus_check_kinto_push_queue:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L81
https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L103

We should only be calling this once per invocation for two reasons:

- reduce unnecessary duplicate API calls
- prevent the case where the collection changes in between the two calls

Ideally we'd call the status once, check it against a possible set of cases, handle exactly one, and move on.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-998)
┆Issue Number: EXP-998

jodyheavener:
This PR updates `KintoClient` to call `_fetch_collection_data` first in certain methods, which will cache the `get_collection` response or use it if it's already cached.

@jaredlockhart this isn't exactly what we talked about in the comments below, but I like this approach because we don't need collection data any time the client is initialized, so this saves us requests.

~This PR updates the Kinto task `nimbus_check_kinto_push_queue` to only perform a call to Kinto one time and handle the collection data/statuses in the task.~

### DIFF
--- a/app/experimenter/kinto/client.py
+++ b/app/experimenter/kinto/client.py
@@ -15,6 +15,15 @@ class KintoClient:
             auth=(settings.KINTO_USER, settings.KINTO_PASS),
         )
         self.review = review
+        self.collection_data = None
+
+    def _fetch_collection_data(self):
+        self.collection_data = (
+            self.collection_data
+            or self.kinto_http_client.get_collection(
+                id=self.collection, bucket=settings.KINTO_BUCKET_WORKSPACE
+            )["data"]
+        )
 
     def _patch_collection(self):
         if self.review:
@@ -57,18 +66,13 @@ class KintoClient:
         self._patch_collection()
 
     def has_pending_review(self):
-        collection = self.kinto_http_client.get_collection(
-            id=self.collection, bucket=settings.KINTO_BUCKET_WORKSPACE
-        )
-        return collection["data"]["status"] == KINTO_REVIEW_STATUS
+        self._fetch_collection_data()
+        return self.collection_data["status"] == KINTO_REVIEW_STATUS
 
     def get_rejected_collection_data(self):
-        collection = self.kinto_http_client.get_collection(
-            id=self.collection, bucket=settings.KINTO_BUCKET_WORKSPACE
-        )
-
-        if collection["data"]["status"] == KINTO_REJECTED_STATUS:
-            return collection["data"]
+        self._fetch_collection_data()
+        if self.collection_data["status"] == KINTO_REJECTED_STATUS:
+            return self.collection_data
 
     def get_rejected_record(self):
         main_records = self.kinto_http_client.get_records(


app/experimenter/kinto/tasks.py
```diff
@@ -95,12 +99,12 @@ def nimbus_check_kinto_push_queue():
...
                 experiment,
                 get_kinto_user(),
-                message=f'Rejected: {rejected_collection_data["last_reviewer_comment"]}',
+                message=f'Rejected: {collection_data["last_reviewer_comment"]}',
             )
 
             kinto_client.rollback_changes()
 
-        if kinto_client.has_pending_review():
+        if collection_data["status"] == KINTO_REVIEW_STATUS:
```
jaredlockhart, line 107: If you're doing the check explicitly here can't you just get rid of the `has_pending_review` method altogether?

jodyheavener: If we also want to update [this](https://github.com/mozilla/experimenter/blob/4b149db457b3b8eaedb151d1d1755cf8f37782f6/app/experimenter/kinto/tasks.py#L294) then yes

jaredlockhart: Oh I seeeeeee.... the method is used elsewhere.... but if we try to use it here then it makes another call.  So we'd either have to remove the method and do the check explitily in the other spot the way we're doing here, or change this to cache the collection data on the client at init time, which is probably reasonable since within a single invocation of a task if it shifts during the middle of the task invocation we'd want the task to just fail hard and start again on its next invocation.

Which is basically what we're doing in the task you're updating, but then we kind of have two modes of using the client, either calling methods on it results in multiple collections calls, which I think we basically never want, or you explicitly pull in the collection data once and then just interact with it directly.

So that might be a better solution here, and then we can encapsulate all those status checks without having to expose the internals to the callers, and don't leave it up to the caller to manage how many calls get made to the collection.  What do you think?

jodyheavener: Sorry, just want to confirm: you’re saying we _should_ cache the collection data on init, and then update those methods to just access the cached data? And then leave the calls in the tasks untouched?

jaredlockhart: Yeah I think that seems right?  Then within a specific task it looks like


```py
client = KintoClient(collection=, bucket=)
if client.has_some_state():
    client.do_some_thing()
```

and you know you're only interacting with the snapshotted state at the time the client was initialized, and then we can hide things like the kinto status checks entirely within the client module.  Right?

jodyheavener: Yeah I like that better than inspecting the raw collection data in the tasks. Lets the client do it’s job.

app/experimenter/kinto/client.py
```diff
@@ -15,6 +15,15 @@ def __init__(self, collection, review=True):
...
         self.review = review
+        self.collection_data = None
+
+    def _fetch_collection_data(self):
+        self.collection_data = (
+            self.collection_data
+            or self.kinto_http_client.get_collection(
+                id=self.collection, bucket=settings.KINTO_BUCKET_WORKSPACE
+            )["data"]
+        )
```
jaredlockhart, line 26: Oh yeah you can do it this way or just call it right in __init__ 🤷‍♀️ either works

I guess doing it this way prevents calling it for cases where it's not needed.  Seems good!

jaredlockhart (approved):

app/experimenter/kinto/client.py
jaredlockhart: Also a very unlikely case here is if the data is `{}` it'll always refetch but I don't think that's likely, not worth worrying about for now.

jodyheavener: Yeah I noticed a handful of spots where we init a KintoClient but never need the collection data, so if the goal is to reduce API calls this seems 👌

jaredlockhart: Yep perfect very nice 👍

Issue #4680, opened 2021-02-23
Don't return early in nimbus_check_kinto_push_queue


jaredlockhart:
I just noticed when we check whether we should push any experiments to kinto, if one collection has a pending review we return early instead of continuing on to the next collection.  We should continue, not return.

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L105

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-996)
┆Issue Number: EXP-996

data-sync-user:
➤ Jared Kerim commented:

Let's split nimbus_check_kinto_push_queue into two tasks, a parent task that loops over each application and dispatches a child task that takes in an application and performs all the relevant checks for that one application, and then it CAN return early while also not blocking the other application updates.  And then we don't even have to test/handle the case where returning early from one prevents updating the other.

PR #4793, opened 2021-03-11
fixes #4686 feat(project): split nimbus_check_kinto_push_queue into multiple tasks

### LINKED ISSUE #4686: Split nimbus_check_kinto_push_queue into multiple tasks
The nimbus_check_kinto_push_queue task is becoming large and unruly, we should split it into multiple smaller tasks that are invoked by the parent task, which will help isolate logic and testing.

So something like

```py
    for application, collection in NimbusExperiment.KINTO_APPLICATION_COLLECTION.items():
        kinto_client = KintoClient(collection)

        if kinto_client.has_pending_review:
            continue
        elif kinto_client.has_rejected_review:
            nimbus_handle_rejected_review.delay()
        elif queued_launch_experiments.exists():
            nimbus_handle_launch_experiment.delay(queued_launch_experiments.first().id)
        elif queued_pause_experiments.exists():
            nimbus_handle_pause_experiment.delay(queued_pause_experiments.first().id)
        elif queued_end_experiments.exists():
            nimbus_handle_end_experiment.delay(queued_end_experiments.first().id)
```

so that we only have a single scheduled task that can initiate any change to remote settings (rather than having the pause update be its own scheduled task right now, for instance, that could potentially race against the other launch/end task) and we only initiate one change per collection per scheduled invocation.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1000)
┆Issue Number: EXP-1000

jodyheavener:
Fixes #4686

This PR breaks the Kinto task `nimbus_check_kinto_push_queue` into a handful of individual tasks that can be independently run. It also introduces a `NimbusExperimenterManager` for collecting groups of queued experiments.

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -16,6 +16,18 @@ from experimenter.experiments.constants import NimbusConstants
 from experimenter.projects.models import Project
 
 
+class NimbusExperimentManager(models.Manager):
+    def launch_queue(self, application):
+        return self.filter(status=NimbusExperiment.Status.REVIEW, application=application)
+
+    def end_queue(self, application):
+        return self.filter(
+            status=NimbusExperiment.Status.LIVE,
+            application=application,
+            is_end_requested=True,
+        )
+
+
 class NimbusExperiment(NimbusConstants, models.Model):
     owner = models.ForeignKey(
         get_user_model(),
@@ -72,6 +84,8 @@ class NimbusExperiment(NimbusConstants, models.Model):
     reference_branch = models.OneToOneField(
         "NimbusBranch", blank=True, null=True, on_delete=models.CASCADE
     )
+
+    objects = NimbusExperimentManager()
 
     class Meta:
         verbose_name = "Nimbus Experiment"
--- a/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
+++ b/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
@@ -19,6 +19,58 @@ from experimenter.experiments.tests.factories import (
 from experimenter.openidc.tests.factories import UserFactory
 
 
+class TestNimbusExperimentManager(TestCase):
+    def test_launch_queue_returns_queued_experiments_with_correct_application(self):
+        experiment1 = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.REVIEW,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.REVIEW,
+            application=NimbusExperiment.Application.FENIX,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        self.assertEqual(
+            list(
+                NimbusExperiment.objects.launch_queue(
+                    NimbusExperiment.Application.DESKTOP
+                )
+            ),
+            [experiment1],
+        )
+
+    def test_end_queue_returns_ending_experiments_with_correct_application(self):
+        experiment1 = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            is_end_requested=True,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            is_end_requested=True,
+            application=NimbusExperiment.Application.FENIX,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            is_end_requested=False,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.REVIEW,
+            is_end_requested=True,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        self.assertEqual(
+            list(
+                NimbusExperiment.objects.end_queue(NimbusExperiment.Application.DESKTOP)
+            ),
+            [experiment1],
+        )
+
+
 class TestNimbusExperiment(TestCase):
     def test_str(self):
         experiment = NimbusExperimentFactory.create(slug="experiment-slug")
--- a/app/experimenter/kinto/client.py
+++ b/app/experimenter/kinto/client.py
@@ -69,6 +69,10 @@ class KintoClient:
         self._fetch_collection_data()
         return self.collection_data["status"] == KINTO_REVIEW_STATUS
 
+    def has_rejection(self):
+        self._fetch_collection_data()
+        return self.collection_data["status"] == KINTO_REJECTED_STATUS
+
     def get_rejected_collection_data(self):
         self._fetch_collection_data()
         if self.collection_data["status"] == KINTO_REJECTED_STATUS:
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -78,54 +78,43 @@ def nimbus_check_kinto_push_queue():
     for application, collection in NimbusExperiment.KINTO_APPLICATION_COLLECTION.items():
         kinto_client = KintoClient(collection)
 
-        rejected_collection_data = kinto_client.get_rejected_collection_data()
-        if rejected_collection_data:
-            rejected_slug = kinto_client.get_rejected_record()
-            experiment = NimbusExperiment.objects.get(slug=rejected_slug)
-            if (
-                experiment.status == NimbusExperiment.Status.LIVE
-                and experiment.is_end_requested
-            ):
-                experiment.is_end_requested = False
-            else:
-                experiment.status = NimbusExperiment.Status.DRAFT
-
-            experiment.save()
-
-            generate_nimbus_changelog(
-                experiment,
-                get_kinto_user(),
-                message=f'Rejected: {rejected_collection_data["last_reviewer_comment"]}',
-            )
-
-            kinto_client.rollback_changes()
-
         if kinto_client.has_pending_review():
-            metrics.incr(f"check_kinto_push_queue.{collection}_pending_review")
             return
 
-        queued_experiments = NimbusExperiment.objects.filter(
-            status=NimbusExperiment.Status.REVIEW, application=application
-        )
-        end_requested_experiments = NimbusExperiment.objects.filter(
-            status=NimbusExperiment.Status.LIVE,
-            application=application,
-            is_end_requested=True,
-        )
-        if queued_experiments.exists():
-            nimbus_push_experiment_to_kinto.delay(queued_experiments.first().id)
-            metrics.incr(
-                f"check_kinto_push_queue.{collection}_queued_experiment_selected"
-            )
-        elif end_requested_experiments.exists():
-            nimbus_end_experiment_in_kinto.delay(end_requested_experiments.first().id)
-            metrics.incr(
-                f"check_kinto_push_queue.{collection}_end_requested_experiment_deleted"
-            )
-        else:
-            metrics.incr(f"check_kinto_push_queue.{collection}_no_experiments_queued")
+        if kinto_client.has_rejection():
+            handle_rejection(kinto_client)
+
+        if queued_launch_experiment := NimbusExperiment.objects.launch_queue(
+            application
+        ).first():
+            nimbus_push_experiment_to_kinto.delay(queued_launch_experiment.id)
+        elif queued_end_experiment := NimbusExperiment.objects.end_queue(
+            application
+        ).first():
+            nimbus_end_experiment_in_kinto.delay(queued_end_experiment.id)
 
     metrics.incr("check_kinto_push_queue.completed")
+
+
+def handle_rejection(kinto_client):
+    rejected_slug = kinto_client.get_rejected_record()
+    collection_data = kinto_client.get_rejected_collection_data()
+    experiment = NimbusExperiment.objects.get(slug=rejected_slug)
+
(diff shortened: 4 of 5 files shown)


jodyheavener:
@jaredlockhart needs lots of test changes, but is this generally inline with what you had in mind?

app/experimenter/kinto/tasks.py
```diff
@@ -112,22 +86,60 @@ def nimbus_check_kinto_push_queue():
...
+@app.task
+def nimbus_handle_launch_experiment(collection, experiment_id):
+    nimbus_push_experiment_to_kinto.delay(experiment_id)
+    metrics.incr(f"check_kinto_push_queue.{collection}_queued_experiment_selected")
+
+
+@app.task
+def nimbus_handle_end_experiment(collection, experiment_id):
+    nimbus_end_experiment_in_kinto.delay(experiment_id)
+    metrics.incr(f"check_kinto_push_queue.{collection}_end_requested_experiment_deleted")
```
jaredlockhart, line 140: Don't need these, just rename the other tasks

jaredlockhart:
Yeah exactly.  I have a couple other ideas too to simplify this.

Move these:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L107-L114

into manager methods (we can pair and i can explain, [example here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/legacy.py#L29-L45))

so that we can do something like

```py
if queued_launch_experiment := NimbusExperiment.objects.queued_launch_experiments():
    handle_launch_experiment.delay(queued_launch_experiment.id)
```

etc

jaredlockhart:
I'd like to basically have it look like

```py
@app.task
def update_kinto():
   for application, collection, etc:
        if kinto_client.has_rejection():
            handle_rejection.delay()
        elif queued_launch_experiment := NimbusExperiment.objects.queued_launch_experiments().first():
            handle_launch_experiment.delay(queued_launch_experiment.id)
        elif queued_pause_experiment := NimbusExperiment.objects.queued_pause_experiments().first():
            handle_pause_experiment.delay(queued_pause_experiment.id)
        elif queued_end_experiment := NimbusExperiment.objects.queued_end_experiments().first():
            handle_end_experiment.delay(queued_end_experiment.id)

def handle_this
def handle_that
etc
```

app/experimenter/kinto/tasks.py
jodyheavener: Can you elaborate?

app/experimenter/kinto/tasks.py
```diff
@@ -78,54 +78,46 @@ def nimbus_check_kinto_push_queue():
...
+            handle_rejection.delay(collection, rejected_collection_data)
+        elif queued_launch_experiment := NimbusExperiment.objects.launch_queue(
+            application
+        ).first():
+            nimbus_push_experiment_to_kinto.delay(queued_launch_experiment.id)
+        elif queued_pause_experiment := NimbusExperiment.objects.pause_queue(
+            application
+        ).first():
+            nimbus_end_experiment_in_kinto.delay(queued_pause_experiment.id)
+        elif queued_end_experiment := NimbusExperiment.objects.end_queue(
```
jodyheavener, line 93: flake8 is not happy with this syntax Screen Shot 2021-03-12 at 3 27 03 PM

app/experimenter/experiments/models/nimbus.py
```diff
@@ -73,6 +92,8 @@ class NimbusExperiment(NimbusConstants, models.Model):
         "NimbusBranch", blank=True, null=True, on_delete=models.CASCADE
     )
 
+    objects = ExperimentQueueManager()
```
jodyheavener, line 95: Is `ExperimentQueueManager` an appropriate name for this? Or should it just be `ExperimentManager`? Does it supersede the existing manager? Did some quick reading on it and I think that is true, but wanna be sure.

app/experimenter/kinto/tasks.py
```diff
@@ -78,54 +78,46 @@ def nimbus_check_kinto_push_queue():
...
-
-            generate_nimbus_changelog(
-                experiment,
-                get_kinto_user(),
-                message=f'Rejected: {rejected_collection_data["last_reviewer_comment"]}',
-            )
+        if kinto_client.has_pending_review():
+            continue
+        elif rejected_collection_data := kinto_client.get_rejected_collection_data():
+            handle_rejection.delay(collection, rejected_collection_data)
```
jodyheavener, line 84: In your example you weren't passing the rejected data through to the other task. Is this what you meant? Or did you intend to have kinto make another call in the new task?

jodyheavener (commented):
@jaredlockhart some more futzing around for you to take a peak at

app/experimenter/experiments/models/nimbus.py
jaredlockhart: So.... You can have multiple managers on a model, so like we could have 

```py
objects = ExperimentManager()
queued_objects = ExperimentQueueManager()
```

but we don't need multiple, so just one is fine, and if it's only one it's usually just called `<Model>Manager` so yeah `NimbusExperimentManager` is fine here.

app/experimenter/kinto/tasks.py
jaredlockhart: Huh that's funny, sometimes flake8 and black don't get along, black knows what to do so whatever black formats it to if flake8 is still complaining just disable the flake8 warning.

app/experimenter/kinto/tasks.py
jaredlockhart: Oh yeah, I mean we have to instantiate the client and get the rejection data inside the rejection task anyways so I don't think you need to pass it through, since there will always only ever be one pending rejection, so you can just say

```py
elif kinto_client.has_rejection()
```

and invoke the task and then find the rejection data from the collection inside the rejection task.


The other thing is.... In theory, in a single invocation of the parent task we could both clear the rejection out of the collection AND push a new change to it.  Structuring it like this will only clear the rejection out then end, and the next invocation would push something.

So... we might not want to make the rejection handling a separate task in which case it'd just be something like


```py
if kinto_client.has_pending_review():
    continue

if kinto_client.has_rejection():
    handle_rejection(kinto_client)

if queued_launch......
```


In *theory* we could make the rejection handling a separate task, AND do two tasks in a single invocation by using some celery mechanisms to dispatch multiple ordered tasks, but I don't think we need to get that fancy.  In theory we don't need to dispatch any subtasks from this parent task at all, but it's kind of a nice way to isolate things from a metrics/blocking/error handling perspective.

So yeah that's a long winded way of saying, try my example above of rejection handling just as a function call rather than a task, but for the rest of these yes looks great 👏 👏 👏

app/experimenter/experiments/models/nimbus.py
```diff
@@ -16,6 +16,20 @@
 from experimenter.projects.models import Project
 
 
+class NimbusExperimentManager(models.Manager):
+    def launch_queue(self, application):
+        return super().filter(
```
jaredlockhart, line 21: You don't actually need to call super here, you can just call `self.filter` or `self.all().filter`.  You only need to call `super` if you're overriding the same method and need to invoke the parent implementation of it.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -16,6 +16,20 @@
 from experimenter.projects.models import Project
 
 
+class NimbusExperimentManager(models.Manager):
```
jaredlockhart, line 19: Very nice 👍

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -19,6 +19,58 @@
...
+            NimbusExperiment.Status.REVIEW,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.REVIEW,
+            application=NimbusExperiment.Application.FENIX,
+        )
+        NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            application=NimbusExperiment.Application.DESKTOP,
```
jaredlockhart, line 34: Yep exactly the perfect way to test this 👏

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -19,6 +19,58 @@
...
+        self.assertEqual(
+            list(
+                NimbusExperiment.objects.launch_queue(
+                    NimbusExperiment.Application.DESKTOP
+                )
+            ),
+            [experiment1],
+        )
+
+    def test_end_queue_returns_ending_experiments_with_correct_application(self):
```
jaredlockhart, line 45: Perfect 💯

app/experimenter/kinto/client.py
```diff
@@ -69,6 +69,10 @@ def has_pending_review(self):
         self._fetch_collection_data()
         return self.collection_data["status"] == KINTO_REVIEW_STATUS
 
+    def has_rejection(self):
+        self._fetch_collection_data()
+        return self.collection_data["status"] == KINTO_REJECTED_STATUS
```
jaredlockhart, line 74: Beautiful 🌹

app/experimenter/kinto/tasks.py
```diff
@@ -78,54 +78,41 @@ def nimbus_check_kinto_push_queue():
...
-            if (
-                experiment.status == NimbusExperiment.Status.LIVE
-                and experiment.is_end_requested
-            ):
-                experiment.is_end_requested = False
-            else:
-                experiment.status = NimbusExperiment.Status.DRAFT
+        if kinto_client.has_pending_review():
+            continue
+        elif kinto_client.has_rejection():
```
jaredlockhart, line 83: This should be its own if clause, that way we can pull the rejection out of the collection AND send a new change to it in the same task invocation.

app/experimenter/kinto/tasks.py
```diff
@@ -78,54 +78,41 @@ def nimbus_check_kinto_push_queue():
...
+            handle_rejection(kinto_client)
+        elif queued_launch_experiment := NimbusExperiment.objects.launch_queue(
```
jaredlockhart, line 85: New if clause starts here

jaredlockhart (changes requested):
Very nice, this already looks much nicer thank you!  Small change to the task if ordering.  I wonder if we should have a test to capture that it handles the rejection AND launches a change....  What do you think?

app/experimenter/kinto/tasks.py
```diff
@@ -78,54 +78,41 @@ def nimbus_check_kinto_push_queue():
...
-            experiment = NimbusExperiment.objects.get(slug=rejected_slug)
-            if (
-                experiment.status == NimbusExperiment.Status.LIVE
-                and experiment.is_end_requested
-            ):
-                experiment.is_end_requested = False
-            else:
-                experiment.status = NimbusExperiment.Status.DRAFT
+        if kinto_client.has_pending_review():
+            continue
```
jaredlockhart, line 82: Oh also I'd leave this as return here, and switch it to continue with a test in https://github.com/mozilla/experimenter/issues/4680

app/experimenter/experiments/models/nimbus.py
```diff
@@ -16,6 +16,20 @@
...
+        return super().filter(
+            status=NimbusExperiment.Status.REVIEW, application=application
+        )
+
+    def end_queue(self, application):
+        return super().filter(
+            status=NimbusExperiment.Status.LIVE,
+            application=application,
+            is_end_requested=True,
+        )
```
jaredlockhart, line 30:
```py
def pause_queue(self, applicaiton):
    return self.filter(id__in=[e.id for e in self.all() if e.should_pause])
```

app/experimenter/kinto/tasks.py
```diff
@@ -78,54 +78,41 @@ def nimbus_check_kinto_push_queue():
...
+        elif kinto_client.has_rejection():
+            handle_rejection(kinto_client)
+        elif queued_launch_experiment := NimbusExperiment.objects.launch_queue(
+            application
+        ).first():
+            nimbus_push_experiment_to_kinto.delay(queued_launch_experiment.id)
+        elif queued_end_experiment := NimbusExperiment.objects.end_queue(
```
jaredlockhart, line 89: 1) Move the pause check into the manager like above
2) Add the pause queue check here and dispatch a pause task (from below, pad it out the same way the end task is padded)
3) You can delete the update pause in kinto task altogether

Rename the push to, end in, pause in tasks to something similar like `handle_whatever`

jodyheavener:
@jaredlockhart  This is ready to roll again with the exception that I can't figure out how to test that it handles the rejection and launches a change. I think I understand what I should be testing, but haven't been able to piece it together.

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -226,6 +226,59 @@ def test_check_live_end_requested_with_reject_review(self):
         self.assertFalse(latest_change.experiment_data["is_end_requested"])
         self.assertEqual(latest_change.message, f"Rejected: {rejection_message}")
 
+    def test_check_live_end_requested_with_reject_review_and_with_review_pushes(self):
```
jaredlockhart, line 229: I can't quite parse this test name....  Maybe another name or a docstring?

jaredlockhart (commented):
Yeah that test looks exactly right thank you!  What about the pause stuff are you still gonna move that around?

app/experimenter/kinto/tests/test_tasks.py
jodyheavener: Hah, fair enough, updated.

jodyheavener:
> What about the pause stuff are you still gonna move that around?

Yup! I have it more or less [ready to go](https://github.com/mozilla/experimenter/commit/22761f1c33fa0551aa2c5a40af41b4c81aeedb89), but I'm gonna open a new PR for it since it's out of scope and is large-ish

jaredlockhart (approved):
Okay cool then let's get this landed and keep movin forward, great job @jodyheavener this is already getting so much cleaner 🎉

PR #4808, opened 2021-03-12
chore(nimbus): use curl/unzip instead of gitsubmodule to track jetstream-config

jaredlockhart:
Because

* Using a git submodule will require that we always commit the latest commit hash which muddies the git history

This commit

* Uses curl/unzip to fetch the latest jetstream config archive and unpacks it into the outcomes folder
* Removes the gitsubmodule

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,3 +98,6 @@ app/coverage_html_report/
 node_modules
 
 google-credentials.json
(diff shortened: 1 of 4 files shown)


jodyheavener (approved):
ya dude, ya

PR #4822, opened 2021-03-15
fixes #4803 fix(nimbus): switch to explicitly-defined names for targeting dropdown

### LINKED ISSUE #4803: Change the targeting "Us Only" drop down option to "US Only"
From user interviews, I've noticed that when users click on the drop down menu for Targeting they often get confused by "Us Only" and read it as "us" rather than "US".

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-1065)
┆Issue Number: EXP-1065

lmorchard:
Because:

- From user interviews, I've noticed that when users click on the drop
  down menu for Targeting they often get confused by "Us Only" and read it
  as "us" rather than "US".

This commit:

- Switches from Django auto-conversion of member name to explicitly
  defined descriptive name


lmorchard:
Turns out the labels in the targeting dropdown were [an auto-conversion from an ID](https://docs.djangoproject.com/en/3.1/ref/models/fields/#enumeration-types), and not using the actual names we had defined.

I didn't literally change "Us Only" to "US Only" - instead I switched to using the "US users (en)" name we already had defined there. We can change that name if we really want "US Only".

Additionally, this changes the labels for all the other targeting options, too

Before:

Capture2

After:

Capture

lmorchard:
Also, this is going to fail lint since I forgot to `make code_format` before pushing. Will fix that right now...

jodyheavener (approved):
🤠

PR #4823, opened 2021-03-16
fixes #4685 feat(project): only put accepted experiments into draft on rejection

### LINKED ISSUE #4685: Handle RS rejections gracefully for all possible rejection cases
There are three times we make a review request to Remote Settings:

- On launch
- On pause
- On end

We currently handle the case where a launch is rejected and capture the rejection reason, but we don't handle the case where it's rejected on pause/end.

I think we should

1) Filter the launch rejection case to make sure we only attempt to roll back an experiment that's in the accepted state
2) In all other cases, throw out the rejection reason and put the collection back in the rollback state so that the pause/end tasks will just attempt the change again.  So then if someone rejects a pause/end (which should not be part of the workflow), the request will just come back on the next invocation.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-999)
┆Issue Number: EXP-999

jodyheavener:
Closes #4685

This work is slightly different from the issue description per my discussion with @jaredlockhart. This PR updates `handle_rejection` (kicked off by the `nimbus_check_kinto_push_queue` task) to:

- Add a condition to the elif that changes the experiment's status back to Draft only if the experiment was in an Accepted state to begin with
- Conditionally saves the experiment in the DB to avoid unnecessary writes

As we discussed, any reject will still 1) create a changelog, and 2) trigger a rollback

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -101,12 +101,16 @@ def handle_rejection(kinto_client):
     collection_data = kinto_client.get_rejected_collection_data()
     experiment = NimbusExperiment.objects.get(slug=rejected_slug)
 
+    has_changes = False
     if experiment.status == NimbusExperiment.Status.LIVE and experiment.is_end_requested:
         experiment.is_end_requested = False
-    else:
+        has_changes = True
+    elif experiment.status == NimbusExperiment.Status.ACCEPTED:
         experiment.status = NimbusExperiment.Status.DRAFT
+        has_changes = True
 
-    experiment.save()
+    if has_changes:
+        experiment.save()
 
     generate_nimbus_changelog(
         experiment,
--- a/app/experimenter/kinto/tests/test_tasks.py
+++ b/app/experimenter/kinto/tests/test_tasks.py
@@ -174,6 +174,55 @@ class TestCheckKintoPushQueue(MockKintoClientMixin, TestCase):
             ).exists()
         )
 
+    def test_check_with_reject_review_doesnt_draft_non_accepted(self):
+        experiment = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            is_end_requested=False,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+
+        self.mock_kinto_client.get_collection.side_effect = [
+            # Desktop responses
+            {
+                "data": {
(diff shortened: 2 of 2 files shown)


app/experimenter/kinto/tasks.py
```diff
@@ -101,12 +101,16 @@ def handle_rejection(kinto_client):
     collection_data = kinto_client.get_rejected_collection_data()
     experiment = NimbusExperiment.objects.get(slug=rejected_slug)
 
+    has_changes = False
```
jaredlockhart, line 104: 👍

jaredlockhart (commented):
Wait I'm confused, this is PRing into your other branch and not into main?

jodyheavener:
@jaredlockhart yeah because the first one isn’t merged and relies on it, this just keeps those commits out of this PR

jaredlockhart:
But why not just make this change against main and then rebase your other pr on top of it?

jodyheavener:
Wouldn't that add those commits to this PR because I'd be comparing against main? idk I've just always done it this way, don't see the harm?

jodyheavener:
So I chnanged it back to main, but because main doesn't have those commits they get included here and I wouldn't want this PR to merge them?

jaredlockhart:
No I mean make this change in isolation against main with 1 commit with the rejection logic where it is, then rebase and incorporate it into your other PR.  With continuous delivery every PR needs to be isolated and revertable, so if you PR this into your other PR and then we merge that into main and have to revert it we lose both changes.

jaredlockhart:
For reference: https://github.com/mozilla/experimenter/blob/main/contributing.md#continuous-deployment-process

jodyheavener:
Oh I see - I wasn't going to merge it into the other PR branch, I was going to wait until that branch was merged and then change this base branch back to main. Since this PR is not urgent it's just a shortcut to avoiding potential conflicts.

jaredlockhart:
Okay sure then just move this into a draft and we can merge the other one first?

jodyheavener:
That's fine, I just wanted to signify that the code technically is ready for review and I don't think you can r+ if it's in Draft.

jaredlockhart:
Yeah I'd rather wait until things are actually landable before approving

jaredlockhart (approved):
Works perfectly tytyty 👏 👏 👏

PR #4837, opened 2021-03-17
Fix #4807 - fix(visualization) Add weekly significance computation.

### LINKED ISSUE #4807: Add weekly significance computation
Right now, significance is only computed as an "overall" result. There should also be weekly significance

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1070)
┆Issue Number: EXP-1070

emtwo:
Because:
* Significance was only computed overall and incorrectly shown for weekly results too

This commit:
* Computes weekly significance to show for weekly results

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
@@ -188,6 +188,7 @@ const TableVisualizationRow: React.FC<{
   displayType?: DISPLAY_TYPE;
   branchComparison?: string;
   tooltip?: string;
+  window?: string;
 }> = ({
   metricKey,
   results,
@@ -196,6 +197,7 @@ const TableVisualizationRow: React.FC<{
   displayType,
   branchComparison,
   tooltip = "",
+  window = "overall",
 }) => {
   const { branch_data, is_control } = results;
   const metricData = branch_data[metricKey];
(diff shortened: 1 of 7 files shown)


jaredlockhart (approved):
Looks good 👍

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
```diff
@@ -217,14 +219,14 @@ const TableVisualizationRow: React.FC<{
...
     const metricDataList = metricData[branchComparison]["all"];
-    const significance = metricData["significance"];
 
     userCountsList.sort(formattedAnalysisPointComparator);
     metricDataList.sort(formattedAnalysisPointComparator);
 
     metricDataList.forEach((dataPoint: FormattedAnalysisPoint, i: number) => {
       const { lower, upper, point, count } = dataPoint;
       const userCountMetric = userCountsList[i]["point"];
+      const significance = metricData["significance"]?.[window][i + 1];
```
LZoog, line 229: Should we update any Storybook states in `TableVisualizationRow` to show what `overall` vs `weekly` looks like?

LZoog (approved):
LGTM, just one question about Storybook!

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.tsx
emtwo: the items in `TableVisualizationRow` don't look different if it's `weekly` vs. `overall`. But the difference can be seen in the storybook of `TableResultsWeekly`

Issue #4785, opened 2021-03-10
Outcomes should not be required in admin


emtwo:
They're currently required in the admin but some experiments don't need/have them.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1047)
┆Issue Number: EXP-1047

PR #4838, opened 2021-03-17
fixes #4585 fix(nimbus): Allow Results sub-links to navigate to Results page

### LINKED ISSUE #4585: Results sidebar loads on Design page when Results are available
https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/multi-stage-onboarding-set-default-v42-experiment/

Click the fancy new "Go to Design" navigation near the top. This should take you to https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/multi-stage-onboarding-set-default-v42-experiment/design

Multi-Stage_Onboarding__Set_Default_v4_2_Experiment___Experimenter

Note, clicking any of the side navigation under "Results" seems to be a noop and append an invalid anchor to the URL (ie: "#tagged_search_count"). My guts are telling me that it's because my page slug is still ./design and those sidebar links should be going to ./results#slugy-slug. A theory which might be confirmed by the fact I can click on that "Results" heading to go to the ./results page, and then clicking on the other subnav seems to work.

Multi-Stage_Onboarding__Set_Default_v4_2_Experiment_–_Design___Experimenter

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-936)
┆Issue Number: EXP-936

LZoog:
Instead of linking to `currentpage#id` we now link to `/results#id`.

Because of a bug requiring Outcomes in the BE admin (#4785), I wasn't able to load Results data locally to double check this works as expected. This approach should fix the problem though and it LGTM in Storybook. If it doesn't fix the issue in prod we can reopen, I don't think we should block on #4785.

EDIT: We have a PR up for #4785 with #4847! I'll wait to merge this until I confirm things work as expected.

fixes #4585

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
@@ -1,5 +1,4 @@
 import { RouteComponentProps, useParams } from "@reach/router";
-import classnames from "classnames";
 import React from "react";
 import Col from "react-bootstrap/Col";
 import Container from "react-bootstrap/Container";
@@ -20,27 +19,11 @@ import { LinkNav } from "../LinkNav";
 import { ReactComponent as BarChart } from "./bar-chart.svg";
 
 export const RESULTS_LOADING_TEXT = "Checking results availability...";
-const baseLinkClass = "d-block inherit-color mb-2";
 
-const getSidebarItems = (
-  metrics: { [metric: string]: string },
-  title: string,
-) => {
-  const sidebarItems = Object.keys(metrics).map((sidebarKey) => (
-    <a
-      href={`#${sidebarKey}`}
-      key={sidebarKey}
-      className={classnames(baseLinkClass, "font-weight-normal ml-3")}
-    >
-      {metrics[sidebarKey]}
-    </a>
-  ));
-  sidebarItems.unshift(
-    <p key={title} className="mb-2">
-      {title}
-    </p>,
-  );
-  return sidebarItems;
+const analysisNotRequiredLinkProps = {
+  storiesOf: "pages/Results",
+  textColor: "inherit-color",
+  className: "mx-1 mb-2 ml-3",
 };
 
 const outcomeToMapping = (outcomes: OutcomesList) => {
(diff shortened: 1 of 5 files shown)


emtwo (approved):
lgtm pending confirmation that it works as expected locally. Thanks!

LZoog:
Reverted this back to a draft. The solution wasn't working as expected and I did a little refactoring and things work now but `Scrollspy` isn't playing nicely.

LZoog:
OK so, this one ended up causing some unexpected headaches due to the scrollspy package and I left a couple of comments in the code about it.

We will probably want a follow up issue that I can make when this is r+. The navigation on the Summary page now takes you to the Results page which fixes the bug, but the user is not scrolled down to the section of the Results page. I thought this was a router issue at first, but I'm pretty sure it's because we reload the experiment data on each page so the `#id` gets lost. I think we can fix this by setting a `ref` on each section and checking what's in the URL after the experiment loaded and then `ref.current.scrollIntoView()` on that element.

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
```diff
@@ -91,18 +76,141 @@ export const AppLayoutSidebarLocked = ({
...
+    return sidebarItems;
+  };
+
   const sidebarKeys = [
     "monitoring",
     "overview",
-    "results-summary",
-    "primary-metrics",
+    "results_summary",
+    "primary_metrics",
```
LZoog, line 124: Ah @emtwo I could revert these if preferred, but I don't mind leaving them since the data is `snake_case` and could match better this way. I created a helper function for these when I tried DRY'ing all this up that was useful with all of these as snake... until I discovered how wacky Scrollspy is.

emtwo:
@LZoog I looked through it again and tried it out locally. Looks great! Thanks for your patience with this!!

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLocked/index.tsx
emtwo: this wfm

LZoog:
Follow up: EXP-1120 / #4912

PR #4839, opened 2021-03-17
fixes #4624 feat(nimbus): allow audience page select boxes to not select anything

### LINKED ISSUE #4624: Can't reset enum-based select fields once set
Once you choose and save an experiment channel, FF min version, or targeting config, you cannot only then change it to another value - you cannot unset it.

Steps to reproduce:

* Create an experiment
* Choose a Channel (or FF Min Version, or Targeting Config, either will work) and save
* Now try to set the dropdown back to "Select..." and save.
* You will get the "an error occurred while submitting" message. If you inspect the response message it'll indicate that it was expecting a value from the enum.

This is pretty low stakes, but considering all fields are optional until the very end and it is confusing to receive an error message for this we should consider fixing it.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-957)
┆Issue Number: EXP-957

jodyheavener:
Closes #4624

This PR updates the select dropdowns on the Audience page so that they can all select "nothing" (that is, return the selection to the "Select..." item).

Since the backend requires one of an enum value be chosen and each enum has a "no selection" value, just default to that value any time "no" value is provided.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -192,6 +192,28 @@ describe("FormAudience", () => {
     }
   });
 
+  it("does not require a selection for channel, ff min version, targeting config (EXP-957)", async () => {
+    const onSubmit = jest.fn();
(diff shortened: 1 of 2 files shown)


jodyheavener:
Wait... we should just remove the "Select..." option instead.

PR #4840, opened 2021-03-17
fixes #4841 feat(project): move pausing experiment check into nimbus_check_kinto_push_queue

### LINKED ISSUE #4841: Combine nimbus_update_paused_experiments_in_kinto and nimbus_check_kinto_push_queue
Spun off from EXP-1000, we should combine these two tasks.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1079)
┆Issue Number: EXP-1079

jodyheavener:
Closes #4841

Spun off from #4686, these two tasks both iterate over the collections, so @jaredlockhart suggested we combine them.

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -20,6 +20,16 @@ class NimbusExperimentManager(models.Manager):
     def launch_queue(self, application):
         return self.filter(status=NimbusExperiment.Status.REVIEW, application=application)
 
+    def pause_queue(self, application):
+        return self.filter(
+            status=NimbusExperiment.Status.LIVE,
+            is_paused=False,
+            application=application,
+            id__in=[
+                experiment.id for experiment in self.all() if experiment.should_pause
+            ],
+        )
+
     def end_queue(self, application):
         return self.filter(
             status=NimbusExperiment.Status.LIVE,
--- a/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
+++ b/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
@@ -70,6 +70,53 @@ class TestNimbusExperimentManager(TestCase):
             [experiment1],
         )
 
+    def test_pause_queue_returns_experiments_that_should_pause_by_application(self):
+        def rewind_launch(experiment):
+            launch_change = experiment.changes.get(
+                old_status=NimbusExperiment.Status.ACCEPTED,
+                new_status=NimbusExperiment.Status.LIVE,
+            )
+            launch_change.changed_on = datetime.datetime.now() - datetime.timedelta(
+                days=11
+            )
+            launch_change.save()
+
+        # Should end, with the correct application
+        experiment1 = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            is_paused=False,
+            proposed_enrollment=10,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
+        rewind_launch(experiment1)
+        # Should end, but wrong application
+        experiment2 = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            proposed_enrollment=10,
+            application=NimbusExperiment.Application.FENIX,
+        )
+        rewind_launch(experiment2)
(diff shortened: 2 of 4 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -20,6 +20,14 @@ class NimbusExperimentManager(models.Manager):
     def launch_queue(self, application):
         return self.filter(status=NimbusExperiment.Status.REVIEW, application=application)
 
+    def pause_queue(self, application):
+        return self.filter(
+            id__in=[
+                experiment.id for experiment in self.all() if experiment.should_pause
+            ],
+            application=application,
```
jaredlockhart, line 28: Oh yeah we also gotta filter for `status` and `is_paused` here, from the previous task:

```py
        live_experiments = NimbusExperiment.objects.filter(
            status=NimbusExperiment.Status.LIVE,
            application=application,
            is_paused=False,
        )
```

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -70,6 +70,44 @@ def test_end_queue_returns_ending_experiments_with_correct_application(self):
             [experiment1],
         )
 
+    def test_pause_queue_returns_experiments_that_should_pause_by_application(self):
```
jaredlockhart, line 73: And then update this with `status` and `is_paused`

app/experimenter/kinto/tasks.py
```diff
@@ -92,6 +92,12 @@ def nimbus_check_kinto_push_queue():
             application
         ).first():
             nimbus_end_experiment_in_kinto.delay(queued_end_experiment.id)
+        elif queued_pause_experiment := NimbusExperiment.objects.pause_queue(
+            application
+        ).first():
+            records = {r["id"]: r for r in kinto_client.get_main_records()}
+            record = records[queued_pause_experiment.slug]
```
jaredlockhart, line 99: Oh I would do all this in the task and just pass the experiment id in to make this more consistent with the other task signatures

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -320,6 +326,29 @@ def test_check_experiment_pushes_experiment_before_ending_experiment(
...
         self.mock_end_task.assert_called_with(experiment_2.id)
 
+    def test_check_experiment_that_should_pause_does_pause(
+        self,
+    ):
+        experiment = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.LIVE,
+            proposed_enrollment=10,
+            application=NimbusExperiment.Application.DESKTOP,
+        )
```
jaredlockhart, line 336: And `is_paused=False` here too

jaredlockhart (changes requested):
Yep looking really good, just gotta factor in the `is_paused` bool, and I think move the record lookup into the task.  Great work @jodyheavener 🎉

app/experimenter/kinto/tests/test_tasks.py
jaredlockhart: @jodyheavener bump?

jodyheavener: oh whoops

jodyheavener: what shouldn't that test be failing now?

jodyheavener: Oh I guess not if is_paused defaults to False

jodyheavener: @jaredlockhart updated

jaredlockhart: Yeah it defaults to false but it's just a good practice to specify all the priors for a test to pass explicitly so if we ever change the factories the test won't suddenly fail.  Thnx!

jaredlockhart (approved):
Looks great! Hurricane of PRS from @jodyheavener !! 🔥 🔥 🔥

app/experimenter/kinto/tests/test_tasks.py
jodyheavener: Agree, good call

PR #4853, opened 2021-03-17
fixes #4680 feat(nimbus): add new task that dispatches to kinto queue check by application

### LINKED ISSUE #4680: Don't return early in nimbus_check_kinto_push_queue

jodyheavener:
Closes #4680

Per @jaredlockhart's comment:

> Let's split nimbus_check_kinto_push_queue into two tasks, a parent task that loops over each application and dispatches a child task that takes in an application and performs all the relevant checks for that one application, and then it CAN return early while also not blocking the other application updates.  And then we don't even have to test/handle the case where returning early from one prevents updating the other.

I _think_ I've done that here. This change also added a lot of newlines/characters.. if you want me to change task names or constantize things like `NimbusExperiment.Application.DESKTOP` in the tests for example I'm game.

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -62,42 +62,53 @@ def nimbus_push_experiment_to_kinto(experiment_id):
 @metrics.timer_decorator("check_kinto_push_queue")
 def nimbus_check_kinto_push_queue():
     """
+    A scheduled task that passes each application to a new scheduled
+    task for working with kinto
+    """
+    for application in NimbusExperiment.KINTO_APPLICATION_COLLECTION:
+        nimbus_check_kinto_push_queue_by_application.delay(application)
+
+
+@app.task
+@metrics.timer_decorator("check_kinto_push_queue_by_application")
+def nimbus_check_kinto_push_queue_by_application(application):
+    """
     Because kinto has a restriction that it can only have a single pending review, this
     task brokers the queue of all experiments ready to be pushed to kinto and ensures
     that only a single experiment is ever in review.
 
     A scheduled task that
     - Checks the kinto collection for a single rejected experiment from a previous push
-      - If one exists, pull it out of the collection and mark it as rejected
+       - If one exists, pull it out of the collection and mark it as rejected
     - Checks if there is still a pending review and if so, aborts
     - Gets the list of all experiments ready to be pushed to kinto and pushes the first
       one
+    - Checks for experiments that should be paused but are not paused in the kinto
+      collection and marks them as paused and updates the record in the collection.
     """
-    metrics.incr("check_kinto_push_queue.started")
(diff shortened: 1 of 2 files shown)


app/experimenter/kinto/tasks.py
```diff
@@ -61,43 +61,54 @@ def nimbus_push_experiment_to_kinto(experiment_id):
...
-      - If one exists, pull it out of the collection and mark it as rejected
+       - If one exists, pull it out of the collection and mark it as rejected
     - Checks if there is still a pending review and if so, aborts
     - Gets the list of all experiments ready to be pushed to kinto and pushes the first
       one
+    - Checks for experiments that should be paused but are not paused in the kinto
+      collection and marks them as paused and updates the record in the collection.
     """
-    metrics.incr("check_kinto_push_queue.started")
+    metrics.incr("check_kinto_push_queue_by_application.started")
```
jaredlockhart, line 89: I would include the application in the metrics

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -73,6 +73,29 @@ def test_push_experiment_to_kinto_reraises_exception(self):
...
+
+    def test_dispatches_check_push_queue(self):
+        mock_application_collections = {
+            "FOO": "foo",
+            "BAR": "bar",
+        }
+
+        with mock.patch.object(
+            NimbusExperiment, "KINTO_APPLICATION_COLLECTION", mock_application_collections
+        ):
```
jaredlockhart, line 92: Do you need to mock this out in the test?

jaredlockhart (changes requested):
Yes nice and clean this is really shaping up thank you @jodyheavener !

app/experimenter/kinto/tests/test_tasks.py
jodyheavener: @jaredlockhart I think yes since the task doesn't care which applications it uses, only that it does something with them, and with this we can freely modify the applications while being confident the task is still doing its job.

jodyheavener: @jaredlockhart nvm I just realized how pointless is

jodyheavener: Classic case of me being excited that I learned something and looking for any opportunity to use it😜

jaredlockhart (approved):
👏 👏 👏

PR #4855, opened 2021-03-17
chore(docs): Add link to Nimbus Documentation Hub to README

LZoog:
This link did not feel right in the "Important Links" table because those seem developer-specific for _this_ repo and wouldn't stand out for a user looking for this link. Do we like the table for that section, anyway? 🤔 I can file a follow up issue if needed there, this doesn't have an associated issue but I thought we could use the cross-link now even if we want to tweak this section later.


k88hudson (approved):
🌩️

Issue #4799, opened 2021-03-11
Create new "timed out" component to signal that experiment owner should re-request review


data-sync-user:
If a request for experiment approval goes too long without a reviewer approving it, it times out and reverts to draft status.

Create a new component to signal this to the experiment owner.

Use a feature flag to demo the component in Storybook while leaving it disabled in production rendering until final integration of FE / BE. Documentation on how to do this is TBD with EXP-1063, but that issue has some links to pointers & examples.

Figma mockups: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=217334%3A6

!Capture.PNG|thumbnail!

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1061)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15531/Capture.PNG">Capture.PNG</a>
┆Issue Number: EXP-1061

data-sync-user:
➤ Jody Heavener commented:

[~jkerim@mozilla.com] [~lorchard@mozilla.com] what is the frontend using to determine this state? publish_status = idle?

jaredlockhart:
@jodyheavener we'll have to add an additional property to V5 API to look up the timeout in the changelogs which the UI can then use to pouplate the component.  Same for review rejection.

PR #4865, opened 2021-03-19
fix #4849 feat(nimbus): add publish_status to nimbus experiment

### LINKED ISSUE #4849: Add new publish status to NimbusExperiment
Add a new field to NimbusExperiment called `publish_status` as a charfield with an enum in constants with the following choices:

* idle
* review
* approved
* waiting

And add `old_publish_status` `new_publish_status` to NimbusChangeLog and populate them in `generate_changelog`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1085)
┆Issue Number: EXP-1085

jaredlockhart:
Because

* With the new review flow we need to track the review/publish status for an experiment separately from its lifecycle status

This commit

* Adds publish_status constants
* Adds a publish_status field to NimbusExperiment
* Tracks old and new publish status on the changelog


jodyheavener (commented):
@jaredlockhart general look-through makes sense. Is there a way I should be manually testing this at this point?

Couple other questions:

- Do the lifecycle and publish states not care about each other at all? As in, there's no combination of publish/lifecycle that should never occur?
- Asked [here](https://github.com/mozilla/experimenter/issues/4799#issuecomment-802050789), but I'm curious if this needs to account for when a review has timed out?

jaredlockhart:
> @jaredlockhart general look-through makes sense. Is there a way I should be manually testing this at this point?
> 
I don't think there's any need to manually test this PR.

> Couple other questions:
> 
>     * Do the lifecycle and publish states not care about each other at all? As in, there's no combination of publish/lifecycle that should never occur?

That will be handled in https://github.com/mozilla/experimenter/issues/4866
> 
>     * Asked [here](https://github.com/mozilla/experimenter/issues/4799#issuecomment-802050789), but I'm curious if this needs to account for when a review has timed out?

I'm filing a followup for that.

jodyheavener (approved):
🤠

PR #4847, opened 2021-03-17
fix #4785 chore(nimbus): make admin fields optional

jaredlockhart:
Because

* All admin fields for a newly created experiment should be optional
* Default empty fields being required makes using the admin cumbersome

This commit

* Makes all fields that have default values optional in the admin
* Updates tests

PR #4884, opened 2021-03-22
fix(project): include publish_status in admin form tests

jodyheavener:
Looks like this is failing main, and I assume publish_status is required, so I just updated the tests that needed it to include it.


lmorchard (approved):

jaredlockhart:
What in the heck how did that pass CI?

jaredlockhart:
Wow it is broken on main....how did that land....

jaredlockhart:
Oh I see what happened.  These were both merged without being rebased first:

https://github.com/mozilla/experimenter/pull/4865
https://github.com/mozilla/experimenter/pull/4847

so they only broke in main but not together.  Interesting, good to know.  Thanks for catching!

PR #4887, opened 2021-03-22
fixes #4866 feat(project): validate experiment changes and transitions relating to publish_status

### LINKED ISSUE #4866: Validate Publish Status in V5 API
* Generalize the StatusRestrictionMixin to work for both status and publish_status with the following restrictions
  * publish_status can only be mutated during draft/preview/live
  * experiments can not be modified in any publish_status other than idle
  * publish_status can transition (from the V5 API)
    * Idle > Review
    * Review > Idle
    * Review > Approved
  * Split out the tests against the StatusRestrictionMixin into a separate test class
  * A single integration test in the NimbusExperimentSerializer test should be sufficient to make sure status/publish_status are included

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1093)
┆Issue Number: EXP-1093

jodyheavener:
Closes #4866

This PR adds validates experiment changes and transitions relating to an experiment's publish_status. Refer to #4866 for the rules I followed.

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -210,35 +210,20 @@ class NimbusStatusValidationMixin:
         }
 
         if self.instance:
-            status = self.instance.status
-            check_status_only = True
-
-            for status_field, statuses in restrictive_statuses.items():
-                instance_value = getattr(self.instance, status_field)
-                if instance_value in statuses:
-                    if set(data.keys()) == {status_field}:
-                        check_status_only = False
-                    else:
-                        raise serializers.ValidationError(
-                            {
-                                "experiment": [
-                                    f"Nimbus Experiment has {status_field} "
-                                    f"'{instance_value}', only {status_field} "
-                                    "can be changed."
-                                ]
-                            }
-                        )
-
-            if check_status_only and status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
-                required_statuses = ", ".join(NimbusConstants.STATUS_ALLOWS_UPDATE)
-                raise serializers.ValidationError(
-                    {
-                        "experiment": [
-                            f"Nimbus Experiment has status '{status}', but can only "
-                            f"be changed when set to '{required_statuses}'."
-                        ]
-                    }
-                )
+            for status_field, restricted_statuses in restrictive_statuses.items():
+                current_status = getattr(self.instance, status_field)
+                is_locked = current_status in restricted_statuses
+                is_modifying_other_fields = set(data.keys()) != {status_field}
+                if is_locked and is_modifying_other_fields:
+                    raise serializers.ValidationError(
+                        {
+                            "experiment": [
+                                f"Nimbus Experiment has {status_field} "
+                                f"'{current_status}', only {status_field} "
+                                "can be changed."
+                            ]
+                        }
+                    )
 
         return data
 
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -175,61 +175,6 @@ class NimbusExperimentBranchMixin:
         return experiment
 
 
-class NimbusStatusRestrictionMixin:
-    ALLOWS_STATUS_CHANGE_ONLY = (NimbusExperiment.Status.PREVIEW,)
-    ALLOWS_UPDATE = (NimbusExperiment.Status.DRAFT,)
-    VALID_STATUS_TRANSITIONS = {
-        NimbusExperiment.Status.DRAFT: (
-            NimbusExperiment.Status.PREVIEW,
-            NimbusExperiment.Status.REVIEW,
-        ),
-        NimbusExperiment.Status.PREVIEW: (
-            NimbusExperiment.Status.DRAFT,
-            NimbusExperiment.Status.REVIEW,
-        ),
-    }
-
-    def validate(self, data):
-        data = super().validate(data)
-        if self.instance:
-            status = self.instance.status
-            if status in self.ALLOWS_STATUS_CHANGE_ONLY:
-                if set(data.keys()) != {"status"}:
-                    raise serializers.ValidationError(
-                        {
-                            "experiment": [
-                                f"Nimbus Experiment has status '{status}', only "
-                                "status can be changed."
-                            ]
-                        }
-                    )
-            elif status not in self.ALLOWS_UPDATE:
-                required_statuses = ", ".join(self.ALLOWS_UPDATE)
-                raise serializers.ValidationError(
-                    {
-                        "experiment": [
-                            f"Nimbus Experiment has status '{status}', but can only "
-                            f"be changed when set to '{required_statuses}'."
-                        ]
-                    }
-                )
-
-        return data
-
-    def validate_status(self, value):
-        if (
-            self.instance
-            and value != self.instance.status
-            and value not in self.VALID_STATUS_TRANSITIONS.get(self.instance.status, ())
-        ):
-            status = self.instance.status
-            raise serializers.ValidationError(
-                f"Nimbus Experiment status cannot transition from {status} to {value}."
-            )
-
-        return value
-
-
 class NimbusDocumentationLinkSerializer(serializers.ModelSerializer):
     class Meta:
         model = NimbusDocumentationLink
@@ -255,10 +200,74 @@ class NimbusExperimentDocumentationLinkMixin:
         return experiment
 
 
+class NimbusStatusValidationMixin:
+    def validate(self, data):
+        data = super().validate(data)
+
+        restrictive_statuses = {
+            "status": NimbusConstants.STATUS_CHANGE_ONLY,
+            "publish_status": NimbusConstants.PUBLISH_STATUS_CHANGE_ONLY,
+        }
+
+        if self.instance:
+            status = self.instance.status
+
+            for status_field, statuses in restrictive_statuses.items():
+                instance_value = getattr(self.instance, status_field)
+                if instance_value in statuses:
+                    if set(data.keys()) == {status_field}:
+                        return data
+                    else:
+                        raise serializers.ValidationError(
+                            {
+                                "experiment": [
+                                    f"Nimbus Experiment has {status_field} "
+                                    f"'{instance_value}', only {status_field} "
+                                    "can be changed."
+                                ]
+                            }
+                        )
+
+            if status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
+                required_statuses = ", ".join(NimbusConstants.STATUS_ALLOWS_UPDATE)
+                raise serializers.ValidationError(
+                    {
+                        "experiment": [
+                            f"Nimbus Experiment has status '{status}', but can only "
+                            f"be changed when set to '{required_statuses}'."
+                        ]
+                    }
+                )
+
+        return data
+
+
+class NimbusStatusTransitionValidator:
+    requires_context = True
+
+    def __init__(self, transitions):
+        self.transitions = transitions
+
+    def __call__(self, value, serializer_field):
+        field_name = serializer_field.source_attrs[-1]
+        instance = getattr(serializer_field.parent, "instance", None)
+
+        if instance and value:
+            instance_value = getattr(instance, field_name)
+
+            if value != instance_value and value not in self.transitions.get(
+                instance_value, ()
+            ):
+                raise serializers.ValidationError(
+                    f"Nimbus Experiment {field_name} cannot transition "
+                    f"from {instance_value} to {value}."
+                )
+
+
 class NimbusExperimentSerializer(
     NimbusExperimentBranchMixin,
+    NimbusStatusValidationMixin,
     NimbusExperimentDocumentationLinkMixin,
-    NimbusStatusRestrictionMixin,
     serializers.ModelSerializer,
 ):
     name = serializers.CharField(
@@ -294,8 +303,23 @@ class NimbusExperimentSerializer(
     population_percent = serializers.DecimalField(
         7, 4, min_value=0.0, max_value=100.0, required=False
     )
+    status = serializers.ChoiceField(
+        choices=NimbusExperiment.Status.choices,
+        required=False,
+        validators=[
+            NimbusStatusTransitionValidator(
+                transitions=NimbusConstants.VALID_STATUS_TRANSITIONS,
+            )
+        ],
+    )
     publish_status = serializers.ChoiceField(
-        choices=NimbusExperiment.PublishStatus.choices, required=False
+        choices=NimbusExperiment.PublishStatus.choices,
+        required=False,
+        validators=[
+            NimbusStatusTransitionValidator(
+                transitions=NimbusConstants.VALID_PUBLISH_STATUS_TRANSITIONS
+            )
+        ],
     )
 
     class Meta:
@@ -450,9 +474,7 @@ class NimbusExperimentSerializer(
             return experiment
 
 
-class NimbusReadyForReviewSerializer(
-    NimbusStatusRestrictionMixin, serializers.ModelSerializer
-):
+class NimbusReadyForReviewSerializer(serializers.ModelSerializer):
     public_description = serializers.CharField(required=True)
     proposed_duration = serializers.IntegerField(required=True, min_value=1)
     proposed_enrollment = serializers.IntegerField(required=True, min_value=1)
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -211,12 +211,13 @@ class NimbusStatusValidationMixin:
 
         if self.instance:
             status = self.instance.status
+            check_status_only = True
 
             for status_field, statuses in restrictive_statuses.items():
                 instance_value = getattr(self.instance, status_field)
                 if instance_value in statuses:
                     if set(data.keys()) == {status_field}:
-                        return data
+                        check_status_only = False
                     else:
                         raise serializers.ValidationError(
                             {
@@ -228,7 +229,7 @@ class NimbusStatusValidationMixin:
                             }
                         )
 
-            if status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
+            if check_status_only and status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
                 required_statuses = ", ".join(NimbusConstants.STATUS_ALLOWS_UPDATE)
                 raise serializers.ValidationError(
                     {
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -205,14 +205,14 @@ class NimbusStatusValidationMixin:
         data = super().validate(data)
 
         restrictive_statuses = {
-            "status": NimbusConstants.STATUS_CHANGE_ONLY,
-            "publish_status": NimbusConstants.PUBLISH_STATUS_CHANGE_ONLY,
+            "status": NimbusConstants.STATUS_ALLOWS_UPDATE,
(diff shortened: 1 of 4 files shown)


app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -455,6 +455,59 @@ def test_update_experiment_audience_error(self):
             },
         )
 
+    def test_end_experiment_in_kinto(self):
```
jodyheavener, line 458: I don't know why this is diffing this way 😒

lmorchard:
Heh, I was just looking through this PR and about to comment that the experiment vs publish statuses seemed a bit confusing mingled together - and then in comes a force-push that took care of that. Stop spying on my brain

lmorchard (dismissed):
Makes sense to me, probably just that one flake8 complaint keeping CI from going green

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -522,50 +581,59 @@ def test_update_experiment_publish_status(self):
...
+            },
+        )
 
-        experiment = NimbusExperiment.objects.get(id=experiment.id)
-        self.assertEqual(experiment.is_end_requested, True)
-        latest_change = experiment.changes.order_by("-changed_on").first()
-        self.assertEqual(latest_change.experiment_data["is_end_requested"], True)
-
-    def test_end_experiment_in_kinto_fails_with_nonlive_status(self):
+    def test_update_publish_status_from_approved_to_review_error(self):
```
lmorchard, line 612: Oh wait, hmm: I just got this failing locally, so maybe more than a flake8 thing

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -451,7 +498,9 @@ def save(self, *args, **kwargs):
 
 
 class NimbusReadyForReviewSerializer(
-    NimbusStatusRestrictionMixin, serializers.ModelSerializer
+    StatusValidationMixin,
+    PublishStatusValidationMixin,
```
jaredlockhart, line 502: Wait isn't this serializer read only?  Why does it need the status validators?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -191,43 +217,63 @@ class NimbusStatusRestrictionMixin:
...
+                raise serializers.ValidationError(
+                    {
+                        "experiment": [
+                            f"Nimbus Experiment has publish status {publish_status} but "
+                            f"can only be changed with {required_publish_statuses}."
+                        ]
+                    }
+                )
+
+        return data
```
jaredlockhart, line 276: Isn't this case the same in both as well?

jaredlockhart:
See if you can [use this](https://www.django-rest-framework.org/api-guide/validators/#class-based) rather than making them mixins.  If you make them parameterized class based validators you should be able to pass in all the parameters rather than making them class attributes.

jaredlockhart:
So then it would look roughly like

```py
class StatusValidator:
    def __init__(self, field_name, allowed_changes, valid_transitions, etc...):
        self.field_name = field_name
        etc....

    def __call__(self, value):
        if whatever:
            raise ValidationError


class NimbusExperimentSerializer(....):
    ....
    status = serializers.CharField(..., validators=[StatusValidator(field_name='status', allowed_whatever=....)])
    publish_status = serializers.CharField(..., validators=[StatusValidator(field_name='publish_status', allowed_whatever=...)])
```

lmorchard:
Thing is, though, these shouldn't just be validating status field. They also control whether any other property can be changed based on which status the experiment is in.

app/experimenter/experiments/api/v5/serializers.py
lmorchard: I think this is rejecting general experiment updates outside of IDLE publishing status. 

The other serializer rejects general experiment updates outside of DRAFT experiment status.

app/experimenter/experiments/api/v5/serializers.py
lmorchard: Hmm, yeah I think these mixins in NimbusExperimentSerializer should cover it. This serializer should just be for final data validation, the other should catch attempts to update at the wrong points

app/experimenter/experiments/tests/api/v5/test_mutations.py
lmorchard: Ah, I think I worked out the problem - tl;dr: multiple inheritance is tricky and I don't think these mixin classes can work this way with a shared base class.

A fix might be to extract that shared state-change validator as a standalone function and just use that in each of the mixins.

To elaborate: This is a base class for two mixins. Each of those mixins gets used as one of many base classes for the serializer.

Both of these mixins define a `FIELD` property. But, when the multiple inheritance is resolved, it looks like just one `FIELD` survives in the serializer - i.e. that of `StatusValidationMixin`. Each mixin does not have its own isolated scope for `FIELD`

So, when `PublishStatusValidationMixin` tries to validate this state transition in its base class parent `validate`, it's actually looking at `status` rather than `publish_status`.

jodyheavener:
@jaredlockhart @lmorchard Okay so I think I'm a little lost here - the mixins explanation makes sense and using a common validator to validate transitions makes sense, but because there are other cases (such as invalidating the whole experiment when a change is made outside of a valid publish status, or validating when only the status can be changed) it seems like we still need to create individual `validate_status`/`validate_publish_status` methods or individual mixins. Is that correct?

lmorchard:
> it seems like we still need to create individual validate_status/validate_publish_status methods or individual mixins. Is that correct?

I think that's correct. I don't think we can just use the per-field validators - we still need general `validate` method code (i.e. via mixins) to enforce what those validators reject outside just the status change.

jodyheavener:
Makes sense. Currently stuck on my validator class looking up the model instance 🤔

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -208,57 +240,43 @@ def validate(self, data):
...
-        fields = (
-            "title",
-            "link",
-        )
+    def __init__(self, field, transitions):
+        self.field = field
+        self.transitions = transitions
 
+    def __call__(self, value, serializer):
+        if serializer.instance and value:
```
lmorchard, line 260: The third parameter here in a class-based per-field validator __call__ isn't the serializer - it's an individual serializer_field definition. Scratching my head at the docs, but I'm not sure you can access an existing value or determine whether the parent serializer has a instance from a per-field validator context.

My suggestion was to define this transition validation as a standalone function and use that within each mixin - e.g. something like def validate_status_transition(status_from, status_to, allowed_transitions)

jodyheavener: I think this makes sense. Thank you!

lmorchard: Oh, but also I just poked around the DRF source a bit and found [this promising stretch of code](https://github.com/encode/django-rest-framework/blob/d635bc9c71b6cf84d137a68610ae2e628f8b62b3/rest_framework/validators.py#L67-L68):

```python
    def __call__(self, value, serializer_field):
        # Determine the underlying model field name. This may not be the
        # same as the serializer field name if `source=<>` is set.
        field_name = serializer_field.source_attrs[-1]
        # Determine the existing instance, if this is an update operation.
        instance = getattr(serializer_field.parent, 'instance', None)
```

jodyheavener:
@lmorchard @jaredlockhart Alright, took another stab at this and got the validator class set up, as well as a mixin that combines the status/publish_status checks guarding against experiment changes. There are three errors that I'm a little stumped on but generally I like this. What do you think?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -175,34 +175,68 @@ def update(self, experiment, data):
...
-                                f"Nimbus Experiment has status '{status}', only "
-                                "status can be changed."
-                            ]
-                        }
-                    )
+
+            for field, statuses in restrictive_statuses.items():
+                instance_value = getattr(self.instance, field)
+                if instance_value in statuses and set(data.keys()) != {str(field)}:
+                    raised_field = (field, instance_value)
```
jaredlockhart, line 227: Ah so this is one of the status fields ya?  I might name `field` as `status_field` I thought as I was reading this it could be any field

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -294,8 +312,40 @@ class NimbusExperimentSerializer(
...
+                transitions={
+                    NimbusExperiment.Status.DRAFT: (
+                        NimbusExperiment.Status.PREVIEW,
+                        NimbusExperiment.Status.REVIEW,
+                    ),
+                    NimbusExperiment.Status.PREVIEW: (
+                        NimbusExperiment.Status.DRAFT,
+                        NimbusExperiment.Status.REVIEW,
+                    ),
+                },
```
jaredlockhart, line 329: Pls move these transitions over to the Nimbus constants file next to their respective enums cuz we'll have to update those together 🙏

lmorchard: We have other related constant lists in here too - should they move also?

https://github.com/mozilla/experimenter/pull/4887/files#diff-da5192d8fdd27a14107b117c1f4fbfbbc6cbee34c61bc19dd273fc612781f7ebR204-R210

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -255,10 +200,75 @@ def update(self, experiment, data):
...
+
+        if self.instance:
+            status = self.instance.status
+
+            for status_field, statuses in restrictive_statuses.items():
+                instance_value = getattr(self.instance, status_field)
+                if instance_value in statuses and set(data.keys()) != {str(status_field)}:
+                    raised_status_field = (status_field, instance_value)
+
+            if raised_status_field:
```
lmorchard, line 221: > There are three errors that I'm a little stumped on

One of the errors is in `test_preview_draft_transition_invokes_kinto_task_1_Preview`, where the serializer is not valid and instead reports "Nimbus Experiment has status 'Preview', but can only be changed when set to 'Draft'."

I think the issue is that we've lost [the `if status in self.ALLOWS_STATUS_CHANGE_ONLY:` conditional](https://github.com/mozilla/experimenter/pull/4887/files#diff-da5192d8fdd27a14107b117c1f4fbfbbc6cbee34c61bc19dd273fc612781f7ebL196) in this refactor.

This "raised_status_field" check rejects what *can't* be updated. But it doesn't allow what can. If nothing is raised, it falls through to the next branch of the conditional, which guards against general updates.

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -255,10 +200,75 @@ def update(self, experiment, data):
...
+            "status": NimbusConstants.STATUS_CHANGE_ONLY,
+            "publish_status": NimbusConstants.PUBLISH_STATUS_CHANGE_ONLY,
+        }
+
+        if self.instance:
+            status = self.instance.status
+
+            for status_field, statuses in restrictive_statuses.items():
+                instance_value = getattr(self.instance, status_field)
+                if instance_value in statuses and set(data.keys()) != {str(status_field)}:
```
lmorchard, line 218: Maybe we also want the opposite here?

If the experiment status is in the list - and the set of keys proposed for update exactly match what's allowed for this status - bail out of this validator entirely because this is an allowed update?

lmorchard: FWIW, I went ahead and tried this tweak locally - pytest passes for me

lmorchard: A further tweak might be to just go ahead and hoist that `if raised_status_fields` conditional into that loop too.

So, like:

```diff
diff --git a/app/experimenter/experiments/api/v5/serializers.py b/app/experimenter/experiments/api/v5/serializers.py
index f9f9aa52..f97b3265 100644
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -215,21 +215,20 @@ class NimbusStatusValidationMixin:
 
             for status_field, statuses in restrictive_statuses.items():
                 instance_value = getattr(self.instance, status_field)
-                if instance_value in statuses and set(data.keys()) != {str(status_field)}:
-                    raised_status_field = (status_field, instance_value)
+                if instance_value in statuses:
+                    if set(data.keys()) == {str(status_field)}:
+                        return data
+                    else:
+                        raise serializers.ValidationError(
+                            {
+                                "experiment": [
+                                    f"Nimbus Experiment has {status_field} '{instance_value}', "
+                                    f"only {status_field} can be changed."
+                                ]
+                            }
+                        )
 
-            if raised_status_field:
-                status_field = raised_status_field[0]
-                value = raised_status_field[1]
-                raise serializers.ValidationError(
-                    {
-                        "experiment": [
-                            f"Nimbus Experiment has {status_field} '{value}', "
-                            f"only {status_field} can be changed."
-                        ]
-                    }
-                )
-            elif status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
+            if status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
                 required_statuses = ", ".join(NimbusConstants.STATUS_ALLOWS_UPDATE)
                 raise serializers.ValidationError(
                     {
```

jodyheavener: Very nice, thanks for the pointer!

lmorchard (approved):
Dismissed my old review just to r+ again, but you might want a second opinion since I jumped the r+ gun back then :)

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -255,10 +200,74 @@ def update(self, experiment, data):
...
+
+        if self.instance:
+            status = self.instance.status
+
+            for status_field, statuses in restrictive_statuses.items():
+                instance_value = getattr(self.instance, status_field)
+                if instance_value in statuses:
+                    if set(data.keys()) == {status_field}:
+                        return data
+                    else:
```
jaredlockhart, line 220: You can take this whole early return out if you want and just

```py
if not set(.....):
    raise serializers.ValidationError(...)
```

I'm not a huge fan of early returns if they can be avoided, makes it harder to trace the flow of a function.

jaredlockhart (approved):
Yeah this is really great @jodyheavener the logic is clear and modular, constants are well laid out, tests are clear and concise, A+ 10/10 would R+ again great job 👏 👏 👏

app/experimenter/experiments/api/v5/serializers.py
jodyheavener: I think the early return is to avoid catching on the following case?

jaredlockhart: Don't you WANT to check that case if the first case is valid?

jaredlockhart: If you only want to check the second case if the first case is valid then i'd write it as a single if/else block instead of the early return to make it clearer?

jodyheavener: If I'm understanding this correctly, if the first check passes (the status or publish_status permits any field to be changed), we don't care if the status is not in the list of statuses that only allows status updates.

So yes, I could remove the early return, but to avoid hitting that second check I'd also need to set up a variable inside the for-loop that determines whether or not to perform that second check. I think?

jodyheavener: a la https://github.com/mozilla/experimenter/pull/4887/commits/e3775186f73bfe6c6f7b332a0dd6ea2f9be603a8

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -255,10 +200,75 @@ def update(self, experiment, data):
...
+                required_statuses = ", ".join(NimbusConstants.STATUS_ALLOWS_UPDATE)
+                raise serializers.ValidationError(
+                    {
+                        "experiment": [
+                            f"Nimbus Experiment has status '{status}', but can only "
+                            f"be changed when set to '{required_statuses}'."
+                        ]
+                    }
+                )
+
```
jaredlockhart, line 242: Wait can't this whole thing just be

```py
            for status_field, restricted_statuses in restrictive_statuses.items():
                current_status = getattr(self.instance, status_field)
                is_locked = current_status in restricted_statuses
                is_modifying_other_fields = set(data.keys()) == {status_field}
                if is_locked and is_modifying other_fields:
                            {
                                "experiment": [
                                    f"Nimbus Experiment has {status_field} "
                                    f"'{instance_value}', only {status_field} "
                                    "can be changed."
                                ]
                            }
                        )
```

jaredlockhart:
🧁 for you

jodyheavener:
@jaredlockhart took your example and flipped the statuses we check from "status"-only changes to "any field" statuses, which I _think_ was the intent of your suggestion

jaredlockhart:
Oh lol we both did it XD

```
-        restrictive_statuses = {
-            "status": NimbusConstants.STATUS_CHANGE_ONLY,
-            "publish_status": NimbusConstants.PUBLISH_STATUS_CHANGE_ONLY,
+        allow_update_statuses = {
+            "status": NimbusConstants.STATUS_ALLOWS_UPDATE,
+            "publish_status": NimbusConstants.PUBLISH_STATUS_ALLOWS_UPDATE,
         }
 
         if self.instance:
-            status = self.instance.status
-            check_status_only = True
-
-            for status_field, statuses in restrictive_statuses.items():
-                instance_value = getattr(self.instance, status_field)
-                if instance_value in statuses:
-                    if set(data.keys()) == {status_field}:
-                        check_status_only = False
-                    else:
-                        raise serializers.ValidationError(
-                            {
-                                "experiment": [
-                                    f"Nimbus Experiment has {status_field} "
-                                    f"'{instance_value}', only {status_field} "
-                                    "can be changed."
-                                ]
-                            }
-                        )
-
-            if check_status_only and status not in NimbusConstants.STATUS_ALLOWS_UPDATE:
-                required_statuses = ", ".join(NimbusConstants.STATUS_ALLOWS_UPDATE)
-                raise serializers.ValidationError(
-                    {
-                        "experiment": [
-                            f"Nimbus Experiment has status '{status}', but can only "
-                            f"be changed when set to '{required_statuses}'."
-                        ]
-                    }
-                )
+            for status_field, allows_update in allow_update_statuses.items():
+                current_status = getattr(self.instance, status_field)
+                is_locked = current_status not in allows_update
+                is_modifying_other_fields = set(data.keys()) != {status_field}
+                if is_locked and is_modifying_other_fields:
+                    raise serializers.ValidationError(
+                        {
+                            "experiment": [
+                                f"Nimbus Experiment has {status_field} "
+                                f"'{current_status}', only {status_field} "
+                                "can be changed."
+                            ]
+                        }
+                    )
 
         return data
 
diff --git a/app/experimenter/experiments/constants/nimbus.py b/app/experimenter/experiments/constants/nimbus.py
index 178857a3..2bb430eb 100644
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -111,7 +111,6 @@ class NimbusConstants(object):
             Status.REVIEW,
         ),
     }
-    STATUS_CHANGE_ONLY = (Status.PREVIEW,)
     STATUS_ALLOWS_UPDATE = (Status.DRAFT,)
 
     class PublishStatus(models.TextChoices):
@@ -127,11 +126,7 @@ class NimbusConstants(object):
             PublishStatus.APPROVED,
         ),
     }
-    PUBLISH_STATUS_CHANGE_ONLY = (
-        PublishStatus.REVIEW,
-        PublishStatus.APPROVED,
-        PublishStatus.WAITING,
-    )
+    PUBLISH_STATUS_ALLOWS_UPDATE = (PublishStatus.IDLE,)
```

jaredlockhart (approved):
Yep it's good!

PR #4889, opened 2021-03-22
fix #4867 feat(nimbus): validate reviewer is different from review requester in V5 API

### LINKED ISSUE #4867: Validate requester/reviewer identity in V5 API
We must validate that a reviewer is different than the review requester in the V5 API.

We should:

* add a method to NimbusExperiment called `can_review` that takes in a user and makes sure it's not the same user that requested the review, which can be found by looking up the most recent changelog which moved the experiment publish status from Idle to Review
* Add a property to the NimbusExperiment type in the V5 API which takes in the current request flow user and passes it into that method to populate that field, which will then be used by the frontend to enable/disable the review UI
* Validate in the V5 NimbusExperiment serializer that changes to the `publish_status` field are made by a user that can review

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1094)
┆Issue Number: EXP-1094

jaredlockhart:
Because

* We want reviews to be made by a different user than the user that requested review

This commit

* Adds logic to the NimbusExperiment to determine whether a user can review
* Exposes whether the current user can review an experiment in the V5 Query API
* Validates whether the current user can approve an experiment in the V5 Mutation API

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -337,6 +337,16 @@ class NimbusExperimentSerializer(
         )
         super().__init__(instance=instance, data=data, **kwargs)
 
+    def validate_publish_status(self, publish_status):
+        if (
+            publish_status == NimbusExperiment.PublishStatus.APPROVED
+            and not self.instance.can_review(self.context["user"])
+        ):
+            raise serializers.ValidationError(
+                f'{self.context["user"]} can not review this experiment.'
+            )
+        return publish_status
+
     def validate_name(self, name):
         if not (self.instance or name):
             raise serializers.ValidationError("Name is required to create an experiment")
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -135,6 +135,7 @@ class NimbusExperimentType(DjangoObjectType):
     computed_end_date = graphene.DateTime()
     is_enrollment_paused = graphene.Boolean()
     enrollment_end_date = graphene.DateTime()
+    can_review = graphene.Boolean()
 
     class Meta:
         model = NimbusExperiment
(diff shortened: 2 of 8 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -337,6 +337,16 @@ def __init__(self, instance=None, data=None, **kwargs):
...
 
+    def validate_publish_status(self, publish_status):
+        if (
+            publish_status == NimbusExperiment.PublishStatus.APPROVED
+            and not self.instance.can_review(self.context["user"])
+        ):
+            raise serializers.ValidationError(
+                f'{self.context["user"]} can not review this experiment.'
+            )
+        return publish_status
```
jodyheavener, line 348: should this go in the StatusRestrictionMixin?

jodyheavener (approved):
LGTM!

pdehaan:
Forgive my general ignorance, but would this apply to local/dev/stage/prod, or just prod?
I can imagine in cases on stage where I want to approve my own experiment for testing without having to bother somebody else. Or for e2e tests, where we might just be testing locally and don't want to have to create multiple users for tests.

jaredlockhart:
> Forgive my general ignorance, but would this apply to local/dev/stage/prod, or just prod?
> I can imagine in cases on stage where I want to approve my own experiment for testing without having to bother somebody else. Or for e2e tests, where we might just be testing locally and don't want to have to create multiple users for tests.

Yep great points but I don't think disabling this validation logic for everywhere except prod would be the right approach, then you're not really testing the actual workflow on stage and if that logic breaks because of the edge cases we lose the security constraint.   I'd rather find a way to keep the validation logic intact and find simple ways to test around it.  The free method would just be changing users in the admin while you test it.  If that proves to be too much of a hassle we can find something fancier.

PR #4896, opened 2021-03-23
fix #4868 feat(nimbus): add rejection to v5 api

### LINKED ISSUE #4868: Expose rejection data in V5 API
To support the new rejection UI we need to look up the rejection reason in the changelogs and expose it in a convenient field in the V5 API.

We should

* Add a manager method to NimbusChangeLog to look up the most recent rejection changelog by finding the changelog which moves from publish_status either Review>Idle (rejected in experimenter)or Waiting>Idle (rejected in remote settings)
* Add a field in the V5 NimbusExperiment type called `rejection_reason` that is None or includes the entire changelog object so the frontend can display the reason, the user, and the date

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1095)
┆Issue Number: EXP-1095

jaredlockhart:
Because

* We need a way to populate the rejection UI with the latest rejection data

This commit

* Looks up the most recent rejection data and adds it to a field in the V5 API called 'rejection'

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -9,6 +9,7 @@ from django.contrib.postgres.fields import ArrayField
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import MaxValueValidator
 from django.db import models
+from django.db.models import Q
 from django.urls import reverse
 from django.utils import timezone
 
@@ -426,8 +427,25 @@ class NimbusChangeLogManager(models.Manager):
         return (
             self.all()
             .filter(
+                old_status=NimbusExperiment.Status.DRAFT,
                 old_publish_status=NimbusExperiment.PublishStatus.IDLE,
+                new_status=NimbusExperiment.Status.DRAFT,
                 new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
+            )
+            .order_by("-changed_on")
+        ).first()
+
+    def latest_rejection(self):
+        return (
+            self.all()
+            .filter(
+                Q(old_publish_status=NimbusExperiment.PublishStatus.REVIEW)
+                | Q(old_publish_status=NimbusExperiment.PublishStatus.WAITING),
+            )
+            .filter(
+                old_status=NimbusExperiment.Status.DRAFT,
+                new_status=NimbusExperiment.Status.DRAFT,
+                new_publish_status=NimbusExperiment.PublishStatus.IDLE,
             )
             .order_by("-changed_on")
         ).first()
--- a/app/experimenter/experiments/tests/api/v5/test_queries.py
+++ b/app/experimenter/experiments/tests/api/v5/test_queries.py
@@ -385,6 +385,70 @@ class TestNimbusQuery(GraphQLTestCase):
         experiment_data = content["data"]["experimentBySlug"]
         self.assertTrue(experiment_data["canReview"])
 
+    def test_experiment_no_rejection_data(self):
+        user_email = "user@example.com"
+        experiment = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.DRAFT,
+            publish_status=NimbusExperiment.PublishStatus.IDLE,
+        )
+
+        response = self.query(
+            """
+            query experimentBySlug($slug: String!) {
+                experimentBySlug(slug: $slug) {
+                    rejection {
+                        changedBy {
+                            email
+                        }
+                    }
+                }
+            }
+            """,
+            variables={"slug": experiment.slug},
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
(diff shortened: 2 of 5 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -426,12 +427,29 @@ def latest_review_request(self):
...
             .filter(
+                old_status=NimbusExperiment.Status.DRAFT,
                 old_publish_status=NimbusExperiment.PublishStatus.IDLE,
+                new_status=NimbusExperiment.Status.DRAFT,
                 new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
             )
             .order_by("-changed_on")
         ).first()
 
+    def latest_rejection(self):
```
lmorchard, line 438: Don't we want this to be per-experiment, rather than the latest rejection across the system? Otherwise, then won't every draft experiment show a rejection if one gets rejected for review?

jaredlockhart: When you call it from an experiments related manager it applies the per experiment filter

`experiment.changes.latest_rejection()` expands to

```sql
 SELECT "experiments_nimbuschangelog"."id",
       "experiments_nimbuschangelog"."experiment_id",
       "experiments_nimbuschangelog"."changed_on",
       "experiments_nimbuschangelog"."changed_by_id",
       "experiments_nimbuschangelog"."old_status",
       "experiments_nimbuschangelog"."old_publish_status",
       "experiments_nimbuschangelog"."new_status",
       "experiments_nimbuschangelog"."new_publish_status",
       "experiments_nimbuschangelog"."message",
       "experiments_nimbuschangelog"."experiment_data"
FROM   "experiments_nimbuschangelog"
WHERE  ( "experiments_nimbuschangelog"."experiment_id" = 8
         AND ( "experiments_nimbuschangelog"."old_publish_status" = 'Review'
                OR "experiments_nimbuschangelog"."old_publish_status" =
                   'Waiting' )
         AND "experiments_nimbuschangelog"."new_publish_status" = 'Idle'
         AND "experiments_nimbuschangelog"."new_status" = 'Draft'
         AND "experiments_nimbuschangelog"."old_status" = 'Draft' )
ORDER  BY "experiments_nimbuschangelog"."changed_on" DESC
LIMIT  1;  
```

lmorchard: Ah, I forgot about that magic bit of Django ORM

lmorchard (approved):
LGTM

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -385,6 +385,70 @@ def test_experiment_in_review_can_review(self):
...
+            publish_status=NimbusExperiment.PublishStatus.IDLE,
+        )
+
+        for publish_status in (
+            NimbusExperiment.PublishStatus.REVIEW,
+            NimbusExperiment.PublishStatus.IDLE,
+        ):
+            experiment.publish_status = publish_status
+            experiment.save()
+            generate_nimbus_changelog(experiment, experiment.owner)
```
jodyheavener, line 428: I've seen this pattern in a lot of tests. Worth turning it into a helper?

app/experimenter/nimbus-ui/schema.graphql
```diff
@@ -47,6 +47,8 @@ input ExperimentInput {
   totalEnrolledClients: Int
 }
 
+scalar JSONString
```
jodyheavener, line 50: The frontend won't know what to do with this, but we can map it to a type [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/react-app-env.d.ts#L7:L7). Something like `type JSONString = Record<any, any>;`

jodyheavener (commented):
Had a couple thoughts.

app/experimenter/experiments/tests/api/v5/test_queries.py
jaredlockhart: Yeah I was thinking that I think when we move everything over to actually using publish_status we'll update the `create_with_status` method on the factory to include moving through the publish_statuses but when I tried now it didn't make sense without all the other bits being hooked up so we can do it after we wrap everything else up as an optimization.  But yes 👍

app/experimenter/nimbus-ui/schema.graphql
jaredlockhart: Oh yeah... it's doing that because we store the entire DTO on the changelog but we won't need it in the frontend until we build out the changelog UI which is coming soon in an epic so we can probably deal with it then.  But good call out 👍

app/experimenter/experiments/tests/api/v5/test_queries.py
jaredlockhart: Filed https://github.com/mozilla/experimenter/issues/4902 as a followup

PR #4897, opened 2021-03-24
fixes #4869 feat(nimbus): expose timeout data in V5 API

### LINKED ISSUE #4869: Expose timeout data in V5 API
To support the new timeout UI we need to look up the timeout in the changelogs and expose it in a convenient field in the V5 API.

We should

* Add a manager method to NimbusChangeLog to look up the most recent timeout changelog by finding the changelog which moves from publish_status Approved>Idle
* Add a field in the V5 NimbusExperiment type called `timeout` that is None or includes the entire changelog object so the frontend can display the date of the timeout

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1096)
┆Issue Number: EXP-1096

lmorchard:
Because:

- To support the new timeout UI we need to look up the timeout in the
  changelogs and expose it in a convenient field in the V5 API.

This commit:

- Adds a `timeout` property to the NimbusExperiment model that looks up
  the latest changelog entry that transitions from publish status waiting
  to review, as long as the experiment is still in review status

- Exposes the `timeout` model property as a field in GQL queries

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -450,6 +450,26 @@ class NimbusChangeLogManager(models.Manager):
             .order_by("-changed_on")
         ).first()
 
+    def latest_timeout(self):
+        return (
+            self.all()
+            .filter(
+                Q(
+                    old_status=NimbusExperiment.Status.DRAFT,
+                    new_status=NimbusExperiment.Status.DRAFT,
+                )
+                | Q(
+                    old_status=NimbusExperiment.Status.LIVE,
+                    new_status=NimbusExperiment.Status.LIVE,
+                )
+            )
+            .filter(
+                old_publish_status=NimbusExperiment.PublishStatus.WAITING,
+                new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
+            )
+            .order_by("-changed_on")
+        ).first()
+
 
 class NimbusChangeLog(models.Model):
     def current_datetime():
--- a/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
+++ b/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
@@ -453,6 +453,34 @@ class TestNimbusExperiment(TestCase):
 
         self.assertFalse(experiment.can_review(UserFactory.create()))
 
+    @parameterized.expand(
+        [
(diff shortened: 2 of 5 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -172,6 +172,19 @@ def treatment_branches(self):
             branches = branches.exclude(id=self.reference_branch.id)
         return list(branches)
 
+    @property
```
lmorchard, line 175: I might be over-thinking this. But, I figured we'd want some more logic here:

* Only report a timeout if the experiment has gotten kicked from WAITING back to REVIEW - and is *still* in REVIEW publishing status.
* The timeout reported is None if/when the experiment gets another attempt going from REVIEW -> APPROVED -> WAITING

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -453,6 +453,49 @@ def test_can_review_false_for_non_review_publish_status(self, publish_status):
 
         self.assertFalse(experiment.can_review(UserFactory.create()))
 
+    @parameterized.expand(
+        [
+            NimbusExperiment.Status.DRAFT,
+            NimbusExperiment.Status.LIVE,
+        ]
+    )
+    def test_timeout_changelog_for_timedout_publish_flow(self, experiment_status):
```
lmorchard, line 462: I *think* all the case we wanted to cover here are within DRAFT and LIVE experiment statuses

lmorchard (commented):
Dropped some notes on some further thinking I did around logic here

jaredlockhart (commented):
Thanks for hopping on this @lmorchard !  I'd suggest waiting until https://github.com/mozilla/experimenter/pull/4896 lands, rebasing, and following the same pattern (ie putting the lookup method in the NimbusChangeLogManager)

lmorchard:
Hmm, if I move from an experiment model property to a changelog manager method, I think that would always give us the last recent timeout but have to shift the decision on whether the timeout is currently relevant to the frontend (e.g. is the experiment timed out or waiting for another approval)

lmorchard:
Rebased, reworked to follow in the footsteps of PR #4896

jaredlockhart (approved):
Awesome looks right to me!

PR #4942, opened 2021-03-29
fix #4888 feat(nimbus): time out launch and end reviews in remote settings

### LINKED ISSUE #4888: Back out timed out changes from Remote Settings
We need to update our Kinto tasks to back out any pending changes from a collection rather than just returning early.  This involves rolling back the state of the collection, and updating the pending experiment to the relevant to the correct state.  Refer to the Kinto documentation for the relevant states in the timeout cases.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1100)
┆Issue Number: EXP-1100

jaredlockhart:
Becuase

* Remote settings can only have a single review up at a time
* This means a neglected review can block other experiments from launching/ending
* We would like unattended reviews to time out and get backed out

This commit

* Adds a new handle_pending case to the check collection task
* If a pending review is for a launching or ending experiment, it gets rolled back
* If a pending review is for a pause update, it does not get rolled back and must be accepted (for now, until we add live experiment updates then it will behave like launch/end cases)
* I added a new way to check experiment states using Django Q objects so that we can define aggregate states as constants rather than dumping them all as methods, in theory this should let us continue to define and modify aggregate states without needing to add/modify new methods/tests all the time

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -3,6 +3,7 @@ from typing import Dict
 
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
 
 
 @dataclass
@@ -193,6 +194,15 @@ class NimbusConstants(object):
         "publish_status",
         "is_end_requested",
     )
+
+    # State Predicates
+    IS_LAUNCHING = Q(status=Status.DRAFT, publish_status=PublishStatus.WAITING)
+    IS_ENDING = Q(
+        status=Status.LIVE,
+        publish_status=PublishStatus.WAITING,
+        is_end_requested=True,
+    )
+    SHOULD_TIMEOUT = Q(IS_LAUNCHING | IS_ENDING)
 
     class Application(models.TextChoices):
         DESKTOP = (APPLICATION_CONFIG_DESKTOP.slug, APPLICATION_CONFIG_DESKTOP.name)
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -38,12 +38,16 @@ class NimbusExperimentManager(models.Manager):
     def end_queue(self, application):
         return self.filter(
             status=NimbusExperiment.Status.LIVE,
+            publish_status=NimbusExperiment.PublishStatus.APPROVED,
             application=application,
             is_end_requested=True,
         )
 
-    def waiting_queue(self):
-        return self.filter(publish_status=NimbusExperiment.PublishStatus.WAITING)
+    def waiting_to_launch_queue(self):
+        return self.filter(
+            status=NimbusExperiment.Status.DRAFT,
+            publish_status=NimbusExperiment.PublishStatus.WAITING,
+        )
 
 
 class NimbusExperiment(NimbusConstants, models.Model):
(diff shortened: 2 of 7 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -113,6 +113,9 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse("nimbus-detail", kwargs={"slug": self.slug})
 
+    def has_state(self, state):
+        return type(self).objects.filter(id=self.id).filter(state).exists()
```
lmorchard, line 117: This seems kind of roundabout, though: We have the experiment and its data fetched from the DB, but we're doing another DB query to determine whether it has a particular state. Maybe it'll make more sense if/when we use these for aggregate queries? Though that seems covered in the `{launch,pause,end}_queue` methods

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -128,6 +129,15 @@ class PublishStatus(models.TextChoices):
     }
     PUBLISH_STATUS_ALLOWS_UPDATE = (PublishStatus.IDLE,)
 
+    # State Predicates
+    IS_LAUNCHING = Q(status=Status.DRAFT, publish_status=PublishStatus.WAITING)
+    IS_ENDING = Q(
+        status=Status.LIVE,
+        publish_status=PublishStatus.WAITING,
+        is_end_requested=True,
+    )
```
lmorchard, line 138: FWIW, these might actually be handy as model methods if we plumb them through GQL to the frontend as booleans. Then the frontend won't need to duplicate this logic. Not a huge deal either way though

lmorchard (approved):
LGTM, though I'm 🤷‍♂️ on the state queries

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Yeah that's a good point this might not scale to the frontend, we can keep an eye out for that.

app/experimenter/experiments/models/nimbus.py
lmorchard: Yeah, not a colossal change either way, just kind of projecting forward. I wouldn't (haven't) r- on it so we can burn that bridge when we get to it.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -38,12 +38,16 @@ def pause_queue(self, application):
...
             is_end_requested=True,
         )
 
-    def waiting_queue(self):
-        return self.filter(publish_status=NimbusExperiment.PublishStatus.WAITING)
+    def waiting_to_launch_queue(self):
+        return self.filter(
+            status=NimbusExperiment.Status.DRAFT,
+            publish_status=NimbusExperiment.PublishStatus.WAITING,
+        )
```
jodyheavener, line 50: 👍

app/experimenter/kinto/tasks.py
```diff
@@ -215,7 +247,9 @@ def nimbus_check_experiments_are_live():
                 experiment.publish_status = NimbusExperiment.PublishStatus.IDLE
                 experiment.save()
 
-                generate_nimbus_changelog(experiment, get_kinto_user())
+                generate_nimbus_changelog(
+                    experiment, get_kinto_user(), message="Experiment is now live!"
```
jodyheavener, line 251: What are your thoughts on making changelog messages constant keys and letting the frontend decide which message to display? Or even storing this and the `Timed Out` messages as constants in the backend? Would love to see these messages grouped/managed together

jaredlockhart: Yeah I filed another ticket to dig into this more deeply https://github.com/mozilla/experimenter/issues/5022 you're right we should store them as constants.

jodyheavener (approved):
beau travail

jaredlockhart:
merci beaucoup!

PR #4986, opened 2021-04-05
Let's see if we can cause the timeout in circle

jaredlockhart:


jaredlockhart:
Okay we figured this out, the v5 serializer tries to call the celery task to synchronize the preview collection and by updating the v5 mutation test from DRAFT > REVIEW to DRAFT > PREVIEW the serializer was invoking the celery task but it wasn't mocked out in that test so it was trying to reach redis and just hanging.  I don't know why it wouldn't just error out that it can't connect but I can't seem to force it to error out so we just need to pay attention to tests that call into anything that'll invoke celery tasks.

PR #4988, opened 2021-04-05
Initial commit adding e2e integration tests for nimbus frontend.

jrbenny35:
Basic e2e test. Also added some ID's to pages. There are 2 makefile integration test jobs now, as well as 2 circleci jobs.


.circleci/config.yml
```diff
@@ -77,7 +77,29 @@ jobs:
...
+          name: Docker compose info
+          command: docker-compose -v
+      - checkout
+      - run:
+          name: Run integration tests
+          command: |
+            cp .env.sample .env
+            make refresh
+            make up_prod_detached
+            make integration_test_nimbus
```
jaredlockhart, line 102: Very nice 👏 👏 👏

Makefile
```diff
@@ -145,13 +145,16 @@ integration_build: build_prod ssl
...
 integration_vnc_up_detached: integration_build
 	$(COMPOSE_INTEGRATION) up -d firefox
 
-integration_test: integration_build
-	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -- -n 4"
+integration_test_legacy: integration_build
+	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-legacy -- -n 4"
+
+integration_test_nimbus: integration_build
+	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-nimbus -- -n 4"
```
jaredlockhart, line 160: Yep yep yep

app/tests/integration/nimbus/pages/audience.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def set_min_version(self,  version=80):
+        el = self.find_element(*self._min_version_select_locator)
+        select = Select(el)
+        select.select_by_value(f"FIREFOX_{version}")
+
+    def set_targeting(self, targeting="US_ONLY"):
+        el = self.find_element(*self._targeting_select_locator)
+        select = Select(el)
+        select.select_by_value(targeting)
```
jaredlockhart, line 48: Maybe we could use the same setter/getter pattern everywhere?  So like the getter/setter same name with @thing.setter like below, just for consistency?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -0,0 +1,35 @@
...
+    audience = metrics.save_and_continue()
+    audience.set_channel()
+    audience.set_min_version()
+    audience.set_targeting()
+    audience.percentage = 50.0
+    audience.expected_clients = 50
+    audience.save_btn()
+    time.sleep(1)
+    audience.save_and_continue()
+    selenium.find_element_by_css_selector("#PageRequestReview")
```
jaredlockhart, line 34: Awesome 🎉 🎉 🎉 🎉 

Maybe just add some new lines to break it up by page so it's a bit easier to visually parse?

jaredlockhart (commented):
This is so great 👏 👏 👏 🎉 🎉 🎉 

Looks great to me, no blockers just a few style suggestions.  Let's get @jodyheavener or @LZoog or @lmorchard to sanity check the frontend bits and then get this landed!

app/tests/integration/nimbus/pages/audience.py
jrbenny35: Hmm yeah, wonder what I was thinking here lol

jaredlockhart:
Ah I just noticed if you try to run it twice in a row it uses the same name so it fails the second time, maybe we could use a random name so you can run it multiple times?  I expect during local debugging you'll want to run the tests over and over.

app/tests/integration/nimbus/pages/branches.py
```diff
@@ -0,0 +1,49 @@
...
+from selenium.webdriver.support import expected_conditions as EC
+
+
+
+class BranchesPage(Base):
+    """Experiment Branches Page."""
+
+    _reference_branch_name_locator = (By.CSS_SELECTOR, "#referenceBranch-name")
+    _reference_branch_description_locator = (By.CSS_SELECTOR, "#referenceBranch-description")
+    _remove_branch_locator = (By.CSS_SELECTOR, ".bg-transparent")
```
pdehaan, line 13: Q: Do we need to add a better classname or something to this button?

app/tests/integration/nimbus/pages/new_experiment.py
```diff
@@ -0,0 +1,70 @@
...
+from nimbus.pages.base import Base
+
+
+class NewExperiment(Base):
+    """New Experiment Page."""
+
+    _public_name_locator = (By.CSS_SELECTOR, "#name")
+    _hypothesis_locator = (By.CSS_SELECTOR, "#hypothesis")
+    _application_select_locator = (By.CSS_SELECTOR, "#application")
+    _cancel_btn_locator = (By.CSS_SELECTOR, ".btn-light")
```
pdehaan, line 17: Q: Do we need to add a better classname or something to this button?

app/tests/integration/nimbus/pages/metrics.py
```diff
@@ -0,0 +1,23 @@
+from nimbus.pages.base import Base
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+
+
+class MetricsPage(Base):
+    """Experiment Metrics Page.."""
```
pdehaan, line 8: nit: two ".." 🤷

app/tests/integration/nimbus/pages/new_experiment.py
```diff
@@ -0,0 +1,70 @@
...
+        name.send_keys(f"{text}")
+
+    @property
+    def hypothesis(self):
+        return self.find_element(*self._hypothesis_locator).text
+
+    @hypothesis.setter
+    def hypothesis(self, text=None):
+        name = self.find_element(*self._hypothesis_locator)
+        name.send_keys(f" {text}")
```
pdehaan, line 49: Probably doesn't matter, but i think this will add a leading space to the text.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -0,0 +1,38 @@
...
+    # Fill Metrics page
+    metrics = branches.save_and_continue()
+    # Fill Audience page
+    audience = metrics.save_and_continue()
+    audience.channel = "Nightly"
+    audience.min_version = 80
+    audience.targeting = "US_ONLY"
+    audience.percentage = 50.0
+    audience.expected_clients = 50
+    audience.save_btn()
```
pdehaan, line 35: Do we need a comment here on why we need the double save => save_and_continue?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
```diff
@@ -253,6 +253,7 @@ export const FormBranches = ({
             <button
               data-testid="next-button"
               className="btn btn-secondary"
+              id="save-and-continue-btn"
```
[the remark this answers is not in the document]
jrbenny35: hmm, good point

app/tests/integration/nimbus/pages/audience.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def wait_for_page_to_load(self):
+        self.wait.until(EC.presence_of_element_located(self._page_wait_locator))
+
+        return self
+
+    def save_and_continue(self):
+        element = self.selenium.find_element(*self._save_continue_btn_locator)
+        element.click()
+        from nimbus.pages.review import ReviewPage
```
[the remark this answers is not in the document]
jrbenny35: Yeah that would cause a circular import

app/tests/integration/nimbus/pages/branches.py
jrbenny35: maybe, but its the only button with that class name on the page

app/tests/integration/nimbus/pages/new_experiment.py
jrbenny35: we could but I don't think it's a big deal.

app/tests/integration/nimbus/pages/new_experiment.py
jrbenny35: yes, this is intentional

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: I could add it

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -141,7 +143,7 @@ const AppLayoutWithExperiment = ({
         experiment,
       }}
     >
-      <section data-testid={testId}>
+      <section data-testid={testId} id={pageId}>
```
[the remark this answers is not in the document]
jrbenny35: this was a good catch. I completely zoned out on this when I was looking for css elements and literally added my own.

app/tests/integration/nimbus/pages/branches.py
jaredlockhart: Yeah good catch @pdehaan probably a good idea to give it an ID

app/tests/integration/nimbus/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+
+@pytest.fixture(scope="session", autouse=True)
+def _verify_url(request, base_url):
+    """Verifies the base URL"""
+    verify = request.config.option.verify_base_url
+    if base_url and verify:
+        session = requests.Session()
+        retries = Retry(backoff_factor=0.1, status_forcelist=[500, 502, 503, 504])
+        session.mount(base_url, HTTPAdapter(max_retries=retries))
+        session.get(base_url, verify=False)
```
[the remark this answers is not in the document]
jaredlockhart: These should be getting caught by flake8/black....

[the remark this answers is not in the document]
jaredlockhart: Ah whoops nope, I thought I added that a while go but looks like I didn't, I'll do that now.

app/tests/integration/nimbus/pages/audience.py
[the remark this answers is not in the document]
jaredlockhart: I would generally put the import at the top of the function, but that's just a style preference.

jaredlockhart (approved):
Yep looks great to me!  I'll add a fix as a separate pr to make sure these files get caught by flake8 and black so we don't have to worry about formatting.  And thank you for making the getters/setters consistent.  We can worry about things like adding explicit ids to every elements the tests touch as little tweaks and followups.  The important thing is getting this landed and then expanding to include the rest of the workflow.

Thanks for taking a close look @pdehaan !  If there's anything not yet addressed please file followups so we don't lose ideas 👍 

@jrbenny35 great job, just get this passing all the checks and it should be good to land 🎉

jaredlockhart:
Oh! Sorry @jrbenny35 I just realized, durr, because this moves the circle jobs around and those are also configured in github we'll have to do this in two passes.  Put the `integration` circle job back in in its old form with an updated command to run the legacy tests, and keep the two new ones `integration_legacy` `integration_nimbus` in as well, and that way we can keep the old checks in place and then we can update github after this lands.

PR #4989, opened 2021-04-05
chore(nimbus): reset outcome cache once during test class

jaredlockhart:
Because

* I noticed while debugging other tests that we were clearing the outcome cache in setUp, which means it's reloading all the outcomes from disk at the start of each test, which only needs to happen once at the start of the test class
* Also I noticed that sometimes we were doing that without overriding to use the test fixtures so some tests were completely testing the wrong thing

This commit

* calls Outcomes.clear_cache() in setUpClass instead of setUp to only hit the disk once
* mocks out the outcomes fixture path for any test that uses them


lmorchard (approved):
LGTM

PR #4990, opened 2021-04-05
chore(nimbus): add make kill to all make test commands

jaredlockhart:
Because

* I just noticed while debugging another issue that circle only has app and db containers running while it runs tests, and if you make bash or up and have other containers running like redis or kinto or autograph, those can cause stateful changes to the test environment such as if a test is attempting to invoke a celery that hasn't been mocked out, having redis running in the background will allow that test to pass, so then it looks like the test is failing in circle and passing locally becuase you have redis up which can be confusing

This commit

* Runs make kill before starting any test so that if we hit a situation like that again, local testing should more closely replicate circle


lmorchard (approved):
:+1:

Makefile
```diff
@@ -63,15 +63,9 @@ build_test: jetstream_config
 build_prod: jetstream_config
 	docker build --target deploy -f app/Dockerfile -t app:deploy app/
 
-test_build: build_test
+compose_build_test: kill build_test
```
lmorchard, line 66: Oh, I missed that [.circleci/config.yml executes `make test_build`](https://github.com/mozilla/experimenter/blob/84db87252deb14bd75c18f8a47702bf75ca3e33d/.circleci/config.yml#L20)

PR #5003, opened 2021-04-09
fix #4972: fix(directory table) - Show N/A for live experiments without results.

### LINKED ISSUE #4972: Show "not yet" for experiments that don't have results in the directory view
We shouldn't show a results link for experiments that won't have results yet in the directory view

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1165)
┆Issue Number: EXP-1165

emtwo:


app/experimenter/nimbus-ui/src/lib/dateUtils.ts
```diff
@@ -48,6 +48,26 @@ export function stringDateSubtract(dateA: string, dateB: string): string {
...
+  const experimentStartDate = new Date(String(startDate));
+
+  const resultsReadyDate = new Date();
+  resultsReadyDate.setDate(
+    experimentStartDate.getDate() + proposedEnrollment + DAYS_UNTIL_ANALYSIS,
+  );
+
+  return today.getDate() >= resultsReadyDate.getDate();
+}
+
```
jaredlockhart, line 70: You could do this on the backend, particularly if there's a constant involved.  All the constants right now live in the NimbusConstants file in Python, so it might get confusing later if we have a constant hidden off in the frontend?

emtwo: Sure we can move this to the backend

emtwo: Actually, I should note that the reason I didn't initially do it on the backend in the visualization API is because it takes some time to load the visualization data and I don't want to do that in the home page just to check if results are available.

So this logic would have to attach a field to the `experiment` object

jaredlockhart: Yeah exactly, make it a method on the NimbusExperiment class to look up the dates and compare to the constant since there's already lots of methods in there that do that same kind of thing and then attach it as a query property in the NimbusExperiment query type.

jaredlockhart (commented):
Oh good yes looks great!  Wasn't there some logic on the frontend that did this? Did that already land in main? Does it need to be taken out?

emtwo:
> Oh good yes looks great! Wasn't there some logic on the frontend that did this? Did that already land in main? Does it need to be taken out?

No it didn't land. I had it in initially in this PR then I removed it. You can look at the resolved comment you had above for that.

jaredlockhart (approved):
Oh I see yes I think this is much better, g/j 👍 👍 👍

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -444,6 +444,28 @@ def test_can_review_true_for_non_requesting_user(self):
...
+    def test_results_ready_true(self):
+        experiment = NimbusExperimentFactory.create()
+
+        NimbusChangeLogFactory.create(
+            experiment=experiment,
+            old_status=NimbusExperiment.Status.DRAFT,
+            new_status=NimbusExperiment.Status.LIVE,
+            changed_on=datetime.date(2019, 5, 1),
+        )
+        self.assertTrue(experiment.results_ready)
```
jaredlockhart, line 456: @jodyheavener Here take a look at this, it would be nice if we could also incorporate providing like a start date for things that need to take place in teh past.....

jaredlockhart: I did something similar in the legacy `create_with_status` where it picks a random day in teh past and does each lifecycle step 1 day later

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/factories/legacy.py#L144-L164

Issue #5004, opened 2021-04-09
Migrate nimbus-ui stories to Storybook 6 format


lmorchard:
We're still using the old Storybook 5 format for all our stories.

Since then, Storybook 6 has introduced a new [Component Story Format](https://storybook.js.org/docs/react/api/csf) based on plain ES6 exports.

We should consider reworking all the stories to match this new format, since the documentation for the old API-based approach is pretty hard to dig up now.

Bad news: we have over 43 story modules to convert.

Good news: the conversion process [is largely mechanical](https://github.com/mozilla/experimenter/commit/c7d88ad1d9a9b546522080a9db644b6cd2309118) and they can be done one at a time - old & new formats can co-exist with no configuration changes.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1186)
┆Issue Number: EXP-1186

lmorchard:
Now that I gave it a try and landed a PR, here's an example of what the mechanical transformation of old-to-new looks like:

https://github.com/mozilla/experimenter/commit/c7d88ad1d9a9b546522080a9db644b6cd2309118

PR #5005, opened 2021-04-09
issue #5004 chore(nimbus): convert ChangeApprovalOperation stories to v6 format

lmorchard:
Because:

- Storybook v6 introduces a new story format

This commit:

- Converts ChangeApprovalOperation stories to the new format

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.stories.tsx
@@ -4,7 +4,6 @@
 
 import { action } from "@storybook/addon-actions";
 import { withLinks } from "@storybook/addon-links";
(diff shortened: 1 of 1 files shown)


lmorchard:
See also: https://storybook.js.org/docs/react/api/csf

Maybe this should be a draft? I just converted this set of stories over to the new format in the laziest way possible. Worth discussing how this could be done better.

In particular, naming exports results in annoying auto-magically produced labels. I could use the `.storyName` property to define a more customized name, but that seems annoying.

jodyheavener (approved):
Love this, and if you wanna merge go for it, also happy to wait and have a convo about it. I think we're headed in that direction anyway. Maybe some docs (or a little blurb about which format you should write stories in) would be helpful, but I don't think that's necessary here

lmorchard:
Rebased, may go ahead and merge after CI green

PR #5018, opened 2021-04-12
Fix #5017 chore(project): disable codeql

### LINKED ISSUE #5017: CodeQL fails main on Dependabot merge
Womp womp:

Screen Shot 2021-04-12 at 11 13 52 AM

These failures are `CodeQL / Analyze (javascript)` and `CodeQL / Analyze (python)`

We don't need to worry about these failures because they're just failing due to the Action not being able to run. But they're still not nice to see and it would be in our best interested to have main _not_ failing where possible. The failure log actually leaves us a handy little note:

> Workflows triggered by Dependabot on the "push" event run with read-only access. Uploading Code Scanning results requires write access. To use Code Scanning with Dependabot, please ensure you are using the "pull_request" event for this workflow and avoid triggering on the "push" event for Dependabot branches. See https://docs.github.com/en/code-security/secure-coding/configuring-code-scanning#scanning-on-push for more information on how to configure these events.

We should follow this advice and fix our setup.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1189)
┆Issue Number: EXP-1189

clouserw:


jaredlockhart (approved):
bye codeql, we knew you too little 👋 👋 👋 👋 👋

Issue #4894, opened 2021-03-23
Redesign results API


emtwo:
The results API is a tad messy code to read and modify. It needs a bit of a redesign to make it more readable and maintainable.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1105)
┆Issue Number: EXP-1105

PR #5025, opened 2021-04-13
Fix #5024: follow-up to #4894: Fix checks for available analysis.

### LINKED ISSUE #5024: Nimbus results page shows "General application error"
E.g., https://experimenter.services.mozilla.com/nimbus/1-click-pin-during-onboarding/results

results error

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1194)
┆Issue Number: EXP-1194

emtwo:
This PR reverts frontend changes made when moving over to pydantic: https://github.com/mozilla/experimenter/commit/0dddf336def53fccb6aee1bb18e083e9679210bd

Instead it makes the server-side set `None` for empty values in the API response as it did prior to that change.

Tested this with 2 broken experiments and it's resolved the issue

### DIFF
--- a/app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
+++ b/app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
@@ -8,9 +8,7 @@ import { AnalysisData } from "./types";
 // `show_analysis` is the feature flag for turning visualization on/off.
 // `overall` will be `null` if the analysis isn't available yet.
 export const analysisAvailable = (analysis: AnalysisData | undefined) =>
-  analysis?.show_analysis === true &&
-  (Object.keys(analysis?.overall).length > 0 ||
-    Object.keys(analysis?.weekly).length > 0);
+  analysis?.show_analysis && (analysis?.overall || analysis?.weekly);
 
 export const analysisUnavailable = (analysis: AnalysisData | undefined) =>
   analysis && !analysisAvailable(analysis);
--- a/app/experimenter/visualization/api/v3/views.py
+++ b/app/experimenter/visualization/api/v3/views.py
@@ -22,11 +22,8 @@ METADATA_FOLDER = "metadata"
 
 
 def load_data_from_gcs(path):
-    return (
-        json.loads(default_storage.open(path).read())
-        if default_storage.exists(path)
-        else []
-    )
+    if default_storage.exists(path):
+        return json.loads(default_storage.open(path).read())
 
 
 def get_results_metrics_map(data, primary_outcomes, secondary_outcomes):
@@ -111,7 +108,9 @@ def analysis_results_view(request, slug):
     }
 
     for window in windows:
-        data = raw_data[window] = JetstreamData(__root__=get_data(recipe_slug, window))
+        data = raw_data[window] = JetstreamData(
+            __root__=(get_data(recipe_slug, window) or [])
+        )
         result_metrics, primary_metrics_set, other_metrics = get_results_metrics_map(
             data, experiment.primary_outcomes, experiment.secondary_outcomes
         )
@@ -130,6 +129,8 @@ def analysis_results_view(request, slug):
             ResultsObjectModel = create_results_object_model(data)
             data = ResultsObjectModel(result_metrics, data, experiment, window)
 
-        experiment_data[window] = data.dict(exclude_none=True)
+        response_data = data.dict(exclude_none=True) or None
+
+        experiment_data[window] = response_data
 
     return Response(experiment_data)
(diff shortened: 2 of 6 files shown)


app/experimenter/visualization/api/v3/views.py
```diff
@@ -111,12 +111,14 @@ def analysis_results_view(request, slug):
...
-        data = raw_data[window] = JetstreamData(__root__=get_data(recipe_slug, window))
+        data = raw_data[window] = JetstreamData(
+            __root__=(get_data(recipe_slug, window) or [])
+        )
         result_metrics, primary_metrics_set, other_metrics = get_results_metrics_map(
             data, experiment.primary_outcomes, experiment.secondary_outcomes
         )
 
-        if data and window == "overall":
+        if len(data) > 0 and window == "overall":
```
jaredlockhart, line 121: Ya, also fyi you can use [`__nonzero__`](https://docs.python.org/2/reference/datamodel.html#object.__nonzero__) to implement the bool check 👍

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -8,9 +8,7 @@ import { AnalysisData } from "./types";
 // `show_analysis` is the feature flag for turning visualization on/off.
 // `overall` will be `null` if the analysis isn't available yet.
 export const analysisAvailable = (analysis: AnalysisData | undefined) =>
-  analysis?.show_analysis === true &&
-  (Object.keys(analysis?.overall).length > 0 ||
-    Object.keys(analysis?.weekly).length > 0);
+  analysis?.show_analysis === true && (analysis?.overall || analysis?.weekly);
```
jaredlockhart, line 11: Wait do you have to do `=== true`?

app/experimenter/visualization/api/v3/views.py
```diff
@@ -25,7 +25,7 @@ def load_data_from_gcs(path):
     return (
         json.loads(default_storage.open(path).read())
         if default_storage.exists(path)
-        else []
+        else None
     )
```
jaredlockhart, line 29: Oh I just realized you could also write this as

```py
if default_storage.exists(path):
    return json.loads(default_storage.open(path).read())
```

since a function will return `None` by default

app/experimenter/visualization/api/v3/views.py
```diff
@@ -126,10 +128,14 @@ def analysis_results_view(request, slug):
...
             experiment_data["other_metrics"] = other_metrics
-        elif data and window == "weekly":
+        elif len(data) > 0 and window == "weekly":
             ResultsObjectModel = create_results_object_model(data)
             data = ResultsObjectModel(result_metrics, data, experiment, window)
 
-        experiment_data[window] = data.dict(exclude_none=True)
+        response_data = data.dict(exclude_none=True)
+        if len(response_data) == 0:
+            response_data = None
```
jaredlockhart, line 137: Should be able to do this as 

```py
response_data = data.dict(exclude_none=True) or None
```

right?

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -47,11 +47,10 @@ def test_analysis_results_view_no_data(self, status, mock_exists):
...
-                "metadata": [],
-                "weekly": {},
-                "overall": {},
-                "other_metrics": {},
+                "daily": None,
+                "metadata": None,
+                "weekly": None,
+                "overall": None,
                 "show_analysis": False,
             },
```
jaredlockhart, line 55: Is this the only test change we need to capture the failures this addresses?

jaredlockhart (commented):
Generally looks good, couple ideas for some optimizations and wondering whether the test coverage is sufficient?

app/experimenter/visualization/api/v3/views.py
jaredlockhart: Sorry sorry it's obviously [`__bool__`](https://docs.python.org/3/reference/datamodel.html#object.__bool__) in py3 XD my bad

app/experimenter/visualization/tests/api/test_views.py
emtwo: I believe so. All the failures were happening because the frontend was expecting None values instead of empty objects.

jaredlockhart: Sounds good 👍

jaredlockhart (approved):
Yep looks great! 👏 👏 👏 👏 
Thanks for the fast fix! 🎉

Issue #4800, opened 2021-03-11
Integrate front-end components for new review flow with real back-end GQL calls


data-sync-user:
Once the back-end work to support the new review flow has been completed, we should be able to wire up the components created throughout the rest of this epic with real GQL queries & mutations.

This should also be the point where any temporary type or test coverage ignore comments or mock API calls get cleaned up and removed. Documentation on how to do this is TBD with EXP-1063, but that issue has some links to pointers & examples.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1062)
┆Issue Number: EXP-1062

lmorchard:
A note on tweaks to isLaunchRequested and launchRequestedBy fields required here:

https://github.com/mozilla/experimenter/pull/4943#pullrequestreview-623613514

jodyheavener:
Note when working on this we need to take [this line](https://github.com/mozilla/experimenter/pull/4946#discussion_r608239693) out.

PR #5029, opened 2021-04-13
issue #4800 chore(nimbus): add publish_status to changelog listings in admin

lmorchard:
Because:

- it's useful to see full status transitions in admin

This commit:

- adds old_publish_status and new_publish_status to changelog listings in admin


jodyheavener (approved):
makes sense

PR #5030, opened 2021-04-13
fixes #4800 feat(nimbus): implement new review flow UI against GQL backend

### LINKED ISSUE #4800: Integrate front-end components for new review flow with real back-end GQL calls

lmorchard:
Because:

* we want to enable the new launch experiment review flow in the UI

This commit:

* start using real experiment data in ChangeApprovalOperations component

* rework ChangeApprovalOperations to switch default state based on
  experiment publishState, rather than changelog events

* tweaks to PageRequestReview to call real GQL mutations in response to
  launch review actions

* refactor ChangeApprovalOperations and PageRequestReview stories to
  storybook v6 format, move more data into mocks shared with tests

* adds a new KINTO_ADMIN_URL setting, defaulting to the production URL

* exposes kintoAdminUrl in GQL config query for use by UI to open remote
  settings

* removes the reviewApproval changelog event exposed in GQL, since it
  became unnecessary after another rework of ChangeApprovalOperations

* adds a new SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER setting for
  disabling access control to reviews in local development


.env.sample
```diff
@@ -43,4 +44,5 @@ STATSD_BACKEND=markus.backends.logging.LoggingMetrics
 STATSD_HOST=
 STATSD_PORT=
 STATSD_PREFIX=experimenter
-USE_GOOGLE_ANALYTICS=false
\ No newline at end of file
+USE_GOOGLE_ANALYTICS=false
+SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER=true
```
lmorchard, line 48: This setting has been really handy for local dev: repeatedly changing the `changedBy` in admin to unlock review access started getting old after the first few rounds.

app/experimenter/experiments/api/v5/mutations.py
```diff
@@ -65,8 +66,20 @@ def mutate(cls, root, info, input: ExperimentInput):
         return handle_with_serializer(cls, serializer)
 
 
+class TriggerCheckKintoPushQueueNow(graphene.Mutation):
+    message = ObjectField()
+
+    @classmethod
+    def mutate(cls, root, info):
+        nimbus_check_kinto_push_queue()
+        return cls(message="Triggered nimbus_check_kinto_push_queue task")
```
lmorchard, line 75: We didn't have an issue for adding this to the backend. But, [the Figma mock](https://www.figma.com/proto/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=218363%3A30350&scaling=min-zoom) has a note to call the Remote Settings API to immediately create the new record. That seemed a bit dicey to do literally, so I punted and just made an attempt to kick off the Celery task that leads to that result.

Capture

app/experimenter/experiments/models/nimbus.py
```diff
@@ -270,6 +270,12 @@ def allocate_bucket_range(self):
         )
 
     def can_review(self, reviewer):
+        if (
+            settings.SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER
+            and reviewer.email == settings.DEV_USER_EMAIL
+        ):
+            return True
```
lmorchard, line 277: Trying to keep this dev flag constrained to just the local dev user email.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx
```diff
@@ -49,30 +57,55 @@ export const ChangeApprovalOperations: React.FC<
...
+          ? ChangeApprovalOperationsState.ApproveOrReject
+          : ChangeApprovalOperationsState.ApprovalPending;
+      case EventKinds.Timeout:
+        return canReview
+          ? ChangeApprovalOperationsState.ApproveOrReject
+          : ChangeApprovalOperationsState.ApprovalPending;
+      default:
+        return ChangeApprovalOperationsState.None;
+    }
+  }, uiStateDependencies);
```
lmorchard, line 94: This ended up expanding in complexity a bit. Best solution I could come up with is to find which is the most recent of all the significant changelog events and treat that as the current state. That seems to correctly handle the case of a retry request after a rejection and a retry approval after a timeout.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx
```diff
@@ -49,30 +57,55 @@ export const ChangeApprovalOperations: React.FC<
...
-  const resetDraftUIState = useCallback(() => setUIState(defaultUIState), [
+
+  const resetUIState = useCallback(() => setUIState(defaultUIState), [
     defaultUIState,
     setUIState,
   ]);
 
+  // Whenever any of the UI default state dependencies change (i.e. via
+  // polling or refetch), override and reset the state.
+  useEffect(() => resetUIState(), uiStateDependencies);
```
lmorchard, line 107: Ran into a small bug fixed by this, trying to work out how to trigger it in tests to ensure it's exercised.

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.stories.tsx
```diff
@@ -3,134 +3,106 @@
...
+  decorators: [withLinks],
+};
+
+export const draftStatus = storyWithExperimentProps({
+  status: NimbusExperimentStatus.DRAFT,
+});
+
+export const previewStatus = storyWithExperimentProps({
+  status: NimbusExperimentStatus.PREVIEW,
+});
```
lmorchard, line 42: Tackling a bit of issue #5004 here by refactoring all the stories to v6 format, which also helps DRY some things up.

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -65,52 +47,59 @@ const PageRequestReview = ({
...
         },
-      ].map((changes: Partial<ExperimentInput>) => async () => {
+        {
+          status: NimbusExperimentStatus.DRAFT,
+          publishStatus: NimbusExperimentPublishStatus.IDLE,
+        },
+      ].map((baseDataChanges: Partial<ExperimentInput>) => async (
+        inputEvent?: any, // ignored, allows direct use as event handler
+        submitDataChanges?: Partial<ExperimentInput>,
+      ) => {
```
lmorchard, line 92: Kind of a hack, but helps keep this common mutation handler generic.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.stories.tsx
```diff
@@ -35,81 +44,61 @@ export default {
...
-      approvalEvent: mockChangelog("def@mozilla.com"),
-      timeoutEvent: mockChangelog("ghi@mozilla.com"),
-      canReview: true,
-    }}
-  />
-);
+export const ReviewPendingInRemoteSettingsUserCanReview = subjectWithProps({
+  ...reviewPendingInRemoteSettingsBaseProps,
+  canReview: true,
+});
```
lmorchard, line 66: Refactoring toward issue #5004 but also extracting some more mocks to share with tests.

lmorchard (commented):
Took a first pass through as self-review to point out some weird things. Coming back tomorrow to finish up tests.

jaredlockhart:
I just landed some of the new backend changes that this'll link to so rebase this to get those and then I think all the hookup points should be there 👍

app/experimenter/experiments/api/v5/mutations.py
jaredlockhart: Oh oh.  Yeah okay so, I wasn't thinking of it like that.  If we want to do that we'd have to move some things around and change some logic.  The way it's setup now is

- Owner requests review
- Reviewer approves
- UI now shows above, but experiment is not yet in RS
- Next scheduled invocation of push task kicks in 0-5 minutes later
- Now it's available in RS to review

If we want to set it up so that when it moves to APPROVED we call the task immediately, we'd have to do a bunch of refactoring.  Like probably a lot.  So I would hazard that we stick with the scheduled approach for now and if we want to do it more immediately we'll have to scope and plan that separately.  cc @AnaMedinac

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Oh interesting cool yeah this is very important not just for local testing but also for the integration tests.  If we do it this way that means we don't need to deal with faking multiple accounts in the integration tests. cc @jrbenny35 this means we should be able to do the whole test with one account the way we did for legacy 👍 

Also @lmorchard instead of flagging this on a new setting we could also just flag it on DEBUG to obviate the need for a new setting?

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx
jaredlockhart: Shouldn't you be able to switch this on the pair of status/publish_status rather than looking at the most recent changelog?

app/experimenter/experiments/models/nimbus.py
lmorchard: Well, I didn't want to tie it DEBUG in general, because I *did* want to enable the access control for *some* of my dev work.

jaredlockhart: Oh interesting, are there times when you're doing dev with DEBUG=False?

app/experimenter/experiments/api/v5/mutations.py
jodyheavener: If we keep the scheduled approach I'd probably recommend a note (tooltip?) in this container to explain the delay. Could be pretty confusing otherwise since you might not realize what's happening -- is it broken? do I need to talk to someone? why does it sometimes show up immediately and sometimes I have to wait 5 minutes?

app/experimenter/experiments/models/nimbus.py
lmorchard: No, I'm doing dev with DEBUG=True but I want the review access control enabled

jaredlockhart: Oh I see waht you mean you want to shut off while you're deving right yep makes sense 👍

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx
jaredlockhart: Oh yeah that makes sense, it should show the current UI state based solely on the publish_status, the status should just determine whether it's launching or ending.  I'm hesitant about building logic strongly around changelogs, since they can change and move around a lot more than the model does.

.env.sample
jodyheavener: This is SO dang handy

jodyheavener (commented):
Few thoughts here and there, but the flows all work for me. Looks and works great.

I'm also kind of thinking that when the experiment is not IDLE we should lower the polling delay. In almost every case I reloaded the page manually and it makes me question if the polling is beneficial at the current reload speed. 🤷

lmorchard:
> I'm also kind of thinking that when the experiment is not IDLE we should lower the polling delay. In almost every case I reloaded the page manually and it makes me question if the polling is beneficial at the current reload speed. 🤷

Could be worth a follow-up issue? Wouldn't want to tackle that here

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.stories.tsx
```diff
@@ -3,134 +3,89 @@
...
+);
+
+export const reviewPendingCanReview = storyWithExperimentProps(
+  { ...reviewPendingBaseProps, canReview: true },
+  "Review pending in Remote Rettings, user can review",
+);
+
+export const reviewTimedoutCanReview = storyWithExperimentProps(
+  { ...reviewTimedoutBaseProps, canReview: true },
+  "Review timed out in Remote Settings, user can review",
```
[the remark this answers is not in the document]
lmorchard: I think this is a bug in the timeout celery task - it rolls back to IDLE rather than REVIEW publishStatus. I've got a PR here:

https://github.com/mozilla/experimenter/pull/5035

Both the UI component and the `latest_timeout` model method are looking for a transition from WAITING to REVIEW. Transitioning from WAITING to IDLE looks like a rejection.

jaredlockhart:
> > I'm also kind of thinking that when the experiment is not IDLE we should lower the polling delay. In almost every case I reloaded the page manually and it makes me question if the polling is beneficial at the current reload speed. 🤷
> 
> Could be worth a follow-up issue? Wouldn't want to tackle that here

@jodyheavener Do you want to file this so we don't lose this idea? 🙏

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.test.tsx
```diff
@@ -215,18 +243,58 @@ describe("PageRequestReview", () => {
...
+    render(<Subject mocks={[mock, mutationMock]} />);
+    const rejectButton = await screen.findByTestId("reject-request");
+    fireEvent.click(rejectButton);
+    const rejectSubmitButton = await screen.findByTestId("reject-submit");
+    const rejectReasonField = await screen.findByTestId("reject-reason");
+    expect(rejectSubmitButton).not.toBeEnabled();
+    fireEvent.change(rejectReasonField, {
+      target: { value: expectedReason },
+    });
+    fireEvent.blur(rejectReasonField);
```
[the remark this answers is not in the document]
jaredlockhart: Yeah I thought we discussed this collectively and agreed not to do this?

.env.sample
```diff
@@ -27,6 +27,7 @@ HOSTNAME=localhost
 KINTO_HOST=http://kinto:8888/v1
 KINTO_PASS=experimenter
 KINTO_USER=experimenter
+KINTO_ADMIN_URL=http://localhost:8888/v1/admin
```
jaredlockhart, line 30: We can derive this in settings like this:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L374

app/experimenter/experiments/api/v5/types.py
```diff
@@ -186,9 +185,6 @@ def resolve_can_review(self, info):
     def resolve_review_request(self, info):
         return self.changes.latest_review_request()
 
-    def resolve_review_approval(self, info):
-        return self.changes.latest_review_approval()
-
```
jaredlockhart, line 191: Didn't we just add this?  Why are we taking it out?

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.test.tsx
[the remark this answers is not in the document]
lmorchard: IIRC, we don't disable buttons so work in progress can be saved. But we *do* disable when the mutation itself can't be submitted because the required data is missing. Here, we don't get a usable server-side submission error if we allow submission - it's a GQL type error that we can't parse into something sensible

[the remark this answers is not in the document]
lmorchard: I can poke at this a bit though and see if I can switch this to a client-side error

app/experimenter/experiments/api/v5/types.py
lmorchard: Removing the reviewApproval changelog event exposed in GQL, since it became unnecessary after another rework of ChangeApprovalOperations to use publishStatus

.env.sample
lmorchard: Derive from what? I tried KINTO_HOST, but it's using the within-docker URL and not the externally-accessible URL

app/experimenter/experiments/api/v5/mutations.py
lmorchard: So, this mutation should be taken out?

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.stories.tsx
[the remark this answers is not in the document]
lmorchard: Just rolled PR #5035 into this PR.

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.test.tsx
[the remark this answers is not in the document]
lmorchard: Okay, I removed the button disabling logic. Also noticed a bug where the client-side validation message that *should* have been there wasn't showing up. So now, you should be able to click the Reject button and see the client-side message but not have the mutation attempted until the form is valid.

app/experimenter/experiments/api/v5/mutations.py
lmorchard: I'm just going to remove the mutation - we can follow up with an enhancement, possibly as part of [EXP-1200](https://jira.mozilla.com/browse/EXP-1200), since responsiveness of the background tasks is related here too.

lmorchard: https://jira.mozilla.com/browse/EXP-1201

jodyheavener (approved):
As far as frontend goes I think this is great! 👏

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -65,55 +47,63 @@ const PageRequestReview = ({
...
-  /* istanbul ignore next until EXP-1055 & EXP-1062 done */
-  const [
-    startRemoteSettingsApproval,
-    { loading: startRemoteSettingsApprovalLoading },
-  ] = useFakeMutation();
-
-  // TODO: EXP-1062 wrap these new mutations in setSubmitError handling like updateExperiment uses below.
+  const startRemoteSettingsApproval = async () => {
+    await triggerCheckKintoPushQueueNow();
+    window.open(kintoAdminUrl!, "_blank");
```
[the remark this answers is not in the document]
lmorchard: https://jira.mozilla.com/browse/EXP-1202

jodyheavener:
Oh also I wanted to add that the Storybook changes are just... :chefs-kiss:

.env.sample
jaredlockhart: Oh consarnit it's different on local but not on stage/prod.  Well if we make it a separate env var then we've put the remote settings domain in twice, so if we ever go to update it we'll have to remember to do it in both places.  Or we could move the localhost hard coded into settings and then just do something like

```py
KINTO_ADMIN_URL = 'https://localhost:8888/v1/admin' if DEBUG else urljoin(KINTO_HOST, 'admin')
```

Both have their disadvantages.

jaredlockhart: And we'll have to file a bugzilla to deploy the new env var before this lands if we add it.

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Oh okay that makes sense 👍 Bye code!

.env.sample
jaredlockhart, line 48: Also sort alphabetically here

.env.sample
lmorchard: Hmm, if KINTO_HOST is accessible on stage & prod, I could derive it in the default and just set it explicitly for dev.

jaredlockhart: oh hype great thinking 👏 👏 👏 👏 👏  then we just have to be aware that there's an env var in the sample that doesn't apply to stage/prod, but sample doesn't get used on stage/prod so that should probably just work itself out 👍

jaredlockhart:
Okay first off WOW this is slick @lmorchard !  Looks great, working great so far!

Just found one thing.

1) Request launch
2) Approve
3) Wait for timeout
4) Timeout message appears, back to approve/reject
5) Reject
7) Request launch
8) Now it says timeout even though it's a new launch flow, I'd expect it not to say timeout now

lmorchard:
> Now it says timeout even though it's a new launch flow, I'd expect it not to say timeout now

Yeah... I think that's because we have a "latest" timeout. All the changelog events are the latest of their kind, without really a way to decide if one has been made irrelevant by subsequent actions. I was kind of trying to get around that by sorting the events, but that didn't quite do the trick either

jaredlockhart:
> > Now it says timeout even though it's a new launch flow, I'd expect it not to say timeout now
> 
> Yeah... I think that's because we have a "latest" timeout. All the changelog events are the latest of their kind, without really a way to decide if one has been made irrelevant by subsequent actions. I was kind of trying to get around that by sorting the events, but that didn't quite do the trick either

Oh then yeah we should modify the latest_timeout to return none if it's not the most recent changelog entry?

lmorchard:
> If that's out of scope here we can file that as a followup

I'd suggest as a followup, if only because it's a little edge-casey and this PR has gotten pretty chonky 😅

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -491,6 +492,24 @@ def test_results_ready_false(self):
         )
         self.assertFalse(experiment.results_ready)
 
+    @parameterized.expand([(settings.DEV_USER_EMAIL, True), ("jdoe@mozilla.org", False)])
+    @mock.patch("django.conf.settings.SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER", True)
```
jaredlockhart, line 496: Oh you don't have to use mock for this django has a dealy for it

﻿﻿https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.override_settings

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx
```diff
@@ -40,39 +39,44 @@ export const ChangeApprovalOperations: React.FC<
...
+          ? ChangeApprovalOperationsState.ShowRemoteSettingsPending
+          : ChangeApprovalOperationsState.ApprovalPending;
+      case NimbusExperimentPublishStatus.REVIEW:
+        return canReview
+          ? ChangeApprovalOperationsState.ShowFormApproveOrReject
+          : ChangeApprovalOperationsState.ApprovalPending;
+      default:
+        return ChangeApprovalOperationsState.None;
+    }
+  }, [publishStatus, canReview]);
```
jaredlockhart, line 65: So good!!! 🎉 🎉 🎉 🎉

app/experimenter/settings.py
```diff
@@ -409,7 +409,7 @@
 KINTO_COLLECTION_NIMBUS_DESKTOP = "nimbus-desktop-experiments"
 KINTO_COLLECTION_NIMBUS_MOBILE = "nimbus-mobile-experiments"
 KINTO_COLLECTION_NIMBUS_PREVIEW = "nimbus-preview"
-
+KINTO_ADMIN_URL = config("KINTO_ADMIN_URL", default=urljoin(KINTO_HOST, "/v1/admin/"))
```
jaredlockhart, line 412: Host already has v1 in it https://github.com/mozilla/experimenter/blob/main/.env.sample#L27

jaredlockhart (approved):
This is so fantastic @lmorchard The UI elements are really clean and well laid out, everything behaves correctly, I'm sitting here clicking through it all going 'WOW WOW WOW' seeing it all finally work.  This whole PR is very readable and clear.  This is really great work.

Couple tiny fixups R+WC 🏆 🏆 🏆 🏆

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
lmorchard: Oh.. 🤦 I've totally used that before too

lmorchard:
Made some more review tweaks, rebased against main, let's see what CI thinks

lmorchard:
>Now it says timeout even though it's a new launch flow, I'd expect it not to say timeout now

Also filed another follow-up: https://jira.mozilla.com/browse/EXP-1204

lmorchard:
Recorded a fresh demo walking through this and didn't see major issues, I think I'll take the approvals and run

PR #5033, opened 2021-04-14
fix(project): regenrate gql schema

jodyheavener:
This was out of date on main.


jaredlockhart (commented):
How did this land on main being out of date?  This change just looks like it's reordering?  What's happening here?

jodyheavener:
ghosts that's how

jodyheavener:
damn ghosts

jaredlockhart:
Just to capture what we talked about in slack, it looks like this is happening when Apollo converts the client side gql queries into types, which happens outside of the check that runs in `make check`.  I'm inclined to leave this alone and let changes like this linger on main and sneak in with other prs that actually change the gql schema, since I don't think it breaks anything.

If it comes up again in a more breaking way we can revist it.  Thanks for catching!

PR #5035, opened 2021-04-14
issue #4800 fix(nimbus): tweak timeout to rollback to REVIEW publish status

lmorchard:
Because:

* we want the reviewer to retry approval and not the experimenter to
  retry review request

This commit:

* switches to REVIEW publish status instead of IDLE on timeout

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -87,7 +87,7 @@ def handle_pending_review(application, kinto_client):
     )
 
     if experiment.has_state(experiment.SHOULD_TIMEOUT):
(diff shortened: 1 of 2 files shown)


lmorchard:
Noticed this didn't work quite as expected while recording a demo. Seems we want this to be REVIEW on timeout, not IDLE

Capture

jaredlockhart:
Yeah I was expecting this to land as part of your other PR so that it rolled back to REVIEW when the UI to move forward from there is landed.  If we land this now it'll roll back to a state where there is no UI and experiments will just get stuck?

lmorchard:
Okay, I can roll this into the other PR

PR #5036, opened 2021-04-15
fix(project): code format & generate types

jodyheavener:
Looks like we needed a code_format and generate_types


jaredlockhart (commented):
Oh thanks for catching this @jodyheavener but we need to actually fix this or it'll just happen again, ticket here: https://github.com/mozilla/experimenter/issues/5038

jodyheavener:
@jaredlockhart Totally but can we manually update it until then? Just want to avoid a larger than necessary diff 😀

jaredlockhart:
But without fixing it, this pr accomplishes nothing?

jaredlockhart (approved):
Nvm let's just land it 👍

jodyheavener:
@jaredlockhart I mean, it allows other PRs to be opened and merged without creating unnecessarily larger diffs?

jaredlockhart:
This might not rebased on latest main with the new circle definitions

Issue #4403, opened 2021-01-11
Summary/review table shows extra "not set" branch with only one branch saved


jaredlockhart:
If I create a new experiment, hit the branches form, delete the second empty branch, and just create a control branch, save, navigate away, then navigate back, it re-adds the second empty branch on top of the control branch I saved.  This could be confusing for experiments that are designed to only have a single branch.

Also on the review page, it shows the second unsaved branch as being incomplete:

<img width="1144" alt="Screen Shot 2021-01-11 at 4 24 00 PM" src="https://user-images.githubusercontent.com/119884/104240074-7cda3680-5429-11eb-9bc4-6fec08ea0b2c.png">

You should be able to create an experiment with only a single branch and have that be the only branch saved/displayed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-821)
┆Issue Number: EXP-821

jaredlockhart:
It doesn't show up in the recipe so that party's correct:


```json
  "branches": [
    {
      "slug": "c",
      "ratio": 1
    }
  ],
```

LZoog:
This is kind of a dupe of #4403 but since this issue also mentions the summary table, I'll leave them both open and close them in one PR and consider this issue the summary page issue and the former the one that deals with the form.

Issue #4150, opened 2020-12-07
If an experiment has saved branches, it shouldn't show empty default branches


jaredlockhart:
Right now if you create an experiment, giv eit only a control branch, save, and reload the page, you'll se a second empty branch.  If an experiment has saved branches loading the branches form should only show the saved branches.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-675)
┆Issue Number: EXP-675

data-sync-user:
➤ Lauren Zugai commented:

I'm going to close this as a dupe of EXP-821 / #4403

PR #5046, opened 2021-04-16
 fixes #4959 feat(nimbus): Display default branch names, alter TableBranches

### LINKED ISSUE #4959: Branches should be called "control" and "treatment" by default
Based on a feature request from Grisha:
> It’ll be nice if the control/treatment were standard options in a dropdown menu, to prevent me from having to input these manually & making a typo

I don't think dropdown is right since we need to support other values in desktop (and will in mobile soon), but having `control` and `treatment` be the default values seems like an obvious win

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1156)
┆Issue Number: EXP-1156

LZoog:
fixes #4959

Because:
* It's more convenient for users to have these filled out by default
* The branches table shouldn't display info until users have saved a branch, which requires a name

This commit:
* Displays 'control' in the branch 'name' input and 'treatment' in the first reference branch if experiment branch details have not already been saved
* Displays custom NotSet copy in TableBranches if at least one branch name hasn't been set
* Moves some logic out of Summary into TableBranches
* Adds extra space at the bottom of the Summary component for readability
* Removes 'New' from 'New treatment X' when new treatment branches are added

We can't just add `value="whatever"` in the input because it has to go through `react-hook-form` as a default value.

===

[note: no longer applicable, but leaving this note since there were some comments in response to it]
I'm opening this as a draft before adding a new Storybook state and test to make sure we want to do this client-side. Is it confusing to users if they see "control" and "treatment" filled out already, when if they go to the Summary page, they're shown as "not set"? I feel perhaps this is fine since they have to fill out branch descriptions and when they click "save" the branches will _actually_ have their names set, but I wonder if it's preferable to set the names when an experiment is created, though maybe we want the experiment data in the BE to be "empty" until a user has actually saved it.

branches-default

===

[update]
I spoke with Kate and Ana to confirm this direction and this copy.

**Please feel free to push to this branch and/or merge this PR since I'm out next week!** Or, leave it and I'll get to it on 4/26.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/NotSet/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/NotSet/index.stories.tsx
@@ -6,4 +6,6 @@ import { storiesOf } from "@storybook/react";
 import React from "react";
 import NotSet from ".";
 
-storiesOf("Components/NotSet", module).add("basic", () => <NotSet />);
+storiesOf("Components/NotSet", module)
+  .add("basic", () => <NotSet />)
+  .add("custom copy", () => <NotSet copy="YOU'RE the one that's not set." />);
--- a/app/experimenter/nimbus-ui/src/components/NotSet/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/NotSet/index.tsx
@@ -5,12 +5,14 @@
 import React from "react";
 
 const NotSet = ({
+  copy = "Not set",
   "data-testid": testid = "not-set",
 }: {
+  copy?: string;
   "data-testid"?: string;
 }) => (
   <span className="text-danger" data-testid={testid}>
-    Not set
+    {copy}
   </span>
 );
 
--- a/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
@@ -42,6 +42,39 @@ describe("FormBranches", () => {
     expect(onNext).not.toHaveBeenCalled();
   });
 
+  it("sets expected default name values", async () => {
+    render(
+      <SubjectBranches
+        {...{
+          experiment: {
+            ...MOCK_EXPERIMENT,
+            referenceBranch: {
+              ...MOCK_EXPERIMENT.referenceBranch!,
+              name: "",
+            },
+            treatmentBranches: [
+              {
+                name: "",
+                slug: "",
+                description: "",
+                ratio: 0,
+                featureValue: null,
+                featureEnabled: false,
+              },
+            ],
+          },
+        }}
+      />,
+    );
+    expect(
+      (screen.getByTestId("referenceBranch.name") as HTMLInputElement).value,
+    ).toEqual("control");
+    expect(
+      (screen.getByTestId("treatmentBranches[0].name") as HTMLInputElement)
+        .value,
+    ).toEqual("treatment");
+  });
+
   it("calls onSave with extracted update when save button clicked", async () => {
     const onSave = jest.fn();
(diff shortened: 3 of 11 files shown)


k88hudson:
I think also it would make sense in the Summary view just to say "Branches are not saved yet" or similar, since before they are saved they aren't real branches anyway

jaredlockhart (commented):
Oh oh, we can set those defaults on the backend by setting defaults in the serializer, rather than hardcoding a bunch of comparisons in the frontend

LZoog:
> Oh oh, we can set those defaults on the backend by setting defaults in the serializer, rather than hardcoding a bunch of comparisons in the frontend

Right, that's what I was thinking with my initial comment. If we still want to show "branch data has not yet been saved" or whatever in the branches table, we can check if a branch's description has been filled out... but if we set these as defaults in the serializer, since the names will actually already be set, will we _want_ the table to be hidden until they add descriptions? 🤔 @k88hudson

k88hudson:
I don't have strong opinions either way (neither would be bad), but doing it this way and hiding the table until the branches are saved are probably what I would do 🤷

LZoog:
PR has been updated based on comments here and a quick discussion in Slack.

app/experimenter/nimbus-ui/src/components/NotSet/index.stories.tsx
```diff
@@ -6,4 +6,6 @@ import { storiesOf } from "@storybook/react";
 import React from "react";
 import NotSet from ".";
 
-storiesOf("Components/NotSet", module).add("basic", () => <NotSet />);
+storiesOf("Components/NotSet", module)
+  .add("basic", () => <NotSet />)
+  .add("custom copy", () => <NotSet copy="YOU'RE the one that's not set." />);
```
jaredlockhart, line 11: 😨

jaredlockhart (commented):
Very nice @LZoog 👍 

Dunno if we wanna make these more consistent in terms of capitalization/naming:

Screen Shot 2021-04-19 at 8 54 51 AM

?

LZoog:
@jaredlockhart 

> Dunno if we wanna make these more consistent in terms of capitalization/naming:

Good call out. I messaged Ana, she'd prefer we use "treatment A" "treatment B" etc., so for this PR I took out the word "new" and created a follow up (#5105) to change those to letters instead of numbers:

jaredlockhart (approved):
Looking good 👍 

Noticed a couple things, some of them are already filed as separate issues:

1) Saving on the branches page doesn't update the 'ready for review' status, but navigating to another page does. Maybe worth filing as a followup?

Screen Shot 2021-04-26 at 1 05 46 PM



2) The summary page seems to expect at least 2 branches but the branch page lets you have only a single page, so it shows a phantom second branch that's 'not set'.  Filed as https://github.com/mozilla/experimenter/issues/4403

Screen Shot 2021-04-26 at 1 06 50 PM



3) If you go down to 1 branch, navigate away from the branches page, then navigate back, there's now a second 'treatment' branch sitting there, even though we saved it having only a single 'control' branch.  Filed as https://github.com/mozilla/experimenter/issues/4150

Screen Shot 2021-04-26 at 1 07 05 PM



Happy to R+ this as it is and then address stuff in followups.

LZoog:
> Saving on the branches page doesn't update the 'ready for review' status, but navigating to another page does. Maybe worth filing as a followup?

This should be addressed when we redo the entire "required for launch" errors epic.

> If you go down to 1 branch, navigate away from the branches page, then navigate back, there's now a second 'treatment' branch sitting there, even though we saved it having only a single 'control' branch. Filed as #4150

Huh yeah, that crossed my mind but I thought it may be wanted behavior. Before this change, if you save only one branch and navigate back to the branches page, the 2nd new/blank branch still shows.
image

PR #5072, opened 2021-04-20
fixes #5044 fix(nimbus): enforce review flow order in reported events

### LINKED ISSUE #5044: Review timeout still displayed for a new launch flow after rejection
Context from [PR comment](https://github.com/mozilla/experimenter/pull/5030#issuecomment-820717909):

1. Request launch
1. Approve
1. Wait for timeout
1. Timeout message appears, back to approve/reject
1. Reject
1. Request launch
1. Now it says timeout even though it's a new launch flow, I'd expect it not to say timeout now

Think that we will need to finesse the model methods that report latest changelog events for various transitions (i.e. request, rejection, timeout).

Maybe if there's a review request that's newer than a rejection or a timeout, the rejection or timeout doesn't get returned?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1204)
┆Issue Number: EXP-1204

lmorchard:
Because:

* we don't want to report rejections & timeouts for the wrong statuses
  or after retried review cycles

This commit:

* checks to ensure that every reported review flow changelog event
  happens after the previous expected event

* ignores changelog events for review flows not associated with the
  current experiment status

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -9,7 +9,7 @@ from django.contrib.postgres.fields import ArrayField
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import MaxValueValidator
 from django.db import models
-from django.db.models import Q
+from django.db.models import F, Q
 from django.urls import reverse
 from django.utils import timezone
 
@@ -179,7 +179,7 @@ class NimbusExperiment(NimbusConstants, models.Model):
             return self.TARGETING_CONFIGS[self.targeting_config_slug]
 
     def latest_change(self):
-        return self.changes.order_by("-changed_on").first()
+        return self.changes.latest_change()
 
     @property
     def treatment_branches(self):
@@ -458,53 +458,49 @@ class NimbusFeatureConfig(models.Model):
 
 
 class NimbusChangeLogManager(models.Manager):
+    def latest_change(self):
+        return self.all().order_by("-changed_on").first()
+
     def latest_review_request(self):
         return (
             self.all()
             .filter(
-                Q(old_status=NimbusExperiment.Status.DRAFT)
-                | Q(old_status=NimbusExperiment.Status.PREVIEW),
+                Q(old_status=F("new_status"))
+                | Q(
+                    old_status=NimbusExperiment.Status.PREVIEW,
+                    new_status=NimbusExperiment.Status.DRAFT,
+                )
             )
             .filter(
                 old_publish_status=NimbusExperiment.PublishStatus.IDLE,
-                new_status=NimbusExperiment.Status.DRAFT,
                 new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
             )
             .order_by("-changed_on")
         ).first()
 
     def latest_rejection(self):
-        return (
-            self.all()
-            .filter(
-                Q(old_publish_status=NimbusExperiment.PublishStatus.REVIEW)
-                | Q(old_publish_status=NimbusExperiment.PublishStatus.WAITING),
+        change = self.latest_change()
+        if (
+            change
+            and change.old_status == change.new_status
+            and change.old_publish_status
+            in (
+                NimbusExperiment.PublishStatus.REVIEW,
+                NimbusExperiment.PublishStatus.WAITING,
             )
-            .filter(
-                new_publish_status=NimbusExperiment.PublishStatus.IDLE,
-            )
-            .order_by("-changed_on")
-        ).first()
+            and change.new_publish_status == NimbusExperiment.PublishStatus.IDLE
+        ):
+            return change
 
     def latest_timeout(self):
-        return (
-            self.all()
-            .filter(
-                Q(
-                    old_status=NimbusExperiment.Status.DRAFT,
-                    new_status=NimbusExperiment.Status.DRAFT,
-                )
-                | Q(
-                    old_status=NimbusExperiment.Status.LIVE,
-                    new_status=NimbusExperiment.Status.LIVE,
-                )
-            )
-            .filter(
-                old_publish_status=NimbusExperiment.PublishStatus.WAITING,
-                new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
-            )
-            .order_by("-changed_on")
-        ).first()
+        change = self.latest_change()
+        if (
+            change
+            and change.old_status == change.new_status
+            and change.old_publish_status == NimbusExperiment.PublishStatus.WAITING
+            and change.new_publish_status == NimbusExperiment.PublishStatus.REVIEW
+        ):
+            return change
 
 
 class NimbusChangeLog(models.Model):
--- a/app/experimenter/experiments/tests/api/v5/test_queries.py
+++ b/app/experimenter/experiments/tests/api/v5/test_queries.py
@@ -543,15 +543,17 @@ class TestNimbusQuery(GraphQLTestCase):
         user_email = "user@example.com"
         experiment = NimbusExperimentFactory.create_with_status(
             NimbusExperiment.Status.DRAFT,
-            publish_status=NimbusExperiment.PublishStatus.APPROVED,
+            publish_status=NimbusExperiment.PublishStatus.IDLE,
         )
-        experiment.publish_status = NimbusExperiment.PublishStatus.WAITING
-        experiment.save()
-        generate_nimbus_changelog(experiment, experiment.owner)
-
-        experiment.publish_status = NimbusExperiment.PublishStatus.REVIEW
-        experiment.save()
-        generate_nimbus_changelog(experiment, experiment.owner)
+        for publish_status in (
+            NimbusExperiment.PublishStatus.REVIEW,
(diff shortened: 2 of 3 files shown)


lmorchard:
I *think* this resolves the issue of stale / out-of-sequence review flow lifecycle events. I don't think any of these events should happen outside the same experiment status (except for preview-to-draft) and should always happen in the same sequence

app/experimenter/experiments/api/v5/types.py
```diff
@@ -185,10 +185,10 @@ def resolve_can_review(self, info):
...
-        return self.changes.latest_review_request()
+        return self.review_request
 
     def resolve_rejection(self, info):
-        return self.changes.latest_rejection()
+        return self.rejection
 
     def resolve_timeout(self, info):
-        return self.changes.latest_timeout()
+        return self.timeout
```
jodyheavener, line 194: Interesting, I thought there was some magic that could infer `resolve_*` when no additional work/params/calls were involved to retrieve a property of the same name

app/experimenter/experiments/models/nimbus.py
```diff
@@ -458,51 +494,73 @@ def __str__(self):  # pragma: no cover
...
+
+    PUBLISH_STATUS_FROM_REVIEW_TO_IDLE = Q(
+        old_publish_status=NimbusExperiment.PublishStatus.REVIEW,
+        new_publish_status=NimbusExperiment.PublishStatus.IDLE,
+    )
+
+    PUBLISH_STATUS_FROM_WAITING_TO_IDLE = Q(
+        old_publish_status=NimbusExperiment.PublishStatus.WAITING,
+        new_publish_status=NimbusExperiment.PublishStatus.IDLE,
+    )
```
jodyheavener, line 527: Should these live alongside the [other ones](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L199) in Nimbus constants?

jodyheavener (approved):
Couple clarifying thoughts, but this makes sense to me, though I will defer to the wisdom of @jaredlockhart

app/experimenter/experiments/models/nimbus.py
lmorchard: I stuck them here per [this advice over here](https://github.com/mozilla/experimenter/issues/5044#issuecomment-821355051):

> it occurs to me if we're going to have ones for different models they shoudln't live in NimbusConstants they should live right on the model

app/experimenter/experiments/api/v5/types.py
lmorchard: Oh hey! Yeah, you're right, I just kind of tweaked my way to this code change but forgot they could just be removed

app/experimenter/experiments/models/nimbus.py
```diff
@@ -301,6 +301,42 @@ def results_ready(self):
...
+    def _enforce_review_changelog_sequence(
+        self, previous_changelog_event, changelog_event
+    ):
+        if (
+            previous_changelog_event is None
+            or changelog_event is None
+            or previous_changelog_event.changed_on > changelog_event.changed_on
+        ):
+            return None
+        return changelog_event
```
jaredlockhart, line 320: Yeah I see what you're getting at here, but I think there's a simpler way to do this.

Down in the manager, just

1) get only the most recent changelog
2) check if it's a review request
3) if it is, return it, otherwise return None

rathern than doing this on the instance and comparing statuses and dates, etc

That should accomplish the same thing in less work, no?

jaredlockhart (commented):
Yep I like these patterns, I want to do more of this Q based filtering!

lmorchard:
Alright, reworked to just fetch the latest changelog event and check whether it meets the condition of review request / rejection / timeout. Passes all the same tests as the bit I was doing with date comparisons.

The one thing that's maybe not a showstopper is that the rejection reason disappears from the UI after the first next change to the experiment, whereas before it would persist until the next review request. Hadn't written a test asserting that though

app/experimenter/experiments/models/nimbus.py
lmorchard: Reworking to just use the latest changelog event makes all these queries obsolete

jaredlockhart:
Yep looks right to me, but I think the frontend isn't setup to handle the null case properly:

Screen Shot 2021-04-20 at 4 22 24 PM

app/experimenter/experiments/models/nimbus.py
```diff
@@ -458,53 +457,48 @@ def __str__(self):  # pragma: no cover
...
-                    new_status=NimbusExperiment.Status.LIVE,
-                )
-            )
-            .filter(
-                old_publish_status=NimbusExperiment.PublishStatus.WAITING,
-                new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
-            )
-            .order_by("-changed_on")
-        ).first()
+        event = self.latest_change()
```
jaredlockhart, line 494: Also I usually call this `change`, but that's just preference 🤷‍♀️

jaredlockhart:
The types list them as nullable

https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/types/getExperiment.ts#L116-L118

so maybe we're just not checking them properly?

Maybe here?

https://github.com/mozilla/experimenter/blob/938444a5548a12eec08eb0ce9fd1d94b40441933/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormApproveOrReject.tsx#L44

lmorchard:
> Yep looks right to me, but I think the frontend isn't setup to handle the null case properly:

Ah, no, this is a problem with fetching just the latest change.

Now that I think it through, the issue is that *only* the latest life cycle change is not-null. So, the original review request becomes null once there's a rejection or a timeout. But we need it to display the "user@foo.com requested to end this experiment"

app/experimenter/experiments/models/nimbus.py
```diff
@@ -179,7 +178,7 @@ def targeting_config(self):
             return self.TARGETING_CONFIGS[self.targeting_config_slug]
 
     def latest_change(self):
-        return self.changes.order_by("-changed_on").first()
+        return self.changes.latest_change()
```
jaredlockhart, line 181: Oh maybe this whole method is redundant and we should just replace it with `experiment.changes.latest_change()` wherever it's called?  NBD we can do that as cleanup later

app/experimenter/experiments/models/nimbus.py
```diff
@@ -458,53 +457,48 @@ def __str__(self):  # pragma: no cover
...
-            self.all()
-            .filter(
-                Q(old_status=NimbusExperiment.Status.DRAFT)
-                | Q(old_status=NimbusExperiment.Status.PREVIEW),
-            )
-            .filter(
-                old_publish_status=NimbusExperiment.PublishStatus.IDLE,
-                new_status=NimbusExperiment.Status.DRAFT,
-                new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
+        change = self.latest_change()
```
jaredlockhart, line 464: Oh technically this could return `None` so in theory you'd wanna check if it's truthy before accessing it fields.  In theory, we never hit this case cuz as soon as an experiment is created it has the creation change, so it's probably not a big deal, just wanted to point it out in case you see something wonky in testing.

jaredlockhart:
Right okay so then for rejection/timeout we want it to behave the way it is now, only return non null if it is the most recent change, and for review request, it should always return the most recent review request?

lmorchard:
Updated to return the latest review request in history, rather than just the latest change

lmorchard:
> Right okay so then for rejection/timeout we want it to behave the way it is now, only return non null if it is the most recent change, and for review request, it should always return the most recent review request?

Yup, that should be the change I just pushed

app/experimenter/experiments/models/nimbus.py
```diff
@@ -458,53 +458,51 @@ def __str__(self):  # pragma: no cover
...
+    def latest_change(self):
+        return self.all().order_by("-changed_on").first()
+
     def latest_review_request(self):
         return (
             self.all()
             .filter(
-                Q(old_status=NimbusExperiment.Status.DRAFT)
-                | Q(old_status=NimbusExperiment.Status.PREVIEW),
+                Q(old_status=F("new_status"))
```
jaredlockhart, line 468: Fancy

app/experimenter/experiments/models/nimbus.py
```diff
@@ -458,53 +458,51 @@ def __str__(self):  # pragma: no cover
...
+            in (
+                NimbusExperiment.PublishStatus.REVIEW,
+                NimbusExperiment.PublishStatus.WAITING,
             )
-            .order_by("-changed_on")
-        ).first()
+            and change.new_publish_status == NimbusExperiment.PublishStatus.IDLE
+        ):
+            return change
+        return None
```
jaredlockhart, line 494: Functions return `None` by default, so you can make this explicit or just omit.  I always omit, the only time I've seen anything complain about it is mypy which makes you always explicitly declare a return for all branches.

jaredlockhart (approved):
Yep all looks right and works right on my end! \o/ Man this flow is so slick.

PR #5074, opened 2021-04-20
fix #5028 feat(nimbus): support rejection during launch/pause/end

### LINKED ISSUE #5028: Handle pause/end rejection
Currently, if you reject a pause change, the task will crash because it only expects launch rejections.  We need to be able to handle rejections at

1) launch
2) live update
3) end

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1196)
┆Issue Number: EXP-1196

jaredlockhart:
Because

* We need to be able to handle remote settings rejection at any point during an experiments lifecycle

This commit

* Handles the case of rejection during launch/pause/end
* Pauses can not really be rejected so the rejection is ignored and the pause is resent
* We can't use the collection record diff to determine which record is being rejected so we look for an experiment in WAITING instead
* If no experiment is found in WAITING we roll back the collection and proceed as normal
* Reorganized the tests to cover all cases more clearly

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -55,76 +55,79 @@ def nimbus_check_kinto_push_queue_by_application(application):
 
     kinto_client = KintoClient(collection)
 
-    can_update_collection = True
+    should_rollback = False
     if kinto_client.has_pending_review():
         logger.info(f"{collection} has pending review")
-        can_update_collection = handle_pending_review(application, kinto_client)
+        should_abort = handle_pending_review(application, kinto_client)
+
+        if should_abort:
+            return
+
+        should_rollback = True
 
     if kinto_client.has_rejection():
         logger.info(f"{collection} has rejection")
-        handle_rejection(kinto_client)
+        handle_rejection(application, kinto_client)
+        should_rollback = True
 
-    if can_update_collection:
-        if queued_launch_experiment := NimbusExperiment.objects.launch_queue(
-            application
-        ).first():
-            nimbus_push_experiment_to_kinto.delay(queued_launch_experiment.id)
-        elif queued_end_experiment := NimbusExperiment.objects.end_queue(
-            application
-        ).first():
-            nimbus_end_experiment_in_kinto.delay(queued_end_experiment.id)
-        elif queued_pause_experiment := NimbusExperiment.objects.pause_queue(
-            application
-        ).first():
-            nimbus_pause_experiment_in_kinto.delay(queued_pause_experiment.id)
+    if should_rollback:
+        kinto_client.rollback_changes()
+
+    if queued_launch_experiment := NimbusExperiment.objects.launch_queue(
+        application
+    ).first():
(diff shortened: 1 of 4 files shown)


jodyheavener (approved):
LGTM with one suggestion

app/experimenter/kinto/tasks.py
```diff
@@ -162,60 +165,67 @@ def nimbus_push_experiment_to_kinto(experiment_id):
 
 
 @app.task
-@metrics.timer_decorator("end_experiment_in_kinto")
-def nimbus_end_experiment_in_kinto(experiment_id):
+@metrics.timer_decorator("pause_experiment_in_kinto")
+def nimbus_pause_experiment_in_kinto(experiment_id):
```
lmorchard, line 169: Swapping the order of the pause & edit tasks here makes the diff confusing to interpret. I think I see what's up, though

app/experimenter/kinto/tasks.py
```diff
@@ -55,76 +55,79 @@ def nimbus_check_kinto_push_queue_by_application(application):
...
 
 
-def handle_rejection(kinto_client):
-    rejected_slug = kinto_client.get_rejected_record()
+def handle_rejection(application, kinto_client):
     collection_data = kinto_client.get_rejected_collection_data()
-    experiment = NimbusExperiment.objects.get(slug=rejected_slug)
+    experiment = NimbusExperiment.objects.filter(
+        application=application, publish_status=NimbusExperiment.PublishStatus.WAITING
+    ).first()
```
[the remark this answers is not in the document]
jaredlockhart: Yeah not a bad idea, but we'd have to pass application in so then it's just the one field with a single constant.  If the logic there starts to grow in complexity we should definitely think about moving it into a method 👍

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -116,57 +54,108 @@ def test_check_with_empty_queue_pushes_nothing(self):
...
-        for publish_status in [
-            NimbusExperiment.PublishStatus.IDLE,
-            NimbusExperiment.PublishStatus.REVIEW,
-            NimbusExperiment.PublishStatus.WAITING,
-        ]:
-            NimbusExperimentFactory.create(
-                publish_status=publish_status,
-                application=NimbusExperiment.Application.DESKTOP,
-            )
+    def test_check_with_no_approved_publish_status_pushes_nothing(self):
```
lmorchard, line 62: Yeesh, github really doesn't know what to do with these changes in a diff does it? 🤷‍♂️

lmorchard (approved):
LGTM too, just grousing at the diff algo a bit here but I don't think I find fault with the code

app/experimenter/kinto/tasks.py
jaredlockhart: Bleargh sorry, i was trying to reorganize it so it always goes

launch
pause
end

it made the diff much uglier than it should've been.  Maybe next time I'll just do reordering in a separate commit so it's easier to see the logical changes.

PR #5076, opened 2021-04-20
fix #5023 chore(nimbus): expose all admin fields

### LINKED ISSUE #5023: Expose all fields in Nimbus admin
Rather than allow listing certain fields and then forgetting to add new fields to the list, we should just `exclude = ('id',)` for all Nimbus models in the admin.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1192)
┆Issue Number: EXP-1192

jaredlockhart:
Because

* We just want to be able to edit the entire nimbus objects from the admin

This commit

* Removes any explicit field lists from nimbus admin

### DIFF
--- a/app/experimenter/experiments/admin/nimbus.py
+++ b/app/experimenter/experiments/admin/nimbus.py
@@ -19,22 +19,11 @@ class NimbusBranchInlineAdmin(admin.StackedInline):
 class NimbusDocumentationLinkInlineAdmin(admin.TabularInline):
     model = NimbusDocumentationLink
     extra = 1
-    fields = ("title", "link")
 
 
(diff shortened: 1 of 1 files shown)


lmorchard:
FWIW, this makes the changelog rows for each experiment pretty huge now. We might want to consider reworking the layout there at some point

jaredlockhart:
Oh yeah I think there's an easy way to do that one sec

jaredlockhart:
They're a little chonkier but at least they fit on the screen

Screen Shot 2021-04-20 at 5 48 07 PM

lmorchard (approved):
LGTM FWIW

PR #5078, opened 2021-04-21
fix #5070 bug(nimbus): set publish status to idle when complete

### LINKED ISSUE #5070: Check experiments are complete must set publish_status to Idle
Just noticed that ending an experiment kept the publish status at WAITING
Screen Shot 2021-04-19 at 6 51 30 PM

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1213)
┆Issue Number: EXP-1213

jaredlockhart:
Because

* We recently introduced publish_status
* publish_status gets set to WAITING as something is in review, such as when ending
* We neglected to set it back to IDLE after the completion flow is done

This commit

* Sets publish_status to IDLE when the experiment completes

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -345,6 +345,7 @@ def nimbus_check_experiments_are_complete():
                 )
 
                 experiment.status = NimbusExperiment.Status.COMPLETE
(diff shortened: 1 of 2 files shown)


jodyheavener (approved):
😅

PR #5080, opened 2021-04-21
fix #4210 chore(nimbus): use separate sentry dsn for nimbus ui

### LINKED ISSUE #4210: Split sentry into separate projects for frontend and backend
I was just looking at sentry for stage:

https://sentry.prod.mozaws.net/operations/experimenter-stage/

And there's a lot of entries and it's hard to reason about which ones are frontend and which ones are backend other than just by looking at the type of error and the filename (.py or .js).  Maybe we should split into two separate projects, one for each of frontend or backend?  Alternatively maybe there's some way to use tags/labels to identify them, but I'm not sure how to do that.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-730)
┆Issue Number: EXP-730

jaredlockhart:
Because

* We'd like to track Nimbus UI as a separate Sentry project

This commit

* Adds a second Sentry DSN just for Nimbus UI


jaredlockhart:
Blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1706739

jodyheavener (approved):
YEAAHH (pending unblocking)

PR #5088, opened 2021-04-22
fix #5081 chore(nimbus): use more declarative filters

### LINKED ISSUE #5081: Use Q objects for all filtering (which reasonable)
I'd like to try using this pattern of defining filters as Q objects on models so we can refer to them as constants instead of littering filters all over the place.  It might make things more readable/testable/reusable.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1218)
┆Issue Number: EXP-1218

jaredlockhart:
Because

* I like this idea of declaring filters as reusable constants
* Enhances code reuse, makes things simpler to test, more readable

This commit

* Finds as many filters that can be moved to constants as reasonable

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -17,18 +17,21 @@ from experimenter.experiments.constants import NimbusConstants
 from experimenter.projects.models import Project
 
 
+class FilterMixin:
+    def has_filter(self, filter):
+        return type(self).objects.filter(id=self.id).filter(filter).exists()
+
+
 class NimbusExperimentManager(models.Manager):
     def launch_queue(self, applications):
         return self.filter(
-            status=NimbusExperiment.Status.DRAFT,
-            publish_status=NimbusExperiment.PublishStatus.APPROVED,
+            NimbusExperiment.Filters.IS_LAUNCH_QUEUED,
             application__in=applications,
         )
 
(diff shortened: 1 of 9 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -124,31 +122,53 @@ class Meta:
...
+            status=NimbusConstants.Status.LIVE,
+            publish_status=NimbusConstants.PublishStatus.APPROVED,
+            is_end_requested=True,
+        )
+        IS_ENDING = Q(
+            status=NimbusConstants.Status.LIVE,
+            publish_status=NimbusConstants.PublishStatus.WAITING,
+            is_end_requested=True,
+        )
+        SHOULD_TIMEOUT = Q(IS_LAUNCHING | IS_ENDING)
```
jodyheavener, line 148: Love it, so you can just infinitely nest `Q` calls?

jodyheavener (approved):
Sweet. Makes sense. Thanks for the chat about the DB hits. 👍

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Yeeeee they're composable, makes declaring stuff like this very nice and succinct, like algebraic types in TS or Rust or any ML variant.

PR #5101, opened 2021-04-26
fix #4260 bug(nimbus): feature value should always be an object

### LINKED ISSUE #4260: Disabled feature has null value, kinto publishing expects it to be JSON-serializable
We ran into this when publishing an experiment with an enabled=false branch I think https://sentry.prod.mozaws.net/operations/experimenter-stage/issues/10261719/

Is it possible it's expecting JSON for the feature value?

Changing the value on django admin to `{}`fixed the problem

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-750)
┆Issue Number: EXP-750

jaredlockhart:
Because

* Clients expect feature value to be a js object

This commit

* Defaults to {} instead of null for empty feature value


jaredlockhart:
Eh I'll just do this all at once in https://github.com/mozilla/experimenter/pull/5104

Issue #5103, opened 2021-04-26
Modify Branch UI to make Feature Config required


jaredlockhart:
Because we now want all experiments to include a feature config, we should change the branch UI to make it clearer that users must select a feature config.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1231)
┆Issue Number: EXP-1231

LZoog:
Oooh, this relates to #5171 / could even be considered a dupe.

@AnaMedinac tagging you in case you can help give some direction on what should be displayed here.

LZoog:
I'm going to throw this into the Required for Launch errors epic, I don't think we necessarily need input from Ana here but we should show this field as required.

PR #5104, opened 2021-04-26
fix #4694 feat(nimbus): make feature config required

### LINKED ISSUE #4694: Experiments should have at least one featureID to launch
Since all experiments in our system are now targeting features (we are deprecating the non-feature API), we should require at a featureId to be associated with all experiments.

This means we should change our branches UI to make sure that branches require choosing a feature.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-908)
┆Issue Number: EXP-908

jaredlockhart:
Because

* We now want all experiments to include a feature config

This commit

* Makes feature_config required to be not null to launch

Filed https://github.com/mozilla/experimenter/issues/5103 as a followup


lmorchard (approved):
LGTM!

PR #5107, opened 2021-04-26
chore(nimbus): always return a dict from kinto get_main_records

jaredlockhart:
Because

* I noticed we're always treating the kinto records as a dict

This commit

* Just returns a dict from get_main_records so we don't do that transformation everywhere


lmorchard (approved):
LGTM, nice little clean up

PR #5110, opened 2021-04-27
fix #4851 feat(nimbus): store published dto on experiment model

### LINKED ISSUE #4851: Store published DTO on NimbusExperiment
We currently expose the NimbusExperiment through the V6 API by serializing the current state of the experiment every time the V6 API is accessed.  However, the state in the database may not match the current published version of the experiment in Remote Settings as the experiment is modified while also being published (such as disabling enrollment).

Rather than serializing the database state when V6 is called, we should store the currently published DTO as a field on NimbusExperiment which is written to after any change is published to Remote Settings, and then use that to populate the V6 API.

This should allow us to delete the nimbus_check_experiments_are_paused task entirely, and simplify future workflows that involve making changes to published experiments.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1087)
┆Issue Number: EXP-1087

jaredlockhart:
Because

* We want to capture a historical record of the DTO as it was specified when the experiment ran

This commit

* Stores the published DTO on the experiment
* This will provide us with a historical record of the exact DTO that was used to run an experiment even if we change the serializer/schema in the future
* This will also allow us to more easily manage making changes to live experiments in the future when we expand the ability to modify live experiments


jodyheavener (approved):
👍

PR #5111, opened 2021-04-27
chore(nimbus): mozilla-nimbus-shared 1.5.0

jaredlockhart:
Because

* The latest mozilla nimbus shared package has the latest schema that requires feature config to be present

This commit

* Updates to latest mozilla nimbus shared


jodyheavener (approved):
noice

PR #5130, opened 2021-04-29
fixes #5022 feat(nimbus): set a changelog message for all experiment changes

### LINKED ISSUE #5022: Always set a changelog message
In the past we would use the pair of old_status/new_status and a lookup table to look up a bunch of templated changelog messages rather than using the `message` field right on the model, and only use `message` if a changelog contained some special/user entered message.  But the lookup table method doesn't scale very well, and having lots of changelogs with empty message and only old/new statuses can be confusing to review, especially if we set a message at changelog creation time in some cases.  

We should just always set a message, it'll make reviewing an experiments changelog history easier and make the UI easier to populate.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1193)
┆Issue Number: EXP-1193

lmorchard:
Because:

* always setting a changelog message will make reviewing an experiment's
  changelog history easier and make the UI easier to populate

This commit:

* makes changelog_message a required field for
  NimbusExperimentSerializer saves and updates

* ensures all Kinto tasks supply a changelog message on changes

* ensures each mutation call in nimbus-ui supplies a changelog message

### DIFF
--- a/app/experimenter/kinto/tasks.py
+++ b/app/experimenter/kinto/tasks.py
@@ -7,7 +7,7 @@ from experimenter.celery import app
 from experimenter.experiments.api.v6.serializers import NimbusExperimentSerializer
 from experimenter.experiments.changelog_utils import generate_nimbus_changelog
 from experimenter.experiments.email import nimbus_send_experiment_ending_email
-from experimenter.experiments.models import NimbusExperiment
+from experimenter.experiments.models import NimbusChangeLog, NimbusExperiment
 from experimenter.kinto.client import KintoClient
 
 logger = get_task_logger(__name__)
@@ -100,7 +100,7 @@ def handle_pending_review(applications, kinto_client):
             generate_nimbus_changelog(
                 experiment,
                 get_kinto_user(),
-                message="Timed Out",
+                message=NimbusChangeLog.Messages.TIMED_OUT_IN_KINTO,
             )
 
             logger.info(f"{experiment} timed out")
@@ -151,7 +151,11 @@ def nimbus_push_experiment_to_kinto(collection, experiment_id):
         experiment.publish_status = NimbusExperiment.PublishStatus.WAITING
         experiment.save()
 
-        generate_nimbus_changelog(experiment, get_kinto_user())
+        generate_nimbus_changelog(
+            experiment,
+            get_kinto_user(),
+            message=NimbusChangeLog.Messages.PUSHED_TO_KINTO,
+        )
 
         logger.info(f"{experiment.slug} pushed to Kinto")
         metrics.incr("push_experiment_to_kinto.completed")
@@ -217,7 +221,11 @@ def nimbus_end_experiment_in_kinto(collection, experiment_id):
         experiment.publish_status = NimbusExperiment.PublishStatus.WAITING
         experiment.save()
 
-        generate_nimbus_changelog(experiment, get_kinto_user())
+        generate_nimbus_changelog(
+            experiment,
+            get_kinto_user(),
+            message=NimbusChangeLog.Messages.DELETED_FROM_KINTO,
+        )
 
         logger.info(f"{experiment.slug} deleted from Kinto")
         metrics.incr("end_experiment_in_kinto.completed")
@@ -254,7 +262,9 @@ def nimbus_check_experiments_are_live():
                 experiment.save()
 
                 generate_nimbus_changelog(
-                    experiment, get_kinto_user(), message="Experiment is now live!"
+                    experiment,
+                    get_kinto_user(),
+                    message=NimbusChangeLog.Messages.LIVE,
                 )
 
                 logger.info(f"{experiment.slug} status is set to Live")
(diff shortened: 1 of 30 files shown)


lmorchard:
I *think* I caught all the places that need a changelog message? Could probably use another manual runthrough of all the flows

app/experimenter/kinto/tasks.py
```diff
@@ -296,7 +296,9 @@ def nimbus_check_experiments_are_paused():
                 experiment.is_paused = True
                 experiment.save()
 
-                generate_nimbus_changelog(experiment, get_kinto_user())
+                generate_nimbus_changelog(
+                    experiment, get_kinto_user(), "is_paused is set to True"
```
jaredlockhart, line 300: Eventually these will show up in a history page, so we might want to make them more end user consumable like 'Enrollment was paused' or something?

jaredlockhart (approved):
Yes awesome @lmorchard fantastic.  I love the constants, great idea about localizing them to which page was edited, previously we just had any edit be 'so and so edited', so that's a nice little touch.  Only thing that stands out to me is the is_paused one being a little technical, but that's just a nitpick.  Thanks for hopping on this! 👏 👏 👏

jaredlockhart (commented):
Oh oh @lmorchard just take out the `= None` here to force the message to be required and then tests will highlight if you missed anything:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/changelog_utils/nimbus.py#L41

app/experimenter/kinto/tasks.py
lmorchard: Yeah, that makes more sense. I had something like that earlier, but figured I should try to follow the other log messages in that neighborhood.

jaredlockhart: Yeah if others in that area are too technical we should make them more human readable too, but the ones you put for created/deleted/etc all looked fine 👍

jaredlockhart: And we can change these messages later so it's not the end of the world

lmorchard:
> then tests will highlight if you missed anything

Oh, yup! That highlighted about 25 other spots 😅

app/experimenter/kinto/tasks.py
```diff
@@ -151,7 +151,7 @@ def nimbus_push_experiment_to_kinto(collection, experiment_id):
         experiment.publish_status = NimbusExperiment.PublishStatus.WAITING
         experiment.save()
 
-        generate_nimbus_changelog(experiment, get_kinto_user())
+        generate_nimbus_changelog(experiment, get_kinto_user(), "Pushed to Kinto")
```
jaredlockhart, line 154: Also in theory you could move these to constants on either NimbusConstants or the NimbusChangeLog model 🤷‍♀️

lmorchard: Yup, [just did that in my latest update](https://github.com/mozilla/experimenter/pull/5130/files#diff-962a3f92cae878ce45f5b3dee466d34b97749882d62bf5ddcce0881482ead1f3R193-R199) after realizing I'd done it client-side but not server-side

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -190,6 +190,14 @@ class PublishStatus(models.TextChoices):
         "is_end_requested",
     )
 
+    class ChangelogMessages(object):
+        TIMED_OUT_IN_KINTO = "Timed Out"
+        PUSHED_TO_KINTO = "Pushed to Kinto"
+        DELETED_FROM_KINTO = "Deleted from Kinto"
+        LIVE = "Experiment is now live!"
+        PAUSED = "Enrollment was paused"
+        COMPLETED = "Experiment is complete"
```
jaredlockhart, line 199: 👏 👏 👏 👏 👏

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -190,6 +190,14 @@ class PublishStatus(models.TextChoices):
         "is_end_requested",
     )
 
+    class ChangelogMessages(object):
```
jaredlockhart, line 193: Just a tiny nit I think the model is called `ChangeLog` with capital `L`

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -574,6 +575,7 @@ def test_end_experiment_in_kinto_deletes_experiment(self):
             experiment.changes.filter(
                 old_publish_status=NimbusExperiment.PublishStatus.APPROVED,
                 new_publish_status=NimbusExperiment.PublishStatus.WAITING,
+                message="Deleted from Kinto",
```
jaredlockhart, line 578: Just noticed some tests referring to literals and some referring to the constants, dunno if you wanna make it all one or the other 🤷‍♀️

lmorchard: Yeah, I'll fix that - should be all consistent

PR #5135, opened 2021-04-29
fix #5134 bug(nimbus): intermittent integration test failure

### LINKED ISSUE #5134: Intermittent test: Nimbus integration tests depending on a feature config existing for every application
The integration tests run against the dev server, which is populated with 'dummy' data created by the 'create_dummy_experiments' management command.  For nimbus it creates experiments with random applications and feature configs, but this may leave an application without a feature config.  When the integration tests run, they may choose an application that doesn't have a dummy feature.  We should make the dummy data exhaustive for all applications.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1253)
┆Issue Number: EXP-1253

jaredlockhart:
Because

* The integration tests rely on there being a feature config for all applications to successfully create a test experiment

This commit

* Updates load_dummy_experiments to create an experiment for each application, which will create feature configs for each application as well

### DIFF
--- a/app/experimenter/base/management/commands/load_dummy_experiments.py
+++ b/app/experimenter/base/management/commands/load_dummy_experiments.py
@@ -22,6 +22,9 @@ class Command(BaseCommand):
             experiment = ExperimentFactory.create_with_status(status, type=random_type)
             logger.info("Created {}: {}".format(experiment, status))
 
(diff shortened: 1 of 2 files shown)


jodyheavener (approved):
Makes sense 👍

PR #5031, opened 2021-04-14
Fix #4927: fix(visualization): broken weekly table headers.

emtwo:
Because:
* Segmented experiments were resulting broken weekly tables
* Missing weekly analysis were not correctly reflected in table headers

This commit:
* Only looks at "all" segment for experiments (for now)
* Makes sure headers are accurate even when data is missing

PR #5136, opened 2021-04-30
Fix #5045: Only include All segment for percent calculations.

### LINKED ISSUE #5045: Population percentages are off on results page
I only see this for this experiment: https://experimenter.services.mozilla.com/nimbus/multi-stage-onboarding-ask-and-show-to-pin/results

the percentages should add up to 100% but they do not.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1205)
┆Issue Number: EXP-1205

emtwo:
Because:
* Not filtering out other segments was making incorrect population % calculations

This commit:
* Does this correct filtering

This is related to https://github.com/mozilla/experimenter/pull/5031 which was also related to segmentation filtering

### DIFF
--- a/app/experimenter/visualization/api/v3/models.py
+++ b/app/experimenter/visualization/api/v3/models.py
@@ -67,6 +67,9 @@ class JetstreamData(BaseModel):
         total_population = 0
         branches = {}
         for jetstream_data_point in self.__root__:
+            if jetstream_data_point.segment != Segment.ALL:
+                continue
+
             if jetstream_data_point.metric == Metric.USER_COUNT:
                 total_population += jetstream_data_point.point
                 branches[jetstream_data_point.branch] = jetstream_data_point.point
--- a/app/experimenter/visualization/tests/api/constants.py
+++ b/app/experimenter/visualization/tests/api/constants.py
@@ -113,7 +113,7 @@ class TestConstants:
         ABSOLUTE_METRIC_DATA_A = cls.get_absolute_metric_data(DATA_POINT_A)
         ABSOLUTE_METRIC_DATA_F = cls.get_absolute_metric_data(DATA_POINT_F)
         ABSOLUTE_METRIC_DATA_F_WITH_PERCENT = ABSOLUTE_METRIC_DATA_F.copy()
-        ABSOLUTE_METRIC_DATA_F_WITH_PERCENT.percent = 33.0
+        ABSOLUTE_METRIC_DATA_F_WITH_PERCENT.percent = 50.0
 
         DIFFERENCE_METRIC_DATA_WEEKLY_NEUTRAL = cls.get_difference_metric_data(
             DATA_POINT_B, SignificanceData(weekly={"1": Significance.NEUTRAL}, overall={})


emtwo:
The bug can be seen here (look at left column for % - should add up to 100%): https://experimenter.services.mozilla.com/nimbus/multi-stage-onboarding-ask-and-show-to-pin/results 

Screenshot before fix: 
Screen Shot 2021-04-30 at 10 19 18 AM

Screenshot after fix:
Screen Shot 2021-04-30 at 10 21 31 AM

app/experimenter/visualization/api/v3/models.py
```diff
@@ -67,6 +67,9 @@ def append_population_percentages(self):
         total_population = 0
         branches = {}
         for jetstream_data_point in self.__root__:
+            if jetstream_data_point.segment != Segment.ALL:
+                continue
```
jaredlockhart, line 71: I love the way this is written now.  Logic clearly laid out, properties clear, constants clear, it's so much nicer to review now @emtwo 🎉 🎉 🎉 🎉 🎉

app/experimenter/visualization/tests/api/constants.py
```diff
@@ -113,7 +113,7 @@ def get_test_data(cls):
         ABSOLUTE_METRIC_DATA_A = cls.get_absolute_metric_data(DATA_POINT_A)
         ABSOLUTE_METRIC_DATA_F = cls.get_absolute_metric_data(DATA_POINT_F)
         ABSOLUTE_METRIC_DATA_F_WITH_PERCENT = ABSOLUTE_METRIC_DATA_F.copy()
-        ABSOLUTE_METRIC_DATA_F_WITH_PERCENT.percent = 33.0
+        ABSOLUTE_METRIC_DATA_F_WITH_PERCENT.percent = 50.0
```
jaredlockhart, line 116: How does this trigger the logic above?

jaredlockhart (approved):
Very nice 👍 Just curious about the test.

app/experimenter/visualization/tests/api/constants.py
emtwo: This segment is added: https://github.com/mozilla/experimenter/blob/e7892de3b55c4832df9bc2b19c46bf91487f0259/app/experimenter/visualization/tests/api/constants.py#L54 and it extends `DATA_IDENTITY_ROW` which is a user count metric: https://github.com/mozilla/experimenter/blob/e7892de3b55c4832df9bc2b19c46bf91487f0259/app/experimenter/visualization/tests/api/constants.py#L43

So we included it in the `total_population` count here: https://github.com/mozilla/experimenter/blob/e7892de3b55c4832df9bc2b19c46bf91487f0259/app/experimenter/visualization/api/v3/models.py#L73

And then when we did this division it caused the population % to be a 3rd because we included that segment: https://github.com/mozilla/experimenter/blob/e7892de3b55c4832df9bc2b19c46bf91487f0259/app/experimenter/visualization/api/v3/models.py#L83

When we exclude the segment, it goes to 50%.

PR #5139, opened 2021-04-30
bug(nimbus): SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER must be a bool

jaredlockhart:
Because

* We added a setting SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER to bypass the two party signoff for local testing
* It must be a bool to be able to disable it

This commit

* Casts SKIP_REVIEW_ACCESS_CONTROL_FOR_DEV_USER to a bool


lmorchard (approved):
LGTM, ~but what bug did this cause?~ oh wait, no I think I saw the bug happen locally when I tried to disable it in .env, nevermind!

PR #5146, opened 2021-05-02
fix #5145 bug(reporting): allow old_status to be None and str rep change

### LINKED ISSUE #5145: [Reporting]: Model Changes to allow old_status to be None and string representation
We need old_status to be able to be None, for creation changelogs.
We also should change the string representation of Reportlog since it's currently `Reportlog Object (1)`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1258)
┆Issue Number: EXP-1258

tiftran:
Because:

* old_status needs to be ble to be None for creation changelogs
* String representation change would allow for easier debugging

This commit:

* allows old_status to be None
* Changes the str rep of ReportLog object to something more descriptive


jaredlockhart (approved):
👍 👍 👍 

Also just want to point out something interesting, if you look at the `dependencies` this only depends on the previous reporting migration, not any of the `experiments` migrations, so this won't conflict with another migration that targets `experiments` when you land it.  Just a handy thing to know when you're thinking about how to land PRs with migrations without creating conflicts on main.

PR #5158, opened 2021-05-03
Add risk mitigation fields to Nimbus experiment model, GQL mutation

### LINKED ISSUE #4919: (BE) Add risk mitigation fields to v5 mutation
EXP-1127 adds risk mitigation boolean fields to the Nimbus model. For this task, expose these fields in the update experiment mutation so the FE can update them.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1128)
┆Issue Number: EXP-1128

### LINKED ISSUE #4920: (BE) Add risk mitigation fields to Nimbus model
In the legacy model, risk mitigation questions look like this: https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/legacy.py#L218-L232

We can do this similarly with the 3 new risk mitigation questions. For this task, add these boolean fields to the Nimbus model. We'll expose them in a mutation and query in follow ups.

Figma: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=218778%3A23479

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1127)
┆Issue Number: EXP-1127

jodyheavener:
Closes #4920
Closes #4919

This PR adds:

- Risk mitigation fields to the Nimbus Experiment model
- The ability to change each of these fields via GQL mutation

The fields in question:

- `riskBrand` - "If the public, users or press, were to discover this experiment and description, do you think it would negatively impact their perception of the brand?"
- `riskPartnerRelated` - "Does this experiment impact or rely on a partner or outside company (e.g: Google, Amazon, Etc.)"
- `riskRevenue` - "Does your experiment have a risk to negatively impact revenue (e.x: search, pocket revenue)?"


jodyheavener:
Ah whoops I need to add these to the `NimbusReadyForReviewSerializer` still.

Also, @jaredlockhart should I migrate all existing experiments not in Draft/Preview to set these three fields as `false`?

jaredlockhart:
@jodyheavener I think for

Already launched experiments
- Will default to None, which is falsy
- When we show them on summary table, that's the same as False

Not yet launched experiments
- Default to None
- Will fail ready for review check
- They'll have to fill them in

Also I just realized we'll need to use radios for these, not checks, since we need to distinguish the None and False cases

Does that sound right to you?

jaredlockhart:
Also for the ready for review check, we'll have to add that AFTER we set them up in the form, otherwise no experiments will be launchable?

jaredlockhart:
Oh you're doing the models and gqls in one

I'd say we file a separate issue to add them to ready for review and do that last?

jodyheavener:
> Does that sound right to you?

Yup, makes sense. So no migrating existing experiments to False (leave them to the default None). And UI will be handled down the line 🤘

> Also for the ready for review check, we'll have to add that AFTER we set them up in the form, otherwise no experiments will be launchable?

Snap good thinking. I will file a separate ticket right now. So this actually is ready for review 😅

jaredlockhart:
@jodyheavener don't forget to mark this with `Data Models` label, becuase there's another `Data Models` PR up as well from @tiftran , and after it's approved make sure to rebase on main, repush, and then land 👍

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -633,3 +633,104 @@ def test_end_experiment_fails_with_nonlive_status(self):
...
+        result = content["data"]["updateExperiment"]
+        message = result["message"]["experiment"][0]
+        self.assertIn("risk_brand", message)
+        self.assertIn("risk_revenue", message)
+        self.assertIn("risk_partner_related", message)
+
+        experiment = NimbusExperiment.objects.get(id=experiment.id)
+        self.assertEqual(experiment.risk_brand, False)
+        self.assertEqual(experiment.risk_revenue, False)
+        self.assertEqual(experiment.risk_partner_related, False)
```
jaredlockhart, line 736: Are these really necessary as separate tests?  We have tests that test the update functionality of the mutation, you could fold these fields in there?

And the last test is repeating other tests to make sure you can't edit at non draft status?  Maybe a duplicate test?

jaredlockhart (approved):
You might be able to simplify/remove those tests altogether, but otherwise looks good 👍

app/experimenter/experiments/tests/api/v5/test_mutations.py
jodyheavener: Yeah, thanks for caling this out. Wasn't sure how thorough I should be with the non-draft updates and testing everything individually. Will adjust.

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -136,6 +142,9 @@ def test_update_experiment_overview(self):
         self.assertEqual(experiment.hypothesis, "new hypothesis")
         self.assertEqual(experiment.public_description, "new public description")
         self.assertEqual(experiment.risk_mitigation_link, "https://example.com/risk")
+        self.assertEqual(experiment.risk_brand, True)
+        self.assertEqual(experiment.risk_revenue, True)
+        self.assertEqual(experiment.risk_partner_related, True)
```
jaredlockhart, line 147: Yeah exactly perfect 👍

PR #5162, opened 2021-05-03
fixes #4921 feat(project): add signoff recommendations to query based on risk fields

### LINKED ISSUE #4921: (BE) Add mapping logic, expose fields to v5 query
We need to map the risk questions added in the Nimbus model in EXP-1127 to what triggers a legal or VP sign off. QA sign off will always be recommended. See https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/legacy.py#L808-L826 for how this was done in the legacy model. The question copy (referred to as 'label' in legacy code) will live in the FE. 

We then need to get the list of recommended reviews, see https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/legacy.py#L831-L837

What triggers a recommended sign off (like `{qa_signoff: true, vp_signoff: false, legal_signoff: true}`) should be exposed in the experiment query.

Figma, which has text describing which questions trigger what sign-off: https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=218778%3A23479

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1129)
┆Issue Number: EXP-1129

jodyheavener:
Closes #4921

This PR adds a new field, `signoffRecommendations`, to the experiment query, which provides the sub-fields `qaSignoff`, `vpSignoff`, and `legalSignoff` (all booleans)

The issue description mentions collecting all the required signoffs to compare them against completed signoffs, but since we don't have a mechanism to verify completed signoffs and they are all recommendations I simplified this a bit. Let me know if I'm missing something.

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -325,6 +325,17 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
             )
             return datetime.date.today() >= resultsReadyDate
 
+    @property
+    def signoff_recommendations(self):
+        return {
+            # QA signoff is always recommended
+            "qa_signoff": True,
+            "vp_signoff": any(
+                (self.risk_brand, self.risk_revenue, self.risk_partner_related)
+            ),
+            "legal_signoff": any((self.risk_revenue, self.risk_partner_related)),
+        }
+
 
 class NimbusBranch(models.Model):
     experiment = models.ForeignKey(
--- a/app/experimenter/experiments/tests/api/v5/test_queries.py
+++ b/app/experimenter/experiments/tests/api/v5/test_queries.py
@@ -701,3 +701,34 @@ class TestNimbusQuery(GraphQLTestCase):
         experiment_data = content["data"]["experimentBySlug"]
         self.assertEqual(experiment_data["isEnrollmentPaused"], True)
         self.assertEqual(experiment_data["enrollmentEndDate"], "2021-01-08")
+
+    def test_signoff_recommendations(self):
+        user_email = "user@example.com"
+        experiment = NimbusExperimentFactory.create_with_status(
+            NimbusExperiment.Status.DRAFT,
+            risk_brand=True,
+            risk_revenue=True,
+            risk_partner_related=True,
+        )
+
+        response = self.query(
+            """
+            query experimentBySlug($slug: String!) {
+                experimentBySlug(slug: $slug) {
+                    signoffRecommendations {
+                        qaSignoff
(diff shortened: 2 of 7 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -325,6 +325,17 @@ def results_ready(self):
...
+    @property
+    def signoff_recommendations(self):
+        return {
+            # QA signoff is always recommended
+            "qa_signoff": True,
+            "vp_signoff": any(
+                (self.risk_brand, self.risk_revenue, self.risk_partner_related)
+            ),
+            "legal_signoff": any((self.risk_revenue, self.risk_partner_related)),
+        }
```
jaredlockhart, line 337: Beautiful, perfect 👏 👏 👏 👏 👏 👏 👏

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -701,3 +701,118 @@ def test_paused_experiment_returns_date(self):
...
+            """,
+            variables={"slug": experiment.slug},
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200, response.content)
+        content = json.loads(response.content)
+        experiment_data = content["data"]["experimentBySlug"]
+        self.assertEqual(experiment_data["signoffRecommendations"]["qaSignoff"], True)
+        self.assertEqual(experiment_data["signoffRecommendations"]["vpSignoff"], True)
+        self.assertEqual(experiment_data["signoffRecommendations"]["legalSignoff"], True)
```
jaredlockhart, line 818: 1) I would do this combinatorial testing agaist the model method and then just a single test against the query to make sure it's hooked up

2) Might be able to compress this into a single parameterized test, something like

```py
@parameterized.expand([
    [True, False, True, False, True, False],
    .....
])
def test_testy_mctestface(self, risk_this, risk_that, risk_another, signoff_this, signoff_that, signoff_another):
    experiment = ExperimentFactory(risk_this=risk_this, etc....)
    self.assertEqual(experiment.signoff_recommendations, {
        'signoff_this': signoff_this',
        etc....
    })
```

jaredlockhart (approved):
Wonderful once again, some ideas for the tests 👍

app/experimenter/experiments/tests/api/v5/test_queries.py
jodyheavener: Hot damn I forgot about the parameterized thing. Thanks!

PR #5168, opened 2021-05-04
fix #5167 feat(nimbus): display recipe json on summary page

### LINKED ISSUE #5167: Display recipe json in summary/review page
We have a link for previewing the recipe json that links out to the V6 API, but with the recent lifecycle/review changes, experiments won't show there until they've live, so that link won't work anymore.  I think the simplest thing to do is just fetch the recipe json through the V5 API and display it right on the summary table.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1270)
┆Issue Number: EXP-1270

jaredlockhart:
Because

* We need a way to display the recipe json before an experiment launches for review
* The link to the V6 API will no longer work in the new lifecycle flow

This commit

* Exposes the recipe json via the V5 API
* Displays it on the summary table

Screenshot 2021-05-04 at 16-14-53 Public-key content-based leverage – Review Launch Experimenter

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -1,9 +1,12 @@
+import json
+
 import graphene
 from django.contrib.auth import get_user_model
 from graphene_django import DjangoListField
 from graphene_django.types import DjangoObjectType
 
 from experimenter.experiments.api.v5.serializers import NimbusReadyForReviewSerializer
+from experimenter.experiments.api.v6.serializers import NimbusExperimentSerializer
 from experimenter.experiments.constants.nimbus import NimbusConstants
 from experimenter.experiments.models.nimbus import (
     NimbusBranch,
@@ -162,6 +165,7 @@ class NimbusExperimentType(DjangoObjectType):
     rejection = graphene.Field(NimbusChangeLogType)
     timeout = graphene.Field(NimbusChangeLogType)
     signoff_recommendations = graphene.Field(NimbusSignoffRecommendationsType)
+    recipe_json = graphene.String()
 
     class Meta:
         model = NimbusExperiment
@@ -205,3 +209,6 @@ class NimbusExperimentType(DjangoObjectType):
 
     def resolve_timeout(self, info):
         return self.changes.latest_timeout()
+
+    def resolve_recipe_json(self, info):
+        return json.dumps(NimbusExperimentSerializer(self).data, indent=2, sort_keys=True)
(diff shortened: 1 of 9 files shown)


app/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
```diff
@@ -76,6 +82,17 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
...
           </tr>
         )}
+        {experiment.recipeJson && (
+          <tr>
+            <th>Recipe JSON</th>
+            <td data-testid="experiment-recipe-json">
+              <JSONPretty
+                id="json-pretty"
+                data={experiment.recipeJson}
+              ></JSONPretty>
```
jodyheavener, line 92: Is `JSONPretty` doing anything for us beyond prettifying and optional syntax highlighting? Since we're not using highlighting I think we can do without the extra dependency and just do something like:

```jsx
<code>
  <pre>
    {JSON.stringify(JSON.parse(experiment.recipeJson), null, 2)}
  </pre>
</code>
```

jodyheavener (commented):
This looks good, I think we could ditch the extra package though. What do you think?

app/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
jaredlockhart: Good call 👍

jodyheavener: If the server can return it prettified we could even forgo the whole `JSON.stringify(JSON.parse(` business

jaredlockhart: I wish bootstrap had a class to highlight json.  It used to.  Oh well.

jaredlockhart:
Screenshot 2021-05-05 at 11-41-37 Cross-platform 5thgeneration groupware – Review Launch Experimenter

app/experimenter/experiments/api/v5/types.py
```diff
@@ -205,3 +209,6 @@ def resolve_rejection(self, info):
 
     def resolve_timeout(self, info):
         return self.changes.latest_timeout()
+
+    def resolve_recipe_json(self, info):
+        return json.dumps(NimbusExperimentSerializer(self).data, indent=2, sort_keys=True)
```
jodyheavener, line 214: ayyyy

jodyheavener (approved):
🤠

PR #5170, opened 2021-05-05
fixes #4924 feat(nimbus): add UI for new risk mitigation fields

jodyheavener:
Closes #4924

This PR:

- Adds the new UI for the 3 risk mitigation questions to the overview page
- Adds the 3 fields to the "missing fields" check (this won't do anything until #5169 is merged)

PR #5174, opened 2021-05-05
fix #4925 feat(nimbus): display recommended signoffs on review page

### LINKED ISSUE #4925: (FE) Display "Recommended actions before launch" table
After EXP-1129 is complete, the FE should have easy access to data like `{qa_signoff: true, vp_signoff: false, legal_signoff: true}` to display green `Recommended:` text where the sign off recommendation is `true`.

Read this data and display the able according to Figma on the Request Review page.

See Figma:  https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=218778%3A23479

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1132)
┆Issue Number: EXP-1132

jaredlockhart:
Because

* After risk mitigation questions are filled out we need to show which sign offs are recommended

This commit

* Displays the recommended sign offs table on the review page

Screenshot 2021-05-05 at 12-39-28 Expanded didactic access – Review Launch Experimenter

Blocked on getting the three `Learn More` links from @AnaMedinac

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
@@ -4,10 +4,11 @@
 
 import { RouteComponentProps } from "@reach/router";
 import React, { useRef, useState } from "react";
+import { Table } from "react-bootstrap";
 import Alert from "react-bootstrap/Alert";
 import { useChangeOperationMutation } from "../../hooks";
 import { useConfig } from "../../hooks/useConfig";
-import { CHANGELOG_MESSAGES } from "../../lib/constants";
+import { CHANGELOG_MESSAGES, EXTERNAL_URLS } from "../../lib/constants";
 import { getStatus } from "../../lib/experiment";
 import { getExperiment_experimentBySlug } from "../../types/getExperiment";
 import {
@@ -16,6 +17,7 @@ import {
 } from "../../types/globalTypes";
 import AppLayoutWithExperiment from "../AppLayoutWithExperiment";
 import ChangeApprovalOperations from "../ChangeApprovalOperations";
+import LinkExternal from "../LinkExternal";
 import Summary from "../Summary";
 import FormLaunchDraftToPreview from "./FormLaunchDraftToPreview";
 import FormLaunchDraftToReview from "./FormLaunchDraftToReview";
@@ -160,6 +162,52 @@ const PageRequestReview = ({
                 )}
               </ChangeApprovalOperations>
             )}
+
+            <h3 className="h5 mb-3">Recommended actions before launch</h3>
+            <Table bordered data-testid="table-signoff" className="mb-4">
+              <tr data-testid="table-signoff-qa">
+                <td>
+                  <strong>QA Sign - Off</strong>
+                </td>
+                <td>
+                  {experiment.signoffRecommendations?.qaSignoff && (
+                    <span className="text-success">Recommended: </span>
+                  )}
+                  Describe what they should do.{" "}
+                  <LinkExternal href={EXTERNAL_URLS.SIGNOFF_QA}>
+                    Learn More
+                  </LinkExternal>
+                </td>
+              </tr>
+              <tr data-testid="table-signoff-vp">
+                <td>
+                  <strong>VP Sign - Off</strong>
+                </td>
+                <td>
+                  {experiment.signoffRecommendations?.vpSignoff && (
+                    <span className="text-success">Recommended: </span>
+                  )}
+                  Describe what they should do.{" "}
+                  <LinkExternal href={EXTERNAL_URLS.SIGNOFF_VP}>
+                    Learn More
+                  </LinkExternal>
+                </td>
+              </tr>
+              <tr data-testid="table-signoff-legal">
+                <td>
+                  <strong>Legal Sign - Off</strong>
+                </td>
+                <td>
+                  {experiment.signoffRecommendations?.legalSignoff && (
+                    <span className="text-success">Recommended: </span>
+                  )}
+                  Describe what they should do.{" "}
+                  <LinkExternal href={EXTERNAL_URLS.SIGNOFF_LEGAL}>
+                    Learn More
+                  </LinkExternal>
+                </td>
+              </tr>
+            </Table>
 
             <Summary {...{ experiment }} />
           </>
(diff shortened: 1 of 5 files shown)


app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -161,6 +162,46 @@ const PageRequestReview = ({
               </ChangeApprovalOperations>
             )}
 
+            <h5>Recommended actions before launch</h5>
+            <Table bordered data-testid="table-signoff" className="mb-4">
+              <tr data-testid="table-signoff-qa">
+                <td>
+                  <strong>QA Sign - Off</strong>
```
jodyheavener, line 169: Drive-by feedback: can we pretty please make "Sign - Off" one word, "signoff", or make it a hyphenated word, "sign-off"? The way it currently is makes it seem like an en-dash

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -161,6 +162,46 @@ const PageRequestReview = ({
               </ChangeApprovalOperations>
             )}
 
+            <h5>Recommended actions before launch</h5>
```
LZoog, line 165: This should actually be an `h3`. We probably need to tweak some of these since I see some are `h2`s when "Review and Launch" is an `h2`, could be a low priority maintenance item.

Also! We use `mb-3` on all other headers on this page.

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -161,6 +162,46 @@ const PageRequestReview = ({
...
+              <tr data-testid="table-signoff-qa">
+                <td>
+                  <strong>QA Sign - Off</strong>
+                </td>
+                <td>
+                  {experiment.signoffRecommendations?.qaSignoff && (
+                    <span className="text-success">Recommended: </span>
+                  )}
+                  Describe what they should do.
+                  <a href=""> Learn more</a>
```
LZoog, line 176: I know you'll use `LinkExternal` here when you've got these😎

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -161,6 +162,46 @@ const PageRequestReview = ({
               </ChangeApprovalOperations>
             )}
 
+            <h5>Recommended actions before launch</h5>
+            <Table bordered data-testid="table-signoff" className="mb-4">
```
LZoog, line 166: This is not entirely necessary but it might be nice to have this as a separate component, mostly because the Request Review page has a ton of stories already and I'd like to see the various states reflected there!

image

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
```diff
@@ -161,6 +162,46 @@ const PageRequestReview = ({
...
+              <tr data-testid="table-signoff-qa">
+                <td>
+                  <strong>QA Sign - Off</strong>
+                </td>
+                <td>
+                  {experiment.signoffRecommendations?.qaSignoff && (
```
LZoog, line 172: I think `qaSignoff` will always be `true` right? I guess it's good to leave the check in here anyway alongside the test, though if there was any possibility of this being `false` I can't help but think it'd be a little confusing without at least one `Recommended:` text guaranteed. As a user I'd just assume all of these were recommended🤷‍♀️

image

LZoog (commented):
Looking gewd so far! Mostly nits, though it would be really nice to see Storybook states for a few different states (all, none (or only `qaSignoff` if that'll always be `true`), and maybe two recommendations just to cover some basis).

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
jaredlockhart: That's not too big? In the mockups it looked like it should be smaller than the other titles?

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
jaredlockhart: Yeah I wasn't sure if it was worth pulling into its own component

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
jaredlockhart: Yeah but the assumption that it's always true is already hard coded in the backend, so if we hard code it here then it's done in two places.  And yeah I think what @AnaMedinac  was going for is that like, they're all kinda recommended, but if a risk mitigation question is true then they're a bit MORE recommended?  I think?

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
LZoog: If the font needs to be smaller it can be done with classes, semantically speaking an `h5` should only occur if it's under an `h4` on the page!

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
LZoog: Yeah I agree, just leave 'em in one place so if it ever changes we don't have to update the FE.

Also that sounds right as well, it's probably also fine since it says "recommended" above and not "required"

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
LZoog: Actually maybe you can sneak it in like, "Preview status, one recommended sign off," since you already have "Review with recommended sign offs" which I somehow missed since the story list there is so long😂

jodyheavener:
@jaredlockhart once you get the updated Learn More links you'll want to update the [constants](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/lib/constants.ts#L29:L34) I introduced in #5170. Tbh I forgot about the outdated links before merging, so I'm glad your PR updates them 😆

jaredlockhart:
Screenshot 2021-05-06 at 11-39-17 Reduced logistical system engine – Review Launch Experimenter

LZoog (approved):
LGTM besides those missing links, I'll go ahead and r+ so you can merge as soon as Ana has those to prevent more PR blockage 👍

app/experimenter/nimbus-ui/src/components/PageRequestReview/index.tsx
jodyheavener: @jaredlockhart 😢

PR #5177, opened 2021-05-06
fixes #4902 chore(nimbus): Refactor NimbusExperimentFactory to incorporate publish_status

### LINKED ISSUE #4902: Refactor NimbusExperimentFactory to incorporate publish_status
Once we finish migrating everything over to using publish_status we'll want to add some logic to iterate the experiment through its status/publish_status stages to simplify the tests, and then we can remove explicitly creating those changelogs in the new tests.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1110)
┆Issue Number: EXP-1110

lmorchard:
Because:

* We want to add some logic to iterate the experiment through its status /
  publish_status stages to simplify the tests

This commit:

* Declare publishing lifecycles and states as constants

* Generate dummy experiments using with new create_with_lifecycle method
  to replace create_with_status

* Rework tests to use new lifecycle-based factory

* Tweak naming of PAUSE_QUEUE to PAUSE_CANDIDATE for clarity

### DIFF
--- a/app/experimenter/base/management/commands/load_dummy_experiments.py
+++ b/app/experimenter/base/management/commands/load_dummy_experiments.py
@@ -22,9 +22,6 @@ class Command(BaseCommand):
             experiment = ExperimentFactory.create_with_status(status, type=random_type)
             logger.info("Created {}: {}".format(experiment, status))
 
-        for application in NimbusExperiment.Application:
-            for status in NimbusExperiment.Status:
-                experiment = NimbusExperimentFactory.create_with_status(
-                    status, application=application
-                )
-                logger.info("Created {}: {}".format(experiment, status))
+        for lifecycle in NimbusExperiment.Lifecycles:
+            experiment = NimbusExperimentFactory.create_with_lifecycle(lifecycle)
+            logger.info("Created {}: {}".format(experiment, lifecycle))
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from enum import Enum
 from typing import Dict
 
 from django.conf import settings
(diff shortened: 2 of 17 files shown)


lmorchard:
@jaredlockhart Does this look at all like what you were thinking? Looks like there's 150 or so uses of create_with_status, so I wanted to get a sanity check before diving into replacing all those.

I ended up just going ahead and defining lifecycle flows, because just looping toward a target failed to handle a few flows where a state is visited twice. (e.g. rejections & timeouts)

app/experimenter/base/management/commands/load_dummy_experiments.py
```diff
@@ -23,8 +23,8 @@ def handle(self, *args, **options):
             logger.info("Created {}: {}".format(experiment, status))
 
         for application in NimbusExperiment.Application:
-            for status in NimbusExperiment.Status:
-                experiment = NimbusExperimentFactory.create_with_status(
-                    status, application=application
+            for lifecycle in NimbusExperiment.Lifecycles:
+                experiment = NimbusExperimentFactory.create_with_lifecycle(
+                    lifecycle, application=application
                 )
```
lmorchard, line 29: Also, trying to decide if this is a good idea. It generates a *lot* of dummy experiments, but it does result in examples at every stage of launch & end flows

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -156,25 +157,115 @@ class ApplicationConfig:
...
+    REVIEW = "Review"
+    APPROVED = "Approved"
+    WAITING = "Waiting"
+
+
+class LifecycleStates(Enum):
+    DRAFT_IDLE = dict(
+        status=Status.DRAFT,
+        publish_status=PublishStatus.IDLE,
+    )
```
jaredlockhart, line 178: Why use this syntax instead of dict literal syntax? I don't think we use this anywhere else?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -156,25 +157,115 @@ class ApplicationConfig:
...
+        is_end_requested=True,
+    )
+    COMPLETE_IDLE_ENDED = dict(
+        status=Status.COMPLETE,
+        publish_status=PublishStatus.IDLE,
+        is_end_requested=True,
+    )
+
+
+class Lifecycles(Enum):
```
jaredlockhart, line 230: First of all, INTERESTING! I hadn't thought of structuring it like this, with each entry being just the end state with the ordered tuple of preceding states.  I was thinking of it as just the list of states and then iterating over the enum leads you through the states, but that doesn't let you have multiple different branching end states like the timeouts and rejections and etc you have here, so that's a really clever design!

Second quick question, why does this have to be defined at the module level instead of inside the constants class? You can't refer to names the same way?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -130,28 +130,11 @@ class Meta:
...
-        IS_ENDING = Q(
-            status=NimbusConstants.Status.LIVE,
-            publish_status=NimbusConstants.PublishStatus.WAITING,
-            is_end_requested=True,
-        )
+        IS_LAUNCH_QUEUED = Q(**NimbusConstants.LifecycleStates.DRAFT_APPROVED.value)
+        IS_LAUNCHING = Q(**NimbusConstants.LifecycleStates.DRAFT_WAITING.value)
+        IS_PAUSE_QUEUED = Q(**NimbusConstants.LifecycleStates.PAUSE_QUEUED.value)
+        IS_END_QUEUED = Q(**NimbusConstants.LifecycleStates.LIVE_APPROVED_ENDING.value)
+        IS_ENDING = Q(**NimbusConstants.LifecycleStates.LIVE_WAITING_ENDING.value)
```
jaredlockhart, line 137: This is even more interesting!!!  Makes total sense, I like that you can define filters in relation to lifecycles.  So interesting.

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -156,25 +157,115 @@ class ApplicationConfig:
...
+        publish_status=PublishStatus.WAITING,
+    )
+    LIVE_IDLE = dict(
+        status=Status.LIVE,
+        publish_status=PublishStatus.IDLE,
+    )
+    PAUSE_QUEUED = dict(
+        status=Status.LIVE,
+        is_paused=False,
+    )
```
jaredlockhart, line 202: This kind of stands out from the others that are named as the pair of status/publishstatus, maybe it's unavoidable, but it stands out a bit to me 🤷‍♀️

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -170,6 +170,30 @@ def create_with_status(cls, target_status, **kwargs):
...
+            if experiment.has_filter(experiment.Filters.SHOULD_ALLOCATE_BUCKETS):
+                experiment.allocate_bucket_range()
+
+            generate_nimbus_changelog(
+                experiment,
+                experiment.owner,
+                f"set lifecycle {lifecycle} state {state}",
+            )
+
+        return NimbusExperiment.objects.get(id=experiment.id)
```
jaredlockhart, line 195: This is 100% exactly what I was imaginging so yes you exactly nailed it.  The only thing I'd question is whether to have the CREATED as the default.  On the one hand, yeah totally, seems like a safe default.  On the other hand, at the call sites in the tests, reading just `create_with_lifecycle()` doesn't make it exactly clear to me what state it's gonna wind up in, reading it I'd like to always see the target state.  But CREATED is a sensible default so, could probably go either way on it.

jaredlockhart (commented):
Yes @lmorchard not only did you capture exactly what I was hoping for but added a bunch of really clear, readable, but also smart ways to capture more than I was even thinking, so yep 10/10 really great job 👏 👏 👏 👏 👏 👏

app/experimenter/experiments/constants/nimbus.py
lmorchard: > Second quick question, why does this have to be defined at the module level instead of inside the constants class? You can't refer to names the same way?

Yeah, within the scope of the constants class, the nested classes can't refer to each other.

So, without hoisting up to the module scope, LifecycleStates couldn't refer to Status or PublishStatus and Lifecycles couldn't refer to LifecycleStates

app/experimenter/experiments/constants/nimbus.py
lmorchard: Yeah, I thought maybe this one should also have `publishStatus`, but I lifted it as-is from the `IS_PAUSE_QUEUED` filter and didn't want to mess with it further

app/experimenter/experiments/tests/factories/nimbus.py
lmorchard: Yeah, I might get rid of the default just to be explicit everywhere

lmorchard:
Alright, I think I finished replacing all the instances of `create_with_status` with `create_with_lifecycle`. Had to rework a handful of tests, but hopefully not much more than just the experiment data setup

jaredlockhart (approved):
Yeah looks really great @lmorchard thank you for digging into this.  Excited to give it a try in some new tests.  Couple little suggestions but nothing major, j'approve! 🎉 🎉 🎉

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -156,25 +157,122 @@ class ApplicationConfig:
...
+    }
+    DRAFT_WAITING = {
+        "status": Status.DRAFT,
+        "publish_status": PublishStatus.WAITING,
+    }
+    LIVE_IDLE = {
+        "status": Status.LIVE,
+        "publish_status": PublishStatus.IDLE,
+    }
+    PAUSE_CANDIDATE = {
```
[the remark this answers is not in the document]
lmorchard: Well, the reason I changed to "candidate" is that this differs from the other named things - paused state is not determined solely from the DB query, there's also [a `should_pause` method check](https://github.com/mozilla/experimenter/blob/19802da5767b014f9cb063a2c39b8763b7156381/app/experimenter/experiments/models/nimbus.py#L37-L39) that figures into determining what's in the pause queue. That tripped me up a bit while trying to work out exactly how this set of experiment properties somehow defined pausing.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -161,6 +144,10 @@ class Filters:
     def __str__(self):
         return self.name
 
+    def set_lifecycle_state(self, lifecycle_state):
+        for name, value in lifecycle_state.value.items():
+            setattr(self, name, value)
```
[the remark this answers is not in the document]
lmorchard: I could rename to apply. I didn't want to save here in case other changes needed to happen along with setting lifecycle state. But could probably save & generate a changelog record here to DRY up code elsewhere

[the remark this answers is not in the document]
lmorchard: I just renamed to "apply", we can tweak to more fully saving if it's handy

lmorchard:
Think I'll merge after green CI

PR #5185, opened 2021-05-07
fixes #4291 feat(nimbus): display server review messages instead of client ones

### LINKED ISSUE #4291: Display server-generated messages when determining review-readiness

jodyheavener:
Closes #4291

This PR's main goal is to, in those review-readiness tooltips, stop displaying static client messages and start displaying messages generated by the server. A couple additional preparation changes needed to be made first.

- I split the review-readiness checks out of the `useExperiment` hook and into its own `useReviewCheck` hook that could be called directly in any component that needs it.
  - In doing this I also moved `refetch`, the function to re-query the experiment, out of the review checks and into the top level, since it's use has grown beyond re-checking the review status
- Then the main part was fairly simple; take the messages sent up by the server and display them in that tooltip presentation.
  - I tweaked how we display these tooltips, as well, by rendering a `FieldReview` component that performs the checks internally and just renders null if there's no reason to display the tooltip.
  - Was able to simplify tests a little as well


jaredlockhart:
Neat!

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -175,15 +176,16 @@ type LayoutProps = {
...
   status,
   analysis,
   analysisRequired,
   analysisLoadingInSidebar,
   analysisError,
   experiment,
-}: LayoutProps) =>
-  status?.launched ? (
+}: LayoutProps) => {
+  const review = useReviewCheck(experiment);
```
LZoog, line 186: Nit: since we only need `review` if `status.launched` is falsey, we _could_ revert L185-186 and update `<AppLayoutWithSidebar {...{ status, review }}>` to `<AppLayoutWithSidebar review={useReviewCheck(experiment)} {...{ status }}>`

app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
```diff
@@ -359,20 +359,34 @@ describe("FormOverview", () => {
...
-      expect(screen.queryByTestId("missing-description")).toBeInTheDocument();
+    const { experiment } = mockExperimentQuery("boo", {
+      readyForReview: {
+        ready: false,
+        message: invalidFields,
+      },
     });
+    render(<Subject {...{ experiment }} />);
+
+    for (const [key, value] of Object.entries(invalidFields)) {
```
LZoog, line 384: Since `invalidFields` is defined right above it's easy to tell what these should be so less applicable here, but normally I'd prefer `[key, value]` to be something like `[field, error]` when using `Object.entries` but maybe that's just me 😄

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -20,6 +20,12 @@ import {
 } from "../../../types/globalTypes";
 import { MOCK_EXPERIMENT, Subject } from "./mocks";
 
+// jest.mock("../../../hooks/useReviewCheck", () => ({
+//   useReviewCheck: () => ({
+//     isMissingField: () => true,
+//   }),
+// }));
```
LZoog, line 27: Missed commented out stuff?

app/experimenter/nimbus-ui/src/hooks/useReviewCheck.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  // EXP-981 should remove this component entirely once fieldReviewMessages is used
+  // to feed these review-readiness messages directly into useCommonForm
+  const FieldReview: React.FC<FieldReviewProps> = ({ field, children }) => {
+    const messages = fieldReviewMessages(field);
+    if (!messages.length || !window.location.search.includes("show-errors")) {
+      return null;
+    }
+
+    return <>{children({ field, message: messages.join(", ") })}</>;
```
LZoog, line 59: It looks like we only use this component like...

```
<FieldReview field="channel">
  {({ field, message }) => (
    <InlineErrorIcon {...{ field, message }} />
  )}
</FieldReview>
```

or like this...

```
<FieldReview field="risk_brand">
  {({ field, message }) => (
    <span className="align-text-bottom ml-1">
      <InlineErrorIcon {...{ field, message }} />
    </span>
  )}
</FieldReview>
```

I know your comment says EXP-981 should remove these in favor of getting these from `useCommonForm`, but do we really need `field, message` returned from `FieldReview`? Can we take a prop instead of accepting `children`?

```
<FieldReview field="channel" />
```
and (or whatever prop makes sense to pass in):
```
<FieldReview field="field_risk" alignBottom />
```

LZoog (commented):
Looking good so far, really just one question that might help simplify this a hair😀

app/experimenter/nimbus-ui/src/hooks/useReviewCheck.tsx
jodyheavener: @LZoog I totally agree except for the fact that in some cases we need to wrap the error icon in a span for styling. Otherwise I'd say let's make it all internal to the component. We _could_ place the spans on the outside, but it would result in empty spans whenever the icon wasn't rendered. Thoughts?

jodyheavener: Further to Slack chat - this looks good with the classes applied on all instances.

Issue #5186, opened 2021-05-07
Experiment detail page broken on stage


jaredlockhart:
https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/pdehaan-20210506-ddd-application-error/edit/overview

"'NoneType' object has no attribute 'slug'"

in v6 serializer

in sentry:

https://sentry.prod.mozaws.net/operations/experimenter-stage/issues/11194593/?query=is%3Aunresolved

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1282)
┆Issue Number: EXP-1282

pdehaan:
See also #5180 / EXP-1280

PR #5188, opened 2021-05-07
fix #4158 bug(nimbus): experiment with branches but no feature won't v6 serializer

### LINKED ISSUE #4158: Audience form looks a bit glitchy between 434px and ~590px wide.
[/edit/audience](https://stage.experimenter.nonprod.dataops.mozgcp.net/experiments/nimbus/pdehaan-2020-12-07-005/edit/audience)

Looks good at 434px wide because things are on their own lines...
pdehaan_2020-12-07_005_–_Audience___Experimenter

At 1px wider (435px) some media query or something takes over and we switch to two columns for the "Percent of clients" and "Expected number of clients" fields. But now the "%" suffix on the "Percent of clients" field is moved to a different line...
pdehaan_2020-12-07_005_–_Audience___Experimenter

At 443px wide, the "%" moves back up to be adjacent to the "Percent of clients" input field (although you cannot see the value in the text field)...
pdehaan_2020-12-07_005_–_Audience___Experimenter

Things start to look normal again around 590px wide since now we can see double digit values for "Percent of clients" and "Experiment duration" (even with the extra padding/spacing needed to display the validation icons on the fields):

pdehaan_2020-12-07_005_–_Audience___Experimenter

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-683)
┆Issue Number: EXP-683

jaredlockhart:
Because

* When we changed experiments to always have a feature config we made it required to render the v6 serializer becuase we only used it when an experiment was fully filled out and ready to launch
* We recently added the recipe json to the V5 API and an experiment may have branches without a feature config while editing, so now we must check whether a feature config is set while rendering the v6 serializer

This commit

* Checks if a feature config is set while rendering the v6 serializer for a branch


pdehaan:
@jaredlockhart Can you confirm the "fix #nnnn" bug number? I think you mean #5186 (since 4158 is a minor responsive nit).

jaredlockhart:
oh good catch @pdehaan ! lol how did that happen

LZoog (approved):
👍 Thank you for the hot fix!

PR #5194, opened 2021-05-07
bug(nimbus): update test to use lifecycles

jaredlockhart:
Because

* We refactored our tests to use a new lifecycle method
* Missed one test!

This commit

* Updates the stray test to use the new lifecycle method


lmorchard (approved):
LGTM!

PR #5195, opened 2021-05-07
fix #5192 bug(nimbus): set defaults for risk questions for existing experiments

### LINKED ISSUE #5192: Add a data migration to set non null values for risk mitigation questions for existing experiments
We recently added new risk mitigation question fields to the NimbusExperiment model, which are now required to launch.  For new experiments, this is fine.  For launched experiments, this is fine.  But for exeriments that are ready to review but not yet launched, they're caught in a state where they're both not editable and not ready to review.  For those experiments we should set the risk questions to False becuase the risk mitigation information will be captured in the risk mitigation document that was filled out when the experiment was created.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1286)
┆Issue Number: EXP-1286

jaredlockhart:
Because

* We recently introduced new risk question fields
* They will default to None for existing experiments
* Experiments that are in review will be unable to review/launch without them filled in but also not be editable

This commit

* Sets the new risk questions to False for experiments that aren't in DRAFT/IDLE
* Note existing experiments that are in review will have already detailed their risk reviews in their risk document whose url we still preserve


lmorchard (approved):
:+1:

PR #5211, opened 2021-05-10
fixes #5114 feat(nimbus): Format feature config JSON, add syntax highlighter

### LINKED ISSUE #5114: Format feature value json nicely
We can probably use some library to format the json prettily, like [this](https://www.npmjs.com/package/react-json-view) or [this](https://www.npmjs.com/package/react-json-pretty)

LZoog:
fixes #5114 

Because:
* This was the last remaining bit of JSON in our summary tables that was not formatted

This commit:
* Creates a new Code component to use for when displaying code, use it in tables, add to Storybook
* Adds react-json-pretty for prettifying and syntax highlighting
* Suppresses a table warning on the request review page by adding tbody to recommended actions table


app/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
```diff
@@ -70,26 +71,23 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
...
-              data-testid="experiment-target-expression"
-              className="text-monospace"
-            >
-              <code>
-                <pre>{experiment.jexlTargetingExpression}</pre>
-              </code>
-            </td>
-          </tr>
-        )}
+        {experiment.jexlTargetingExpression &&
```
pdehaan, line 74: Interesting&hellip; Was this an issue before, or is this just so we don't try and JSON.stringify `null` or `undefined`? Curious why the extra check.

app/experimenter/nimbus-ui/src/components/FormattedJson/index.tsx
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+
+export const FormattedJson = ({ json }: { json: string }) => (
+  <code>
+    <pre className="mb-0">{json}</pre>
```
pdehaan, line 9: Oh! So this isn't _formatting_ the JSON, it's just a `<pre>` wrapper for pre-formatted JSON strings?

app/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -322,7 +322,7 @@ export function mockExperiment<
...
+        '{\n  "appId": "firefox-desktop",\n  "appName": "firefox_desktop",\n  "application": "firefox-desktop",\n  "arguments": {},\n  "branches": [\n    {\n       "feature": {\n       "enabled": false,\n       "featureId": null,\n       "value": {}\n     },\n     "ratio": 1,\n     "slug": "control"\n    }\n  ],\n  "bucketConfig": null,\n  "channel": "",\n  "endDate": null,\n  "featureIds": [],\n  "id": "pre-emptive-zero-tolerance-data-warehouse",\n  "isEnrollmentPa… "control",\n  "schemaVersion": "1.5.0",\n  "slug": "pre-emptive-zero-tolerance-data-warehouse",\n  "startDate": null,\n  "targeting": "\'app.shield.optoutstudies.enabled\'|preferenceValue",\n  "userFacingDescription": "Analysis art mean sort serve stuff. Scene alone current television up write company. Without admit she occur total generation by mother. Environmental remember account huge drive policy play strong.",\n  "userFacingName": "Pre-emptive zero tolerance data-warehouse"\n}',
```
pdehaan, line 325: My guts are a bit scared around here...

```js
,\n  "isEnrollmentPa… "control",\n  "schemaVersion": "1.5.0",\n
```

Not sure what is up with the ellipsis in the middle of `"isEnrollmentPa… "control"`.

app/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
jodyheavener: That's how I am reading it, you'll get into trouble if you try to stringify a null value

jodyheavener (approved):
LGTM save for Peter's comment about the malformed JSON mock string

app/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
LZoog: TypeScript wanted `experiment.jexlTargetingExpression!` sent into `<FormattedJson json={` to assert it's not undefined so I added the check here instead since we shouldn't display it otherwise.

app/experimenter/nimbus-ui/src/components/FormattedJson/index.tsx
LZoog: `<code>` and `<pre>`, right, though it does transform strings with `\n` into exactly what the output is. Could use a different name if there's a better one 🤔 

We may need another issue if we want long full targeting expressions or feature configs to format like the `recipeJson`. Check out what's returned by the server between `recipeJson` and other values: https://experimenter.services.mozilla.com/nimbus/howdy-doody

app/experimenter/nimbus-ui/src/lib/mocks.tsx
LZoog: Huh, that's a great catch, it must have come from when I copied the response. o.O

LZoog:
@pdehaan Great catch on `pre` needing to be outside of `code`. I'll take a look into the syntax coloring options since there's no rush on getting this in right this sec and this would be the PR to do that in 😎

LZoog:
OK so, I tried a couple different syntax highlighters thinking that's all we'd need, but then I poked around at some experiments and found at least this one: https://experimenter.services.mozilla.com/nimbus/multi-stage-onboarding-set-default-v42-experiment

We definitely need some sort of lib for formatting JSON, sticking the feature config with `<pre>` and `<code`> outputs this which we don't want at least for JSON:
image

I'll circle back to this first thing in the AM, sounds like we want prettifying _and_ syntax highlighting, but I don't want to introduce unnecessary or heavy deps.

LZoog:
I ended up using `json-react-pretty` to prettify some of the JSON which looks nice in Storybook and when checking it out locally. In [this discussion](https://github.com/mozilla/experimenter/pull/5168#discussion_r626577065) we opted not to use it, but since we wanted syntax highlighting, it seemed like one of the better options, as I tried out a couple others.

I also made a custom theme since the ones provided only looked good with a dark background.

PR #5216, opened 2021-05-10
fix #5214 bug(nimbus): restore original feature configs

### LINKED ISSUE #5214: Our directory view shows "No Feature" for all experiments under the Feature column
In the prod directory view in Nimbus all our experiments are showing "No Feature" under the feature column. 

Marking as a bug since it's not intentional behaviour. 

!image-2021-05-10-16-33-25-042.png|width=672,height=455!

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/EXP-1291)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/16212/image-2021-05-10-16-33-25-042.png">image-2021-05-10-16-33-25-042.png</a>
┆Issue Number: EXP-1291

jaredlockhart:
Because

* We recently made feature configs required for all experiments
* Past experiments existed with no feature config
* We wrote a migration to set experiments with no feature config to a default
* The migration missed the filter to check for feature_config=None and set ALL experiments to the default

This commit

* Restores the originally set feature config by looking it up in the changelog data
* Includes a data migration test, which we should always do from now on


app/experimenter/experiments/migrations/0170_restore_feature_configs.py
```diff
@@ -0,0 +1,34 @@
...
+from django.db import migrations
+
+
+def restore_feature_configs(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+    NimbusFeatureConfig = apps.get_model("experiments", "NimbusFeatureConfig")
+    for experiment in NimbusExperiment.objects.all():
+        feature_slug = (
+            experiment.changes.all()
+            .order_by("changed_on")
```
lmorchard, line 12: TL;DR: I think this should be `.order_by("-changed_on")` in order to catch the latest change and not the earliest change of feature config.

I was able to see my hunch happen:

1. `git co 91cd41d4`
2. `git co HEAD^` (i.e. the commit before the bad migration)
3. Create a draft experiment, change the feature config a few times
4. `git co main`
5. Run migrations
6. `git co upstream/5216`
7. Run migrations
8. The experiment was reset to the *first* feature config and not the *final* feature config before the migration

lmorchard (changes requested):
Don't want to block this too hard, but I think this would be good to go after flipping the order on `changed_on`. Ideally also tweak the test to change the feature config a few times before setting the expected and incorrect configs and verifying

jaredlockhart:
@lmorchard Yeah I was thinking about that but I don't think it's a case we need to worry about for now. BUT.  That being said, I realized, since we already know the names of the errant feature configs, we can just exclude those and take the most recent like you're suggesting, so I went ahead and did it anyways 👍

PR #5239, opened 2021-05-12
Fix #4755: feat(visualization) Confidence interval bars should reflect the scale of change.

### LINKED ISSUE #4755: Confidence interval bars should reflect the scale of the change
The confidence interval bars in the Experiment Console are schematic; the size and location of the bars indicate whether zero is included in the interval but don't actually reflect the size of the interval.

It would be great if the span of the bars reflected the size of the confidence interval.

It could be useful to choose a symmetric set of limits so that zero is always in the middle. Rounding the extreme value to a "pretty" number p and having the limits of the scale fall from (-p, p) is one way to achieve that.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1040)
┆Issue Number: EXP-1040

emtwo:
Because:
* The schematic for confidence interval bars was a fixed size and sometimes misleading

This commit:
* Makes the bars scale

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.tsx
@@ -6,108 +6,88 @@
 // TODO: EXP-638 remove this ^
 
 import React from "react";
-import { SIGNIFICANCE } from "../../../lib/visualization/constants";
 
-const renderBounds = (lower: number, upper: number, significance: string) => {
-  const padding = <div className="col p-0" />;
-  const numbers = (
-    <>
-      <div
-        className={`${significance}-significance col text-left p-0 h6 font-weight-normal`}
-      >
-        {lower}%
-      </div>
-      <div
-        className={`${significance}-significance col text-right p-0 h6 font-weight-normal`}
-      >
-        {upper}%
-      </div>
-    </>
-  );
+const BUFFER = 5;
 
-  switch (significance) {
-    case SIGNIFICANCE.NEGATIVE:
-      return (
-        <>
-          {numbers}
-          {padding}
-          {padding}
-        </>
-      );
-    case SIGNIFICANCE.POSITIVE:
-      return (
-        <>
-          {padding}
-          {padding}
-          {numbers}
-        </>
-      );
-    case SIGNIFICANCE.NEUTRAL:
-      return (
-        <>
-          {padding}
-          {numbers}
-          {padding}
-        </>
-      );
-  }
-};
+const renderBounds = (
+  lower: number,
+  upper: number,
+  leftPercent: number,
+  barWidth: number,
+  significance: string,
+) => (
+  <div
+    className="position-absolute"
+    style={{
+      // Add some buffer to space out the rendered bound values.
+      left: `${leftPercent - BUFFER * 2}%`,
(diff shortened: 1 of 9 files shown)


emtwo:
Reference screenshot with real data:

Screen Shot 2021-05-12 at 1 24 42 PM

app/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.tsx
```diff
@@ -6,108 +6,89 @@
...
+  barWidth: number,
+  significance: string,
+  buffer: number,
+) => (
+  <div
+    className="position-absolute"
+    style={{
+      // Add some buffer to space out the rendered bound values.
+      left: `${leftPercent - buffer * 2}%`,
+      width: `${barWidth + buffer * 3}%`,
```
LZoog, line 23: Normally I'm like, grrr no inline styles! But... makes sense here. 😆

app/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.tsx
```diff
@@ -6,108 +6,89 @@
...
-  if (SIGNIFICANCE.NEUTRAL === significance) {
-    position = "py-1 mt-2";
-  }
+const renderBounds = (
+  lower: number,
+  upper: number,
+  leftPercent: number,
+  barWidth: number,
+  significance: string,
+  buffer: number,
```
LZoog, line 16: I think it'd make sense to take `const BUFFER = 5` out of `ConfidenceInterval` and stick it above `renderBounds`, that way you don't have to pass it as a prop here.

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -55,3 +65,29 @@ export const getSortedBranches = (analysis: AnalysisData) => {
   });
   return sortedBranches;
 };
+
+/**
+ * Find the most extreme upper or lower bound value
+ * for an outcome across all branches.
+ *
+ * This is used to scale the confidence interval bars
+ * shown for a metric.
```
LZoog, line 74: TY for explainer comment😁

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -55,3 +65,29 @@ export const getSortedBranches = (analysis: AnalysisData) => {
...
+    const metricDataList =
+      results[branch].branch_data[outcomeSlug][branchComparison]["all"];
+    metricDataList.forEach((dataPoint: FormattedAnalysisPoint) => {
+      const { lower, upper } = dataPoint;
+      const max = Math.max(Math.abs(lower!), Math.abs(upper!));
+      extreme = max > extreme ? max : extreme;
+    });
+  });
+  return extreme;
+};
```
LZoog, line 93: It almost feels like this could use a test to make sure it's setting this as expected but I know it's one of our `utils` so it's excluded from coverage... Maybe we could use a test somewhere else to test that `bounds` prop?

LZoog (approved):
I have a couple suggestions (one minor, one test), but wouldn't block those on a r+. 🙂

jaredlockhart:
Neat @emtwo looks really good!

PR #5251, opened 2021-05-13
fixes #5249 chore(project): add some default reviewers for DB migrations

### LINKED ISSUE #5249: For extra coverage of DB migrations, add some additional code owners
Since we've had some bugs happen recently in the neighborhood of database migrations, it was suggested in retro that we put some more attention on them in terms of review. One way to do this is to assert the need for at least two reviewers on PRs with migrations. Another is adding some DB-informed folks to CODEOWNERS.

This task is about the latter, adding folks to CODEOWNERS.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1318)
┆Issue Number: EXP-1318

lmorchard:
Because:

* We'd like some elevated attention to PRs involving database migrations

This commit:

* Adds some folks to CODEOWNERS for changes to DB migrations

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,6 @@
 app/tests @jrbenny35 @pdehaan
+app/experimenter/base/migrations @jaredlockhart @lmorchard
+app/experimenter/experiments/migrations @jaredlockhart @lmorchard
+app/experimenter/notifications/migrations @jaredlockhart @lmorchard
+app/experimenter/projects/migrations @jaredlockhart @lmorchard
+app/experimenter/reporting/migrations @jaredlockhart @lmorchard


lmorchard:
Just as a stake in the ground, I added myself and @jaredlockhart to the CODEOWNERS here.

Should anyone else be added?

.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
-app/tests @jrbenny35 @pdehaan
\ No newline at end of file
+app/tests @jrbenny35 @pdehaan
+app/experimenter/base/migrations @jaredlockhart @lmorchard
```
pdehaan, line 2: @lmorchard What about the migrations at [app/experimenter/experiments/migrations](https://github.com/mozilla/experimenter/tree/main/app/experimenter/experiments/migrations)?

lmorchard: Good catch! I missed 'em!

I just ran `find -type d -name 'migrations'` and added everything I found, although I don't know what some of these are for

pdehaan: Oh, you found way more folders. Well, now I'm wondering if they support complex globs and we could do something like `app/experimenter/**/migrations`.

The docs don't explicitly show double star glob in https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax, but they do say they follow the .gitignore syntax (with a few exceptions).

pdehaan (approved):
R+ from me. Not that anybody was asking/waiting; but in case you wanted a set of 👀  before merging.

lmorchard:
I think I'll just merge this - we can further iterate as needed

jaredlockhart:
Awesome @lmorchard ! Neat!

PR #5276, opened 2021-05-18
fixes #5116 feat(nimbus): filter targeting choices by application

### LINKED ISSUE #5116: Audiences should be filtered by application
Audiences should:

* Apply to one or more application 
* Only show up in the UI for matching applications

Note that all audiences currently in that file are **Desktop only**

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1242)
┆Issue Number: EXP-1242

lmorchard:
Because:

* audience targeting should be considered as application-dependent

This commit:

* associates each NimbusTargetingConfig with an ApplicationConfig -
  currently all APPLICATION_CONFIG_DESKTOP

* adds NimbusExperimentTargetingConfigChoice type in GQL to represent
  targeting choices augmented with application filtering context

* filters targeting choices in nimbus-ui by current experiment
  application

* make NO_TARGETING a proper NimbusTargetingConfig instance for
  consistency

### DIFF
--- a/app/experimenter/experiments/api/v5/queries.py
+++ b/app/experimenter/experiments/api/v5/queries.py
@@ -2,6 +2,7 @@ import graphene
 from django.conf import settings
 
 from experimenter.experiments.api.v5.types import (
+    NimbusExperimentTargetingConfigSlugChoice,
     NimbusExperimentType,
     NimbusFeatureConfigType,
     NimbusLabelValueType,
@@ -22,7 +23,7 @@ class NimbusConfigurationType(graphene.ObjectType):
     feature_config = graphene.List(NimbusFeatureConfigType)
     firefox_min_version = graphene.List(NimbusLabelValueType)
     outcomes = graphene.List(NimbusOutcomeType)
-    targeting_config_slug = graphene.List(NimbusLabelValueType)
+    targeting_config_slug = graphene.List(NimbusExperimentTargetingConfigSlugChoice)
     hypothesis_default = graphene.String()
     max_primary_outcomes = graphene.Int()
     documentation_link = graphene.List(NimbusLabelValueType)
@@ -53,7 +54,16 @@ class NimbusConfigurationType(graphene.ObjectType):
         return Outcomes.all()
 
     def resolve_targeting_config_slug(root, info):
-        return root._text_choices_to_label_value_list(NimbusExperiment.TargetingConfig)
+        return [
+            NimbusExperimentTargetingConfigSlugChoice(
+                label=choice.label,
+                value=choice.name,
+                application_values=NimbusExperiment.TARGETING_CONFIGS[
+                    choice.value
+                ].application_choice_names,
+            )
+            for choice in NimbusExperiment.TargetingConfig
+        ]
 
     def resolve_hypothesis_default(root, info):
         return NimbusExperiment.HYPOTHESIS_DEFAULT
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -464,6 +464,21 @@ class NimbusExperimentSerializer(
                     )
                 }
             )
+
+        if self.instance and "targeting_config_slug" in data:
+            targeting_config_slug = data["targeting_config_slug"]
+            application_choice = NimbusExperiment.Application(self.instance.application)
+            targeting_config = NimbusExperiment.TARGETING_CONFIGS[targeting_config_slug]
+            if application_choice.name not in targeting_config.application_choice_names:
+                raise serializers.ValidationError(
+                    {
+                        "targeting_config_slug": (
+                            f"Targeting config '{targeting_config.name}' is not "
+                            f"available for application '{application_choice.label}'"
+                        )
+                    }
+                )
+
         return data
 
     def update(self, experiment, validated_data):
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -6,105 +6,6 @@ from django.conf import settings
 from django.db import models
 
 
-@dataclass
-class NimbusTargetingConfig:
-    name: str
-    slug: str
-    description: str
-    targeting: str
-    desktop_telemetry: str
-
-
-TARGETING_ALL_ENGLISH = NimbusTargetingConfig(
-    name="All English users",
-    slug="all_english",
-    description="All users in en-* locales.",
-    targeting="localeLanguageCode == 'en'",
-    desktop_telemetry="STARTS_WITH(environment.settings.locale, 'en')",
-)
-
-TARGETING_US_ONLY = NimbusTargetingConfig(
-    name="US users (en)",
-    slug="us_only",
-    description="All users in the US with an en-* locale.",
-    targeting="localeLanguageCode == 'en' && region == 'US'",
-    desktop_telemetry=(
-        "STARTS_WITH(environment.settings.locale, 'en') "
-        "AND normalized_country_code = 'US'"
-    ),
-)
-
-TARGETING_FIRST_RUN = NimbusTargetingConfig(
-    name="First start-up users (en)",
-    slug="first_run",
-    description=("First start-up users (e.g. for about:welcome) with an en-* " "locale."),
-    targeting=("{en} && (({is_first_startup} && {not_see_aw}) || {sticky})").format(
-        en=TARGETING_ALL_ENGLISH.targeting,
-        is_first_startup="isFirstStartup",
-        not_see_aw="!('trailhead.firstrun.didSeeAboutWelcome'|preferenceValue)",
-        sticky="experiment.slug in activeExperiments",
-    ),
-    desktop_telemetry=(
-        "STARTS_WITH(environment.settings.locale, 'en') "
-        "AND payload.info.profile_subsession_counter = 1"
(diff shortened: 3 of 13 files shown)


app/experimenter/experiments/api/v5/queries.py
```diff
@@ -53,7 +54,26 @@ def resolve_outcomes(root, info):
...
+                or ""
+            )
+            resolved_configs.append(
+                NimbusExperimentTargetingConfigSlugChoice(
+                    label=targeting_config.name,
+                    value=value,
+                    application_value=application_value,
+                )
+            )
+        return resolved_configs
```
lmorchard, line 76: Trying to think of a more elegant way to do this. But, we've got a bit of a tangle here between our dataclasses and `models.TextChoices` members.

We use slugs in the database. But, forms & mutations use `name` from `TextChoices` members. These are not the same values. (i.e. targeting slug "first_run_win1903" is represented by a choice named `TARGETING_FIRST_RUN_WINDOWS_1903_NEWER` and application slug "firefox-desktop" corresponds to choice `DESKTOP`)

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -6,21 +6,95 @@
...
+    RELEASE = "release"
+
+
+class BucketRandomizationUnit(models.TextChoices):
+    NORMANDY = "normandy_id"
+    NIMBUS = "nimbus_id"
+
+
+@dataclass
+class ApplicationConfig:
```
lmorchard, line 23: Weird diff, but just moving application definitions before targeting so they can specify applications.

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -6,21 +6,95 @@
...
 class NimbusTargetingConfig:
     name: str
     slug: str
     description: str
     targeting: str
     desktop_telemetry: str
+    application: ApplicationConfig
 
 
+TARGETING_NO_TARGETING = NimbusTargetingConfig(
```
lmorchard, line 82: Adding this as a full-fledged targeting choice made some other code simpler. Shouldn't work the same otherwise

lmorchard (commented):
Gave things a self-review and called out a couple points of interest

app/experimenter/experiments/api/v5/queries.py
jaredlockhart: Hold on let's rethink this for a second.  In the UI, we only care about application or targeting `name`, we don't use `slug` at all for either right?

So if in our targeting configs we say

```py
    application = Application.DESKTOP.name
```

and just pass that off verbatim to the frontend, we do the filtering on the frontend, frontends `name` back, we already resolve that back to the value, then we don't need to do this transformation right?

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Oh interesting.  Here we have `application=None` meaning it applies to all of them, I just asked @k88hudson and she suggested we may have targeting that applies to more than one application, so maybe this should be a list and in this example we put all three in?

jaredlockhart:
Also we'll want to validate this in the V5 serializer.  We can either do that here or as a followup, if we wanna do it as a followup we should file another ticket.

jaredlockhart (commented):
Added a couple comments.

app/experimenter/experiments/api/v5/queries.py
lmorchard: I can do that, though I figured that just pushes the hack around a bit. At some point if we need to refer to the rest of the application associated in targeting, we need to do a different mapping. But, for the sake of YAGNI in this current PR, just using the name would be cleaner.

jaredlockhart: Yeah exactly we can deal with that later if we need to.

lmorchard: Oh, yeah, and I was also hoping that specifying `application: ApplicationConfig` would get some type-checking benefit. But that doesn't seem to do what I thought it would

app/experimenter/experiments/constants/nimbus.py
lmorchard: Switched over to using a list of application choice names for this

app/experimenter/experiments/api/v5/queries.py
lmorchard: Reworked this to use a list of choice names specified for each targeting config

lmorchard:
> Also we'll want to validate this in the V5 serializer.

Going to add that today after some meetings

app/experimenter/experiments/api/v5/queries.py
```diff
@@ -53,7 +54,19 @@ def resolve_outcomes(root, info):
...
+                label=member.label,
+                value=choice_name,
+                application_values=NimbusExperiment.TARGETING_CONFIGS[
+                    member.value
+                ].application_choice_names,
+            )
+            for (
+                choice_name,
+                member,
+            ) in NimbusExperiment.TargetingConfig.__members__.items()
```
jaredlockhart, line 68: You can just loop over an Enum and get the individual entries and call `.name` `.value` on them instead of digg into private attributes

jaredlockhart, line 68: Enums are iterable, you don't need to dig into private attributes.

```py
In [7]: for targeting in NimbusConstants.TargetingConfig:
   ...:     print(targeting.name, targeting.value)
   ...: 
NO_TARGETING 
ALL_ENGLISH all_english
US_ONLY us_only
TARGETING_FIRST_RUN first_run
TARGETING_FIRST_RUN_CHROME_ATTRIBUTION first_run_chrome
TARGETING_FIRST_RUN_WINDOWS_1903_NEWER first_run_win1903
TARGETING_HOMEPAGE_GOOGLE homepage_google_dot_com
TARGETING_URLBAR_FIREFOX_SUGGEST urlbar_firefox_suggest_us_en
```

app/experimenter/experiments/api/v5/queries.py
```diff
@@ -53,7 +54,19 @@ def resolve_outcomes(root, info):
         return Outcomes.all()
 
     def resolve_targeting_config_slug(root, info):
-        return root._text_choices_to_label_value_list(NimbusExperiment.TargetingConfig)
+        return [
+            NimbusExperimentTargetingConfigSlugChoice(
+                label=member.label,
+                value=choice_name,
+                application_values=NimbusExperiment.TARGETING_CONFIGS[
```
jaredlockhart, line 61: Maybe `application_names` is more appropriate?

lmorchard: I went with `application_values` here since it's essentially the same kind of data as `value` - i.e. choice names. It's all kind of a tangle of terms.

app/experimenter/experiments/api/v5/queries.py
lmorchard: Ah, could have sworn that didn't work when I tried it

app/experimenter/experiments/api/v5/queries.py
```diff
@@ -53,7 +54,16 @@ def resolve_outcomes(root, info):
...
-        return root._text_choices_to_label_value_list(NimbusExperiment.TargetingConfig)
+        return [
+            NimbusExperimentTargetingConfigSlugChoice(
+                label=choice.label,
+                value=choice.name,
+                application_values=NimbusExperiment.TARGETING_CONFIGS[
+                    choice.value
+                ].application_choice_names,
+            )
+            for choice in NimbusExperiment.TargetingConfig
```
jaredlockhart, line 65: 👍

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -464,6 +464,25 @@ def validate(self, data):
...
                 }
             )
+
+        if self.instance and "targeting_config_slug" in data:
+            targeting_config_slug = data["targeting_config_slug"]
+            application_choice = next(
+                choice
+                for choice in NimbusExperiment.Application
+                if choice.value == self.instance.application
+            )
```
jaredlockhart, line 474: Oh thank you for adding!  You can look up the name of an enum from the value like this:

```py
In [7]: experiment.Application(experiment.application).name
Out[7]: 'FENIX'
```

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -419,6 +442,7 @@ class EmailType(models.TextChoices):
     TARGETING_CHANNEL = 'browserSettings.update.channel == "{channel}"'
 
     TARGETING_CONFIGS = {
+        TARGETING_NO_TARGETING.slug: TARGETING_NO_TARGETING,
```
jaredlockhart, line 445: I forget why we have this dict and the choices below it, we should only need one I think...  Not for this PR I mean, just something to maybe cleanup later.

jaredlockhart (approved):
Yep looks right, tested locally, g/j 👍 Thanks for adding the serializer part too 🎉 

You can simplify that enum lookup but otherwise g2g 💯 💯 💯

app/experimenter/experiments/api/v5/serializers.py
lmorchard: Oh hey that worked, I was trying `[]` subscript notation and getattr and getting nowhere

Issue #4899, opened 2021-03-24
[Nimbus] A 2nd branch of an experiment is redisplayed in the creation form after it is removed


SoftVision-CiprianMuresan:
**[Steps to reproduce]:**
1. Navigate to the https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/ website.
2. Create a new experiment.
3. Fill all sections except the Branches one with valid data and save at each step.
4. Navigate to the Branches section and fill out one branch with valid data.
5. Remove the second branch and Save.
6. Refresh the page and observe the behavior.

**[Expected result]:**
- The experiment form remains unchanged and only shows a single branch.

**[Actual result]:**
- A 2nd empty branch is displayed.

**[Notes]:**
- The 2nd branch is also displayed on the `request-review` page in red.
- The 2nd branch does not appear in the Remote Settings 
- Attached is a screen recording of the issue.

![the 2nd branch of an experiment is redisplayed in the creation form after it is removed](https://user-images.githubusercontent.com/18304511/112313262-54d04480-8cb0-11eb-848f-2fb346f2b69d.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1107)
┆Issue Number: EXP-1107

lmorchard:
I think this is working as intended: experiments need to have at least two branches - control and one treatment branch. [So, we create a default for control and the first treatment branch if either is missing](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/types.py#L152-L160).

SoftVision-CiprianMuresan:
I understand, but in this case, we would need to have validation on the Branches page to expect 2 or more branches and not let users continue with a single branch. As it is now, the form heavily implies that you _need_ 2 branches, but lets you build an experiment with a single branch nonetheless.

SoftVision-CiprianMuresan:
I have verified that the issue is no longer reproducible on the Experimenter stage platform. A second branch is no longer displayed in the "Branches" or "Review" pages when a single branch experiment is set up.

PR #5315, opened 2021-05-27
fixes #4289 feat(nimbus): display review-readiness messages on branch fields

### LINKED ISSUE #4289: Update UI of review-readiness message for individual branch fields

### LINKED ISSUE #4653: Use faked “missing fields” messages when control branch is null
When an experiment is missing its control branch we currently display the missing ( ! ) icon in the top right of the branch container. This is because the entire branch object is missing, instead of individual fields being missing.

In an effort to be more clear to the user about what exactly is required, instead of indicating that the control branch is missing, we need to update this so that it highlights each required field (name and description) of the missing branch.

This will require us to create fake review-readiness messages and insert them into the readyForReview data as it passes through the FormBranches component when it sees that reference_branch is in the list of errors. Both fields then need to render the messages in the same visual style other review-readiness messages are presented, only visible when the ?show-errors query param is present (refer to other required fields to see how this is implemented).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-978)
┆Issue Number: EXP-978

jodyheavener:
Closes #4289
Closes #4653

This PR...

Frontend:

- Updates the Branches page to display review-readiness messages directly on the individual branch fields, making this consistent with the rest of our new review message style.
  - As a result, this removes the last (!) warning icon from the code.
- Updates the `SerializerMessages` type declaration to account for an object as a Record value (it currently only supports values that are an array of strings or an array of objects).
- Updates any stories touched to modern Storybook format

Backend:

- Adds an empty control branch one experiment creation (so the control is never null).
- Updates the `NimbusReadyForReviewSerializer`'s validation of branch descriptions to return a dict containing a key/value pair instead of an array of strings.

[Storybook preview](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/fb11e53e5850fb02110fc59e4abea26808361dd2/nimbus-ui/index.html?path=/story/pages-editbranches--missing-fields)


jodyheavener:
Ahh @jaredlockhart it looks like creating that initial reference branch is resetting the empty treatment branch (hence the integration failure). Hmm. Since that treatment branch is really only necessary until you create your own we could fake it in the client?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -553,18 +561,20 @@ class Meta:
 
     def validate_reference_branch(self, value):
         if value["description"] == "":
-            raise serializers.ValidationError("Description cannot be blank.")
+            raise serializers.ValidationError(
+                {"description": ["This field cannot be blank."]}
```
LZoog, line 565: This is unrelated to your PR but a teeny win you could get in at least here - I met with Ana earlier to discuss review messaging and since the default error messages use "may not" instead of "cannot," that's the nomenclature we went with on the risk mitigation questions. This should probably be "This field may not be blank." to match the other generic blank field messages.

It may also be worth adding this to the [`NimbusConstants` serializer error messages](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L500).

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -156,6 +156,14 @@ def _validate_feature_configs(self, data):
 
         return data
 
+    def create(self, validated_data):
+        experiment = super().create(validated_data)
+        experiment.reference_branch = NimbusBranch.objects.create(
+            experiment=experiment, slug="control", name="Control", feature_enabled=False
```
LZoog, line 162: > Adds an empty control branch one experiment creation (so the control is never null).

Why do we need to do this?

Assuming we leave this in, we need to:
* Update [the logic](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx#L105) around showing both branches initially and only showing one if the user only saved one, which might be tricky, and would otherwise re-open #4899
* Remove `NO_BRANCHES_COPY` and logic around it
* I think we can [take "control" out of `annotateExperimentBranch`](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/state.ts#L70-L73)
* `FormBranches` Storybook "empty branches" is no longer a valid state, and makes me wonder if we should search for other mock reference branches and fill in the slug/name/feature config to match

app/experimenter/nimbus-ui/src/lib/test-utils.tsx
```diff
@@ -76,24 +82,32 @@ export const assertSerializerMessages = async (
...
-    if (typeof errors[0] === "object") {
-      let index = 0;
-      for (const set of errors) {
-        for (const [innerField, innerErrors] of Object.entries(set)) {
-          await screen.findByText(innerErrors.join(", "), {
-            selector: `[data-for="${snakeToCamelCase(
-              field,
-            )}[${index}].${snakeToCamelCase(innerField)}"]`,
-          });
+    // - an object containing this same thing
```
LZoog, line 85: Oh man, this is getting dicey lol. I don't have a suggestion for improvement though, I think we definitely need the individual errors on the branches the way you did it for the UI.

jodyheavener: Yeah I am not a fan of the variety of shapes a serializer response can be

app/experimenter/experiments/api/v5/serializers.py
jodyheavener: Basically, the goal of this is to surface validation feedback on the individual fields that make up a branch, and in order to do that the branch needs to first exist -- otherwise the only validation that can occur is on `reference_branch` itself, which would just be null.

There are likely alternatives here, such as looking for the null reference branch validation message in the client and replacing it with an object containing the individual field messages, or finding some way to intercept the null reference branch validation message in the serializer and replace it with a dict containing the individual field messages. But this appears to be the cleanest way. Definitely open to other options that can accomplish this

LZoog: > ...  otherwise the only validation that can occur is on `reference_branch` itself, which would just be null.

Huh, I didn't dig but I wonder why the entire branch comes back `null` since if you create a new experiment, a `referenceBranch` object is created with default values. Sure the branch isn't valid, but it's also still invalid without a branch description filled out.

> ... finding some way to intercept the null reference branch validation message in the serializer and replace it with a dict containing the individual field messages

I don't think this is a bad option, I like it better than doing this in the FE at least. 🤔

I'm OK with the approach if it's the cleanest way but I do think my points in my initial comment have to be addressed. 😩 (or at least the first point, others could be handled in maintenance follow ups, but bleh)

jodyheavener: > I guess I'm wondering if the branch is actually ever null? Seems like at experiment creation time we get this guy back on the client?

That's coming from [this field resolver](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/types.py#L180:L183). On field lookup if there is a reference branch return it, other instantiate a new one. But this only occurs on field lookup. Actually, if we do keep this change of creating a reference branch on experiment creation I should should remove this bit here as it will be redundant.

> I don't think this is a terrible option, I like it better than doing this in the FE at least.

Yeah this was my initial approach but we thought it might be cleaner to just do the initial branch creation. So I'm a little torn!

@jaredlockhart what do you make of this?

LZoog: So actually, is it possible to just create the branch with `slug` and `name` of the branch set to the defaults, which are empty strings? I think that would actually fix everything. 😆

LZoog: Following up here - seems like we can get away without setting the `slug`, which is great since we can change the `savedBranches`/other similar references in the FE to check branch `slug` instead of `name`, and it doesn't mess with that first bullet point I laid out here.

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -566,18 +574,20 @@ class Meta:
...
         for branch in value:
-            error = None
+            error = {}
             if branch["description"] == "":
-                error = ["Description cannot be blank."]
+                error["description"] = [NimbusConstants.ERROR_REQUIRED_FIELD]
             errors.append(error)
 
-        if any(x is not None for x in errors):
+        if any(x for x in errors):
```
jaredlockhart, line 590: Oh then you can just say `any(errors)`

app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.test.tsx
```diff
@@ -150,15 +150,15 @@ describe("TableBranches", () => {
...
+              slug: "treatment-1",
               description: "",
               ratio: 0,
               featureValue: null,
               featureEnabled: true,
             },
             {
               name: "",
-              slug: "",
+              slug: "treatment-2",
```
LZoog, line 161: Why change this one? I don't think you can have a slug set without a name set.

jodyheavener: You're right, and my change made the test invalid anyway, so I've updated it 👍

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -156,6 +156,14 @@ def _validate_feature_configs(self, data):
 
         return data
 
+    def create(self, validated_data):
+        experiment = super().create(validated_data)
+        experiment.reference_branch = NimbusBranch.objects.create(
+            experiment=experiment, name="control", feature_enabled=False
```
jaredlockhart, line 162: Shouldn't you set slug here?  Or are you relying on the fact that they'll have to set the description, which will force them to save, which will then generate the slug?

LZoog (approved):
Not sure if you want to wait for Jared's review but LGTM! Tried it out locally and everything works as expected. Thanks for the discussions. 😁

jaredlockhart (commented):
Looks good!

I ran through it locally, and I see it goes in three steps:

1) Control exists, treatment does not, so only control warns about requiring description:

Screen Shot 2021-05-28 at 12 21 21 PM

2) Save control, now it's fine, treatment exists, so its description is now required:

Screen Shot 2021-05-28 at 12 21 34 PM

3) Save treatment, now feature config is required but we don't warn at all:

Screen Shot 2021-05-28 at 12 21 46 PM


So we could either leave that as it is and call it fine, or we could create the one treatment branch at creation time too so then it warns about both up front?  And if they don't want treatment they can delete it.  Or we could just leave it as it is.

But we should find some way to warn about feature config being required?  Otherwise it just says branches is incomplete but doesn't tell you what's missing?

app/experimenter/experiments/api/v5/serializers.py
LZoog: Setting the slug requires a bunch of other edits: https://github.com/mozilla/experimenter/pull/5315#discussion_r640941887

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Yeah I think it can go either way, since leaving the description blank forces the user to save which then sets the slug.  Probably fine like this.

jodyheavener:
> But we should find some way to warn about feature config being required?

@jaredlockhart we have that [here](https://jira.mozilla.com/browse/EXP-1231) :)

jaredlockhart:
> > But we should find some way to warn about feature config being required?
> 
> @jaredlockhart we have that [here](https://jira.mozilla.com/browse/EXP-1231) :)

Perf 👍

jaredlockhart (approved):
Yep looks good 👍

PR #5318, opened 2021-05-28
fixes #5321: fix(visualization): Make results page load faster. 

### LINKED ISSUE #5321: Speed up results page load
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1361)
┆Issue Number: EXP-1361

emtwo:
Note that given the way this caching is done, we will need to run the task once on prod right after deploying to pull in the latest data.


app/experimenter/visualization/api/v3/views.py
```diff
@@ -1,136 +1,11 @@
...
-        elif data and window == "weekly":
-            ResultsObjectModel = create_results_object_model(data)
-            data = ResultsObjectModel(result_metrics, data, experiment, window)
-
-        response_data = data.dict(exclude_none=True) or None
-
-        experiment_data[window] = response_data
-
-    return Response(experiment_data)
+    return Response(experiment.results_data or {})
```
jaredlockhart, line 11: If we wanna do this maybe make it the default value on the field?

jaredlockhart (changes requested):
Very nice!  This will be such a good change.  Couple ideas on moving some stuff around, adding logging to the task, but generally looks good 👍

tdsmith:
I was looking at the v3 views the other day and realized that we hadn't thought about how splitting the results into multiple files was creating challenges. An architectural alternative to caching here could be to use [signed URLs](https://cloud.google.com/storage/docs/access-control/signing-urls-manually) and [gzip-encoded data](https://cloud.google.com/storage/docs/transcoding) to push small blobs directly to the client from GCS instead of round-tripping them through Django at all.

Will there be a way to trip this by hand? It'd be nice to have less than 24 h of latency, especially if Experimenter deploys and the ETL milk run get out of sync or if we run new results by hand.

jaredlockhart:
@tdsmith good points.  The other motivating factor here was a recent conversation with @shell1 where she indicated we'll want to be able to sort/filter based on experiment results, so for that we'd need them in the database so I was hoping to kill two birds with one stone here, fetch them into the db for fast lookups and make them available to other future use cases.

emtwo:
We can also add a check for the timestamp on the data files and if it's sooner than the most recent cache, then pull in the update

jaredlockhart:
@emtwo if jetstream only runs once a day and we check once a day what's the case where they'll be more up to date?

emtwo:
> @emtwo if jetstream only runs once a day and we check once a day what's the case where they'll be more up to date?

If we update the cache in the morning, for example, and Jetstream updates the data at noon. Then if someone queries results from experimenter at 4pm the jetstream files will be more recently updated than the cache.

jaredlockhart (changes requested):
Yeah all looks right but we should add all the metrics and logging to the tasks so we can track them

app/experimenter/jetstream/tasks.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@app.task
+@metrics.timer_decorator("fetch_experiment_data")
+def fetch_experiment_data(experiment_id):
+    metrics.incr("fetch_experiment_data.started")
+    try:
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        experiment.results_data = get_experiment_data(experiment)
+        experiment.save()
```
jaredlockhart, line 22: `metrics.incr("fetch_experiment_data.completed")` here

app/experimenter/jetstream/tasks.py
```diff
@@ -0,0 +1,59 @@
...
+                logger.info(
+                    "Skipping cache refresh for old experiment {name}".format(
+                        name=experiment.name
+                    )
+                )
+
+                continue
+
+            logger.info("Fetching Jetstream data for {name}".format(name=experiment.name))
+            fetch_experiment_data.delay(experiment.id)
```
jaredlockhart, line 55: `metrics.incr("fetch_jetstream_data.completed")` here

jaredlockhart (approved):
Yep, just two `completed` metrics but g2g! Thanks @emtwo !  This is gonna really speed things up 👏 👏 👏 👏 👏

PR #5352, opened 2021-05-31
fix #5351 bug(nimbus): paused experiments are stuck in waiting

### LINKED ISSUE #5351: Experiments not pausing
Ah, I found why experiments aren't pausing.  We added all these fancy declarative filters for all the states of the experiment, including one for IS_PAUSE_CANDIDATE:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L237-L241

which is used here to find experiments to pause:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L35

but we also used it here in the task to update experiments to paused:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L293

but at that point the experiment is in `WAITING` not `IDLE` so no experiments are getting updated, that's why we have all those experiments sitting in `WAITING` even though they've been successfully paused in RS.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1375)
┆Issue Number: EXP-1375

jaredlockhart:
Because

* We recently refactored everything to use these handy declarative lifecycles
* We accidentally used the wrong one when considering experiments to update their pause state

This commit

* Renames some of the lifecycles to be more consistent
* Adds the necessary lifecycle for experiments waiting to be paused
* Updates tasks/tests

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -234,12 +234,17 @@ class LifecycleStates(Enum):
         "status": Status.LIVE,
         "publish_status": PublishStatus.IDLE,
     }
-    PAUSE_CANDIDATE = {
+    LIVE_IDLE_ENROLLING = {
         "status": Status.LIVE,
         "publish_status": PublishStatus.IDLE,
         "is_paused": False,
     }
-    PAUSED = {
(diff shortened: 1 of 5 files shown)


jodyheavener (approved):
Neat!

PR #5354, opened 2021-06-01
fix #5353 chore(nimbus): split lifecycles and filters

### LINKED ISSUE #5353: Reusing state constants between logic and tests can hide errors
While working on https://github.com/mozilla/experimenter/issues/5351 I realized one of the reasons we missed this is that sharing the state definitions between test and logic masked the error.  The state definition was incorrect but tests were passing.  Generally I think tests should hard code all their expectations independently of the implementation they're testing, so I think in this case that just means not reusing the lifecycle definitions in the experiment filters, which isn't that bad since there's not that many.  Hopefully that will help expose errors like this in the future.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1378)
┆Issue Number: EXP-1378

jaredlockhart:
Because

* While working on a recent bug I discovered that one of our lifecycle states was malformed
* Tests were passing becuase the same lifecycle state was used as part of logic and part of tests
* Using separate state definitions for logic and tests can help catch future bugs

This commit

* Moves the lifecycles over to the test factory and hard codes definitions in the model filters

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass
-from enum import Enum
 from typing import Dict
 
 from django.conf import settings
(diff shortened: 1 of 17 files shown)


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -195,118 +194,19 @@ class NimbusTargetingConfig:
 )
 
 
-class Status(models.TextChoices):
```
lmorchard, line 198: > While working on a recent bug I discovered that one of our lifecycle states was malformed

Which one was that? This move obscures the change

lmorchard (approved):
I'd rather see the constants duplicated rather than move & hardcode, but we can revisit that if/when the state patterns are used more extensively on the front-end.

Might also be good to split the bug fix from the refactor in separate commits. It's hard to work out what changed here in the combined diff

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Oh sorry I didn't link there's no bugfix here, just moving the constants.  Bugfix was here https://github.com/mozilla/experimenter/issues/5351

PR #5359, opened 2021-06-01
fix #5356 bug(nimbus): handle multiple launch/end changelogs

### LINKED ISSUE #5356: Handle multiple changelogs in start date
Looking at the V5 API on stage, we're seeing the following, I suspect it's related to computing dates:

```json
{
	"errors": [{
		"message": "get() returned more than one NimbusChangeLog -- it returned 2!",
		"locations": [{
			"line": 9,
			"column": 5
		}],
		"path": ["experiments", 4, "startDate"]
	}, {
		"message": "get() returned more than one NimbusChangeLog -- it returned 2!",
		"locations": [{
			"line": 12,
			"column": 5
		}],
		"path": ["experiments", 4, "computedEndDate"]
	}, {
		"message": "get() returned more than one NimbusChangeLog -- it returned 2!",
		"locations": [{
			"line": 16,
			"column": 5
		}],
		"path": ["experiments", 4, "monitoringDashboardUrl"]
	}, {
		"message": "get() returned more than one NimbusChangeLog -- it returned 2!",
		"locations": [{
			"line": 17,
			"column": 5
		}],
		"path": ["experiments", 4, "resultsReady"]
	}]
}
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1380)
┆Issue Number: EXP-1380

jaredlockhart:
Because

* We expect an experiment to only launch or end once and only have single launch/end changelogs
* Real data is muddy and messy
* We should elegantly handle the case where there's multiple starts/ends

This commit

* Uses the most recent launch or end changelog to compute the start/end dates

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -239,20 +239,20 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
 
     @property
     def start_date(self):
-        start_changelog = self.changes.filter(
-            new_status=NimbusExperiment.Status.LIVE,
+        start_changelogs = self.changes.filter(
             old_status=NimbusExperiment.Status.DRAFT,
+            new_status=NimbusExperiment.Status.LIVE,
         )
-        if start_changelog.exists():
-            return start_changelog.get().changed_on
+        if start_changelogs.exists():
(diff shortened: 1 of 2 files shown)


jaredlockhart:
Actually hold on let me do this in another order.

jodyheavener (approved):
sick 🚢

PR #5373, opened 2021-06-04
feat(nimbus): proxy nimbus-ui webpack dev server through nginx

lmorchard:
Because:

* The way we link to the nimbus-ui JS in local development differs from
  production in a way that confused at least one tool (i.e. Sentry)

* We're not taking advantage of webpack dev server features like hot
  reloading and richer exception handling

This commit:

* Switches from using the nimbus-ui webpack dev server URL in the Django
  nimbus-ui template to proxying the webpack webpack dev server and
  websocket in nginx

### DIFF
--- a/nginx/nginx-dev.conf
+++ b/nginx/nginx-dev.conf
@@ -1,0 +1,68 @@
+user  nginx;
+worker_processes  4;
+
+error_log  /var/log/nginx/error.log warn;
+pid        /var/run/nginx.pid;
+
+
+events {
+    worker_connections  19000;
+}
+
+
+http {
+    include       /etc/nginx/mime.types;
+    default_type  application/octet-stream;
+
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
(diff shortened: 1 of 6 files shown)


nginx/nginx-dev.conf
```diff
@@ -0,0 +1,68 @@
...
+        ssl_certificate_key key.pem;
+        client_max_body_size 20M;
+
+        location ~ ^/api/v1/experiments/(?:[\w-]+/recipe/)?$ {
+            proxy_pass http://app:7001;
+            proxy_set_header X-Forwarded-Proto $scheme;
+            proxy_set_header Host $host;
+        }
+
+        # Proxy the nimbus-ui static assets from the CRA webpack dev server
```
lmorchard, line 44: Not super happy that `nginx-dev.conf` is a copy of `nginx.conf` with just these two location blocks added. Was trying to work out a way to add this conditionally on an env var or something, but gave up and did this.

lmorchard:
Not sure why this is failing integration tests - looks like somehow nginx isn't starting up? Works locally. May just close this and go back to the drawing board if I can't manage to reproduce

lmorchard:
Yeah, I'll go back to the drawing board with this one.

jaredlockhart:
It's a good idea, especially if it can get the websocket working.  You just need separate configs for dev and prod, I think.  I wonder if there's a way to pass an env var into nginx and then selectively apply a server config based on that?

jaredlockhart:
Hrmm apparently the only way to actually use envvars in server blocks is with openresty.  That's annoying.

Issue #4099, opened 2020-12-02
Add nimbus-ui sourcemaps to Sentry


data-sync-user:
The nimbus-ui React app should supply sourcemaps to Sentry so exceptions can point to the correct lines of code where they occurred. Setup instructions [here](https://docs.sentry.io/platforms/javascript/sourcemaps/) (heads up, you'll need someone to get you a Sentry API token, and you'll need to set up env vars for stage and prod).

Also because it's a create-react-app you won't have direct access to the Webpack config, but you can inject new config in our [Rescripts config](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/.rescriptsrc.js).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-654)
┆Issue Number: EXP-654

data-sync-user:
➤ Jody Heavener commented:

Btw, marked this as High priority because it's darn near impossible ( https://sentry.prod.mozaws.net/operations/experimenter-stage/issues/10082384/?query=is%3Aunresolved ) to see what's going on in Sentry right now with the minified JS.

data-sync-user:
➤ Chelsea Lewis commented:

Added flag as this is blocked by exp-730

data-sync-user:
➤ Les Orchard commented:

Let's see if I can figure this out

data-sync-user:
➤ Les Orchard commented:

{quote}heads up, you'll need someone to get you a Sentry API token, and you'll need to set up env vars for stage and prod{quote}

Dug around today and it looks like we'll need some ops support for this part? Or, at least, access to configure Jenkins and create internal integrations on Sentry? (Neither of which I have)

The plugin itself looks easy enough to configure, but it needs an auth token and per-environment vars at build time:

  1. Need an auth token for SentryWebpackPlugin for use in uploading a source map. This sounds like an internal integration auth token ( https://docs.sentry.io/product/integrations/integration-platform/#internal-integrations ). Unless I'm missing it, I don't think I have access to see or use the UI to create this
  2. Need a SENTRY_AUTH_TOKEN env var at build (i.e. in Jenkins?) to supply the internal integration auth token to webpack
  3. Need SENTRY_ORG and SENTRY_PROJECT env vars to be accessible during build - i.e. SENTRY_ORG="operations", and then SENTRY_PROJECT="nimbus-ui-prod" and SENTRY_PROJECT="nimbus-ui-stage" respectively

[...]

PR #5374, opened 2021-06-05
issue #4099 chore(nimbus): upload sourcemaps to Sentry releases

lmorchard:
Because:

* The nimbus-ui React app should supply sourcemaps to Sentry so
  exceptions can point to the correct lines of code where they occurred.

This commit:

* Uses @sentry/webpack-plugin on nimbus-ui webpack build to upload
  sourcemaps to a release corresponding to current git commit hash

* Supplies the git commit hash of the current build to the nimbus-ui
  frontend so it can initialize Sentry reporting using it as a release

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,10 @@ google-credentials.json
 
 # Jetstream config
 app/experimenter/outcomes/jetstream-config*
+
+# Versioning assets generated on build
+app/commit-description.txt
+app/commit-summary.txt
+app/experimenter/version.json
+app/version.json
+
--- a/app/experimenter/nimbus-ui/.rescriptsrc.js
+++ b/app/experimenter/nimbus-ui/.rescriptsrc.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const SentryWebpackPlugin = require("@sentry/webpack-plugin");
 
 // Rescripts allows us modify the default CRA
 // Webpack config without needing to eject.
@@ -27,6 +28,27 @@ module.exports = [(config) => {
   ] = new MiniCssExtractPlugin({
     filename: 'static/css/[name].css',
   });
+
+  if (process.env.SENTRY_UPLOAD_SOURCEMAPS && process.env.SENTRY_AUTH_TOKEN) {
+    let versionInfo = null;
+    try {
+      versionInfo = require("../../version.json");
+    } catch (e) {
+      /* no-op */
+    }
+    if (versionInfo) {
+      config.plugins.push(
+        new SentryWebpackPlugin({
+          authToken: process.env.SENTRY_AUTH_TOKEN,
+          org: process.env.SENTRY_ORG,
+          project: process.env.SENTRY_PROJECT,
+          release: versionInfo.commit,
+          include: "./build",
+          urlPrefix: "~/static/nimbus",
+        })
+      );
+    }
+  }
 
   return config;
 }];
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true rescripts start",
-    "build": "SKIP_PREFLIGHT_CHECK=true rescripts build",
+    "build": "SKIP_PREFLIGHT_CHECK=true SENTRY_UPLOAD_SOURCEMAPS=true rescripts build",
     "test": "SKIP_PREFLIGHT_CHECK=true rescripts test",
     "test:cov": "yarn test --coverage",
     "lint": "yarn lint:eslint && yarn lint:tsc && yarn lint:styles",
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -101,3 +101,10 @@ google-credentials.json
 
 # Jetstream config
 app/experimenter/outcomes/jetstream-config*
+
+# Versioning assets generated on build
```
lmorchard, line 105: Ignoring these, since I was generating them a lot while figuring this stuff out

app/experimenter/nimbus-ui/.rescriptsrc.js
```diff
@@ -28,5 +29,26 @@ module.exports = [(config) => {
...
+          org: process.env.SENTRY_ORG,
+          project: process.env.SENTRY_PROJECT,
+          release: versionInfo.commit,
+          include: "./build",
+          urlPrefix: "~/static/nimbus",
+        })
+      );
+    }
+  }
+
```
lmorchard, line 52: This code should end up doing nothing until we manage to get these env vars set in Jenkins (or CircleCI?) for build:

* `SENTRY_ORG=operations`
* `SENTRY_PROJECT=nimbus-ui-prod` / `SENTRY_PROJECT=nimbus-ui-stage` (respectively)
* `SENTRY_AUTH_TOKEN=XXXX`
  * this value will need to be an auth token from [a new internal integration created within Developer Settings found under Sentry organizational settings](https://docs.sentry.io/product/integrations/integration-platform/#internal-integrations)

app/experimenter/nimbus-ui/package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true rescripts start",
-    "build": "SKIP_PREFLIGHT_CHECK=true rescripts build",
+    "build": "SKIP_PREFLIGHT_CHECK=true SENTRY_UPLOAD_SOURCEMAPS=true rescripts build",
```
lmorchard, line 8: This ensures that the sourcemap upload is skipped for local dev

app/experimenter/settings.py
```diff
@@ -19,6 +20,13 @@
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 
+APP_VERSION = ""
+try:
+    version_json = json.load(open(os.path.join(BASE_DIR, "version.json")))
+    APP_VERSION = version_json["commit"]
+except:
+    # No-op, because version.json is usually only generated for real builds
+    pass
```
lmorchard, line 29: Kind of ugly, tried to work out how to handle the absence of version.json in local dev but when it's available in build.

lmorchard (commented):
Some comments for self-review. 

Main gist here is threading through the current build commit to the frontend so that reported exceptions get mapped to a release on the Sentry side with the approriate uploaded sourcemap artifacts.

lmorchard:
Oh, I see the failure - I need to write a test to cover the new `APP_VERSION` in settings. I'll get to that on Monday

lmorchard:
Coming back to this after the weekend, I'm thinking probably the sourcemap upload to Sentry shouldn't happen at build time in the Webpack plugin but as a separate Makefile task done as part of deploy.

If we build a single image that we deploy to stage & prod, then Sentry sourcemap upload should happen seperately with sentry-cli based on env vars for each environment

jaredlockhart:
I would think we would want to trigger this in circle? Like somewhere around here:

https://github.com/mozilla/experimenter/blob/main/.circleci/config.yml#L116-L120

Maybe as part of the deploy task or as a separate circle task that happens at the end of the flow.  We'd only want it to happen on `main`, not pr branches.  And then we'd put those sentry env vars into the circle config to access them.

jaredlockhart:
It's really too bad we can't get around this with the thing where it just discovers the source maps.  I don't know why that's not working, it seems like it should.

The files declare their source map at the bottom:

https://experimenter.services.mozilla.com/static/nimbus/static/js/main.js

```
//# sourceMappingURL=main.js.map
```

which is accessible:

https://experimenter.services.mozilla.com/static/nimbus/static/js/main.js.map

OH I KNOW WHY MAYBE??

Is it becuase the source maps are behind the auth gate so sentry can't fetch it??

Oh that might be it.  Maybe if we allowlist the static assets subpaths, sentry will be able to discover them and then we don't have to do the upload at all?  Maybe worth trying?

lmorchard:
Yeah, going to pull this one down too to rework - its need to be triggered in CircleCI to deploy the sourcemaps to Sentry after the Docker images are built.

lmorchard:
> Is it becuase the source maps are behind the auth gate so sentry can't fetch it??

Yes, the [sourcemaps need to be uploaded because they are otherwise inaccessible to Sentry](https://github.com/mozilla/experimenter/issues/4099#issuecomment-852573740):

> Oh that might be it. Maybe if we allowlist the static assets subpaths, sentry will be able to discover them and then we don't have to do the upload at all? Maybe worth trying?

@jaredlockhart Maybe worth trying? I don't know how that's set up. Can you try that or know who can?

lmorchard:
> Maybe as part of the deploy task or as a separate circle task that happens at the end of the flow. 

Yup, [that's what I meant here](https://github.com/mozilla/experimenter/pull/5374#issuecomment-856057257).

PR #5389, opened 2021-06-08
Initial commit adding e2e live test.

jrbenny35:
This adds an e2e test that approves an experiment on remote settings and checks its status on the experiments Summary page.


Makefile
```diff
@@ -157,4 +157,4 @@ integration_test_legacy: integration_build
 	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-legacy $(TOX_ARGS) -- -n 4 $(PYTEST_ARGS)"
 
 integration_test_nimbus: integration_build
-	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-nimbus $(TOX_ARGS) -- -n 4 $(PYTEST_ARGS)"
+	$(COMPOSE_INTEGRATION) up -d; MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod -R a+rwx /code/app/tests/integration/;tox -c app/tests/integration -e integration-test-nimbus $(TOX_ARGS) -- --html=report.html --self-contained-html  $(PYTEST_ARGS)"
```
jaredlockhart, line 160: Oh I don't think you should add this here?  I kept starting the environment and running the tests separate on purpose.  Does this have to be here?

.env.sample
```diff
@@ -1,7 +1,7 @@
 BUGZILLA_API_KEY=
 BUGZILLA_CC_LIST=
 BUGZILLA_HOST=https://bugzilla.allizom.org/
-CELERY_SCHEDULE_INTERVAL=300
+CELERY_SCHEDULE_INTERVAL=30
```
jaredlockhart, line 4: Probably don't need to change this here if you're using a separate one for integration tests?

jaredlockhart (commented):
Do all those file permission changes need to be in here?

Makefile
jrbenny35: No, it shouldn't my bad!

.env.sample
jrbenny35: TRUE

jrbenny35:
> 
> 
> Do all those file permission changes need to be in here?

No, I think I will remove the pytest-html report and add that to another PR and fix the permissions then.

app/tests/integration/nimbus/pages/review.py
```diff
@@ -1,8 +1,52 @@
...
 class ReviewPage(Base):
     """Experiment Review Page."""
 
+    _root_locator = (By.CSS_SELECTOR, "#PageRequestReview")
+    _approve_request_button_locator = (By.CSS_SELECTOR, "#approve-request-button")
+    _experiment_status_icon_locator = (
+        By.CSS_SELECTOR,
+        ".header-experiment-status .boarder-primary",
+    )
     _page_wait_locator = (By.CSS_SELECTOR, "#PageRequestReview")
```
pdehaan, line 16: Should we just recycle `_root_locator` above (new L10), or does this serve a different purpose?

app/tests/integration/nimbus/remote_settings/pages/login.py
```diff
@@ -0,0 +1,34 @@
...
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+
+
+class Login(Page):
+
+    _root_locator = (By.CSS_SELECTOR, ".container-fluid")
+    _auth_type_checkbox_locator = (
+        By.CSS_SELECTOR,
+        "div.radio:nth-child(2) label span input",
```
pdehaan, line 11: `:nth-child()` scares me, but my guts tell me this might be the RS login page which is probably not worth trying to do upstream PRs to add better selectors (unless we need to automate remote settings for multiple projects).

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -42,3 +47,84 @@ def test_create_new_experiment(selenium, base_url):
...
+    selenium.get(base_url)
+    # refresh until the experiment shows up
+    for attempt in range(45):
+        try:
+            home = HomePage(selenium, base_url).wait_for_page_to_load()
+            new_experiments = len(home.tables[0].experiments)
+            assert new_experiments != current_experiments
+        except AssertionError:
+            time.sleep(2)
+            selenium.refresh()
```
pdehaan, line 116: I was going to say this looks risky, but from the looks of L108-109, it seems like we'll attempt to see if the experiment is live 45 times before failing, with a 2 second pause between checks. Seems fair (if I understand correctly). I guess worst case is that we fail and the test takes ~90s to sleep intermittently. Best case we find it on the first or second iteration and break the loop.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -42,3 +47,84 @@ def test_create_new_experiment(selenium, base_url):
...
+        else:
+            print("live")
+            break
+    # Check it's live
+    home = HomePage(selenium, base_url).wait_for_page_to_load()
+    live_experiments = home.tables[0]
+    assert "live experiments" in live_experiments.table_name.lower()
+    for item in live_experiments.experiments:
+        if experiment_name in item.text:
+            print(item.text)
```
pdehaan, line 126: Is this for local debugging, or do we need to keep these for CI troubleshooting reasons? (this and L118 above, which prints "live" to STDOUT).

pdehaan (commented):
Looking good.
Not sure about the chmod changes. Looks like we have about 41 instances of chmod 100644 → 100755 (which is odd, since I expected 777 from my interpretation of the diff), and 29 instances of `/legacy/` in this PR (I think all chmod changes).

app/experimenter/nimbus-ui/src/components/PageRequestReview/FormLaunchConfirmationCheckboxes.tsx
```diff
@@ -28,6 +28,7 @@ const FormLaunchConfirmationCheckboxes = ({
       {checkboxLabels.map((labelText, index) => (
         <Form.Group
           key={index}
+          id="checkbox"
```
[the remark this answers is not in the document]
jrbenny35: since its a loop it adds a number after and that is #goodenough for this in my opinion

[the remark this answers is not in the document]
pdehaan: Oh, if it's adding a number, I wonder if this line is ignored and we're talking about the `controlId` below on L33 (I'd have to dig into the docs for this `Form.Group` and `controlId` stuff in React). I'll see if I can reproduce on stage though and tell what the current values are.

app/tests/integration/nimbus/pages/home.py
```diff
@@ -16,9 +16,31 @@ def wait_for_page_to_load(self):
...
     def create_new_button(self):
         el = self.find_element(*self._page_wait_locator)
         el.click()
         from nimbus.pages.new_experiment import NewExperiment
 
         return NewExperiment(self.driver, self.base_url).wait_for_page_to_load()
+
+    class Tables(Region):
+
+        _table_name_locator = (By.CSS_SELECTOR, ".font-weight-bold")
```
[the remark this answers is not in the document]
jrbenny35: I could look into this, but that class is only on the 1st element if you look at this function https://github.com/mozilla/experimenter/blob/768018f57c810a6bcfc6e48a65dffdaf5efb2510/app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx#L97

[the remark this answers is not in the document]
pdehaan: Good point. I think the closest I came in browser DevTools was

```js
$$(".directory-table thead th:first-child")
```

Which returns 5 results (for "Live Experiments", "In Review", "In Preview", "Completed", and "Drafts" `<th>` children).

app/tests/integration/nimbus/remote_settings/pages/login.py
jrbenny35: Yeah, I may do that though in the future.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Yeah so this is a weird one. I need to do a ```refresh()``` and selenium seems to act weird when I do that within a ```WebdriverWait``` function.

Makefile
```diff
@@ -157,4 +157,4 @@ integration_test_legacy: integration_build
 	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-legacy $(TOX_ARGS) -- -n 4 $(PYTEST_ARGS)"
 
 integration_test_nimbus: integration_build
-	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-nimbus $(TOX_ARGS) -- -n 4 $(PYTEST_ARGS)"
+	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run firefox sh -c "sudo chmod -R a+rwx /code/app/tests/integration/;tox -c app/tests/integration -e integration-test-nimbus $(TOX_ARGS) -- $(PYTEST_ARGS)"
```
pdehaan, line 160: Wouldn't running <kbd>make integration_test_nimbus</kbd> locally chmod a bunch of app/tests/integration/** files that Git would want to commit?

jrbenny35: good catch, I just need to remote the recursive flag

pdehaan (approved):
🎉  Great job, @jrbenny35!

jaredlockhart (approved):
Now we're cooking with fire!! 🔥 🔥 🔥 🔥 🔥 

Great job @jrbenny35 !!!

Issue #4945, opened 2021-03-29
Timeline shows wrong durations


jaredlockhart:
In the Nimbus UI, the timeline shows the projected durations and not the actually occurring durations.  Ie I just launched and ended this on the same day but still shows 28/7 days:

<img width="1156" alt="Screen Shot 2021-03-29 at 2 45 35 PM" src="https://user-images.githubusercontent.com/119884/112884504-93457500-909d-11eb-9914-a7b28d0b7b3c.png">

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1146)
┆Issue Number: EXP-1146

jodyheavener:
Huh, yeah makes sense. We could use the `computedEndDate`-`startDate` for the total duration, but enrollment might be a little wonky to calculate. Maybe the lesser of `computedEndDate`-`startDate` _or_ proposed duration 🤔

jaredlockhart:
No we can figure out the real enrollment period by looking at the changelogs.  We should just add some computed properties for those to the model that use the real values from teh changelogs if they exist or fallback to teh proposed ones, expose those in V5 and change the timelie UI to always use the computed values and then they'll be the proposed ones until teh real values start filling in and then update automagically.

PR #5399, opened 2021-06-09
chore(nimbus): spread factory lifecycle changelog events out over a random timespan

lmorchard:
Because:

* Life cycle events in experiments created via load_dummy_experiments
  appear to happen all within the same minute

This commit:

* Spreads the life cycle events out over a much longer random timespan
  similar to legacy experimenter

### DIFF
--- a/app/experimenter/experiments/tests/factories/nimbus.py
+++ b/app/experimenter/experiments/tests/factories/nimbus.py
@@ -1,3 +1,4 @@
+import datetime
 import decimal
 import json
 import random
@@ -5,6 +6,7 @@ from collections.abc import Iterable
 from enum import Enum
 
 import factory
+from django.utils import timezone
 from django.utils.text import slugify
 from faker import Factory as FakerFactory
 
@@ -246,8 +248,9 @@ class NimbusExperimentFactory(factory.django.DjangoModelFactory):
             self.countries.add(*extracted)
 
     @classmethod
-    def create_with_lifecycle(cls, lifecycle, **kwargs):
+    def create_with_lifecycle(cls, lifecycle, with_random_timespan=False, **kwargs):
         experiment = cls.create(**kwargs)
+        now = timezone.now() - datetime.timedelta(days=random.randint(100, 200))
 
         for state in lifecycle.value:
             experiment.apply_lifecycle_state(state)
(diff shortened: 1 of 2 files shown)


lmorchard:
@lzoog This might help things look more realistic for #4945

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -256,11 +259,15 @@ def create_with_lifecycle(cls, lifecycle, **kwargs):
...
-            generate_nimbus_changelog(
+            change = generate_nimbus_changelog(
                 experiment,
                 experiment.owner,
                 f"set lifecycle {lifecycle} state {state}",
             )
+            if with_random_timespan:
+                change.changed_on = now
+                change.save()
+                now += datetime.timedelta(days=random.randint(5, 20))
```
lmorchard, line 270: This is slightly hacky. But, I had an earlier attempt that modified `generate_nimbus_changelog` to add an optional `changed_on` parameter that seemed to add more complexity than necessary IMO

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -246,8 +248,9 @@ def countries(self, create, extracted, **kwargs):
             self.countries.add(*extracted)
 
     @classmethod
-    def create_with_lifecycle(cls, lifecycle, **kwargs):
+    def create_with_lifecycle(cls, lifecycle, with_random_timespan=False, **kwargs):
```
lmorchard, line 251: Oh yeah, and I made it an optional switch here because the random timing threw a lot of tests off

app/experimenter/experiments/tests/factories/nimbus.py
jaredlockhart: Oh oh! Thank you! I've actually been craving something like this lately, so thank you for hopping in and doing it.  This is a totally fine way to do it, but one thing I've been wanting lately is the ability to explicitly set like 'make this this exact many days in the past' and I was thinking about sane ways to do it, one is you could pass in a negative delta paramter, but that's a bit meh, but what about passing in a start date datetime object, which would basically take the place of your 'now' initializer, so if you don't pass it in, it just starts at now now, and if you do, it starts in the past (or future if you needed it to I guess).  And then I think combined with this you could basically cover all the cases I've been thinking of.  What do you think?

jaredlockhart (approved):
Yeah totally, super helpful, thank you.  Totally good to land it like this, or we could think about adding some kind of start date param now, or we could just do that later, I'm good with either.  👍

lmorchard:
> we could think about adding some kind of start date param 

Do you mean a start date for `create_with_lifecycle` or for `generate_nimbus_changelog`? Because the start date for `create_with_lifecycle` would get random days added to it, so that's not exact beyond the first change. Might be worth an issue with some more details / use cases

lmorchard:
I'll just merge so this can be used for #4945 sooner than later. We can further iterate and clarify

PR #5402, opened 2021-06-09
fixes #4945 feat(nimbus): Add/consume computedEnrollmentDays and computedDuration

### LINKED ISSUE #4945: Timeline shows wrong durations

LZoog:
closes #4945 

Because:
* The timeline showed proposed duration/days of enrollment and not the actual duration/days of enrollment

This commit:
* Checks the experiment changelogs for accurate duration/enrollment and expose them with fallbacks
* Updates SummaryTimeline to use the new properties

Co-authored-by: Les Orchard <me@lmorchard.com>


app/experimenter/experiments/models/nimbus.py
```diff
@@ -268,6 +268,26 @@ def proposed_end_date(self):
...
+    @property
+    def computed_duration(self):
+        if self.start_date:
+            return (self.computed_end_date - self.start_date).days
+        else:
+            return self.proposed_duration
+
     @property
     def computed_end_date(self):
         return self.end_date or self.proposed_end_date
```
LZoog, line 293: @lmorchard I reverted this back to what we had before because it seems to be fine? Not sure what hackery we ran into yesterday but I checked a bunch of completed experiments locally and things look as expected to me.

app/experimenter/experiments/tests/factories/nimbus.py
```diff
@@ -110,9 +110,7 @@ class Lifecycles(Enum):
     LIVE_ENROLLING = LAUNCH_APPROVE_APPROVE + (LifecycleStates.LIVE_IDLE_ENROLLING,)
     LIVE_ENROLLING_WAITING = LIVE_ENROLLING + (LifecycleStates.LIVE_WAITING_ENROLLING,)
     LIVE_PAUSED = LIVE_ENROLLING + (LifecycleStates.LIVE_IDLE_PAUSED,)
-    ENDING_REVIEW_REQUESTED = LAUNCH_APPROVE_APPROVE + (
-        LifecycleStates.LIVE_REVIEW_ENDING,
-    )
+    ENDING_REVIEW_REQUESTED = LIVE_PAUSED + (LifecycleStates.LIVE_REVIEW_ENDING,)
```
LZoog, line 113: @jaredlockhart I'd like to double check this is right with you. It seems like when an experiment is ending,`sEnrollmentPaused` would be set to `true`.

app/experimenter/experiments/models/nimbus.py
lmorchard: Ah, yeah, this is probably fine. A proper if/else might be clearer, but if it ain't broke don't poke it

jaredlockhart: Oh yeah python short circuiting, it'll return the first truthy value.  I tend to use it for very simple cases like this, you can technically build complex logical expressions and get values out of it but that's usually better written as if/else, but for simple things like this I just like to use the short circuited version

app/experimenter/experiments/tests/factories/nimbus.py
jaredlockhart: It may or may not, you can end an experiment before you pause it.

jaredlockhart: But this is fine either way.

jaredlockhart (changes requested):
Generally looks good but looks like something went sideways:

Screen Shot 2021-06-10 at 12 35 19 PM

jaredlockhart:
Looks like a bug in the existing implementation of `start_date` `computed_end_date`.

```py
In [4]: experiment.computed_duration
TypeError                                 Traceback (most recent call last)
<ipython-input-4-d21d9d247f4c> in <module>
----> 1 experiment.computed_duration

/app/experimenter/experiments/models/nimbus.py in computed_duration(self)
    285     def computed_duration(self):
    286         if self.start_date:
--> 287             return (self.computed_end_date - self.start_date).days
    288         else:
    289             return self.proposed_duration

TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.datetime'

In [5]: experiment.computed_end_date
Out[5]: datetime.date(2021, 7, 8)

In [6]: experiment.start_date
Out[6]: datetime.datetime(2021, 6, 10, 16, 34, 35, 580747, tzinfo=<UTC>)
```

How did our tests and your new tests miss this I wonder?

jaredlockhart:
Oh that's why our tests didn't catch it, becuase we don't HAVE tests for `computed_end_date`.  Looks like we gotta clean a bit of this up.

jaredlockhart:
@LZoog did you test this locally?

jaredlockhart:
We should probably standardize all these date methods on using `datetime.date` instead of `datetime.datetime`

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -352,6 +352,54 @@ def test_should_end_returns_True_after_proposed_end_date(self):
...
+    def test_computed_duration_returns_fallback(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+        )
+
+        self.assertEqual(
+            experiment.computed_duration,
+            experiment.proposed_duration,
+        )
+
```
jaredlockhart, line 402: Oh I see we're missing a test that hits the case where the experiment launched and has a real start date, but hasn't ended so the end date is computed.

LZoog:
Ohh, @jaredlockhart what you're seeing on https://github.com/mozilla/experimenter/pull/5402#issuecomment-858775717 is the "hackery" I was referring to with my comment here: https://github.com/mozilla/experimenter/pull/5402#discussion_r648733793

We were seeing a similar error and then changed that line to something like ` datetime.combine()` with `+ time()` to do some sort of conversion there, but that caused the `test_experiment_computed_end_date_proposed` test to fail. Poked at it further and reverted the change we made because it seemed to be working locally for me but clearly there's still a problem, thanks for catching it 👀

jaredlockhart:
Aha!  @LZoog if you rebase this on main with @jrbenny35 's new test it catches that case!

```
nimbus/test_e2e_create_experiment.py::test_create_new_experiment PASSED                                                                                                       [ 50%]
nimbus/test_e2e_create_experiment.py::test_create_new_experiment_remote_settings FAILED                                                                                       [100%]
```

That's so awesome 🎉 🎉 🎉 🎉

LZoog:
Not sure this is going to make the sprint, got tied up in some other things today and I can't seem to reproduce the error locally but I do see the failure in the integration tests after rebasing.

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
LZoog: I went ahead and added `computed_end_date` tests to the model.

LZoog:
@jaredlockhart I think this is ready for another look. We do have a little awkwardness around `datetime` and `time` types on the nimbus model, filed #5446 for that.

app/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -41,7 +41,7 @@ def test_serializer_outputs_expected_schema_with_feature(self):
                 # DRF manually replaces the isoformat suffix so we have to do the same
                 "endDate": experiment.end_date.isoformat().replace("+00:00", "Z"),
                 "id": experiment.slug,
-                "isEnrollmentPaused": False,
+                "isEnrollmentPaused": True,
```
lmorchard, line 44: ~Hmm, trying to remember why this flipped from False to True.~ Oh! I think I remember: It's because we made the change to factory-generated experiments to include pausing enrollment as part of the canned lifecycles.

LZoog: Right, maybe I shouldn't have resolved [this comment](https://github.com/mozilla/experimenter/pull/5402/files#r649338991) as it would have been more obvious. Guess this wasn't really necessary but 🤷‍♀️

lmorchard:
Took another skim through this again myself and things make sense to me. Best to defer to Jared's set of eyes, though, since I also helped write some of this

jodyheavener:
I know I helped work on this but I can't not see the two different durations - the one under the timeline and the one in the top-right. The one in the top-right is doing the math in the browser using `stringDateSubtract`, but now that we have the duration from the server maybe we can use it there as well (or remove that one)?

Screen Shot 2021-06-17 at 4 21 29 PM

jodyheavener (approved):
Other than the two duration discrepancies I think this looks good and works for me in testing.

PR #5405, opened 2021-06-10
fixes #5404 fix(reporting): ReportLog timestamp to changelog time

### LINKED ISSUE #5404: [Reporting] ReportLog model timestamp value should not be time when created
Currently the timestamp value for reportlogs is the time when the reportlog object is created not the time of the corresponding changelog.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1414)
┆Issue Number: EXP-1414

tiftran:
This commit:
* changes the model field timestamp to not autogenerate
* timestamp field is now mandatory for reportlogs

Because:
* we don't want to know when the reportlog was created rather the time of the corresponding changelog


app/experimenter/reporting/tests/factories.py
```diff
@@ -12,6 +13,7 @@
 
 
 class ReportLogFactory(ReportLogConstants, factory.django.DjangoModelFactory):
```
jaredlockhart, line 15: Django's been hassling me about using `timezone.now()` instead of `datetime.now()`, should probably use that so it doens't give you grief later

app/experimenter/reporting/tests/test_models.py
```diff
@@ -19,6 +21,7 @@ def test_reportlog_saves_correctly(self):
     def test_reportlog_saves_with_null_old_status(self):
         self.assertEqual(ReportLog.objects.count(), 0)
         ReportLog.objects.create(
+            timestamp=datetime.now(),
```
jaredlockhart, line 24: same here

app/experimenter/reporting/tests/test_models.py
```diff
@@ -27,3 +30,16 @@ def test_reportlog_saves_with_null_old_status(self):
...
+        ReportLog.objects.create(
+            timestamp=yesterday,
+            experiment_slug="experiment-slug",
+            experiment_name="experiment name",
+            experiment_type="Nimbus IOS",
+            experiment_new_status="Draft",
+            event="Create",
+            event_reason="New",
+        )
+        self.assertEqual(ReportLog.objects.filter(timestamp=yesterday).count(), 1)
```
jaredlockhart, line 45: I don't think you really need to test this 🤷‍♀️

jaredlockhart (approved):
👍

PR #5416, opened 2021-06-10
fix(nimbus): Fix nimbus-ui compile warning

robhudson:


LZoog (approved):
Thanks for the quick fix! :D

PR #5429, opened 2021-06-14
fixes #5105 feat(nimbus): use letters when adding new treatment branches

### LINKED ISSUE #5105: New treatment branch names should default to letters instead of numbers
Right now if you click "New branch," the name is "treatment 2", "treatment 3," etc. Since it's more typical for branches to be saved as "treatment A," "treatment B" etc., we should change these to letters.

[todo: confirming with Ana if the first treatment branch should be "treatment" or "treatment A" since the next new branch will likely be "treatment B"]

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1232)
┆Issue Number: EXP-1232

jodyheavener:
Closes #5105

This PR:

- Updates the default branches to use capitalized titles
- Changes the generation of treatment branches to increment in letters instead of numbers

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
+++ b/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
@@ -57,7 +57,11 @@ function addBranch(state: FormBranchesState) {
   } else {
     treatmentBranches = [
       ...(treatmentBranches || []),
-      createAnnotatedBranch(state.lastId, `treatment ${lastId}`),
+      createAnnotatedBranch(
+        state.lastId,
+        // 65 is A, but we generate a control, so start at 64
+        `Treatment ${String.fromCharCode(64 + lastId)}`,
+      ),
     ];
   }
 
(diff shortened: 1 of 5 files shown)


app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
```diff
@@ -57,7 +57,11 @@ function addBranch(state: FormBranchesState) {
   } else {
     treatmentBranches = [
       ...(treatmentBranches || []),
-      createAnnotatedBranch(state.lastId, `treatment ${lastId}`),
+      createAnnotatedBranch(
+        state.lastId,
+        // 65 is A, but we generate a control, so start at 64
+        `Treatment ${String.fromCharCode(64 + lastId)}`,
```
lmorchard, line 63: Things will get weird if we ever go over 26 branches.

(but that's a weird number of branches)

Capture

lmorchard (approved):
LGTM, only comment was on an edge case we'll most probably never ever see

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
jodyheavener: Ehh, I'd be willing to revisit if we're regularly going past 26 branches. Not super concerned at this point!

PR #5450, opened 2021-06-16
fixes #5041 feat(nimbus): Use more specific remote settings review url

### LINKED ISSUE #5041: kinto_admin_url should link to the exact /simple-review path on remote settings
Let's have kinto_admin_url return the exact remote settings review URL.

Let's start with on **staging only right now** (the new UI hasn't shipped in remote settings prod yet)*,* let's link to the new UI. This should be

settings-writer.stage.mozaws.net/v1/admin/#/buckets/main-workspace/collections/{collection\}/simple-review

Where {collection\} is nimbus-desktop-experiments for desktop and nimbus-mobile-experiments for ios/android.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/EXP-1202)
┆Issue Number: EXP-1202

k88hudson:
Because

* We have a new UI for reviewing we'd like to try on staging

This commit

* Adds a review_url property to the nimbus model
* Replaces kinto_admin_url usages in the front end to use the review_url
* Sets the review_url path to /{collection}/simple-review for staging
* Sets the review+url path to /{collection}/records for prod

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -39,7 +39,7 @@ APPLICATION_CONFIG_DESKTOP = ApplicationConfig(
         Channel.RELEASE: "firefox-desktop",
     },
     default_app_id="firefox-desktop",
-    rs_experiments_collection="nimbus-desktop-experiments",
+    rs_experiments_collection=settings.KINTO_COLLECTION_NIMBUS_DESKTOP,
     randomization_unit=BucketRandomizationUnit.NORMANDY,
 )
 
@@ -53,7 +53,7 @@ APPLICATION_CONFIG_FENIX = ApplicationConfig(
         Channel.RELEASE: "org.mozilla.firefox",
     },
     default_app_id="",
-    rs_experiments_collection="nimbus-mobile-experiments",
+    rs_experiments_collection=settings.KINTO_COLLECTION_NIMBUS_MOBILE,
     randomization_unit=BucketRandomizationUnit.NIMBUS,
 )
 
@@ -67,7 +67,7 @@ APPLICATION_CONFIG_IOS = ApplicationConfig(
         Channel.RELEASE: "org.mozilla.ios.Firefox",
     },
     default_app_id="",
-    rs_experiments_collection="nimbus-mobile-experiments",
+    rs_experiments_collection=settings.KINTO_COLLECTION_NIMBUS_MOBILE,
     randomization_unit=BucketRandomizationUnit.NIMBUS,
 )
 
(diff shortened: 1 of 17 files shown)


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -64,6 +67,7 @@ class ApplicationConfig:
         Channel.RELEASE: "org.mozilla.ios.Firefox",
     },
     default_app_id="",
+    rs_experiments_collection="nimbus-mobile-experiments",
```
jaredlockhart, line 70: We have the collection name constants in settings:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L416-L418

which is already imported in this file:

https://github.com/mozilla/experimenter/blob/d0aef25a9a261276098d5225c7d317a3b330295e/app/experimenter/experiments/constants/nimbus.py#L4

app/experimenter/experiments/models/nimbus.py
```diff
@@ -300,6 +300,18 @@ def to_timestamp(date):
             slug=self.slug, from_date=start_date, to_date=end_date
         )
 
+    @property
+    def review_url(self):
+        is_staging = "settings-writer.stage" in settings.KINTO_ADMIN_URL
```
jaredlockhart, line 305: This should definitely be in settings.py as a global constant

app/experimenter/experiments/models/nimbus.py
```diff
@@ -300,6 +300,18 @@ def to_timestamp(date):
...
+
+        return "{base_url}/{collection_path}/{collection}/{review_path}".format(
+            base_url=settings.KINTO_ADMIN_URL,
+            collection_path="#/buckets/main-workspace/collections",
+            collection=self.application_config.rs_experiments_collection,
+            # TODO: Update when simple-review is on prod in remote settings
```
jaredlockhart, line 311: I'd file this as an actual ticket and put the issue ID right here in the comment

jaredlockhart (changes requested):
Thnx for hopping on this @k88hudson !  This new flow is gonna be so faaaaannnccccyyyy.....

Also we should figure out how/when that makes its way into our local RS docker image and flag it on for local as well as stage so we can test/use it locally, but we can do that as a followup if you don't wanna dig into it.  If you don't wanna do it here pls file that also as a followup.

app/experimenter/experiments/constants/nimbus.py
k88hudson: yeah but you want it to be associated with the application, no?

k88hudson: or do you mean just like `rs_experiments_collection=settings.whatever`

app/experimenter/experiments/models/nimbus.py
k88hudson: as in you want a global `IS_STAGING` constant?

jaredlockhart (approved):
Awesome looks good! 👍

PR #5816, opened 2021-06-21
Fix #5813: fix(visualization): Pull in Jetstream data for complete experiments with null data.

### LINKED ISSUE #5813: Completed experiments should pull in jetstream results if results_data is null
This is a follow up to https://github.com/mozilla/experimenter/commit/60de609806100c9f71a21690e4dc6df8815e17c9
The should look up jetstream data if `results_data` is null for old experiments so that they are stored once but later not updated

emtwo:


emtwo:
For some extra context, in this PR: https://github.com/mozilla/experimenter/commit/60de609806100c9f71a21690e4dc6df8815e17c9 I added caching using a celery task to pull in jetstream data periodically. The data was not pulled in for experiments that finished 3 or more days ago. This ended up meaning that all old experiments no longer had their data available on the results page.

To address this, experiments that have their field `results_data` set as `None` should be updated so that even old experiments can have data available.

app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -181,13 +181,36 @@ def test_data_fetch_in_loop(self, lifecycle, mock_delay):
...
             new_status=NimbusExperiment.Status.COMPLETE,
         ).update(changed_on=datetime.date.today() - datetime.timedelta(days=4))
         tasks.fetch_jetstream_data()
         mock_delay.assert_not_called()
 
+    @parameterized.expand(
+        [
+            (NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE,),
+        ]
+    )
```
jodyheavener, line 203: Naive question.. is it worth writing this in the parameterized format if we're only testing one lifecycle? Or do we plan to add more later?

jodyheavener (approved):
From my understanding of what's going on I think this makes sense 👍

app/experimenter/jetstream/tests/test_tasks.py
emtwo: Fair question, I think it's not necessary but it does make it easier to add more later.

PR #5819, opened 2021-06-21
e2e nimbus remote settings deny test

jrbenny35:


app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,95 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
+            selenium.refresh()
+        else:
+            home.tabs[-1].click()
+            draft_experiments = home.tables[0]
+            for item in draft_experiments.experiments:
+                if experiment_name in item.text:
+                    item.click()
+                    break
+        finally:
+            break
```
pdehaan, line 213: Q: What does this do?
What happens if it fails all 45 attempts?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,95 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
+    experiment_url = experiment_name.replace(" ", "-")
+    print(experiment_url)
+    selenium.get(f"{base_url}/{experiment_url}/request-review")
+    for attempt in range(30):
+        try:
+            summary_page = ReviewPage(selenium, base_url).wait_for_page_to_load()
+            assert summary_page.rejected_text, "Rejected text box did not load"
```
pdehaan, line 220: Sorry, I'm not good w/ Python...

If we have a test, and the `try` succeeds and our assert passes and no exception is thrown on the first attempt... is that it? Or will it keep doing the `for..in` loop another 29 times? Do we need to break out of the loop somehow on success? Or am I missing something?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,95 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
+        except AssertionError:
+            time.sleep(2)
+            selenium.refresh()
+        else:
+            home.tabs[-1].click()
+            draft_experiments = home.tables[0]
+            for item in draft_experiments.experiments:
+                if experiment_name in item.text:
+                    item.click()
+                    break
```
pdehaan, line 211: Just to confirm I understand this correctly&hellip;

So, we loop <45 times, and if the home page loads, we click on the last tab, get the length of the first table's experiments, and assert that the numbers are different (so an experiment was added). If our assertion fails, we fall into the `except` block, sleep for 2 seconds, and refresh the browser (1). Now, in the `else` flow, we click on the last tab (drafts) and assert that our experiment name is visible (and if so, we break out of the loop).

(1) Does the `finally` block get called after `except` is done? If so, what does the `break` statement on L213 do? Cause it to try again in the `for attempt in range(45)` loop? Or try once and exit the loop? Maybe I'm getting confused as to whether we're trying to break out of a loop, or `try..except` block

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,95 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
+        finally:
+            break
+    experiment_url = experiment_name.replace(" ", "-")
```
pdehaan, line 214: I wonder if we can use the built-in Nimbus sluggifier, but it might not matter since our experiment names don't have mix of uppercase or lowercase.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,95 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
+    audience.expected_clients = 50
+    audience.save_btn()
+    review = audience.save_and_continue()
+
+    # Review and approve
+    selenium.find_element_by_css_selector("#PageRequestReview")
+    review.launch_without_preview.click()
+    review.request_review.click_launch_checkboxes()
+    review.request_review.request_launch_button.click()
+    review.approve()
```
pdehaan, line 178: Q: Should this be `.action()`?

LZoog:
Just a heads up, I have a draft PR for combining the summary page with request review up (#5821), but, there's a few `needs-product` questions and Ana isn't available until Wednesday so that probably won't be merged until Thursday-ish, and maybe I can figure out the needed changes fairly easily. 👀

app/tests/integration/nimbus/pages/home.py
```diff
@@ -6,13 +6,14 @@
 class HomePage(Page):
     """Nimbus Home page."""
 
-    _page_wait_locator = (By.CSS_SELECTOR, "#create-new-button")
+    _create_new_btn_locator = (By.CSS_SELECTOR, "#create-new-button")
+    _page_wait_locator = (By.CSS_SELECTOR, ".table")
```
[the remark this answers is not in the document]
jrbenny35: I can do that

app/tests/integration/nimbus/remote_settings/pages/dashboard.py
```diff
@@ -67,5 +73,20 @@ def _wait_for_it(selenium):
...
             el.click()
+
+    class Modal(Region):
+        _decline_changes_locator = (By.CSS_SELECTOR, ".modal-dialog .btn-primary")
+
+        def decline_changes(self):
+            self.selenium.execute_script(
+                """
+                document.activeElement.blur()
+                document.querySelector('.modal-dialog .btn-primary').click()
```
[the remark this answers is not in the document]
jrbenny35: I'll just remove that line

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,95 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
     summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
     assert "live" in summary_page.experiment_status.lower()
+
+
+def test_create_new_experiment_remote_settings_reject(selenium, base_url):
+    experiment_name = f"name here remote {random.randint(0, 1000)}"
+
+    selenium.get(base_url)
+    home = HomePage(selenium, base_url).wait_for_page_to_load()
+    home.tabs[-1].click()  # Click drafts
```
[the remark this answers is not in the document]
jrbenny35: Which line exactly?

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Good catch, I meant to throw an error

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Another good catch, I should exit the loop using an else clause.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: I was thinking this too but this is just a workaround until the link loads the correct page. I think @LZoog is working on that.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Action is a function for remote settings.

app/tests/integration/nimbus/test_e2e_create_experiment.py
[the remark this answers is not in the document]
pdehaan: L139 `home.tabs[-1].click()` (see also L199, and L206 below).
I mean, it's a super nit and possibly very much over-optimizing so feel free to ignore and we can DRY it out in the future if it ever comes up.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,3 +129,100 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
+    experiment_url = experiment_name.replace(" ", "-")
+    selenium.get(f"{base_url}/{experiment_url}/request-review")
+    for attempt in range(30):
+        try:
+            summary_page = ReviewPage(selenium, base_url).wait_for_page_to_load()
+            assert summary_page.rejected_text, "Rejected text box did not load"
+        except NoSuchElementException:
+            time.sleep(2)
+            selenium.refresh()
+        if attempt == 30:
```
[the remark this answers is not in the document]
pdehaan: OK, some lessons learned:

```python
attempts = range(4)
for attempt in attempts:
  if attempt == attempts.start:
    print("First iteration, who dis?")
  print(attempt)
  if attempt == attempts.stop - 1:
    print("END")
```

### OUTPUT

```
First iteration, who dis?
0
1
2
3
END
```

So `range(4)` loops from 0..3, which makes sense; but it confused me for a bit since the maximum `attempt` value (3) never matched `attempts.stop` value (4), so I had to use `-1` to detect the last iteration.
Also, it looks like `range(4)` returns a `range` instance, where as my `attempt` value is an integer, so I can't use something like `attempt.first` since that's ... not how any of this works. So I declared the `range(4)` outside the `for..in` loop because I couldn't think of another way.

Although, further Binging shows I can do a `for..in` with an `else` that might work (although arguably not entirely intuitive, but I'm also an idiot):

```python
for attempt in range(4):
  print(attempt)
else:
  print("END")
```

### OUTPUT

```
0
1
2
3
END
```

So, in summary, not sure if `attempt` will ever equal the specified target value for `range()`, or if we need to subtract 1. Also, my use cases were super simple, so might not account for nested `try..except` and other real world logic.

Thanks for coming to my TED talk.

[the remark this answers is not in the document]
jrbenny35: I had it at range -1 but changed it. This is how it should be.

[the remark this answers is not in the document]
pdehaan: I tried a smaller test case which is closer to the current implementation and I don't think my `if attempt == 45:` is working.

```python
attempts = range(5)
for attempt in attempts:
    try:
        assert 1 != 1
    except AssertionError:
        print(f"AssertionError; retrying {attempt} of {attempts.stop}")
    else:
        print("success!")
        break # exit the `for..in` loop.

    if attempt == 5:
        raise AssertionError("Experiment was not found")
else:
    print("for..else")
```

### OUTPUT

```
AssertionError; retrying 0 of 5
AssertionError; retrying 1 of 5
AssertionError; retrying 2 of 5
AssertionError; retrying 3 of 5
AssertionError; retrying 4 of 5
for..else
```

Note that it prints the debug output from the `except AssertionError:` block, but never seems to hit the `raise AssertionError("Experiment was not found")` clause.

Raising the assertion in the `else` block worked, or using `attempts.stop - 1`

```python
    if attempt == attempts.stop - 1:
        raise AssertionError("Experiment was not found")
```

app/tests/integration/nimbus/remote_settings/pages/dashboard.py
```diff
@@ -67,5 +73,20 @@ def _wait_for_it(selenium):
...
             self.wait.until(_wait_for_it)
-            el = self.find_element(*self._approve_locator)
+            if "approve" in action:
+                el = self.find_element(*self._approve_locator)
+            elif "reject" in action:
+                el = self.find_element(*self._reject_locator)
             el.click()
+
+    class Modal(Region):
+        _decline_changes_locator = (By.CSS_SELECTOR, ".modal-dialog .btn-primary")
```
pdehaan, line 83: ✂️

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -103,14 +110,14 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
-    record.approve()
+    record.action()
     selenium.get(base_url)
     # refresh until the experiment shows up
     for attempt in range(45):
         try:
             home = HomePage(selenium, base_url).wait_for_page_to_load()
             new_experiments = len(home.tables[0].experiments)
-            assert new_experiments != current_experiments
+            assert new_experiments > current_experiments
```
pdehaan, line 120: nit: Do we want to add some assertion message to this? (similar to `assert summary_page.rejected_text, "Rejected text box did not load"` below). Or maybe unneeded since we're catching the exception anyways below

pdehaan (approved):
R+! 
Awesome work! Left a few nit-level comments which you're free to ignore, and then merge at your leisure.

<kbd>/giphy "huzzah!"</kbd>

PR #5822, opened 2021-06-22
fixes #5712 feat(project): add country and locale to v5 input/serializer

### LINKED ISSUE #5712: (BE) Add Country and Locale fields to ExperimentInput, NimbusExperimentSerializer
To the [ExperimentInput](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/inputs.py#L39:L39) type add two new _countries_ and _locales_ fields as Lists of Int types.

To the [NimbusExperimentSerializer](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L267:L267) add the new _countries_ and _locales_ fields using a  PrimaryKeyRelatedField serializer.

This should make these fields modifiable via the _updateExperiment_ GraphQL mutations.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1248)

jodyheavener:
Closes #5712

This PR adds Countries and Locales to the Nimbus experiment input and serializer classes. With this change an experiment's countries and locales can be updated via GraphQL mutations.

### DIFF
--- a/app/experimenter/experiments/api/v5/inputs.py
+++ b/app/experimenter/experiments/api/v5/inputs.py
@@ -63,3 +63,5 @@ class ExperimentInput(graphene.InputObjectType):
     risk_partner_related = graphene.Boolean()
     risk_revenue = graphene.Boolean()
     risk_brand = graphene.Boolean()
+    countries = graphene.List(graphene.Int)
+    locales = graphene.List(graphene.Int)
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -5,6 +5,7 @@ from django.db import transaction
 from django.utils.text import slugify
 from rest_framework import serializers
 
+from experimenter.base.models import Country, Locale
 from experimenter.experiments.changelog_utils import generate_nimbus_changelog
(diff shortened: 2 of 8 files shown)


jodyheavener:
Thought: this is set up so you supply a country/locale ID in the mutation, but I'm sure there's a way to do it all with a country/locale's `code` field. I'm guessing this would occur somewhere in the create/update method of the serializer, but my Django chops aren't there though, so if anyone has pointers let me know. Or if using an ID is sufficient we can leave it.

lmorchard:
> this is set up so you supply a country/locale ID in the mutation, but I'm sure there's a way to do it all with a country/locale's code field. I'm guessing this would occur somewhere in the create/update method of the serializer, but my Django chops aren't there though, so if anyone has pointers let me know. Or if using an ID is sufficient we can leave it.

I'd say let's see how it works out for EXP-1250 / #5078. I think we already use ID for things like feature config selection via select box, we should be able to do the same thing with a multi-select I think?

lmorchard (approved):
LGTM. Don't see any visible effect on creating & launching an experiment, as expected since this is just groundwork for the next bits. Also was able to manually query the locale & countries in GraphQL playground, so :+1:

PR #5826, opened 2021-06-22
fixes #5818 fix(nimbus): attempt to normalize time and timezone on human-readable dates

### LINKED ISSUE #5818: End date in timeline / header is off by 1 day
The timeline end date appears to be off by 1 day. The data by the server is correct so hopefully `humanDate` just needs a quick adjustment in the front-end.

!image-2021-06-21-12-42-41-390.png|thumbnail! 

A recent PR merge (#5817) probably introduced this.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-1439)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/15532/Capture.PNG">Capture.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/15526/image-2021-06-21-12-42-41-390.png">image-2021-06-21-12-42-41-390.png</a>

jodyheavener:
Something like this until #5804 is addressed?

### DIFF
--- a/app/experimenter/nimbus-ui/src/lib/dateUtils.test.ts
+++ b/app/experimenter/nimbus-ui/src/lib/dateUtils.test.ts
@@ -24,6 +24,12 @@ describe("humanDate", () => {
   it("with year set to 'past', should produce the date, month, and year", () => {
     expect(humanDate(FAKE_ISO_DATE, "past")).toEqual("Dec 25, 2020");
   });
+
+  it("normalizes time and timezone", () => {
+    expect(humanDate("2021-06-21T10:00:00Z")).toEqual("Jun 21");
+    expect(humanDate("2021-06-21T00:00:00Z")).toEqual("Jun 21");
+    expect(humanDate("2021-06-21")).toEqual("Jun 21");
+  });
 });
 
 describe("addDaysToDate", () => {
--- a/app/experimenter/nimbus-ui/src/lib/dateUtils.ts
+++ b/app/experimenter/nimbus-ui/src/lib/dateUtils.ts
@@ -12,9 +12,16 @@ export function humanDate(
 ): string {
   const today = new Date();
   const parsedDate = new Date(date);
+  // Dates currently arrive both with and without time and
+  // timezone, so reset time to start of day and force GMT
+  parsedDate.setUTCMinutes(0);
+  parsedDate.setUTCHours(0);
+  parsedDate.setUTCSeconds(0);
+  parsedDate.setUTCMilliseconds(0);
   const options: Intl.DateTimeFormatOptions = {
     month: "short",
     day: "numeric",
+    timeZone: "GMT",
   };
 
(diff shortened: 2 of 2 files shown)


app/experimenter/nimbus-ui/src/lib/dateUtils.ts
```diff
@@ -12,9 +12,15 @@ export function humanDate(
 ): string {
   const today = new Date();
   const parsedDate = new Date(date);
+  // Dates currently arrive both with and without time and
+  // timezone, so reset time to start of day and force GMT
+  parsedDate.setUTCMinutes(0);
+  parsedDate.setUTCHours(0);
+  parsedDate.setUTCSeconds(0);
```
lmorchard, line 19: Super-picky, but maybe we also want to add `parseDate.setUTCMilliseconds(0)` here too? Technically we could still end up with a start date that's after the end date if they're not absolutely equal. But I'm hard-pressed to come up with a real problem millisecond difference would cause.

app/experimenter/nimbus-ui/src/lib/dateUtils.test.ts
```diff
@@ -24,6 +24,12 @@ describe("humanDate", () => {
   it("with year set to 'past', should produce the date, month, and year", () => {
     expect(humanDate(FAKE_ISO_DATE, "past")).toEqual("Dec 25, 2020");
   });
+
+  it("normalizes time and timezone", () => {
+    expect(humanDate("2021-06-21T10:00:00Z")).toEqual("Jun 21");
+    expect(humanDate("2021-06-21T00:00:00Z")).toEqual("Jun 21");
+    expect(humanDate("2021-06-21")).toEqual("Jun 21");
```
lmorchard, line 31: Hmm, just out of curiosity I commented out the changes to `dateUtils.ts` to see this test fail.

First case passes. Second case fails as I'd expect.

This third case doesn't fail, at least not right at this moment. That makes sense, because the issue is tied to client-side timezone. Not sure there's really a way to test that. so 🤷

lmorchard (approved):
LGTM :+1: Tried to really be picky about what could go weird here but I think it works fine. Only gotcha might be if we use non-GMT timestamps somewhere, but we have dumber problems if that's the case.

app/experimenter/nimbus-ui/src/lib/dateUtils.test.ts
jodyheavener: Interesting - the third case fails for me. But yeah I'm 4 hours ahead of you?
Screen Shot 2021-06-22 at 3 47 23 PM

lmorchard: > But yeah I'm 4 hours ahead of you?

Yup, I'd guess that's why. I hate timezones.

PR #5833, opened 2021-06-23
fix #5773 chore(nimbus): ensure release version is supplied to Sentry

### LINKED ISSUE #5773: Should be able to hide/show branches on weekly graphs
This would require making Vegalite graphs interactive

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1398)

lmorchard:
Because:

* Sentry works better if it has context on the deployed release version

This commit:

* Ensures that the git commit included on build in version.json is
  available to Sentry error report calls for nimbus-ui

### DIFF
--- a/app/experimenter/base/__init__.py
+++ b/app/experimenter/base/__init__.py
@@ -1,0 +1,25 @@
+import json
+import os
+
+from django.conf import settings
+
+APP_VERSION = None
+
+
+def app_version():
+    global APP_VERSION
+
+    if APP_VERSION is None:
+        if settings.APP_VERSION:
+            APP_VERSION = settings.APP_VERSION
+        else:
+            try:
+                version_json_path = os.path.join(settings.BASE_DIR, "version.json")
+                with open(version_json_path) as version_json_file:
+                    version_json = json.load(version_json_file)
+                    APP_VERSION = version_json["commit"]
+            except IOError:
+                # EXP-1384: Preserve default blank version if version.json unavailable
+                APP_VERSION = ""
+
+    return APP_VERSION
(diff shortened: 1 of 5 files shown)


lmorchard:
To try this out locally:

* run `./scripts/store_git_info.sh` to generate `version.json`
* run `make up`
* load any `/nimbus/` page and you should see a git commit version in the log

Capture

lmorchard:
Oh, I see: Forgot to run pytest with coverage on. This code wants a test

app/experimenter/base/__init__.py
```diff
@@ -0,0 +1,25 @@
+import json
+import os
+
+from django.conf import settings
+
+APP_VERSION = None
+
+
+def app_version():
```
lmorchard, line 9: I had this code living in `context_processors.py` but refactored out into its own function. Seemed easier to test that way and possibly reusable if we need this info anywhere else.

jodyheavener (approved):
LGTM and makes sense.

If I understand this correctly, this is going to let Sentry know what the release version is at the time an event is reported. Do we also need to do something on the sourcemaps side (like add a header) to help it pair events with the correct sourcemaps? I see this is done when [manually](https://docs.sentry.io/platforms/javascript/sourcemaps/) uploading maps, but I don't see an equivalent for auto-detected maps.

lmorchard:
> but I don't see an equivalent for auto-detected maps.

Honestly I'm crossing my fingers and hoping Sentry does the right thing with the hint. Otherwise, this is half the solution necessary anyway and the other half is to go back to looking at uploading sourcemaps tagged with release versions

PR #5821, opened 2021-06-21
closes #5608 feat(nimbus): Combine Summary Page and Request Review Page

LZoog:
also fixes #5547 and fixes #5552

Because:
* It's confusing to users to have Summary and Request Review separately because the only difference is the review & launch actions section (#5608)
* The Monitoring link contains no data if the experiment hasn't launched (#5552)
* Correct heading levels improve our HTML semantics and a11y (#5547)

This commit:
* Combines the summary page with request review page and removes request review page, redirects, and other references
* Moves some actions from `Summary` to `PageSummary` since they now occur above the `Summary` component, moves "experiment will end as soon as it is approved” feedback out of `Summary` and under the action header
* Creates `getSummaryAction` to return an action string based on experiment status and if user can review, displays on `PageSummary` header, browser tab, and sidebars
* Creates `LinkNavSummary` component to share across launched/non-launched sidebars that displays summary action
* Displays launch signoff table at bottom of `Summary` component instead of at the top of the page if experiment has been launched, scoots it into its own component with tests
* Renames `TableSummary` to `TableOverview` since the header above it now displays “Overview” per Product
* Removes “Experiment end requested” blue bill per product request since the header reads “Requested End” so it'd be redundant, adds `endRequested` status to statusCheck
* Hides monitoring link if the experiment hasn't been launched
* Adjusts heading levels for HTML semantics
* Adjusts text-success green to be more ADA friendly
* Adjusts tests and Storybook states accordingly

PR #5835, opened 2021-06-23
fixes #5834 fix(nimbus): Fix review url

### LINKED ISSUE #5834: Fix remote settings review URL
Right now there is a double slash situation going on and a button that should be an external link

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1445)

k88hudson:
Because

* The reviewUrl is wrong for "Open Remote Settings"

This commit

* Changes the FormRemoteSettingsPending to be an anchor
* Removes a the leading slash on the remote settings URL

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import classNames from "classnames";
 import React from "react";
 import Alert from "react-bootstrap/Alert";
 import Form from "react-bootstrap/Form";
@@ -28,7 +29,10 @@ const FormRemoteSettingsPending = ({
           <div>
             <LinkExternal
               data-testid="open-remote-settings"
-              className={`mr-2 btn btn-primary${isLoading ? " disabled" : ""}`}
+              className={classNames(
+                "mr-2 btn btn-primary",
+                isLoading && "disabled",
+              )}
               href={reviewUrl}
             >
               Open Remote Settings
--- a/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx
@@ -4,16 +4,16 @@
 
 import React from "react";
 import Alert from "react-bootstrap/Alert";
-import Button from "react-bootstrap/Button";
 import Form from "react-bootstrap/Form";
+import LinkExternal from "../LinkExternal";
 
-const FormApproveConfirm = ({
+const FormRemoteSettingsPending = ({
   isLoading,
-  onConfirm,
+  reviewUrl,
   actionDescription,
 }: {
   isLoading: boolean;
-  onConfirm: () => void;
+  reviewUrl: string;
   actionDescription: string;
 }) => {
   return (
@@ -26,14 +26,13 @@ const FormApproveConfirm = ({
 
         <div className="d-flex bd-highlight">
           <div>
-            <Button
+            <LinkExternal
               data-testid="open-remote-settings"
-              className="mr-2 btn btn-primary"
-              disabled={isLoading}
-              onClick={onConfirm}
+              className={`mr-2 btn btn-primary${isLoading ? " disabled" : ""}`}
+              href={reviewUrl}
             >
               Open Remote Settings
-            </Button>
+            </LinkExternal>
           </div>
         </div>
       </Form>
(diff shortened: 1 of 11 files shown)


LZoog:
This is going to cause conflicts with #5821 btw👀

jodyheavener:
@k88hudson a rebase should help that failure

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/mocks.tsx
```diff
@@ -36,7 +35,8 @@ export const BaseSubject = ({
       timeoutEvent,
       rejectChange,
       approveChange,
-      startRemoteSettingsApproval,
+      reviewUrl:
+        "http://localhost:8888/v1/admin/#/buckets/main-workspace/collections/nimbus-mobile-experiments/records",
```
jodyheavener, line 39: Nit: could try and use [`REVIEW_URL`](https://github.com/mozilla/experimenter/pull/5835/files#diff-430239984ea0d023eca01d487e1e14285726983f13689b031f324e529d398c27R17) here

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx
```diff
@@ -26,19 +26,18 @@ const FormApproveConfirm = ({
 
         <div className="d-flex bd-highlight">
           <div>
-            <Button
+            <LinkExternal
               data-testid="open-remote-settings"
-              className="mr-2 btn btn-primary"
-              disabled={isLoading}
-              onClick={onConfirm}
+              className={`mr-2 btn btn-primary${isLoading ? " disabled" : ""}`}
```
jodyheavener, line 31: This is clever in that the `disabled` class applies `pointer-events: none` to the button so it is not clickable. What it doesn't do is prevent focusing and triggering via the keyboard. Disabling links is [tricky](https://css-tricks.com/how-to-disable-links/). We _could_ remove the href as well, but even that doesn't do the whole trick. This is probably fine as-is.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.stories.tsx
```diff
@@ -136,7 +134,8 @@ export const FormRemoteSettingsPendingStory = () => (
   <FormRemoteSettingsPending
     {...{
       isLoading: false,
-      onConfirm: action("confirm"),
+      reviewUrl:
+        "https://settings-writer.prod.mozaws.net/v1/admin/#/buckets/main-workspace/collections/nimbus-mobile-experiments/records",
```
jodyheavener, line 138: Is this story intended to use a prod URL? Should it also use [`REVIEW_URL`](https://github.com/mozilla/experimenter/pull/5835/files#diff-430239984ea0d023eca01d487e1e14285726983f13689b031f324e529d398c27R17)?

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.test.tsx
```diff
@@ -14,17 +14,20 @@ import {
...
 } from "./mocks";
 
+const REVIEW_URL =
+  "http://localhost:8888/v1/admin/#/buckets/main-workspace/collections/nimbus-mobile-experiments/records";
+
 const Subject = ({
   rejectChange = () => {},
   approveChange = () => {},
-  startRemoteSettingsApproval = () => {},
+  reviewUrl = REVIEW_URL,
```
jodyheavener, line 23: I don't see any tests modifying this prop, so it could probably be removed and REVIEW_URL passed directly into the component.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.test.tsx
```diff
@@ -58,7 +59,7 @@ describe("ChangeApprovalOperations", () => {
       "open-remote-settings",
     );
     fireEvent.click(openRemoteSettingsButton);
```
jodyheavener, line 61: I don't think this fireEvent is useful anymore and could probably be removed as you've done elsewhere

jodyheavener (commented):
Works great for me, and as we discussed I can't see any reason this should stay a button over an anchor. Couple nits in tests/stories and I think we should be good.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx
k88hudson: yeah, personally I don't think maintaining custom code for that is worth it for our use case

LZoog: We could use `classNames` here, something like `className={classNames("mr-2 btn btn-primary", isLoading && "disabled"))`

jodyheavener (approved):
LGTM after that prettier failure is fixed

jaredlockhart:
>  why is vscode not helping me find these 

@k88hudson you might have to `yarn install` on your host environment to get the latest version, dependabot updates will automatically apply inside the container but not outside, so you can wind up with different version inside and outside the container with conflicting/inconsistent rules.

PR #5836, opened 2021-06-23
fixes #5804 feat(project): standardize on date over datetime in nimbus date properties

### LINKED ISSUE #5804: Standardize Nimbus model methods on date/datetime
Our Nimbus model methods return a mix of `datetime` and `date` types. We should standardize them to return one type or the other, probably `date`, since we don't display any time info; [see this comment](
https://github.com/mozilla/experimenter/pull/5402#issuecomment-858787168). ~(but, note: we should probably wait until we get some clarity from product on #5357, I want to double check we don't intend to display times)~ Edit: confirmed just to use `date`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1430)

jodyheavener:
Closes #5804

Only pass around and return `date` and not `datetime` across the nimbus backend


app/experimenter/experiments/models/nimbus.py
```diff
@@ -283,21 +279,24 @@ def computed_enrollment_days(self):
...
         )
         if paused_change:
-            return (paused_change.changed_on - self.start_date).days + 1
+            days = (paused_change.changed_on.date() - self.start_date).days
         else:
-            return self.proposed_enrollment
+            days = self.proposed_enrollment
+
+        # Always return at least 1
+        return max(days, 1)
```
jaredlockhart, line 287: Wait what's this for?

jaredlockhart (commented):
Generally looks right to me.  Why set the min enrollment days to 1?  I specifically added the ability to set it to 0 to trigger stuff in testing.

app/experimenter/experiments/models/nimbus.py
jodyheavener: Going off the [test](https://github.com/mozilla/experimenter/blob/jh/5804/lets-date/app/experimenter/experiments/tests/test_models/test_models_nimbus.py#L371:L371), looks like we always want to return at least 1. I imagine this is so when you have an experiment that starts and ends on the same day, comparing the enrollment days would result in 0, which I guess looks a little odd maybe?

jodyheavener:
Suppose this is ready now.. definitely open to feedback on my use of `max`. Everything appears to be in order / working?

app/experimenter/experiments/models/nimbus.py
jaredlockhart: No I would take this out.  The serializer validates that users can't enter less than 1, and if we want to set it to 0 for some other reason either through teh admin or through testing we should be able to.  I don't know any reason to have this and I can imagine it creating confusing behaviour in the future.

jaredlockhart (changes requested):

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Actually Just move it up to the paused_change exists case, and that should preserve the case the test is talking about and otherwise we just return the proposed_enrollment verbatim.  I think that's probably what we want.

jaredlockhart (approved):

jaredlockhart:
Thanks @jodyheavener !

PR #5837, opened 2021-06-23
e2e timeout test

jrbenny35:
This is an e2e time-out test. It creates an experiment and waits for it to timeout in remote settings.


app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -228,3 +228,62 @@ def test_create_new_experiment_remote_settings_reject(selenium, base_url):
...
+    experiment_name = f"name here remote {random.randint(0, 1000)}"
+
+    selenium.get(base_url)
+    home = HomePage(selenium, base_url).wait_for_page_to_load()
+    # home.tabs[-1].click()  # Click drafts
+    # current_experiments = None
+    # try:
+    #     current_experiments = len(home.tables[0].experiments)
+    # except TimeoutException:
+    #     current_experiments = 0
```
pdehaan, line 243: Wanna just ✂️  this?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -228,3 +228,62 @@ def test_create_new_experiment_remote_settings_reject(selenium, base_url):
...
+    review.request_review.request_launch_button.click()
+    review.approve()
+    for attempt in range(35):
+        try:
+            review.timeout_text.is_displayed()
+        except NoSuchElementException:
+            time.sleep(2)
+            selenium.refresh()
+        else:
+            assert review.timeout_text.is_displayed(), "Timeout text not shown."
```
pdehaan, line 289: Does this need a `break` or some other loop control?
Like, if it finds the timeout_text on attempt 2, will it loop 33 more times, or can/do we exit early?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -228,3 +228,62 @@ def test_create_new_experiment_remote_settings_reject(selenium, base_url):
...
+
+    # Fill Audience page
+    audience = metrics.save_and_continue()
+    audience.channel = "Nightly"
+    audience.min_version = 80
+    audience.targeting = "US_ONLY"
+    audience.percentage = 50.0
+    audience.expected_clients = 50
+    audience.save_btn()
+    review = audience.save_and_continue()
```
pdehaan, line 274: Unrelated to this PR, so feel free to swiftly ignore this... But wondering if this is very boilerplatey and we want to eventually just have some ridiculous "make_fake_experiment()" function which does all this repetitive heavy lifting. Although I guess that means all the experiments would be boilerplate and hard to change specific things like "channel" for a specific test (unless we can easily override certain fields). I'd have to revisit the previous PR and see how much of this is similar.

Or, possibly that's a terrible idea, and I should be shunned by the community.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -228,3 +228,62 @@ def test_create_new_experiment_remote_settings_reject(selenium, base_url):
...
+
+    # Review and approve
+    selenium.find_element_by_css_selector("#PageRequestReview")
+    review.launch_without_preview.click()
+    review.request_review.click_launch_checkboxes()
+    review.request_review.request_launch_button.click()
+    review.approve()
+    for attempt in range(35):
+        try:
+            review.timeout_text.is_displayed()
```
pdehaan, line 284: How... does this work?
Is the timeout window just a very small number in CI (enough that we are reasonably confident we'd hit it in 35 attempts with a 2s delay; or I guess ~70 seconds)?
If we ran the tests locally or on stage, would the test fail if the review timeout is like 5-10 minutes?

pdehaan (approved):
Looks good!
A few comments/questions, but otherwise 👍 .

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: dang I missed that

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Break wouldn't  hurt

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: The worker should run every 30 seconds, so I am waiting for it to push to kinto and then timeout. So I need a little over a minute, but I should probably add more time.

app/tests/integration/nimbus/test_e2e_create_experiment.py
pdehaan: Does it also need a final `else:` like your previous PR to cover the case where we loop 35-50 times, but never find the element?
Or is it possible that if the element is never displayed we catch the NoSuchElementException on the last iteration, and it sleeps/refreshes and just exits the `for..in` loop silently?

```python
    else:
        raise AssertionError("Experiment timeout text never displayed")
```

app/tests/integration/nimbus/test_e2e_create_experiment.py
pdehaan: Yeah, might be safer to either increase the range stop value, or possibly the sleep interval.
50 loops @ 2s pause = 100s (1m:40s)
35 loops @ 3s pause = 105s (1m:45s)

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: > 
> 
> Unrelated to this PR, so feel free to swiftly ignore this... But wondering if this is very boilerplatey and we want to eventually just have some ridiculous "make_fake_experiment()" function which does all this repetitive heavy lifting. Although I guess that means all the experiments would be boilerplate and hard to change specific things like "channel" for a specific test (unless we can easily override certain fields). I'd have to revisit the previous PR and see how much of this is similar.
> 
> Or, possibly that's a terrible idea, and I should be shunned by the community.

No this is great. It is the first thing I am doing next quarter before we add more tests.

PR #5839, opened 2021-06-24
Fetch Jetstream data more frequently.

emtwo:
Because
* We only fetch Jetstream data once per day and this is causing delays for catching up on unprocessed data

This commit
* Fetches 3 times per day (ever 8 hours) instead of once


jodyheavener (approved):
Sure!

Issue #5851, opened 2021-06-25
(UI) Move common prop sets (based on flows/experiment status) to central location


data-sync-user:
We have a lot of sets of properties for various experiment flows and/or states for use in Storybook, testing, and in our mutations. It feels like these could go in some sort of shared location, or a shared flow/prop mock file and another for reference in the UI.

We have some great "[state]BaseProps" in [ChangeApprovalOperations mocks](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/mocks.tsx). I think we can pull these out to a higher level to use across stories where applicable, there's likely other stories (and potentially tests) like [in AppLayoutWithSidebar](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.stories.tsx#L56-L75) and [LinkNavSummary tests](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/LinkNavSummary/index.test.tsx#L59) that could pull from a shared place, and we have other "base props" [set up in Summary mocks](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/Summary/mocks.tsx#L64) as well as [in PageSummary](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageSummary/mocks.tsx#L70).

Additionally, we have a lot of flow actions for mutations that correspond to similar sets of "flow props". See this [in PageSummary](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageSummary/index.tsx#L55-L101) as well as one [in Summary](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/Summary/index.tsx#L41-L43).

There may be other similar instances - just mostly feels like this could use some consolidation and something that can make these properties/objects correspond with a kind of flow or action would feel a bit cleaner.

Related, fixing up how we're checking for statuses: EXP-1158 / #5643

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1454)

jaredlockhart:
@lmorchard if you feel like doing this soon then great, if not then just close this as a wontfix 👍

PR #5838, opened 2021-06-24
fix #5739 refactor(nimbus): remove pause timeout special case

lmorchard:
Because:

* we want to make pausing enrollment a manual task soon

This commit:

* makes pausing enrollment in a live experiment subject to timeout in
  remote settings just like other updates

PR #5852, opened 2021-06-26
fix #5742 feat(nimbus): implement UI for manual enrollment end 

### LINKED ISSUE #5742: Add end enrollment UI flow
After we switch the backend to end enrollments based on publish_status, we need to add a flow to the UI to trigger it.  It'll following the exact same flow as ending the experiment except setting a different set of states to trigger the tasks.  And we'll need a second button for 'End Enrollment' next to 'End Experiment'.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1349)

lmorchard:
Because:

* we want to add UI to enable manual enrollment end

This commit:

* add isEnrollmentPaused to GQL experiment input for consistency with
  serialized experiment type, mapped to is_paused in Django model

* add support for isEnrollmentPausePending to indicate when the review
  is in progress

* rework useChangeOperationMutation to indicate loading while awaiting
  experiment refetch

* update "rejecting this review" copy to "rejecting the request to
  {actionDescription}"

* refactors handling of strings describing lifecycle review flows, moves
  all the strings into shared constants

* refactors the Settings page and ChangeApprovalOperations components

* adds more exhaustive stories for launch, end enrollment, and end cases
  on PageSummary

* reworks rejection reason display to more properly use old_status and
  old_status_next to determine what kind of review was rejected

* adds more status-based logic to make the correct GQL queries on
  approval / rejection of review

Co-authored-by: Lauren Zugai <lauren@zugai.com>


lmorchard:
Opening as a draft with work in progress because:

* There's still work to be done around tests & coverage

* I'll be out for another week after the wellness week and someone else might want to grab the baton from me

There's also possibly some some thinking to do about how manual enrollment end is handled on the backend, which could require some further tweaks.

For automated pause, we *do not* set is_paused on the Experimenter side first.

Instead, we set "isEnrollmentPaused" in the Remote Settings record. Then, we wait until it's approved in Remote Settings before setting is_paused=True in Experimenter. That means, on the Experimenter side, we never show that enrollment is paused until it's fully resolved on the Remote Settings side.

For manual pause, we might want to set is_paused=True on the Experimenter side first.

Otherwise, if we follow that same pattern as automated pause, then the only thing that represents a request to pause is status=LIVE / status_next=LIVE. (Because is_paused=False until the very end.)

Could also be a thing to punt to follow-ups, if/when we want other status=LIVE / status_next=LIVE updates other than enrollment ending.

lmorchard:
Rebased against master to pick up the changes to PageSummary and approve / reject handlers

lmorchard:
Looks like I got the storybook to build, so here are some interesting places to look:

* Enrollment active (with new button) - https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/07b3a25088c07c202b64c2ef25f48efe2587c6fa/nimbus-ui/index.html?path=/story/components-summary--enrollment-active
* Enrollment ended (no button) - https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/07b3a25088c07c202b64c2ef25f48efe2587c6fa/nimbus-ui/index.html?path=/story/components-summary--enrollment-ended
* Lots of new launch, pause, and end stories -  https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/07b3a25088c07c202b64c2ef25f48efe2587c6fa/nimbus-ui/index.html?path=/story/pages-summary--live-update-pending-can-review

LZoog:
Responding to [Les' comment](https://github.com/mozilla/experimenter/pull/5852#issuecomment-868895053) + leaving some notes for myself and the reviewer:

Per [Jared's comment here](https://github.com/mozilla/experimenter/pull/5838#issuecomment-874905288), automatically triggered pausing will be removed in favor of requiring manual end enrollment.

> Otherwise, if we follow that same pattern as automated pause, then the only thing that represents a request to pause is status=LIVE / status_next=LIVE. (Because is_paused=False until the very end.)

I think for now, this is OK since there's not any other statuses/flows that use this set and we won't need to differentiate between automatic or manual enrollment requests. That status set (which we're calling `pauseRequested`) does seem a little more fragile than say, `endRequested`, which checks for status=LIVE / status_next=COMPLETE and I could see that perhaps `is_paused` (`isEnrollmentPaused`) could potentially do a little more in the FE for us but we _are_ using `isEnrollmentPaused` in conjunction with `pauseRequested` to conditionally show the `EndEnrollment` component. 🤔 FWIW we also aren't using it for end enrollment RS timeouts either since we just check for `timeout` on the experiment.

If we do have another flow later that needs status=LIVE / status_next=LIVE, we should be able to alter `pauseRequested` on the FE fairly easily. With all of this said, I think this PR mostly just needed some hard squinting at, test fixes and additions, warning fixes, and a couple follow up issues filed, all of which I've got underway. Just wanted to dump my thoughts here before finishing those up.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -155,11 +152,14 @@ class Filters:
         )
         IS_PAUSE_QUEUED = Q(
             status=NimbusConstants.Status.LIVE,
-            publish_status=NimbusConstants.PublishStatus.IDLE,
+            status_next=NimbusConstants.Status.LIVE,
+            publish_status=NimbusConstants.PublishStatus.APPROVED,
+            # TODO: EXP-1325 Need to check for something else here for end enrollment?
```
LZoog, line 157: This comment was Les' and I'm not sure what else we need here?

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -306,6 +310,7 @@ describe("PageSummary", () => {
     );
     render(<Subject mocks={[mock, mutationMock]} />);
     const rejectButton = await screen.findByTestId("reject-request");
+    await screen.findByText("Approve and End Experiment");
```
LZoog, line 313: I added these `findByText` tests to double check the test was actually testing the correct flow.

app/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -28,20 +28,32 @@ import TableOverview from "./TableOverview";
...
+      publishStatus: NimbusExperimentPublishStatus.REVIEW,
+      status: NimbusExperimentStatus.LIVE,
+      statusNext: NimbusExperimentStatus.COMPLETE,
+      changelogMessage: CHANGELOG_MESSAGES.REQUESTED_REVIEW_END,
+    },
+    {
+      publishStatus: NimbusExperimentPublishStatus.REVIEW,
+      status: NimbusExperimentStatus.LIVE,
+      statusNext: NimbusExperimentStatus.LIVE,
+      // TODO: EXP-1325 Need to set something else here for end enrollment in particular?
```
LZoog, line 53: I'm guessing this ties in with [this TODO](https://github.com/mozilla/experimenter/pull/5852/files#r666457588) where at the moment I'm not sure what else we need

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -126,6 +126,7 @@ describe("Summary", () => {
         NimbusExperimentPublishStatus.REVIEW,
         {
           changelogMessage: CHANGELOG_MESSAGES.REQUESTED_REVIEW_END,
+          status: NimbusExperimentStatus.LIVE,
           statusNext: NimbusExperimentStatus.COMPLETE,
```
LZoog, line 130: FWIW we should end up using shared common props for flows when we do #5851. I didn't pull in [the mocks from PageSummary](https://github.com/mozilla/experimenter/pull/5852/files#diff-a22e33c18e87900a21ac78b38246317600edd36c1cc9885b6d322d4e1fb170f5R114) since we probably need to pull those shared props into a separate file at a higher level which would require some more refactoring and make this PR even longer, plus that's what #5851 is for.

LZoog:
To follow up here, yesterday I pushed test fixes and additions plus console warnings and a couple of other tweaks but left BE test fixes since we need to wait until there's a PR closing #5739.

I also already rebased/squashed the commits because of habit and I didn't realize until after pushing that this PR is probably going to stick around by the time @lmorchard gets back. 😅

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/RejectionReason.tsx
```diff
@@ -0,0 +1,51 @@
...
+      return LIFECYCLE_REVIEW_FLOWS.LAUNCH.description;
+    }
+  }, [oldStatus, oldStatusNext]);
+
+  return (
+    <Alert variant="warning" data-testid="rejection-notice">
+      <div className="text-body">
+        <p className="mb-2">
+          The request to {rejectionActionDescription} was{" "}
+          <strong>Rejected</strong> due to:
```
LZoog, line 40: Oh @lmorchard I see what you were saying in planning, #5927 covers the other side of this, "You are rejecting..." instead of "The request to". Could easily be done in this PR or in that issue separately.

app/experimenter/experiments/models/nimbus.py
lmorchard: This comment points toward basically [what I mentioned in this PR comment](https://github.com/mozilla/experimenter/pull/5852#issuecomment-868895053) - we don't set `is_paused=True` here, just a LIVE->LIVE status change that gets interpreted as a request to pause.

lmorchard:
Going to pick this back up tomorrow - PR #5935 will result in a totally different backend situation along with requiring we send `is_paused={True,False}` from the frontend.

So, some reworking will be required. But, hopefully not more than ripping out the backend changes in this PR and just updating the GQL queries on the frontend though.

lmorchard:
Alright, I think this is ready for review now. Stripped out the defunct backend changes, added some to allow mutating isEnrollmentPaused / is_paused. Also tossed in a fix for #5927 as a quick bonus. The `Co-authored-by` might disqualify @LZoog from reviewing, since she helped carry this PR for the week I was out and wrapped up test coverage.

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/RejectionReason.tsx
lmorchard: Added a fix for this https://github.com/mozilla/experimenter/pull/5852/files#diff-c86b7877156b25769ecb6879ff40aef78f66ac82fa8ff115b6e679b6870a92b9R54-R57

jaredlockhart:
Clicking the End Experiment button sends an API request but doesn't update the UI.  Refreshing the page kicked it over, but looks like we're missing something there.

jaredlockhart:
Approve dialog missing some text for End Enrollment:

Screen Shot 2021-07-14 at 10 22 58 AM

jaredlockhart:
Clicking reject also seems to not update UI:

jaredlockhart (changes requested):
Hit a bunch of snags up front, added some gifs.  I'll stop testing there for now and look again when I can get further through the flow.

LZoog:
> Clicking the End Experiment button sends an API request but doesn't update the UI. Refreshing the page kicked it over, but looks like we're missing something there.

@lmorchard FWIW I think that's taken care of in #5939 if you rebase

lmorchard:
I see what happened with the missing text & GQL query: I added in an `isEnrollmentPaused` condition to `status.pauseRequested` and that broke things. Turns out `isEnrollmentPaused` is now derived from `is_paused_published` rather than `is_paused`, and that only turns true after review completion. Might need to add another property to represent `is_paused` vs `is_paused_published`

lmorchard:
> @lmorchard FWIW I think that's taken care of in #5939 if you rebase

Hmm, that only takes care of the end experiment UI, not this new UI. Need to copy that over / see if it can be more generalized

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -745,6 +746,38 @@ def test_paused_experiment_returns_date(self):
...
+            [NimbusExperimentFactory.Lifecycles.PAUSING_REVIEW_REQUESTED, False, True],
+            [NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE, False, True],
+            [NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_WAITING, False, True],
+            [NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_TIMEOUT, False, True],
+            [NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_APPROVE, True, False],
+            [NimbusExperimentFactory.Lifecycles.PAUSING_REJECT, False, False],
+            [NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_REJECT, False, False],
+        ]
+    )
+    def test_experiment_pause_pending(self, lifecycle, expected_paused, expected_pending):
```
lmorchard, line 760: Figured I might as well try exhausting all the cases here to make sure the pending vs published pause states worked as expected.

app/experimenter/nimbus-ui/src/hooks/useChangeOperationMutation.tsx
```diff
@@ -58,7 +61,11 @@ export function useChangeOperationMutation(
                 return void setSubmitError(message.status.join(", "));
               }
 
-              refetch && refetch();
+              if (refetch) {
+                setIsLoadingRefetch(true);
+                await refetch();
+                setIsLoadingRefetch(false);
+              }
```
lmorchard, line 68: This seems to better ensure that the UI reflects that something's still in-flight for the refetch after the mutation

app/experimenter/nimbus-ui/src/hooks/useChangeOperationMutation.test.tsx
```diff
@@ -20,28 +19,65 @@ import { useChangeOperationMutation } from "./useChangeOperationMutation";
...
 
-  it("indicates when loading", async () => {
-    const {
-      isLoading,
-      callbacks: [callback],
-    } = setupTestHook();
-    expect(isLoading).toBeFalsy();
-    await act(async () => void callback());
-    waitFor(() => expect(isLoading).toBeTruthy());
+  it("indicates when loading mutation", async () => {
```
lmorchard, line 37: Reworking these tests to use [`await waitForNextUpdate()`](https://react-hooks-testing-library.com/usage/advanced-hooks#async) seems to exercise the functionality better. But I suspect there might be an intermittent race condition in here still that I haven't been able to flush out

app/experimenter/nimbus-ui/src/hooks/useChangeOperationMutation.test.tsx
```diff
@@ -56,39 +92,42 @@ const setupTestHook = (customMocks: MockedResponse[] = []) => {
     },
   ];
 
-  const mocks = customMocks.length
-    ? customMocks
-    : mutationSets.reduce<MockedResponse[]>((acc, cur) => {
```
lmorchard, line 61: Seemed like `customMocks` was never used in these tests?

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -18,7 +18,7 @@ import PageLoading from "../PageLoading";
 
 type AppLayoutWithExperimentChildrenProps = {
   experiment: getExperiment_experimentBySlug;
-  refetch: () => void;
+  refetch: () => Promise<unknown>;
   analysis?: AnalysisData;
 };
```
lmorchard, line 23: Not sure where else we're using `refetch` - this change doesn't seem to have broken any other tests, but we might need to update this type elsewhere eventually. Technically the refetch from Apollo *is* an async function and not `() => void`, but we didn't have that plumbed through for use in `useChangeOperationMutation`

app/experimenter/nimbus-ui/src/hooks/useChangeOperationMutation.test.tsx
```diff
@@ -20,28 +19,68 @@ import { useChangeOperationMutation } from "./useChangeOperationMutation";
...
+    const refetchState = {
+      called: false,
+      resolve: null as null | (() => void),
+    };
+    const refetch = () => {
+      refetchState.called = true;
+      return new Promise(
+        (resolve) => (refetchState.resolve = () => resolve(null)),
+      );
+    };
```
lmorchard, line 59: This feels a little dirty, but seems like a decent way to predictably walk through the states around refetch.

app/experimenter/experiments/api/v5/types.py
```diff
@@ -222,6 +223,9 @@ def resolve_jexl_targeting_expression(self, info):
     def resolve_is_enrollment_paused(self, info):
         return self.is_paused_published
 
+    def resolve_is_enrollment_pause_pending(self, info):
+        return self.is_paused and not self.is_paused_published
+
```
lmorchard, line 228: This combined with status=LIVE / status_next = LIVE should give us all the hints needed on the frontend

lmorchard (commented):
This PR has become monstrous, but I think it's got some needed refactorings. Commenting on some additions that should fix the UI with respect text, loading indication, and using the correct GQL mutation for end enrollment

jaredlockhart:
@lmorchard If you feel like there's refactors in here that are necessary but don't actually implement the new UI directly you could pull them out into separate more isolated PRs, review and land those first, and then rebase this PR to just contain the actual logic to implement the new UI.  Up to you.

lmorchard:
> If you feel like there's refactors in here that are necessary but don't actually implement the new UI directly you could pull them out into separate more isolated PRs, review and land those first, and then rebase this PR to just contain the actual logic to implement the new UI. Up to you.

No, it's all pretty much inseparable at this point. The refactors were necessary to get the shared logic working between the now three review cases we have

lmorchard:
Filed #9852 to capture a refactor idea for `useChangeOperationMutation` and keep this PR from getting bigger.

app/experimenter/nimbus-ui/src/lib/constants.ts
```diff
@@ -60,6 +60,33 @@ export const RISK_QUESTIONS = {
...
+  LAUNCH: {
+    buttonTitle: "Launch Experiment",
+    description: "launch this experiment",
+    requestSummary: "Requested Launch",
+    reviewSummary: "Review Launch Request",
+  },
+  PAUSE: {
+    buttonTitle: "End Enrollment for Experiment",
+    description: "end enrollment for this experiment",
+    requestSummary: "Requested End Enrollment",
```
jodyheavener, line 73: Nit: I might phrase these as "X Requested", but def feel free to ignore :P

jodyheavener (commented):
Stared at this for a bit and had a few thoughts, but nothing major. I was able to go through the flow from start to end (as well as rejections) and didn't get hung up on anything. 🥳

app/experimenter/nimbus-ui/src/components/Summary/EndEnrollment.tsx
```diff
@@ -0,0 +1,63 @@
...
+          </div>
+        </Alert>
+      ) : (
+        <Button
+          variant="primary"
+          onClick={toggleShowEndConfirmation}
+          disabled={isLoading}
+          data-testid="end-enrollment-start"
+        >
+          End Enrollment for Experiment
```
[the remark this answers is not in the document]
lmorchard: I can't find an issue (which I thought existed) but I think we've been pushing to be more verbose and explicit about these messages. Might be worth a follow-up issue to revisit though

app/experimenter/nimbus-ui/src/components/Summary/EndEnrollment.tsx
```diff
@@ -0,0 +1,63 @@
...
+}) => {
+  const [showEndConfirmation, setShowEndConfirmation] = useState(false);
+
+  const toggleShowEndConfirmation = useCallback(
+    () => setShowEndConfirmation(!showEndConfirmation),
+    [showEndConfirmation, setShowEndConfirmation],
+  );
+
+  return (
+    <div className="mb-4" data-testid="enrollment-end">
```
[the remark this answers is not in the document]
lmorchard: Could be worth a follow-up issue, e.g. to combine these two buttons into a component side-by-side with state to display the confirmation form mutually exclusively for one or the other. Seems a decent enough chunk of work to avoid making this PR bigger

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx
```diff
@@ -5,51 +5,66 @@
...
   ShowFormRejectReason,
   ShowRemoteSettingsPending,
 }
 
 export type ChangeApprovalOperationsProps = {
+  actionButtonTitle: string;
   actionDescription: string;
   isLoading: boolean;
   canReview: boolean;
+  // TODO: refactor to just take `experiment` rather than all these separate props?
```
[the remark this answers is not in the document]
lmorchard: Yeah, I keep thinking some kind of Context with experiment could be a good refactor, maybe also take the place of that functional child pattern we use for all the experiment-using pages

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Wait explain this.  Shouldn't you just be able to infer all this based on publish_status? Why do you need to check the difference between the pause and published pause state?

app/experimenter/nimbus-ui/src/lib/experiment.ts
```diff
@@ -31,7 +32,14 @@ export function getStatus(
     approved: publishStatus === NimbusExperimentPublishStatus.APPROVED,
     review: publishStatus === NimbusExperimentPublishStatus.REVIEW,
     waiting: publishStatus === NimbusExperimentPublishStatus.WAITING,
-    endRequested: experiment?.statusNext === NimbusExperimentStatus.COMPLETE,
+    // TODO: EXP-1325 Need to check something else here for end enrollment in particular?
+    pauseRequested:
+      status === NimbusExperimentStatus.LIVE &&
+      statusNext === NimbusExperimentStatus.LIVE &&
+      isEnrollmentPausePending === true,
```
jaredlockhart, line 39: The only LIVE>LIVE update we have right now is pausing so that should be sufficient to know we're in a pausing flow?

lmorchard: Wanted to be specific here and only catch the pending end enrollment case

app/experimenter/experiments/api/v5/types.py
lmorchard: Can't infer specifically that this is a pending change to end enrollment, and I wanted to be specific https://github.com/mozilla/experimenter/pull/5852#discussion_r670715048

app/experimenter/nimbus-ui/src/lib/experiment.ts
jaredlockhart: Oh I see what you mean.  I think we'll end up having to do a lot more infrastructure before we do any non pause live updates, so we shouldn't really need to worry about it for now.

app/experimenter/nimbus-ui/src/components/Summary/EndEnrollment.tsx
```diff
@@ -0,0 +1,63 @@
...
+      {showEndConfirmation ? (
```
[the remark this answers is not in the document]
jaredlockhart: Yeah good point, interesting thought.  You're right that they're pretty heavily guarded by the multiple review steps.  I'd say leave it and if it starts tripping people up we can revisit it, but good thing to point out. 👍

app/experimenter/nimbus-ui/src/hooks/useChangeOperationMutation.test.tsx
jaredlockhart: Circle reveals all.

jaredlockhart (approved):
Tested locally, everything went off without a hitch.  Looks great, works great, thanks @lmorchard ! 🎉 🎉 🎉 🎉

lmorchard:
Alright, I'm gonna hit the merge button to get this monster landed, then file a few follow up issues that are hopefully quick maintenance sized things

lmorchard:
Tried to capture some of @jodyheavener's feedback in follow ups:

* Refining the end action buttons - https://github.com/mozilla/experimenter/issues/5970
* Thinking about Context to pass experiment around - https://github.com/mozilla/experimenter/issues/5971

PR #5879, opened 2021-07-05
fixes #5866 feat(reporting): Add admin normandy launch reportlog filter

### LINKED ISSUE #5866: [Reporting]: add ability to filter normandy related launch filter for reportlogs in admin
We would like to filter normandy related reportlogs, so that we can see what launch, update or pause occurred in a given time range.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1457)

tiftran:
Because:

* This would allow to view normandy related updates for cpr reports

This commit:

* Adds a filter to filter normandy launch reportlogs

### DIFF
--- a/app/experimenter/reporting/admin/reporting.py
+++ b/app/experimenter/reporting/admin/reporting.py
@@ -10,7 +10,7 @@ from experimenter.reporting.models import ReportLog
 
 def download_csv(modeladmin, request, queryset):
     response = HttpResponse(content_type="text/csv")
-    response["Content-Disposition"] = "attachment;" "filename=reportlog.csv"
+    response["Content-Disposition"] = "attachment; filename=reportlog.csv"
     writer = csv.writer(response)
     writer.writerow(
         [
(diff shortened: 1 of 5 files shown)


app/experimenter/reporting/admin/reporting.py
```diff
@@ -1,8 +1,85 @@
...
+from django.db.models import Q
+from django.http import HttpResponse
+from rangefilter.filters import DateTimeRangeFilter
 
 from experimenter.reporting.models import ReportLog
 
 
+def download_csv(modeladmin, request, queryset):
+    response = HttpResponse(content_type="text/csv")
+    response["Content-Disposition"] = "attachment;" "filename=reportlog.csv"
```
jaredlockhart, line 13: This is formatted a bit wonky?  It's supposed to be one string?

app/experimenter/reporting/admin/reporting.py
```diff
@@ -1,8 +1,85 @@
...
+
+class CPRReportListFilter(admin.SimpleListFilter):
+    title = "CPR Related Report Log Filter"
+    parameter_name = "CPR ReportLogs"
+
+    def lookups(self, request, model_admin):
+        return [("cpr", "CPR Only")]
+
+    def queryset(self, request, queryset):
+        if self.value() == "cpr":
```
jaredlockhart, line 58: What's this "cpr" string, is this a constant anywhere?

jaredlockhart (approved):
Ah @tiftran .  I'm having so many flashbacks.  Do you have any IDEA how many admin csv loaders/exporters I've built?? XDDDD  I seriously need badges for this stuff.  You've earned your admin CSV export badge.

Actually you know what's funny about this.  Back in the day in py2, the default strings were ASCII not UNICODE, so every SINGLE time you had to do ANYTHING with CSV, you'd build it naively, like this, and then it would crash the second it got to prod because there'd be a unicode character SOMEWHERE that someone had copy pasted from Microsoft Excel and you'd have to go in and add all these explicit utf8 decode/encode steps all over the place.  One of the SPECIFIC reasons for the switch to python 3 was to change the default string to unicode, so in theory you shouldn't ever have to deal with that.  You kids.  So spoiled.  Okay, grandpa story time over.  G/j @tiftran !

app/experimenter/reporting/admin/reporting.py
tiftran: Decided to change it to "Normandy Launch Updates" instead

PR #5880, opened 2021-07-05
fixes #5810 fix(legacy): Remove firefox version 100

### LINKED ISSUE #5810: [Legacy] Remove Firefox Version 100
Normandy doesn't handle firefox version 100 using the version filter object very well.
As a precaution, we'll remove firefox version 100. Users will need to specify version 100+ in additional filtering.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1427)

tiftran:
This commit:

* removes firefox version 100

Because:

* Normandy isn't able to handle ff 100 using standard version filter object


jaredlockhart (approved):
@tiftran so what are we gonna do when we hit 100?  Also LGTM for now 👍

jaredlockhart:
Also make sure to rebase this on `main` before merging 👍

tiftran:
@jaredlockhart hitting 100, would mean adding versioning notes to additional filtering in experimenter and then a normandy operator would be using the version range filter object, rather than the version filter object to set versions.

PR #5887, opened 2021-07-06
chore(nimbus): decrease integration test celery timing

jaredlockhart:
Because

* We should be able to decrease the celery task timeouts so speed up the integration tests

This commit

* Turns them down to 5s to update and 10s to timeout


pdehaan (approved):
R+ 👍 
Can confirm that 5<30 and 10<60.

PR #5891, opened 2021-07-07
chore(project): silence new django model warning

jaredlockhart:
Because

* Django added a new warning about auto generated primary key fields
* It's noisy and unnecessary

This commit

* Silences the warning


jodyheavener (approved):

app/experimenter/settings.py
```diff
@@ -404,7 +404,7 @@
 SECURE_REFERRER_POLICY = config("SECURE_REFERRER_POLICY", default="origin")
 
 # Silenced ssl_redirect and sts checks
```
pdehaan, line 406: Need to update comment too?

jaredlockhart: ;)

pdehaan (approved):

PR #5896, opened 2021-07-07
fixes #5864 feat(reporting): Add serializer/endpoint for reportlog and stats

### LINKED ISSUE #5864: [Reporting]: Add serializer/endpoint for reportlogs and statistics
We would like to have a serializer/endpoint for reportlogs and statistics so that it can later be displayed in the frontend.
Endpoint should be able to take two dates for filtering

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1455)

tiftran:


jaredlockhart (commented):
Nice.  This API only pertains to reporting right?  If it's only about reporting and talks to reporting models it should live in the `/reporting/` app right?

tiftran:
@jaredlockhart  ooh you're right, it only deals with reporting

jaredlockhart:
@tiftran Yeah so call it v7 but just put it in `/reporting/api/`

app/experimenter/reporting/api/v7/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+
+from experimenter.reporting.api.v7.views import ReportDataView
+
+urlpatterns = [
+    url(
+        r"^(?P<start_date>\d{4}-\d{2}-\d{2})/(?P<end_date>\d{4}-\d{2}-\d{2})$",
```
jaredlockhart, line 7: You know this is kind of a neat idea.  Usually I'd use query parameters for parameters like this, but putting it in the url lets you enforce the date format.  There's some [discussion](https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api) here related.  Fundamentally it's just about patterns there's no hard and fast rule.  

Maybe according to what they're saying here the reportlogs is the resource and the dates are filters, so they'd go in query parameters.  If you're the only one calling it then you can structure it however you like, but one thing to think about is, if some somewhere starts calling this API, and you need to move something around that would break the calling interface, how flexible is this design?  For instance what if in the future you need to filter on date with time?  Or make either start/end date optional?  Something to think about.

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+
+from django.conf import settings
+from rest_framework import views
+from rest_framework.response import Response
+
+from experimenter.reporting.constants import ReportLogConstants
+from experimenter.reporting.models import ReportLog
+
+
+class ReportDataView(views.APIView):
```
jaredlockhart, line 14: You don't wanna use a ListAPIView and just get a bunch of this logic for free?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+        end_date = kwargs.get("end_date", None)
+        data = self.get_data(start_date, end_date)
+        results = Response(data)
+        return results
+
+    def get_data(self, start_date, end_date):
+
+        experiment_names = (
+            ReportLog.objects.filter(
+                timestamp__range=[start_date, end_date], experiment_new_status="Live"
```
jaredlockhart, line 27: Looks like `"Live"` should be a constant here?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+            )
+            .exclude(projects__name__in=ReportLog.EXCLUDED_PROJECTS)
+            .values_list("experiment_name", flat=True)
+            .distinct()
+        )
+        reportLogs = ReportLog.objects.filter(experiment_name__in=experiment_names)
```
jaredlockhart, line 33: Why name and not slug? or ID?

jaredlockhart, line 33: Camel Casing??? 🐫

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+        nimbus_durations = []
+        status_durations = defaultdict(lambda: {})
+
+        for name in experiment_names:
+            status_duration_times = {}
+            experiment_reportlogs = reportLogs.filter(experiment_name=name).order_by(
+                "timestamp"
+            )
+            reportlog = experiment_reportlogs.first()
+            type = reportlog.experiment_type
```
jaredlockhart, line 48: `type` is a reserved word

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+            for i in range(len(experiment_reportlogs) - 1):
+                current_status = experiment_reportlogs[i].experiment_new_status
+                next_status_ts = experiment_reportlogs[i + 1].timestamp
+                current_ts = experiment_reportlogs[i].timestamp
+                time_delta = next_status_ts - current_ts
+                current_time = status_duration_times.get(
+                    current_status, datetime.timedelta(hours=0)
+                )
+                status_duration_times[current_status] = time_delta + current_time
+                if current_status != "Live":
```
jaredlockhart, line 66: `"Live"` constant?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+                    pre_live_duration += time_delta
+
+            data = {
+                "name": name,
+                "url": url,
```
jaredlockhart, line 71: If `url` is only used once I'd just move line 55 right into here

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+        status_durations = defaultdict(lambda: {})
+
+        for name in experiment_names:
+            status_duration_times = {}
+            experiment_reportlogs = reportLogs.filter(experiment_name=name).order_by(
+                "timestamp"
+            )
+            reportlog = experiment_reportlogs.first()
+            type = reportlog.experiment_type
+            projects_string = " "
```
jaredlockhart, line 49: Is the space necessary?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+                "time_in_draft": format_time(status_duration_times.get("Draft", None)),
+                "time_in_preview": format_time(
+                    status_duration_times.get("Preview", None)
+                ),
+                "time_in_review": format_time(status_duration_times.get("Review", None)),
+                "time_in_ship": format_time(status_duration_times.get("Ship", None)),
+                "time_in_accepted": format_time(
+                    status_duration_times.get("Accepted", None)
+                ),
+                "time_in_live": format_time(status_duration_times.get("Live", None)),
```
jaredlockhart, line 83: Constants

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+            }
+            results.append(data)
+
+            if "Normandy" in type and "Rollout" not in type:
+                normandy_durations.append(pre_live_duration)
+            if "Nimbus" in type:
```
jaredlockhart, line 89: More constants?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+                nimbus_durations.append(pre_live_duration)
+
+            for status, time in status_duration_times.items():
+                if status_durations[type].get(status):
+                    status_durations[type][status].append(time)
+                else:
+                    status_durations[type][status] = [time]
+
+        status_medians = {}
+        for type in status_durations:
```
jaredlockhart, line 99: Now you've reused `type` here and above in the same namespace?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+                "num_of_launches": type_launch,
+                "num_of_launch_by_project": project_launch,
+            },
+        }
+
+        return response
+
+
+def format_url(slug, type):
+    exp_type = f"experiments/{slug}"
```
jaredlockhart, line 132: urljoin instead of fstring here?

app/experimenter/reporting/api/v7/views.py
```diff
@@ -0,0 +1,140 @@
...
+    if "Nimbus" in type:
```
jaredlockhart, line 133: constant

jaredlockhart (changes requested):
Generally looks good, couple little cleanups for names/constants.  Worth thinking about the URL parameters but up to you.  Maybe the get_data method could be broken into a couple smaller chunks?

app/experimenter/reporting/api/v7/views.py
tiftran: current structure of the data is 
```

app/experimenter/reporting/api/v7/urls.py
tiftran: so I initially wanted to do query parameters where it would've been like `/v7/?startdate=2020-11-11&enddate=2020-12-12` had some trouble, don't remember why (probs cause I suck at regex), but I think I might take another stab at it as another issue?

jaredlockhart: We all suck at regex's, @tiftran 🙏 😆

jaredlockhart (approved):
Ship it! 🎉 🎉 🎉 🎉 🎉 🎉

PR #5916, opened 2021-07-08
fixes #5915 fix(reporting): Change event pair mapping to use event cl…

### LINKED ISSUE #5915: [Reporting]: Change Reporting constants so that pair mapping uses event classes
EVENT_PAIRS current uses strings instead of values from the event classes.
We would need to move the Event and EventReason Class out of ReportlogConstants

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1471)

tiftran:
…asses

Because:

 * It was using strings before

This commit:

 * replaces event pairing from strings to event classes values

### DIFF
--- a/app/experimenter/reporting/constants.py
+++ b/app/experimenter/reporting/constants.py
@@ -1,6 +1,28 @@
 from django.db import models
 
 
+class Event(models.TextChoices):
+    CREATE = "Create"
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
👍 👍 👍 👍 👍

PR #5918, opened 2021-07-09
fix #5917 chore(nimbus): display bucketing/isolation groups in admin as read only

### LINKED ISSUE #5917: Add bucketing/isolation groups to NimbusExperiment admin as read only
To help understand population allocations, we should add the bucketing/isolation group models to a NimbusExperiment admin but set them to read only to prevent them accidentally (or intentionally) being modified.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1472)

jaredlockhart:
Because

* It will be helpful in some cases to inspect the assigned bucket ranges/isolation groups for an experiment
* They should never be edited by hand since they are always automatically assigned by the bucket allocator

This commit

* Displays the bucket and isolation groups assigned to an experiment in the nimbus experiment admin as read only

### DIFF
--- a/app/experimenter/experiments/admin/nimbus.py
+++ b/app/experimenter/experiments/admin/nimbus.py
@@ -4,6 +4,7 @@ from django.contrib.postgres.forms import SimpleArrayField
 
 from experimenter.experiments.models import (
     NimbusBranch,
+    NimbusBucketRange,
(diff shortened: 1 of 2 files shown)


jodyheavener (approved):
yar

PR #5923, opened 2021-07-09
fix #5922 chore(nimbus): use cache instead of global for loading app version

### LINKED ISSUE #5922: Version file tests fail in manage.py test
I just noticed the new tests for loading the version file are failing in manage.py test, but passing in pytest.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1475)

jaredlockhart:
Because

* Using a global to store the version so we don't load the file twice can lead to non deterministic stateful behaviour such as state persisting across tests
* Manifests in the Django test runner but not the pytest runner, but I suspect that might be related to the thread scheduling and memory locality

This commit

* Uses the builtin python functools cache which explicitly allows clearing between tests

### DIFF
--- a/app/experimenter/base/__init__.py
+++ b/app/experimenter/base/__init__.py
@@ -1,25 +1,20 @@
 import json
-import os
+from functools import cache
 
 from django.conf import settings
 
-APP_VERSION = None
 
-
+@cache
 def app_version():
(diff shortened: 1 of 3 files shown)


jodyheavener (approved):
Sure

PR #5925, opened 2021-07-09
Fix #5886: fix(visualization) - only load primary or secondary outcomes when overall results are available.

### LINKED ISSUE #5886: Default Browser Adoption experiment has error for its results
Need to look into the error showing up on this experiment: https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment-release/results

I tried it out locally and did not get an error so further investigation is needed.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1460)

emtwo:
Because:
* Experiments with primary metrics fail to load the weekly view since overall data is required to show them

This commit:
* Only show primary and secondary metrics if overall data is available.

I tested this locally by confirming https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment-release/results works (it's currently broken on prod)

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
@@ -104,21 +104,22 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
             </div>
 
             <div>
-              {experiment.primaryOutcomes?.map((slug) => {
-                const outcome = configOutcomes!.find((set) => {
-                  return set?.slug === slug;
-                });
+              {analysis?.overall &&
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Oh awesome thank you @emtwo ! 🎉 🎉

jaredlockhart:
Not that we need to block this pr since we should get a fix out but is there any way to add/modify any tests we have to catch this error in the future?

emtwo:
> Not that we need to block this pr since we should get a fix out but is there any way to add/modify any tests we have to catch this error in the future?

I filed a follow-up to not block this: https://github.com/mozilla/experimenter/issues/5926

PR #5938, opened 2021-07-12
fixes #5783 feat(nimbus): Add duration tooltip to Audience & Summary

### LINKED ISSUE #5783: Clarification between enrollment period and duration period
I've heard some feedback that it's confusing to understand if enrollment period is included in the duration period calculations. 

In the UI, it's calculated including it, but often we want additional X days after enrollment, so it's causing confusion if we should continue the experiment longer or end it on "calculated" time.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-1381)

LZoog:
Because:
* Users were not sure if total duration included the enrollment period

This commit:
* Adds a tooltip displaying the same copy for "Experiment duration" on the Audience page and "Total duration" in SummaryTimeline


jodyheavener (approved):
LGTM!

PR #5939, opened 2021-07-12
fixes #5921 fix(nimbus): allow refetching after end experiment, disable buttons while waiting

### LINKED ISSUE #5921: The Nimbus UI can take between 3 and 25 seconds to be refreshed when requesting to end an experiment
**[Prerequisites]:**
- Have an Desktop experiment in the `Live` state. (eg [here](https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/cmuresan-testing-end-experiment-ui-response-time))

**[Steps to reproduce]:**
1. Navigate to the Nimbus experiment page from prerequisites.
2. Click the "End Experiment" and "Yes, end the experiment" buttons.
3. Observe the page.

**[Expected result]:**
- The page is instantly refreshed and the "All set, the experiment will end as soon as it is approved" message is displayed.

**[Actual result]:**
- The page can take between 3 and 25 seconds to refresh, the buttons to become unavailable, and the message to be displayed.

**[Notes]:**
- Users could potentially request the end of an experiment multiple times before the UI is refreshed and the buttons are unavailable.

it can take between 3 and 25s for the UI to change when requesting an experiment to end

can make multiple end requests before the UI changes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1474)

jodyheavener:
Closes #5921 

This PR updates the flow for ending an experiment to immediately reload the experiment after the request to end confirmation button is clicked.

**What was happening?** We actually try to refetch experiment data after one of the change approval operation mutations is called, but only if the refetch function is provided. In this case it wasn't being provided, so we were falling back to the regular polling interval, which is why it was taking a varying amount of time to update the UI.

I also added a faux loading state when this button is clicked. It's not really waiting on anything to finish (like a network request), but when the reftehc occurs it's going to result in the UI updating anyway, so it can't hurt to making things look "loading".

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
@@ -195,7 +195,7 @@ const PageContent: React.FC<{
       )}
 
       <h2 className="mt-3 mb-4 h4">Summary</h2>
-      <Summary {...{ experiment }} />
+      <Summary {...{ experiment, refetch }} />
     </>
   );
 };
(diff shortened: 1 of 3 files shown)


LZoog (approved):
Well, that's super tricky. I almost feel like we need a comment somewhere explaining what `refetch` does, but I also don't think we're going to run into this problem much more often. Nice find.

PR #5941, opened 2021-07-12
Docker Shared Layer Cache: Does this work?

jaredlockhart:
Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,10 @@ jobs:
       - run:
           name: Build docker images
           command: |
-            make build_prod
+            docker login -u $DOCKER_USER -p $DOCKER_PASS
+            make build_write_dev
+            make build_write_test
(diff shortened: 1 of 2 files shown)


jaredlockhart:
So it technically works insomuch as it pushes the layers to dockerhub and then locally when you build it'll discover and pull them in much more quickly than it would take to build them from scratch *EXCEPT* it does a tarball/export step ~every time~ that takes ~90s for our ~2gb dev image, regardless of whether the layers are already cached locally and haven't changed.  So that pretty much negates the benefit of doing this.  There's relevant tickets filed about this :

https://github.com/docker/buildx/issues/626
https://github.com/docker/build-push-action/issues/401

So we can revisit this in the future if that behaviour changes.

jaredlockhart:
I did manage to get this working here https://github.com/mozilla/experimenter/pull/5944

PR #5945, opened 2021-07-13
fixes #5542 fix(nimbus): Allow .0001 population_percent to be valid in ReadyForReview serializer

### LINKED ISSUE #5542: Population percent of 0.0001 not considered valid
It looks like if you set your Audience population percent to 0.0001 this is considered invalid even though the server says you can use it. Under the hood the GQL readyForReview check is saying "Ensure this value is greater than or **equal to** 0.0001". 0.0002 does work, so my guess it's actually checking greater than and not equal to.

As an aside this should also be showing that message as an error on the field, but that's a separate issue (EXP-773).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-774)

LZoog:
fixes #5542

Because:
* 0.0001 is a valid value that users can save, but the "Ready for review" validation errors show this to be an error

This commit:
* Adds a custom population_percent validation method to check for a min value rather than rely on min_value

---

@jaredlockhart I fully expect you to be like "uhhh no" - totally up for an alternative here. 😉

If you print the value of `value` in `validate_population_percent` when `min_value=0.0001`, this new test never appears to hit the validator despite `experiment.population_percent` equaling `0.0001`. It just spits back the "Ensure this value is greater than 0.001" error.

If you print the value of `value` in `validate_population_percent` when `min_value=0.0`, it comes back as `0.0001` correctly. However, when I compare `if value < 0.0001`, the if statement is truthy somehow and `value == 0.0001` is `False`. I double checked both values are decimal types. Checking the value against something a smidge smaller than `0.0001` instead (like what was done here) works. 🙃

Is there some esoteric Python/Django stuff happening here?

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -560,7 +560,12 @@ class NimbusReadyForReviewSerializer(serializers.ModelSerializer):
     proposed_duration = serializers.IntegerField(required=True, min_value=1)
     proposed_enrollment = serializers.IntegerField(required=True, min_value=1)
     population_percent = serializers.DecimalField(
-        7, 4, min_value=0.0001, max_value=100.0, required=True
+        7,
+        4,
+        min_value=0.00009,
+        max_value=100.0,
+        required=True,
+        error_messages={"min_value": NimbusConstants.ERROR_POPULATION_PERCENT_MIN},
     )
     total_enrolled_clients = serializers.IntegerField(required=True, min_value=1)
     firefox_min_version = serializers.ChoiceField(
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -418,6 +418,7 @@ Optional - We believe this outcome will <describe impact> on <core metric>
     ERROR_REQUIRED_FEATURE_CONFIG = (
         "You must select a feature configuration from the drop down."
     )
+    ERROR_POPULATION_PERCENT_MIN = "Ensure this value is greater than or equal to 0.0001."
 
     # Analysis can be computed starting the week after enrollment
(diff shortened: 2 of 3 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -630,3 +630,10 @@ def validate_hypothesis(self, value):
         if value == NimbusExperiment.HYPOTHESIS_DEFAULT.strip():
             raise serializers.ValidationError("Hypothesis cannot be the default value.")
         return value
+
+    def validate_population_percent(self, value):
+        # HACK - setting min_value=0.0001 throws a validation error with that value
+        if value < 0.000099:
```
jaredlockhart, line 636: Can't you just change the min value in the serializer field above to that?

LZoog: Should have commented here - we can, but it changes the error message to "Ensure this value is greater than or equal to 9.9e-05" instead of `0.0001`.

jaredlockhart: Should just be able to pass in a custom error message [like so](https://stackoverflow.com/questions/26943985/custom-error-messages-in-django-rest-framework-serializer/26975268#26975268)

LZoog: Sure, I can make that change, I mostly threw this PR up wondering if you'd point something out I was missing (I'm still not sure _why_ we need the `0.000099` solution) and it was easier to debug by seeing if it ran through a validator. I'll push that up this morning. 😎

LZoog: Pushed👌

jaredlockhart (approved):
Beauty 👩‍🍳 🍰 👍

PR #5948, opened 2021-07-13
fix #5946 chore(project): cache docker layers in dockerhub

### LINKED ISSUE #5946: Cache docker layers between ci and local dev
Whenever we land changes to package manifests, the local Docker cache will become invalidated, which requires a lengthy rebuild process.  We already rebuild the containers with the new package manifests in CI, so it would be handy to make the layers built there available to local dev.  The new Docker Buildkit supports [fetching from remote caches](https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources) during build, so we should be able to build in CI, push the layers to DockerHub, and configure the build commands in the Makefile to use them as a remote cache source.

Some helpful discussion is found here:

https://andrewlock.net/caching-docker-layers-on-serverless-build-hosts-with-multi-stage-builds---target,-and---cache-from/
https://github.com/moby/moby/issues/39003
https://gist.github.com/kekru/8ac61cd87536a4355220b56ae2f4b0a9

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1486)

jaredlockhart:
Because

* It would save time if we push the docker build layers from circleci to dockerhub to reuse locally

This commit

* Uses the new buildkit --cache-from flag to pull in build layers from dockerhub when applicable
* Builds all stages and pushes them with their layers to dockerhub on every merge to main
* Configures local builds to pull those layers from dockerhub
* Removes the build circle stage since each circle task will now be able to pull from dockerhub rather than the local circle docker cache


Makefile
```diff
@@ -54,20 +54,15 @@ jetstream_config:
 	curl -LJ -o app/experimenter/outcomes/jetstream-config.zip https://github.com/mozilla/jetstream-config/archive/main.zip
 	unzip -o -d app/experimenter/outcomes app/experimenter/outcomes/jetstream-config.zip
 
-build_dev: jetstream_config
-	docker build --target dev -f app/Dockerfile -t app:dev app/
+build_dev: jetstream_config ssl
```
jrbenny35, line 57: Why are we adding SSL? Just curious

app/Dockerfile
```diff
@@ -1,3 +1,15 @@
+FROM alpine:3.12.0 as file-loader
+
+# To preserve layer caching across machines which may have different local file properties
+# such as permissions, timestamps, etc, all files are copied into a container and their
+# permissions and timestamps are reset to consistent values
+# Credit: https://gist.github.com/kekru/8ac61cd87536a4355220b56ae2f4b0a9
+COPY . /app/
+RUN chmod -R 555 /app \
+ && chown -R root:root /app \
+ && find /app -exec touch -a -m -t 201512180130.09 {} \;
```
jrbenny35, line 10: some 1337 stuff here

jrbenny35 (approved):

jodyheavener (approved):
The specifics are a little over my head, but it works well for me :)

Makefile
jaredlockhart: Oh I had a `compose_build` step that would rebuild all the compose containers, and that would check for the ssl certs, but I realized we don't actually need the `compose_build` step at all since the only container we need to worry about rebuilding is the experimenter container, so I just took the `compose_build` out and lifted the `ssl` up here.

app/Dockerfile
jaredlockhart: giphy

PR #5950, opened 2021-07-14
fix #5949 chore(project): add makefile command to approve dependabot prs

### LINKED ISSUE #5949: Add makefile command to approve dependabot PRs
We can at least simplify the current dependabot workflow by using the [Github CLI](https://github.com/cli/cli) to approve all dependabot prs in one command.  Then maybe we can wrap that in a GH action.

```
gh pr list | grep dependabot | awk '{print $1}' | xargs -n1 gh pr review -a -b "@dependabot squash and merge"
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1487)

jaredlockhart:
Because

* Rather than approving each dependabot pr by hand, it'd be handy to just do them all at once with the github cli

This commit

* Adds a makefile command to approve all dependabot prs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,10 @@ bash: compose_build
 refresh: kill compose_build
 	$(COMPOSE) run -e SKIP_DUMMY=$$SKIP_DUMMY app bash -c '$(WAIT_FOR_DB) $(PYTHON_MIGRATE)&&$(LOAD_LOCALES)&&$(LOAD_COUNTRIES)&&$(LOAD_DUMMY_EXPERIMENTS)'
 
+dependabot_approve:
+	echo "Install and configure the Github CLI https://github.com/cli/cli"
+	gh pr list --author app/dependabot | awk '{print $$1}' | xargs -n1 gh pr review -a -b "@dependabot squash and merge"
+
 # integration tests
 integration_build: build_prod ssl
 	$(COMPOSE_INTEGRATION) build
--- a/contributing.md
+++ b/contributing.md
@@ -122,14 +122,27 @@ Dependencies are automatically updated by [Dependabot](https://dependabot.com/) 
 directly into GitHub.  Each week Dependabot will create a large number of individual PRs that update
 each dependency.  To merge those into main, use the following process:
 
-
 ### Merge Dependabot PRs
-1. Dependabot will create many individual PRs against the `main` branch, which has no
-branch protections and so each of those PRs can be merged automatically by approving each PR with the comment:
 
-        @dependabot squash and merge
+#### Manually
+1. Dependabot will create many individual PRs against the `main` branch, which must pass all CI checks and be approved before merging.  Approve each PR with the following comment and if they pass CI they will merge automatically:
+
+```
+@dependabot squash and merge
+```
 
 All done!
+
+#### Automatically
+1. Install and configure the [Github CLI](https://github.com/cli/cli)
+1. From your local Experimenter repo run
+
+```
+make dependabot_approve
+```
+
+### Failed Dependabot PRs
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -140,6 +140,10 @@ bash: compose_build
 refresh: kill compose_build
 	$(COMPOSE) run -e SKIP_DUMMY=$$SKIP_DUMMY app bash -c '$(WAIT_FOR_DB) $(PYTHON_MIGRATE)&&$(LOAD_LOCALES)&&$(LOAD_COUNTRIES)&&$(LOAD_DUMMY_EXPERIMENTS)'
 
+dependabot_approve:
+	echo "Install and configure the Github CLI https://github.com/cli/cli"
+	gh pr list | grep "chore(deps" | awk '{print $$1}' | xargs -n1 gh pr review -a -b "@dependabot squash and merge"
```
jodyheavener, line 145: Hmm, doesn't appear this is an [option](https://cli.github.com/manual/gh_pr_list#options)*, but it would be nice if were filtering by PRs that are passing CI, so we're _trying_ to merge failing PRs (I know it wouldn't allow us anyway, but still).

* Unless we make CircleCI apply a `passing` label to PRs that are passing, then we could filtering by label, but feels like more effort than it's worth

jaredlockhart: They can't merge if they don't pass so even if they're approved nothing will happen.

jaredlockhart: And this step can be done before they even run in CI.  And they'll often rebase themselves as others merge and re run CI so, I don't think worrying about whether they're passing when this is run matters.

pdehaan, line 145: Fascinating!
https://cli.github.com/manual/gh_pr_list

Looks like instead of the `grep "chore(deps"`, we could possibly do a safer `--author` query:

```sh
gh pr list --author app/dependabot
```

I was also trying to play with the `--json` and `--jq` flags to see if you could avoid `awk`, but not sure this is any better or nein.

```sh
gh pr list --author app/dependabot --state=closed --json number --jq ".[] | .number"
```

pdehaan:
Random Q: Are you/we planning on running this manually? Or making a [cron-GitHub action](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events) which runs daily/weekly? Adding this as a CircleCI job so every new PR silently triggers this autobot merge?

jaredlockhart:
@pdehaan I'm gonna try wrapping it in an action but if that doesn't work it can be run manually.

Makefile
jodyheavener: Yup. I mostly suggested this as a nice-to-have since someone not familiar with this process could go to the PR page and get confused seeing `@dependabot squash and merge` comments on failing PRs. Ignore me.

contributing.md
```diff
@@ -122,15 +122,24 @@ Dependencies are automatically updated by [Dependabot](https://dependabot.com/)
...
 
         @dependabot squash and merge
 
 All done!
 
+#### Automatically
+1. Install and configure the [Github CLI](https://github.com/cli/cli)
+1. From your local Experimenter repo run
+
+        make dependabot_approve
```
jodyheavener, line 138: Nit: excessive rendered indentation

jaredlockhart: This always confuses me.  It says 4 spaces in the doc but then it doesn't render properly?

jaredlockhart: test
    four spaces
        eight spaces

pdehaan: It's 4 spaces for a code block (if not using fences).
... But if you're in a list, it might need another 2-4 spaces indentation if you want the code block to be part of the bulleted list.

But yeah, code fences are always better/clearer (and support syntax highlighting, which is a nice bonus).

jaredlockhart: Right exactly

jodyheavener (approved):
Nice!

PR #5964, opened 2021-07-15
chore(deps): Security 'trim'

LZoog:
Because:
* Dependabot found that 'trim' has a security vulnerability

This commit:
* Updates the trim package to the recommended remediation version

Per our new [security warning docs](https://github.com/mozilla/experimenter/blob/main/contributing.md#security-warnings). 😎

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -11,6 +11,7 @@
   },
   "devDependencies": {},
   "resolutions": {
+    "trim": "^0.0.3",
     "postcss": "^7.0.36",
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
1) Okay cool so that process worked!
2) This should clear out the security warning
3) After this deploys, our local environments should rebuild from the remote cache instead of rebuilding locally.  Moment of truth 🤞 
4) Thanks @LZoog !

PR #5968, opened 2021-07-15
fix #5967 chore(project): set shell in makefile

### LINKED ISSUE #5967: Set SHELL=/bin/bash in Makefile
For the Makefile to work portably we need to explicitly set its shell to /bin/bash.  We can just stick

```
SHELL = /bin/bash
```

at the top

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1500)

jaredlockhart:
Because

* The makefile depends on /bin/bash to run correctly
* That is not the default on all environments

This commit

* Explicitly sets SHELL = /bin/bash in the Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+SHELL = /bin/bash
+
 WAIT_FOR_DB = /app/bin/wait-for-it.sh -t 30 db:5432 &&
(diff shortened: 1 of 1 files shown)


lmorchard (approved):
Yup, that seems to fix the error for me!

PR #5969, opened 2021-07-15
fixes #5954 feat(nimbus): group home page experiment tabs by all statuses

### LINKED ISSUE #5954: Add all statuses as directory view tabs
On the Nimbus directory page, we currently have the following tabs:

Active | Completed | Drafts

where Active has the three subheadings: Live, In Review, In Preview

Let's reorganize that into the following tabs in the following order:

Live | Review | Preview | Draft | Complete

with no subheadings under any tab.

The first column of each tabs table is currently labeled the same as the tab name so let's change the title of the first column to `Name`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1490)

jodyheavener:
Closes #5954 

This PR changes the directory/home page view to display for _all_ experiment lifecycle statuses, instead of grouping them together.

As a bonus, I also added a mechanism to remember which tab you were last on.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.test.tsx
@@ -107,12 +107,8 @@ function expectTableCells(testId: string, cellTexts: string[]) {
 describe("DirectoryTable", () => {
   it("renders as expected with default columns", () => {
     const experiments = [experiment];
-    render(<DirectoryTable title="Woozle Wuzzle" {...{ experiments }} />);
-    expectTableCells("directory-table-header", [
-      "Woozle Wuzzle",
-      "Owner",
-      "Feature",
-    ]);
+    render(<DirectoryTable {...{ experiments }} />);
+    expectTableCells("directory-table-header", ["Name", "Owner", "Feature"]);
     expectTableCells("directory-table-cell", [
       experiment.name,
       experiment.owner!.username,
@@ -124,17 +120,13 @@ describe("DirectoryTable", () => {
   });
 
   it("renders as expected without experiments", () => {
-    render(<DirectoryTable title="Sweet Pea" experiments={[]} />);
-    expectTableCells("directory-table-header", ["Sweet Pea"]);
-    expect(
-      screen.getByTestId("directory-table-no-experiments"),
-    ).toBeInTheDocument();
+    render(<DirectoryTable experiments={[]} />);
+    expect(screen.getByTestId("no-experiments")).toBeInTheDocument();
   });
 
   it("renders as expected with custom columns", () => {
     render(
       <DirectoryTable
-        title="Record Numbers"
         experiments={[experiment]}
         columns={[
           { label: "Cant think", component: DirectoryColumnTitle },
@@ -157,14 +149,9 @@ describe("DirectoryTable", () => {
 
 describe("DirectoryLiveTable", () => {
   it("renders as expected with custom columns", () => {
-    render(
-      <DirectoryLiveTable
-        title="Live Experiments"
-        experiments={[experiment]}
-      />,
-    );
+    render(<DirectoryLiveTable experiments={[experiment]} />);
     expectTableCells("directory-table-header", [
-      "Live Experiments",
+      "Name",
       "Owner",
       "Feature",
       "Enrolling",
@@ -186,11 +173,9 @@ describe("DirectoryLiveTable", () => {
 
 describe("DirectoryCompleteTable", () => {
   it("renders as expected with custom columns", () => {
-    render(
-      <DirectoryCompleteTable title="Completed" experiments={[experiment]} />,
-    );
+    render(<DirectoryCompleteTable experiments={[experiment]} />);
     expectTableCells("directory-table-header", [
-      "Completed",
+      "Name",
       "Owner",
       "Feature",
       "Started",
@@ -210,8 +195,8 @@ describe("DirectoryCompleteTable", () => {
 
 describe("DirectoryDraftsTable", () => {
   it("renders as expected with custom columns", () => {
-    render(<DirectoryDraftsTable title="Drafts" experiments={[experiment]} />);
-    expectTableCells("directory-table-header", ["Drafts", "Owner", "Feature"]);
+    render(<DirectoryDraftsTable experiments={[experiment]} />);
+    expectTableCells("directory-table-header", ["Name", "Owner", "Feature"]);
     expectTableCells("directory-table-cell", [
       experiment.name,
       experiment.owner!.username,
--- a/app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
@@ -71,70 +71,49 @@ interface Column {
 }
 
 interface DirectoryTableProps {
-  title: string;
   experiments: getAllExperiments_experiments[];
   columns?: Column[];
 }
 
 const DirectoryTable: React.FunctionComponent<DirectoryTableProps> = ({
-  title,
   experiments,
   columns: customColumns,
 }) => {
   const columns = customColumns || [
-    { label: title, component: DirectoryColumnTitle },
+    { label: "Name", component: DirectoryColumnTitle },
     { label: "Owner", component: DirectoryColumnOwner },
     { label: "Feature", component: DirectoryColumnFeature },
   ];
   return (
-    <div className="directory-table pb-2" data-testid="DirectoryTable">
-      <table className="table">
-        <thead>
-          <tr>
-            {experiments.length ? (
-              columns.map(({ label }, i) => (
+    <div className="directory-table pb-2 mt-4">
+      {experiments.length ? (
+        <table className="table" data-testid="DirectoryTable">
+          <thead>
+            <tr>
+              {columns.map(({ label }, i) => (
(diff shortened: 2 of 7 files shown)


jaredlockhart (changes requested):
Awesome!  Thank you @jodyheavener , so much easier to navigate.  The session thing is a neat idea but please pull that out of this pr and file a separate ticket for it, we explicitly wanted it to always land on `Live` on a fresh load.  We should look at the idea of storing the current tab separately.  I'd like it to persist in the URL, we can look at that separately.

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.test.tsx
```diff
@@ -105,14 +105,10 @@ function expectTableCells(testId: string, cellTexts: string[]) {
 }
 
 describe("DirectoryTable", () => {
-  it("renders as expected with default columns", () => {
+  fit("renders as expected with default columns", async () => {
```
pdehaan, line 108: Re: `it` => `fit`,

What dat about? I'm a bit unclear after https://jestjs.io/docs/api#testonlyname-fn-timeout about which tests will be run with that tweaked. If this `describe("DirectoryTable", ...)` suite has three `it` tests set, but one of them is `fit`, does that mean it will only run the 1 of the 3 tests now?

✅  `fit("renders as expected with default columns", async () => {...}`
❌  `it("renders as expected without experiments", () => {...}`
❌  `it("renders as expected with custom columns", () => {...}`

jodyheavener: Whoops! Leftover from testing. Good catch.

jaredlockhart (approved):
Yeyeyeyeyeye hypehypehypehype 🙏 🙏 🙏 🙏 🙏

app/tests/integration/nimbus/pages/home.py
```diff
@@ -7,7 +7,7 @@ class HomePage(Page):
     """Nimbus Home page."""
 
     _create_new_btn_locator = (By.CSS_SELECTOR, "#create-new-button")
-    _page_wait_locator = (By.CSS_SELECTOR, ".table")
+    _page_wait_locator = (By.CSS_SELECTOR, ".directory-table")
```
pdehaan, line 10: I might not be understanding the exact role of `data-testid`.
Above we have the following snippet:
```html
<table className="table" data-testid="DirectoryTable">
```

I don't know I'd expect this CSS selector to work if we're only setting a `.table` class, but then trying to use `.directory-table` here. Not sure if the selector should be something weird like `table.table[data-testid="DirectoryTable"]` or if just `[data-testid="DirectoryTable"]` works here (or heck, even just `.table` since we're now following The Highlander rules of "there can be only one").

But now that we've split the 3 sub-tables on live into separate tabs, the tests might need to be tweaked further if we were trying to do a lookup of experiments in "Review" or "Preview" states. /cc @jrbenny35

pdehaan (changes requested):
Preemptively requesting changes until the integration tests are greenish.

jaredlockhart:
@jodyheavener rebasing this on main may clear up the integration failure

jodyheavener:
@jrbenny35 @pdehaan I'm a little stuck on these integration test failures. It appears to be having trouble finding the element to indicate the page is loaded. This was previously `.table`, but since that element is now conditionally rendered I switched it to `.directory-table`, an element that is always visible when the experiment request has finished. It is unable to find this element. I even tried using `#PageHome-page`, an element that is immediately visible, as the selector just to see if it could find that, with no luck. Kinda stumped 🤔

pdehaan:
Ah, curious, so we should always have a `.directory-table` wrapper div, but don't always have a `.table` since we now conditionally render it based on `experiments.length`, and otherwise render `<p data-testid="no-experiments">No experiments found.</p>`.

It looks like the current `integration_nimbus` error is:

```
nimbus/test_e2e_create_experiment.py:118: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
nimbus/pages/home.py:23: in tables
    self.wait.until(EC.presence_of_element_located(_table_locator))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
```

https://github.com/mozilla/experimenter/blob/568f4cf35a8c86ac10ba44e42d87777e6392e43b/app/tests/integration/nimbus/pages/home.py#L20-L25

So, re: `_table_locator = (By.CSS_SELECTOR, ".active .directory-table .table")`, I see the `.directory-table` and [conditional] `.table` class names below:

```jsx
    <div className="directory-table pb-2 mt-4">
      {experiments.length ? (
        <table className="table" data-testid="DirectoryTable">
```

I'll have to poke around a bit more and see where the `.active` className is being added. I'm guessing that's the active section tab, which I don't expect changed behavior (or I didn't see it in this PR).

jodyheavener:
@pdehaan Oh that's interesting, locally my tests fail with:

```
nimbus/test_e2e_create_experiment.py:236:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
nimbus/pages/home.py:16: in wait_for_page_to_load
    self.wait.until(EC.visibility_of_element_located(self._page_wait_locator))
```

pdehaan:
> ```
> nimbus/pages/home.py:16: in wait_for_page_to_load
>    self.wait.until(EC.visibility_of_element_located(self._page_wait_locator))
> ```

Curious, since that code is now:

```python
_page_wait_locator = (By.CSS_SELECTOR, ".directory-table")
```

And I thought that `<div className="directory-table pb-2 mt-4">` always existed. I might have to defer to @jrbenny35's expertise here in adding some pauses or screenshots to help debug locally.

jaredlockhart:
Eyyyyyy green finally 🍏 📗 🟢 💚 🥗

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -124,7 +124,7 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
     # Check it's live
     home = HomePage(selenium, base_url).wait_for_page_to_load()
     live_experiments = home.tables[0]
-    assert "live experiments" in live_experiments.table_name.lower()
+    assert home.active_tab_text, "Live"
```
jrbenny35, line 127: This should check that the active tab has "live experiments" within it, currently, this assert will always pass.

See: https://www.w3schools.com/python/ref_keyword_assert.asp

jrbenny35 (changes requested):

app/tests/integration/nimbus/test_e2e_create_experiment.py
jodyheavener: @jrbenny35 Hmm, would it always pass? I want to assert that the tab is selected (contains the `.active` class). So if "Review" was selected the `active_tab_text` would not equal `"Live"`. I changed it to this since we removed the element/text that previously displayed "live experiments"

jodyheavener: Edit: oh I see what you meant about the usage of `assert`. Updated that, but I'm still checking for "Live" in the tab text and since "Live experiments" no longer exists.

pdehaan (commented):
Great job!
Just 2 small questions.

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.test.tsx
```diff
@@ -105,14 +105,10 @@ function expectTableCells(testId: string, cellTexts: string[]) {
 }
 
 describe("DirectoryTable", () => {
-  it("renders as expected with default columns", () => {
+  it("renders as expected with default columns", async () => {
```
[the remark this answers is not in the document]
jodyheavener: Whoops, must have thought I was going to add an async method here and then didn't. Will update.

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -71,70 +71,49 @@ interface Column {
...
+          <thead>
+            <tr>
+              {columns.map(({ label }, i) => (
                 <th
-                  className={`border-top-0 ${
-                    i === 0 ? "font-weight-bold" : "font-weight-normal"
-                  }`}
+                  className="border-top-0"
                   key={label}
                   data-testid="directory-table-header"
```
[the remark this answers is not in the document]
jodyheavener: > Q: does a data-testid have to be unique?

Nope! You can [use](https://testing-library.com/docs/queries/about/#types-of-queries) the `queryAllBy*` / `findAllBy*` / `getAllBy*` methods to look up multiple elements of the same criteria. In fact, we are doing [exactly that](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.test.tsx#L100) in this test file.

jrbenny35 (approved):

pdehaan (approved):
:shipit:

PR #5973, opened 2021-07-16
fix #5972 chore(project): revert cache docker layers in dockerhub 😢 

### LINKED ISSUE #5972: Revert docker layer remote caching
After much exhaustive research I have determined there's no good way to share docker layers between CircleCI and a local machine on OSX.  Which means we can never achieve the dream of rebuilding containers in Circle and pulling them in locally.  Except we can.  The dev image ONLY contains system/python/node packages, it doesn't actually contain the code, that gets volume mapped in.  So rather than attempting to rebuild locally with remote caching, we should just build the whole image in circle and ONLY use the circle built image locally, never rebuild locally, unless you explicitly modify the package manifests as part of a PR.  But for PRs that don't make package changes, which is most PRs, we can just reuse the remotely build dev image.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1503)

jaredlockhart:
Becuase

* After digging into it I find that even if the file contents and visible metadata (owner, group, timestamp) are identical between circle and local, docker will generate different layer hashes, which means the layers built on circle won't be reused locally
* Something else about the file metadata/host environment is contributing to different hashing
* So while layers built on circle can be reused within circle, they can't be reused locally, and we already have shared layer caching within circle using the circle docker build cache

This commit

* This reverts commit 4c0df6b262508a21fa67c1251b79eb13e2f964d8.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,24 @@
 version: 2.1
 jobs:
+  build:
+    machine:
+      docker_layer_caching: true
+      image: ubuntu-2004:202104-01 # Ubuntu 20.04, Docker v20.10.6, Docker Compose v1.29.1
+    resource_class: large
+    working_directory: ~/experimenter
+    steps:
+      - run:
+          name: Docker info
+          command: docker -v
+      - run:
+          name: Docker compose info
+          command: docker-compose -v
+      - checkout
+      - run:
+          name: Build docker images
+          command: |
+            make build_prod
+
   check:
     machine:
       docker_layer_caching: true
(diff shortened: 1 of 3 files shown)


jrbenny35 (approved):
Thou fought bravely but the whale twas too great

jaredlockhart:
Screen Shot 2021-07-16 at 11 09 09 AM

jaredlockhart:
Okay, new plan.  While we can't reuse the layers from circle locally, they are reused within circle, and it pretty significantly speeds up our overall circle runs, so let's keep it.  Closing this.

PR #5982, opened 2021-07-16
fix #5981 chore(nimbus): remove deprecated update celery task

### LINKED ISSUE #5981: Remove deprecated celery task reference
We recently removed `experimenter.kinto.tasks.nimbus_check_experiments_are_updated` but neglected to remove it from the celery scheduler.  We should take this out.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1508)

jaredlockhart:
Because

* We recently refactored the live update/pause kinto tasks
* We removed the separate update task and folded that logic into the check collection task
* We neglected to remove the reference to it from the beat scheduler

This commit

* Removes the deprecated reference


jodyheavener (approved):
🧁

Issue #5976, opened 2021-07-16
Intermittent test failure: TestTask.test_generate_reportlogs_creates_for_both_types_of_changelogs


jaredlockhart:
https://app.circleci.com/pipelines/github/mozilla/experimenter/10370/workflows/01bf0ece-816f-4708-97b3-5cc74eb45b84/jobs/30594

```
=================================== FAILURES ===================================
____ TestTask.test_generate_reportlogs_creates_for_both_types_of_changelogs ____
[gw0] linux -- Python 3.9.6 /usr/local/bin/python3

self = <experimenter.reporting.tests.test_tasks.TestTask testMethod=test_generate_reportlogs_creates_for_both_types_of_changelogs>

    def test_generate_reportlogs_creates_for_both_types_of_changelogs(self):
        mock_response_data = []
        mock_response = mock.Mock()
        mock_response.json = mock.Mock()
        mock_response.json.return_value = mock_response_data
        mock_response.raise_for_status = mock.Mock()
        mock_response.raise_for_status.side_effect = None
        mock_response.status_code = 200
        self.mock_normandy_requests_get.return_value = mock_response
        ExperimentFactory.create_with_status(target_status=Experiment.STATUS_LIVE)
        NimbusExperimentFactory.create_with_lifecycle(
            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE
        )
        self.assertEqual(ReportLog.objects.count(), 0)
        generate_reportlogs()
        self.assertEqual(
            ReportLog.objects.filter(experiment_type__startswith="Normandy").count(), 5
        )
        self.assertEqual(
            ReportLog.objects.filter(experiment_type__startswith="Nimbus").count(), 3
        )
        self.assertEqual(ReportLog.objects.count(), 8)
    
        ExperimentChangeLogFactory.create(changed_on=timezone.now())
    
        self.assertEqual(ExperimentChangeLog.objects.count(), 6)
        generate_reportlogs()
    
>       self.assertEqual(ReportLog.objects.count(), 9)
E       AssertionError: 8 != 9

experimenter/reporting/tests/test_tasks.py:204: AssertionError
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1505)

PR #5985, opened 2021-07-16
fix #5983 bug(nimbus): modify timing on timeout integration test

### LINKED ISSUE #5983: Intermittent integration test failure: test_create_new_experiment_remote_settings_timeout
Seen locally:

```
______________ test_create_new_experiment_remote_settings_timeout ______________                                                                                                                                                
>           raise AssertionError("Timeout text was never shown.")
E           AssertionError: Timeout text was never shown.

nimbus/test_e2e_create_experiment.py:286: AssertionError
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1510)

jaredlockhart:
Because

* With the settings we shipped, timeout=60s, refresh=30s, wait=45*2=90s, we expect that after two invocations of the celery task the experiment will be marked as timed out and 90s gives us that extra 30s to wait, but I've observed that since the timeout calculation is detached from the celery schedule, it may occur that on the second invocation it can be milliseconds below the timeout point and so it won't timeout until the third invocation, and since that's 90s which butts up right against the 90s the test waits it can randomly happen just after the test ends, I verified that the test did in fact end just before the experiment was marked as timed out

This commit

* Changes the integration test timing to timeout=40s, refresh=20s, wait=60*2=120s which should give enough time for three invocations of the task to occur


app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -11,7 +10,7 @@
 
 @pytest.mark.nondestructive
 def test_create_new_experiment(selenium, base_url):
-    experiment_name = f"name here remote {random.randint(0, 100)}"
+    experiment_name = "test_create_new_experiment"
```
jaredlockhart, line 13: Oh and I changed the names from random to the name of the test because I couldn't figure out which experiment was created by which test when inspecting in VNC

pdehaan: Lovin it! And seems like potentially fewer risks of an experiment naming collision if `random.randomint` ever happened to return the same 0..100 value for different tests in a run.

Does Python/Pytest have a way to get the current function's name from within a function?
I found https://stackoverflow.com/questions/17726954/py-test-how-to-get-the-current-tests-name-from-the-setup-method; but maybe not worth the hassle.

pdehaan (commented):
the "Run tests and linting" CI error seems&hellip; unrelated.

```
>       self.assertEqual(ReportLog.objects.count(), 9)
E       AssertionError: 8 != 9

experimenter/reporting/tests/test_tasks.py:204: AssertionError

✂️ ✂️ ✂️ 

=========================== short test summary info ============================
FAILED experimenter/reporting/tests/test_tasks.py::TestTask::test_generate_reportlogs_creates_for_both_types_of_changelogs
=========== 1 failed, 882 passed, 326 warnings in 160.17s (0:02:40) ============
INFO 2021-07-16 19:55:48,256 raven.contrib.django.client.DjangoClient Raven is not configured (logging is disabled).
```

And now it looks like the CI builds are re-running and I can't see what the Nimbus integration test error was.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jaredlockhart: Nah no need to be clever about it.

jaredlockhart:
> the "Run tests and linting" CI error seems… unrelated.
> 
> ```
> >       self.assertEqual(ReportLog.objects.count(), 9)
> E       AssertionError: 8 != 9
> 
> experimenter/reporting/tests/test_tasks.py:204: AssertionError
> 
> ✂️ ✂️ ✂️ 
> 
> =========================== short test summary info ============================
> FAILED experimenter/reporting/tests/test_tasks.py::TestTask::test_generate_reportlogs_creates_for_both_types_of_changelogs
> =========== 1 failed, 882 passed, 326 warnings in 160.17s (0:02:40) ============
> INFO 2021-07-16 19:55:48,256 raven.contrib.django.client.DjangoClient Raven is not configured (logging is disabled).
> ```
> 
> And now it looks like the CI builds are re-running and I can't see what the Nimbus integration test error was.

@pdehaan the `make check` failure was https://github.com/mozilla/experimenter/issues/5976

the integration nimbus failure was https://github.com/mozilla/experimenter/issues/5986

jaredlockhart:
@pdehaan and you can see past circle runs here https://app.circleci.com/pipelines/github/mozilla/experimenter

jrbenny35:
This is nice, thanks for that knowledge of  the tasks timing.

jrbenny35 (approved):

Issue #5847, opened 2021-06-25
Legacy/nimbus integration tests failing with HTTP timeout


data-sync-user:
I've seen this come up a lot the last few days, the legacy integration tests fail when the legacy files weren't touched.
See:

* https://app.circleci.com/pipelines/github/mozilla/experimenter/10081/workflows/899234d6-7660-4982-80c8-61c448676ba7/jobs/29424
* https://app.circleci.com/pipelines/github/mozilla/experimenter/10083/workflows/22cd466a-3e57-4571-884e-e73dd27131b9/jobs/29437

Edit: also seeing this on nimbus integration tests at least here:
* https://app.circleci.com/pipelines/github/mozilla/experimenter/10085/workflows/690282f5-ba32-4c8b-a51a-0660573c576b/jobs/29442

Maybe we can just increase `COMPOSE_HTTP_TIMEOUT`?

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-1450)

data-sync-user:
➤ Jared Lockhart commented:

I think this is resource constraint hiccuping in circle, I don’t think there’s anything we can change to help this, 120s for the compose timeout is already more than reasonable.  If we see this in the future just try rerunning from failed should clear it up.  I’m gonna close this and if it becomes a more persistent problem we’ll have to dig in more.

PR #5987, opened 2021-07-16
fix #5986 chore(project): increase COMPOSE_HTTP_TIMEOUT

### LINKED ISSUE #5986: Increase COMPOSE_HTTP_TIMEOUT
We've seen a lot of intermittent test failures on circle complaining about `COMPOSE_HTTP_TIMEOUT` being too low so let's take docker at its word and try increasing it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1511)

jaredlockhart:
Because

* We've seen a lot of intermittent failures in circle complaining about COMPOSE_HTTP_TIMEOUT being too low

This commit

* Let's try increasing it and see if that helps


pdehaan (approved):
Can confirm, 16.67 minutes is definitely larger than 2 minutes.

Wondering if we can add any logging anywhere in the system to capture how long this COMPOST_HTTP_TIMEOUT is actually taking. Clearly it occasionally wants more than 2 minutes, but it might be interesting to chart over time to see if it runs anywhere between 30 seconds and 8 minutes.

Merge away! :+1:

LZoog:
Huh, did we want to dig in more [per this comment](https://github.com/mozilla/experimenter/issues/5847#issuecomment-875041282)? Or maybe you did already👀

jaredlockhart:
> Huh, did we want to dig in more [per this comment](https://github.com/mozilla/experimenter/issues/5847#issuecomment-875041282)? Or maybe you did already👀

I still think that's true so I'm not sure if this is actually going to help, but it's happening frequently enough and we can't reasonably increase the resources on these nodes so maybe we do actually just have to increase this, worth a shot.

PR #5989, opened 2021-07-16
chore(nimbus): disable intermittently failing test

jaredlockhart:
Because

* test_generate_reportlogs_creates_for_both_types_of_changelogs is intermittently failing frequently in CI

This commit

* Disables it until a fix can land so CI isn't blocked


tiftran (approved):
should be chore(reporting) i think?

jaredlockhart:
Oh!  Actually maybe it should be `project` since it's project wide.... is `reporting` one of our commit names?

jaredlockhart:
SO IT IS

reporting: Anything scoped to Reporting

PR #6002, opened 2021-07-19
Add proposed duration changes.

### LINKED ISSUE #6000: All checks for proposed enrollment/duration should check for not None
Every time we check for proposed_duration/enrollment, we should check that it is not None rather than simply truthy.  This will allow us to use `0` as a valid time during integration testing.  Example

No

```py
if self.proposed_duration:
```

Yes

```py
if self.proposed_duration is not None:
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1514)

jrbenny35:
Closes #6000


jaredlockhart (approved):
giphy

PR #5362, opened 2021-06-01
fixes #5355 feat(nimbus): display gql error on homepage directory

jodyheavener:
Closes #5355

This PR updates the experiment directory page to display a GQL request error if it occurs.

PR #6008, opened 2021-07-20
fix #5765 feat(nimbus): Display more detailed error output across pages in alert

### LINKED ISSUE #5765: Display GQL error on all pages
In #5362 we displayed the GQL error at the top of the home directory page.

For dev purposes it would be nice to see the error immediately on other pages instead of seeing a "General application error" and having to open dev tools > Network > Response to see what the problem is.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1410)

LZoog:
fixes #5765 

Because:
* We want to display more detailed error messaging where an error would break the app for handiness across envs

This commit:
* Adds a new ApolloErrorAlert component to parse through various ApolloErrors and display helpful output, displays this if an error occured when querying an experiment in PageHome and AppLayoutWithExperiment, as well as if the config query errors
* Adjusts AppErrorBoundary to display an error.message if there is one

---

This was slightly trickier than expected. I wanted to create a more universal handler so we wouldn't need to `return <ApolloErrorAlert ...` every time, but:
* using an ApolloError version [ErrorBoundary](https://reactjs.org/docs/error-boundaries.html) would be awkward since you have to rethrow the error for it to catch (plus it seems one `ErrorBoundary` is typical per the docs)
* we wouldn't want to do all this error message parsing in the `AppErrorBoundary` since any error within that render would be uncaught
* using [`link`](https://www.apollographql.com/docs/react/api/link/introduction/) for our errors is hit per GQL mutation and doesn’t appear to be a way to access that anyway inside components (see [this related issue](https://github.com/apollographql/apollo-link/issues/1022))

Added Storybook states to account for the various kinds of `ApolloError`. ~Also, might change the name to `ApolloErrorAlert`. 🤔~ Updated to `ApolloErrorAlert.

👉 [Storybook link](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/82490fdc534d01381fc269672add106e4ff4d154/nimbus-ui/index.html?path=/story/components-apolloerroralert--with-error)

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.stories.tsx
@@ -1,0 +1,33 @@
+import React from "react";
+import { ApolloErrorAlert } from ".";
+import {
+  GQL_ERROR,
+  GQL_ERROR_MULTIPLE,
+  NETWORK_SERVER_ERROR,
+  NETWORK_SERVER_PARSE_ERROR,
+} from "./mocks";
+
+export default {
+  title: "components/ApolloErrorAlert",
+  component: ApolloErrorAlert,
+};
+
+export const withError = () => (
+  <ApolloErrorAlert
+    error={new Error("Uh oh. You made the app crashy crash.")}
+  />
+);
+
+export const withNetworkServerError = () => (
+  <ApolloErrorAlert error={NETWORK_SERVER_ERROR} />
+);
+
+export const withNetworkServerParseError = () => (
+  <ApolloErrorAlert error={NETWORK_SERVER_PARSE_ERROR} />
+);
+
+export const withGQLError = () => <ApolloErrorAlert error={GQL_ERROR} />;
+
+export const withMultipleGQLErrors = () => (
+  <ApolloErrorAlert error={GQL_ERROR_MULTIPLE} />
+);
--- a/app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.tsx
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { ApolloError, ServerError, ServerParseError } from "@apollo/client";
+import { GraphQLError } from "graphql";
+import React from "react";
+import { Alert } from "react-bootstrap";
+import sentryMetrics from "../../services/sentry";
+
+export type NetworkError = ApolloError["networkError"];
+
+// Type guards - If the `networkError` is a `ServerError`, we want to display its `result.errors`
+// array containing extra info
+function isNetworkServerError(
+  networkError: NetworkError,
+): networkError is ServerError {
+  return (networkError as ServerError)?.result !== undefined;
+}
+
+// If the `networkError` is a `ServerParseError`, we want to display `bodyText` and `statusCode`
+function isNetworkServerParseError(
+  networkError: NetworkError,
+): networkError is ServerParseError {
+  return (
+    (networkError as ServerParseError)?.bodyText !== undefined &&
+    (networkError as ServerParseError)?.statusCode !== undefined
+  );
+}
+
+export const ApolloErrorAlert = ({ error }: { error: ApolloError | Error }) => {
+  const { graphQLErrors = [], networkError } = error as ApolloError;
+  const networkServerErrors =
+    networkError && isNetworkServerError(networkError)
+      ? networkError.result.errors
+      : [];
+
+  // Report Errors and NetworkErrors, graphQLErrors should be captured server-side
+  if (networkError || error instanceof Error) {
+    sentryMetrics.captureException(error);
+  }
+
+  let heading = "General GQL Error";
+  if (graphQLErrors.length) {
+    heading = "GraphQL Apollo Error";
+  } else if (networkError) {
+    heading = "Network Error";
+  }
+
+  return (
+    <Alert variant="danger" data-testid="apollo-error-alert">
+      <Alert.Heading>{heading}</Alert.Heading>
+      <p className="mb-6">Something went wrong. Please try again later.</p>
+      <hr />
+      {error.message && (
+        <p>
+          <b>Message:</b> {error.message}
+        </p>
+      )}
+      {graphQLErrors.length > 0 &&
+        graphQLErrors.map((gqlError: GraphQLError) => (
+          <p key={gqlError.message}>
+            <b>graphQL error:</b> {gqlError.message}
+          </p>
+        ))}
+      {networkServerErrors.length > 0 &&
+        networkServerErrors.map((serverError: ServerError) => (
+          <p key={serverError.message}>
+            <b>Network error:</b> {serverError.message}
+          </p>
+        ))}
+      {networkError && isNetworkServerParseError(networkError) && (
+        <>
+          <p>
+            <b>Status code:</b> {networkError.statusCode}
+          </p>
+          <p>
+            <b>Body text:</b> {networkError.bodyText}
+          </p>
+        </>
+      )}
+    </Alert>
+  );
+};
+
+export default ApolloErrorAlert;
--- a/app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/mocks.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/mocks.tsx
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { ApolloError } from "@apollo/client";
+import { GraphQLError } from "graphql";
+
+export const NETWORK_SERVER_ERROR = new ApolloError({
+  graphQLErrors: [],
+  networkError: {
+    name: "ServerError",
+    message: "Response not successful: Received status code 400",
+    response: new Response(),
+    statusCode: 400,
+    result: {
+      errors: [
+        {
+          message: 'Cannot query field "WOOT" on type "NimbusExperimentType".',
+          locations: [
+            {
+              line: 3,
+              column: 5,
+            },
+          ],
+        },
+      ],
+    },
+  },
+});
+
+export const NETWORK_SERVER_PARSE_ERROR = new ApolloError({
+  graphQLErrors: [],
+  networkError: {
+    name: "ServerParseError",
+    message:
+      "JSON.parse: unexpected character at line 1 column 1 of the JSON data",
+    response: new Response(),
+    statusCode: 404,
+    bodyText:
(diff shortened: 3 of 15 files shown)


app/experimenter/nimbus-ui/src/components/AppErrorBoundary/index.tsx
```diff
@@ -26,17 +26,21 @@ class AppErrorBoundary extends React.Component {
...
-      <Alert.Heading>General application error</Alert.Heading>
-      <p>Something went wrong. Please try again later.</p>
-    </Alert>
-  );
-};
+export const AppErrorAlert = ({ error }: { error: Error }) => (
+  <Alert variant="danger" data-testid="error-loading-app">
+    <Alert.Heading>General application error</Alert.Heading>
+    <p>Something went wrong. Please try again later.</p>
+    {error?.message && (
```
LZoog, line 37: Seems like we want to keep any logic in here _super_ minimal since anything that errors in the furthest most outer error boundary will be unhandled, but this seemed safe enough to me.

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.test.tsx
```diff
@@ -32,6 +33,19 @@ describe("AppLayoutWithExperiment", () => {
     await screen.findByText("Loading...");
   });
 
+  // TODO: when this test is present, there's an Error: Uncaught [TypeError: stopPolling is not a function]
```
LZoog, line 36: Will address this before merging

app/experimenter/nimbus-ui/src/components/GQLErrorAlert/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+): networkError is ServerParseError {
+  return (
+    (networkError as ServerParseError)?.bodyText !== undefined &&
+    (networkError as ServerParseError)?.statusCode !== undefined
+  );
+}
+
+export const GQLErrorAlert = ({ error }: { error: ApolloError }) => {
+  // without this, we only have access to error.message
+  const allErrors = JSON.parse(JSON.stringify(error));
```
LZoog, line 33: This was a bit awkward, I got it from [this comment](https://github.com/apollographql/apollo-client/issues/1496#issuecomment-290303638).

Seems like an alternative is to `.catch` the error after every query and deconstruct `networkError`/`graphQLErrors`/`error` there and then send those in instead, but this seemed slightly preferable to me.

app/experimenter/nimbus-ui/src/components/GQLErrorAlert/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { networkError } = allErrors as ApolloError;
+  const networkServerErrors =
+    networkError && isNetworkServerError(networkError)
+      ? networkError.result.errors
+      : [];
+
+  // we want to report Errors and NetworkErrors
+  // graphQLErrors should be captured on the server-side
+  if (networkError || error instanceof Error) {
+    sentryMetrics.captureException(error);
```
LZoog, line 44: Prior to this change, if we [caught the error and returned anything](https://github.com/mozilla/experimenter/pull/5362/files#diff-77ef9304606052b73cebb4b85f40b658b05bed9a51065504a32148a563aa0de7R25), it wasn't reported to Sentry since it didn't bubble up to `AppErrorBoundary` any longer.
 
This seemed right to me, followed some of what we laid out in [`fxa-settings` docs](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings#gql-error-handling). Let me know otherwise. 😎

app/experimenter/nimbus-ui/src/components/GQLErrorAlert/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    <Alert variant="danger" data-testid="gql-error-alert">
+      <Alert.Heading>{heading}</Alert.Heading>
+      <p className="mb-6">Something went wrong. Please try again later.</p>
+      <hr />
+      {error.message && (
+        <p>
+          <b>Message:</b> {error.message}
+        </p>
+      )}
+      {graphQLErrors.length > 0 &&
```
LZoog, line 64: Without these explicit `> 0` checks, sometimes a "0" would render.

app/experimenter/nimbus-ui/src/components/GQLErrorAlert/mocks.tsx
```diff
@@ -0,0 +1,72 @@
...
+  message: "Response not successful: Received status code 400",
+};
+
+export const NETWORK_SERVER_PARSE_ERROR = {
+  graphQLErrors: [],
+  networkError: {
+    name: "ServerParseError",
+    response: {},
+    statusCode: 404,
+    bodyText:
```
LZoog, line 34: I forced each of these errors to happen and copied them exactly as-is to have mocks matching exactly what we should receive in practice. Thought about taking a chunk of this one out, but🤷‍♀️

app/experimenter/nimbus-ui/src/components/GQLErrorAlert/mocks.tsx
```diff
@@ -0,0 +1,72 @@
...
+  },
+  message:
+    "JSON.parse: unexpected character at line 1 column 1 of the JSON data",
+};
+
+export const NETWORK_ERROR = {
+  graphQLErrors: [],
+  networkError: {
+    name: "Error",
```
LZoog, line 44: I may have been able to check for this `name` string instead of the [type guards functions](https://github.com/mozilla/experimenter/pull/6008/files#diff-ab35b3e1a6dc836a528565f1ed49e372648c894bb4ab2164fda7b4c92bfb13abR13-R29), but it seems like we wanted this to handle regular `Error` instances too and the typeguards made TS happy.

LZoog:
Oh strange, seems like CI found one test failure I wasn't seeing locally, will look tomorrow AM!

LZoog:
OK, this should be good to go now. Sorry for the "Outdated" comments, they're still relevant if I didn't resolve them; I changed the name from `GQLErrorAlert` to `ApolloErrorAlert`.

jaredlockhart:
> Oh strange, seems like CI found one test failure I wasn't seeing locally, will look tomorrow AM!

If you find a new intermittent don't forget to file a ticket 👍

LZoog:
> If you find a new intermittent don't forget to file a ticket 👍

Will do, this was actually one inside the Summary page that I was able to see/fix this morning (still not sure why I didn't see it yesterday, but it was a quick fix).

app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    (networkError as ServerParseError)?.bodyText !== undefined &&
+    (networkError as ServerParseError)?.statusCode !== undefined
+  );
+}
+
+export const ApolloErrorAlert = ({ error }: { error: ApolloError }) => {
+  // without this, we only have access to error.message
+  const allErrors = JSON.parse(JSON.stringify(error));
+  const { graphQLErrors = [] } = allErrors as ApolloError;
+  const { networkError } = allErrors as ApolloError;
```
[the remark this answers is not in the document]
lmorchard: Fell down a rabbit hole with this a bit and I think it maybe needs a little type juggling?

https://github.com/mozilla/experimenter/compare/issues/5765...lmorchard:issues/5765

lmorchard (approved):
Works for me and looks good... Just had one mystery with that `JSON.parse(JSON.stringify())` line that I think *maybe* isn't necessary after a few tweaks:

https://github.com/mozilla/experimenter/compare/issues/5765...lmorchard:issues/5765

app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.tsx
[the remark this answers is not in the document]
jodyheavener: Hahaha I also fell down this rabbit hole a bit, went the route of creating the classes https://gist.github.com/jodyheavener/c3becc6cf9ed7e7a79960f4477f1541c

app/experimenter/nimbus-ui/src/components/AppErrorBoundary/index.tsx
```diff
@@ -26,17 +26,21 @@ class AppErrorBoundary extends React.Component {
...
+      <>
+        <hr />
+        <small>{error.message}</small>
```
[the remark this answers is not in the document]
LZoog: Good call though, I'd rather eliminate as much risk as possible of something erroring inside what catches all our unhandled errors😬

app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/mocks.tsx
```diff
@@ -0,0 +1,72 @@
...
...     <li>\n            \n                ^admin/\n                \n            \n          </li>\n        \n          <li>\n            \n                ^experiments/\n                \n            \n          </li>\n        \n          <li>\n            \n                ^nimbus/\n                [name=\'nimbus-list\']\n            \n          </li>\n        \n          <li>\n            \n                ^nimbus/(?P&lt;slug&gt;[\\w-]+)/\n                [name=\'nimbus-detail\']\n            \n          </li>\n        \n          <li>\n            \n                ^$\n                [name=\'home\']\n            \n          </li>\n        \n          <li>\n            \n                ^404/\n                \n            \n          </li>\n        \n      </ol>\n      <p>\n        \n          The current path, <code>graphql</code>,\n        \n        didn’t match any of these.\n      </p>\n    \n  </div>\n\n  <div id="explanation">\n    <p>\n      You’re seeing this error because you have <code>DEBUG = True</code> in\n      your Django settings file. Change that to <code>False</code>, and Django\n      will display a standard 404 page.\n    </p>\n  </div>\n</body>\n</html>\n',
```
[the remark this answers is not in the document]
LZoog: Heh right, I mention that [here](https://github.com/mozilla/experimenter/pull/6008#discussion_r672724465) 😆

app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.tsx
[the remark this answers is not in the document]
LZoog: Ha, you're both so great. I spent a little bit of time on this as well and then just threw my arms up and went with what worked.

Though, WTF. When I was trying to _console.log_ the error to grab the mock...

> And peeked at the source for ApolloError and can't see how or why graphQLErrors or networkError would be inaccessible otherwise.

... I'd run into this, in which case I was also not sure why I couldn't access it. I've just realized that If I simply replace:

```
  const allErrors = JSON.parse(JSON.stringify(error));
  const { graphQLErrors = [] } = allErrors as ApolloError;
  const { networkError } = allErrors as ApolloError;
```

with what we'd expect would work,

```
  const { graphQLErrors = [], networkError } = error;
```

It _does_ seem to work when the error is thrown in development, for at least two errors I've tried so far.

I'll refactor this a tad knowing this works in practice and see if there's other TS wins I can grab from your branch @lmorchard, thanks for looking. 🙂 Also when purposefully throwing an error checking this, I see if there's multiple errors, perhaps the UI can be adjusted slightly (even if just to remove "errors" plural):

image

I think I'll still show both `message` and the GQL errors individually since I have no idea if those will differ with some kinds of errors.

>  This passes, but in this I removed the Network Error one, since I was confused by it (it looks like it's just an Error instance, but it also has other properties, but Apollo doesn't have a corresponding type for it, so idk where it comes from)

Wouldn't this just be `ApolloError` since `networkError` can be an `Error` (specifically, Error | ServerParseError | ServerError | null`)? On a similar note, I'm actually not sure if a regular ol' `Error` instance will ever be passed into this new component. I figured we could protect against it anyway just in case🤷‍♀️

app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const withError = () => (
+  <ApolloErrorAlert
+    error={new Error("Uh oh. You made the app crashy crash.") as ApolloError}
+  />
+);
+
+export const withNetworkServerError = () => (
+  <ApolloErrorAlert error={NETWORK_SERVER_ERROR as unknown as ApolloError} />
```
pdehaan, line 22: Q: Instead of casting `as unknown as ApolloError` in multiple files is it possible to define the data types in the ./mocks file when you're exporting the `const`s?

LZoog: Heh yeah good question, I'm copying what [Les did here](https://github.com/mozilla/experimenter/compare/issues/5765...lmorchard:issues/5765#diff-7eb756946081987e4b6fdab1479633ea511c0d72edcd4a69cb86ddb51e5b975dR30) which is basically what you suggested 😎

LZoog:
Alrighty, made a couple changes based on comments. I'll merge at the end of my day if there aren't any more. 🙂

app/experimenter/nimbus-ui/src/components/ApolloErrorAlert/index.tsx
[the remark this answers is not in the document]
LZoog: Pushed some changes and did end up using `const { graphQLErrors = [], networkError } = error as ApolloError;` but that's just to coerce `Error` instances into `ApolloError` type. All things are working properly for me in Storybook and actually triggering the errors in practice.

I'm facepalming that `JSON.parse(JSON.stringify(error))` was only needed to console log values😂 And then, they weren't even entirely correct as `ApolloError` because creating a new Apollo error shuffles around the placement of the `.message`, so, `= new ApolloError(` was def the way to go in those mocks.

PR #6017, opened 2021-07-21
Fix #6016: fix(visualization): Make results page work with outcomes.

### LINKED ISSUE #6016: Make results page work with outcomes
There are some issues that came up with the first end-to-end run using outcomes. Essentially, the results page was initially assuming that a given outcome would have at least 1 conversion metric associated with it with the suffix `ever_used`. There was an implicit assumption that primary metrics would always be these `ever_used` conversion metric types.

However, we now have an experiment in production (https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment-release/results) where there is no conversion `ever_used` metric associated with it and, in theory, there could be an unlimited number of metrics for an outcome. But the way the highlights and results tables up top should work is to only display 1-2 default metrics for a given outcome. The rest would be shown in further detail down below.

Furthermore, there is some missing outcomes data not being sent to the frontend such as slug, description, and friendly name for the metrics associated with an outcome. These are needed 

We can address these issues by doing the following:

1) Don't show outcome metrics in the hightlights or results table if they are not default. Right now, this means not showing any at all since the implementation in Jetstream for default metrics is not yet merged (https://github.com/mozilla/jetstream/pull/821/files)

2) Send the slug, description, and friendly name associated with each outcome from the backend to be displayed in the results page

3) Show all the primary outcome metrics in the detailed results.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1524)

emtwo:
This PR addresses the issue of experiments with outcomes not loading (e.g. https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment-release/results). Here are some things the PR does:

1) Adds an `isDefault` and `metrics` field to the backend outcomes object to be sent to the frontend for visualization
2) Removes use of hardcoded `ever_used` suffix
3) Primary outcomes that are not default are not shown in either the highlights or results tables
4) There was an implicit assumption in the visualization code that primary metrics were conversion metrics and secondary metrics were count metrics - these components (e.g. `TablePrimaryMetric`) have been re-labelled to reflect their nature (e.g. `TablePrimaryMetric` -> `TableConversionMetric`)
5) All metrics, including primary, are shown by default using a `TableCountMetric` component. In the future, Jetstream would be able to specify a default way to display a given metric such as count, conversion, or something else and then the display table could be changed.

I've attached a couple of screenshots showing the `default-browser-adoption-experiment-release` experiment working locally with primary metrics displayed

Screen Shot 2021-07-22 at 10 41 15 AM

Screen Shot 2021-07-22 at 10 41 03 AM

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -135,11 +135,19 @@ class NimbusBucketRangeType(DjangoObjectType):
         exclude = ("id", "experiment")
 
 
+class NimbusOutcomeMetricType(graphene.ObjectType):
+    slug = graphene.String()
+    friendly_name = graphene.String()
+    description = graphene.String()
+
+
 class NimbusOutcomeType(graphene.ObjectType):
     friendly_name = graphene.String()
     slug = graphene.String()
     application = NimbusExperimentApplication()
     description = graphene.String()
+    isDefault = graphene.Boolean()
+    metrics = graphene.List(NimbusOutcomeMetricType)
 
 
 class NimbusLabelValueType(graphene.ObjectType):
--- a/app/experimenter/jetstream/tests/__init__.py
+++ b/app/experimenter/jetstream/tests/__init__.py
@@ -1,0 +1,10 @@
+import os
+import pathlib
+
+from django.test import override_settings
+
+mock_valid_outcomes = override_settings(
+    JETSTREAM_CONFIG_OUTCOMES_PATH=os.path.join(
+        pathlib.Path(__file__).parent.absolute(), "fixtures", "valid_outcomes"
+    )
+)
--- a/app/experimenter/jetstream/tests/test_tasks.py
+++ b/app/experimenter/jetstream/tests/test_tasks.py
@@ -16,12 +16,19 @@ from experimenter.jetstream.models import (
     MetricData,
     SignificanceData,
 )
+from experimenter.jetstream.tests import mock_valid_outcomes
 from experimenter.jetstream.tests.constants import TestConstants
+from experimenter.outcomes import Outcomes
 
 
+@mock_valid_outcomes
 @override_settings(FEATURE_ANALYSIS=False)
 class TestFetchJetstreamDataTask(TestCase):
     maxDiff = None
+
+    def setUp(self):
+        super().setUp()
+        Outcomes.clear_cache()
 
     def get_metric_data(self, data_point):
         return MetricData(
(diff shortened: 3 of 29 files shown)


app/experimenter/experiments/api/v5/types.py
```diff
@@ -135,11 +135,19 @@ class Meta:
...
+    description = graphene.String()
+
+
 class NimbusOutcomeType(graphene.ObjectType):
     friendly_name = graphene.String()
     slug = graphene.String()
     application = NimbusExperimentApplication()
     description = graphene.String()
+    isDefault = graphene.Boolean()
+    metrics = graphene.List(NimbusOutcomeMetricType)
```
jaredlockhart, line 150: Very nice 👍

app/experimenter/jetstream/tests/__init__.py
```diff
@@ -0,0 +1,10 @@
+import os
+import pathlib
+
+from django.test import override_settings
+
+mock_valid_outcomes = override_settings(
```
jaredlockhart, line 6: Ah good, ya.  I might call this something like `mock_jetstream_outcomes` just so it doesn't get confused with the other `mock_valid_outcomes` decorator.

app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -16,13 +16,20 @@
...
 
 
+@mock_valid_outcomes
 @override_settings(FEATURE_ANALYSIS=False)
 class TestFetchJetstreamDataTask(TestCase):
     maxDiff = None
 
+    def setUp(self):
+        super().setUp()
+        Outcomes.clear_cache()
```
jaredlockhart, line 31: Very nice.  Could probably also stick this in `setUpClass` since it only needs to load once for this whole suite instead of per test, but not a big deal.

app/experimenter/outcomes/__init__.py
```diff
@@ -43,6 +52,19 @@ def _load_outcomes(cls):
...
+                                        metric,
+                                        outcome_data["metrics"][metric].get(
+                                            "friendly_name"
+                                        ),
+                                        outcome_data["metrics"][metric].get(
+                                            "description"
+                                        ),
+                                    )
+                                    for metric in outcome_data["metrics"].keys()
+                                ],
```
jaredlockhart, line 67: Very nice 👍

jaredlockhart (approved):
Very nice @emtwo this looks great.  Thank you!

jodyheavener (approved):
I'm happy with this :) 🚢

PR #6018, opened 2021-07-21
fix 5974 feat(nimbus): handle manual remote settings rollback

jaredlockhart:
Because

* There is no way to prevent a Remote Settings reviewer from manually rolling back the collection after a rejection before Experimenter has a chance to query the state of the collection
* This is currently an unhandled case and leaves the experiment in a stranded Waiting state
* We can treat this as a rejection without a rejection reason and show a generic rejection message

This commit

* Updates the kinto tasks to handle the case where there is a Waiting experiment but the collection is in to-sign
* To prevent race conditions with the live/ended check tasks we must move that logic into the check collection task
* Moves enrollment/end emails into a single task
* Updates docs to distinguish between the reject and reject+rollback fows
* Updates tests as necessary


lmorchard (approved):
Looks like a good refactor and reorg. Thought I found a problem, but figured out it was something unrelated that I filed as issue #6023

PR #6030, opened 2021-07-22
fixes #5718 feat(nimbus): save countries and locales in experiment form

### LINKED ISSUE #5718: (FE) Hook up Country and Locale fields’ UI to existing values and mutation
Now that the FormAudience page has the new Locales and Countries UI we need to hook it up.

Start by updating the [getExperiment](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/gql/experiments.ts#L27:L27) GraphQL query to retrieve the new locales and countries fields, so they can be accessed on experiment data.

For both fields’ Select components apply _formSelectAttrs_ returned by the _useCommonForm_ hook. This will require a _useState_ callback to manage each field’s selected values. Refer to [FormMetrics](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx) to see how we’ve previously done it. The two fields should now populate with existing experiment data if it’s available. The values passed to the mutation should be country or locale IDs.

In the [PageEditAudience](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.tsx) component update the _onFormSubmit_ callback to pass along country and locale data to the _updateExperiment_ mutation. You should now be able to save the data for each of these fields.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1250)

jodyheavener:
Closes #5718

This PR adds locales and countries to the mutation call that saves experiment data.

Also, I forgot that we were saving countries/locales by their `id`, not their `code`, so I swapped that around.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
@@ -34,8 +34,8 @@ type FormAudienceProps = {
 };
 
 type AudienceFieldName = typeof audienceFieldNames[number];
-type SelectCodeItems = {
-  code: string;
+type SelectIdItems = {
+  id: number;
   name: string;
 }[];
 
@@ -51,10 +51,10 @@ export const audienceFieldNames = [
   "locales",
 ] as const;
 
-const selectOptions = (items: SelectCodeItems) =>
+const selectOptions = (items: SelectIdItems) =>
   items.map((item) => ({
     label: item.name!,
-    value: item.code!,
+    value: item.id!,
   }));
 
 export const FormAudience = ({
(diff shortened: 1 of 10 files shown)


app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -107,10 +107,19 @@ export const FormAudience = ({
...
       [false, true].map((next) =>
         handleSubmit(
-          (dataIn: DefaultValues) => !isLoading && onSubmit(dataIn, next),
+          (dataIn: DefaultValues) =>
+            !isLoading &&
+            onSubmit(
+              {
+                ...dataIn,
+                locales,
+                countries,
```
jodyheavener, line 116: I would have thought `countries` and `locales` would appear in the `dataIn` object automatically because they are registered components. But alas, this is not happening, and is even how we do it for the Metrics page.

jaredlockhart (approved):
Works perfectly thnx @jodyheavener ! 🎉 🎉 🎉

PR #6040, opened 2021-07-23
fixes #6047 type(feat): Add a new targeting attribute for non enterprise users

### LINKED ISSUE #6047: Add targeting for users with enterprise policies
We want to be able to target users with enterprise policies because some experiments might not make sense in that case.

piatra:


jaredlockhart (approved):
Thanks @piatra looks good 👍 

The test failure is an unrelated intermittent, so you should just be able to restart it.  I filed a ticket for it.

Also commit formatting guidelines [are here](https://github.com/mozilla/experimenter/blob/main/contributing.md#git-commit-guidelines) 🙏

PR #6056, opened 2021-07-24
fixes #6055 fix(legacy): change jira DS url to new url

### LINKED ISSUE #6055: Jira DS Url Validation need to change correspond to new url
Since the Jira root url has change, we need to change the validation to correspond to the new url.
Since stage and prod are the same, we can hardcode this value

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1545)

tiftran:
Because:
* Jira migration therefore new jira url

This commit:
* Hardcodes the new DS jira url in settings.py


.env.integration-tests
```diff
@@ -9,7 +9,7 @@ DB_PASS=postgres
 DB_USER=postgres
 DEBUG=True
 DELIVERY_CONSOLE_HOST=
-DS_ISSUE_HOST=https://jira.example.com/browse/
+DS_ISSUE_HOST=https://mozilla-hub.atlassian.net/browse/
```
jaredlockhart, line 12: You should just be able to take `DS_ISSUE_HOST` out of both `.env.sample` and `.env.integration-test` shouldn't you?

jaredlockhart (approved):
Yep looks right, ty.  You'll also have to update the URLs in the legacy integration tests.  Thnx @tiftran !

.env.integration-tests
tiftran: oooh you're right

PR #6065, opened 2021-07-26
fix #6023 fix(nimbus-ui): Skip lifecycles involving Remote Settings for dummy experiments

### LINKED ISSUE #6023: Some lifecycle states are problematic in loaded dummy experiments
Some lifecycle states should only be reached when there is a corresponding record in Kinto. Iterating through all possible lifecycle states in load_dummy_experiments results in some experiments in states that violate this expectation.

For example, ENDING_APPROVE without a corresponding record in Kinto results in a repeating celery task that keeps trying to delete the experiment but [failing with an exception each time](https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/tasks.py#L262-L265).

Solution is probably to figure out which lifecycle states produce this situation and just filter those out of the list used to generate dummy experiments.

Off the top of my head, this is at least ENDING_APPROVE and PAUSING_APPROVE. Might also include any WAITING cases.

Could also look at making the Celery task handle this data inconsistency more robustly, though it's probably an edge case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1527)

lmorchard:
Because:

* Some lifecycle states should only be reached when there is a
  corresponding record in Remote Settings.

* Loading dummy experiments doesn't update the local instance of Remote
  Settings

This commit:

* Only generates dummy experiments for lifecycle states that do not
  assume an corresponding experiment record exists in Remote Settings

* Preview is included as an exception because the synchronize task will
  push the experiment to Remote Settings shortly after generation

### DIFF
--- a/app/experimenter/base/management/commands/load_dummy_experiments.py
+++ b/app/experimenter/base/management/commands/load_dummy_experiments.py
@@ -21,7 +21,7 @@ class Command(BaseCommand):
             experiment = ExperimentFactory.create_with_status(status, type=random_type)
             logger.info("Created {}: {}".format(experiment, status))
 
-        for lifecycle in NimbusExperimentFactory.Lifecycles:
+        for lifecycle in NimbusExperimentFactory.LocalLifecycles:
             experiment = NimbusExperimentFactory.create_with_lifecycle(
                 lifecycle, with_random_timespan=True
             )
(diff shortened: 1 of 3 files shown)


jaredlockhart (approved):
Yep looks good thnx 👍 One thing I've been thinking of doing is using the lifecycle title as the actual experiment name instead of the random names so it's easier to pick the right one when you're doing local testing, but we could do that whenever.  Thnx @lmorchard !

PR #6078, opened 2021-07-28
fix #6069 feat(nimbus): support deprecated targeting configs

### LINKED ISSUE #6069: Handle missing targeting slugs in Nimbus UI
Because advanced targeting configs can come and go, we need to handle the case that a past experiment has a targeting config that isn't in the list of currently available targeting configs anymore.  We can probably just spit out the slug in that case.

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/EXP-1550)

jaredlockhart:
Because

* Targeting configs can come, change, and go
* We need to be able to render completed experiments with targeting configs that no longer exist
* We need to be able to edit draft experiments that were saved with a suddenly deprecated targeting config

This commit

* Removes the enum choices from the targeting config slug field in the V5 API
* Valid choices will still be validated by the serializer validation layer
* The frontend will display deprecated targeting configs with a Deprecated label
* Published experiments will use their published targeting string in the UI instead of attempting to regenerate it

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -183,7 +183,7 @@ class NimbusExperimentType(DjangoObjectType):
     channel = NimbusExperimentChannel()
     documentation_links = DjangoListField(NimbusDocumentationLinkType)
     treatment_branches = graphene.List(NimbusBranchType)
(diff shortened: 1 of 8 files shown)


lmorchard (approved):
👍

PR #6085, opened 2021-07-29
fix #6083 chore(project): update docker instructions in readme

### LINKED ISSUE #6083: Update Docker docs in README
We need to update the Docker docs to mention a couple things:

* We should recommend using [Docker Desktop 3.3.2 on MacOS](https://docs.docker.com/docker-for-mac/release-notes/#docker-desktop-332) since more recent versions have been buggy (I wish there was a proper LTS version)
* We need to mention increasing the default resources since running `make check` with the default memory cap of 2gb will result in eslint exiting with 137 (OOM)

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/EXP-1558)

jaredlockhart:
Because

* The most recent Docker Desktop builds have been buggy so we should recommend a known stable version
* The default resource settings won't allow make check to pass so we have to recommend increasing them

This commit

* Links to Docker Desktop 3.3.2
* Provides resource guidance

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,14 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or 
 
 1.  Install [docker](https://www.docker.com/) on your machine
 
-- On linux, [setup docker to run as non-root](https://docs.docker.com/engine/security/rootless/)
+- On Linux [setup docker to run as non-root](https://docs.docker.com/engine/security/rootless/)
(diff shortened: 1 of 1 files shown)


pdehaan (approved):
R+.

We should try and set up a periodic reminder to spot check newer Docker Desktop versions to see if the bugginess gets sorted so we don't need to pin to an archived version.

PR #6082, opened 2021-07-28
fix #5815 feat(nimbus): add support for focus/klar on android

jaredlockhart:
Because

* We'd like to support Focus and Klar on android

This commit

* Adds application configs for Focus and Klar
* We need to remember to add the `No Feature` fixtures for every new application
* Added a test to make sure those fixtures are added so we don't forget in the future
* This is only temporary until we move feature configs over to application repos and consume them the same way we do outcomes and take them out of the database altogether

Issue #6032, opened 2021-07-22
Visualization overlapping text


jaredlockhart:
On https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment/results you can see some of the results numbers are overlapping:

<img width="321" alt="Screen Shot 2021-07-22 at 3 36 23 PM" src="https://user-images.githubusercontent.com/119884/126699255-74e97b95-41a6-41b6-8abc-bef4afaeeb6d.png">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1532)

LZoog:
This was closed in #6088

PR #6088, opened 2021-07-29
fixes #6082 fix(nimbus): Address ConfidenceInterval overlapping text

LZoog:
fixes #6082

Because:
* When there are a large number of digits in upper/lower bounds, confidence interval numbers overlapped

This commit:
* Updates the space buffer from a static 5 to calculate based on number of digits + 3

===

This is one of those things where I tried several different values (including the values on the page) and things LGTM with this calculation, but it's possible it may need tweaking later when there's another edge case. Could maybe use a test, but we've got test coverage and this is just for a visual change so I added some stories instead.  

image


LZoog:
OK, I fiddled with this a lot and added a bunch of cases in Storybook and think it's looking pretty good. I won't be surprised if there's maybe another edgecase in the future since this kind of thing tends to be finicky but I tried out other values too and feel like this is good enough for now and definitely addresses the bug.

jaredlockhart (approved):
Awesome thank you @LZoog 👍 👍 👍 💯 💯 💯

LZoog:
Arrrgh, wrong issue reference here. This closed #6032.

PR #6090, opened 2021-07-30
fix #6081 feat(nimbus): filter channel by application

### LINKED ISSUE #6081: Disable some channels for some applications
Some applications might not support all channels so we'll have to add a list of channels to each application and then use that to populate the forms in the frontend.

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/EXP-1557)

jaredlockhart:
Because

* Some applications may not support all channels

This commit

* Uses the channel/app_id mapping as the canonical definition of which channels an application supports
* Removes NO_CHANNEL as the default value for the model field
* Instead sets an initial channel choice at creation time from the channels available for the chosen application
* Exposes the application/channels mapping through the config query
* Filters the channel options on the frontend based on the chosen application


jodyheavener (approved):
WFM

PR #6091, opened 2021-07-30
fix #6034 chore(nimbus): remove locale/country filtering from targeting configs

### LINKED ISSUE #6034: Remove locale filtering from targeting configs
Now that we have the ability to explicitly set locale/country filtering, we no longer need locale/country filtering baked into the targeting configs.  Let's update the targeting configs to remove any locale/country specific filtering.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1533)

jaredlockhart:
Becuase

* We can now explicitly target by locale/country independently of targeting configs
* Allowing locale/country filtering in targeting configs in addition to the specific fields can create invalid targeting expressions

This commit

* Removes all locale/country filtering from targeting configs


jrbenny35 (approved):

jodyheavener (approved):
noice

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -101,50 +101,25 @@ class NimbusTargetingConfig:
...
-TARGETING_ALL_ENGLISH = NimbusTargetingConfig(
-    name="All English users",
-    slug="all_english",
-    description="All users in en-* locales.",
-    targeting="localeLanguageCode == 'en'",
-    desktop_telemetry="STARTS_WITH(environment.settings.locale, 'en')",
-    application_choice_names=(Application.DESKTOP.name,),
-)
-
-TARGETING_US_ONLY = NimbusTargetingConfig(
```
pdehaan, line 113: I think this is covered by #6078, but what's the experiment owner impact of deleting these?
Previous experiments that are draft/live/completed will still show their targeting configs at the time they were created, along with some deprecation notices?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -101,50 +101,25 @@ class NimbusTargetingConfig:
     ),
 )
 
-TARGETING_ALL_ENGLISH = NimbusTargetingConfig(
-    name="All English users",
-    slug="all_english",
-    description="All users in en-* locales.",
-    targeting="localeLanguageCode == 'en'",
```
pdehaan, line 108: And now if an experiment owner wants to target the old `localeLanguageCode == 'en'`, they'd need to explicitly select all 3 English* locales, correct? "English (British)", "English (Canadian)", and "English (US)"

That seems to give me the following JEXL (when I have "US users (en)" Advanced Targeting on stage):

```jexl
'app.shield.optoutstudies.enabled'|preferenceValue &&
  localeLanguageCode == 'en' &&
  region == 'US' &&
  locale in ['en-CA', 'en-GB', 'en-US']
```

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: @pdehaan Yeah that's what https://github.com/mozilla/experimenter/pull/6078 was for, so now it'll show `Deprecated` next to the targeting config for past experiments and treat drafts with deprecated targeting configs as invalid so they're forced to go back and select a valid one 👍

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Exactly 👍

PR #6092, opened 2021-07-30
fix #6038 chore(project): cache docker ui layer

### LINKED ISSUE #6038: Cache build target in dockerhub
I just realized that the work done in the `build` target of the Dockerfile is never pushed to Dockerhub so it can't be rehydrated locally from the shared cache.  While I'm at it I should add the facility to cache by commit hash and branch so increase the chances of hits.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1536)

jaredlockhart:
Because

* We have a fourth layer in our Dockerfile that we're not remotely caching which slows down local builds/ci

This commit

* Builds and pushes the ui build target in our Dockerfile in circle

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,13 +92,16 @@ jobs:
             docker login -u $DOCKER_USER -p $DOCKER_PASS
             make build_dev
             make build_test
+            make build_ui
             ./scripts/store_git_info.sh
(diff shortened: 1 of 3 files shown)


jrbenny35 (approved):
All the cache 🤑

PR #6094, opened 2021-07-30
fix #6093 bug(nimbus): look up targeting config name in v5 api if valid

### LINKED ISSUE #6093: Cannot save experiment with empty Advanced Targeting selection
Not sure when this started happening but it looks like by default an experiment has no selection for the "Advanced Targeting" field on the Audience page. You cannot save the form with this field empty; you need to explicitly select the "No Targeting" option.

Furthermore, it looks like when you do selected the "No Targeting" option, save the form, and then reload the page the field returns to an empty state, preventing further saves.

This field should default to the "No Targeting" option, and use that option when there is no default value. It should not be possible to save this field with a truly empty value.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1563)

jaredlockhart:
Because

* To support deprecated targeting configs we removed the enum slug/name conversion from the targeting_config_slug field in the V5 API
* This breaks the form since it can no longer map the stored slug to the name value in the options

This commit

* Optionally returns the enum name in the V5 API if it the stored slug maps to a currently valid enum value
* Split the V5 API query tests into separate test classes by query rather than having them all mixed together

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_queries.py
+++ b/app/experimenter/experiments/tests/api/v5/test_queries.py
@@ -15,7 +15,7 @@ from experimenter.experiments.tests.factories.nimbus import NimbusFeatureConfigF
 from experimenter.outcomes import Outcomes
 
 
-class TestNimbusQuery(GraphQLTestCase):
+class TestNimbusExperimentsQuery(GraphQLTestCase):
     GRAPHQL_URL = reverse("nimbus-api-graphql")
 
     def test_experiments(self):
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -559,40 +598,157 @@ def test_recipe_json_returns_serialized_data(self):
...
+            variables={"slug": experiment.slug},
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200, response.content)
+        content = json.loads(response.content)
+        experiment_data = content["data"]["experimentBySlug"]
+        self.assertEqual(
+            experiment_data["targetingConfigSlug"],
+            "deprecated_targeting",
+        )
```
jaredlockhart, line 747: Sorry I muddied the pr by reorganizing the tests.  These are the new tests.

jodyheavener (approved):

pdehaan (approved):
👍 LGTM locally.

PR #6098, opened 2021-07-30
fix #6097 fix(nimbus): Position ConfidenceInterval numbers relatively

### LINKED ISSUE #6097: Small significance negative bounds in ConfidenceInterval looking wonky, causing horizontal scrollbar
follow up to #6032

I was poking around the Results page to make sure everything looked good after merging that PR but I see a couple edge cases I need to address.

https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment/results

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1566)

LZoog:
fixes #6097

Because:
* The previous fix for overlapping numbers was causing other wonky edgecases

This commit:
* Tweaks the ConfidenceInterval to position numbers relatively instead of absolute with non-breaking spaces to account for small significances so they'll never be squished

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.stories.tsx
@@ -3,178 +3,132 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { withLinks } from "@storybook/addon-links";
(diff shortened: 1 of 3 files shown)


LZoog:
@jaredlockhart This should be way less flaky than the previous fix. The non-breaking spaces serve as a minimum width between the numbers so they'll never be squished.

robhudson (approved):

PR #6099, opened 2021-08-01
fix #6064 chore(nimbus): Cleanup of nimbus end-to-end tests

### LINKED ISSUE #6064: Make integration test setup cleaner using pytest fixtures
Right now we repeat a lot of steps within the nimbus e2e suite. We can reduce these by using fixtures, and a similar setup to the legacy e2e tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1548)

lennartschoch:
Because

* code in the nimbus e2e-tests is duplicated frequently (especially the initial experiment creation)

This commit

* moves the logic for the creation of experiments into a pytest fixture that can be re-used


lennartschoch:
Oops, two of the e2e tests are also failing in CircleCI... so seems like I actually broke something.
Since there are more tests failing on my machine (3/4) than on the pipeline (2/4), I guess there is still something wrong with my setup though.

lennartschoch:
Made some progress regarding this today:
- The tests on `main` are now passing on my machine. Seems like I just used the wrong .env file
- The tests on my branch are still failing. Weirdly, they still seem to fail because of `#approve-request-button` not being visible (which isn't a problem in CircleCI). I also discovered `make integration_vnc_up`, but since the test abort the moment they fail, it was hard to see what exactly is happening.

jrbenny35:
Hey @lennartschoch! Thanks for working on this. For your local testing there is a few steps to run before the ```make integration_test_nimbus``` command will work. You can follow these commands https://github.com/mozilla/experimenter/blob/main/.circleci/config.yml#L77


Let me know if that helps.

lennartschoch:
Hey @jrbenny35, thanks for the advice! I went through these steps yesterday and now the tests pass on `main` on my machine. For some reason, the test on my feature branch still fail for different reasons on my machine than on the pipeline. For example, the test seem to be unable to find the `#approve-request-button` in the nimbus UI on my machine, but on the pipeline the tests went past that point and only failed when approving or rejecting the experiment on Kinto.

jrbenny35:
> 
> 
> Hey @jrbenny35, thanks for the advice! I went through these steps yesterday and now the tests pass on `main` on my machine. For some reason, the test on my feature branch still fail for different reasons on my machine than on the pipeline. For example, the test seem to be unable to find the `#approve-request-button` in the nimbus UI on my machine, but on the pipeline the tests went past that point and only failed when approving or rejecting the experiment on Kinto.

Gotcha, let me look through the PR and see if you missed something. The Approve button won't show if the experiment forms aren't filled out correctly.

lennartschoch:
> > Hey @jrbenny35, thanks for the advice! I went through these steps yesterday and now the tests pass on `main` on my machine. For some reason, the test on my feature branch still fail for different reasons on my machine than on the pipeline. For example, the test seem to be unable to find the `#approve-request-button` in the nimbus UI on my machine, but on the pipeline the tests went past that point and only failed when approving or rejecting the experiment on Kinto.
> 
> Gotcha, let me look through the PR and see if you missed something. The Approve button won't show if the experiment forms aren't filled out correctly.

Oh, that's good to know! Maybe I was paying too much attention to the screen with the approve button and not enough attention to the experiment forms when trying to debug this. Will also have another look! I guess you're referring to the Approve button in Nimbus, not Kinto, correct?

jrbenny35:
> 
> 
> > Gotcha, let me look through the PR and see if you missed something. The Approve button won't show if the experiment forms aren't filled out correctly.
> 
> Oh, that's good to know! Maybe I was paying too much attention to the screen with the approve button and not enough attention to the experiment forms when trying to debug this. Will also have another look! I guess you're referring to the Approve button in Nimbus, not Kinto, correct?

Hmm. Looks like the Experiment Not Found error is after it has been rejected on Kinto and should be showing on the drafts page. 

The other error points to a page not loading. Seems like the Summary Page isn't loading

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -1,57 +1,81 @@
...
+            channel="Nightly",
+            min_version=80,
+            targeting="TARGETING_MAC_ONLY",
+            percentage=50.0,
+            expected_clients=50,
+        ),
+    )
+
+
+def create_experiment(selenium, home_page, data):
```
jrbenny35, line 39: This could be a fixture and moved to ```conftest.py```.

jrbenny35 (changes requested):
Some initial thoughts, thanks for starting on this!

app/tests/integration/nimbus/test_e2e_create_experiment.py
lennartschoch: The problem is that the experiment name of each test has to be different to make sure there are no conflicts between tests. I saw that there is a way of [parameterizing fixtures using pytest annotations](https://docs.pytest.org/en/latest/example/parametrize.html), but it looks a bit hacky to me. Should I still use it to define the experiment name for each test or do you have another idea?

lennartschoch:
@jrbenny35 thanks for the feedback! :) I implemented all the changes requested by you except the one I [commented](https://github.com/mozilla/experimenter/pull/6099#discussion_r682172251) on.

Also, I managed to get the tests to pass both locally and on the pipeline. Still, I believe there might be something a bit flaky in the experiment creation part (at some point, different tests were failing locally and on CircleCI for different reasons, even though I was running the exact same commands on my machine that run on the pipeline).

 But for now everything seems to be fine! 🎉

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Maybe concatenate some random numbers or a shortened UUID on the end of the assignment. 

```suggestion
data.public_name + UUID generation or something
```

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -8,50 +8,55 @@
...
     # Review
     selenium.find_element_by_css_selector("#PageSummary")
+    return review
 
 
-def test_create_new_experiment_remote_settings(selenium, base_url):
-    experiment_name = "test_create_new_experiment_remote_settings"
+@pytest.mark.nondestructive
+def test_create_new_experiment(selenium, base_url, default_data):
+    default_data.public_name = "test_create_new_experiment"
```
jrbenny35, line 51: Love this

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Do you still plan to move this to ```conftest.py```?

jrbenny35 (changes requested):
This is looking good!

app/tests/integration/nimbus/test_e2e_create_experiment.py
lennartschoch: I do! Will take some time tomorrow to move this into a fixture and implement your random number / UUID idea!

jaredlockhart: I liked having the experiment name be the name of the test so it's easier to debug the tests after the fact?

app/tests/integration/nimbus/test_e2e_create_experiment.py
jaredlockhart: What's being passed in here?  An instance of a data class?  Does mutating it share data between tests?

jaredlockhart (commented):
Generally looking really good to me!  Some ideas for naming/structure.  Thanks for working on this @lennartschoch !

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -126,57 +100,25 @@ def test_create_new_experiment_remote_settings(selenium, base_url):
...
             item.click()
             break
     summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
     assert "live" in summary_page.experiment_status.lower()
 
 
-def test_create_new_experiment_remote_settings_reject(selenium, base_url):
-    experiment_name = "test_create_new_experiment_remote_settings_reject"
+def test_create_new_experiment_remote_settings_reject(selenium, base_url, default_data):
+    default_data.public_name = "test_create_new_experiment_remote_settings_reject"
```
jaredlockhart, line 111: Oh I think my comment vanished somehow.  Are we mutating the same object that's passed to multiple tests?  Does this share data between tests?

app/tests/integration/nimbus/test_e2e_create_experiment.py
lennartschoch: @jaredlockhart I agree. In case we actually go for the generated experiment names, I think we would at least have to log them.

@jrbenny35 I just started putting the `create_experiment` function into a fixture. I think this might not be as easy as it seems. In some tests, we have steps that have to be executed **before** the creation of the experiment, e.g. counting the number of experiments in the `test_create_new_experiment_remote_settings` test or navigating to the Drafts tab in `test_create_new_experiment_remote_settings_reject`. Probably, in order to keep this behavior, we would need to move those parts of the code into fixtures as well. If the only reason for putting this function into a fixture is to make it available in multiple test files, I would suggest we don't create a fixture that executes `create_experiment`, but instead returns the `create_experiment` function itself, so we can execute it whenever we want. This pattern is also explained in the [pytest docs](https://docs.pytest.org/en/latest/how-to/fixtures.html#factories-as-fixtures). What do you think?

lennartschoch:
Thanks for the feedback @jaredlockhart! Will create some commits based on it tomorrow :)

app/tests/integration/nimbus/test_e2e_create_experiment.py
lennartschoch: This should not be the case. According to the [docs](https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes), fixtures are destroyed after each test by default.

app/tests/integration/nimbus/models/base_dataclass.py
```diff
@@ -0,0 +1,35 @@
...
+
+@dataclass
+class BaseDataClass:
+    hypothesis: str
+    application: str
+    public_description: str
+    branches: Optional[List[BaseBranchDataClass]]
+    metrics: BaseMetricsDataClass
+    audience: BaseAudienceDataClass
+    public_name: Optional[str] = None
```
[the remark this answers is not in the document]
lennartschoch: Fixed in https://github.com/mozilla/experimenter/pull/6099/commits/b71f79a2ed8b031ae43b0b75ca14f4359eab64b8

app/tests/integration/nimbus/models/base_dataclass.py
```diff
@@ -0,0 +1,35 @@
+from dataclasses import dataclass
+from typing import List, Optional
+
+
+@dataclass
+class BaseBranchDataClass:
+    name: str
+    description: str
+    config: str
```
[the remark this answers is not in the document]
lennartschoch: Fixed in https://github.com/mozilla/experimenter/pull/6099/commits/b71f79a2ed8b031ae43b0b75ca14f4359eab64b8

app/tests/integration/nimbus/conftest.py
```diff
@@ -38,3 +44,27 @@ def _verify_url(request, base_url):
...
+        ],
+        metrics=BaseMetricsDataClass(primary_outcomes=[], secondary_outcomes=[]),
+        audience=BaseAudienceDataClass(
+            channel="Nightly",
+            min_version=80,
+            targeting="TARGETING_MAC_ONLY",
+            percentage=50.0,
+            expected_clients=50,
+        ),
+    )
```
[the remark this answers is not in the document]
lennartschoch: Fixed in https://github.com/mozilla/experimenter/pull/6099/commits/b71f79a2ed8b031ae43b0b75ca14f4359eab64b8

app/tests/integration/nimbus/conftest.py
```diff
@@ -2,6 +2,12 @@
 
 import pytest
 import requests
+from nimbus.models.base_dataclass import (
+    BaseAudienceDataClass,
+    BaseBranchDataClass,
+    BaseDataClass,
```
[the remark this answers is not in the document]
lennartschoch: Fixed in https://github.com/mozilla/experimenter/pull/6099/commits/b71f79a2ed8b031ae43b0b75ca14f4359eab64b8

app/tests/integration/nimbus/conftest.py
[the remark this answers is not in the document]
lennartschoch: One more thing I'd like to do here is using the value of the `channel` select instead of the name for consistency reasons. I'll wait for this until #6132 is resolved though.

lmorchard:
A fix for #6132 landed, so I'll give CI a kick

jaredlockhart:
@lennartschoch You should be able to rebase on `main` now to clear up the integration test failures 👍

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: @lennartschoch I like that. Let's do that.

jrbenny35:
@lennartschoch Hey! Do you have any time to work on this for the remainder of the week? This PR is blocking some tests I am looking to add, but I want to give you some time to finish it up. If you don't have the time I can add the final changes needed. Please let me know :)

jaredlockhart:
This is looking really good @lennartschoch !  I'll let @jrbenny35 give the final sign off but it all looks right to me 👍 Thank you!

app/tests/integration/nimbus/conftest.py
```diff
@@ -38,3 +49,94 @@ def _verify_url(request, base_url):
...
+
+        # Review
+        selenium.find_element_by_css_selector("#PageSummary")
+        return review
+
+    return _create_experiment
+
+
+@pytest.fixture
+def perform_kinto_action():
```
jrbenny35, line 124: Oh you went for extra points! Love it!

app/tests/integration/nimbus/conftest.py
```diff
@@ -38,3 +49,94 @@ def _verify_url(request, base_url):
...
+            targeting=BaseExperimentAudienceTargetingOptions.TARGETING_MAC_ONLY,
+            percentage=50.0,
+            expected_clients=50,
+        ),
+    )
+
+
+@pytest.fixture
+def create_experiment():
+    def _create_experiment(selenium, home_page, data):
```
jrbenny35, line 83: Beautiful!

jrbenny35 (approved):
This is really great work here! R+, I hope you stick around for more! Awesome job!!

jrbenny35:
@lennartschoch I'll give you the honors of hitting the ```squash and merge``` button.

lennartschoch:
@jrbenny35 @jaredlockhart Thank you! :) It's a really cool product you created there, will definitely stick around! Unfortunately I don't have the permissions to merge though, so I'll leave that to one of you ;)

PR #6107, opened 2021-08-03
fix #6039 fix(nimbus): Display warning for polling error, allow query refetch with useRefetchOnError hook

### LINKED ISSUE #6039: An error occurred while looking up experiments: JSON.parse
via https://mozilla.slack.com/archives/C011MC5EB0V/p1626799999145500

> not sure if someone just fixed it or it’s an intermittent? https://experimenter.services.mozilla.com/nimbus/ showed a yellow
>> _"An error occurred while looking up experiments: `JSON.parse`: unexpected character at line 1 column 1 of the JSON data"_
>
> (under the blue “not sure where to start?“)

It's rare, but intermittent. Our collective guts think the likely culprit is that the server is mid-deploy and all the assets on stage/prod aren't ready yet.

Not sure if a possible workaround patch is wrapping the JSON fetch/parse in a `try..catch` block w/ auto retry and backoff so we can reattempt after 0.5s, 1s, 2s, 4s before failing with a huge 🤷  emoji and better messaging than a JSON parse error.

I personally saw this on stage the other day during testing, but then an immediate refresh cleared the error up and I wasn't able to reproduce again after a few dozen pageviews.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1537)

LZoog:
fixes #6039 

Because:
* When a polling error occurs, users are seeing the Apollo error alert on the entire page, and showing an alert is preferred since this error can happen when Experimenter is in mid-deploy
* Users are also sometimes seeing an error on the Home page that goes away after refresh, likely also due to Experimenter mid-deploy

This commit:
* Checks in AppLayoutWithExperiment for an apollo error after experiment data is present, signalling a polling error, and displays a warning-style alert above the rest of the page rather than only an alert. Polling is retried after the pollInterval passes.
* Adds useRefetchOnError hook and implements it on PageHome and App to cover getAllExperiments and the config query. This displays a warning-style alert informing the user the query will be retries, retries the query after a settimeout (5 seconds), and returns the Apollo error alert if an error occurs again
* Refactors AppLayoutWithExperiment stories and gives the component a default testId, adds tests for both pages

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.stories.tsx
@@ -3,9 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { withLinks } from "@storybook/addon-links";
-import { storiesOf } from "@storybook/react";
 import React from "react";
-import AppLayoutWithExperiment from ".";
+import AppLayoutWithExperiment, { AppLayoutWithExperimentProps } from ".";
 import { mockExperimentQuery } from "../../lib/mocks";
 import { RouterSlugProvider } from "../../lib/test-utils";
 import {
@@ -13,63 +12,61 @@ import {
   NimbusExperimentStatus,
 } from "../../types/globalTypes";
 
-storiesOf("components/AppLayoutWithExperiment", module)
-  .addDecorator(withLinks)
-  .add("status: draft", () => {
-    const { mock } = mockExperimentQuery("demo-slug");
-    return (
-      <RouterSlugProvider mocks={[mock]}>
-        <AppLayoutWithExperiment
-          title="Howdy!"
-          testId="AppLayoutWithExperiment"
-        >
-          {({ experiment }) => <p>{experiment.name}</p>}
-        </AppLayoutWithExperiment>
-      </RouterSlugProvider>
-    );
-  })
-  .add("status: preview", () => {
-    const { mock } = mockExperimentQuery("demo-slug", {
-      status: NimbusExperimentStatus.PREVIEW,
-    });
-    return (
-      <RouterSlugProvider mocks={[mock]}>
-        <AppLayoutWithExperiment
-          title="Howdy!"
-          testId="AppLayoutWithExperiment"
-        >
-          {({ experiment }) => <p>{experiment.name}</p>}
-        </AppLayoutWithExperiment>
-      </RouterSlugProvider>
-    );
-  })
-  .add('status: launched ("live" or "complete")', () => {
-    const { mock } = mockExperimentQuery("demo-slug", {
-      status: NimbusExperimentStatus.LIVE,
-    });
-    return (
-      <RouterSlugProvider mocks={[mock]}>
-        <AppLayoutWithExperiment
-          title="Howdy!"
-          testId="AppLayoutWithExperiment"
-        >
-          {({ experiment }) => <p>{experiment.name}</p>}
-        </AppLayoutWithExperiment>
-      </RouterSlugProvider>
-    );
-  })
-  .add("publish status: review", () => {
-    const { mock } = mockExperimentQuery("demo-slug", {
-      publishStatus: NimbusExperimentPublishStatus.REVIEW,
-    });
-    return (
-      <RouterSlugProvider mocks={[mock]}>
-        <AppLayoutWithExperiment
-          title="Howdy!"
-          testId="AppLayoutWithExperiment"
-        >
-          {({ experiment }) => <p>{experiment.name}</p>}
-        </AppLayoutWithExperiment>
-      </RouterSlugProvider>
-    );
-  });
+const Subject = ({
+  ...props
+}: Omit<AppLayoutWithExperimentProps, "children">) => (
+  <AppLayoutWithExperiment title="Howdy!" {...props}>
+    {({ experiment }) => <p>{experiment.name}</p>}
+  </AppLayoutWithExperiment>
+);
+
+export default {
+  title: "components/AppLayoutWithExperiment",
+  component: Subject,
+  decorators: [withLinks],
+};
+
+const storyWithProps = (
+  { mock } = mockExperimentQuery("demo-slug"),
+  props: React.ComponentProps<typeof Subject> = {},
+  storyName?: string,
+) => {
+  const story = () => (
+    <RouterSlugProvider mocks={[mock]}>
+      <Subject {...props} />
+    </RouterSlugProvider>
+  );
+  if (storyName) story.storyName = storyName;
+  return story;
+};
+
+export const StatusDraft = storyWithProps();
+
+export const StatusPreview = storyWithProps(
+  mockExperimentQuery("demo-slug", {
+    status: NimbusExperimentStatus.PREVIEW,
+  }),
+);
+
+export const StatusLaunched = storyWithProps(
+  mockExperimentQuery("demo-slug", {
+    status: NimbusExperimentStatus.LIVE,
+  }),
+  {},
+  'Status: Launched ("Live" or "Complete")',
+);
+
+export const PublishStatusReview = storyWithProps(
+  mockExperimentQuery("demo-slug", {
+    publishStatus: NimbusExperimentPublishStatus.REVIEW,
+  }),
+);
+
+export const PollingError = storyWithProps(
+  mockExperimentQuery("demo-slug"),
+  {
+    polling: true,
+    pollInterval: 2000,
+  },
+  "Polling error (wait 2 seconds)",
+);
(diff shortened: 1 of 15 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.stories.tsx
```diff
@@ -72,4 +72,24 @@ storiesOf("components/AppLayoutWithExperiment", module)
...
+        >
+          {({ experiment }) => (
+            <p>
+              Wait 5 seconds and you'll see what this page looks like when we
+              render a polling error.
+            </p>
+          )}
+        </AppLayoutWithExperiment>
+      </RouterSlugProvider>
+    );
```
LZoog, line 94: Hmm guess I could update this file to the new Storybook syntax, will do

LZoog:
Alright, I've refactored the storybook file to reflect the new syntax.

The ticket suggests a try/catch and retry after X number of seconds, but the error from polling just comes back as `error` from `useQuery` and I tried some things that made this PR a lot more complicated that I was having trouble getting working.

This just displays an alert to the user and states polling will be retried to whatever we have set on the `pollInterval`, which defaults to 30 seconds.

See Storybook: https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/2f0c6066eee306df3967568f867cfc25e98c0de4/nimbus-ui/index.html?path=/story/components-applayoutwithexperiment--polling-error

LZoog:
Oh _man_, in checking out the ticket again, I just realized this was first reported in the **directory** view and secondary in the summary page. 😩

Right now, this only covers the Summary page. I'm going to revert this to a draft and keep hacking away at it to cover both pages.

jaredlockhart:
@LZoog does it poll on the directory view?

LZoog:
@jaredlockhart we don't poll on that page, but we can try to execute the query again after X number of seconds if it errors. From that Slack thread it seems like an immediate refresh tends to fix the problem, so maybe we try requerying after 5 seconds and if it still errors, just show the Apollo GQL error.

app/experimenter/nimbus-ui/src/components/PageHome/index.stories.tsx
```diff
@@ -41,6 +41,20 @@ export const Basic = storyTemplate([mockDirectoryExperimentsQuery()]);
 
 export const Loading = storyTemplate([]);
 
+export const QueryError = () => {
+  const mockWithError = {
+    ...mockDirectoryExperimentsQuery(),
+    error: new Error("boop, something's actually wrong"),
+  };
+  return (
+    <MockedCache mocks={[mockWithError, mockWithError]}>
```
LZoog, line 50: FWIW I think we have room for a little cleanup here and I noticed the "Loading" story hasn't worked since we started handling the errors; filed #6113 as a follow up maintenance task.

LZoog:
Okay, I made the modifications on `PageHome` and fixed several merge conflicts that emerged. See my updated commit message/comment for details.

Storybooks of interest:
* [AppLayoutWithExperiment](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/7508e50ea8d1ee87f00e4d0f74fe628cae8bf146/nimbus-ui/index.html?path=/story/components-applayoutwithexperiment--polling-error) polling error
* [PageHome](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/7508e50ea8d1ee87f00e4d0f74fe628cae8bf146/nimbus-ui/index.html?path=/story/pages-home--query-error) getAllExperiments error/retry + error fallback

P.S. normally I hate disappearing messages that users don't have a chance to read if they wanted, but I didn't think we should let this hang out for 10-15 seconds and with Experimenter being an internal tool and this being a weird edge case 🤷‍♀️

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
```diff
@@ -40,6 +43,13 @@ export const Body = () => {
   }
 
   if (error) {
+    if (!hasRefetched) {
+      setTimeout(() => {
+        refetch();
+        setHasRefetched(true);
+      }, REFETCH_DELAY);
+      return <RefetchAlert />;
+    }
```
LZoog, line 52: A few things probably worth calling out:

* This does not catch and retry for JSON.parse-errors only and instead retries if anything fails, but I could change the conditional on L46 to something like `!hasRefetched && error.message.includes("JSON.parse")` if preferred, but not sure if there's ever any other kind error that occurs mid-deploy
* Errors should be captured by Sentry in `ApolloErrorAlert` (if the refetch fails), but not if something fails just once. It didn't seem like we wanted to capture it if it went away after refetch since it's likely a false positive.
* We're also not sending the polling-specific errors to Sentry since it feels edge casey and shouldn't error after repoll, but can if we want to

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
```diff
@@ -106,4 +116,17 @@ const PageHome: React.FunctionComponent<PageHomeProps> = () => {
...
+  <Alert variant="warning" data-testid="refetch-alert" className="mt-4">
+    <Alert.Heading>Error fetching experiments</Alert.Heading>
+    <p>
+      An error occured. This usually happens when Experimenter is mid-deploy.
+    </p>
+    <p>
+      Fetching experiments will be retried one time automatically in{" "}
+      {REFETCH_DELAY / 1000} seconds.
+    </p>
+  </Alert>
```
jaredlockhart, line 129: Can we reuse this component?

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
```diff
@@ -40,6 +43,13 @@ export const Body = () => {
   }
 
   if (error) {
+    if (!hasRefetched) {
+      setTimeout(() => {
+        refetch();
+        setHasRefetched(true);
+      }, REFETCH_DELAY);
+      return <RefetchAlert />;
```
jaredlockhart, line 51: So we use this logic for fetching experiments but not configuration on the home page?  What's the reasoning there?

jaredlockhart (commented):
Tested locally, looks great, works great, just curious how we should handle errors on the initial load of the landing page?

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
LZoog: Ohh, good call out. I actually just forgot that I need to add this to `App/index.tsx` as well to cover both queries, will add it there.

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
LZoog: Now that I'm going to add this to `App/index.tsx` we can. With the heading, body, and alert `variant` differing, I didn't share it with `ApolloErrorAlert` or `AppErrorBoundary` alert.

jodyheavener (commented):
Just one thought!

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
```diff
@@ -40,6 +43,13 @@ export const Body = () => {
   }
 
   if (error) {
+    if (!hasRefetched) {
+      setTimeout(() => {
```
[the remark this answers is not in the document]
LZoog: Oh funny, I didn't see your comment but just pushed something similar since I saw the memory leak while testing.

LZoog:
Pushed another update - I've created a new hook, `useRefetchOnError`, since we needed to refetch the config as well as getAllExperimentsQuery, and added two stories for it.

I didn't end up reusing `RefetchAlert` with a polling error since the entire message heading/body differs and the `if (error) { return ErrorAlert }` bits wouldn't be as clean/simple. I also used a more general refetch message so it can be used for the config and all experiments query error.

I can remove any tests that might seem superfluous, I left the ones I wrote before this push except for one I believe that I moved into the hook test, but it seemed we should double check it works as expected on the pages we're using it, and we need the line coverage.

* [hook story](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/6335fe007812efd1d3880f4bfc27656e24bdc09d/nimbus-ui/index.html?path=/story/hooks-userefetchonerror--error-and-refetch-error)
* [home page story](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/6335fe007812efd1d3880f4bfc27656e24bdc09d/nimbus-ui/index.html?path=/story/pages-home--query-error)
* [polling story](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/6335fe007812efd1d3880f4bfc27656e24bdc09d/nimbus-ui/index.html?path=/story/components-applayoutwithexperiment--polling-error

jodyheavener (approved):
LGTM!

PR #6109, opened 2021-08-03
fix #5958 feat(nimbus): directory view filters

### LINKED ISSUE #5958: Add filtering to directory view
Let's add a horizontal filter bar on the top of the directory view (either above or below the status tabs, whichever looks better) with the following fields:

* Feature
* Application
* Owner
* Version

Each filter select should be a multiselect, and changing it should dynamically update the visible table.  Swapping status tabs should automatically apply the currently selected filters.  The displayed number of experiments in each tab should update as the filters change.  Currently applied filters should be reflected in the url as query parameters, and we’ll probably want to also put the currently selected status tab in there as well.

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/EXP-1494)

lmorchard:
Because:

* We'd like to filter the directory view listing by various criteria

This commit:

* Adds a new FilterBar component to PageHome with multi-select boxes

* Adds utilities to manage search query parameters as filter state and
  applying filters to the list of fetched experiments

* Adds featureConfig, application, and firefoxMinVersion to the
  getAllExperiments query for directory view

* Further expands mock data to provide something a little more
  interesting to filter in stories

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.stories.tsx
@@ -1,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import FilterBar from ".";
+import { EVERYTHING_SELECTED_VALUE, Subject } from "../mocks";
+
+const withPadding = (Story: React.FC) => (
+  <div className="p-3">
+    <Story />
+  </div>
+);
+
+export default {
+  title: "pages/Home/FilterBar",
+  component: FilterBar,
+  decorators: [withPadding],
+};
+
+export const NothingSelected = () => <Subject />;
+
+export const EverythingSelected = () => (
+  <Subject value={EVERYTHING_SELECTED_VALUE} />
+);
--- a/app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
@@ -1,0 +1,133 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React, { useMemo } from "react";
+import Nav from "react-bootstrap/Nav";
+import Navbar from "react-bootstrap/Navbar";
+import Select, { OptionsType, OptionTypeBase } from "react-select";
+import { NullableObjectArray } from "../../../lib/types";
+import { FilterOptions, FilterValue } from "../types";
+
+export type FilterBarProps = {
+  options: FilterOptions;
+  value: FilterValue;
+  onChange: (value: FilterValue) => void;
+};
+
+export const FilterBar: React.FC<FilterBarProps> = ({
+  options,
+  value: filterValue,
+  onChange,
+}) => {
+  return (
+    <Navbar variant="light" bg="light" className="nav-fill mt-4 mb-4">
+      <Nav className="w-100">
+        <FilterSelect
+          fieldLabel="Feature"
+          fieldOptions={options.featureConfig!}
+          filterValueName="featureConfig"
+          optionLabelName="name"
+          optionValueName="slug"
+          {...{ filterValue, onChange }}
+        />
+        <FilterSelect
+          fieldLabel="Application"
+          fieldOptions={options.application!}
+          filterValueName="application"
+          optionLabelName="label"
+          optionValueName="value"
+          {...{ filterValue, onChange }}
+        />
+        <FilterSelect
+          fieldLabel="Owner"
+          fieldOptions={options.owners!}
+          filterValueName="owners"
+          optionLabelName="username"
+          optionValueName="username"
+          {...{ filterValue, onChange }}
+        />
+        <FilterSelect
+          fieldLabel="Version"
+          fieldOptions={options.firefoxMinVersion!}
+          filterValueName="firefoxMinVersion"
+          optionLabelName="label"
+          optionValueName="value"
+          {...{ filterValue, onChange }}
+        />
+      </Nav>
+    </Navbar>
+  );
+};
+
+type FilterSelectProps<T extends NullableObjectArray> = {
+  filterValue: FilterValue;
+  onChange: (value: FilterValue) => void;
+  filterValueName: keyof FilterOptions;
+  fieldLabel: string;
+  fieldOptions: T;
+  optionLabelName: keyof NonNullable<T[number]>;
+  optionValueName: keyof NonNullable<T[number]>;
+};
+
+const FilterSelect = <T extends NullableObjectArray>({
+  filterValue,
+  onChange,
+  filterValueName,
+  fieldLabel,
+  fieldOptions,
+  optionLabelName,
+  optionValueName,
+}: FilterSelectProps<T>) => {
+  const filterOptions = useMemo(
+    () =>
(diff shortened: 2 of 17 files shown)


lmorchard:
Opening as a draft to get some storybook links published. Lots of work left on test coverage, at least.

But also a couple of details that could still use some general polish:

* The filter bar layout is functional but pretty ugly. Could use some further thinking on how to arrange it
* I'm lazily throwing the filter criteria into URL params as a JSON blob. That could be better broken up into per-filter params, if the JSON blob is too offensive.

lmorchard:
And some initially relevant stories:

* https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/24ae0d2fdc00f14955ad314e4ee7eb083f9baeeb/nimbus-ui/index.html?path=/story/pages-home-filterbar--no-selections
* https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/24ae0d2fdc00f14955ad314e4ee7eb083f9baeeb/nimbus-ui/index.html?path=/story/pages-home--basic

Functioning but ugly. Still poking around at maybe giving the react-select elements a default min-width or a flex grow style. But, even with that, things get dicey in the other direction when all the filters have many selections.

lmorchard:
Updated with a few style tweaks to de-uglify, don't want to putter too much more with it in absence of a second opinion. Will add some storybook links here as soon as they publish. Meanwhile, will work to get test coverage handled.

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  filterValue: FilterValue;
+  onChange: (value: FilterValue) => void;
+  filterValueName: keyof FilterOptions;
+  fieldLabel: string;
+  fieldOptions: T;
+  optionLabelName: keyof NonNullable<T[number]>;
+  optionValueName: keyof NonNullable<T[number]>;
+};
+
+const FilterSelect = <T extends NullableObjectArray>({
```
lmorchard, line 73: Most of the trickiness in this component is maintaining lists of string values to be used in `filterValue` so its easier to convert to / from query params.

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      ),
+    [fieldOptions],
+  );
+  const fieldValue = filterValue[filterValueName];
+  const selectedOptions = useMemo(
+    () =>
+      filterOptions.filter(
+        (option) =>
+          typeof option[optionValueName] === "string" &&
+          fieldValue?.includes("" + option[optionValueName]),
```
lmorchard, line 95: This is dirty, but I spent way more time than I liked trying to get something cleaner type-wise here

lmorchard (commented):
Hopefully CI is green shortly? In the meantime, giving this a self-review for any weird spots.

lmorchard:
Gonna just go and say this is ready for review. The integration test failures in CI look unrelated, but I'll poke at them too.

jaredlockhart:
Yeah something just quietly bumped in pypi land looks like. So much for deterministic builds XD

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.stories.tsx
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import FilterBar from ".";
+import { EVERYTHING_SELECTED_VALUE, Subject } from "../mocks";
+
+const withPadding = (Story: React.FC) => (
```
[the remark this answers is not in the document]
jodyheavener: (could be a quick win too)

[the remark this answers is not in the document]
lmorchard: Yeah, I think we do use this in a bunch of places, and decorators seem to have gotten easier to use with the new storybook file format

lmorchard:
Rebased against main, hopefully that resolves the integration test failures!

jaredlockhart (changes requested):
Awesome work @lmorchard !  Works perfectly!  Couple UI tweaks:

* Left align the drop down options
Screen Shot 2021-08-06 at 4 17 58 PM

* Used a fixed width for each filter option so their sizes aren't jumping around every time you change one
Screen Shot 2021-08-06 at 4 18 22 PM

* Is putting JSON in the url necessary? It's only four fields can't you just say `field=a,b,c&&`?
Screen Shot 2021-08-06 at 4 19 15 PM

lmorchard:
Pushed some tweaks:

* Left align the drop down options ✅
* Used a fixed width for each filter option
    * Not really a such thing as fixed width in a responsive flex layout. I think I managed to get it close, but it will jump in size a little as the boxes fill up with options
* Is putting JSON in the url necessary? It's only four fields can't you just say field=a,b,c&?
    * [Wanted to punt that to a follow up](https://github.com/mozilla/experimenter/pull/6109#discussion_r683781978), but that was when I thought this would make the last sprint. As long as we're careful that these values never themselves contain commas, this should work now

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+      ),
+    [fieldValue, filterOptions],
+  );
+
+  return (
+    <Nav.Item
+      className="m-1 text-left flex-grow-1 flex-shrink-1"
+      style={{ flexBasis: "0px" }}
```
LZoog, line 103: I think instead of flex-grow/flex-shrink + flex-basis, we only need `flex-basis: 1`. Can we add a custom class in `styles/index.scss` and maybe call it `flex-1`? (hmm, that SCSS file is starting to feel a bit messy)

LZoog (approved):
Probably should wait for Jared before merge, but this LGTM with one comment. Checked it out locally and it's pretty awesome.

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
jaredlockhart: Why are we using flex explicitly instead of using `Row` and `Col` to do columnar layout?

jaredlockhart (approved):
Fantastic great and wonderful.  Thank you for incorporating the feedback!  Every looks perfect and works great, everything is snappy and the behaviour all makes sense, urls look right.  Anything else I think can be done as followups but let's ship this out, it's gonna make so many workflows so much simpler now.  Thanks a million @lmorchard 🎉

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
lmorchard: Used these specific class names after wrestling with row, col, and other flex properties. This was the only combination that seemed to land on the exact semi-fixed width behavior we want for the selectors. Could be worth continuing to poke at, but maybe in a follow up

app/experimenter/nimbus-ui/src/components/PageHome/filterExperiments.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+  beforeEach(() => {
+    mockWarn = jest.fn();
+    global.console.warn = mockWarn;
+  });
+
+  afterEach(() => {
+    global.console.warn = origWarn;
+  });
+
+  it("converts JSON representation from ?filters into a filter value", () => {
```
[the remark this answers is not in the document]
lmorchard: Oh, I need to update that comment - this is no longer JSON, it's a comma-separated list in a param per filter field

lmorchard:
One more push to rebase against main and correct some test descriptions, will merge if/when CI goes green

PR #6112, opened 2021-08-03
fix #6096 feat(nimbus): branches with enabled features always have a value

### LINKED ISSUE #6096: Always show feature value on branch with feature enabled
If someone creates a NimbusFeatureConfig but leaves schema completely empty, users will not be able to enter a value for a branch at all.  There needs to at least be an empty JSON object in the schema field for the value field to be displayed.  While we initially considered the case that some features would be controlled solely through their slug and the 'enabled' flag, we have since shifted our thinking to deprecating the 'enabled' flag altogether in favour of always requiring a feature, and always passing in a value for it.  So we should just always show the value field if the feature is enabled in the branch, regardless of the schema.  Later we will deprecate the enabled flag altogether.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1565)

jaredlockhart:
Because

* In the original design, some branches would have a feature enabled but not specify a value
* Since then, the client has migrated towards always expecting a feature value rather than looking at the enabled bool
* Not all features specify a schema but should always expect a value

This commit

* Always shows the feature value input regardless of whether a schema is specified
* Always validate the feature value as json regardless of whether a schema is specified
* Continue to validate against the schema if one is specified

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -30,17 +30,21 @@ class NimbusBranchSerializer(serializers.ModelSerializer):
             )
         return value
 
+    def validate_feature_value(self, value):
+        if value:
+            try:
+                json.loads(value)
+            except Exception as e:
+                raise serializers.ValidationError(f"Invalid JSON: {e.msg}")
+        return value
+
     def validate(self, data):
         data = super().validate(data)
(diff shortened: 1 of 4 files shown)


lmorchard (approved):
LGTM! (thought I'd r+'d this yesterday but no!)

PR #6117, opened 2021-08-04
fix #6116 chore(nimbus): remove default_app_id

### LINKED ISSUE #6116: Remove default_app_id
Since we've moved towards requirig a channel that is explicitly defined for an application, we won't ever hit a case where a chanel is chosen without an app_id so we can remove the default_app_id logic.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1579)

jaredlockhart:
Because

* We now enforce that experiments must be created with a channel the application supports
* We will never fallback to the default case

This commit

* Removes default_app_id


lmorchard (approved):
:+1:

PR #6130, opened 2021-08-05
fix #5476 chore(project): delete non ldap users

### LINKED ISSUE #5476: Remove non ldap emails from database
Because of a configuration error, we introduced many emails to the database for non mozilla ldap users.  These accounts are not necessary and add unnecessary noise to any user selection drop down, so we should write a data migration to remove them.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-114)

jaredlockhart:
Because

* Long ago our auth gate was misconfigured and ~1k non mozillians accessed the production site
* This created a User entry for each person
* A small number (~5) of these users interacted with site elements which created relevant changelogs
* Even after fixing the auth gate, these user accounts persist in the production database
* This pollutes the User list with a large number of irrelevant users when selecting users from drop downs
* We should remove these unnecessary account records
* Deletion in Django will default to cascading the delete so any foreign keyed objects may also be deleted
* We need to be careful to filter any users with foreign keys out from the deletion

This commit

* Creates a data migration which deletes all users with non mozilla emails
* Excludes users with any foreign keys on to any Legacy or Nimbus models


app/experimenter/experiments/migrations/0181_remove_non_ldap_emails.py
```diff
@@ -0,0 +1,29 @@
...
+    User.objects.filter(
+        analyzed_experiments__isnull=True,
+        experimentchangelog__isnull=True,
+        experimentcomment__isnull=True,
+        nimbuschangelog__isnull=True,
+        notifications__isnull=True,
+        owned_experiments__isnull=True,
+        owned_nimbusexperiments__isnull=True,
+        subscribed_experiments__isnull=True,
+    ).exclude(email__icontains="mozilla").delete()
```
tiftran, line 17: pocket people would have @getpocket.com?

jaredlockhart: I've tried asking the auth folks for the list of all valid ldap domains and I couldn't get anything.  Good catch @tiftran 🙏

jaredlockhart: Also the fallback here is if they don't have any foreign keys, they get deleted, next time they access, account gets recreated, no data is lost.

lmorchard:
FWIW, looks like [this hit](https://app.circleci.com/pipelines/github/mozilla/experimenter/10641/workflows/d8ca1ea0-1b68-43a1-b360-3432c26ef898/jobs/31647) the same [integration test pip install error](https://app.circleci.com/pipelines/github/mozilla/experimenter/10639/workflows/f906910f-5ec1-4e96-ad1d-34bba336f801/jobs/31638) as PR #6109

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,118 +1,42 @@
...
-            **props,
-        ).id
-        for props in args
-    ]
-
-
-class TestMigration0174Forward(MigratorTestCase):
-    migrate_from = ("experiments", "0174_nimbusexperiment_results_data")
-    migrate_to = ("experiments", "0175_switch_to_status_next")
+class TestMigration0181Forward(MigratorTestCase):
```
lmorchard, line 15: Probably ignorable out-loud thinking: Do we need to delete old migration tests? Could be handy to keep them around, if only as example code. Then again, migrations shouldn't ever change or have cause to fail from other future changes - so I guess that makes the tests rather stale code. (And git history is a thing) 🤷‍♂️

app/experimenter/experiments/migrations/0181_remove_non_ldap_emails.py
```diff
@@ -0,0 +1,29 @@
...
+    User.objects.filter(
+        analyzed_experiments__isnull=True,
+        experimentchangelog__isnull=True,
+        experimentcomment__isnull=True,
+        nimbuschangelog__isnull=True,
+        notifications__isnull=True,
+        owned_experiments__isnull=True,
+        owned_nimbusexperiments__isnull=True,
+        subscribed_experiments__isnull=True,
+    ).exclude(email__icontains="mozilla").exclude(email__icontains="getpocket").delete()
```
lmorchard, line 17: Looks like codeowners roped me in as a second on data migrations. This LGTM

lmorchard (approved):
LGTM

app/experimenter/experiments/tests/test_migrations.py
jaredlockhart: No I don't think so, they only need to run once against prod and then never again, so I've always deleted them after they're used.  And yeah if you want to see them they're in the history.

PR #6143, opened 2021-08-09
fixes #6031 fix(nimbus): make results sidebar scroll with the page

### LINKED ISSUE #6031: Results table of contents doesn't scroll overflow
If the results table of contents exceeds the page height, there's no way to scroll it to see the items below the fold.  I know we had it locked so it was always visible while the right hand content is scrolled, but then you can't see the items on the left.  We could make the table of contents independently scrollable, but that seems needlessly complicated.  Maybe we should just make them a single scroll pane?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1531)

jodyheavener:
Closes #6031

Un-fix the results/launched sidebar so it scrolls with the page. This will stop Very Long Results menu items from being permanently out of view.

Look at it [here](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/fa71d7a2b62518a02120d019a21559c1d31a18b0/nimbus-ui/index.html?path=/story/components-applayoutsidebarlaunched--has-analysis-results) by making your window height smoller

Yeehaw

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
@@ -233,10 +233,7 @@ export const AppLayoutSidebarLaunched = ({
           xl="2"
           className="bg-light pt-2 border-right shadow-sm"
         >
-          <nav
-            data-testid="nav-sidebar"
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Smol change for a smol window 💯 💯 💯 

kitten-small

PR #6155, opened 2021-08-09
fix #6154 feat(nimbus): Tweak useRefetchOnError to only refetch if not in dev

### LINKED ISSUE #6154: Tweak useRefetchOnError to only refetch if not in development
Follow up to #6107 

Just realized this morning/when merging this would be preferable so we don't have to wait 5 seconds to see the error message every time when we're developing.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1607)

LZoog:
fixes #6154, this is a quick follow up to #6107 that I thought about after the fact

Because:
* In development, if an error occurs where this hook is used, devs will have to wait the refetch time to see the helpful error message (ApolloErrorAlert)

This commit:
* Skips refetching in useRefetchOnError if the environment is development and just returns ApolloErrorAlert

### DIFF
--- a/app/experimenter/nimbus-ui/src/hooks/useRefetchOnError/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useRefetchOnError/index.test.tsx
@@ -25,6 +25,14 @@ describe("useRefetchOnError", () => {
     });
     await screen.findByText("No errors!");
   });
+
+  it("does not refetch in development env", async () => {
+    Object.defineProperty(process.env, "NODE_ENV", {
+      value: "development",
+    });
+    render(<Subject />);
+    await screen.findByTestId("apollo-error-alert");
+  });
 });
 
 jest.useFakeTimers();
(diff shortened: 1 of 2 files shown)


jaredlockhart:
Wait sorry what's the rationale here?

LZoog:
@jaredlockhart Without this, if you're developing and you cause an error in the config/experiments query (and anything else in the future that may use that hook), you'll see that "this usually happens when Experimenter is in mid-deploy" message alert alert, a refetch will be tried 5 seconds later, and then once it fails again, you'll see that more helpful ApolloErrorAlert message. It seems much more useful to just display that ApolloErrorAlert right away (which has the error message details) rather than attempt to refetch since the error isn't happening from Experimenter being mid-deploy.

jaredlockhart (approved):
Oh oh I see what you mean.  On the one hand, I don't like hard coding in special cases for separate environments, more cases to think about and differences from prod.  BUT this kind of more verbose error logging on dev is exactly what Django does and it will make debugging easier so YES to this thank you! 🎉

PR #6163, opened 2021-08-10
updated docker version 

Datalog-Mindhack:


jaredlockhart (changes requested):
Oh interesting I didn't know they bumped their VMs, thank you!  But if you're going to update it please update every vm in the circle file to the same version 🙏

Datalog-Mindhack:
Sorry for the late reply, will definitely do it, thank you.

PR #6173, opened 2021-08-10
fix #5706 feat(nimbus): update is_archived in v5 serializer

### LINKED ISSUE #5706: (BE) Update NimbusExperimentSerializer to support new archived field, prevent editing while archived
Add the new archived field to [NimbusExperimentSerializer](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L267:L267) so it can be updated via GQL mutation.

Validate the archived field to prevent it from being set to true when an experiment has a Live lifecycle or non-Idle publish status.

The serializer’s main [validate](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L455:L455) method should also be updated (or maybe a new mixin created) to prevent changes to all experiment properties (including publish/lifecycle statuses) while archived is true.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1293)

jaredlockhart:
Because

* We'd like to be able to save changes to is_archived
* Archived experiments should be uneditable
* Archived experiments should be unarchivable

This commit

* Adds is_archived to the V5 serializer/mutations
* Validates that only draft/complete experiments can be archived
* Validates that archived experiments can't be edited
* Validates that archived experiments can be unarchived


lmorchard (approved):
Looks good, ship it!

PR #6174, opened 2021-08-10
fix #5708 feat(nimbus): show archived experiments on the directory view

### LINKED ISSUE #5708: (FE) Split archived experiments into new “Archived” directory view tab
Archived experiments will still be returned in the `experiments` GQL query, however we don’t want them mixed into any of the unarchived experiments.

First, update the [GQL query](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/gql/experiments.ts#L138:L138) to return the new `archived` property.

Update how we sort experiments to split archived experiments of any status into their own group. It looks like that happens [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageHome/sortByStatus.ts#L13:L13).

With the new selection of experiments, create a new [tab](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx#L41:L59) in the directory view and render them there.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1294)

jaredlockhart:
Because

* We want to show archived experiments on a separate tab on the directory view

This commit

* Queries isArchived from V5 API
* Sorts archived experiments out from other statuses
* Adds an archived tab to the directory view

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/index.test.tsx
@@ -52,11 +52,12 @@ describe("PageHome", () => {
       ["completed", screen.getByText("Completed (4)")],
       ["drafts", screen.getByText("Draft (3)")],
       ["live", screen.getByText("Live (3)")],
+      ["archived", screen.getByText("Archived (1)")],
     ] as const;
 
   it("displays five Directory Tables (one for each status type)", async () => {
     await renderAndWaitForLoaded();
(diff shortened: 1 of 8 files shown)


lmorchard (approved):
LGTM! Might be nice to throw an archived experiment or two into load_dummy_experiments. But, that could come in a later issue (e.g. EXP-1296)

PR #6180, opened 2021-08-11
fix #5709 feat(nimbus): display archived pill on experiment page

### LINKED ISSUE #5709: (FE) Add Archived pill to HeaderExperiment component
First, update the `experimentBySlug` [query](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/gql/experiments.ts#L26:L26) to return the new `archived` property.

With this new value update the [HeaderExperiment](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/HeaderExperiment/index.tsx#L17:L17) component to display an “Archived” pill in the experiment title. Refer to [this](https://www.figma.com/file/5SLa07XqQb0ttB1ay1gami/Nimbus-Console-UI?node-id=219297%3A36297) Figma design.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1295)

jaredlockhart:
Because

* We need a way to show experiments that are archived

This commit

* Adds a red Archived pill on the experiment summary page for experiments that are archived

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
@@ -140,7 +140,8 @@ const AppLayoutWithExperiment = ({
     return <PageExperimentNotFound {...{ slug }} />;
   }
 
-  const { name, startDate, computedEndDate, computedDurationDays } = experiment;
+  const { name, startDate, computedEndDate, computedDurationDays, isArchived } =
+    experiment;
 
   return (
     <Layout
@@ -169,6 +170,7 @@ const AppLayoutWithExperiment = ({
             computedEndDate,
             status,
             computedDurationDays,
+            isArchived,
           }}
         />
         {hasPollError && (
--- a/app/experimenter/nimbus-ui/src/components/HeaderExperiment/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/HeaderExperiment/index.tsx
@@ -11,7 +11,12 @@ import "./index.scss";
 
 type HeaderExperimentProps = Pick<
   getExperiment_experimentBySlug,
-  "name" | "slug" | "startDate" | "computedEndDate" | "computedDurationDays"
+  | "name"
+  | "slug"
+  | "startDate"
+  | "computedEndDate"
+  | "computedDurationDays"
+  | "isArchived"
 > & { status: StatusCheck };
 
 const HeaderExperiment = ({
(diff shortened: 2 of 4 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
```diff
@@ -169,6 +170,7 @@ const AppLayoutWithExperiment = ({
             computedEndDate,
             status,
             computedDurationDays,
+            isArchived,
```
jodyheavener, line 173: Feels like we're getting pretty close to just passing the whole experiment down lol

app/experimenter/nimbus-ui/src/components/HeaderExperiment/index.tsx
```diff
@@ -74,17 +91,25 @@ const StatusPill = ({
   label,
   active,
   padded = true,
+  weight = "primary",
+  className = "",
 }: {
   label: string;
   active: boolean;
   padded?: boolean;
+  weight?: string;
```
jodyheavener, line 100: Nit: `weight` could be a little unclear; since it's only applicable to the active class maybe `activeVariant` could be clearer (`variant` is an established prop in React Bootstrap). What do you think?

jodyheavener (commented):
One question for ya

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
jaredlockhart: Oh yeah I realized this after I wrote it.  Like I could put `isArchived` into the little handy dandy `status` thingy, but then whenever we want some other random bool we gotta stick it in there?  I like just passing the whole experiment around, I mean why not?  Also I'm not exactly sure how we have the hooks setup, but in theory can't some random component just use `useExperiment` directly without needing it from the parent?  Or does invoking the hook make the fetch all?

app/experimenter/nimbus-ui/src/components/HeaderExperiment/index.tsx
jaredlockhart: Yeah I wasn't sure what to call it

jaredlockhart: Actually looking at the bootstrap docs probably `color` is right

app/experimenter/nimbus-ui/src/components/AppLayoutWithExperiment/index.tsx
jodyheavener: Yeah useExperiment will make the fetch call. But soon, once we move to Context, we'll be able to pull experiment data from the context and not need to pass the experiment object or individual properties down as props :)

jodyheavener (approved):

PR #6187, opened 2021-08-12
fix #5715 feat(nimbus): attach archive button to mutation

### LINKED ISSUE #5715: (FE) Send mutation when un/archive button is clicked
In EXP-1296 we added the UI for toggling an experiment’s archived state. This ticket should create a new `updateExperiment` GQL mutation call that sets the experiment’s archived state.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1297)

jaredlockhart:
Because

* We want the archive button to do the archiving!

This commit

* Adds a useArchive hook to wrap the mutation
* Attaches the useArchive hook to the button in the SidebarActions
* Tests tests tests

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/LinkNav/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/LinkNav/index.tsx
@@ -17,6 +17,7 @@ type LinkNavProps = {
   className?: string;
   textColor?: string;
   title?: string;
+  onClick?: () => void;
 };
 
 export const LinkNav = ({
@@ -28,6 +29,7 @@ export const LinkNav = ({
   className = "mx-1 my-2",
   textColor,
   title,
+  onClick,
 }: LinkNavProps) => {
   const to = route ? `${BASE_PATH}/${route}` : BASE_PATH;
   // an alternative to reach-router's `isCurrent` with identical
@@ -56,6 +58,7 @@ export const LinkNav = ({
           data-sb-kind={storiesOf}
           className={classNames(textColor, "d-flex align-items-center")}
           data-testid={testid}
+          onClick={onClick}
         >
           {children}
         </Link>
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
@@ -53,6 +53,7 @@ const Subject = ({
             }
           : undefined,
         experiment,
+        refetch: async () => {},
       }}
     >
       <p data-testid="test-child">Hello, world!</p>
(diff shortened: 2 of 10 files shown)


jaredlockhart:
Look at it go!

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.stories.tsx
```diff
@@ -50,7 +52,8 @@ storiesOf("components/AppLayoutSidebarLaunched", module)
           overall: {},
           other_metrics: mockAnalysis().other_metrics,
         }}
-        {...{ experiment }}
+        experiment={experiment}
+        refetch={async () => {}}
```
jodyheavener, line 56: Since this is an empty function and it's repeated in a bunch of spots, we should either probably make it optional and just supply `async () => {}` as the prop default, or.. just make it optional and not call `refetch` if it's undefined?

jodyheavener (changes requested):
Couple thoughts, I think we should do something with all those refetchs

jaredlockhart:
Yeah I think it would be nice to change the useExperiment hook around so that it can be imported directly into arbitrary components without that triggering a call so you could import it directly without needing to pass it down.

app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/mocks.tsx
```diff
@@ -21,7 +21,7 @@ export const Subject = ({
 
   return (
     <RouterSlugProvider mocks={[mock]} path="/my-special-slug/edit">
-      <AppLayoutWithSidebar {...{ experiment }}>
+      <AppLayoutWithSidebar {...{ experiment }} refetch={async () => {}}>
```
jodyheavener, line 24: Any reason `AppLayoutWithSidebar` couldn't also have this default refetch value?

app/experimenter/nimbus-ui/src/components/LinkNav/index.tsx
```diff
@@ -17,6 +17,7 @@ type LinkNavProps = {
   className?: string;
   textColor?: string;
   title?: string;
+  onClick?: any;
```
jodyheavener, line 20: I'd like to avoid `any` where possible - I believe this could be `() => void`

jaredlockhart: Sploops! Forgot I left that in there 👍

jodyheavener (approved):
🤠

lmorchard:
> Yeah I think it would be nice to change the useExperiment hook around so that it can be imported directly into arbitrary components without that triggering a call so you could import it directly without needing to pass it down.

I think that's [EXP-1597](https://mozilla-hub.atlassian.net/browse/EXP-1597)?

jaredlockhart:
> > Yeah I think it would be nice to change the useExperiment hook around so that it can be imported directly into arbitrary components without that triggering a call so you could import it directly without needing to pass it down.
> 
> I think that's [EXP-1597](https://mozilla-hub.atlassian.net/browse/EXP-1597)?

Oh yes beauty 👏

PR #6193, opened 2021-08-12
fix #6124 fix #5556 feat(nimbus-ui): preserve / restore search param state on navigation

### LINKED ISSUE #5556: Remember last active tab on Nimbus homepage
When you click to an experiment from the homepage directory and then navigate back you are taken to the first available tab, which is currently “Active”. This is confusing and annoying.

Update the homepage directory to “remember” the last tab you were on when you left. I think we have two options for this:

* When the tab changes, update the location hash, and check for the hash on page load.
* This would also allow you to link directly to a tab.
* If you want to get really fancy, use Reach Router or fire a push state so that it updates the path instead of the hash.

* Local storage; when the tab changes store it, and check for it on page load.

I imagine the first option would be best. Either option should allow you to change to the non-first tab in the homepage directly, click to an experiment page, and then click back and be back on the tab you left from.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-834)

### LINKED ISSUE #6124: [Enhancement] Return users to the main Nimbus page on the initial tab they were on
**[Prerequisites]:**
- Be on the main Nimbus page on a tab other than “Live” (e.g. “Draft”).

**[Steps to reproduce]:**
1. Click one of the experiments to open its page.
2. Click the “Experiments” button to return to the main page.
3. Observe the tab you are on.

**[Proposed result]:**
- The user is directed to the tab they started on (i.e. “Draft”). If someone wants to review more than one experiment (in "Preview" or "Review"), see the results of multiple completed experiments, or wants to create more than one draft, needing to constantly click the same tab could get annoying.

**[Current result]:**
-  The user is always directed to the “Live” tab.

**[Notes]:**
- The same case appears when clicking “Create new” and then “Cancel” (when creating experiments).
- Attached is a recording of the current behavior.

review more than one

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1584)

lmorchard:
Because:

* We want to update the homepage directory to “remember” the last tab
  you were on when you left.

This commit:

* adds the option to useSearchParamsState to preserve search params in
  session storage on navigation and to restore the params on return with
  an empty search string.

* uses the preserve / restore option on PageHome

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
@@ -36,7 +36,7 @@ type PageHomeProps = Record<string, any> & RouteComponentProps;
 
 export const Body = () => {
   const config = useConfig();
-  const [searchParams, updateSearchParams] = useSearchParamsState();
+  const [searchParams, updateSearchParams] = useSearchParamsState("PageHome");
   const { data, loading, error, refetch } = useQuery<{
     experiments: getAllExperiments_experiments[];
   }>(GET_EXPERIMENTS_QUERY);
--- a/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.test.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.test.tsx
@@ -2,15 +2,45 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { useNavigate } from "@reach/router";
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import React from "react";
 import { RouterSlugProvider } from "../lib/test-utils";
-import useSearchParamsState from "./useSearchParamsState";
+import useSearchParamsState, { genStorageKey } from "./useSearchParamsState";
 
 // TODO: Work out how to test this stuff with @testing-library/react-hooks?
 // Depends on being wrapped by @reach/router, so that seems to make it difficult
 
 describe("hooks/useSearchParamsState", () => {
+  let originalSessionStorage: typeof window.sessionStorage;
+
+  beforeAll(() => {
+    originalSessionStorage = window.sessionStorage;
+    // @ts-ignore excuse this hackery for mocking sessionStorage
+    delete window.sessionStorage;
+    Object.defineProperty(window, "sessionStorage", {
+      writable: true,
+      value: {
+        getItem: jest.fn().mockName("getItem"),
+        setItem: jest.fn().mockName("setItem"),
+      },
+    });
+  });
+
+  beforeEach(() => {
+    // @ts-ignore excuse this hackery for mocking sessionStorage
+    sessionStorage.getItem.mockClear();
+    // @ts-ignore excuse this hackery for mocking sessionStorage
+    sessionStorage.setItem.mockClear();
+  });
+
+  afterAll(() => {
+    Object.defineProperty(window, "sessionStorage", {
+      writable: true,
+      value: originalSessionStorage,
+    });
+  });
+
   it("returns the current search parameters", async () => {
     const expected = { foo: "bar", baz: "quux" };
     const params = new URLSearchParams(expected);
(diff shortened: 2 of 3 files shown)


jaredlockhart:
Can't we just store the parameters in a useState somewhere without needing to store them in session/local storage?  I'd like to avoid persistent storage as much as we can.

jodyheavener:
> Can't we just store the parameters in a useState somewhere without needing to store them in session/local storage? I'd like to avoid persistent storage as much as we can.

Maybe a state + context rendered really high up? Also kinda curious why you no like persistent storage because I have another thing coming down the line that uses it

lmorchard:
> Can't we just store the parameters in a useState somewhere without needing to store them in session/local storage? I'd like to avoid persistent storage as much as we can.

There's not really a place for that state, unless maybe we add an overall app context with the state. State for PageHome is nuked whenever it's removed from rendering. We can do that, but the storage here does essentially the same thing and gets cleaned up on tab close.

jaredlockhart:
@jodyheavener what's the thing?  I'm not a fan of persistent storage because it gets in the way of resetting your flow for local dev and integration testing.  There's some times you have to store stuff in there like if you're handling auth/sessions right in your app, but since that's handled outside our code I think everything we're doing should be essentially stateless.  As much as possible I think we should try to make everything stateless and put parameters in queryparams so the flow always goes

1) load params in from queryparams
2) rehydrate internal state
3) proceed as normal
4) user updates something
5) update internal state, query params
6) go to either 3 within a session or 1 if we're at a fresh navigation

As long as we can stick to that cycle we don't have much to worry about

If we already unpack the query params from the url into an internal obj, we can stick that in a useState somewhere, and then just use it to format the link on the summary page back to the directory page, right?

jaredlockhart:
> > Can't we just store the parameters in a useState somewhere without needing to store them in session/local storage? I'd like to avoid persistent storage as much as we can.
> 
> There's not really a good local place for that state, unless maybe we add an overall App context with the state. State for PageHome is nuked whenever it's removed from rendering. We can add that App-level state context, but the storage here does essentially the same thing and gets cleaned up on tab close.

Yeah sessionStorage isn't too bad because it cleans itself up quickly like you said but are we just using it here as a shortcut around useState/context becuase it's less code?

lmorchard:
> Yeah sessionStorage isn't too bad because it cleans itself up quickly like you said but are we just using it here as a shortcut around useState/context becuase it's less code?

Well, less code, and I think a context wrapping the whole app to manage `useState` means the whole app re-renders whenever that state changes. That will probably have some performance impact and side effects like making [this change I just merged](https://github.com/mozilla/experimenter/commit/59eaa667e796bf426bd8c9a0b2166f862ff1e773) re-fetch data every time a tab, sort, or filter changes (i.e. because the parent component re-rendered)

jaredlockhart (approved):
Yeah works perfectly locally, seems reasonable if it's scoped to a session, maybe later when we do more refactors to move things into contexts we'll be able to lift this in there more easily.  Thnx @lmorchard 🎉

lmorchard:
> a context wrapping the whole app to manage useState means the whole app re-renders whenever that state changes

Probably worth a follow-up issue, because some more head scratching can probably do better than that first thought of mine. Maybe possible to use `useRef` for a non-changing value ref whose *content* changes and only gets checked in certain conditions like we do with sessionStorage now

lmorchard:
Actually, the more I looked at it, the more I disliked the hacks to make this sessionState thing work. [This combo of `useRef` and `useContext`](https://github.com/mozilla/experimenter/pull/6193/commits/c302a05172384acdf909c88a26ce8d00768efb74) should do essentially the same thing and without the performance / side effect issues I'd initially thought up.

If we like that, I can squash it before merging.

lmorchard:
Actually, since I already got an r+ here and want to start on a demo, I'm gonna merge this and file the follow up issue with a quick PR. Then we can get the feature landed and maybe beat up on the enhancement as a fast-follow

jaredlockhart:
Very reasonable plan @lmorchard 💯 💯 💯 💯

lmorchard:
There we go, part 2: https://github.com/mozilla/experimenter/pull/6202

PR #6196, opened 2021-08-12
fixes #6195 feat(reporting): add link to legacy to get to reporting

### LINKED ISSUE #6195: [Reporting]: Add a link to legacy home page so reporting is accessible
We should have way that users can navigate to reporting from the legacy home page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1626)

tiftran:
Because

* It would allow users to navigate to reporting

This commit:

* adds a link to legacy home page to reporting

### DIFF
--- a/app/experimenter/legacy-ui/templates/base.html
+++ b/app/experimenter/legacy-ui/templates/base.html
@@ -69,7 +69,7 @@
               {{ request.user.subscribed_experiments.count }} Subscribed
               Deliver{{ request.user.subscribed_experiments.count|pluralize:"y,ies" }}
             </a>
-            <a class="nocolorstyle d-block" href="{% url "home" %}reporting">
+            <a class="nocolorstyle d-block" href="{% url "reporting" %}">
               <span class="fas fa-table"></span>
               Go to Reporting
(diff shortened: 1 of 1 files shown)


tiftran:
wasn't entirely sure if this was the best place for it...
Screen Shot 2021-08-12 at 12 20 34 PM

app/experimenter/legacy-ui/templates/base.html
```diff
@@ -69,6 +69,10 @@ <h3 class="pt-2">
               {{ request.user.subscribed_experiments.count }} Subscribed
               Deliver{{ request.user.subscribed_experiments.count|pluralize:"y,ies" }}
             </a>
+            <a class="nocolorstyle d-block" href="{% url "home" %}reporting">
```
jaredlockhart, line 72: Ah stick a name on the url in urls.py and then reverse it here 👍

jaredlockhart (changes requested):

jaredlockhart (approved):
Oh yeah it already has a name XD

PR #6197, opened 2021-08-12
fix #6134 fix(nimbus-ui): always fetch experiments when navigating to directory view

### LINKED ISSUE #6134: Experiments that change states are displayed in their respective directory views only after the page is refreshed
**[Steps to reproduce]:**
1. Navigate to the Nimbus page and create a new experiment.
2. Fill all fields with valid data and save the experiment.
3. Click the "< Experiments" button.
4. Click the "Draft (#)" directory view.
5. Search for the newly created experiment in the list.

**[Expected result]:**
- The experiment is displayed in the list.

**[Actual result]:**
- The experiment is not found.

**[Notes]:**
- The experiment is displayed only after the page is refreshed. 
- The issue is reproducible whenever the state of the experiment is changed (Draft to Preview, Preview - Review, Review - Live, Live - Complete).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1590)

lmorchard:
Because:

* we want to see the directory view updated after performing an action
  that changes an experiment's status

This commit:

* uses the "network-only" fetch policy for the query fetching
  experiments for the directory view, which should cause a fresh fetch
  every time navigation to the page happens

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
@@ -39,7 +39,7 @@ export const Body = () => {
   const [searchParams, updateSearchParams] = useSearchParamsState();
   const { data, loading, error, refetch } = useQuery<{
     experiments: getAllExperiments_experiments[];
-  }>(GET_EXPERIMENTS_QUERY);
+  }>(GET_EXPERIMENTS_QUERY, { fetchPolicy: "network-only" });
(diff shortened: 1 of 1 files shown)


lmorchard:
I *think* this should do the trick without any undue side-effects. Should result in a fresh fetch only when the component is rendered overall - i.e. not as tab navigation happens, but only on initial page load or a return to visibility on a within-app navigation event.

jodyheavener (approved):
neat!

PR #6199, opened 2021-08-12
fix #6198 bug(nimbus): support old changelog data

### LINKED ISSUE #6198: Unable to archive some experiments
Attempting to archive some experiments on prod results in 

```json
{
	"errors": [{
		"message": "'published_dto'",
		"locations": [{
			"line": 2,
			"column": 3
		}],
		"path": ["updateExperiment"]
	}],
	"data": {
		"updateExperiment": null
	}
}
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1627)

jaredlockhart:
Because

* We store a serialization of the experiment on each changelog
* As we introduce new fields, those old serializations will not contain them
* If we build logic that inspects these fields, we have to be careful that some field may not exist on old changelogs

This commit

* Changes key lookups to expect they may not exist


lmorchard (approved):
:+1:

PR #6202, opened 2021-08-12
fix #6201 refactor(nimbus-ui): use App-global storage for useSearchParamsState

### LINKED ISSUE #6201: Refine state management for useSearchParamsState to use App global state rather than sessionStorage
Per https://github.com/mozilla/experimenter/pull/6193#issuecomment-897939204:

> I'm not a fan of persistent storage because it gets in the way of resetting your flow for local dev and integration testing. There's some times you have to store stuff in there like if you're handling auth/sessions right in your app, but since that's handled outside our code I think everything we're doing should be essentially stateless. 

Let's see if we can rework storage for useSearchParamsState to *not* use sessionStorage

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1629)

lmorchard:
Because:

* sessionStorage may offer some undesirable properties for this hook

This commit:

* switches useSearchParamsState over to use state managed by an
  App-global Context provider

### DIFF
--- a/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.tsx
@@ -3,28 +3,49 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { useLocation, useNavigate } from "@reach/router";
-import { useEffect, useMemo } from "react";
+import React, {
+  createContext,
+  MutableRefObject,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+} from "react";
 
-export const genStorageKey = (subkey: string) => `searchParamState:${subkey}`;
+export type SearchParamStorage = Record<string, string>;
+export type SearchParamsContextState =
+  | MutableRefObject<SearchParamStorage>
+  | undefined;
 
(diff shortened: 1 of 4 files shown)


lmorchard:
Figured this would be better as a quick follow-up since it might not be a perfect solution without some extra attention

app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.tsx
```diff
@@ -3,28 +3,49 @@
...
+  createContext<SearchParamsContextState>(undefined);
+
+export const SearchParamsStateProvider: React.FC = ({ children }) => {
+  const storage = useRef<SearchParamStorage>({});
+  return (
+    <SearchParamsContext.Provider value={storage}>
+      {children}
+    </SearchParamsContext.Provider>
+  );
+};
```
lmorchard, line 30: An alternative I'd considered for this was to just define a context with a ref in the App component itself. But, I thought creating this hook-specific context provider would help contain the madness to this module and naturally scope the data stored in the ref.

jaredlockhart (approved):
Booyah, looks great, works great, love it ❤️

lmorchard:
Gonna just merge this so I can drop another bugfix after it next

Issue #6200, opened 2021-08-12
Experiment end date can be incorrect if overridden in external config


scholtzan:
External configuration files allow to change the `end_date` of the experiment. For example, for `1-click-pin-experiment` see https://github.com/mozilla/jetstream-config/blob/cbf385c231ce8d830ccf22153a753c580047ff2e/1-click-pin-experiment.toml#L6

Experimenter still shows the end date that was configured when the experiment was initially created:

![Screen Shot 2021-08-12 at 14 44 12](https://user-images.githubusercontent.com/1239705/129274178-683fa511-492a-487f-8f57-8791b527c1a2.png)

I've had a couple of cases where stakeholders/people looking at the result were confused and thought results were missing or experiments should have already completed.

This is related to an issue we found with configs overriding the `reference_branch`: https://github.com/mozilla/experimenter/issues/6129
One way of solving this is adding the `end_date` to the experiment metadata that Jetstream produces and use it as the source of truth.

LZoog:
@scholtzan What else can be overridden in the config?

We show the end date on the summary page too. Right now the metadata is returned with the analysis fetch, and if the experiment has been launched, we hit that endpoint to check if analysis data is available yet; so while it'd require some refactoring we _can_ rework the end date to point at the metadata on the analysis page and summary page, but is this something that will keep coming up with other experiment properties? If it is, I wonder if we should or if it's plausible to pull in that metadata and change the experiment data from within Experimenter so it matches and then we don't have to have a bunch of conditionals around if there's overrides in the metadata. 😕

scholtzan:
The config can override `reference_branch`, `end_date`, `start_date`, and `proposed_enrollment`. I think more properties could be added to that list in the future, but nothing is planned at the moment.
Yeah, updating the experiment settings in Experimenter would probably be ideal.

lmorchard:
This is probably a function of me not yet being familiar with this part of Nimbus, but: Does this point to a general problem with source of truth for an experiment?

*Should* the end date of an experiment be overridden in external config? What's the use case & rationale for that? 

Not asserting that this *shouldn't* happen, so much as trying to understand when & why it would happen. e.g. Is it expected that Experimenter Console should poll external configs to update its database and published DTOs? Do changes need to be pushed to Remote Settings based on external configs?

scholtzan:
[...]

Issue #6071, opened 2021-07-27
Enable relative/absolute Results toggle for entire page


data-sync-user:
In EXP-1404, we're implementing a toggle for absolute/relative values in the Results Summary portion of the Results page.

Let's extend this functionality to include the entire page in case users want to see an absolute comparison for the other tables and because otherwise users may be confused at what comparison they're looking at in other parts of the page.

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/EXP-1551)

PR #6204, opened 2021-08-13
fix #6129 refactor/feat(nimbus): Results page - refactor with context, reference metadata for control branch name

### LINKED ISSUE #6129: Get control branch for visualization from experiment metadata
The reference_branch can be changed via the config files living in jetstream-config. This can result in the visualization breaking since Experimenter uses a different control branch (the one initially configured when the experiment was created) than jetstream did use for the analysis (the one from the custom config).

To fix this, Jetstream will add the `reference_branch` it used for the analysis to the experiment metadata. Experimenter can use this `reference_branch` as the source of truth when creating the visualization. Experimenter already uses this metadata to get the metric descriptions.

Blocked on https://github.com/mozilla/jetstream/issues/840

cc @LZoog

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1587)

LZoog:
fixes #6129 

I'm opening this as a draft in case someone wants to take an early peak since this is a mid-large PR. I separated the commits to make it easier to see the refactor bits and the actual issue fix - at the time of writing, storybook + tests all work in the first commit, but I need to work on the stories + tests for the second commit, which shouldn't take too long tomorrow. I'm probably going to punt the story refactor-to-new-syntax to #6071.

The refactor felt necessary now because we were already prop-drilling several things down from `PageResults` into the tables, to fix this issue we would have needed another prop drill or to feed every table the entire `analysis` object where the `metadata` exists (which can be JSON 40k+ lines long), and then we would have needed yet another prop drilled down for #6071.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricConversion/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricConversion/index.tsx
@@ -89,7 +89,13 @@ const TableMetricConversion = ({
                       results={results[branch]}
                       group={GROUP.OTHER}
                       tableLabel={TABLE_LABEL.PRIMARY_METRICS}
-                      {...{ metricKey, displayType, branchComparison, bounds }}
+                      // isControlBranch={branch === controlBranchName}
+                      {...{
+                        metricKey,
+                        displayType,
+                        branchComparison,
+                        bounds,
+                      }}
                     />
                   ),
                 )}
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricConversion/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TableMetricConversion/index.tsx
@@ -2,14 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React from "react";
+import React, { useContext } from "react";
+import { ResultsContext } from "../../../lib/contexts";
 import {
   CONVERSION_METRIC_COLUMNS,
   DISPLAY_TYPE,
   GROUP,
   TABLE_LABEL,
 } from "../../../lib/visualization/constants";
-import { AnalysisDataOverall } from "../../../lib/visualization/types";
 import { getExtremeBounds } from "../../../lib/visualization/utils";
 import { getConfig_nimbusConfig_outcomes } from "../../../types/getConfig";
 import TableVisualizationRow from "../TableVisualizationRow";
@@ -22,9 +22,7 @@ type ConversionMetricStatistic = {
 };
 
 type TableMetricConversionProps = {
-  results: AnalysisDataOverall;
   outcome: getConfig_nimbusConfig_outcomes;
-  sortedBranches: string[];
 };
 
 const getStatistics = (slug: string): Array<ConversionMetricStatistic> => {
@@ -41,16 +39,18 @@ const getStatistics = (slug: string): Array<ConversionMetricStatistic> => {
   return conversionMetricStatisticsList;
 };
 
-const TableMetricConversion = ({
-  results = {},
-  outcome,
-  sortedBranches,
-}: TableMetricConversionProps) => {
+const TableMetricConversion = ({ outcome }: TableMetricConversionProps) => {
+  const {
+    analysis: { overall },
+    sortedBranches,
+    controlBranchName,
+  } = useContext(ResultsContext);
+  const overallResults = overall!;
   const conversionMetricStatistics = getStatistics(outcome.slug!);
   const metricKey = `${outcome.slug}_ever_used`;
   const bounds = getExtremeBounds(
     sortedBranches,
-    results,
+    overallResults,
     outcome.slug!,
     GROUP.OTHER,
   );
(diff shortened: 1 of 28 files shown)


app/experimenter/nimbus-ui/src/components/PageResults/TableMetricConversion/index.tsx
```diff
@@ -89,7 +89,13 @@ const TableMetricConversion = ({
                       results={results[branch]}
                       group={GROUP.OTHER}
                       tableLabel={TABLE_LABEL.PRIMARY_METRICS}
-                      {...{ metricKey, displayType, branchComparison, bounds }}
+                      // isControlBranch={branch === controlBranchName}
```
LZoog, line 92: @emtwo it doesn't look like we're actually using this table anywhere. Should we remove it or am I missing something (maybe we plan to use it again later)? 😬

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -73,12 +75,18 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
         // For testing - users will be redirected if the analysis is unavailable
         // before reaching this return, but tests reach this return and
         // analysis.overall is expected to be an object (EXP-800)
-        if (analysisUnavailable(analysis)) return;
+        if (!analysis || analysisUnavailable(analysis)) return;
```
LZoog, line 78: I took `!analysis` out of the `analysisUnavailable` check and placed it here to silence Typescript down the entire page (I was able to remove a lot of `?`s and `!` where we know `analysis` can't be undefined due to this check). `analysisAvailable` checks if it's undefined by way of `analysis?.whatever` chaining so it shouldn't change anything functionally.

app/experimenter/nimbus-ui/src/lib/utils.ts
```diff
@@ -19,3 +19,10 @@ export const optionalBoolString = (
   }
   return String(value);
 };
+
+// function getKeyByValue(
```
LZoog, line 23: Oops, thought I needed this at one point but I don't, will remove tomorrow when I push again.

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -47,23 +68,24 @@ export const getTableDisplayType = (
   return displayType;
 };
 
+// Returns [control branch name, reference branch names]
```
LZoog, line 71: treatment branch names*

app/experimenter/nimbus-ui/src/components/PageResults/TableMetricConversion/index.tsx
emtwo: Initially we were using it for primary metrics but it was a bit hard coded. This used to be called `TableMetricPrimary` but it only had conversion metrics. However, primary metrics are not always conversion metrics and now primary metrics are defaulting to `TableMetricCount`. 

In the future, Jetstream would specify whether a primary or secondary metric is a count or conversion metric (or something else) and that's how we would decide what table to use.Right now there is no logic in there for making that decision because it's not part of Jetstream yet either

LZoog: I'll include this in the Context refactor then since it sounds like we may need to use it later, thanks. 😄

LZoog:
Sooo, #6200 was just filed and highlights a larger problem here... more than just the `reference_branch` can be overridden in the metadata. I think the more ideal solution is going to be something like updating experiment data to match Jetstream overrides, but that seems like a task that needs to be thought through and probably discussed.

Since I'm working to finish this patch and it fixes a problem with weekly tables in production now, I think it should be merged for now - I still think using Context here was the right call, it really cleaned some things up and eliminated some prop drilling/future prop drilling on this page, but some of the logic added in my 2nd commit can probably be taken out if we go the updating experiment data to match jetstream route to fix #6200.

LZoog:
There's two tests I'm having trouble getting passing (the ones I added with `selector: table tr:first-of-type`) and I need to properly fill out the commit message, but this is otherwise ready.

I wanted to actually finish this prior to the weekend since I'm out on Monday but it's late and I have to jet in 5 minutes to another obligation. I'll wrap those tests up and address any comments on Tuesday.

jodyheavener (commented):
It might be too much to try and re-do this into separate commits, so def feel free to push back: given the size and number of places this touches, I wonder if this PR could be split up a little? One for adding the context setup/provider, one for mock utils, then ones for the various areas it touches? We have [an epic](https://mozilla-hub.atlassian.net/browse/EXP-1597) for exactly this, but with experiment data.

LZoog:
> It might be too much to try and re-do this into separate commits, so def feel free to push back: given the size and number of places this touches, I wonder if this PR could be split up a little? One for adding the context setup/provider, one for mock utils, then ones for the various areas it touches? We have an epic for exactly this, but with experiment data.

I could split this up by way of Context vs metadata fix (that's why I sorta left this hanging out as two commits, and could still do that), but I don't know about further than that. Adding the setup/provider was only maybe two files which could be separate, but I don't think we can separate mock utils/implementation or at least not here since we need the mock for tests.

Looks like I've got a meeting on my calendar to discuss the metadata today, so I'll go ahead and pull out the Context upgrade here and put it in as a new PR.

LZoog:
I'm actually going to close this now to prevent any confusion. There's going to be a bit more (not much though) needed from my 2nd commit to close this issue but I'll have that up after #6231 is merged and I work off that.

PR #6205, opened 2021-08-13
fixes #5768 fix(nimbus): changed percent of clients step increment to one

### LINKED ISSUE #5768: Remove .0001 step increment in "Percent of Clients" or change to 1
I ran [this comment](https://github.com/mozilla/experimenter/issues/4298#issuecomment-833740133) by Ana...

> While we're at it here, is `.0001` a good increment in the input UI? I kind of wonder if hitting the "up" arrow key or clicking the "up" arrow in the input make sense to step up/down in value more substantially, like 1-2 decimal places, but I'm not sure what would be preferred by users.

... she would prefer that we remove the step and up/down arrows and have users manually enter the number they want, but if that's messy to do then we can just change the step to `1` which is more usable than `.0001`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1411)

anthonytran7757:
Because

* the percent of clients step increment of 1 is preferred over .0001

This commit 

* changes the percent of clients step increment to 1

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
@@ -248,7 +248,7 @@ export const FormAudience = ({
                 type="number"
                 min="0"
                 max="100"
-                step="0.0001"
+                step="1"
               />
               <InputGroup.Append>
                 <InputGroup.Text id="populationPercent-unit">%</InputGroup.Text>


jodyheavener (changes requested):
Hi @anthonytran7757 this looks good to me, however can I get you to format your commit messages according to our [Git Commit Guidelines](https://github.com/mozilla/experimenter/blob/main/contributing.md#git-commit-guidelines)? It could look just as you have titled the PR itself.

jaredlockhart:
Thank you for picking this up @anthonytran7757 !

jodyheavener (approved):
Thanks @anthonytran7757 !

PR #6209, opened 2021-08-13
fix #6207 fix(nimbus-ui): replace history entry when restoring search query

### LINKED ISSUE #6207: After sorting/filtering you can't navigate back to an experiment after using the "< Back to Experiments" button
**[Steps to reproduce]:**
1. Navigate to the Nimbus homepage and sort the page a few times.
2. Create a new experiment.
3. Fill the fields with valid data and save the experiment.
4. Click the "< Back to Experiments" button.
5. Click the Browser's "Back" button and observe the behavior.

**[Expected result]:**
- The previously created experiment page is displayed.

**[Actual result]:**
- The homepage is refreshed.

**[Additional Notes]:**
- The issue is also reproducible when navigating to the homepage from an already existing experiment.
- You can get back to the experiment page through the Back button's context menu.

cant go back to an experiment after using sorting or filtering

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1631)

lmorchard:
Because:

* restoring search params interfered with back button navigation

This commit:

* switches to `{ replace: true }` when restoring search params so that
  an additional history entry is not created

* improves tests for useSearchParamsState to assert exact router
  navigation call parameters

### DIFF
--- a/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.tsx
@@ -42,7 +42,7 @@ export function useSearchParamsState(storageKey?: string) {
     } else {
       const savedParams = storage.current[storageKey];
       if (savedParams) {
-        navigate(`${location.pathname}?${savedParams}`);
+        navigate(`${location.pathname}?${savedParams}`, { replace: true });
       }
     }
   }, [navigate, storage, storageKey, location.pathname, location.search]);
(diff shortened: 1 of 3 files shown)


app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.tsx
```diff
@@ -42,7 +42,7 @@ export function useSearchParamsState(storageKey?: string) {
     } else {
       const savedParams = storage.current[storageKey];
       if (savedParams) {
-        navigate(`${location.pathname}?${savedParams}`);
+        navigate(`${location.pathname}?${savedParams}`, { replace: true });
```
lmorchard, line 45: This is the actual fix, just a one-liner. The rest of this is mainly test enhancements to assert exactly when navigate gets called with `{ replace: true }`

jodyheavener (approved):
🤠

PR #6210, opened 2021-08-13
fix #6206 fix(nimbus-ui): constrain max width of filter bar selectors

### LINKED ISSUE #6206: Entering a string much longer than a Nimbus filter field will make it overflow out of its section until deselected
**[Prerequisites]:**
- Have the Nimbus main page open.

**[Steps to reproduce]:**
1. Click the Feature filter field.
2. Enter a very long string (e.g. 128 characters).
3. Observe the filter fields.

**[Expected result]:**
- Entering a long string doesn’t cause UI issues.

**[Actual result]:**
- Entering a long string will expand the field until it pushes either other fields or itself off their section.

**[Notes]:**
- The fields return to their default sizes after clicking out of them.
- If the page has been resized to be smaller, the number of characters required to reproduce this issue decreases.
- The fields in the Metrics section don’t seem to display this behavior, while the fields on Audience will enlarge until they move to a new line and start behaving like the fields on Metrics.

long string

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1630)

lmorchard:
Because:

* a filter bar selector with extremely wide contents wrecks the layout
 of the overall bar

This commit:

* applies a max-width of 25% to each filter bar selector

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
@@ -98,10 +98,7 @@ const FilterSelect = <T extends NullableObjectArray>({
   );
 
   return (
-    <Nav.Item
-      className="m-1 text-left flex-grow-1 flex-shrink-1"
-      style={{ flexBasis: "0px" }}
-    >
+    <Nav.Item className="m-1 mw-25 text-left flex-basis-0 flex-grow-1 flex-shrink-1">
       <label className="ml-1 mr-1" htmlFor={`filter-${filterValueName}`}>
         {fieldLabel}
       </label>
--- a/app/experimenter/nimbus-ui/src/styles/index.scss
+++ b/app/experimenter/nimbus-ui/src/styles/index.scss
@@ -71,6 +71,14 @@ select.form-control.is-warning {
   font-weight: 600;
 }
 
+.flex-basis-0 {
+  flex-basis: 0px;
+}
+
+.mw-25 {
+  max-width: 25%;
+}
+
 .w-33 {
   width: 33.33%;
 }


app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
```diff
@@ -100,7 +100,7 @@ const FilterSelect = <T extends NullableObjectArray>({
   return (
     <Nav.Item
       className="m-1 text-left flex-grow-1 flex-shrink-1"
-      style={{ flexBasis: "0px" }}
+      style={{ flexBasis: "0px", maxWidth: "25%" }}
```
lmorchard, line 103: This feels kind of grungy. I tried using [a sizing class of `mw-25`](https://getbootstrap.com/docs/4.0/utilities/sizing/), but that seemed to do nothing. Also don't like the arbitrary 25%, since that will need changing if/when we add or remove filter selectors. But, this seemed to clamp down on the expansion without wrecking the responsive layout

jodyheavener: Looks like `mw-25` doesn't exist; [only](https://github.com/twbs/bootstrap/blob/main/dist/css/bootstrap-utilities.css#L542-L544) `mw-100` exists.

How about we make our own `mw-25` and stick it with the [other](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/styles/index.scss) hand-rolled SCSS? If nothing else but to help keep custom styles class-based

jodyheavener (commented):
One suggestion for ya

app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
lmorchard: Ohhh, I guess I got that confused with `w-25` ugh

lmorchard:
Moved both those two inline styles into `index.scss`, seems to do the trick

jodyheavener (approved):
🙌

PR #6211, opened 2021-08-13
fix #6141 chore(nimbus): use the same eslint config between react-scripts and eslint

### LINKED ISSUE #6141: react-scripts and lint:eslint use different ESLint configs
When we set up the `create-react-app`, instead of extending/using the existing ESLint config provided by `react-scripts` we went and added our own `.eslintrc.js`, which is what `yarn lint:eslint` (which has the power to fail CI) is using.

As a result our code is passing `lint:eslint` but is throwing all kinds of warnings when `react-scripts` (`yarn start`) is running:

```
./src/components/AppLayoutWithSidebar/index.scss (/app/node_modules/react-scripts/node_modules/css-loader/dist/cjs.js??ref--6-oneOf-5-1!/app/node_modules/postcss-loader/src??postcss!/app/node_modules/resolve-url-loader??ref--6-oneOf-5-3!/app/node_modules/react-scripts/node_modules/sass-loader/dist/cjs.js??ref--6-oneOf-5-4!./src/components/AppLayoutWithSidebar/index.scss)
Greetings, time traveller. We are in the golden age of prefix-less CSS, where Autoprefixer is no longer needed for your stylesheet.

./src/components/PageHome/DirectoryTable/index.tsx
src/components/PageHome/DirectoryTable/index.tsx
  Line 124:8:  React Hook useCallback has a missing dependency: 'descending'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

./src/hooks/useSearchParamsState.ts
src/hooks/useSearchParamsState.ts
  Line 19:6:  React Hook useMemo has a missing dependency: 'location.pathname'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

Some options:

* Try to disable the `react-scripts` linter since we don't need two and `lint:eslint` is already set up for CI (but we should probably copy over some of its config like `react-hooks/exhaustive-deps` since they're useful)
* Find a way for `lint:eslint` to extend and use the `react-script` config - this could be possible via `.rescriptsrc.js`

If we decide to do nothing, though, we should at least be monitoring the output of the `react-scripts` linter since it's providing useful feedback.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1596)

lmorchard:
Because:

* our code is passing lint:eslint but is throwing all kinds of warnings
  when react-scripts (yarn start) is running

This commit:

* switches away from .eslintrc.js to a revised config in package.json
  that extends from eslint-config-react-app from create-react-app

* fixes a few small new lint issues that popped up


lmorchard:
This was bothering me, so... I think this fixes it?

app/experimenter/nimbus-ui/package.json
```diff
@@ -103,7 +103,30 @@
...
+      "@typescript-eslint/no-non-null-assertion": "off",
+      "@typescript-eslint/camelcase": "off",
+      "@typescript-eslint/ban-ts-ignore": "off",
+      "@typescript-eslint/no-empty-function": "off",
+      "react/display-name": "off",
+      "react/prop-types": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/explicit-module-boundary-types": "off",
+      "@typescript-eslint/ban-ts-comment": "off",
+      "import/no-anonymous-default-export": "off"
```
lmorchard, line 128: These don't seem redundant with [the rules in eslint-config-react-app](https://github.com/facebook/create-react-app/blob/main/packages/eslint-config-react-app/index.js)? Weirdly, I picked up a few *new* lint issues that seemed legit, but not sure why we didn't see them before

jodyheavener (approved):

PR #6215, opened 2021-08-16
fix(numbus): update out-of-date yarn lock file

jodyheavener:
yarn.lock was out of date on main for me

before approving please verify this is also true on your side and that it's not just me or my yarn setup


jaredlockhart (approved):
Oh yeah same on mine I'll bet we forgot to do this when we added the reporting frontend, good catch 👍

PR #6226, opened 2021-08-16
 fixes #6194 feat(nimbus): Consolidate Results and Grafana columns 

### LINKED ISSUE #6194: Consolidate Results and Grafana columns
In the Directory View, conslidate the Results and Grafana columns into a single column, conditionally displaying either link. The column title can still be titled "Results". This will allow us to remove a column.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1625)

lennartschoch:
Because

* we want to keep our tables clean and readable

This commit

* consolidates the "results" and "monitoring" columns


app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -301,6 +286,13 @@ export const DirectoryLiveTable: React.FC<DirectoryTableProps> = (props) => (
               >
                 Results
               </Link>
+            ) : experiment.monitoringDashboardUrl ? (
```
jaredlockhart, line 289: Oh I see, yeah I wouldn't do it as either results OR monitoring, while an experiment is live both could be available, this should just be a second check so if both results and monitoring are available both links appear 🙏

jaredlockhart (changes requested):
Thank you for picking this up @lennartschoch Just a small change pls 🙏

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
lennartschoch: Good point, @jaredlockhart! I wasn't sure if that condition was possible. Fixed in https://github.com/mozilla/experimenter/pull/6226/commits/f8393dbccf87555106eda13de92b3face880caf8.

jaredlockhart (approved):
Yes!! Perfect thank you 🙏

PR #6231, opened 2021-08-17
fix #6230 refactor(nimbus): Use Context on the Results page

### LINKED ISSUE #6230: Refactor Results page to use Context
I did this as part of EXP-1587 but I'm separating this out for reviewing purposes and in case the solution tacked on to that PR needs to be tweaked (which we should know from a meeting later today). Will reduce points over there to reflect this.

See context for why this refactor was needed in PR #6204.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1640)

LZoog:
Because:
* We are already prop-driling several things down from PageResults into the tables, and at least two upcoming visualization fixes and tasks would need to do the same. This change cleans up the prop-drilling by adding ResultsContext referenced by tables on this page and sets us up for easier work in the future.

This commit:
* Creates a ResultsContext set in PageResults and consumes what's needed in each table
* Creates a mock context, updates stories and tests
* Contains other minor cleanups

---

Original PR for this was #6204. This was split up 1) for review purposes and 2) because we're discussing Jetstream metadata later today, I'll create a follow up PR for the remaining work in #6204 and will close that first PR at that time.

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import classNames from "classnames";
 import React, { useState } from "react";
 import Collapse from "react-bootstrap/Collapse";
 import vegaEmbed, { VisualizationSpec } from "vega-embed";
@@ -104,8 +105,8 @@ const GraphsWeekly = ({
 
   const [open, setOpen] = useState(false);
   const [embedded, setEmbedded] = useState(false);
-  const graphsVisibleClass = !open ? "d-none" : "";
-  const graphsHiddenClass = open ? "d-none" : "";
+  const graphsVisibleClass = classNames(!open && "d-none");
+  const graphsHiddenClass = classNames(open && "d-none");
 
   const handleCollapse = () => {
     setOpen(!open);
--- a/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
@@ -8,6 +8,7 @@ import Collapse from "react-bootstrap/Collapse";
 import { useConfig } from "../../hooks";
 import { ReactComponent as CollapseMinus } from "../../images/minus.svg";
 import { ReactComponent as ExpandPlus } from "../../images/plus.svg";
+import { ResultsContext, ResultsContextType } from "../../lib/contexts";
 import {
   BRANCH_COMPARISON,
   GROUP,
@@ -73,12 +74,17 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
         // For testing - users will be redirected if the analysis is unavailable
         // before reaching this return, but tests reach this return and
         // analysis.overall is expected to be an object (EXP-800)
-        if (analysisUnavailable(analysis)) return;
+        if (!analysis || analysisUnavailable(analysis)) return;
+
+        const resultsContextValue: ResultsContextType = {
+          analysis,
+          sortedBranches: getSortedBranches(analysis),
+        };
 
         const slugUnderscored = experiment.slug.replace(/-/g, "_");
-        const sortedBranches = getSortedBranches(analysis!);
+
         return (
-          <>
+          <ResultsContext.Provider value={resultsContextValue}>
             <LinkMonitoring {...experiment} />
             <h3 className="h5 mb-3 mt-4" id="overview">
               Overview
@@ -95,12 +101,7 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
             </p>
             <h3 className="h6">Hypothesis</h3>
             <p>{experiment.hypothesis}</p>
-            {analysis?.overall && (
-              <TableHighlights
-                results={analysis}
-                {...{ experiment, sortedBranches }}
-              />
-            )}
+            {analysis.overall && <TableHighlights {...{ experiment }} />}
             <TableHighlightsOverview {...{ experiment }} />
 
             <div id="results_summary">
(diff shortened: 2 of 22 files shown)


app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
```diff
@@ -104,8 +107,8 @@ const GraphsWeekly = ({
 
   const [open, setOpen] = useState(false);
   const [embedded, setEmbedded] = useState(false);
-  const graphsVisibleClass = !open ? "d-none" : "";
-  const graphsHiddenClass = open ? "d-none" : "";
+  const graphsVisibleClass = !open ? "d-none" : undefined;
+  const graphsHiddenClass = open ? "d-none" : undefined;
```
LZoog, line 111: Empty React classNames should be set to `undefined` so `className=""` isn't rendered on the element.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -73,12 +74,17 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
         // For testing - users will be redirected if the analysis is unavailable
         // before reaching this return, but tests reach this return and
         // analysis.overall is expected to be an object (EXP-800)
-        if (analysisUnavailable(analysis)) return;
+        if (!analysis || analysisUnavailable(analysis)) return;
```
LZoog, line 77: Bringing a comment over from the original PR:

I took `!analysis` out of the `analysisUnavailable` check and placed it here to silence Typescript down the entire page (I was able to remove a lot of `?`s and `!` where we know `analysis` can't be undefined due to this check). `analysisAvailable` checks if it's undefined by way of `analysis?.whatever` chaining so it shouldn't change anything functionally.

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -47,25 +47,14 @@ export const getTableDisplayType = (
   return displayType;
 };
 
+// Ensures the control branch is first
 export const getSortedBranches = (analysis: AnalysisData) => {
```
LZoog, line 51: This function is likely going to be changed when we fix #6129, but I changed it here because it seems like my change from `analysis?.overall` to `analysis.overall` caused a line coverage complaint on what was previously L68. 

I don't feel we need a test for that since we should never hit what was previously on L68/what would be returned from what is now `|| {}` because of our `analysisAvailable` check, which checks for truthiness of `analysis.overall` or `analysis.weekly` and redirects the user if one of these isn't available. We could also add this utils file to the ignore list like we did the experimenter utils file, but seems like it's been okay up until this point leaving it in the coverage.

app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
jodyheavener: Curious, can the `classnames` utility be used here? Like `const graphsHiddenClass = classNames(open && "d-none");`

LZoog: Probably, I saw this in a drive-by sort of way editing another part of the file and didn't necessarily want to tweak further but hey might as well since I already changed the line and you called it out here. 😎

jodyheavener (commented):
Nice! I tried to squint real hard at everything, and nothing is popping out at me. With the number of files touched it'll be good to get a second set of eyes.

I think my only real question (and this may not apply here, I'm less familiar with our Visualization components) is: do all the components here need to be pulling from the context? It may be better to pull the data out in main/layout/container components, and still pass props down to individual UI/child components. If that's already happening here then carry on, just wanted to check to see what you think.

LZoog:
> I think my only real question (and this may not apply here, I'm less familiar with our Visualization components) is: do all the components here need to be pulling from the context? It may be better to pull the data out in main/layout/container components, and still pass props down to individual UI/child components. If that's already happening here then carry on, just wanted to check to see what you think.

Yeah, I think that makes perfect sense to ask. Right now, this is what the structure of this page looks like (at least as far as components that use something from `analysis` - filling this out for my benefit too 😄):

PageResults
* TableHighlights
  * is now pulling data from `useContext`, using it to display things, + passes data into `TableVisualizationRow`
* TableResults
  *  is now pulling data from `useContext`, passes data into `TableVisualizationRow`
* TableResultsWeekly
  * is now pulling data from `useContext` for displaying a tooltip - I think that this could be changed to not use Context since the data is converted to a boolean (`hasOverallResults`), however, we'll want to use Context already for #6129 so I think we should just leave it.
  * Also returns `TableWeekly`
* TableMetricCount
  * is now pulling data from `useContext`, passes this data into `TableVisualizationRow` as well as `GraphsWeekly`
  * GraphsWeekly
    * Is currently using Context for `weekly` results, but **I totally see that this could be passed in by `TableMetricCount` which is pulling that same piece of data**

`VisualizationRow` isn't using Context. Seems to me like I can change `GraphsWeekly` to not use it at least, thanks @jodyheavener for calling that out👍

app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
LZoog: I did end up changing this, but it feels a tad awkward living in this PR since I no longer touch this file otherwise. Oh well. 😆

app/experimenter/nimbus-ui/src/components/PageResults/GraphsWeekly/index.tsx
```diff
@@ -104,8 +105,8 @@ const GraphsWeekly = ({
 
   const [open, setOpen] = useState(false);
   const [embedded, setEmbedded] = useState(false);
-  const graphsVisibleClass = !open ? "d-none" : "";
-  const graphsHiddenClass = open ? "d-none" : "";
+  const graphsVisibleClass = classNames(!open && "d-none");
+  const graphsHiddenClass = classNames(open && "d-none");
```
LZoog, line 109: See [here](https://github.com/mozilla/experimenter/pull/6231#discussion_r690536055) for context (pun intended)

jodyheavener (approved):
homer-woohoo.gif

If you're feeling good about this I'm on board, fire away 🚀

emtwo (approved):
This is such a perfect use of `Context`! It makes me so happy! :')

Issue #6229, opened 2021-08-17
[Intermittent] There is a delay in archiving an experiment if several fields were edited but not saved


SoftVision-CarmenFat:
**[Prerequisites]:**
- Have a Nimbus experiment saved in Draft mode.
- Have the drafted experiment page open.

**[Steps to reproduce]:**
1. Go to the “Overview” section from the sidebar.
2. Edit the section by changing the strings in the fields, selecting radio buttons, and adding links.
3. Do **NOT** save the changes.
4. Click the “Archive Experiment” option from the sidebar.
5. Observe the behavior.

**[Expected result]:**
- You are redirected to the "Summary" page and the “Archived” tag is added next to the title almost instantly.
- The options from the “Edit" field (Overview, Branches, Metrics, Audience) are greyed out.

**[Actual result]:**
- You are redirected to the "Summary" page but the “Archived” tag is added next to the title after ~30 seconds.
- The options from the “Edit” field (Overview, Branches, Metrics, Audience) remain clickable during those ~30 seconds.

**[Notes]:**
- If one of the “Edit” options from the sidebar is clicked during those 30 seconds of waiting, the clicked page is displayed instead of the Summary one, the “Edit” options from the sidebar are greyed out and the “Archived” tag in the title is displayed.
- This issue is intermittently reproducible with _Firefox Release 91_, _Chrome (v. 92.0.4515.159)_, and _Chromium (v. 92.0.4515.131)_ browsers, across platforms (_Windows 10 x64, Ubuntu 20.04,_ and _macOS 11.5.1_).
- Attached is a recording of the issue.

![nimbus example](https://user-images.githubusercontent.com/26000308/129754003-b1d7acf1-09e6-4b48-9ae2-bac316af15ea.gif)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1639)

lmorchard:
The 30 second delay is the polling refetch we do on these pages to pick up state changes.

Seems like this does *not* happen on the Summary page, just on any editing page. And it doesn't seem to matter if the fields have been modified or not. The whole page gets re-built with stale data after a refetch in the useArchive hook, rather than with fresh.

lmorchard:
Looks like we're using a `<LinkNav>` here which seems to cause a router navigation event that tears down the page before the refetched data gets applied. Should be a `<Button>` which doesn't trigger router changes, though the styling will need to be tweaked. But, switching to a button resolves the issue.

[...]

PR #6233, opened 2021-08-17
refactor(nimbus): switch to useChangeOperationMutation for archiving

lmorchard:
Because:

* useArchive hook seems to duplicate useChangeOperationMutation

This commit:

* switches archive action to useChangeOperationMutation hook

* simplifies loading state management in useChangeOperationMutation

* moves changelog message strings for archiving into lib/constants.ts


lmorchard:
Noticed this while poking into #6229: I think we can just use `useChangeOperationMutation` for archiving rather than a separate unique hook

jodyheavener (approved):
Oh heck yeah 👏 🧁

jaredlockhart:
Oh I missed this! +33 −208 👏 👏 👏  that's much better 👍

PR #6248, opened 2021-08-18
Add targeting for various types of profile activity

piatra:
@Mardak what do you think?

Here are some of the targeting attributes we have right now that might be relevant for the classification:

`userMonthlyActivity` - returns an array where each entry represents # of URIs visited from that day. Data goes back 30 days. No activity for a certain day means no entry. Eg. user browsed on 7 distinct days => 7 entries in the array.

`mainPingSubmissions` - returns an array where each entry signals that the user sent a main telemetry ping for a particular day. Data goes back 60 days. Similarly no entry means no submission.

We also have:
`profileAge` - unix timestamp when the profile was created

We have basic math functions available to use in the targeting condition along with the `currentDate` which means we can compute the profile age in days for example. Here is a complete list of [all targeting attributes with descriptions](https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/targeting-attributes.html)

For example the way I've currently expressed regular user using `userMonthlyActivity` is:
`userMonthlyActivity|length >= 14 && userMonthlyActivity|length <= 20 && (currentDate|date - profileAgeCreated|date) / 86400000 >= 28"` - Which reads "User browsed between 14 and 20 days in the past 30 days and the profile is at least 28 days old".
One issue is I don't know how to do the same in terms of `mainPingSubmissions` (because data window is 60 days and we cannot filter out and keep just the past 30).

```
* regular: [14, 20] urls in the past 28 days
* casual: [7, 13] in the past 28 days
* infrequent: [1, 6] in the past 28 days
```


jaredlockhart:
Oh @piatra you have to run `make generate_types&&make code_format` on this commit to turn CI green

piatra:
@jaredlockhart When I do that I get
> KINTO_REVIEW_TIMEOUT not found. Declare it as envvar or define a default value.

irrationalagent:
It looks (to me) like the criteria are different from those we use, which are documented [here](https://docs.telemetry.mozilla.org/concepts/segments.html#activity-segments-informal). 

1. The segments we have been using are defined based on the *number of days* in the past 28 days that a client submitted *a main ping* with  >= 1 URI count. So to qualify as regular, the client must have sent a main ping indicating >= 1 URI on at least 14 distinct days out of the previous 28.
2. There is also a 4th segment, arguably the most important, which is `core_user`: ```client that browsed at least 1 URI in at least 21 of the past 28 days.```

In these definitions, the most important thing is the distinct number of days that the client sent a main ping; the >= 1 URI part of the definition is not as important, though ideally we would keep both if possible. 

If I understand correctly, what you are doing here is looking to see if the client had between (for example) 14 and 20 URIs IN TOTAL across the previous 28 days. So they could have had this all on one day, for instance. This will yield a very different set of clients than the criteria I describe above. (The main point of the segments is track *how many days* the user was active).

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -215,6 +215,86 @@ class NimbusTargetingConfig:
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_INFREQUENT_USER_URIS = NimbusTargetingConfig(
+    name="Infrequent user (uris)",
+    slug="infrequent_user_uris",
+    description="Between 1 and 6 URIs visited in the past 28 days",
```
Mardak, line 221: @irrationalagent This description is indeed misleading as it's actually counting 1-6 days (that have at least 1 URI visited). The data does contain the URI counts from a given day, but here we're looking at `userMonthlyActivity|length` which is the total length of the array where each array item is a pair `[uri count, date]`

Here's an example value of `userMonthlyActivity`:
```js
[
  [
    1038,
    "17-08-2021"
  ],
  [
    830,
    "18-08-2021"
  ],
  [
    140,
    "19-08-2021"
  ],
]
```

You can view your own data by:
1) toggling `browser.newtabpage.activity-stream.asrouter.devtoolsEnabled` from about:config
2) opening a new tab
3) click wrench icon top right corner
4) select Targeting on the left side
5) scroll down (or find in page) "userMonthlyActivity" or "mainPingSubmissions"

irrationalagent: Thanks for walking me through that. 

RE: the 60 day issue for  `mainPingSubmissions` - I wonder about the extent to which we really need to use this, if we already have `userMonthlyActivity`. 

In other words, based on my understanding of things, if the user has non-0 URIs for a given date, then they should have, if everything worked properly, submitted a main ping on that day as well. I get there are edge cases there where a user browses on a day but e.g. kills their machine before they send a ping. And we may want to know the extent to which things *do* operate this way.

That said, `userMonthlyActivity` represents the client's direct understanding of how many URIs were browsed on a day. That's what we're trying to *estimate* on the server side by counting URIs per (server) submission date. So, even though just relying solely on `userMonthlyActivity` will to some extent conflict with our server-side understanding, maybe it would be better because its closer to what we are *intending* to target on. 

And, once we have telemetry that rolls up `userMonthlyActivity` and sends it to us, we won't have to rely on server-based understandings anymore, anyway. 

So to be clear, assuming I'm not misunderstanding things here (big if), I think I'm advocating that we rely only on `userMonthlyActivity` for the segment targeting.

jaredlockhart (approved):
@piatra Oh you also have to add these all to the dict `TARGETING_CONFIGS` and the class `TargetingConfig`. I think you had it there before but it looks like it got taken out?

jaredlockhart (changes requested):
Comment above ^

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -215,6 +215,86 @@ class NimbusTargetingConfig:
...
+    targeting="mainPingSubmissions|length >= 27 && mainPingSubmissions|length <= 40 &&"
+    "(currentDate|date - profileAgeCreated|date) / 86400000 >= 60",
+    desktop_telemetry="",
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_CORE_USER_TELEMETRY = NimbusTargetingConfig(
+    name="Regular user (telemetry)",
+    slug="regular_user_telemetry",
+    description="Regular user determined by telemetry",
```
pdehaan, line 291: Should these be "Regular" or "Core" user? (Looks like a copy-pasta from L279-281 above)

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -215,6 +215,86 @@ class NimbusTargetingConfig:
...
+    description="Between 14 and 20 URIs visited in the past 28 days",
+    targeting="userMonthlyActivity|length >= 14 && userMonthlyActivity|length <= 20 && "
+    "(currentDate|date - profileAgeCreated|date) / 86400000 >= 28",
+    desktop_telemetry="",
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_CORE_USER_URIS = NimbusTargetingConfig(
+    name="Regular user (uris)",
+    slug="regular_user_uris",
```
pdehaan, line 250: Should these say "Regular users" or "Core users"?

Mardak:
> One issue is I don't know how to do the same in terms of `mainPingSubmissions` (because data window is 60 days and we cannot filter out and keep just the past 30).

@jaredlockhart Is it okay to implement a 1-time-use targeting as merge-to-deploy seems to be relatively quick? With JEXL targeting, we can do something like `mainPingSubmissions[.timestampCreated > 1627073164266]|length` to reduce a 60-length array down to entries that newer than July 23rd, and this is only useful for clients that we enroll on August 20th if we want the past 28 days. (This is working around 2 issues: 1) JEXL doesn't allow referencing other variables like `currentDate` within a filter and 2) targeting set on the server is statically defined.)

jaredlockhart:
@Mardak Sure if you'd like, or we could build a tiny bit of infrastructure around delaying setting that date until the experiment is deployed, either works.

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -215,6 +215,86 @@ class NimbusTargetingConfig:
...
+    "(currentDate|date - profileAgeCreated|date) / 86400000 >= 28",
+    desktop_telemetry="",
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_INFREQUENT_USER_TELEMETRY = NimbusTargetingConfig(
+    name="Infrequent user (telemetry)",
+    slug="infrequent_user_telemetry",
+    description="Infrequent users determined by telemetry",
+    targeting="mainPingSubmissions|length >= 1 && mainPingSubmissions|length <= 12 && "
```
Mardak, line 262: Sounds like we can get rid of this second set of `mainPingSubmissions` related targeting as `userMonthlyActivity` is actually based on what was the intended measurement (at least 1 URI visited) as well as closer to expected size (target 28 days).

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -215,6 +215,86 @@ class NimbusTargetingConfig:
...
+    "(currentDate|date - profileAgeCreated|date) / 86400000 >= 28",
+    desktop_telemetry="",
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_CORE_USER_URIS = NimbusTargetingConfig(
+    name="Regular user (uris)",
+    slug="regular_user_uris",
+    description="More than 20 URIs visited in the past 28 days",
+    targeting="userMonthlyActivity|length >= 20 && "
```
Mardak, line 252: This should be `>= 21` to not overlap with "Regular"

And looks like I misspoke about this being 30 days. A quick check on my main profile `userMonthlyActivity|length` evaluates `Status: ✅, Result: 31` which I suppose makes sense as we include "-30 days" before today, so inclusive of today is 31 days. Let's have a followup on Firefox side to make this 28 days total to be a multiple of a whole week, which avoids some bias if the extra non-week-multiple days happens to fall on a weekend vs weekday.

But these targeting in general should be good as they will work across Fx91 and future versions with the client fix.

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -215,6 +215,86 @@ class NimbusTargetingConfig:
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_INFREQUENT_USER_URIS = NimbusTargetingConfig(
+    name="Infrequent user (uris)",
+    slug="infrequent_user_uris",
+    description="Between 1 and 6 URIs visited in the past 28 days",
+    targeting="userMonthlyActivity|length >= 1 && userMonthlyActivity|length <= 6 && "
```
Mardak, line 222: A quick test of clearing all browsing history does result in `userMonthlyActivity|length` evaluating to 0. Even after browsing a page, at least via asrouter devtools, it evaluates 0 until restarting Firefox. I guess in general, this suggests this data can be stale for 6 hours (frecency update interval). So someone starting firefox on a new day and browsing and closes before 6 hours will likely report 1 fewer day because it likely excludes today's activity?

piatra: We can look at it another way: when you start firefox on a new day what is your exact classification at that moment (which is also the moment when enrollments are evaluated)? Handling the caching for this targeting will probably not help much because a similar 6 hour refresh window applies to evaluating recipes (`app.normandy.run_interval_seconds`).

piatra:
> @piatra Oh you also have to add these all to the dict `TARGETING_CONFIGS` and the class `TargetingConfig`. I think you had it there before but it looks like it got taken out?

I might be imagining this but I think it got removed after I ran `make generate_types&&make code_format`

app/experimenter/experiments/constants/nimbus.py
piatra: https://bugzilla.mozilla.org/show_bug.cgi?id=1726739

piatra:
@jaredlockhart good to merge?

jaredlockhart (approved):
Yep looks good to me 👍

PR #6251, opened 2021-08-18
Add end experiment approve and reject e2e tests.

jrbenny35:


app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+        if default_data.public_name in item.text:
+            item.click()
+            break
+    summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
+    summary_page.end_experiment()
+    summary_page.approve()
+    perform_kinto_action(selenium, base_url, "approve")
+    # refresh page until kinto updates
+    for _ in range(45):
+        selenium.get(f"{base_url}/{request.node.name}")
```
pdehaan, line 187: Should this be `default_data.public_name` (per L145 alias above)?
Which, I guess works as long as the `public_name` always matches the slugified version.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+        if default_data.public_name in item.text:
+            item.click()
+            break
+    summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
+    summary_page.end_experiment()
+    summary_page.approve()
+    perform_kinto_action(selenium, base_url, "reject")
+    # refresh page until kinto updates
+    for _ in range(45):
+        selenium.get(f"{base_url}/{request.node.name}")
```
pdehaan, line 241: Same note as above, we aliased this to `default_data.public_name` above on L199. Not sure which we want to use here.

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+        summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
+        try:
+            summary_page.rejected_text
```
pdehaan, line 244: Q: Should this be an `assert` of some sort?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+        except NoSuchElementException:
+            time.sleep(2)
+            selenium.refresh()
+        else:
+            break
```
pdehaan, line 249: Interestingly, these two new tests look pretty similar, with a couple small diffs:

a_txt_↔_b_txt_—_tmp

Not sure if they're worth trying to DRY out, or if we leave that for a future over-optimization.

app/tests/integration/nimbus/conftest.py
```diff
@@ -124,9 +125,12 @@ def _create_experiment(selenium, home_page, data):
...
         selenium.get("http://kinto:8888/v1/admin")
-        kinto_login = Login(selenium, base_url).wait_for_page_to_load()
-        kinto_login.kinto_auth.click()
-        kinto_dashbard = kinto_login.login()
+        try:
+            kinto_login = Login(selenium, base_url).wait_for_page_to_load()
+            kinto_login.kinto_auth.click()
+            kinto_dashbard = kinto_login.login()
+        except NoSuchElementException:
+            kinto_dashbard = Dashboard(selenium, base_url).wait_for_page_to_load()
```
pdehaan, line 133: What's this doing?
Handling a case where we're already current logged into the dashboard?

app/experimenter/nimbus-ui/src/components/Summary/EndExperiment/index.tsx
```diff
@@ -53,6 +54,7 @@ const EndExperiment = ({
         <Button
           variant="primary"
           onClick={toggleShowEndConfirmation}
+          className="EndExperimentStartButton"
```
pdehaan, line 57: (same for L35 above) I think our current convention is kebab-lower-case for class names.
For example, `className="end-experiment-start-button"` and `className="end-experiment-confirm-button"`. Unless you want to try finding a way to reuse the `data-testid` attribute (possibly like `[data-testid='end-experiment-start']`).

pdehaan (approved):
Mostly just silly questions&hellip;

app/tests/integration/nimbus/test_e2e_create_experiment.py
pdehaan: Ha, this was something I learnt the other day via Twitter. VSCode as a diff tool: <kbd>code --diff a.txt b.txt</kbd>

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+    review.launch_without_preview.click()
+    review.request_review.click_launch_checkboxes()
+    review.request_review.request_launch_button.click()
+    review.approve()
+
+    perform_kinto_action(selenium, base_url, "approve")
+
+    selenium.get(base_url)
+    # refresh until the experiment shows up
+    for attempt in range(45):
```
jaredlockhart, line 164: We bumped the number of retries up to 60 and that seemed to help with intermittency so we should probably set these to 60 too.  Or maybe pull it out into a constant at the top?

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+    current_experiments = None
+    try:
+        current_experiments = len(home.tables[0].experiments)
+    except TimeoutException:
+        current_experiments = 0
+    review = create_experiment(selenium, home, default_data)
+    review.launch_without_preview.click()
+    review.request_review.click_launch_checkboxes()
+    review.request_review.request_launch_button.click()
+    review.approve()
```
jaredlockhart, line 158: Wow this is so nice, great job on the refactor @lennartschoch and @jrbenny35 👏 👏 👏

app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -137,3 +137,113 @@ def test_create_new_experiment_remote_settings_timeout(
...
+            break
+    summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
+    summary_page.end_experiment()
+    summary_page.approve()
+    perform_kinto_action(selenium, base_url, "approve")
+    # refresh page until kinto updates
+    for _ in range(45):
+        selenium.get(f"{base_url}/{request.node.name}")
+        summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
+        if "Complete" in summary_page.experiment_status:
```
jaredlockhart, line 189: Could add a method to the `SummaryPage` class to do this complete status check since we'll probably reuse it a bunch?

jaredlockhart (approved):
This is looking really great, I'm so glad more of these are coming in and they're so concise!  Couple ideas for a bit of cleanup, but we should probably bump those retries to 60.  Thanks @jrbenny35 !

jaredlockhart:
Look at it go!!  🎉 🎉 🎉 🎉 

```
nimbus/test_e2e_create_experiment.py::test_create_new_experiment PASSED  [ 16%]
nimbus/test_e2e_create_experiment.py::test_create_new_experiment_remote_settings PASSED [ 33%]
nimbus/test_e2e_create_experiment.py::test_create_new_experiment_remote_settings_reject PASSED [ 50%]
nimbus/test_e2e_create_experiment.py::test_create_new_experiment_remote_settings_timeout PASSED [ 66%]
nimbus/test_e2e_create_experiment.py::test_end_experiment_and_approve_end PASSED [ 83%]
nimbus/test_e2e_create_experiment.py::test_end_experiment_and_reject_end PASSED [100%]

======================== 6 passed in 191.43s (0:03:11) =========================
___________________________________ summary ____________________________________
  integration-test-nimbus: commands succeeded
  congratulations :)
```

app/experimenter/nimbus-ui/src/components/Summary/EndExperiment/index.tsx
jrbenny35: Good point, I'll change this.

app/tests/integration/nimbus/conftest.py
jrbenny35: Exactly that!!

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: lets do a constant _fixture_ 😉

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: this uses the test name as the name for the experiment itself. Just a simpler way, forgot to mention it when we did the refactor.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: I like it

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Yes it should

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Yeah, they are very similar but different enough. No clean way to parameterize them

pdehaan: Yeah, since it only looked like 4-6 lines were different I didnt know if it was just adding one more param for `kinto_action` which is either "approve" or "reject" (which gets passed to `perform_kinto_action()`) and then you'd probably need another `if` statement at the bottom to say:

```python
if kinto_action == "approve":
  if "Complete" in summary_page.experiment_status:
    break;
  else:
    time.sleep(2)
    selenium.refresh()
elif kinto_action == "reject":
  try:
    assert(summary_page.rejected_text)
  except NoSuchElementException:
    time.sleep(2)
    selenium.refresh()
  else:
    break;
else:
  # Unknown kinto_action. Blow up or something.
```

But maybe thats a terrible idea. Since it works, maybe we don't tempt fate by trying to be too clever with one method to rule them all.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: We should use the timeout from the .env file. Let me make that an enhancement.

jaredlockhart: If you're talking about the KINTO_TIMEOUT we'd want double that here.  But we should make this 60 before we land it and then do it as a followup?

pdehaan: Wait, I'm confused. Do we mean this timeout?

https://github.com/mozilla/experimenter/blob/d723f2978f2dd420a661cdd8642925ec62535125/.env.integration-tests#L29

If I'm reading the code correctly, it feels a bit odd to have it serve as both a maximum timeout value (in seconds) and a maximum loop counter value.

jaredlockhart: Yeah that's why I said double it, because it won't happen in exactly that time, it'll happen in some multiple of the kinto sync time, so doubling it to gives the test enough time for the kinto sync to happen and the frontend to pick it up.

app/tests/integration/nimbus/test_e2e_create_experiment.py
jrbenny35: Tried this and there is no clean way to do it without making the page model something more than it is.

app/tests/integration/nimbus/pages/summary.py
```diff
@@ -35,8 +40,28 @@ def request_review(self):
         return self.RequestReview(self)
 
     def approve(self):
-        el = self.find_element(*self._approve_request_button_locator)
+        for _ in range(45):
```
pdehaan, line 43: @jrbenny35 Should this be `range(timeout_length)`?

jrbenny35: doh, missed one. I'll fix tomorrow

PR #6256, opened 2021-08-18
fixes #6054 fix(nimbus): Make TootipWithMarkdown tips visible when they're in pre/code tags

### LINKED ISSUE #6054: Visualization tooltips are empty
Just noticed this [here](https://experimenter.services.mozilla.com/nimbus/default-browser-adoption-experiment-release/results)

Screen Shot 2021-07-23 at 2 10 09 PM

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1544)

LZoog:
fixes #6054 

Because:
* Some tooltips appeared to not be displaying

This commit:
* Sets text in tooltips rendering as code to white

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/TooltipWithMarkdown/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/TooltipWithMarkdown/index.stories.tsx
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import TooltipWithMarkdown from ".";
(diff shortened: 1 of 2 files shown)


jodyheavener (approved):
yeah!

pdehaan:
Verified on prod:

Default_Browser_Adoption_Experiment_-_Release_–_Analysis___Experimenter

... But I think I will open @jodyheavener's https://github.com/mozilla/experimenter/issues/6054#issuecomment-895181635 as a new issue so we can possibly investigate why it's rendered in a `<pre><code>` block to begin with. Having to horizontally scroll a multiline tooltip isn't ideal, so we might just need to poke into some plumbing and see if we need to try overriding some nested markdown-it config somewhere.
If my random theory is correct, we might just need to disable Markdown's weird rule of converting things that are indented by 4+ spaces into a code block, via `markdownIt(options).disable('code');`. But I'm struggling to find this exact string/tooltip to see how it could be indented in a way to trigger that.

PR #6257, opened 2021-08-18
feat(nimbus): Adds first run targeting for mobile

tarikeshaq:
Fixes https://mozilla-hub.atlassian.net/browse/SDK-347.

Adds first run targeting for mobile.

A few notes
- A note about the `JEXL` here, we are opting to allow consumers of the `SDK` to define targeting parameters without blocking on SDK changes (for more context read: https://github.com/mozilla/experimenter-docs/pull/114 and https://mozilla-hub.atlassian.net/browse/SDK-345)). But for now, that's limited to strings, which is why the `isFirstRun` is a string.
- The client changes are not yet merged:
  - ref Fenix: https://github.com/mozilla-mobile/fenix/pull/20642
  - ref iOS: https://github.com/mozilla-mobile/firefox-ios/pull/8952
  
  But we would like to run some local tests with the new targeting
- The way I approached this is by searching for all occurrences of `TARGETING_NO_ENTRPRISE` and added `TARGETING_MOBILE_FIRST_RUN` (making changes wherever appropriate)
- Tested this by running it locally, and seeing it pop up in the UI, then the correct targeting expression was displayed on the overview page when I chose the option. Made sure to also run `make check` and it passed. Here's what the UI looked like

Screen Shot 2021-08-18 at 11 59 26 AM
Screen Shot 2021-08-18 at 11 57 43 AM


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -175,6 +175,17 @@ class NimbusTargetingConfig:
...
 )
 
+TARGETING_MOBILE_FIRST_RUN = NimbusTargetingConfig(
+    name="First start-up users",
+    slug="first_run_mobile",
+    description=(
+        "First start-up users (i.e first run after installing the app) on mobile"
+    ),
+    targeting=("isFirstRun == 'true'"),
+    desktop_telemetry="",
```
tarikeshaq, line 185: I'm not 100% sure of this, since this is the first special targeting parameter on mobile...

jaredlockhart: Yeah you should be able to leave that blank.

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -175,6 +175,17 @@ class NimbusTargetingConfig:
...
+    application_choice_names=(Application.FENIX.name, Application.IOS.name),
```
jaredlockhart, line 186: Not Klar or Focus?

jaredlockhart (approved):
Yep LGTM 👍 Feel free to merge at your convenience 🎉

app/experimenter/experiments/constants/nimbus.py
tarikeshaq: Nope not yet, they don't consume the SDK (as far as I can tell, but either way the work we are doing is targeting iOS and Fenix first 😄)

Issue #6120, opened 2021-08-04
Results relative branch comparison weekly table - bug in specific prod experiment


data-sync-user:
I poked around in production after #6010 was merged and things looked good to me, but in looking at #6071 / EXP-1404, I poked around even more, and there's definitely a bug in at least this experiment if you click to expand the weekly data tables:

https://experimenter.services.mozilla.com/nimbus/multi-stage-onboarding-ask-and-show-to-pin/results#results_summary

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-1581)

LZoog:
After some digging, it appeared like this experiment somehow had the control branch incorrectly set since the control branch viz data from the API endpoint contained relative uplift branch comparison data which didn't make sense; relative branch comparison data on the control branch shouldn't exist since the comparison is relative to the control branch.

After conferring with Anna S./Rob H., we discovered that while in the experiment data the control/reference branch is set to `control-no-show`, this was overridden to become `msw_control` [in the Jetstream config](https://github.com/mozilla/jetstream-config/blob/59ebb94e672c7a62d7d89b39a9c423241348863d/multi-stage-onboarding-ask-and-show-to-pin.toml#L11). The viz API endpoint returns data with an `is_control` boolean which references what the experiment data is set to, but doesn't reflect the override any overrides by the Jetstream config, which apparently [happens often](https://github.com/mozilla/jetstream-config/search?q=reference_branch).

Since I wasn't sure what the problem was at first, I copied the entire server response from the viz API endpoint and the Experiment data and stuck it in Storybook. I never planned on opening a PR for this since the viz data is 43k lines long, but I did [push it up](https://github.com/mozilla/experimenter/compare/results-page-storybook-xl-data) (which I think I'll leave upstream in case I/someone else ever want to test a large data set on this page) to share Storybook stories. [Here's](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/46d8ef0d112f816a110dbff3111d01aa3afb1f5a/nimbus-ui/index.html?path=/story/pages-results--with-xl-data-set) what it looks like in prod today, and [here's](https://storage.googleapis.com/mozilla-storybooks-experimenter/commits/8802a69f9794a4e014a1323edc02932b31fa08ab/nimbus-ui/index.html?path=/story/pages-results--with-xl-data-set) what it looks like if the correct control branch is referenced.

[...]

PR #6264, opened 2021-08-19
fixes #6129 feat(nimbus): Get control branch from metadata config, display in alert

### LINKED ISSUE #6129: Get control branch for visualization from experiment metadata

LZoog:
fixes #6129, and will fix the bug described in #6120

Because:
* The control/reference branch, along with other properties that will be accounted for in a separate PR, can be overridden in the Jetstream config for that experiment which affects only the analysis and Results page

This commit:
* Adds a getControlBranchName fn to return the reference branch from the Jetstream metadata if it exists, or returns the control branch from analysis data
* Adds the new external_config property to the metadata for tests/Storybook and fills in the reference_branch property
* Displays an alert for users on the Results page if any overrides exist

---

Note: I'd like to test this locally with the experiment noted in #6120. Will do before merging (if you see this comment I haven't done that yet - ~~looks like we need the metadata reuploaded, will ping relevant folks~~ pinged Anna S. and she reuploaded it). Our "viewing analysis data locally" docs for any prod experiment are out of date, I may put in another PR up for doc updates there...

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/ExternalConfigAlert/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/ExternalConfigAlert/index.test.tsx
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import ExternalConfigAlert from ".";
+import { MOCK_METADATA_EXTERNAL_CONFIG } from "../../../lib/visualization/mocks";
+
+describe("ExternalConfigAlert", () => {
+  it("renders as expected", () => {
+    render(
+      <ExternalConfigAlert
+        externalConfig={{
+          ...MOCK_METADATA_EXTERNAL_CONFIG,
+        }}
+      />,
+    );
+    expect(
+      screen.getByTestId("external-config-reference-branch"),
+    ).toHaveTextContent("treatment");
+
+    expect(screen.getByTestId("external-config-url")).toHaveProperty(
+      "href",
+      "https://github.com/mozilla/jetstream-config/",
+    );
+  });
+});
--- a/app/experimenter/nimbus-ui/src/components/PageResults/ExternalConfigAlert/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/ExternalConfigAlert/index.tsx
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from "react";
+import { Alert } from "react-bootstrap";
+import { MetadataExternalConfig } from "../../../lib/visualization/types";
+import LinkExternal from "../../LinkExternal";
+
+export const ExternalConfigAlert = ({
+  externalConfig,
+}: {
+  externalConfig: MetadataExternalConfig;
+}) => (
+  <Alert variant="warning" data-testid="external-config-alert">
+    <Alert.Heading>Analysis has manual overrides in Jetstream</Alert.Heading>
+    <p>
+      The results shown on this page are from an analysis ran with at least one
+      experiment override that affects only the <em>analysis</em>, meaning{" "}
+      <strong>
+        experiment data on the Summary page will not reflect this.
+      </strong>
+    </p>
+    <ul className="pl-0">
+      Overrides (
+      <LinkExternal href={externalConfig.url} data-testid="external-config-url">
+        click here
+      </LinkExternal>{" "}
+      to view the config file):
+      {externalConfig.reference_branch && (
+        <li className="ml-3" data-testid="external-config-reference-branch">
+          Baseline branch → <strong>{externalConfig.reference_branch}</strong>
+        </li>
+      )}
+    </ul>
+    <p>
+      If you have questions about this, please ask data science in{" "}
+      <LinkExternal href="https://mozilla.slack.com/archives/C0149JH7C1M">
+        #cirrus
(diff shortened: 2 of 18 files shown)


LZoog:
Ah oops, I think there's one thing I missed that's causing failures, will fix that real quick and reopen as PR when CI is green.

app/experimenter/nimbus-ui/src/components/PageResults/ExternalConfigAlert/index.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import ExternalConfigAlert from ".";
+import { MOCK_METADATA_EXTERNAL_CONFIG } from "../../../lib/visualization/mocks";
+
+describe("ExternalConfigAlert", () => {
+  it("renders as expected", () => {
+    render(
+      <ExternalConfigAlert
+        externalConfig={{
+          ...MOCK_METADATA_EXTERNAL_CONFIG,
+          reference_branch: "foo",
```
LZoog, line 16: I only changed this because I wanted to check for text context other than "control" in case we change "Baseline branch" text to use the word "control" later.

app/experimenter/nimbus-ui/src/components/PageResults/TableVisualizationRow/index.stories.tsx
```diff
@@ -29,6 +29,7 @@ storiesOf("pages/Results/TableVisualizationRow", module)
       metricKey="retained"
       displayType={DISPLAY_TYPE.POPULATION}
       group={GROUP.OTHER}
+      isControlBranch
```
LZoog, line 32: I could have given this a default variable to make Storybook/maybe some future tests more convenient, but `isControlBranch` needs to be explicitely passed in everywhere we use it in the tables, so I'd rather keep it as required without a fallback.

app/experimenter/nimbus-ui/src/components/PageResults/ExternalConfigAlert/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        experiment data on the Summary page will not reflect this.
+      </strong>
+    </p>
+    <ul className="pl-0">
+      Overrides (
+      <LinkExternal href={externalConfig.url} data-testid="external-config-url">
+        click here
+      </LinkExternal>{" "}
+      to view the config file):
+      {externalConfig.reference_branch && (
```
lmorchard, line 30: Will there be any other external config properties mentioned here besides `reference_branch`? The `MetadataExternalConfig` type also includes these:
```
  end_date: string | null;
  enrollment_period: string | null;
  skip: boolean;
  start_date: string | null;
```

LZoog: This PR is just for `reference_branch`, I noted the separate PR bit in the PR description / issue number in [a comment here](https://github.com/mozilla/experimenter/pull/6264/files#diff-1255a403cd47ad0be32d2918921c64471bdcf96d6808ffee9487be30fa0e815fR50) 😁

LZoog:
Pushed one tiny update, changed `reference_branch: "control"` in the mock config to `reference_branch: "treatment"` and added 1 other test since it was overriding but wasn't actually pointing to a different branch than what was already set as the reference branch.

app/experimenter/nimbus-ui/src/lib/visualization/types.ts
```diff
@@ -19,6 +19,16 @@ export type AnalysisDataWeekly = Exclude<AnalysisData["weekly"], null>;
...
+}
+
+export interface MetadataExternalConfig {
+  end_date: string | null;
+  enrollment_period: string | null;
+  reference_branch: string | null;
+  skip: boolean;
+  start_date: string | null;
+  url: string;
 }
```
lmorchard, line 32: Maybe just a note to myself, but: I was a little confused looking at [this change to the Jetstream schema](https://github.com/mozilla/jetstream/pull/841) and [this comment](https://github.com/mozilla/experimenter/issues/6129#issuecomment-893778432) in issue #6129.

But, things have been updated since then. [Here's the Jetstream change that corresponds to this type](https://github.com/mozilla/jetstream/commit/530469c5facf9f1573a3882d831e1dd73c6cb3eb).

And here's what the data in a prod experiment looks like from the analysis endpoint

Capture

lmorchard (approved):
LGTM!

app/experimenter/nimbus-ui/src/lib/visualization/types.ts
LZoog: Ack, yeah, I can see why that's confusing. We originally thought this was a problem with only the `reference_branch` and it was exposed in the metadata as only `reference_branch` (jetstream config schema version 3). Then, a few days later, #6200 was filed, which prompted changing this to that `external_config` object (schema version 4).

The metadata for that experiment was re-uploaded manually end of day yesterday by Anna, as well as a couple of others that had overrides, and this morning I went to double check that `external_config` was coming back on `metadata` on the experiment in that bug and we discovered we needed to "force jetstream data fetch" for that experiment since old completed experiments apparently keep that cache around for 3 days or so (I'm making note of this in a doc update).

LZoog:
I mentioned in my original PR comment that I wanted to check this out locally first for that experiment mentioned in the bug before merging - I'm having trouble doing that as well as even pulling up results for `testing-nimbus` locally per our docs (looks like `testing-nimbus` doesn't exist in production and I thhink it has to?). I'll merge this Monday morning after more investigating (AKA pinging someone that'll have an answer 😄) and will have a PR up to update the docs too.

PR #6266, opened 2021-08-20
e2e end experiment cleanup

jrbenny35:


app/tests/integration/nimbus/test_e2e_create_experiment.py
```diff
@@ -123,7 +133,7 @@ def test_create_new_experiment_remote_settings_timeout(
     review.launch_without_preview.click()
     review.request_review.click_launch_checkboxes()
     review.request_review.request_launch_button.click()
-    review.approve()
+    review.approve(timeout_length)
 
     for attempt in range(60):
```
pdehaan, line 138: Q: Should this use `range(timeout_length)` too now (to be consistent w/ L202 and L262 below, although it also means we'd drop the value from 60 to 45).

If so, we might want to audit the other usages too and see if we use the same value globally in this file:

```
L46: `for attempt in range(45):`
L93: `for attempt in range(45):`
L113: `for attempt in range(30):`
L138: `for attempt in range(60):`
```

pdehaan:
Unrelated, but my eyes saw it:

https://github.com/mozilla/experimenter/blob/320da4ecc65c3472e9758dbee64b55271eaad1ca/app/tests/integration/nimbus/test_e2e_create_experiment.py#L139-L147

Do we need L141? I was going to ask if that should have an assert also, but then I see we have the `assert` on L146 in the `else` statement. Wasn't sure if that makes L141 unneeded, or if L146 should replace L141. Happy to open a new issue and discuss there as well so I don't muddy this thread.

jrbenny35:
> 
> 
> Unrelated, but my eyes saw it:
> 
> https://github.com/mozilla/experimenter/blob/320da4ecc65c3472e9758dbee64b55271eaad1ca/app/tests/integration/nimbus/test_e2e_create_experiment.py#L139-L147
> 
> Do we need L141? I was going to ask if that should have an assert also, but then I see we have the `assert` on L146 in the `else` statement. Wasn't sure if that makes L141 unneeded, or if L146 should replace L141. Happy to open a new issue and discuss there as well so I don't muddy this thread.

Interesting, I agree it should be within the ```try``` block

pdehaan:
Apart from the unrelated note above about the `assert` this looks good to me.
I disabled auto merging because I don't 100% know how that works and if it will merge automatically before you do the assert change or whatever you might be thinking.

pdehaan (approved):

jrbenny35:
> 
> 
> Apart from the unrelated note above about the `assert` this looks good to me.
> I disabled auto merging because I don't 100% know how that works and if it will merge automatically before you do the assert change or whatever you might be thinking.

I'll save that for another PR

pdehaan:
> I'll save that for another PR

Cool. I might do a quick `git grep` on asserts and see what else my lazy eyes might have missed. And if we have error messages for asserts, etc.

Feel free to merge when green.

jrbenny35:
> 
> 
> > I'll save that for another PR
> 
> Cool. I might do a quick `git grep` on asserts and see what else my lazy eyes might have missed. And if we have error messages for asserts, etc.
> 
> Feel free to merge when green.

We can file an issue to add more info to the regular ```assert``` statements.

PR #6274, opened 2021-08-23
fix #6267 chore(docs): Update visualization data nimbus-ui docs

### LINKED ISSUE #6267: Update "Access Visualization Data Locally" docs
The third bullet point [here](https://github.com/mozilla/experimenter/tree/main/app/experimenter/nimbus-ui#access-visualization-data-locally) could use an adjustment to reflect that you can change the experiment name to something that exists in production. I haven't done this outside of that `testing-nimbus` slug per the docs so I'm about to run through this to check out an experiment in prod locally and will document anything else that might be helpful.

Also while we're at it, let's document that analysis data is cached for 6 hours, but if an experiment ended a while ago, it's cached for a few days.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1662)

LZoog:
fixes #6267 

Because:
* The 'testing locally' vis docs were out of date and could use more details

This commit:
* Updates the visualization portion of the nimbus-ui readme for more explicit instructions on testing locally, with other details

### DIFF
--- a/app/experimenter/nimbus-ui/README.md
+++ b/app/experimenter/nimbus-ui/README.md
@@ -236,15 +236,41 @@ While an experiment is in `draft` status all fields are technically allowed to b
 
 No additional action is required; as long as you correctly set the form group's `controlId` and the form error's `name` these review-readiness messages will automatically appear as needed when a user attempts to go into review with incomplete fields.
 
-## Access Visualization Data Locally
+## Results Page and Visualization Data
 
 The Results page renders experiment analysis data and can only be accessed when an experiment is complete, the analysis feature flag is on, and the analysis data returned from the visualization API endpoint is available.
 
-To see the Results page and data locally:
+### Locally Access Visualization Data for Experiments in Prod
+
+At the time of writing, locally created complete experiments don't return data from the visualization endpoint. However, it is possible to either supply a JSON response to essentially mock the endpoint response or to view a locally created experiment that mirrors an experiment in production to test the Results page locally with production visualization data.
+
+#### "Results data" Admin JSON Dump
+
+If you're just looking to test front-end changes, the easiest way to view the Results page with data locally is to:
+
+1. Set `FEATURE_ANALYSIS=true` in your local `.env` file
+2. Find a completed experiment in production that has analysis data ready that you'd like to test your changes against. Either login to `/admin` on production and navigate to that Nimbus Experiment and copy what's in "Results data," or, copy the entire server response from what's returned from the visualization endpoint in production
+3. At `localhost/admin`, add a new Nimbus Experiment with any name/slug and paste your clipboard into the "Results data" for the new experiment
+
+You should now be able to navigate to `localhost/nimbus/your-experiment-slug/results` to test your changes locally.
+
+#### Fetching from GCP
+
+Alternatively to view results locally, if you want to more closely mirror production or if you want to test a Jetstream task:
 
 1. Ensure your [Google Credentials](https://github.com/mozilla/experimenter#google-credentials) are configured
 2. Set `FEATURE_ANALYSIS=true` in your local `.env` file
-3. Create a Nimbus Experiment at `localhost/admin` with experiment slug "testing-nimbus" status "complete," a branch/branch slug named "control", and another branch/branch slug named "treatment". Alternately, change an existing, complete Nimbus experiment with branches in place to reflect the experiment slug `testing-nimbus`.
+3. Find a completed experiment in production that has analysis data ready that you'd like to test your changes against
+4. At `localhost/admin`, add a new Nimbus Experiment with the same experiment slug as the production experiment you'd like to test against. Set the status to "complete," add the matching reference/control branch from the production experiment (name and slug), and save the experiment. Then, select that branch in the "reference branch" drop down and save again.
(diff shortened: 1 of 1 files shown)


app/experimenter/nimbus-ui/README.md
```diff
@@ -236,15 +236,29 @@ While an experiment is in `draft` status all fields are technically allowed to b
...
+
 To see the Results page and data locally:
 
 1. Ensure your [Google Credentials](https://github.com/mozilla/experimenter#google-credentials) are configured
 2. Set `FEATURE_ANALYSIS=true` in your local `.env` file
-3. Create a Nimbus Experiment at `localhost/admin` with experiment slug "testing-nimbus" status "complete," a branch/branch slug named "control", and another branch/branch slug named "treatment". Alternately, change an existing, complete Nimbus experiment with branches in place to reflect the experiment slug `testing-nimbus`.
+3. Find a completed experiment in production that has analysis data ready that you'd like to test your changes against
+4. At `localhost/admin`, add a new Nimbus Experiment with the same experiment slug as the production experiment you'd like to test against. Change the status to "complete," and add every branch (name and slug) to match the production experiment. Also, make sure "reference branch" is set to the reference branch for that production experiment.
```
emtwo, line 252: I think you actually only need to add the reference branch's name and slug and the rest is picked up on its own. This is handy if you want to try out an experiment locally that has a lot of branches so it's less time to have to input all of them.

emtwo, line 252: I think you actually only need to add the reference branch's name and slug and the rest is picked up on its own. This is handy if you want to try out an experiment locally that has a lot of branches so it's less time to have to input all of them.

app/experimenter/nimbus-ui/README.md
```diff
@@ -236,15 +236,29 @@ While an experiment is in `draft` status all fields are technically allowed to b
...
+5. Optionally change `fetch_jetstream_data` in `app/experimenter/settings.py` to `30` so it fetches and stores the data in the local database
```
emtwo, line 253: I think this is not optional (unless you do a force refresh like you describe below). The data won't show up for 8 hours if you don't lower the time it takes to fetch

app/experimenter/nimbus-ui/README.md
```diff
@@ -236,15 +236,29 @@ While an experiment is in `draft` status all fields are technically allowed to b
...
+
+You should now be able to navigate to `localhost/nimbus/your-experiment-slug/results` to test your changes locally.
+
+### Jetstream Metadata and Caching
+
+With the data returned from the visualization endpoint, a `metadata` object is also returned. At the time of writing, it contains a `description` per metric used for tooltips on the Results page as well as an `external_config` object containing some experiment properties than can be overridden for analysis purposes only. Jetstream configs for experiments are located in the [`jetstream-config`](https://github.com/mozilla/jetstream-config/) repository and the metadata schema lives in the [Jetstream repo](https://github.com/mozilla/jetstream/).
+
+Visualization data is cached and refetches every 8 hours for experiments less than 3 days old. After that point, experiment data is not refetched. If a manual refetch needs to occur, in the production `/admin`, go to Nimbus Experiments > Action > "Force jetstream data fetch".
```
emtwo, line 261: it refetches for experiments that are live or whose end date is less than 3 days ago.

emtwo (commented):
thanks so much for making these changes, Lauren!!

app/experimenter/nimbus-ui/README.md
LZoog: Oh, interesting! I'll try this out to double check but that is definitely handy.

app/experimenter/nimbus-ui/README.md
LZoog: Oh oops, that is totally what I meant when I wrote this. 😆

app/experimenter/nimbus-ui/README.md
jaredlockhart: Oh really when I was trying this the other day for an experiment with like 5 branches I did them all, so I don't need to do that?

app/experimenter/nimbus-ui/README.md
jaredlockhart: Oh I see these instructions are for fetching from GCP.  You can also test locally by just pulling the `results_data` from the prod admin and dumping it into the local admin.  Depends on what you're trying to test.  If you're testing the frontend you don't need to worry about the jetstream task, you can just skip this step and dump the processed data right into the db.  If you want to test the jetstream task you gotta do this.  Probably worth describing both?

app/experimenter/nimbus-ui/README.md
LZoog: I just tested this out to confirm and that is correct, not all branches are needed, just the reference branch. 😈

app/experimenter/nimbus-ui/README.md
LZoog: If someone doesn't have access to prod admin, they can probably also just copy the entire response from the endpoint on the experiment they want to look at. I actually haven't ever done it that way by just dumping it via local admin, I've just been following the instructions here so that's definitely seems worth noting, I'll add👍

LZoog: @jaredlockhart turns out if you do it that way, the slug and branches don't even need to match. 😆 PR has been updated.

lmorchard (approved):
LGTM!

PR #6297, opened 2021-08-24
fix #6293 chore(nimbus): don't require generate_types for targeting changes

### LINKED ISSUE #6293: Don't require make generate_types for TargetingConfig changes
To simplify adding new TargetingConfigs, we should remove the necessity to run make generate_types.  We can just change the field to a StringField, the values will still be sent by the config query.

Based on feedback in [https://mozilla-hub.atlassian.net/browse/EXP-1685](https://mozilla-hub.atlassian.net/browse/EXP-1685|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1686)

jaredlockhart:
Because

* We expect many people to be adding/changing targeting configs over time
* It shouldn't be necessary to run make generate_types when landing changes to targeting

This commit

* Removes the enumeration typing from the V5 API for targeting_config_slug


lmorchard (approved):
Hmm, yeah, looks like one of those things graphql makes a bit more complicated than we need?

PR #6301, opened 2021-08-24
fix #5782 feat(nimbus): add parent to nimbus experiment

### LINKED ISSUE #5782: (BE) Add new parent foreign key to experiment model, update query to resolve it
Update [NimbusExperiment](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L58:L58) to add a new optional foreign key linking an experiment back to its parent experiment. This will allow us to link a cloned experiment to its parent.

Update [NimbusExperimentType](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/types.py#L148:L148) so it can resolve the parent experiment field. This will also result in an update to schema.gql. This does not need to be added to the frontend query yet.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1367)

jaredlockhart:
Because

* To track which experiment another experiment was cloned from we need a parent field

This commit

* Adds a parent field to NimbusExperiment


app/experimenter/experiments/changelog_utils/nimbus.py
```diff
@@ -23,15 +21,12 @@ class Meta:
...
     reference_branch = NimbusBranchChangeLogSerializer()
     branches = NimbusBranchChangeLogSerializer(many=True)
     feature_config = NimbusFeatureConfigChangeLogSerializer()
-    owner = serializers.SlugRelatedField(
-        queryset=get_user_model().objects.all(), slug_field="email"
-    )
-    projects = serializers.SlugRelatedField(
-        many=True, queryset=Project.objects.all(), slug_field="slug"
-    )
+    owner = serializers.SlugRelatedField(read_only=True, slug_field="email")
```
lmorchard, line 28: Had a little head-scratch at this change, but makes sense after I read the DRF docs about read only fields

lmorchard (approved):
LGTM! :+1:

app/experimenter/experiments/changelog_utils/nimbus.py
jaredlockhart: Oh yeah exactly I saw that in the docs and realized we didn't need to pass qs at all 👍

PR #6302, opened 2021-08-24
fix #6200 feat(nimbus): Display all analysis external config overrides on Results page

### LINKED ISSUE #6200: Experiment end date can be incorrect if overridden in external config

LZoog:
fixes #6200 

Because:
* Experiment analysis can be ran with overrides on the Jetstream side. As of Jetstream metadata schema 4, these overrides are exposed and we want to convey them to users to eliminate confusion. The reference/control branch was taken care of in a prior commit - this covers the rest of the overrides.

This commit:
* Displays start date, end date, and proposed enrollment overrides in ExternalConfigAlert
* Displays 'analysis was skipped' text in the sidebar if 'skip' is set to true, slightly alters analysis ready text
* Refactors AppLayoutSidebarLaunched stories to new syntax, refactors PageResults stories to new syntax and adds story to better preview the page when the control branch is overridden


app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
```diff
@@ -34,32 +38,37 @@ const Subject = ({
...
-        experiment,
-        refetch: async () => {},
-      }}
-    >
-      <p data-testid="test-child">Hello, world!</p>
-    </AppLayoutSidebarLaunched>
-  </RouterSlugProvider>
-);
+}) => {
+  const mockedAnalysis =
```
LZoog, line 42: This could probably use a bit more refactoring but I didn't want too much time here😬

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
```diff
@@ -79,7 +88,7 @@ describe("AppLayoutSidebarLaunched", () => {
         `${BASE_PATH}/my-special-slug`,
       );
 
-      screen.getByText("Experiment results not yet ready");
+      screen.getByText("Experiment analysis not ready yet");
```
LZoog, line 91: I changed this text to use the word "analysis" since "Results" sits right above this and this just felt nicer since the title on the page is "Analysis"? I can easily revert if preferred.

app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
```diff
@@ -272,8 +272,10 @@ export const AppLayoutSidebarLaunched = ({
                       </LinkExternal>
                       .
                     </>
+                  ) : analysis?.metadata?.external_config?.skip ? (
```
LZoog, line 275: I added this [per this comment](https://github.com/mozilla/experimenter/issues/6200#issuecomment-901398847) in the issue. I believe it'll only affect [this experiment](https://experimenter.services.mozilla.com/nimbus/set-default-as-first-screen-100-roll-out) for now.

app/experimenter/nimbus-ui/src/lib/visualization/mocks.tsx
```diff
@@ -47,12 +47,11 @@ export const MOCK_METADATA = {
  * overrides are present.
  */
 export const MOCK_METADATA_EXTERNAL_CONFIG = {
-  // TODO: account for the rest of these, EXP-1628
-  end_date: null,
-  enrollment_period: null,
+  start_date: "2021-05-26",
+  end_date: "2021-06-06",
+  proposed_enrollment: 9,
```
LZoog, line 52: I actually have no idea where I got `enrollment_period` from my previous commit. 😆 [These](https://github.com/mozilla/experimenter/issues/6200#issuecomment-898685666) are the options we need to account for.

LZoog (commented):
Little self-review here.

BTW, I checked the tables we render on this page and we don't look to be referencing any experiment properties (`startDate`, `endDate` `proposedEnrollment`) in any of the tables so we don't need to point to the metadata as the "source of truth" for these, unlike the control branch, which required a slight refactor in most tables.

app/experimenter/nimbus-ui/src/lib/visualization/mocks.tsx
lmorchard: Hmm, from [when I last looked at the metadata from jetstream](https://github.com/mozilla/experimenter/pull/6264#discussion_r693148062), `enrollment_period` *was* one of the properties included.

PR #6304, opened 2021-08-25
fixes #5781 refactor(nimbus): reorder middleware

### LINKED ISSUE #5781: Security: reorder middleware
Just noticed in the [whitenoise docs](http://whitenoise.evans.io/en/stable/#quickstart-for-django-apps) they suggest putting the [Django Security middleware](https://docs.djangoproject.com/en/3.2/ref/middleware/#module-django.middleware.security) above the whitenoise middleware, so we should probably do that.

So we just need to swap these two:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L82-L83

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1415)

anthonytran7757:
Because

* the whitenoise docs suggests putting the Django Security middleware above the whitenoise middleware

This commit

* reorders the middleware as suggested by whitenoise

### DIFF
--- a/app/experimenter/settings.py
+++ b/app/experimenter/settings.py
@@ -85,8 +85,8 @@ INSTALLED_APPS = [
 
 MIDDLEWARE = [
     "corsheaders.middleware.CorsMiddleware",
-    "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.middleware.security.SecurityMiddleware",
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Awesome thank you @anthonytran7757 ! 🎉

PR #6307, opened 2021-08-25
    fix #5791 feat(nimbus): add clone api

### LINKED ISSUE #5791: (BE) Add mutation to clone an experiment
Create a new [mutation](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/mutations.py) that looks up an experiment and calls its clone method. This mutation should take a new [input](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/inputs.py) type (something like CloneExperimentInput) that allows you to supply an ID int (the experiment to clone) and a name string (optional).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1369)

jaredlockhart:
Because
    
* We want to trigger cloning from the UI
    
This commit
    
* Exposes the cloning method through a V5 API

### DIFF
--- a/app/experimenter/experiments/api/v5/inputs.py
+++ b/app/experimenter/experiments/api/v5/inputs.py
@@ -66,3 +66,8 @@ class ExperimentInput(graphene.InputObjectType):
     risk_brand = graphene.Boolean()
     countries = graphene.List(graphene.Int)
     locales = graphene.List(graphene.Int)
+
+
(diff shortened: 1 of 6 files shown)


lmorchard (approved):
Tests pass, makes sense, LGTM! :+1:

PR #6333, opened 2021-08-31
fix #6305 chore(nimbus): add archiving integration test

### LINKED ISSUE #6305: Integration test for Experiment Archiving
* test that you can archive a draft experiment and then unarchive it successfully
* not sure if you want to do it w/ a completed experiment also (since that requires doing the RS dance). Probably part of the larger discussion of having a set of precanned experiments added to the CI run so not every experiment needs to run through the full approval flow before it can do certain things.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1691)

### LINKED ISSUE #6405: Clicking on one of the subsections of Results in the navigation sidebar while on Summary opens just Results
**[Prerequisites]:**
- Have an experiment with Results available open on Summary.

**[Steps to reproduce]:**
1. Click on “Results Summary” from the navigation sidebar.
2. Observe the behavior.

**[Expected result]:**
- The Results page opens focused on the “Results Summary” section.

**[Actual result]:**
- The Results page opens without being focused on any section. 
- The user has to click again on “Results Summary” to navigate to that section.

**[Notes]:**
- The issue is applicable for all subsections of Results in the navigation sidebar.
click twice on Results

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1752)

jaredlockhart:
Because

* We should add a simple integration test for each new major feature we add

This commit

* Adds an integration test for archiving
* Splits the integration tests into two groups
  * Tests that run for each application (for remote settings integration)
  * Tests that only run once (for things constrained just to experimenter UI)
* Adds a random identifier to the end of integration test experiments to allow rerunning tests locally
* Cleans up the page load wait locators


app/tests/integration/nimbus/pages/home.py
```diff
@@ -6,15 +6,14 @@
 class HomePage(Page):
     """Nimbus Home page."""
 
+    _page_wait_locator = (By.CSS_SELECTOR, "#PageHome-page")
```
pdehaan, line 9: Interesting. This moved+changed. Why wait for `#pageHome-page` instead of `.directory-table`?

jaredlockhart: 1) Consistency, looking for the named outer container on every page makes it easier to reason about what we're waiting for
2) This is actually a bug in the test setup, `.directory-table` returns an element for each of the 6 tables, and if you just pass that into the wait function, it just picks the first one, which is fine if you land on the page with the Live tab highlighted, but if you set it to any of the other tabs, it picks the first one which is hidden, which fails the wait for check.

.circleci/config.yml
```diff
@@ -68,17 +68,17 @@ jobs:
       docker_layer_caching: true
     resource_class: xlarge
     working_directory: ~/experimenter
-    parallelism: 5
+    parallelism: 6
```
pdehaan, line 71: I assume this is because the `parallel_pytest_args.txt` file now has 6 lines instead of 5?
If so, could/should this be semi-scripted like some `$(wc -l "app/tests/integration/nimbus/parallel_pytest_args.txt")`? Or should we add some comment about the pseudo-magic number?

app/tests/integration/nimbus/test_experimenter.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Create experiment
+    summary_page = create_experiment(selenium, home, default_data)
+
+    # Archive it on the summary page
+    summary_page.archive_button.click()
+
+    # Check it's archived on the summary page
+    time.sleep(2)
+    assert summary_page.archive_label.is_displayed()
```
pdehaan, line 26: Should this retry a couple in a loop at all in case the label isn't available within ~2s?

app/tests/integration/nimbus/test_experimenter.py
```diff
@@ -0,0 +1,62 @@
...
+            break
+
+    # Unarchive it on the summary page
+    summary_page.archive_button.click()
+
+    # Check it's unarchived on the summary page
+    time.sleep(2)
+    try:
+        assert summary_page.archive_label
+    except Exception:
```
pdehaan, line 47: Does this need to catch a more specific exception? (like NoSuchElementException)

app/tests/integration/nimbus/test_experimenter.py
pdehaan: Should this also have some error text if the assertion fails? Something like https://github.com/mozilla/experimenter/blob/19298d682fcbb5e38cc8b49fd0def8d414764b5d/app/tests/integration/nimbus/test_e2e_create_experiment.py#L125

But not sure that it's a pattern we use globally in the test assertions, so not a huge deal.

app/tests/integration/nimbus/pages/home.py
```diff
@@ -6,15 +6,14 @@
...
     _create_new_btn_locator = (By.CSS_SELECTOR, "#create-new-button")
-    _page_wait_locator = (By.CSS_SELECTOR, ".directory-table")
 
-    def __init__(self, selenium, base_url, **kwargs):
-        super(HomePage, self).__init__(selenium, base_url, timeout=30, **kwargs)
+    def __init__(self, selenium, base_url, timeout=30, **kwargs):
+        super(HomePage, self).__init__(selenium, base_url, timeout=timeout, **kwargs)
```
jrbenny35, line 13: We don't need this still, do we?

jrbenny35 (changes requested):

app/tests/integration/nimbus/test_experimenter.py
jaredlockhart: Ah I learned an even better pattern for all of this!

```py
        self.wait.until(
            EC.presence_of_all_elements_located(self._archive_label_locator),
            message="Summary Page: could not find archive label",
        )
```

app/tests/integration/nimbus/test_experimenter.py
jaredlockhart: Changed this to use the wait.until style, much nicer 👍

pdehaan (approved):
:chefs_kiss:

Will leave final 👍 to @jrbenny35.

jrbenny35 (approved):

jaredlockhart:
Now we just pray it passes without hitting the rs intermittent XD

PR #6336, opened 2021-09-01
fix #6312 feat(nimbus): Connect Clone modal to mutation

### LINKED ISSUE #6312: (FE) Connect Clone modal to clone mutation
Once we have both the mutation and the modal landed, we'll have to hook them up, including displaying errors that the API returns.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1696)

lmorchard:
Because:

* we want to enable cloning of experiments

This commit:

* adds the clone experiment GQL mutation in front end

* creates a useCloneDialog hook with CloneDialog display and mutation logic that can be partially controlled and watched from SidebarActions

* adds display of general submit errors to CloneDialog

* updates types & tests as necessary

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/SidebarActions/CloneDialog/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/SidebarActions/CloneDialog/index.test.tsx
@@ -2,6 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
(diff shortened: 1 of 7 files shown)


lmorchard:
Opening as draft until I see CI go green, but I think this will do the trick

jaredlockhart (approved):
Aawwwweesssooommmeee.  This is so slick.  Works perfectly.  I did find a bug in the clone method that I filed, but the UI works perfectly 🎉 Thanks @lmorchard !

PR #6339, opened 2021-09-01
fix #6295 bug(nimbus): Fix intermittency in remote settings integration tests

### LINKED ISSUE #6295: Intermittent integration test failure: test_end_experiment_and_approve_end
Seen here: https://app.circleci.com/pipelines/github/mozilla/experimenter/10892/workflows/45afd741-97f5-4680-af8e-84fa3b8f9e78/jobs/32607

```AssertionError: Approve button was never shown```

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-1688)

### PR DESCRIPTION
**Pull request by jaredlockhart**
Because

* We've seen persistent intermittent failures in the remote settings integration tests
* It turns out the remote settings pages can render in stages so it's not sufficient to just wait for the page wait locator for all the UI elements to be available
* Also we were using the old remote settings review page instead of the new simple one that Experimenter links to

This commit

* Adds a wait locator to every UI element in the Remote Settings pages so interactions shouldn't occur until the element is visible
* Adds helpful messages to each of the waits so if one fails the test should clearly say what it was waiting for and couldn't find
* Use the new simple-review page in Remote Settings instead of the old review page

Note: I found a lot of useful patterns while I was working on this that I think we could apply to other parts of the tests, like adding per element waits with helpful messages to the forms.  I'll do those as a followup.

### CODE REVIEW

File: app/tests/integration/nimbus/conftest.py [hunk 1]
```diff
@@ -54,32 +75,39 @@ def _verify_url(request, base_url):
...
-    params=[
-        (BaseExperimentApplications.DESKTOP, "No Feature Firefox Desktop"),
-        (BaseExperimentApplications.FENIX, "No Feature Fenix"),
-        (BaseExperimentApplications.IOS, "No Feature iOS"),
-        (BaseExperimentApplications.KLAR, "No Feature Klar for Android"),
-        (BaseExperimentApplications.FOCUS, "No Feature Focus for Android"),
-    ],
-    ids=["DESKTOP", "FENIX", "IOS", "KLAR", "FOCUS"],
+    params=list(BaseExperimentApplications),
+    ids=[application.name for application in BaseExperimentApplications],
```
jrbenny35, line 95: I like this but I don't at the same time

app/tests/integration/nimbus/conftest.py
```diff
@@ -13,16 +12,38 @@
...
+        "#/buckets/main-workspace/collections/nimbus-mobile-experiments/simple-review"
+    ),
+    BaseExperimentApplications.FOCUS: (
+        "#/buckets/main-workspace/collections/nimbus-mobile-experiments/simple-review"
+    ),
+}
+
+
+def slugify(input):
+    return input.lower().replace(" ", "-").replace("[", "").replace("]", "")
```
jrbenny35, line 46: love this

app/tests/integration/nimbus/pages/base.py
```diff
@@ -12,13 +13,25 @@ class Base(Page):
...
+                self.wait_for_page_to_load()
+                selenium.find_element(*locator)
+            except NoSuchElementException:
+                selenium.refresh()
+                return False
+            else:
+                return True
+
+        self.wait.until(_wait_for_it, message=message)
+
```
jrbenny35, line 34: 🔥

app/tests/integration/nimbus/remote_settings/pages/dashboard.py
```diff
@@ -1,97 +1,37 @@
-from pypom import Page, Region
-from selenium.common.exceptions import NoSuchElementException
+from nimbus.pages.base import Base
```
jrbenny35, line 1: Why are you importing the nimbus base page here? I think, if anything, we should create a separate kinto base page.

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -1,261 +1,172 @@
...
-    assert "Live" in home.active_tab_text
-    for item in live_experiments.experiments:
-        if default_data.public_name in item.text:
-            item.click()
-            break
-    summary_page = SummaryPage(selenium, base_url).wait_for_page_to_load()
-    assert "live" in summary_page.experiment_status.lower()
+    selenium.get(kinto_review_url)
+    kinto_dashboard = Dashboard(selenium, kinto_review_url).wait_for_page_to_load()
+    kinto_dashboard.approve()
```
jrbenny35, line 33: Did you try using ```.open()```?

jrbenny35 (changes requested):
I like where this is going!

app/tests/integration/nimbus/conftest.py
jaredlockhart: Yeah we could leave it as it is, I did this in a bunch of other places in the rest of the Nimbus project just so the list of applications is written in one place instead of in a bunch places.

app/tests/integration/nimbus/remote_settings/pages/dashboard.py
jaredlockhart: Oh it was to reuse the same base logic with the timeouts and the waits and stuff.  Is there something we need to put in a separate base?

app/tests/integration/nimbus/test_remote_settings.py
jaredlockhart: I did but something else had to change so I left out and was gonna do it as a followup.

app/tests/integration/nimbus/conftest.py
pdehaan: Nice to have this DRY. Should we reuse Django’s slugify function vs rolling a second one w/ potentially slightly different rules?

jaredlockhart: We totally should but unfortunately django's not in this container :( We could lift the source of it and just copy it over, which I thought about, but I wanted to (try XD) to change as little as I could so I just left it with the one we have now.  We can do that as a followup though.

app/tests/integration/nimbus/pages/base.py
jaredlockhart: No the built in wait is very nice and has fixed timeouts, the duration is specified [here](https://github.com/mozilla/experimenter/pull/6339/files#diff-2bb50b01c7c0f75eb7f6408f89b9c14ae62d0d5ecacdb072aacd8f8014595ae1R16) and it'll display the handy message when it fails :D

app/tests/integration/nimbus/conftest.py
jrbenny35: Can we add a comment just to help with readability?

jaredlockhart: Shore 👍

app/tests/integration/nimbus/remote_settings/pages/dashboard.py
jrbenny35: Since remote settings is _technically_ different than nimbus, I would rather have remote settings stuff be agnostic to nimbus.

jaredlockhart: I see your point conceptually, we could have a separate base but then we'd either repeat the same stuff we put in the nimbus base, or inherit from it and there wouldn't be anything specific to remote settings in it.  I generally don't add a class until it's necessary, and if/when there's something specific to remote settings we could add its own base class but I don't think there is anything yet?

jaredlockhart: We could also move the nimbus pages into a sub folder like we did with remote settings and leave base at the root so it's clear structurally that it's a shared base?

jrbenny35: There isn't anything yet that is specific to remote settings but there are things in that base class that don't belong to remote settings. It's not a huge deal but it will bother me 😁

jaredlockhart: Ya I think if we do it the way I'm suggesting we move the experimenter specific stuff out, one sec....

jaredlockhart: Actually there was something to put in a remote settings base!  So now there's Base (timeouts/waits), ExperimenterBase (all the save stuff), RemoteSettingsBase (with a shared locator!)

jaredlockhart: Much nicer 👍

PR #6344, opened 2021-09-02
chore(nimbus): use page.open in integration tests

jaredlockhart:
Because

* The integration test PyPom library provides a handy .open method that opens the page and waits for it to load
* This can clean up a lot of code in the tests

This commit

* Uses .open in all the integration tests


jrbenny35 (approved):
Looks good, don't think this hurts readability that much.

PR #6349, opened 2021-09-02
fix #6314 fix(nimbus): do not restore previous search parameters when navigation resets them to empty

### LINKED ISSUE #6314: The "descending" sorting filter cannot be removed once applied to a category if the "tab=" parameter is missing from the address bar
*[Steps to reproduce]:*
1. Open the Nimbus main page without clicking on any of the tabs (e.g. Live, Review, Preview, Completed, Draft, Archived)
2. Click on one of the categories to sort by it twice.
3. Click on the same category a third time.

*[Expected result]:*
- The "descending" sorting filter is removed.

*[Actual result]:*
- The "descending" sorting filter is removed and then reapplied instantly.

*[Notes]:*
- The issue is not reproducible if one of the tabs is opened first and the "tab=" parameter is added. (e.g. tab=live)
- Attached a screen recording of the issue:

descending_not_removed

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1697)

lmorchard:
Because:

* resetting sort order on the index page clears query parameters

This commit:

* applies a bugfix to correctly handle reset query parameters, rather
  than restoring the previous non-empty stored value

### DIFF
--- a/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.test.tsx
+++ b/app/experimenter/nimbus-ui/src/hooks/useSearchParamsState.test.tsx
@@ -7,7 +7,6 @@ import {
   createMemorySource,
   History,
   HistorySource,
-  useLocation,
   useNavigate,
 } from "@reach/router";
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
LGTM 👍

PR #6352, opened 2021-09-02
fix #6337 fix(nimbus): Cloning an archived experiment should not archive the clone

### LINKED ISSUE #6337: Cloning an archived experiment should not archive the clone
I just noticed a small error landed in the clone method.

`cloned.archived = False`

should be

`cloned.is_archived = False`

I guess we're missing a test since the factory always defaults that to false.  Also I'm annoyed that nothing complained about assigning to a non existent field.  

VSCode seems to at least know it's not a field because it highlights it white instead of blue:

signal-2021-09-01-124656_001

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1708)

lmorchard:
Because:

* clone experiments are intended to be fresh drafts

This commit:

* ensures that cloned experiments are properly set to not be archived

### DIFF
--- a/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
+++ b/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
@@ -1027,6 +1027,63 @@ class TestNimbusExperiment(TestCase):
             self.assertEqual(child_branch.feature_value, parent_branch.feature_value)
         self.assertEqual(child.changes.all().count(), 1)
 
+    def test_clone_archived_experiment(self):
+        parent = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE
+        )
+
+        parent.is_archived = True
+        generate_nimbus_changelog(parent, parent.owner, "Archiving experiment")
+
+        child = parent.clone("Child Experiment", parent.owner)
+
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -1027,6 +1027,63 @@ def test_clone_completed_experiment(self):
...
+            set(child.countries.all().values_list("code", flat=True)),
+            set(parent.countries.all().values_list("code", flat=True)),
+        )
+        for parent_branch in parent.branches.all():
+            child_branch = child.branches.get(slug=parent_branch.slug)
+            self.assertEqual(child_branch.name, parent_branch.name)
+            self.assertEqual(child_branch.description, parent_branch.description)
+            self.assertEqual(child_branch.ratio, parent_branch.ratio)
+            self.assertEqual(child_branch.feature_value, parent_branch.feature_value)
+        self.assertEqual(child.changes.all().count(), 1)
```
jaredlockhart, line 1085: We probably don't need to reassert all this in the test since we have this in the other tests, probably just checking `is_archived` is True for the parent and False for the clone is sufficient

jaredlockhart (approved):
I'm so annoyed that nothing caught that this field was completely mispelled >_<  Thnx for fixing 👍

PR #6354, opened 2021-09-02
fix #6243 fix(nimbus): do not display invalid pages alert for an archived experiment

### LINKED ISSUE #6243: Archived experiment still prompts me to fill in missing fields
Possibly related to https://github.com/mozilla/experimenter/issues/6242 and experiment states...

Should an Archived experiment still have a banner that says:

> "Before this experiment can be reviewed or launched, all required fields must be completed. Fields on the Audience page are missing details."

Or should we hide that banner if we're in an Archived state since it's unlikely the experiment will escape its untimely Archived state?

---

Sauce:
Before this experiment can be reviewed or launched, all required fields must be completed. Fields on the Audience page are missing details.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1649)

lmorchard:
Because:

* archived experiments cannot be changed

This commit:

* prevents the display of the alert listing pages with missing details
  for an archived experiment

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.stories.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.stories.tsx
@@ -5,12 +5,14 @@
 import { action } from "@storybook/addon-actions";
 import { withLinks } from "@storybook/addon-links";
 import React from "react";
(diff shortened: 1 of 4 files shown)


jaredlockhart (approved):
👍 👍 👍 👍

PR #6365, opened 2021-09-03
fix #6338 fix(nimbus): use experiment duration as enrollment duration if experiment ends before pause

### LINKED ISSUE #6338: Enrollment duration for complete experiment uses proposed duration
Just noticed a complete experiment is showing the wrong enrollment duration, it's using proposed instead of computed:

Screen Shot 2021-09-01 at 3 59 14 PM

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1709)

lmorchard:
Because:

* experiments can be ended before enrollment is paused

This commit:

* switches to experiment duration rather than proposed duration if
  enrollment wasn't explicitly ended


jaredlockhart (approved):
Oh right yeah if it ends without a pause, good catch 👍

PR #6377, opened 2021-09-07
Fix #6361: fix(visualization) Skip over outcome metrics with no data.

### LINKED ISSUE #6361: Results not loading for experiment
A user just raised that the results are not loading for this experiment:

https://experimenter.services.mozilla.com/nimbus/default-browser-onboarding/results

Console shows:

```
Error while computing statistic bootstrap_mean for metric default_browser_card_go_to_settings_pressed: 'data' contains null values
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1726)

emtwo:
Because:
* Results pages were failing to load if a metric was listed as an outcome metric but it had no data

This commit:
* Will gracefully ignore outcome metrics that don't have data.

### DIFF
--- a/app/experimenter/jetstream/tests/test_tasks.py
+++ b/app/experimenter/jetstream/tests/test_tasks.py
@@ -39,7 +39,7 @@ class TestFetchJetstreamDataTask(TestCase):
         ).dict(exclude_none=True)
 
     def add_outcome_data(self, data, overall_data, weekly_data, primary_outcome):
-        primary_metrics = ["mozilla_default_browser", "default_browser_action"]
+        primary_metrics = ["default_browser_action"]
         range_data = DataPoint(lower=2, point=4, upper=8)
 
         for primary_metric in primary_metrics:
--- a/app/experimenter/jetstream/models.py
+++ b/app/experimenter/jetstream/models.py
@@ -236,7 +236,11 @@ class ResultsObjectModelBase(BaseModel):
                 primary_metric_data = getattr(
                     getattr(branch_data, METRIC_GROUP.get(primary_metric, Group.OTHER)),
                     primary_metric,
+                    None,
                 )
+                if primary_metric_data is None:
+                    continue
+
                 absolute_primary_metric_vals = getattr(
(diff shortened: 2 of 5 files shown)


app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -39,7 +39,7 @@ def get_metric_data(self, data_point):
         ).dict(exclude_none=True)
 
     def add_outcome_data(self, data, overall_data, weekly_data, primary_outcome):
-        primary_metrics = ["mozilla_default_browser", "default_browser_action"]
+        primary_metrics = ["default_browser_action"]
```
jaredlockhart, line 42: So the case we're looking for is where a metric is listed in the outcome but DOESN'T appear in the data right?  So we want to test the cases where it's 
- defined in the outcome and present in the data
- defined in the outcome and missing in the data
- missing in the outcome and present in the data

Which cases are we testing? Do we need to add/modify any other tests?

emtwo:
Screenshot of locally loading the experiment:

Screen Shot 2021-09-07 at 12 36 01 PM

app/experimenter/jetstream/tests/test_tasks.py
emtwo: Since we set the primary outcome in this test like so `primary_outcome = "default-browser"`, there are 2 metrics associated with it, `mozilla_default_browser` and `default_browser_action` based on the toml file. In this function, `add_outcome_data()` we only add one of the 2 outcome metrics to the mock jetstream data, this mimicks having Jetstream return data for `default_browser_action` but not for `mozilla_default_browser`. In the `constants.py` file I also removed the expectation that there will be an API response with data for `mozilla_default_browser`

So ..
`default_browser_action` is defined in the outcome and present in the data
`mozilla_default_browser` is defined in the outcome and missing in the data

Any other metric that is missing in the outcome and present in the data is treated as a default metric

jaredlockhart (approved):
Awesome looks good thank you! 🎉

PR #6383, opened 2021-09-07
chore(nimbus): rename config api fields

jaredlockhart:
Because

* Some of the config API fields were inconsistently singular or plural
* Some fields names didn't reflect the type of data they contained

This commit

* Renames fields that return a collection to pluralized
* Renames fields to match the type of data they return


lmorchard (approved):
Lots of little changes, but something I think TypeScript makes pretty safe

jaredlockhart:
Yeah typescript was a lifesaver here.  There was exactly ONE spot where a field was untyped and I tried to figure out how to get it to typecheck properly but... it was getting gritty so I just left it.

PR #6387, opened 2021-09-08
fix #6381 feat(nimbus): sort directory view options

### LINKED ISSUE #6381: Sort Directory Drop Downs
We should sort the options in all the directory drop downs lexicographically.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1737)

jaredlockhart:
Because

* Its easier to reason about options in a drop down when they're sorted

This commit

* Sorts the options in all the directory view filter drop downs
* Default sorts the experiment list by latest change before applying column specific sorting

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -186,6 +186,7 @@ class NimbusConfigurationType(graphene.ObjectType):
     locales = graphene.List(NimbusLocaleType)
     max_primary_outcomes = graphene.Int()
     outcomes = graphene.List(NimbusOutcomeType)
+    owners = graphene.List(NimbusUser)
     targeting_configs = graphene.List(NimbusExperimentTargetingConfig)
 
     def _text_choices_to_label_value_list(root, text_choices):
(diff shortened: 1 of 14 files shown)


app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
```diff
@@ -68,12 +67,7 @@ export const Body = () => {
     applications: config!.applications!,
     featureConfigs: config!.featureConfigs!,
     firefoxVersions: config!.firefoxVersions!,
-    // Populating owners from fetched experiments since it might be expensive
-    // to fetch and manage a list of all users
```
lmorchard, line 72: Yeah, I guess this was kind of premature optimization. We could always lazy-load this later if it turns onerous to performance.

app/experimenter/experiments/api/v5/types.py
```diff
@@ -221,14 +222,21 @@ def resolve_application_configs(root, info):
...
         return root._text_choices_to_label_value_list(NimbusExperiment.Version)
 
     def resolve_outcomes(root, info):
         return Outcomes.all()
 
+    def resolve_owners(root, info):
+        return (
+            get_user_model()
+            .objects.filter(owned_nimbusexperiments__isnull=False)
+            .order_by("email")
```
lmorchard, line 237: And this is better than "fetch all users" anyway, so :+1:

lmorchard (approved):
:+1: LGTM

app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
jaredlockhart: Oh! I just realized this was the only use of `uniqueByProperty` so I should get rid of that too.

PR #6389, opened 2021-09-08
fix #6366 bug(nimbus): validate clone name length

### LINKED ISSUE #6366: The “Public name” field front-end validation of the Nimbus clone modal does not catch a name length which is too long
**[Prerequisites]:**
- Have a Nimbus experiment with an 80 character name open.

**[Steps to reproduce]:**
1. Click the “Clone” button.
2. Click inside the “Public name” field and then somewhere else on the modal.
3. Click “Save” and observe the modal.

**[Expected result]:**
- A front-end error stating that the name exceeds the 80 character limit is displayed.

**[Actual result]:**
- The front-end validation has a green check for a 81+ character name.
- A “Sorry, an error occurred while submitting. Please try again.” server error is displayed when trying to save the experiment.

**[Notes]:**
81 clone

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1730)

jaredlockhart:
Because

* We neglected to validate that the clone name can't exceed 80 characters

This commit

* Validates clone name is <= 80 characters
* Noticed we were only validating experiment length on the client side, not the server side, so added it in the serializer too

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -310,7 +310,7 @@ class NimbusExperimentSerializer(
     serializers.ModelSerializer,
 ):
     name = serializers.CharField(
-        min_length=0, max_length=255, required=False, allow_blank=True
+        min_length=1, max_length=80, required=False, allow_blank=True
     )
     slug = serializers.ReadOnlyField()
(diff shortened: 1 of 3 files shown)


lmorchard (approved):
:+1:

PR #6390, opened 2021-09-08
fix #6329 feat(nimbus): validate targeting config jexl syntax

### LINKED ISSUE #6329: Validate Targeting JEXL with pyjexl
To help validate that the JEXL expressions in TargetingConfigs are syntactically valid before they land, we can use [PyJEXL](https://github.com/mozilla/pyjexl) to validate each expression.  This doesn't help us validate that they are fully valid in Firefox, for that we'll need to [run them in Firefox](https://github.com/mozilla/experimenter/issues/6308) as part of CI.  But this can at least help prevent simple syntactic errors until we have more complete validation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1704)

jaredlockhart:
Because

* We'd like to prevent targeting configs with invalid JEXL syntax from landing

This commit

* Uses the PyJEXL validator to validate the syntax of all targeting configs
* Adds an extra test to make sure new targeting configs are added to both constant classes

### DIFF
--- a/app/experimenter/experiments/tests/test_targeting_configs.py
+++ b/app/experimenter/experiments/tests/test_targeting_configs.py
@@ -1,0 +1,35 @@
+from django.test import TestCase
+from parameterized import parameterized
+from parsimonious.exceptions import ParseError
+from pyjexl.jexl import JEXLConfig
+from pyjexl.operators import default_binary_operators, default_unary_operators
+from pyjexl.parser import Parser, jexl_grammar
+
+from experimenter.experiments.constants import NimbusConstants
+
+default_config = JEXLConfig({}, default_unary_operators, default_binary_operators)
+
+
+class DefaultParser(Parser):
+    grammar = jexl_grammar(default_config)
+
+    def __init__(self):
+        super().__init__(default_config)
+
+
+class TestTargetingConfigs(TestCase):
+    def test_all_targeting_configs_defined_in_constants(self):
+        self.assertEqual(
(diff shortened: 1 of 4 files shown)


app/experimenter/experiments/tests/test_targeting_configs.py
```diff
@@ -0,0 +1,35 @@
...
+        super().__init__(default_config)
+
+
+class TestTargetingConfigs(TestCase):
+    def test_all_targeting_configs_defined_in_constants(self):
+        self.assertEqual(
+            set([t.value for t in NimbusConstants.TargetingConfig]),
+            set(t for t in NimbusConstants.TARGETING_CONFIGS.keys()),
+            "Targeting Configs must be defined in both NimbusConstants.TargetingConfig "
+            "and NimbusConstants.TARGETING_CONFIGS",
```
lmorchard, line 26: Hmm, too bad we can't DRY these up, but I can see why there are two lists

lmorchard (approved):
:+1:

app/experimenter/experiments/tests/test_targeting_configs.py
jaredlockhart: I know I tried, the problem is you can't define the name if you do it automatically >_<

[It is also possible to make use of the Enum Functional API with the caveat that labels are automatically generated as highlighted above](https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types)

PR #6401, opened 2021-09-09
fix #6394 bug(nimbus): set default branches on backend

### LINKED ISSUE #6394: Modifying the Branches section for a single branch experiment will trigger missing details
**[Prerequisites]:**
- Have a draft ready for launch.

**[Steps to reproduce]:**
1. Open the Branches section and remove all but one branch.
2. Click “Save”.
3. Refresh page.
4. Edit the existing branch (change description).
5. Click "Save"
6. Refresh page.

**[Expected result]:**
- The branch is saved and the experiment is still ready for launch.

**[Actual result]:**
- The Branch section is seen as missing details and can no longer be launched.
-  A second empty branch is displayed.

**[Notes]:**
- The behavior is also reproducible if you launch the recipe after step 2 to Preview, return it to draft state and then edit it and save it as initially found by @piatra  . 
edit and refresh one branch

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1744)

jaredlockhart:
Because

* We recently found a bug where an experiment with 1 branch keeps inserting a second branch, even if it's not visible in the UI
* This will become extra critical when we start launching more rollouts since they only have a single branch
* We persistently had issues like this in legacy, where we'd try to populate empty defaults in the frontend, especially for branches, and the pattern we ended up landing on was removing all special cases from the frontend and letting the frontend treat everything passed from the backend verbatim and then passing empty default values back from the backend

This commit

* Removes the special "Treatment A" cases from the frontend
* Removes creating an empty control branch at creation time
* Passes an empty reference branch from the API if no reference branch exists
* Passes an empty treatment branch from the API if no branches exist (so we pass no empty treatment if only a reference branch does exist)

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -186,14 +186,6 @@ class NimbusExperimentBranchMixin:
 
         return data
 
-    def create(self, validated_data):
-        experiment = super().create(validated_data)
-        experiment.reference_branch = NimbusBranch.objects.create(
-            experiment=experiment, name="Control", feature_enabled=False
-        )
-        experiment.save()
-        return experiment
-
     def update(self, experiment, data):
         with transaction.atomic():
             if set(data.keys()).intersection({"reference_branch", "treatment_branches"}):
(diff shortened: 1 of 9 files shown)


lmorchard (approved):
:+1:

PR #6436, opened 2021-09-14
fix #6430 feat(nimbus): load desktop features from mozilla central

### LINKED ISSUE #6430: Add features app that fetches configs from application repos
Let's add an app called features that

* Fetches the feature manifest for each app's repo
* Loads it into memory at init time
* Provides an interface for it to be consumed in forms

We should be able to do this exactly the same way that we pull in outcomes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1771)

jaredlockhart:
Because

* We'd like to consume feature manifests directly from client repos instead of storing them in the database

This commit

* Downloads the desktop feature manifest at container build time
* Adds a features app that parses the desktop feature manifest into memory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,10 @@ LOAD_LOCALES = python manage.py loaddata ./experimenter/base/fixtures/locales.js
 LOAD_DUMMY_EXPERIMENTS = [[ -z $$SKIP_DUMMY ]] && python manage.py load_dummy_experiments || echo "skipping dummy experiments"
 PUBLISH_STORYBOOKS = npx github:mozilla-fxa/storybook-gcp-publisher --commit-summary commit-summary.txt --commit-description commit-description.txt --version-json experimenter/version.json
 
+
+JETSTREAM_CONFIG_URL = https://github.com/mozilla/jetstream-config/archive/main.zip
+FEATURE_MANIFEST_DESKTOP_URL = https://hg.mozilla.org/mozilla-central/raw-file/default/toolkit/components/nimbus/FeatureManifest.js
+
 ssl: nginx/key.pem nginx/cert.pem
 
 nginx/key.pem:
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -58,22 +62,30 @@ secretkey:
 	openssl rand -hex 24
 
 jetstream_config:
-	curl -LJ -o app/experimenter/outcomes/jetstream-config.zip https://github.com/mozilla/jetstream-config/archive/main.zip
+	curl -LJ -o app/experimenter/outcomes/jetstream-config.zip $(JETSTREAM_CONFIG_URL)
 	unzip -o -d app/experimenter/outcomes app/experimenter/outcomes/jetstream-config.zip
 
+feature_manifests:
+	curl -LJ --create-dirs -o app/experimenter/features/manifests/firefox-desktop.js $(FEATURE_MANIFEST_DESKTOP_URL)
+	echo "console.log(JSON.stringify(FeatureManifest));" >> app/experimenter/features/manifests/firefox-desktop.js
```
lmorchard, line 70: Well that's a nifty quick & dirty JSON conversion. Might be worth wrapping this up in a shell script? But it's only 2 lines, so also probably not worth thinking about until/unless it grows.

Might also be worth running inside a container in case the host doesn't have node? Also probably not a practical concern.

lmorchard (approved):
Looks good, makes sense, tests pass :+1:

Scratched my head at the Makefile a little, but I mostly thinking out loud about potential gotchas we'll likely never see

Makefile
jaredlockhart: Yeah I know I was thinking the same but we don't currently have an entry point to stick it... and I wanted to keep it all together.  And yeah tacking console.log on to the end is super hacky but the js files in mc are weird and I couldn't get it to be importable by a separate script. 

And yes I just realized I should update the README to indicate installing node as a build requirement which is only listed under the semi dockerized setup.

PR #6437, opened 2021-09-15
fix 6308 - Validate targetting configs for desktop on CI.

jrbenny35:


app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -90,3 +94,49 @@ def test_end_experiment_and_reject_end(
...
+    base_url,
+    kinto_url,
+    kinto_review_url,
+    experiment_url,
+    default_data,
+    create_experiment,
+    slugify,
+):
+    default_data.audience.targeting = (
+        BaseExperimentAudienceTargetingOptions.TARGETING_MAC_ONLY
```
jaredlockhart, line 111: So it looks like this test is using only a single targeting config option, but the ticket says

"So ideally we'd have a test for each targeting config that ... Creates a Nimbus experiment for Desktop with that targeting config and no other custom targeting (we might be able to do this from the command line without needing to start the dev server at all)"

So we'll need to find a way to capture every targeting config that is landed in experimenter and do a test for each one.  Ideally we'll be able to discover all the targeting configs that are defined in NimbusConstants rather than hard coding them here so that when someone adds a new one they don't also have to come modify the integration tests.

app/tests/integration/nimbus/pages/experimenter/audience.py
```diff
@@ -39,6 +44,41 @@ def min_version(self, version=80):
         select = Select(el)
         select.select_by_value(f"FIREFOX_{version}")
 
+    @property
+    def locale(self):
```
jaredlockhart, line 48: The instructions in the ticket say:

"So ideally we'd have a test for each targeting config that ... Creates a Nimbus experiment for Desktop with that targeting config and no other custom targeting (we might be able to do this from the command line without needing to start the dev server at all)"

So I don't think we need to touch locales/countries here at all since that falls under 'other custom targeting'.  The only thing we're interested in validating here is the JEXL body of the TargetingConfig entries in NimbusConstants.

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -90,3 +94,49 @@ def test_end_experiment_and_reject_end(
...
+    default_data.audience.locale = "English (Canadian)"
+    default_data.audience.countries = "Canada"
+    create_experiment(selenium).launch_and_approve()
+
+    Login(selenium, kinto_url).open().login()
+    Dashboard(selenium, kinto_review_url).open().approve()
+    SummaryPage(selenium, experiment_url).open().wait_for_live_status()
+    title = default_data.public_name
+    base_url = urlparse(base_url)
+    json_url = f"https://{base_url.netloc}/api/v6/experiments/{slugify(title)}"
```
jaredlockhart, line 122: This could probably be a fixture?

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -90,3 +94,49 @@ def test_end_experiment_and_reject_end(
...
+    if environment is False:
+        environment = "en-US"
+    expression = "&&".join(expression)
+    environment = "&&".join(environment)
+    # Inject filter expression
+    selenium.get("about:blank")
+    with open("nimbus/utils/filter_expression.js") as js:
+        with selenium.context(selenium.CONTEXT_CHROME):
+            script = selenium.execute_script(js.read(), 90.1, environment, expression)
+    assert script is not None
```
jaredlockhart, line 142: It might be nice to create a Page object that wraps this so we can interact with it at a higher level in the test body?

app/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -0,0 +1,20 @@
...
+Components.utils.import("resource://gre/modules/components-utils/ClientEnvironment.jsm");
+
+// Targeting
+
+async function remoteSettings(arguments) {
+
+    const { TargetingContext } = ChromeUtils.import(
+        "resource://messaging-system/targeting/Targeting.jsm"
+    );
+    const ENVIRONMENT = {version: arguments[0], localeLanguageCode: arguments[1]};
```
jaredlockhart, line 11: So here we're building an environment explicitly, but this won't be the same environment that Firefox desktop actually uses when a real experiment is launched.  

The goal of this test is to validate that the JEXL in each TargetingConfig will run as expected in Firefox, which includes both being syntactically valid (which is now checked in unit tests using PyJEXL) and that each clause in the expression will evaluate to a symbol in the environment and not undefined (ie if someone mispells a variable).

So we'll need to pull in the real environment that Firefox provides rather than specifying it here manually.

jaredlockhart (changes requested):
Awesome, this is a good start @jrbenny35 !  But it looks like it doesn't quite line up with what the ticket is asking for, I added more context but let me know if you still need more information 👍

app/tests/integration/nimbus/utils/filter_expression.js
jaredlockhart: Looks like this should do it:

https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/lib/RemoteSettingsExperimentLoader.jsm#110-113

app/tests/integration/nimbus/test_remote_settings.py
jaredlockhart: Couple ways I can think of to do it are

1) Discover the list from the drop down options in the UI, but that would depend on creating an experiment first to get to that page then pulling all the options out and then creating an experiment for each one

2) Use a management command to create an experiment with each targeting config during the setup phase in the Makefile which can then just import the TargetingConfig list directly and create them with predictable names that the tests can then discover

3) Query the configuration endpoint with the graphql API which will return them all as JSON

app/tests/integration/nimbus/utils/filter_expression.js
jrbenny35: So how do I do this?

app/tests/integration/nimbus/test_remote_settings.py
jrbenny35: "capture every targeting config", where would this be defined?

jaredlockhart: This is probably the easiest way to do it, call the V5 graphql API and it'll return them all:

```python
In [14]: requests.post('https://localhost/api/v5/graphql', json={'operationName': 'getConfig',
    ...:  'variables': {},
    ...:  'query': '\nquery getConfig {\n  nimbusConfig {\n    targetingConfigs {\n      label\n      value\n      applicationValues\n    }\n  }\n}\n'}, verify=Fals
    ...: e).content
/Users/jaredlockhart/Library/Caches/pypoetry/virtualenvs/app-9BVIRsKI-py3.8/lib/python3.8/site-packages/urllib3/connectionpool.py:981: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
Out[14]: b'{"data":{"nimbusConfig":{"targetingConfigs":[{"label":"No Targeting","value":"","applicationValues":["DESKTOP","FENIX","IOS","FOCUS_ANDROID","KLAR_ANDROID"]},{"label":"First start-up users","value":"first_run","applicationValues":["DESKTOP"]},{"label":"First start-up users from Chrome","value":"first_run_chrome","applicationValues":["DESKTOP"]},{"label":"First start-up users on Windows 10 1903 (build 18362) or newer","value":"first_run_win1903","applicationValues":["DESKTOP"]},{"label":"Homepage set to google.com","value":"homepage_google_dot_com","applicationValues":["DESKTOP"]},{"label":"Urlbar (Firefox Suggest)","value":"urlbar_firefox_suggest","applicationValues":["DESKTOP"]},{"label":"Mac OS users only","value":"mac_only","applicationValues":["DESKTOP"]},{"label":"No enterprise users","value":"no_enterprise_users","applicationValues":["DESKTOP"]},{"label":"First start-up users","value":"first_run_mobile","applicationValues":["FENIX","IOS"]},{"label":"Infrequent user (uris)","value":"infrequent_user_uris","applicationValues":["DESKTOP"]},{"label":"Casual user (uris)","value":"casual_user_uris","applicationValues":["DESKTOP"]},{"label":"Regular user (uris)","value":"regular_user_uris","applicationValues":["DESKTOP"]},{"label":"Core user (uris)","value":"core_user_uris","applicationValues":["DESKTOP"]},{"label":"No enterprise or past VPN use","value":"no_enterprise_or_past_vpn","applicationValues":["DESKTOP"]}]}}}'
```

app/tests/integration/nimbus/utils/filter_expression.js
jaredlockhart: Well you probably just need to import the RemoteSettingsExperimentLoader module and call that method?

jrbenny35: yup, I added that comment before I refreshed the page.

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -90,3 +94,49 @@ def test_end_experiment_and_reject_end(
...
+    create_experiment(selenium).launch_and_approve()
+
+    Login(selenium, kinto_url).open().login()
+    Dashboard(selenium, kinto_review_url).open().approve()
+    SummaryPage(selenium, experiment_url).open().wait_for_live_status()
+    title = default_data.public_name
+    base_url = urlparse(base_url)
+    json_url = f"https://{base_url.netloc}/api/v6/experiments/{slugify(title)}"
+    # Get experiment JSON and parse
+    experiment_json = requests.get(f"{json_url}", verify=False).json()
```
jaredlockhart, line 124: This is `experiment recipe`

app/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -0,0 +1,20 @@
...
+    let targeting = new TargetingContext(ENVIRONMENT);
+    const EXPRESSION = `${arguments[2]}`;
+    const actual = await targeting.evalWithDefault(EXPRESSION, ENVIRONMENT);
+    return actual
```
jaredlockhart, line 15: I would actually do the undefined comparison right in here so we don't lose it or map it onto a Python type:

```
typeof myVar !== 'undefined'
```

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -90,3 +94,49 @@ def test_end_experiment_and_reject_end(
...
+    default_data.audience.targeting = (
+        BaseExperimentAudienceTargetingOptions.TARGETING_MAC_ONLY
+    )
+    default_data.audience.locale = "English (Canadian)"
+    default_data.audience.countries = "Canada"
+    create_experiment(selenium).launch_and_approve()
+
+    Login(selenium, kinto_url).open().login()
+    Dashboard(selenium, kinto_review_url).open().approve()
+    SummaryPage(selenium, experiment_url).open().wait_for_live_status()
```
jaredlockhart, line 119: We don't actually need to do this

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -90,3 +94,49 @@ def test_end_experiment_and_reject_end(
...
+    default_data,
+    create_experiment,
+    slugify,
+):
+    default_data.audience.targeting = (
+        BaseExperimentAudienceTargetingOptions.TARGETING_MAC_ONLY
+    )
+    default_data.audience.locale = "English (Canadian)"
+    default_data.audience.countries = "Canada"
+    create_experiment(selenium).launch_and_approve()
```
jaredlockhart, line 115: Actually instead of launching/approving just send it to preview and then you don't need to do the whole remote settings flow but it will still show up in the V6 API.

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -1,9 +1,38 @@
...
+            "{\n    targetingConfigs {\n      "
+            "label\n      value\n      applicationValues\n    "
+            "}\n  }\n}\n",
+        },
+        verify=False,
+    ).json()
+    for item in data["data"]["nimbusConfig"]["targetingConfigs"]:
+        if "DESKTOP" in item["applicationValues"]:
+            apps.append(item["value"])
+    return apps
```
jaredlockhart, line 28: List comprehensions are your friend 👬 

```py
return [item for item in data["data"]["nimbusConfig"]["targetingConfigs"] if "DESKTOP" in item["applicationValues"]]
```

jaredlockhart (changes requested):
Yes awesome, loading the the targeting configs seems to work great.  It's still not actually testing the right thing though, if you go into one of the targeting configs and break the targeting string, like deleting a single character from one of the variables, the test still passes, so it looks like we need to more strongly assert the output of the jexl evaluator.  

Testing locally it looks like if it hits an undefined value it returns `null`, example:

```js
await experiment_loader.evaluateJexl('isFirstStartup', {experiment: {}});
false

await experiment_loader.evaluateJexl('isFirstSartup', {experiment: {}});
null
```

so we probably need to strongly check we're not getting null back on the js side.

jaredlockhart:
Actually we probably can't use this function at all since for malformed filter expressions it'll throw, catch, and return `false` which is no good for us.  See if you can copy the actual implementation of that function here:

https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/lib/RemoteSettingsExperimentLoader.jsm#249-265

and handle the `catch` case explicitly.

So then we should get something like:

1) `"isFirstStartup"` > pass
2) `"isFirstSartup"` > fail
3) `"'app.shield.optoutstudies.enabled'|peferenceValue"` > fail

app/tests/integration/nimbus/test_remote_settings.py
jrbenny35: Yeah this is something I am not good at lol

jrbenny35:
> 
> 
> Actually we probably can't use this function at all since for malformed filter expressions it'll throw, catch, and return `false` which is no good for us. See if you can copy the actual implementation of that function here:
> 
> https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/lib/RemoteSettingsExperimentLoader.jsm#249-265
> 
> and handle the `catch` case explicitly.
> 
> So then we should get something like:
> 
>     1. `"isFirstStartup"` > pass (correctly spelled variable)
> 
>     2. `"isFirstSartup"` > fail (incorrectly spelled variable)
> 
>     3. `"'app.shield.optoutstudies.enabled'|peferenceValue"` > fail (incorrectly spelled filter expression)

So this will return true or false not undefined/null?

jaredlockhart:
> > Actually we probably can't use this function at all since for malformed filter expressions it'll throw, catch, and return `false` which is no good for us. See if you can copy the actual implementation of that function here:
> > https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/lib/RemoteSettingsExperimentLoader.jsm#249-265
> > and handle the `catch` case explicitly.
> > So then we should get something like:
> > ```
> > 1. `"isFirstStartup"` > pass (correctly spelled variable)
> > 
> > 2. `"isFirstSartup"` > fail (incorrectly spelled variable)
> > 
> > 3. `"'app.shield.optoutstudies.enabled'|peferenceValue"` > fail (incorrectly spelled filter expression)
> > ```
> 
> So this will return true or false not undefined/null?

Work with @piatra to figure out the exact behaviour of the function and use those three cases I listed above as test criteria for this PR to be complete.  I'll hand review duty of this over to @lmorchard

jaredlockhart:
And here's some extra clarity on the three acceptance criteria:

     1. `"isFirstStartup"` > function returns `bool`, test passes (correctly spelled variable)
 
     2. `"isFirstSartup"` > function returns `null`, test fails (incorrectly spelled variable)
 
     3. `"'app.shield.optoutstudies.enabled'|peferenceValue"` > function throws error, test fails (incorrectly spelled filter expression)

app/tests/integration/nimbus/pages/browser/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Browser Model"""
+
+
+class Browser:
+    def execute_script(selenium, *args, script=None, context=None):
+        if "chrome" in context:
```
pdehaan, line 6: Python Q: Is `context` a string, or array/object? I see one usage below which seems to specify `"chrome"` exact string, but curious why the `in` instead of exact string matching here.

app/tests/integration/nimbus/pages/experimenter/summary.py
```diff
@@ -13,6 +13,7 @@ class SummaryPage(ExperimenterBase):
 
     _page_wait_locator = (By.CSS_SELECTOR, "#PageSummary")
     _approve_request_button_locator = (By.CSS_SELECTOR, "#approve-request-button")
+    _launch_to_preview = (By.CSS_SELECTOR, "#launch-to-preview-button")
```
pdehaan, line 16: nit: should this have a `_locator` suffix for consistency?

app/tests/integration/nimbus/parallel_pytest_args.txt
```diff
@@ -4,3 +4,5 @@
 -k KLAR -m run_per_app
 -k FOCUS -m run_per_app
 -k DESKTOP -m run_once
+-k DESKTOP -m run_parallel -n 2
```
pdehaan, line 7: Fasinating! Per above:

> parallelism: 7 # This should match the number of entries in parallel_pytest_args.txt

Does "-n 2" mean we should increase parallelism by another level? (so 8 instead of 7). Or maybe I'm not fully grasping what `-n` is doing here.

app/tests/integration/nimbus/test_remote_settings.py
pdehaan: Not sure if you want to do this or not. I'm indifferent to syntax if you think one is better or more legible than the other option.

app/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -0,0 +1,25 @@
...
+
+// Targeting
+
+async function remoteSettings(arguments) {
+
+    const TargetingContext = ChromeUtils.import(
+        "resource://messaging-system/targeting/Targeting.jsm"
+    )
+
+    _experiment = JSON.parse(arguments[1])
```
pdehaan, line 12: Not sure if you want to add any JavaScript comments around here on what `arguments[1]` (and `arguments[0]` indexes represent below). Ie: One's an expression, one's the data/context or whatever.

app/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -0,0 +1,25 @@
...
+    const targetingContext = new TargetingContext.TargetingContext(_experiment);
+    let result = false;
+    try {
+        result = await targetingContext.evalWithDefault(arguments[0]);
```
pdehaan, line 16: What will this return? A Boolean `true` or `false` or throw an error which gets caught below on L17?

app/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -0,0 +1,25 @@
+Components.utils.import("resource://gre/modules/components-utils/FilterExpressions.jsm");
+Components.utils.import("resource://gre/modules/components-utils/ClientEnvironment.jsm");
+
+// Targeting
+
+async function remoteSettings(arguments) {
+
+    const TargetingContext = ChromeUtils.import(
+        "resource://messaging-system/targeting/Targeting.jsm"
+    )
```
pdehaan, line 10: nit: missing semi-colon. I wonder if these are getting linted by ESLint. Possibly not if they're new .js files in a new location.

Same nit would apply to L12, L18, and L23 below.

app/tests/integration/nimbus/utils/filter_expression.js
piatra: Overall I think it's better that we're using this custom code and not calling into `evaluateJexl` directly.
This way we can catch the error and return it to python for a better error message.

> What will this return? A Boolean true or false or throw an error which gets caught below on L17?

Depends on the argument `a == 1` will return `true/false`, something like `true && foo` where `foo` is an undefined variable will return `undefined` or that variables value. Any syntax errors: typos or unknown functions will throw. It should match [our docs](https://experimenter.info/desktop-targeting-debug#targeting).

app/tests/integration/nimbus/pages/browser/__init__.py
jrbenny35: ```context``` is a string. Using the ```in``` operator, python will compare the strings.

app/tests/integration/nimbus/parallel_pytest_args.txt
jrbenny35: this is for pytest, the parallelism thing in the circleci config is for circleci. This adds another circleci parallel job and then tells pytest to run things with 2 workers in parallel.

app/tests/integration/nimbus/test_remote_settings.py
jrbenny35: I like list comprehensions but sometimes I don't. Not sure here, I'll just add it.

app/tests/integration/nimbus/pages/browser/__init__.py
pdehaan: Ah! OK, so it looks like `in` does substring checking (so `context` could be "chrome" or "somethinglikechrome").

app/tests/integration/nimbus/test_remote_settings.py
jrbenny35: On second thought, the way I have the test setup is looking for the data structure I had. I don't think this will add much adding it.

pdehaan (approved):
I think this is awesome. Great job, @jrbenny35!

There are 2 unresolved @jaredlockhart comments above (in test_remote_settings.py). Otherwise, 👍 .

PR #6440, opened 2021-09-16
fix(nimbus): Update infrequent windows users targeting to check for specific build number

piatra:
We're using this targeting condition for a message that asks users to pin Firefox to the taskbar and that API is not available on earlier Windows builds.


jaredlockhart (approved):
👍

PR #6441, opened 2021-09-16
fix(nimbus): Infrequent non-default windows users; increment build ve…

piatra:
…rsion


piatra:
Another one for you @jaredlockhart thanks 😁

lmorchard (approved):
LGTM insofar as tests pass and JEXL validation seems happy

jaredlockhart:
@piatra One thing that I'm curious about is whether modifying targeting configurations in place will inadvertently or adversely other experiments which may also be using them.  If only one body of experiments uses them then it's probably fine but I'm curious if we might in the future see cases where one team modifies a targeting configuration and accidentally disrupts the experiments of another.

piatra:
I think you're right and targeting expressions should never update (unless a bug is found maybe) from their original usecase.
This one that I've changed here has never been in use yet and these updates are the result of the review and QA process.

PR #6444, opened 2021-09-16
fix #6421 feat(nimbus): add support for screenshot upload and in-place update for branches

### LINKED ISSUE #6421: Implement backend API for attaching a screenshot to a branch
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1767)

lmorchard:
Because:

* we want to support uploading and managing screenshots associated with
  branches

This commit:

* switches from GraphQLView to FileUploadGraphQLView in v5 API urls to
  add support for multipart file uploads

* reduces length of screenshot filenames by removing the branch slug

* adds BranchScreenshotType under BranchType for GQL experiment input

* enhancements to NimbusBranchSerializer to support for specifying
  branch ID to perform in-place update

* adds NimbusBranchScreenshotSerializer to handle screenshot management

* adds NimbusBranchReadyForReviewSerializer to validate screenshots for
  review readiness without requiring screenshot content

* sets up a new settings_test.py module for testing-specific overrides
  to Django settings

* adds UPLOADS_FILE_STORAGE setting to override file storage for uploads
  with a named class (e.g. inmemorystorage.InMemoryStorage)

* adds NimbusBranchScreenshotFactory to generate screenshots

* updated GQL types to account for new screenshots model, with temporary
  stub for Upload type in nimbus-ui (to be defined in EXP-1766)

* adds graphene-file-upload and dj-inmemorystorage dependencies

### DIFF
--- a/app/experimenter/base/__init__.py
+++ b/app/experimenter/base/__init__.py
@@ -2,7 +2,8 @@ import json
 from functools import cache
 
 from django.conf import settings
-from django.core.files.storage import FileSystemStorage
+from django.core.files.storage import FileSystemStorage, get_storage_class
+from django.utils.functional import LazyObject
 from google.oauth2 import service_account
 from storages.backends.gcloud import GoogleCloudStorage
 
@@ -23,8 +24,11 @@ def app_version():
     return app_version
 
 
-@cache
 def get_uploads_storage():
+    if settings.UPLOADS_FILE_STORAGE:
+        cls = get_storage_class(settings.UPLOADS_FILE_STORAGE)
+        return cls()
+
     if settings.UPLOADS_GS_CREDENTIALS:
         credentials = service_account.Credentials.from_service_account_file(
             settings.UPLOADS_GS_CREDENTIALS
@@ -36,3 +40,10 @@ def get_uploads_storage():
         )
 
     return FileSystemStorage()
+
+
+# Borrowing Django's DefaultStorage lazy object approach
+# https://github.com/django/django/blob/ddae36700d1c07cbca15a6821bd99d480294c137/django/core/files/storage.py#L271-L273
+class UploadsStorage(LazyObject):
+    def _setup(self):
+        self._wrapped = get_uploads_storage()
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -11,6 +11,7 @@ from experimenter.experiments.constants.nimbus import NimbusConstants
 from experimenter.experiments.models import NimbusExperiment
 from experimenter.experiments.models.nimbus import (
     NimbusBranch,
+    NimbusBranchScreenshot,
     NimbusDocumentationLink,
     NimbusFeatureConfig,
 )
@@ -46,7 +47,36 @@ class ExperimentNameValidatorMixin:
         return name
 
 
+class NimbusBranchScreenshotSerializer(serializers.ModelSerializer):
+    id = serializers.IntegerField(required=False)
+    description = serializers.CharField(max_length=1024, required=True)
+    image = serializers.ImageField(required=False, allow_empty_file=False, use_url=True)
+
+    class Meta:
+        model = NimbusBranchScreenshot
+        fields = (
+            "id",
+            "description",
+            "image",
+        )
+
+
 class NimbusBranchSerializer(serializers.ModelSerializer):
+    id = serializers.IntegerField(required=False)
+    screenshots = NimbusBranchScreenshotSerializer(many=True, required=False)
+
+    class Meta:
+        model = NimbusBranch
+        fields = (
+            "id",
+            "name",
+            "description",
+            "ratio",
+            "feature_enabled",
+            "feature_value",
+            "screenshots",
+        )
+
     def validate_name(self, value):
         slug_name = slugify(value)
         if not slug_name:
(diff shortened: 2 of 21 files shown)


lmorchard:
The big rework here is to branch updates: Rather than fully delete and re-create branches, this adds support for specifying per-branch ID from the front-end to indicate an in-place change. That way, we can edit branches without throwing away all the child screenshots.

Should be backwards compatible with the current front-end, though: If no IDs are supplied, as per the current implementation, the result is functionally the same - i.e. delete all branches and re-create from GQL submission.

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -355,6 +359,79 @@ def test_update_experiment_branches_with_feature_config_error(self):
...
+        self.assertEqual(experiment.reference_branch.name, reference_branch["name"])
+        self.assertEqual(experiment.reference_branch.screenshots.count(), 2)
+
+        treatment_branch = experiment.treatment_branches[0]
+        self.assertEqual(treatment_branch.name, treatment_branches[0]["name"])
+        self.assertEqual(treatment_branch.screenshots.count(), 1)
+
+        screenshot = treatment_branch.screenshots.first()
+
+        logging.info(f"SCREEN {screenshot} {screenshot.image}")
```
lmorchard, line 433: Oh, that doesn't belong there...

jaredlockhart:
@lmorchard Okay crazy idea, do you think you can pull the change to modifying the branches in place with ID into a separate PR and rebase this one on top of it?

lmorchard:
The branch modifications is really the main change here. The rest are just small type changes, really

lmorchard:
Looking through it again, I could maybe separate out the storage class change for the in-memory testing stuff?

But yeah, the bulk of this is updates to serializers to handle the nested resources for upload with the in-place updates added along the way. I don't think the in-place stuff is easily separable from the nested resources handling without some rewrites

app/experimenter/experiments/tests/api/v5/test_serializers.py
```diff
@@ -291,6 +295,54 @@ def test_branch_with_invalid_feature_value_jso(self):
...
+        self.assertTrue(branch_serializer.is_valid(), branch_serializer.errors)
+        branch_serializer.save()
+
+        branch = NimbusBranch.objects.get(pk=branch.id)
+        self.assertEqual(branch.screenshots.count(), 2)
+
+        for screenshot in deleted_screenshots:
+            self.assertFalse(branch.screenshots.filter(pk=screenshot.id).exists())
+
+        screenshots = branch.screenshots.order_by("id")
```
jaredlockhart, line 333: I wouldn't depend on this for a stable ordering, I've seen countless teams that the db can reorder inserts/queries depending on the phase of the moon, so see if there's a way you can distinguish some other way?

app/experimenter/settings_test.py
```diff
@@ -0,0 +1,3 @@
+from .settings import *
+
+UPLOADS_FILE_STORAGE = "inmemorystorage.InMemoryStorage"
```
jaredlockhart, line 3: You know we had one of these and then got rid of it a while back.  You can also decorate the relevant classes with the override.  I'm hesitant to add this back in now just for one setting.

jaredlockhart (changes requested):
Yep everything looks broadly correct to me, just a couple nits.  Gonna test locally now.

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -603,6 +688,17 @@ def save(self, *args, **kwargs):
...
 
+class NimbusBranchScreenshotReadyForReviewSerializer(NimbusBranchScreenshotSerializer):
+    # Round-trip serialization & validation for review can use a string path
+    image = serializers.CharField(required=True)
+
+
+class NimbusBranchReadyForReviewSerializer(NimbusBranchSerializer):
+    screenshots = NimbusBranchScreenshotReadyForReviewSerializer(
+        many=True, required=False
+    )
```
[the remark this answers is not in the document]
lmorchard: This only triggers when there are actually screenshots and then only to ensure that the screenshots have descriptions & images. 

Needed to add this to address a bunch of failing tests that fell over when screenshots *were* defined (e.g. via factory) and the validated data was a string rather than an image blob - e.g. [this round-trip pattern](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/api/v5/test_serializers.py#L1735-L1742):

```python
        serializer = NimbusReadyForReviewSerializer(
            experiment,
            data=NimbusReadyForReviewSerializer(
                experiment,
                context={"user": self.user},
            ).data,
            context={"user": self.user},
        )
```

app/experimenter/settings_test.py
lmorchard: I tried decorating individual tests. Two problems with that:

1. This is just about everything under nimbus models, since I added a factory to generate screenshots with branches. 

2. The storage is cached and gets used at the time of model class definition. I found that per-test or per-class override of storage type had no effect, so I ended up resorting to this.

Could poke at more of a lazy-loading storage class, but figured this was simpler.

app/experimenter/experiments/tests/api/v5/test_serializers.py
lmorchard: We use it over here in [the `treatment_branches` property](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L281), so I figured it was convenient enough here.

app/experimenter/base/__init__.py
```diff
@@ -36,3 +40,10 @@ def get_uploads_storage():
         )
 
     return FileSystemStorage()
+
+
+# Borrowing Django's DefaultStorage lazy object approach
+# https://github.com/django/django/blob/ddae36700d1c07cbca15a6821bd99d480294c137/django/core/files/storage.py#L271-L273
+class UploadsStorage(LazyObject):
+    def _setup(self):
+        self._wrapped = get_uploads_storage()
```
lmorchard, line 49: Made some tweaks that *should* make overriding upload file storage easier on a test-by-test level. Still thinking about how to do it without having to decorate a lot of tests or make changes to otherwise unaffected tests

app/experimenter/settings_test.py
jaredlockhart: Oh yeah, fair point.  Okay yeah better to get something working landed and figure it out later if it becomes a hassle.  The reason we got rid of it was every so often you'd forget it exists and have to poke around to figure out where something is getting set, but yeah I see your point it happens at init time so too late to decorate.

jaredlockhart (approved):
Yep yip yep yep to all of this, all looks right to me.  Great job 👏 👏 👏 👏

PR #6450, opened 2021-09-17
fix #6447 feat(nimbus): move feature select to top of branch page

### LINKED ISSUE #6447: Move feature select to top of branch form
Let's move the feature select out of each branch and into the top of the form so we can add the space for adding custom features which will require a couple fields.

jaredlockhart:
Because

* We will soon need to add a form for custom feature configs
* We'll want to stick it right under the feature config select

This commit

* Moves the feature config select to the top of the branch form


lmorchard:
Weird: Made some tweaks and `yarn test` passes for me, but `make check` doesn't and seems to fail on unrelated tests. Still digging at this PR.

jaredlockhart:
Oh I can't review this XD

jrbenny35 (approved):

PR #6470, opened 2021-09-22
fix #6468 fix(nimbus): do not delete experiment when deleting reference_branch

### LINKED ISSUE #6468: The links from Overview and the Locales and Countries filters from Audience are deleted when the Branches section is saved
**[Prerequisites]:**
- Have a draft of a Nimbus experiment open on Overview.

**[Steps to reproduce]:**
1. Save a few links in the “Additional Links” section.
2. Go to the “Audience” section.
3. Save options in the “Locales” and “Countries” fields.
4. Go to the “Branches” section.
5. Click “Save”.
6. Go to “Summary” and observe the “Additional links”,  “Locales” and “Countries” sections.

**[Expected result]:**
- Saved links, country and locale filters remain saved.

**[Actual result]:**
- Saved links, country and locale filters are deleted when the “Branches” section is saved.

**[Notes]:**
- The data is deleted from the “Audience” and “Overview” sections as well, not just Summary.
- The data does not seem to be deleted when other sections are saved.
- Whether the "Save" or "Save and Continue" buttons are used does not affect the result.

deleted data

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1788)

lmorchard:
Because:

* deleting a reference_branch cascaded to deleting the parent experiment

This commit:

* switches to `on_delete=SET_NULL` so that the experiment is not deleted


lmorchard:
I'm not sure why the previous deletion code causes issue #6468, but this tweak seems to fix it. Maybe `on_delete=models.CASCADE`, but that seems like the inverse of what should happen?

lmorchard:
On reviewing [the Django docs](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.CASCADE):
> Django emulates the behavior of the SQL constraint ON DELETE CASCADE *and also deletes the object containing the ForeignKey.*

So, actually, it seems like we want `on_delete=models.SET_NULL` because we *don't* want the experiment deleted along with deleting a reference branch. I think what's happening here is that the experiment is deleted but then recreated with the same old ID, and thus all the other foreign key associations (i.e. links, etc) are lost

lmorchard:
Hmm, we might also want to make this change for `feature_config` too - i.e. deleting a feature config shouldn't delete all experiments pointing at it.

lmorchard:
Shotgunning some review requests out there, hopefully this is a quick one

jodyheavener (approved):
Don't know _toooo_ much about this, but I went through the repro steps and this works as described, and I don't see anything glaring

Issue #6424, opened 2021-09-10
Create GCP buckets for branch screenshot storage for stage & production


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1770)

data-sync-user:
➤ Les Orchard commented:

Filed https://mozilla-hub.atlassian.net/browse/DSRE-162 ( https://mozilla-hub.atlassian.net/browse/DSRE-162|smart-link ) to hopefully kick off this process.

PR #6472, opened 2021-09-23
fix #6420 feat(nimbus): branch screenshot management and display

### LINKED ISSUE #6420: Implement front end UI for attaching screenshots to a branch
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1766)

lmorchard:
Because:

* we want to support managing and displaying screenshots associated
  with branches

This commit:

* adds apollo-upload-client dependency and switches the Apollo client to
  use it for multipart submission support and file upload

* adds FormScreenshot subcomponent to FormBranches to support uploading
  screenshots with descriptions

* updates to FormBranches reducer to handle managing screenshots

* implements add / remove screenshot support in FormBranch and renders
  FormScreenshot as appropriate

* misc type updates to accommodate screenshots and in-place updates of
  branches using IDs

* adds support to copy screenshot files as part of experiment cloning

* sets the default ordering for screenshots to use ID

* resolves screenshot images to URLs in GQL query data

* relaxes required fields to allow saving partially composed screenshots


lmorchard:
Opening as a draft to have something posted before I close up shop for the day. Seems there's still an integration test failing?

Trying to think if there's any way I could split this into smaller pieces, but it's all pretty integral and more complex than I'd initially estimated. This will take a bit to review - I can do a walkthrough on zoom if necessary

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/index.test.ts
```diff
@@ -61,6 +64,120 @@ describe("extractUpdateState", () => {
...
+
+  it("sets screenshot images to undefined if not changed", () => {
+    const branch = {
+      ...baseBranch,
+      screenshots: [
+        { id: 123, description: "first", image: "/some/image1" },
+        { id: 456, description: "second", image: "/some/image2" },
+      ],
+    };
+    const file = new File(["blah"], "blah.png");
```
jodyheavener, line 120: Looks unused in this test?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormScreenshot/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+        {...{
+          defaultValues: { image: {} },
+        }}
+      />,
+    );
+    expect(screen.queryByTestId("upload-label")).toHaveTextContent("");
+    expect(screen.queryByTestId("upload-preview")).not.toBeInTheDocument();
+  });
+
+  it("displays a preview of existing screenshot image", async () => {
```
jodyheavener, line 81: Pretty minor, so feel free to ignore, but most of these new tests don't need to be `async`

app/experimenter/nimbus-ui/src/services/apollo.ts
```diff
@@ -29,9 +30,11 @@ export const cacheConfig: InMemoryCacheConfig = {
 export const cache = new InMemoryCache(cacheConfig);
 
 export function createApolloClient() {
-  const httpLink = createHttpLink({
+  // HACK: the type coercion here might be a bad idea, but it seems functional?
+  // https://github.com/jaydenseric/apollo-upload-client/issues/268#issuecomment-893907474
+  const httpLink = createUploadLink({
     uri: config.graphql_url,
-  });
+  }) as unknown as ApolloLink;
```
jodyheavener, line 37: 👍

jodyheavener (commented):
I think I was able to go through everything, and it all works as I expected it to. Nothing major in the code, just questions and nitpicks.

One thing I did notice was that I was able to click "Add screenshot" a bunch and save the form without selecting images - should totally empty screenshot sets be stripped?

👏

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormScreenshot/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+}
+
+const UploadPreviewImage = ({ upload }: { upload: Upload | null }) => {
+  let previewUrl = null;
+  if (upload) {
+    if (typeof upload === "string") {
+      previewUrl = upload;
+    } else if (typeof upload === "object" && "name" in upload) {
+      previewUrl = URL.createObjectURL(upload);
+    }
```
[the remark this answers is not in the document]
lmorchard: Yup, exactly. Existing image is the URL to the file in storage, while a client-side image pending upload is a [File](https://developer.mozilla.org/en-US/docs/Web/API/File) object. But, I couldn't find anything as obvious as a `File.isFile()` API, so this sort of duck typing was the next best thing I could think of

app/experimenter/experiments/api/v5/types.py
```diff
@@ -103,7 +104,20 @@ class Meta:
...
 
 
+class NimbusBranchScreenshotType(DjangoObjectType):
+    image = graphene.String()
+
+    class Meta:
+        model = NimbusBranchScreenshot
+
+    def resolve_image(root, info):
+        if root.image and root.image.name:
```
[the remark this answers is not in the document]
lmorchard: Pretty much like most other fields, the image is optional in the serializer for saving but required for review. So, it's possible for the screenshot record to have been created but the image to not yet have been uploaded while the author is still working on the experiment.

lmorchard:
> One thing I did notice was that I was able to click "Add screenshot" a bunch and save the form without selecting images - should totally empty screenshot sets be stripped?

They should be okay to save and not get stripped, but "Missing details in: Branches" review feedback should appear and disallow a launch request. Then the empty fields should report "This field may not be blank." when the errors are shown

lmorchard:
@jodyheavener Thanks for taking a look at this beastie!

app/experimenter/experiments/api/v5/types.py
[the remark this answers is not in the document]
jodyheavener: Gotcha!

lmorchard:
Hot diggity, that integration test revealed a real issue that I managed to fix.

Going to open this for review, but actually merging it is blocked by resolution of issue #6424

app/experimenter/experiments/models/nimbus.py
```diff
@@ -554,6 +550,18 @@ class Meta:
...
+        cloned = copy.copy(self)
+        cloned.id = None
+        cloned.experiment = to_experiment
+        cloned.save()
+
+        screenshots = self.screenshots.all()
+        for screenshot in screenshots:
+            screenshot.clone(cloned)
+
+        return cloned
```
jaredlockhart, line 563: Ah yep good thinking to break this out 👍

app/experimenter/experiments/models/nimbus.py
```diff
@@ -574,10 +582,28 @@ class NimbusBranchScreenshot(models.Model):
...
+    def clone(self, to_branch):
+        image_copy = ContentFile(self.image.read())
+        image_copy.name = self.image.name
+
+        cloned = copy.copy(self)
+        cloned.id = None
+        cloned.branch = to_branch
+        cloned.image = image_copy
+        cloned.save()
+        return cloned
```
jaredlockhart, line 605: 👍 👍 👍 👍

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.tsx
```diff
@@ -235,6 +240,47 @@ export const FormBranch = ({
...
+                {}) as FormScreenshotProps["errors"],
+              //@ts-ignore react-hook-form types seem broken for nested fields
+              touched: (touched?.screenshots?.[idx] ||
+                {}) as FormScreenshotProps["touched"],
+              //@ts-ignore react-hook-form types seem broken for nested fields
+              reviewErrors: (reviewErrors?.screenshots?.[idx] ||
+                {}) as FormScreenshotProps["reviewErrors"],
+            }}
+          />
+        ))}
```
jaredlockhart, line 282: Can we flow these into separate rows for each experiment instead of packing horizontally? 🙏

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormScreenshot/index.tsx
```diff
@@ -0,0 +1,170 @@
...
```
jaredlockhart, line 156: How come we have to handle both the string and object case?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
```diff
@@ -116,6 +120,21 @@ export const FormBranches = ({
...
+  const handleAddScreenshot = (branchIdx: number) => () => {
+    commitFormData();
+    dispatch({ type: "addScreenshotToBranch", branchIdx });
+  };
+
+  const handleRemoveScreenshot =
+    (branchIdx: number) => (screenshotIdx: number) => {
+      commitFormData();
+      dispatch({
+        type: "removeScreenshotFromBranch",
```
jaredlockhart, line 132: Do we have constants somewhere for these action names?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
```diff
@@ -250,17 +257,117 @@ function commitFormData(
...
+  });
+  return {
+    ...branch,
+    ...formData,
+    screenshots,
+  };
+}
+
+type AddScreenshotToBranchAction = {
+  type: "addScreenshotToBranch";
```
jaredlockhart, line 302: Constant?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
```diff
@@ -250,17 +257,117 @@ function commitFormData(
...
-    }));
+  return {
+    ...state,
+    referenceBranch,
+    treatmentBranches,
+  };
+}
+
+type RemoveScreenshotFromBranchAction = {
+  type: "removeScreenshotFromBranch";
```
jaredlockhart, line 339: Constant?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts
```diff
@@ -250,17 +257,117 @@ function commitFormData(
...
+      treatmentBranches,
+      branchIdx,
+      (selectedBranch) => ({
+        ...selectedBranch,
+        screenshots: withoutScreenshot(
+          selectedBranch.screenshots,
+          screenshotIdx,
+        ),
+      }),
+    );
```
jaredlockhart, line 370: It would be nice if there was some way to treat all branches holistically for cases like this where you don't really care if it's the reference or treatment branches.  Nothing we should address here but maybe there's a pattern we can find later so we don't always have to treat reference as a special case?

jaredlockhart (changes requested):
Awesome @lmorchard great job, everything looks and works great.  Only big thing that stands out to me is flowing them vertically instead of packed horizontally.  Maybe some action name constants but that's just a suggestion.

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
lmorchard: No variable constants defined, but they are constants within [a type defining all valid actions](https://github.com/mozilla/experimenter/blob/4deaedc748fad269e01ddbbd925142811c7969b6/app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/actions.ts#L43-L52). So, if any of the strings are typo'd, linting will fail.

jaredlockhart: Oh interesting truuuueeee.... Okay cool that's good enough 👍

lmorchard: Yeah, it's a little weird, but seems to be a convention to reduce some boilerplate and lean into TypeScript

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormScreenshot/index.tsx
lmorchard: The string case represents an existing screenshot on the server, where an upload doesn't happen at time of submission and we should set the GQL property to null so as not to replace or delete it.

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
jaredlockhart: Yeah interesting I always think of it as `const SOMETHING = "whatever"` and then just refering to the variable so you never retype the string anywhere, but I have also done it this way with declaring the string literal right in the type declaration.  TypeScripts structural typing is super fancy and blows my mind every time I use it.  VSCode will probably autocomplete that string for you even.

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.tsx
lmorchard: Tweaked the form to go vertically, though the screenshots take up a lot more scrolling now. That was part of why I tried horizontally. But, it does more closely match the eventual appearance on the Summary page, so that's probably a better situation

lmorchard:
Tweaked the screenshot form layout and rebased against main

jaredlockhart:
Oh sorry I wasn't clear enough, or I commented on the wrong section of code!  I meant on the summary page:


Screen Shot 2021-09-29 at 7 22 01 PM


Also I just noticed two things:

1) We should definitely move the add screenshot button to the bottom becuase if the bottom is off the viewport and you click it it looks like nothing happens

2) I just noticed if you add multiple empty screenshot forms and save them, they'll save as empty screenshots, set the alt text to `c` for some reason, and then they show up on the summary page

If we don't want to block this we can land it and file them as followups, but if this is still blocked on ops then we might as well just fix them here?

lmorchard:
Rebased against main; added `display: block` style to screenshot figured on overview so they don't flow in-line; move "add screenshot" button to the end of the list of screenshots in branch form

lmorchard:
Oh, and also re-created the DB migration since another has landed ahead of this PR on main

PR #6478, opened 2021-09-27
fix #6475 chore(nimbus): move targeting config expression to front of whole targeting expression

### LINKED ISSUE #6475: Targeting errors aren't detected by integration tests
I just played around with the new targeting validation integration tests and noticed that if you purposefully malform one of the targeting expressions, it doesn't fail the test, which is what those tests are trying to accomplish.  When I dug in a bit I realized it's not becuase the tests aren't doing what they're supposed to, ie running the JEXL in Firefox and checking for errors/undefined, it's becuase the JEXL evaluator will short circuit boolean expressions, ie

```js
eval("false && someUndefined") -> false
eval("true && someUndefined") -> undefined
```

and since we stick the targeting config expression after all the version/channel checks, those can fail first and never hit the targeting config expression, so we have to move the targeting config expression to the front of the expression chain.

jaredlockhart:
Because

* For our new integration tests to properly check the validity of targeting config jexl expressions, the targeting config expression must be the first expression otherwise the jexl evaluator might short circuit before it gets to the targeting config expression

This commit

* Moves the targeting config expression up to the front of whole expression
* This exposes that the homepage targeting config is no longer valid in the current Firefox, so it is removed

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -186,19 +186,6 @@ TARGETING_MOBILE_FIRST_RUN = NimbusTargetingConfig(
     application_choice_names=(Application.FENIX.name, Application.IOS.name),
 )
 
-TARGETING_HOMEPAGE_GOOGLE = NimbusTargetingConfig(
-    name="Homepage set to google.com",
-    slug="homepage_google_dot_com",
-    description="Users with their Homepage set to google.com",
-    targeting=(
-        "!homePageSettings.isDefault && "
-        "homePageSettings.isCustomUrl && "
-        "homePageSettings.urls[.host == 'google.com']|length > 0"
-    ),
-    desktop_telemetry="",
-    application_choice_names=(Application.DESKTOP.name,),
-)
-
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -186,19 +186,6 @@ class NimbusTargetingConfig:
...
 )
 
-TARGETING_HOMEPAGE_GOOGLE = NimbusTargetingConfig(
-    name="Homepage set to google.com",
-    slug="homepage_google_dot_com",
-    description="Users with their Homepage set to google.com",
-    targeting=(
-        "!homePageSettings.isDefault && "
-        "homePageSettings.isCustomUrl && "
-        "homePageSettings.urls[.host == 'google.com']|length > 0"
```
jaredlockhart, line 196: @ncloudioj `homePageSettings` evaluates to `undefined` in Firefox 93 so it will have to be changed or removed for our integration tests to pass.

jaredlockhart:
Ah this is the wrong solution.  Closing and updating the ticket.

PR #6491, opened 2021-09-29
Change jexl integration test script.

jrbenny35:
This is now causing some failures but it matches a bit better to what Firefox does.

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -247,9 +247,6 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
             # TODO: Remove opt-out after Firefox 84 is the earliest supported Desktop
             expressions.append("'app.shield.optoutstudies.enabled'|preferenceValue")
 
-        if self.targeting_config and self.targeting_config.targeting:
-            expressions.append(self.targeting_config.targeting)
-
         if self.locales.count():
             locales = [locale.code for locale in self.locales.all().order_by("code")]
             expressions.append(f"locale in {locales}")
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -247,6 +247,9 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
             # TODO: Remove opt-out after Firefox 84 is the earliest supported Desktop
             expressions.append("'app.shield.optoutstudies.enabled'|preferenceValue")
 
+        if self.targeting_config and self.targeting_config.targeting:
+            expressions.append(self.targeting_config.targeting)
+
         if self.locales.count():
             locales = [locale.code for locale in self.locales.all().order_by("code")]
             expressions.append(f"locale in {locales}")
@@ -323,7 +326,7 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
     def computed_enrollment_days(self):
         paused_change = (
             self.changes.all()
-            .filter(experiment_data__is_paused=True)
+            .filter(NimbusChangeLog.Filters.IS_APPROVED_PAUSE)
             .order_by("changed_on")
             .first()
         )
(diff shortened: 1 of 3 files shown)


jaredlockhart (changes requested):
Oh this misses the other thing I mentioned about reordering the generation of the targeting strings here: https://github.com/mozilla/experimenter/issues/6475#issuecomment-928055547

app/experimenter/experiments/models/nimbus.py
```diff
@@ -323,7 +323,7 @@ def proposed_end_date(self):
     def computed_enrollment_days(self):
         paused_change = (
             self.changes.all()
-            .filter(NimbusChangeLog.Filters.IS_APPROVED_PAUSE)
```
jaredlockhart, line 326: Oh that all snuck in from another PR, just revert those lines

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -542,34 +539,6 @@ def test_computed_enrollment_days_returns_fallback(self):
...
-            new_status=NimbusExperiment.Status.LIVE,
-        ).update(changed_on=datetime.datetime.now() - datetime.timedelta(days=5))
-        experiment.changes.filter(experiment_data__is_paused=True).update(
-            changed_on=datetime.datetime.now()
-        )
-        self.assertEqual(
-            experiment.computed_enrollment_days,
-            expected_days,
-        )
-
```
jaredlockhart, line 572: And all this stuff

jrbenny35: Where the heck did this come from

jaredlockhart: Just landed in main after I closed my PR

jaredlockhart: If you're using VSCode it has a thing where you can revert individual blocks of changes within files

jrbenny35: Oh I get it.

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -490,25 +490,22 @@ def test_should_end_enrollment_returns_True_after_proposed_enrollment_end_date(s
...
 
         experiment.changes.filter(experiment_data__is_paused=True).update(
             changed_on=datetime.datetime.now()
         )
 
         self.assertEqual(
             experiment.computed_enrollment_days,
-            expected_days,
+            3,
         )
```
jaredlockhart, line 509: This needs to get reverted too

jaredlockhart:
hey @piatra  in the ExperimentManager jsm, there's this bit: https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/lib/ExperimentManager.jsm#77-82

but when i try it locally like this

https://github.com/mozilla/experimenter/blob/62136e96ab1b6d64d3cf636c549c981df01569e7/app/tests/integration/nimbus/utils/filter_expression.js#L25-L32

i get 

Screen Shot 2021-09-30 at 4 32 13 PM

which is used in the first run targeting jexls like this:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants/nimbus.py#L143

so I think this is causing all the first run targeting tests to fail atm.  

Is there something we're not setting up correctly in the test?

piatra:
yeah so you're doing this:
```js
const experimentManager = new ExperimentManager._ExperimentManager() 
```
This creates a new uninitialized experiment manager. `activeExperiments` will never resolve because it needs the experiment store and a whole bunch of setup to be useful.
You need a reference to the actual `ExperimentManager` in use (that should be `ExperimentManager.ExperimentManager` based on how the imports look/work in that script)
Alternatively maybe give another try to using the `evaluateJexl` method now that I've changed the return type (not sure if that was the only problem preventing this option)

jaredlockhart:
Aha! Thank you @piatra let me try your new modified `evaluateJexl`

jaredlockhart:
AWESOME this works!!

```js
-    const experimentManager = new ExperimentManager._ExperimentManager()
-
     const context = TargetingContext.TargetingContext.combineContexts(
         _experiment,
-        experimentManager.createTargetingContext(),
+        ExperimentManager.ExperimentManager.createTargetingContext(),
```

jaredlockhart (approved):
YESSSSSS 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #6493, opened 2021-09-29
fix #6492 chore(nimbus): delete media files before loading new dummy experiments

### LINKED ISSUE #6492: Clear media folder out during make refresh
I just noticed that my local `/media` folder has thousands of folders/images from test runs.  We should switch the unit tests to only use the in memory storage provider so nothing persists between test runs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1798)

lmorchard:
Because:

* `make refresh` generates new media files (i.e. branch screenshots)
  when loading dummy experiments, but never deletes them

This commit:

* deletes media files before loading dummy experiments in `make refresh`
  to avoid accumulating old orphaned files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ CHECK_DOCS = python manage.py generate_docs --check=true
 GENERATE_DOCS = python manage.py generate_docs
 LOAD_COUNTRIES = python manage.py loaddata ./experimenter/base/fixtures/countries.json
 LOAD_LOCALES = python manage.py loaddata ./experimenter/base/fixtures/locales.json
+DELETE_MEDIA_FILES = rm -rvf ./experimenter/media/*
 LOAD_DUMMY_EXPERIMENTS = [[ -z $$SKIP_DUMMY ]] && python manage.py load_dummy_experiments || echo "skipping dummy experiments"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -157,7 +158,7 @@ bash: build_dev
 	$(COMPOSE) run app bash
 
 refresh: kill build_dev
-	$(COMPOSE) run -e SKIP_DUMMY=$$SKIP_DUMMY app bash -c '$(WAIT_FOR_DB) $(PYTHON_MIGRATE)&&$(LOAD_LOCALES)&&$(LOAD_COUNTRIES)&&$(LOAD_DUMMY_EXPERIMENTS)'
+	$(COMPOSE) run -e SKIP_DUMMY=$$SKIP_DUMMY app bash -c '$(WAIT_FOR_DB) $(PYTHON_MIGRATE)&&$(LOAD_LOCALES)&&$(LOAD_COUNTRIES)&&$(DELETE_MEDIA_FILES)&&$(LOAD_DUMMY_EXPERIMENTS)'
```
lmorchard, line 161: Had thought about sticking this in the `kill` task, but thought it might be better if it were more specifically handled where the files are generated

jaredlockhart (approved):
Yep perfect ty 👍

PR #6497, opened 2021-09-30
fix #6495 chore(nimbus): allow feature slugs to be shared by aplications

### LINKED ISSUE #6495: Allow multiple applications to have the same feature config slug
There may be cases where a similar feature is implemented in multiple applications, ie in multiple mobile clients like iOS and Android, and in that case it would be helpful to enforce uniqueness within an application but not globally.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1800)

jaredlockhart:
Because

* Some applications may have the same feature config slug, ie same feature across ios and fenix

This commit

* Changes feature config slug from globally unique to unique per application


lmorchard (approved):
:+1: LGTM!

PR #6498, opened 2021-09-30
Adds new user targeting for MR2

tarikeshaq:
Changes the currently unused and misleading "First Run" targeting to be a "New User" targeting with the definition from https://mozilla-hub.atlassian.net/browse/SDK-373

The work to add the jexl values in the client is in: https://github.com/mozilla/application-services/pull/4491 and https://github.com/mozilla/application-services/pull/4490

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -175,13 +175,11 @@ TARGETING_FIRST_RUN_WINDOWS_1903_NEWER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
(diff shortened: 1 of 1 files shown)


lmorchard:
This looks good to me insofar as it passes JEXL validation, but I recall [this comment on an earlier similar PR](https://github.com/mozilla/experimenter/pull/6441#issuecomment-921171794):

> One thing that I'm curious about is whether modifying targeting configurations in place will inadvertently or adversely other experiments which may also be using them. If only one body of experiments uses them then it's probably fine but I'm curious if we might in the future see cases where one team modifies a targeting configuration and accidentally disrupts the experiments of another.

Looks like this doesn't add a new target but replaces an existing one. Is that because the existing one was never used?

lmorchard (approved):

PR #6499, opened 2021-10-01
Seperates recently updated users from new user targeting

tarikeshaq:
Updating the targeting after chatting with @jhugman 

The idea is that there will be multiple experiments, some targeting new users, and others targeting users who recently updated - but not both in the same experiment.

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -178,11 +178,26 @@ TARGETING_FIRST_RUN_WINDOWS_1903_NEWER = NimbusTargetingConfig(
 TARGETING_MOBILE_NEW_USER = NimbusTargetingConfig(
     name="New Users on Mobile",
     slug="mobile_new_users",
-    description=("New users on mobile"),
-    targeting=("is_already_enrolled || days_since_install < 7 || days_since_update < 7"),
+    description=("New users on mobile who installed the app less than a week ago"),
+    targeting=("is_already_enrolled || days_since_install < 7"),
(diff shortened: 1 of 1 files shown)


jaredlockhart:
@tarikeshaq You gotta run `make code_format` and the tests will clear up 👍

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -178,12 +178,26 @@ class NimbusTargetingConfig:
...
+    targeting=("is_already_enrolled || days_since_install < 7"),
     desktop_telemetry="",
     application_choice_names=(Application.FENIX.name, Application.IOS.name),
 )
 
+TARGETING_MOBILE_RECENTLY_UPDATED_USER = NimbusTargetingConfig(
+    name="Recently Updated Users",
+    slug="mobile_recently_updated_users",
+    description=(
+        "Users who updated their app within the last week. This excludes users who are new users"
```
jaredlockhart, line 191: Actually you'll probably have to brake this up by hand since `black` won't do that automatically.  Just change it to 

```py
    description=(
        "Users who updated their app within the last week. "
        "This excludes users who are new users"
    )
```

tarikeshaq: Awesome thank you! yeah `code_format` didn't fix this one 😄

tarikeshaq:
Got feedback from @jhugman on slack and addressed it! @lmorchard do have a minute to approve and merge this 🙏

lmorchard (approved):
:+1:

PR #6543, opened 2021-10-12
fix #6542 chore(nimbus): store media files in docker volume

### LINKED ISSUE #6542: Store media in docker volume for local dev
Rather than storing media files on the host filesystem and explicitly `rm -Rf`ing them on `make refresh` we can just put them inside a volume and that will hide them from the hostfs and automatically reset them in `make refresh`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1819)

jaredlockhart:
Because

* Running rm -Rf as part of your build process should be concerning for anyone

This commit

* Moves media files into a docker volume and never exposes them to the host filesystem, and they will get reset automatically in make refresh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,7 +81,6 @@ target/
 !.vscode/extensions.json
 
 # Hosted assets
-app/experimenter/media/
 app/experimenter/served/
 app/experimenter/legacy-ui/assets/
 app/experimenter/reporting/reporting-ui/assets/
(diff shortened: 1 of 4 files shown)


lmorchard:
Hmm, I see a nimbus integration test failure - maybe an intermittent? Gave it a kick to re-run

PR #6549, opened 2021-10-13
fixes #6335 fix(nimbus): remove duplicate tooltip

### LINKED ISSUE #6335: Archive experiment tooltip has a curious right border on it
https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/this-tests-experiment-creation-on-1908

This_tests_experiment_creation_on_19_08_–_Summary___Experimenter

This is kind of funny. I can see it more prominently on one of my monitors, so I think it might be a brightness issue. Anyways, there is a weird 15px slightly greyish border on the right side of the tooltip. But the curious part is that on my overly bright monitor seems to show some overlapping text. In fact, here's a screenshot of the zoomed screenshot:
Untitled

So I _think_ the tooltip is being applied twice with a slightly different offset, which explains the dual text and slight gray background border

anthonytran7757:
Because

* the tooltip is rendered twice

This commit

* moves the ReactToolTip component into the Info component

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/SidebarActions/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/SidebarActions/index.tsx
@@ -78,16 +78,15 @@ export const SidebarActions = ({
           <TrashIcon className="sidebar-icon" />
           {experiment.isArchived ? "Unarchive" : "Archive"}
           {archiveDisabled && (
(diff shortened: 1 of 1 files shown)


anthonytran7757:
Here is a screenshot of the screenshot now!
screenshot

PR #6620, opened 2021-11-03
fix #6249 fix #6188 refactor(nimbus): move most links into the sidebar

### LINKED ISSUE #6188: (FE) View as stakeholder/engineer on Summary page
There's a desire to have a slimmer version of the Summary page for users such as stakeholders to be able to skim the most relevant pieces to them on the summary page. 

Let’s add a second summary page that only includes the hypothesis, application, experiment owner, and additional links. There should likely be a new component to render since they will probably look nicer outside of a table which such little information on the page.  Let’s have two navigation links under Summary for the overview and detail pages.

If there’s any information that should be shared between the overview and detail pages let’s pull it up into a shared header maybe?

Pull links out into a persistent area for links

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1620)

### LINKED ISSUE #6249: Remove all scrollspy stuff
In #6031 a bug was reported where if the sidebar nav was too long, users couldn't see the bottom of the list.

We solved this by making the sidebar unfixed. However, now the sidebar seems significantly less useful as 1) you can only click on sections when you're scrolled to the very top part of the page and 2) you can't see where you are when you're scrolling down the page.

We can either add an independent scrollbar for the sidebar and leave it at that, or add the 2nd scrollbar and try adjusting font-size/margins to accommodate more cases with a lot of results in the sidebar, or add the 2nd scrollbar and have it scroll up/down when the user is at that part of the page. Or, if we leave this as-is, we should just remove all the scrollspy pieces/package.

Personally I'm thinking we try the 1st or 2nd solution, the 3rd is ideal but seems like a much higher level of effort (maybe we can look at it in a separate maintenance task?).

Willing to hear pushback of course, but the nav just seems way less helpful if we leave the sidebar as-is. There will also be new "actions" to consider that will live under this section, however.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1652)

lmorchard:
Because:

* There's a desire to have a slimmer version of the Summary page

This commit:

* moves many links scattered inline throughout summary pages into the a
  new "Links" section in the sidebar

* renames the old links section to "Help"

* removes ScrollSpy from the results page sidebar


app/experimenter/nimbus-ui/package.json
```diff
@@ -51,7 +51,6 @@
     "react-json-pretty": "^2.2.0",
     "react-markdown": "^5.0.3",
     "react-scripts": "3.4.0",
-    "react-scrollspy": "^3.4.3",
```
jaredlockhart, line 54: Oh good, finally getting rid of this.  But also try not to fold too many things into a single pr you never know if you'll have to revert it and then lose multiple changes 🙏

jaredlockhart (changes requested):
Awesome, this'll make things so much cleaner.  Also earns a cupcake 🧁 

Couple little tweaks:

1) The Preview Recipe JSON doesn't seem to jump to the anchor tag if it navigates in from another page?

2) Is there some way to make the external link icon smaller/lighter?
Screen Shot 2021-11-03 at 5 37 58 PM

app/experimenter/nimbus-ui/package.json
lmorchard: Well, this folded into [this bit here](https://github.com/mozilla/experimenter/pull/6620/files#diff-89aaf3f025dcf630551a25637e625f6ab63a8f9b1967abed0a1443de2ff7c8d1L159-L225) where I was already removing a section and removing ScrollSpy simplified the whole thing.

lmorchard:
>  The Preview Recipe JSON doesn't seem to jump to the anchor tag if it navigates in from another page?

Hmm, looks like the problem is that the anchor doesn't actually exist when navigating from another page. We refresh the data and toss up a loading spinner, so that anchor in the DOM hasn't appeared yet. Might need to do something funky to make that work.

>    Is there some way to make the external link icon smaller/lighter?

Took a shot at reducing by a couple pixels and opacity in CSS

lmorchard:
Alright, since we'll probably need this again, I cobbled together a quick `useScrollToLocationHash` hook to ensure the page scrolls down to the hash in the window location after page load.

app/experimenter/nimbus-ui/package.json
jaredlockhart: Yeah if it makes sense together then is good 👍

jaredlockhart (approved):
Oh yep good, thnx for lightening the icons.  Later we can go in and put in specific ones for each link type.  And yeah the scroll thing works perfectly 🎉

PR #6631, opened 2021-11-05
fix #6623 chore(nimbus): revert remove feature enabled

### LINKED ISSUE #6623: Revert 'remove feature enabled'
We thought that the 'feature enabled' flag was deprecated in desktop clients so we removed it, however we have since learned that it is still being used by some components in desktop and so we'll have to revert it for now until those components can migrate away from it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1854)

jaredlockhart:
Because

* Some desktop client functionality still depends on the feature_enabled flag

This commit

* Restores the feature_enabled flag and all ui/api/tests


app/experimenter/experiments/migrations/0195_nimbusbranch_feature_enabled.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("experiments", "0194_auto_20211102_1749"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="nimbusbranch",
+            name="feature_enabled",
+            field=models.BooleanField(default=True),
```
lmorchard, line 16: In the components that still use this flag, is an empty feature config (i.e. `{}`) equivalent to a non-enabled feature? Or do we need to re-introduce explicit True / False values with some logic in this migration?

lmorchard (approved):
LGTM, code-wise. But a question about restoring the field in the migration - i.e. is all True just fine for practically solving the issue in desktop components, or should we restore some True / False values based on some logic?

app/experimenter/experiments/migrations/0195_nimbusbranch_feature_enabled.py
jaredlockhart: No, empty value isn't the same as enabled, many experiments have empty in both branches with enabled off/on in each.  Actually I was originally gonna just set everything back to `True` as the default and update any if anyone asked in the admin, but then I remembered we can recover that data from the `published_dto` so I added that to the migration.

lmorchard (approved):
Migration & test look keen :+1:

Issue #6649, opened 2021-11-08
Buckets should not be reallocated during end enrollment


jaredlockhart:
I just noticed an experiment ending enrolment and its buckets appear to be reallocated.  Allocation should only happen at launch, not any time after, so we may be inadvertently reallocating buckets during republishing?

![image](https://user-images.githubusercontent.com/119884/140819536-80e6497c-fa7a-451d-8a89-e90b8daa7db5.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1870)

PR #6657, opened 2021-11-09
fix #6628 feat(nimbus): experiments flagged as rollouts should use a separate bucket namespace

### LINKED ISSUE #6628: Namespace buckets by exp/rol per is_rollout value
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1859)

lmorchard:
Because:

* we want rollouts to live in separate buckets than general experiments

This commit:

* adds "-rollout" to the bucket namespace when is_rollout=True

* leaves non-rollout experiment namespace formulation alone to constrain
  the impact of this change

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -410,6 +410,9 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
         if self.channel:
             keys.append(self.channel)
 
+        if self.is_rollout:
+            keys.append("rollout")
+
         return "-".join(keys)
 
     def allocate_bucket_range(self):
--- a/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
+++ b/app/experimenter/experiments/tests/test_models/test_models_nimbus.py
@@ -823,6 +823,26 @@ class TestNimbusExperiment(TestCase):
         self.assertEqual(NimbusBucketRange.objects.count(), 2)
         self.assertEqual(NimbusIsolationGroup.objects.count(), 1)
 
+    def test_bucket_namespace_changes_for_rollout(self):
+        feature = NimbusFeatureConfigFactory(slug="feature")
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            application=NimbusExperiment.Application.DESKTOP,
+            channel=NimbusExperiment.Channel.RELEASE,
+            feature_config=feature,
+            population_percent=Decimal("50.0"),
+        )
+        original_namespace = experiment.bucket_namespace
+
+        experiment.is_rollout = True
+        experiment.save()
+
+        self.assertNotEqual(original_namespace, experiment.bucket_namespace)
+        self.assertEqual(
+            experiment.bucket_namespace,
+            "firefox-desktop-feature-release-rollout",
+        )
+
     def test_proposed_enrollment_end_date_without_start_date_is_None(self):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.CREATED,


lmorchard:
Expected a little more complexity here, but I *think* this is sufficient?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -410,6 +410,9 @@ def bucket_namespace(self):
         if self.channel:
             keys.append(self.channel)
 
+        if self.is_rollout:
+            keys.append("rollout")
```
jaredlockhart, line 414: Yes totally but I was thinking we should also do the other side and add `experiment` if it's not a rollout so you can quickly tell just by looking at it what it is.

jaredlockhart (changes requested):

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Oh and also I just realized we should wait on landing this until I get the fix for #6649 up because otherwise we'll have the same problem where the namespace will change midflight.

lmorchard: I wanted to avoid that in order to not change all the namespaces everywhere

jaredlockhart: Yeah but changing the buckets for inflight is fully a bug so we should just fix it first and then land this after with both keys instead of doing it in two parts, I'm wrapping up the fix now.

lmorchard: I don't mean inflight so much as the shared bucket namespace between one live experiment and one launched after this. Before the change, they would have taken up buckets in the same namespace. After, they end up in different namespaces even though everything else is equal

lmorchard:
Easy enough to switch to adding "experiment-" or "rollout-" to the namespace.

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Ooohhhh yes.  You are right that is a very valid point.   Welllllll....we could either leave it out and either do some kind of migration and/or wait for some lull later, or just make the cutover now and deal with a bit of overlap until all the old experiments are done.  What do you think?

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -823,6 +823,22 @@ def test_allocate_buckets_deletes_buckets_preserves_occupied_isolation_group(sel
...
+            channel=NimbusExperiment.Channel.RELEASE,
+            feature_config=feature,
+            population_percent=Decimal("50.0"),
+        )
+        original_namespace = experiment.bucket_namespace
+
+        experiment.is_rollout = True
+        experiment.save()
+
+        self.assertNotEqual(original_namespace, experiment.bucket_namespace)
```
jaredlockhart, line 840: Oh instead of doing this I would just do it like above in 

https://github.com/mozilla/experimenter/pull/6657/files#diff-b7f8cfcb629fab7e38ae45d46aa600746219a2fe6e73f6e8b98e3a95f21da0ebR748-R751

and explicitly assert the actual content of the namespace that includes the `rollout` key

jaredlockhart (changes requested):

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Yeah nevermind what I said your original approach was correct, it's better to be as non disruptive as possible.  If we want to add `experiment` as a key we'll find a way to do it later that doesn't disrupt users workflows.

lmorchard:
Alright, tweaked back to the original impl of just tacking on "-rollout" for rollouts. Also added an explicit string equality check to the test

jaredlockhart (approved):
Wonderful good, sorry for the big roundabout XD but very good points raised and discussed.  We should think about how to handle this for cases where we want to change namespacing in the future 👍

PR #6658, opened 2021-11-09
fix #6649 bug(nimbus): only generate bucket allocation at launch time

### LINKED ISSUE #6649: Buckets should not be reallocated during end enrollment

jaredlockhart:
Because

* We had only specified the review state for when to allocate buckets, buckets were being erroneously regenerated at pause time
* We recently modified the namespacing keys, and this resulted in buckets being regenerated for a new namespace at pause time

This commit

* Modifies the SHOULD_ALLOCATE_BUCKETS filter to explicitly include the draft status
* Adds a data migration to revert bucket allocations that may have been erroneously regenerated at pause time back to the values stored in the published_dto


lmorchard (approved):
Oof 💯 for that tricky migration

jaredlockhart:
> Oof 💯 for that tricky migration

I know it was a bit of a headache 😅

PR #6686, opened 2021-11-19
fixes mozilla/jetstream#940 adjust date range for Looker link

scholtzan:
Fixes https://github.com/mozilla/jetstream/issues/940

Data shown in the Looker dashboard is in UTC timezone. The Nimbus Experimenter Looker links don't account for timestamps being in UTC, so it looks like dashboards are missing data when clicking on them. This fixes this issue


jaredlockhart (approved):
Thnx @scholtzan !

PR #6707, opened 2021-11-24
chore(nimbus): split v5 api serializer tests

jaredlockhart:
Because

* There are too many API V5 serializer tests in one file
* Makes it hard to navigate
* Turns out we also had a bunch of duplicated tests and nobody noticed

This commit

* Splits that big file up into separate test files


jaredlockhart:
Here's some other files that have more than 1k lines in case we wanna do something similar in the future:

```
    1230 ./experiments/constants/legacy.py
    1089 ./experiments/tests/test_forms.py
    1629 ./experiments/tests/test_models/test_models_nimbus.py
    1807 ./experiments/tests/test_models/test_models_legacy.py
    2077 ./experiments/tests/api/v2/test_serializers.py
    1145 ./experiments/tests/api/v5/test_queries.py
    1093 ./experiments/tests/api/v5/test_mutations.py
    1326 ./experiments/models/legacy.py
```

lmorchard (approved):
Good stuff!

PR #6708, opened 2021-11-25
fix #6575: add option to warn on feature value schema validation failure

### LINKED ISSUE #6575: Add support for soft feature schema validation
Some experiments will need to launch with feature values that may not validate with the stored schema.  To support those cases, we should add a checkbox to the branch page under the feature select which allows people to 'Warn only on feature schema validation failure'.  

On the review page, if the experiment is set to warn only, we should still display the schema validation errors as warnings but not block launching.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1834)

lmorchard:
Because:

* Some experiments will need to launch with feature values that may not validate with the stored schema.

This commit:

* Adds a checkbox to the branches edit page - 'Warn only on feature
  schema validation failure'

* Adds warnOnFeatureValueSchemaInvalid to the branch page form reducer

* Adds warn_on_feature_value_schema_invalid to experimend model and plumbs it
  through GQL for queries and mutations

* Adds support to forms in the UI to display warnings along with errors


lmorchard:
This change snowballed a little bit, mainly due to needing to work in new plumbing to handle warnings vs errors. Initially tried just kind of hacking warnings into the list of errors, but that kind of complicated other bits of review UI. So, this just adds warnings as a new class of review messages.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -151,6 +151,7 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
         "NimbusFeatureConfig",
         blank=True,
     )
+    warn_on_feature_value_schema_invalid = models.BooleanField(default=False)
```
jaredlockhart, line 154: That's a bit long for a field name, something shorter would probably be fine like `warn_schema` or something.

jaredlockhart (changes requested):
Okay yep, this looks right and works right.  That field name is too long.

Also I just realized that.... I'm gonna have to move all this around as soon as it lands because the multifeature V5 API change moves the schema validation out of the top level serializer wayyyy down into the new BranchFeatureValue serializer, but, it's fine, gotta do it anyways.

Thanks @lmorchard all the behaviour and UI is correct 👍

jaredlockhart:
> This change snowballed a little bit, mainly due to needing to work in new plumbing to handle warnings vs errors. Initially tried just kind of hacking warnings into the list of errors, but that kind of complicated other bits of review UI.
> 
> So, this just adds warnings as a new class of review messages. Seems like we _might_ need that more in the future?

Yeah and actually, thinking about it more, squishing the errors and warnings into a single serializer I think might not even be the right approach because then you can't recompose the serializers around.  I *think* another way to do it is to have two separate parallel chains of serializers starting from the top all the way down, one for errors, one for warnings?  

Don't try it here, this is fine for now, better to get it out than spin our wheels on it.  But already I have to move all this around for the multifeature V5 change so maybe I'll give it a whirl in there?

app/experimenter/experiments/models/nimbus.py
lmorchard: I can change this, though FWIW I do tend to favor long descriptive names wherever possible

lmorchard:
Alright, renamed that field to `warn_feature_schema` and hopefully VSCode's cross-file search-and-replace did the needful. Rebased against main and pushed

jaredlockhart (approved):
Awesome looks great thank you, gonna be heck to rebase my pr on it here we go XD

PR #6723, opened 2021-12-01
feat(nimbus): add targeting for Firefox needing pin for infrequent and casual activity level

Mardak:
r?@piatra Wasn't sure about modifying existing "pin capable" which checks for windows build as a proxy for can pin (which is still different from actually needing pin), but I guess we might want to keep that as a separate targeting anyway.

Generates jexl like…
```js
(
  userMonthlyActivity|length >= 7 &&
  userMonthlyActivity|length <= 13 &&
  doesAppNeedPin ||
  experiment.slug in activeExperiments
) &&
browserSettings.update.channel == "release" &&
version|versionCompare('48.!') >= 0 &&
'app.shield.optoutstudies.enabled'|preferenceValue &&
locale in ['ach', 'af', 'en-US', 'sq'] &&
region in ['AF', 'AL', 'AX', 'US']
```


piatra (approved):
Looks good. Good to have the `sticky` in there as well.

jaredlockhart (approved):
LGTM 👍

PR #6727, opened 2021-12-02
chore(nimbus): hardcode last mc version with feature manifest

jaredlockhart:
Because

* We have moved the desktop feature manifest location
* The new path is not yet available
* Containers won't build or start without it

This commit

* Uses the last mc commit with the feature manifest at the old path until we can get the new one to unblock dev


piatra (approved):
LGTM. I’ll see what we can do to improve thisi.

PR #6731, opened 2021-12-06
fix #6728 feat(nimbus): support desktop multifeature in V6 API

### LINKED ISSUE #6728: Update V6 API for Multifeature
Let's adapt the V6 API to emit the new multifeature schema with `features`, with the caveat that desktop clients will need dummy values for `feature` to not break.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1911)

jaredlockhart:
Because

* We'd like to be able to launch multifeature experiments to desktop as soon as possible
* Client implementation of multifeature support is available in Firefox 95
* Schema 1.6.2 supports multifeature

This commit

* Updates mozilla nimbus shared to 1.6.2
* Adds a multifeature serializer specifically for desktop in the V6 API
* Wraps the single and multifeature V6 serializers in a proxy object that will dispatch to the correct serializer based on application and version
* Adds example tests to dilineate between the single and multifeature cases


lmorchard (approved):
:+1:

jaredlockhart:
I just realized I should make sure that in the listview context it uses the right serializer for the right experiment

PR #6752, opened 2021-12-06
fix #6750 chore(nimbus): use eslint-plugin-testing-library to lint js tests

### LINKED ISSUE #6750: Lint for screen.debug in nimbus-ui
Just noticed a stray `screen.debug()` landed in main that makes it look like tests are failing, we should lint against landing those in main.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1924)

jaredlockhart:
Because

* A screen.debug statement snuck into our nimbus-ui tests that made it look like tests were failing
* eslint-plugin-testing-library has a rule to lint for that
* It also has a couple handy rules for enforcing some good patterns in tests that might catch past or future intermittents

This commit

* Adds eslint-plugin-testing-library
* Adds a couple handy rules
* Cleans up a couple small test lint failures it caught

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
@@ -128,7 +128,6 @@ describe("AppLayoutSidebarLaunched", () => {
       expect(
         screen.queryByText(RESULTS_WAITING_FOR_LAUNCH_TEXT),
       ).not.toBeInTheDocument();
-      screen.debug();
     });
 
     it("when analysis is skipped", () => {
--- a/app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
@@ -13,10 +13,10 @@ it("renders text with newlines converted to <br> tags", () => {
 
   We are not ready`;
 
-  const result = render(<RichText {...{ text }} />);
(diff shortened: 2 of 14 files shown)


app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
```diff
@@ -128,7 +128,6 @@ describe("AppLayoutSidebarLaunched", () => {
       expect(
         screen.queryByText(RESULTS_WAITING_FOR_LAUNCH_TEXT),
       ).not.toBeInTheDocument();
-      screen.debug();
```
lmorchard, line 131: D'oh

app/experimenter/nimbus-ui/src/components/RichText/index.test.tsx
```diff
@@ -13,10 +13,10 @@ it("renders text with newlines converted to <br> tags", () => {
 
   We are not ready`;
 
-  const result = render(<RichText {...{ text }} />);
+  const view = render(<RichText {...{ text }} />);
```
lmorchard, line 16: Huh, we used `result` because the type is called `RenderResult`. We could also probably destructure `container` from here instead. Seems like [an arbitrary rule](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/render-result-naming-convention.md), but 🤷‍♂️ I like a robot keeping the code clean

lmorchard (approved):
LGTM, maybe this will help keep the awaits & asyncs in better order too

PR #6759, opened 2021-12-07
fix #6749 feat(nimbus): support multiple features with single feature apis/ui

### LINKED ISSUE #6749: Gracefully support not yet implemented multifeature experiments in current nimbus-ui
Before we fully implement support for multifeature experiments in nimbus-ui, let's make sure that the current UI gracefully supports a multifeature experiment in the database defined through the admin without catastrophically breaking.  It's probably sufficient to just show the first feature and omit the others, as long as people can still load the UI and enact the launch/pause/end workflows and review results.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1923)

jaredlockhart:
Because

* While full multifeature support is being developed, we may want to create a multifeature experiment via the admin and launch it
* Existing V5 API and nimbus-ui expect only a single feature, so we should gracefully degrade rather than break the UI altogether
* Editing can be done via the admin and the V6 serializer already supports the new multifeature schema

This commit
* Updates V5 API to gracefully degrade by picking a single feature (first feature ordered by slug) and displaying it in the UI


lmorchard (approved):

jaredlockhart:
I wanna wait till this lands first so I can make sure I don't regress the feature value form https://github.com/mozilla/experimenter/pull/6794

PR #6760, opened 2021-12-08
fixes #6683 fix(Nimbus): hide edit button on takeaways section of arc…

### LINKED ISSUE #6683: [Enhancement] Consider hiding the Edit button on the Takeaways section as long as an experiment is archived
**[Prerequisites]:**
- Have an experiment in the `Complete` state with `Takeaways` added. (eg https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/cmuresandesktop-nimbus-smoke-111921-copy)

**[Steps to reproduce]:**
1. Navigate to the experiment from prerequisites.
2. Archive the experiment and observe the `Takeaways` section.

**[Proposed result]:**
- The `Edit` button is no longer displayed.

**[Current result]:**
- The `Edit` button is displayed and is actionable.

**[Notes]:**
- As long as an experiment is Archived the Takeaways section can't be updated anyway.

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1891)

anthonytran7757:
…hived completed experiment

Because

* Takeaways section cannot be updated on archived experiments

This commit

* hides the edit takeaways button if the experiment is archived

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageSummary/Takeaways/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageSummary/Takeaways/index.test.tsx
@@ -48,6 +48,7 @@ describe("Takeaways", () => {
     expect(
       screen.queryByTestId("conclusion-recommendation-status"),
     ).not.toBeInTheDocument();
+    expect(screen.queryByTestId("edit-takeaways")).toBeInTheDocument();
     expect(
       screen.queryByTestId("takeaways-summary-rendered"),
     ).not.toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


jaredlockhart (changes requested):
Thank you @anthonytran7757 !  Since this is new conditional logic, I'd expect there to be a new test case checking that it's hidden when archive is True and probably an update to an existing test case to check that it's visible when false.

So these test cases could probably add an assert that the button is visible:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageSummary/Takeaways/index.test.tsx#L44-L68

and then you could copy one, pass `isArchived={false}` into `Subject` and assert that it's not visible.

Give that a shot and let me know here if it gives you trouble 🙏

jaredlockhart (approved):
Perfect g/j @anthonytran7757 thank you for updating tests! 🎉

PR #6764, opened 2021-12-08
fix #6763 bug(nimbus): feature_config must be set before branches are saved

### LINKED ISSUE #6763: GraphQL error when creating an experiment for Firefox Desktop 95+
Steps to reproduce:

1) Create a new experiment for desktop
1) Select a desktop feature on the branches page and fill in feature values for the branches
1) Fill in the rest of the fields and select Firefox 96 for firefox version 
1) Attempt to launch
1) Observe 500 graphql error

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1930)

jaredlockhart:
Because

* We recently added the feature_values field for branches that must foreign key onto a feature_config at save time
* feature_configs must be set on the experiment before the branches are saved
* Unit tests that should have caught this were not setting feature_values and asserting that they're saved

This commit

* Moves saving feature_configs in the NimbusExperimentSerializer to before super().save() is called
* Updates the relevant tests to include feature_values and assert that their values and fks on to feature_config are saved correctly

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
@@ -3,213 +3,211 @@ from django.test import TestCase
 from experimenter.experiments.api.v5.serializers import NimbusExperimentSerializer
 from experimenter.experiments.constants.nimbus import NimbusConstants
 from experimenter.experiments.models import NimbusExperiment
-from experimenter.experiments.tests.factories import (
-    NimbusBranchFactory,
-    NimbusExperimentFactory,
-)
+from experimenter.experiments.tests.factories import NimbusExperimentFactory
 from experimenter.experiments.tests.factories.nimbus import NimbusFeatureConfigFactory
 from experimenter.openidc.tests.factories import UserFactory
 
 
 class TestNimbusExperimentBranchMixin(TestCase):
     maxDiff = None
-    BASIC_JSON_SCHEMA = """\
-    {
-    "$schema": "http://json-schema.org/draft-07/schema#",
-    "description": "Password autocomplete",
-    "type": "object",
-    "properties": {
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
```diff
@@ -55,162 +49,166 @@ def test_serializer_replace_branches(self):
...
+        self.assertEqual(experiment.feature_configs.get(), feature_config)
 
-        for key, val in reference_branch.items():
-            self.assertEqual(getattr(experiment.reference_branch, key), val)
-
-        for branch_data in treatment_branches:
-            branch = experiment.branches.get(name=branch_data["name"])
-            for key, val in branch_data.items():
-                self.assertEqual(getattr(branch, key), val)
+        self.assertEqual(experiment.reference_branch.name, "control")
```
lmorchard, line 54: Hmm, could be nice to source this "control" from `reference_branch["name"]` just to retain some of the DRY-ness of the previous version. But, not a win on the verbosity side of things to change all these literal values to references

lmorchard (approved):
LGTM!

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
jaredlockhart: Yeah I've been waffling back and forth on this recently as I work through tests, whether to reference the value or hard code it in the assertion.  I can kinda convince myself of either but I was trying to lean towards the explicit side here just to see if that makes me feel better about the tests.  Sometimes you can use references in assertions that are wrong on both sides and then miss it and then the test passes even though it's broken and you miss it because the reference looks right.  I dunno, something to think about 🤷‍♀️

Issue #6775, opened 2021-12-13
Need Audience section to set max version of an experiment


data-sync-user:
Current Nimbus experimenter console only allows min version to set , and doesn't have capability to set Max version which block running experiment targeting specific version only

┆Issue is synchronized with this [Jira User Request](https://mozilla-hub.atlassian.net/browse/EXP-1938)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18552/Screen+Shot+2021-12-13+at+9.49.36+AM.png">Screen Shot 2021-12-13 at 9.49.36 AM.png</a>

data-sync-user:
➤ Punam Dahiya commented:

UI for More From Mozilla changes in Fx 96 and we don’t want users upgrading to 96 to stay in the experiment

punamdahiya:
With fix of this issue custom targeting added in #6776 can be removed

data-sync-user:
➤ Punam Dahiya commented:

Tried using max version from experimenter console. It seems there is client validation set that doesn’t allow to user to select same release as both min and max version which is needed to run experiment on a selected release. See attached screenshot Thanks

PR #6789, opened 2021-12-13
Fix #6776 - Add new custom targeting for Fx95 users

### LINKED ISSUE #6776: Add new custom targeting for Fx 95 users only
Till #6775 is resolved, we need custome targeting to target 95 users only

`version|versionCompare('95.!') == 0`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1939)

punamdahiya:


jaredlockhart (approved):
LGTM 👍

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -213,6 +213,13 @@ class NimbusTargetingConfig:
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_FX95_DESKTOP_USERS = NimbusTargetingConfig(
+    name="Desktop Users on Fx95",
+    slug="fx95_desktop_users",
+    description=("Firefox 95 Desktop users"),
+    targeting=("version|versionCompare('95.!') == 0"),
```
Mardak, line 220: This matches exactly the version "95.!" where I believe "!" is used as a "dummy" character ("less than" `0`) to relatively compare against. If we want any 95 (e.g., including "a" and "b" and dot releases), we probably want a range, e.g., compare(95) >= 0 && compare(96) < 0

Mardak (changes requested):
https://searchfox.org/mozilla-central/source/xpcom/base/nsIVersionComparator.idl

app/experimenter/experiments/constants/nimbus.py
punamdahiya: +1, should the range be "version|versionCompare('95') >= 0 && version|versionCompare('96') < 0"

Or with dummy character  "version|versionCompare('95.!’) >= 0 && version|versionCompare('96.!’) < 0" or either works

jaredlockhart: Oh you're right, that doesn't work.  On my local:

```js
await targetingContext.evalWithDefault("version")
"96.0b4"

await targetingContext.evalWithDefault("version|versionCompare('95.!') == 0")
false

await targetingContext.evalWithDefault("version|versionCompare('96.!') == 0")
false

await targetingContext.evalWithDefault("version|versionCompare('97.!') == 0")
false
```

Using the method @Mardak is suggesting:

```js
await targetingContext.evalWithDefault("(version|versionCompare('95.!') >= 0) && (version|versionCompare('96.!') < 0)")
false

await targetingContext.evalWithDefault("(version|versionCompare('96.!') >= 0) && (version|versionCompare('97.!') < 0)")
true

await targetingContext.evalWithDefault("(version|versionCompare('97.!') >= 0) && (version|versionCompare('98.!') < 0)")
false
```

jaredlockhart (changes requested):
Actually I think this doesn't work as intended, see my comment above

app/experimenter/experiments/constants/nimbus.py
punamdahiya: thanks for verifying,  patch updated with feedback

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -213,6 +213,15 @@ class NimbusTargetingConfig:
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_FX95_DESKTOP_USERS = NimbusTargetingConfig(
+    name="Desktop Users on Fx95",
+    slug="fx95_desktop_users",
+    description=("Firefox 95 Desktop users"),
+    targeting=("(version|versionCompare('95.!') >= 0) && (version|versionCompare('96.!') < 0)"),
```
Mardak, line 220: Pretty sure operator precedence will check the comparisons before the AND, but this is fine.

FYI (i.e., don't update patch) If you really wanted to "optimize" 😝  and maximize the mozilla version comparison quirks.. `version|versionCompare('95+') < 0` works for the second part and `version|versionCompare('95pre') >= 0` for the first………

Mardak (approved):
Just making sure, you don't need any other "advanced targeting" for the experiment (i.e., it's "No Targeting")? If not, you'll need to combine the two advanced targeting as only 1 can be selected.

punamdahiya:
+1, yes the experiment was not using any advanced targeting. With this custom targeting available , we can set Version drop down to 'No version' and pick 'TARGETING_FX95_DESKTOP_USERS' from Advanced Targeting

Mardak:
> we can set Version drop down to 'No version'

I suppose technically we could have just exposed this as "max 95 version" and reused the existing minimum version targeting if for some reason we need to target say Fx94 & Fx95 for some reason for another experiment. I'm not familiar with how these one-off advanced targeting should get cleaned up -- especially with this particular targeting is probably less likely to be reused.

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: Yeah I would've thought so too but it didn't work in my local testing without them

jaredlockhart: Also interesting good to know 👍

punamdahiya:
Eventually we should get capability to set Max version from console (#6775). Just left comment in #6775 to remove this custom targeting with its fix

jaredlockhart (approved):

jaredlockhart:
Hrmmm something's gone wrong with the tests that validate the targeting in Firefox.  I'll have to poke at it some more in the morning.

punamdahiya:
hmm why are the tests failing for , where it should skip?
[gw0] [ 87%] FAILED nimbus/test_remote_settings.py::test_check_targeting[IOS-fx95_desktop_users] 
nimbus/test_remote_settings.py::test_check_targeting[KLAR-fx95_desktop_users] 
[gw1] [ 91%] FAILED nimbus/test_remote_settings.py::test_check_targeting[FENIX-fx95_desktop_users] 
nimbus/test_remote_settings.py::test_check_targeting[FOCUS-fx95_desktop_users] 
[gw0] [ 95%] FAILED nimbus/test_remote_settings.py::test_check_targeting[KLAR-fx95_desktop_users] 
[gw1] [100%] FAILED nimbus/test_remote_settings.py::test_check_targeting[FOCUS-fx95_desktop_users]

jaredlockhart:
@punamdahiya yeah it's a weird edge case in our pytest setup, I added a hack to skip those tests (which shouldn't be running in the first place) and filed a followup https://github.com/mozilla/experimenter/issues/6791  The incomparable @jrbenny35 is gonna look into it.

app/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -7,6 +7,7 @@
 from nimbus.pages.experimenter.summary import SummaryPage
 from nimbus.pages.remote_settings.dashboard import Dashboard
 from nimbus.pages.remote_settings.login import Login
+from nimbus.models.base_dataclass import BaseExperimentApplications
```
punamdahiya: @jaredlockhart looks like test failed bcause of import order 

ERROR: /app/tests/integration/nimbus/test_remote_settings.py Imports are incorrectly sorted and/or formatted.

punamdahiya:
Thank you!

PR #6798, opened 2021-12-14
fix #6775 feat(nimbus): add firefox max version field for targeting

### LINKED ISSUE #6775: Need Audience section to set max version of an experiment

lmorchard:
Because:

* experimenter console only allows min version to set and doesn't have
  capability to set Max version which block running experiment targeting
  specific version only

This commit:

* adds firefoxMaxVersion as a field to the audience editing page

* adds firefoxMaxVersion display to overview tables

* adds firefox_max_version to nimbus DB model & admin

* includes maximum version comparison in targeting expression based on
  presence and value of firefox_max_version

* adds firefox_max_version to GQL experiment query & input types


lmorchard:
Tried catching all the spots where min version appears and added max version alongside. This doesn't add filtering for max version on the directory view, but that might be worth a follow-up issue.

app/experimenter/experiments/migrations/0201_nimbusexperiment_firefox_max_version.py
```diff
@@ -0,0 +1,115 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="nimbusexperiment",
+            name="firefox_max_version",
+            field=models.CharField(
+                choices=[
+                    ("", "No Version"),
+                    ("11.!", "Firefox 11"),
+                    ("12.!", "Firefox 12"),
```
lmorchard, line 20: Was tempted to remove the choices here from both min & max version in the DB model to avoid migrations like this in the future when versions change. But, didn't want to rock the boat too much. If it's worth it, I can tweak things to drop the choices though

jaredlockhart: No yeah good thinking leaving it alone.  I don't know how much value there really is in explicitly declaring choices right on models, it auto populates the admin but since we don't use djangos forms/templates in nimbus the value basically stops there.  Maybe in January we can do that as a nice cleanup step.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -257,6 +262,12 @@ def targeting(self):
                         version=self.firefox_min_version
                     )
                 )
+            if self.firefox_max_version:
+                expressions.append(
+                    "version|versionCompare('{version}') <= 0".format(
```
jaredlockhart, line 267: Ah heck this isn't exactly right.  Looking at Normandy for reference, it looks like this is the pattern:

```
((version|versionCompare(\"70.!\") >=0 ) && (version|versionCompare(\"70.*\") < 0))
```

so we'll need to replace the `!` with a `*` and compare `< 0`

So we could either leave the constants as they are with `!` for the min case, explicitly replace it with a `*` here in the max version clause, and call that good enough.  

Or we could take `!` out of the min case and then explicitly add it in the min case and add `*` in the max case.  Of course we have to remember to support cases where neither `!` and `*` are used when a specific version is hard coded.  But this seems like a hairer solution so maybe the first option is better.

jaredlockhart (changes requested):
Oh wow that was fast! Thank you!  Everything looks right except

1) We'll have to change the emitted JEXL
2) Testing locally, the max version clause isn't showing up in the recipe/targeting expression for some reason:
Screen Shot 2021-12-15 at 2 19 50 PM
Screen Shot 2021-12-15 at 2 19 55 PM

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Found the normandy implementation:

https://github.com/mozilla/normandy/blob/master/normandy/recipes/filters.py#L634-L635

app/experimenter/experiments/models/nimbus.py
```diff
@@ -257,6 +262,14 @@ def targeting(self):
                         version=self.firefox_min_version
                     )
                 )
+            if self.firefox_max_version:
+                # HACK: tweak the min version to better match max version pattern
```
jaredlockhart, line 266: 👍

app/experimenter/experiments/models/nimbus.py
lmorchard: Yeah, I'll just go with a string replace of "!" with "*" for now.

lmorchard:
Pushing another update: Looks like I left out a bit in the v5 serializer to accept `firefox_max_version` - added that and a mutation test.

jaredlockhart (approved):
Awesome yes, works right, looks right, thanks for hopping on this so quickly!

Just remembered we also have to add a validation that min version must be less than or equal to max version but that's small.  Could either do it in here or as a followup.  If you wanna do it as a followup just file an extra ticket.

Thanks for the fast turnaround @lmorchard !

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -881,6 +883,22 @@ def _validate_feature_configs(self, data):
 
         return data
 
+    def _validate_versions(self, data):
```
lmorchard, line 886: Went ahead and added the version validation

PR #6806, opened 2021-12-17
fix #6805 feat(nimbus): disable launches should not prevent ending enrolment

### LINKED ISSUE #6805: Launch locking shouldn't prevent ending enrolment or experiment
The launch lock should only prevent new launches, not ending enrolment or experiments.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1950)

jaredlockhart:
Becuase

* We want to be able to disable any new launches without preventing people from ending enrolment or experiments

This commit

* Checks that an experiment is moving from draft to live when using the launch disabled check

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -340,6 +340,15 @@ class NimbusStatusValidationMixin:
                         }
                     )
 
+            if (
+                SiteFlag.objects.value(SiteFlagNameChoices.LAUNCHING_DISABLED)
+                and self.instance.status == NimbusExperiment.Status.DRAFT
+                and data.get("status_next") == NimbusExperiment.Status.LIVE
+            ):
+                raise serializers.ValidationError(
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -340,6 +340,15 @@ def validate(self, data):
                         }
                     )
 
+            if (
+                SiteFlag.objects.value(SiteFlagNameChoices.LAUNCHING_DISABLED)
+                and self.instance.status == NimbusExperiment.Status.DRAFT
```
lmorchard, line 345: Request for launch is also possible from PREVIEW status - probably want to add a check for that status here too. 

This will also allow launch review approval in a case where launch was requested and then launch was disabled just after. That might be more of an edge case?

lmorchard (approved):
Looks good, though I'd say we probably want to add PREVIEW to the status check for blocking launch

PR #6820, opened 2022-01-03
fix #6819 bug(nimbus): support same min max firefox version

### LINKED ISSUE #6819: Unable to create experiment with same min and max firefox version
We recently added firefox max version, but the validation does not allow an experiment with the same min and max version, which it should.

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1957)

jaredlockhart:
Because

* We recently added max firefox version
* We need to support min and max targeting the same version

This commit

* Updates the serializer to allow min and max being the same version


lmorchard (approved):
LGTM! :+1:

And FWIW, it looks like the check failure here is resolved in PR #6817

PR #6829, opened 2022-01-03
fix #6828 chore(project): disable legacy tests and dev services

### LINKED ISSUE #6828: Disable legacy tests and dev servers
Now that legacy is really in low maintenance mode, and hasn't required any work in over a year, we can safely start disabling its test suites and removing it from the dev server stack.  We should keep everything intact should we need to run it but we can at least move it out of our critical path.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1959)

jaredlockhart:
Because

* The legacy workflow is in maintenance only mode and has not had any changes or failures in over a year

This commit

* Moves the legacy tests out of make check but preserves them in make check_legacy
* Moves the legacy dev services out of make up but preserves them in make up_legacy
* Removes the integration_legacy job from circle

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,28 +40,6 @@ jobs:
             ./scripts/store_git_info.sh
             make publish_storybooks
 
-  integration_legacy:
-    machine:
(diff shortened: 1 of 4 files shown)


lmorchard (approved):
✂️ it out

PR #6832, opened 2022-01-04
fix #6746 chore(nimbus): rename gql types with input/type/enum

### LINKED ISSUE #6746: Rename V5 API types to dilineate between queries/enums/inputs
The current naming for the graphql types in the V5 API does not consistently refer to query types as `Type`, enums as `Enum`, or inputs as `Input`, which can make it difficult to reason about which type is being referred to on the frontend.  Let's rename those types and update their references in the frontend.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1920)

jaredlockhart:
Because

* Not all of our graphql types were consistently named
* Sometimes it can be hard to tell whether something is a read type, a write input, or an enum just from the name without a suffix

This commit

* Renames all the v5 api graphql types to have either type, input, or enum as a suffix

### DIFF
--- a/app/experimenter/experiments/api/v5/inputs.py
+++ b/app/experimenter/experiments/api/v5/inputs.py
@@ -2,43 +2,43 @@ import graphene
 from graphene_file_upload.scalars import Upload
 
 from experimenter.experiments.api.v5.types import (
-    NimbusExperimentApplication,
-    NimbusExperimentChannel,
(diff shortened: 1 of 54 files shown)


lmorchard (approved):
Kind of skimmed over the changes, looks good overall. This seems like a thing that linting should catch if anything's amiss. (And it passed, so...)

PR #6964, opened 2022-01-11
fix #6145 fix #6146 fix #6152 refactor(nimbus): introduce new experiment context and migrate PageSummary to using it

### LINKED ISSUE #6145: Set up Context for experiment properties
In a new file, `lib/contexts.ts`, create and export a React Context instance with `React.createContext()`; to hold experiment-related properties. Eventually this could also have a context to hold analysis-related properties.

`ExperimentContext` should contain the following properties:

* `experiment?: getExperiment_experimentBySlug;` - An object containing the current experiment data.
* `refetch: () => void;` - A function to re-fetch the experiment data.
* `status: StatusCheck;` - An object containing the results of a call to the `getStatus` helper with experiment data.
* `review: ReviewCheck;` - An object containing the results of a call to the `useReviewCheck` hook with experiment data.
* `polling: boolean` - A boolean to determine if we should be polling for changes to the experiment.

(Since `status` and `review` just take an experiment object it could be argued that they don’t need to be put into the context and could just be called as needed with experiment data, but since they’re so frequently used I figure it’s best to stick them in here. Open to pushback.)

With this context set up it should now be populated with properties: in the `AppLayoutWithExperiment` component wrap the returned `<Layout>` component in the context’s provider component. The end should look something like:

```
<ExperimentContext.Provider value={experimentContext}>
  <Layout>...</Layout>
</ExperimentContext.Provider>
```

Where the `value` prop contains an object of the context’s properties.

Once complete any child component will now have access to this context’s’ properties via the `useContext` hook.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1598)

### LINKED ISSUE #6146: Create mock Context Provider component
Both our tests and Storybook will benefit from mocked contexts, allowing us to mostly remove dependence on GQL mocks since values can now be mocked on the context instead of the query response.

In `lib/mocks.tsx` create and export a new component, `MockExperimentContextProvider` (or similar). It should take and render children, wrapping them in context’s provider component using dummy data.

[Here’s an example](https://gist.github.com/jodyheavener/63ede34122b9d51bde1b935d38b7ee13) of what this could look like (ignore the analysis example for now).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1599)

### LINKED ISSUE #6152: Migrate PageSummary component to ExperimentContext
Update the PageSummary component, tests, and Storybook stories so that they use React Context.

Largely, update the component to pull properties out of the ExperimentContext. For example: 

```java
const { experiment, status, review } = useContext(ExperimentContext);
```

Some things to keep in mind:

* Some larger refactoring will be needed to replace function-as-child setups in some components (e.g. [PageEditAudience](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.tsx#L24-L25)), but generally we should be able to eliminate this along with the hacky refs to experiment data and the refetch function.
* Usage of `getStatus` and `useReviewCheck` in components can now be replaced by pulling these properties out of the ExperimentContext.
* The component should no longer take props such as `experiment`, `status`, or anything review-related since these can be extracted from the context. However it should still provide these props to child components until they are migrated to context as well.
* Tests and stories that use `mockExperimentQuery` or similar mocks should instead now wrap the component in `MockExperimentContextProvider` to mock the properties the component retrieves from the context.

For bonus points (but not necessary if the diff is already getting large), when updating stories you could consider transitioning them to the new format. Check [these stories](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.stories.tsx) out for an example.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1605)

lmorchard:
Because:

- we want to transition the Nimbus UI React app to use React Context for
  passing around Experiment properties.

This commit:

- creates a new React for experiment and analysis fetch

- introduces an ExperimentRoot wrapper in App router that populates the
  experiment context for routes that deal with an experiment

- convert PageSummary over to use context

- some transitional hacks in tests to apply context instead of Apollo
  mock queries


lmorchard:
Opening as a draft because I want to bang on it a bit more tomorrow, but I *think* it's a good incremental first step. 

Will probably file another issue or two for things I ran into along the way. In particular:

* There's more controller logic in AppLayoutWithExperiment than I expected - turns out that's where we handle a few things like experiment not found, loading errors, and experiment polling. Those probably should live in the new ExperimentRoot in the router, but pulling on that thread unraveled a bunch of things better pushed off to a follow up

* Providing experiment via context can simplify things, but a lot of tests rely on simulating a GQL query to fetch that experiment. Got a hack to patch over that in this PR, but actually switching those tests over to using a context would be a better eventual follow up.

app/experimenter/nimbus-ui/src/components/PageResults/index.test.tsx
```diff
@@ -128,11 +128,13 @@ jest.mock("../AppLayoutWithExperiment", () => ({
 
     return (
       <div data-testid="PageResults">
-        {props.children({
-          experiment,
-          analysis,
-          refetch: () => Promise.resolve(),
-        })}
+        {typeof props.children === "function"
```
jaredlockhart, line 131: It's too bad there isn't a better way to handle this case

jaredlockhart:
Awesome this is a good start. It'll be good to migrate more pages over to it.  Thanks for getting started on this!

app/experimenter/nimbus-ui/src/components/PageResults/index.test.tsx
lmorchard: Well, this case should go away entirely once all the `Page*` components are using the context rather than the functional child pattern

jaredlockhart: Oh wonderful 👍

lmorchard:
Just going to mark this ready for review. Had one bit of code marked as ignored for test coverage, but I figured out why it wasn't getting hit by tests.

PR #6990, opened 2022-01-17
fix #6981 chore(nimbus): add firefox version 96.0.2

### LINKED ISSUE #6981: Add a custom targeting for Firefox 96.0.2
This is a follow-up to #6940.

Per relman, they're planning on cutting a dot release now to fix some other urgent issues, and that will not include the patches for our experiment. As a result, we have to bump our minimal Fx version accordingly.

Relman also suggested that our patches will be included in a dot release scheduled on Jan 20th. Not sure if we will have another dot release(s) before that. I guess we can add this filter before that gets finalized next week.

@jaredlockhart What do you think?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2122)

jaredlockhart:
Because

* Some experiments need to target Firefox 96.0.2

This commit

* Adds Firefox version 96.0.2


lmorchard (approved):

ncloudioj:
Doh, looks like there are some CI check failures.

@jaredlockhart Can you take a look as we're looking to lunch this tomorrow.? Thanks in advance!

PR #7023, opened 2022-01-28
fix #6739 feat(nimbus): support multifeature in v5 api

### LINKED ISSUE #6739: Add feature_configs/feature_values to V5 API
Let's add a new field to the V5 API to pass the `feature_configs` field from NimbusExperiment to the frontend and back, without deprecating the existing `feature_config` field.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1913)

jaredlockhart:
Because

* We need to update the V5 API to support setting multiple features on the experiment and per branch
* We need to keep the single feature interface intact while we make the transition

This commit

* Adds featureConfigs and featureValues to the V5 API and updates tests as necessary

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -64,6 +64,35 @@ class NimbusBranchScreenshotSerializer(serializers.ModelSerializer):
         )
 
 
+class NimbusBranchFeatureValueSerializer(serializers.ModelSerializer):
+    feature_config = serializers.PrimaryKeyRelatedField(
+        queryset=NimbusFeatureConfig.objects.all(), required=False, allow_null=False
+    )
+    enabled = serializers.BooleanField(required=False)
+    value = serializers.CharField(required=False, allow_blank=True)
+
+    class Meta:
+        model = NimbusBranchFeatureValue
+        fields = (
+            "feature_config",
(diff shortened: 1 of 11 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -227,50 +294,51 @@ def _validate_single_branch_for_rollout(self, data):
         return data
 
     def update(self, experiment, data):
-        with transaction.atomic():
-            reference_branch_data = data.pop("reference_branch", None)
-            treatment_branches_data = data.pop("treatment_branches", None)
+        data.pop("reference_branch", None)
```
lmorchard, line 297: Good clean refactor here. :+1: Scratching my head at why this was complicated originally. IIRC, I think [this bugfix for on_delete=SET_NULL](https://github.com/mozilla/experimenter/commit/7385e50dbc0a2651b8ffb349da37c4d41b6c5b8f#diff-36bd393fce70ef5fc7ad821bc3c232a4b00882f000583f222fc3041f735d3bdbR156) needed to land first. The experiment itself would get deleted without explicitly setting `experiment.reference_branch = None` first. But, this works now so yay!

lmorchard (approved):
Looking pretty good, some nice refactors 👍

PR #7033, opened 2022-02-01
fixes #6967 fix(Nimbus): removes choices from Firefox version model f…

### LINKED ISSUE #6967: Remove Firefox Version from model choices to prevent migrations
Since we've had a couple cases already where we needed to add custom firefox versions for particular experiments, there's no need to introduce a migration for such a change.  We should remove the choices parameter from the version model fields.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2107)

anthonytran7757:
…ield

Because

* there were cases where a custom Firefox version was required for a particular experiment

This commit

* removes choices from Firefox version model field


app/experimenter/experiments/admin/nimbus.py
```diff
@@ -98,12 +98,8 @@ class NimbusExperimentAdminForm(forms.ModelForm):
     application = forms.ChoiceField(choices=NimbusExperiment.Application.choices)
     channel = forms.ChoiceField(choices=NimbusExperiment.Channel.choices)
     public_description = forms.CharField(required=False, widget=forms.Textarea())
-    firefox_min_version = forms.ChoiceField(
-        choices=NimbusExperiment.Version.choices, required=False
-    )
-    firefox_max_version = forms.ChoiceField(
-        choices=NimbusExperiment.Version.choices, required=False
-    )
```
jaredlockhart, line 106: Oh sorry no we do want to keep it here in the admin 🙏

jaredlockhart (changes requested):
Thank you @anthonytran7757 !  Looks good, just pls revert the admin change since we want to keep the choices there just take them off the model 🙏

jaredlockhart (approved):
Tytytyty 🙏

PR #7074, opened 2022-02-09
fix #6740 chore(nimbus): update nimbus ui to use feature_configs

### LINKED ISSUE #6740: Use feature_configs in nimbus-ui

jaredlockhart:
Because

* In anticipation of adding multifeature to nimbus ui, we should update the core nimbus ui infrastructure to use feature_configs instead of feature_config

This commit

* Updates the branches page to use and set feature_configs instead of feature_config

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.tsx
@@ -41,8 +41,6 @@ const TableBranches = ({
 }: {
   experiment: getExperiment_experimentBySlug;
 }) => {
-  const { featureConfig } = experiment;
-  const hasSchema = featureConfig?.schema !== null;
   const branches = [
     experiment.referenceBranch,
     ...(experiment.treatmentBranches || []),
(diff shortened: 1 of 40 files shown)


app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.tsx
```diff
@@ -59,7 +57,7 @@ const TableBranches = ({
       ) : (
         <>
           {savedBranches.map((branch, key) => (
-            <TableBranch key={key} {...{ hasSchema, experiment, branch }} />
```
eliserichards, line 62: We no longer need to know if it has a schema?

eliserichards (approved):
Looks good! 🎉

app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.tsx
jaredlockhart: That's actually a holdover from a long since passed intermediate implementation where we thought we only wanted to enter feature values for features that were enabled and had schemas, but that concept has gone and away and this should've been removed a while ago.

PR #7076, opened 2022-02-10
Use assertEqual for comparisons, not assertEqual

### LINKED ISSUE #7075: Some tests misusing assertTrue for comparisons
`assertTrue` is not for comparing arguments. Consider `assertEqual` instead.

The developer's intent of the test was to compare argument 1 with argument 2, which is not happening. Really what is happening is the test is passing because first argument is truthy. The correct method to use is `assertEqual`. [more details](https://codereview.doctor/features/python/best-practice/avoid-misusing-unittest-assert-true)

https://github.com/mozilla/experimenter/blob/7c5f876c884a7ef6603937a1fb09b4e8712a5f4d/app/experimenter/experiments/tests/test_models/test_models_legacy.py#L1655
https://github.com/mozilla/experimenter/blob/7c5f876c884a7ef6603937a1fb09b4e8712a5f4d/app/experimenter/experiments/tests/test_models/test_models_legacy.py#L1662
https://github.com/mozilla/experimenter/blob/7c5f876c884a7ef6603937a1fb09b4e8712a5f4d/app/experimenter/experiments/tests/api/v2/test_serializers.py#L172

I found this issue automatically, see other issues [here](https://codereview.doctor/mozilla/experimenter)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2206)

code-review-doctor:
Fixes #7075

### DIFF
--- a/app/experimenter/experiments/tests/api/v2/test_serializers.py
+++ b/app/experimenter/experiments/tests/api/v2/test_serializers.py
@@ -169,7 +169,7 @@ class TestExperimentDesignMultiPrefSerializer(MockRequestMixin, TestCase):
         self.assertEqual(self.experiment.changes.count(), 0)
 
         experiment = serializer.save()
-        self.assertTrue(experiment.variants.all().count(), 1)
+        self.assertEqual(experiment.variants.all().count(), 1)
         self.assertEqual(experiment.changes.count(), 1)
 
         control = ExperimentVariant.objects.get(experiment=experiment, is_control=True)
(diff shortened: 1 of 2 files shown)


code-review-doctor:
the test is failing because the test is now actually checking the logic the dev wanted to check when the test was written.

question is: does this failure indicate that there is a bug in the logic the test checks, or is the test wrong and should be checking if the value equals 2?

jaredlockhart (approved):

PR #7088, opened 2022-02-16
fix #7087 feat(nimbus): add pip targeting with sticky

### LINKED ISSUE #7087: Create an audience filter for Picture-in-Picture
Hi,

Can you please help us create an audience filter for a Picture-in-Picture Nimbus experiment?
Here are the targeting criteria:

```
# media.videocontrols.picture-in-picture.video-toggle.has-used = false
# media.videocontrols.picture-in-picture.enabled = true
# media.videocontrols.picture-in-picture.video-toggle.enabled = true
```

We want to keep users in the experiment after they use PiP for the first time OR disable PiP while in the experiment, so the filter set needs to be sticky. 

With the existing filter (PiP never used) i[n this experiment|https://experimenter.services.mozilla.com/nimbus/picture-in-picture-first-time-toggle-text]:

* users are unenrolled from the study after using the Picture in Picture for the first time (bug [1754957|https://bugzilla.mozilla.org/show_bug.cgi?id=1754957],
* users are unenrolled after disabling PiP while in the study (bug [1754986|https://bugzilla.mozilla.org/show_bug.cgi?id=1754986])

cc @Niklas Baumgardner

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2217)

jaredlockhart:
Because

* The PIP targeting depends on prefs that may change during the experiment
* We need to add a sticky clause

This commit

* Adds a new PIP sticky targeting that includes the pref checks from the PIP targeting with a sticky clause

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -448,8 +448,22 @@ TARGETING_PIP_NEVER_USED = NimbusTargetingConfig(
         !('media.videocontrols.picture-in-picture.video-toggle.has-used'|preferenceValue)
         &&
         ('media.videocontrols.picture-in-picture.enabled'|preferenceValue)
+        &&
+        ('media.videocontrols.picture-in-picture.video-toggle.enabled'|preferenceValue)
     )
     """,
+    desktop_telemetry="",
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_PIP_NEVER_USED_STICKY = NimbusTargetingConfig(
+    name="PiP Never Used (Sticky)",
+    slug="pip_never_used_sticky",
+    description="Users that have never used Picture in Picture, with sticky enrollment",
+    targeting="(({pip}) || ({sticky}))".format(
+        pip=TARGETING_PIP_NEVER_USED.targeting,
+        sticky=TARGETING_STICKY,
+    ),
     desktop_telemetry="",
     application_choice_names=(Application.DESKTOP.name,),
 )
@@ -671,6 +685,7 @@ class NimbusConstants(object):
             TARGETING_INFREQUENT_OR_CASUAL_WIN_USER_CAN_PIN
         ),
         TARGETING_PIP_NEVER_USED.slug: TARGETING_PIP_NEVER_USED,
+        TARGETING_PIP_NEVER_USED_STICKY.slug: TARGETING_PIP_NEVER_USED_STICKY,
     }
 
     class TargetingConfig(models.TextChoices):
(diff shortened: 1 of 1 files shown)


jaredlockhart:
@eliserichards For context, this PR is adding a 'targeting config' which is a constant that generates the JEXL that will be used when clients enroll.  The JEXL is validated in two ways:

1) We use the pyjexl library to validate the syntax
2) We load the jexl expression into a firefox in the integration tests and run it through the actual firefox JEXL evaluator to make sure that both the syntax and variable names will resolve correctly in desktop Firefox

Additionally, we have tests to make sure that targeting configs are exposed to the forms correctly, so generally these targeting changes are pretty rigorously tested.

There will also be another reviewer from the PIP team to validate that the prefs names are correct.

So really the thing to look out for here is that the names/ids make sense, and that the format string parameters look okay.  But please feel free to ask questions about anything you're not sure of.  Thank you!

niklasbaumgardner:
New pref is correct

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -448,12 +448,26 @@ class NimbusTargetingConfig:
...
     desktop_telemetry="",
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_PIP_NEVER_USED_STICKY = NimbusTargetingConfig(
+    name="PiP Never Used (Sticky)",
+    slug="pip_never_used_sticky",
+    description="Users that have never used Picture in Picture, with sticky enrollment",
+    targeting="(({pip}) || ({sticky}))".format(
+        pip=TARGETING_PIP_NEVER_USED.targeting,
```
eliserichards, line 464: My understanding here would be that `TARGETING_PIP_NEVER_USED` should point it's targeting to `TARGETING_PIP_NEVER_USED_STICKY` (here `_STICKY` points to the old never used config). That way we don't need to change any references to the old targeting config but they would still get sticky 🤔 

Or is this just adding a new targeting config that can be used for future stuff?

jaredlockhart: Well, we could add in the sticky config to the old one, update it in place, except if existing experiments point to it then they'll be using a config whose behaviour is different than they expected to be when they chose it.  So instead we're adding a new one that adds the sticky behaviour next to the previous ones config.  

I'm glossing over a bit about keeping the old one intact since we're also adding a third pref to it but my understanding from the PIP team is that that pref should've always been present in the first one for it to perform the filtering they expected from it.

eliserichards: Totally makes sense 👍

eliserichards (approved):
LGTM! 🚀

PR #7100, opened 2022-02-25
fixes #6446 feat(Nimbus): include public name in rollout promotion

### LINKED ISSUE #6446: Prefix rollout experiment name w/ parent experiment name when doing a rollout
Currently if I click the "Promote to Rollout" button on a specific branch, it will use the branch's name as the experiment name. But this can be confusing and tricky considering a lot of the experiments have [non-descriptive] branch names like "control" or "treatment" or "a1/a2" or "var2", which leads to lots of naming collisions or just awkwardly named experiments like "Treatment Rollout".

So should we set a default "Public name" of "control Rollout" or should it say something like "Default Browser Adoption Experiment - control Rollout" to make it more obvious/unique?

```
import axios from "axios";

// NOTE: This fetches PRODUCTION data.
const { data: experiments } = await axios.get("https://experimenter.services.mozilla.com/api/v6/experiments/");
console.log(`${experiments.length} experiments`);

for (const exp of experiments) {
  console.log(exp.slug);
  for (const branch of exp.branches) {
    console.log(`  - ${branch.slug}`);
  }
}
```

!This_tests_experiment_creation_on_19_08_–_Summary__Review_End_Request___Experimenter.png|thumbnail!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1781)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/16475">This_tests_experiment_creation_on_19_08_–_Summary__Review_End_Request___Experimenter.png</a>

anthonytran7757:
Because

* branch names are not great rollout names

This Commit

* adds public name to rollout promotion

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/CloneDialog/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/CloneDialog/index.tsx
@@ -58,7 +58,7 @@ export const CloneDialog = ({
 }: CloneDialogProps) => {
   const defaultValues: CloneParams = {
     name: rolloutBranch
-      ? `${rolloutBranch.name} Rollout`
+      ? `${experiment!.name} - ${rolloutBranch.name} Rollout`
       : `${experiment!.name} Copy`,
   };
 


jaredlockhart (approved):
Perfect thank you @anthonytran7757 !  I just tested locally to make sure it handles the case of a long name exceeding 80 characters with `- {branch name} Rollout` tacked onto the end but it all works as expected so this is great thank you!

PR #7105, opened 2022-03-04
Add logic to build a correct JEXL expression for Mobile experiments.

jrbenny35:
Should fix https://mozilla-hub.atlassian.net/browse/EXP-2216

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -247,7 +247,12 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
         expressions = []
 
         if self.targeting_config and self.targeting_config.targeting:
-            expressions.append(self.targeting_config.targeting)
+            if self.application == self.Application.DESKTOP:
+                expressions.append(self.targeting_config.targeting)
+            else:
+                expressions.append(f"({self.targeting_config.targeting})")
+
+
 
         if self.application == self.Application.DESKTOP:
             if self.channel:
@@ -285,12 +290,12 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
                 )
                 iso_languages_expression = " || ".join(
                     [
-                        f"{language} in locale"
+                        f"'{language}' in locale"
                         for language in list(filter(lambda s: "-" not in s, locales))
                     ]
                 )
                 locale_expressions = (
-                    f"{iso_locales_expression} || {iso_languages_expression}"
+                    f"({iso_locales_expression} || {iso_languages_expression})"
                 )
                 expressions.append(locale_expressions)
 
@@ -300,9 +305,14 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
                 country.code for country in self.countries.all().order_by("code")
             ]
             if self.application == self.Application.DESKTOP:
-                expressions.append(f"region in {countries}")
+                expressions.append(f"region in '{countries}'")
             else:
-                expressions.append(f"locale in {countries}")
+                iso_region_expression = " || ".join(
+                    [
+                        f"'{region}' in locale" for region in countries
+                    ]
+                )
+                expressions.append(f"({iso_region_expression})")
 
         #  If there is no targeting defined all clients should match, so we return "true"
         if len(expressions) == 0:
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -280,8 +280,11 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
             if self.application == self.Application.DESKTOP:
                 expressions.append(f"locale in {locales}")
             else:
-                iso_locales = [locale for locale in locales if "-" in locale]
-                iso_locales_expression = f"locale in {iso_locales}"
+                iso_locale_list = [locale for locale in locales if "-" in locale]
+                iso_locales = set([locale[:2] for locale in iso_locale_list])
+                iso_locales_expression = " || ".join(
+                    [f"'{language}' in locale" for language in list(iso_locales)]
+                )
                 iso_languages = [locale for locale in locales if "-" not in locale]
                 iso_languages_expression = " || ".join(
                     [f"'{language}' in locale" for language in iso_languages]
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -284,10 +284,7 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
                 iso_locales_expression = " || ".join(
                     [f"'{language}' in locale" for language in sorted(iso_locales)]
                 )
-                locale_expressions = (
-                    f"({iso_locales_expression})"
-                )
-                expressions.append(locale_expressions)
+                expressions.append(f"({iso_locales_expression})")
 
         if self.countries.count():
             countries = [
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -280,17 +280,12 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
             if self.application == self.Application.DESKTOP:
                 expressions.append(f"locale in {locales}")
             else:
-                iso_locale_list = [locale for locale in locales if "-" in locale]
-                iso_locales = set([locale[:2] for locale in iso_locale_list])
+                iso_locales = {locale[:2] for locale in locales}
                 iso_locales_expression = " || ".join(
-                    [f"'{language}' in locale" for language in list(iso_locales)]
-                )
-                iso_languages = [locale for locale in locales if "-" not in locale]
-                iso_languages_expression = " || ".join(
-                    [f"'{language}' in locale" for language in iso_languages]
+                    [f"'{language}' in locale" for language in sorted(iso_locales)]
                 )
                 locale_expressions = (
-                    f"({iso_locales_expression} || {iso_languages_expression})"
+                    f"({iso_locales_expression})"
                 )
                 expressions.append(locale_expressions)
 
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -247,12 +247,7 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
         expressions = []
 
         if self.targeting_config and self.targeting_config.targeting:
-            if self.application == self.Application.DESKTOP:
-                expressions.append(self.targeting_config.targeting)
-            else:
-                expressions.append(f"({self.targeting_config.targeting})")
-
-
+            expressions.append(self.targeting_config.targeting)
 
         if self.application == self.Application.DESKTOP:
             if self.channel:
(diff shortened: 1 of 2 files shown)


jhugman (changes requested):
This isn't quite ready, but it's a great start! Well done!

jrbenny35:
@jhugman is the expression in the new test I added correct?

app/experimenter/experiments/models/nimbus.py
```diff
@@ -276,13 +276,33 @@ def targeting(self):
...
+        # Region matching
         if self.countries.count():
             countries = [
                 country.code for country in self.countries.all().order_by("code")
             ]
-            expressions.append(f"region in {countries}")
+            if self.application == self.Application.DESKTOP:
+                expressions.append(f"region in {countries}")
+            else:
+                expressions.append(f"locale in {countries}")
```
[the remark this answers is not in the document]
jrbenny35: Doesn't JEXL match a list? This would currently be ```["sk", "cz"] in locale```. Would that not work?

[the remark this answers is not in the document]
jhugman: yes, but `locale` is the ISO format (`sk-SK`), so you need a series of substring matches.

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -372,6 +372,31 @@ def test_targeting_with_locales_and_countries(self):
...
+            channel=NimbusExperiment.Channel.NO_CHANNEL,
+            locales=[locale_ca, locale_ro],
+            countries=[country_ca, country_ro],
+        )
+        self.assertEqual(
+            experiment.targeting,
+            (
+                "is_already_enrolled || days_since_install < 7 "
+                "&& locale in ['en-CA'] || ro in locale "
+                "&& locale in ['CA', 'RO']"
```
jhugman, line 396: This should read:

```suggestion
                "(is_already_enrolled || days_since_install < 7) "
                "&& (locale in ['en-CA'] || 'ro' in locale) "
                "&& ('CA' in locale || 'RO' in locale)"
```

NB: I also noticed `ro` should  be quoted.

jrbenny35: so it must have the parentheses?

jhugman: :+1: yes.

jrbenny35:
Extended the test case a bit

jaredlockhart (changes requested):
Thank you @jrbenny35 and @jhugman for digging into this!  After our discussion yesterday it looks like a path forward is to fix the locale case immediately here until we have a more rigorous solution that involves capturing the language codes used by iOS/Android and adding them to a separate mobile language field in Experimenter, and dealing with the region case separately.  Tickets/Epics for that work are forthcoming.

jaredlockhart:
Also @jrbenny35 note the contributing guidelines here about commit formatting:

https://github.com/mozilla/experimenter/blob/main/contributing.md#subject

jrbenny35:
> Also @jrbenny35 note the contributing guidelines here about commit formatting:
> 
> https://github.com/mozilla/experimenter/blob/main/contributing.md#subject

So the proper JEXL would be, 
```(is_already_enrolled || days_since_install < 7) "
    "&& (locale in ['en-CA'] || 'de' in locale || 'en' in locale || 'ro' in locale)"
```

app/experimenter/experiments/models/nimbus.py
```diff
@@ -247,7 +247,12 @@ def targeting(self):
         expressions = []
 
         if self.targeting_config and self.targeting_config.targeting:
-            expressions.append(self.targeting_config.targeting)
+            if self.application == self.Application.DESKTOP:
+                expressions.append(self.targeting_config.targeting)
+            else:
+                expressions.append(f"({self.targeting_config.targeting})")
```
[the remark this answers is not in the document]
jrbenny35: No, the parentheses are needed.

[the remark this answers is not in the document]
jrbenny35: oh, not sure but good point. I guess it wouldn't hurt. Wonder if this would mean I'd need to update all the JEXL tests though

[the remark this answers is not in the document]
jaredlockhart: Actually how does this relate specifically to the locale targeting?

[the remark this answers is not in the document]
jrbenny35: Not sure. If you read above, I asked James for a correct JEXL string, it included these parentheses.

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
jaredlockhart: Oh I see, this is an entirely separate issue.  We should file this separately.

jaredlockhart: Filed https://github.com/mozilla/experimenter/issues/7120

app/experimenter/experiments/models/nimbus.py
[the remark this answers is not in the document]
jaredlockhart: Oh I see yeah that's an entirely unrelated issue, not specific to mobile or locales/regions.  Filed it separately as https://github.com/mozilla/experimenter/issues/7120 so don't worry about it here.

jrbenny35:
I can update the commit before merging

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -372,6 +372,30 @@ def test_targeting_with_locales_and_countries(self):
...
+            targeting_config_slug=targeting_config.TARGETING_MOBILE_NEW_USER,
+            channel=NimbusExperiment.Channel.NO_CHANNEL,
+            locales=[locale_ca, locale_ro, locale_de, locale_en],
+        )
+        self.assertEqual(
+            experiment.targeting,
+            (
+                "is_already_enrolled || days_since_install < 7 "
+                "&& (locale in ['en-CA'] || 'de' in locale || 'en' in locale "
+                "|| 'ro' in locale)"
```
jaredlockhart, line 395: @jhugman @travis79 Can you please help verify that this looks correct as a workaround to match Experimenter's locale codes to the mobile ones?

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -372,6 +372,30 @@ def test_targeting_with_locales_and_countries(self):
...
+            firefox_min_version=NimbusExperiment.Version.NO_VERSION,
+            firefox_max_version=NimbusExperiment.Version.NO_VERSION,
+            targeting_config_slug=targeting_config.TARGETING_MOBILE_NEW_USER,
+            channel=NimbusExperiment.Channel.NO_CHANNEL,
+            locales=[locale_ca, locale_ro, locale_de, locale_en],
+        )
+        self.assertEqual(
+            experiment.targeting,
+            (
+                "is_already_enrolled || days_since_install < 7 "
```
jaredlockhart, line 393: The point @jhugman made about wrapping this in parens will be handled separately by https://github.com/mozilla/experimenter/issues/7120

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
jaredlockhart: Supplemental @jrbenny35 is there a way we can test/verify this against a mobile client before we deploy it?

jaredlockhart:
@jrbenny35 Amazing this looks fantastic and exactly right to me, let's see if there's some way we can test/verify this against a mobile client and then get it merged? 🥳 🙏

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
travis79: Yes, this format looks correct to me for `en-CA` as to what we would currently have in the `locale` of our  context.

jrbenny35: Looking at the checks here, https://github.com/mozilla/application-services/pull/4857 this looks fine

jaredlockhart (approved):
Okay this looks great let's get it landed and then we can test it on stage.  Filed a followup QA request here:

Filed a QA request to have this QAed here:

https://mozilla-hub.atlassian.net/browse/QA-1421

Thanks @jrbenny35!!! 🎉 🎉 🎉 🎉 🎉

jaredlockhart (changes requested):
Oh I just realized one last thing before we land this we should do this:

jaredlockhart (approved):
Actually we can do it as a followup https://github.com/mozilla/experimenter/issues/7124

jaredlockhart (changes requested):
I need to think about this for a minute.

jaredlockhart:
Okay so thinking about the way mobile does this, it constructs its code by querying the language and region from the phone, sticking them together with a `-`, and then calling it `locale` in the JEXL context.  So for an English phone in Romania the code would be `en-RO`.  And since Experimenter doesn't have an 'English' code ie `en`, it only has region specific English codes ie `en-CA`, `en-US`, `en-UK`, etc. then by targeting 'English' language you would inadvertently also be targeting only those geogrpahic countries as well.

So what we need to do is

1. Split all the Experimenter locale codes on `-` and pull out the first element to get just the language code
1. Take the unique set of language codes by wrapping in `set(...)`
1. Construct the `||` expression out of those
1. Ignore the `locale in [....]` case altogether since that's inadvertently doing region matching


So then if you create an experiment like

Experiment Locales: English(US), English(CA), German(DE)
Experiment Regions: <none, disabled for mobile>

JEXL: `('en' in locale || 'de' in locale)`

jrbenny35:
How about this? ```(locale in ['en', 'es'] || 'de' in locale || 'ro' in locale)```

app/experimenter/experiments/models/nimbus.py
```diff
@@ -276,7 +276,23 @@ def targeting(self):
...
+                iso_locales_expression = " || ".join(
+                    [f"'{language}' in locale" for language in list(iso_locales)]
+                )
+                iso_languages = [locale for locale in locales if "-" not in locale]
+                iso_languages_expression = " || ".join(
+                    [f"'{language}' in locale" for language in iso_languages]
+                )
+                locale_expressions = (
+                    f"({iso_locales_expression} || {iso_languages_expression})"
+                )
```
jaredlockhart, line 294: I think you'll find if you take out the `"-" in locale` check on line 283 you can take this second case out altogether.

app/experimenter/experiments/models/nimbus.py
```diff
@@ -276,7 +276,23 @@ def targeting(self):
 
         if self.locales.count():
             locales = [locale.code for locale in self.locales.all().order_by("code")]
-            expressions.append(f"locale in {locales}")
+
+            if self.application == self.Application.DESKTOP:
+                expressions.append(f"locale in {locales}")
+            else:
+                iso_locale_list = [locale for locale in locales if "-" in locale]
+                iso_locales = set([locale[:2] for locale in iso_locale_list])
```
jaredlockhart, line 284: Yes good idea just taking the first two characters 👍

app/experimenter/experiments/models/nimbus.py
```diff
@@ -276,7 +276,23 @@ def targeting(self):
...
+                iso_locales_expression = " || ".join(
+                    [f"'{language}' in locale" for language in list(iso_locales)]
```
jaredlockhart, line 286: You don't need to wrap it in a `list()`, `set` also implements an iterator iterface 👍

app/experimenter/experiments/tests/test_models/test_models_nimbus.py
```diff
@@ -372,6 +372,31 @@ def test_targeting_with_locales_and_countries(self):
...
+            targeting_config_slug=targeting_config.TARGETING_MOBILE_NEW_USER,
+            channel=NimbusExperiment.Channel.NO_CHANNEL,
+            locales=[locale_ca, locale_ro, locale_de, locale_es, locale_us],
+        )
+        self.assertEqual(
+            experiment.targeting,
+            (
+                "is_already_enrolled || days_since_install < 7 "
+                "&& ('en' in locale || 'es' in locale || 'de' in locale "
+                "|| 'ro' in locale)"
```
jaredlockhart, line 396: 👏 👏 👏 👏 👏  looks good to me 👍

jaredlockhart (changes requested):

PR #7116, opened 2022-03-08
 fixes #7113: Limits integration test generated names to 76 chars 

### LINKED ISSUE #7113: Limit integration test name/slug to 80 characters
If an integration test creates an experiment whose name is longer than 80 characters it will fail validation and hang.  This can cause inadvertent integration test failures on valid PRs.  We should trim the names in our create experiment fixture to prevent this case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2263)

jrbenny35:


app/tests/integration/nimbus/conftest.py
```diff
@@ -121,9 +121,10 @@ def experiment_url(base_url, default_data, slugify):
 def default_data(request):
     application = request.param
     feature_config = APPLICATION_FEATURES[application]
+    public_name = f"{request.node.name}-{str(uuid.uuid4())[:4]}"
 
     return BaseExperimentDataClass(
-        public_name=f"{request.node.name}-{str(uuid.uuid4())[:4]}",
+        public_name=public_name[:80],
```
jaredlockhart, line 127: Oh yeah I was thinking about this and I realized if we do this way then we could chop off the UUID portion altogether, I think we should be trimming `request.node.name[:76]` so we always include the UUID at the end to prevent name collisions on test reruns.

jaredlockhart (dismissed):

app/tests/integration/nimbus/conftest.py
jrbenny35: doh, good one!

eliserichards (approved):
Looks good! 🚀

PR #7118, opened 2022-03-09
fixes #7117 chore(project): Add firefox versions up to 120

### LINKED ISSUE #7117: Add new firefox versions up to 120
As we are at firefox 99 (legacy) and firefox 100 (nimbus). We should add more firefox versions for future experiments that want to target these versions

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2269)

tiftran:
Because:
 * In the future we will need to target experiments above 100

This commit:
 * adds to the version lists up to version 120


jaredlockhart (approved):
Tytytyty 🙏

PR #7123, opened 2022-03-10
fix #7122 bug(project): disable reporting task

### LINKED ISSUE #7122: Disable reporting changelog tasks
We're seeing some unusual database load that looks like a periodically scheduled task that's hitting the changelog table.  I suspect it's the [reporting task](https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L348-L351) that periodically rolls up changelogs.  I'm going to try disabling that task and see if that clears it up.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2273)

jaredlockhart:
Because

* We're seeing intermittent database load spikes that are preventing Experimenter from loading
* The spikes are happening on queries to the changelog table
* It may be the reporting task hitting that table that is causing the issue

This commit

* Disables the reporting task

### DIFF
--- a/app/experimenter/settings.py
+++ b/app/experimenter/settings.py
@@ -345,10 +345,10 @@ CELERY_BEAT_SCHEDULE = {
         "task": "experimenter.jetstream.tasks.fetch_jetstream_data",
         "schedule": 28800,
     },
-    "reporting_generate_report_logs": {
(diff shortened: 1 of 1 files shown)


tiftran (approved):
🙈

PR #7131, opened 2022-03-16
fix #6769 feat(nimbus): enable version targeting on mobile

### LINKED ISSUE #6769: Support version targeting on mobile
When version targeting is supported on mobile we need to remove the condition that only applies it to desktop experiments.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1934)

jaredlockhart:
Because

* Mobile clients now support version targeting

This commit

* Enables version targeting for mobile on supported clients
* Only includes version targeting for mobile clients above the minimum supported version

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -238,6 +238,32 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
             "https://{host}".format(host=settings.HOSTNAME), self.get_absolute_url()
         )
 
+    def _get_targeting_versions(self):
+        expressions = []
+
+        version_key = "version"
+        if self.application != self.Application.DESKTOP:
+            version_key = "app_version"
+
+        version_supported = True
+        if self.application in self.TARGETING_APPLICATION_SUPPORTED_VERSION:
+            supported_version = self.TARGETING_APPLICATION_SUPPORTED_VERSION[
+                self.application
+            ]
+            version_supported = self.firefox_min_version >= supported_version
+
+        if version_supported:
+            if self.firefox_min_version:
+                expressions.append(
+                    f"{version_key}|versionCompare('{self.firefox_min_version}') >= 0"
+                )
+            if self.firefox_max_version:
+                # HACK: tweak the min version to better match max version pattern
+                max_version = self.firefox_max_version.replace("!", "*")
+                expressions.append(f"{version_key}|versionCompare('{max_version}') < 0")
+
+        return expressions
+
     # This is the full JEXL expression processed by clients
     @property
     def targeting(self):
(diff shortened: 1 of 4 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -238,6 +238,34 @@ def experiment_url(self):
...
+        if self.application in self.TARGETING_APPLICATION_SUPPORTED_VERSION:
+            supported_version = self.TARGETING_APPLICATION_SUPPORTED_VERSION[
+                self.application
+            ]
+            version_supported = self.firefox_min_version >= supported_version
+
+        if version_supported:
+            if self.firefox_min_version:
+                expressions.append(
+                    "version|versionCompare('{version}') >= 0".format(
```
jrbenny35, line 254: Why are you using ```.format``` vs an f-string?

jrbenny35 (approved):

eliserichards (approved):
Looks good! This is going to be really useful 📱

app/experimenter/experiments/models/nimbus.py
tarikeshaq, line 254:
```suggestion
                    "app_version|versionCompare('{version}') >= 0".format(
```

This is annoying, but I think it should be `app_version` from https://github.com/mozilla/application-services/blob/main/components/nimbus/src/tests/test_evaluator.rs#L109 

it's changeable, I'll file a ticket to create another `version` parameter with the same value, so we can end up depricating the `app_version` for consistency with desktop

tarikeshaq (changes requested):
Came across this and I think the parameter in mobile is `app_version` (which really it shouldn't be, it's always been like this in the SDK but I think we should make it consistent with desktop)

Starting to think that maybe we should prioritize the SDK targeting integration tests to catch stuff like this 😕

app/experimenter/experiments/models/nimbus.py
jaredlockhart: Oh yeah good point 👍

jaredlockhart:
> Came across this and I think the parameter in mobile is `app_version` (which really it shouldn't be, it's always been like this in the SDK but I think we should make it consistent with desktop)
> 
> Starting to think that maybe we should prioritize the SDK targeting integration tests to catch stuff like this 😕

Yep I've already been talking to @jrbenny35 about exactly this for exactly this reason 👍

app/experimenter/experiments/models/nimbus.py
```diff
@@ -238,6 +238,32 @@ def experiment_url(self):
             "https://{host}".format(host=settings.HOSTNAME), self.get_absolute_url()
         )
 
+    def _get_targeting_versions(self):
+        expressions = []
+
+        version_key = "version"
+        if self.application != self.Application.DESKTOP:
+            version_key = "app_version"
```
eliserichards, line 246: 👏

eliserichards (commented):
LGTM! I'll leave the final r+ to @tarikeshaq though :)

tarikeshaq (approved):
The versions and targeting parameter look good to me! Do you know if there be any type of QA or A/A experiment since this is the first time we expose this for mobile?

jaredlockhart:
> The versions and targeting parameter look good to me! Do you know if there be any type of QA or A/A experiment since this is the first time we expose this for mobile?

Filed https://mozilla-hub.atlassian.net/browse/QA-1438

PR #7132, opened 2022-03-16
fix #7120 chore(nimbus): wrap all JEXL expressions in parens

### LINKED ISSUE #7120: Wrap targeting config JEXL in outer parens
We always want to treat a targeting configs expression as an atomic unit, so we should wrap it in outer parens here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L250

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2271)

jaredlockhart:
Because

* Arbitrary JEXL expressions can contain arbitrary subclauses
* To ensure the JEXL evaluator groups them the way is intended by the subclause authors

This commit

* Wraps all JEXL subclauses in outer parens

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -284,7 +284,7 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
                 iso_locales_expression = " || ".join(
                     [f"'{language}' in locale" for language in sorted(iso_locales)]
                 )
-                expressions.append(f"({iso_locales_expression})")
+                expressions.append(iso_locales_expression)
 
         if self.countries.count():
             countries = [
@@ -296,7 +296,7 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
         if len(expressions) == 0:
             return "true"
 
-        return " && ".join(expressions)
+        return " && ".join([f"({expression})" for expression in expressions])
 
     @property
     def application_config(self):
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/models/nimbus.py
```diff
@@ -296,7 +296,7 @@ def targeting(self):
         if len(expressions) == 0:
             return "true"
 
-        return " && ".join(expressions)
+        return " && ".join([f"({expression})" for expression in expressions])
```
eliserichards, line 299: 👏

app/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -54,9 +54,9 @@ def test_expected_schema_with_desktop_single_feature(self):
                 # DRF manually replaces the isoformat suffix so we have to do the same
                 "startDate": experiment.start_date.isoformat().replace("+00:00", "Z"),
                 "targeting": (
-                    'browserSettings.update.channel == "nightly" '
-                    "&& version|versionCompare('94.!') >= 0 "
-                    "&& 'app.shield.optoutstudies.enabled'|preferenceValue"
```
eliserichards, line 59: Should we be testing both cases (with and without parens)?

eliserichards (approved):
Looks good! Just one q

app/experimenter/experiments/tests/api/v6/test_serializers.py
jaredlockhart: No no there won't be a non paren case anymore

PR #7138, opened 2022-03-23
fix #7137 feat(nimbus): disable country filtering for mobile

### LINKED ISSUE #7137: Disable region filtering for mobile
The region filter is not yet implemented for mobile clients so we should disable the region option in the UI until it is implemented.

jaredlockhart:
Becuase

* Country(region) filtering is not yet implemented for mobile

This commit

* Disables the country field for mobile experiments

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
@@ -22,6 +22,7 @@ import {
   getConfig_nimbusConfig_targetingConfigs,
 } from "../../../types/getConfig";
 import { getExperiment_experimentBySlug } from "../../../types/getExperiment";
+import { NimbusExperimentApplicationEnum } from "../../../types/globalTypes";
 import LinkExternal from "../../LinkExternal";
 
 type FormAudienceProps = {
@@ -136,6 +137,9 @@ export const FormAudience = ({
     [config, experiment],
   );
 
+  const isDesktop =
+    experiment.application === NimbusExperimentApplicationEnum.DESKTOP;
+
   return (
     <Form
       noValidate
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -201,9 +205,15 @@ export const FormAudience = ({
             <Select
               placeholder="All Countries"
               isMulti
+              isDisabled={!isDesktop}
               {...formSelectAttrs("countries", setCountries)}
               options={selectOptions(config.countries as SelectIdItems)}
             />
+            {!isDesktop ? (
+              <p className="text-secondary">
```
eliserichards, line 213: nit: is there any sort of `text-error` that we could use here?

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -201,9 +205,15 @@ export const FormAudience = ({
             <Select
               placeholder="All Countries"
```
eliserichards, line 206: Could we add the caveat as secondary text underneath the `Form.label` (in all scenarios) instead of doing it only when it's `!isDesktop`?

.. I guess that might be a "don't let the user know that a feature doesn't exist" situation, though

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
jaredlockhart: Oh yes there is, but I like to reserve using 'error' styled messages for cases where the user has entered something invalid they need to change, whereas here there's nothing they can do wrong we're just notifying them of why it's not available.  Generally we've used 'text-secondary' for notices like that in both old and new Experimenter, but I'm not opposed to rethinking how we use typography to signal different kinds of messages to users but I would hazard we approach that as a separate body of work.

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
eliserichards: Totally makes sense 😄

eliserichards (approved):
LGTM! 🚢 🚀

PR #7143, opened 2022-03-24
fix #7139 bug(nimbus): check min and max supported versions separately in targeting

### LINKED ISSUE #7139: It is not made clear to experiment owners that some mobile version targeting will not be reflected in the experiment recipe
**[Prerequisites]:**

* Have a mobile Focus draft open.

**[Steps to reproduce]:**

1. Open the draft from prerequisites on the “Audience” page.
1. Set the “Min Version” field to “No Version”.
1. Set the “Max Version” field to a value (e.g. Firefox 99).
1. Click “Save and Continue”.
1. Click “Preview Recipe JSON” to observe the “targeting” part of the recipe.

**[Expected result]:**

* Users are notified in some way that only setting a Max Version will not be functional.

**[Actual result]:**

* Users can set only a Max Version, without a Min Version, despite that targeting will not be updated to respect that choice.

**[Notes]:**

* The same issue appears when users try to set a Minimum version <98. They are not informed by the UI that it is not possible. It is especially problematic if they also want to set an acceptable Maximum version (Min 97 & Max 99), as the Maximum version will not be accepted either.

* Attached is a recording of the issue.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-2304)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/22060">max version only.gif</a>

jaredlockhart:
Because

* We recently added checks to include version targeting for mobile based on which specific versions of each app supported version targeting
* We were only checking if the minimum version was supported to determine whether to include version targeting
* This prevented experiments that only target a max version from including version targeting if the max version is above the supported version

This commit

* Separately checks whether each of the min and max versions are supported and includes the relevant targeting string if they are

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -245,22 +245,24 @@ class NimbusExperiment(NimbusConstants, FilterMixin, models.Model):
         if self.application != self.Application.DESKTOP:
             version_key = "app_version"
 
-        version_supported = True
+        min_version_supported = True
+        max_version_supported = True
         if self.application in self.TARGETING_APPLICATION_SUPPORTED_VERSION:
             supported_version = self.TARGETING_APPLICATION_SUPPORTED_VERSION[
                 self.application
             ]
-            version_supported = self.firefox_min_version >= supported_version
+            min_version_supported = self.firefox_min_version >= supported_version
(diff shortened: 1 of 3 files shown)


eliserichards (approved):
Lookin' good! 🚀

PR #7145, opened 2022-03-25
    fix #7142 bug(nimbus): use version comparators for supported version checks

### LINKED ISSUE #7142: Don't compare versions as strings
Now that we have to do version comparisons in several places in Experimenter, and we have versions >100, we can't rely on string comparison, so we'll need a better way to do proper version comparisons.  In legacy we would parse out integers using regexs, but now that we also hard code several minor/patch versions, that probably won't work.  Desktop and mobile have more complex version comparison modules that support special wildcard characters like `!` and `*`, I'm not sure we need to get that fine grained.  I wonder if we can parse them into tuples and compare those?

jaredlockhart:
Because
    
* We recently added checks for the minimum supported version for version targeting on mobile
* Any version comparison needs to use package.version to compare rather than string comparisons
    
This commit
    
* Uses package.version to compare minimum supported mobile targeting versions

### DIFF
--- a/app/experimenter/experiments/models/nimbus.py
+++ b/app/experimenter/experiments/models/nimbus.py
@@ -18,6 +18,7 @@ from django.db.models.constraints import UniqueConstraint
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.text import slugify
+from packaging import version
 
 from experimenter.base import UploadsStorage
 from experimenter.base.models import Country, Locale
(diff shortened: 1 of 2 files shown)


jaredlockhart:
Blocked on https://github.com/mozilla/experimenter/pull/7143

eliserichards (changes requested):
Needs a rebase 🙏

jaredlockhart:
@eliserichards oh yes I wouldn't treat 'needs rebase' as a request for change, if you review just the commit at the top of the stack you can review it that way and then approve or request changes on that, and if they're all approved you can combine the 'update pr' and 'auto merge' buttons to do kind of a fake merge stack.

eliserichards:
Ahh it didn't give me the "update PR" options, only "resolve conflicts".

PR #7154, opened 2022-03-28
fixes #7119 fix(Nimbus): Hide end experiment when experiment is not idle

### LINKED ISSUE #7119: End Experiment should not appear while experiment is not Idle
I just noticed while reviewing an experiment that the 'End Experiment' button was available while I was reviewing an 'End Enrollment' request.  When an experiment is not in the `Idle` publish status, there shouldn't be other actions available.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2270)

anthonytran7757:
Because

* end experiment button should not be available unless the experiment is idle

This commit

* adds a condition to only show the end experiment button when the experiment is idle

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
@@ -89,6 +89,30 @@ describe("Summary", () => {
     await screen.findByTestId("pill-enrolling-complete");
   });
 
+  it("renders the end experiment button if the experiment is live and idle", async () => {
+    render(
+      <Subject
+        props={{
+          status: NimbusExperimentStatusEnum.LIVE,
+          publishStatus: NimbusExperimentPublishStatusEnum.IDLE,
+        }}
+      />,
+    );
+    await screen.findByText("End Experiment");
+  });
+
+  it("does not renders the end experiment button if the experiment is live and not idle", async () => {
+    render(
+      <Subject
+        props={{
+          status: NimbusExperimentStatusEnum.LIVE,
+          publishStatus: NimbusExperimentPublishStatusEnum.APPROVED,
+        }}
+      />,
+    );
+    expect(screen.queryByText("End Experiment")).not.toBeInTheDocument();
+  });
+
   describe("ending an experiment request", () => {
     const origWindowOpen = global.window.open;
     let mockWindowOpen: any;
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -89,6 +89,30 @@ describe("Summary", () => {
...
+    render(
+      <Subject
+        props={{
+          status: NimbusExperimentStatusEnum.LIVE,
+          publishStatus: NimbusExperimentPublishStatusEnum.APPROVED,
+        }}
+      />,
+    );
+    expect(screen.queryByText("End Experiment")).not.toBeInTheDocument();
+  });
```
jaredlockhart, line 114: Fantastic, this very clearly lays out the expectations of the change, very well written 👍

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -89,6 +89,30 @@ describe("Summary", () => {
...
+  it("renders the end experiment button if the experiment is live and idle", async () => {
+    render(
+      <Subject
+        props={{
+          status: NimbusExperimentStatusEnum.LIVE,
+          publishStatus: NimbusExperimentPublishStatusEnum.IDLE,
+        }}
+      />,
+    );
+    await screen.findByText("End Experiment");
```
jaredlockhart, line 101: You don't need to use expect to be in the document here?  Because of the await it'll throw if it can't find it?

app/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -83,10 +82,9 @@ const Summary = ({
           />
         )}
 
-      {status.live && !status.review && !status.endRequested && (
+      {status.live && !status.review && !status.endRequested && status.idle && (
```
jaredlockhart, line 85: I wonder if the review and endRequested checks are redundant now...  Don't change it, just curious and thinking out loud.

jaredlockhart (approved):
Yep, tested locally, everything looks correct, thank you @anthonytran7757 ! 🎉

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
jaredlockhart: Oh I see this is the pattern used in all the other tests.  Yep looks good 👍

PR #7161, opened 2022-03-30
Fixes ios version targeting version to be 98

tarikeshaq:
I mistakingly thought the minimum supported version for the version targeting was 97, when it was actually shipped in 98. This fixes that to update the version to 98.

ref: https://mozilla-hub.atlassian.net/browse/EXP-2331


jaredlockhart (approved):
Oh wonderful thank you for catching that! 🙏

PR #7166, opened 2022-03-31
fix #7165 bug(nimbus): fix experimenter on m1

### LINKED ISSUE #7165: Experimenter won't build on M1
Now that everyone is refreshing to fancy new M1 macbooks, experimenter isn't building.  Looks like the issue is some node packages.  I think it looks like the base image we're using on arm64 has some different setup to x86 and it's complaining.  Let's get this fixed up before more people switch to M1.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2342)

jaredlockhart:
Because

* Experimenter is not building on M1 computers
* Looks like a problem with node-sass

This commit

* Replaces node-sass with sass

### DIFF
--- a/app/experimenter/nimbus-ui/package.json
+++ b/app/experimenter/nimbus-ui/package.json
@@ -96,10 +96,10 @@
     "mini-css-extract-plugin": "0.9.0",
     "mockdate": "^3.0.5",
     "mutationobserver-shim": "^0.3.7",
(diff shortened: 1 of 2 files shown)


eliserichards (approved):
This is working for me! Woohoo! 🎉 🙌 🚀

PR #7187, opened 2022-04-07
fix #7186 chore(project): only copy outcomes from jetstream-config

### LINKED ISSUE #7186: Ignore jetstream code in linting
I just had a local `make check` failure because

1) We load in a dump of the jetstream-config repo before building local containers to capture all the outcome configs
2) This repo may contain python files
3) The linter may reject those python files

So we need to ignore the jetstream repo path in our linting configs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2361)

jaredlockhart:
Because

* We copy jetstream-config repo into the local filesystem to collect the outcome definition TOMLs when we build the Docker containers
* The jetstream-config repo may contain python files which are superfluous to Experimenter's needs
* Those python fails may cause Experimenter's linting and other checks to fail

This commit

* Copies only the outcome TOMLs into the local repo rather than the entire jetstream-config repo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,12 @@ secretkey:
 	openssl rand -hex 24
 
 jetstream_config:
-	curl -LJ -o app/experimenter/outcomes/jetstream-config.zip $(JETSTREAM_CONFIG_URL)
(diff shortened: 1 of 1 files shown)


scholtzan (approved):
looks good to me 👍

PR #7202, opened 2022-04-19
chore(docs): Update install instructions for mac in the README

eliserichards:
Update the readme install instructions for mac:
- installing yarn
- docker install location 
- node version
- configuring vscode workspaces

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,51 +37,61 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or 
 1. Prerequisites
 
     On all platforms:
-    - Install [Docker](https://www.docker.com/)
-    - Install [Node](https://nodejs.org/en/) ^14.0.0
-
+    - Install [Node](https://nodejs.org/en/download/releases/) to match [current version](https://github.com/mozilla/experimenter/blob/main/app/Dockerfile#L29)
+     
     On Linux:
+    - Install [Docker](https://www.docker.com/)
+    - Install [yarn](https://classic.yarnpkg.com/lang/en/docs/install)
     - [Setup docker to run as non-root](https://docs.docker.com/engine/security/rootless/)
 
     On MacOS:
-      - [Install Docker Desktop](https://www.docker.com/products/docker-desktop)
+    - Install [Docker](https://docs.docker.com/desktop/mac/install/)
       - Adjust resource settings
         - CPU: Max number of cores
         - Memory: 50% of system memory
         - Swap: Max 4gb
         - Disk: 100gb+
-
-1.  Clone the repo
+    - Install [yarn](https://github.com/yarnpkg)
+            
+1. Clone the repo
 
         git clone <your fork>
 
-1.  Copy the sample env file
+1. Copy the sample env file
 
         cp .env.sample .env
 
-1.  Set DEBUG=True for local development
+1. Set DEBUG=True for local development
 
         vi .env
 
-1.  Create a new secret key and put it in .env
+1. Create a new secret key and put it in .env
 
         make secretkey
 
-1.  Run tests
+	vi .env
+
+	```
+	...
+	SECRETKEY=mynewsecretkey
+	...
+	```
+
+1. Run tests
 
         make check
 
-1.  Setup the database
+1. Setup the database
 
         make refresh
 
 #### Fully Dockerized Setup (continuation from General Setup 1-7)
 
-1.  Run a dev instance
+1. Run a dev instance
 
         make up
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,51 +37,74 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or
...
         make check
 
-1.  Setup the database
+1. Setup the database
 
         make refresh
 
+1. If using VSCode, configure workspace folders
+
+   - Add `app` folder to your workspace (File -> Add Folder to Workspace -> `path/to/experimenter/app`)
```
jaredlockhart, line 91: You have to add both folders to workspace:

`/experimenter/`
`/experimenter/app`

README.md
```diff
@@ -37,51 +37,74 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or
 1. Prerequisites
 
     On all platforms:
-    - Install [Docker](https://www.docker.com/)
-    - Install [Node](https://nodejs.org/en/) ^14.0.0
-
+    - Install [Node](https://nodejs.org/en/) ^14.0.0 <17.0 (recommended: 16.9.1)
```
jaredlockhart, line 40: We might want to keep [these consistent](https://github.com/mozilla/experimenter/blob/main/app/Dockerfile#L29)?

jaredlockhart (changes requested):

README.md
```diff
@@ -37,51 +37,74 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or
...
+   - Add `app` folder to your workspace (File -> Add Folder to Workspace -> `path/to/experimenter/app`)
+   - From the `app` folder, run `yarn install`
+       - Make sure you are using the correct version of node
+          
+           node -v
+
+       - Troubleshooting: 
+           - [Changing node version](https://stackoverflow.com/a/50817276/12178648)
+           - Clear npm cache: `npm cache clean --force`
+
```
mikewilli, line 100: A couple thoughts, feel free to take or leave...

Later in the README (step 2 under Semi Dockerized Setup Steps) is where we currently mention `yarn install`, would it be better to put this block there? Or at least everything under the second item about running `yarn install` and troubleshooting?
- If this is moved, then also don't need the `brew install yarn` for MacOS Prereqs
- If this is _not_ moved, we should add a link to [yarn's install guide](https://classic.yarnpkg.com/lang/en/docs/install/) to the Linux prereqs section

The VSCode workspace configuration could potentially be its own section if we want a spot to put general IDE setup tips.

mikewilli (approved):
Made a couple suggestions, but overall looks good to me!

README.md
eliserichards: Ahh that's a good point. I figured this out when running `yarn install` -- it gave me an error that it needed to be less than 17.0. Keeping it consistent with that version works better

README.md
```diff
@@ -37,51 +37,74 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or
 1. Prerequisites
 
     On all platforms:
-    - Install [Docker](https://www.docker.com/)
```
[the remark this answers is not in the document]
eliserichards: Ooo good to know!

README.md
eliserichards: I haven't used the semi-dockerized instructions, so it might be better to pull this `yarn` stuff out into its own section. I was the same thing about the vscode info too 🤔

README.md
eliserichards: blah I don't like my wording. thoughts?

README.md
```diff
@@ -37,51 +37,61 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or
...
-    - Install [Node](https://nodejs.org/en/) ^14.0.0
-
+    - Install [Node](https://nodejs.org/en/download/releases/) to match [current version](https://github.com/mozilla/experimenter/blob/main/app/Dockerfile#L29)
```
jaredlockhart, line 40: Nah that's fine that's good 👍

README.md
```diff
@@ -37,51 +37,61 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or
...
+     
     On Linux:
+    - Install [Docker](https://www.docker.com/)
+    - Install [yarn](https://classic.yarnpkg.com/lang/en/docs/install)
```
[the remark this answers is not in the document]
eliserichards: I'll chat about this with Jared and then check in with you if we make those changes 👍

PR #7206, opened 2022-04-21
fix #7204 chore(nimbus): remove unique constraint on featureconfig name

### LINKED ISSUE #7204: Remove unique constraint on feature config name
Now that feature configs are unique by (application, slug) we don't need the name to be globally unique, and since we'll be loading feature configs from multiple sources for mobile applications they may collide so we should remove that constraint.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2383)

jaredlockhart:
Because

* We will now be loading feature configs from multiple repos
* We uniquely identify a feature config by the pair of slug and application
* There's no way to prevent feature configs in different repos using the same name

This commit

* Removes the uniqueness constraint on feature config name


yashika-khurana (approved):
LGTM!

PR #7208, opened 2022-04-21
 fix #7207 chore(nimbus): remove python scripts from jetstream config

### LINKED ISSUE #7207: Builds broken locally while importing outcomes
We recently changed the way we loud outcomes in #7186 to prevent python scripts in the jetstream config repo from sneaking into the Experimenter codebase and failing linting, but it turns out the fix worked in Circle but not locally.  We need to take another stab at this.

jaredlockhart:
Because

 * Jetstream config added some python scripts recently
 * We import the jetstream config repo to load outcomes
 * The jetstream config python scripts were causing experimenter's linting to fail
 * We landed an attempt a fix that ended up working in circle but not locally

 This commit

 * Reverts the previous attempt at a fix
 * Removes the python scripts explicitly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,12 +63,9 @@ secretkey:
 	openssl rand -hex 24
 
 jetstream_config:
-	curl -LJ -o /tmp/jetstream-config.zip $(JETSTREAM_CONFIG_URL)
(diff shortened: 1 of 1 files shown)


yashika-khurana (approved):
LGTM!

PR #7216, opened 2022-04-22
fix #7215 bug(nimbus): handle existing features with different names

### LINKED ISSUE #7215: Look up existing feature configs by slugs only
In our feature loading script, we `get_or_create` features by `(name, slug, applicationSlug)`, but if a feature already exists with the same slug but a different name it'll try to create a new one with the new name but the same slug which will violate uniqueness constraint.  So we should look up only by `(slug, applicationSlug)` and set the name at save time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2389)

jaredlockhart:
Because

* A feature config may already exist in the database when attempting to load from a remote feature manifest
* If the existing feature has the same slug but a different name then the loader would attempt to create a new feature and then violate the slug uniqueness constraint

This commit

* Looks existing features up only by (slug, applicationSlug) and then resets the name

### DIFF
--- a/app/experimenter/features/management/commands/load_feature_configs.py
+++ b/app/experimenter/features/management/commands/load_feature_configs.py
@@ -16,10 +16,10 @@ class Command(BaseCommand):
 
         for feature in Features.all():
             feature_config, _ = NimbusFeatureConfig.objects.get_or_create(
-                name=feature.slug,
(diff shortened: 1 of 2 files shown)


yashika-khurana (approved):
LGTM!

PR #7226, opened 2022-04-26
fix bug 1766436 feat(nimbus): add "need pin" targeting for core activity level

Mardak:
Like #6723 but for core. r?@jaredlockhart  https://bugzilla.mozilla.org/show_bug.cgi?id=1766436

Results in targeting like…
```js
(
  userMonthlyActivity|length >= 21 &&
  doesAppNeedPin ||
  experiment.slug in activeExperiments
) &&
(browserSettings.update.channel == "release") &&
('app.shield.optoutstudies.enabled'|preferenceValue) &&
(version|versionCompare('98.!') >= 0) &&
(locale in ['ach', 'af', 'sq']) &&
(region in ['AF', 'AL', 'AX'])
```

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -335,7 +335,7 @@ TARGETING_INFREQUENT_USER_NEED_PIN = NimbusTargetingConfig(
     name="Infrequent user (need pin)",
     slug="infrequent_user_need_pin",
     description="Between 1 and 6 days of activity in the past 28 days needing pin",
-    targeting="({infrequent_user} && doesAppNeedPin || {sticky})".format(
+    targeting="{infrequent_user} && doesAppNeedPin || {sticky}".format(
         infrequent_user=TARGETING_INFREQUENT_USER_URIS.targeting,
         sticky=TARGETING_STICKY,
     ),
@@ -347,7 +347,7 @@ TARGETING_INFREQUENT_WIN_USER_NEED_PIN = NimbusTargetingConfig(
     name="Infrequent Windows user (need pin)",
     slug="infrequent_windows_user_need_pin",
     description="Between 1 and 6 days of activity in the past 28 days needing pin on Win",
-    targeting="({infrequent_user} && os.isWindows && doesAppNeedPin || {sticky})".format(
+    targeting="{infrequent_user} && os.isWindows && doesAppNeedPin || {sticky}".format(
         infrequent_user=TARGETING_INFREQUENT_USER_URIS.targeting,
         sticky=TARGETING_STICKY,
     ),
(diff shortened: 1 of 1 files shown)


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -432,13 +432,24 @@ class NimbusTargetingConfig:
...
     "(currentDate|date - profileAgeCreated|date) / 86400000 >= 28",
     desktop_telemetry="",
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_CORE_USER_NEED_PIN = NimbusTargetingConfig(
+    name="Core user (need pin)",
+    slug="core_user_need_pin",
+    description="At least 21 days of activity in the past 28 days needing pin",
+    targeting="({core_user} && doesAppNeedPin || {sticky})".format(
```
Mardak, line 446: I see #7120 wraps with parens, so maybe I should drop these (and the other places where we explicitly added them?)

jaredlockhart: Yeah you don't need the outermost parens, those'll get added later, but do you want parents around the `&&` clause so both operands are `||`ed with the sticky clause?

Mardak: I removed the unnecessary parens from the activity level targeting

Mardak:
In general, there can be some odd behavior of `|| sticky` if used within another advanced targeting. The simplest error is the operator precedence mixup, e.g., `a || sticky` reused as `{a_sticky} && b` incorrectly expands to `a || sticky && b`

But even wrapping it as `(a || sticky) && b` has a separate odd behavior as `b` can cancel the stickiness, e.g., if `b` was "today is Monday" results in non-sticky unenrolling on Tuesday. It seems like sticky might want to be immediately outside advanced targeting, e.g., `a && b || sticky`. But not too far out that it would override other targeting like max version.

app/experimenter/experiments/constants/nimbus.py
Mardak: To be clear, the `core && pin || sticky` is the same as `(core && pin) || sticky`, so parens aren't necessary but can be added. Most of the activity level targeting does not have the unnecessary parens.

jaredlockhart:
> In general, there can be some odd behavior of `|| sticky` if used within another advanced targeting. The simplest error is the operator precedence mixup, e.g., `a || sticky` reused as `{a_sticky} && b` incorrectly expands to `a || sticky && b`
> 
> But even wrapping it as `(a || sticky) && b` has a separate odd behavior as `b` can cancel the stickiness, e.g., if `b` was "today is Monday" results in non-sticky unenrolling on Tuesday. It seems like sticky might want to be immediately outside advanced targeting, e.g., `a && b || sticky`. But not too far out that it would override other targeting like max version.

Oooohhh very interesting @Mardak thank you for the deep dive.  We've had a backlogged ticket to pull sticky out into its own targeting option rather than treating it as an artifact of advanced targeting, we should probably do that soon and take into account the behaviours you're describing.

jaredlockhart (approved):
Looks good 👍

PR #7227, opened 2022-04-26
Fixes #7167 chore(reporting): Deprecate existing reporting infrastructure

### LINKED ISSUE #7167: Deprecate existing reporting infrastructure
We added a `reporting` module last year as a way to gather information on both Normandy and Nimbus experiments and provide it in CSV to incorporate into executive reporting dashboards.  With the continued ramp down and eventual deprecation of Normandy, this infrastructure is no longer in use and can be safely removed, as confirmed by @shell1 and @tiftran.  We are currently developing a new reporting infrastructure that will replace this functionality with a focus on Nimbus experiments.

Let's deprecate the current reporting infrastructure.  Here are some places to look for code that can be removed:

- [x] https://github.com/mozilla/experimenter/blob/main/Makefile#L29-L31
- [x] https://github.com/mozilla/experimenter/blob/main/Makefile#L35
- [x] https://github.com/mozilla/experimenter/blob/main/app/Dockerfile#L67-L68
- [x] https://github.com/mozilla/experimenter/blob/main/app/Dockerfile#L86-L87
- [x] https://github.com/mozilla/experimenter/blob/main/app/Dockerfile#L115
- [x] https://github.com/mozilla/experimenter/blob/main/docker-compose.yml#L21
- [x] https://github.com/mozilla/experimenter/blob/main/docker-compose.yml#L38
- [x] https://github.com/mozilla/experimenter/blob/main/docker-compose-legacy.yml#L14
- [x] https://github.com/mozilla/experimenter/blob/main/docker-compose-legacy.yml#L19-L32
- [x] https://github.com/mozilla/experimenter/blob/main/app/package.json#L8
- [ ] https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L84
- [x] https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L108
- [x] https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L194
- [x] https://github.com/mozilla/experimenter/blob/main/app/experimenter/settings.py#L348-L351
- [x] https://github.com/mozilla/experimenter/blob/main/app/experimenter/urls.py#L11
- [x] https://github.com/mozilla/experimenter/blob/main/app/experimenter/urls.py#L19
- [x] https://github.com/mozilla/experimenter/blob/main/app/experimenter/urls.py#L24
- [ ] https://github.com/mozilla/experimenter/tree/main/app/experimenter/reporting

Note: We might not be able to remove the `reporting` django app altogether because it has database migrations, it might be necessary to keep the app and its `models.py` file in place, but empty, with a new migration that deprecates the tables, and then remove the app and its `models.py` entirely as a separate step.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2344)

eliserichards:
Because
* we are no longer using the legacy reporting infrastructure 

This commit
* removes all of the current now-deprecated reporting models
* maintains the db migration that removes them
* keeps the `reporting` folder with empty `models.py` and the additional db migration (which will be removed separately)
   ```
       reporting/
           ├── models.py
           └── migrations/
               ├── ...
               └── 0005_delete_reportlog.py
   ```


eliserichards:
Looks like it's complaining about test coverage missing on 2 lines: 

```experimenter/normandy/client.py                                        46      2    96%   73-74```

This is the same when running test cov locally (worth a try to see if it was only failing on CI 😛 ). What do we usually do in this case?

eliserichards:
This needs a test for client.py ✍️

jaredlockhart (changes requested):
-10k.  We have a word for that, it's 🧁 

Okay so this looks like it removes everything but keeps the migrations and models file in place to remove the model.  That looks exactly correct to me.

Just played around with it locally, couple small changes:

For the migration to run, `experimenter.reporting` still needs to be in `INSTALLED_APPS` in `settings.py` so we have to keep that in there and remove it later. 

So put that back in and I'll play with it again, but broadly I think this all looks correct, great job! 🎉

eliserichards:
Just added it back to `INSTALLED_APPS` and rebased and it went from 10k to 1k... _panics_

jaredlockhart (approved):
Fantastic, everything looks right to me, migration runs successfully.  You might have to regenerate the migration after you rebase on main, maybe not.  CI will know.  Great job @eliserichards 👏 👏 👏 👏

PR #7228, opened 2022-04-26
Fixes #7159 feat(nimbus): Include target_configs in api endpoint 

### LINKED ISSUE #7159: Expose targeting config in API
Each targeting config has a description field but we don't show it anywhere on the audience form or on the detail/summary page.  We should show it in the drop down if we can finagle it into the react-select item component, and we should add it on the detail page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2328)

yashika-khurana:
- Because
We ask the user to select the `Targeting configs` and currently we are only returning the name of `targeting_configs` as `target_config_slug`. 
- Change
    In this PR following changes has been done:
    - To return the `target_configs` as part of API in `getConfig` and `target_config` as part of `experimentBySlug` API
    - If there is no `target_config_slug` selected, it will return the default Targeting config options which is No Targeting config, else the selected one including label, value, description and application values
   - Includes both frontend and backend tests

Refer to screenshot below for the api response: 
- `nimbusConfig` api response returning targeting configs 
image

 - `experimentBySlug` api response if targeting config slug is selected by the user
image

- `experimentBySlug` api response if targeting config slug is not selected by the user (will return default Targeting config- No Targeting option, case: when creating new experiment)
image

fixes #7159


app/experimenter/experiments/api/v5/types.py
```diff
@@ -302,6 +303,7 @@ def resolve_targeting_configs(root, info):
                 application_values=NimbusExperiment.TARGETING_CONFIGS[
                     choice.value
                 ].application_choice_names,
+                description=NimbusExperiment.TARGETING_CONFIGS[choice.value].description,
```
eliserichards, line 306: Possible nit: is it normal to not put spaces around the = sign? I've seen this format in a few places but I'm not sure where it should be used 😅

app/experimenter/experiments/api/v5/types.py
```diff
@@ -405,6 +408,39 @@ def resolve_targeting_config_slug(self, info):
...
+                    value=choice.value,
+                    application_values=NimbusExperiment.TARGETING_CONFIGS[
+                        choice.value
+                    ].application_choice_names,
+                    description=NimbusExperiment.TARGETING_CONFIGS[
+                        choice.value
+                    ].description,
+                )
+                for choice in NimbusExperiment.TargetingConfig
+            ]
```
eliserichards, line 442: Is it possible to remove some duplicate code here?

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1005,6 +1005,99 @@ def test_targeting_config_slug_for_valid_targeting_config_returns_name(self):
             NimbusExperiment.TargetingConfig.TARGETING_FIRST_RUN.value,
         )
 
+    def test_targeting_configs(self):
```
eliserichards, line 1008: 👏

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1005,6 +1005,99 @@ def test_targeting_config_slug_for_valid_targeting_config_returns_name(self):
...
+                    "applicationValues": list(
+                        NimbusExperiment.TARGETING_CONFIGS[
+                            NimbusExperiment.TargetingConfig.TARGETING_FIRST_RUN.value
+                        ].application_choice_names
+                    ),
+                }
+            ],
+        )
+
+    def test_targeting_configs_with_empty_targeting_config_slug(self):
```
eliserichards, line 1054: 👏

eliserichards (commented):
This looks fantastic! Just a few comments and then we can 🚢

app/experimenter/experiments/api/v5/types.py
```diff
@@ -405,6 +408,39 @@ def resolve_targeting_config_slug(self, info):
             return self.TargetingConfig(self.targeting_config_slug).value
         return self.targeting_config_slug
 
+    def resolve_targeting_configs(self, info):
```
jaredlockhart, line 411: Drive by sick review 😷 

This looks great except for one thing:

This should be in the `getConfig` query, not `getExperimentBySlug`, since it's all targeting configs, not just the one specific to this experiment.

yashika-khurana: The field is added both in `getConfig` and `getExperimentBySlug`, the idea is when we set the target config slug to the experiment then it will return that particular targeting config, else will return list of targeting config

yashika-khurana: so should we restrict to `getConfig` or can we keep in both?

app/experimenter/experiments/api/v5/types.py
jaredlockhart: All the formatting is automatic, you don't have to worry about it ;)

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Oh yeah you updated the `getConfig` query, so you don't need the second case (lines 428-442) here.  The NimbusExperimentType only needs to serialize the data for a single experiment.  And instead of looping and matching the slug you can just instantiate it with the value the same way you did on line 306.

jaredlockhart, line 411: Also calling this method `resolve_targeting_configs` doesn't make sense on this object since a single experiment can only have a single targeting config, it should be singular.

yashika-khurana:
@eliserichards  @jaredlockhart  So now `experimentBySlug` returns `targeting_conifg` whereas `getConfig` API return all targeting fields as `targeting_configs`.

eliserichards:
This looks good to me! I'm going to wait for @jaredlockhart to give the final ✅ just in case there's something I missed 🙌 🚀 🚢

jaredlockhart (approved):
Fantastic, looks great 👍

app/experimenter/experiments/api/v5/types.py
```diff
@@ -338,6 +340,7 @@ class NimbusExperimentType(DjangoObjectType):
     documentation_links = DjangoListField(NimbusDocumentationLinkType)
     treatment_branches = graphene.List(NimbusBranchType)
     targeting_config_slug = graphene.String()
+    targeting_config = graphene.List(NimbusExperimentTargetingConfigType)
```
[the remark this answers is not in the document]
yashika-khurana: correct, that's why I just added additional stuff, I will file a issue for the refactoring of this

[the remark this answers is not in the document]
yashika-khurana: here it is https://github.com/mozilla/experimenter/issues/7251

PR #7229, opened 2022-04-27
fix #6639 feat(nimbus): add import/export to nimbus experiment admin page

### LINKED ISSUE #6639: Support exporting/importing experiments JSON through admin
It would be nice to be able to export an experiment to JSON from the admin on some instance (like prod), and then be able to load it into another instance (like local) to make it easier to reproduce prod issues locally.  Can either piggy back on `dumpdata` `loaddata` management commands, or maybe there's a library to do this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1867)

mikewilli:
Because

* it is helpful for development and testing to be able to copy experiments from prod to dev

This commit

* adds the ability to export selected experiments to a file
* adds the ability to import experiments from a file

Other Info
* "experiment" here refers to Nimbus Experiments
* Note that import will perform an update if the experiment identifier matches an existing experiment
* We are not concerned about exact matches for all nested objects, so in some cases using a dev user is acceptable

Export Context
* Admin page Action (available by selecting the desired experiments and selecting the export option in the dropdown)
* Should produce valid output in the selected format, including the experiment info + `changes` (list/array), `branches` (list/array), and `reference_branch_slug` (string)

Import Context
* Should be able to take the output file produced by Export and create/update an experiment, including change logs and branches
* Available via Nimbus Experiments admin page (button in upper right)
* Import button directs to import form with dropdown list of file type options
    * Select the file and type that you exported previously

Screenshot of additions to the Nimbus Experiment Admin page (see Action selection and Import button):
image


jaredlockhart:
@mikewilli Awesome!!  Hooray for libraries that do stuff for us 🎉 

Looks like there's a hiccup though, some fields use the Python type `Decimal`, and it looks like it's choking on that:

```
RepresenterError at /admin/experiments/nimbusexperiment/

('cannot represent an object', Decimal('100.0000'))
```

I just did a `make refresh up` and tried to export one of the dummy experiments.  See if you can poke around at it and find a simple solution, if nothing jumps out at you as a simple solution let's look at it together and see if we can't find something that doesn't involve breaking this thing to bits.

mikewilli:
@jaredlockhart Where are you seeing the RepresenterError? I tried `make refresh up` followed by exporting an experiment without seeing any errors, and it looks like the export is representing Decimal fields ok in the output (they're in the JSON as strings - I checked `population_percent` which should be a Decimal field). I also tried exporting all 6 dummy experiments, and import them back, and everything seemed to work fine.

Let's look together to try and figure out what the difference is -- once I can reproduce the error I'll work on resolving it.

jaredlockhart:
@mikewilli Aha, I see the problem.  Export to JSON works, but export to YAML doesn't.  Is there any way to restrict the output formats?

mikewilli:
Ah that's my bad, I did not test all formats. There's an [open bug](https://github.com/django-import-export/django-import-export/issues/1324) on this with a suggested fix/workaround that I will try. I'll also look into whether we can limit the output formats.

jaredlockhart:
@mikewilli Yeah so the thing I was worried about is this exports/imports the top level of the object but not any of the foreign keyed objects, which are necessary.  I was hoping this library might do it for us automagically.  Can you see if there's any way for this thing to easily serialize related objects?

mikewilli:
@jaredlockhart Yep I'll look into that. I think I remember seeing something in the docs about it.

I haven't found an easy way to limit the export file formats, but I was able to implement that workaround I mentioned. The workaround is to override the DecimalWidget's `render` function (which is what converts the Decimal to the format in the output representation) to just return `str(value)`. That seems to be what is happening with the other output formats anyway, but let me know if you think that would present any problems.

jaredlockhart:
Yeah `str(value)` should be okay since that'll become a JSON `Number` and ideally get reparsed.  It's not super critical since this is only for a debugging workflow.

jaredlockhart (changes requested):
Let's see if we can figure out a quick/easy way to include related objects in the serialization and if nothing's immediately doable we'll land it like this and figure it out later.

jaredlockhart:
@mikewilli You might have to do this to get circle working again:

But if you have hit that button, and you use CircleCI, you may be noticing some problems.  The fix appears relatively simple:

    Log out of CircleCI
    Clear all CircleCI related cookies
    Log back into CircleCI

That appears to resolve the issues. 
People have seen 404's on CircleCI pages, PR's not processing, not having permissions to restart jobs, etc.  But again, clearing cookies has resolved this for them.

app/experimenter/experiments/admin/nimbus.py
```diff
@@ -22,6 +30,122 @@ def force_fetch_jetstream_data(modeladmin, request, queryset):
...
+        model = NimbusExperiment
+        exclude = ("id", "reference_branch")
+        import_id_fields = ("slug",)
+
+    def dehydrate_changes(self, experiment):
+        all_changes = experiment.changes.all()
+
+        changelogs = []
+        for c in all_changes:
+            changelogs.append(dict(NimbusChangeLogSerializer(c).data))
```
jaredlockhart, line 84: I'm not a huge fan of single letter variable names outside of comprehensions, but that's just me.

app/experimenter/experiments/admin/nimbus.py
```diff
@@ -22,6 +30,122 @@ def force_fetch_jetstream_data(modeladmin, request, queryset):
...
+        branches = []
+        for c in all_branches:
+            branches.append(dict(NimbusBranchChangeLogSerializer(c).data))
+
+        return branches
+
+    def dehydrate_reference_branch_slug(self, experiment: NimbusExperiment):
+        if experiment.reference_branch is not None:
+            return experiment.reference_branch.slug
+        return None
```
jaredlockhart, line 100: This is superfluous, it'll return `None` by default, but there's no harm in being explicit either.

app/experimenter/experiments/admin/nimbus.py
```diff
@@ -22,6 +30,122 @@ def force_fetch_jetstream_data(modeladmin, request, queryset):
...
+        for c in all_changes:
+            changelogs.append(dict(NimbusChangeLogSerializer(c).data))
+
+        return changelogs
+
+    def dehydrate_branches(self, experiment):
+        all_branches = experiment.branches.all()
+
+        branches = []
+        for c in all_branches:
```
jaredlockhart, line 92: I would've said `for branch in all_branches`

app/experimenter/experiments/tests/test_admin/test_admin_nimbus.py
```diff
@@ -145,3 +159,98 @@ def test_admin_force_jetstream_fetch(self, mock_fetch_experiment_data):
...
+            widget=NimbusBranchForeignKeyWidget(NimbusBranch, "slug"),
+        )
+        experiment = NimbusExperimentFactory.create()
+        experiment.id = None
+        experiment.reference_branch = None
+        test_row = {"slug": experiment.slug, "reference_branch_slug": "control"}
+        try:
+            reference_branch_slug.widget.get_queryset(value=None, row=test_row)
+        except ValueError:
+            self.fail("get_queryset raised an Exception!")
```
jaredlockhart, line 183: Oh that's a fun little pattern, I would've just made the call and let the exception be the failure case, but this way you get to add a nice little message.  Interesting.

jaredlockhart: Well I guess the flip side is, the exception might contain some useful information that gets elided here, maybe rather than wrapping it just let the exception fail the test itself.

app/experimenter/experiments/tests/test_admin/test_admin_nimbus.py
```diff
@@ -145,3 +159,98 @@ def test_admin_force_jetstream_fetch(self, mock_fetch_experiment_data):
...
+
+    def test_resource_get_diff_headers(self):
+        ner = NimbusExperimentResource()
```
jaredlockhart, line 186: Ah I see `ner` == `nimbus experiment resource`, I would've just called it `resource`

jaredlockhart (changes requested):
Fantastic!!  Everything looks broadly right to me, bunch of nitty code style rambling, but nothing significant.

I think the user association issue needs some attention though, if I do the following

```
make refresh up
<export experiment>
make refresh SKIP_DUMMY=1 up
<import experiment>
```

then the random factory user created in the first pass won't exist in the db in the second pass which is roughly analagous to some real user in prod not existing in local dev, and it fails with

```
Errors

    Line number: 1 - User matching query does not exist. 
```

so probably do need to either

1) get_or_create the Auth.User at import time
2) force override the user with one that will exist in local dev ie `dev@example.com`

to get around it.

But this is great, thank you @mikewilli for taking the time to dig into this this is gonna save so much hassle in the future 🙏 🙏 🙏 🙏

app/experimenter/experiments/admin/nimbus.py
mikewilli: I was getting an error on import when trying to access `slug` from None if `experiment.reference_branch` didn't exist.

app/experimenter/experiments/tests/test_admin/test_admin_nimbus.py
mikewilli: Works for me

app/experimenter/experiments/tests/test_admin/test_admin_nimbus.py
mikewilli: Yea I agree, especially in this case where the custom message is not really adding any value.

app/experimenter/experiments/admin/nimbus.py
```diff
@@ -22,6 +32,124 @@ def force_fetch_jetstream_data(modeladmin, request, queryset):
...
+
+    class Meta:
+        model = NimbusExperiment
+        exclude = ("id", "reference_branch")
+        import_id_fields = ("slug",)
+
+    def dehydrate_changes(self, experiment):
+        return [dict(NimbusChangeLogSerializer(c).data) for c in experiment.changes.all()]
+
+    def dehydrate_branches(self, experiment):
```
eliserichards, line 84: What does `dehydrate` refer to here?

eliserichards (commented):
Code's looking good 🔥 I'm going to do some manual testing as well. Would you mind adding a little more context about what the import feature should do (file types, etc), and maybe screenshots in the description if applicable?

app/experimenter/experiments/admin/nimbus.py
mikewilli: `dehydrate_*` functions come from the `django-import-export` library, used to tell the ModelResource how to populate custom fields. See [here](https://django-import-export.readthedocs.io/en/latest/getting_started.html#advanced-data-manipulation-on-export) for an example in the docs.

eliserichards (approved):
Wonderful! 💯 Thanks for adding some context to the description @mikewilli 😄

jaredlockhart:
Thanks for getting this landed @eliserichards and @mikewilli ! 🎉 🎉 🎉 🎉 🎉

PR #7230, opened 2022-04-27
fix #7225 feat(nimbus): Show target config field data on summary page

### LINKED ISSUE #7225: Show targeting config field in Summary/details page
User can select the targeting config when setting up the experiment, but currently we don't show the selected targeting config in summary or details page, so in this task add that field along with its description

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2392)

yashika-khurana:
Because

*  We ask the user to set the `Advanced Targeting` option when creating the experiment, but we don't show that value back anywhere

This commit

* Adds the `slug of targeting config` with its `description`
* Updates the test cases to reflect this new field added in the table

Sample screenshot for the new field added on `Details page`

image

closes #7225


app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1005,6 +1005,98 @@ def test_targeting_config_slug_for_valid_targeting_config_returns_name(self):
...
+                        NimbusExperiment.TARGETING_CONFIGS[
+                            NimbusExperiment.TargetingConfig.NO_TARGETING.value
+                        ].application_choice_names
+                    ),
+                    "description": NimbusExperiment.TARGETING_CONFIGS[
+                        NimbusExperiment.TargetingConfig.NO_TARGETING.value
+                    ].description,
+                }
+            ],
+        )
```
jaredlockhart, line 1098: This test is redundant since it's just testing a particular value rather than a special case, you can take this out I think.

yashika-khurana: I added this to show if we don't have slug, in that case we get the default t`arget_config  as NO_TARGETING`, should we remove it?

app/experimenter/nimbus-ui/src/components/Summary/TableOverview/index.tsx
```diff
@@ -116,6 +116,18 @@ const TableOverview = ({
                 )}
               </td>
             </tr>
+            <tr>
+              <th>Advanced Targeting</th>
+              <td data-testid="experiment-targeting-config">
+                {experiment.targetingConfig?.length ? (
+                  experiment.targetingConfig.map((t) => (
+                    <p key={t?.label}>{t?.label + "- " + t?.description}</p>
```
jaredlockhart, line 124: You need a space at the start of the dash, and also in general, for both js and python, instead of string concatenation I'd tend to prefer format strings like so

```js
{`${t?.label} - ${t?.description}`}
```

jaredlockhart (changes requested):
Looks good! Just a small change 🙏

app/experimenter/experiments/tests/api/v5/test_queries.py
jaredlockhart: Yeah, if the field defaulted to `None` then we'd be concerned about handling that case correctly since the dict key lookup would fail, but since the field defaults to a valid choice and isn't nullable, we never need to worry about that case so we can get away with the single test 👍

yashika-khurana: got it makes sense :)

jaredlockhart (approved):
Ah okay, we should try to organize these sub branch prs a little better next time 😅 

Looks good 👍

PR #7233, opened 2022-04-28
Fixes #7232 fix(project): Test locale factory should reference locale instead of country

### LINKED ISSUE #7232: Test locale factory should reference locale instead of country
https://github.com/mozilla/experimenter/blob/43e7d22bd47e3ec4ac0f97dbb64d990c4aa86ba5/app/experimenter/base/tests/factories.py#L19 should reference `faker.locale()`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2396)

eliserichards:
Because...

* The locale factory was incorrectly referencing `country`

This commit...

* Uses the correct field, `locale`, instead

### DIFF
--- a/app/experimenter/base/tests/factories.py
+++ b/app/experimenter/base/tests/factories.py
@@ -16,7 +16,7 @@ class CountryFactory(factory.django.DjangoModelFactory):
 
 
 class LocaleFactory(factory.django.DjangoModelFactory):
-    name = factory.LazyAttribute(lambda o: faker.country())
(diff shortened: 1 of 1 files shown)


yashika-khurana (approved):
LGTM ! 🥳

jaredlockhart:
Oh interesting catch @eliserichards.  I was trying to think if this might break some validation, but I don't think it will.  Thanks for catching!

PR #7245, opened 2022-05-02
fix #7244 chore(nimbus): add targeting for Fx100+, Win10+

### LINKED ISSUE #7244: Add custom targeting for Fx100+, Windows 10+
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2399)

ncalexan:
Because

* Some experiments need to target Firefox 100+ but only apply to
  Windows 10+, and in fact only build 15063 plus.

This commit

* Adds targeting for Fx100+, Windows build number 15063+.

Notes:

* The build number was chosen because the `UserChoice` Windows feature
  changed in that build, see [this
  comment](https://searchfox.org/mozilla-central/rev/87ecd21d3ca517f8d90e49b32bf042a754ed8f18/browser/components/shell/WindowsUserChoice.cpp#86).

* It's not clear to me why there's no version comparison in the
  `desktop_telemetry` section of my model,
  `TARGETING_FX95_DESKTOP_USERS`, but I followed suit.

Fix #7244

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -206,6 +206,32 @@ TARGETING_FIRST_RUN_WINDOWS_1903_NEWER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_WINDOWS_WITH_USERCHOICE = NimbusTargetingConfig(
+    name="Users on Windows with UserChoice support (Windows build 15063 or newer)",
+    slug="windows_userchoice",
+    description="Users on Windows with UserChoice support (i.e., Windows build 15063+)",
+    targeting="os.windowsBuildNumber >= 15063",
+    desktop_telemetry="environment.system.os.windows_build_number >= 15063",
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_WINDOWS_WITH_USERCHOICE_FIRST_RUN = NimbusTargetingConfig(
+    name="First start-up users on Windows with UserChoice support "
+    "(Windows build 15063 or newer)",
+    slug="windows_userchoice_first_run",
+    description="First start-up users (e.g. for about:welcome) on Windows with "
+    "UserChoice support (i.e., Windows build 15063+)",
+    targeting="{first_run} && {user_choice}".format(
+        first_run=TARGETING_FIRST_RUN.targeting,
+        user_choice=TARGETING_WINDOWS_WITH_USERCHOICE.targeting,
+    ),
+    desktop_telemetry=("{first_run} AND {user_choice}").format(
+        first_run=TARGETING_FIRST_RUN.desktop_telemetry,
+        user_choice=TARGETING_WINDOWS_WITH_USERCHOICE.desktop_telemetry,
+    ),
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
 TARGETING_FX95_DESKTOP_USERS = NimbusTargetingConfig(
     name="Desktop Users on Fx95",
     slug="fx95_desktop_users",
@@ -728,6 +754,10 @@ class NimbusConstants(object):
         TARGETING_MAC_ONLY.slug: TARGETING_MAC_ONLY,
         TARGETING_NO_ENTERPRISE.slug: TARGETING_NO_ENTERPRISE,
         TARGETING_FX95_DESKTOP_USERS.slug: TARGETING_FX95_DESKTOP_USERS,
+        TARGETING_WINDOWS_WITH_USERCHOICE.slug: TARGETING_WINDOWS_WITH_USERCHOICE,
+        TARGETING_WINDOWS_WITH_USERCHOICE_FIRST_RUN.slug: (
+            TARGETING_WINDOWS_WITH_USERCHOICE_FIRST_RUN
+        ),
         TARGETING_MOBILE_NEW_USER.slug: TARGETING_MOBILE_NEW_USER,
         TARGETING_MOBILE_RECENTLY_UPDATED.slug: TARGETING_MOBILE_RECENTLY_UPDATED,
         TARGETING_NO_ENTERPRISE_OR_PAST_VPN.slug: TARGETING_NO_ENTERPRISE_OR_PAST_VPN,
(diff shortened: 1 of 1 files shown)


ncalexan:
I'd appreciate a close review of this, since the meat of the change is string-ly typed and I don't really know how to test this in any meaningful way.  Could somebody explain why my model (see commit message) doesn't appear to have the `desktop_telemetry` field spelled out?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -206,6 +206,21 @@ class NimbusTargetingConfig:
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_FIRST_RUN_WINDOWS_10_NEWER_FX100_NEWER = NimbusTargetingConfig(
+    name="Fx100+ first start-up users on Windows 10 Creators (build 15063) or newer",
+    slug="first_run_win10_fx100",
+    description="Fx100+ first start-up users (e.g. for about:welcome) on Windows 1507+",
+    targeting=("{first_run} "
+               "&& os.windowsBuildNumber >= 15063 "
```
Mardak, line 214: Version 1507 seems to have a build number of 10240 according to https://en.wikipedia.org/wiki/Windows_10_version_history

But if you generally want Windows 10 (10+ including 11? and not necessarily a build number), you can use `os.windowsVersion >= 10` https://searchfox.org/mozilla-central/rev/ea1234192518e01694a88eac8ff090e4cadf5ca4/toolkit/components/utils/ClientEnvironment.jsm#189,213,225

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -206,6 +206,21 @@ class NimbusTargetingConfig:
...
+               "&& (version|versionCompare('100.!') >= 0)"
```
Mardak, line 215: No need to hardcode the version as the advanced targeting can be combined with "regular" audience targeting, e.g., min version, channel, locales.

Mardak (changes requested):
You'll also need to update the targeting configs further down in the file. See https://github.com/mozilla/experimenter/commit/585d7b5dd52ce243c377c12288679d5fa19b1d3d for an example

app/experimenter/experiments/constants/nimbus.py
Mardak: Oh, I see the comment about version 1703, so the description is out of date.

Mardak:
I don't think the telemetry field is actually used. Some advanced targeting just puts in empty string.

One way I've tested the targeting is using `browser.newtabpage.activity-stream.asrouter.devtoolsEnabled` -> open new tab -> wrench in the top right -> Targeting in the left -> paste/edit "Evaluate JEXL expression" input -> Evaluate

It shares a lot with what nimbus uses for targeting but it's not actually the same but seems to work well enough for quick tests of these expressions.

ncalexan:
@Mardak thanks so much for the prompt feedback.  I woke this morning and realized I didn't wire my definition into anything, which you pointed out -- I've now done that.  (If we're looking to streamline this process: it would be nice if there was a Python decorator that did the wiring automatically, so that copy-pasting the "obvious block" would do everything needed.)

Since the Firefox version targeting is not needed, I looked ahead to when we want to roll these default PDF experiments out to existing users and added just the Windows build number filter.  I also changed the copy to include `UserChoice`, since that's the semantic meaning of the syntactic build number.

jaredlockhart:
@Mardak Thank you for hopping in with all that helpful review!

@ncalexan Yes I know, I've tried a couple different things to remove the burden of having to explicitly add it to the dicts, but a decorator is a neat idea, I'll give that a whirl.

You're right I don't think the desktop telemetry is currently being used, it's there to assist in population sizing but developing the tooling around that is still backlogged.

As far as testing goes, we do two things
1) We run the targeting strings through a JEXL evaluator as part of the Python unit tests to ensure syntactic validity
2) We automatically do the thing @Mardak is suggesting, run each targeting expression through the Firefox Desktop targeting evaluator to sure both syntactic and semantic validity, as part of the CI tests.

The last step is doing code formatting, but I can do that for you.

ncalexan:
> @ncalexan Yeah it uses black, I ran it for you so don't worry about it. I cleaned up the string formatting for you too, so if it looks good to you and @Mardak it should be ready to merge. I retriggered the integration tests, Remote Settings can be moody.

I would flatten everything into the main patch and rebase, but not all projects do that.  Otherwise, everything looks good -- thanks for getting to this so quickly!

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -206,6 +206,31 @@ class NimbusTargetingConfig:
...
+    "(Windows build 15063 or newer)",
+    slug="windows_userchoice_first_run",
+    description="First start-up users (e.g. for about:welcome) on Windows with "
+    "UserChoice support (i.e., Windows build 15063+)",
+    targeting="{first_run} && {user_choice}".format(
+        first_run=TARGETING_FIRST_RUN.targeting,
+        user_choice=TARGETING_WINDOWS_WITH_USERCHOICE.targeting,
+    ),
+    desktop_telemetry=(
+        "{first_run} " "AND environment.system.os.windows_build_number >= 15063"
```
Mardak, line 229: nit: this could be a single string and reuse `TARGETING_WINDOWS_WITH_USERCHOICE.desktop_telemetry`

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -728,6 +753,10 @@ class EmailType(models.TextChoices):
         TARGETING_MAC_ONLY.slug: TARGETING_MAC_ONLY,
         TARGETING_NO_ENTERPRISE.slug: TARGETING_NO_ENTERPRISE,
         TARGETING_FX95_DESKTOP_USERS.slug: TARGETING_FX95_DESKTOP_USERS,
+        TARGETING_WINDOWS_WITH_USERCHOICE.slug: (TARGETING_WINDOWS_WITH_USERCHOICE),
```
Mardak, line 756: nit: unnecessary parens now that it's a single line

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -206,6 +206,31 @@ class NimbusTargetingConfig:
...
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+TARGETING_WINDOWS_WITH_USERCHOICE_FIRST_RUN = NimbusTargetingConfig(
+    name="First start-up users on Windows with UserChoice support "
+    "(Windows build 15063 or newer)",
+    slug="windows_userchoice_first_run",
+    description="First start-up users (e.g. for about:welcome) on Windows with "
+    "UserChoice support (i.e., Windows build 15063+)",
+    targeting="{first_run} && {user_choice}".format(
```
Mardak, line 224: This expands to `((isFirstStartup && !('trailhead.firstrun.didSeeAboutWelcome'|preferenceValue)) || experiment.slug in activeExperiments) && os.windowsBuildNumber >= 15063` so I think it should be fine as users probably aren't often downgrading from 10+ to 7/8 (and getting unenrolled).

Mardak (approved):

ncalexan:
I will fold and address review comments shortly, ready for merge by EoD.  Thanks!

ncalexan:
OK, I've folded this, addressed the last few review comments,  and rebased it onto `main`.  Should be ready to land.  Thanks, all!

PR #7250, opened 2022-05-03
Fixes #7249 chore(project): Add 99.1 to min version targeting

### LINKED ISSUE #7249: Add 99.1 to Firefox iOS minimum version filtering
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2401)

eliserichards:
Because...

* The iOS would like to target a minimum version of 99.1

This commit...

* Adds the constants for v99.1

image

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -708,6 +708,7 @@ class NimbusConstants(object):
         FIREFOX_98 = "98.!"
         FIREFOX_9830 = "98.3.0"
         FIREFOX_99 = "99.!"
+        FIREFOX_9910 = "99.1.0"
         FIREFOX_100 = "100.!"
         FIREFOX_101 = "101.!"
         FIREFOX_102 = "102.!"
--- a/app/experimenter/nimbus-ui/schema.graphql
+++ b/app/experimenter/nimbus-ui/schema.graphql
@@ -342,6 +342,7 @@ enum NimbusExperimentFirefoxVersionEnum {
   FIREFOX_98
   FIREFOX_9830
   FIREFOX_99
+  FIREFOX_9910
   FIREFOX_100
   FIREFOX_101
   FIREFOX_102
--- a/app/experimenter/nimbus-ui/src/types/globalTypes.ts
+++ b/app/experimenter/nimbus-ui/src/types/globalTypes.ts
@@ -175,6 +175,7 @@ export enum NimbusExperimentFirefoxVersionEnum {
   FIREFOX_98 = "FIREFOX_98",
   FIREFOX_9830 = "FIREFOX_9830",
   FIREFOX_99 = "FIREFOX_99",
+  FIREFOX_9910 = "FIREFOX_9910",
   NO_VERSION = "NO_VERSION",
 }
 


eliserichards:
A few questions... 

1) should we be adding this for legacy? my guess is "no"
2) in [schema.graphql](https://github.com/mozilla/experimenter/pull/7250/files#diff-94e150b136f1203751536073d038fabd6bece66ad8c0f85efb4942a5af26e0e2) it looks like the default is to have the full version number (`99.1.0`) instead of the shortened version (`99.1`). Is this right?

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -682,6 +682,7 @@ def parse(version_str):
         FIREFOX_98 = "98.!"
         FIREFOX_9830 = "98.3.0"
         FIREFOX_99 = "99.!"
+        FIREFOX_9910 = "99.1.0"
```
yashika-khurana, line 685: Just curious to know why `99.1.0`, as in other places its `99.1`? does it makes a difference?

jaredlockhart:
> A few questions...
> 
>     1. should we be adding this for legacy? my guess is "no"

Yeah you can ignore legacy, in general, but also particularly becuase this is a mobile request and legacy doesn't support mobile.

>     2. in [schema.graphql](https://github.com/mozilla/experimenter/pull/7250/files#diff-94e150b136f1203751536073d038fabd6bece66ad8c0f85efb4942a5af26e0e2) it looks like the default is to have the full version number (`99.1.0`) instead of the shortened version (`99.1`). Is this right?

This is a very good question and has to be directed to the requesters to specify exactly what version string they need.

>     3. if we're adding this version to the min's, should we also be adding it to the maxs?

If we add it to the Version enum shouldn't it automatically go into both?

app/experimenter/experiments/constants/legacy.py
```diff
@@ -175,6 +175,7 @@ def FEATURE_TYPE_CHOICES(cls):  # pragma: no cover
         ("97.0", "Firefox 97.0"),
         ("98.0", "Firefox 98.0"),
         ("99.0", "Firefox 99.0"),
+        ("99.1", "Firefox 99.1"),
```
jaredlockhart, line 178: No need to add here.

app/experimenter/experiments/constants/nimbus.py
jaredlockhart: It does make a difference, it's a bit tricky I can explan when next we chat.

eliserichards: I added it as 99.1.0 here to follow the convention above. Looks like Jared says (below) that's a question for the person who requested the version to be added

eliserichards:
> If we add it to the Version enum shouldn't it automatically go into both?

Yes, yes it does 😂 (oops)

eliserichards:
@yashika-khurana I removed the new version in `/legacy-ui/core/js/components/constants.js` and `experiments/constants/legacy.py`, which also removes the db migration and the min/maxes in `openapi-schema.json` and `swagger-ui.html`. This is much smaller now 👍

yashika-khurana:
> @yashika-khurana I removed the new version in `/legacy-ui/core/js/components/constants.js` and `experiments/constants/legacy.py`, which also removes the db migration and the min/maxes in `openapi-schema.json` and `swagger-ui.html`. This is much smaller now 👍

perfect makes sense 👍

yashika-khurana (approved):
Yeah good job 🥇

PR #7253, opened 2022-05-04
fix #7129 feat(nimbus): load remote json schemas

### LINKED ISSUE #7129: Load schemas for individual feature values
Some desktop features include schemas for their feature values rather than explicitly declaring each field right in the feature manifest.  We should load those schemas in and use them to validate feature values.

Example:

https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/FeatureManifest.yaml#467

jaredlockhart:
Because

* Some features define a rich json schema for their value types
* We can load and use those json schemas instead of generating them from the variables

This commit

* Looks for remotely defined feature schemas
* Loads and stores them if they're defined

### DIFF
--- a/app/experimenter/features/__init__.py
+++ b/app/experimenter/features/__init__.py
@@ -2,13 +2,17 @@ import json
 import os
 from enum import Enum
 from typing import Literal, Optional, Union
+from urllib.parse import urljoin
 
+import requests
 import yaml
 from django.conf import settings
 from django.core.checks import Error, register
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from experimenter.experiments.constants import NimbusConstants
+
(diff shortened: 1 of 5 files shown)


yashika-khurana:
@jaredlockhart I have once question, trying to understand better internal functionality, is the schema is present, we are using that, but is that schema already validated somewhere?

jaredlockhart:
And then about validation, that happens in two places:

* In Experimenter when the experiment owner is creating the experiment, and they can choose whether to strictly validate the feature value input or just warn on it (ie even if they fail validation they can still launch)
* In the client, ie in desktop, when the experiment is loaded

yashika-khurana (approved):
@jaredlockhart  thank you for helping me understand, its clear now :)  and yes changes looks good too 👏

PR #7263, opened 2022-05-05
fix #7262 feat(nimbus): add ESR channel for desktop

### LINKED ISSUE #7262: Add ESR channel
Some upcoming experiments need to explicitly target the ESR channel so we should add that to our list of channels.

The channel string is `"esr"`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2412)

jaredlockhart:
Because

* Some experiments need to target the ESR desktop channel

This commit

* Adds the ESR channel to desktop


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -12,6 +12,7 @@ class Channel(models.TextChoices):
     NIGHTLY = "nightly"
     BETA = "beta"
     RELEASE = "release"
+    ESR = "esr"
```
jaredlockhart, line 15: this is the string that goes to clients in targeting

app/experimenter/experiments/migrations/0208_alter_nimbusexperiment_channel.py
```diff
@@ -0,0 +1,29 @@
...
+            model_name="nimbusexperiment",
+            name="channel",
+            field=models.CharField(
+                choices=[
+                    ("", "No Channel"),
+                    ("default", "Unbranded"),
+                    ("nightly", "Nightly"),
+                    ("beta", "Beta"),
+                    ("release", "Release"),
+                    ("esr", "Esr"),
```
Mardak, line 23: I'm guessing "Esr" is some automatic title casing? We could name it "Extended Support Release"

Mardak (approved):
I don't quite know what all these other "esr" entries are for but functionality looks to be working:
esr channel

app/experimenter/experiments/migrations/0208_alter_nimbusexperiment_channel.py
jaredlockhart: Oh yeah tehre is a way to specify a custom label but we're not doing it yet.  If we wanna do that we'd have to do it as a followup.

PR #7264, opened 2022-05-05
feat(nimbus): add targeting to exclude users with certain tcp study codes

Mardak:
Because 
- TCP Phase 2 & 3 don't want to include users in the normandy study.

This commit
- adds negative targeting for certain search codes on each of the normandy branches for phase 3
- also adds targeting for new users for phase 2

tcp false targeting
tcp true targeting

### DIFF
--- a/app/experimenter/experiments/constants/nimbus.py
+++ b/app/experimenter/experiments/constants/nimbus.py
@@ -208,6 +208,29 @@ TARGETING_FIRST_RUN_WINDOWS_1903_NEWER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_NOT_TCP_STUDY = NimbusTargetingConfig(
+    name="Exclude users in the TCP revenue study",
+    slug="not_tcp_study",
+    description="Exclude users with certain search codes set",
+    targeting="!'browser.search.param.google_channel_us'|preferenceValue('')|regExpMatch"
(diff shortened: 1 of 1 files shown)


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -206,6 +206,29 @@ class NimbusTargetingConfig:
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TARGETING_NOT_TCP_STUDY = NimbusTargetingConfig(
+    name="Exclude users in the TCP revenue study",
+    slug="not_tcp_study",
+    description="Exclude users with certain search codes set",
+    targeting="!'browser.search.param.google_channel_us'|preferenceValue('')|regExpMatch"
+    "('^[ntc]us5$') && !'browser.search.param.google_channel_row'|preferenceValue('')|"
+    "regExpMatch('^[ntc]row5$')",
```
Mardak, line 215: Just copying over the targeting from https://github.com/mozilla/messaging-system-inflight-assets/pull/235/files#diff-c41aca0e731ed15191b531d6cfc8434c4b70695a96ad97f86b3c96946bbcfef9R355-R356

jaredlockhart (approved):
It's times like these I'm so glad we have both JEXL syntax and JEXL browser evaluation as part of our test suites.  LGTM 👍

jaredlockhart:
Oh @Mardak you'll need to do this to get circle to run:

https://support.circleci.com/hc/en-us/articles/4410418394523-Getting-Started-with-GitHub-SSO

and then resign your commit with `git commit --amend` and force push

PR #7266, opened 2022-05-05
fix #7265 chore(nimbus) Make custom targeting for UserChoice select W…

### LINKED ISSUE #7265: Make custom targeting for `UserChoice` select Windows version 1809+/build ID 17763+
This is follow-up to #7244.  The existing `UserChoice` feature appears to not work before 1809, so we only want to target 1809+ until it is fixed on earlier versions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2413)

ncalexan:
…indows version 1809+/build ID 17763+.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1767967.


ncalexan:
While in the area, I streamlined the name and description pair so that the UI looks better.

jaredlockhart (approved):
Thank you @ncalexan !

PR #7268, opened 2022-05-06
fix #7267 feat(nimbus): increase feature schema strictness

### LINKED ISSUE #7267: Increase feature value schema strictness
Now that we have the ability to disable strict schema validation and simply warn on validation errors, we should change the default schema validation to be strict, ie require all defined feature variables as properties, validate their types, and prevent additional fields.  If for whatever reason a user needs to deviate from the stored schema they can set the schema validation to warn only.

In addition, we recently discovered that remote settings does not allow float values in json payloads at all, and the feature schema defines only `int` as a numeric type, so we should not be using `number` but rather `integer` in the generated schema.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2414)

jaredlockhart:
Because

* We now have the ability to disable schema validation failure preventing launches
* As our customer base grows, the possibility for error is increasing
* We have seen recent cases where invalid feature data can cause difficult to trace errors

This commit

* Uses the correct integer type for numbers instead of Number
* Increases the strictness of the auto generated feature value schemas

### DIFF
--- a/app/experimenter/features/__init__.py
+++ b/app/experimenter/features/__init__.py
@@ -23,7 +23,7 @@ class FeatureVariableType(Enum):
 
 
 FEATURE_SCHEMA_TYPES = {
-    FeatureVariableType.INT: "number",
+    FeatureVariableType.INT: "integer",
     FeatureVariableType.STRING: "string",
     FeatureVariableType.BOOLEAN: "boolean",
 }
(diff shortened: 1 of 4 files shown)


yashika-khurana (approved):
LGTM! 😊 🎉

PR #7269, opened 2022-05-06
7150/add languages field nimbus UI

yashika-khurana:
Because

* Desktop Firefox uses a ‘locale’ to determine its language, and Experimenter includes a locale field to target desktop clients based on that locale/language.  However these locales do not match the locale/language codes used by mobile clients.  Rather than attempting to repurpose/mutate the locale field in Experimenter to support both desktop and mobile clients=> so now for the `Mobile Client`  we are enabling `languages` field 

Change
- Add the option to select languages  and remove option to select locales for mobile client
- Add the option to select locales  and remove option to select languages for Desktop client
- Show language field in summary if the application is Mobile, else show locales
- Test cases to support new changes

Refer to screenshot below to see new changes: 

For mobile client- Language field is enabled, and locale removed and user can choose option from dropdown
image

Languages selected option for Mobile client
image

If no language is selected, it will show All languages for mobile client
image

For the desktop
image

image

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -186,6 +186,32 @@ describe("FormAudience", () => {
     ).toHaveAttribute("disabled");
   });
 
+  it("enables language field for mobile", async () => {
+    render(
+      <Subject
+        experiment={{
+          ...MOCK_EXPERIMENT,
+          application: NimbusExperimentApplicationEnum.FENIX,
+        }}
+      />,
+    );
+
+    expect(screen.queryByTestId("languages")).toBeInTheDocument();
+  });
+
+  it("disables language field for desktop", async () => {
+    render(
+      <Subject
+        experiment={{
+          ...MOCK_EXPERIMENT,
+          application: NimbusExperimentApplicationEnum.DESKTOP,
+        }}
+      />,
+    );
+
+    expect(screen.queryByTestId("languages")).not.toBeInTheDocument();
+  });
+
   it("calls onSubmit when save and next buttons are clicked", async () => {
     const onSubmit = jest.fn();
     const expected = {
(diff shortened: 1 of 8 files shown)


app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -186,6 +186,32 @@ describe("FormAudience", () => {
...
+      <Subject
+        experiment={{
+          ...MOCK_EXPERIMENT,
+          application: NimbusExperimentApplicationEnum.DESKTOP,
+        }}
+      />,
+    );
+
+    expect(screen.queryByTestId("languages")).not.toBeInTheDocument();
+  });
```
jaredlockhart, line 213: Might also be worth asserting the opposite in each test, ie languages is shown, locales isn't, and visa versa?

jaredlockhart (commented):
Yep looks exactly right, works perfectly.  I just realized we need to add the review check with the version blocking, and also add it to the targeting serialization, before we can land this.  So let's mark this as blocked until those land and then we can land this.  Great job, thnx @yashika-khurana !

yashika-khurana:
> Yep looks exactly right, works perfectly. I just realized we need to add the review check with the version blocking, and also add it to the targeting serialization, before we can land this. So let's mark this as blocked until those land and then we can land this. Great job, thnx @yashika-khurana !

perfect, got the idea, thank you for mentioning that :)

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
yashika-khurana: absolutely, that check would be better to have, I just added it :)

jaredlockhart (commented):
Blocked until backend prs land

yashika-khurana:
@jaredlockhart should we release this now???? 🎉

jaredlockhart:
@yashika-khurana Looks like something's not hooked up right, I can set a low version and include a language but I'm not receiving the review error?

Screen Shot 2022-06-03 at 2 07 59 PM

yashika-khurana:
@jaredlockhart yeah I am looking into it, thanks for noticing it 🙏

yashika-khurana:
@jaredlockhart fixed now!
image

yashika-khurana:
@jaredlockhart  here we go 🚀  , thank you again 🙏 
image

jaredlockhart (approved):
Fantastic!  Very well executed change.  Clear method updates, clear succinct tests.  Tested locally and everything works correctly.  Great job @yashika-khurana !  🎉 🎉 🎉 🎉

PR #7277, opened 2022-05-11
fix #7275 bug(nimbus): remove required fields from feature schemas

### LINKED ISSUE #7275: Remove required properties from JSON schema validation
We recently modified the auto generated JSON schemas to enforce all required properties, however for some features not all properties are required.  Let's take the required fields out but continue to leave additional properties disabled.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2422)

jaredlockhart:
Because

* We recently increased the strictness of the auto generated feature value schemas to force all fields to be required
* Some variables are optional for some features
* Requiring all fields for features with some optional fields can lead to awkward errors

This commit

* Removes the required field list from the feature value schemas

### DIFF
--- a/app/experimenter/features/__init__.py
+++ b/app/experimenter/features/__init__.py
@@ -60,7 +60,6 @@ class Feature(BaseModel):
             "type": "object",
             "properties": {},
             "additionalProperties": False,
-            "required": [],
         }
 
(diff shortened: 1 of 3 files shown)


eliserichards (approved):
👏 Looks good!

Issue #7185, opened 2022-04-06
Transition from long lived gcp key to ADC


jaredlockhart:
Presently, for local dev to interact with the GCS buckets that store jetstream data, we provision a long lived GCP key, [documented here](https://github.com/mozilla/experimenter#google-credentials-for-jetstream).

However, @jasonthomas-mozilla-owner [suggests an alternative using ADC](https://mozilla-hub.atlassian.net/browse/DSRE-654?focusedCommentId=533248&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-533248).

Let's prototype this workflow and then update the docs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2359)

PR #7282, opened 2022-05-11
Fixes #7281 chore(project): Add new google credentials to the setup instructions in the readme

### LINKED ISSUE #7281: Add google creds to setup instructions
We need to add the [new google credentials](https://github.com/mozilla/experimenter/commit/20aa3cc93f6bb3357365d749e8a53522f4f8d349) from #7185 to the README instructions:

```
In your `.env` file, add the following: 

GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/application_default_credentials.json
GOOGLE_ADC_FILE=~/.config/gcloud/application_default_credentials.json
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2429)

eliserichards:
Because...

* #7185 updated our google credentials
* these `.env` vars are used as the volume mount point for Docker

This commit...

* Updates the README setup instructions to include these credentials

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,13 @@ Check out the [🌩 **Nimbus Documentation Hub**](https://experimenter.info) or 
 	...
 	```
 
+1. Add Google credentials to your .env
+
+	```
+        GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/application_default_credentials.json
+        GOOGLE_ADC_FILE=~/.config/gcloud/application_default_credentials.json
+	```
+
 1. Run tests
 
         make check


jaredlockhart (changes requested):
Ah thank you @eliserichards for making a point to update the README!  But I don't think this is quite right.  This shouldn't affect new setups, since those keys/values are in .env.sample and the new setup will automatically get them when they

```
cp.env.sample .env
```

The issue is for existing local devs when .env.sample changes, so as @mikewilli suggested maybe this belongs in the troubleshooting section and some examples are

Local dev won't start?  Try:

- Temporarily try using .env.sample:
  - `mv .env .env.bak`
  - `cp .env.sample .env`
  - `make refresh`
- Diff your env with sample
  - `diff .env .env.sample`
  - look for missing keys

jaredlockhart:
Also you might have to do this to get circle working:

https://support.circleci.com/hc/en-us/articles/4410418394523-Getting-Started-with-GitHub-SSO

PR #7283, opened 2022-05-12
fix #7261 chore(project): move legacy models into legacy app

### LINKED ISSUE #7261: Split nimbus and legacy experiments apps
Right now, the legacy and nimbus experiments models/views/serializers/etc are all contained in a single app called `experiments`.  Since legacy is in maintenance mode and being deprecated soon, we should split these up so that continued work on nimbus doesn't involve having to wade through legacy files.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2411)

jaredlockhart:
Because

* During the initial phases of Nimbus development we put everything in the existing Experiments app
* Having both Legacy and Nimbus experiments in the same app creates a lot of noise when navigating the project
* Legacy experiments are being deprecated entirely soon
* It would help to move all the legacy experiment concerns into a separate app

This commit

* Begins this migration process by moving only the Legacy models into a new app
* To keep this process sane, we'll do it in many pieces
* By using the db_name Meta field, we can move the models in code without moving the actual tables in the database


app/experimenter/base/tests/test_load_dummy_experiments.py
```diff
@@ -1,8 +1,9 @@
 from django.core.management import call_command
 from django.test import TestCase
 
-from experimenter.experiments.models import Experiment, NimbusExperiment
+from experimenter.experiments.models import NimbusExperiment
 from experimenter.experiments.tests.factories import NimbusExperimentFactory
+from experimenter.legacy.legacy_experiments.models import Experiment
```
eliserichards, line 6: Looks like we might need to stop using `Experiment.STATUS_CHOICES` and replicate it on the nimbus side 🤔

eliserichards (commented):
Lookin' good so far! I'd love your thoughts on choosing `experimenter.legacy.legacy_experiments` vs `experimenter.legacy.experiments`?

jaredlockhart:
> Lookin' good so far! I'd love your thoughts on choosing `experimenter.legacy.legacy_experiments` vs `experimenter.legacy.experiments`?

I know I wanted to, but turns out django gets mad at you if two apps have the same name even if they're in sub paths, I guess because there's many places where you just refer to `app.Model` regardless of its import path so they need to be absolutely unique.  Shame.

app/experimenter/base/tests/test_load_dummy_experiments.py
jaredlockhart: Good catch!  I'm finding a couple little weird overlap things, like there was bucketing defined in Legacy that wasn't being used for anything at all.  So hopefully we'll find more of these kinds of things as we continue this process 🙏

eliserichards: Do we have any issues filed for these lil overlaps? Do you think that should be part of this work, or can it come later?

yashika-khurana (approved):
LGTM, Great job 🎉 👍

app/experimenter/base/tests/test_load_dummy_experiments.py
jaredlockhart: I don't have individual issues filed for the overlaps/duplication, some were already handled as part of this PR, and I suspect that they'll have to be resolved as I move things around, so I'm thinking of them more as side effects of the changes rather than the focus of the changes themselves.  If I get through a whole pass on this and we still find some them we can try filing tickets for them more individually.

eliserichards: That sounds good to me. I just want to make sure nothing falls through the cracks 😄

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,47 +1,104 @@
...
+        VariantPreferences = self.new_state.apps.get_model(
+            "legacy_experiments", "VariantPreferences"
+        )
+        RolloutPreference = self.new_state.apps.get_model(
+            "legacy_experiments", "RolloutPreference"
         )
 
-        self.assertFalse(NimbusBranchFeatureValue.objects.filter(value=None).exists())
-        self.assertTrue(NimbusBranchFeatureValue.objects.filter(value="value").exists())
-        self.assertTrue(NimbusBranchFeatureValue.objects.filter(value="").exists())
```
eliserichards, line 47: Are we no longer `assert`ing in these tests? Does `XXX.objects.get` cover these cases for us?

eliserichards (changes requested):

eliserichards (approved):
Gorgeous! Let's get this merged! 💪

yashika-khurana (approved):
LGTM! 🚀

PR #7299, opened 2022-05-20
fixes #7298 feat(nimbus): add targeting for infrequent bookmarks exp

### LINKED ISSUE #7298: Add targeting for desktop bookmark experiment
https://bugzilla.mozilla.org/show_bug.cgi?id=1767380

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2454)

dmose:
This adds targeting for infrequent users on mac or windows who only have 5 bookmarks, which is the default number of bookmarks in Firefox 100, acting as a proxy for "hasn't performed an import to this profile".

It's in service of [this experiment](https://bugzilla.mozilla.org/show_bug.cgi?id=1767380) and will allow us to fix [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1770396).


app/experimenter/experiments/constants/nimbus.py
```diff
@@ -421,6 +421,20 @@ class NimbusTargetingConfig:
...
 
+TARGETING_INFREQUENT_MAC_OR_WIN_USER_WITH_5_BOOKMARKS = NimbusTargetingConfig(
+    name="Infrequent Mac or Windows users with 5 bookmarks",
+    slug="infrequent_mac_or_win_users_with_5_bookmarks",
+    # 5 is the new profile bookmark count in Fx 100, a proxy for nothing imported
+    description="Between 1 and 6 days activity in the past 28 days, w/5 bookmarks",
+    targeting="({infrequent_user} && {win_or_mac}) || {sticky}".format(
+        win_or_mac="(os.isWindows || os.isMac)",
+        infrequent_user=TARGETING_INFREQUENT_USER_URIS.targeting,
+        sticky=TARGETING_STICKY,
```
Mardak, line 432: Description mentions bookmarks, but targeting doesn't? Does the targeting actually need to be limited to windows and mac and exclude linux?

Mardak (changes requested):

app/experimenter/experiments/constants/nimbus.py
dmose: Targeting fixed.  I've asked akash about the linux choice, and will update or comment here once I hear back.

dmose: PM is happy to have linux, so I've added it in.

app/experimenter/experiments/constants/nimbus.py
```diff
@@ -421,6 +421,22 @@ class NimbusTargetingConfig:
...
 
+TARGETING_INFREQUENT_MAC_WIN_OR_LINUX_USER_WITH_5_BOOKMARKS = NimbusTargetingConfig(
+    name="Infrequent Mac, Win, or Linux user with 5 bookmarks",
+    slug="infrequent_mac_win_or_linux_user_with_5_bookmarks",
+    # 5 is the new profile bookmark count in Fx 100, a proxy for nothing imported
+    description="Between 1 and 6 days activity in the past 28 days, w/5 bookmarks",
+    targeting="({infrequent_user} && {mac_win_or_linux} && {fiveBookmarks})"
+    " || {sticky}".format(
+        infrequent_user=TARGETING_INFREQUENT_USER_URIS.targeting,
+        mac_win_or_linux="(os.isMac || os.isWindows || os.isLinux)",
```
Mardak, line 432: Why this way instead of omitting this?

dmose: Since we didn't also include the various Tier >1 platorms (e.g. FreeBSD, NetBSD, etc, etc.).  Otherwise default queries will include them unless the query writer goes the to trouble of excluding them. That said, presumably these numbers would be small enough to get lost in the noise, but I figured I'd default to eliminating an extra variable, given that there's already plenty of complexity in data analysis.

I'm happy to do any of a) leave it alone, b) add a comment explaining the above, or c) simply drop that clause, whatever you'd prefer...

Mardak: If the concern is introducing an extra variable, it looks like this approach does so by being different from all the other targeting we've used.

dmose: Heh, fair enough.  I'll remove it.

jaredlockhart (approved):
Looks good 👍

dmose:
Rather than deal with the merge conflicts here, I'm just going to close this PR and submit a new one, which will be much simpler.

Issue #7064, opened 2022-02-04
Intermittent integration test failure: test_create_new_experiment_timeout_remote_settings


jaredlockhart:
Seen in https://app.circleci.com/pipelines/github/mozilla/experimenter/11983/workflows/d3534cc9-89c5-449d-b686-3c722ffc9457/jobs/36751

```py
=================================== FAILURES ===================================
_________ test_create_new_experiment_timeout_remote_settings[DESKTOP] __________

selenium = <selenium.webdriver.firefox.webdriver.WebDriver (session="d4a7d654-e2f8-4b93-b529-7f9ffd4de5e6")>
create_experiment = <function create_experiment.<locals>._create_experiment at 0x7effd791f9d0>

    @pytest.mark.run_once
    def test_create_new_experiment_timeout_remote_settings(
        selenium,
        create_experiment,
    ):
        summary = create_experiment(selenium)
        summary.launch_and_approve()
>       summary.wait_for_timeout_alert()

nimbus/test_remote_settings.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
nimbus/pages/experimenter/summary.py:95: in wait_for_timeout_alert
    self.wait_with_refresh(
nimbus/pages/base.py:27: in wait_with_refresh
    self.wait.until(_wait_for_it, message=message)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <selenium.webdriver.support.wait.WebDriverWait (session="d4a7d654-e2f8-4b93-b529-7f9ffd4de5e6")>
method = <function Base.wait_with_refresh.<locals>._wait_for_it at 0x7effd7968ca0>
message = 'Summary Page: Unable to find timeout alert'

    def until(self, method, message=''):
        """Calls the method provided with the driver as an argument until the \
        return value is not False."""
        screen = None
        stacktrace = None
    
        end_time = time.time() + self._timeout
        while True:
            try:
                value = method(self._driver)
                if value:
                    return value
            except self._ignored_exceptions as exc:
                screen = getattr(exc, 'screen', None)
                stacktrace = getattr(exc, 'stacktrace', None)
            time.sleep(self._poll)
            if time.time() > end_time:
                break
[...]

PR #7301, opened 2022-05-23
fixes #7300 feat(nimbus): Add integration testing between the SDK and experimenter.

### LINKED ISSUE #7300: Integration testing between the SDK and Experimenter
https://mozilla-hub.atlassian.net/browse/EXP-2360

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2455)

jrbenny35:
This adds configuration to allow us to test the mobile sdk in combination with experimenter. With the help of @travis79 I was able to get a fresh set of 20 app contexts to use when creating the sdk nimbus config. I am also separating the jobs on circleci to not hold up the main integration test job. Using a regular python 3 docker 7image I created a script to set it up correctly each time.


jrbenny35:
Think I am running into https://github.com/mozilla/experimenter/issues/7064

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,158 @@
...
+
+def locale_number_loader(locales=None):
+    locale_list = []
+    path = Path().resolve()
+    path = str(path)
+    path = path.strip("/tests/integration/nimbus")
+    path = f"/{path}/experimenter/base/fixtures/locales.json"
+    with open(path) as file:
+        data = json.loads(file.read())
+        for locale in locales:
```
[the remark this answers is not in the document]
jrbenny35: no, it will error. I'd like it to fail hard here as if we don't load any locales then we don't have any app contexts and creating the nimbus object will fail.

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,158 @@
...
+                    locale_list.append(item["pk"])
+    return locale_list
+
+
+def client_info_list():
+    def poll_job(s, redash_url, job):
+        while job["status"] not in (3, 4):
+            response = s.get("{}/api/jobs/{}".format(redash_url, job["id"]))
+            job = response.json()["job"]
+            time.sleep(1)
```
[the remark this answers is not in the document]
jrbenny35: Hmm I guess this is possible...i'd have to ask @travis79 about it.

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if job["status"] == 3:
+            return job["query_result_id"]
+
+        return None
+
+    redash_url = "https://sql.telemetry.mozilla.org"
+    query_id = 86023
```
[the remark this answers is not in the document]
jrbenny35: Well it is a saved query. I am assuming if it changes then the ```query_id``` would change too. Is that correct @travis79 ?

docker-compose-integration-test.yml
```diff
@@ -17,3 +17,14 @@ services:
     ports:
       - "5900:5900"
     shm_size: 2g
+  rust-sdk:
+    image: b4handjr/nimbus-rust-image:latest
```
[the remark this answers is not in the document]
jrbenny35: I like the idea of a dockerfile...I kinda want application-services to have a dockerfile that we could use as mine is very specific to nimbus (it builds ```nimbus.py```). I could certainly put it in this repo and we maintain it here since it is just for us.

jaredlockhart (commented):
Hopefully rebasing on the remote settings changes clears this up 🙏

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,128 @@
...
+    return _load_app_context_helper
+
+
+@pytest.fixture(name="ar_units")
+def fixture_ar_units():
+    return nimbus.AvailableRandomizationUnits(None, 0)
+
+
+@pytest.fixture(name="sdk_client")
+def fixture_sdk_client(ar_units):
```
jaredlockhart, line 64: What's `ar_units`?

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,158 @@
...
+            value = targetting_helper.eval_jexl(expression)
+        except nimbus.NimbusError.EvaluationError:
+            return None
+        else:
+            return value
+
+    return _eval_jexl
+
+
+@pytest.mark.parametrize("context", client_info_list())
```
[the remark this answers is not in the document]
jaredlockhart: Yeah testing the JEXL expression from the mobile advanced targetings is specifically what I'd like to test here since Experimenter has no way of determining whether they're 'valid' or not, only the SDK can determine that.

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,128 @@
...
+    jexl_evaluator, sdk_client, load_app_context, slugify, experiment_name
+):
+    client = json.loads(client_info_list()[0]["app_context"])
+    helpers.create_mobile_experiment(
+        slugify(experiment_name),
+        client["app_name"],
+        locale_number_loader([client["locale"]]),
+    )
+    expression = ""  # represents a malformed jexl expression
+    assert jexl_evaluator(sdk_client(load_app_context(client)), expression) is None
```
jaredlockhart, line 128: We don't need to test these cases here since that's covered by the SDK unit tests, the only thing we need to worry about here is the JEXL defined in the advanced targeting for mobile.

app/tests/integration/nimbus/utils/helpers.py
```diff
@@ -0,0 +1,86 @@
...
+                        }
+                    }
+                    """,
+        }
+    )
+    targeting_configs = []
+    for item in data["data"]["nimbusConfig"]["targetingConfigs"]:
+        if "DESKTOP" in item["applicationValues"]:
+            targeting_configs.append(item["value"])
+    return targeting_configs
```
jaredlockhart, line 47: Is this being used anywhere?  This is filtering for desktop experiments but we want the opposite?

jaredlockhart (changes requested):
Awesome, this has come so far, and it runs so fast!  The only thing left I think is to create an experiment with each advanced targeting and then include that in the evaluation so we have a way to know that they're valid.  Great job @jrbenny35 and @tarikeshaq !! 🎉 🎉 🎉 🎉

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,128 @@
...
+from pathlib import Path
+
+import pytest
+from nimbus.models.base_app_context_dataclass import BaseAppContextDataClass
+from nimbus.utils import helpers
+
+nimbus = pytest.importorskip("nimbus_rust")
+
+
+def locale_number_loader(locales=None):
```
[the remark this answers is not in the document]
jrbenny35: yeah I guess its the database ID

app/tests/integration/nimbus/test_mobile_integration.py
jrbenny35: um.... @tarikeshaq ? I know it was needed to create the rust client

app/tests/integration/nimbus/test_mobile_integration.py
jrbenny35: Kinda what I figured but the acceptance criteria on the JIRA ticket mentioned these. Should I remove them?

app/tests/integration/nimbus/utils/helpers.py
jrbenny35: I plan to replace the setup in `test_firefox_targeting.py` with this but that will follow this PR.

jaredlockhart: Okay cool can you file the ticket and link it here as a followup pls? 🙏

app/tests/integration/nimbus/test_mobile_integration.py
tarikeshaq: We probably want to include the full name here 😄 It's the `available_randomization_units` - it helps with randomization, in the case a bucket configuration wants to randomize using a client_id (useful in some tests too) 

if the experiments all use `nimbus_id` as the `randomizationUnit`, we can just pass a default value (i.e construct it with None as the `client_id` as Ben is doing here!)

app/tests/integration/nimbus/test_mobile_integration.py
jrbenny35: Malforming those will result in a `None` not a `False`

jaredlockhart: Perfect then we can check for that similar to how we do for desktop 👍

docker-compose-integration-test.yml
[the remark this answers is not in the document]
tarikeshaq: Ooo I like that idea. I think there is a small decision to make around how versioning would work (app services versions vs app versions when it comes to testing)

once that's a little clearer we can have the dockerfile in app services and have it rebuild on a regular cadence, whether it's cron-job, on release, etc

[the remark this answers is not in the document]
tarikeshaq: But for now, I'm happy with this 😄

app/tests/integration/nimbus/test_mobile_integration.py
jrbenny35: Thanks Tarik...I added a docstring

app/tests/integration/nimbus/test_mobile_integration.py
```diff
@@ -0,0 +1,115 @@
...
+):
+    client = json.loads(client_info_list()[0]["app_context"])
+    helpers.create_mobile_experiment(
+        slugify(experiment_name),
+        client["app_name"],
+        locale_databse_id_loader([client["locale"]]),
+    )
+    data = helpers.load_experiment_data(slugify(experiment_name))
+    expression = data["data"]["experimentBySlug"]["jexlTargetingExpression"]
+    assert jexl_evaluator(sdk_client(load_app_context(client)), expression[::-1]) is None
```
jaredlockhart, line 115: Won't this test always pass?

jrbenny35:
Blocked on a rust error

```
value = True, buf = <nimbus_rust.RustBufferBuilder object at 0x7faa613ba4f0>

    @staticmethod
    def write(value, buf):
>       utf8Bytes = value.encode("utf-8")
E       AttributeError: 'bool' object has no attribute 'encode'
```

app/tests/integration/nimbus/test_mobile_integration.py
jrbenny35: Yeah I removed it

jaredlockhart (approved):
YES!!!!!@#($*#()@$)(

Tested locally, everything behaves exactly as hoped.  This is so fantastic, this is such a huge step forward, I can't believe we've gotten this far.

Fantastic work everyone, thank you so much @jhugman @travis79 @tarikeshaq and massive props to @jrbenny35 for struggling and suffering and pulling it all together.  This is epic.  Great job.

giphy

jrbenny35:
damn requirements file

PR #7309, opened 2022-05-24
fixes #7307 feat(nimbus): Add languages to JEXL targeting

### LINKED ISSUE #7307: Add languages to JEXL targeting
After we add the `languages` field to the NimbusExperiment Model, we need to add it to the targeting context here:

[https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L276](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models/nimbus.py#L276|smart-link) 

similar to how we handle locales and countries, ie the JEXL expression should be

`(language in [....])`

and add/update tests as necessary.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2418)

yashika-khurana:
fixes #7307 

Because...

  We want to add languages to JEXL targeting, like `(language in [....])`

This commit...

-  Adds the languages in JEXL expression for non-desktop applications
-  Tests


jaredlockhart (changes requested):
Looks great, couple small tweaks but nothing major.  Thnx @yashika-khurana 🙏 🙏 🙏 🙏

yashika-khurana:
> Looks great, couple small tweaks but nothing major. Thnx @yashika-khurana 🙏 🙏 🙏 🙏

@jaredlockhart good point to remove redundant checks. thank you 🙏 😊

app/experimenter/experiments/models/nimbus.py
```diff
@@ -294,14 +294,14 @@ def targeting(self):
...
 
-            if self.application == self.Application.DESKTOP:
-                expressions.append(f"locale in {locales}")
-            else:
-                iso_locales = {locale[:2] for locale in locales}
-                iso_locales_expression = " || ".join(
-                    [f"'{language}' in locale" for language in sorted(iso_locales)]
-                )
-                expressions.append(iso_locales_expression)
+            expressions.append(f"locale in {locales}")
```
jaredlockhart, line 297: Sorry @yashika-khurana I just realized if we take this out now then in the time between landing this and landing the UI for languages mobile experiments won't work properly, let's split this into two.  We can add the language part below here and leave the language code untouched, and then file a follow up to remove the special case for mobile locales after we add the languages UI.

jaredlockhart (changes requested):

app/experimenter/experiments/models/nimbus.py
yashika-khurana: Got it, but we do have UI ready as well https://github.com/mozilla/experimenter/pull/7269. Am I missing something? 🤔

yashika-khurana: Yes makes sense, thank you for explaining it 😊 🙏

jaredlockhart (approved):
Fantastic thank you @yashika-khurana !  🎉 🎉 🎉 🎉

PR #7323, opened 2022-05-26
fix bug 1770367 feat(nimbus): add "uses FxA" targeting for regular activity level

Mardak:
```
experiment.slug in activeExperiments ||
userMonthlyActivity|length >= 14 &&
userMonthlyActivity|length <= 20 &&
isFxAEnabled &&
usesFirefoxSync
```

I was thinking of moving sticky into the base `TARGETING_REGULAR_USER_URIS` but maybe that's a later cleanup or maybe it'll be handled as part of some general sticky toggle or maybe all targeting will be sticky?


jaredlockhart (approved):
Looks good 👍

PR #7349, opened 2022-06-01
fix #7347 fix(project): handle null values for experiment enums in import/export

### LINKED ISSUE #7347: Nimbus UI error for imported experiments
Certain enum fields with null values are serialized as `""` (in JSON, at least) on export, but this causes an error after import when loading the UI (see screenshot below).

We need the import/export to handle these null values properly for fields that allow null as a value. This could mean updating the serializer (or deserializer) for the experiment, or changing the widget used by the `django-import-export` library for the relevant fields (either via monkeypatch like we did to fix a bug with the DecimalWidget's `render` fn, or by changing the widget used for these nullable fields).

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2482)

mikewilli:
Because

* experiment import/export library does not properly handle nulls for nullable enums

This commit

* creates special cases for these fields in order to properly handle them as null instead of empty string


jaredlockhart (approved):
Looks good!

PR #7360, opened 2022-06-02
fix #7357 feat(nimbus): move targeting into its own app

### LINKED ISSUE #7357: Move targeting into its own app
We have a lot of logic crammed into the `experiments` app, so it would be nice to break some things out into their own apps, and I think targeting is high traffic and isolated concern enough that we can move it out into its own little app.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2487)

jaredlockhart:
Because

* Targeting has fairly isolated concerns
* The targeting configurations are very high traffic for changes
* Having to explicitly configure the targeting dictionaries when making changes is annoying

This commit

* Moves targeting configs into their own app
* Auto generates the targeting config dictionaries instead of explicitly declaring them
* Changes the default 'no targeting' to an explicit slug 'no_targeting' instead of an empty string
* Migrates existing experiments from the old to new default


yashika-khurana:
@jaredlockhart integration test seems to be failing 🤔

yashika-khurana:
@jaredlockhart I see there are some conflicts with main 🤔

app/experimenter/experiments/migrations/0211_alter_nimbusexperiment_targeting_config_slug.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.2.13 on 2022-06-02 19:10
+
+from django.db import migrations, models
+
+
+def update_no_targeting_defaults(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+    NimbusExperiment.objects.filter(targeting_config_slug="").update(
+        targeting_config_slug="no_targeting"
```
yashika-khurana, line 9: 🎉

yashika-khurana (approved):
LGTM! 🎉 👏 
Just a conflict-regarding constant.

PR #7373, opened 2022-06-07
fixes #7364 feat(nimbus): sort report by start date

### LINKED ISSUE #7364: sort reports by start date
We want reports to be sorted out based on start date, need to modify sorting so that we see latest experiments at the top.
Hint: `compare them as date objects and instead of None use a date in the past, like Jan 1 1971`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2491)

yashika-khurana:
Because

* We want to see reports sorted by start date of the experiments

This commit

* Modifies the query to sort results
* Test to verify new changes

fixes #7364


jaredlockhart (dismissed):
Looks good but looks like the test isn't testing the logic rigorously enough, if I comment the sorting out:

```py
    def get_queryset(self):
        return super().get_queryset()

        # return sorted(
        #     key=lambda experiment: (
        #         experiment.start_date and experiment.start_date.strftime("%Y-%m-%d")
        #     )
        #     or "",
        #     reverse=True,
        # )
```

the test still passes so it probably needs to be a little stricter about its setup and assertions.

yashika-khurana:
> Looks good but looks like the test isn't testing the logic rigorously enough, if I comment the sorting out:
> 
> ```python
>     def get_queryset(self):
>         return super().get_queryset()
> 
>         # return sorted(
>         #     key=lambda experiment: (
>         #         experiment.start_date and experiment.start_date.strftime("%Y-%m-%d")
>         #     )
>         #     or "",
>         #     reverse=True,
>         # )
> ```
> 
> the test still passes so it probably needs to be a little stricter about its setup and assertions.

@jaredlockhart I totally agree on this, I added one more experiment to support changes. 😊 🙏  thank you 🙏

jaredlockhart (approved):
Fantastic good job!

PR #7374, opened 2022-06-07
fix #7362 feat(nimbus): filter archived experiments

### LINKED ISSUE #7362: Filter archived experiments out of CSV report
Looks like we're not filtering out the archived experiments in the CSV report so there's a lot of stuff in there that probably doesn't need to be in there.  Let's exclude archived experiments for now.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2489)

yashika-khurana:
Because

    We want to see experiments that are not archived

This commit

-     Modifies the query to get non-archived experiments
-     Test to verify new changes

fixes #7362

### DIFF
--- a/app/experimenter/experiments/api/v5/views.py
+++ b/app/experimenter/experiments/api/v5/views.py
@@ -12,14 +12,20 @@ class NimbusExperimentCsvRenderer(CSVRenderer):
 
 class NimbusExperimentCsvListView(ListAPIView):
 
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Yep, code looks correct, behaviour is correct, test is correct, good job 👍 I'll mark this blocked until https://github.com/mozilla/experimenter/pull/7373 lands.  Thanks @yashika-khurana !

PR #7379, opened 2022-06-08
fix #7377 feat(legacy): remove create delivery button

### LINKED ISSUE #7377: Disable the "create delivery" button in Normandy
I'd like to try removing the button from the interface.  Let's try directing everyone through office hours and shell can manually create in the admin tool if it truly is a case that only normandy can support still.

If people think they may need Normandy - they should attend the weekly [Experiment Tooling office hours ](https://docs.google.com/document/d/1fj4BqG0N-cN_Fx5sq-AozdiOdKO7asaDWR4B7sH88F0/edit#heading=h.yguiolmttiw2)to see if it can be done with Nimbus instead.  If not - it will be manual with Shell to create.

https://mozilla-hub.atlassian.net/browse/EXP-2501

jaredlockhart:
Because

* We continue to transition away from Normandy towards Nimbus
* Any new experiment/rollout should be launched with Nimbus
* Any delivery that requires Normandy support can be created without needing Experimenter

This commit

* Removes the Create Delivery button in the Legacy UI

Screenshot 2022-06-08 at 11-02-49 Mozilla Experimenter

### DIFF
--- a/app/experimenter/legacy/legacy-ui/templates/experiments/list.html
+++ b/app/experimenter/legacy/legacy-ui/templates/experiments/list.html
@@ -66,17 +66,6 @@
   <a href="{% url "experiments-api-csv" %}?{{ request.GET.urlencode }}"><i class="fas fa-file-csv"></i> Export as CSV</a>
 {% endblock %}
 
-{% block header_sidebar %}
-  <div class="row">
-    <div class="col">
-      <a class="col btn btn-primary" href="{% url "experiments-create" %}">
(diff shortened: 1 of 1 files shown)


tiftran (approved):
yay! 🥳 

So the new workflow in the event for a rare normandy experiment/rollout would be:
Admin -> Experiments Model-> Add -> Enter basic info -> move back to legacy ui to finish it off?

jaredlockhart:
@tiftran That's a good question, I'm not sure, @shell1 what do you think?

PR #7381, opened 2022-06-08
fix #7380 feat(project): nimbus default landing page

### LINKED ISSUE #7380: Make Nimbus the default landing page
Now that we're disabling the create button in the legacy UI, it's probably time to make Nimbus the default landing page.  We have to make sure we preserve all the existing urls for things that people have bookmarked.  We also have to make sure we have a button in the Nimbus landing page to get back to the legacy UI.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2504)

jaredlockhart:
Because

* We continue to transition away from Normandy towards Nimbus
* All new experiments/rollouts should be launched with Nimbus
* We are disabling the ability to launch Normandy deliveries in the Legacy UI

This commit

* Makes Nimbus the new default landing page
* Adds a link back to legacy in the Nimbus UI

Screen Shot 2022-06-08 at 11 22 53 AM

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageHome/index.tsx
@@ -153,6 +153,14 @@ const PageHome: React.FunctionComponent<PageHomeProps> = () => {
         .
       </Alert>
 
+      <Alert variant="warning" className="mb-4">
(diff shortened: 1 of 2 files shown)


tiftran (approved):
🎉

PR #7389, opened 2022-06-09
fix #7152 feat(nimbus): Prevent country targeting below min version supported

### LINKED ISSUE #7152: Prevent launching experiments targeting countries below minimum supported versions
When the `regions` key is added to the SDK targeting context, we can add a check to the NimbusReviewSerializer to prevent launching if an experiment targeting mobile clients includes country targeting and targets a version below the minimum supported version.

[AS v93.0.3](https://github.com/mozilla/application-services/releases/tag/v93.0.3) adds targeting attributes for `country`. We need to figure out which version(s) of iOS and Fenix started to use this version, keeping in mind there might have been reverts.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2325)

yashika-khurana:
Because

* We want to prevent launching experiments targeting countries below minimum supported versions
 

This commit

*  Add the minimum version required to support countries

          FENIX: FIREFOX_102,
          FOCUS_ANDROID: FIREFOX_102,
          IOS: FIREFOX_101,
          FOCUS_IOS: FIREFOX_101,

-     Validation in NimbusReviewSerializer to support country targeting
-     Test cases to support new changes

fixes #7152

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -445,7 +445,9 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
             ),
         ]
     )
-    def test_valid_experiments_with_all_countries(self, application, firefox_version):
+    def test_valid_experiments_with_country_unsupported_version(
+        self, application, firefox_version
+    ):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.CREATED,
             application=application,
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -360,7 +360,7 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
             ),
         ]
     )
-    def test_valid_experiments_supporting_countries_versions_selecting_all_countires_default(
+    def test_valid_experiments_supporting_countries_versions_default_as_all_countries(
         self, application, firefox_version
     ):
 
@@ -369,6 +369,7 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
             application=application,
             channel=NimbusExperiment.Channel.RELEASE,
             firefox_min_version=firefox_version,
+            countries=[],
         )
 
         serializer_1 = NimbusReviewSerializer(
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -370,7 +370,7 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
             channel=NimbusExperiment.Channel.RELEASE,
             firefox_min_version=firefox_version,
         )
-        experiment_1.save()
+
         serializer_1 = NimbusReviewSerializer(
             experiment_1,
             data=NimbusReviewSerializer(
@@ -389,7 +389,7 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
             firefox_min_version=firefox_version,
             countries=[country.id],
         )
-        experiment_2.save()
+
         serializer_2 = NimbusReviewSerializer(
             experiment_2,
             data=NimbusReviewSerializer(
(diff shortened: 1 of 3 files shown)


app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -340,6 +340,148 @@ def test_invalid_experiments_with_specific_languages(
...
+        self, application, firefox_version
+    ):
+
+        experiment_1 = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            application=application,
+            channel=NimbusExperiment.Channel.RELEASE,
+            firefox_min_version=firefox_version,
+        )
+        experiment_1.save()
```
jaredlockhart, line 373: You don't need to `save` here, it's already saved by the `create_with_lifecycle` method

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -340,6 +340,148 @@ def test_invalid_experiments_with_specific_languages(
...
+        serializer_1 = NimbusReviewSerializer(
+            experiment_1,
+            data=NimbusReviewSerializer(
+                experiment_1,
+                context={"user": self.user},
+            ).data,
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer_1.is_valid())
```
jaredlockhart, line 382: What is this case testing?  It seems like there's two tests in here, and one of them doesn't involve countries?

jaredlockhart (changes requested):
Yep looks good except one test seems to be doing some unnecessary work?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
yashika-khurana: It means the above-mentioned versions are valid, and we can test two scenarios based on that:

- When you select a specific country with valid versions
- When you don't select a specific country i.e. All countries with valid versions

jaredlockhart: Right but I mean I expect a single test to have

1) Setup
2) Assertions

for a single case.  So if you're doing that cycle twice in a single experiment, when an assertion fails, you don't know which of the two cases it's trying to say are wrong, so if you want to keep both of these cases I'd suggest splitting them into two separate experiments 🙏

yashika-khurana: sure sounds like a good plan 💃 🙏  thanks 😊

jaredlockhart: Oh and if in one of your cases you specifically want to test the 'all countries' case, you'd have to explicitly specify `countries=[]` in the ExperimentFactory otherwise it will pick some at random:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/factories.py#L321-L322

yashika-khurana: got it all done, thank you @jaredlockhart 🙏

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -340,6 +340,174 @@ def test_invalid_experiments_with_specific_languages(
...
+
+        serializer = NimbusReviewSerializer(
+            experiment,
+            data=NimbusReviewSerializer(
+                experiment,
+                context={"user": self.user},
+            ).data,
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer.is_valid())
```
jaredlockhart, line 465: So isn't this a duplicate of `test_valid_experiments_supporting_countries_versions_default_as_all_countries` above now?

yashika-khurana: Kind of no, I was thinking in this way- the above one includes valid versions for the mobile client, whereas this includes invalid versions but this shows that you can select all countries by default for this.

jaredlockhart: Ah yes, then I would included something like `with_language_unsupported_version` in the name 🙏

yashika-khurana: okey dokey, all set 🚀 , thank you 🙏

jaredlockhart (approved):
👏 👏 👏 👏 👏 👏  Great job @yashika-khurana.  Thank you for updating the tests to be clear and concise!  Looks great let's get country targeting going! 🥳

PR #7392, opened 2022-06-09
fix #7391 fix(nimbus): retrieve expected outcome metrics from jetstream and allow all stats types

### LINKED ISSUE #7391: Nimbus should handle partial results for outcome metrics without failing
Currently Jetstream does not send all outcome metrics (e.g., if the metric was not processed), and Nimbus does not check for existence of metrics results before attempting to display them. We should validate the metrics in Nimbus before attempting to display in order to avoid errors, and to avoid confusing users with confusing/incomplete results.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2499)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/24842">image-20220608-213146.png</a>

mikewilli:
Because

- Nimbus could potentially expect different outcome metrics from what Jetstream actually processed and sends
- Nimbus was filtering out processed outcome metrics from results due to a too-restrictive filter on statistics options based on the type of metric (primary, default, etc)
- there is no reason that outcome metrics could not be any of the Statistic options

This commit

- requests Jetstream metadata to update the outcome metrics expected by Nimbus
- allows outcome metrics to be any of the Statistic options

### DIFF
--- a/app/experimenter/jetstream/client.py
+++ b/app/experimenter/jetstream/client.py
@@ -59,23 +59,19 @@ def get_results_metrics_map(
         )
     )
 
-    try:
-        valid_metrics = set(
-            chain.from_iterable(
-                [
-                    experiment_metadata["outcomes"][slug]["metrics"]
-                    + experiment_metadata["outcomes"][slug]["default_metrics"]
-                    for slug in experiment_metadata["outcomes"]
-                ]
-            )
+    metrics_set_from_jetstream = set(
+        chain.from_iterable(
+            [
+                experiment_metadata["outcomes"][slug]["metrics"]
+                + experiment_metadata["outcomes"][slug]["default_metrics"]
+                for slug in experiment_metadata["outcomes"]
+            ]
         )
-    except (TypeError, KeyError):
-        # could not retrieve metadata from Jetstream
-        valid_metrics = None
+    )
 
     for metric in primary_outcome_metrics:
         # validate against jetstream metadata unless we couldn't get it
-        if valid_metrics is None or metric.slug in valid_metrics:
+        if metric.slug in metrics_set_from_jetstream:
             RESULTS_METRICS_MAP[metric.slug] = set([Statistic.BINOMIAL])
             primary_metrics_set.add(metric.slug)
 
--- a/app/experimenter/jetstream/client.py
+++ b/app/experimenter/jetstream/client.py
@@ -37,7 +37,9 @@ def get_metadata(slug):
     return load_data_from_gcs(path)
 
 
-def get_results_metrics_map(data, primary_outcome_slugs, secondary_outcome_slugs):
+def get_results_metrics_map(
+    data, primary_outcome_slugs, secondary_outcome_slugs, outcomes_metadata
+):
     # A mapping of metric label to relevant statistic. This is
     # used to see which statistic will be used for each metric.
     RESULTS_METRICS_MAP = {
@@ -57,9 +59,17 @@ def get_results_metrics_map(data, primary_outcome_slugs, secondary_outcome_slugs
         )
     )
 
+    metrics_set = set()
+    for slug in primary_outcome_slugs:
+        for metric in outcomes_metadata[slug]["metrics"]:
+            metrics_set.add(metric)
+        for metric in outcomes_metadata[slug]["default_metrics"]:
+            metrics_set.add(metric)
+
     for metric in primary_outcome_metrics:
-        RESULTS_METRICS_MAP[metric.slug] = set([Statistic.BINOMIAL])
-        primary_metrics_set.add(metric.slug)
+        if metric.slug in metrics_set:
+            RESULTS_METRICS_MAP[metric.slug] = set([Statistic.BINOMIAL])
+            primary_metrics_set.add(metric.slug)
 
     for outcome_slug in secondary_outcome_slugs:
         RESULTS_METRICS_MAP[outcome_slug] = set([Statistic.MEAN])
@@ -111,9 +121,11 @@ def get_experiment_data(experiment):
     windows = ["daily", "weekly", "overall"]
     raw_data = {}
 
+    experiment_metadata = get_metadata(recipe_slug)
+
     experiment_data = {
         "show_analysis": settings.FEATURE_ANALYSIS,
-        "metadata": get_metadata(recipe_slug),
+        "metadata": experiment_metadata,
     }
 
     for window in windows:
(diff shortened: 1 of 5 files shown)


jaredlockhart:
Fantastic @mikewilli , this is a nice clear fix.  Now comes the fun part, updating the tests XD

Right now there's just some big hard coded test fixtures so you'll have to go wading through those, but when we have some down time I think it would help if we converted those into proper test factories to make testing easier.

app/experimenter/jetstream/client.py
```diff
@@ -57,9 +59,25 @@ def get_results_metrics_map(data, primary_outcome_slugs, secondary_outcome_slugs
...
+                [
+                    experiment_metadata["outcomes"][slug]["metrics"]
+                    + experiment_metadata["outcomes"][slug]["default_metrics"]
+                    for slug in experiment_metadata["outcomes"]
+                ]
+            )
+        )
+    except (TypeError, KeyError):
+        # could not retrieve metadata from Jetstream
+        valid_metrics = None
```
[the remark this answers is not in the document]
mikewilli: Yea, thinking about this a little more we should be able to make this less convoluted. The keys we are accessing should always exist as long as we can retrieve metadata from Jetstream. If we can't retrieve from Jetstream, we'll fail before getting here, so that's covered. If somehow we got here and the expected keys don't exist in the metadata, we should probably fail anyway because it means the metadata looks different than we expect.

I was also thinking that we would be OK passing through without validating against Jetstream metadata if the experiment_metadata was not available for some reason, but I'm not sure that's the case. The examples I've tried would all be fine doing this, but I don't know that it's a safe assumption. Better to just use the metadata if we have it, and fail if not (and like I mentioned earlier, we should likely be failing earlier if we don't have it anyway).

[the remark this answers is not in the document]
jaredlockhart: Yeah exactly, making assumptions about data and then completely falling over and erroring out to sentry is a completely valid strategy, especially in background tasks like this, so I'm totally good with that 👍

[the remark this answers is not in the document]
mikewilli: Question: we have this function earlier in `client.py`:

```
def load_data_from_gcs(path):
    if default_storage.exists(path):
        return json.loads(default_storage.open(path).read())
```

In the case that `default_storage.exists(path)` returns `False`, the metadata will be `None`, but we don't currently fail. Should we a) handle this in the code block referenced above, or b) remove the `default_storage.exists(path)` check and let an exception happen (leave the check but explicitly raise an exception in the else case)?

For now, I went with (a), but happy to update or discuss if there's something I missed.

mikewilli:
There was an issue with the original fix that caused all experiments' results pages to fail to load, instead displaying a CORS error from react. The check may not have been important for outcome metrics, but seems to have been needed for other results. Per Daniel's comment on the issue, the primary and secondary outcome metrics should not be restricted to a particular Statistics type, so I instead reverted to the original logic, but added the missing Statistics types to the allowed list for these metrics.

app/experimenter/jetstream/client.py
```diff
@@ -57,12 +59,32 @@ def get_results_metrics_map(data, primary_outcome_slugs, secondary_outcome_slugs
...
+    else:
+        BYPASS_JETSTREAM_CHECK = True
+
     for metric in primary_outcome_metrics:
-        RESULTS_METRICS_MAP[metric.slug] = set([Statistic.BINOMIAL])
-        primary_metrics_set.add(metric.slug)
+        # validate against jetstream metadata unless we couldn't get it
+        if BYPASS_JETSTREAM_CHECK or metric.slug in metrics_set_from_jetstream:
+            RESULTS_METRICS_MAP[metric.slug] = set(
+                [Statistic.BINOMIAL, Statistic.MEAN, Statistic.COUNT, Statistic.PERCENT]
```
jaredlockhart, line 80: If these two lists are the same, and static, can we move it up to the top and make it a constant?

app/experimenter/jetstream/client.py
```diff
@@ -57,12 +59,32 @@ def get_results_metrics_map(data, primary_outcome_slugs, secondary_outcome_slugs
...
+            chain.from_iterable(
+                [
+                    outcomes_metadata[slug]["metrics"]
+                    + outcomes_metadata[slug]["default_metrics"]
+                    for slug in outcomes_metadata
+                ]
+            )
+        )
+    else:
+        BYPASS_JETSTREAM_CHECK = True
```
jaredlockhart, line 74: For patterns like this another way to reduce the number of cases is treat the `else` case here as the default ie

```py
BYPASS_JETSTSREAM_CHECK = True
if outcomes_metadata is not None:
    BYPASS_JETSTREAM_CHECK = False
    ...
```

Just saves an extra clause.

jaredlockhart: But also good idea to set this check here like this.

Oh but also, sidebar, [all caps is usually used to denote a constant](https://peps.python.org/pep-0008/#constants).

app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -201,6 +202,10 @@ def open_file(filename):
         experiment = NimbusExperiment.objects.get(id=experiment.id)
         self.assertIsNone(experiment.results_data)
 
+        mock_get_metadata.return_value = {
+            "outcomes": {"default-browser": {"metrics": ["test"], "default_metrics": []}}
+        }
```
jaredlockhart, line 207: Does this recreate the failure we see in prod?  Testing this locally, reverting the changes to `client.py` still lets the tests pass?

Ideally we'd modify/extend the tests to force the case we see in sentry:

```py
IndexError

list index out of range

    experimenter/jetstream/models.py in append_conversion_count at line 261

                conversion_count = 0.0
                if None not in (population_count, conversion_percent):
                    conversion_count = population_count * conversion_percent
                absolute_primary_metric_vals.first.count = conversion_count
                absolute_primary_metric_vals.all[0].count = conversion_count  <<<<<<<<<
```

jaredlockhart (changes requested):
Looking good, nice and clear and simple.  Can we extend the tests to recreate the prod failure case?

app/experimenter/jetstream/tests/test_tasks.py
mikewilli: This fixes the existing test cases, but you're right -- it doesn't cover the failure fixed by this PR. I added another commit updating the tests and test data to cover this.

jaredlockhart (approved):
Fantastic, everything looks good and correct, tests recreate the expected failure, fix looks nice and clean.  Great job @mikewilli 🎉 🎉 🎉 🎉 🎉  Long road to get here, but huge impact change 🙏 Thank you!

PR #7405, opened 2022-06-14
fix #7338 chore(nimbus): transform locales to languages for mobile experiments

### LINKED ISSUE #7388: The Locale field of older mobile experiments has been removed
**[Prerequisites]:**

* Have an older experiment with Locales filters set (e.g. [+Stage experiment+](https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/locales-and-languages-in-fenix/details) or [+Prod Experiment+](https://experimenter.services.mozilla.com/nimbus/viewpoint-survey-beta-validation-control-fix/details))

**[Steps to reproduce]:**

1. Open a page from prerequisites on the “Details” section.
1. Observe the “Audience” subsection.

**[Expected result]:**

* Older experiments using the “Locales” filter retain the populated “Locales” fields.

**[Actual result]:**

* Older experiments using the “Locales” filter display a “Languages” field displaying the “All Languages” option.

**[Notes]:**

* The issue does not affect the “Full targeting expression” field or the recipe itself.

* The issue is not reproducible for desktop locales.

* Attached is a screenshot of the issue.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-2508)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/24859">Screenshot 2022-06-09 at 13-02-20 cmuresan ios focus qa-1494 05_13_2022 Locale – Details Experimenter.png</a>

yashika-khurana:
Because

* Since we added the languages field for the mobile client, but still, we have use of locales in older experiments

This commit

* Migrate the locales fields for mobile experiments to the Languages field, if the code exists for that Language, else it will not add to the languages field.

fix #7388


app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,37 +1,79 @@
...
+        # get desktop experiment
+        desktop_experiment = NimbusExperiment.objects.get(
+            application=Experiment.Application.DESKTOP
+        )
+
+        self.assertEqual(
+            desktop_experiment.locales.all()[0].code, Locale.objects.get(code="en").code
+        )
+
+        # get mobile experiment included locale filed
```
eliserichards, line 64: Nit: 

```suggestion
        # get mobile experiment included locale field
```

app/experimenter/experiments/migrations/0212_alter_nimbusexperiment_languages_field_for_mobile_client.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def update_languages_field_for_mobile_client(apps, schema_editor):
+    NimbusExperiments = apps.get_model("experiments", "NimbusExperiment")
+
+    for experiment in NimbusExperiments.objects.all():
+        if experiment.application != NimbusExperiment.Application.DESKTOP:
+            for locale in experiment.locales.all():
+                locale_code = locale.code[:2]
+                language = Language.objects.filter(code=locale_code).first()
```
eliserichards, line 14: Is it possible to have more than one language per locale?

app/experimenter/experiments/migrations/0212_alter_nimbusexperiment_languages_field_for_mobile_client.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations
+
+from experimenter.base.models import Language
+from experimenter.experiments.models import NimbusExperiment
+
+
+def update_languages_field_for_mobile_client(apps, schema_editor):
+    NimbusExperiments = apps.get_model("experiments", "NimbusExperiment")
```
jaredlockhart, line 8: This should be singular 🙏

app/experimenter/experiments/migrations/0212_alter_nimbusexperiment_languages_field_for_mobile_client.py
```diff
@@ -0,0 +1,30 @@
...
+
+    for experiment in NimbusExperiments.objects.all():
```
jaredlockhart, line 10: We don't want this to run for experiments that have already launched since then the database wouldn't agree with the recipe that was sent to users.  We should filter this to only draft 🙏

app/experimenter/experiments/migrations/0212_alter_nimbusexperiment_languages_field_for_mobile_client.py
```diff
@@ -0,0 +1,30 @@
...
+        if experiment.application != NimbusExperiment.Application.DESKTOP:
+            for locale in experiment.locales.all():
+                locale_code = locale.code[:2]
+                language = Language.objects.filter(code=locale_code).first()
+                if language:
+
+                    experiment.languages.add(language.id)
+            experiment.locales.clear()
+            experiment.save()
```
jaredlockhart, line 19: I think `save` is redundant here since both `locales` and `languages` are related entities so the `add` and `clear` calls both already hit the database.  `save` will only persist changes to local fields right on the instance.

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,37 +1,79 @@
...
         User = self.old_state.apps.get_model("auth", "User")
+        Locale = self.old_state.apps.get_model("base", "Locale")
+        Language = self.old_state.apps.get_model("base", "Language")
         NimbusExperiment = self.old_state.apps.get_model(
             "experiments", "NimbusExperiment"
         )
         user = User.objects.create(email="test@example.com")
-        NimbusExperiment.objects.create(
+        locale_en = Locale.objects.create(code="en", name="English")
+        Language.objects.create(code="en", name="English")
```
jaredlockhart, line 24: Do you also want to test the case of a four character locale code ie `en-US` or similar, in addition to the two character case?

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,37 +1,79 @@
...
+        locale_ts = Locale.objects.create(code="ts", name="does not exist")
+
+        desktop_experiment = NimbusExperiment.objects.create(
```
jaredlockhart, line 27: Since we only want to run this for draft experiments, we should probably put one in here with a mobile application and some locales and then assert that it doesn't get modified 🙏

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,37 +1,79 @@
...
-        self.assertFalse(
-            NimbusExperiment.objects.filter(targeting_config_slug="").exists()
+
+        # get desktop experiment
+        desktop_experiment = NimbusExperiment.objects.get(
+            application=Experiment.Application.DESKTOP
+        )
+
+        self.assertEqual(
+            desktop_experiment.locales.all()[0].code, Locale.objects.get(code="en").code
```
jaredlockhart, line 61: We know we're only setting up a single locale on this experiment so hardcoding `[0]` here is technically fine, but in general when comparing two collections of data it's better to convert both collections to `set`s and compare that way, just as a general pattern.

jaredlockhart (dismissed):
Yes generally looks correct, just a couple cases to check for ie the draft case 🙏 Thanks @yashika-khurana 🎉

app/experimenter/experiments/migrations/0212_alter_nimbusexperiment_languages_field_for_mobile_client.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def update_languages_field_for_mobile_client(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+
+    for experiment in NimbusExperiment.objects.all():
+        if (
+            experiment.application != Experiment.Application.DESKTOP
+            and experiment.status == NimbusConstants.Status.DRAFT
+        ):
```
jaredlockhart, line 15: This filtering should be done above in the queryset?

jaredlockhart (changes requested):
Looking good, one last change 🙏

app/experimenter/experiments/migrations/0212_alter_nimbusexperiment_languages_field_for_mobile_client.py
```diff
@@ -0,0 +1,32 @@
...
+from experimenter.experiments.models import NimbusExperiment as Experiment
+
+
+def update_languages_field_for_mobile_client(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+
+    for experiment in NimbusExperiment.objects.filter(
+        status=NimbusConstants.Status.DRAFT
+    ):
+        if experiment.application != Experiment.Application.DESKTOP:
```
jaredlockhart, line 14: This can be done in the queryset with `.exclude(application='firefox-desktop')`

In general anything that can be filtered on in the queryset should be done there rather than in memory so that
1) all the filters are in one place if you need to change them later
2) the database is faster than Python (not that really matters but just worth pointing out)
3) fewer cases to handle in code with logic

jaredlockhart (changes requested):

jaredlockhart (approved):
Okay yes, tested locally, code looks correct, the test cases are all good, I think this is good to go thank you for addressing all the feedback, looks great good job @yashika-khurana 🎉 🎉 🎉

PR #7414, opened 2022-06-15
fixes #7409 chore(nimbus): Allow integration testing against multiple firefox versions

### LINKED ISSUE #7409: Run circleci integration tests against multiple Firefox Versions
We currently test on Nightly Firefox, we should extend that to Beta and Release

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2514)

jrbenny35:
Because

* We are currently testing only on nightly (and it's kinda old). We would like to run targeting tests against release and beta, and then use nightly for everything else.

This commit

* Adds Firefox Beta and Release to circleci
* Adds a script to build a "fresh" nightly image for testing
* Creates a cron job to build Firefox Release and Beta weekly at 00:00 UTC on Sunday.


jrbenny35:
Running into a little issue locally.

jaredlockhart:
Okay I took a quick first look at this @jrbenny35 and wow, 1) it all works 2) the way you've setup the version discovery and caching is very clever, I was actually surprised to see the way it works and checks versions and reuses the cache, very well done.

I'd like to chat over zoom so I can understand better exactly how it's working but my quick high level feedback without understanding enough of how it works is

1) Can we move the build task into a separate circle cron task that fires maybe once a week and have the actual tests always depend on the cache?

2) Now I think we need to reorganize the way our tests are grouped, right now it looks like it's running the entire 'desktop' suite in both beta and release but inside those suites there's three subgroups of tests

- Testing Experimenter UI
- Testing Experimenter/Remote Settings integration
- Testing targeting expressions inside the desktop evaluation context

I think only the last one needs to be done across different versions/channels, so is there a way for us to reorganize how they're grouped so we run the first two in release only, and the last one across channels/versions (right now the latest version but later some set of grouped versions)?  Primarily this will save costs and time during CI runs so we don't blow up our circle budget.  But I think that should be a straight forward change on top of this foundation.

Overall really great work and another step closer to being able to maintain this system healthily in the long term.  Fantastic work @jrbenny35 🙏 🙏 🙏 🙏 🙏

jaredlockhart:
Oh also could you please fill in the PR description from the [commit guidelines](https://github.com/mozilla/experimenter/blob/main/contributing.md#body) pls 🙏

jaredlockhart:
Let's try adding nightly back in here 🙏

jaredlockhart (approved):
Yep, played around with this and looks like it's doing all the things I hoped it would so I think this is good to go, great job @jrbenny35 🎉 🎉 🎉

PR #7416, opened 2022-06-16
fix #7410 feat(nimbus): add enrollmentEndDate to V6 API

### LINKED ISSUE #7410: add enrollmentEndDate to V6 API
From @danielkberry:

"Jetstream bases the automated analysis off that proposedEnrollment  field, not whatever the actual enrollments were. Oftentimes owners say they’re not getting enough enrollments and leave enrollments open longer than planned, but those later enrollees aren’t included for the automated analysis."

To assist jetstream in setting the right enrollment period we should add a field to the V6 API called 'enrollmentEndDate' that sets the date enrollments actually ended which jetstream can then use to include the correct set of clients when performing analysis.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2515)

jaredlockhart:
Because

* We currently expose the proposed enrollment period in the V6 API
* Jetstream uses this to determine the enrollment window to collect the set of enrolled clients
* Some experiments enrollment period will vary from the proposed enrollment period depending on the enrollment characteristics
* Jetstream has no way to determine the actual enrollment period and so may miss some clients in its analysis

This commit

* Adds a computed_enrollment_end_date property to NimbusExperiment model
* Exposes that property in the V6 API as enrollmentEndDate

### DIFF
--- a/app/experimenter/experiments/tests/test_models.py
+++ b/app/experimenter/experiments/tests/test_models.py
@@ -865,6 +865,35 @@ class TestNimbusExperiment(TestCase):
             expected_days,
         )
 
+    def test_computed_enrollment_end_date_returns_start_date_plus_enrollment_days(self):
+        start_date = datetime.date(2022, 1, 1)
+        enrollment_end_date = start_date + datetime.timedelta(days=3)
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_APPROVE,
+            proposed_enrollment=7,
+        )
+
+        experiment.changes.filter(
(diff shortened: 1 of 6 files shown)


app/experimenter/experiments/tests/test_models.py
```diff
@@ -865,6 +865,35 @@ def test_computed_enrollment_days_returns_fallback_while_pause_pending_approval(
             expected_days,
         )
 
+    def test_computed_enrollment_end_date_returns_start_date_plus_enrollment_days(self):
```
eliserichards, line 868: 👏

eliserichards (approved):
Looks good! Love the tests! 👍

PR #7430, opened 2022-06-20
fix #7421 feat(nimbus): add sticky clause to jexl targeting

### LINKED ISSUE #7421: Add sticky clause to NimbusExperiment JEXL
Let’s update the JEXL generation for NimbusExperiment to include the sticky clause

`experiment.slug in activeExperiments`

such that the entire targeting expression follows this structure

`(({sticky} || {enrollment_targeting}) && {continuous_targeting})`

Enrollment Targeting

* Advanced targeting expression
* Application
* Min version
* Locales
* Languages
* Countries

Continuous Targeting

* Max version

And update the tests as necessary.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2527)

jaredlockhart:
Because

* It is a very common pattern for experiments to want to have "sticky" enrollment
* Sticky enrollment means that once a client is enrolled in an experiment they won't be unenrolled until the experiment ends
* Since clients constantly re evaluate the targeting expression, if some clause in the targeting may become false for an enrolled client during the experiment, they would tthen unenroll
* The sticky expression checks if a client is already enrolled in a given experiment and then ignores other targeting clauses
* Sticky enrollment should supercede some targeting expressions including
    - Min version
    - Locale/Language/Country
    - Advanced targeting
* Sticky enrollment should not supercede other targeting expressions including
    - Channel
    - Max version

This commit

* Adds the sticky clause to the auto generated JEXL targeting expression
* Updates the tests


jaredlockhart:
@Mardak Yeah I'm going to fix all the constants separately in here https://github.com/mozilla/experimenter/issues/7440

PR #7446, opened 2022-06-21
fix #7440 feat(nimbus): migrate sticky targeting configs

### LINKED ISSUE #7440: Remove sticky clause from advanced targeting configs
Once we have a sticky enrollment clause as part of the auto generated JEXL it will no longer be necessary to include it in the advanced targeting.  We should remove it from the advanced targeting constants.  We’ll also have to migrate any past experiments which used an advanced targeting that included a sticky clause to have is_sticky set to true.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2535)

jaredlockhart:
Because

* Now that the sticky targeting clause is moved right into the NimbusExperiment targeting method
* There's no need for the sticky clause to be included in individual targeting configs

This commit

* Removes the sticky clauses from all desktop and mobile targeting configs
* Removes the now redundant 'pip_never_used_sticky' targeting config and migrates experiments that used it to 'pip_never_used'
* Migrates all existing experiments that use a previously sticky targeting config to have is_sticky=True
* Discovered that some JEXL clauses were not being evaluated at all as part of the integration tests becuase of short circuiting behaviour in the JEXL evaluator
* Reworked the SDK integration test to pass values in as their expected types rather than as strings which was triggering a suppressed error about comparing strings to integers
* Filed https://github.com/mozilla/experimenter/issues/7449 as a followup


app/tests/integration/nimbus/app_contexts.json
```diff
@@ -1,149 +1,144 @@
...
-              "installation_date": 1641967200,
-              "home_directory": null,
-              "custom_targeting_attributes":
-                {
-                  "is_already_enrolled": "true",
-                  "days_since_update": "1",
-                  "days_since_install": "1"
-                }
+  "query_result": {
+    "id": 15734423,
```
jrbenny35, line 3: why is this changed?

jaredlockhart: Oh vscode moved the indentation around

jaredlockhart: The actual change was trying to get the `custom_targeting_attributes` to have the expected types, ie integers and bools, rather than strings.  But it turns out this whole thing gets parsed and expects only string types.  But you can pass in `custom_targeting_attributes` separately into the helper encoded as JSON with the actual types, so I ended up just taking it out of here entirely.

yashika-khurana (approved):
LGTM! 🚀

app/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,105 +1,189 @@
...
+        for desktop_sticky_targeting in DESKTOP_STICKY_TARGETING_SLUGS:
+            NimbusExperiment.objects.create(
+                owner=user,
+                name=desktop_sticky_targeting,
+                slug=desktop_sticky_targeting,
+                application=Experiment.Application.DESKTOP,
+                status=NimbusConstants.Status.DRAFT,
+                targeting_config_slug=desktop_sticky_targeting,
+                is_sticky=False,
+            )
```
dmose, line 88: Should this be `is_sticky=True`?

dmose:
The desktop sticky migration stuff looks good to me.  r=dmose

jrbenny35 (approved):

PR #7451, opened 2022-06-27
fix #7449 bug(nimbus): test jexl subexpressions for mobile

### LINKED ISSUE #7449: Invalid variables in left hand of JEXL or expression are not evaluated
While working on #7440 I noticed that the left hand operands of an `||` expression in JEXL are not evaluated in either the Rust or JS implementations, see below:

In Rust:
```py
(Pdb) targetting_helper.eval_jexl('asdf || true')
True
(Pdb) targetting_helper.eval_jexl('asdf || false')
False
(Pdb) targetting_helper.eval_jexl('true || asdf')
True
(Pdb) targetting_helper.eval_jexl('false || asdf')
*** nimbus_rust.NimbusError.EvaluationError: EvaluationError: Identifier 'asdf' is undefined
```

In JS:
```js
await targetingContext.evalWithDefault('asdf || true');
true
await targetingContext.evalWithDefault('asdf || false');
false
await targetingContext.evalWithDefault('true || asdf');
true
await targetingContext.evalWithDefault('false || asdf');
null
```

This means that if someone includes an invalid variable in their JEXL expression it will not be caught by the integration tests.  In theory we could make changes to both the JS and Rust implementations of the JEXL evaluator to change this behaviour, but in the mean time we can account for this by tokenizing the expression and evaluating them each individually in addition to the entire expression.

Let's update both the desktop and sdk integration tests to account for this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2553)

jaredlockhart:
Because
    
* We recently learned that JEXL will not evaluate the left hand side of an || expression
* When we validate JEXL expressions in our integration tests we want to know if any part of the expression is invalid
    
This commit
    
* Parses all JEXL expressions in the integration tests and collects all their subexpressions
* Evaluates every subexpression in the integration tests in the mobile SDK
* Desktop support requires changes to the JEXL runtime and will have to happen in a followup


app/tests/integration/nimbus/jexl/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        return f"{to_str(node.subject)}[{to_str(node.expression)}]"
+    else:
+        raise Exception(f"Unhandled node type: {node}")
+
+
+def collect_exprs(expr):
+    """
+    Collect all the sub expressions of a JEXL expression
+    """
+    nodes = [DefaultParser().parse(expr)]
```
jrbenny35, line 61: what is a "node" supposed to be?

jrbenny35 (changes requested):
Some questions

app/tests/integration/nimbus/jexl/__init__.py
jaredlockhart: [Node](https://en.wikipedia.org/wiki/Node_%28computer_science%29) is the word used to describe an element of a tree structure, which is what the parser returns.

app/tests/integration/nimbus/jexl/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    exprs = []
+
+    while nodes:
+        node = nodes.pop()
+
+        if node.children:
+            nodes += list(node.children)
```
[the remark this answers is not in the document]
jaredlockhart: You mean like `[node.children]`?  That would be a list of one element, which is the generator returned by `node.children`.  You could maybe do something like `[*node.children]` but I think that's less readable than just `list(...)`.

yashika-khurana (commented):
LGTM!, just I am wondering if we have any test case to cover try/Exception block? PS:  Any test case invalidating the jexl expression

jrbenny35 (approved):

PR #7467, opened 2022-06-29
Add check to circleci skip job if no changes are detected.

b4handjr:
Because

* With our extended test suite we are using a lot of circleci credits now.

This commit

* Will check the git diff against the main branch to see which file paths have changes and run the needed tests
* The default check step is always run in full (for now).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,16 +99,6 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Check file paths
-          command: |
-            if git diff --name-only main HEAD | grep -E 'app/experimenter/integration'
-              then
-                echo "Continuing"
-              else
-                echo "No targeting changes, skipping"
-                circleci-agent step halt
-            fi
-      - run:
           name: Run integration tests
           command: |
             PYTEST_ARGS=$(circleci tests split "app/tests/integration/nimbus/parallel_pytest_args.txt")


.circleci/config.yml
```diff
@@ -31,6 +31,16 @@ jobs:
       PYTEST_ARGS: -k DESKTOP -m run_targeting -n 2
     steps:
       - checkout
+      - run:
+          name: Check file paths
+          command: |
+            if git diff --name-only main HEAD | grep -E 'app/experimenter/targeting|app/experimenter/experiments'
```
jaredlockhart, line 37: It's too bad we can't narrow it down even further.... I guess we could move the targeting logic out of the experiment model and into a mixin in the targeting app....  Maybe we can do that as a followup.

jaredlockhart (changes requested):
Wow this is gonna save a LOT of time and credits.  I think we just need to take the UI filter off that one job until we can split it up more granularly, but the targeting ones are perfect.

.circleci/config.yml
jrbenny35: We can narrow it down to the file name itself can't we?

jaredlockhart: No it's like specific methods inside the file that we care about in this context, but I was also already thinking of moving that bit of logic into the targeting app anyways.

.circleci/config.yml
```diff
@@ -78,6 +98,16 @@ jobs:
       FIREFOX_VERSION: :latest
     steps:
       - checkout
+      - run:
+          name: Check file paths
+          command: |
+            if git diff --name-only main HEAD | grep -E 'app/experimenter/integration|app/experimenter/nimbus-ui'
```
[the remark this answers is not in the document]
jaredlockhart: Yeah like right now we have one job that uses the 'parallel' option to invoke pytest with many different configurations that each are targeting different aspects of the nimbus-ui, the kinto(remote settings) integration, some mobile configuration, so the job in its entirety is so broad there's no way to say which paths it applies to.  

So either we'd have to move the path/grep check inside each parallel job, but then we'd have some checks at the in this file and others in other files, or we can just split each of those parallel jobs into a whole separate circle job, so we have one for all the remote settings tests that is gated on `experiments|kinto` apps, and one that's testing just the nimbus-ui maybe gated on `experiments|nimbus-ui`.  That way we can keep all these path diff checks in one place.

.circleci/config.yml
```diff
@@ -78,6 +98,16 @@ jobs:
       FIREFOX_VERSION: :latest
     steps:
       - checkout
+      - run:
+          name: Check file paths
+          command: |
+            if git diff --name-only main HEAD | grep -E 'app/experimenter/integration'
```
jaredlockhart, line 104: 1) The path `app/experimenter/integration` doesn't exist?
2) My point was that we can not gate this circle job on any path until we more clearly separate it into more specific groups of tests

jaredlockhart (changes requested):

.circleci/config.yml
jrbenny35: oooooooooh okay, so we run these everytime

PR #7470, opened 2022-06-29
fix #7469 chore(nimbus): simplify google auth process for local jetstream dev

### LINKED ISSUE #7469: Consolidate Google Cloud auth for local Jetstream
We have the Google Cloud authentication for Jetstream as a command in the Makefile, which spawns two browser tabs. It would be nice to consolidate these so that only one tab is launched.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2579)

mikewilli:
Because

* `make auth_gcloud` launches two separate browser tabs
* there is a simpler gcloud command to consolidate the authentication

This commit

* consolidates the commands run by `make auth_gcloud` into a single command that launches a single tab

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,8 +63,7 @@ secretkey:
 	openssl rand -hex 24
 
 auth_gcloud:
-	gcloud auth login
-	gcloud auth application-default login
+	gcloud auth login --update-adc
 
 jetstream_config:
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Neat!

PR #7471, opened 2022-06-29
Fix weekly firefox builds.

jrbenny35:
Because:
* The weekly builds for our firefox cache are broken

This commit:
* Adds the crucial step of comparing the version files created to see if we need to build firefox.


jrbenny35:
With how long the cache lives...we might need to have it build release/beta every 14 days just to be sure

PR #6635, opened 2021-11-05
fix #6625 fix #6627 feat(nimbus): initial support for is_rollout flag in experiment data model

lmorchard:
Because:

* we want to distinguish rollouts from normal experiments

This commit:

* add is_rollout boolean field to experiment DB model

* migration to add is_rollout=False to all experiments

* exposes isRollout in experiments for GQL queries in v5 API

* exposes isRollout in experiment DTO for v6 API

* adds a is_rollout column to admin list table for nimbus experiments

Issue #183, opened 2017-08-16
Create an ExperimentChangeLog at experiment creation time


jaredkerim:
When someone creates a new experiment, we should create an ExperimentChangeLog to show that they created it.

This will obviate the need for the datetime fields on the experiment itself.

jaredkerim:
This was done long ago.

PR #7476, opened 2022-06-30
fix #6625 feat(nimbus): Add isRollout to v6 api

### LINKED ISSUE #6625: Add is_rollout bool to experiment model
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1856)

eliserichards:
Because...

* [nimbus-shared #183](https://github.com/mozilla/nimbus-shared/issues/183) added `isRollout` to the schema, and #6635 added `isRollout` to the v5 API

This commit...

* adds `isRollout` to the v6 API so that it can be accessed by jetstream/clients


app/experimenter/docs/openapi-schema.json
```diff
@@ -3040,6 +3040,10 @@
             "type": "string",
             "readOnly": true
           },
+          "isRollout": {
+            "type": "string",
```
eliserichards, line 3044: Question: should this be `string` or `boolean`? I was going off the example of `isEnrollmentPaused` right above, which is boolean other places but string here 🤔

brennie: The Experiment JSONSChema has boolean

https://github.com/mozilla/nimbus-shared/blob/main/types/experiments.ts#L50-L57

yashika-khurana (dismissed):
LGTM, 🎉  Might be worth considering a test case with `isRollout` set to `True`

app/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -395,3 +396,31 @@ def test_serializer_outputs_empty_targeting(self):
...
+            publish_status=NimbusExperiment.PublishStatus.APPROVED,
+            targeting_config_slug=NimbusExperiment.TargetingConfig.NO_TARGETING,
+            application=NimbusExperiment.Application.FENIX,
+            firefox_min_version=NimbusExperiment.Version.FIREFOX_94,
+            is_rollout=False,
+        )
+
+        serializer = NimbusExperimentSerializer(experiment)
+        self.assertEqual(serializer.data["isRollout"], False)
+        check_schema("experiments/NimbusExperiment", serializer.data)
```
jaredlockhart, line 426: I don't think you need these tests, the field is using entirely built in serializer functionality so just updating the above tests that the field is included should be sufficient.

app/experimenter/docs/openapi-schema.json
jaredlockhart: Yeah that's a bit wonky.  We should look into that separately.

jaredlockhart (approved):
Approved thnx @eliserichards !  You can probably take those two extra test cases out but up to you.

PR #7486, opened 2022-07-06
fix #7475 feat(nimbus): Handle schema load failure

### LINKED ISSUE #7475: Handle schema load failure in load_feature_configs
If we try to load a feature config schema from a path that doesn't resolve or results in a json decode error, we should just silently pass over it instead of failing to start all of experimenter.  See failure here:

https://sentry.io/organizations/mozilla/issues/3392208353/?query=is%3Aunresolved

Becuase this feature

https://hg.mozilla.org/mozilla-central/raw-file/default/toolkit/components/nimbus/FeatureManifest.yaml

is referring to this schema path

https://hg.mozilla.org/mozilla-central/raw-file/tip/browser/components/newtab/schemas/MessagingExperiment.schema.json

which is not found which is triggering a JSON decode error, so we should wrap this one line:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/features/management/commands/load_feature_configs.py#L24

in a try/except so that if, a feature exists in the database, and has a schema, and we fail to find the new one, we don't overwrite it and just fall back to what's in the database.

Test case is to use mock to force either an http request error or a json decode error here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/features/__init__.py#L55

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2581)

yashika-khurana:
Because

* If we try to load a feature config schema from a path that doesn't resolve or results in a json decode error, we should just silently pass over it instead of failing to start all of experimenter. See failure here:

https://sentry.io/organizations/mozilla/issues/3392208353/?query=is%3Aunresolved 

Because this feature

https://hg.mozilla.org/mozilla-central/raw-file/default/toolkit/components/nimbus/FeatureManifest.yaml

is referring to this schema path

https://hg.mozilla.org/mozilla-central/raw-file/tip/browser/components/newtab/schemas/MessagingExperiment.schema.json

which is not found which is triggering a JSON decode error

This commit

* Adds the try/except block to deal with JSON decode error
* Test case to verify new changes 

fixes #7475

### DIFF
--- a/app/experimenter/features/tests/test_load_feature_configs.py
+++ b/app/experimenter/features/tests/test_load_feature_configs.py
@@ -6,7 +6,10 @@ from django.test import TestCase
 from experimenter.experiments.models import NimbusExperiment, NimbusFeatureConfig
 from experimenter.experiments.tests.factories import NimbusFeatureConfigFactory
 from experimenter.features import Features
-from experimenter.features.tests import mock_valid_features
+from experimenter.features.tests import (
+    mock_invalid_remote_schema_features,
+    mock_valid_features,
+)
 
 
 @mock_valid_features
@@ -98,3 +101,18 @@ class TestLoadFeatureConfigs(TestCase):
             feature_config.description,
             "Some Firefox Feature",
         )
+
+
+@mock_invalid_remote_schema_features
+class TestLoadInvalidRemoteSchemaFeatureConfigs(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        Features.clear_cache()
+
+    def test_invalid_remote_schema_raises_json_error(self):
+        call_command("load_feature_configs")
+
+        for feature in Features.all():
+            with self.assertRaises(json.JSONDecodeError):
+                feature.get_jsonschema()
--- a/app/experimenter/features/tests/test_load_feature_configs.py
+++ b/app/experimenter/features/tests/test_load_feature_configs.py
@@ -1,8 +1,10 @@
 import json
 
+import mock
 from django.core.management import call_command
 from django.test import TestCase
 
+from experimenter.experiments.constants import NimbusConstants
 from experimenter.experiments.models import NimbusExperiment, NimbusFeatureConfig
 from experimenter.experiments.tests.factories import NimbusFeatureConfigFactory
 from experimenter.features import Features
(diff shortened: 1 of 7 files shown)


jaredlockhart (changes requested):
Yes good start!  But you missed one important thing, you didn't mock out the actual network request.  So this test is making a real network request to mozilla central and we just expect that that path doesn't actually resolve, which triggers the exception.  But this means the tests are now susceptible to network failure, or maybe that path actually does start to resolve to something in json, which means they can fail for reasons outside what they're testing.

There's an example of how to setup a remote json failure mock already here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/features/tests/test_features.py#L167-L175

So try using that in your test.

Also another thing to think about is whether the exception handling should happen in the management command or in the feature loader, since the line that might throw is here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/features/__init__.py#L56

but then the try/catch is in another file, so there's kind of a hidden/indirect relationship between the two, ie it's not clear at the call site that the call might throw, unlike in say Java where a method will explicitly declare which exceptions it might throw and forces you to handle them.

So you might handle the exception at the json load line and then send some other signal that it didn't load to handle in the management command like returning None instead of a string?

But good start!

yashika-khurana:
@jaredlockhart  thank you for taking this over 🙏 🙏

app/experimenter/features/tests/test_load_feature_configs.py
```diff
@@ -110,9 +112,23 @@ def setUpClass(cls):
...
         Features.clear_cache()
 
-    def test_invalid_remote_schema_raises_json_error(self):
+    def setUp(self):
+        self.remote_schema = "{invalid json"
+
+        mock_requests_get_patcher = mock.patch("experimenter.features.requests.get")
+        self.mock_requests_get = mock_requests_get_patcher.start()
+        self.addCleanup(mock_requests_get_patcher.stop)
+        mock_response = mock.MagicMock()
```
eliserichards, line 121: 🪄 🧙 ✨ magic mock!

eliserichards (approved):
This is great! Thanks for the thorough explanation in the PR description + @jaredlockhart 's comments! 🚢

PR #7499, opened 2022-07-08
fixes #7497: Update atomicwrites dependency version to 1.4.1

### LINKED ISSUE #7497: Update atomicwrites dependency to 1.4.1
The dependency `atomicwrites` has published a new version and removed all old versions. We are currently using 1.4.0, so we need to update to the latest versions so we aren't breaking CI.

https://github.com/untitaker/python-atomicwrites/issues/61

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2597)

eliserichards:
Because...

* the `atomicwrites` package has published a new version

This commit...

* updates our dependencies to the latest version 1.4.1

#7497

### DIFF
--- a/app/tests/requirements.txt
+++ b/app/tests/requirements.txt
@@ -83,7 +83,7 @@ wcwidth==0.2.5 \
 more-itertools==8.12.0 \
     --hash=sha256:43e6dd9942dffd72661a2c4ef383ad7da1e6a3e968a927ad7a6083ab410a688b \
     --hash=sha256:7dc6ad46f05f545f900dd59e8dfb4e84a4827b97b3cfecb175ea0c7d247f6064
-atomicwrites==1.4.0 \
+atomicwrites==1.4.1 \
     --hash=sha256:6d1784dea7c0c8d4a5172b6c620f40b6e4cbfdf96d783691f2e1302a7b88e197 \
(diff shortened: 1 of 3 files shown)


eliserichards:
I'm not sure if this is the right way to do things.. when I updated the version number using poetry it's changing so many things in the `poetry.lock` file 👀 🤔

app/tests/requirements.txt
```diff
@@ -83,7 +83,7 @@ wcwidth==0.2.5 \
 more-itertools==8.12.0 \
     --hash=sha256:43e6dd9942dffd72661a2c4ef383ad7da1e6a3e968a927ad7a6083ab410a688b \
     --hash=sha256:7dc6ad46f05f545f900dd59e8dfb4e84a4827b97b3cfecb175ea0c7d247f6064
-atomicwrites==1.4.0 \
+atomicwrites==1.4.1 \
```
jaredlockhart, line 86: The poetry lock changes are expected yes.  For the integration tests, you have to use a tool called [hashin](https://pypi.org/project/hashin/) that's a little tricky to use I can show you and @yashika-khurana how it works this afternoon.

yashika-khurana:
@jaredlockhart  right they restored 🎉

PR #7509, opened 2022-07-11
7494/sticky warning

### LINKED ISSUE #7494: Add a warning for targeting configs that require sticky ie new users
Previously, some custom targeting configs were specifically designed to target new users using targeting expressions that will only be true the first time they are evaluated.  Those targeting configs explicitly used a sticky clause and so it was not possible to target first run users without also having sticky enrollment.  Now sticky enrollment is a separate checkbox and so it's possible for someone to create an experiment targeting new users without enabling sticky, and then every user would unenroll on their second session and so the experiment would be completely broken.

To prevent this we should

1) Add a new boolean field to the TargetingConfig object called stickyRequired
2) Explicitly flag each targeting config with whether they require sticky, the list of existing targeting configs that require sticky can be found [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/migrations/0214_migrate_sticky.py#L11-L38)
3) Add a check to the review serializer that checks if an experiment is using a targeting config that requires sticky but does not have sticky enabled and warns the user that their selected targeting expression requires sticky to function correctly (but does not prevent launching)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2594)

yashika-khurana:
Because

* Previously, some custom targeting configs were specifically designed to target new users using targeting expressions that will only be true the first time they are evaluated. Those targeting configs explicitly used a sticky clause and so it was not possible to target first run users without also having sticky enrollment. Now sticky enrollment is a separate checkbox and so it's possible for someone to create an experiment targeting new users without enabling sticky, and then every user would unenroll on their second session and so the experiment would be completely broken.

This commit

* Add a warnings if an experiment is using a targeting config that requires sticky but does not have sticky enabled and warns the user that their selected
Screen Shot 2022-07-11 at 1 35 02 PM
Screen Shot 2022-07-11 at 1 34 47 PM
 targeting expression requires sticky to function correctly (but does not prevent launching)
Sample screenshot 
fixes #7494

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1162,6 +1162,19 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
                 )
         return data
 
+    def _validate_sticky_enrollment(self, data):
+        targeting_config_slug = data.get("targeting_config_slug")
+        targeting_config = NimbusExperiment.TARGETING_CONFIGS[targeting_config_slug]
+        is_sticky = data.get("is_sticky")
+        sticky_required = targeting_config.sticky_required
+        if sticky_required and (not is_sticky):
+            self.warnings["is_sticky"].append(
+                "Selected targeting expression requires sticky enrollment to function\
+                    correctly"
+            )
+
+        return data
+
     def validate(self, data):
         application = data.get("application")
         channel = data.get("channel")
@@ -1173,6 +1186,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         data = self._validate_feature_config(data)
         data = self._validate_feature_configs(data)
         data = self._validate_versions(data)
+        data = self._validate_sticky_enrollment(data)
         if application != NimbusExperiment.Application.DESKTOP:
             data = self._validate_languages_versions(data)
             data = self._validate_countries_versions(data)
--- a/app/experimenter/experiments/tests/api/v5/test_queries.py
+++ b/app/experimenter/experiments/tests/api/v5/test_queries.py
@@ -440,6 +440,7 @@ class TestNimbusExperimentBySlugQuery(GraphQLTestCase):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.CREATED,
             application=NimbusExperiment.Application.DESKTOP,
+            targeting_config_slug=NimbusExperiment.TargetingConfig.NO_TARGETING,
             feature_configs=[
                 NimbusFeatureConfigFactory(
                     application=NimbusExperiment.Application.DESKTOP
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -541,6 +541,52 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
 
         self.assertIn("countries", serializer.errors)
 
+    @parameterized.expand(
+        [
+            (
+                NimbusExperiment.TargetingConfig.MAC_ONLY,
+                True,
+                0,
+            ),
+            (
+                NimbusExperiment.TargetingConfig.MAC_ONLY,
+                False,
+                0,
+            ),
+            (
+                NimbusExperiment.TargetingConfig.MOBILE_NEW_USERS,
+                False,
+                1,
+            ),
+            (
+                NimbusExperiment.TargetingConfig.MOBILE_NEW_USERS,
+                True,
+                0,
+            ),
+        ]
+    )
+    def test_experiments_with_is_sticky_warning(
+        self, targeting_config, is_sticky, warnings
+    ):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            targeting_config_slug=targeting_config,
+            channel=NimbusExperiment.Channel.RELEASE,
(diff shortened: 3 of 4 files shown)


jaredlockhart (changes requested):
Thank you @yashika-khurana !  This all looks good, but we probably need to do something more.  As it stands now, the user can completely miss this warning, there's no mention of it on the review page.  Let's try changing this from a warning to a review error for now, and see how people react to that.  


Screen Shot 2022-07-12 at 11 25 25 AM

yashika-khurana:
> @yashika-khurana Actually no that's kind of a silly workflow. Let's just make it so if an advanced targeting is selected where sticky is required, just force check the box, disable it, and add a note saying 'Sticky enrollment is required for this targeting configuration.'

@jaredlockhart  sure I am making the changes according to above mention comment, I agree that would be good. But I am wondering if there can be a scenario if people want to run advance targeting where sticky is recommended but still they don't want experiments to be sticky?? 🤔

yashika-khurana:
So I have done the following changes:
- Added the parameter in the api- `stickyRequired ` when passing` targeting config object`
- So if the sticky required is true for the selected targeting, sticky enrollment will get auto-selected with the note (see screenshot below)
image
- If the sticky is not required for the selected targeting, it will uncheck the sticky enrollment, but still user can select sticky enrollment if they want too (see two screenshot below)
- 
Screen Shot 2022-07-13 at 8 21 20 PM
Screen Shot 2022-07-13 at 8 21 14 PM

yashika-khurana:
@jaredlockhart, please see my above comment for the description of the change 🙏 :) Also, I have not removed the check from the review serializer. I thought it good to have validation on the backend side too. Please let me know your opinion on this :)

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1162,6 +1162,19 @@ def _validate_countries_versions(self, data):
                 )
         return data
 
+    def _validate_sticky_enrollment(self, data):
+        targeting_config_slug = data.get("targeting_config_slug")
+        targeting_config = NimbusExperiment.TARGETING_CONFIGS[targeting_config_slug]
+        is_sticky = data.get("is_sticky")
+        sticky_required = targeting_config.sticky_required
+        if sticky_required and (not is_sticky):
+            self.warnings["is_sticky"].append(
```
jaredlockhart, line 1171: We should make this an error instead of a warning.

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -110,6 +113,152 @@ describe("FormAudience", () => {
...
+    // selecting targeting which requires stick enrollment
+    fireEvent.change(screen.getByTestId("targetingConfigSlug"), {
+      target: { value: MOCK_CONFIG!.targetingConfigs![1]!.value },
+    });
+    expect(screen.getByTestId("isSticky")).toHaveProperty("checked", true);
+    expect(screen.getByTestId("isSticky")).toBeDisabled();
+    await expect(
+      screen.getByTestId("stickyRequiredWarning"),
+    ).toBeInTheDocument();
+  });
```
jaredlockhart, line 260: These are all the right cases but I would split each of these into separate tests rather than putting them all together in one.  Testing too many things in one test can make it hard to know which specific thing is broken when the test fails since then you have to go digging around at which line is failing its assertion rather than just looking at the test name itself.

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -148,6 +163,18 @@ export const FormAudience = ({
...
+  const TargetingOnChange = (ev: React.ChangeEvent<HTMLInputElement>) => {
+    const checkStickyRequired = targetingConfigSlugOptions.find(
+      (config) => config.value === ev.target.value,
+    );
+    setIsSticky(checkStickyRequired?.stickyRequired || false);
+    if (checkStickyRequired?.stickyRequired === true) {
+      setStickyRequiredWarning(true);
+    } else {
+      setStickyRequiredWarning(false);
+    }
```
jaredlockhart, line 175:
```js
      setStickyRequiredWarning(checkStickyRequired?.stickyRequired);
```

😉

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -256,8 +284,16 @@ export const FormAudience = ({
...
               {...formControlAttrs("isSticky")}
               type="checkbox"
+              checked={isSticky}
+              onChange={(e) => setIsSticky(e.target.checked)}
+              disabled={stickyRequiredWarning}
               label="Sticky Enrollment (Clients remain enrolled until the experiment ends)"
             />
+            {stickyRequiredWarning && (
+              <FormLabel data-testid="stickyRequiredWarning">
+                Sticky enrollment is required for this targeting configuration.
```
jaredlockhart, line 294: Let's use `.text-warning` for this.

jaredlockhart (changes requested):
Looking really good @yashika-khurana ! Just a couple small little changes then this should be ready to go 🙏

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
jaredlockhart: Sorry actually no, not text warning, a [warning alert](https://react-bootstrap.github.io/components/alerts/#examples).

yashika-khurana:
@jaredlockhart  sorry about that, I missed the file to commit. 🙏

PR #7513, opened 2022-07-12
fix #7510 bug(nimbus): update removed targeting config slug

### LINKED ISSUE #7510: Targeting constants were changed without migrating existing references
The following experiments show a broken detail page:

https://experimenter.services.mozilla.com/nimbus/firefox-suggest-history-rollout-expansion
https://experimenter.services.mozilla.com/nimbus/firefox-suggest-history-rollout
https://experimenter.services.mozilla.com/nimbus/firefox-suggest-a-ux-treatment-of-history-suggestions-from-firefox
https://experimenter.services.mozilla.com/nimbus/firefox-suggest-v1

They all refer to a targeting constant `urlbar_firefox_suggest_us_en` that was removed here:

https://github.com/mozilla/experimenter/commit/88ec57f5b958fbe3a96449187132194de692c718

No data migration was attached to update existing references to that slug.  This is a good reminder that we should always include a data migration when we move constants like this around.

So all we need to do is write a data migration that updates slugs that were changed/removed in that commit to their replacements.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2601)

jaredlockhart:
Because

* We removed a targeting config slug ("urlbar_firefox_suggest_us_en") in a change a while ago but neglected to write a data migration to update references to it
* Experiments with that removed targeting config slug don't render at all

This commit

* Adds a data migration to update references to that targeting config slug


yashika-khurana (approved):
LGTM! 🎉 🎉

PR #7521, opened 2022-07-14
fix #7518 feat(nimbus): Store external config files in experimenter repo.

### LINKED ISSUE #7518: Store external config files in repo
We've recently seen cases where changes to external config files (like feature manifests and jetstream configs) can introduce bugs/crashes in Experimenter, so rather than loading them in at container build time we should store them right in the repo and then use a PR/CI workflow to pull changes to them in so possible regressions can be caught by tests in CI.  

So for this part we should

1. First remove [these .gitignore rules](https://github.com/mozilla/experimenter/blob/main/.gitignore#L101-L105) so that the jetstream config and manifest files are tracked by git
2. Runs make jetstream_config feature_manifests to fetch the latest versions of the config files

and commit that.  Then we can add a task for updating them in https://github.com/mozilla/experimenter/issues/7496

jrbenny35:
Because

* In an effort to cause less breakages due to external file changes, we are implementing a system that will allow us to test any changes done to Jetstream config files as well as feature manifests files, before they're added to nimbus.

This commit

* Adds those external files to the repo for tracking their changes.


jaredlockhart (changes requested):
Awesome thank you @jrbenny35 .  Two things

1) We should remove anything other than the `outcomes` folder in the jetstream config otherwise random other changes in files will trigger a pr when we only care about `outcomes`.  Maybe just like find | grep <exclude outcomes> | rm type of thing right [here](https://github.com/mozilla/experimenter/blob/main/Makefile#L71)

2) Also gotta remove `fetch_external_resources` from the build commands in [the makefile](https://github.com/mozilla/experimenter/blob/main/Makefile#L85-L94)

jrbenny35:
Tried to be sneaky but can't ```. It is not possible to re-include a file if a parent directory of that file is excluded.```

jaredlockhart (changes requested):
Perfect looks exactly right, the tree view in the PR UI really helps with this 🙏 

*Oh no nvm me we do need to keep that line, nvm me

jaredlockhart (approved):
Perfect yes, this is correct, g/j 💯 💯 💯 💯

PR #7539, opened 2022-07-26
fix #7496 feat(nimbus): load external configs hourly

### LINKED ISSUE #7496: Move loading external configuration files into an automated PR step
Presently, we load external configuration files for

1) Feature manifests from external client repos
2) Jetstream configs from jetstream-config repo

at container build time.  However, errors in these files can result in any of the following

1) PRs breaking in CI before merging regardless of whether the PR code is healthy
2) Deploys breaking after merging regardless of whether the main code is healthy
3) Production servers breaking at runtime regardless of whether the main code is healthy

This means any stage of our workflow can break because of an external error that is not fixable by any process in Experimenter or its contributors.

To prevent this from happening in the future, we should move these external configuration fetching steps into a different workflow like so:

0) Local healthy copies of all external configuration files live in the Experimenter repo
1) A change to a configuration file lands in an external repo
2) The change is automatically detected and a PR is automatically created against the copies stored in the Experimenter repo
3) This PR step will catch any errors introduced in changes to those configuration files that would break some stage of the Experimenter test/deploy/live cycle
4) This PR can be automerged if it passes CI

This way we can separate code changes and configuration changes so that a breakage in one does not prevent deploying the other.

As part of this change, we should move the step that parses the desktop feature manifest files for externally linked resource schemas and fetches them out of the startup task that loads feature configurations at runtime, and into the container build stage so that we reduce the risk that breaking changes are introduced at runtime.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2596)

jaredlockhart:
Because

* We want to update the external configs for jetstream and features automatically
* We want updates to those files to be tested as part of CI

This commit

* Adds a new circle task to check for updates in external configs hourly
* If changes are detected, automatically create a PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
+  gh: circleci/github-cli@2.0
 jobs:
   check:
     machine:
@@ -46,7 +48,7 @@ jobs:
           command: |
             cp .env.integration-tests .env
             make refresh SKIP_DUMMY=1 up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
-  
+
   integration_nimbus_desktop_beta:
     machine:
       image: ubuntu-2004:202107-02 # Ubuntu 20.04, Docker v20.10.7, Docker Compose v1.29.2
@@ -73,6 +75,7 @@ jobs:
           command: |
             cp .env.integration-tests .env
             make refresh SKIP_DUMMY=1 up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
+
   integration_nimbus_desktop:
     machine:
       image: ubuntu-2004:202107-02 # Ubuntu 20.04, Docker v20.10.7, Docker Compose v1.29.2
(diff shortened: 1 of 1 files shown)


yashika-khurana (commented):
Looks good to me 🚀 , just I have two questions, as we are doing `git checkout -B external-config`, won't this be an issue the next time circle ci creates the branch- it will say branch already exist or something something. 2nd- Can we increase the cron job schedule to 1 hour to may be couple of hours (I am missing how often does it change)?

jaredlockhart:
> Looks good to me 🚀 , just I have two questions, as we are doing `git checkout -B external-config`, won't this be an issue the next time circle ci creates the branch- it will say branch already exist or something something. 2nd- Can we increase the cron job schedule to 1 hour to may be couple of hours (I am missing how often does it change)?

- If you use `-b` and the branch already exists it'll fail but if you use `-B` it'll force overwrite it
- If there's no changes it'll just pass quietly, it's not exactly about how often it changes but how quickly after a change lands we want it to show up in Experimenter, even this past week someone landed a new feature and wanted it basically immediately.  I don't think there's any harm to checking every hour since if there's no changes it'll just pass quietly.

yashika-khurana (approved):
perfect, I missed -B completely. looks good 🚀

PR #7541, opened 2022-07-27
fix #7536 feat(nimbus): Clone locales only if the application is desktop

### LINKED ISSUE #7536: Don't clone locales for mobile experiments where languages are supported
If someone clones a mobile experiment that has `locales` defined from before the earliest version that supported `languages`, then the `locales` will appear as part of the targeting expression when they are not supported by targeting on mobile clients.  We should only include `locales` when cloning a desktop experiment.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2623)

yashika-khurana:
Because

* If someone clones a mobile experiment that has locales defined from before the earliest version that supported languages, then the locales will appear as part of the targeting expression when they are not supported by targeting on mobile clients. We should only include locales when cloning a desktop experiment.

This commit

* Adds a check to clone locale if the application is desktop only

fix #7536


jaredlockhart (changes requested):
Thank you @yashika-khurana , can you please also add a test 🙏

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_clone_serializer.py
```diff
@@ -77,3 +79,43 @@ def test_invalid_rollout_branch_slug(self):
...
+                "parent_slug": parent.slug,
+                "name": "New Name",
+            },
+            context={"user": parent.owner},
+        )
+        self.assertTrue(serializer.is_valid())
+        child = serializer.save()
+        self.assertEqual(list(child.locales.all()), [locale])
+
+    def test_locales_not_clone_for_mobile_application(self):
```
jaredlockhart, line 103: Yep, could've used `parameterize` to test all the mobile applications here, but I don't think it's super critical, just something to remember whenever want something to be true for all mobile applications.

jaredlockhart (approved):
Thanks for adding tests @yashika-khurana ! Looks great 🎉🎉🎉🎉

PR #7549, opened 2022-07-27
fix #7548 feat(nimbus): add is_first_run to NimbusExperiment

### LINKED ISSUE #7548: Add first_run boolean field to NimbusExperiment
Let’s add a is_first_run boolean field to NimbusExperiment, and then add it to the `list_filter` on the V6 API.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2626)

jaredlockhart:
Because

* To support first run mobile experiments, we will fetch experiment recipes from the V6 API and include them in the built bundles that are deployed
* We need a way to filter experiments in the V6 API by whether they are first run or not so we only include the first run experiments in the bundle

This commit

* Adds a is_first_run bool field to NimbusExperiment
* Adds a query parameter to the V6 list api to filter by is_first_run


yashika-khurana (approved):
LGMT! 🚀 🌷  👍

PR #7554, opened 2022-07-29
fixes #6697 feat(nimbus): Review validation errors for rollouts

### LINKED ISSUE #6697: Add ready-for-review validation to ensure is_rollout is supported on application version selected for experiment
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1895)

eliserichards:
For #6697

Because:

* not all clients supported by Nimbus support `isRollout`

This commit:

* adds a review validator error if the selected combination of application and min version does not support rollouts

* adds a list to NimbusConstants describing clients by application and min version combinations that support rollouts

* implements review validation of the experiment against the list of supported clients

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -7,7 +7,6 @@ import jsonschema
 from django.contrib.auth import get_user_model
 from django.db import transaction
 from django.utils.text import slugify
-from packaging import version
 from rest_framework import serializers
 from rest_framework_dataclasses.serializers import DataclassSerializer
 
@@ -472,7 +471,6 @@ class NimbusExperimentBranchMixin:
         data = super().validate(data)
         data = self._validate_duplicate_branch_names(data)
         data = self._validate_single_branch_for_rollout(data)
-        data = self._validate_rollout_version_support(data)
         return data
 
     def _validate_duplicate_branch_names(self, data):
@@ -512,20 +510,6 @@ class NimbusExperimentBranchMixin:
                     ],
                 }
             )
-        return data
-
-    def _validate_rollout_version_support(self, data):
-        if not self.instance or not self.instance.is_rollout:
-            return data
-
-        for support in NimbusExperiment.ROLLOUT_SUPPORT:
-            if version.parse(self.instance.firefox_min_version) < version.parse(
-                support.firefox_min_version
-            ):
-                raise serializers.ValidationError(
-                    {"is_rollout": NimbusConstants.ERROR_BRANCH_NO_VALUE}
-                )
-
         return data
 
     def update(self, experiment, data):
@@ -1392,6 +1376,20 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
 
         return data
 
+    def _validate_rollout_version_support(self, data):
+        if not self.instance or not self.instance.is_rollout:
+            return data
+
+        for support in NimbusExperiment.ROLLOUT_SUPPORT:
+            if NimbusExperiment.Version.parse(
+                self.instance.firefox_min_version
+            ) < NimbusExperiment.Version.parse(support.firefox_min_version):
+                raise serializers.ValidationError(
+                    {"is_rollout": NimbusConstants.ERROR_ROLLOUT_VERSION_SUPPORT}
+                )
+
+        return data
+
     def validate(self, data):
         application = data.get("application")
         channel = data.get("channel")
@@ -1404,6 +1402,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         data = self._validate_feature_configs(data)
         data = self._validate_versions(data)
         data = self._validate_sticky_enrollment(data)
+        data = self._validate_rollout_version_support(data)
         if application != NimbusExperiment.Application.DESKTOP:
             data = self._validate_languages_versions(data)
             data = self._validate_countries_versions(data)
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -7,6 +7,7 @@ import jsonschema
 from django.contrib.auth import get_user_model
 from django.db import transaction
 from django.utils.text import slugify
+from packaging import version
 from rest_framework import serializers
 from rest_framework_dataclasses.serializers import DataclassSerializer
 
@@ -471,6 +472,7 @@ class NimbusExperimentBranchMixin:
         data = super().validate(data)
         data = self._validate_duplicate_branch_names(data)
         data = self._validate_single_branch_for_rollout(data)
+        data = self._validate_rollout_version_support(data)
         return data
 
     def _validate_duplicate_branch_names(self, data):
@@ -510,6 +512,20 @@ class NimbusExperimentBranchMixin:
                     ],
                 }
             )
+        return data
+
+    def _validate_rollout_version_support(self, data):
+        if not self.instance or not self.instance.is_rollout:
+            return data
+
+        for support in NimbusExperiment.ROLLOUT_SUPPORT:
+            if version.parse(self.instance.firefox_min_version) < version.parse(
+                support.firefox_min_version
+            ):
+                raise serializers.ValidationError(
+                    {"is_rollout": NimbusConstants.ERROR_BRANCH_NO_VALUE}
+                )
+
         return data
 
     def update(self, experiment, data):
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1384,7 +1384,10 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         rollout_version_supported = NimbusExperiment.ROLLOUT_SUPPORT_VERSION.get(
             self.instance.application
         )
-        if min_version < NimbusExperiment.Version.parse(rollout_version_supported):
+        if (
+            rollout_version_supported is not None
+            and min_version < NimbusExperiment.Version.parse(rollout_version_supported)
+        ):
             raise serializers.ValidationError(
                 {"is_rollout": NimbusConstants.ERROR_ROLLOUT_VERSION_SUPPORT}
             )
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1380,13 +1380,14 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         if not self.instance or not self.instance.is_rollout:
             return data
 
-        for support in NimbusExperiment.ROLLOUT_SUPPORT:
-            if NimbusExperiment.Version.parse(
-                self.instance.firefox_min_version
-            ) < NimbusExperiment.Version.parse(support.firefox_min_version):
-                raise serializers.ValidationError(
-                    {"is_rollout": NimbusConstants.ERROR_ROLLOUT_VERSION_SUPPORT}
-                )
+        min_version = NimbusExperiment.Version.parse(self.instance.firefox_min_version)
+        rollout_version_supported = NimbusExperiment.ROLLOUT_SUPPORT_VERSION.get(
+            self.instance.application
+        )
+        if min_version < NimbusExperiment.Version.parse(rollout_version_supported):
+            raise serializers.ValidationError(
+                {"is_rollout": NimbusConstants.ERROR_ROLLOUT_VERSION_SUPPORT}
+            )
 
         return data
 
--- a/app/experimenter/experiments/constants.py
+++ b/app/experimenter/experiments/constants.py
@@ -147,12 +147,6 @@ class Application(models.TextChoices):
     )
 
 
-@dataclass
-class RolloutSupport:
-    application: object
-    firefox_min_version: object
-
-
 class NimbusConstants(object):
     class Status(models.TextChoices):
         DRAFT = "Draft"
@@ -372,28 +366,13 @@ class NimbusConstants(object):
         Application.FOCUS_IOS: Version.FIREFOX_101,
     }
 
-    ROLLOUT_SUPPORT = (
-        RolloutSupport(
-            application=Application.DESKTOP,
-            firefox_min_version=Version.FIREFOX_105,
-        ),
-        RolloutSupport(
-            application=Application.FENIX,
-            firefox_min_version=Version.FIREFOX_105,
-        ),
-        RolloutSupport(
-            application=Application.IOS,
-            firefox_min_version=Version.FIREFOX_105,
-        ),
-        RolloutSupport(
-            application=Application.FOCUS_ANDROID,
-            firefox_min_version=Version.FIREFOX_105,
-        ),
-        RolloutSupport(
-            application=Application.FOCUS_IOS,
-            firefox_min_version=Version.FIREFOX_105,
-        ),
-    )
+    ROLLOUT_SUPPORT_VERSION = {
+        Application.DESKTOP: Version.FIREFOX_105,
+        Application.FENIX: Version.FIREFOX_105,
+        Application.FOCUS_ANDROID: Version.FIREFOX_105,
+        Application.IOS: Version.FIREFOX_105,
+        Application.FOCUS_IOS: Version.FIREFOX_105,
+    }
 
     # Telemetry systems including Firefox Desktop Telemetry v4 and Glean
     # have limits on the length of their unique identifiers, we should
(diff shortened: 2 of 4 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1376,20 @@ def _validate_sticky_enrollment(self, data):
...
         return data
 
+    def _validate_rollout_version_support(self, data):
+        if not self.instance or not self.instance.is_rollout:
+            return data
+
+        for support in NimbusExperiment.ROLLOUT_SUPPORT:
+            if NimbusExperiment.Version.parse(
+                self.instance.firefox_min_version
+            ) < NimbusExperiment.Version.parse(support.firefox_min_version):
```
jaredlockhart, line 1386: Wait this is a bit wonky, it's not matching by application?  This is matching all supported version for all applications against the experiment.  I think what you want here is the constant being a dict that maps application to minimum supported version and then look up the version from the experiments application?

app/experimenter/experiments/constants.py
```diff
@@ -365,6 +371,29 @@ class EmailType(models.TextChoices):
         Application.FOCUS_IOS: Version.FIREFOX_101,
     }
 
+    ROLLOUT_SUPPORT = (
```
jaredlockhart, line 374: I think you want this to be a dictionary that maps application to version?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -1130,6 +1130,74 @@ def test_serializer_feature_config_validation_treatment_value_no_schema(self):
         )
         self.assertTrue(serializer.is_valid())
 
+    def test_rollout_valid_version_support(self):
```
jaredlockhart, line 1133: So you could parameterize these tests to test every application/version pair like we do [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py#L239-L259) for both the valid and invalid cases, that way they're all covered.

jaredlockhart (changes requested):
Ya good idea breaking this out into a separate PR 👍

app/experimenter/experiments/constants.py
eliserichards: yes I do!

app/experimenter/experiments/api/v5/serializers.py
eliserichards: ahaha yeah that's totally my bad. 👍

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
eliserichards: I wanted to ask you about that. I have a few questions (e.g. do we want to include klar in this list) so I think I'd like to land this in the meantime and chat about it the version #s

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1376,21 @@ def _validate_sticky_enrollment(self, data):
 
         return data
 
+    def _validate_rollout_version_support(self, data):
+        if not self.instance or not self.instance.is_rollout:
```
jaredlockhart, line 1380: Ah good ol [DeMorgans](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1376,21 @@ def _validate_sticky_enrollment(self, data):
...
+            return data
+
+        min_version = NimbusExperiment.Version.parse(self.instance.firefox_min_version)
+        rollout_version_supported = NimbusExperiment.ROLLOUT_SUPPORT_VERSION.get(
```
jaredlockhart, line 1384: So this is the tiniest nit that ever nitted but here comes my speech.

So in Python there's two ways to look up a key in a dict, either with `a_dict['key']` or `a_dict.get('key')`, but there's an important behaviour difference between the two.  If the key doesn't exist in the dict, then the square brackets will raise a `KeyError`, whereas `.get()` will return `None`.  

So often times people might use the two interchangeably, but this failure case difference will matter later when something breaks either intentionally or unintentionally, ie do you expect not to be able to find the key?  If you expect to be able to find the key, then you want to use square brackets and if the key isn't found then something is actually wrong and you want that to fail and throw to Sentry so you know.  If you expect to possibly not find the key, you do want to use `.get()` and get `None` back, and then explicitly check for it when you use the value.

So in this case since there's a finite number of keys (ie the values of the enum) and we've explicitly setup the dict with each of those enum values as keys, I'd expect that we always expect to find the key, in which case if the key isn't found we want to throw.  If we use `.get()` here and it returns `None`, then by not checking for `None` and just using the value on the following line throwing it to `parse` then that would fail but it would be masking the actual failure, that we failed a key lookup we expected to pass.

Anyways that's my long rant.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
```diff
@@ -264,35 +264,73 @@ def test_no_duplicate_branch_names(self):
...
+            is_sticky=True,
+            targeting_config_slug=NimbusExperiment.TargetingConfig.MAC_ONLY,
         )
         experiment.is_rollout = True
         experiment.save()
         for branch in experiment.treatment_branches:
             branch.delete()
 
         data = {
+            "application": f"{NimbusExperiment.Application.DESKTOP}",
```
jaredlockhart, line 285: Oh this is another one of *Jared's Patented Nits*.  If what you want is to convert something to a string, then `str(...)` is more idiomatic than wrapping it in a format string where you're not actually doing any formatting.

jaredlockhart: Alternatively, if what you want is the string value of the enum, you can say `NimbusExperiment.Application.DESKTOP.value`

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
```diff
@@ -264,35 +264,73 @@ def test_no_duplicate_branch_names(self):
...
-            serializer.errors,
-            {
-                "treatment_branches": [
-                    {"name": NimbusConstants.ERROR_SINGLE_BRANCH_FOR_ROLLOUT}
-                    for i in data["treatment_branches"]
-                ],
-            },
+        self.assertEqual(len(serializer.errors), 1)
+
+        error = serializer.errors["treatment_branches"][0].get("name")
```
jaredlockhart, line 302: See just reiterating what I was saying earlier, here `"name"` may or may not be in the dict, so we either get back the error or None.  We check it on the next line so we'll still see a test failure so in theory either works here, but it's also okay to just assume it's there for the test and let it throw as the test failure.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
```diff
@@ -264,35 +264,73 @@ def test_no_duplicate_branch_names(self):
...
+        )
+        experiment.save()
+        for branch in experiment.treatment_branches:
+            branch.delete()
+        data = {
+            "application": f"{NimbusExperiment.Application.DESKTOP}",
+            "is_sticky": "true",
+            "is_rollout": "true",
+            "targeting_config_slug": f"{NimbusExperiment.TargetingConfig.MAC_ONLY}",
+            "firefox_min_version": f"{NimbusExperiment.Version.FIREFOX_108}",
```
jaredlockhart, line 323: 😬

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_branch_mixin.py
```diff
@@ -264,35 +264,73 @@ def test_no_duplicate_branch_names(self):
...
+            "changelog_message": "test changelog message",
+            "channel": "",
+        }
+        serializer = NimbusExperimentSerializer(
+            experiment, data=data, partial=True, context={"user": self.user}
         )
 
+        self.assertTrue(serializer.is_valid())
+        self.assertEqual(len(serializer.errors), 0)
```
jaredlockhart, line 332: Don't really need to check errors here at all if `is_valid` is `True`

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
jaredlockhart: Ya totes, not a blocking thing, just worth pointing out.

jaredlockhart (changes requested):
Yep thnx for the dict fix, generally looking good but the `.get` thing would cause shady runtime error if left alone 🙏

app/experimenter/experiments/constants.py
```diff
@@ -390,6 +398,9 @@ class EmailType(models.TextChoices):
     # Serializer validation errors
     ERROR_DUPLICATE_BRANCH_NAME = "Branch names must be unique."
     ERROR_SINGLE_BRANCH_FOR_ROLLOUT = "A rollout may have only a single reference branch"
+    ERROR_ROLLOUT_VERSION_SUPPORT = (
+        "Rollouts are not supported for the given application and version number."
```
yashika-khurana, line 402: This is just an improvement suggestion that QA pointed me out in the earlier PR to mention the version no. in the errors messages like rollouts are supported for this and this version for this particular application. So that when we use this error message in the frontend we can help the users which versions it clearly supports 💪

yashika-khurana (commented):
Just one minor suggestion

app/experimenter/experiments/api/v5/serializers.py
eliserichards: ooooooo that's so good to know! ty!!

app/experimenter/experiments/constants.py
eliserichards: that's a great suggestion! I'll bring this up with jared when we dig into [which applications and versions](https://github.com/mozilla/experimenter/pull/7554#discussion_r937178467) we want to support

yashika-khurana: perfect 🎉 🎉

jaredlockhart: Great suggestion @yashika-khurana 🙏

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
eliserichards: https://github.com/mozilla/experimenter/issues/7584

app/experimenter/experiments/api/v5/serializers.py
eliserichards: heh heh heh

eliserichards:
hrmmm integration test errors: 

```
*** You are running in headless mode.
Mozilla Firefox 104.0b6
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/s/six/python3-six_1.14.0-2_all.deb  Could not connect to archive.ubuntu.com:80 (185.125.190.36), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.91.38), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.91.39), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.39), connection timed out
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/j/javascript-common/javascript-common_11_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/j/jquery/libjs-jquery_3.3.1~dfsg-3_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/s/sphinx/libjs-sphinxdoc_1.8.5-7ubuntu3_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/p/python-pip/python-pip-whl_20.0.2-5ubuntu1.6_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/a/appdirs/python3-appdirs_1.4.3-2.1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/d/distlib/python3-distlib_0.3.0-1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/p/python-filelock/python3-filelock_3.0.12-2_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/m/more-itertools/python3-more-itertools_4.2.0-1build1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python-zipp/python3-zipp_1.0.0-1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python-importlib-metadata/python3-importlib-metadata_1.5.0-1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/pyparsing/python3-pyparsing_2.4.6-1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python-packaging/python3-packaging_20.3-1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/p/python-pluggy/python3-pluggy_0.13.0-2_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/s/setuptools/python3-setuptools_45.2.0-1_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/p/python-toml/python3-toml_0.10.0-3_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/t/tox/tox_3.13.2-2_all.deb  Unable to connect to archive.ubuntu.com:http:
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
sh: 1: tox: not found
ERROR: 127
make: *** [Makefile:192: integration_test_nimbus] Error 127

Exited with code exit status 2

```

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1376,24 @@ def _validate_sticky_enrollment(self, data):
...
+            self.instance.application
+        )
+        if (
+            rollout_version_supported is not None
```
jaredlockhart, line 1388: Forcing you to check for `None` with `Optional` return values is the only useful thing I've seen mypy do, but it doens't work well with Django so we can't use it 😖

jaredlockhart: And yeah so this will pass quietly if we didn't specify a minimum version for some given application, which may be what we want, may not.

app/experimenter/experiments/constants.py
```diff
@@ -366,6 +366,14 @@ class EmailType(models.TextChoices):
         Application.FOCUS_IOS: Version.FIREFOX_101,
     }
 
+    ROLLOUT_SUPPORT_VERSION = {
+        Application.DESKTOP: Version.FIREFOX_105,
+        Application.FENIX: Version.FIREFOX_105,
+        Application.FOCUS_ANDROID: Version.FIREFOX_105,
+        Application.IOS: Version.FIREFOX_105,
+        Application.FOCUS_IOS: Version.FIREFOX_105,
```
jaredlockhart, line 374: Actually the other good thing mypy can do is force you to be exhaustive about enums (in some cases anyways).  That's also useful.

jaredlockhart (approved):
Okay good thank you @eliserichards this is looking good.  Yay rollouts! 🎉

jaredlockhart:
Oh and I disabled those broken integration tests until we can get them fixed so this should be mergeable.

PR #7562, opened 2022-08-02
fix #7559 chore(nimbus): fail silently in external config circle task

### LINKED ISSUE #7559: Change CircleCI external config update task to quietly end if there's no changes
Instead of exiting non zero for the case where there are no changes, we should quietly end the circle task so we can distinguish between the case where it ended early successfully or throws because of some error.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2648)

jaredlockhart:
Because

* The hourly circle task to check for changes to external config files will fail if there are no changes
* The failures are logged and notified as circle failures
* This can mask an actual failure

This commit
* Changes the task to exit 0 if no changes are detected

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,16 +165,16 @@ jobs:
             git checkout main
             git pull origin main
             make fetch_external_resources
-            if git diff --name-only
+            if (($(git status --porcelain | wc -c) > 0))
               then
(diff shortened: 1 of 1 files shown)


yashika-khurana (approved):
@jaredlockhart  this seems awesome 🚀

PR #7566, opened 2022-08-02
fix #7565 chore(nimbus): only suppress exit code for pr creation in external config task

### LINKED ISSUE #7565: Only suppress pr creation failure in external config task
We only need to suppress the error code on the pr create command rather than all of the git commands.  If we suppress all of them then something other than pr creation could fail and we'd never know.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2651)

jaredlockhart:
Because

* We only need to suppress the error code of the pr creation command in the external config circle task
* If we suppress the others they could fail for some unexpected reason and we'd never be notified

This commit

* Only suppresses the exit code of the pr creation command

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,12 +167,11 @@ jobs:
             make fetch_external_resources
             if (($(git status --porcelain | wc -c) > 0))
               then
-                git checkout -B external-config &&\
-                git add . &&\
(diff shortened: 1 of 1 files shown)


yashika-khurana (approved):
🚀

PR #7578, opened 2022-08-08
fix #7577 bug(nimbus): check for is_paused in changelog in computed enrollment days

### LINKED ISSUE #7577: 500 error in V6 API
Just noticed after #7574 we're getting a 500 from the V6 API, sentry says:

https://sentry.io/organizations/mozilla/issues/3487680627/?query=is%3Aunresolved

jaredlockhart:
Because

* To compute the enrollment days we must first find the changelog when is_paused became True
* The is_paused field may not be present in very old changelogs
* We must check for the keys presence before looking it up

This commit

* Checks that the is_paused key is present in the changelog data before accessing it

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -437,7 +437,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         paused_changelogs = [
             c
             for c in changes
-            if c.experiment_data["is_paused"]
+            if "is_paused" in c.experiment_data
+            and c.experiment_data["is_paused"]
             and c.new_status == NimbusExperiment.Status.LIVE
             and c.new_status_next is None
             and c.new_publish_status == NimbusExperiment.PublishStatus.IDLE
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/models.py
```diff
@@ -437,7 +437,8 @@ def computed_enrollment_days(self):
         paused_changelogs = [
             c
             for c in changes
-            if c.experiment_data["is_paused"]
+            if "is_paused" in c.experiment_data
+            and c.experiment_data["is_paused"]
```
jrbenny35, line 441: Aren't these redundant?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -932,6 +932,20 @@ def test_computed_enrollment_days_returns_fallback_while_pause_pending_approval(
             expected_days,
         )
 
+    def test_computed_enrollment_days_returns_duration_if_is_paused_missing(self):
```
eliserichards, line 935: ❤️

eliserichards (approved):

app/experimenter/experiments/models.py
jaredlockhart: Negative, the 500 error in prod is happening because that key may not always be present in which case it will throw a `KeyError` so you need to check that the key is present first, and because Python short circuits boolean operations (properly I might add, looking at you JEXL), it won't continue evaluating the following clauses is the first one fails.

PR #7580, opened 2022-08-08
fix #7579 bug(nimbus): check for None in changelog experiment data when computing enrollment days

### LINKED ISSUE #7579: 500 on V6 API... again
Ah the greatest bug of them all, `None`.

Here we check for the presence of a key in a dict, but when we added that field rather than setting a default value of `{}` we set it to be nullable, so we have to check that it's `not None` before looking up a key in it 😅  

https://sentry.io/organizations/mozilla/issues/3487873693/?query=is%3Aunresolved

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2671)

jaredlockhart:
Because

* We changed the computed_enrollment_days method to do filters in memory rather than in the database for performance reasons
* Not only must we check that the key 'is_paused' is in the experiment_data dict, we must also check that the experiment_data field is not None

This commit

* Adds a None check to the 'is_paused' key lookup in changelog.experiment_data

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -437,7 +437,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         paused_changelogs = [
             c
             for c in changes
-            if "is_paused" in c.experiment_data
+            if c.experiment_data is not None
+            and "is_paused" in c.experiment_data
             and c.experiment_data["is_paused"]
             and c.new_status == NimbusExperiment.Status.LIVE
             and c.new_status_next is None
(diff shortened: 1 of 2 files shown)


jrbenny35 (commented):
Is this still needed?

jaredlockhart:
> Is this still needed?

Yep, see here: https://sentry.io/organizations/mozilla/issues/3487873693/?query=is%3Aunresolved

app/experimenter/experiments/models.py
```diff
@@ -437,7 +437,8 @@ def computed_enrollment_days(self):
         paused_changelogs = [
             c
             for c in changes
-            if "is_paused" in c.experiment_data
+            if c.experiment_data is not None
+            and "is_paused" in c.experiment_data
             and c.experiment_data["is_paused"]
```
jrbenny35, line 442: Is this needed?

jrbenny35 (changes requested):

yashika-khurana (approved):
👏 👏

jrbenny35 (approved):

PR #7581, opened 2022-08-09
fix #7525 feat(nimbus): force feature value for mobile experiments

### LINKED ISSUE #7525: Hide feature enabled flag for mobile
The feature enabled flag was never implemented for mobile and having it exposed in the UI can create misleading experiment designs we should hide it entirely for mobile experiments until we later deprecate it entirely on desktop.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2612)

jaredlockhart:
Because

* The feature enabled flag was never implemented on mobile
* Allowing it in the UI creates the false impression that it has some meaning for mobile experiments
* We are deprecating it on desktop and eventually removing it altogether

This commit

* Removes the feature enabled flag for mobile experiments and forces them to specify a feature value for every branch, which may explicitly be an empty object

Screenshot 2022-08-09 at 11-59-22 asdf – Branches Experimenter

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -359,20 +359,6 @@ class NimbusBranchSerializer(serializers.ModelSerializer):
     def validate(self, data):
         data = super().validate(data)
 
-        if data.get("feature_enabled") and not data.get("feature_value"):
-            raise serializers.ValidationError(
-                {"feature_value": NimbusConstants.ERROR_BRANCH_NO_VALUE}
-            )
-
-        if data.get("feature_value") and not data.get("feature_enabled"):
-            raise serializers.ValidationError(
-                {
-                    "feature_value": (
-                        "feature_enabled must be specified to include a feature_value."
-                    )
-                }
-            )
-
         feature_values = data.get("feature_values")
 
         if feature_values is not None:
@@ -409,6 +395,9 @@ class NimbusBranchSerializer(serializers.ModelSerializer):
             )
 
         branch.feature_values.all().delete()
+
+        if branch.experiment.application != NimbusExperiment.Application.DESKTOP:
+            feature_enabled = True
 
         if feature_value is not None:
             NimbusBranchFeatureValue.objects.create(
@@ -1111,6 +1100,19 @@ class NimbusBranchReviewSerializer(NimbusBranchSerializer):
             except Exception as e:
                 raise serializers.ValidationError(f"Invalid JSON: {e.msg}")
         return value
+
+    def validate(self, data):
+        if data.get("feature_enabled") and not data.get("feature_value"):
+            raise serializers.ValidationError(
+                {"feature_value": NimbusConstants.ERROR_BRANCH_NO_VALUE}
+            )
+
+        if data.get("feature_value") and not data.get("feature_enabled"):
+            raise serializers.ValidationError(
+                {"feature_enabled": NimbusConstants.ERROR_BRANCH_NO_ENABLED}
+            )
+
+        return data
 
 
 class NimbusReviewSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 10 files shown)


yashika-khurana:
@jaredlockhart one of the job got failed :(

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1112,6 +1101,19 @@ def validate_feature_value(self, value):
...
+            raise serializers.ValidationError(
+                {"feature_value": NimbusConstants.ERROR_BRANCH_NO_VALUE}
+            )
+
+        if data.get("feature_value") and not data.get("feature_enabled"):
+            raise serializers.ValidationError(
+                {"feature_enabled": NimbusConstants.ERROR_BRANCH_NO_ENABLED}
+            )
+
+        return data
```
eliserichards, line 1115: making this it's own `validate` method 👏

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -350,7 +350,7 @@ def test_update_experiment_branches_with_feature_config(self):
         self.assertEqual(experiment.feature_configs.get(), feature)
         self.assertEqual(experiment.branches.count(), 2)
         self.assertEqual(experiment.reference_branch.name, reference_branch_data["name"])
-        self.assertEqual(experiment.reference_branch.feature_values.get().enabled, False)
+        self.assertEqual(experiment.reference_branch.feature_values.get().enabled, True)
```
eliserichards, line 353: Tiny little nit, feel free to ignore: should we add "mobile" to the test name here? i.e. is there any reason to test for feature configs in both the desktop and mobile cases?

eliserichards: Ahh, I see you added a mobile test in the other file. Can we add "desktop" to this one then?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.tsx
```diff
@@ -160,28 +163,39 @@ export const FormBranch = ({
...
-      </Form.Group>
+        </Form.Group>
+      )}
+
       <Form.Group
         className="p-1 mx-3 mt-2 mb-0"
         data-testid="feature-config-edit"
       >
-        {featureEnabled ? (
+        {featureEnabled || !isDesktop ? (
```
eliserichards, line 198: (I think I'm completely misunderstanding what this block does, but..) should this be `isDesktop` instead of `!isDesktop`?

eliserichards:
```
E       selenium.common.exceptions.TimeoutException: Message: Edit Branches Page: could not find treatment branch enabled
```

Re-running CI to see if it times out again..

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: Oh yeah, this one is mobile, I'll add it to the name 👍

jaredlockhart: And yeah now the feature configs behave differently between desktop and mobile so we need separate tests.

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.tsx
jaredlockhart: No no other way around, if it's not desktop we just always show the feature value input box.  In desktop check whether it's explicitly enabled.

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Well not just that but also, it was happening at save time, which it doesn't really need to do, but also now we actually can't do this validation at all at this time since we need to force enabled to True for mobile experiments, and we can either do that from the frontend or from the backend, and I tried both and backend was easier/better, so the frontend thinks feature enabled is false and sends that and a value over and it's the backend that forces enabled to True, so if we try to do this validation at save them it'll always fail for mobile experiments.  But there was no reason to do it at save time so I moved it to review time.

yashika-khurana (approved):
LGTM! 🚀 🚀 , great work @jaredlockhart 👏 👏

PR #7586, opened 2022-08-10
fix #7519 chore(project): Re-enable Legacy Integration and JS tests on CI

### LINKED ISSUE #7519: Re enable legacy integration tests
While we continue to maintain Normandy support, we should have the legacy integration tests continue to run to ensure we don't introduce any regressions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2608)

jrbenny35:
Because

* We are still supporting the use of legacy for the foreseeable future, so we should make sure that changes added to the code base don't affect legacy in a negative way.

This commit

* Re-enables legacy integration tests on CI.
* Re-enables legacy JS tests as part of the check job on CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
           name: Run tests and linting
           command: |
             cp .env.sample .env
-            make check
+            make check_legacy
 
   integration_nimbus_desktop_release:
     machine:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           name: Run tests and linting
           command: |
             cp .env.sample .env
-            make check
+            make check_legacy
```
jaredlockhart, line 23: Oh right....  Actually....  Instead of doing it this way, let's just take out `check` from the makefile and rename `check_legacy` to `check` so that way running `make check` locally is the same as in CI and we don't have to worry about remembering which one to use.

jaredlockhart (changes requested):

jaredlockhart (approved):
Awesome thnx @jrbenny35 !

PR #7590, opened 2022-08-11
fix #7466 feat(nimbus):Disable feature enabled for 104+ versions

### LINKED ISSUE #7466: Disable the "enabled" switch in Web UI when targeting Fx 104+
Firefox 104 is removing support for Feature.enabled, so recipes targeting it should not use the Feature.enabled flag.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2576)

yashika-khurana:
Because

* Firefox 104 is removing support for Feature.enabled, so recipes targeting it should not use the Feature.enabled flag.

This commit

*  Add a check to the review serializer that 'enabled' can't be false for desktop 104+ and test cases to verify new changes

fix #7466


yashika-khurana:
@jaredlockhart @eliserichards I am having issue showing the errors message on the frontend, though the serializer is returning the error correctly  (find the image below)
image
and it warns here as well 
image
but the error message doesn't come on the toggle/switch, I tried using Form.Errors but it gave the error. I am sure missing something 🙏 🙏

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1257,7 +1257,9 @@ def _validate_feature_config(self, data):
 
         schema = json.loads(feature_config.schema)
         error_result = {}
-        if data["reference_branch"].get("feature_enabled"):
```
jaredlockhart, line 1260: Wait what's the reason for this?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1380,51 @@ def _validate_sticky_enrollment(self, data):
...
         return data
 
+    def _validate_feature_enabled_version(self, data):
+        application = data.get("application")
+        min_version = data.get("firefox_min_version", "")
+
+        min_supported_version = NimbusConstants.FEATURE_ENABLED_MIN_UNSUPPORTED_VERSION[
+            application
+        ]
+        if NimbusExperiment.Version.parse(min_version) > NimbusExperiment.Version.parse(
```
jaredlockhart, line 1390: Should this be `>=`?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1380,51 @@ def _validate_sticky_enrollment(self, data):
...
+            min_supported_version
+        ):
+            return True
+        return False
+
+    def _validate_feature_enabled_for_treatment_branches(self, data):
+        if self._validate_feature_enabled_version(data) and "treatment_branches" in data:
+            treatment_branches_error = []
+            for i in data["treatment_branches"]:
+                if not i["feature_enabled"]:
```
jaredlockhart, line 1400: `i` isn't a very informative variable name here ;)

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1380,51 @@ def _validate_sticky_enrollment(self, data):
...
+        min_version = data.get("firefox_min_version", "")
+
+        min_supported_version = NimbusConstants.FEATURE_ENABLED_MIN_UNSUPPORTED_VERSION[
+            application
+        ]
+        if NimbusExperiment.Version.parse(min_version) > NimbusExperiment.Version.parse(
+            min_supported_version
+        ):
+            return True
+        return False
```
jaredlockhart, line 1394: These returns are redundant, you can just return the value of the comparison

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1376,6 +1380,51 @@ def _validate_sticky_enrollment(self, data):
...
+                    {"treatment_branches": treatment_branches_error}
+                )
+
+        return data
+
+    def _validate_feature_enabled_for_reference_branch(self, data):
+        if (
+            "reference_branch" in data
+            and not data["reference_branch"]["feature_enabled"]
+            and self._validate_feature_enabled_version(data)
```
jaredlockhart, line 1416: How come we check the version here for the reference branch but not above for the treatment branches? Shouldn't we do both?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1410,6 +1459,9 @@ def validate(self, data):
         if application != NimbusExperiment.Application.DESKTOP:
             data = self._validate_languages_versions(data)
             data = self._validate_countries_versions(data)
+        else:
+            data = self._validate_feature_enabled_for_treatment_branches(data)
+            data = self._validate_feature_enabled_for_reference_branch(data)
```
jaredlockhart, line 1464: 👍

app/experimenter/experiments/constants.py
```diff
@@ -365,6 +365,7 @@ class EmailType(models.TextChoices):
         Application.IOS: Version.FIREFOX_101,
         Application.FOCUS_IOS: Version.FIREFOX_101,
     }
+    FEATURE_ENABLED_MIN_UNSUPPORTED_VERSION = {Application.DESKTOP: Version.FIREFOX_103}
```
jaredlockhart, line 368: Oh i see you wrote it here as 103 and used `>` above.  The change was actually introduced in 104 wasn't it?  Looking at this I would think it was introduced in 103, it would probably be clearer to make it match here the version it was actually changed in.  

Also since we're only doing desktop here, a dictionary is probably unnecessary?

app/experimenter/experiments/constants.py
```diff
@@ -421,6 +422,8 @@ class EmailType(models.TextChoices):
     ERROR_FIREFOX_VERSION_MAX = (
         "Ensure this value is greater than or equal to the minimum version"
     )
+    ERROR_FEATURE_ENABLED = f"Feature enabled is only supported for\
+         {FEATURE_ENABLED_MIN_UNSUPPORTED_VERSION[Application.DESKTOP]} version and less"
```
jaredlockhart, line 426: Oh yeah, I see what you mean.  To match the language we're using for the mobile, we might say it something more like

```
Experiments for Firefox Desktop {version} and above must have the feature enabled and specify a feature value.
```

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -1326,6 +1326,158 @@ def test_serializer_feature_config_validation_bad_json_value(self):
...
+            feature_value.enabled = feature_enabled
+            feature_value.value = (
+                """\
+            {
+            "bar": {
+                "baz": "baz",
+                "qux": 123
+            }
+            }
+        """.strip()
```
jaredlockhart, line 1384: Indentation's a bit wonky here.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -1326,6 +1326,158 @@ def test_serializer_feature_config_validation_bad_json_value(self):
...
+            (
+                NimbusExperiment.Application.DESKTOP,
+                NimbusExperiment.Version.FIREFOX_105,
+                True,
+                True,
+            ),
+        ]
+    )
+    def test_serializer_feature_enable_version_support_reference_branch(
+        self, application, firefox_version, feature_enabled, valid
```
jaredlockhart, line 1358: Oh here I might call `valid` as `expected_valid`

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -1326,6 +1326,158 @@ def test_serializer_feature_config_validation_bad_json_value(self):
...
+                NimbusExperiment.Version.FIREFOX_101,
+                False,
+                True,
+            ),
+            (
+                NimbusExperiment.Application.DESKTOP,
+                NimbusExperiment.Version.FIREFOX_105,
+                True,
+                True,
+            ),
```
jaredlockhart, line 1354: Yeah 101 and 105 are on either side of the supported version, but it might be clearer if they're right on the line, ie 103 and 104.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -1326,6 +1326,158 @@ def test_serializer_feature_config_validation_bad_json_value(self):
...
+        ]
+    )
+    def test_serializer_feature_enable_version_support_treatment_branches(
```
jaredlockhart, line 1432: I feel like you could fold these two tests together since they use the same setup.  I know I'm always rambling about not testing too much in a single test, but I feel like these are testing essentially the same thing.  Not a hard requirement, just a suggestion.

jaredlockhart (dismissed):
Awesome thank you @yashika-khurana this is perfectly timed 🙏  Some tweaks to the logic but broadly correct 👍

app/experimenter/experiments/api/v5/serializers.py
yashika-khurana: yes modified according to the comments

app/experimenter/experiments/api/v5/serializers.py
yashika-khurana: we are verifying in the treatment branch as well 
image

app/experimenter/experiments/constants.py
yashika-khurana: makes sense completely :)

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
yashika-khurana: I completely agree with your point, initially I kept one but then I thought to separate out signifying two different functions in the serializer validating that depending on the type of branch

eliserichards (approved):
🚢 LGTM!

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Oh I missed it at the top! XD

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1272,7 +1273,8 @@ def _validate_feature_config(self, data):
         for branch_data in data["treatment_branches"]:
             branch_error = None
             branch_warning = None
-            if branch_data.get("feature_enabled", False):
+            feature_enabled = branch_data.get("feature_enabled", False)
+            if feature_enabled:
```
jaredlockhart, line 1277: I still don't know what this change is accomplishing?

jaredlockhart (changes requested):
Thanks for updating with feedback, but I don't see the error showed on the branch page?

Screenshot 2022-08-15 at 12-20-59 asdf – Branches Experimenter

app/experimenter/experiments/api/v5/serializers.py
yashika-khurana: sorry, forgot to get rid of that

yashika-khurana:
@jaredlockhart here is the error message
image

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.tsx
```diff
@@ -174,6 +174,14 @@ export const FormBranch = ({
               type="switch"
               label={featureEnabled ? "On" : "Off"}
             />
+
+            {reviewErrors !== undefined && reviewErrors.feature_enabled && (
```
jaredlockhart, line 178: Wait can't you write this as `reviewErrors?.feature_enabled`?

jaredlockhart (approved):
Okay good looks great works great thank you @yashika-khurana ! I think you might be able to simplify that `undefined` check but otherwise looks great 🎉 🎉 🎉 🎉

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormBranch.tsx
jaredlockhart: Yeah I think you should be able to:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

yashika-khurana: right right, just wrote code while talking to my brain logically, good point @jaredlockhart  thank you :)

PR #7594, opened 2022-08-15
fix #7572 chore(nimbus): prevent force pushing external config if pr exists

### LINKED ISSUE #7572: External config bot repushing unchanged PRs
Looks like the external config bot is repushing the same pr many times, not sure that's super disastrous but definitely unnecessary so we should try to prevent that if possible.

Screen Shot 2022-08-08 at 12 09 36 PM

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2667)

jaredlockhart:
Because

* When external configs change, we will automatically commit and create a pr
* If a pr already exists, we are force pushing changes on every run which forces CI to needlessly re run

This commit

* Checks if the new commit is different from the commit that exists before force pushing to prevent CI from re running

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,7 +117,7 @@ jobs:
           command: |
             cp .env.integration-tests .env
             make refresh SKIP_DUMMY=1 up_prod_detached integration_test_nimbus_rust PYTEST_ARGS="$PYTEST_ARGS"
-  
+
   integration_legacy:
     machine:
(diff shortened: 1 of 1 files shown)


jaredlockhart:
Ran a test run [here](https://app.circleci.com/pipelines/github/mozilla/experimenter/14046/workflows/7ef78477-25d1-46d7-a40a-9750a933f32d/jobs/42544) while https://github.com/mozilla/experimenter/pull/7593 is open

jeddai (approved):
Looks good! The test run hits the else and I verified the diff check locally.

Issue #6696, opened 2021-11-22
Add a checkbox to branches page that controls is_rollout=true


data-sync-user:
* adds a checkbox to the branches page to control isRollout
* adds isRollout to the form reducer for FormBranches
* adds isRollout to the GQL ExperimentInput type
* slightly refactors onFormSave in PageEditBranches to handle whatever
new properties might be added to the FormBranches reducer
* moves the single-branch enforcement to the ready-for-review serializer

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-1894)

data-sync-user:
➤ Les Orchard commented:

A few wrinkles that might need working through: If the checkbox is checked, should it immediately update the UI such that all but the first branch and the “+ Add Branch” button disappears - or should we wait until after save? Possible dataloss if the checkbox is checked then unchecked, but we’ve discard the treatment branches.

data-sync-user:
➤ Janet Dragojevic commented:

🚩 Flag added

blocked on nimbus-shared/issues/183

PR #7606, opened 2022-08-17
Clone rollouts

yashika-khurana:


yashika-khurana:
@jaredlockhart I added some work to Elise PR

jaredlockhart:
Oh @yashika-khurana I'm not sure this was captured properly somewhere, it's [here](https://github.com/mozilla/experimenter/issues/6696#issuecomment-975972066) as a comment, but when the rollout checkbox is checked it should hide all the branches but control and remove the 'add branch' button altogether.

yashika-khurana:
absolutely @jaredlockhart , the rollout is checked, add branch button is gone and telling user to get rid of this branch 
image

jaredlockhart:
@yashika-khurana No I mean there should never be more than the control branch if it's checked, rather than prompting the user to do it.

yashika-khurana:
oh so you mean we should remove that branch voluntarily?

jaredlockhart:
@yashika-khurana exactly, there's no reason to prompt the user to do it becuase there is no case where there would be more than 1 branch for a rollout.

yashika-khurana:
perfect, thanks @jaredlockhart I will make the necessary changes

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
```diff
@@ -248,6 +257,23 @@ export const FormBranches = ({
...
         </Form.Group>
 
+        <Form.Row>
+          <Form.Group as={Col} controlId="isRollout">
+            <Form.Check
+              data-testid="is-rollout-checkbox"
+              onChange={handleIsRollout}
+              checked={!!isRollout}
+              type="checkbox"
+              label="Handle this experiment as a rollout"
```
jaredlockhart, line 267: Let's write it as 

`This is a rollout (single branch)`

yashika-khurana:
@jrbenny35 I see one of my check is not passing, any insights would be great to have 🙏 , I think its related to some image

jaredlockhart:
okay do a re run, see if that works, lots of people are available to review, make sure:

- checking the rollout box hides the other branches from the ui AND removes their data from the API call so the branches get removed on the server side
- sets the is_rollout bool on the model

yashika-khurana:
@jaredlockhart yup it does that :)

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
```diff
@@ -248,6 +257,23 @@ export const FormBranches = ({
...
+              onChange={handleIsRollout}
+              checked={!!isRollout}
+              type="checkbox"
+              label="This is a rollout (single branch)"
+            />
+            {fieldMessages?.is_rollout && (
+              <Alert data-testid="sticky-required-warning" variant="danger">
+                {fieldMessages?.is_rollout}
+              </Alert>
+            )}
```
mikewilli, line 273: Would it be better to use the [Form.Check props](https://react-bootstrap.github.io/forms/checks-radios/#form-check-props) `isInvalid` and `feedback` instead of the Alert?

```typescript
isInvalid={!!(fieldMessages?.is_rollout)}
feedback={fieldMessages?.is_rollout}
```

mikewilli (approved):
Couple suggestions but nothing major. Overall looks great!

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
yashika-khurana: good point thanks @mikewilli I have updated as per suggestion :)

yashika-khurana:
@mikewilli do you wanna give second pass before I merge, I have updated as per your suggestions

mikewilli:
Looks good to me!

Issue #7613, opened 2022-08-19
Canceling the review request for ending enrollment modifies the Enrollment duration of the Timeline


data-sync-user:
**[Prerequisites]:**

* Have a Live experiment with enrollment still enabled.

**[Steps to reproduce]:**

1. Open the page of the experiment from prerequisites.
1. Click “End Enrollment for Experiment”.
1. Click “Yes, end enrollment for the experiment”.
1. Make a note of the “Enrollment” portion of the Timeline.
1. Click “Cancel Review Request”.
1. Observe the “Enrollment” portion of the Timeline.

**[Expected result]:**

* The “Enrollment” duration of the Timeline remains unchanged.

**[Actual result]:**

* The “Enrollment” duration of the Timeline is changed to one corresponding to when the request was canceled. 

**[Notes]:**

* The issue is not reproducible for “Total duration” (canceling End Experiment).
* The issue is not reproducible when rejecting the request from the experiment page or from Remote Settings.
* Attached is a recording of the issue.

!Enrollment duration.gif|width=1830,height=818!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-2716)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/27643">Enrollment duration.gif</a>

data-sync-user:
➤ Janet Dragojevic commented:

Yashika Khurana can you please take a look.

data-sync-user:
➤ Yashika Khurana commented:

Janet Dragojevic thanks Janet, sure I will take this

data-sync-user:
➤ Yashika Khurana commented:

Maria Heres can you please review it, I just pushed the fix. Thank you 🙂

data-sync-user:
➤ Maria Heres commented:

Yashika Khurana I can still reproduce this issue on Stage.

data-sync-user:
➤ Yashika Khurana commented:

thanks Maria Heres , I have added another patch, please let me know if you still see that issue.

data-sync-user:
➤ Maria Heres commented:

I have verified that this issue is no longer reproducible. Tested on Stage using Firefox 103.0.2 (Build ID: 20220808125904) - Windows 10.

PR #7630, opened 2022-08-23
7613/cancel review enrollment bug

yashika-khurana:
Because

    When the experiment is in a live state and we cancel the review request, it was changing the enrollment days (as it is a property and we compute this from the changelogs)

This commit

    fixes the issue regarding changing enrollment days and test cases to verify new changes

fix https://github.com/mozilla/experimenter/issues/7613


eliserichards (approved):
Thanks for the explanation! This looks great 💪

PR #7640, opened 2022-08-24
fix #7273 feat(nimbus) test(nimbus): Add new opMon link for rollout dashboards

### LINKED ISSUE #7273: Link to Looker OpMon rollout dashboards
The dashboards for Nimbus and Normandy rollouts that OpMon generates follow this URL pattern:

`https://mozilla.cloud.looker.com/dashboards/operational_monitoring::<normandy slug with dashes to underscores>`

For example:

[https://mozilla.cloud.looker.com/dashboards/operational_monitoring::bug_1716693_rollout_doh_canada_rollout_release_91_94](https://mozilla.cloud.looker.com/dashboards/operational_monitoring::bug_1716693_rollout_doh_canada_rollout_release_91_94)

We made some improvements for the new viz to render a little bit faster (it takes about 30s still for the dashboard to completely appear, but we are working on more improvements), there are also crash-related default metrics available (the default metrics can easily be changed). So I think it's time to add links to these dashboards. 

They will be available **the day after** a rollout has been launched.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2419)

eliserichards:
Because...

* Rollout dashboards live in a different place than the experiment looker dashboards

This commit...

* Adds the new rollout link format (with underscores instead of dashes in the slug)
* Adds the rollout dashboard to models
* Only shows the rollout dashboard after the rollout is Live
* Shows rollout dashboards for rollouts and experiment dashboards for experiments

image

#7273

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -522,6 +522,13 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         )
 
     @property
+    def rollout_monitoring_dashboard_url(self):
+        if self.start_date and ((datetime.date.today() - self.start_date).days >= 1):
+            return settings.ROLLOUT_MONITORING_URL.format(
+                slug=self.slug.replace("-", "_")
+            )
+
+    @property
     def review_url(self):
         return "{base_url}{collection_path}/{collection}/{review_path}".format(
             base_url=settings.KINTO_ADMIN_URL,
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -523,7 +523,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def rollout_monitoring_dashboard_url(self):
-        if self.start_date and ((datetime.date.today() - self.start_date).days >= 1):
+        if self.status == (NimbusExperiment.Status.LIVE or NimbusExperiment.Status.COMPLETE):
             return settings.ROLLOUT_MONITORING_URL.format(
                 slug=self.slug.replace("-", "_")
             )
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -525,7 +525,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
     def rollout_monitoring_dashboard_url(self):
         if self.is_rollout and (
             self.status
-            == (NimbusExperiment.Status.LIVE or NimbusExperiment.Status.COMPLETE)
+            in (NimbusExperiment.Status.LIVE, NimbusExperiment.Status.COMPLETE)
         ):
             return settings.ROLLOUT_MONITORING_URL.format(
                 slug=self.slug.replace("-", "_")
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -523,7 +523,10 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def rollout_monitoring_dashboard_url(self):
-        if self.status == (NimbusExperiment.Status.LIVE or NimbusExperiment.Status.COMPLETE):
+        if self.is_rollout and (
+            self.status
+            == (NimbusExperiment.Status.LIVE or NimbusExperiment.Status.COMPLETE)
+        ):
             return settings.ROLLOUT_MONITORING_URL.format(
                 slug=self.slug.replace("-", "_")
             )
--- a/app/experimenter/experiments/tests/test_models.py
+++ b/app/experimenter/experiments/tests/test_models.py
@@ -1116,16 +1116,18 @@ class TestNimbusExperiment(TestCase):
         )
 
     def test_rollouts_monitoring_dashboard_url_is_none_when_rollout_not_begun(self):
-        experiment = NimbusExperimentFactory.create(
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
             slug="rollout-1-slug",
             status=NimbusExperiment.Status.DRAFT,
         )
-        self.assertIsNone(experiment.rollouts_monitoring_dashboard_url)
+        self.assertIsNone(experiment.rollout_monitoring_dashboard_url)
 
     def test_rollouts_monitoring_dashboard_url_is_none_on_first_day(self):
-        experiment = NimbusExperimentFactory.create(
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
             slug="rollout-1-slug",
-            status=NimbusExperiment.Status.LIVE,
+            status=NimbusExperiment.Status.DRAFT,
         )
 
         NimbusChangeLogFactory.create(
@@ -1135,14 +1137,15 @@ class TestNimbusExperiment(TestCase):
             changed_on=datetime.date.today(),
         )
 
-        self.assertIsNone(experiment.rollouts_monitoring_dashboard_url)
+        self.assertIsNone(experiment.rollout_monitoring_dashboard_url)
 
     def test_rollouts_monitoring_dashboard_returns_url_one_day_after_start(self):
-        experiment = NimbusExperimentFactory.create(
-            slug="rollout-1-slug",
-            status=NimbusExperiment.Status.LIVE,
-        )
         yesterday = datetime.date.today() - datetime.timedelta(days=1)
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            slug="rollout-1-slug",
+            status=NimbusExperiment.Status.DRAFT,
+        )
         NimbusChangeLogFactory.create(
             experiment=experiment,
             old_status=NimbusExperiment.Status.DRAFT,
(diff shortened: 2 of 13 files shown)


eliserichards:
for testing:

```
make bash
---
./manage.py shell
---
   ...: for i in range(100):
   ...:     NimbusExperimentFactory.create_with_lifecycle(NimbusExperimentFactory.Lifecycl
   ...: es.LAUNCH_APPROVE_APPROVE, is_rollout=True)
```

app/experimenter/experiments/models.py
```diff
@@ -521,6 +521,13 @@ def monitoring_dashboard_url(self):
             to_date=end_date.strftime("%Y-%m-%d"),
         )
 
+    @property
+    def rollout_monitoring_dashboard_url(self):
+        if self.start_date and ((datetime.date.today() - self.start_date).days >= 1):
```
jaredlockhart, line 526: Why do we need the `days >= 1` check?  And if we're checking for the presence of a `start_date`, couldn't we just check that `self.status == self.STATUS.LIVE`?

app/experimenter/experiments/models.py
```diff
@@ -521,6 +521,13 @@ def monitoring_dashboard_url(self):
...
+            return settings.ROLLOUT_MONITORING_URL.format(
+                slug=self.slug.replace("-", "_")
```
jaredlockhart, line 528: Sigh 😅

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1115,6 +1115,106 @@ def test_monitoring_dashboard_url_returns_url_when_experiment_is_complete(self):
...
+            experiment=experiment,
+            old_status=NimbusExperiment.Status.DRAFT,
+            new_status=NimbusExperiment.Status.LIVE,
+            changed_on=yesterday,
+        )
+
+        expected_slug = experiment.slug.replace("-", "_")
+        url = experiment.rollout_monitoring_dashboard_url
+        actual_slug = url[(url.index("::") + 2) :]  # take a slice of the url after '::'
+        self.assertEqual(expected_slug, actual_slug)
```
jaredlockhart, line 1181: I would just hard code the `expected_slug` here instead of transforming it since we're explicitly setting the original slug so we know what the transformed version is going to look like.  And it's fine to just check the whole URL instead of the slug separately, this kind of string querying is finnicky, if you wanted to really do this you'd want to use a regex, but I don't think it's necessary.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1115,6 +1115,106 @@ def test_monitoring_dashboard_url_returns_url_when_experiment_is_complete(self):
...
+            slug="rollout-1-slug",
+            status=NimbusExperiment.Status.DRAFT,
+        )
+
+        NimbusChangeLogFactory.create(
+            experiment=experiment,
+            old_status=NimbusExperiment.Status.DRAFT,
+            new_status=NimbusExperiment.Status.LIVE,
+            changed_on=datetime.date.today(),
+        )
```
jaredlockhart, line 1138: Oh yeah, this isn't really the right way to do this.  Then you've got an experiment whose status is `DRAFT` but has a `DRAFT` to `LIVE` changelog, so the experiments in kind of a wonky state, which could produce other weird unintended behaviours.  What you'd wanna do is use the lifecycle that puts it into the end state you want, and then lookup the changelog for `DRAFT` to `LIVE` and update its date.

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -288,10 +288,21 @@ export const DirectoryLiveTable: React.FC<DirectoryTableProps> = (props) => (
                 Looker
               </LinkExternal>
             )}
+            {experiment.monitoringDashboardUrl &&
+              experiment.resultsReady &&
+              experiment.rolloutMonitoringDashboardUrl && <br />}
```
jaredlockhart, line 293: That's a bit specific, if we change those states around later this'll be awkward to keep in sync?  There's no simpler way to do that?

jaredlockhart (dismissed):
Looking good but I think you can skip checking dates altogether.  Thnx @eliserichards !

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
eliserichards: This is just taken from below, line 302, with the addition of `rolloutMonitoringDashboardUrl`. I'm open to suggestions 😄

app/experimenter/experiments/tests/test_models.py
eliserichards: ohhh I see what you mean 👍

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
jaredlockhart: Okay maybe let's just leave it and maybe @yashika-khurana can find a better way to do it when she tackles the sidebars.

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -288,10 +288,21 @@ export const DirectoryLiveTable: React.FC<DirectoryTableProps> = (props) => (
...
+            {experiment.rolloutMonitoringDashboardUrl && (
```
jaredlockhart, line 294: Oh I just realized, we have to guard this somewhere that it is `isRollout` right?  We can either do it here in the UI or we can do it above in the model method.  We had this same pattern all the time in legacy, do you check in the UI whether to show something or do it in the model method.  Doing it in the model method means you don't hagve to worry about it later when you request and use it in the UI in arbitrary places, so it's maybe a little more reusable?

eliserichards: I don't think we need to guard here. This is following the same pattern as the normal monitoring dashboard, as in it's either there or it's not. It might save us a line or two, but I don't think it's super necessary

jaredlockhart: No I mean, if we don't guard the model method or the UI then there will be an opmon link for experiments that goes nowhere?

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -288,10 +288,21 @@ export const DirectoryLiveTable: React.FC<DirectoryTableProps> = (props) => (
...
+                data-testid="link-rollout-monitoring-dashboard"
+              >
+                Rollout dashboard
+              </LinkExternal>
+            )}
             {experiment.monitoringDashboardUrl && experiment.resultsReady && (
               <br />
             )}
-            {experiment.resultsReady && (
+            {!experiment.isRollout && experiment.resultsReady && (
```
yashika-khurana, line 305: We don't want to show results when the experiment is rollout?

eliserichards: I don't think that rollouts have a Results page the way that experiments do (but now that you mention it... @jaredlockhart pls confirm)

jaredlockhart: Yes that's correct, there is no 'results' for rollouts

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
eliserichards: oh, duh. adding that

yashika-khurana (approved):
LGTM! 🚀

app/experimenter/experiments/models.py
```diff
@@ -505,21 +505,32 @@ def is_paused_published(self):
 
     @property
     def monitoring_dashboard_url(self):
-        start_date = (self.start_date or datetime.date.today()) - datetime.timedelta(
-            days=1
-        )
+        if not self.is_rollout:
```
jaredlockhart, line 508: Oh no no no, experiments and rollouts both have monitoring for enrollment.  It's only the opmon link that needs to be guarded.

app/experimenter/experiments/models.py
```diff
@@ -505,21 +505,32 @@ def is_paused_published(self):
...
+                from_date=start_date.strftime("%Y-%m-%d"),
+                to_date=end_date.strftime("%Y-%m-%d"),
+            )
+
+    @property
+    def rollout_monitoring_dashboard_url(self):
+        if self.is_rollout and (
+            self.status
+            == (NimbusExperiment.Status.LIVE or NimbusExperiment.Status.COMPLETE)
+        ):
```
jaredlockhart, line 530: Yes exactly 👍

jaredlockhart (changes requested):

app/experimenter/experiments/models.py
jaredlockhart: Oh wait what, does that work?  I've never seen that syntax before....  I would've written it as

`self.status in (LIVE, COMPLETE)`

jaredlockhart: Where did you get that syntax?  I don't think it works like that?

```py
ipdb> 1 == (1 or 2)
True
ipdb> 2 == (1 or 2)
False
```

jaredlockhart: Another thing to think about is parametrically testing both statuses would probably catch that this doesn't work as intended.

eliserichards: DANG IT. Python got me there.

PR #7650, opened 2022-08-29
fix #7522 feat(nimbus): First run experiment UI

### LINKED ISSUE #7522: Make first run targeting a selectable option
Now that we've pulled sticky out, I think the next major flag we should promote is 'first run'.  Many of the advanced targeting configs include first run in some way, and some are duplicates with and without first run.  First run also has a strong interaction with sticky, since first run targeting only works with sticky enabled.

Let's add another checkbox to the audience page for first run targeting, and follow the same process we did for sticky, ie

1) Add a bool to the model
2) Expose it through the V5 API into the UI
3) Pull the first run clause out of advanced targeting and into the model targeting method
4) Update existing advanced targeting configs to remove the first run clause where it's used and identify all the advanced targeting configs that included it
5) Migrate all experiments that use any of the targeting configs that included first run to have first run set to true in the model
6) Change the UI/serializers to force sticky if first run is enabled rather than based on a flag on the particular targeting config (and remove the sticky required field altogether)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2610)

yashika-khurana:
Because

* We want to enable first-run experiments

This commit

* Adds a checkbox to enable the capability for first-run experiments and test cases to verify new changes
image

fix #7522


jaredlockhart (approved):
Yep looks good, works good, let's make sure to file a follow up to manage the interactions between like

- is_first_run
- sticky_targeting
- advanced_targeting

Since these all have a strong interdependency.

yashika-khurana:
thanks @jaredlockhart  perfect, filled a ticket https://mozilla-hub.atlassian.net/browse/EXP-2739

PR #7662, opened 2022-09-01
fix #7661 chore(nimbus): Update integration_sdk_shell command

### LINKED ISSUE #7661: Update integration_sdk_shell command
When we try to run `make integration_sdk_shell` locally, it will give us this error 
`Pulling app (app:deploy)...
ERROR: pull access denied for app, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
make: *** [integration_sdk_shell] Error 1
`

We need to build prod while running those test cases, so update make command and config to include `build_prod` command

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2741)

yashika-khurana:
Because

* When we try to run `make integration_sdk_shell` locally, it will give us this error
`Pulling app (app:deploy)... ERROR: pull access denied for app, repository does not exist or may require 'docker login': denied: requested access to the resource is denied make: *** [integration_sdk_shell] Error 1`

We need to build prod while running those test cases

This commit

* Update make command and config to include `build_prod` command

fix #7661


jrbenny35 (changes requested):
I see changes in ```.circleci/``` in here, did you mean to include those?

Makefile
```diff
@@ -173,7 +173,7 @@ dependabot_approve:
 integration_shell:
```
jaredlockhart, line 173: Sorry I meant to add it to all of the integration commands

jaredlockhart (changes requested):

jaredlockhart:
> I see changes in `.circleci/` in here, did you mean to include those?

Yeah adding the `build_prod` as a dependency to those make commands means if they're run individually as they were before it'll rebuild the containers in between `make up_prod_detached` and `make integration_test_legacy` whereas if you fold them together into a single command make will run `build_prod` once at the start.

Makefile
```diff
@@ -182,11 +182,11 @@ integration_vnc_up:
 integration_vnc_up_detached:
```
jaredlockhart, line 182: What about these two?

jaredlockhart (approved):
Tytytyty 🙏

PR #7634, opened 2022-08-23
fixes #7592 test(nimbus): Added end-to-end test for telemetry verification during enrollment and unenrollment

jrbenny35:
Because
* Making sure telemetry is sent when a client enrolls and unenrolls is very important. We did not have automated tests for this though

This Commit
* Adds automated checks for telemetry enroll/unenroll events.

Since this test can take a varying amount of time to complete, I utilized pytest-xdist to group tests so that this test can be run immediately, to save time.
I also had to write a ping server to collect and verify the pings, that is included in this PR.

PR #7664, opened 2022-09-01
fix #7663 chore(nimbus): Run integration test on updation of tests

### LINKED ISSUE #7663: Run integration tests if test files change
In our integration tests we check for changes in some file paths to determine whether we should run tests, however we're not checking for changes to the test files so we can make changes there that might break the tests but not be run as part of the PR, so we should update all the path checks to include `app/tests/` so that any changes to the test files/requirements/etc force the tests to be run.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2742)

yashika-khurana:
Because

* In our integration tests we check for changes in some file paths to determine whether we should run tests, however, we're not checking for changes to the test files so we can make changes there that might break the tests but not be run as part of the PR

This commit

* Update all the path checks to include app/tests/ so that any changes to the test files/requirements/etc force the tests to be run.


.circleci/config.yml
```diff
@@ -328,4 +328,4 @@ workflows:
             branches:
               only: main
           requires:
-           - check
+            - check
```
jaredlockhart, line 331: Wait how was this valid before?

.circleci/config.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - run:
           name: Check file paths
           command: |
-            if git diff --name-only main HEAD | grep -E 'app/experimenter/targeting|app/experimenter/experiments'
+            if git diff --name-only main HEAD | grep -E 'app/experimenter/targeting|app/experimenter/experiments|app/tests'
```
jrbenny35, line 68: I don't think we should change this. We shouldn't need to run the integration tests against beta/release for every commit.

jrbenny35 (changes requested):

.circleci/config.yml
yashika-khurana: yeah I am thinking that too

.circleci/config.yml
jaredlockhart: What do you mean on every commit?

yashika-khurana:
@jaredlockhart @jrbenny35 okay so found the issue- the reason why these test cases were failing on other PR- https://app.circleci.com/pipelines/github/mozilla/experimenter/14647/workflows/47c166e9-c327-49f5-8232-8cf64e1f257a/jobs/43970 
The reason is: that when we merged this PR
https://github.com/mozilla/experimenter/pulls?q=is%3Apr+is%3Aclosed, in this it removed the languages field as a parameter, where as mobile experiments were passing languages, our check didn't run at all during that period and we merged it

jaredlockhart:
@yashika-khurana which PR? the link you added goes to the list of all PRs

yashika-khurana:
@jaredlockhart my bad, here it is https://github.com/mozilla/experimenter/pull/7634

.circleci/config.yml
jaredlockhart: @yashika-khurana Why did you take this out?

yashika-khurana: oh I read the above @jrbenny35 as he suggested, I can bring it back

jaredlockhart: Right but adding `app/tests` won't cause it to run on every commit, only changes to the tests folder, which is specifically what this ticket is requesting.  Just because someone makes a suggestion in a PR does not mean it should be followed, and @jrbenny35 I would recommend framing feedback on PRs as questions rather than statements in general.

yashika-khurana: @jaredlockhart its back now

jrbenny35: > Right but adding `app/tests` won't cause it to run on every commit, only changes to the tests folder, which is specifically what this ticket is requesting. Just because someone makes a suggestion in a PR does not mean it should be followed, and @jrbenny35 I would recommend framing feedback on PRs as questions rather than statements in general.

Sorry, was sort of in a rush, should've saved my feedback for a bit later. So yes, it wont run on every commit but running the tests for beta/release for changes in app/tests is a not needed in my opinion. Running the tests against nightly should be fine but we also don't do that many changes to app/tests. 

Another point here is the path should be ```app/tests/integration/nimbus``` as we have a separate job for ```app/tests/integration/legacy```

app/tests/integration/nimbus/conftest.py
```diff
@@ -217,14 +217,15 @@ def _create_experiment(selenium):
 
 @pytest.fixture
 def create_basic_experiment():
-    def _create_basic_experiment(name, app, targeting):
+    def _create_basic_experiment(name, app, targeting, languages=[]):
```
jrbenny35, line 220: whoops 😅

.circleci/config.yml
jaredlockhart: @jrbenny35 specifically the issue here is that changes in `app/tests` including:
 
- dependency changes
- test file changes

can break the tests, which is what just happened and resulted in the integration tests being broken in main.  So we could say that the beta/release tests are so close to the nightly tests that if the nightly tests pass we expect the beta/release tests to work, but there could still be cases where one passes and the other fails, so the goal here is to say ANY changes to the tests folder should force ALL integration tests to run so we don't introduce another breaking regression into main.

jrbenny35 (approved):

jaredlockhart:
@yashika-khurana Awesome, tests are all fixed, and this should prevent future test dependency changes from accidentally breaking them on prod, great job 🙏 And thank you for taking a look and digging in @jrbenny35 ❤️

PR #7681, opened 2022-09-06
fix #7653 feat(nimbus): cache start/end date on nimbus experiment model

### LINKED ISSUE #7653: Remove changelog queries from GQL queries altogether
So looking at the database query 'insights' page on GCP, the longest query we have right now is the query that pulls out all the changelogs and attaches them to experiments to lookup the start/end dates, at ~3.5s avg.  We can cut this query out altogether if we denormalize the start/end dates right on to the experiment model.  So let's:

* Add start_date/end_date datetime fields (nullable) to the NimbusExperiment model
* In the start/end date methods, check those fields first and if they're set, return them otherwise proceed to query them as normal and then at the end of the lookup, write them and save them to those fields
* Remove the changelog from the prefetched list in the GQL queries

Cutting the changelog query/join out should shave about half the time off those queries altogether.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2737)

jaredlockhart:
Because

* The slowest part of the landing page API call is the call/join to the changelog table
* This call is used to compute the start/end dates of experiments
* We can cache these values right on the experiment since they are immutable once set

This commit

* Adds _start_date/_end_date fields to NimbusExperiment
* Adds logic to read/write them to the start_date/end_date property methods
* Updates the NimbusExperimentFactory to include writing to those fields
* Adds new tests to NimbusExperiment tests to check the cached/uncached cases
* Updates all other necessary tests to use the new start_date/end_date methods


yashika-khurana (approved):

yashika-khurana:
LGTM! 🚀

PR #7696, opened 2022-09-07
For #7273 test(nimbus): Add parameterized tests for rollouts dashboard

eliserichards:
Because...

* We want to ensure that the rollouts dashboard url works for certain statuses

This commit...

* Parameterizes the model tests so that we are checking all of the status states

### DIFF
--- a/app/experimenter/experiments/tests/test_models.py
+++ b/app/experimenter/experiments/tests/test_models.py
@@ -1064,10 +1064,18 @@ class TestNimbusExperiment(TestCase):
             ),
         )
 
-    def test_monitoring_dashboard_returns_url_when_rollout(self):
+    @parameterized.expand(
+        [
+            NimbusExperiment.Status.DRAFT,
+            NimbusExperiment.Status.COMPLETE,
+            NimbusExperiment.Status.LIVE,
+            NimbusExperiment.Status.PREVIEW,
+        ]
+    )
+    def test_monitoring_dashboard_returns_url_when_rollout(self, status):
         experiment = NimbusExperimentFactory.create(
             slug="experiment",
-            status=NimbusExperiment.Status.DRAFT,
+            status=status,
             is_rollout=True,
         )
 
@@ -1126,15 +1134,6 @@ class TestNimbusExperiment(TestCase):
             ),
         )
 
-    def test_rollouts_monitoring_dashboard_url_is_none_when_rollout_not_begun(self):
-        experiment = NimbusExperimentFactory.create_with_lifecycle(
-            NimbusExperimentFactory.Lifecycles.CREATED,
-            slug="rollout-1-slug",
-            is_rollout=True,
-            status=NimbusExperiment.Status.DRAFT,
-        )
-        self.assertIsNone(experiment.rollout_monitoring_dashboard_url)
-
     def test_rollouts_monitoring_dashboard_returns_correct_formatted_url(self):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.LAUNCH_APPROVE_APPROVE,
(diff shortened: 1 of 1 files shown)


jaredlockhart (commented):
Ah yeeeeee

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1064,16 +1064,27 @@ def test_monitoring_dashboard_url_is_valid_when_experiment_not_begun(self):
             ),
         )
 
-    def test_monitoring_dashboard_returns_url_when_rollout(self):
+    @parameterized.expand(
+        [
+            (True, NimbusExperiment.Status.DRAFT),
+            (True, NimbusExperiment.Status.COMPLETE),
+            (True, NimbusExperiment.Status.LIVE),
+            (True, NimbusExperiment.Status.PREVIEW),
```
jaredlockhart, line 1072: If it's always true then that parameter is probably superfluous.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 import os.path
 from decimal import Decimal
+from logging.config import valid_ident
```
jaredlockhart: Did vscode throw that in?  Does it all the time to me, silly thing.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1064,16 +1065,25 @@ def test_monitoring_dashboard_url_is_valid_when_experiment_not_begun(self):
...
             slug="experiment",
-            status=NimbusExperiment.Status.DRAFT,
+            status=status,
             is_rollout=True,
         )
 
         from_date = datetime.date.today() - datetime.timedelta(days=1)
         to_date = datetime.date.today() + datetime.timedelta(days=1)
 
+        self.assertIsNotNone(experiment.monitoring_dashboard_url)
```
jaredlockhart, line 1085: You don't really need this if you're checking it again on the next line?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1155,21 +1156,56 @@ def test_rollouts_monitoring_dashboard_returns_correct_formatted_url(self):
...
             is_rollout=True,
-            status=NimbusExperiment.Status.COMPLETE,
-        )
-        expected_slug = "rollout_1_slug"
-        self.assertEqual(
-            experiment.rollout_monitoring_dashboard_url,
-            settings.ROLLOUT_MONITORING_URL.format(
-                slug=expected_slug,
-            ),
+            status=status,
```
jaredlockhart, line 1194: You don't need to explicitly set the status here, that's part of the lifecycles.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1155,21 +1155,56 @@ def test_rollouts_monitoring_dashboard_returns_correct_formatted_url(self):
...
         )
 
+        if valid_status:
+            expected_slug = "rollout_1_slug"
```
jaredlockhart, line 1198: The slug is hardcoded here twice, just move that up to the top of the test, or refer to `experiment.slug` here.

jaredlockhart (changes requested):
Looking good!

app/experimenter/experiments/tests/test_models.py
eliserichards: I already deleted it hehe

app/experimenter/experiments/tests/test_models.py
eliserichards: Oh fair

app/experimenter/experiments/tests/test_models.py
eliserichards: It's a different formatted slug here. It's formatted from "-" to "_" (https://github.com/mozilla/experimenter/pull/7640#discussion_r958853787 hehe)

eliserichards: I wanted to leave it outside of the `assert` so that it's clear that it's not just the normal slug as-is, but I'm open to just hardcoding it inside the `assert` since it's only used once. WDYT?

app/experimenter/experiments/tests/test_models.py
eliserichards: 🤯

app/experimenter/experiments/tests/test_models.py
jaredlockhart: Oh the conversion only goes one way, `-` to `_` so if you use the `_` version for both then they'll just match and nothing will get transformed, but then you're not really testing the conversion in the logic.  Yeah okay I see your point I forgot about the `-` `_` thing.  Then yeah this is fine.

jaredlockhart (approved):
👍 👏 🎉 👍 👏 🎉 👍 👏 🎉 👍 👏 🎉 👍 👏 🎉 👍 👏 🎉

PR #7702, opened 2022-09-08
fix #7666 feat(nimbus): More info on landing page

### LINKED ISSUE #7666: Add more information to the main page
1. Adding more fields to the main page- channel, version (both min and max in table view and in filter), application, population, rollout

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2704)

### LINKED ISSUE #7703: Remove redundant info on Landing page
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2748)

yashika-khurana:
Because

* We want to show more data on the landing page and make landing page less noisy by removing the slug value

This commit

* Adds new fields on the main page- `Application`, `Channel`, `Population percentage`, `Firefox min version`, `Firefox max version` with an ability to sort them as well
* Removes slug value from the table
* Test cases to verify new changes

PS: It fixes 2 issues in single PR

Before changes: 
image

After changes:
image

fix #7666 
fix #7703


jaredlockhart (approved):
Yes this is awesome so much better.  

We'll probably have to deal with spacing/layout at some point but at least now we have the information we need on that page so thank you.  

I was testing this locally and couldn't figure out why I couldn't see the changes until I realized it's only on the Live tab, but really we want to know this for all the tabs.  Can you please file a followup to use this same table on all tabs and get rid of the other tables that have less information? 🙏

PR #7716, opened 2022-09-08
fix #7715 bug(nimbus): don't clone start/end dates

### LINKED ISSUE #7715: Exclude _start_date/_end_date from clone method
We do not want to clone the cached start/end date fields we recently added, so let's reset those to `None` in the clone method.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2777)

jaredlockhart:
Because

* We recently added _start_date/_end_date fields to the NimbusExperiment model to cache the start/end dates for faster lookups in the API
* The clone method copies the entire NimbusExperiment object and requires explicitly setting to None any fields which should not be cloned
* This results in newly added fields being unintentionally cloned unless they're explicitly excluded

This commit

* Explicitly excludes the _start_date/_end_date fields from being cloned
* Reorganizes the clone test assertions to show which fields are being explicitly reset and which are being cloned

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -683,6 +683,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         cloned.results_data = None
         cloned.takeaways_summary = None
         cloned.conclusion_recommendation = None
+        cloned._start_date = None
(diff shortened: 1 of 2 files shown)


yashika-khurana (approved):
LGTM! 🚀

PR #7719, opened 2022-09-08
fix #7582: feat(nimbus): add custom targeting for background task mes…

### LINKED ISSUE #7582: Add custom targeting for background task messaging
This tracks adding custom targeting for background task messaging:
- `defaultProfile.currentDate` is 28+ days in the past
- `defaultProfile.version` is older than this version
There may be some other restrictions as well.

ncalexan:
…saging.

This is intended to target lapsed users who are background updated to Firefox 106.  We check for:
- Win 10+; and
- either:
  - no serialized targeting state from the default profile; or
  - both:
    - the default profile was run more than 28 days ago; and
    - it was an olderversion than the current background update task

We'll target exactly Firefox 106 in the Experimenter audience independently.


ncalexan:
@jaredlockhart so, it rather looks like there is some kind of test that runs a Desktop Firefox and tries to ensure that the given targeting configuration doesn't cause an error.  It's timing out, which isn't helpful, but this targeting won't succeed in any circumstance on Desktop because one of the variables, namely `defaultProfile`, is only populated for background tasks.  The expedient workaround might be to do one of:
1.  Not test this directly at all (i.e., not against Desktop)
2.  Teach Desktop to include `defaultProfile: {}` in its targeting context, silly though that is in general.
If you can point me to more details of the failure I might be able to think about other ways to work around this.  Guidance appreciated!

jaredlockhart:
@ncalexan Hrmmm yes interesting.... Let me think about this for a minute and get back to you.

jaredlockhart:
@ncalexan Okay yeah, I think the simplest solution for now is to inject

`{defaultProfile: {}}` 

here

https://github.com/mozilla/experimenter/blob/main/app/tests/integration/nimbus/utils/filter_expression.js#L19-L23

until we can find a more rigorous solution so at least you're not blocked.

app/tests/integration/nimbus/conftest.py
```diff
@@ -126,7 +126,7 @@ def experiment_url(base_url, default_data, slugify):
 
 @pytest.fixture
 def experiment_name(request):
-    return f"{request.node.name[:76]}-{str(uuid.uuid4())[:4]}"
+    return f"{request.node.name[:75]}-{str(uuid.uuid4())[:4]}"
```
jaredlockhart, line 129: Oh I see what happened, we had a latent bug in this name formatting that this targeting config accidentally triggered because it has a long name.

jaredlockhart: That's what was causing the timeout @ncalexan , unrelated to the actual targeting failure, which fails with a more specific message.

jaredlockhart:
Anyways adding the `defaultProfile` key to the targeting context does indeed make the test pass, so until we can have a more rigorous validation against the background task process, this should unblock you.  Does that work for your @ncalexan ?

app/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -18,6 +18,7 @@ async function remoteSettings(arguments) {
 
     const context = TargetingContext.TargetingContext.combineContexts(
         _experiment,
+        {defaultProfile: {}}, // Workaround for supporting background tasks
```
ncalexan, line 21: Can I rewrite the comment, or bulk out the commit comment?  Say something like:

In Firefox Desktop, the targeting context is not fixed.  From the perspective of Experimenter there is only one `firefox-desktop` application, but the client splits Nimbus features between two sub-application IDs: `firefox-desktop` and `firefox-desktop-background-task`.  The `firefox-desktop-background-task` sub-application applies to `firefox --backgroundtask ...` invocations.  Not all features apply to both sub-application IDs, and the targeting context for the `firefox-desktop-background-task` sub-application ID includes details about the default browsing profile not present to the regular `firefox-desktop` ID.  Ideally, this `defaultProfile` would only appear when the test experiment's feature applies to `firefox-desktop-background-task`, but since the test experiments created here use a constant feature, this isn't possible at this time.

ncalexan (commented):
Thanks for jumping on this, @jaredlockhart -- I appreciate it.

app/tests/integration/nimbus/utils/filter_expression.js
jaredlockhart: Yeah we use squash commits so we can add that information to the commit at squash time.

app/tests/integration/nimbus/conftest.py
jaredlockhart: Yeah I was thinking that but that might have obfuscated the bug since it would be trimming the salt at the end that we inject to prevent name collisions during test runs, so that might have inadvertently caused other confusing failures.

jaredlockhart (approved):
Okay I added the necessary fixes, the targeting configuration itself looks correct to me, R+ from me pending @jrbenny35 double checking.

jrbenny35 (approved):
Nice find

PR #7722, opened 2022-09-09
fix #7712 feat(nimbus): Add first run required in targeting

### LINKED ISSUE #7712: Add is_first_run_required property to TargetingConfigs
Let's add another property to the Targeting Configs called `is_first_run_required` similar to how we did `sticky_required`, such that if it is true, we force set the Is First Run checkbox on the audience page the same way we do for sticky required.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2774)

yashika-khurana:
Because

* We want to support first run experiments and auto enable based on targeting selected

This commit

* Add another property to the Targeting Configs called` is_first_run_required` that if it is true, we force to set the Is First Run checkbox on the audience page
* Test cases to verify changes

fix #7712 

Screenshot for new changes (temporary enabled first run required for the targeting)
image

### DIFF
--- a/app/experimenter/docs/openapi-schema.json
+++ b/app/experimenter/docs/openapi-schema.json
@@ -3399,6 +3399,9 @@
                 },
                 "stickyRequired": {
                   "type": "boolean"
+                },
+                "isFirstRunRequired": {
+                  "type": "boolean"
                 }
               },
               "required": [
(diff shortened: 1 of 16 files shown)


jaredlockhart (approved):
Wow @yashika-khurana , fastest coder in the west!  Looks all exactly right, tested locally, works perfectly.  Fantastic great job A+ 10/10 cats 🐱 

grumpy-cat-tarder-sauce

PR #7724, opened 2022-09-12
fix #7599 chore(nimbus): explicitly convert snake to camel case

### LINKED ISSUE #7599: Move camelCase fields from graphql to serializers
The GraphQL framework we are using (Graphene) will automatically convert snake_case fields to camelCase fields, so the fields in the serializer are all in snake_case but are referenced in the frontend by camelCase.  To prepare for removing the GraphQL layer, let’s move the camelcasing back a layer to the serializer layer.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2686)

jaredlockhart:
Because

* snake_case is idiomatic for Python
* camelCase is idiomatic for javascript/typescript
* The graphql library graphene automatically converts snake_case to camelCase
* After we remove graphene, we want to keep the frontend interface the same to not have to modify the entire frontend
* We need to disable the automatic snake/camel case conversion and explicitly handle it in the serializers

This commit

* Disables the automatic snake/camel case conversion
* Updates all the graphene/serializer layers to explicitly map snake_case fields to their camelCase counterparts while keeping the frontend interface unmodified


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -717,54 +734,124 @@ class NimbusExperimentSerializer(
...
-            "takeaways_summary",
-            "targeting_config_slug",
-            "total_enrolled_clients",
-            "treatment_branches",
+            "statusNext",
+            "takeawaysSummary",
+            "targetingConfigSlug",
+            "totalEnrolledClients",
+            "treatmentBranches",
+            "warnFeatureSchema",
```
yashika-khurana, line 854: I really liked 🎉  the sorted fields it increases readability 👀 , I think we should always try to write/add fields in a sorted way

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1090,104 +1178,172 @@ def validate_feature_value(self, value):
...
+        required=False,
+        allow_blank=True,
+    )
+    warnFeatureSchema = serializers.BooleanField(source="warn_feature_schema")
 
     class Meta:
         model = NimbusExperiment
-        exclude = ("id",)
+        fields = [
+            "application",
```
yashika-khurana, line 1294: I think we should write this too in a sorted way 🤔

app/experimenter/experiments/api/v5/types.py
```diff
@@ -79,6 +79,7 @@ class NimbusCountryType(DjangoObjectType):
 
     class Meta:
         model = Country
+        fields = ("id", "code", "name")
```
yashika-khurana, line 82: since we are adding all fields, how about `fields = "__all__"`

app/experimenter/experiments/api/v5/types.py
```diff
@@ -88,6 +89,7 @@ class NimbusLocaleType(DjangoObjectType):
 
     class Meta:
         model = Locale
+        fields = ("id", "code", "name")
```
yashika-khurana, line 92: `fields="__all__"`?

app/experimenter/experiments/api/v5/types.py
```diff
@@ -97,14 +99,17 @@ class NimbusLanguageType(DjangoObjectType):
 
     class Meta:
         model = Language
+        fields = ("id", "code", "name")
```
yashika-khurana, line 102: `fields ="__all__"`?

app/experimenter/experiments/api/v5/types.py
```diff
@@ -429,54 +562,31 @@ def resolve_ready_for_review(self, info):
...
-    def resolve_targeting_config_slug(self, info):
+    def resolve_targetingConfigSlug(self, info):
         if self.targeting_config_slug in self.TargetingConfig:
             return self.TargetingConfig(self.targeting_config_slug).value
         return self.targeting_config_slug
 
-    def resolve_targeting_config(self, info):
-
+    def resolve_targetingConfig(self, info):
+        targeting_config = NimbusExperiment.TARGETING_CONFIGS[self.targeting_config_slug]
```
yashika-khurana, line 571: oh that's a great change, I mean its a small change, but it is for sure more readable and avoids re-reading configs again and again

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1042,28 +1042,19 @@ def test_targeting_config(self):
         self.assertEqual(response.status_code, 200, response.content)
         content = json.loads(response.content)
         experiment_data = content["data"]["experimentBySlug"]
+        targeting_config = NimbusExperiment.TARGETING_CONFIGS[
```
yashika-khurana, line 1045: 🎉

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/RejectionReason.tsx
```diff
@@ -21,13 +18,13 @@ export const RejectionReason = ({
     rejectionEvent!;
 
   const rejectionActionDescription = useMemo(() => {
-    if (oldStatus === NimbusChangeLogOldStatus.LIVE) {
-      if (oldStatusNext === NimbusChangeLogOldStatusNext.LIVE) {
```
yashika-khurana, line 25: Oh wait, I want to understand why we are removing `NimbusChangeLogOldStatus`?? 💭

yashika-khurana (changes requested):
Overall great job @jaredlockhart 🎉 👏 , we are more close to get rid of Graphql, I have one question about why we are getting rid of `NimbusExperimentLogOldStatus`? and a few minor suggestions.....

app/experimenter/experiments/api/v5/serializers.py
yashika-khurana: and I think everywhere where we have `fields`

jaredlockhart: Agree, let's do it as a followup https://github.com/mozilla/experimenter/issues/7735

app/experimenter/experiments/api/v5/types.py
jaredlockhart: Well, the pattern I was finding was that in a bunch of places we were doing essentially that but by using `exclude` and then fields were unintentionally showing up in the objects and it wasn't clear because there wasn't an explicit list of fields, so I started feeling like we should avoid that pattern and always explicitly list fields.

yashika-khurana: that's a great point

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/RejectionReason.tsx
yashika-khurana: great explanation 🥳 👏 , now its clear, thank you @jaredlockhart 🙏

yashika-khurana (approved):
Thank you for filling the issue and for a great explanation 🎉

Issue #7464, opened 2022-06-28
Evaluate JEXL subexpressions in desktop integration tests 


jaredlockhart:
As a followup to https://github.com/mozilla/experimenter/issues/7449 we discovered that while we can evaluate subexpressions in the Rust JEXL evaluator and meaningfully determine if they include undefined variables or invalid comparisons, we can't do the same on desktop without changes to the JS JEXL evaluator.

After we make the following changes:

https://github.com/mozilla/mozjexl/issues/25

And get them landed into MC:

https://bugzilla.mozilla.org/show_bug.cgi?id=1773388#c2

then we can add the subexpression evaluation to desktop the way we do for the sdk.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2573)

jaredlockhart:
Bump

jaredlockhart:
Still worth doing

PR #7741, opened 2022-09-13
fix #7740 type(test): Extend targeting integration tests

### LINKED ISSUE #6791: Integration test failure: targeting tests being tricked by targeting slug
Just saw the weirdest behaviour in the targeting integration tests.  We pass a couple parameters into the test, one is an application slug like `DESKTOP`, and the other is a targeting config slug.  If the targeting config slug also includes the word `desktop` like `something_desktop_something` then it will trick pytest into running the tests with applications other than desktop which will then fail.

https://app.circleci.com/pipelines/github/mozilla/experimenter/11753/workflows/223c1fac-3cd5-4d94-a25b-6c07d0bbd489/jobs/35944

We've probably got to dig into a better way to parameterize the tests so it doesn't get tricked into running the targeting tests on applications other than desktop.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-1940)

### LINKED ISSUE #7740: [Desktop] extend tartgeting tests to use additional audience fields
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2785)

jrbenny35:
Because
* We currently check each targeting expression validates in the browser, but it is a very basic expression

This Commit
* Extends that test with additional fiends such as channel, locale, region and versions.

Also closes #6791


jrbenny35:
An example of what the full targeting expression will look like:

```
(browserSettings.update.channel == "nightly") && ('app.shield.optoutstudies.enabled'|preferenceValue) && (version|versionCompare('120.*') <= 0) && ((isFirstStartup && !('trailhead.firstrun.didSeeAboutWelcome'|preferenceValue))) && (version|versionCompare('100.!') >= 0) && (locale in ['en-CA']) && (region in ['CA'])
```

app/tests/integration/nimbus/test_telemetry.py
```diff
@@ -97,8 +98,37 @@ def test_check_telemetry_enrollment_unenrollment(
...
+        #     {
+        #         "description": "treatment branch",
+        #         "name": "Branch 2",
+        #         "ratio": 50,
+        #         "featureEnabled": False,
+        #         "featureValue": "",
+        #     }
+        # ],
+        # population_percent="100",
+        # total_enrolled_clients=55,
```
jaredlockhart, line 130: Left this in by mistake?

jaredlockhart (changes requested):
Oh awesome!  Thanks for getting to this, increased coverage ftw 🎉

jaredlockhart:
Ah yeah, unfortunately this isn't going to work the way I'd hoped, just testing this out locally.  I tried injecting some invalid parts into one of the targeting constants like so

```js
await targetingContext.evalWithDefault("(browserSettings.update.channel == 'release') && asdfasdf && ('app.shield.optoutstudies.enabled'|preferenceValue) && asdfasdf && (version|versionCompare('80.!') >= 0)")
false
```

but since the channel test comes first, and it doesn't match and returns `false` that short circuits the whole expression and returns false so we never get around to evaluating the invalid part.  You can see if that if you put the invalid part at the start of the expression:

```
await targetingContext.evalWithDefault("asdfasdf && (browserSettings.update.channel == 'release') && ('app.shield.optoutstudies.enabled'|preferenceValue) && asdfasdf && (version|versionCompare('80.!') >= 0)")
null
```

So we're gonna have to do this a little differently.  Instead of setting all the fields on every targeting test, we have to keep those with no additional audience fields so that only the targeting expression from the targeting constant evaluated.  And then we should have a separate suite of tests that tests each of the audience fields individually and sets it and checks for validity so that we can check their emitted jexl expression in isolation.

Sorry to make you write all this up, but the other changes in here, like switching the desktop targeting tests to use the graphql endpoint instead of the ui, and fixing the name collision thing, are all good, so maybe a thing to do is just modify this to take out setting the additional fields but keep the graphql/naming part, land that, and then add the additional tests I just described above as a followup.

The other thing this highlights to me is we SHOULD do this https://github.com/mozilla/experimenter/issues/7464 even though it won't catch invalid dotted lookups, it would still catch the examples I'm showing above since it would evaluate each key in isolation.

jrbenny35:
Closing in favor of https://github.com/mozilla/experimenter/pull/7754

PR #7745, opened 2022-09-14
fix #7739 bug(nimbus): filter related entities (screenshots, documentation links) by parent

### LINKED ISSUE #7739: Duplicate document links
I just noticed all the experiment detail pages show a huge number of duplicated links.  Looking at the graphql data, it looks like every documentation link is being returned rather than just those specific to the experiment.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2788)

jaredlockhart:
Because

* We are in the process of refactoring the graphql and rest apis
* This involves moving from implicitly converting snake_case to camelCase fields which requires explicitly mapping the serialized field back to its model field
* It seems that the DjangoListField that was being used falls back to returning all entities from a model rather than those that foreign key on to a parent
* The tests did not explicitly test against this so it was missed

This commit

* Changes from using the DjangoListField to resolve foreign keys to explicitly declaring a resolver that does
* Adds tests to ensure parent foreign keys are followed in related entities

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -2,7 +2,6 @@ import json
 
 import graphene
 from django.contrib.auth import get_user_model
-from graphene_django import DjangoListField
 from graphene_django.types import DjangoObjectType
 
 from experimenter.base.models import Country, Language, Locale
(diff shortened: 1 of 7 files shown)


jaredlockhart:
Let's block this on https://github.com/mozilla/experimenter/pull/7748 then update it to the pre snake/camel conversion state, and then when we reintroduce the snake/camel conversion this case will be covered.

jaredlockhart:
Actually I'll just fold this into reapplying https://github.com/mozilla/experimenter/issues/7599

PR #7751, opened 2022-09-14
fix #7750 chore(nimbus): add env var for printing more DOM during testing

### LINKED ISSUE #7750: Testing library does not print entire DOM if too large
There is an environment variable we can set to show more of the DOM. The default for the testing library is 7000 (characters, I think), per docs here: https://testing-library.com/docs/dom-testing-library/api-debugging

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2800)

mikewilli:
Because

* it can be difficult to figure out why new tests fail to find DOM elements as expected
* the testing library defaults to only printing 7000 DOM characters

This commit

* adds the environment variable explicitly to the Makefile so that it is easier to change for local testing
* documents this change in the README


jaredlockhart (approved):
Awesome, very handy, I've hit this multiple times so this should help debugging 🙏 Thanks @mikewilli !

PR #7754, opened 2022-09-15
fix #7753 type(chore): Update targeting integration tests to use graphQL API

### LINKED ISSUE #6791: Integration test failure: targeting tests being tricked by targeting slug

### LINKED ISSUE #7753: Update targeting integration tests to use graphql API
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2801)

jrbenny35:
Because:
* Using the UI to create experiments for the targeting tests can take a lot of time

This Commit:
* Switches to using the graphQL API to create experiments resulting in faster tests.

Also fixes #6791

### DIFF
--- a/app/tests/integration/nimbus/conftest.py
+++ b/app/tests/integration/nimbus/conftest.py
@@ -104,7 +104,7 @@ def application(request):
     Returns the current application to use for testing
     Will also parametrize the tests
     """
-    return request.param 
+    return request.param
 
 
 @pytest.fixture(scope="session", autouse=True)
--- a/app/tests/integration/nimbus/conftest.py
+++ b/app/tests/integration/nimbus/conftest.py
@@ -24,7 +24,7 @@ from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
 APPLICATION_FEATURE_IDS = {
-    BaseExperimentApplications.DESKTOP: "1",
+    BaseExperimentApplications.FIREFOX_DESKTOP: "1",
     BaseExperimentApplications.FENIX: "2",
     BaseExperimentApplications.IOS: "3",
     BaseExperimentApplications.FOCUS_ANDROID: "4",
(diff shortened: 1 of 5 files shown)


app/tests/integration/nimbus/conftest.py
```diff
@@ -320,6 +313,48 @@ def _language_database_id_loader(languages=None):
...
+        path = os.path.join("/", path, "experimenter/base/fixtures/locales.json")
+        with open(path) as file:
+            data = json.loads(file.read())
+            for locale in locales:
+                for item in data:
+                    if locale in item["fields"]["code"]:
+                        locale_list.append(item["pk"])
+        return locale_list
+
+    return _locales_database_id_loader
```
jrbenny35, line 355: I could probably write one function to do all of this, but maybe I save that for a contributor??

jaredlockhart: Actually the other thing I was thinking is you could also query this info from the configuration endpoint.  I thought we were doing that at some point but now I don't see it.  Or did we figure out that it was too slow/janky?  Reading them from the disk works too.

jrbenny35: I'd need help with that if you think it's worth doing.

jaredlockhart (approved):
Yep all looks good, thank you, nice optimization.  I think we should keep this pattern up and use the api for anything that isn't explicitly testing the nimbus ui.  Thanks @jrbenny35 !

Oh also I just noticed, the commit formatting is slightly off, it's not

```
type(chore)
```

it should be

```
chore(nimbus)
```

jrbenny35:
> Yep all looks good, thank you, nice optimization. I think we should keep this pattern up and use the api for anything that isn't explicitly testing the nimbus ui. Thanks @jrbenny35 !
> 
> Oh also I just noticed, the commit formatting is slightly off, it's not
> 
> ```
> type(chore)
> ```
> 
> it should be
> 
> ```
> chore(nimbus)
> ```

Sorry about that!

PR #7758, opened 2022-09-15
fix #7740 test(nimbus): Extend targeting integration tests

### LINKED ISSUE #7740: [Desktop] extend tartgeting tests to use additional audience fields

jrbenny35:
Because
* We currently check each targeting expression validates in the browser, but it is a very basic expression

This Commit
* Adds an addition test with additional fiends such as channel, locale, region and versions.


app/tests/integration/nimbus/test_firefox_targeting.py
```diff
@@ -74,7 +74,7 @@ def targeting_config_slug(request):
 
 
 @pytest.mark.run_targeting
```
yashika-khurana, line 76: do you think we should rename this to `run_advance_targeting`? 🤔  just to signify it contains more than required fields

yashika-khurana (changes requested):
Just a minor suggestion, well done @jrbenny35  👏

app/tests/integration/nimbus/test_firefox_targeting.py
jrbenny35: Oh this marker is just for selecting tests when pytest runs. These tests should run with the old targeting tests.

yashika-khurana: yup absolutely, I mean to say we are changing the meaning of marker from `targeting` to `advance targeting`, so would recommend refactoring that

jaredlockhart: @yashika-khurana Yes I agree, however I think we should do a larger refactor of all the integration tests to more clearly separate and identify different testing concerns now that we've expanded the scope so much, so I'll file a followup for that with ideas for a better layout.

app/tests/integration/nimbus/test_firefox_targeting.py
```diff
@@ -133,3 +133,82 @@ def test_check_targeting(
...
+@pytest.mark.parametrize(
+    "audience_field",
+    [
+        {"channel": "NIGHTLY"},
+        {"firefoxMinVersion": "FIREFOX_100"},
+        {"firefoxMaxVersion": "FIREFOX_120"},
+        {"locales": [37]},
+        {"countries": [42]},
+        {"proposedEnrollment": "14"},
+        {"proposedDuration": "30"},
```
jaredlockhart, line 147: These don't go into targeting, so we don't need them here.

app/tests/integration/nimbus/test_firefox_targeting.py
```diff
@@ -133,3 +133,82 @@ def test_check_targeting(
...
+                "description": "treatment branch",
+                "name": "Branch 2",
+                "ratio": 50,
+                "featureEnabled": False,
+                "featureValue": "",
+            }
+        ],
+        "populationPercent": "100",
+        "totalEnrolledClients": 55,
+    }
```
jaredlockhart, line 192: We don't have like a fixture for this in conftest somewhere?

app/tests/integration/nimbus/test_firefox_targeting.py
```diff
@@ -133,3 +133,82 @@ def test_check_targeting(
...
+    data.update(audience_field)
```
jaredlockhart, line 193: Perf 👍

app/tests/integration/nimbus/test_firefox_targeting.py
```diff
@@ -133,3 +133,82 @@ def test_check_targeting(
...
+    selenium.get("about:blank")
+    with open("nimbus/utils/filter_expression.js") as js:
+        result = Browser.execute_script(
+            selenium,
+            targeting,
+            json.dumps({"experiment": recipe}),
+            script=js.read(),
+            context="chrome",
+        )
+    assert result is not None, "Invalid Targeting, or bad recipe"
```
jaredlockhart, line 214: I love how succinct we've gotten these 🙏

jaredlockhart (changes requested):
Awesome thank you @jrbenny35 🙏 Just take those proposed durations out, and maybe there's a way to move that json fixture out?

app/tests/integration/nimbus/test_firefox_targeting.py
jrbenny35: Good idea, we can add that

jaredlockhart (approved):
Fantastic thank you @jrbenny35 🎉 🎉 🎉 🎉

PR #7767, opened 2022-09-21
fix #7766 nimbus(chore): Update integration test ping server for Mobile clients

### LINKED ISSUE #7766: Update integration test ping server to allow for Mobile clients
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2810)

jrbenny35:
Because
* Our current implementation of the Ping server used to capture telemetry is quite specific to Desktop clients

This Commit
* Expands the implementation to allow for both Desktop and Mobile clients to submit pings to the server.

### DIFF
--- a/app/tests/integration/nimbus/utils/ping_server/ping_server.py
+++ b/app/tests/integration/nimbus/utils/ping_server/ping_server.py
@@ -25,10 +25,10 @@ def pings():
 
 
 @app.route(
-    "/submit/telemetry/<uuid>/<type>/<browser>/<version>/<browser_type>/<date>/",
+    "/submit/<path:telemetry>",
     methods=["POST"],
 )
-def submit(uuid, type, browser, version, browser_type, date):
+def submit_desktop(telemetry):
 
     if request.method == "POST":
         request_data = request.get_data()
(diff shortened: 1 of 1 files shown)


yashika-khurana (commented):
Can you help me gain more context about ping and its use of it here, and how it gonna be helpful for mobile applications, I think I am missing context 🤔

jrbenny35:
> Can you help me gain more context about ping and its use of it here, and how it gonna be helpful for mobile applications, I think I am missing context 🤔

Sure thing. So when we do our telemetry integration tests, specifically the tests in [here](https://github.com/mozilla/experimenter/blob/main/app/tests/integration/nimbus/test_mobile_integration.py), we collect telemetry to verify it. The current ```ping_server``` only has a route for desktop pings, this PR will change it to allow for both Desktop and Mobile clients to submit to the same route. I plan to add integration tests for Mobile clients in the future so this just gets the ping server setup for that.

yashika-khurana (approved):
thanks @jrbenny35 for explaining it, it makes more sense to me 🎉

jaredlockhart (approved):
Yep looks good and thank you for the very clear explanation 👍

PR #7774, opened 2022-09-22
fix #7765 feat(nimbus):Type filter

### LINKED ISSUE #7765: Add new filter for experiment/rollouts
We want to filter experiments on the basis that if its experiment or rollouts. So add a new filter TYPE on the landing page with 2 values experiment/rollout

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2808)

yashika-khurana:
Because

* We want to filter according to the type- ` EXPERIMENT` OR `ROLLOUT`

This commit

* Adds a new filter `TYPE` on the landing page which allows seeing results based on choices-` EXPERIMENT/ROLLOUT`
Screenshot of changes:
image
 fix #7765


jaredlockhart (approved):
Yep looks right, works perfectly.  `Type` is a really tricky word to use for something in Python becuase `type` is a reserved word so it's easy to accidentally say `type = ....` and override the builtin, you've avoided it here by saying `types` everywhere, but it's way too easy to mess up by accident.  It's doubly risky here becuase `Type` is also used in GraphQL to refer to the outgoing serialization classes, again avoided here becuase we're calling it `TypeEnum` but a hair away from being tricky.  If we weren't immediately getting rid of GraphQL I'd probably ask you to name it something different but hopefully it'll all be gone before it matters.

Anyways thank you, this is great 🎉 🎉 🎉 🎉

PR #7786, opened 2022-10-01
feat(nimbus): add targeting for background task

Mardak:
r? @jaredlockhart cc @ncalexan  Like #7719 except just `isBackgroundTaskMode` to try to debug enrollment issues… or even potentially launch and do the other targeting via message targeting.

### DIFF
--- a/app/experimenter/targeting/constants.py
+++ b/app/experimenter/targeting/constants.py
@@ -628,6 +628,16 @@ RALLY_CORE_ADDON_USER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+BACKGROUND_TASK_NOTIFICATION = NimbusTargetingConfig(
+    name="Background task notification",
+    slug="Background_task_notification",
+    description="Firefox running a background task",
+    targeting="isBackgroundTaskMode",
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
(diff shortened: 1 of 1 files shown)


app/experimenter/targeting/constants.py
```diff
@@ -628,6 +628,16 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+BACKGROUND_TASK_NOTIFICATION = NimbusTargetingConfig(
+    name="Background task notification",
+    slug="Background_task_notification",
+    description="Firefox running a background task",
+    targeting="isBackgroundTaskMode",
+    desktop_telemetry="",
+    sticky_required=True,
```
yashika-khurana, line 637: Question: Do you need` sticky required` to be `True`?

yashika-khurana (changes requested):

app/experimenter/targeting/constants.py
ncalexan: No.  In fact, it would be fascinating if we witnessed unenrollment from these experiments due to `targeting-mismatch`, i.e., `isBackgroundTaskMode` oscillating.

Mardak: I copied the stickiness of `WINDOWS_10_PLUS_BACKGROUND_TASK_NOTIFICATION_` but doesn't need to be.

yashika-khurana (approved):

PR #7804, opened 2022-10-04
7777/css id overview page

yashika-khurana:
Because

* Some of the sections eg. the `Additional Links section`, don’t have easy-to-access CSS attributes.

This commit

*  Add the` id` parameter which will help aide in our UI integration tests.


jrbenny35 (approved):
😍

PR #7814, opened 2022-10-06
fix #7775 feat(nimbus): Move filter to left panel

### LINKED ISSUE #7775: Add a sidebar and move all buttons/filters to it (same style as detail page)
We have all the filters and buttons on the top of the landing page, we want to make consistent flow and move filters/buttons section to the right hand of the screen

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2749)

yashika-khurana:
Because

* We want to improve the UX experience, we are moving the filters panel to the left side of the screen

This commit

* Moves filter to the left side of the screen, including new hover and responsive effect on the table

Before changes: 
image
After changes:
image
fixes #7775


app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -340,12 +340,15 @@ const DirectoryTable: React.FunctionComponent<DirectoryTableProps> = ({
   }
 
   return (
-    <div
-      className="directory-table pb-2 mt-4"
-      style={{ width: "auto", maxWidth: "100%" }}
-    >
+    <>
```
eliserichards, line 343: Can we take this out?

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -340,12 +340,15 @@ const DirectoryTable: React.FunctionComponent<DirectoryTableProps> = ({
...
       {experiments.length ? (
-        <table className="table" data-testid="DirectoryTable">
+        <Table
+          data-testid="DirectoryTable"
+          size="sm"
+          responsive
+          hover
```
eliserichards, line 349: ❤️

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -340,12 +340,15 @@ const DirectoryTable: React.FunctionComponent<DirectoryTableProps> = ({
...
+          style={{ width: "100%" }}
```
eliserichards, line 350: Does this need double `{}`?

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -365,11 +368,11 @@ const DirectoryTable: React.FunctionComponent<DirectoryTableProps> = ({
               </tr>
             ))}
           </tbody>
-        </table>
+        </Table>
       ) : (
         <p data-testid="no-experiments">No experiments found.</p>
       )}
-    </div>
+    </>
```
eliserichards, line 375: Same as above, can this be removed?

eliserichards (commented):
This looks great so far! I'm going to test it locally now 💯

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
yashika-khurana: interesting, actually no 😆 , this is a fragment, as the div is removed we need to include this

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
yashika-khurana: yup, because - To style an element with the inline style attribute `style={{width:"100%"}}`, the value must be a JavaScript object. In JSX, JavaScript expressions are written inside curly braces, and since JavaScript objects also use curly braces, the styling in the example above is written inside two sets of curly braces `{{}}.`

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
yashika-khurana: it is required to get rid of the additional wrapping div

yashika-khurana:
@eliserichards modified design now :) 
image

app/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
eliserichards: 🤔 🤔 🤔

eliserichards (approved):
I love it! Great job! 🚀

jaredlockhart (changes requested):
Yessssss YESSSSSSS this is so sick @yashika-khurana 🎉 🎉 🎉 🎉 

I think there's something wonky with the buttons though:
Screenshot 2022-10-11 at 10-24-04 Experiments Experimenter

Maybe just need to explicitly wrap each one in a row/col so they're forced to be separated vertically and full width?

jaredlockhart:
Also that's actually a lot of padding around each of the filters, they should be able to fill more of the sidebar?

jaredlockhart:
ALSO

Another way to save space in a design like this, is to put the label inside each drop down as a default option

Screen Shot 2022-10-11 at 10 27 07 AM

But that's just a suggestion.

yashika-khurana:
@jaredlockhart thank you for the suggestions, I have incorporated all the requested changes, and here is the new look
image

eliserichards:
I love it!! It looks amazing. r+ 🚀 

One small concern I have about putting the labels for the filters inside the filters instead of above is that, once you choose filter options, you no longer can tell what the filter is for. I'm not sure if this is an actual problem, since our users seem to be really familiar with the data, but .... 

image

jaredlockhart (approved):
First of all, awesome.  Looks fantastic, thank you @yashika-khurana 🎉 

I have two tiny nits, you can do them or leave them if you wanna move on.

Screen Shot 2022-10-26 at 4 22 17 PM

- `new` should be capitalized
- there's still too much padding around the buttons, it looks like it's a `<Button>` with some explicit margins right inside a `<Row>` but it should be more like
```js
            <Row>
              <Col>
                <Link
                  to="new"
                  data-sb-kind="pages/New"
                  className="btn btn-primary btn-small w-100"
```
then it looks a little more correct like:
Screen Shot 2022-10-26 at 4 27 46 PM
- the padding between the buttons and the filter drop downs isn't aligned, I think if you do what I said above for everything then they should all line up and the padding should be a little more reasonable

Just a small thing, up to you, but looks great, the labels inside the filters look fantastic, all looks great, great job @yashika-khurana !  🎉 🎉 🎉

yashika-khurana:
@jaredlockhart @eliserichards  thank you for reviewing the PRs, @eliserichards that's a really good point but I am not sure what's the best way to show once the dropdown is selected and accept for the labels on the top, may be we can release this and get feedback from the user if they are having difficulty we can add that lables later too. @jaredlockhart made changes as per your suggestions, here is the new look and feel 
image

jaredlockhart:
Almost!

You want to always include a `<Col>` inside a `<Row>` and then you never have to specify margins they're automatic.

Like so:

```js
            <Row>
              <Col>
                <Link
                  to="new"
                  data-sb-kind="pages/New"
                  className="btn btn-primary btn-small mt-2 w-100"
                  id="create-new-button"
                >
                  <CreateNewIcon
                    width="20"
                    height="20"
                    fill="white"
                    dominantBaseline="start"
                    role="img"
                    aria-label="download icon"
                  />
                  <span className="ml-1">Create New </span>
                </Link>
              </Col>
            </Row>
```

See the difference:
Screen Shot 2022-10-26 at 4 58 30 PM

yashika-khurana:
@jaredlockhart sorry I enable auto merged, let me revert the changes and put col inside it

PR #7819, opened 2022-10-07
Update custom targeting for Mozilla Rally users

aaga:
We want to change how the custom targeting works for Mozilla Rally users, and also add a new category of targeting for users with our new extension


eliserichards (commented):
This looks good so far! Looks like `make check` is complaining, so you'll need to fix those issues before we get this merged 👍 

Just to confirm, there will be no use for `RALLY_CORE_ADDON_USER` targeting in the future? We can always keep that config if it might be useful

jrbenny35:
image
Thats 🔥

aaga:
@eliserichards Thanks for your comments! I added back `RALLY_CORE_ADDON_USER` just to be safe like you suggested, and after some tinkering was able to re-frame the targeting string so that `make check` is happy

jaredlockhart (approved):
Awesome looks good to me 👍

PR #7821, opened 2022-10-07
fix #7759 feat(nimbus): parses jetstream analysis results into segments

### LINKED ISSUE #7759: Retrieve and organize segment data from Jetstream
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2804)

mikewilli:
Because

* Jetstream performs segmentation analysis already
* these results would be useful to view in Experimenter

This commit

* parses the Jetstream results and puts them into segments instead of only using the 'all' segment


mikewilli:
Outstanding question: is it OK to modify the existing schema of `experiment_data` in the name of keeping the parsing consistent across segments, or would it be better to special-case the 'all' segment so that existing UI code doesn't break?

Current thought is to put back in (at time of writing, this PR modifies it) the existing schema for now to avoid breaking changes, while also putting 'all' in as a segment (as this PR already does). Once the UI can be updated (to be done next as part of the same Epic), we can then remove the deprecated parts of the schema.

mikewilli:
Responding to my own question above for posterity:

For this PR, I opted to just move forward with the new schema and make minimal UI-side updates to accommodate it. The only pieces this affects are under PageResults section, and my next work will be to flesh it out so it is worth spending more upfront time now moving to the updated schema than spending similar time in implementing two different schemas temporarily.

app/experimenter/jetstream/client.py
```diff
@@ -152,33 +157,40 @@ def get_experiment_data(experiment):
...
-
-        transformed_data = data.dict(exclude_none=True) or None
-
-        experiment_data[window] = transformed_data
+        experiment_data[window] = {}
+        data_from_jetstream = get_data(recipe_slug, window) or []
+        segments = list(set(point["segment"] for point in data_from_jetstream))
+
+        for segment in segments:
+            segment_data = [d for d in data_from_jetstream if d["segment"] == segment]
```
jaredlockhart, line 165: Not that it really matters, the scale of this data isn't really anything to worry about, but just a fun little optimization to think about.  

Here we're looping over the points in `data_from_jetstream` twice, once to get the unique set of segments, and then once again to get the points of each segment type.  You could do that in one pass like

```py
segment_points = defaultdict(list)
for point in data_from_jetstream:
    segment_points[point['segment']].append(point)

for segment, segment_data in segment_points.items():
    ...
```

app/experimenter/jetstream/client.py
```diff
@@ -137,7 +138,11 @@ def get_other_metrics_names_and_map(data, RESULTS_METRICS_MAP):
 def get_experiment_data(experiment):
     recipe_slug = experiment.slug.replace("-", "_")
     windows = ["daily", "weekly", "overall"]
-    raw_data = {}
+    raw_data = {
+        "daily": {},
+        "weekly": {},
+        "overall": {},
```
jaredlockhart, line 144: Strings like this, and references to them everywhere, should generally be moved to constants.  These would probably fit nicely in an enum.

app/experimenter/jetstream/models.py
```diff
@@ -78,6 +78,12 @@ class JetstreamDataPoint(BaseModel):
 
 
 class JetstreamData(BaseModel):
+    """
+    Parameters:
+        __root__: List[JetstreamDataPoint] = []
+            The list should be filtered as needed coming in (e.g., by a given segment).
+    """
```
jaredlockhart, line 85: Do you need this if the property below is already type annotated?  Or just to add the description?

app/experimenter/nimbus-ui/src/components/PageResults/TableMetricCount/index.tsx
```diff
@@ -34,6 +34,7 @@ type TableMetricCountProps = {
   outcomeDefaultName: string;
   group: string;
   metricType?: MetricTypes;
+  segment?: string;
```
jaredlockhart, line 37: In TypeScript there's this awesome thing where you can declare a value as an enum of string constants and then later even if you refer to something as a literal value it'll still validate that it's one of the valid enum choices.  If we know the segment keys here we should declare them explicitly as like an enum or a string union, and then even if you use the literal string value anywhere in the code it'll still know whether it's a valid choice.

jaredlockhart: Oh you know I just realized, segments will be arbitrary keys that come from the data, so we can't do what I just said.  Oh well.

jaredlockhart (approved):
Yep, looks great.  Couple thoughts but nothing at all major.  Great job 👏 👍 💯 Thanks @mikewilli !

app/experimenter/nimbus-ui/src/components/PageResults/TableMetricCount/index.tsx
mikewilli: Yea... I had the exact same thought process and decided that `string` was the best I could do.

app/experimenter/jetstream/models.py
mikewilli: The main thing I wanted to get across is that the list in `__root__` should already be filtered at instantiation time. Previously we were doing some filtering to make sure the data points were `Segment.ALL`, but I removed that and placed the responsibility on the code creating the object. Happy to discuss a better way to accomplish that goal though.

app/experimenter/jetstream/client.py
mikewilli: Perfect, just the kind of easy optimization I like!

PR #7841, opened 2022-10-12
fixes #7840 test(nimbus): Adds tests for outcomes on the Metrics Page.

### LINKED ISSUE #7840: Update Metrics page tests
Currently we don’t do anything on the metrics page. We should add that capability.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2845)

jrbenny35:
Because
* We have been skipping adding anything on the metrics page within the integration tests because it isn't needed to create a valid experiment

This commit
* Adds the capability to set metrics within the integration tests
* Asserts the metrics are set and are not empty.

### DIFF
--- a/app/tests/integration/nimbus/conftest.py
+++ b/app/tests/integration/nimbus/conftest.py
@@ -177,6 +177,8 @@ def default_data(application, experiment_name, load_experiment_outcomes):
             primary_outcomes=[load_experiment_outcomes["firefox_ios"][0]],
             secondary_outcomes=[load_experiment_outcomes["firefox_ios"][1]],
         )
+    else:
+        metrics_data = BaseExperimentMetricsDataClass(primary_outcomes=[],secondary_outcomes=[])
 
     return BaseExperimentDataClass(
         public_name=experiment_name,
@@ -209,7 +211,7 @@ def default_data(application, experiment_name, load_experiment_outcomes):
 
 
 @pytest.fixture
-def create_experiment(base_url, default_data):
+def create_experiment(base_url, application, default_data):
     def _create_experiment(selenium):
         home = HomePage(selenium, base_url).open()
         experiment = home.create_new_button()
@@ -246,10 +248,11 @@ def create_experiment(base_url, default_data):
 
         # Fill Metrics page
         metrics = branches.save_and_continue()
-        metrics.set_primary_outcomes(values=default_data.metrics.primary_outcomes[0])
-        assert metrics.primary_outcomes.text != "", "The primary outcome was not set"
-        metrics.set_secondary_outcomes(values=default_data.metrics.secondary_outcomes[0])
-        assert metrics.secondary_outcomes.text != "", "The seconday outcome was not set"
+        if "focus" not in str(application).lower():
+            metrics.set_primary_outcomes(values=default_data.metrics.primary_outcomes[0])
+            assert metrics.primary_outcomes.text != "", "The primary outcome was not set"
+            metrics.set_secondary_outcomes(values=default_data.metrics.secondary_outcomes[0])
+            assert metrics.secondary_outcomes.text != "", "The seconday outcome was not set"
 
         # Fill Audience page
         audience = metrics.save_and_continue()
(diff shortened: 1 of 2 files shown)


jrbenny35:
Ah just remembered this will fail on Focus applications. I'll have to add a skip

jrbenny35:
image
Example

app/tests/integration/nimbus/conftest.py
```diff
@@ -143,10 +143,45 @@ def experiment_name(request):
...
+    outcomes = {"firefox_desktop": "", "fenix": "", "firefox_ios": ""}
+    base_path = "/code/app/experimenter/outcomes/jetstream-config-main/outcomes"
+
+    for k in list(outcomes):
+        outcomes[k] = [
+            name.split("_")[0].rsplit(".")[0]
+            for name in os.listdir(f"{base_path}/{k}")
+            if "example" not in name
+        ]
+    return outcomes
```
jaredlockhart, line 158: Can't you just discover them from the list in the html?  Like active the drop down and just click wahtever the first option is?

app/tests/integration/nimbus/conftest.py
```diff
@@ -143,10 +143,45 @@ def experiment_name(request):
...
+        )
+    elif "ios" in str(application).lower():
+        metrics_data = BaseExperimentMetricsDataClass(
+            primary_outcomes=[load_experiment_outcomes["firefox_ios"][0]],
+            secondary_outcomes=[load_experiment_outcomes["firefox_ios"][1]],
+        )
+    else:
+        metrics_data = BaseExperimentMetricsDataClass(
+            primary_outcomes=[], secondary_outcomes=[]
+        )
```
jaredlockhart, line 183: This looks like it could be better represented with a dict?

jaredlockhart (changes requested):
Yessss, more UI tests!  Couple thoughts 🙏

app/tests/integration/nimbus/conftest.py
jrbenny35: No. The Select element on the page isn't a "normal" select. It seems to be hidden within some react element ```id="react-select-126-input"```, and all of the CSS is munged on the page as well. I am open to thoughts on this, but I didn't think changing the page was worth the effort.

jaredlockhart: Hrmm yeah that sucks.  Oh well.  We'll just have to do it this way 👍

app/tests/integration/nimbus/conftest.py
jrbenny35: Hmm, I think I can see what you mean. I would still have to do a application check but it would be cleaner. Good idea.

app/tests/integration/nimbus/conftest.py
```diff
@@ -143,10 +143,51 @@ def experiment_name(request):
...
+            secondary_outcomes=[load_experiment_outcomes["firefox_ios"][1]],
+        ),
+    }
+
+    application_str = str(application).lower()
+
+    for _ in list(outcomes):
+        if _ in application_str and "focus" not in application_str:
+            metrics_data = outcomes[_]
+            break
```
jaredlockhart, line 185: Shouldn't this just be

```py
if application_str in outcomes:
    metrics_data = outcomes[application_str]
else:
    ...
```

?

jaredlockhart (changes requested):

app/tests/integration/nimbus/conftest.py
jaredlockhart: Or you could add the empty entry to `focus` in the dict so then it's just

```py
metrics_data = outcomes[application_str]
```

jrbenny35: > Shouldn't this just be
> 
> ```python
> if application_str in outcomes:
>     metrics_data = outcomes[application_str]
> else:
>     ...
> ```
> 
> ?

Ah yes, I added an extra step. Reminds me to not code while sick

jaredlockhart: Brain 🧠 not found 😴  Me toooooooooo 😓

jrbenny35:
I think I've got everything

jaredlockhart (approved):
👏 👏 👏 👏 👏 👏 🙏 🙏 🙏 🙏 🙏 🙏  thnx @jrbenny35 🎉

PR #7842, opened 2022-10-13
fix #7760 feat(nimbus): add segment selector to results page

### LINKED ISSUE #7760: Allow user to group Results page by segmentation
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2805)

mikewilli:
Because

* users can currently configure segments for Jetstream analysis
* there is no way for users to view the results of this analysis in Experimenter

This commit

* adds a segment selector to the Results page

**Screenshots** (note that the results are different when different segments are selected)
segment-ui_0
segment-ui_1


danielkberry (approved):
I can't review the code (or the functionality) but the UI screenshots look great and seem to be what we want.

jaredlockhart (approved):
Thank you, very concise.  You know we had storybooks to be able to visually inspect ui elements outside of the application context, but we weren't using it so I took it out, but this is one case where it would've been helpful for testing.  But looks good thank you! 🎉

PR #7858, opened 2022-10-18
fix #7855 fix(nimbus): ensure branch keys exist before retrieval

### LINKED ISSUE #7855: Check for keys before getting branches from results data
Sentry is reporting an issue getting branch from results: https://sentry.io/organizations/mozilla/issues/3677564899/?query=is%3Aunresolved

This is related to the schema changes made in #7821, and is related to other recent issues #7839, #7847, #7850

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2860)

mikewilli:
Because

* results may not be available when retrieving the control branch from `analysis`
* unavailable results error prevents the Results page from loading

This commit

* checks for valid results before retrieving control branch
* throws and handles an error when a control branch cannot be determined
* adds tests for this situation

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
@@ -61,8 +61,33 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
   const resultsContextValue: ResultsContextType = {
     analysis,
     sortedBranchNames,
-    controlBranchName: sortedBranchNames[0],
+    controlBranchName:
+      sortedBranchNames.length > 0 ? sortedBranchNames[0] : undefined,
   };
+
+  const controlBranchError =
+    sortedBranchNames.length === 0 ? (
+      <AnalysisErrorAlert
+        errors={[
+          {
+            metric: null,
+            message: "Analysis could not be performed",
+            filename: null,
+            exception: null,
+            func_name: null,
+            log_level: null,
+            statistic: null,
+            timestamp: null,
+            experiment: null,
+            exception_type: null,
+          },
+        ]}
+      />
+    ) : (
+      false
+    );
+
+  console.log(controlBranchError);
 
   const { external_config: externalConfig } = analysis.metadata || {};
 
@@ -221,6 +246,7 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
 
         <div>
           <h2 className="h4 mb-3">Outcome Metrics</h2>
+          {controlBranchError && controlBranchError}
           {analysis.overall?.[selectedSegment] &&
           Object.keys(analysis.overall?.[selectedSegment]).length > 0
             ? experiment.primaryOutcomes?.map((slug) => {
--- a/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
@@ -187,38 +187,42 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
           <b>Hypothesis</b>: {experiment.hypothesis}
         </p>
 
-        {analysis.overall?.[selectedSegment] && (
-          <TableWithTabComparison
-            {...{ experiment }}
-            Table={TableHighlights}
-            className="mb-2 border-top-0"
-            segment={selectedSegment}
-          />
-        )}
+        {analysis.overall?.[selectedSegment] &&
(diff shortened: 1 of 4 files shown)


app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -38,22 +38,41 @@ export const getControlBranchName = (analysis: AnalysisData) => {
...
+      if (branchData.is_control) {
+        return branchName;
+      }
+    }
+  }
+  // last option - try to find a unique branch name in the daily results
+  const daily = analysis.daily?.all || [];
+  if (daily.length > 0) {
+    const branches = new Set();
+    daily.forEach((point) => branches.add(point.branch));
```
jeddai, line 53:
```suggestion
    const branches = new Set(daily.map((point) => point.branch));
```
Not a necessary change, but you could swap this to construct the set with an array of the branches on the daily points.

app/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -38,22 +38,40 @@ export const getControlBranchName = (analysis: AnalysisData) => {
...
+    const branches = new Set(daily.map((point) => point.branch));
+    if (branches.size === 1) {
+      return branches.values().next().value; // return the first and only value
     }
   }
+
+  throw new Error(
+    "Invalid argument 'analysis': no branch name could be found in the results.",
+  );
 };
```
jaredlockhart, line 61: Yeah this seems reasonable.  If we can't find any control branch then we can't really show any meaningful analysis so probably in that case all we could do is show a big red error that we can't analyze this experiment and to contact an admin?

jaredlockhart (approved):
Yeah this seems reasonable for now, how does it show in the UI if no control branch is found?

mikewilli:
@jaredlockhart If it doesn't find a control branch, it basically just shows an empty looking Results page. If Jetstream logged any errors when trying to analyze then those would display, but as of now I have not added any explicit errors. However, after you asked, I added a quick error to show what that would look like. See the screenshot below -- right now the page looks the same, but without the bottom error. The bottom error is what I added as an example.

image

jaredlockhart:
Yep looks good ty.  Looks like CI is still being cranky, we're looking at it over here but in the mean time I disabled the two failing checks so you should be able to land this.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -61,9 +61,34 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
...
+      sortedBranchNames.length > 0 ? sortedBranchNames[0] : undefined,
   };
 
+  const controlBranchError =
+    sortedBranchNames.length === 0 ? (
+      <AnalysisErrorAlert
+        errors={[
+          {
+            metric: null,
+            message: "Analysis could not be performed",
```
jaredlockhart, line 74: Maybe be a little more specific here about the problem being that no control branch was found?

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -221,6 +246,7 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
 
         <div>
           <h2 className="h4 mb-3">Outcome Metrics</h2>
+          {controlBranchError && controlBranchError}
```
jaredlockhart, line 249: Is this a trick to show if it's not null or hide it otherwise?

mikewilli: Yes. Might be a better way, but I tested and don't think there are any issues with doing it this way.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
mikewilli: Oh yea, I meant to update that, thanks for the reminder! Updated in new commit.

app/experimenter/nimbus-ui/src/components/PageResults/index.tsx
jeddai: I feel like I remember cases where these evaluate to `null` or `undefined` will prevent them from being displayed, I may be wrong on that though. Not sure that this is a more clear solution though if it does work

```suggestion
          {controlBranchError || null}
```

mikewilli: Yep, this seems right to me too, I knew there was probably a better way thanks all.

PR #7859, opened 2022-10-18
Fixes #7698 feat(nimbus): Add dirty publish state for rollouts

### LINKED ISSUE #7698: Add dirty publish state for rollouts
See [https://mozilla-hub.atlassian.net/browse/EXP-2562](https://mozilla-hub.atlassian.net/browse/EXP-2562|smart-link)for state diagram.

Once a rollout has been **published** (live), we need to define the state for whether is has been edited or not. We will add a new publish status, `dirty`, which will sit in between the `idle`/ `review` publish states.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2563)

eliserichards:
#7698 

Because..

* We need to differentiate the state what a rollout is Live and changes have been made

This commit...

* Adds a new `PublishStatus` option, `PublishStatus.DIRTY`
* Updates the status transitions and the lifecycle states
* Adds lots of tests for the status transitions :)

image


app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -968,7 +968,7 @@ def test_cant_review_for_requesting_user(self):
         self.assertFalse(serializer.is_valid(), serializer.errors)
         self.assertIn("publish_status", serializer.errors)
 
-    def test_can_review_for_requesting_user_when_idle(self):
```
eliserichards, line 971: Was this added so that we can manually approve? Should this still be allowed?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -56,3 +60,308 @@ def test_update_experiment_with_invalid_publish_status_error(self):
         )
         self.assertFalse(serializer.is_valid())
         self.assertIn("experiment", serializer.errors)
+
+    @parameterized.expand(
```
eliserichards, line 64: These are all the same test contents, but I broke it out into a few different ones by the `Lifecycle` it was testing because it was easier to debug. I'd love thoughts on whether I should make it one (huge) test 🤔

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
yashika-khurana: I am not sure about this 🤔

yashika-khurana: if they still uses manually approve thing

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
yashika-khurana: I am happy with the split tests as the test name focuses on functionality, but I don't mind having huge test too, up to you

jaredlockhart:
Let's block this on https://github.com/mozilla/experimenter/pull/7878

app/experimenter/experiments/constants.py
```diff
@@ -177,7 +178,11 @@ class Type(models.TextChoices):
         Status.DRAFT: (Status.PREVIEW,),
         Status.PREVIEW: (Status.DRAFT,),
     }
-    STATUS_ALLOWS_UPDATE = (Status.DRAFT,)
+
+    STATUS_ALLOWS_UPDATE = (
+        Status.DRAFT,
+        Status.LIVE,
+    )
```
jaredlockhart, line 185: Hrmmm interesting.  We want to allow edits for live rollouts but not live experiments, I wonder if we'll have to expand this logic to be more specific?  Can figure it out later as we get more into the weeds with it, just something to think about.

app/experimenter/experiments/constants.py
```diff
@@ -187,13 +192,30 @@ class Type(models.TextChoices):
...
 
     VALID_PUBLISH_STATUS_TRANSITIONS = {
-        PublishStatus.IDLE: (PublishStatus.REVIEW, PublishStatus.APPROVED),
+        PublishStatus.IDLE: (
+            PublishStatus.REVIEW,
+            PublishStatus.DIRTY,
+        ),
+        PublishStatus.DIRTY: (
+            PublishStatus.REVIEW,
+            PublishStatus.DIRTY,
```
jaredlockhart, line 201: Do we need to have Dirty > Dirty here?  If so why isn't there Idle > Idle above?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -596,7 +596,7 @@ def test_status_with_invalid_target_status_next(self):
         )
 
     def test_status_restriction(self):
-        experiment = NimbusExperimentFactory(status=NimbusExperiment.Status.LIVE)
+        experiment = NimbusExperimentFactory(status=NimbusExperiment.Status.COMPLETE)
```
jaredlockhart, line 599: This should probably be parameterized for all the statuses that don't allow updates.  Might as well do it now while we're futzing with these things.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
jaredlockhart: Oh interesting yeah I don't think that workflow exists anymore, I don't see any reason to keep that.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -968,7 +968,7 @@ def test_cant_review_for_requesting_user(self):
         self.assertFalse(serializer.is_valid(), serializer.errors)
         self.assertIn("publish_status", serializer.errors)
 
-    def test_can_review_for_requesting_user_when_idle(self):
+    def test_cannot_approve_for_requesting_user_when_idle(self):
```
jaredlockhart, line 971: It's not even really that, it's that Idle > Approved isn't even an allowed transition.  I'd nuke this particular test altogether and just make sure the allowed transition tests are parameterized to cover all the transitions.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -25,7 +26,8 @@ def test_update_experiment_publish_status_while_in_preview(self):
             },
             context={"user": self.user},
         )
-        self.assertTrue(serializer.is_valid())
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("publish_status", serializer.errors)
```
jaredlockhart, line 30: Interesting this is another one of those cases where we allowed a transition that doesn't actually make sense.  There is no publish flow for moving into or out of preview so you shouldn't be able to update publish status while in preview.  Good catch 👍

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -56,3 +58,308 @@ def test_update_experiment_with_invalid_publish_status_error(self):
...
+                False,
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.CREATED,
+                NimbusExperiment.PublishStatus.WAITING,
+                False,
+            ],
+        ]
+    )
+    def test_update_publish_status_errors_for_create(self, lifecycle, status, valid):
```
jaredlockhart, line 91: I would call this parameter `publish_status` just to make it really clear it's not a `status` which is a different thing.  When I added publish_status I wasn't sure if I should rename `status` to something that also has a qualifier like `lifecycle_status` or something, I just left it alone cuz we'd have to refactor basically the whole app.  But yeah generally I try to name parameters like this after the field they're referring to on the model.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -56,3 +58,308 @@ def test_update_experiment_with_invalid_publish_status_error(self):
...
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.CREATED,
+                NimbusExperiment.PublishStatus.APPROVED,
+                False,
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.CREATED,
+                NimbusExperiment.PublishStatus.WAITING,
+                False,
```
jaredlockhart, line 87: If they're all `CREATED` then it doesn't need to be a parameter at all.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -56,3 +58,308 @@ def test_update_experiment_with_invalid_publish_status_error(self):
...
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.PREVIEW,
+                NimbusExperiment.PublishStatus.APPROVED,
+                False,
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.PREVIEW,
+                NimbusExperiment.PublishStatus.WAITING,
+                False,
```
jaredlockhart, line 130: Same here for PREVIEW

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -56,3 +58,308 @@ def test_update_experiment_with_invalid_publish_status_error(self):
...
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertEquals(serializer.is_valid(), valid)
+
+    @parameterized.expand(
+        [
+            [
+                NimbusExperimentFactory.Lifecycles.LIVE_IDLE,
```
jaredlockhart, line 151: Don't need LIVE_IDLE here

jaredlockhart: Oh there's more, yeah if it's the same in all of them then it doesn't need to be a parameter.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -56,3 +58,308 @@ def test_update_experiment_with_invalid_publish_status_error(self):
...
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertEquals(serializer.is_valid(), valid)
+
+    @parameterized.expand(
+        [
+            [
+                NimbusExperimentFactory.Lifecycles.PREVIEW,
```
jaredlockhart, line 108: The other thing I'd suggest here is, using the lifecycles is a nice shortcut for when you want something to be in a particular set of statuses, but since we're explicitly testing the combinations of statuses here, the lifecycles are actually hiding which pair of (status, publish_status) we're really talking about, so maybe it would be clearer not to use the lifecycles at all and just explicitly refer to the pair of (status, publish_status)

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
jaredlockhart: Hrmmmm.....  So yes I like this, and if we test the combinations of statuses here to stress test the allowed status/publish_status transitions, then we don't need more than one test cover that in the actual serializer test above, just a single one to test integration is sufficient and then we can rigorously test all the combinations here and then we're covered.

As far as whether to make these separate or make a single test with a giant list of all the combinations.....  I feel like I could go either way on it.  On the one hand, the test body is the same in all of them so there's no real reason to make them separate, but then the parameter list gets MASSIVE.  Is a massive list of status combinations okay?  Could be?  Makes it very clear which combinations are allowed and which aren't.  I dunno, I could go either way on it.  

If you keep them separate then I would take `status` out of the parameter list since then it's redundant.  And I would use `status` directly instead of bothering with the lifecycles since then it's really clear what exact combination is being talked about.

app/experimenter/experiments/tests/factories.py
```diff
@@ -178,6 +208,13 @@ class Lifecycles(Enum):
     LAUNCH_APPROVE_TIMEOUT = LAUNCH_APPROVE_WAITING + (LifecycleStates.DRAFT_REVIEW,)
     LIVE_ENROLLING = LAUNCH_APPROVE_APPROVE + (LifecycleStates.LIVE_IDLE_ENROLLING,)
     LIVE_PAUSED = LIVE_ENROLLING + (LifecycleStates.LIVE_IDLE_PAUSED,)
+
+    LIVE_IDLE = (LifecycleStates.LIVE_IDLE,)  # why didn't we have this one before?
+    LIVE_DIRTY = (LifecycleStates.LIVE_DIRTY,)
+    LIVE_REVIEW_REQUESTED = (LifecycleStates.LIVE_REVIEW,)
+    LIVE_REVIEW_APPROVED = (LifecycleStates.LIVE_APPROVED,)
+    LIVE_REVIEW_WAITING = (LifecycleStates.LIVE_WAITING,)
```
jaredlockhart, line 216: Oh this isn't exactly how the lifecycles work.  This would create an experiment that goes straight from nothing straight to each of these states without passing through any of the states before it.  Like it just popped into existence like a virtual quantum particle.  That's fine for the quantum vacuum of outer space, but Experimenter is not so stochastic.  Look above at how they each take one of the previous states and then tack on another, that's what we'd want here.

Also a plain `LIVE_IDLE` is the `LIVE_ENROLLING` one above, since there's two cases where an experiment would be `LIVE_IDLE`, with `is_paused=False` and `is_paused=True`.  Also just looking at it here I realize the `LIVE_PAUSED` one should have these `LIVE_REVIEW_....` ones in between `LIVE_ENROLLING` and `LIVE_PAUSED` since you have to go through the full review flow (minus DIRTY) to pause it.  Wouldn't hurt to update that while you're adding the other review states for live.

jaredlockhart (changes requested):
Generally looking good!  Good updates to statuses, catching some invalid cases and transitions we should definitely get rid of, adding some nice exhaustive tests to all the expected transitions, I like it.  Having the diagrams up to date and correct makes reviewing this much much nicer 🙏  Good going @eliserichards !

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
jaredlockhart: Oh on second thought re: below, we can be exhaustive in the mixin tests and then we only need a single test in the serializer to test integration.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
eliserichards: When I was writing these, it was really helpful to keep them split up to debug them. With the parameterized tests, it just fails everything and you have to test each parameter line individually to see which was the failing line. 

I like the idea of using `status` directly -- that makes it much more clear

app/experimenter/experiments/tests/factories.py
eliserichards: 👩‍🚀 that makes so much more sense. I was a little confused with the additions here and why they were defined like that

eliserichards:
Closing for a ✨ fresh ✨ PR

PR #7865, opened 2022-10-19
fix #7864 bug(project): freeze remote settings docker image

### LINKED ISSUE #7864: Internal server error in Remote Settings in CI
Looks like we're getting a bunch of CI failures coming from Remote Settings.  Let's try the previous tag and see if that clears it up.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2863)

jaredlockhart:
Because

* A regression was recently introduced into the latest remote settings docker image
* This is breaking our CI runs

This commit

* Freezes the remote settings docker image to the last working tag

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -86,7 +86,7 @@ services:
       - "6379:6379"
 
   kinto:
-    image: mozilla/remote-settings:latest
+    image: mozilla/remote-settings:29.1.1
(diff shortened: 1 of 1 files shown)


jaredlockhart:
This doesn't seem to be the solution.

Issue #7479, opened 2022-06-30
Define state changes for publishing rollout updates


data-sync-user:
Define state machine for 1. rollouts integrated into experiments and 2. rollouts separate

See [https://github.com/mozilla/experimenter/tree/main/app/experimenter/kinto](https://github.com/mozilla/experimenter/tree/main/app/experimenter/kinto|smart-link) for current diagrams.

The new diagrams should be added to readme’s on experimenter and [experimenter.info](http://experimenter.info) once they are approved.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2562)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/29546">rollout_state_v2.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/29294">rollouts_state.png</a>

data-sync-user:
➤ Elise Richards commented:

!rollouts_state.png|width=961,height=231!

Per our discussion during the workweek, this is the proposed state diagram for rollouts. 

Rollouts may have the following states (highlighted in bold in the diagram above):

  * Draft
    * Feature: editable
    * Population: editable

  * Preview
    * Feature: frozen
    * Population: frozen

  * Published:
    * Feature: frozen
    * Population: frozen

  * Unpublished
    * Feature: frozen
    * Population: editable

When a rollout is in a draft state, it must be reviewed before it can be published. In addition, when a change is made to a published rollout, it must be reviewed again before it can be re-published.

An unpublished state can only exist after the rollout has moved from draft → published (_clean_) → published (_dirty_) → unpublished (dirty). Once a published rollout is edited and becomes dirty, it is not considered unpublished until those changes are “saved” and submitted for review.

data-sync-user:
➤ Benjamin Forehand Jr. commented:

A rollout can go from Draft to Published with a review? What would make it unpublished?

data-sync-user:
➤ Elise Richards commented:

One question that I have is should we require a review before we “turn off” the rollout (going from published to unpublished)?

data-sync-user:
➤ Elise Richards commented:

Updated version (hopefully this is easier to read):

!rollout_state_v2.png|width=722,height=549!

data-sync-user:
➤ Elise Richards commented:

Final state diagram (for both rollouts and experiments) to be put in the docs:

!final_no-worker.svg|thumbnail!

PR #7867, opened 2022-10-24
For #7479 docs(nimbus): Add new state diagram to kinto readme

eliserichards:
#7479

Because...

* We are going to be fleshing out the "request update" section of the experimenter state machine with rollouts being added

This commit...

* Adds an updated version of the state diagram and descriptions to the `app/experimenter/kinto/` folder
* Adds .xml for this state diagram so that we can easily change/edit it in the future (we used the drawio app on confluence)
* Note: another PR will add the sequence diagrams (#7878)

--------
states-v2


jaredlockhart (approved):
Yes this all looks right to me, thank you for updating!  Is there some way to export the diagram in an editable form like some xml thing or something so we can re edit it later if we need to?

yashika-khurana:
@eliserichards thank you it looks good, just one question do we need one arrow between remote settings and experiment publish state?

eliserichards:
> 

Yes! Good point, I forgot to add the `XXX.sequence` to this PR

eliserichards:
> @eliserichards thank you it looks good, just one question do we need one arrow between remote settings and experiment publish state?

I see what you mean, maybe a dotted line like we have above 🤔

PR #7871, opened 2022-10-24
fix #7879 bug(project): update remote settings configuration

### LINKED ISSUE #7879: Update docs with remote settings state specifics
We don't have much documentation about the [remote settings states](https://github.com/mozilla/experimenter/blob/main/app/experimenter/kinto/client.py#L4-L7) and how they impact our stuff. Maybe add this in the [state diagrams section](https://github.com/mozilla/experimenter/tree/main/app/experimenter/kinto#remote-settings-kinto-integration)?  

Related to #7479

jaredlockhart:
Because

* Remote Settings 30.0.0 introduced breaking configuration changes around removing sentry
* This was causing RS to fail to start locally/in CI

This commit

* Updates our Remote Settings configuration with the 30.0.0 changes described here https://github.com/mozilla/remote-settings/blob/main/CHANGELOG.rst#3000-2022-10-17

### DIFF
--- a/kinto/server.ini
+++ b/kinto/server.ini
@@ -135,7 +135,7 @@ plugin = dogstatsd
 keys = root, kinto
 
 [handlers]
-keys = console, sentry
+keys = console
 
 [formatters]
 keys = color, json, generic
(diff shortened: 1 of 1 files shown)


yashika-khurana (approved):
oh totally makes sense now, I am wondering how could we prevent this in the future?

jaredlockhart:
Yeah it's a good question, I don't think there's a good way to prevent it, but at least now we know if we see breaking changes in RS we can check the release notes for first clues.

PR #7876, opened 2022-10-25
fix #7809 feat(nimbus): add schema validation opt out

### LINKED ISSUE #7809: Add a field for opting out of schema validation on Desktop
There have been several occurrences of validation logic and schema issues with experiments on Desktop, so it would be nice to have a way to disable schema validation in cases where experiments need to launch and updated validation logic or schema cannot ride the train.

This would be an admin-console only field in.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2832)

jaredlockhart:
Because

* We want a way to disable client side feature schema validation for individual experiments/rollouts for cases where the client side schema may contain errors even if the payload is valid

This commit

* Adds a boolean field called is_client_schema_disabled to the NimbusExperiment model which is configurable via the admin
* Serializes that field in the V6 serializer as featureValidationOptOut


app/experimenter/experiments/tests/test_changelog_utils.py
```diff
@@ -35,24 +35,25 @@ def test_outputs_expected_schema_for_empty_experiment(self):
...
                 "channel": NimbusExperiment.Channel.NO_CHANNEL,
                 "conclusion_recommendation": None,
                 "countries": [],
                 "feature_configs": [],
-                "firefox_min_version": NimbusExperiment.Version.NO_VERSION,
                 "firefox_max_version": NimbusExperiment.Version.NO_VERSION,
+                "firefox_min_version": NimbusExperiment.Version.NO_VERSION,
                 "hypothesis": NimbusExperiment.HYPOTHESIS_DEFAULT,
                 "is_archived": experiment.is_archived,
+                "is_client_schema_disabled": experiment.is_client_schema_disabled,
```
yashika-khurana, line 50: by default it takes the` False` value, I think we should add test case with `True` value too, what do you think @jaredlockhart 🤔

app/experimenter/experiments/tests/test_changelog_utils.py
```diff
@@ -142,8 +144,8 @@ def test_outputs_expected_schema_for_complete_experiment(self):
                 "risk_revenue": experiment.risk_revenue,
                 "secondary_outcomes": [secondary_outcome],
                 "slug": experiment.slug,
-                "status": experiment.status,
                 "status_next": experiment.status_next,
+                "status": experiment.status,
```
yashika-khurana, line 148: another question did you manually changed the sequence of the variable or with formatter? 🤔

yashika-khurana (changes requested):
Overall looks good and makes sense to disable validation on the client side and apply experiments/rollouts if the payload is valid ❤️ , just some minor suggestion to add test case with true value

app/experimenter/experiments/tests/test_changelog_utils.py
jaredlockhart: Oh yeah I understand your point but that gets huge really fast if you want to try to do the combinatoric values of every field.  We trust that the serializer will do its job and that the field is declared as a `BooleanField` so no I don't think it's necessary.

app/experimenter/experiments/tests/test_changelog_utils.py
jaredlockhart: Oh yeah I sorted the list alphabetically.

PR #7878, opened 2022-10-25
For #7479 docs(nimbus): Updating sequence diagrams

eliserichards:
#7479 
Related to #7867

This commit...

* Updates the sequence diagrams to include rollouts update/dirty state
* Shows the state, published state, and remote settings state more clearly for all published states
* Since rollouts doesn't have an "end"/"completed" state, those diagrams are just updated to the new format of "status: XXX, publish status: XXX, next status: XXX"

----
Notes:
* We may update these further in the future if we implement a Complete state for rollouts
* We also want to use Mermaid (https://github.com/mozilla/experimenter/issues/7956) for any future changes instead of doing it manually like this


jaredlockhart (changes requested):
Okay!  Generally looking good and right.  Two things

1)  Can you pls reorganize it so it goes
- All Draft > Preview/Live flows
- All Live > Live flows
- All Live > Complete flows

Right now the rollout update is in the middle of the draft > live ones

2) I noticed you kept the pause ones intact still referring to `is_paused` as a special status rather than just treating live experiment pausing as a case of Live/dirty update.  Do you want to still keep those separate or try to fold them together?  We could call it outside of the scope of this and port pausing over to dirty later, I'm fine with either.

eliserichards:
Yeah, wrt the pause ones I want to take that out in a separate PR (filing that issue now... #...) 

Let me move the rollouts down to the Live section ✍️

app/experimenter/kinto/README.md
```diff
@@ -127,35 +127,54 @@ The following actors are involved:
...
 - Remote Settings Backend: The backend server that hosts the Remote Settings API and database and is only accessible via allow listed VPN access
 
+### Draft (Create)
+
+A new experiment or rollout which has yet to be sent for review or put into preview is marked for Draft.
+![draft](diagrams/create_draft_idle.png)
+
 ### Preview
 
 A draft experiment that has been validly completed is marked for Preview, is published to the preview collection in Remote Settings, and is then accessible to specially configured clients.
```
yashika-khurana, line 137: draft experiment/rollout

yashika-khurana (changes requested):
overall looks awesome @eliserichards thank you 🎉 , the only thing I feel like in some place is missing is `experiment/rollout`, it just says `experiment`, if you can give a quick pass over that.

eliserichards:
I went through and added more specificity when referring to experiments and/or rollouts :) Hopefully this makes it a little clearer 😄

jaredlockhart (dismissed):
It's coming along really nicely!  But there's a formatting issue here.

The original formatting I used was

```
lifecycle state/publish state(next lifecycle state)
```

which is mentioned [here](https://github.com/mozilla/experimenter/tree/main/app/experimenter/kinto#workflows)

Now some of the diagrams use this formatting:

Screen Shot 2022-10-31 at 4 32 46 PM

Which is very nice and clear, and I like it.

But then others use:

Screen Shot 2022-10-31 at 4 33 08 PM

Which kind of mixes them together, but `(published)` is not a lifecycle status?

I'd say either stick with the original formatting everywhere, or if you want to use the newer one (which I like!) then apply it everywhere pls 🙏

jaredlockhart (dismissed):
It looks like most of the status references are missing the next status?

Screen Shot 2022-11-09 at 12 26 33 PM

The old format was `status/publish_status(status_next)`

You need all three everywhere.

jaredlockhart:
There's still a bunch of incorrect states and diagrams in here, but it looks like we can [use Mermaid 🧜‍♀️](https://github.com/mozilla/experimenter/blob/1b7a00f3043b187665a493422348aa812350bf71/app/experimenter/kinto/README.md) in the docs so we don't need to bother with this whole png screenshot situation.  So let's close this in favour of https://github.com/mozilla/experimenter/issues/7956

eliserichards:
Since there were only a few things that needed to be fixed on these, we're going to get these landed now and work on 🧜‍♀️ when we have some more time

jaredlockhart (dismissed):
Okay almost done!  Couple things

- Under Update(Approve/Approve) it ends on Live/Idle/Live but I think idle should always have `next_status: <none>` right?
- Now that I think about it, Dirty is kind of like an Idle state, since it hasn't really kicked off any kind of publish flow yet until it hits review.  So I see a bunch of places where the first thing it does is go `Live/Idle/None` to `Live/Dirty/Live`, but if we think of Dirty more like an idle state, then it should be `Live/Dirty/None`?  What do you think?  

Actually I see in one of the reject flows it does land on Live/Dirty/None.  So we should pick whether the 'post user edit, pre review request' state is `Live/Dirty/Live` or `Live/Dirty/None`, I think I prefer the latter, what do you think?

eliserichards:
> 



> Okay almost done! Couple things
> 
>     * Under Update(Approve/Approve) it ends on Live/Idle/Live but I think idle should always have `next_status: <none>` right?

Yes, agree with that

 
>     * Now that I think about it, Dirty is kind of like an Idle state, since it hasn't really kicked off any kind of publish flow yet until it hits review.  So I see a bunch of places where the first thing it does is go `Live/Idle/None` to `Live/Dirty/Live`, but if we think of Dirty more like an idle state, then it should be `Live/Dirty/None`?  What do you think?

User publishes a rollout -> `Live/Idle/None`
User changes population size -> `Live/Dirty/None`
User clicks review -> `Live/Review/Live`

Yeah I think that makes sense to me. There will never be a state where it is `Live/Dirty/Live`

 
> Actually I see in one of the reject flows it does land on Live/Dirty/None. So we should pick whether the 'post user edit, pre review request' state is `Live/Dirty/Live` or `Live/Dirty/None`, I think I prefer the latter, what do you think?

I totally agree, `Live/Dirty/None` makes the most sense to me there

eliserichards:
whoops, didn't mean to close that

jaredlockhart (dismissed):
- Under `End Enrollment (Approve/Timeout)` it ends with

Screen Shot 2022-11-18 at 12 42 56 PM

But that should be Live/Review/Live

jaredlockhart (approved):
Alright.  Everything looks right to me 🙏 Coming in for a landing! ✈️

PR #7884, opened 2022-10-27
feat(nimbus): Add exclusionary targeting for pref-setting features

brennie:
When launching an experiment that uses a feature that sets prefs, there is now the option to exclude any users that set the prefs set by the feature. This is enabled via a checkbox in the branches page. The checkbox only shows up when one of the features sets prefs.


app/experimenter/experiments/models.py
```diff
@@ -309,6 +312,21 @@ def _get_targeting_max_version(self):
...
         return expressions
 
+    def _get_targeting_pref_conflicts(self):
+        prefs = []
+
+        if self.prevent_pref_conflicts:
+            app_features = {
+                feature.slug: feature
+                for feature in Features.by_application(self.application)
+            }
```
jaredlockhart, line 322: Can't you use the `find` method you added below for this?

app/experimenter/experiments/api/v5/types.py
```diff
@@ -129,10 +130,17 @@ class NimbusExperimentTargetingConfigType(graphene.ObjectType):
 class NimbusFeatureConfigType(DjangoObjectType):
     id = graphene.Int()
     application = NimbusExperimentApplicationEnum()
+    sets_prefs = graphene.Boolean()
 
     class Meta:
         model = NimbusFeatureConfig
 
+    def resolve_sets_prefs(root, info):
+        if feature := Features.find(root.application, root.slug):
```
jaredlockhart, line 139: image

app/experimenter/experiments/api/v5/types.py
```diff
@@ -129,10 +130,17 @@ class NimbusExperimentTargetingConfigType(graphene.ObjectType):
...
+            return bool(feature.sets_prefs)
```
jaredlockhart, line 140: Is this not already a bool?  Or you mean it could be `None`?

app/experimenter/experiments/models.py
```diff
@@ -202,6 +203,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
     _start_date = models.DateField(blank=True, null=True)
     _end_date = models.DateField(blank=True, null=True)
 
+    prevent_pref_conflicts = models.BooleanField(blank=True, null=True)
```
jaredlockhart, line 206: Why not default `False`?

app/experimenter/experiments/models.py
```diff
@@ -368,6 +386,12 @@ def targeting(self):
         else:
             expressions.extend(sticky_expressions)
 
+        if prefs := self._get_targeting_pref_conflicts():
```
jaredlockhart, line 389: image

app/experimenter/experiments/models.py
```diff
@@ -368,6 +386,12 @@ def targeting(self):
...
+            expressions.extend(
+                f"'!({pref}'|preferenceIsUserSet)"
```
jaredlockhart, line 391: 👍

app/experimenter/features/__init__.py
```diff
@@ -128,6 +136,13 @@ def all(cls):
     def by_application(cls, application):
         return [f for f in cls.all() if f.applicationSlug == application]
 
+    @classmethod
+    def find(cls, application, slug):
+        return next(
+            (f for f in cls.by_application(application) if f.slug == slug),
+            None
+        )
```
jaredlockhart, line 144: Oheck, the `next` pattern.  I never do this but I've seen people do it.  I guess it's nicer than `[t for t in things if t.whatever == something][0]` cuz it falls back to the default so you don't hav to do a check on two lines or get an IndexError.  Maybe I should do this more.

app/experimenter/experiments/models.py
```diff
@@ -309,6 +312,21 @@ def _get_targeting_max_version(self):
...
+
+        if self.prevent_pref_conflicts:
+            app_features = {
+                feature.slug: feature
+                for feature in Features.by_application(self.application)
+            }
+
+            for config in self.feature_configs.all():
+                for pref in app_features[config.slug].sets_prefs:
+                    prefs.append(pref)
```
jaredlockhart, line 326: Could move this right into the Features object 🤷‍♀️  up to you

app/experimenter/experiments/models.py
jaredlockhart: Wait I think something's off here.

```
('!(nimbus.testing.testSetString'|preferenceIsUserSet))
```

I think this should be

```py
f"!('{pref}'|preferenceIsUserSet)"
```

jaredlockhart (changes requested):
Okay!!  Tested locally, all works perfectly, code all looks good.  I think there's a formatting error in the JEXL expression but other than that I think it's all good to go.  One thing is making sure that generated JEXL gets tested in the desktop integration tests, but we can do that as a followup.  Thanks @brennie g/j 🙏

app/experimenter/experiments/api/v5/types.py
brennie: `feature.set_prefs` is a list of strings, which we don't want to return becuase graphql is expecting bool

app/experimenter/experiments/models.py
brennie: many find() calls vs 1 dict construction

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -192,6 +194,7 @@ def _get_feature_configs(self):
                 application=NimbusExperiment.Application(f.application).name,
                 ownerEmail=f.owner_email,
                 schema=f.schema,
+                setsPrefs=bool(Features.find(f.application, f.slug).sets_prefs),
```
jaredlockhart, line 197: You can get rid of this

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -192,6 +193,7 @@ def _get_feature_configs(self):
                 application=NimbusExperiment.Application(f.application).name,
                 ownerEmail=f.owner_email,
                 schema=f.schema,
+                setsPrefs=len(f.sets_prefs) > 0 if f.sets_prefs is not None else False,
```
jaredlockhart, line 196: I think you can just

```py
>>> bool(None)
False
>>> bool([])
False
>>> bool([1, 2, 3])
True
```

app/experimenter/experiments/migrations/0221_nimbusexperiment_prevent_pref_conflicts.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.16 on 2022-11-01 00:39
```
jaredlockhart, line 1: Oh yeah fold these two together, you can just delete them and rerun `make makemigrations` and it'll smush them together.

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1251,6 +1251,37 @@ def test_branches(self):
                 experiment_data["treatmentBranches"],
             )
 
+    @parameterized.expand([[True], [False]])
```
jaredlockhart, line 1254: Funny I just had this conversation with @yashika-khurana the other day, I tend to use parameterized for cases where different states will trigger different code paths.  Here we're just always spitting out verbatim the value, it's declared as a `boolean` field type which will enforce values and so on, so I wouldn't consider it necessary to parametrically test both values.

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1251,6 +1251,37 @@ def test_branches(self):
...
+    def test_prevent_pref_conflicts(self, prevent_pref_conflicts):
```
jaredlockhart, line 1255: I like to use test names that include like, the action that it's doing, so like `test_query_prevent_pref_conflicts`, but since this is just looking up the value of the field, you could just fold that into another test that's already querying the rest of the experiment?

app/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -1414,6 +1446,11 @@ def assertChoices(data, text_choices):
                     ).name,
                     "ownerEmail": feature_config.owner_email,
                     "schema": feature_config.schema,
+                    "setsPrefs": (
+                        len(feature_config.sets_prefs) > 0
+                        if feature_config.sets_prefs is not None
+                        else False
```
jaredlockhart, line 1452: Same here with `bool(...)`?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_configuration_serializer.py
```diff
@@ -15,7 +15,15 @@
...
     def test_expected_output(self):
-        feature_configs = NimbusFeatureConfigFactory.create_batch(10)
+        feature_configs = NimbusFeatureConfigFactory.create_batch(8)
+        feature_configs.append(
+            NimbusFeatureConfigFactory.create(sets_prefs=["foo.bar.baz"])
+        )
+        feature_configs.append(
+            NimbusFeatureConfigFactory.create(
+                sets_prefs=["qux.quux.corge", "grault.garply.waldo"]
+            )
```
jaredlockhart, line 25: Nice.

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -356,6 +356,45 @@ def test_saves_branches_single_feature(self):
         self.assertTrue(treatment_feature_value.enabled)
         self.assertEqual(treatment_feature_value.value, "{'this': 'that'}")
 
+    def test_saves_prevent_pref_conflict(self):
```
jaredlockhart, line 359: You could just fold this into an existing saving test couldn't you?

Also I try to avoid putting these kinds of multiple steps, multiple cases into a single test, because if it fails you're not sure which of the cases is failing without inspecting further, so if we kept this I would split it into multiple tests, but even then it would kind of be testing the same thing twice.  All we need to know is that it'll update the db for some new value, what the value is kind of doesn't matter since that's all enforced by the serializer anyways.

app/experimenter/experiments/tests/test_models.py
```diff
@@ -692,6 +696,71 @@ def test_targeting_with_sticky_and_no_advanced_targeting_omits_sticky_expression
...
+            ],
+        )
+
+        self.assertEqual(
+            experiment.targeting,
+            (
+                '(browserSettings.update.channel == "release") && '
+                "(!('nimbus.test.string'|preferenceIsUserSet)) && "
+                "(!('nimbus.test.int'|preferenceIsUserSet)) && "
+                "(!('nimbus.test.boolean'|preferenceIsUserSet))"
```
jaredlockhart, line 729: Look at it go! 👏 👏 👏 👏

app/experimenter/experiments/tests/test_models.py
```diff
@@ -692,6 +696,71 @@ def test_targeting_with_sticky_and_no_advanced_targeting_omits_sticky_expression
         )
         JEXLParser().parse(experiment.targeting)
 
+    @mock_valid_features
+    def test_targeting_with_prevent_pref_conflicts_set_prefs(self):
+        Features.clear_cache()
+        call_command("load_feature_configs")
```
jaredlockhart, line 702: Oh yeah exactly, easy peasy 👍

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
```diff
@@ -596,6 +599,92 @@ describe("FormBranches", () => {
...
+  describe("preventPrefConflicts checkbox", () => {
+    it("doesn't render if no feature config set", () => {
+      render(
+        <SubjectBranches
+          experiment={{ ...MOCK_EXPERIMENT, featureConfigs: [] }}
+        />,
+      );
+
+      expect(
+        screen.queryByTestId("prevent-pref-conflicts-checkbox"),
```
jaredlockhart, line 612: 👌

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
```diff
@@ -596,6 +599,92 @@ describe("FormBranches", () => {
...
+        <SubjectBranches
+          experiment={{
+            ...MOCK_EXPERIMENT,
+            featureConfigs: [MOCK_FEATURE_CONFIG],
+          }}
+        />,
+      );
+
+      expect(
+        screen.queryByTestId("prevent-pref-conflicts-checkbox"),
```
jaredlockhart, line 627: 🙏

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
```diff
@@ -596,6 +599,92 @@ describe("FormBranches", () => {
...
+      ).not.toBeInTheDocument();
+    });
+
+    it("does render if the set feature config does not set prefs", () => {
```
jaredlockhart, line 631: Shouldn't this be if it does set prefs?

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
```diff
@@ -596,6 +599,92 @@ describe("FormBranches", () => {
...
+            featureConfigs: [MOCK_FEATURE_CONFIG_WITH_SETS_PREFS],
+          }}
+        />,
+      );
+
+      const checkbox = screen.getByTestId(
+        "prevent-pref-conflicts-checkbox",
+      ) as HTMLInputElement;
+      expect(checkbox).toBeInTheDocument();
+      expect(checkbox.checked).toEqual(false);
```
jaredlockhart, line 645: 💯

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
```diff
@@ -596,6 +599,92 @@ describe("FormBranches", () => {
...
+
+      {
+        const checkbox = screen.getByTestId(
+          "prevent-pref-conflicts-checkbox",
+        ) as HTMLInputElement;
+
+        expect(checkbox).toBeInTheDocument();
+        expect(checkbox.checked).toEqual(false);
+      }
+    });
```
jaredlockhart, line 686: Yeah like I'd treat these two transitions as two separate tests.  Moving from setsprefs to not setsprefs hides the check box, moving from not setsprefs to setsprefs reveals the checkbox.

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/reducer/index.test.ts
```diff
@@ -606,4 +607,14 @@ describe("formBranchesReducer", () => {
       ).toBeTruthy();
     };
   });
+
+  it("setPreventPrefConflicts sets preventPrefConflicts", () => {
+    const oldState = { ...MOCK_STATE };
```
jaredlockhart, line 612: Any way to explicitly set the initial state with `preventPrefConflicts` to `false`, just so its explicit here and changes to the default mock later don't break this test?

jaredlockhart (changes requested):
Ya awesome, that was faster than I expected 👏 👏 👏 👏 

Couple ideas for tests and the little `bool` thing instead of the ternary, pick and choose as you like.

Looks real good @brennie great job 🎉

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
brennie: Its testing that if you switch away to not-sets-prefs, the checkbox will be unchecked the next time you select a sets-prefs-feature, which is not the same as a test that switches from not-sets-prefs to sets-prefs (as there is no state to reset)

app/experimenter/experiments/api/v5/serializers.py
brennie: Good catch.

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
jaredlockhart: Ah yep that's a good point 👍

app/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.test.tsx
brennie: Copy / paste error :)

jaredlockhart (changes requested):

jaredlockhart (approved):
Yep, all looks right to me, tested locally everything works perfectly.  Great job getting this all together quickly, all very idiomatic and concise.  🎉 🎉 🎉 🎉

app/experimenter/docs/openapi-schema.json
```diff
@@ -3288,7 +3291,8 @@
                 "description",
                 "application",
                 "ownerEmail",
-                "schema"
+                "schema",
+                "setsPrefs"
```
yashika-khurana, line 3295: @jaredlockhart it does include this file

Issue #7885, opened 2022-10-27
The "Preview Recipe JSON" link leads to a blank page


data-sync-user:
*[Prerequisites]:*

* Have an experiment page open.

*[Steps to reproduce]:*

# Click the “Preview Recipe JSON” link from the sidebar.
# Observe the page displayed.

*[Expected result]:*

* The “Recipe JSON #” field is focused.

*[Actual result]:*

* A blank page is displayed.

*[Notes]:*

* The URL of the link is set for the Details page.
* Attached is a recording of the issue.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-2883)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/30108">recipe link.mov</a>

data-sync-user:
➤ Maria Heres commented:

I have verified that the issue is no longer reproducible. Clicking the link now scrolls down to the “Recipe JSON” field. Tested using Firefox Release 106.0.2 (Build ID: 20221025065831) on Windows 10.

PR #7886, opened 2022-10-27
For #7885 fix(nimbus): Link the preview json button in the sidebar to the correct place

eliserichards:
#7885

Because...

* The preview recipe json button in the sidebar was still linking to the details page

This commit...

* Updates the link to navigate to the json on the summary page
* Adds a test :)

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/SidebarActions/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/SidebarActions/index.test.tsx
@@ -235,4 +235,20 @@ describe("SidebarActions", () => {
       expect(screen.queryByTestId("CloneDialog")).not.toBeInTheDocument();
     });
   });
+
+  it("scrolls to json when preview recipe json button is clicked", async () => {
+    const experiment = mockExperiment({ isArchived: false, canArchive: true });
+    const refetch = jest.fn();
+    render(<Subject {...{ experiment, refetch }} />);
+
+    const jsonButton = await screen.findByTestId("button-recipe-json");
+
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/SidebarActions/index.test.tsx
```diff
@@ -235,4 +235,20 @@ describe("SidebarActions", () => {
...
+    render(<Subject {...{ experiment, refetch }} />);
+
+    const jsonButton = await screen.findByTestId("button-recipe-json");
+
+    fireEvent.click(jsonButton);
+
+    await waitFor(() => {
+      expect(
+        screen.queryByTestId("button-recipe-json"),
+      ).not.toBeInTheDocument();
```
jaredlockhart, line 251: Wait why is this the assertion?

jaredlockhart: That's so weird, the fake test viewport has some invisible fixed size that makes this true for this particular layout?  That's so weird.  Okay well if it works for now let's just go with it.

eliserichards: 😂

jaredlockhart (approved):
Okay ya so this gets us back to where we were where clicking it the first time scrolls you down, but scrolling back up and clicking it again doesn't, which was already the behaviour.  So this is fine.  Thanks for the quick fix @eliserichards ! 🎉

PR #7892, opened 2022-10-31
fix #7891 feat(nimbus): update branch slug on every save

### LINKED ISSUE #7891: Update branch slugs on every save
Originally, we set branch slugs to only be set once at create time and never again at any future save time because an early design of the client interface allowed querying by specific branch ids and so they had to be stable.  That client interface is gone now and so there should be no external consumers of branch slugs, so there should be no reason why we can't update them on every save.  Conversely it has caused repeated confusion for users that their branch slugs don't match their names.  So let's allow regenerating branch slugs on every save.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2888)

jaredlockhart:
Because

* Originally we set the branch slugs to only be set once at branch creation time
* An early version of the mobile client API allowed querying by branch so it was necessary to freeze branch slugs to keep them in sync
* That API is gone now and so there are no external consumers of branch slugs
* This allows branch slugs to be updated freely
* Having the branch names and slugs out of sync creates confusion for users

This commit

* Moves branch slug saving from create to every save

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -387,10 +387,6 @@ class NimbusBranchSerializer(serializers.ModelSerializer):
 
         return data
 
-    def create(self, data):
-        data["slug"] = slugify(data["name"])
-        return super().create(data)
-
     def _save_feature_values(
         self, feature_enabled, feature_value, feature_values, branch
     ):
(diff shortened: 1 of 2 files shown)


yashika-khurana (changes requested):
overall looks good, thank you @jaredlockhart , just thought about a test case to include replicating and changing the slug?? what do you think?

jaredlockhart:
Oh you mean the clone flow?  Yeah but just having it on save should cover that case as well.  If we see an edge case come up in production we can add more tests.

yashika-khurana:
@jaredlockhart perfect agreed, approve approve 🎉

yashika-khurana (approved):

PR #7902, opened 2022-10-31
fix #7901 chore(nimbus): remove opt out pref

### LINKED ISSUE #7901: Remove optoutstudies pref from targeting
We previously had to explicitly check for the opt out pref in our targeting to prevent enrolling clients that had opted out of studies.  Since Firefox 84, that opt out logic is implemented elsewhere and so it is no longer necessary to include it in targeting.  Now that it's been a sufficient amount of time since 84 and telemetry shows no significant number of clients reporting that version, we can safely deprecate that check.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2890)

jaredlockhart:
Because

* We used a pref check in targeting to look for user opt out from studies
* After Firefox 84 this check was moved to not depend on the pref
* Now that a sufficient amount of time has passed and we no longer need to support 84, we can remove this targeting check

This commit

* Removes the opt out pref check from targeting


jrbenny35:
> Now that a sufficient amount of time has passed and we no longer need to support 84, we can remove this targeting check

Would be lovely to have some sort of deprecation structure we follow

jrbenny35 (approved):

brennie (approved):

jaredlockhart:
> > Now that a sufficient amount of time has passed and we no longer need to support 84, we can remove this targeting check
> 
> Would be lovely to have some sort of deprecation structure we follow

We do kinda have something like that here:

https://github.com/mozilla/nimbus-shared/blob/main/.github/ISSUE_TEMPLATE/schema-change.md

but that wouldn't catch this since it's not a schema change.  So yeah there's no good way to track this right now.  It's an interesting question.  We should collect a couple examples of deprecations and figure out a nice little pattern we can follow and track.  Good idea.

jaredlockhart:
Oh and I referred to https://mozilla.cloud.looker.com/explore/firefox_desktop/client_counts?qid=2uPpBvwJW5j9PH4OCe8w9q&origin=share-slug to look for which versions are active.  I hope in the future we have something like this as part of our new Nimbus health dashboards so we can keep track of which versions we need to support and then we can use that as a guide for when we can safely deprecate things.

Issue #7815, opened 2022-10-06
Make Audience and Branches on summary/detail page as grid structure to avoid empty space


data-sync-user:
We have a lot of empty space on the details page which can be improved using grid strucutre

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2709)

PR #7904, opened 2022-11-01
fix #7816 feat(nimbus): Use cards on Summary page

### LINKED ISSUE #7816: Use cards on the details page
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2713)

yashika-khurana:
Because

* We want to improve UX for the summary pages

This commit

* Use cards layout on the `Overview, audience and branches page`

`Overview page`
before
Screenshot 2022-11-01 at 3 04 04 PM
After
Screenshot 2022-11-01 at 3 04 27 PM
`Audience page`
before
Screenshot 2022-11-01 at 3 04 43 PM
After
Screenshot 2022-11-01 at 3 04 56 PM
`Branches Page`
Before
Screenshot 2022-11-01 at 3 05 10 PM
After
Screenshot 2022-11-01 at 3 05 34 PM

fixes #7816 #7815


app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.tsx
```diff
@@ -50,18 +52,18 @@ const TableBranches = ({
   const hasOneBranchNameSet = Boolean(savedBranches);
 
   return (
-    <>
+    <Card className="mt-4" border="light">
```
jaredlockhart, line 55: Maybe instead of doing an outer 'Branches' card with multiple 'Branch' cards within, maybe just a single flat card where the header is `Branch: <branch name>`?

Something like this (but with the fonts fixed, etc)

Screenshot 2022-11-02 at 11-42-24 Upgradable motivating software – Summary Experimenter

jaredlockhart (changes requested):
Oh yeah this helps so much.  Maybe try taking out the outer card for the branches.  What do you think of the borders?  I'm kinda iffy on them.  Not sure if it's better with them or without them.

@yashika-khurana @eliserichards what do you think?

Screenshot 2022-11-02 at 11-47-03 Upgradable motivating software – Summary Experimenter

eliserichards:
I think these can be moved into a separate issue: 

> 1. Make the 'takeaways' section also into a card
> 2. Move it under the timeline?

app/experimenter/nimbus-ui/src/components/Summary/TableBranches/index.tsx
eliserichards: There should definitely be one "group" with multiple "groups" within it, but I'm unsure if it should be a card or not (what that would look like visually, I gotta look it up). What I do love about Yashika's implementation is the whitespace in between the cards -- we don't want the users' eyes to glaze over while they're looking at this page; we want them to be able to find things quickly

yashika-khurana: yeah idea was to pack them on the main component, internally it will be in cards, but let me look if I can make outer component wrapping branches better, thank you for the feedback @jaredlockhart @eliserichards 🎉

yashika-khurana:
@jaredlockhart @eliserichards I have addressed the changes and also did the takeaways part in the same PR, here is the new screenshot 
image

jaredlockhart (approved):
I love it.  I love it.  Fantastic.  Interesting choice taking the side borders off the cards!  I think it totally works.  Full width works.  Checked in all states.  Everything looks wonderful.  Such an improvement on this page.  Great job @yashika-khurana thank you!  🎉 🎉 🎉

eliserichards (approved):
this is perfection 🥇 let's get this landed! 🚀

PR #7912, opened 2022-11-03
Add easy setup targeting

AllegroFox:
Targeting to support https://bugzilla.mozilla.org/show_bug.cgi?id=1798094 - New users/installs with a profile <24h old, with Firefox not yet set to default


jaredlockhart (approved):
Looks good 👍

app/experimenter/targeting/constants.py
```diff
@@ -85,6 +86,33 @@ def __post_init__(self):
...
+        "{first_run} && os.windowsBuildNumber >= 18362 && {new_profile} && "
+        "{need_default}".format(
+            first_run=FIRST_RUN.targeting,
+            new_profile=NEW_PROFILE,
+            need_default=NEED_DEFAULT,
+        )
+    ),
+    desktop_telemetry=(
+        "{first_run} AND environment.system.os.windows_build_number >= 18362 AND "
+        "!isDefaultBrowser AND doesAppNeedPin"
```
punamdahiya, line 109: noticed desktop_telemetry has doesAppNeedPin extra check, curious where desktop_temetry is used and if difference of condition in desktop_telemetry is ok?

PR #7915, opened 2022-11-03
Add targeting for Firefox Suggest data collection

ncloudioj:
We need this targeting for the upcoming experiments / rollouts of Firefox Suggest.


jaredlockhart (approved):
Looks good 👍

PR #7916, opened 2022-11-03
7905/show application with features

yashika-khurana:
Because

* The feature select drop-down on the list page lists all features for all applications, but since some features exist on multiple applications and may have the same name there's currently no way to distinguish which is which, for example:

Screenshot 2022-11-03 at 2 41 05 PM

This commit

* Adds the application name along with the feature name in the dropdown
Screenshot 2022-11-03 at 2 42 56 PM


app/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
```diff
@@ -124,7 +128,12 @@ const FilterSelect = <
           value: fieldValue,
           placeholder: "All " + fieldLabel + "s",
           getOptionLabel: (item: OptionTypeBase) =>
-            item[optionLabelName as string],
+            fieldLabel === "Feature"
+              ? item[optionLabelName as string] +
+                " (" +
+                displayConfigLabelOrNotSet(item["application"], applications) +
+                ")"
```
jaredlockhart, line 135: Oh try not to use string concatenation like this, you can use a template literal just like a format string in Python:

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

jaredlockhart (dismissed):
Awesome 🙏

eliserichards (approved):
YES 🙌 🥇

jaredlockhart (approved):
👏 👏 👏 👏 👏  awesome ty 🙏

PR #7933, opened 2022-11-08
fix #7917 feat(nimbus): show errors for metrics that would otherwise be hidden

### LINKED ISSUE #7917: Errors not displaying for certain metrics
If a metric has errors, but no data, it is possible that the metric slug will not appear in our currently expected places (metadata and/or data of viz and config objects), so the metric will not render on the Results page. We use the same logic for rendering errors, so there are scenarios where we have metrics with errors, but those errors do not display on the Results page, negating much of the value of recording these errors in the first place.

We should have some additional logic that checks for errors that would not otherwise be displayed, and display them.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2918)
┆Link To Issue: https://mozilla-hub.atlassian.net/browse/EXP-2918

mikewilli:
Because

* metrics that are not part of outcomes and that do not have data are not displayed on the Results page
* these metrics may have errors

This commit

* adds logic to retrieve and display errors for metrics that would otherwise be hidden
* displays these errors at the bottom of the Results page

image


scholtzan (approved):

app/experimenter/nimbus-ui/src/components/PageResults/index.test.tsx
```diff
@@ -222,6 +222,50 @@ describe("PageResults", () => {
...
+    );
+
+    expect(screen.getByText("Other Metric Errors"));
+
+    expect(screen.getByText("BadThingException calculating test_statistic"));
+
+    expect(
+      screen.getByText(
+        "Error while computing statistic test_statistic for metric bad_metric",
+        { exact: false },
```
jaredlockhart, line 264: Exact: false?? I didn't know you could do that 😱

jaredlockhart (approved):
Yup looks good 👍 🎉

PR #7960, opened 2022-11-17
fix #7959 feat(nimbus): add targeting for users who have prefers-redu…

### LINKED ISSUE #7959: Create targeting for user who don't prefer reduced motion
This issue is to add targeting for if users who don't have prefers-reduced-motion enabled, to be used in future experiments. There is  a [bugzilla tracking bug here](https://bugzilla.mozilla.org/show_bug.cgi?id=1800128).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2934)

jwayn:
…ced-motion off

Because

* we would like to run experiments that target users who don't have prefers-reduced-motion enabled

This commit

* adds targeting for users who don't have prefers-reduced-motion enabled


jaredlockhart (approved):
Looks good! Thanks @jwayn 🎉

Issue #7982, opened 2022-11-28
Implement tests for current state changes


data-sync-user:
Our current publish states are not being fully tested. This ticket should cover the state transitions that are currently defined in our state diagrams and sequence diagrams.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2951)

PR #7987, opened 2022-11-28
For #7982 test(nimbus): Add tests for status transitions and publish states

eliserichards:
For #7982

Because...

* We aren't fully testing the state transitions that are defined in the state diagram and sequence diagrams

This commit...

* Adds serializer tests to help boost our surety that we are allowing the correct state changes

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
@@ -1,4 +1,5 @@
 from django.test import TestCase
+from parameterized import parameterized
 
 from experimenter.experiments.api.v5.serializers import NimbusExperimentSerializer
 from experimenter.experiments.models import NimbusExperiment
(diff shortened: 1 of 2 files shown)


app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
```diff
@@ -42,17 +44,138 @@ def test_update_experiment_with_invalid_status_error(self):
...
+        )
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "publish_status": publish_status,
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertEquals(serializer.is_valid(), valid)
```
yashika-khurana, line 181: @eliserichards thank you for adding these tests they are much much easier to understand and makes more stronger test suite, I think we should also add more test cases like what if status is `IDLE`, `REVIEW`, `APPROVED` and `WAITING`, like what are the valid next and previous states are 🤔

yashika-khurana (dismissed):
thank you for adding this @eliserichards 🔥, I added a comment for more test cases like that 🙈

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_validation_mixin.py
eliserichards: That's a great idea! I'll file issues for those 💪

yashika-khurana (approved):
@eliserichards  thank you for adding extra tickets :), this looks great 🚀

Issue #7983, opened 2022-11-28
Serializer tests for status and publish statuses


data-sync-user:
Our current implementation of the lifecycle states and publish statuses are not fully covered by our unit tests. Let's get these transitions covered before we work on DIRTY editability.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2952)

data-sync-user:
➤ Elise Richards commented:

There’s some conflicts between what is defined in our sequence diagrams and state diagrams and what is actually implemented. This review has been on hold for a few days waiting for Jared’s return so we don’t accidentally miss something / break something due to missing context. Should be unblocked now 👍

PR #7989, opened 2022-11-29
For #7983 test(nimbus): Update tests and test factories to match state and sequence diagrams

eliserichards:
For #7983

Because...

* We are not consistently showing all of the possible state transitions

This commit...

* Removes invalid factory `LifecycleState`:
   * LIVE_IDLE_REJECT_ENDING
* Adds `LifecycleState`:
   * LIVE_REVIEW
   * LIVE_APPROVED_WAITING
* Adds missing factory `Lifecycles`: 
   * LAUNCH_APPROVE_REJECT
   * LAUNCH_APPROVE_REJECT_ROLLBACK
   * LIVE_ENROLLING_PAUSED
   * ENDING_REJECT
   * ENDING_APPROVE_REJECT_ROLLBACK
* Adds tests for:
   * v5 serializer (`test_nimbus_status_transition_validator`, `test_nimbus_experiment_serializer`)
   * kinto tasks (`test_tasks`)
* Refactors some tests to use the factory `create_with_lifecycle` instead of just using `create` and a `Status`


app/experimenter/experiments/constants.py
```diff
@@ -173,26 +173,29 @@ class Type(models.TextChoices):
...
-        Status.PREVIEW: (Status.DRAFT,),
-    }
     STATUS_ALLOWS_UPDATE = (Status.DRAFT,)
 
     # Valid status_next values for given status values
     VALID_STATUS_NEXT_VALUES = {
-        Status.DRAFT: (None, Status.LIVE),
-        Status.PREVIEW: (None, Status.LIVE),
+        Status.DRAFT: (None, Status.PREVIEW, Status.LIVE),
+        Status.PREVIEW: (None, Status.DRAFT, Status.LIVE),
```
yashika-khurana, line 181: 🎉

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -850,8 +850,8 @@ def test_reject_ending_experiment(self):
 
     def test_end_experiment(self):
         user_email = "user@example.com"
-        experiment = NimbusExperimentFactory.create(
-            status=NimbusExperiment.Status.LIVE,
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            lifecycle=NimbusExperimentFactory.Lifecycles.PUBLISH_APPROVE_APPROVE
```
yashika-khurana, line 854: just a suggestion: this name is little confusing `PUBLISH_APPROVE_APPROVE` 😆

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_transition_validator.py
```diff
@@ -15,21 +15,21 @@ def setUp(self):
...
-                "status": NimbusExperiment.Status.LIVE,
+                "status": NimbusExperiment.Status.COMPLETE,
                 "changelog_message": "test changelog message",
             },
             context={"user": self.user},
         )
         self.assertFalse(serializer.is_valid())
         self.assertEqual(
             serializer.errors["status"][0],
-            "Nimbus Experiment status cannot transition from Draft to Live.",
```
yashika-khurana, line 32: I think we should keep the test for draft to live not valid transition

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -93,12 +99,23 @@ def test_check_with_no_approved_publish_status_pushes_nothing(self):
...
+                NimbusExperimentFactory.Lifecycles.PUBLISH_APPROVE_WAITING,
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_WAITING,
+            ],
+            [
+                NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_WAITING,
+            ],
+        ]
+    )
```
yashika-khurana, line 114: 🎉

yashika-khurana (commented):
I have given the first pass, I need to sit focused to review it again 😆 , it is more clear now, amazing work @eliserichards  👏 , I really like the `DRAFT_CREATED` status, just some minor suggestion ❤️

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_transition_validator.py
eliserichards: This was a weird test, because we **can** go from Draft to Live (see [Publish Approve/Approve](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/README.md#publish-approveapprove) workflow). I'm not sure why this test was here in the first place 🤷‍♀️

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: I agree that it's a bit confusing, especially when there are so many lifecycle states

The thought in changing these names from `LIVE_` to `PUBLISH_` was to match the names to the sequence diagram names: https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/README.md#publish-approveapprove

eliserichards: If you think it's too confusing I can change them back though

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_transition_validator.py
eliserichards: This might be a "wait for Jared because there might be a reason we're missing" case

eliserichards:
Sounds good! I'll do that now :D

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -877,15 +877,15 @@ def test_end_experiment(self):
...
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            lifecycle=NimbusExperimentFactory.Lifecycles.PREVIEW
         )
         response = self.query(
             UPDATE_EXPERIMENT_MUTATION,
             variables={
                 "input": {
                     "id": experiment.id,
-                    "status": NimbusExperiment.Status.DRAFT.name,
+                    "status": NimbusExperiment.Status.PREVIEW.name,
```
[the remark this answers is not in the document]
eliserichards: This is true. I removed `LIVE` from the allowed list for `PREVIEW`. Let's see if that breaks the integration tests 👀 (and make sure it still works)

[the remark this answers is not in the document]
eliserichards: womp womp: looks like it was actually going from Preview to Live

image

The question is, do we want to update our diagrams, or do we want to update the code 🤔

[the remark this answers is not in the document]
eliserichards: that's what I'm thinking too... @jaredlockhart ?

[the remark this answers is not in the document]
jaredlockhart: Experiment is in preview:

image

Click the launch button, request:

image

Experiment is in draft/review(live)

image

It doesn't go to live until after it's already approved in remote settings.

I don't understand why we'd be changing the launch flow at all in this PR?  Shouldn't this only be scoped to the live update changes?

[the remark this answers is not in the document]
jaredlockhart: > womp womp: looks like it was actually going from Preview to Live
> 
> image
> 
> The question is, do we want to update our diagrams, or do we want to update the code 🤔

Also this is talking about `status_next` not `status`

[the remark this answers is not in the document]
jaredlockhart: The change you made in the PR is to go to `preview/review(live)` but that's not the right state, the docs and the code both describe `draft/review(live)`, which is the state it would go to either from `draft/idle(none)` or `preview/idle(none)`.   None of that should have changed?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -700,7 +700,7 @@ class NimbusExperimentSerializer(
         required=False,
         validators=[
             NimbusStatusTransitionValidator(
-                transitions=NimbusConstants.VALID_STATUS_TRANSITIONS,
+                transitions=NimbusConstants.VALID_STATUS_NEXT_VALUES,
```
jaredlockhart, line 703: Wait, why would we use the `status_next` transitions on the `status` field?

app/experimenter/experiments/constants.py
```diff
@@ -173,26 +173,29 @@ class Type(models.TextChoices):
...
 
     # Valid status_next values for given status values
     VALID_STATUS_NEXT_VALUES = {
-        Status.DRAFT: (None, Status.LIVE),
-        Status.PREVIEW: (None, Status.LIVE),
+        Status.DRAFT: (None, Status.PREVIEW, Status.LIVE),
+        Status.PREVIEW: (None, Status.DRAFT),
         Status.LIVE: (None, Status.LIVE, Status.COMPLETE),
+        Status.COMPLETE: (None),
     }
```
jaredlockhart, line 184: Wait why are we getting rid of the `VALID_STATUS_TRANSITIONS` here?

eliserichards: Because they didn't match the sequence diagrams. That's the main purpose of this PR 👍

app/experimenter/experiments/api/v5/serializers.py
eliserichards: The status transitions should be identical to the status next transitions

app/experimenter/experiments/constants.py
eliserichards: I went through our sequence diagrams, and everywhere it starts as, for example, `DRAFT`, it either transitions to `None`, `PREVIEW`, or `LIVE`. I did these for each of the statuses to make sure we are matching the diagrams in all cases

jaredlockhart: Right but for instance, starting at  `status=Draft`:

1) `status` can't be `None`?
2) These validators are used to limit which statuses the UI can set, not all possible state transitions, so for instance, the UI isn't allowed to set `status` to `Live`, only the workers can do that.

`status` and `status_next` share the same values but they're tracking two different things.  The aggregate state is the tuple of `(status, publish_status, status_next)` but each one is separate and unique, I don't think you can fold `status` and `status_next` together the way you're suggesting?

app/experimenter/experiments/api/v5/serializers.py
eliserichards: I guess the question is, why were we using VALID_STATUS_TRANSITIONS only for draft <-> preview?

eliserichards: The assumption that I made is that the `NimbusStatusTransitionValidator` should test _all_ status transitions, not just that one

jaredlockhart: > The status transitions should be identical to the status next transitions

I don't think that's true?

jaredlockhart: > The assumption that I made is that the `NimbusStatusTransitionValidator` should test _all_ status transitions, not just that one

Ah yeah see, two different concerns.  The V5 API acts as a broker between the UI and the database, so it only needs to be concerned with what the UI can do, and the UI only allows users to move between Draft and Preview. 

If we wanted to have a broader status transition validator, that would live at the model layer so that anywhere in the code that tries to change the status is forced to go through it so statuses are always valid.  That is a pattern we've explored in the past, but the big problem with that is that every so often you want to actually break out of the usual flow and hop between states that aren't usually moved between, maybe in tests, maybe in code, or most importantly, from the admin where you need to be able to reset some state.  Actually that just happened with me this week.  So in general when we've tried to do like global state transition enforcement it hurt more than it helped.  

The other way we've kind of soft approached this is with the way we're storing the states as those Q object constants and then just referring to them everywhere.  That way you can easily look at the list of all states as those constants and compare them to the diagrams more easily.

So yeah I get what you're thinking of, but it's not something I'd recommend, and even if it was, it wouldn't go in here.

app/experimenter/experiments/constants.py
eliserichards: ahh #2 makes sense. I think that was unclear to me when I was going through this 👍

jaredlockhart: Also it would be helpful for me instead of framing this as 'make them line up with the diagrams' as explictly listing out the concrete changes like 'Adding state X' 'Updating tests to include new lifecycle Y' etc

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_transition_validator.py
eliserichards: Ahhh this is making sense now. This test is because this status validator is only checking what the UI is able to do, and it is not allowed to change draft to live

PR #7995, opened 2022-12-01
fix #7993 test(nimbus): Add test for homepage table check

### LINKED ISSUE #7993: Create an integration test that checks the homepage
Right now, our integration tests don’t actually check that the table on the home page loads correctly. We recently had a failure that would have been caught if our tests had a check for the home page table.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2955)

jrbenny35:
Because:
* We don't currently have any checks that make sure experiments are loading properly on the homepage, and within the tables

This commit:
* Adds a test that creates an experiment and publishes it, and then checks the live table for it.


yashika-khurana (approved):
This is great @jrbenny35 🎉 , we do have UI test for the table, but not with complete integration with kinto, thank you @jrbenny35 🎉

yashika-khurana (changes requested):
oops one check is failing 👀

jrbenny35:
> oops one check is failing 👀

Ahh looks like its running in parallel and isn't approving the right experiments..ill fix that

jrbenny35:
@yashika-khurana I moved the test to the ```remote_settings``` file and gave it a specific run group when it is running in parallel. This file is mainly used for any tests that use remote settings, so this fits fine imo

yashika-khurana (approved):
makes clear sense, thank you @jrbenny35 🎉

Issue #7994, opened 2022-12-01
Population slider won't allow '.' or numbers to be entered directly


shell1:
The slider won't let us enter '.' so we can enter non-whole number or even enter numbers directly - so is taking away capabilities.  I found a way to hack in .5% by doing to 5% and adding a '.' before.  

I get NaN when I do anything outside of dragging the bar and then can't type until delete NaN.

I can't add 5.5% - as it won't let me type a . after a number.  It won't let me type in 7% - but i can type in 5 and then type of that number with a 7.  

also it will let you target over 100% - so 1000% you can save.

![image](https://user-images.githubusercontent.com/8192232/205149290-5798c6b9-1039-429e-ac16-8802b658eda2.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2956)

data-sync-user:
➤ Elise Richards commented:

Backspacing all the way makes the field show NaN instead of empty

PR #7996, opened 2022-12-01
For #7994 fix(nimbus): Fixes the population slider to allow decimals and not show NaN

eliserichards:
Because ...

* We were trying to be clever by only allowing number-based input

This commit...

* Fixes the text input so it doesn't show `NaN` as the value if a non-number input is typed or the text is deleted 
* Allows `""` and doesn't force it to `0`
* Adds a default value in the text input (0)


app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -103,20 +103,25 @@ export const FormAudience = ({
...
-    return parseFloat(value ?? "0");
-  }
+  // function parsePercent(value: string | null): string {
+  //   if (value === "") {
+  //     return value;
+  //   } else {
+  //     let newVal = parseFloat(value).toString;
+  //     return parseFloat(value ?? "").toString();
+  //   }
+  // }
```
yashika-khurana, line 113: do you want to keep new function 👀 , it is commented out

yashika-khurana (changes requested):
Thank you @eliserichards for the quick fix 🎉 , Can you add some test cases regarding the scenarios reported just to make sure we are handling it 👀

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
eliserichards: just fixed it 😂

yashika-khurana (approved):
Awesome Awesome \o/, thank you for removing clicks as well 😆

eliserichards:
This still needs to be fixed: 

`nimbus/test_remote_settings.py::test_check_live_status_on_home_page[FENIX] FAILED `

eliserichards:
^ Looks like it was an intermittent failure 🤔

PR #8002, opened 2022-12-05
fix #7966 feat(nimbus): Option to select team projects

### LINKED ISSUE #7966: Team filtering option to select on overview page
After adding the backend support to get the project options, add a new field on the overview page to choose the appropriate team

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2937)

yashika-khurana:
Because

* We want to map experiments to the project/team 

This commit

* Allows the option to `select team` on the `Overview page` and test cases to verify changes
Screenshot 2022-12-05 at 12 53 31 PM
Screenshot 2022-12-05 at 12 53 37 PM

fix #7966

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/FormOverview/index.test.tsx
@@ -6,7 +6,7 @@ import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import React from "react";
 import { DOCUMENTATION_LINKS_TOOLTIP } from ".";
 import { FIELD_MESSAGES, RISK_QUESTIONS } from "../../lib/constants";
-import { mockExperimentQuery } from "../../lib/mocks";
+import { mockExperimentQuery, MOCK_CONFIG } from "../../lib/mocks";
 import { assertSerializerMessages } from "../../lib/test-utils";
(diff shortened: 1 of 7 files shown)


jaredlockhart (changes requested):
Yep, all looks correct to me, tested locally and works and looks great.  Little confused about why to create that type when we could just refer to the type we already have?  

In theory there's nothing wrong with this, but one of the patterns I see is detaching type definitions from the GQL supplied types, and then the contract between the backend and frontend breaks down becuase they're not enforced, so I worry about a similar thing happening here.  

In theory TypeScript is smart enough to know that the type signature of your new type matches the type that comes from the API so if one would break it should break here, so your new type is really only an alias, but I'm not sure of the exact semantics of `as` and whether it blindly casts or actually tries to enforce homomorphy.  I think we should try to avoid that as much as possible and just use the GQL types directly?  

I want to go in and actually refactor a bunch of things to more explicitly use the GQL types to prevent this kind of issue.

eliserichards (commented):
This is looking great! Looks like the code coverage is complaining now ☹️ I'm happy to approve once that is green

yashika-khurana:
@eliserichards all done now!

jaredlockhart (approved):
Awesome, thnx for updating, looks good, g/j @yashika-khurana 🎉

PR #8029, opened 2022-12-09
fix #8027 feat(nimbus): Update Existing Win7 user targeting

### LINKED ISSUE #8027: Update" Existing Windows 7+ user" advanced Experimenter targeting to exclude users signed into an FX account
[Tracking on Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1804935)

> After discussion with product, we'd like to update the advanced targeting of an upcoming existing user device migration experiment. The targeting should also exclude users signed into a Firefox account.
> 
> We may be able to use something like: `!("services.sync.username" | preferenceValue)`

Refs: 97cbd97

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2978)

aminomancer:
Fixes #8027
Refs: 97cbd97

### DIFF
--- a/app/experimenter/targeting/constants.py
+++ b/app/experimenter/targeting/constants.py
@@ -809,6 +809,22 @@ EXISTING_WINDOWS_USER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+EXISTING_WINDOWS_USER_NO_FX_ACCOUNT = NimbusTargetingConfig(
+    name="Existing Windows 7+ user without Fx account",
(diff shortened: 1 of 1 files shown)


app/experimenter/targeting/constants.py
```diff
@@ -802,7 +802,12 @@ def __post_init__(self):
     name="Existing Windows 7+ user",
```
lmegviar, line 802: Let's update the name and description to make it clear this doesn't include users signed in to FXA sync

jaredlockhart (approved):
Thank you looks good!

PR #8043, opened 2022-12-13
fix #7571 feat(nimbus): parse exposure events analysis and display results

### LINKED ISSUE #7571: Results grouped by exposed segments
exposure events can be configured in Jetstream, but once we start calculating metrics from them, we need an option to view the analyses group by exposed segments on the results page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2661)

mikewilli:
Because

* Jetstream can perform exposure events analysis
* there is no way to see those results outside of BigQuery

This commit
* parses exposure events analysis results
* allows user to display default enrollment results or exposures results (if available)


jaredlockhart (changes requested):
Let's try doing this in a couple more steps, ideas for a broken down version of this:

1. New jetstream loader logic to parse the new data into a new set of parallel keys that coexist with the existing ones, with a data migration to prepopulate those keys for existing data

1. Update UI to use new data structure

1. Remove old code/data

mikewilli:
Closing as this work is being done in smaller discrete PRs

Issue #8012, opened 2022-12-07
Allow Live rollouts to update


data-sync-user:
Change serializer constants to add `LIVE` status: 

`STATUS_ALLOWS_UPDATE = (Status.DRAFT, Status.LIVE)`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2968)

PR #8047, opened 2022-12-13
Fixes #8022 test(nimbus): Add lifecycles and lifecycle states for dirty

eliserichards:
Because...

* We want our test factories to cover the full extent of our status and publish status lifecycles

This commit...

* Adds the following `LifecycleStates` for the `DIRTY` publish status:
   * LIVE_DIRTY
   * LIVE_REVIEW
   * LIVE_APPROVED
   * LIVE_APPROVED_UNPAUSED
   * LIVE_APPROVED_WAITING
   * LIVE_WAITING_UNPAUSED
   * LIVE_IDLE_REJECT

* Adds the following `Lifecycles` for `DIRTY` (these cover all scenarios in the [sequence diagrams](https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/README.md#update-approveapprove)):
   * LIVE_DIRTY
   * LIVE_REVIEW_REQUESTED
   * LIVE_REJECT
   * LIVE_APPROVE
   * LIVE_APPROVE_WAITING
   * LIVE_APPROVE_APPROVE
   * LIVE_APPROVE_REJECT
   * LIVE_APPROVE_TIMEOUT
   * LIVE_REJECT_MANUAL_ROLLBACK

* Adds a few missing `Lifecycles` (these are in our sequence diagrams but were not represented here):
   * LAUNCH_APPROVE_REJECT
   * LAUNCH_REJECT_MANUAL_ROLLBACK
   * ENDING_REJECT
   * ENDING_APPROVE_REJECT
   * ENDING_REJECT_MANUAL_ROLLBACK

For #8022

Issue #7823, opened 2022-10-10
UI for published/unpublished dirty changes on Summary page


data-sync-user:
See [https://mozilla-hub.atlassian.net/browse/EXP-2562](https://mozilla-hub.atlassian.net/browse/EXP-2562|smart-link)for state diagram.

Once a rollout has been **published**, we need some sort of UI to show whether is has unpublished (dirty) changes. 

[https://mozilla-hub.atlassian.net/browse/EXP-2954](https://mozilla-hub.atlassian.net/browse/EXP-2954|smart-link) adds the existing Audience page to the sidebar for Live rollouts. Once an edit has been made and saved on the Audience page, we should see the existing prompt on the Summary page to send to Review

!image-20221130-203507.png|width=754,height=190!

!image-20221130-203536.png|width=1438,height=228!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2839)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/31628">image-20221130-203507.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/31629">image-20221130-203536.png</a>

jaredlockhart:
Okay how about this.

On the detail page, the Audience button is still available while live:

<img width="1526" alt="Screen Shot 2022-11-03 at 2 50 18 PM" src="https://user-images.githubusercontent.com/119884/199808996-5b31d821-fdc7-469d-8744-d9333992255b.png">

All fields except population are locked, slider for population

<img width="1544" alt="image" src="https://user-images.githubusercontent.com/119884/199809091-c9f45f1a-d9b2-436e-83fa-2927b24dbca3.png">

Save will save without navigating, save and continue saves and navigates back to detail page?

Now the review buttons appear because we're 'dirty'

<img width="1528" alt="image" src="https://user-images.githubusercontent.com/119884/199809364-9b3e2308-8a13-4519-93e2-8956ae61b6c8.png">

PR #8048, opened 2022-12-13
For #8012 feat(nimbus): Allow edits on the backend while rollouts are LIVE

eliserichards:
Because...

* We want to allow edits to Live rollouts (and NOT live experiments)

This commit...

* Adds the `LIVE` status to the list `PUBLISH_STATUS_ALLOWS_UPDATE`
* Adds a serializer check to determine if the `Status.LIVE` item we are editing is a Rollout
* Updates tests to ensure that we are only allowing these Live updates in the correct states

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -591,33 +591,45 @@ class NimbusStatusValidationMixin:
         if self.instance:
             for status_field, restricted_statuses in restrictive_statuses.items():
                 current_status = getattr(self.instance, status_field)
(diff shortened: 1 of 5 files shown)


eliserichards:
This should wait for #8047 to land first

jaredlockhart:
Oh it just occurred to me we shouldn't update the UI to allow live edits for rollouts until the kinto tasks are updated to broker those changes.  I think we need to make some updates there otherwise a user could make a live edit but the tasks would never kick in to actually publish those changes.  We should update that first before we expose it in the UI?

eliserichards:
Exactly right! The UI will be added in #8050 and #7823 👍

PR #8056, opened 2022-12-14
For #7983 test(nimbus): Add serializer tests for publish_status and status

eliserichards:
Because...

* We are about to allow Live rollouts to be Dirty 👷 

This commit...

* Bulks up our existing tests for our serializers so that we can easily adjust these status and publish_status transitions for `DIRTY`

### DIFF
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
@@ -56,12 +56,8 @@ class TestNimbusStatusNextTransitionValidator(TestCase):
             ],
         ]
     )
-    def test_update_status_errors_for_status_next_live(
-        self, status, valid
-    ):
-        experiment = NimbusExperimentFactory.create(
-            status=status
-        )
+    def test_update_status_errors_for_status_next_live(self, status, valid):
+        experiment = NimbusExperimentFactory.create(status=status)
         serializer = NimbusExperimentSerializer(
             experiment,
             data={
@@ -71,7 +67,7 @@ class TestNimbusStatusNextTransitionValidator(TestCase):
             context={"user": self.user},
         )
         self.assertEquals(serializer.is_valid(), valid)
-    
+
     @parameterized.expand(
         [
             [
@@ -92,12 +88,8 @@ class TestNimbusStatusNextTransitionValidator(TestCase):
             ],
         ]
     )
-    def test_update_status_errors_for_status_next_complete(
-        self, status, valid
-    ):
-        experiment = NimbusExperimentFactory.create(
-            status=status
-        )
+    def test_update_status_errors_for_status_next_complete(self, status, valid):
+        experiment = NimbusExperimentFactory.create(status=status)
         serializer = NimbusExperimentSerializer(
             experiment,
             data={
@@ -107,7 +99,7 @@ class TestNimbusStatusNextTransitionValidator(TestCase):
             context={"user": self.user},
         )
         self.assertEquals(serializer.is_valid(), valid)
-        
+
     @parameterized.expand(
         [
             [
@@ -128,12 +120,8 @@ class TestNimbusStatusNextTransitionValidator(TestCase):
             ],
         ]
     )
-    def test_update_status_errors_for_status_next_draft(
-        self, status, valid
-    ):
-        experiment = NimbusExperimentFactory.create(
-            status=status
-        )
+    def test_update_status_errors_for_status_next_draft(self, status, valid):
+        experiment = NimbusExperimentFactory.create(status=status)
         serializer = NimbusExperimentSerializer(
             experiment,
             data={
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
@@ -1,0 +1,196 @@
+from django.test import TestCase
+from parameterized import parameterized
+
+from experimenter.base.models import SiteFlag, SiteFlagNameChoices
+from experimenter.experiments.api.v5.serializers import NimbusExperimentSerializer
+from experimenter.experiments.models import NimbusExperiment
+from experimenter.experiments.tests.factories import NimbusExperimentFactory
+from experimenter.openidc.tests.factories import UserFactory
+
+
+class TestNimbusStatusNextTransitionValidator(TestCase):
+    maxDiff = None
+
+    def setUp(self):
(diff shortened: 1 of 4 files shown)


app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_transition_validator.py
```diff
@@ -50,57 +51,180 @@ def test_update_publish_status_from_approved_to_review_error(self):
             "Nimbus Experiment publish_status cannot transition from Approved to Review.",
         )
 
-    def test_launch_request_while_disabled_error(self):
```
eliserichards, line 53: This side-by-side view is not being very nice. It looks like this changed, but it didn't -- 

This test (`test_launch_request_while_disabled_error`) is moved into the status_next validator tests, and `test_end_enrolment_request_while_disabled_error` and `test_end_experiment_request_while_disabled_error` have not changed.

eliserichards:
This PR tries to cover all possible transitions for `status`, `publish_status`, and `status_next` for the v5 serializer. Do you think I should even be testing things like `WAITING` since those aren't handled by the serializer at all? @jaredlockhart

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
```diff
@@ -0,0 +1,184 @@
...
+            experiment,
+            data={
+                "status": NimbusExperiment.Status.LIVE,
+                "publish_status": NimbusExperiment.PublishStatus.DIRTY,
+                "status_next": NimbusExperiment.Status.LIVE,
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer.is_valid())
```
eliserichards, line 167: This will be `False` once we add validation (#8012)

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
```diff
@@ -0,0 +1,184 @@
...
+            [
+                NimbusExperiment.Status.DRAFT,
+                True,
+            ],
+            [
+                NimbusExperiment.Status.PREVIEW,
+                True,
+            ],
+            [
+                NimbusExperiment.Status.LIVE,
```
yashika-khurana, line 50: is it possible `status as Live` and `status_next` as live too if its an experiment????

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
```diff
@@ -0,0 +1,184 @@
...
+        experiment = NimbusExperimentFactory.create(status=status)
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "status_next": NimbusExperiment.Status.LIVE,
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertEquals(serializer.is_valid(), valid)
```
yashika-khurana, line 69: its worth testing the error message too as you did in first test case 🖖

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
yashika-khurana: makes clear sense

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_status_next_transition_validator.py
```diff
@@ -0,0 +1,184 @@
...
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            lifecycle=NimbusExperimentFactory.Lifecycles.LAUNCH_APPROVE_APPROVE,
+            is_rollout=True,
+        )
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "status": NimbusExperiment.Status.LIVE,
+                "publish_status": NimbusExperiment.PublishStatus.DIRTY,
+                "status_next": NimbusExperiment.Status.LIVE,
```
yashika-khurana, line 179: here I understand the transition for live as status to live as status_next for the rollout

yashika-khurana (changes requested):
thank you @eliserichards for doing this 👏 , its much much stronger now 💪 🚀  , just have couple of suggestions

jaredlockhart (changes requested):
Ah okay I see what you're doing here but I have a couple thoughts.  So we have the transition validator here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L623

and then we're using it for `status` and `publish_status` here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L698-L715

but we're not using it for status_next, we're just doing that with a validation method here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/api/v5/serializers.py#L879-L891


So the way I generally try to setup the test files and classes is to be one to one with the classes they're testing, so here we're referencing `status_next_transition_validator` but no such class `StatusNextTransitionValidator` exists, it's just methods on the `NimbusExperimentSerializer` class and so I would expect those tests to live in the corresponding test file.

It looks to me like we should be able to use the same StatusTransitionValidator class for `status_next` to since the transitions are formatted the same way, so I think another way to approach this is to change the `status_next` field on `NimbusExperimentSerializer` to use the transition validator the same way `status` and `publish_status` do, and then we can remove that `validate_status_next` method, and then you can fold the `status_next` tests into the `test_status_transition_validator_mixin` tests?  That should make things a little more reusable and consistent?

What do you think?

jaredlockhart:
The other thing I try to be careful of is not testing the same logic multiple times in different places, so we're already testing the status next transitions in the NimbusExperimentSerializer tests and the NimbusStatusTransitionValidator tests, so try to keep an eye out for where something's being tested so you don't accidentally re-test the same thing twice.

jaredlockhart (commented):
Also it's kind of tough to reason about a PR that introduces tests for existing logic without any logic changes.  We're already at 100% coverage and tests are all passing.  This doesn't introduce any logic changes I don't think so it's tough to know what these should be covering.  Maybe it would make more sense to couple the new tests with the change to allow dirty edits in the status_next transitions constant?

eliserichards:
That makes sense to me @jaredlockhart. I was pretty unsure about this as I was doing it, since _yes_, we are testing `status_next` and it's covered by our tests, it's also not testing the full breadth of possible combinations. I like the idea of rolling this into #8012.. I think that'll make this more clear 👍

PR #8060, opened 2022-12-14
fix #8055 refactor(nimbus): refactor visualization schema to include versioning

### LINKED ISSUE #8055: Improve ease of changing visualization data schema
Add a top level versioning key to the visualization data schema, allowing us to more easily have old and new schemas side by side.

Before:
```JSON
{
  daily: {},
  weekly: {},
  overal: {},
  // etc
}
```

After:
```JSON
{
  v3: {
    daily: {},
    weekly: {},
    overal: {},
    // etc
  }
}
```

This allows us to simultaneously and discretely add a schema change without disrupting the UI or existing schema. We can simply have them both exist side by side, and flip the API return statement to use the new version when UI is ready.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3002)

mikewilli:
Because

* we want to make schema changes without breaking other code/UI

This commit

* adds a top-level key to the visualization schema to allow us to have different schemas side-by-side


yashika-khurana (commented):
I was just going through all the PRs and have one question. From what I understood is that you are adding a new key `current` and reassigning the existing value to that key, but I am not sure what's the benefit of doing that or why we are doing this. My apologies in advance if its a stupid question 😆

mikewilli:
No problem, thanks for looking! You're correct on what this PR does. The reasoning is that this allows us to easily have side-by-side schema versions for intermediate changes. Without something like this, we would need to make schema changes together with UI changes and data migration in one giant PR (see #8043 -- this ticket is a refactor that is part of splitting that PR into multiple smaller discrete PRs).

jaredlockhart (changes requested):
Oh yes this is correct but, `current` suggests to me that that would be reused since there's always some current schema, I think you'd want some absolute identifier that you'd use now and then throw away later?  Like a version number or word that describes the current structure?

mikewilli:
I was picturing `current` always being the in-use schema, and that would allow for `old` or `new` or whatever to be a deprecated or not-yet-adopted version on a rolling basis.

That said, I don't feel strongly about it -- I'll update to use `v1` and `v2`.

yashika-khurana:
@mikewilli thank you for explaining that, it makes sense now :)

app/experimenter/jetstream/client.py
```diff
@@ -217,4 +217,7 @@ def get_experiment_data(experiment):
 
     experiment_data["errors"] = errors_by_metric
 
-    return experiment_data
+    ret = {}
+    ret["v1"] = experiment_data
+
+    return ret
```
jaredlockhart, line 223: What's wrong with `return {'v1': experiment_data}`?

app/experimenter/visualization/tests/api/test_views.py
```diff
@@ -36,6 +36,16 @@ def test_analysis_results_view_200(self, lifecycle, mock_exists):
...
 
+        # test with results_data object
+        experiment.results_data = {"v1": {}}
+        experiment.save()
+
+        response = self.client.get(
+            reverse("visualization-analysis-data", kwargs={"slug": experiment.slug}),
+            **{settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
```
jaredlockhart, line 47: This doesn't need to be a separate step, just set the results data on the create and assert the contents after the response code?

jaredlockhart (changes requested):
Crazy that api test only ever asserted the status code but not the contents, a view test like that should always do both.  Update the test and this should be good to go.

app/experimenter/jetstream/client.py
mikewilli: Nothing! I'll do that. The answer for how I got here is that I started with the ret object containing v1 and v2 and it was declared elsewhere. Realized that should happen in a future PR and didn't think about how this didn't make sense anymore :)

app/experimenter/visualization/tests/api/test_views.py
mikewilli: I left it in to test the `None` case, but updated now to check the contents for both.

jaredlockhart (approved):
Awesome looks good thank you! 🎉

app/experimenter/visualization/tests/api/test_views.py
jaredlockhart: I still would've made those two separate tests, but not a big deal 👍

jaredlockhart:
Oh it just occurred to me we could've made that `"v1"` a constant somewhere like in settings.py but it's not a big deal if we need to we can do it whenever.

PR #8062, opened 2022-12-15
fix #7571 feat(nimbus): add analysis_bases to analysis results data

### LINKED ISSUE #7571: Results grouped by exposed segments

mikewilli:
Because

* Jetstream can compute exposure events analysis
* Experimenter does not account for exposures analysis_basis

This commit
* adds analysis_basis to the visualization schema (alongside the existing schema)

### DIFF
--- a/app/experimenter/jetstream/client.py
+++ b/app/experimenter/jetstream/client.py
@@ -8,6 +8,7 @@ from django.core.files.storage import default_storage
 
 from experimenter.jetstream.models import (
     METRIC_GROUP,
+    AnalysisBasis,
     Group,
     JetstreamData,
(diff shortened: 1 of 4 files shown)


jaredlockhart (approved):
Yeah you know this makes it so easy to see that the v2 data is sitting healthily next to the v1 data and what changes are being made to the processing, etc.  This is great 👍 👍 👍

PR #8064, opened 2022-12-15
fix #8063 feat(nimbus): enable analysis bases in results page

### LINKED ISSUE #8063: Results page allows user to filter by analysis basis
Jetstream can compute exposures analysis_basis, and theoretically other ones in the future, and Experimenter accounts for this in the parsing of Jetstream results, so we should allow these to be displayed.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3005)

mikewilli:
Because

* Jetstream can perform exposures analysis
* we have prepared the Jetstream client to handle different analysis bases in the viz data

This commit
* uses the new schema with analysis_bases in the Results page to allow users to select exposures if available


app/experimenter/experiments/migrations/0224_analysis_add_basis.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 3.2.15 on 2022-10-13 17:50
+
+import copy
+
+from django.db import migrations
+
+
+def analysis_add_basis(apps, schema_editor):
```
jaredlockhart, line 8: Wait shouldn't this have been done in a previous commit?

mikewilli: Sure, that should work. Right now (without this PR), the `v2` schema is optional and not used. This PR would make it required and used, so I included the data migration here to enforce it. Doesn't hurt to split out the data migration first though, I can do that.

mikewilli:
I'll split this PR out per comment from @jaredlockhart. Step one is the data migration [here](https://github.com/mozilla/experimenter/pull/8069)

PR #8070, opened 2022-12-16
fix #8063 feat(nimbus): enable analysis bases in results page

### LINKED ISSUE #8063: Results page allows user to filter by analysis basis

mikewilli:
Because

* Jetstream is able to calculate exposures analysis for configured experiments

This commit

* allows the user to view results by exposures only for experiments with these results available

Screenshot from Results page:
image

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
@@ -47,8 +47,8 @@ const Subject = ({
     (withAnalysis
       ? {
           show_analysis: true,
-          daily: { all: [] },
-          weekly: { all: {} },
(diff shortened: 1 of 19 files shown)


jaredlockhart (approved):
Okay all looks right to me.  It'll be nice when we can finally factor out those test mocks into something less noisy.  Tested locally with the migrated prod data and it looked okay even though it didn't have any exposure data but at least the existing pages should be fine.  It'll be exciting when something with exposures comes through and we can see it in action.  Great job @mikewilli 🎉 🎉 🎉  and thank you for splitting this up, makes it much more manageable to reason about 🙏 🙏 🙏

PR #8073, opened 2022-12-16
Add new advanced desktop targeting for new users with infrequent use

lmegviar:
This new targeting will support out follow up import bookmarks experiments. The goal is to target users who used Firefox 6 or fewer times over the past 28 days and who have a profile age of less than 28 days.

PR #8076, opened 2022-12-19
Update new users with infrequent use targeting profile age value

lmegviar:
See [original PR here](https://github.com/mozilla/experimenter/pull/8073)

This targeting has not yet been used and was newly added for an upcoming experiment. The original PR incorrectly used the `PROFILE28DAYS` variable. What we actually want is to target users with profile ages of _less_ than 28 days.


app/experimenter/targeting/constants.py
```diff
@@ -464,8 +464,11 @@ def __post_init__(self):
 NEW_USERS_WITH_INFREQUENT_USE = NimbusTargetingConfig(
     name="New users with infrequent use",
     slug="new_users_with_infrequent_use",
-    description="0 - 6 days activity in past 28 days and profile age <= 28 days",
-    targeting=(f"{PROFILE28DAYS} " "&& userMonthlyActivity|length <= 6"),
+    description="0 - 6 days activity in past 28 days and profile age < 28 days",
+    targeting=(
+        "(currentDate|date - profileAgeCreated|date) / 86400000 < 28 "
```
yashika-khurana, line 469: I would suggest moving this calculation to a variable like `PROFILELESSTHAN28DAYS`

yashika-khurana (changes requested):
Thank you for the clear description, I just have one suggestion, other than that it looks good!

lmegviar:
Good call @yashika-khurana, I've moved that logic into a variable.

yashika-khurana (approved):
thank you @lmegviar looks perfect 💯

PR #8077, opened 2022-12-19
fixes #8045 tests(nimbus): Add an integration test for sticky targeting

### LINKED ISSUE #8045: Integration test for Sticky Targeting
Let’s write an e2e integration test for sticky targeting

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2998)

jrbenny35:
Because:
* Sticky targeting is alive and well within nimbus

This Commit:
* Ensures we have proper automated testing of it.


app/experimenter/targeting/constants.py
```diff
@@ -825,6 +825,17 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+TEST_STICKY_TARGETING = NimbusTargetingConfig(
+    name="Test targeting",
+    slug="test_targeting",
+    description="Config for sticky targeting",
+    targeting="'sticky.targeting.test.pref'|preferenceValue",
+    desktop_telemetry="",
+    sticky_required=False,
```
eliserichards, line 834: Why wouldn't this be `sticky_required=True`?

eliserichards (commented):
Looks good! One tiny question for you :D

app/experimenter/targeting/constants.py
jrbenny35: So in the test I set the sticky field itself, but seeing as this is a sticky targeting pref, it should be set here too I guess. I can change it

eliserichards: Ok sweet. I wasn't sure if that would be a separate test or something 👍

Issue #7189, opened 2022-04-07
Deprecate storybook


jaredlockhart:
Our team structure and workflow has changed greatly, and storybook is no longer being used by anyone to validate and review the frontend.  In addition, it creates a large overhead for changes in the frontend.  Rather than maintaining storybook as a resource, we can divert the attention and effort that would have gone into it towards maintaining and expanding our fantastic documentation infrastructure which is providing tremendous value to our users.  Let's deprecate storybook altogether.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2363)

PR #8089, opened 2022-12-19
No issue chore(nimbus): Cleaning up some of the leftover storybook lines

eliserichards:
Because...

* We removed storybook a while ago (#7189)
* And we still had some `storiesOf`'s hanging around

This commit...

* Removes the remaining references to `storiesOf`


eliserichards:
I think this one is safe to land before code freeze @yashika-khurana 😋

yashika-khurana (approved):
absolutely @eliserichards 💯 👏

jaredlockhart (approved):
Ah good catch!

PR #8094, opened 2022-12-20
Add new targeting and update new user with infrequent user targeting

lmegviar:
This patch updates an existing targeting expression that was added for a specific experiment set which has not been run yet (so should be safe to update). `NEW_USER_WITH_INFREQUENT_USE` now includes a clause of having a profile age greater than 7 days. I also changed the name to use "user" instead of "users" to be consistent with other targeting names.

I also added new targeting, `NEW_WITH_INFREQUENT_USE_FIVE_BOOKMARKS`, which builds on the former, but adds clauses related to bookmark count and OS.

These will support the import bookmarks follow up experiments:

- https://mozilla-hub.atlassian.net/browse/OMC-283
- https://mozilla-hub.atlassian.net/browse/OMC-330

### DIFF
--- a/app/experimenter/targeting/constants.py
+++ b/app/experimenter/targeting/constants.py
@@ -26,6 +26,7 @@ HAS_PIN = "!doesAppNeedPin"
 NEED_DEFAULT = "!isDefaultBrowser"
 PROFILE28DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 >= 28"
 PROFILELESSTHAN28DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 < 28"
+PROFILEMORETHAN7DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 > 7"
 NEW_PROFILE = "(currentDate|date - profileAgeCreated|date) / 3600000 <= 24"
 WIN1903 = "os.windowsBuildNumber >= 18362"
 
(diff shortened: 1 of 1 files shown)


app/experimenter/targeting/constants.py
```diff
@@ -462,11 +463,30 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
-NEW_USERS_WITH_INFREQUENT_USE = NimbusTargetingConfig(
-    name="New users with infrequent use",
-    slug="new_users_with_infrequent_use",
-    description="0 - 6 days activity in past 28 days and profile age < 28 days",
-    targeting=(f"{PROFILELESSTHAN28DAYS} " "&& userMonthlyActivity|length <= 6"),
```
yashika-khurana, line 469: I hope the above targeting is not referenced in any experiment??

yashika-khurana (changes requested):

app/experimenter/targeting/constants.py
yashika-khurana: oh, I just checked that there are 4 experiments in prod 3 are in preview and 1 in draft using that targeting, I would suggest sending all back to draft first and then merging this PR, and then selecting this new targeting.
Just for future reference, we avoid removing targeting as it can be referenced in old experiments.

yashika-khurana (approved):
thank you @lmegviar 🎉 , just make sure you send 3 preview experiments in draft and then merge this PR and select the new targeting :)

lmegviar:
Thanks, @yashika-khurana! We'll be updating those experiments to use the new `NEW_WITH_INFREQUENT_USE_FIVE_BOOKMARKS` targeting.

lmegviar:
Ah, sorry @yashika-khurana looks like I don't have merge access. Are you able to merge on your end?

yashika-khurana:
yeah sure to make sure it goes smoothly can you send preview experiments to draft first?
image


 I can merge after that

lmegviar:
> yeah sure to make sure it goes smoothly can you send preview experiments to draft first? image
> 
> I can merge after that

Those are back in draft state now. Thank you!

yashika-khurana:
awesome @lmegviar  thank you 🎉

lmegviar:
So sorry @yashika-khurana  - I just realized the slug and name for the new targeting weren't updated. I've filed [a follow-up PR to correct.](https://github.com/mozilla/experimenter/pull/8098/)

yashika-khurana:
@lmegviar no worries approved and merged

Issue #8095, opened 2022-12-20
Add ability to disable Audience fields


eliserichards:
Related to #8050 

We want to be able to "lock" the Audience page fields that are not editable, so that we can have the Audience page available for Live Rollouts without allowing edits to be made.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3018)

PR #8096, opened 2022-12-20
For #8095 feat(nimbus): Add ability to disable fields on Audience page

eliserichards:
Because...

* We are going to reuse the Audience page to allow Live rollouts to be edited
* And we want to disable fields on the Audience page that we don't want to be edited

This commit...

* Checks whether each field on the Audience page should be locked
* Adds tests for disabled fields

image

vs

image

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -24,6 +24,7 @@ import {
   NimbusExperimentApplicationEnum,
   NimbusExperimentChannelEnum,
   NimbusExperimentFirefoxVersionEnum,
+  NimbusExperimentStatusEnum,
 } from "../../../types/globalTypes";
 import { MOCK_EXPERIMENT, Subject } from "./mocks";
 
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -197,6 +198,8 @@ export const FormAudience = ({
 
   const isDesktop =
     experiment.application === NimbusExperimentApplicationEnum.DESKTOP;
+  const isLiveRollout = experiment.isRollout && getStatus(experiment).live;
+  const isLocked = isLiveRollout && !getStatus(experiment).draft;
```
jaredlockhart, line 202: Wait, we DO lock it if it's a rollout and it's live?  Shouldn't it be the other way around?  Also is the idea here that users can get back to this page, and see the fields being locked in some case where it's not a live rollout, ie a completed rollout or a live experiment?  I'm not sure that's the right approach?  Couldn't we set it up so that users can only navigate to the page in the case that it's a live rollout, and then they can get to the page altogether in the cases of

1) It's a draft of any type
2) It's a live rollout

and in both of those cases the fields are all unlocked, and in all other cases they just can't get to the page altogether?

jaredlockhart: Oh I see this is meant as a no-op change and the additional logic to expose it for rollouts will come in a followup.  Sounds good!

eliserichards: Yes, exactly :) The PRs following this will ensure that only the correct fields are unlocked for live rollouts

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -1227,6 +1228,74 @@ describe("FormAudience", () => {
...
+  );
+  // testing one checkbox, one dropdown, and one text field
+  expect(screen.getByTestId("isSticky")).toBeDisabled();
+  expect(screen.getByTestId("firefoxMinVersion")).toBeDisabled();
+  expect(
+    within(
+      screen.queryByTestId("population-percent-top-row") as HTMLElement,
+    ).getByTestId("population-percent-text"),
+  ).toBeDisabled();
+});
```
yashika-khurana, line 1263: I think it would be worth including fields which will not be disabled along with disabled fields 🙈

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -1227,6 +1228,74 @@ describe("FormAudience", () => {
...
+    />,
+  );
+  // testing one checkbox, one dropdown, and one text field
+  expect(screen.getByTestId("isSticky")).toBeEnabled();
+  expect(screen.getByTestId("firefoxMinVersion")).toBeEnabled();
+  expect(
+    within(
+      screen.queryByTestId("population-percent-top-row") as HTMLElement,
+    ).getByTestId("population-percent-text"),
+  ).toBeEnabled();
```
yashika-khurana, line 1296: 🎉

yashika-khurana (approved):
@eliserichards thank you 👏 , test cases are awesome just would like to have a combo of the test cases in which disabled and enabled fields are together so that in case we want to have some fields locked and some not. Apart from that ✅

app/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
eliserichards: that's a good point! I will add that 👍

eliserichards: Actually I think that can be added in the next part, cuz I will be adding the unlocked fields for rollouts

Issue #8107, opened 2023-01-03
Expose behavioral targeting in Experimenter


jeddai:
Add advanced targeting options to Experimenter for the various behavioral targeting transforms. These advanced targeting options should be available for both Firefox iOS and Fenix.

- Core Active Users - Users who have been active 21 out of the last 28 days
  - This event is app_cycle.foreground in Firefox iOS, and events.app_opened in Fenix.
- Recently Logged In Users - Users who have completed a Sync login within the last 12 weeks

[EXP-2903](https://mozilla-hub.atlassian.net/browse/EXP-2903)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3037)

PR #8108, opened 2023-01-03
For #8107 feat(nimbus): add advanced targeting options for mobile behavioral targeting

jeddai:
Because...

    The behavioral event store now exists within Firefox iOS and Android

This commit...

    Adds advanced targeting options that leverage the events being recorded.

**A note:** We need two distinct targeting configs for each behavior, one for iOS and one for Android. These are required because the same event is named differently across the two apps. Something I noticed is Experimenter seems to be using the targeting config `slug` field split+capitalized for displaying the config, and I don't think we'd want "Android" and "Ios" showing up in the list. If we do, something worth noting is that "Ios" should be "iOS".

iOS set as the app
image

Android set as the app
image

### DIFF
--- a/app/experimenter/targeting/constants.py
+++ b/app/experimenter/targeting/constants.py
@@ -29,6 +29,8 @@ PROFILELESSTHAN28DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 
 PROFILEMORETHAN7DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 > 7"
 NEW_PROFILE = "(currentDate|date - profileAgeCreated|date) / 3600000 <= 24"
 WIN1903 = "os.windowsBuildNumber >= 18362"
+CORE_ACTIVE_USERS_TARGETING = "'{event}'|eventCountNonZero('Days', 28, 0) >= 21"
+RECENTLY_LOGGED_IN_USERS_TARGETING = "'{event}'|eventCountNonZero('Weeks', 12, 0) >= 1"
 
 NO_TARGETING = NimbusTargetingConfig(
     name="No Targeting",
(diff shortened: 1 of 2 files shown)


app/experimenter/targeting/constants.py
```diff
@@ -868,6 +868,56 @@ def __post_init__(self):
     application_choice_names=[a.name for a in Application],
 )
 
+CORE_ACTIVE_USERS_TARGETING = "'{event}'|eventsCountNonZero('Days', 28, 0) >= 21"
```
yashika-khurana, line 871: nitpick: Can you move these constants to the top 🆙

app/experimenter/targeting/constants.py
```diff
@@ -868,6 +868,56 @@ def __post_init__(self):
...
+    slug="ios_core_active_users",
+    description="Users who have been active at least 21 out of the last 28 days",
+    targeting=CORE_ACTIVE_USERS_TARGETING.format(event="app_cycle.foreground"),
+    desktop_telemetry="",
+    sticky_required=True,
+    is_first_run_required=False,
+    application_choice_names=(Application.IOS.name, Application.FOCUS_IOS.name),
+)
+
+RECENTLY_LOGGED_IN_USERS_TARGETING = "'{event}'|eventsCountNonZero('Weeks', 12, 0) >= 1"
```
yashika-khurana, line 895: same as before to move this 🆙

yashika-khurana (changes requested):
Meanwhile , just a nit pick

app/tests/integration/nimbus/conftest.py
```diff
@@ -141,7 +141,7 @@ def experiment_url(base_url, default_data, slugify):
 
 @pytest.fixture
 def experiment_name(request):
-    return f"{request.node.name[:75]}-{str(uuid.uuid4())[:4]}"
+    return f"{request.node.name[:75]}{str(uuid.uuid4())[:4]}"
```
jrbenny35, line 144: 1 char too much

jrbenny35 (approved):

app/tests/integration/nimbus/conftest.py
yashika-khurana: 😆

yashika-khurana (approved):

PR #8127, opened 2023-01-09
fix #7810 feat(nimbus): Sort versions

### LINKED ISSUE #7810: Sort versions from high to low on Min/Max Version dropdown on Audience page
Related to #6332 

image

Change the sorting of Min Version and Max Version dropdown so that the most recent (highest) version number is at the top, keeping "No version" as the first item in the list.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2833)

yashika-khurana:
Because

* Versions dropdown on the landing page and on the Audience page shows versions from firefox 11, firefox 12, and so on, and in the end shows no versions.

This commit

* Sort the versions list and return the result as `NO VERSION` as the top result and the highest firefox versions....till firefox 11 at the end.
Screenshot 2023-01-09 at 12 32 45 PM
Screenshot 2023-01-09 at 12 32 55 PM

fixes #7810


app/experimenter/experiments/api/v5/types.py
```diff
@@ -309,13 +309,34 @@ def resolve_all_feature_configs(root, info):
...
+            NimbusLabelValueType(
+                label=name.label,
+                value=name.label.replace(" ", "_").upper(),
+            )
+            for name in sorted_versions
+        ]
+        no_version_option = sorted_versions.pop()
+        sorted_versions.insert(0, no_version_option)
+
+        return sorted_versions
```
jaredlockhart, line 338: 1) Why isn't this using the [existing parsing method](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/constants.py#L243-L245) that lets you sort by semantic versioning?

2) Why do we need to sort this list at all?  The list is static and already ordered, if all you want to do is reverse it you can just `[::-1]`?

jaredlockhart (changes requested):

app/experimenter/experiments/api/v5/types.py
yashika-khurana: yeah for point 1, I could have used the parse method, totally missed that. For point 2, options are already ordered, reversing it is a quick option, I was overthinking if the list is not sorted in the future. I will do the second point reversing list and add no version as the first option, thanks @jaredlockhart

eliserichards (approved):
LGTM! 🚢

app/experimenter/experiments/api/v5/types.py
```diff
@@ -309,13 +309,28 @@ def resolve_all_feature_configs(root, info):
...
```
jaredlockhart, line 332: Very nice 👍

PR #8132, opened 2023-01-10
fix #8130 feat(nimbus): Remove enrollment date to be cloned

### LINKED ISSUE #8130: Remove enrollmentEndDate from clone method
We recently added an `_enrollmentEndDate` property to the NimbusExperiment model but neglected to exclude it from the clone method [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models.py#L703-L704).  Let's make sure to also set that field to `None` in the clone method.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3048)

yashika-khurana:
Because

* We recently added an` _enrollment_end_date `property to the NimbusExperiment model but neglected to exclude it from the clone method [here](https://github.com/mozilla/experimenter/blob/main/app/experimenter/experiments/models.py#L703-L704). 

This commit

* Set the` _enrollment_end_date = None `in the clone method and test case to support changes

fixes #8130

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -702,6 +702,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         cloned.conclusion_recommendation = None
         cloned._start_date = None
         cloned._end_date = None
+        cloned._enrollment_end_date = None
         cloned.save()
 
         if rollout_branch_slug:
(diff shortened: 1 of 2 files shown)


eliserichards (approved):
Perfect! 🎉

jaredlockhart:
Thank you! 🙏

PR #8087, opened 2022-12-19
For #8012 refactor(nimbus): Move transition constants into serializer

eliserichards:
📢 Requires #8081 and #8086 to be merged first

This is a small refactor to move some constants for status and publish status out of the `NimbusExperiments` constants.py file into the v5 serializer where they are being used.

This commit...

* Creates `TransitionConstants` class in the v5 serializer
* Moves the following constants into the class:
   * `VALID_STATUS_TRANSITIONS`
   * `VALID_STATUS_NEXT_VALUES`
   * `VALID_PUBLISH_STATUS_TRANSITIONS`
   * `STATUS_ALLOWS_UPDATE`
   * `PUBLISH_STATUS_ALLOWS_UPDATE`
   * `STATUS_UPDATE_EXEMPT_FIELDS`

PR #8086, opened 2022-12-19
For #7983 feat(nimbus) test(nimbus): Serializer validation for publish status and tests

eliserichards:
📢 Requires #8081 to be merged first 

Because...

* We are allowing `LIVE` rollouts to have `DIRTY` publish status and go through review

This commit...

* Checks for `DIRTY` status when the v5 serializer validates `publish_status`
* Adds some tests to the serializer: expanding what we already have to include tests for both experiments and rollouts, and adding some tests for dirty publish status:
   * `[...]publish_status_doesnt_invoke_push_task` - for rollouts and experiments
   * `test_rollout_can_request_review_from_live`
   * `test_rollout_can_request_review_from_live_dirty`
   * `test_allow_live_rollout_to_be_dirty`
   * `test_cant_review_for_requesting_user_dirty`
   * `test_can_review_for_requesting_user_when_idle` - for rollouts and experiments
   * `test_can_review_for_requesting_user_when_dirty` - for rollouts and experiments

Issue #8050, opened 2022-12-14
Add Audience page to sidebar for Live rollouts


data-sync-user:
We are adding the `dirty` state on the backend to signify when a rollout is `Live` and has unpublished changes (see EXP-2563). 

We want to allow users to edit the population percent while their rollouts are live.

* Add an “Audience” item to the sidebar (just like in Draft)
* When clicked, everything on the Audience page should be disabled 
* Save button should be disabled
* Save and Continue should be disabled

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2954)

PR #8136, opened 2023-01-11
For #8050 feat(nimbus): Add Audience page to sidebar

eliserichards:
📢 Requires #8087 and #8086 to be merged first

Because...
* We want to reuse the Audience page to allow users to edit rollouts while live

This commit...
* Adds the Audience page to the sidebar **for both rollouts and experiments**
* **Hides** the audience link for experiments, **enables** it for rollouts
* Because of #8096, the fields on the audience page are **disabled** 
   * we will enable them once we implement the ability to send the edits through review - #7823

---- 

Rollout with fully disabled audience page:
image

Experiment with no audience link in sidebar: 
   
image

image

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/App/ExperimentRoot/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/App/ExperimentRoot/index.tsx
@@ -74,15 +74,17 @@ export const ExperimentRoot = ({
   // after experiment has been loaded
   const useRedirectCondition = (redirect: RedirectCondition) => {
     let redirectPath: string | undefined, redirectResult: string | void;
-    if (
-      !loading &&
-      status &&
-      (redirectResult = redirect!({ status, analysis })) != null
-    ) {
-      redirectResult = redirectResult.length
-        ? `/${redirectResult}`
-        : redirectResult;
-      redirectPath = `${BASE_PATH}/${slug}${redirectResult}`;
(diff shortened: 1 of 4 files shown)


eliserichards:
Actually, I'm gonna close this for now until the other ones get merged 🤞

jaredlockhart (dismissed):
Looks good, where's tests? 🤔

yashika-khurana:
> Looks good, where's tests? 🤔

Interesting and code coverage is 100% 🤔

yashika-khurana:
what about not showing the Audience option at all for the experiments instead of disabling it (like that we have now) and only showing the Audience option for the rollouts? @jaredlockhart @eliserichards

jaredlockhart:
> what about not showing the Audience option at all for the experiments instead of disabling it (like that we have now) and only showing the Audience option for the rollouts? @jaredlockhart @eliserichards

Yeah I think that's a good option too, since experiments should never be able to get to it.

eliserichards:
Makes sense to me. I wasn't sure if we'd want to see Audience from an experiment in the future, but I changed it so it's only shown for rollouts now 😄 I also added some tests. It's kinda weird that code cov wasn't complaining 🤔

yashika-khurana (approved):
tested locally works awesome, audience option is gone for experimenter, thank you for adding test as well, 🎉

PR #8155, opened 2023-01-16
fixes #8148 feat(nimbus): Change targeting from installed engines to …

### LINKED ISSUE #8148: Modify additional targeting based on only certain search engines to instead target based on certain current default engines
The experiment: https://experimenter.services.mozilla.com/nimbus/persist-search-term-staged-rollout-phase-1/edit/audience

[The request](https://github.com/mozilla/experimenter/issues/8141) resulted in creating a condition for users who have one of DDG/Bing/Google installed, but the experiment should only affect users who have either Bing, DDG, or Google as their current default search engine. Thus, we should just check the current engine in `searchEngines` against search engine telemetry id names.

I'll submit a PR with the change.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3063)

jamesteow:
…current default engine

Because...

* The original implementation checked the client had Google, Bing, and DDG installed whereas the experiment wants to only apply to users with either of those search engines as the current default search engine.

This commit...

* Changes the targeting to default engines matching Google, Bing, and DDG.
* Google can have variety of names based off of telemetryId's hence the use of "in".
* Changes descriptors and variable name to match the behaviour.

### DIFF
--- a/app/experimenter/targeting/constants.py
+++ b/app/experimenter/targeting/constants.py
@@ -927,14 +927,14 @@ IOS_RECENTLY_LOGGED_IN_USER = NimbusTargetingConfig(
     application_choice_names=(Application.IOS.name, Application.FOCUS_IOS.name),
 )
 
-HAS_GOOGLE_BING_DDG_SEARCH_ENGINE = NimbusTargetingConfig(
-    name="Has Google, Bing, or DuckDuckGo search engine",
-    slug="has_search_engine_google_bing_or_ddg",
-    description="Users who have the Google, Bing, or DuckDuckGo search engines installed",
+HAS_GOOGLE_BING_DDG_AS_CURRENT_DEFAULT_SEARCH_ENGINE = NimbusTargetingConfig(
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Looks good, thank you 👍

jamesteow:
Thanks for the approval @jaredlockhart, not sure why the tests are failing on CI. 

As far I can tell (not at an expert of Docker or this repo) by reading the errors, both **Test SDK Targeting (Release Firefox)** and **Test Focus Android and Remote Settings (Release Firefox)** had connection resets when accessing 34.205.13.154, which a reverse look up is Amazon.

Do you think this was a random network error solvable by re-running the tests? I can't imagine how my proposed modifications to the existing code could've resulted in those tests resulting in a network error.

eliserichards:
Very weird 🤔 I just tried re-running them. Let's see if it fails again...

jaredlockhart:
@jamesteow @eliserichards Yeah we're seeing a lot of intermittent network failures with Dockerhub, not related to the contents of this (or other) PRs.

PR #8159, opened 2023-01-17
fix #8072 feat(nimbus): Record update datetime for the experiments

### LINKED ISSUE #8072: Default sort V5 by last updated date
Let's cache a last updated date right on the Nimbus model and then default sort the list being returned by that. The default ordering right now is oldest first and it's very annoying.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3010)

yashika-khurana:
Because

* The default ordering for the experiments on the landing page is oldest first and but we want the latest updated experiments to be ranked first so that more edits or currently working experiments will be on top 🔝 

This commit

* Adds the new field in Nimbus Model `updated`, which will record the last updated datetime of the experiments (as the field is `auto_now`).
* Sorts the result based on the updated datetime
* Test cases to verify new changes

Note`As currently implemented, setting auto_now or auto_now_add to True will cause the field to have editable=False and blank=True set.` ( [find more details here](https://docs.djangoproject.com/en/4.1/ref/models/fields/)) and updated field will be initialized with current datetime
Migration logs: 
image

fixes #8072


app/experimenter/experiments/models.py
```diff
@@ -196,6 +198,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         null=True,
     )
     takeaways_summary = models.TextField(blank=True, null=True)
+    updated = models.DateTimeField(auto_now=True)
```
jaredlockhart, line 201: Oh can you please follow the same naming pattern as the other cached dates below ie `_updated_date` 🙏

app/experimenter/experiments/migrations/0225_nimbusexperiment_updated.py
```diff
@@ -0,0 +1,18 @@
...
+        ("experiments", "0224_analysis_add_basis"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="nimbusexperiment",
+            name="updated",
+            field=models.DateTimeField(auto_now=True),
+        ),
+    ]
```
jaredlockhart, line 18: Do you want to pre populate it with the datetime of the most recent changelog?

jaredlockhart (dismissed):
Ty, couple thoughts 🙏

app/experimenter/experiments/migrations/0225_nimbusexperiment_updated.py
yashika-khurana: the thing is that the option `auto_now and default` are mutually exclusive. we might not be able to provide a default value.

yashika-khurana: ❗ ❗ Added migration file to prepopulate `_updated_date_time` from the changelogs

app/experimenter/experiments/models.py
```diff
@@ -196,6 +198,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         null=True,
     )
     takeaways_summary = models.TextField(blank=True, null=True)
+    _updated_date_time = models.DateTimeField(auto_now=True)
```
eliserichards, line 201: Why does this one have an underscore at the beginning? Should it be `updated_date_time` to match the others?

eliserichards: oh I see the other date fields are formatted like this too 🤔

eliserichards (approved):
This looks good to me! I'm curious why there's a different format for the date fields vs others, but everything else is 💯

app/experimenter/experiments/migrations/0226_alter_nimbusexperiment_updated_date_time.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 3.2.14 on 2022-07-12 21:19
```
jaredlockhart, line 1: You should do this in one migration pls 🙏

app/experimenter/experiments/migrations/0226_alter_nimbusexperiment_updated_date_time.py
```diff
@@ -0,0 +1,31 @@
...
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+    for experiment in NimbusExperiment.objects.all():
+
+        # disabling auto_now
+        experiment._meta.get_field("_updated_date_time").auto_now = False
+
+        changes = experiment.changes.all()
+        if changes:
+
+            last_changed_on = changes.order_by("-changed_on")[0]
```
jaredlockhart, line 16:
```py
experiment.changes.all().order_by('-changed_on').first()
```

app/experimenter/experiments/models.py
```diff
@@ -61,8 +61,10 @@ def latest_changed(self):
         )
 
     def with_owner_features(self):
-        return self.get_queryset().prefetch_related(
-            "owner", "feature_configs", "projects"
+        return (
+            self.get_queryset()
+            .prefetch_related("owner", "feature_configs", "projects")
+            .order_by("-_updated_date_time")
```
jaredlockhart, line 67: 🎉

app/experimenter/experiments/models.py
jaredlockhart: A single underscore is a Python convention for saying 'this is kinda private'

jaredlockhart (changes requested):
Looks good just a couple small points 🙏

app/experimenter/experiments/models.py
yashika-khurana: its private 🔒

jaredlockhart: Oh I thought I mentioned moving it down with the other fields.  Not a huge deal, just nice to keep things grouped together.

app/experimenter/experiments/tests/test_changelog_utils.py
```diff
@@ -30,14 +32,20 @@ def setUpClass(cls):
 
     def test_outputs_expected_schema_for_empty_experiment(self):
         owner = UserFactory.create()
-        experiment = NimbusExperiment.objects.create(owner=owner)
+        updated_time = timezone.datetime(
+            year=2022, month=1, day=2, hour=0, minute=0, second=0
+        )
+        with mock.patch("django.utils.timezone.now") as mock_now:
+            mock_now.return_value = updated_time
+            experiment = NimbusExperiment.objects.create(owner=owner)
```
jaredlockhart, line 40: Oh sorry I wasn't focusing on this earlier, I don't think you need to go through all the trouble of mocking out now, you can just use `experiment._updated_date_time.strftime(....)` for whatever the value is here.  We're not explicitly testing the saving behaviour here, just that whatever is in the db is serialized.

jaredlockhart (approved):
Yes all looks good and correct, thank you.  I don't think you need to bother with the mocks in the tests but up to you.  Thnx @yashika-khurana this will definitely make the UI easier to grok 🙏 🙏 🙏

app/experimenter/experiments/models.py
jaredlockhart: @eliserichards 

https://peps.python.org/pep-0008/#descriptive-naming-styles

```
_single_leading_underscore: weak “internal use” indicator. E.g. from M import * does not import objects whose names start with an underscore.
```

app/experimenter/experiments/tests/test_changelog_utils.py
yashika-khurana: oh so if I don't add the field we are getting that field in the test case, so that's why I mocked the field

PR #8170, opened 2023-01-19
fix #8168 test(nimbus): Add better rerun capability to integration tests

### LINKED ISSUE #8168: Fix enrollment integration test rerun capability
Rerunning the e2e integration tests sometimes doesn’t work due to the previous experiment still existing in the database and there is a chance that experiment enrolls and not the newly created, experiment. We should either end enrollment or clear the remote settings db.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3072)

jrbenny35:
Because:
* We have implemented reruns with pytest-sentry to help track intermittent failures.

This commit:
* Adds a clean up step to each test run so that the remote settings server is clean before each test run.

This mainly has to do with the enrollment tests.

### DIFF
--- a/app/tests/integration/nimbus/conftest.py
+++ b/app/tests/integration/nimbus/conftest.py
@@ -93,6 +93,7 @@ def firefox_options(firefox_options):
     firefox_options.log.level = "trace"
     return firefox_options
 
+
 @pytest.fixture
 def selenium(selenium, experiment_name, kinto_client, base_url, slugify):
     yield selenium
@@ -108,8 +109,6 @@ def selenium(selenium, experiment_name, kinto_client, base_url, slugify):
         summary.wait_for_complete_status()
     except Exception as e:
         raise e
-    else:
-        pass
 
 
 @pytest.fixture(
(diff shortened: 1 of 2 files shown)


yashika-khurana:
thanks @jrbenny35 this would be really great to have as we have seen this error often.

app/tests/integration/nimbus/conftest.py
```diff
@@ -94,6 +94,23 @@ def firefox_options(firefox_options):
...
+
+    experiment_slug = str(slugify(experiment_name))
+    try:
+        summary = SummaryPage(selenium, urljoin(base_url, experiment_slug)).open()
+        summary.end_and_approve()
+        kinto_client.approve()
+        summary = SummaryPage(selenium, urljoin(base_url, experiment_slug)).open()
+        summary.wait_for_complete_status()
+    except Exception:
+        pass
```
jaredlockhart, line 111: Oh so this will run on every experiment end, the name `selenium` means it'll override the existing selenium fixture so it'll automatically be included in every test, and it'll pull the experiment_name out of the other fixture so it'll arleady be the same for each test, and then we just explicitly end it....  So that tacks on the time to do the end on top of the length of each test run but cleans the experiment up at the end....

Yeah that's cool, let's try it and see if it helps.

jaredlockhart (approved):
Interesting, thanks for writing this up, and thank you for a very clear issue and clear description in the PR, makes this very easy to grok 💯 💯 💯 💯

PR #8175, opened 2023-01-19
fix #8174 bug(nimbus): properly parse retention data by analysis_basis 

### LINKED ISSUE #8174: Intermittent issue with Retention metric on Results page
Results page for some experiments shows “2-Week Browser Retention is not available”. Initial investigation points to this being a problem parsing analysis for experiments that have some exposures results. Investigation was only on Firefox Desktop experiments with Results in order to keep things as similar as possible to the original experiment where this was noticed.

Examples showing the error:
https://experimenter.services.mozilla.com/nimbus/import-infrequent-make-yourself-at-home/results
https://experimenter.services.mozilla.com/nimbus/first-run-easy-as-one-click-ver-1/results

Examples without the error:
https://experimenter.services.mozilla.com/nimbus/pocket-newtab-new-spoc-positions/results
https://experimenter.services.mozilla.com/nimbus/set-to-default-wording-infrequent-eco2206/results
https://experimenter.services.mozilla.com/nimbus/mr2-aboutwelcome-holdback/results

cc @danielkberry

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3075)

mikewilli:
Because

* it was noticed that some experiments were not showing overall retention metric results
* these experiments appeared to *have* overall retention results available

This commit

* fixes the parsing of overall retention results so they can be displayed properly


jaredlockhart (approved):
Awesome thank you 🎉 🎉 🎉

PR #8181, opened 2023-01-23
fix #8180 chore(project): apply sourcery to integration tests

### LINKED ISSUE #8180: Apply Sourcery to Integration Tests
We can also use Sourcery.ai to autoformat the integration tests, that'll keep them more in sync with the code style in the rest of the project now that we've applied it everywhere else.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3080)

jaredlockhart:
Because

* We recently applied the Sourcery.ai autoformatter to the bulk of the application code
* We can also apply it to the integration tests to keep them stylistically in sync with the rest of the code

This commit

* Applies Sourcery.ai to the integration tests


app/tests/integration/nimbus/jexl/__init__.py
```diff
@@ -44,9 +42,7 @@ def to_str(node):
     elif type(node) == BinaryExpression:
         return f"({to_str(node.left)} {node.operator.symbol} {to_str(node.right)})"
     elif type(node) == Transform:
-        args = ""
-        if node.args:
-            args = f'({", ".join([to_str(a) for a in node.args])})'
+        args = f'({", ".join([to_str(a) for a in node.args])})' if node.args else ""
```
jrbenny35, line 45: what the heck

app/tests/integration/nimbus/kinto/client.py
```diff
@@ -30,7 +30,7 @@ def _has_pending_review(self):
         return self._fetch_collection_data()["status"] == KINTO_REVIEW_STATUS
 
     def approve(self):
-        for i in range(60):
+        for _ in range(60):
```
jrbenny35, line 33: really??

app/tests/integration/nimbus/pages/browser/__init__.py
```diff
@@ -7,12 +7,12 @@
...
-    def execute_script(selenium, *args, script=None, context=None):
+    def execute_script(self, *args, script=None, context=None):
         if "chrome" in context:
-            with selenium.context(selenium.CONTEXT_CHROME):
-                return selenium.execute_script(script, *args)
+            with self.context(self.CONTEXT_CHROME):
+                return self.execute_script(script, *args)
         else:
-            return selenium.execute_script
+            return self.execute_script
```
jrbenny35, line 15: I don't think this will work

app/tests/integration/nimbus/pages/browser/__init__.py
```diff
@@ -42,7 +42,7 @@ def wait_for_pref_flip(self, pref, pref_value):
                 assert pref_value in [
                     element.text for element in elements
                 ], "Pref not found"
-            except (Exception, AssertionError) as e:
+            except Exception as e:
```
jrbenny35, line 45: I guess an ```AssertionError``` is just a fancy ```Exception```?

jrbenny35 (commented):
I think there's only 1 that wont work, lets see what the tests say

app/tests/integration/nimbus/jexl/__init__.py
jaredlockhart: This is a [conditional expression](https://docs.python.org/3/reference/expressions.html#conditional-expressions) which are usually called 'ternarny' expression in some other languages.

app/tests/integration/nimbus/pages/browser/__init__.py
jrbenny35: Wow, so this passed. I guess it is technically ```self``` then....huh...

app/tests/integration/nimbus/kinto/client.py
jaredlockhart: `_` is usually a placeholder for a value you must unpack but then don't care to reference again.  It's more of a [pattern](https://stackoverflow.com/a/5893946) than an actual syntax, but actually it is now a formal syntax for the new [match statement](https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords)

app/tests/integration/nimbus/pages/browser/__init__.py
jaredlockhart: All exception subclass `Exception` so here you're saying 'match all exceptions that are either a subclass of Exception (which matches all exceptions), or AssertionError', so the assertion error part there is redundant since everything will match the first type.

app/tests/integration/nimbus/pages/browser/__init__.py
jaredlockhart: The first argument to a class method is always the instance of the class the method is being called on, it can be called anything but it'll always point to that, so conventionally it's called 'self':

https://docs.python.org/3/tutorial/classes.html

```
Often, the first argument of a method is called self. This is nothing more than a convention: the name self has absolutely no special meaning to Python. Note, however, that by not following the convention your code may be less readable to other Python programmers, and it is also conceivable that a class browser program might be written that relies upon such a convention.
```

jrbenny35: Ah so this "self" is from the original class I used to call this function. Makes sense!

app/tests/integration/nimbus/jexl/__init__.py
jrbenny35: well I think the ```", "``` part is weird to read. The line just looks a bit confusing to read

jaredlockhart: It is a confusing line, I would generally avoid putting so much logic on a single line even though you can I try to keep it as like 'each line does one thing'.  Also the ", ".join(...) was already there, all its doing here is folding the if statement into a conditional statement?

jaredlockhart: But we can throw this change out, I agree it's a very busy line.

jrbenny35 (approved):
Thanks Jared!!

PR #8192, opened 2023-01-24
fix #8191 chore(nimbus): Remove reruns from sdk integration tests 

### LINKED ISSUE #8191: Remove reruns from SDK integration tests
These aren't needed as they can add significant overhead to the tests

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3089)

jrbenny35:
Because:
* We added reruns to some of the tests within the integration test suite including the sdk integration tests.

This commit:
* Removes the reruns from the SDK integration tests.

If a SDK integration test fails, it's probably for a good reason and rerunning it won't fix it.


jaredlockhart (approved):
🎉 🎉 🎉 🎉

PR #8194, opened 2023-01-25
fix #8137 chore(nimbus): remove deprecated v1 viz schema

### LINKED ISSUE #8137: Remove deprecated visualization API schema
We added a top level version to the visualization data (`v1` and `v2` right now) and filter to the correct one. The `v2` schema has been active for a couple weeks now, which means that `v1` should be safe to remove.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3051)

mikewilli:
Because

* v2 viz schema (with analysis_basis) has been in production without issue

This commit

* removes deprecated v1 viz schema


app/experimenter/experiments/migrations/0226_remove_viz_api_v1.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.db import migrations
+
+
+def remove_viz_api_v1(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+
+    for experiment in NimbusExperiment.objects.all():
+        if experiment.results_data is not None and "v1" in experiment.results_data:
+            del experiment.results_data["v1"]
```
jaredlockhart, line 11: I've always hated that this is the syntax to remove a key from a dictionary.  It's the only time you use a keyword like this, it should just be a method.  And I mean yeah you can just do experiment.results_data.pop('v1') and just ignore the return value, but I wish it was a little clearer.

jaredlockhart: Also do we know for sure that every dictionary in the db has this key?  Should we check for its presence here?

jaredlockhart: Oh nevermind you already tested for those cases 🎉 🎉 🎉 🎉

jaredlockhart (approved):
Awesome looks great ty 🙏 🙏 🙏

app/experimenter/experiments/migrations/0226_remove_viz_api_v1.py
mikewilli: Agreed, it's odd syntax for Python, but at least as far as uncommon syntax goes it's pretty obvious what it's doing.

PR #8195, opened 2023-01-25
fix 8189 tests(nimbus): Add missing fields to integration tests

jrbenny35:
Because:
* We were just filling out the required fields for the validator to pass and allow us to launch.

This commit:
* Adds the missing fields in so that we can properly test all the fields if needed.


app/tests/integration/nimbus/conftest.py
```diff
@@ -246,6 +246,7 @@ def _create_experiment(selenium):
         overview.select_risk_brand_false()
         overview.select_risk_revenue_false()
         overview.select_risk_partner_false()
+        overview.projects = ["Advanced hybrid forecast"]
```
jaredlockhart, line 249: You can't hardcore a name like this, they'll be different every time.

jaredlockhart (changes requested):

app/tests/integration/nimbus/conftest.py
jrbenny35: Ugh, is there a way to add a test project that will always be there?

jaredlockhart: In theory, this is why I was asking about how to trigger the multiselect.  Why can't we just hit it and select the first option?

app/tests/integration/nimbus/conftest.py
```diff
@@ -246,6 +246,7 @@ def _create_experiment(selenium):
         overview.select_risk_brand_false()
         overview.select_risk_revenue_false()
         overview.select_risk_partner_false()
+        overview.projects = ["A"]
```
jaredlockhart, line 249: 🤣 🤣 🤣 👏 👏 👏

app/tests/integration/nimbus/conftest.py
jaredlockhart: Could also do it the same way we do for countries/locales/languages, except instead of reading the fixture file off the db, query it from the gql to get the ids.  Actually we could/should be doing that for countries/locales/languages as well.  Reading off those fixture files is messy.

jaredlockhart: Actually yeah that's just easier, let's do that.

https://github.com/mozilla/experimenter/issues/8196

app/tests/integration/nimbus/conftest.py
```diff
@@ -240,17 +241,18 @@ def _create_experiment(selenium):
...
         overview.select_risk_partner_false()
+        overview.public_description = default_data.public_description
         overview.set_additional_links(value="DESIGN_DOC")
         overview.add_additional_links()
         overview.set_additional_links(value="DS_JIRA", url="https://jira.jira.com")
         overview.add_additional_links()
         overview.set_additional_links(
             value="ENG_TICKET", url="https://www.smarter-engineering.eng"
         )
+        overview.projects = [helpers.load_config_data()["projects"][0]["name"]]
```
jaredlockhart, line 255: 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

jrbenny35:
Had to remove the ```SKIP_DUMMY``` flag as it wasn't setting up projects

.circleci/config.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           name: Run integration tests
           command: |
             cp .env.integration-tests .env
-            make refresh SKIP_DUMMY=1 PYTEST_SENTRY_DSN=$PYTEST_SENTRY_DSN up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
```
jaredlockhart, line 50: Yeah don't take this out, this'll create a bunch of experiments in a bunch of states including various live states and they'll get synced into remote settings and that could weirdly screw up the integration tests.  I'll just add a step to add projects for the integration tests https://github.com/mozilla/experimenter/issues/8198

jaredlockhart: Here use this https://github.com/mozilla/experimenter/pull/8200

jaredlockhart:
Oh and we need advanced targeting too.

jrbenny35:
> Oh and we need advanced targeting too.

What do you mean? We test with the ```No Targeting - All Users``` config now

jaredlockhart:
That's what I mean, set it to some other value.

app/tests/integration/nimbus/conftest.py
```diff
@@ -218,8 +219,8 @@ def default_data(application, experiment_name, load_experiment_outcomes):
         metrics=outcomes[str(application).lower().rsplit(".")[-1]],
         audience=BaseExperimentAudienceDataClass(
             channel=BaseExperimentAudienceChannels.RELEASE,
-            min_version=80,
-            targeting="no_targeting",
+            min_version=102,
+            targeting="test_targeting",
```
jrbenny35, line 223: @jaredlockhart is this what you meant?

jaredlockhart: Oh sure yeah that works thank you!

jaredlockhart (approved):
Perfect thank you!  🎉 🎉 🎉

PR #8197, opened 2023-01-25
fix #8196 chore(nimbus): query config data for integration tests

### LINKED ISSUE #8196: Read configuration from V5 API in integration tests
Presently, we read configuration for language, locale, and country ids by parsing the fixture files stored on disk, because they happen to include their db pks.  But not everything we need to set in the tests exists as a fixture on disk, but everything can be found by querying the configuration V5 API.  So let's query that API to get the ids for everything, and then we'll have all the ids we need and won't need to parse those files on disk at all.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3098)

jaredlockhart:
Because

* To fill out all the fields on all of the forms, we need the ids of all the select choices
* We currently query some configuration data in the integration tests, but not all of it

This commit

* Expands the config query in the integration tests to query all of the config data
* Refactors the graphql calls in the integration tests to be in one place


app/tests/integration/nimbus/test_mobile_targeting.py
```diff
@@ -1,109 +1,107 @@
...
-            # Wrap the sub expression in a boolean test because the evaluator will throw
-            # if the return type is not bool
-            sub_expr = f"{sub_expr} == true"
-            targeting_helper.eval_jexl(sub_expr)
-        except Exception as e:
-            raise Exception(f"Error evaluating: '{sub_expr}': {e}") from e
+import json
+import os
+
+import pytest
```
jrbenny35, line 4: Wonder why github is showing so many changes

app/tests/integration/nimbus/utils/helpers.py
```diff
@@ -1,89 +1,215 @@
...
+        except json.JSONDecodeError:
+            if retry + 1 >= LOAD_DATA_RETRIES:
+                raise
+            else:
+                time.sleep(LOAD_DATA_RETRY_DELAY)
+
+
+@lru_cache(maxsize=None)
+def load_config_data():
+    return load_graphql_data(
```
jrbenny35, line 28: friggen awesome

jrbenny35 (approved):
Looks good! Thanks for taking some time to get the graphQL stuff figured out. This will help us in the future

PR #8200, opened 2023-01-26
fix #8198 chore(project): create dummy projects for integration tests

### LINKED ISSUE #8198: Create projects when SKIP_DUMMY=1
Now that we have support for projects in nimbus, we want to set them in the integration tests.  But right now the only place that creates them is load_dummy_experiments, but the integration tests skip that step so then there's no projects to use during the integration tests.  Let's add a command to add some projects and then use that in the integration tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3099)

jaredlockhart:
Because

* Nimbus now integrates with the Projects models
* We need to test these in the integration tests
* The integration tests skip creating dummy experiments which will then not create any projects

This commit

* Adds a new management command to create dummy projects when SKIP_DUMMY is true for the integration tests


jrbenny35 (approved):
Thank you!

PR #8202, opened 2023-01-26
For #7823 feat(nimbus): Update can_edit to include live rollouts

eliserichards:
Because...

* We are going to allow users to edit live rollouts

This commit...

* Updates the definition of `can_edit` in the model
* Fixes some typos in the test factory which referenced `LAUNCH` instead of `LIVE`
* Adds tests for live in `test_models`


app/experimenter/experiments/models.py
```diff
@@ -613,8 +613,17 @@ def allocate_bucket_range(self):
...
+                self.status == self.Status.DRAFT
+                and self.publish_status == self.PublishStatus.IDLE
+            )
+            or (
+                self.status == self.Status.LIVE
+                and (
+                    self.publish_status == self.PublishStatus.DIRTY
+                    or self.publish_status == self.PublishStatus.IDLE
+                )
+            )
```
jaredlockhart, line 626: Shouldn't this be guarded on `is_rollout`?

jaredlockhart (dismissed):

app/experimenter/experiments/models.py
eliserichards: good point 👍

eliserichards: I split out the tests so there's one for experiments and one for rollouts. The parameterization would have been too huge otherwise

yashika-khurana (approved):
🎉

PR #8203, opened 2023-01-27
For #7823 feat(nimbus): Add form for live reviews

eliserichards:
📢📢📢 Requires #8202 to be merged first

Because...

* We want to be able to make edits while a rollout is live
* And we want to send those edits through the review process before we launch the edits

This commit...

* Creates the Live -> Review form which will appear at the top of the summary page
* Adds tests


eliserichards:
This is branched off of #8202, so let's get that merged first 👍

jaredlockhart (commented):
Why does this require a separate form instead of just retriggering the same 'Request Review' flow we have now?

eliserichards:
The current forms have checkbox acknowledgments built in, and I removed those for this form

yashikakhurana (changes requested):
Same as Jared's comment, we might not need new form, existing form can be re-used but if you want to remove some fields, you can add conditional check to those fields and can potentially use same form 🤔

eliserichards:
Since there is a specific form for each of the other four use cases and this one is very different from the others, I would prefer to keep them separate. It also makes the testing clearer imo, and doesn't rely on a bunch of `if`s

jaredlockhart (approved):
Okey doke, makes sense and looks good 👍

PR #8208, opened 2023-01-27
fix #8206 chore(nimbus): add python static typechecking

### LINKED ISSUE #8206: Add static type checking to Python
After many years of experimenting with Python static typing, particularly in the context of a Django project, I think I've finally found a combination that seems to work well.  We can use the [pyright](https://github.com/microsoft/pyright) type checker from Microsoft both in the backend and in the editor for those that use VS Code, and we can use [django-types](https://github.com/sbdchd/django-types) with that for django specific types.  Let's start by just enabling the checks that use as much of the automatic type inference from django-types as much as possible and then we can incrementally enable more features over time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3102)

jaredlockhart:
Because

* The Python static type checking environment is maturing
* Django has been historically very difficult to statically type check
* The django-types package now seems to add benefit without requiring a significant investment

This commit

* Adds the pyright type checker to our check suite
* Adds the django-types extension to get automatic type inference
* Enables some checks that will provide value without requiring massive refactors
* Cleans up some existing failures that this type checking has already caught


jaredlockhart:
Note, all those `pyi` files at the bottom are auto generated stubs.

app/experimenter/kinto/tests/test_tasks.py
```diff
@@ -608,6 +608,25 @@ def test_ending_experiment_completed_when_record_is_not_in_main(self):
             ).exists()
         )
 
+    def test_updating_experiment_with_published_dto_none_is_skipped(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.PAUSING_APPROVE_WAITING,
+            application=NimbusExperiment.Application.DESKTOP,
+            published_dto=None,
+        )
```
jrbenny35, line 616: A new test?

app/experimenter/settings.py
```diff
@@ -9,12 +9,26 @@
...
+    ForeignKey,
+    JSONField,
+    ManyToManyField,
+    ModelAdmin,
+    StackedInline,
+    TabularInline,
+]:
+    cls.__class_getitem__ = classmethod(lambda cls, *args, **kwargs: cls)
+
+
```
jrbenny35, line 26: What is this for?

jrbenny35 (commented):
Just a few questions

app/experimenter/kinto/tests/test_tasks.py
jaredlockhart: One of the things the typechecker gets you is looking for where things could be `None` but the code isn't handling the `None` case and then throws a big stink about it, which is good.  So now there's a bunch of cases where we check for `None` that were never covered by existing tests, to then the coverage checker will say 'these lines aren't tested', so you gotta add a new test.  This is all part of the dream of having static type analysis in Python 🙏

app/experimenter/settings.py
jaredlockhart: So this is an unfortunate part of where typechecking in Python is at right now.  So the theory is adding static type checking in Python is, the type annotations should be valid Python code, but also basically do nothing at runtime, ie be ignored.  Usually that's things like

```py
x: int = 1
```

or

```py
def my_func() -> int:
    return 1
```

The type annotations there `: int` or `-> int` can be ignored at runtime.  But another syntax that Python static typing uses is to add a generic parameter to something which looks like

```py
things: Dict[str, int] = {}
```

Notice the square brackets there.  The type checker is using those as a way to say 'these are the generic type specifications for this type', but square brackets also have runtime meaning in Python, which is it triggers a `__get_item__` method call, which is what's happening when you say `some_dictionary[some_key]`.  So what happens is at runtime, it tries to actually call `__get_item__` even though you're not trying to actually do like a dictionary lookup, you're just specifying generic types.  

So what this means is

```py
    cls.__class_getitem__ = classmethod(lambda cls, *args, **kwargs: cls)
```

means 'If someone adds a generic type parameter to these classes using [...], just treat it as a no op and do nothing at runtime'

It's kind of an annoying hack that's a side effect of the design of Python static type checking.

eliserichards: Interesting 🤔

app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -762,8 +762,8 @@ def test_results_data_with_null_conversion_percent(
 
         (
             DAILY_DATA,
-            WEEKLY_DATA,
-            OVERALL_DATA,
```
eliserichards, line 766: I'm curious why these were replaced?

eliserichards (commented):
Momma mia that's a lotta generated files 😳

app/experimenter/settings.py
jrbenny35: Great explanation, ill need to save that one!

jrbenny35 (approved):

eliserichards (approved):

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -1199,7 +1199,7 @@ def test_update_experiment_branches_with_feature_configs_error(self):
         reference_branch = {"name": "control", "description": "a control", "ratio": 1}
         treatment_branches = [{"name": "treatment1", "description": "desc1", "ratio": 1}]
         invalid_feature_config_id = (
-            NimbusFeatureConfig.objects.all().order_by("-id").first().id + 1
+            NimbusFeatureConfig.objects.all().order_by("-id").all()[0].id + 1
```
yashikakhurana, line 1202: is there any reason to change first from all()[0]

yashikakhurana (approved):
lot of files 😨 , looks good, lot of changes and is great to have, thanks @jaredlockhart this is very helpful, just have one question, otherwise looks 💯

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: Yeah it's probably because `.first()` is `T | None` and you can't add +1 to None whereas `.all()[0]` is just `T`

PR #8218, opened 2023-01-31
fix #feat(nimbus): Swap branches name

yashikakhurana:
Because

* We have seen this (error)[https://github.com/mozilla/experimenter/issues/8163], whenever you swap the branch's name i.e. treatment and control, it will start giving the error. This is happening because we have a unique constraint on the `NimbusBranch` model ` unique_together = (("slug", "experiment"),)`, so whenever we are swapping the names slug experiment combo still existed and raises an error

This commit
* Add check to the serializer to detect the branch swap case

---

Update:
* I made changes on the serializer to check the swap case and if that happens it will show the message to the user to select a random name
image


eliserichards (dismissed):
Excellent!!! 👏

app/experimenter/experiments/models.py
```diff
@@ -767,7 +767,7 @@ class NimbusBranch(models.Model):
     class Meta:
         verbose_name = "Nimbus Branch"
         verbose_name_plural = "Nimbus Branches"
-        unique_together = (("slug", "experiment"),)
+        unique_together = (("slug", "experiment", "id"),)
```
jaredlockhart, line 770: Oh no this completely breaks the goal of the uniqueness constraint.  The point of the uniqueness constraint is to not have two branches with identical slugs in the same experiment.  Adding `id` to this means you CAN have two branches with the same slug for an experiment, which is the same as not have any uniqueness constraint at all.  And since the branch slug is used for the deterministic enrollment in the client, this would break the enrollment determinism.  Similarly for analysis, branch slugs are used as bases for the analysis so this would also break analysis.

This isn't the right approach.  We have to keep that constraint intact and fix it in the branch saving logic.

jaredlockhart (dismissed):

yashikakhurana:
okay, so here is the update: I made changes on the serializer to check the swap case and if that happens it will show the message to the user to select a random name
image

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -577,6 +577,38 @@ def _validate_duplicate_branch_names(self, data):
...
+                        {
+                            "reference_branch": {
+                                "name": NimbusConstants.ERROR_BRANCH_SWAP
+                            },
+                            "treatment_branches": [
+                                {"name": NimbusConstants.ERROR_BRANCH_SWAP}
+                                for _ in data["treatment_branches"]
+                            ],
+                        }
+                    )
```
jaredlockhart, line 610: This only catches the case of a single treatment swapping with control, but doesn't catch the cases of

1) two treatments swapping
2) swapping a treatment with reference when there's more than 1 treatment

See if you can find a more general way to check for swaps for any combination.

jaredlockhart (commented):
Good start but there's a couple uncovered cases 🙏

jaredlockhart (changes requested):
Oops forgot to request change

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -526,29 +526,48 @@ class NimbusExperimentBranchMixin:
...
         if "reference_branch" in data and "treatment_branches" in data:
             ref_branch_name = data["reference_branch"]["name"]
-            treatment_branch_names = [
-                branch["name"] for branch in data["treatment_branches"]
-            ]
+            treatment_branch_names = [b["name"] for b in data["treatment_branches"]]
             all_names = [ref_branch_name, *treatment_branch_names]
-            unique_names = set(all_names)
+            if len(all_names) != len(set(all_names)):
+                error = {"name": NimbusConstants.ERROR_DUPLICATE_BRANCH_NAME}
```
eliserichards, line 538: 👏

app/experimenter/experiments/constants.py
```diff
@@ -406,6 +406,9 @@ class EmailType(models.TextChoices):
     ERROR_FIREFOX_VERSION_MAX = (
         "Ensure this value is greater than or equal to the minimum version"
     )
+    ERROR_BRANCH_SWAP = (
+        "You are trying to swap branches name, Give another random name to the branches"
```
eliserichards, line 410:
```suggestion
        "You are trying to swap branch names. Please choose another name for the branches."
```

eliserichards (commented):
This is looking great! Just to confirm that I understand correctly:
* This is only a valid case that can happen in Status = Draft
* When you switch the branch names, you get the error message, and changing _just one_ of the branch names will clear the error? (you don't have to change both branch names)

jaredlockhart:
@eliserichards Yeah exactly, that's the solution, is if you're going to swap branch names, just change them one at a time.

jaredlockhart:
@eliserichards Also thank you for rewording the error message!

eliserichards (approved):
Excellent!! Great job @yashikakhurana @jaredlockhart 🙌

yashikakhurana (commented):
Neat and clean 💎 , thanks @jaredlockhart 🎉

PR #8222, opened 2023-01-31
fix #8221 feat(nimbus): add search to nimbus experiment admin

### LINKED ISSUE #8221: Add slug and search to NimbusExperiment list view in admin
Because we often want to look up an experiment by its slug, we should add slug to the list table and enable the search bar for name and slug.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3110)

jaredlockhart:
Because

* We often want to find a particular experiment in the admin to make a change
* Usually we know the name or the slug
* It would be helpful to be able to find an experiment from its name or slug
* It would also be helpful to be able to filter by more things like channel or feature

This commit

* Enables search on the nimbus experiment admin page
* Adds some additional filters for rollout, channel, and feature config


app/experimenter/experiments/admin.py
```diff
@@ -309,7 +310,15 @@ class NimbusExperimentAdmin(
...
-    list_filter = ("status", "publish_status", "application")
+    list_filter = (
+        "status",
+        "publish_status",
+        "is_rollout",
+        "application",
+        "channel",
+        "feature_configs",
+    )
+    search_fields = ("name", "slug")
```
yashikakhurana, line 321: I think it would be worth including features too in the search

eliserichards: Agreed 👍

jaredlockhart: Oh that's not super practical since it's many related models deep to get to their fields.  I added features on the right filters for filtering by feature?

yashikakhurana: yup that works

eliserichards (approved):
LGTM!

Issue #8226, opened 2023-01-31
Add targeting for enrolling new users who have Firefox as pinned and need default


nuggetsnegin:
Scope is to add targeting that will check for enrolling new users who have Firefox as pinned and need default (existing targeting only covers needing default).

[Bugzilla #1814263](https://bugzilla.mozilla.org/show_bug.cgi?id=1814263)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3112)

PR #8228, opened 2023-01-31
Issue 8226 - Add targeting for enrolling new users who have Firefox has pinned and need default

nuggetsnegin:
Closes https://github.com/mozilla/experimenter/issues/8226

### DIFF
--- a/app/experimenter/targeting/constants.py
+++ b/app/experimenter/targeting/constants.py
@@ -496,6 +496,17 @@ INFREQUENT_USER_FIVE_BOOKMARKS = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+NEW_USER_WITH_7_28_DAY_PROFILE_AGE = NimbusTargetingConfig(
+    name="New user with 7 to 28 day profile age",
+    slug="new_user_with_7_28_day_profile_age",
+    description="Users with a profile that is between 7-28 days old, inclusive",
+    targeting=f"{PROFILELESSTHAN28DAYS} && {PROFILEMORETHAN7DAYS}",
+    desktop_telemetry="",
+    sticky_required=True,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
 NEW_USER_WITH_INFREQUENT_USE = NimbusTargetingConfig(
     name="New user with infrequent use",
     slug="new_user_with_infrequent_use",
@@ -945,6 +956,32 @@ IOS_RECENTLY_LOGGED_IN_USER = NimbusTargetingConfig(
     sticky_required=True,
     is_first_run_required=False,
     application_choice_names=(Application.IOS.name, Application.FOCUS_IOS.name),
+)
+
+HAS_GOOGLE_BING_DDG_AS_CURRENT_DEFAULT_SEARCH_ENGINE = NimbusTargetingConfig(
+    name="Has Google, Bing, or DuckDuckGo as current default search engine",
+    slug="has_google_bing_or_ddg_as_current_default_search_engine",
+    description="Users with Google, Bing, or DuckDuckGo as current default engine",
+    targeting=(
+        "'google' in searchEngines.current ||"
+        "searchEngines.current == 'bing' ||"
+        "searchEngines.current == 'ddg'"
+    ),
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+NEW_ANDROID_13_USERS = NimbusTargetingConfig(
+    name="New Android 13 Users",
+    slug="new_android_13_users",
+    description="Users who have Android 13 and are on their first run of the application",
+    targeting="(android_sdk_version|versionCompare('33') >= 0) && is_first_run",
+    desktop_telemetry="",
+    sticky_required=True,
+    is_first_run_required=True,
+    application_choice_names=(Application.FENIX.name,),
 )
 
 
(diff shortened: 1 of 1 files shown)


app/experimenter/targeting/constants.py
```diff
@@ -142,6 +142,36 @@ def __post_init__(self):
...
+        "{need_default}".format(
+            first_run=FIRST_RUN.targeting,
+            new_profile=NEW_PROFILE,
+            need_default=NEED_DEFAULT,
+            has_pin=HAS_PIN,
+        )
+    ),
+    desktop_telemetry=(
+        "{first_run} AND environment.system.os.windows_build_number >= 18362 AND "
+        "!isDefaultBrowser AND {new_profile} AND {has_pin}"
```
punamdahiya, line 166: should this have  {need_default} added as well

app/experimenter/targeting/constants.py
```diff
@@ -142,6 +142,36 @@ def __post_init__(self):
...
+    ),
+    slug="first_run_new_profile_need_default_has_pin",
+    description=(
+        "First start-up users (e.g. for about:welcome) on Windows 1903+, "
+        "with a new profile, needing default w/ pin"
+    ),
+    targeting=(
+        "{first_run} && os.windowsBuildNumber >= 18362 && {has_pin} && {new_profile} && "
+        "{need_default}".format(
+            first_run=FIRST_RUN.targeting,
```
punamdahiya, line 158: might be worth reusing `FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903`  similar to how its used in `FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903_PREFER_MOTION` for common checks

jaredlockhart:
LGTM but I'll wait till @punamdahiya has a chance to look again 👍

app/experimenter/targeting/constants.py
```diff
@@ -142,6 +142,33 @@ def __post_init__(self):
...
+            first_run=FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903.targeting,
+            has_pin=HAS_PIN,
+        )
+    ),
+    desktop_telemetry=(
+        "{first_run} AND {has_pin}".format(
+            first_run=FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903.desktop_telemetry,
+            has_pin=HAS_PIN,
+        )
+    ).format(first_run=FIRST_RUN.desktop_telemetry, has_pin=HAS_PIN),
```
punamdahiya, line 166: @jaredlockhart not sure about two formats (line 162 and line 166) syntax for desktop_telemetry. Does that look right?

nuggetsnegin: Oh, hm I based it off `FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903_PREFER_MOTION` targeting, but that is the only one that has two formats. 🤔 I can remove it since it  `FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903` has the targeting I need.

jaredlockhart: Oh wow I completely missed that.  Yeah that won't work like that, the second call is a no-op since the format strings will already be replaced.  You only need to call format once.

jaredlockhart (changes requested):

app/experimenter/targeting/constants.py
nuggetsnegin: Thanks for clarifying, I updated it! Should I also update the existing targeting that uses another format for telemetry on line 139 for `FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903_PREFER_MOTION ` or should that be a separate pr?

```
    desktop_telemetry=(
        "{first_run} AND !userPrefersReducedMotion".format(
            first_run=FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903.desktop_telemetry,
        )
    ).format(first_run=FIRST_RUN.desktop_telemetry),
    ```

**Review by punamdahiya — APPROVED** [review]
LGTM! Thanks

File: app/experimenter/targeting/constants.py (hunk 3 above)
Comment (punamdahiya) [reply app/experimenter/targeting/constants.py]: My understanding we discourage changing existing advanced targeting,  if needed we can remove and add new one when desktop_telemetry feature is available in auto analysis.

File: app/experimenter/targeting/constants.py [hunk 4]
```diff
@@ -142,6 +142,33 @@ def __post_init__(self):
...
+    ),
+    slug="first_run_new_profile_need_default_has_pin",
+    description=(
+        "First start-up users (e.g. for about:welcome) on Windows 1903+, "
+        "with a new profile, needing default w/ pin"
+    ),
+    targeting=(
+        "{first_run} && {has_pin}".format(
+            first_run=FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903.targeting,
+            has_pin=HAS_PIN,
```
jrbenny35, line 158: Any reason we are using ```.format``` over f-strings?

jrbenny35 (commented):
Just had a question

app/experimenter/targeting/constants.py
nuggetsnegin: No particular reason, was basing this targeting off of `FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903_PREFER_MOTION`. But I can update it if f-strings is the more common format.

jaredlockhart: Oh yeah, I've thought about that, in general yes fstrings are preferred over .format but in this case, the lookups for the fields being substituted in can be very long and it would make the fstrings much uglier and difficult to read so I think .format is fine in this context.

app/experimenter/targeting/constants.py
jaredlockhart: I would say in general, adding a new targeting is preferred.  Modifying or removing existing ones can definitely be done and won't break anything, but it might interrupt someone elses planning, so if you can get away with just adding a new one that's generally a better idea.

jaredlockhart (approved):
Looks great thnx @nuggetsnegin 🥳

PR #8235, opened 2023-02-01
For #7823 feat(nimbus): Add dirty to update requests and rejection filters in models

eliserichards:
Because...

* We want to be able to approve/reject update requests for live rollouts the same way we do for experiments

This commit...

* Updates the definition of `latest_review_request` to include update review requests
* Adds `PublishStatus.DIRTY` to the review rejection definition 
* Tests!
   * The test_mutations file has some checks for ending experiments/rejection, so I updated that too

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -1021,13 +1021,19 @@ class NimbusChangeLogManager(models.Manager["NimbusChangeLog"]):
     def latest_review_request(self):
         return (
             self.all()
-            .filter(NimbusChangeLog.Filters.IS_REVIEW_REQUEST)
+            .filter(
+                NimbusChangeLog.Filters.IS_REVIEW_REQUEST
+                | NimbusChangeLog.Filters.IS_UPDATE_REVIEW_REQUEST
+            )
             .order_by("-changed_on")
         ).first()
 
     def latest_rejection(self):
         change = self.latest_change()
-        if change and change.has_filter(NimbusChangeLog.Filters.IS_REJECTION):
+        if change and change.has_filter(
+            NimbusChangeLog.Filters.IS_REJECTION
+            | NimbusChangeLog.Filters.IS_UPDATE_REJECTION
+        ):
             return change
 
     def latest_timeout(self):
@@ -1084,13 +1090,30 @@ class NimbusChangeLog(FilterMixin, models.Model):
             old_publish_status=NimbusExperiment.PublishStatus.IDLE,
             new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
         )
+        IS_UPDATE_REVIEW_REQUEST = Q(
+            old_publish_status=NimbusExperiment.PublishStatus.DIRTY,
+            new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
+        )
         IS_REJECTION = Q(
             Q(old_status=F("new_status")),
             old_publish_status__in=(
                 NimbusExperiment.PublishStatus.REVIEW,
                 NimbusExperiment.PublishStatus.WAITING,
             ),
-            new_publish_status=NimbusExperiment.PublishStatus.IDLE,
+            new_publish_status=(NimbusExperiment.PublishStatus.IDLE),
+            new_status__in=(
+                NimbusExperiment.Status.DRAFT,
+                NimbusExperiment.Status.LIVE,
+            ),
+            published_dto_changed=False,
+        )
+        IS_UPDATE_REJECTION = Q(
+            Q(old_status=F("new_status")),
+            old_publish_status__in=(
+                NimbusExperiment.PublishStatus.REVIEW,
+                NimbusExperiment.PublishStatus.WAITING,
+            ),
+            new_publish_status__in=(NimbusExperiment.PublishStatus.DIRTY,),
             published_dto_changed=False,
         )
         IS_TIMEOUT = Q(
--- a/app/experimenter/experiments/tests/api/v5/test_mutations.py
+++ b/app/experimenter/experiments/tests/api/v5/test_mutations.py
@@ -826,7 +826,8 @@ class TestUpdateExperimentMutationSingleFeature(
     def test_reject_ending_experiment(self):
         user_email = "user@example.com"
         experiment = NimbusExperimentFactory.create_with_lifecycle(
-            NimbusExperimentFactory.Lifecycles.ENDING_REVIEW_REQUESTED
+            NimbusExperimentFactory.Lifecycles.ENDING_REVIEW_REQUESTED,
+            is_rollout=False,
         )
         response = self.query(
             UPDATE_EXPERIMENT_MUTATION,
@@ -834,7 +835,8 @@ class TestUpdateExperimentMutationSingleFeature(
                 "input": {
                     "id": experiment.id,
                     "publishStatus": NimbusExperiment.PublishStatus.IDLE.name,
-                    "statusNext": NimbusExperiment.Status.COMPLETE.name,
+                    "statusNext": None,
+                    "status": NimbusExperiment.Status.LIVE.name,
                     "changelogMessage": "This is not good",
                 }
             },
@@ -847,7 +849,7 @@ class TestUpdateExperimentMutationSingleFeature(
 
         experiment = NimbusExperiment.objects.get(id=experiment.id)
         self.assertEqual(experiment.publish_status, NimbusExperiment.PublishStatus.IDLE)
-        self.assertEqual(experiment.status_next, NimbusExperiment.Status.COMPLETE)
+        self.assertEqual(experiment.status_next, None)
         rejection = experiment.changes.latest_rejection()
         self.assertEqual(rejection.changed_by.email, user_email)
         self.assertEqual(rejection.message, "This is not good")
--- a/app/experimenter/experiments/tests/test_models.py
+++ b/app/experimenter/experiments/tests/test_models.py
@@ -2095,6 +2095,19 @@ class TestNimbusChangeLogManager(TestCase):
 
         self.assertEqual(experiment.changes.latest_review_request(), change)
 
+    def test_latest_review_request_returns_change_for_dirty_to_review(self):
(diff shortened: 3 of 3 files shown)


app/experimenter/experiments/tests/test_models.py
```diff
@@ -2095,6 +2095,18 @@ def test_latest_review_request_returns_change_for_idle_to_review(self):
 
         self.assertEqual(experiment.changes.latest_review_request(), change)
 
+    def test_latest_review_request_returns_change_for_dirty_to_review(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LIVE_DIRTY,
+        )
+
+        experiment.publish_status = NimbusExperiment.PublishStatus.REVIEW
+        experiment.save()
```
jaredlockhart, line 2104: There could be another lifecycle specifically for being in the live/dirty > review state?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -2143,6 +2155,25 @@ def test_latest_rejection_returns_rejection_for_review_to_idle(self):
...
+        changes = []
+        for publish_status in (
+            NimbusExperiment.PublishStatus.REVIEW,
+            NimbusExperiment.PublishStatus.DIRTY,
+        ):
+            experiment.publish_status = publish_status
+            experiment.save()
+            changes.append(
+                generate_nimbus_changelog(experiment, experiment.owner, "test message")
+            )
```
jaredlockhart, line 2172: Oh yeah I would definitely just make lifecycles for these states since that'll do what you're doing here automatically

app/experimenter/experiments/tests/test_models.py
```diff
@@ -2176,6 +2207,18 @@ def test_launch_to_live_is_not_considered_latest_rejection(self):
...
 
+    def test_update_live_is_not_considered_latest_rejection(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_APPROVE,
+        )
+
+        experiment.status = NimbusExperiment.Status.LIVE
+        experiment.publish_status = NimbusExperiment.PublishStatus.IDLE
+        experiment.save()
+        generate_nimbus_changelog(experiment, experiment.owner, "test message")
```
jaredlockhart, line 2218: Shouldn't it already be in this state with that lifecycle?

jaredlockhart (dismissed):
Love this.  Those filter objects are such a nice way to encapsulate stuff like this.  Couple ideas 🙏

app/experimenter/experiments/tests/test_models.py
eliserichards: There is, I could use [LIVE_REVIEW_REQUESTED](https://github.com/mozilla/experimenter/blob/bb461d14e599870808c0e538e75c0c601c71ff05/app/experimenter/experiments/tests/factories.py#L223) instead 👍

eliserichards: Actually, I think that I'll keep these as is. These tests are creating an experiment one step before we want it (in this case, it's a Live Dirty rollout) and then we're manually generating the change log so we can assert on it.

If we created the lifecycle directly, we'd have to do a bunch of fun matching with the way that the [factory generates changelogs](https://github.com/mozilla/experimenter/blob/bb461d14e599870808c0e538e75c0c601c71ff05/app/experimenter/experiments/tests/factories.py#L468). I think this is simpler and easier to understand

app/experimenter/experiments/tests/test_models.py
eliserichards: Same as above

app/experimenter/experiments/tests/test_models.py
eliserichards: Whoops! Yup 👍

app/experimenter/experiments/models.py
```diff
@@ -1025,13 +1025,19 @@ def latest_change(self):
     def latest_review_request(self):
         return (
             self.all()
-            .filter(NimbusChangeLog.Filters.IS_REVIEW_REQUEST)
+            .filter(
+                NimbusChangeLog.Filters.IS_REVIEW_REQUEST
+                | NimbusChangeLog.Filters.IS_UPDATE_REVIEW_REQUEST
+            )
             .order_by("-changed_on")
```
yashikakhurana, line 1032: oh we can change this by `-updated_date_time`

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -826,15 +826,17 @@ def test_reject_draft_experiment(self):
...
+            is_rollout=False,
         )
         response = self.query(
             UPDATE_EXPERIMENT_MUTATION,
             variables={
                 "input": {
                     "id": experiment.id,
                     "publishStatus": NimbusExperiment.PublishStatus.IDLE.name,
-                    "statusNext": NimbusExperiment.Status.COMPLETE.name,
+                    "statusNext": None,
```
yashikakhurana, line 838: why we are changing` statusNext `here?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -2176,6 +2208,22 @@ def test_launch_to_live_is_not_considered_latest_rejection(self):
 
         self.assertIsNone(experiment.changes.latest_rejection())
 
+    def test_update_live_is_not_considered_latest_rejection(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_WAITING,
+            is_rollout=True,
+            published_dto={"id": "cool-cat", "test": False},
```
yashikakhurana, line 2215: 😹

app/experimenter/experiments/tests/test_models.py
```diff
@@ -2143,6 +2156,25 @@ def test_latest_rejection_returns_rejection_for_review_to_idle(self):
         self.assertEqual(experiment.changes.latest_review_request(), changes[0])
         self.assertEqual(experiment.changes.latest_rejection(), changes[1])
 
+    def test_latest_rejection_returns_rejection_for_review_to_dirty(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LIVE_DIRTY,
```
yashikakhurana, line 2161: In the above test, you have included `is_rollout=True`, but not here, If its included in life cycle then its good and would recommend to remove from the above test too, if not I will suggest to add it here too

yashikakhurana (dismissed):
Thank you @eliserichards it includes some wonderful test cases 👏 , some minor suggestions/questions

app/experimenter/experiments/tests/test_models.py
eliserichards: 🐱

app/experimenter/experiments/tests/test_models.py
eliserichards: Ah! great catch!! It should be added here too

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: The test is supposed to be looking at a request to end the experiment being rejected, which means it's statusNext should be None. I'm not sure why it was using `COMPLETE` before, I think it was just an overlook 🤷‍♀️

eliserichards: create an experiment with: 

```
status: LIVE
statusNext: COMPLETE
publishStatus: REVIEW
```

reject -> 

```
status: LIVE
statusNext: None
publishStatus: IDLE
```

eliserichards: https://github.com/mozilla/experimenter/blob/main/app/experimenter/docs/README.md#end-reject------

app/experimenter/experiments/models.py
eliserichards: Looks like there's a few places where we can update `changed_on` -> `updated_date_time`. I'll file a bug for this :D

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: A+ for referring back to the diagrams here 👏 👏 👏 👏

app/experimenter/experiments/models.py
jaredlockhart: I don't think so?  `-_updated_date_time` is on `NimbusExperiment`, but here we're querying `NimbusChangeLog`?

app/experimenter/experiments/models.py
```diff
@@ -1026,13 +1026,19 @@ def latest_change(self):
...
         ).first()
 
     def latest_rejection(self):
         change = self.latest_change()
-        if change and change.has_filter(NimbusChangeLog.Filters.IS_REJECTION):
+        if change and change.has_filter(
+            NimbusChangeLog.Filters.IS_REJECTION
+            | NimbusChangeLog.Filters.IS_UPDATE_REJECTION
```
jaredlockhart, line 1040: Actually we could do this another way and have like

```py
IS_LAUNCH_REJECTION = ....
IS_UPDATE_REJECTION = ....
IS_REJECTION = IS_LAUNCH_REJECTION | IS_UDPATE_REJECTION
```

and then only use `IS_REJECTION` in these methods.  Similar for `IS_REVIEW_REQUEST`.  I guess it depends on how/where we want to treat review/rejection cases the same/different.  We can also just keep an eye on it and move things around later if we want.

jaredlockhart (approved):
Yep, all looks very nice to me.  Very nice clean way to describe the states of interest, consistent with diagrams, nice simple comprehensive tests. A+ 10/10 🎉

app/experimenter/experiments/models.py
yashikakhurana: right, I got confused with Nimbus Experiment Manager

Issue #8239, opened 2023-02-02
Enable population percent fields


data-sync-user:
Once all of the review forms are ready to go, we can enable the population percent fields on the audience page for live rollouts. This will be the final step to allow people to edit live rollouts.

[https://github.com/mozilla/experimenter/issues/8239](https://github.com/mozilla/experimenter/issues/8239|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3123)

PR #8250, opened 2023-02-03
For #7823 feat(nimbus): Add population percent and dirty to constants

eliserichards:
Because...

* We want to allow updates to dirty states and allow editing for population percent for live rollouts

This commit...

* updates the serializer constants to reflect that

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -86,7 +86,9 @@ class TransitionConstants:
             NimbusExperiment.PublishStatus.IDLE,
         ],
         "experiments": [],
-        "rollouts": [],
+        "rollouts": [
+            NimbusExperiment.PublishStatus.DIRTY,
+        ],
     }
 
     STATUS_UPDATE_EXEMPT_FIELDS = {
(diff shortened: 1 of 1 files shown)


yashikakhurana (changes requested):
I think worth adding little test

eliserichards:
These are used in the serializers and already have tests :) Also #8239 will add tests once the fields are enabled in the UI

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -84,6 +84,7 @@ class TransitionConstants:
     PUBLISH_STATUS_ALLOWS_UPDATE = {
         "all": [
             NimbusExperiment.PublishStatus.IDLE,
+            NimbusExperiment.PublishStatus.DIRTY,
```
jaredlockhart, line 87: Shouldn't this be under rollouts?

eliserichards: 🤦‍♀️

yashikakhurana (approved):
looks good, thank you for pointing to the right test cases 👏

PR #8253, opened 2023-02-03
For #7823 feat(nimbus): Update publish status on live update

eliserichards:
⭐ **For testing...** ⭐ 

Since this PR doesn't enable the population fields for editing, you can cherry-pick [this commit](https://github.com/mozilla/experimenter/commit/f788820d022187978385584d1bb7d9753b42b10f) for local testing

----

This commit...

* Updates publish status to `DIRTY` when a live update is made to a rollout
* Navigates to the summary page instead of `../` for live saves
* Adds lots of tests for the review workflow
   * test_mutations
   * ChangeApprovalOperations/index.test.tsx
   * LinkNavSummary/index.test.tsx
   * PageEditAudience/index.test.tsx
   * PageSummary/index.test.tsx
   * Summary/index.test.tsx
* Adds some mock infrastructure so testing live/dirty rollouts is easier in these files

This commit _does not_...

* Enable the population percent field or the save/save & continue buttons on the audience page (#8239)
* Add functionality for the cancel buttons (#TBD)

--- 

https://user-images.githubusercontent.com/43795363/216703509-b7c0f474-b43f-4cf5-a3a0-6286a783fa91.mov

https://user-images.githubusercontent.com/43795363/216703597-a20f353d-d7a9-41d9-85f0-d74f59f18839.mov

image
image

image

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1030,6 +1030,19 @@ class NimbusExperimentSerializer(
         return data
 
     def update(self, experiment, validated_data):
+        if (
+            experiment.is_rollout
+            and not experiment.is_paused
+            and experiment.status == NimbusExperiment.Status.LIVE
+            and experiment.status_next is None
+            and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
+            and validated_data.get("publish_status")
+            != NimbusConstants.PublishStatus.REVIEW
+        ):
+            # can be Live Update (Dirty), End Enrollment, or End Experiment
+            # if we don't check validated_data["publish_status"]
+            validated_data["publish_status"] = NimbusConstants.PublishStatus.DIRTY
+
         self.changelog_message = validated_data.pop("changelog_message")
         return super().update(experiment, validated_data)
 
--- a/app/experimenter/experiments/tests/api/v5/test_mutations.py
+++ b/app/experimenter/experiments/tests/api/v5/test_mutations.py
@@ -433,7 +433,7 @@ class TestUpdateExperimentMutationSingleFeature(
         experiment = NimbusExperiment.objects.get(id=experiment_id)
         self.assertTrue(experiment.is_rollout)
         self.assertEqual(experiment.population_percent, 50.0)
-        self.assertEqual(experiment.publish_status, NimbusConstants.PublishStatus.IDLE)
+        self.assertEqual(experiment.publish_status, NimbusConstants.PublishStatus.DIRTY)
         self.assertEqual(experiment.status, NimbusConstants.Status.LIVE)
         self.assertEqual(experiment.status_next, None)
 
@@ -458,6 +458,10 @@ class TestUpdateExperimentMutationSingleFeature(
             headers={settings.OPENIDC_EMAIL_HEADER: user_email},
         )
         self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        result = content["data"]["updateExperiment"]
+
+        self.assertEqual(result["message"].keys(), {"experiment"})
 
         content = json.loads(response.content)
         result = content["data"]["updateExperiment"]
--- a/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.test.tsx
@@ -145,6 +145,42 @@ describe("ChangeApprovalOperations", () => {
     expect(openRemoteSettingsButton).toHaveProperty("href", REVIEW_URL);
   });
 
+  it("when user can review for live updates, supports approval and opening remote settings", async () => {
+    const approveChange = jest.fn();
+    render(
+      <Subject
+        {...{
+          ...reviewRequestedBaseProps,
+          canReview: true,
+          approveChange,
+          status: {
+            ...getStatus(MOCK_LIVE_ROLLOUT),
+            draft: false,
+            dirty: true,
+            live: true,
+          },
+        }}
+      />,
+    );
+
+    const reviewAlertTitle = await screen.findByTestId("review-request-alert");
+    await waitFor(() => {
+      expect(reviewAlertTitle).toBeInTheDocument();
+    });
+
+    const approveButton = await screen.findByTestId("approve-request");
+    fireEvent.click(approveButton);
+
+    await waitFor(() => {
+      expect(approveChange).toHaveBeenCalled();
+    });
+
+    const openRemoteSettingsButton = await screen.findByTestId(
+      "open-remote-settings",
+    );
+    expect(openRemoteSettingsButton).toHaveProperty("href", REVIEW_URL);
+  });
+
   it("when user cannot review, an approval pending notice is displayed", async () => {
     Object.assign(navigator, {
       clipboard: {
--- a/app/experimenter/nimbus-ui/src/components/LinkNavSummary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/LinkNavSummary/index.test.tsx
@@ -10,6 +10,7 @@ import {
 import { render, screen } from "@testing-library/react";
 import React from "react";
 import LinkNavSummary from "src/components/LinkNavSummary";
+import { LIFECYCLE_REVIEW_FLOWS } from "src/lib/constants";
 import { mockGetStatus } from "src/lib/mocks";
 import {
   NimbusExperimentPublishStatusEnum,
@@ -24,6 +25,7 @@ type SubjectProps = {
   canReview?: boolean;
   showSummaryAction?: boolean;
   isArchived?: boolean;
+  isRollout?: boolean;
 };
 
 const Subject = ({
@@ -34,6 +36,7 @@ const Subject = ({
   showSummaryAction = true,
   canReview = false,
   isArchived = false,
+  isRollout = false,
 }: SubjectProps) => {
   // Create a LocationProvider context since LinkNav uses `useLocation()`
   // to detect current page
@@ -122,14 +125,32 @@ describe("LinkNavSummary", () => {
       expect(screen.queryByText("Review End Request")).toBeInTheDocument();
     });
 
-    it("renders 'Review Launch Request' when expected", () => {
+    it("renders review launch request title when expected", () => {
       render(
         <Subject
+          status={NimbusExperimentStatusEnum.DRAFT}
           publishStatus={NimbusExperimentPublishStatusEnum.REVIEW}
           canReview
         />,
       );
-      expect(screen.queryByText("Review Launch Request")).toBeInTheDocument();
+      expect(
+        screen.queryByText(LIFECYCLE_REVIEW_FLOWS.LAUNCH.reviewSummary),
+      ).toBeInTheDocument();
+    });
+
+    it("renders update launch request title when expected", () => {
+      render(
+        <Subject
+          status={NimbusExperimentStatusEnum.LIVE}
+          statusNext={NimbusExperimentStatusEnum.LIVE}
+          publishStatus={NimbusExperimentPublishStatusEnum.REVIEW}
+          canReview
+          isRollout={true}
+        />,
+      );
+      expect(
+        screen.queryByText(LIFECYCLE_REVIEW_FLOWS.UPDATE.reviewSummary),
+      ).toBeInTheDocument();
     });
 
     it("renders 'Review End Enrollment Request' when expected", () => {
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
@@ -33,6 +33,11 @@ let rolloutMutationMock: ReturnType<
   typeof mockUpdateExperimentAudienceMutation
 >;
 
+let mockDirtyRolloutSubmitData: Partial<ExperimentInput>;
+let dirtyRolloutMutationMock: ReturnType<
+  typeof mockUpdateExperimentAudienceMutation
+>;
+
 describe("PageEditAudience", () => {
   beforeAll(() => {
     fetchMock.enableMocks();
@@ -45,6 +50,9 @@ describe("PageEditAudience", () => {
 
   beforeEach(() => {
     mockSubmitData = { ...MOCK_FORM_DATA };
+    mockRolloutSubmitData = { ...MOCK_ROLLOUT_FORM_DATA };
+    mockDirtyRolloutSubmitData = { ...MOCK_DIRTY_ROLLOUT_FORM_DATA };
+
     mutationMock = mockUpdateExperimentAudienceMutation(
       {
         ...mockSubmitData,
@@ -57,6 +65,15 @@ describe("PageEditAudience", () => {
     rolloutMutationMock = mockUpdateExperimentAudienceMutation(
       {
         ...mockRolloutSubmitData,
+        id: rollout.id,
+        changelogMessage: CHANGELOG_MESSAGES.UPDATED_AUDIENCE,
+      },
+      {},
+    );
+
+    dirtyRolloutMutationMock = mockUpdateExperimentAudienceMutation(
+      {
+        ...mockDirtyRolloutSubmitData,
         id: rollout.id,
         changelogMessage: CHANGELOG_MESSAGES.UPDATED_AUDIENCE,
       },
@@ -161,6 +178,22 @@ const MOCK_ROLLOUT_FORM_DATA = {
   firefoxMaxVersion: NimbusExperimentFirefoxVersionEnum.FIREFOX_95,
   targetingConfigSlug: "FIRST_RUN",
   populationPercent: "40",
+  totalEnrolledClients: 68000,
+  proposedEnrollment: "7",
+  proposedDuration: "28",
+  countries: ["1"],
+  locales: ["1"],
+  languages: ["1"],
+  isSticky: true,
+  isFirstRun: true,
+};
+
+const MOCK_DIRTY_ROLLOUT_FORM_DATA = {
+  channel: NimbusExperimentChannelEnum.NIGHTLY,
+  firefoxMinVersion: NimbusExperimentFirefoxVersionEnum.FIREFOX_83,
+  firefoxMaxVersion: NimbusExperimentFirefoxVersionEnum.FIREFOX_95,
+  targetingConfigSlug: "FIRST_RUN",
+  populationPercent: "60",
   totalEnrolledClients: 68000,
   proposedEnrollment: "7",
   proposedDuration: "28",
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.tsx
@@ -48,6 +48,7 @@ const PageEditAudience: React.FunctionComponent<RouteComponentProps> = () => {
       }: Record<string, any>,
       next: boolean,
     ) => {
+      
       try {
         // issue #3954: Need to parse string IDs into numbers
         const nimbusExperimentId = experiment.id;
--- a/app/experimenter/nimbus-ui/src/components/PageSummary/FormUpdateLiveToReview.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageSummary/FormUpdateLiveToReview.tsx
@@ -9,11 +9,11 @@ import Form from "react-bootstrap/Form";
 const FormUpdateLiveToReview = ({
   isLoading,
   onSubmit,
-  onCancel,
+  onRevert,
 }: {
   isLoading: boolean;
   onSubmit: () => void;
-  onCancel: () => void;
+  onRevert: () => void;
 }) => {
   return (
     <Alert
@@ -36,13 +36,13 @@ const FormUpdateLiveToReview = ({
               Request Update
             </button>
             <button
-              data-testid="cancel"
+              data-testid="revert"
               type="button"
               className="btn btn-secondary"
-              disabled={isLoading}
-              onClick={onCancel}
+              disabled={true}
+              onClick={onRevert}
             >
-              Cancel
+              Revert update
             </button>
           </div>
         </div>
--- a/app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
@@ -85,16 +85,10 @@ describe("PageSummary", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("renders the cancel review button if the experiment is in review state", async () => {
+  it("verify cancel review doesn't change enrollment days", async () => {
+    const refetch = jest.fn();
     const { mock, experiment } = mockExperimentQuery("demo-slug", {
       ...reviewRequestedBaseProps,
-      canReview: true,
-    });
-    render(<Subject mocks={[mock]} />);
-    await screen.findByText("Cancel Review");
-  });
-  it("verify cancel review doesn't change enrollment days", async () => {
-    const { mock, experiment } = mockExperimentQuery("demo-slug", {
       status: NimbusExperimentStatusEnum.LIVE,
       publishStatus: NimbusExperimentPublishStatusEnum.REVIEW,
     });
@@ -107,8 +101,7 @@ describe("PageSummary", () => {
         isEnrollmentPaused: false,
       },
     );
-    render(<Subject mocks={[mock, mutationMock]} />);
-
+    render(<Subject mocks={[mock, mutationMock]} {...{ refetch }} />);
     await screen.findByTestId("cancel-review-start");
     fireEvent.click(screen.getByTestId("cancel-review-start"));
     await waitFor(() => {
@@ -117,27 +110,6 @@ describe("PageSummary", () => {
     expect(screen.queryByTestId("label-enrollment-days")).toHaveTextContent(
       "1 day",
     );
-  });
-
-  it("can cancel review when the experiment is in the review state", async () => {
-    const { mock, experiment } = mockExperimentQuery("demo-slug", {
-      ...reviewRequestedBaseProps,
-      canReview: true,
-    });
-    const mutationMock = createFullStatusMutationMock(
-      experiment.id!,
-      NimbusExperimentStatusEnum.DRAFT,
-      null,
-      NimbusExperimentPublishStatusEnum.IDLE,
-      CHANGELOG_MESSAGES.CANCEL_REVIEW,
-    );
-    render(<Subject mocks={[mock, mutationMock]} />);
-
-    await screen.findByTestId("cancel-review-start");
-    fireEvent.click(screen.getByTestId("cancel-review-start"));
-    await waitFor(() => {
-      expect(screen.queryByTestId("submit-error")).not.toBeInTheDocument();
-    });
   });
 
   it("hides takeaways section if experiment is not complete", async () => {
@@ -266,9 +238,7 @@ describe("PageSummary", () => {
     expect(submitButton!).toBeEnabled();
     await act(async () => void fireEvent.click(submitButton));
 
-    const cancelButton = await screen.findByTestId("cancel");
-    fireEvent.click(cancelButton);
-    await screen.findByTestId("request-live-update-alert");
+    await screen.findByTestId("revert");
   });
 
   it("handles Launch to Preview after reconsidering Launch to Review from Draft", async () => {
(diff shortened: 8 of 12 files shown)


app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -85,16 +85,10 @@ describe("PageSummary", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("renders the cancel review button if the experiment is in review state", async () => {
```
eliserichards, line 88: This was a duplicate test, it's already handled in the Summary and ChangeApprovalOperations tests

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -119,27 +112,6 @@ describe("PageSummary", () => {
     );
   });
 
-  it("can cancel review when the experiment is in the review state", async () => {
```
eliserichards, line 122: same here

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -266,9 +238,7 @@ describe("PageSummary", () => {
     expect(submitButton!).toBeEnabled();
     await act(async () => void fireEvent.click(submitButton));
 
-    const cancelButton = await screen.findByTestId("cancel");
-    fireEvent.click(cancelButton);
-    await screen.findByTestId("request-live-update-alert");
+    await screen.findByTestId("revert");
```
eliserichards, line 241: We will add the revert functionality in another PR

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -408,6 +409,33 @@ def test_update_experiment_is_rollout(self):
...
+            is_rollout=True,
+            population_percent=40,
+        )
+        experiment_id = experiment.id
+        response = self.query(
+            UPDATE_EXPERIMENT_MUTATION,
+            variables={
+                "input": {
+                    "id": experiment.id,
+                    "isRollout": True,
```
jaredlockhart, line 426: This field won't get set on the audience page (it's set on the branches page I think) so I'd leave it out of this particular test to more closely resemble the query that will come from the audience page?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -408,6 +409,33 @@ def test_update_experiment_is_rollout(self):
...
+                    "publishStatus": NimbusConstants.PublishStatus.DIRTY.name,
+                    "changelogMessage": "test changelog message",
+                }
+            },
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        self.assertTrue(experiment.is_rollout)
+        self.assertEqual(experiment.population_percent, Decimal("50"))
```
jaredlockhart, line 437: We're setting the experiment up in IDLE, and I'd expect to assert that it's in DIRTY after the save?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -791,6 +819,29 @@ def test_update_experiment_publish_status(self):
...
+                    "publishStatus": NimbusExperiment.PublishStatus.DIRTY.name,
+                    "changelogMessage": "test changelog message",
+                }
+            },
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+
+        experiment = NimbusExperiment.objects.get(id=experiment.id)
+        self.assertEqual(experiment.publish_status, NimbusExperiment.PublishStatus.DIRTY)
```
jaredlockhart, line 843: Oh yeah, we could have the frontend be responsible for explicitly bumping it from IDLE to DIRTY, but I don't think there's any reason why that would be preferable to having the backend do it.  I'd imagine the backend would be responsible for making sure that an edit can only come through on LIVE(IDLE/DIRTY, is_rollout), and then the backend would bump from IDLE to DIRTY?  Then the frontend doens't need to worry about it?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -893,6 +944,32 @@ def test_launch_experiment_valid_with_preview_status(self):
         result = content["data"]["updateExperiment"]
         self.assertEqual(result["message"], "success")
 
+    def test_launch_experiment_valid_with_dirty_status(self):
+        user_email = "user@example.com"
+        experiment = NimbusExperimentFactory.create(
+            is_rollout=True,
+            publish_status=NimbusExperiment.PublishStatus.DIRTY,
+            status=NimbusExperiment.Status.LIVE,
```
jaredlockhart, line 952: Any reason not to set this up as a Lifecycle?  I think the goal of the lifecycles was so that we could have a shared and discoverable way to reason about the different states we expect to find an experiment in, to prevent multiple far apart tests accidentally making subtly different expectations about what state to find an experiment in?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -893,6 +944,32 @@ def test_launch_experiment_valid_with_preview_status(self):
...
+                    "changelogMessage": "test changelog message",
+                }
+            },
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200, response.content)
+
+        content = json.loads(response.content)
+        result = content["data"]["updateExperiment"]
+        self.assertEqual(result["message"], "success")
```
jaredlockhart, line 971: Generally whenever there's an API call to write something to the db, I'd expect to see an assertion that checks that the data is saved properly.

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -11,20 +11,33 @@ import PageEditAudience from "src/components/PageEditAudience";
...
 
+let mockRolloutSubmitData: Partial<ExperimentInput>;
+let rolloutMutationMock: ReturnType<
+  typeof mockUpdateExperimentAudienceMutation
+>;
+
+let mockDirtyRolloutSubmitData: Partial<ExperimentInput>;
+let dirtyRolloutMutationMock: ReturnType<
+  typeof mockUpdateExperimentAudienceMutation
+>;
```
jaredlockhart, line 39: Wow.  Wow.   Bless your heart.  All I've been doing for weeks is trying to hunt down all the places where `Record<string, any>` is written in stead of `ExperimentInput`.  Thank you for using the actual type 🙏 😭

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -55,12 +89,58 @@ describe("PageEditAudience", () => {
...
     await screen.findByTestId("PageEditAudience");
     fireEvent.click(screen.getByTestId("next"));
+    expect(experiment.status).toEqual(NimbusExperimentStatusEnum.DRAFT);
     await waitFor(() => {
       expect(mockSubmit).toHaveBeenCalled();
       expect(navigate).toHaveBeenCalledWith("../");
     });
   });
 
+  it("handles form next button for live updates", async () => {
```
jaredlockhart, line 99: Oh good case to check 👍

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -55,12 +89,58 @@ describe("PageEditAudience", () => {
...
+    );
+    expect(rollout.populationPercent).toEqual(popPercentOgValue);
+
+    await waitFor(() => {
+      fireEvent.click(screen.getByTestId("submit"));
+      expect(
+        dirtyRolloutMutationMock.request.variables.input.publishStatus,
+      ).toEqual(NimbusExperimentPublishStatusEnum.DIRTY);
+    });
+  });
```
jaredlockhart, line 130: Similar to what I said above about making this the backends responsibility.

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.tsx
```diff
@@ -48,9 +49,17 @@ const PageEditAudience: React.FunctionComponent<RouteComponentProps> = () => {
...
+      // issue #3954: Need to parse string IDs into numbers
+      const nimbusExperimentId = experiment.id;
+      let publish_status = experiment.publishStatus;
+      if (
+        experiment.isRollout &&
+        getStatus(experiment).live &&
+        experiment.populationPercent !== populationPercent
+      ) {
+        publish_status = NimbusExperimentPublishStatusEnum.DIRTY;
+      }
```
jaredlockhart, line 61: Oh yeah see.  It would be so easy to mess this up on the client or change it and then have cases where fields are being changed but DIRTY isn't being set.  And it shouldn't even be dependent on any particular field or anything.  It should just be

1) If it's in LIVE/IDLE
2) Any mutation comes in
3) It's not LIVE/DIRTY

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
jaredlockhart: Leaving the camp site cleaner than you found it 🙏 🔥

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -438,6 +408,54 @@ describe("PageSummary", () => {
     );
   });
 
+  it("handles approval of live update as expected", async () => {
```
jaredlockhart, line 411: You could've split this part out.

1) PR for saving LIVE(IDLE/DIRTY) rollout saves and updates fields and lands in LIVE(DIRTY)
2) PR for LIVE(DIRTY) can go through the review/update flow

I know there was already a PR for the review/update flow.  It's not a big deal to include it in here but that could be a way to isolate PR concerns.

app/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -59,7 +59,12 @@ const Summary = ({ experiment, refetch }: SummaryProps) => {
       changelogMessage: CHANGELOG_MESSAGES.REQUESTED_REVIEW_END_ENROLLMENT,
     },
     {
-      publishStatus: NimbusExperimentPublishStatusEnum.IDLE,
+      publishStatus:
+        experiment.isRollout &&
+        experiment.status === NimbusExperimentStatusEnum.LIVE &&
+        experiment.publishStatus === NimbusExperimentPublishStatusEnum.DIRTY
+          ? NimbusExperimentPublishStatusEnum.DIRTY
+          : NimbusExperimentPublishStatusEnum.IDLE,
```
jaredlockhart, line 67: Oh is this to set what to write publishStatus to on the update flow?  Yeah exactly, now this logic is living in multiple places on the frontend.  No need for that.

jaredlockhart (dismissed):
Awesome.  Generally looks good.  This is getting so close!! 🎉 🎉 🎉

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: Totally agreed 🙌

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.test.tsx
```diff
@@ -105,6 +106,42 @@ describe("ChangeApprovalOperations", () => {
...
+
+    await waitFor(() => {
+      expect(approveChange).toHaveBeenCalled();
+    });
+
+    const openRemoteSettingsButton = await screen.findByTestId(
+      "open-remote-settings",
+    );
+    expect(openRemoteSettingsButton).toHaveProperty("href", REVIEW_URL);
+  });
```
yashikakhurana, line 143: 🎉

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
eliserichards: That's fair. The other PR was pretty much just adding the UI, and I saw this one as actually adding the logic 🤷‍♀️

app/experimenter/nimbus-ui/src/components/LinkNavSummary/index.test.tsx
```diff
@@ -122,14 +125,32 @@ describe("LinkNavSummary", () => {
       expect(screen.queryByText("Review End Request")).toBeInTheDocument();
     });
 
-    it("renders 'Review Launch Request' when expected", () => {
+    it("renders review launch title when expected", () => {
```
yashikakhurana, line 128: Review Launch Request title?

app/experimenter/nimbus-ui/src/components/PageSummary/FormUpdateLiveToReview.tsx
```diff
@@ -36,13 +36,13 @@ const FormUpdateLiveToReview = ({
...
+              data-testid="revert"
               type="button"
               className="btn btn-secondary"
-              disabled={isLoading}
-              onClick={onCancel}
+              disabled={true}
+              onClick={onRevert}
             >
-              Cancel
+              Revert update
```
yashikakhurana, line 45: what about Revert Changes?

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
yashikakhurana: oh I think this test case is not duplicated

yashikakhurana (dismissed):
overall great job @eliserichards 👏

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1030,6 +1030,13 @@ def validate(self, data):
...
 
     def update(self, experiment, validated_data):
+        population_percent = validated_data.get("population_percent")
+        if population_percent != experiment.population_percent and experiment.status in (
+            NimbusConstants.Status.LIVE
+        ):
+            experiment.publish_status = NimbusConstants.PublishStatus.DIRTY
+            experiment.save()
+
         self.changelog_message = validated_data.pop("changelog_message")
```
eliserichards, line 1040: I'm not sure if `update` is the best place for this 🤔 I considered a few other places but didn't think that a validation method would make sense, and this made it clear that the other data was validated before we updated the `publish_status` (though would it make sense to make this change and _then_ validate that the `publish_status` is still valid once we make it `DIRTY`??)

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
eliserichards: it's a little hard to find (so many tests for this stuff in so many places) -- https://github.com/mozilla/experimenter/blob/9a8f02df68452fd3a464914ae1f282bf681abf08/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx#L143

app/experimenter/experiments/tests/api/v5/test_mutations.py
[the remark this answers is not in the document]
eliserichards: it wants the decimal :( 
image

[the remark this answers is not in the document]
yashikakhurana: I wonder how it is working here https://github.com/mozilla/experimenter/blob/9a8f02df68452fd3a464914ae1f282bf681abf08/app/experimenter/experiments/tests/api/v5/test_mutations.py#L704

eliserichards:
Fixing the integration tests ✍️

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1030,6 +1030,12 @@ def validate(self, data):
         return data
 
     def update(self, experiment, validated_data):
+        population_percent = validated_data.get("population_percent")
+        if population_percent != experiment.population_percent and experiment.status in (
+            NimbusConstants.Status.LIVE
+        ):
```
jaredlockhart, line 1036: Oh I don't think there's any need to check individual fields here.  That won't scale well as we continue to allow different fields.  I think the only thing we care about here is if it's live and we're updating, it's dirty now.

app/experimenter/experiments/tests/api/v5/test_mutations.py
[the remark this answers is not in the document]
jaredlockhart: @eliserichards @yashikakhurana Notice something different between those two examples?  What type is `"50"` and what type is `10.0`? 🤔

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: If the backend is the one setting this then we don't need to cover this case at all.  In fact I might go as far as to say the frontend CAN'T do this so you don't want up with accidentally setting the publish_status in the frontend in a way that conflicts with the backend doing it.

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -893,6 +949,38 @@ def test_launch_experiment_valid_with_preview_status(self):
         result = content["data"]["updateExperiment"]
         self.assertEqual(result["message"], "success")
 
+    def test_launch_experiment_valid_with_dirty_status(self):
```
jaredlockhart, line 952: I think we mean `update` here not `launch`?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -408,6 +409,34 @@ def test_update_experiment_is_rollout(self):
         experiment = NimbusExperiment.objects.get(id=experiment_id)
         self.assertTrue(experiment.is_rollout)
 
+    def test_update_live_rollout(self):
```
jaredlockhart, line 412: Do we also have a test case to check that you can't do this for a non rollout?

jaredlockhart (changes requested):

app/experimenter/experiments/api/v5/serializers.py
eliserichards: I was using it as a way to rule out end enrollment and end experiment updates, which are also status=LIVE. I can switch it out though, I think you're right that it doesn't make sense here 👍

jaredlockhart: Oh good point about the ending cases, but you should be able to determine that by looking at `status` and `status_next`.  And in the ending enrollment case, in theory that should go away as a special case altogether and just fold into the DIRTY review/update flow.

app/experimenter/experiments/tests/api/v5/test_mutations.py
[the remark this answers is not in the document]
eliserichards: *facepalms severely

app/experimenter/experiments/api/v5/serializers.py
eliserichards: Nah, cuz we also need to account for end enrollment/end experiment rejection! There's 5 cases total to think about, and I think this is the smallest checks we can do to ensure that all of them are covered:
image

eliserichards: (don't judge my color-coded spreadsheet)

jaredlockhart: First of all. 👏 👏 👏 👏 

Second of all, shouldn't it be (status, status_next, publish_status)`LIVE/LIVE/IDLE` always goes to `LIVE/LIVE/DIRTY`?

jaredlockhart: In the case where an update comes in from V5?

jaredlockhart: Like we only care about the DIRTY UPDATE case in your diagram, and it would have gone from `LIVE/NONE/IDLE` to `LIVE/NONE/IDLE` ie not touching the states at all, but we can just bump `IDLE` to `DIRTY`?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1030,6 +1030,19 @@ def validate(self, data):
...
 
     def update(self, experiment, validated_data):
+        if (
+            experiment.is_rollout
+            and not experiment.is_paused
+            and experiment.status == NimbusExperiment.Status.LIVE
+            and experiment.status_next is None
+            and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
+            and validated_data.get("publish_status")
+            != NimbusConstants.PublishStatus.REVIEW
```
jaredlockhart, line 1040: So yeah I was just thinking about this.  The `self.instance` state will be what state the experiment is in before the new changes are applied.  The states in `validated_data` are what they're going to.  From your handy dandy spreadsheet, the case we want to intercept is the case where

current state: (live, none, idle)
next state: (live, none, idle)

and then we bump idle to dirty.  I think you need to check all six here?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -408,6 +408,63 @@ def test_update_experiment_is_rollout(self):
...
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        self.assertTrue(experiment.is_rollout)
+        self.assertEqual(experiment.population_percent, 50.0)
+        self.assertEqual(experiment.publish_status, NimbusConstants.PublishStatus.DIRTY)
+        self.assertEqual(experiment.status, NimbusConstants.Status.LIVE)
+        self.assertEqual(experiment.status_next, None)
```
jaredlockhart, line 438: Excellent test 👏

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -408,6 +408,63 @@ def test_update_experiment_is_rollout(self):
...
+            variables={
+                "input": {
+                    "id": experiment.id,
+                    "populationPercent": "50",
+                    "changelogMessage": "test changelog message",
+                }
+            },
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
```
jaredlockhart, line 460: I assume we expect some kind of error to be here in the response, I'd expect to check that here in the test.

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -829,6 +887,7 @@ def test_reject_ending_experiment(self):
                     "publishStatus": NimbusExperiment.PublishStatus.IDLE.name,
                     "statusNext": None,
                     "status": NimbusExperiment.Status.LIVE.name,
+                    "populationPercent": "50",
```
jaredlockhart, line 890: What is this testing?  Why would we be mutating the population in the middle of a rejection?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -858,12 +918,16 @@ def test_end_experiment(self):
                     "id": experiment.id,
                     "publishStatus": NimbusExperiment.PublishStatus.REVIEW.name,
                     "statusNext": NimbusExperiment.Status.COMPLETE.name,
+                    "populationPercent": "50",
```
jaredlockhart, line 921: Same here?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -964,13 +966,15 @@ def test_rollout_can_request_review_from_live_dirty(self):
...
             is_rollout=True,
+            population_percent=50,
         )
 
         serializer = NimbusExperimentSerializer(
             experiment,
             data={
                 "status": NimbusExperiment.Status.LIVE,
                 "publish_status": NimbusExperiment.PublishStatus.REVIEW,
+                "population_percent": "50",
```
jaredlockhart, line 977: ?

jaredlockhart (dismissed):

app/experimenter/experiments/api/v5/serializers.py
eliserichards: Yup yup, all of them need to be there. If we don't, when we end the experiment/enrollment or reject them it'll set the state to dirty

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: OH dangarooski. These needed a value for populationPercent when I was checking whether the value had changed in the serializer. Now that we aren't checking that field, these can be removed

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
eliserichards: same

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: same

app/experimenter/experiments/api/v5/serializers.py
eliserichards: Not necessarily, `LIVE/NONE/IDLE` can be from ending an enrollment or ending an experiment as well (see the lefthand side of Dirty Updates, End Enrollment, and End Experiment on my handy-dandy spreadsheet). All three updates can start from the same `LIVE/NONE/IDLE` values, so we need to 
1. add isPaused into the mix to rule out End Enrollment
2. check the new val for publish status (Review for both end exp and end enroll) and/or check the new val for status next (Complete for end exp)

app/experimenter/experiments/tests/api/v5/test_mutations.py
yashikakhurana: I agree it should return or raise some error, instead of 200 status 🤔

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormApproveOrReject.tsx
```diff
@@ -38,7 +38,7 @@ const FormApproveOrReject = ({
           </p>
         </Alert>
       )}
-      <Alert variant="secondary">
+      <Alert variant="secondary" data-testid="review-request-alert">
```
yashikakhurana, line 41: I think it's worth adding a test if the rollout is false, `review-request-alert` is not rendered in that case

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -438,6 +408,54 @@ describe("PageSummary", () => {
...
+    expect(approveButton).toHaveTextContent("Approve and Update Rollout");
+    fireEvent.click(approveButton);
+    const openRemoteSettingsButton = await screen.findByTestId(
+      "open-remote-settings",
+    );
+    expect(openRemoteSettingsButton).toHaveProperty("href", rollout.reviewUrl);
+  });
+
+  it("handles rejection of live update as expected", async () => {
+    const expectedReason = "Oh no.";
```
yashikakhurana, line 434: oh no 😆

yashikakhurana (dismissed):
Super super close 🕺, just one major thought if the experiment is not rollout and api trying to change population size, it should raise an error that `experiment is not allowed to change population size`, otherwise 👏

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: Oh well no I mean graphql will always return a 200, sadly, but there will be errors in the response body we should probably assert on here.

yashikakhurana: yeah right, I mean review errors can be used here 😆

app/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
jaredlockhart: image

app/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormApproveOrReject.tsx
eliserichards: The review request alert is the box that says "email@email.com requested to {...}", so it _should_ be rendered in all cases :D I'll add this to the other tests!

eliserichards: https://github.com/mozilla/experimenter/pull/8317/files#r1108878470

eliserichards:
> Super super close 🕺, just one major thought if the experiment is not rollout and api trying to change population size, it should raise an error that `experiment is not allowed to change population size`, otherwise 👏

Since we don't even show the Audience page for non-rollouts, there is no possible way that a normal experiment could change the population size 😄

PR #8257, opened 2023-02-03
fix #8240 feat(nimbus): Fix none branch

### LINKED ISSUE #8240: TypeError: can't access property "name", c is null
Sentry Issue: [EXPERIMENTER-STAGE-3D](https://sentry.io/organizations/mozilla/issues/3913283100/?referrer=github_integration)

```
TypeError: can't access property "name", c is null
  at formControlAttrs (hooks/useCommonForm/useCommonFormMethods.tsx:55:31)
  at <object>.formControlAttrs (hooks/useCommonForm/useCommonNestedForm.tsx:44:8)
  at FormBranch (components/PageEditBranches/FormBranches/FormBranch.tsx:111:19)
  at sentryWrapped (../../../src/helpers.ts:98:17)
...
(14 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3124)

yashikakhurana:
Because

* We are getting this error https://github.com/mozilla/experimenter/issues/8240, The main issue was that from the backend we initiated the warning and errors as `None`, which leads to null in the frontend. 
* The warning needs an object inside so that we also maintain the warning no. as to which treatment branch it is referring to. for example, if we have warning in treatment branch b, then we need something like that 0-{}, 1-{"warning in treatment branch b"}  so that we can show the warning appropriately too.

This commit

* Instead of null/None returns an empty object
before fix screenshot
Screenshot 2023-02-03 at 3 10 08 PM

After fix 
Screenshot 2023-02-03 at 3 10 17 PM

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1394,8 +1394,8 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         treatment_branches_errors = []
         treatment_branches_warnings = []
         for branch_data in data["treatment_branches"]:
-            branch_error = None
-            branch_warning = None
+            branch_error = {}
+            branch_warning = {}
 
             if branch_data.get("feature_enabled", False):
                 if errors := self._validate_feature_value_against_schema(


jaredlockhart (changes requested):
Thanks @yashika-khurana ! We should update a test case to make sure we keep this behaviour intact.

jaredlockhart (approved):
Ty @yashika-khurana 🙏 🎉

PR #8258, opened 2023-02-06
fix #6740 feat(nimbus): remove single feature case from V5 input api

### LINKED ISSUE #6740: Use feature_configs in nimbus-ui

jaredlockhart:
Because

* The Nimbus UI now writes to the featureConfigs field rather than featureConfig
* We can now safely remove the single feature write case from the V5 API

This commit

* Removes write support for the single feature featureConfig field from the V5 API


jaredlockhart:
Actually I have to close this and do the `feature_enabled`/`feature_value` fields as well.

PR #8269, opened 2023-02-07
fix #8261 feat(nimbus): add timezone to analysis timestamp

### LINKED ISSUE #8261: `Results last calculated: 2/6/2023, 6:12:45 p.m.` should include a timezone
That line is from https://experimenter.services.mozilla.com/nimbus/mr2022-backgroundtaskmessage-notification-release-2pct-with-holdback/results#results_summary.  This run was at 10:12 AM Pacific, so this is UTC, but that's not clear.  Showing both UTC and local time would be a lot more helpful.  (Or show N hours ago, or whatever `moment.js` does.)

mikewilli:
Because

* we show the analysis timestamp without a timezone

This commit

* adds the timezone to remove ambiguity

image


jaredlockhart (approved):
Nice and clean 💯

PR #8272, opened 2023-02-07
fix #8268 chore(nimbus): remove feature enabled flag

### LINKED ISSUE #8268: Remove feature enabled from Nimbus UI
Feature enabled was deprecated in Desktop in 104+ and was never implemented in mobile, so let's go ahead and remove the enabled flag from the UI altogether, remove it from the models, and require a feature value for every branch.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3139)

jaredlockhart:
Because

* We originally had a boolean flag to enable a feature in a branch
* On desktop we realized this was superfluous and we could have features specify their enabled state in the feature value
* We deprecated support for the enabled flag on desktop in 104
* The feature enabled flag was never implemented in the Nimbus SDK

This commit

* Removes the feature enabled flag from the NimbusBranch model, serializers, apis, tests
* Preserves the flag set to default True in the V6 serializer until we no longer need to support desktop clients below 104
* Updates the Nimbus UI to remove the enabled toggle for feature values


app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -60,7 +58,7 @@ def get_feature(self, obj):
 
         return {
             "featureId": feature_config_slug,
-            "enabled": feature_enabled,
+            "enabled": True,
```
yashikakhurana, line 61: I would suggest to add a note as todo that we are setting this value as True and this later will get removed when we no longer support 104 version

app/experimenter/experiments/api/v6/serializers.py
```diff
@@ -77,7 +75,7 @@ def get_features(self, obj):
         for fv in obj.feature_values.all():
             feature_value = {
                 "featureId": fv.feature_config and fv.feature_config.slug or "",
-                "enabled": fv.enabled,
+                "enabled": True,
```
yashikakhurana, line 78: same as here, or may be add a constant and add a note where the constant is declared 🤔

app/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -305,13 +305,12 @@ def test_serializer_with_branches_no_feature_95(self):
...
             branch=experiment.reference_branch,
             feature_config=None,
-            enabled=False,
             value="",
         )
         serializer = NimbusExperimentSerializer(experiment)
         self.assertEqual(
             serializer.data["branches"][0]["features"],
-            [{"featureId": "", "enabled": False, "value": {}}],
+            [{"featureId": "", "enabled": True, "value": {}}],
```
yashikakhurana, line 313: and we could reuse that same constant here too

yashikakhurana (changes requested):
@jaredlockhart great description, awesome approach preserving v6 api thank you 👏 🎉 , just have one minor suggestion

yashikakhurana (approved):
just curious to know why are we removing the entire thing in the integration test? other than that 💯

jaredlockhart:
> just curious to know why are we removing the entire thing in the integration test? other than that 💯

Which entire thing?

app/tests/integration/nimbus/test_desktop_client_integrations.py
```diff
@@ -279,18 +261,9 @@ def test_check_telemetry_sticky_targeting(
...
-    experiment_default_data["treatmentBranches"] = [
-        {
-            "description": "treatment branch",
-            "name": "Branch 2",
-            "ratio": 0,
-            "featureEnabled": False,
-            "featureValue": "",
-        }
-    ]
+    experiment_default_data["treatmentBranches"] = []
```
yashikakhurana, line 266: this thing 🆙

jaredlockhart: Oh the tests don't use those branches so there's no point in them being declared at all.

PR #8285, opened 2023-02-08
fix #8284 chore(nimbus): remove duplicate enums from V5 API

### LINKED ISSUE #8284: Use consistent enums in V5 API
We have some duplicate enums in the V5 API for statuses, etc.  If you don't explicitly set a field as an enum in both the output types and input types, apollo will autogenerate duplicate enum references.  Let's get all of those consistent so we don't have duplicate enums.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3146)

jaredlockhart:
Because

* We have some duplicate enames in the V5 API for status, documentation link titles
* This can cause the typescript validation to mark things as incompatible when they are in fact compatible

This commit

* Explicitly declares enum fields in both the output and input types in the V5 API
* Updates references in Nimbus UI


app/experimenter/experiments/api/v5/types.py
```diff
@@ -241,6 +243,9 @@ class NimbusReviewType(graphene.ObjectType):
 
 
 class NimbusChangeLogType(DjangoObjectType):
+    old_status = NimbusExperimentStatusEnum()
+    old_status_next = NimbusExperimentStatusEnum()
```
eliserichards, line 247: 👏 this is going to make my life much easier w/ rollouts

eliserichards (approved):
Excellent 🙌

PR #8290, opened 2023-02-08
fix #8289 feat(nimbus): always fetch analysis data for Live experiments

### LINKED ISSUE #8289: Fetch Jetstream data based off external config end_date if available
Experimenter fetches Jetstream data up until the experiment's end_date (+ buffer of 3 days), even if an experiment's end_date is extended by a custom config. We should see if there is a custom config available, and if it has an end_date then use that.

The custom config info will not be available until after Jetstream data has been fetched once with the config in place, but this at least allows for automatic fetches after that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3148)

mikewilli:
Because

* we use the computed_end_date to determine when to stop refreshing jetstream results
* the computed_end_date is the proposed end date (based off proposed duration) before experiment ends
* the experiment may be extended beyond the proposed end date
* the experiment will still be LIVE if it is extended

This commit

* does not skip refreshing LIVE experiments


jaredlockhart (changes requested):
There's probably some tests we can add to cover these cases 🙏 

https://github.com/mozilla/experimenter/blob/main/app/experimenter/jetstream/tests/test_tasks.py

app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -837,15 +838,62 @@ def test_data_fetch_in_loop(self, lifecycle, mock_delay):
         tasks.fetch_jetstream_data()
         mock_delay.assert_called_once_with(experiment.id)
 
+    @parameterized.expand(
+        [
+            (NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_APPROVE,),
+        ]
+    )
```
jaredlockhart, line 845: Why parameterize for a single option?

app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -837,15 +838,62 @@ def test_data_fetch_in_loop(self, lifecycle, mock_delay):
...
+    @patch("experimenter.jetstream.tasks.fetch_experiment_data.delay")
+    def test_data_fetch_live(self, lifecycle, mock_delay):
+        # continue to fetch data while LIVE, even after proposed end date
```
jaredlockhart, line 848: Oh your comment here seems to me like it should be the actual name of the test

jaredlockhart (approved):
Yep all looks good and fine and right to me.  Couple nits about the tests but nothing major.  Thanks @mikewilli ! 🎉 🎉 🎉

app/experimenter/jetstream/tests/test_tasks.py
mikewilli: Yea I just kept the same format as all the other existing tests, but you're right no reason to keep this pattern when it's a single option. Removed in latest commit.

app/experimenter/jetstream/tests/test_tasks.py
mikewilli: Smart!

PR #8291, opened 2023-02-09
fix #8286 chore(nimbus): use cached dates in V6 API

### LINKED ISSUE #8286: Use cached dates in V6 API
Now that we're storing the start, enrollment end, and end dates right on the NimbusExperiment model, we don't need to use the changelogs to compute them in the V6 API.  Let's update the serializer to use the cached dates so we can cut down on the SQL time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3147)

jaredlockhart:
Because

* We recently saw some performance degradation in the V6 API
* We attempted to solve it by prefetching the changes
* This adds an expensive call to the changelog table which further regressed performance
* We already have a cache for the experiment dates so a better strategy is to make sure those caches get hit more frequently

This commit

* Removes the prefetch call to the changelog table in the V6 API
* Adds some additional logic checks to make sure we don't attempt to query for changelogs when we know there won't be one ie
  * No need to check for an enrollment end date for an experiment that's not started
  * No need to check for an enrollment end date for an experiment that isn't paused
* Moves away from the prefetch change/filter in memory approach which adds additional database load when prefetching isn't used

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -38,7 +38,6 @@ class NimbusExperimentManager(models.Manager["NimbusExperiment"]):
             super()
             .get_queryset()
             .prefetch_related(
-                "changes",
                 "locales",
                 "languages",
                 "countries",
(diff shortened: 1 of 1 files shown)


eliserichards (approved):
Lookin gorgeous! 👏

yashikakhurana (approved):
great description, super clear, with these changes our test cases are intact ✅

Issue #8238, opened 2023-02-02
Add advanced targeting to buckets


data-sync-user:
We compute buckets right now based rollout, feature, channel — NOT by advanced targeting —

* Advanced targeting can be “New users” or “existing users” — at the targeting level it is exclusive, these two categories will never overlap

We want to add advanced targeting as one of the bucketing identifiers so that every one of those combinations of rollout/feature/channel/adv.targeting would be unique

* Add to bucketing namespace hash function (for rollouts)
  * It’s common to have many experiments in the same bucket
  * What will this change do?
    * Experiments that previously would be on the same track will now have different tracks because they have different advanced targeting
    * Could under enroll experiments, but can pack more experiments in
    * How to avoid that?
      * Add a warning that you’re trying to launch a rollout for the same group of people who already have one — already live for same rollout, feature, channel , advanced targeting
* Prefix the namespace key when things are in preview, so that you don't get the case where an experiment/rollout that isn't live collides with one that's only in preview
  * Currently, switching from preview to live always reallocates the buckets (see `SHOULD_ALLOCATE_BUCKETS` definition)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3122)

PR #8299, opened 2023-02-13
For #8238 feat(nimbus): Extend bucketing to include targeting

eliserichards:
Because...

This commit...

* Extends the bucketing definition to include targeting
* Parses the targeting slug so that it follows the hyphenated format of the rest of the bucket name
* Adds a warning to `PageSummary` when you are attempting to create an experiment that would end up in the same bucket as an existing experiment

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1074,7 +1074,10 @@ class NimbusExperimentSerializer(
             experiment = super().save()
 
             if experiment.has_filter(experiment.Filters.SHOULD_ALLOCATE_BUCKETS):
+                if experiment.bucket_range_exists(): 
+                    experiment.warnings["bucketing"] = [NimbusExperiment.ERROR_BUCKET_EXISTS]
                 experiment.allocate_bucket_range()
+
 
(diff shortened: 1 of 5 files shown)


yashikakhurana:
@eliserichards question for you- `We want rollouts to be consistently enrolled as the population percentage increases (not unenrolled when the value is changed)` what will happen if the owner decided to go population size from 50% to 40%, I am just trying to understand the context

eliserichards:
> @eliserichards question for you- `We want rollouts to be consistently enrolled as the population percentage increases (not unenrolled when the value is changed)` what will happen if the owner decided to go population size from 50% to 40%, I am just trying to understand the context

Great point! If the user decreases the bucket size, it is caught by the new `SHOULD_ALLOCATE_BUCKETS` filter (which now includes live rollouts), and goes into the `allocate_bucket_range` [function on the model](https://github.com/mozilla/experimenter/blob/31704bf65acdb70e73c2e870a353a5ab0924d80b/app/experimenter/experiments/models.py#L610). Here it's saying it should delete if it already exists, which we don't want! if it's just a decrease, the bucket should remain the same and some users should be unenrolled from that existing bucket..

Issue #8315, opened 2023-02-16
Save button navigation for live rollouts


data-sync-user:
The “save & continue” button on the Audience page should navigate to the summary page when clicked

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3160)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/34068">image-20221130-203507.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/34069">image-20221130-203536.png</a>

PR #8316, opened 2023-02-16
For #8315 feat(nimbus) test(nimbus): Save buttons navigate to summary for live rollouts

eliserichards:
Because...

* We are re-using the Audience page for live rollouts
* And we want the "save and continue" button to take us to the summary page in that case

This commit...

* Adds navigation to summary page for live rollouts
* Adds tests!

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
@@ -11,23 +11,32 @@ import PageEditAudience from "src/components/PageEditAudience";
 import FormAudience from "src/components/PageEditAudience/FormAudience";
 import { UPDATE_EXPERIMENT_MUTATION } from "src/gql/experiments";
 import { CHANGELOG_MESSAGES, SUBMIT_ERROR } from "src/lib/constants";
-import { mockExperimentQuery } from "src/lib/mocks";
+import { mockExperimentQuery, mockLiveRolloutQuery } from "src/lib/mocks";
 import { RouterSlugProvider } from "src/lib/test-utils";
 import {
   ExperimentInput,
   NimbusExperimentChannelEnum,
   NimbusExperimentFirefoxVersionEnum,
+  NimbusExperimentPublishStatusEnum,
+  NimbusExperimentStatusEnum,
 } from "src/types/globalTypes";
 import { updateExperiment_updateExperiment } from "src/types/updateExperiment";
 
 const { mock, experiment } = mockExperimentQuery("demo-slug");
+const { mockRollout, rollout } = mockLiveRolloutQuery("demo-slug-1");
 
 let mockSubmitData: Partial<ExperimentInput>;
 let mutationMock: ReturnType<typeof mockUpdateExperimentAudienceMutation>;
 
+let mockRolloutSubmitData: Partial<ExperimentInput>;
+let rolloutMutationMock: ReturnType<
+  typeof mockUpdateExperimentAudienceMutation
+>;
+
 describe("PageEditAudience", () => {
   beforeAll(() => {
     fetchMock.enableMocks();
+    mockRolloutSubmitData = { ...MOCK_ROLLOUT_FORM_DATA };
   });
 
   afterAll(() => {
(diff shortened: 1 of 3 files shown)


eliserichards:
I pulled this stuff out of #8253

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -55,12 +73,36 @@ describe("PageEditAudience", () => {
...
     fireEvent.click(screen.getByTestId("next"));
+    expect(experiment.status).toEqual(NimbusExperimentStatusEnum.DRAFT);
     await waitFor(() => {
       expect(mockSubmit).toHaveBeenCalled();
       expect(navigate).toHaveBeenCalledWith("../");
     });
   });
 
+  it("handles form next button for live updates", async () => {
+    render(<Subject mocks={[mockRollout, rolloutMutationMock]} />);
```
jaredlockhart, line 84: So the `mockRollout` here is already in a live state eh?  I wish the setups here were a little more explicit.  I don't just mean this specific test, I just mean the frontend tests in general.  I like to see a setup where the exact conditions necessary are explicitly declared.  Particularly since we're asserting on the status later.  Not a specific direction, just a thought about the frontend tests in general.

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -55,12 +73,36 @@ describe("PageEditAudience", () => {
...
+      expect(navigate).toHaveBeenCalledWith("summary");
+    });
+  });
+
+  it("handles save button for live updates", async () => {
+    const { mockRollout, rollout } = mockLiveRolloutQuery("demo-slug", {
+      status: NimbusExperimentStatusEnum.LIVE,
+      publishStatus: NimbusExperimentPublishStatusEnum.IDLE,
+      isRollout: true,
+    });
```
jaredlockhart, line 99: Yeah see this is lovely 🙏

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -55,12 +73,36 @@ describe("PageEditAudience", () => {
...
+    render(<Subject mocks={[mockRollout, rolloutMutationMock]} />);
+    await screen.findByTestId("PageEditAudience");
+    fireEvent.click(screen.getByTestId("submit"));
+    await waitFor(() => expect(mockSubmit).toHaveBeenCalled());
```
jaredlockhart, line 103: So what exactly is this case testing?  Is there specific logic that says that this save won't occur if it's live but not a rollout?

jaredlockhart (approved):
Yeah the logic for navigating to the summary on next and test makes perfect sense.  Thanks for breaking this out, nice and simple and clear 🙏

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
```diff
@@ -55,12 +73,36 @@ describe("PageEditAudience", () => {
     render(<Subject mocks={[mock, mutationMock]} />);
     await screen.findByTestId("PageEditAudience");
     fireEvent.click(screen.getByTestId("next"));
+    expect(experiment.status).toEqual(NimbusExperimentStatusEnum.DRAFT);
```
yashikakhurana, line 76: great checking that 👏

yashikakhurana (approved):
thank you for splitting that up, your breakdown work really helping us to review 👏

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
eliserichards: I see what you mean. Maybe something to chat about next time the three of us get together?

app/experimenter/nimbus-ui/src/components/PageEditAudience/index.test.tsx
jaredlockhart: That's what I mean, the logic is preventing navigation to this page, but not preventing saving, so without an explicit logic check to control when save is available, I'm not sure this test is actually testing anything?  Not a big deal though.

PR #8322, opened 2023-02-17
For #7823 feat(nimbus): Updates to live rollouts set dirty publish status

eliserichards:
This commit...

* Updates publish status to `DIRTY` when a live update is made to a rollout
* Updates test_mutations

This commit _does not_...

* Enable the population percent field or the save/save & continue buttons on the audience page (#8239)

### DIFF
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1038,11 +1038,9 @@ class NimbusExperimentSerializer(
             and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
             and validated_data.get("publish_status")
             != NimbusConstants.PublishStatus.REVIEW
-            and validated_data.get("publish_status")
-            != NimbusConstants.PublishStatus.DIRTY
         ):
             # can be Live Update (Dirty), End Enrollment, or End Experiment
-            # if we don't check validated_data["publish_status"]
+            # (including rejections) if we don't check validated_data
             validated_data["publish_status"] = NimbusConstants.PublishStatus.DIRTY
 
         self.changelog_message = validated_data.pop("changelog_message")
--- a/app/experimenter/experiments/api/v5/serializers.py
+++ b/app/experimenter/experiments/api/v5/serializers.py
@@ -1030,6 +1030,21 @@ class NimbusExperimentSerializer(
         return data
 
     def update(self, experiment, validated_data):
+        if (
+            experiment.is_rollout
+            and not experiment.is_paused
+            and experiment.status == NimbusExperiment.Status.LIVE
+            and experiment.status_next is None
+            and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
+            and validated_data.get("publish_status")
+            != NimbusConstants.PublishStatus.REVIEW
+            and validated_data.get("publish_status")
+            != NimbusConstants.PublishStatus.DIRTY
+        ):
+            # can be Live Update (Dirty), End Enrollment, or End Experiment
+            # if we don't check validated_data["publish_status"]
+            validated_data["publish_status"] = NimbusConstants.PublishStatus.DIRTY
+
         self.changelog_message = validated_data.pop("changelog_message")
         return super().update(experiment, validated_data)
 
--- a/app/experimenter/experiments/tests/api/v5/test_mutations.py
+++ b/app/experimenter/experiments/tests/api/v5/test_mutations.py
@@ -433,7 +433,7 @@ class TestUpdateExperimentMutationSingleFeature(
         experiment = NimbusExperiment.objects.get(id=experiment_id)
         self.assertTrue(experiment.is_rollout)
         self.assertEqual(experiment.population_percent, 50.0)
-        self.assertEqual(experiment.publish_status, NimbusConstants.PublishStatus.IDLE)
+        self.assertEqual(experiment.publish_status, NimbusConstants.PublishStatus.DIRTY)
         self.assertEqual(experiment.status, NimbusConstants.Status.LIVE)
         self.assertEqual(experiment.status_next, None)
 
--- a/app/experimenter/experiments/tests/api/v5/test_mutations.py
+++ b/app/experimenter/experiments/tests/api/v5/test_mutations.py
@@ -6,6 +6,7 @@ from django.core.files.uploadedfile import SimpleUploadedFile
 from django.urls import reverse
 from graphene_django.utils.testing import GraphQLTestCase
 from graphene_file_upload.django.testing import GraphQLFileUploadTestCase
+from parameterized import parameterized
 
 from experimenter.base.tests.factories import (
     CountryFactory,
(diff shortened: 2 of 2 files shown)


app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1030,6 +1030,20 @@ def validate(self, data):
...
     def update(self, experiment, validated_data):
+        if (
+            experiment.is_rollout
+            and not experiment.is_paused
+            and experiment.status == NimbusExperiment.Status.LIVE
+            and experiment.status_next is None
+            and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
+            and validated_data.get("publish_status") == NimbusConstants.PublishStatus.IDLE
+            and validated_data.get("publish_status")
+            != NimbusConstants.PublishStatus.DIRTY
```
jaredlockhart, line 1041: This is redundant now with the equality comparison above?

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1030,6 +1030,20 @@ def validate(self, data):
         return data
 
     def update(self, experiment, validated_data):
+        if (
+            experiment.is_rollout
+            and not experiment.is_paused
+            and experiment.status == NimbusExperiment.Status.LIVE
+            and experiment.status_next is None
+            and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
+            and validated_data.get("publish_status") == NimbusConstants.PublishStatus.IDLE
```
jaredlockhart, line 1039: It's probably worth adding a parameterized test to test a bunch of these cases?  No need to be exhaustive but like, a case where it matches everything and then mutate the match everything case for each of those fields and assert false is probably sufficient?

jaredlockhart (dismissed):

app/experimenter/experiments/api/v5/serializers.py
eliserichards: Great idea!! 💯

app/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1030,6 +1030,19 @@ def validate(self, data):
...
 
     def update(self, experiment, validated_data):
+        if (
+            experiment.is_rollout
+            and not experiment.is_paused
+            and experiment.status == NimbusExperiment.Status.LIVE
+            and experiment.status_next is None
+            and experiment.publish_status == NimbusExperiment.PublishStatus.IDLE
+            and validated_data.get("publish_status")
+            != NimbusConstants.PublishStatus.REVIEW
```
eliserichards, line 1040: Ok so I changed this back from `== NimbusConstants.PublishStatus.IDLE` to `!= NimbusConstants.PublishStatus.REVIEW` because it can also = `IDLE` when it is an end enrollment/end experiment _rejection_. The good thing is that the test cases caught it immediately ;)

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -433,10 +434,65 @@ def test_update_live_rollout(self):
...
+            [False, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_APPROVE, True],
+            [True, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_REJECT, False],
+            [True, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_REJECT, True],
+            [True, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_TIMEOUT, False],
+            [True, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_TIMEOUT, True],
+            [True, NimbusExperimentFactory.Lifecycles.LIVE_REJECT_MANUAL_ROLLBACK, False],
+            [True, NimbusExperimentFactory.Lifecycles.LIVE_REJECT_MANUAL_ROLLBACK, True],
+        ]
+    )
+    def test_update_rollout_is_dirty(self, is_dirty_expected, lifecycle, is_paused):
```
eliserichards, line 467: Testing updating population percentage with a combination of lifecycle and paused status, then seeing what is made dirty

app/experimenter/experiments/api/v5/serializers.py
jaredlockhart: The aggregate of all six fields together is the unique state we want:

<google-sheets-html-origin><style type="text/css"></style>

publish status | IDLE | publish status | DIRTY (IDLE)
-- | -- | -- | --
status | LIVE | status | LIVE
status next | None | status next | None

app/experimenter/experiments/tests/api/v5/test_mutations.py
jaredlockhart: Move is_paused back to lifecycles

app/experimenter/experiments/api/v5/serializers.py
eliserichards: wait wait wait. The validated_data won't necessarily have `status`, `publish_status`, or `status_next`. In this case, we are only updating `population_percent`: 

image

eliserichards: So we can't rely on the uniqueness of those six fields. We have to structure it to include the fields that will be coming through `update`

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: https://github.com/mozilla/experimenter/issues/8334

app/experimenter/experiments/api/v5/serializers.py
yashikakhurana: oh do we need this check  
```
validated_data.get("publish_status")
            != NimbusConstants.PublishStatus.REVIEW
``` 
since we are checking all status, status next and publish status?

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -433,10 +434,55 @@ def test_update_live_rollout(self):
...
+            },
+            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+        )
+        self.assertEqual(response.status_code, 200)
+
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        self.assertEqual(experiment.population_percent, 50.0)
+        self.assertEqual(
+            experiment.publish_status == NimbusConstants.PublishStatus.DIRTY,
+            is_dirty_expected,
```
yashikakhurana, line 483: worth asserting status_next and status too, just to double check that it is working as expected

yashikakhurana (dismissed):
few minor questions, thank you @eliserichards making good progress 💪

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: The only thing it should be changing is PublishStatus -> DIRTY, so I don't think we need to add those to the parameters

app/experimenter/experiments/api/v5/serializers.py
eliserichards: Yes, both "End Experiment" and "End Enrollment" go through this `update` (they both have `publish_status=REVIEW`)

eliserichards: If we are making a live update, like the example above, all we will have in `validated_data` is changelog and population_percent

eliserichards: image

If we were guaranteed to always have pub status, status next, and status in `validated_data`, we could use these six fields to determine if it should be dirty. Since we can't guarantee that, we have to rule out these other four cases based on what data comes through `update`:

image

app/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -433,10 +434,55 @@ def test_update_live_rollout(self):
...
+            # [True, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_TIMEOUT],
+            # [True, NimbusExperimentFactory.Lifecycles.LIVE_REJECT_MANUAL_ROLLBACK],
+            # [True, NimbusExperimentFactory.Lifecycles.LIVE_DIRTY],
+            # [True, NimbusExperimentFactory.Lifecycles.LIVE_REJECT],
+            # [False, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE_APPROVE],
+            # [False, NimbusExperimentFactory.Lifecycles.LIVE_ENROLLING],
+            # [False, NimbusExperimentFactory.Lifecycles.LIVE_PAUSED],
+            # [False, NimbusExperimentFactory.Lifecycles.LIVE_REVIEW_REQUESTED],
+            # [False, NimbusExperimentFactory.Lifecycles.LIVE_APPROVE],
+            # [False, NimbusExperimentFactory.Lifecycles.LAUNCH_APPROVE_APPROVE],
```
jaredlockhart, line 454: ?

jaredlockhart (dismissed):

app/experimenter/experiments/tests/api/v5/test_mutations.py
eliserichards: poop. pushing changes now

jaredlockhart (approved):
Okay great, this is nice and clear and concise.  The checks in the update method are nice and clear, your reasoning with the table is very clear, the test is nice and concise and parameterizing it with the appropriate lifecycles makes it very clear to see where it does and doesn't apply.  If we land this and find there are some states that aren't covered correctly, it'll be very clear and easy to adjust this.  This PR looks great, thank you for being patient through all the feedback.  R+!

Issue #8321, opened 2023-02-17
Indicate unpublished changes 


eliserichards:
Should show something on list page (home) and every page for the rollout to indicate that it is Dirty

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3170)

PR #8337, opened 2023-02-22
For #8321 feat(nimbus) test(nimbus): Definitions for update and update reviews in getStatus

eliserichards:
Because...

* We want to signal to the user on PageSummary that a rollout has unpublished updates pending (DIRTY publish status) 
* And we want to use the `getStatus` definitions to determine if it's in the following states:
   * Dirty
   * Update requested
   * Update approved & waiting for remote settings approval

This commit...

* Updates the `getStatus` definitions to include `updateRequested` and `updatedRequestedWaiting`
* Not sure why the test coverage wasn't complaining, but I added tests in `experiment.test.ts`

### DIFF
--- a/app/experimenter/nimbus-ui/src/lib/experiment.test.ts
+++ b/app/experimenter/nimbus-ui/src/lib/experiment.test.ts
@@ -58,6 +58,21 @@ describe("getStatus", () => {
 
     experiment.publishStatus = NimbusExperimentPublishStatusEnum.WAITING;
     expect(getStatus(experiment).waiting).toBeTruthy();
+
+    experiment.publishStatus = NimbusExperimentPublishStatusEnum.DIRTY;
+    expect(getStatus(experiment).dirty).toBeTruthy();
+
+    experiment.publishStatus = NimbusExperimentPublishStatusEnum.REVIEW;
+    experiment.status = NimbusExperimentStatusEnum.LIVE;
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/lib/experiment.test.ts
```diff
@@ -58,6 +58,21 @@ describe("getStatus", () => {
...
+    experiment.status = NimbusExperimentStatusEnum.LIVE;
+    experiment.statusNext = NimbusExperimentStatusEnum.LIVE;
+    experiment.isRollout = true;
+    expect(getStatus(experiment).updateRequested).toBeTruthy();
+
+    experiment.publishStatus = NimbusExperimentPublishStatusEnum.APPROVED;
+    expect(getStatus(experiment).updateRequestedApproved).toBeTruthy();
+
+    experiment.publishStatus = NimbusExperimentPublishStatusEnum.WAITING;
+    expect(getStatus(experiment).updateRequestedWaiting).toBeTruthy();
```
jaredlockhart, line 75: Oh wow this is all in one test?  Woof.  Oh well it's already the pattern.

eliserichards: I know, I'd like to break them out at some point :/

PR #8339, opened 2023-02-22
fix #8338 chore(nimbus): refactor fetch logic

### LINKED ISSUE #8338: Refactor analysis results data fetch
The logic is a bit confusing and is leading to some weird behavior in tests vs production. A refactor to explicitly look for what to include (vs exclude) should make this clearer.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3182)

mikewilli:
Because

* the logic to fetch analysis results looks for the ones to skip, and then falls into fetch logic if we don't want to skip
* this flow may be confusing to parse

This commit

* seeks the experiments to fetch instead of the ones to skip
* makes the logic a little more explicit
* fixes a test that was passing in some cases it shouldn't

### DIFF
--- a/app/experimenter/jetstream/tasks.py
+++ b/app/experimenter/jetstream/tasks.py
@@ -1,4 +1,4 @@
-import datetime
+import datetime as dt
 
 import markus
 from celery.utils.log import get_task_logger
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Much clearer!!! Thank you 🙏

PR #8344, opened 2023-02-23
fix #8332 feat(nimbus): better filtering/UX around experiment errors

### LINKED ISSUE #8332: Results Page: Filter out experiment-level errors to last run only
We filter out all errors shown on the Results page to within the last weekly window, but this can still leave some noise. We should be safe to filter out experiment-level errors that were not in the last run.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3173)

mikewilli:
Because

* we include all error logs going back to the last weekly run
* only last-run errors are relevant for experiment-level errors (as opposed to metric-level errors)
* there is no visibility in the UI about when errors occurred

This commit

* filters experiment-level errors on ingest to only include those from the last run
* adds timestamp to the UI for errors


app/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -350,7 +350,8 @@ def test_valid_results_data_parsed_and_stored(
                             ],
                             "default_metrics": [],
                         }
-                    }
+                    },
+                    "analysis_start_time": "2022-08-31T04:30:03+00:00",
```
jaredlockhart, line 354: The logic is checking for cases where either of `analysis_start_time` or `timestamp` can be present or missing, should we be testing for the four combinations of those cases?  Maybe a parameterized test?

app/experimenter/jetstream/client.py
```diff
@@ -261,7 +262,19 @@ def get_experiment_data(experiment):
...
                 errors_by_metric[metric_slug].append(err)
             else:
-                errors_experiment_overall.append(err)
+                try:
+                    analysis_start_time = datetime.fromisoformat(
+                        experiment_metadata.get("analysis_start_time", "")
+                        if experiment_metadata is not None
+                        else ""
+                    )
+                    timestamp = datetime.fromisoformat(err.get("timestamp", ""))
```
jaredlockhart, line 271: So in both of these cases we pass an empty string to `fromisoformat`, but why not flip it the other way around and say `analysis_start_time` and `timestamp` are `None` unless each are an actual iso datetime in which case parse and compare?  I see this locally:

```
>>> datetime.datetime.fromisoformat("")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: ''
```

jaredlockhart (changes requested):

app/experimenter/jetstream/client.py
mikewilli: The problem is determining in advance (and with certainty) whether the timestamps are going to be valid without passing them to `fromisoformat`, which means we're risking an exception (`ValueError` if string is invalid, `TypeError` if not a string) either way. This seemed like the cleanest way to handle that. The default `""` in the `get()` calls are probably superfluous though.

app/experimenter/jetstream/client.py
```diff
@@ -261,7 +262,19 @@ def get_experiment_data(experiment):
...
+                        if experiment_metadata is not None
+                        else ""
+                    )
+                    timestamp = datetime.fromisoformat(err.get("timestamp"))
+
+                    if timestamp >= analysis_start_time:
+                        errors_experiment_overall.append(err)
+                except (ValueError, TypeError):
+                    # ill-formatted/missing timestamp: default to including the error
+                    errors_experiment_overall.append(err)
```
jaredlockhart, line 278: Oh I see what you mean, you just proceed as normal and if it throws then you just append the error.  Oh then yeah I wouldn't even bother with all the logic to do fallbacks, just look up the key and parse it and add `KeyError` to your exception handler.

jaredlockhart (approved):
Awesome thank you for adding the test 🙏

PR #8348, opened 2023-02-23
For #8238 feat(nimbus): Allocate buckets for rollouts

eliserichards:
Because...

This commit...

* Allocates buckets for live approved rollouts

---- 

This commit _does not_...
* Add a warning to `PageSummary` when you are attempting to create an experiment that would end up in the same bucket as an existing experiment -- this will be the next PR

### DIFF
--- a/app/experimenter/experiments/models.py
+++ b/app/experimenter/experiments/models.py
@@ -256,6 +256,12 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                 status=NimbusConstants.Status.DRAFT,
                 publish_status=NimbusConstants.PublishStatus.APPROVED,
             )
+            | Q(
+                is_rollout=True,
+                status=NimbusConstants.Status.LIVE,
+                status_next=NimbusConstants.Status.LIVE,
+                publish_status=NimbusConstants.PublishStatus.APPROVED,
+            )
         )
 
     def __str__(self):
--- a/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
+++ b/app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
@@ -662,9 +662,81 @@ class TestNimbusExperimentSerializer(TestCase):
         self.assertTrue(NimbusBucketRange.objects.filter(experiment=experiment).exists())
         self.assertEqual(experiment.bucket_range.count, 5000)
 
+    def test_end_enrollment_does_not_change_bucket_allocation(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.ENDING_REVIEW_REQUESTED,
+            is_rollout=True,
+            population_percent=Decimal("50.0"),
+        )
+        bucket_id = experiment.bucket_range.id
+
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "publish_status": NimbusExperiment.PublishStatus.REVIEW,
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer.is_valid())
+        experiment = serializer.save()
+
+        self.assertTrue(NimbusBucketRange.objects.filter(experiment=experiment).exists())
+        self.assertEqual(experiment.bucket_range.count, 5000)
+        self.assertEqual(experiment.bucket_range.id, bucket_id)
+
+    def test_end_experiment_does_not_change_bucket_allocation(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.PAUSING_REVIEW_REQUESTED,
+            is_rollout=True,
+            population_percent=Decimal("50.0"),
+        )
+        bucket_id = experiment.bucket_range.id
+
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "publish_status": NimbusExperiment.PublishStatus.REVIEW,
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer.is_valid())
+        experiment = serializer.save()
+
+        self.assertTrue(NimbusBucketRange.objects.filter(experiment=experiment).exists())
+        self.assertEqual(experiment.bucket_range.count, 5000)
+        self.assertEqual(experiment.bucket_range.id, bucket_id)
+
+    def test_live_rollout_allocates_buckets(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LIVE_APPROVE,
+            is_rollout=True,
+            population_percent=Decimal("50.0"),
+        )
+
+        bucket_id = experiment.bucket_range.id
+
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "publish_status": NimbusExperiment.PublishStatus.APPROVED,
+                "changelog_message": "test changelog message",
+            },
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer.is_valid())
+
+        experiment = serializer.save()
+
+        self.assertTrue(NimbusBucketRange.objects.filter(experiment=experiment).exists())
+        self.assertEqual(experiment.bucket_range.count, 5000)
+        self.assertNotEqual(experiment.bucket_range.id, bucket_id)
+
     def test_live_experiment_does_not_allocate_buckets(self):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.PAUSING_REVIEW_REQUESTED,
+            is_rollout=False,
             population_percent=Decimal("50.0"),
         )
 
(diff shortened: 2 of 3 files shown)


app/experimenter/experiments/models.py
```diff
@@ -595,10 +601,13 @@ def bucket_namespace(self):
             feature_config.slug
             for feature_config in self.feature_configs.all().order_by("slug")
         )
+
         if self.channel:
             keys.append(self.channel)
 
         if self.is_rollout:
+            if self.targeting_config_slug:
+                keys.append(self.targeting_config_slug.replace("_", "-"))
```
jaredlockhart, line 610: Why replace?

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -662,9 +662,56 @@ def test_publish_status_approved_generates_bucket_allocation(self):
         self.assertTrue(NimbusBucketRange.objects.filter(experiment=experiment).exists())
         self.assertEqual(experiment.bucket_range.count, 5000)
 
+    def test_end_enrollment_does_not_change_bucket_allocation(self):
```
jaredlockhart, line 665: Great idea 🙏

app/experimenter/experiments/models.py
```diff
@@ -256,6 +256,12 @@ class Filters:
                 status=NimbusConstants.Status.DRAFT,
                 publish_status=NimbusConstants.PublishStatus.APPROVED,
             )
+            | Q(
+                is_rollout=True,
+                status=NimbusConstants.Status.LIVE,
+                status_next=NimbusConstants.Status.LIVE,
+                publish_status=NimbusConstants.PublishStatus.APPROVED,
```
jaredlockhart, line 263: If we're adding this here, shouldn't there be a test below to test the case that we do reallocate buckets for a live update?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1352,10 +1352,12 @@ def test_allocate_buckets_generates_bucket_range(self):
         feature = NimbusFeatureConfigFactory(slug="feature")
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.CREATED,
+            is_rollout=False,
             application=NimbusExperiment.Application.DESKTOP,
             channel=NimbusExperiment.Channel.RELEASE,
             feature_configs=[feature],
             population_percent=Decimal("50.0"),
+            targeting_config_slug=NimbusExperiment.TargetingConfig.MAC_ONLY,
```
jaredlockhart, line 1360: If `is_rollout` is `False` then why do we need to add a targeting config slug here?

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1384,6 +1386,7 @@ def test_allocate_buckets_deletes_buckets_and_empty_isolation_group(self):
...
```
jaredlockhart, line 1389: And why do we need to add them to all of these?

jaredlockhart (changes requested):

jaredlockhart:
Also it seems like there's two unrelated changes in here

1) When to reallocate buckets
2) To include the targeting config slug in the namespace for rollouts

This could've been split up?

app/experimenter/experiments/models.py
eliserichards: Just for consistency. The rest of the bucketing is `xxx-xxxx-xxx-xxx` so it looked kinda wonky with `xx-xxx-xxx-xxx_xxxx_xxxx`

eliserichards:
Sure, I can do that

app/experimenter/experiments/tests/test_models.py
eliserichards: Not in this one, since it's not a rollout

app/experimenter/experiments/models.py
jaredlockhart: Right but it's not actually necessary for the behaviour and logic, and particularly since if you look at it and go 'where did this part come from' and it's not actually what's in the targeting config, that would be confusing.  I wouldn't do this.

app/experimenter/experiments/tests/test_models.py
jaredlockhart: I don't understand what you mean?

eliserichards: we don't need to add targeting_config_slug here because it's not a rollout

app/experimenter/experiments/models.py
eliserichards: It's below in `test_models`

jaredlockhart (changes requested):
Oh good yes thank you except

1) Now this PR is titled and referencing the wrong issue
2) We're still missing the test to test the new should allocate condition

app/experimenter/experiments/tests/test_models.py
```diff
@@ -1375,6 +1376,23 @@ def test_allocate_buckets_creates_new_bucket_range_if_population_changes(self):
         experiment.allocate_bucket_range()
         self.assertEqual(experiment.bucket_range.count, 2000)
 
+    def test_allocate_buckets_for_live_approved_rollout(self):
```
eliserichards, line 1379: @jaredlockhart here

jaredlockhart: Does this test that condition though?

eliserichards:
@jaredlockhart you re-reviewed right as I was changing the title lol

app/experimenter/experiments/tests/test_models.py
eliserichards:
```
                is_rollout=True,
                status=NimbusConstants.Status.LIVE,
                status_next=NimbusConstants.Status.LIVE,
                publish_status=NimbusConstants.PublishStatus.APPROVED,
```

`NimbusExperimentFactory.Lifecycles.LIVE_APPROVE,`:
```
                status=NimbusConstants.Status.LIVE,
                status_next=NimbusConstants.Status.LIVE,
                publish_status=NimbusConstants.PublishStatus.APPROVED,
```

image

jaredlockhart: Right, and how does explicitly calling `experiment.allocate_bucket_range()` cover that line above?

jaredlockhart: Said another way, if you revert the change to `SHOULD_ALLOCATE_BUCKETS` will this model test still pass?

eliserichards: true! I'll move this check into the serializer test

app/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -662,9 +662,81 @@ def test_publish_status_approved_generates_bucket_allocation(self):
...
+            context={"user": self.user},
+        )
+        self.assertTrue(serializer.is_valid())
+        experiment = serializer.save()
+
+        self.assertTrue(NimbusBucketRange.objects.filter(experiment=experiment).exists())
+        self.assertEqual(experiment.bucket_range.count, 5000)
+        self.assertEqual(experiment.bucket_range.id, bucket_id)
+
+    def test_live_rollout_allocates_buckets(self):
```
eliserichards, line 711: @jaredlockhart here :)

jaredlockhart (changes requested):

app/experimenter/experiments/tests/test_models.py
eliserichards: Experiments are tested in here, so rollouts should be too. I think both tests are valid

jaredlockhart: Right but waht I mean is, if you remove the change to SHOULD_ALLOCATE_BUCKETS, will this test still pass?

eliserichards: Yes, but it increases our consistency in the tests. This case wasn't covered here before

eliserichards: I would like to keep it

jaredlockhart: Sure, but I'm just trying to encourage you to think about the relationship between changes and test cases.  If this test will pass regardless of the changes in a PR, then we know it's unrelated to the changes in this PR.  If it's covering a case that isn't covered at all, like we see some failure in prod but all tests are passing, that's totally valid, but unrelated to this change.  In general there should be some motivation for adding a test that's related to the expected behaviour of the application, whether it's a new change, a bug fix, etc, and not be re testing things that are already covered by other tests.

jaredlockhart (approved):
Looks good thank you! 👍

app/experimenter/experiments/tests/test_models.py
eliserichards: That's a fair point. In that case a lot of the model & serializer tests may be overlapping their coverage; maybe we should take a look at them

jaredlockhart: Totally, I do that all the time, find the same thing being tested in multiple places and then nuke the duplicates.  Like the theory is, you test all the cases at the lowest level, and then each level above you only test integration.  Of course that gets lost over time so you gotta periodically go back in and remove duplicates.  Also in this particular case it occurred to me afterwards that it's the same as the test above it so we could've also accomplished what I think the spirit of what you were going for with just parameterizing against `is_rollout`.  But anyways as usual, I like tests, tests are good, just good to keep in mind what's being tested where 👍

PR #8350, opened 2023-02-24
For #8238 feat(nimbus): Add targeting slug to bucketing namespace

eliserichards:
Because...

This commit...

* Adds the`targeting_config_slug` to the bucketing namespace for rollouts
* Updates tests

---- 

This commit _does not_...
* Add a warning to `PageSummary` when you are attempting to create an experiment that would end up in the same bucket as an existing experiment -- this will be the next PR


jaredlockhart (approved):
A+ 10/10 would namespace a bucket again 🎉 🎉 🎉

eliserichards:
Broken out from #8348

PR #8353, opened 2023-02-27
For #8321 feat(nimbus): Show unpublished changes badge on summary page

eliserichards:
Because.. 

* We want to notify the user if their rollout has unpublished changes (publish_status=DIRTY)

This commit...

* Adds a red status pill to the summary page that shows up next to the "Timeline" section when the experiment is Dirty

image

This commit _does not..._
* Add an "unpublished changes" indicator to the home page. That will be added in a different PR.


jaredlockhart (approved):
Excellent A+ 10/10 would unpublish changes again 🙆‍♀️ 

We might also want to put something like this on the list page but we can think about that separately, this is great thank you. 🙏

Issue #8359, opened 2023-02-28
Add duplicate bucket warning for rollouts


data-sync-user:
We added advanced targeting as one of the bucketing identifiers so that every one of those combinations of rollout/feature/channel/adv.targeting would be unique

We should add a warning that you’re trying to launch a rollout for the same group of people who already have one — already live for same rollout, application feature, channel , advanced targeting

Wording: 

```
"WARNING: A rollout already exists for this combination
of application, feature, channel, and advanced targeting! 
If this rollout is launched, a client meeting the advanced targeting criteria 
will be <enrolled in both/enrolled in one but not the other> and 
you will not be able to adjust the sizing for this rollout."
```

Figure out what happens if someone does launch a rollout in a dupe. Will they be enrolled in both? If enrolled in only one, how is that decided?

* the client enforces per feature:
  * one experiment
  * one rollout
* It would just not enroll in the second rollout, but which one is **undefined behaviour** (it's the order it evaluates them from Remote Settings, which is undefined)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3198)

PR #8362, opened 2023-03-01
For #8359 feat(nimbus): Show warning on pagesummary when a rollout has dupe bucket

eliserichards:
Because

- We want to warn a user if they are about to launch a rollout that has the same **channel, feature, application, and advanced targeting** as another existing rollout 
- And we only want to warn them, not throw an error

This commit

- Shows the user a warning message before they launch their rollout and meets the following criteria:
   - status is Draft or Preview
   - the experiment is a rollout
   - there is at least one other live rollout existing that has the same channel, feature, application, and advanced targeting 

image


yashikakhurana (approved):
💯 amazing test cases, well define PR

Issue #8361, opened 2023-02-28
Update button should appear alongside end buttons


data-sync-user:
Right now, when a rollout has been edited (is publish_status=DIRTY), the end enrollment/end experiment buttons are not shown and the “Update” buttons are shown instead. We should be showing all three options at the same time.

!image-20230224-193429.png|width=447,height=319!

LIVE/IDLE: request end enrollment, request end
LIVE/DIRTY: request update, request end enrollment, request end
LIVE/REVIEW: cancel

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3192)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/34345">image-20230224-193429.png</a>

data-sync-user:
➤ Jared Lockhart commented:

Exactly:

LIVE/IDLE: request end enrollment, request end
LIVE/DIRTY: request update, request end enrollment, request end
LIVE/REVIEW: cancel

PR #8363, opened 2023-03-01
For #8361 feat(nimbus): End enroll and end experiment buttons appear when dirty

eliserichards:
Because

- We want live updates to have the option to end enrollment and end experiment when they are Dirty

This commit

- Adds `status.dirty` check to the end enrollment and end experiment buttons

image

----

This commit _does not_...
* Remove the "cancel" button from the Update actions
* Put the update button down with the end enroll/end exp buttons

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
@@ -76,6 +76,31 @@ describe("Summary", () => {
     await screen.findByText("End Experiment");
   });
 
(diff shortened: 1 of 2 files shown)


jaredlockhart (changes requested):
Can we get rid of the `Cancel` button there that doesn't actually do anything and have all three of those buttons together? 🙏

jaredlockhart (approved):
Very lovely concise little change 🌹

eliserichards:
I just updated the description to make it clear that the cancel button will come in another PR :D

PR #8365, opened 2023-03-01
For #8361 refactor(nimbus): Remove unnecessary cancel button from update form

eliserichards:
Because

- We want live updates to have the option to update when they are Dirty
- We don't care about "cancelling" a dirty update 

This commit

- Removes the "cancel" button from the Update actions

image

----

This commit _does not_...
* Put the update button down with the end enroll/end exp buttons


jaredlockhart (approved):
🙏

PR #8366, opened 2023-03-01
For #8361 feat(nimbus): Put request update button alongside end buttons on Summary

eliserichards:
Because

- We want the "Request update" buttons to appear in the same place as "request launch"
- And we want the end experiment and end enrollment buttons to still be available under "Actions"

This commit

- Moves the "Request update" button from PageSummary →  Summary
   - ~This puts the Update button in the same place that the Launch button is for drafts (see screenshots)~
   - Put the Update button in the Actions section next to end enrollment and end experiment
- Moves tests from PageSummary -> Summary
- Renames the "Request update" from `FormUpdateLiveToReview` → `RequestLiveUpdate`
- Removes `treatmentBranches` and `enrollmentPaused=true` from the test mock for Live Rollouts (this was an oversight)

----

Note: the position of "update" is the same as "launch" buttons for Drafts (above the "actions" section in the grey box)

image

vs Draft: 

image

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
@@ -350,4 +350,38 @@ describe("Summary", () => {
     expect(requestUpdateButton).toBeEnabled();
     await act(async () => void fireEvent.click(requestUpdateButton));
   });
+
+  it("shows update and end buttons for live dirty rollout", async () => {
+    const { mockRollout, rollout } = mockLiveRolloutQuery("demo-slug", {
+      status: NimbusExperimentStatusEnum.LIVE,
+      publishStatus: NimbusExperimentPublishStatusEnum.DIRTY,
+      statusNext: null,
+    });
+
+    const mutationMock = createMutationMock(
+      rollout.id!,
+      NimbusExperimentPublishStatusEnum.DIRTY,
+      {
+        changelogMessage: CHANGELOG_MESSAGES.REQUESTED_REVIEW_UPDATE,
+        statusNext: null,
+        publishStatus: NimbusExperimentPublishStatusEnum.DIRTY,
+        status: NimbusExperimentStatusEnum.LIVE,
+      },
+    );
+    render(<Subject props={rollout} mocks={[mockRollout, mutationMock]} />);
+    const requestUpdateButton = await screen.findByTestId(
+      "request-live-update-alert",
+    );
+    const endExperimentButton = await screen.findByTestId(
+      "end-experiment-start",
+    );
+    const endEnrollmentButton = await screen.findByTestId(
+      "end-enrollment-start",
+    );
+    await waitFor(() => {
+      expect(requestUpdateButton).toBeInTheDocument();
+      expect(endEnrollmentButton).toBeInTheDocument();
+      expect(endExperimentButton).toBeInTheDocument();
+    });
+  });
 });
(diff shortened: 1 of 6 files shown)


eliserichards:
I can break out the renaming of the update form into another PR if y'all want, but I figured since it was already moving to another component it wasn't too much for this 🤔

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -318,4 +324,30 @@ describe("Summary", () => {
...
+      {
+        changelogMessage: CHANGELOG_MESSAGES.REQUESTED_REVIEW_UPDATE,
+        statusNext: null,
+        publishStatus: NimbusExperimentPublishStatusEnum.DIRTY,
+        status: NimbusExperimentStatusEnum.LIVE,
+      },
+    );
+    render(<Subject props={rollout} mocks={[mockRollout, mutationMock]} />);
+
+    const requestUpdateButton = await screen.findByTestId(
```
yashikakhurana, line 347: can you also add a test that end enrollments and other buttons are still visible with this? :)

yashikakhurana (changes requested):

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
yashikakhurana: one thing: I think it's worth saying, instead of End experiment-> End rollout, what do you think?

eliserichards: Ooo yeah we can definitely do that. I'll file a ticket!

eliserichards: https://github.com/mozilla/experimenter/issues/8368

yashikakhurana: fantastic, thank you 🙏

yashikakhurana:
oh also in the screenshot,end experiment is appearing two times 🤔

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -318,4 +324,54 @@ describe("Summary", () => {
...
+      "request-live-update-alert",
+    );
+    expect(requestUpdateButton).toBeEnabled();
+    await act(async () => void fireEvent.click(requestUpdateButton));
+  });
+
+  it("shows update and end buttons for live dirty rollout", async () => {
```
eliserichards, line 354: @yashikakhurana what do you think?

yashikakhurana: yup 💯 , thank you @eliserichards

yashikakhurana: and now when we will change the name from `end experiment` to `end rollout`, your super test will be able to catch this 👏

jaredlockhart:
Why can't we just take this opportunity to put them all together in one place while we're poking around in here?

eliserichards:
It matches what we do for Draft. There's a description up by the screenshots ☝️

jaredlockhart:
Yes, I mean I want to change that design altogether.  Have all the buttons in one place together.

eliserichards:
Is that what we want? I was under the assumption that the Launch (and Update) were separated visually with the grey  box because they took priority over the other actions

eliserichards:
And if we want to group them together, does that apply to draft/preview too? or just update

eliserichards:
This can be closed in favor of #8370

PR #8370, opened 2023-03-01
For #8361 feat(nimbus): Match update button with end enroll and end experiment buttons

eliserichards:
Because

- We want the "Request update" buttons to appear in the "Actions" section alongside the "End experiment" and "End enrollment" buttons

This commit

- Removes the Alert box and text from the Update form
- Renames the "Request update" from `FormUpdateLiveToReview` → `RequestLiveUpdate`
- Adds a test for viewing all three buttons

----

image


eliserichards:
> What's the relationship between this and #8366?

I closed the other one in favor of this PR. We no longer need to move the Update button into Summary if it's going to live with the end enrollmnet/end experiment buttons :)

jaredlockhart (approved):
Perfect thank you! 🎉

PR #8373, opened 2023-03-02
For #8361 feat(nimbus): Add changelog message for request update rejection

eliserichards:
Because

- We want to show a changelog message when a requested update is rejected in Experimenter

This commit

- Adds the changelog message


jaredlockhart (commented):
Oh I see, the frontend defines this and it's passed in as the message when this occurs.  Updating a test case somewhere would've helped me better understand where this is being used.  Is there not a test that can assert on this somewhere?

yashikakhurana (approved):
🍒

eliserichards:
> Oh I see, the frontend defines this and it's passed in as the message when this occurs. Updating a test case somewhere would've helped me better understand where this is being used. Is there not a test that can assert on this somewhere?

Yeah, it doesn't seem like we're asserting anything about the changelog messages in these tests. I'll make a note to update them

PR #8374, opened 2023-03-02
fix #8372 feat(nimbus): add changelogs to v5 api

### LINKED ISSUE #8372: Add all changelogs to V5 API
It can be helpful to inspect the changelog history an experiment for testing and debugging.  The only way to access this right now is through the admin but that can be slow and dangerous.  Exposing it through the V5 API will allow anyone to query it through the graphql web interface.  This should assist in prod debugging and testing.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3210)

jaredlockhart:
Because

* We do not currently expose the changelogs in the UI anywhere
* The only way to inspect them presently is through the admin
* They can provide helpful information for inspecting and testing and debugging experiments
* There should be a way to inspect them without needing admin access
* They can contain confidential information and so they should still be behind an auth gateway

This commit
* Adds all experiment changelogs to the V5 API
* This allows them to be inspected via the graphql web ui
* The V5 API is behind the auth gate

### DIFF
--- a/app/experimenter/experiments/api/v5/types.py
+++ b/app/experimenter/experiments/api/v5/types.py
@@ -245,10 +245,21 @@ class NimbusReviewType(graphene.ObjectType):
 class NimbusChangeLogType(DjangoObjectType):
     old_status = NimbusExperimentStatusEnum()
     old_status_next = NimbusExperimentStatusEnum()
+    new_status = NimbusExperimentStatusEnum()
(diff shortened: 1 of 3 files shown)


yashikakhurana (approved):
yup looks perfect with an awesome test case, thank you @jaredlockhart

Issue #8378, opened 2023-03-02
Integration test for editing live rollout


data-sync-user:
Once all of the review forms are ready to go, we can enable the population percent fields on the audience page for live rollouts. Let’s make an integration test for “create - launch - update - launch”.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3212)

data-sync-user:
➤ Benjamin Forehand Jr. commented:

I can see a few test cases here. Do you mind defining the cases here?

eliserichards:
When an update is requested (“Request update” button in the “Actions” sections), the user can go through the familiar review flow (approve/reject in Experimenter, then approve/reject in Remote Settings). 

We want our tests to cover the following scenarios:

   * Live rollout → Edit population % → Request update → Approve in Exp → Approve in RS
   * Live rollout → Edit population % → Request update → Approve in Exp → Reject in RS
   * Live rollout → Edit population % → Request update → Reject in Exp

eliserichards:
The summary page for the experiment should show Live status, no "Unpublished changes" status or "Request update", and should reflect the updated population percentage

data-sync-user:
➤ Elise Richards commented:

🚩 Flag added

Having some weird errors on CI that I can't reproduce locally. Once Ben is back we can pair on this

PR #8379, opened 2023-03-02
For #7823 feat(nimbus): Enable population percent and save buttons for live rollouts

eliserichards:
Because

- We want to allow users to edit the population percentages for live rollouts

This commit

- Enables population percent for live rollouts

image

----
This commit _does not_...
- Add an integration test. This will be next in #8378 🎉


jaredlockhart:
Okay first bug detected we'll want to file separately and fix first:

Rejecting an update takes the experiment back to IDLE instead of DIRTY:

image

eliserichards:
We'll merge #8383 first

eliserichards:
Closing this for now since we want to do a few other small things first 👍

Issue #8382, opened 2023-03-03
Rejection of live edit should go back to Dirty


eliserichards:
Right now it's going to Idle instead of Dirty when a review is rejected in Remote Settings

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3213)

PR #8383, opened 2023-03-03
For #8382 feat(nimbus): Rejection from RS sets dirty publish status for updates

eliserichards:
Because

- We want a live rollout that is updated, sent to review, and rejected in Remote Settings to be set back to Dirty

This commit

- Sets the publish status to Dirty when rejected from RS

image


jaredlockhart (approved):
Yep this definitely corrects that case, tested locally.  Thanks @eliserichards ! 🎉

PR #8384, opened 2023-03-04
fix #8377 feat(cirrus): Initial commit

### LINKED ISSUE #8377: Set up Cirrus and Fast API
Since we are adding Cirrus service to support non-firefox applications, set up cirrus in the root folder of the experimenter repo as a subfolder server and client. In a server set up FastApi with the docker file.
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3211)

yashikakhurana:
Because

- We want to add Cirrus to the experimenter repo

This commit

-  Set up the cirrus server and requirements
-  Set up Dockerfile and include cirrus in docker-compose file
-  To run the cirrus server- use command `make cirrus_up`


Makefile
```diff
@@ -193,3 +193,11 @@ integration_test_nimbus: build_prod
...
 integration_test_nimbus_rust: build_prod
 	MOZ_HEADLESS=1 $(COMPOSE_INTEGRATION) run rust-sdk sh -c "chmod a+rwx /code/app/tests/integration/.tox;tox -c app/tests/integration -e integration-test-nimbus-rust $(TOX_ARGS) -- -n 2 $(PYTEST_ARGS)"
+
+
+
+cirrus_up:  ## Run Docker Compose services
+	$(COMPOSE) up web
+
+cirrus_down:  ## Shutdown Docker Compose services
+	$(COMPOSE) down
```
jeddai, line 203: Can we reference the docker compose service so it only shuts down Cirrus?

docker-compose.yml
```diff
@@ -101,6 +101,16 @@ services:
     ports:
       - "8000:8000"
 
+  web:
```
jeddai, line 104: Considering this is in a compose with a number of other web services, would it maybe make sense for the service to be called Cirrus?

jeddai (commented):
Just a couple thoughts but this is looking good!

docker-compose.yml
yashikakhurana: good call

docker-compose.yml
```diff
@@ -101,6 +101,16 @@ services:
...
       - "8000:8000"
 
+  cirrus:
+    build:
+      context: .
+      dockerfile: cirrus/server/Dockerfile
+    volumes:
+      - .:/cirrus/server
+    ports:
+      - "8000:8000"
```
jaredlockhart, line 111: 8000 is already in use by autograph above?

jaredlockhart (dismissed):
Yay Cirrus! 🎉

jeddai (approved):
looking great 🎉

Issue #8396, opened 2023-03-06
Live rollout without any changes shows 'Unpublished Changes'


jaredlockhart:
I just noticed a live rollout here:

https://experimenter.services.mozilla.com/nimbus/mozilla-rally-convert-core-add-on-users-to-attention-stream-2nd-rollout/summary

is showing 'Unpublished Changes' even though there can't be any unpublished changes since we haven't landed that functionality yet:

<img width="1263" alt="image" src="https://user-images.githubusercontent.com/119884/223172917-17061c78-5c38-4ad3-91c9-4e91cb11a00b.png">

Currently in status LIVE/LIVE/WAITING

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3225)

data-sync-user:
➤ Elise Richards commented:

This will be invalid once https://mozilla-hub.atlassian.net/browse/EXP-3123 ( https://mozilla-hub.atlassian.net/browse/EXP-3123|smart-link ) is merged

data-sync-user:
➤ Elise Richards commented:

Closing since this is no longer valid.

PR #8398, opened 2023-03-06
For #8396 feat(nimbus): Add feature flag for live rollouts

eliserichards:
Because

- We only want the bucketing warning and "Unpublished changes" status when we enable live rollouts

This commit

- Adds a lil feature flag to prevent these from showing up until we fully enable live rollouts


jaredlockhart:
Oh no yeah I was thinking of this, we actually have a mechanism for this here:

https://github.com/mozilla/experimenter/blob/main/app/experimenter/base/models.py#L4-L29

That way we can turn it on locally and on stage but not on prod, you wanna try that?

Issue #8421, opened 2023-03-07
Remove end enrollment for live rollouts


data-sync-user:
Since we have a conflict between rollouts being perpetually updateable and the uni-directionality of ending enrollment, and the lack of a clear use case of ending unenrollments for rollouts, we want to remove “End Enrollment” for live rollouts.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3236)

PR #8422, opened 2023-03-07
For #8421 feat(nimbus): Don't show end enrollment button for live rollouts

eliserichards:
Because

- Ending enrollment is at odds with updating live rollouts

This commit

- Removes the "End Enrollment" button for live rollouts

### DIFF
--- a/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
+++ b/app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
@@ -379,6 +379,29 @@ describe("Summary", () => {
     await waitFor(() => {
       expect(requestUpdateButton).toBeInTheDocument();
       expect(endExperimentButton).toBeInTheDocument();
+    });
+  });
+
+  it("do not show end enrollment button for live dirty rollout", async () => {
+    const { mockRollout, rollout } = mockLiveRolloutQuery("demo-slug", {
+      status: NimbusExperimentStatusEnum.LIVE,
+      publishStatus: NimbusExperimentPublishStatusEnum.DIRTY,
+      statusNext: null,
+      isEnrollmentPaused: false,
+    });
+
+    const mutationMock = createMutationMock(
+      rollout.id!,
+      NimbusExperimentPublishStatusEnum.DIRTY,
+      {
+        changelogMessage: CHANGELOG_MESSAGES.REQUESTED_REVIEW_UPDATE,
+        statusNext: null,
+        publishStatus: NimbusExperimentPublishStatusEnum.DIRTY,
+        status: NimbusExperimentStatusEnum.LIVE,
+      },
+    );
+    render(<Subject props={rollout} mocks={[mockRollout, mutationMock]} />);
+    await waitFor(() => {
       expect(
         screen.queryByTestId("end-enrollment-start"),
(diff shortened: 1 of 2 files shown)


app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -351,7 +351,7 @@ describe("Summary", () => {
     await act(async () => void fireEvent.click(requestUpdateButton));
   });
 
-  it("shows update and end buttons for live dirty rollout", async () => {
+  it("shows update and end experiment button for live dirty rollout", async () => {
```
yashikakhurana, line 354: don't you mean don't show end experiment button?

yashikakhurana (changes requested):

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
eliserichards: nope! it **should show** end experiment and it **should not** show end enrollment

eliserichards: I could separate that out into another test 🤔

eliserichards: this test could be "show these two buttons" and a new one could be "don't show this button" 🤔

yashikakhurana: right, yeah that's what I thought, test case means show end enrollment but it wasn't doing that :D

yashikakhurana (approved):
nice test cases 🥇

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
```diff
@@ -376,13 +376,35 @@ describe("Summary", () => {
...
-      "end-enrollment-start",
-    );
     await waitFor(() => {
       expect(requestUpdateButton).toBeInTheDocument();
-      expect(endEnrollmentButton).toBeInTheDocument();
       expect(endExperimentButton).toBeInTheDocument();
     });
   });
+
+  it("do not show end enrollment button for live dirty rollout", async () => {
```
jaredlockhart, line 385: Should be for a rollout in any state here

app/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -100,7 +100,8 @@ const Summary = ({ experiment, refetch }: SummaryProps) => {
               !status.review &&
               (status.idle || status.dirty) &&
               !status.pauseRequested &&
-              !experiment.isEnrollmentPaused && (
+              !experiment.isEnrollmentPaused &&
+              !experiment.isRollout && (
```
jaredlockhart, line 104: Ya the logic is correct here, but the test makes it seem like it's only hidden for rollouts in specific states

app/experimenter/nimbus-ui/src/components/Summary/index.test.tsx
eliserichards: The enrollment end button is only shown when live, so it is only hidden for live rollouts

Issue #8416, opened 2023-03-06
Unpublished Changes doesn't show in mid review flow


jaredlockhart:
When a rollout goes from REVIEW > APPROVED the 'Unpublished Changes' pill disappears:

<img width="1268" alt="image" src="https://user-images.githubusercontent.com/119884/223259684-f82dc1ef-1066-425a-aa19-d2a3f2d6e5d7.png">

and then reappears when it goes from APPROVED > WAITING

<img width="1285" alt="image" src="https://user-images.githubusercontent.com/119884/223259847-e4e42657-23c8-4e7a-93f8-4e7cb835451f.png">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3234)

jrbenny35:
Would be nice to have this checked within the integration tests in #8378

PR #8433, opened 2023-03-08
For #8416 feat(nimbus): Show the unpublished changes status pill when a live rollout is being approved

eliserichards:
Because

- We want to show the "Unpublished changes" status pill all through the review until it is live again (for updates to live rollouts)

This commit

- Adds a check for approval--in addition to the existing checks for dirty and waiting--when we are deciding to show the "Unpublished changes" status pill on the summary page
- Breaks up the tests into three parameterized ones (for brevity):
   - Tests for Dirty and for Dirty -> Review
   - Tests for Approve and Waiting -- I broke these out since they needed different starting parameters and it would have been too confusing/verbose to add to the original test 
   - Tests for non-live updates (which shouldn't show the pill)

image

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
@@ -96,6 +96,7 @@ describe("PageSummary", () => {
     render(<Subject mocks={[mock]} />);
     await screen.findByText("Cancel Review");
   });
+
   it("verify cancel review doesn't change enrollment days", async () => {
     const { mock, experiment } = mockExperimentQuery("demo-slug", {
       status: NimbusExperimentStatusEnum.LIVE,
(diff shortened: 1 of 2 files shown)


yashikakhurana (commented):
tested locally, works perfectly, my only concern is do we allow them to decrease the population percent? I mean it does support now from a UI perspective, I am wondering from a technical perspective can it create issues for bucketing and enrollments?

eliserichards:
We do want to allow them to decrease the size, since they want the functionality to be able to turn it to 0% (like a kill switch if something goes wrong). What happens when they decrease the population % is that --
* The bucket range exists already
* We delete it (unenrolling everyone, essentially)
* We re-allocate the bucket with the new size -- since the namespaces for rollouts are now unique (since we [added advanced targeting](https://github.com/mozilla/experimenter/pull/8350) to the namespace) it means that the same group of users will be enrolled again up to the new pop%. 

So, if we are at 100% and we decrease it to 50%:
[XXXXXXXXXXXXXXXXXXX] <- let's say that's the namespace with X's being enrolled users

the namespace is removed:
[________________________]

then 50% is re-allocated:
[XXXXXXXXX_____________]

And it is the same bucket of users, since the namespace is the same. Does that make sense?

yashikakhurana (approved):
thank you for explaining it, all set now

PR #8434, opened 2023-03-08
fix #8402: chore(nimbus) Add dev env test to CI runs

### LINKED ISSUE #8402: Add a dev env test to CI
We recently saw a case where a dependabot bump passed CI and ran fine in prod but broke local dev.  All of the CI tests currently run against the `make up_prod` command which replicates the production environment but none run against `make up` which is what is used in local dev.  Let's do something like run the nimbus ui tests against both local dev and prod environments to make sure a change doesn't break one or the other.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3229)

jrbenny35:
Because
* We are testing using ```make up_prod_detached``` and this gets us a production environment

This commit
* Adds a UI test run using ```make up_detached```, which makes sure we don't ruin our dev environment

This is mainly done for bot PRs


jrbenny35:
Not sure why this is failing on CI

jaredlockhart:
Let's close this until @jrbenny35 has time to to look at it.

PR #8461, opened 2023-03-09
 fix #8367 feat(nimbus): add flag to experiment model for results url availability

### LINKED ISSUE #8367: Update Experiment model to include results ready logic
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3203)

mikewilli:
Because

- a change to the experiment `results_data` schema caused the utility function that determines whether the UI shows the Results URL to break

This commit

- fixes the broken logic to determine whether to show the Results URL
- moves this logic into the Experiment model


experimenter/experimenter/experiments/models.py
```diff
@@ -676,6 +676,34 @@ def results_ready(self):
...
+                    enrollments = weekly_results["enrollments"]
+                    if "all" in enrollments and enrollments["all"] is not None:
+                        return True
+
+        return False
+
+    @property
+    def show_results_url(self):
+        # TODO: more logic needed? if not then return IF directly
+        if self.has_displayable_results and self.results_ready and not self.is_rollout:
```
eliserichards, line 703: 💃

experimenter/experimenter/experiments/models.py
```diff
@@ -676,6 +676,34 @@ def results_ready(self):
...
+                        return True
+
+            if "weekly" in results_data:
+                weekly_results = results_data["weekly"]
+                if "enrollments" in weekly_results:
+                    enrollments = weekly_results["enrollments"]
+                    if "all" in enrollments and enrollments["all"] is not None:
+                        return True
+
+        return False
```
eliserichards, line 698: Oooof so many nested if's. But I can't think of a better way to do this, and you've made it super readable

eliserichards (approved):
Lookin' good! 🚀

experimenter/experimenter/experiments/models.py
yashikakhurana: oh you can do `return self.has_displayable_results and self.results_ready and not self.is_rollout`

experimenter/experimenter/experiments/models.py
yashikakhurana: we can do this 
```python
    if self.results_data and "v2" in self.results_data:
        results_data = self.results_data["v2"]
        for result_type in ["overall", "weekly"]:
            if result_type in results_data:
                enrollments = results_data[result_type].get("enrollments", {}).get("all")
                if enrollments is not None:
                    return True
    return False

**Review by yashikakhurana — CHANGES_REQUESTED** [review]
works perfectly 👏 , just some suggestion for code optimization

File: experimenter/experimenter/experiments/models.py (hunk 2 above)
Comment (jrbenny35) [reply experimenter/experimenter/experiments/models.py:698]: We know what the data structure is right? We could do something like this, where we attempt to check for  the fields and if they return None, we assume there is not displayable results. 

```
while True:
            # Check overall results
            try:
                [item for item in self.results_data["v2"]["overall"]["enrollments"]["all"]]
            except (TypeError, KeyError) as e:
                pass
            else:
                return True
            # Check weekly results
            try:
                [item for item in self.results_data["v2"]["weekly"]["enrollments"]["all"]]
            except (TypeError, KeyError) as e:
                pass
            else:
                return True

            return False
```

File: experimenter/experimenter/experiments/models.py (hunk 1 above)
Comment (mikewilli) [reply experimenter/experimenter/experiments/models.py:703]: Not sure how I missed it on commit.. whoops sorry! Glad I put this in though.

**Review by yashikakhurana — APPROVED** [review]
I like parameterised test cases easy to understand, great job 🎉

File: experimenter/experimenter/experiments/models.py [hunk 3]
```diff
@@ -676,6 +676,23 @@ def results_ready(self):
...
+    @property
+    def has_displayable_results(self):
+        # True if self.results_data has weekly or overall results
+        if self.results_data and "v2" in self.results_data:
+            results_data = self.results_data["v2"]
+            for window in ["overall", "weekly"]:
+                if window in results_data:
+                    enrollments = results_data[window].get("enrollments", {}).get("all")
+                    if enrollments is not None:
+                        return True
```
jaredlockhart, line 688: Oh awesome.  One thing I would've said here is having this logic live somewhere in the jetstream module and be callable from here, since it has to do with the details of the actual structure and content of the jetstream data, which in theory could be opaque to the experiment module, but this is a good step thank you 🙏

mikewilli: Yea totally agree. I was planning to have a setter in the model that the Jetstream client code would call, but it wasn't working with the tests. I can revisit it separately though, and then this function can just be a regular getter.

experimenter/experimenter/experiments/models.py
```diff
@@ -676,6 +676,23 @@ def results_ready(self):
             )
             return datetime.date.today() >= resultsReadyDate
 
+    @property
+    def has_displayable_results(self):
```
jaredlockhart, line 680:
```py
return self.results_data.get('has_results')
```

PR #8470, opened 2023-03-10
feat(nimbus): add targeting for new profiles

Mardak:
r?@jaredlockhart /@AllegroFox /@punamdahiya  Because https://experimenter.services.mozilla.com/nimbus/firefox-accounts-toolbar-button-default-visibility needs a version only for new profiles but broader than `isFirstStartup`

This reuses the existing `NEW_PROFILE` targeting as the only targeting.

new profile created

cc @mikeconley


jaredlockhart (approved):
Looks good 👍

PR #8478, opened 2023-03-13
chore(nimbus): Update to nimbus-shared 2.0 and jsonschema 4.17

brennie:
Because

- a schema in mozilla-central used a modern JSONSchema feature (dependentRequired);
- a recipe was deployed that failed validation in Nightly but passed in Experimenter console;
- jsonschema 3.x is several years old and doesn't support modern JSONSchema features, like dependentRequired;
- nimbus-shared has a transitive dependency on jsonschema; and
- the new nimbu-shared has a schema update for experiment localization.

This commit

- updates jsonschema and nimbus-shared to modern versions so that we can use those features in schemas and validate experiment localization fields.


yashikakhurana (approved):
versions upgrade, thank you @brennie for well description of the PR 🎉

jaredlockhart (approved):
Awesome ty 🙏

PR #8488, opened 2023-03-14
fixes #8451 feat(nimbus): change audience page so that numbers do not…

### LINKED ISSUE #8451: Don't let numbers start with zero on Audience page
Steps to reproduce:
1. Create a new experiment Draft
2. On the audience page, fill in the "Expected number of clients" field with a number starting with zero (e.g. "050")

Expected:
The UI tells you that it isn't a valid number

Actual: 
The number can be saved and is valid. When you navigate to another page and come back, the number has been correctly formatted (if you entered "050" it now shows "50"). 

image

We should either:

1. Show a warning when a number begins with 0
or 
2. When the user saves, it should format the number

kendimuthomi:
… start with zero

Because

- The user input number for the Audience page should not start with a 0

This commit

- Change the Audience page and added a condition that the entry for 'Expected number of clients' cannot begin with a zero


yashikakhurana:
@kendimuthomi integration test are failing because of the changes, try to refer to CI, you will get more insights :)

kendimuthomi:
@yashikakhurana I will look into that. Thank you.

yashikakhurana:
@kendimuthomi refer to the readme doc, there you will find some instructions for how to run the integration test locally
also Ran your PR locally, 
image
I don't think this is the right message, this message is valid if we are providing string like `asadad`, but not for `070` would suggest adding other constant for this and show appropriate message

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -1173,9 +1179,15 @@ describe("FormAudience", () => {
...
-      ).toHaveTextContent(FIELD_MESSAGES.POSITIVE_NUMBER);
+      if (fieldName === "totalEnrolledClients") {
+        expect(
+          container.querySelector(`.invalid-feedback[data-for=${fieldName}]`),
+        ).toHaveTextContent(FIELD_MESSAGES.NON_ZERO_POSITIVE_NUMBER);
+      } else {
+        expect(
+          container.querySelector(`.invalid-feedback[data-for=${fieldName}]`),
+        ).toHaveTextContent(FIELD_MESSAGES.POSITIVE_NUMBER);
+      }
```
yashikakhurana, line 1190: that's not a right approach, in the test cases we should be sure about the outcome since we are setting up test case, it should not have if/else in it, Kindly adjust the test cases based on the input given for totalEnrolledClients

yashikakhurana (changes requested):

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
kendimuthomi: Okay

yashikakhurana:
@kendimuthomi if you are stuck and not working on this, let me know I can close this PR, feel free to submit again once you are able to resolve the issue, thanks

kendimuthomi:
I am running into issues because my PC will not run any commands locally anymore. I can not tell if my code is passing and I cannot run the app locally either. I think it would be best if you reassign the issue to someone else. Thank you for the opportunity.

Issue #8357, opened 2023-02-28
Indicate unpublished changes on Home


data-sync-user:
Should show something on list page (home) for the rollout to indicate that it is Dirty

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3196)

PR #8494, opened 2023-03-14
For #8357 feat(nimbus): Add column to homepage for Unpublished Updates

eliserichards:
Because

- We want users to be able to tell if their rollouts have been edited and have unpublished changes
- And we already can tell from the detail summary of the rollout (#8321)

This commit

- Adds a new column to the end of the home page for Unpublished Updates
- Adds a sort selector for it
- Shows a red "YES" badge if there are unpublished changes (publish_status=DIRTY) 
- Shows empty if there are no unpublished changes

Screen Shot 2023-03-14 at 3 29 16 PM

Note: I confirmed that this is the correct place for the column with Shell.

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
@@ -5,7 +5,7 @@
 import { Link } from "@reach/router";
 import classNames from "classnames";
 import React, { useCallback, useMemo } from "react";
-import { Button, Table } from "react-bootstrap";
+import { Badge, Button, Table } from "react-bootstrap";
 import LinkExternal from "src/components/LinkExternal";
 import NotSet from "src/components/NotSet";
 import { displayConfigLabelOrNotSet } from "src/components/Summary";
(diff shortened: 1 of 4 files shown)


yashikakhurana:
@eliserichards I am just wondering to think better name `unpublished updates` as it takes a lot of space in UI and only has value as `Yes`, the other thing is to consider that this unpublished update is only valid for rollouts, not for experiments, I am just thinking a better way to represent this, maybe adding this tile with the name of the rollout, and changed 'yes' to 'updated' and drop the column for unpublished changes 🤔

eliserichards:
I've OK'd this implementation with Shell, so I think we can go ahead and use this for now. I completely agree that we need a better, less space-intensive way to indicate things like this. Let's brainstorm this a bit more and make some mockups? @yashikakhurana

eliserichards:
I've filed an issue for this: https://github.com/mozilla/experimenter/issues/8506

experimenter/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -140,6 +144,27 @@ export const DirectoryColumnEndDate: ColumnComponent = ({
...
+  <td data-testid="directory-table-cell">
+    {p ? (
+      <>
+        <Badge
+          className={
+            "ml-2 border rounded-pill px-2 bg-white font-weight-normal border-danger text-danger"
+          }
+          data-testid="directory-unpublished-updates"
+        >
+          {p === NimbusExperimentPublishStatusEnum.DIRTY ? "YES" : ""}
```
jaredlockhart, line 159: Actually what we should do is add one of those aggregate states to that status object that computes the aggregate set of states that have 'unpublished changes' and then we can use that on the summary page pill and here.  But we can do that as a followup.

jaredlockhart (approved):
Awesome looks good, tested locally, let's file a followup for moving the aggregate status, thnx @eliserichards ! 🎉

PR #8496, opened 2023-03-14
Fix #8495 chore(nimbus): move docs folder to repository root

### LINKED ISSUE #8495: Move docs to repository root
The docs in `experimenter/experimenter/docs` should be moved to the repository root, as they apply to Nimbus as a whole.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3275)

jeddai:
Because
- The documentation in the `experimenter/experimenter/docs` folder applies to Nimbus as a whole

This commit
- Moves that docs folder to the repository root
- Updates url paths in relevant docs

~Just checked, there will need to be an associated PR in the experimenter-docs repo to update paths as well (to be fair the paths are currently broken).~ Additional PR for related docs: mozilla/experimenter-docs#400


eliserichards:
Good catch on the experimenter-docs stuff being broken 😓 

If we're going to add cirrus stuff to this `docs/` folder, would it make sense to have this stuff as `docs/experimenter/`?

jeddai:
Hm that's a good point – maybe we move the diagrams/swagger/open-api things into `docs/experimenter` and leave the adrs in the root `docs`?

yashikakhurana:
> Hm that's a good point – maybe we move the diagrams/swagger/open-api things into `docs/experimenter` and leave the adrs in the root `docs`?

I second that, in that way we could also add docs/cirrus and keep cirrus documentation there, and thank you for noticing that.

eliserichards:
Yeah yeah 👍 We could add another folder docs/adrs?

jaredlockhart (approved):
Awesome thank you @jeddai 🎉 Just occurred to me we should do the same with the integration tests....

jeddai:
Ah I didn't know about the docker piece, thanks @jaredlockhart!

jeddai:
Opened #8507 – I'm not planning to impl that right now, could it be a reasonable issue to add the outreachy label to?

jaredlockhart:
> Opened #8507 – I'm not planning to impl that right now, could it be a reasonable issue to add the outreachy label to?

Oh no, anything involving the integration tests is too much for outreachy.  Good for @jrbenny35 when he's back from the race track 🚗 🏁

PR #8516, opened 2023-03-15
Add "not mac" as targeting option

danielkberry:
Because

- The new accessibility cache experiment needs to target "non-mac" users

This commit

- Adds an advanced targeting to exclude non-mac users


yashikakhurana:
@danielkberry can you set `desktop_telemetry=""`

yashikakhurana (approved):
🚢 , thanks @danielkberry

PR #8520, opened 2023-03-16
fixes #7845 fix(nimbus):change summary page to show total enrolled cl…

### LINKED ISSUE #7845: Add total enrolled clients to detail page
We're not showing the total enrolled clients on the detail page, we should add that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2849)

Avi-88:
…ients even if zero

Because

- the field doesn't render if the value is zero, which causes irregularity in the details layout

Before:
image

This commit

- change audience section in summary page to allow the total enrolled clients field even if the value is zero so as to not break the layout of information

After:
image


Avi-88:
@yashika-khurana , @eliserichards , @jaredlockhart  please review the code and let me know if any further changes are required , thank you

Issue #8514, opened 2023-03-15
Rollout conflict check returns false positive


danielkberry:
[This rollout](https://experimenter.services.mozilla.com/nimbus/enrollment-volume-debugging/summary) is currently throwing a "conflicting rollout" warning, when it shouldn't. Through conversations with @jaredlockhart and @eliserichards , the issue is [here](https://github.com/mozilla/experimenter/blob/37d08c9348f69033caa263d3327f6c72c0790d90/experimenter/experimenter/experiments/api/v5/serializers.py#L1513) which should also be filtering by feature. 

![Screenshot 2023-03-15 at 1 07 42 PM](https://user-images.githubusercontent.com/2738595/225430344-1713cd41-b21c-426b-9893-dff955fdf9a7.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3288)

PR #8524, opened 2023-03-16
For #8514 fix(nimbus): Check feature configs when looking for duplicate rollouts

eliserichards:
Because

- I forgot to add feature_configs to the check for the duplicate rollout warning in #8359 

This commit

- Adds the check for feature_configs in the serializer
- Fixes the tests

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1507,18 +1507,20 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         return data
 
     def _validate_bucket_duplicates(self, data):
-        if not self.instance or not self.instance.is_rollout:
-            return data
         is_rollout = self.instance.is_rollout
+        if not self.instance or not is_rollout:
+            return data
+
         count = NimbusExperiment.objects.filter(
             status=NimbusExperiment.Status.LIVE,
             channel=self.instance.channel,
             application=self.instance.application,
             targeting_config_slug=self.instance.targeting_config_slug,
+            feature_configs__in=self.instance.feature_configs.all(),
             is_rollout=is_rollout,
         ).count()
 
-        if is_rollout and count > 0:
+        if count > 0:
             self.warnings["bucketing"] = [NimbusConstants.ERROR_BUCKET_EXISTS]
 
         return data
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1507,18 +1507,25 @@ def _validate_rollout_version_support(self, data):
...
             is_rollout=is_rollout,
-        ).count()
+        ):
+            if experiment.feature_configs.exists():
+                for feature_config in experiment.feature_configs.all():
+                    if feature_config in configs:
+                        matches.append(experiment)
 
-        if is_rollout and count > 0:
+        if matches.__len__() > 0:
```
jaredlockhart, line 1528:
```py
NimbusExperiment.objects.filter(
    ...
    feature_configs__in=self.instance.feature_configs.all()
    ...
)
```

;)

jaredlockhart: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#in

eliserichards: `__in`!!! brilliant

jaredlockhart (approved):
Perf looks good 👍

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1507,18 +1507,20 @@ def _validate_rollout_version_support(self, data):
...
             status=NimbusExperiment.Status.LIVE,
             channel=self.instance.channel,
             application=self.instance.application,
             targeting_config_slug=self.instance.targeting_config_slug,
+            feature_configs__in=self.instance.feature_configs.all(),
             is_rollout=is_rollout,
         ).count()
 
-        if is_rollout and count > 0:
+        if count > 0:
```
jaredlockhart, line 1523: Actually another neat trick you could do, but this is super minor, is

```py
# create a queryset object, but don't evaluate it (ie it's not actually executed in the db)
matching_rollouts = NimbusExperiment.objects.filter(....)

# now the query is executed, and since we only care if any exist we don't even need to count, just check existence which is an even lighter db call
if matching_rollouts.exists():
    return data
```

eliserichards: Oh shoot I didn't see this before I merged. Thanks for the tips!! I might open a new PR for that once I'm done this other work I have in progress 🙌

PR #8527, opened 2023-03-16
Fixes #8526: Add adr 0009 for nimbus web sdk architecture

### LINKED ISSUE #8526: Write ADR for Nimbus web SDK architecture changes
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3297)

jeddai:
Because

- We require some large changes to the core structure of the Nimbus SDK to support Nimbus for the web

This commit

- Adds ADR 0009


jeddai:
In its current state, this PR is open for feedback regarding the various options. James and I are thinking one of the two Cargo features options is the best, but which one/to what degree is up for discussion. 

@tarikeshaq @bendk your thoughts/feedback would be greatly appreciated!

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+* Good, because dependencies can be excluded based on enabled features.
+  * As an example, RKV will only be needed for Nimbus, so the RKV dependency will only be included in compilation when the `nimbus` feature is enabled.
+* Good, because it will not increase the size of the existing Nimbus SDK binary.
+* Bad, because onboarding developers learning Rust would likely have a more difficult time understanding Cargo features.
+
+### Cargo features, separate library
+
+This is roughly the same concept as above, but the features will be used in a different way.
+Instead of features for both Nimbus and Cirrus, we would include a `stateless` feature in the Nimbus library to disable state-oriented features and dependencies.
+The Cirrus library will then exist as a separate library from the Nimbus library.
```
jaredlockhart, line 76: It's not clear to me what the role of a separate Cirrus library would be in this scenario?

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+  * See a similar point in the "Separate `nimbus-core`, `nimbus-sdk`, and `cirrus-sdk` libraries" option above.
+* Bad, because methods used across both libraries could result in collisions during development.
+
+### Expose a stateless API alongside the stateful API
+
+In this solution, we would expose static/stateless methods for interacting with experiments.
+The UniFFI UDL would be updated to include the methods, resulting in the methods being visible in Android and iOS where they will not be used. 
+
+* Good, because it's simple and entirely additive.
+* Bad, because it makes Nimbus more prone to developer error.
```
jaredlockhart, line 91: Can you provide some more context or examples about how this would occur?

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+
+Technical Story: https://mozilla-hub.atlassian.net/browse/EXP-3158
+
+## Context and Problem Statement
+
+This is part of a body of work necessary to support the use of Nimbus within web applications. 
+The current Nimbus SDK is written in such a way that it supports client-oriented experimentation — experiments are downloaded, evaluated, and stored on the client, and a feature store is exposed with the experiment branches applied. 
+In order to support web experimentation, the Nimbus SDK will need to be updated to be stateless, to support a more statically defined set of helper methods, and to have additional support for Python.
+
+Ultimately, the problem we're trying to solve can be boiled down to one question — how can we update the Nimbus SDK to support web applications while continuing to support the existing clients?
```
jaredlockhart, line 15: This is a wonderfully succinct summary 👏 👏 👏

jaredlockhart (commented):
This is fantastic @jeddai thank you 🙏 So are we still undecided on the approach here?

docs/adrs/0009-nimbus-web-sdk-architecture.md
jeddai: In this scenario the purpose of the Cirrus library would be a separation of concerns re: the Cirrus UniFFI UDL and exported Rust/Python Cirrus-specific functions. These would likely primarily be helper methods, e.g: parsing JSON from Python before getting to the core function or other things along those lines. 

The Cirrus SDK would import the Nimbus library as a dependency, enabling the `stateless` feature to cut out dependencies like RKV/Nimbus-specific UniFFI and state-oriented features such as the event store and internal database.

jaredlockhart: Makes sense, can you add some of that explanation back to the ADR? 🙏

docs/adrs/0009-nimbus-web-sdk-architecture.md
jeddai: Sure! So the existing UDL includes a NimbusClient class – this class shouldn't ever actually be used by iOS/Android directly, but it is there and _can_ be used. It doesn't include any tooling for handling threading or object builders or anything like that.

By adding the stateless methods for experiment evaluation (and others), they would be available in Python, Kotlin, and Swift, even though they should only be used within the context of Cirrus. 

More than anything else I see this as a "someone finds the function in the Nimbus internals and starts using it incorrectly" concern.

jaredlockhart: Ah interesting.  That makes sense, we should also add some of that context back to the ADR.

docs/adrs/0009-nimbus-web-sdk-architecture.md
jeddai: Sure!

tarikeshaq (commented):
Thank you for writing this up @jeddai 🚀 I wrote down my thoughts as I went through, mostly questions bu this is great!

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,101 @@
...
+The Cirrus library will then exist as a separate library from the Nimbus library, primarily to function as a more visible separation of concerns between the two libraries.
+The new Cirrus library would mostly consist of the Cirrus UDL, as well as helper methods — e.g: parsing JSON received from Python before getting to the core function or other things along those lines.
+
+* Good, because all the good things in the "Cargo features, one library" option above.
+* Good, because the changes required will require minimal churn.
+* Good, because better separation of concerns between Nimbus and Cirrus.
+* Bad, because defining and using types and traits across the libraries could prove difficult.
+  * See a similar point in the "Separate `nimbus-core`, `nimbus-sdk`, and `cirrus-sdk` libraries" option above.
+* Bad, because methods used across both libraries could result in collisions during development.
+
```
yashikakhurana, line 85: its worth mentioning about size impact, whether it will be more or less

yashikakhurana (commented):
Thank you @jeddai for putting this together, overall its looks good 🎉

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,101 @@
...
+  * The core logic of the existing SDK will be re-used for the Nimbus web service (Cirrus)
+  * The SDK needs to support Python
+  * The SDK needs to be stateless
+
+## Considered Options
+
+* Separate `nimbus-core`, `nimbus-sdk`, and `cirrus-sdk` libraries
+* Cargo features, one library
+* Cargo features, separate library
+* Expose a stateless API alongside the stateful API
```
jhugman, line 29: ISTR this is the `Do Nothing` option: compile both Cirrus and Nimbus SDKs for both web apps _and_ mobile apps.

jhugman (commented):
I've added review comments further to our discussion. Good job Charlie!

docs/adrs/0009-nimbus-web-sdk-architecture.md
jeddai: Yep – it might be worth me adding that in parens here

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+## Pros and Cons of the Options
+
+### Separate `nimbus-core`, `nimbus-sdk`, and `cirrus-sdk` libraries
+
+For this option, we would take the core Nimbus business logic currently in the `nimbus-sdk` library and move it to a new `nimbus-core` library. 
+The `nimbus-sdk` and `cirrus-sdk` libraries would depend on the `nimbus-core` library, extending the functionality for their various use cases. 
+
+* Good, because separate libraries allow for strong separation of concerns.
+* Bad, because defining and using types becomes more difficult.
+  * Traits will be needed to extend functionality in the Nimbus Core library. While dynamic typing with traits is supported, it is one of the more complex features of Rust.
```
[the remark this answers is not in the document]
jeddai: I do agree, but I'm also not sure how to go about explaining it succinctly. I've got a code example here: https://github.com/mozilla/application-services/compare/main...jeddai:application-services:create-nimbus-core#diff-02305e8e02a7900352e67be1fd2eef0b5a7c7cf91f4cc3e4559668c124d34e88R11-R19

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,101 @@
...
+Instead of features for both Nimbus and Cirrus, we would include a `stateless` feature in the Nimbus library to disable state-oriented features and dependencies.
+The Cirrus library will then exist as a separate library from the Nimbus library, primarily to function as a more visible separation of concerns between the two libraries.
+The new Cirrus library would mostly consist of the Cirrus UDL, as well as helper methods — e.g: parsing JSON received from Python before getting to the core function or other things along those lines.
+
+* Good, because all the good things in the "Cargo features, one library" option above.
+* Good, because the changes required will require minimal churn.
+* Good, because better separation of concerns between Nimbus and Cirrus.
+* Bad, because defining and using types and traits across the libraries could prove difficult.
+  * See a similar point in the "Separate `nimbus-core`, `nimbus-sdk`, and `cirrus-sdk` libraries" option above.
+* Bad, because methods used across both libraries could result in collisions during development.
```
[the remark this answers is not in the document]
eliserichards: Circular dependencies, maybe?

docs/adrs/0009-nimbus-web-sdk-architecture.md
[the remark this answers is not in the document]
eliserichards: I think linking to an existing example would be good

eliserichards (commented):
Overall, this is looking awesome 👏 

This is totally a nit and feel free to totally ignore, but I would prefer the "Good, because.." and "Bad, because..." to be under "Pros"/"Cons" sub-headers 😬

docs/adrs/0009-nimbus-web-sdk-architecture.md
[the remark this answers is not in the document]
tarikeshaq: Uh I see, yeah adding the snippet you described here makes a ton of sense!

FWIW, if traits extending other traits is a `Bad` we _might_ be able to work around it. For example, we could:
- Only having one `EventStore` trait and have the consumers (i.e NimbusSDK) deal with whether they are DB backed or not. In practice, this means the NimbusSDK can define a type that implements `EventStore`, and holds on to a DB connection in `self`

- Remove the `Box<dyn..>` and have types be generic over `T: EventStore`. This would mean `NimbusClient` becomes `NimbusClient<T>` and `NimbusClient::event_store`'s type becomes `T`


Regardless - The example clarifies your point a lot and your point still stands, playing around with generics/traits like that is complex.

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Cargo features, one library
+
+Cargo features allow for conditional compilation of libraries and library dependencies. 
+Native Rust attributes are applied to sections of Rust code to conditionally compile structs, methods, variables, and more.
+
+For more information, check out the Cargo Features reference[<sup>[1.ii]</sup>](#links) available in the Rust Cargo book.
+
+* Good, because the use of feature configurations is familiar to Rust developers.
+* Good, because all code relating to Nimbus business logic remains in one place.
```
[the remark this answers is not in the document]
jeddai: Sort of? In this case that's going to specifically be just the core Nimbus business logic, because all logic for statefulness and Kotlin/Swift will end up being in the `nimbus-sdk` crate.

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,97 @@
+# Nimbus Web SDK Architecture
+
+* Status: proposed
+* Deciders: James Hugman, Charlie Humphreys, Yashika Khurana, Jared Lockhart
+* Date: 2023-15-03
+
+Technical Story: https://mozilla-hub.atlassian.net/browse/EXP-3158
+
+## Context and Problem Statement
```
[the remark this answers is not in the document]
jeddai: I think I'm going to refrain from explicit code examples within this ADR, but I can definitely add an example of what the API might look like

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,101 @@
...
+  * As an example, RKV will only be needed for Nimbus, so the RKV dependency will only be included in compilation when the `nimbus` feature is enabled.
+* Good, because it will not increase the size of the existing Nimbus SDK binary.
+* Bad, because onboarding developers learning Rust would likely have a more difficult time understanding Cargo features.
+
+### Cargo features, separate library
+
+This is roughly the same concept as above, but the features will be used in a different way.
+Instead of features for both Nimbus and Cirrus, we would include a `stateless` feature in the Nimbus library to disable state-oriented features and dependencies.
+The Cirrus library will then exist as a separate library from the Nimbus library, primarily to function as a more visible separation of concerns between the two libraries.
+The new Cirrus library would mostly consist of the Cirrus UDL, as well as helper methods — e.g: parsing JSON received from Python before getting to the core function or other things along those lines.
```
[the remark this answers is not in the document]
jeddai: Yep!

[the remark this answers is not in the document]
jeddai: Something worth noting here is this same benefit is available in the `cargo features; one library` approach as well. Since cargo features can be used to include/exclude entire dependencies, we could introduce new dependencies for the `cirrus` feature that are not included when compiling with the `nimbus` feature.

jeddai:
Alrighty folks, a first pass on addressing feedback has been pushed, and it includes the decision outcome (chatted with James and we came to a conclusion on it).

bendk (commented):
I don't think I should be making this decision, but it makes a lot of sense to me.  From my uninformed position, feature flags seems nicer than separate crates.

I left one UniFFI-related suggestion.  It could be good to use this as an opportunity to move into the future and try out the proc-macros.

I also think using mulitple UDL files should work and don't expect any technical issues.  It just might be a bit of a pain to maintain them.

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Positive Consequences
+
+* Meets all decision drivers.
+* Small amount of work necessary with a very limited amount of code churn.
+
+### Negative Consequences
+
+* It will be difficult to draw boundaries between Cirrus and Nimbus code.
+* We could run into unexpected issues with UniFFI, as multiple UDL files in a single crate has not yet been tested.
```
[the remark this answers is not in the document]
jeddai: Hm that's a really good point. I'm certainly not opposed, it would work incredibly well in tandem with the feature attributes.

I think it's a great plan but I also don't know the time investment that would require. I'm going to start by setting up the cargo features first with the proc macro piece being an optional thing as we move forward with the project and add the cirrus features.

docs/adrs/0009-nimbus-web-sdk-architecture.md
```diff
@@ -0,0 +1,139 @@
...
+* Deciders: James Hugman, Charlie Humphreys, Yashika Khurana, Jared Lockhart
+* Date: 2023-15-03
+
+Technical Story: https://mozilla-hub.atlassian.net/browse/EXP-3158
+
+## Context and Problem Statement
+
+This is part of a body of work necessary to support the use of Nimbus within web applications. 
+The current Nimbus SDK is written in such a way that it supports client-oriented experimentation — experiments are downloaded, evaluated, and stored on the client, and a feature store is exposed with the experiment branches applied. 
+In order to support web experimentation, the Nimbus SDK will need to be updated to be stateless, to support a more statically defined set of helper methods, and to have additional support for Python.
```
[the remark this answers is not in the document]
jeddai: Yep that's a decision that was made before this ADR, I'll add that in to make it more clear, thanks!

jaredlockhart (approved):
This is fantastic thank you @jeddai 🙏

PR #8528, opened 2023-03-16
change changed_on to updated_date_time on latest_changed function

kennedy-muthaura:
Because
we are required to use _updated_date_time instead of changed_on in NimbusExpriment model
- Reason

This commit
changes the latest_changed function in NimbusExprimentManager to use  the new introduced _updated_date_time field  in NimbusExpriment instead of changed_on field in NimbusChangelog


yashikakhurana (approved):
@kennedy-muthaura thank you, great job, working as expected 👏

kennedy-muthaura:
you are welcome, let me search for another issue so that you can assign me

PR #8540, opened 2023-03-19
fixes #8499 fix(nimbus) store and revisit search result

### LINKED ISSUE #8499: Store search results
image
image
image

KramStyles:
Because

- the filtered search results are not retained after navigating away from the search page, we need to modify the behavior of the "back to experiment" button to retain the search term and filtered results.

This commit

- Ensures the clear button on the search bar also resets the address bar when clicked

@yashikakhurana @eliserichards @jaredlockhart I've implemented the changes but I have two issues. 

- While working on this task I noticed a bug with I wasn't able to fully fix as I had to find a way around it. To reproduce:
   -  On this branch, search for an experiment
   - Click on one of the filtered results to visit the page
   - Click on the `back to experiments` link
   - Clear the search bar using the `X` button
   - Click on any of the experiment currently displayed. They should all be showing since no filter was applied
   - Click on the `back to experiments` link to go back. For a brief second, the former search term is still displayed on the address bar. Even though the search term has been cleared from history and localStorage, it's still saved somewhere and I can't find it. This is the reason for the conditional statement in `line 80` of `experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx`. I would really appreciate knowing where it's saved so I can remove it properly and not make use of the conditional statement.
   

- The second issue comes up when I run `make up`. Docker consumes a lot of resources and it uses up to 12GB of my ram. It takes about 5 minutes and more to completely start the project. It wasn't like this initially as my system has become very slow when I run the project. I also noticed about 100GB of my memory missing and I've done `docker image prune` and `docker container prune` but it didn't help much. Is there anything that can be recommended for me to fix this issue? 

Thank you so much for your help. 

This is a draft PR as I'm currently writing tests to ensure the `back to experiments` button work as expected.


yashikakhurana:
@KramStyles thank you for your message, regarding 1st question- I am not sure 100% where it is getting stored, but my best guess would be maybe it's in the state of the component, would suggest checking that, 2nd would suggest clean your volumes too, and check all the stopped containers and clean them too, same for the images :)

KramStyles:
@yashikakhurana @jaredlockhart @eliserichards I apologize for disturbing. I'm actually having issues with the test. `Line 55` of my tests in [SearchBar/index.test](https://github.com/mozilla/experimenter/blob/75ea4f609b9d09dbc9b2810a3f65c72e852f3bcd/experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.test.tsx) never gets called and `Line 49` fails on [Circle CI](https://app.circleci.com/pipelines/github/mozilla/experimenter/21881/workflows/fd8a8223-d851-42fc-ad9f-5e86f29f99e4/jobs/70388). 

I've not mastered testing in jest and I'd really appreciate any help I can get to move past this and attend to other tasks. Thank you so much.

experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.test.tsx
```diff
@@ -39,6 +42,35 @@ describe("SearchBar", () => {
...
+      search: "?query=test",
+    };
+
+    // Call the function
+    resetWindowLocation();
+
+    // Expect window.location to be updated
+    expect(window.location.search).toBe("");
+  });
 });
```
yashikakhurana, line 74: try this- 
```javascript
 it("calls resetWindowLocation when nimbus-ui-search is not in localStorage", () => {
  const resetWindowLocation = jest.fn();
  jest.spyOn(Subject.prototype, "resetWindowLocation").mockImplementation(resetWindowLocation);

  jest.spyOn(window.localStorage, "getItem").mockReturnValue(null);

  render(<Subject experiments={[]} onChange={{ current: {} }} />);

  expect(resetWindowLocation).toHaveBeenCalledTimes(1);
});

describe("resetWindowLocation", () => {
  it("resets window location to default homepage", () => {
    // Mock the window.location object
    const oldLocation = window.location;
    delete window.location;
    window.location = { ...oldLocation, search: "?query=test" };

    // Call the function
    resetWindowLocation();

    // Expect window.location to be updated
    expect(window.location.search).toBe("");
  });
});

yashikakhurana (changes requested):

experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.test.tsx
KramStyles: > 

Thank you for the help and suggestion @yashikakhurana .  I tried using the code above and I tried to modify it to work but I still got the following errors

 Cannot spyOn on a primitive value; undefined given coming from `jest.spyOn(Subject.prototype, "resetWindowLocation").mockImplementation(resetWindowLocation); (src/components/PageHome/SearchBar/index.test.tsx:47:14)
` and 

 `TypeError: Illegal invocation` coming from `const url = new URL(`${window.location}`); (src/components/PageHome/SearchBar/index.tsx:41:33)`

KramStyles:
@yashikakhurana Thank you for your guidance. I followed your suggestion and I found the `state component` that updates the `window.location`. Working with that implementation save alot of code and I didn't need to use the conditional statement. This led to the removing of the initial conditional statement and tests I made for it. 
 
 I pushed my code and it initially failed because the coverage was at 99.9%. I had to update the initial test to ensure it reached the `100%` benchmark

yashikakhurana:
@KramStyles this is to let you know that we have reverted some code related to search- https://github.com/mozilla/experimenter/pull/8567 and https://github.com/mozilla/experimenter/pull/8568 so you might need to update your branch and see what you can do from there

KramStyles:
@yashikakhurana Ohh. Thanks for the information. I checked earlier and I noticed formRef and the useEffect were removed.  Do you mean I shouldn't work with these or I can add them back and ensure they work with the updates that were made?

yashikakhurana:
@KramStyles it means you have to start from scratch where you store result in URL and use it for search query, but that PR was causing intermittent test failure, so your PR should handle that as well, regarding space pr, someone is working on that so don't worry about that, let me know if you have any questions , thanks

yashikakhurana:
@KramStyles here is the PR for more info regarding the search- https://github.com/mozilla/experimenter/pull/8547, basically it was checking if the term doesn't include space after the term, it will add it, please feel free to use useEffect and refs if you think that is the valid choice

KramStyles:
> @KramStyles here is the PR for more info regarding the search- #8547, basically it was checking if the term doesn't include space after the term, it will add it, please feel free to use useEffect and refs if you think that is the valid choice

Ok. @yashikakhurana Thank you for the information. I would get started on that; I'd try to get it done as soon as possible

yashikakhurana:
@KramStyles I wouls suggest close this PR for now, and create a new PR once you are done that

KramStyles:
@yashikakhurana Thank you. I'd do that immediately

PR #8541, opened 2023-03-20
fixes #8498 fix(nimbus): change the search bar to check if input cont…

### LINKED ISSUE #8498: Search space issue
We added a search box on the landing page,
image
 but if you will write just `pdf` and `pdf ` with space, the results are different, so we want that whenever we search for something, we should add a space after that here- https://github.com/mozilla/experimenter/blob/86c4d1d7e8c9ab20d1a5d07568cb542655238266/experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx#LL94 , so before calling search to modify search term and add space after that

Prithvi-Chanchal:
…ains space at the end, if input does not have space add space at the end

Because

- In the search bar if you type same thing with space and without space at end the result we are getting is different.

This commit

- Change the search bar so that if you type same thing with space and without space at the end the result should be same

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx
@@ -78,9 +78,19 @@ const SearchBar: React.FunctionComponent<SearchBarProps> = ({
   const handleChange = (event: {
     target: { value: React.SetStateAction<string> };
   }) => {
+    let inputValue = event.target.value as string;
+    // Check if the input value already has a space at the end
+    if (
+      inputValue.length > 0 &&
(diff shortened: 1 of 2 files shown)


yashikakhurana:
@Prithvi-Chanchal you are super close, so by your changes expected result of the search query got modified, so you need to modify test cases and review(1) is found and completed (2) experiments are found, you can also check ci for more detailed error

yashikakhurana:
@Prithvi-Chanchal I am reverting this PR because your PR is build on top of this pr changes-https://github.com/mozilla/experimenter/pull/8418, so to revert that PR I have to revert yours too, feel free to submit a again new pr, the reason of reverting is because we are getting intermittent test failure-https://github.com/mozilla/experimenter/issues/8557

Issue #8438, opened 2023-03-08
Indicate dirty changes on backend


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3256)

PR #8550, opened 2023-03-21
For #8438 feat(nimbus): Add flag for isDirty on the v5 serializer

eliserichards:
Because

- We want a way to signal when a rollout has unpublished changes
- And the current way--using publish_status to signal when a rollout is dirty--does not account for publish status being used to signal the review flow for ending, launching, and publishing new changes, which causes confusion

This commit

- Adds a separate flag for `isDirty` to indicate that a rollout has unpublished changes

image


yashikakhurana:
@eliserichards what do you think of renaming this from `isDirty` to `isRolloutDirty` as this dirty state is directly linked to rollouts?

eliserichards:
that's a great idea!

jaredlockhart (approved):
Yep looks good, don't think you need the resolve method 👍

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -754,6 +754,7 @@ class NimbusExperimentSerializer(
         min_length=0, max_length=1024, required=False, allow_blank=True
     )
     is_enrollment_paused = serializers.BooleanField(source="is_paused", required=False)
+    is_rollout_dirty = serializers.BooleanField(required=False)
```
yashikakhurana, line 757: 🎉

yashikakhurana (approved):
thank you @eliserichards 🎉

experimenter/experimenter/experiments/api/v5/serializers.py
eliserichards: 💃💃💃💃💃

PR #8553, opened 2023-03-22
fixes #8549 docs(project): Add firefox version flag to integration te…

### LINKED ISSUE #8549: Add firefox version flag to integration test docs
Let's update our docs to specify that we need the following firefox version flag when running integration tests: 

```FIREFOX_VERSION=nimbus-firefox-release```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3311)

### PR DESCRIPTION
**Pull request by Prithvi-Chanchal**
…st docs

Because

- We have not specified with firefox version flag is used in intergration tests

This commit

- add the firefox version flag to intergration tests, that is being used

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,14 @@ yarn workspace @experimenter/nimbus-ui test:cov
 
 For a full reference of all the common commands that can be run inside the container, refer to [this section of the Makefile](https://github.com/mozilla/experimenter/blob/main/Makefile#L16-L38)
 
+#### Intergration Tests
+
+Note: You need the following firefox version flag when running integration tests
+
+```sh
+FIREFOX_VERSION=nimbus-firefox-release
+```
+
 #### make integration_test_legacy
 
 Run the integration test suite for experimenter inside a containerized instance of Firefox. You must also be already running a `make up` dev instance in another shell to run the integration tests.

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -394,6 +394,14 @@ yarn workspace @experimenter/nimbus-ui test:cov
 
 For a full reference of all the common commands that can be run inside the container, refer to [this section of the Makefile](https://github.com/mozilla/experimenter/blob/main/Makefile#L16-L38)
 
+#### Intergration Tests
+
+Note: You need the following firefox version flag when running integration tests
+
+```sh
+FIREFOX_VERSION=nimbus-firefox-release
+```
```
eliserichards, line 403: I think you could add this under https://github.com/mozilla/experimenter#integration-test-options. What do you think?

eliserichards (changes requested):

README.md
Prithvi-Chanchal: Ya I can add it under there, but I thought we should prompt whoever is running integration tests about this before they start the test so they won't run in any errors and then look for answers. should I change it or leave it as it is? @eliserichards @yashika-khurana

eliserichards: It is already noted in [some of the other commands](https://github.com/mozilla/experimenter#make-firefox_version-integration_test_nimbus), so I think adding it to the "Integration test options" section is the best way :) You can also add an example to make it more clear, like this: 

```
make FIREFOX_VERSION=nimbus-firefox-release integration_test_nimbus PYTEST_ARGS=-ktest_rollout_create_and_update
```

Issue #8346, opened 2023-02-23
 Highlight fields when saving on Metrics page


eliserichards:
The overview and audience pages highlight all the fields in the green border and checkmark **after saving** to show that they were validated and saved, but the metrics page does not. We should fix that so that we show the valid fields as **green** when saving, it's just a nice UI signal that your stuff was saved properly.

Broken out from https://github.com/mozilla/experimenter/issues/6584

Haywayaheadshot:
@yashikakhurana I am an outreachy applicant and I will like to work on this please

yashikakhurana:
@Haywayaheadshot sure assigning it you, let me know if you have any questions

Haywayaheadshot:
@yashikakhurana Thank you very much. Have a great day

Haywayaheadshot:
@yashikakhurana I have cloned the project but I am having a hard time running it. I have tried all I can and I need help.

emmaz12:
@yashikakhurana Attempting to solve this issue. When clicking the save button, I managed to have the form highlight in green for all valid field inputs. However, for some reason this highlighting only happens while the page is "loading" (isLoading = true). I am struggling to find the issue with this and how to fix this. I tried to cross-reference with Audience and Overview page however, the branches page seems to be built in a quite different manner. Any advice?
<img width="330" alt="Screen Shot 2023-03-21 at 7 29 31 PM" src="https://user-images.githubusercontent.com/117552772/226786684-8d08df7a-f663-40f6-92a5-df777c65db4b.png">

https://user-images.githubusercontent.com/117552772/226786770-b4b758b1-57ab-45e0-b28e-fd2b05bc2719.mov

yashikakhurana:
@Avi-88 worked on the similar task- https://github.com/mozilla/experimenter/pull/8539 you can checkout if that helps @emmaz12

emmaz12:
@Avi-88 @yashikakhurana Thanks for the suggestion! However, it seems like the implementation of validating a valid input was already done for the branches page. From what I can tell, the main difference between the audience/overview page and the branches page is that the save button will highlight all the fields in green when clicked whereas the branches page did not. I am attempting to fix this issue, however, running into some challenges. It seems like the way isSubmitted is implemented is causing this issue. Any advice?
[...]

PR #8539, opened 2023-03-18
fixes #8093 fix(nimbus): fixed the arrow icons of input type select t…

Avi-88:
…o match in audience section of summary page

Because

- 3 out of 5 inputs of type select in the audience section of summary page had different styling which caused inconsistency.

Before: 

image

This commit

- Change styling of the inputs of type select to match the desired style by overriding the default styles provided by `react-boostrap` library using the `custom` prop.

After:

image

image

PR #8555, opened 2023-03-22

Haywayaheadshot:
Because

The overview and audience pages highlight all the fields in the green border and checkmark after saving to show that they were validated and saved, but the metrics page does not. We should fix that so that we show the valid fields as green when saving, it's just a nice UI signal that your stuff was saved properly.

This commit

I refactored the Select tag to use Form.Control so I can add is-valid class.

### Video For;
- Before fix

https://user-images.githubusercontent.com/105272189/227390766-037cabdc-641a-4fa1-8861-b6427b3cb2ea.mp4

- After fix

https://user-images.githubusercontent.com/105272189/227390667-3504026c-f3e3-457a-9785-5c65f2b4064c.mp4

- Updated After-fix

https://user-images.githubusercontent.com/105272189/227726999-edd0d64d-fb5d-4c9e-8fc3-a05485b4c634.mp4

### This pull request was generated from [this](https://github.com/mozilla/experimenter/issues/8346) issue.


yashikakhurana (changes requested):
kindly refer to my comment

Avi-88:
> @Haywayaheadshot thank you for working on this, unfortunately, you are trying to change the complete UI, we don't want to change the look of the drop-down, you can refer to this PR- #8539 @Avi-88 did something similar, thanks

Thank you @yashikakhurana to bring this up, I think I missed a previous mention made by @Haywayaheadshot for the same. Sorry about that was just a bit busy with exams. @Haywayaheadshot you can refer to my PR and see if you can implement the same for the currently used `<Select/>` component instead of replacing it with `form-control`. Feel free to tag me again and i'll do my best to help you , good luck :)

Haywayaheadshot:
@yashikakhurana Thank you so much for your feedback.

@Avi-88 Thank you for giving me permission to check out your project.

Haywayaheadshot:
@yashikakhurana I believe the video of the updated UI below conforms with your feedback on the select component validation. I am looking forward to your review.

# Updated Fix;

https://user-images.githubusercontent.com/105272189/227068091-ac608284-175b-416e-a04f-e5cdbb049441.mp4

Avi-88:
> @yashikakhurana I believe the video of the updated UI below conforms with your feedback on the select component validation. Let me know what you think please. I am looking forward to your review.
> 
> ### Updated Fix;

Awesome @Haywayaheadshot , this looks clean. I noticed that you added the css for the `is-valid` class again, did it not work before that ? because I've included it in the main stylesheet already so it'd be a duplicate, unless of course it is updated for this specific usecase. Also I think you accidentally included a change not related to this issue ( the PageHome test file). Are we using react-icons anywhere ? I too thought about using it for a different issue but refrained from adding extra packages just for a single issue. @yashikakhurana ma'am there are a lot of places we are using `svg's` directly which could  possibly be replaced with icon components, please let me know what you think about this.

Haywayaheadshot:
@Avi-88 thank you for the feedback. The issue I was facing before was that I was adding the form-control to the `select` component. But upon checking inspect element, I realized it was rendering different divs with bootstrap classes. So I targeted particular classes and updated them when the user chooses an option.

Also, the only class I added in the index.scss is the `valid-right` class. This is because anytime the component updates, a `border-success` class is added to the component except for the `close button` and the `drop-down arrow`. So I removed it and added it to the component instead. Also, the `is-valid` is a bootstrap class.

yashikakhurana:
@Haywayaheadshot glad you are one more step closer, try to work on solving test issues, thanks

Haywayaheadshot:
> @Haywayaheadshot glad you are one more step closer, try to work on solving test issues, thanks

You're welcome. Thank you also for being patient and for your guidance.

Haywayaheadshot:
Because

The overview and audience pages highlight all the fields in the green border and checkmark after saving to show that they were validated and saved, but the metrics page does not. We should fix that so that we show the valid fields as green when saving, it's just a nice UI signal that your stuff was saved properly.

This commit

- I defined two state variables `primaryValid` and `secondaryValid` using the useState hook. These variables are initially set to false.
- I added three `useEffect` hooks.
- Two `useEffect` hooks are also defined to handle the validation of the outcomes.
- The first `useEffect` hook is responsible for updating the visual feedback of the validation status of the primary and secondary outcomes. It does this by selecting the corresponding HTML elements in the DOM and adding or removing CSS classes depending on the validation status. It runs whenever either primaryValid or secondaryValid changes.
- The second `useEffect` hook listens for changes in the primaryOutcomes array using the dependency array `[primaryOutcomes]`. Whenever `primaryOutcomes` changes, the hook sets `primaryValid` to true if the length of `primaryOutcomes` is greater than `0`, otherwise it sets `primaryValid` to `false`.
- The third `useEffect` hook is similar to the second hook but listens for changes in the `secondaryOutcomes` array instead.

### Video For;

- After fix


https://user-images.githubusercontent.com/105272189/227385027-3b4541e9-a870-4efd-b045-3b4a30b65cb9.mp4

experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx
```diff
@@ -128,6 +128,69 @@ const FormMetrics = ({
...
+    }
+  }, [primaryValid, secondaryValid]);
+
+  useEffect(() => {
+    setPrimaryValid(primaryOutcomes.length > 0);
+  }, [primaryOutcomes]);
+
+  useEffect(() => {
+    setSecondaryValid(secondaryOutcomes.length > 0);
+  }, [secondaryOutcomes]);
```
yashikakhurana, line 192: few things 
Here are some optimizations you can make to this code:
-  Use useRef instead of querySelector:
    querySelector is not recommended for performance reasons because it searches the entire DOM tree. Instead, you can use useRef to reference the specific DOM element you want to manipulate.
- Combine the two useEffect hooks into one:
    Since both useEffect hooks are checking for changes in the same state variables, you can combine them into one hook to reduce code duplication.
  -Use a single state variable instead of two:
    Since the only difference between primaryValid and secondaryValid is the DOM elements they reference, you can use a single state variable to keep track of both. You can also combine the two setPrimaryValid and setSecondaryValid into one function that takes an argument to determine which one to update.
- Avoid unnecessary classList operations:
    If a DOM element already has a certain class, there's no need to add or remove it again. You can use classList.toggle() to add or remove a class based on whether it's already present.

you can do something like this probably, I haven't tested the code its a suggestion
```javascript 
const [valid, setValid] = useState({ primary: false, secondary: false });

const primaryContainerDivRef = useRef();
const secondaryContainerDivRef = useRef();
const primarySelectContainerRef = useRef();
const secondarySelectContainerRef = useRef();

useEffect(() => {
  const primaryContainerDiv = primaryContainerDivRef.current;
  const secondaryContainerDiv = secondaryContainerDivRef.current;
  const primarySelectContainer = primarySelectContainerRef.current;
  const secondarySelectContainer = secondarySelectContainerRef.current;

  if (primaryContainerDiv && primarySelectContainer) {
    primaryContainerDiv.classList.toggle("form-control", valid.primary);
    primaryContainerDiv.classList.toggle("is-valid", valid.primary);
    primaryContainerDiv.classList.toggle("valid-right", valid.primary);
    primarySelectContainer.classList.toggle("border-success", valid.primary);
  }

  if (secondaryContainerDiv && secondarySelectContainer) {
    secondaryContainerDiv.classList.toggle("form-control", valid.secondary);
    secondaryContainerDiv.classList.toggle("is-valid", valid.secondary);
    secondaryContainerDiv.classList.toggle("valid-right", valid.secondary);
    secondarySelectContainer.classList.toggle("border-success", valid.secondary);
  }
}, [valid]);

useEffect(() => {
  setValid((prevState) => ({
    ...prevState,
    primary: primaryOutcomes.length > 0,
  }));
}, [primaryOutcomes]);

useEffect(() => {
  setValid((prevState) => ({
    ...prevState,
    secondary: secondaryOutcomes.length > 0,
  }));
}, [secondaryOutcomes]);

**Review by yashikakhurana — CHANGES_REQUESTED** [review]

File: experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx (hunk 1 above)
Comment (Haywayaheadshot) [reply experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx:192]: @yashikakhurana Thank you for the feedback. I will work on the changes asap.

**Comment by Haywayaheadshot** [comment]
@yashikakhurana Thank you for the review. Your suggested refactor code was great. I just had to tweak it a little.

File: experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx (hunk 1 above)
Comment (Haywayaheadshot) [reply experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx:182]: @yashikakhurana when I placed a ref on the `Select` component. I got an error that it was not compatible with the type expected by the `ref` prop of the `Select` component.

So I placed the ref on the `Form.Group` components. And grabbed the necessary classLists to be updated using `getElementsByClassName`[0]. This is because using `getElementByClassName` instead of `querySelector` can help with performance since it restricts the search to a specific set of elements with the specified class name.

It also allowed me to limit the refs to just two for each `Form.Group`.

File: experimenter/experimenter/nimbus-ui/src/styles/index.scss [hunk 2]
```diff
@@ -163,4 +163,8 @@ select.form-control.is-warning {
     border-color: #28a745;
     padding-right: calc(0.8em + 2.8rem);
     background: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' height='24' width='24' viewBox='0 0 20 20' aria-hidden='true' focusable='false' class='css-tj5bde-Svg'%3E%3Cpath fill='lightgray' d='M4.516 7.548c0.436-0.446 1.043-0.481 1.576 0l3.908 3.747 3.908-3.747c0.533-0.481 1.141-0.446 1.574 0 0.436 0.445 0.408 1.197 0 1.615-0.406 0.418-4.695 4.502-4.695 4.502-0.217 0.223-0.502 0.335-0.787 0.335s-0.57-0.112-0.789-0.335c0 0-4.287-4.084-4.695-4.502s-0.436-1.17 0-1.615z'%3E%3C/path%3E%3C/svg%3E") right 0.60rem center/20px 25px no-repeat,#fff url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' width='6' height='6' viewBox='0 0 8 8'%3e%3cpath fill='%2328a745' d='M2.3 6.73L.6 4.53c-.4-1.04.46-1.4 1.1-.8l1.1 1.4 3.4-3.8c.6-.63 1.6-.27 1.2.7l-4 4.6c-.43.5-.8.4-1.1.1z'/%3e%3c/svg%3e") center right 2rem/calc(0.8em + 0.3rem) calc(0.8em + 0.3rem) no-repeat !important;
-}
\ No newline at end of file
+}
+
+.valid-right {
+  border: none;
+}
```
yashikakhurana, line 170: you can avoid adding css here and can use bootstrap CSS as a form of class, I will suggest checking to the website to see more options- https://react-bootstrap-v4.netlify.app/components/alerts/

yashikakhurana (changes requested):
@Haywayaheadshot one small suggestion :)

Haywayaheadshot:
@yashikakhurana Sorry! I forgot to leave a comment. I added the class with bootstrap. Please let me know if there is anything else I should do.

experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx
```diff
@@ -128,6 +128,59 @@ const FormMetrics = ({
...
+        valueContainer.classList.toggle("is-valid", valid.primary);
+        valueContainer.classList.toggle("border-0", valid.primary);
+        valueContainerDiv?.classList.toggle("border-success", valid.primary);
+      }
+    }
+
+    if (secondaryContainerDiv) {
+      const valueContainer = secondaryContainerDiv.getElementsByClassName(
+        "css-g1d714-ValueContainer",
+      )[0];
```
yashikakhurana, line 158: here you are using container class, which can change, I would suggest using the ids concept, like try to use getElementById() instead of containers id

experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx
```diff
@@ -128,6 +128,59 @@ const FormMetrics = ({
...
+      const valueContainerDiv =
+        secondaryContainerDiv.getElementsByClassName("css-yk16xz-control")[0];
```
yashikakhurana, line 160: same as in here

yashikakhurana (changes requested):

experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx
Haywayaheadshot: Sure. I will work on that asap. Thanks

Haywayaheadshot:
@yashikakhurana The HTML elements for the `Select` component renders a couple of `divs`. Below is a screenshot of the `divs`. It only has classes. I am not sure `getElementById` is possible.

Screen Shot 2023-03-29 at 3 52 43 PM

Haywayaheadshot:
@yashikakhurana The HTML elements for the `Select` component are a couple of `divs`. Below is a screenshot of the `divs`. It only has classes. I am not sure `getElementById` is possible.

Screen Shot 2023-03-29 at 3 52 43 PM

experimenter/experimenter/nimbus-ui/src/components/PageEditMetrics/FormMetrics/index.tsx
Haywayaheadshot: The HTML elements for the `Select` component are a couple of `divs`. Below is a screenshot of the `divs`. It only has classes. I am not sure `getElementById` is possible.

Screen Shot 2023-03-29 at 3 52 43 PM

Haywayaheadshot:
@yashikakhurana As I said earlier. The `Select` components render HTML `div`s with classes. To `getElementById`, I must add an `id` to the rendered div. And I do not think it is a good idea we do that. Below is a short video to further explain my predicament.



https://user-images.githubusercontent.com/105272189/228685637-7d8781c5-0acc-4605-8c79-8139556bcc34.mp4

Avi-88:
> @yashikakhurana As I said earlier. The `Select` components render HTML `div`s with classes. To `getElementById`, I must add an `id` to the rendered div. And I do not think it is a good idea we do that. Below is a short video to further explain my predicament.

Hi @Haywayaheadshot , I understand what you mean, what you could do is add `id` to the <Select/> component and then access the nested div using something like this `const div = select.querySelectorAll('div')[0];`. I'll try and see If I can find something more specific but this is the approach I can think of currently. Hope this helps :)

yashikakhurana:
> > @yashikakhurana As I said earlier. The `Select` components render HTML `div`s with classes. To `getElementById`, I must add an `id` to the rendered div. And I do not think it is a good idea we do that. Below is a short video to further explain my predicament.
> 
> Hi @Haywayaheadshot , I understand what you mean, what you could do is add `id` to the component and then access the nested div using something like this ( const div = select.querySelectorAll('div')[0] ). I'll try and see If I can find something more specific but this is the approach I can think of currently. Hope this helps :)
@Avi-88 yup exactly this is what I am looking for and thank you for helping him too :)

Haywayaheadshot:
@Avi-88 @yashikakhurana Thank you so much for helping and for pointing me in the right direction. But I am a little confused and I have a few questions.

1.) What then is the purpose of using the ref hook if I have to add unique IDs to the `Select` component to `getElementsBy`?
2.) Due to performance reasons, @yashikakhurana advised that I used `getElementsById` instead of `querySelector`. I ended up using `getElementsByClassName` because they had no IDs. Does that mean ( const div = select.querySelectorAll('div')[0] ) is an exception?
3.) Lastly, the `Select` component has `div`s nested in it that are only visible when rendered. The picture below is the `Select` component wrapped in a `Form.Control`.

Screen Shot 2023-03-30 at 5 08 06 PM

And the image below is the rendered `div`s when I inspect element.

Screen Shot 2023-03-30 at 5 16 02 PM

With that, I added an `Id="primary-outcomes-select"` to the primary `Select` component and used it to validate the inputs. Like in the image below.
Screen Shot 2023-03-30 at 5 31 32 PM

@yashikakhurana Does this match what you are looking for?

Avi-88:
> 

Ok, so 
1. You can absolutely use useRef hook instead of `getElementById` here (just for the <Select/> component )because of the `div` you want to access being nested in the component and is visible only in the dom. If this was not the case and the div was accessible via `id` then `getElementById` is the better choice.

2. No `querySelectorAll` is not an exception, but its still better ( in terms of flexibility ) than `getElementByClassName` which is the next thing to consider after performance.

The last one I'll leave it to @yashikakhurana ma'am to answer as well as correct me if I'm wrong on the above mentioned points, hopefully this helped a little @Haywayaheadshot  good luck :)

Haywayaheadshot:
> Ok, so
> 
> [ 1 ] You can absolutely use useRef hook instead of `getElementById` here (just for the component )because of the div you want to access being nested in the component and is visible only in the dom. If this was not the case and the div was accessible via id then getElementById is the better choice [ 2 ] No ,querySelectorAll is not an exception, but its still better ( in terms of flexibility ) than getElementByClassName which is the next thing to consider after performance. The last one I'll leave it to @yashikakhurana ma'am to answer as well as correct me if I'm wrong on the above mentioned points, hopefully this helped a little @Haywayaheadshot good luck :) (ps - sorry for the bad indentation, github markup is acting wierd in this one )

@Avi-88 Thank you so much. I really do appreciate the effort you have put into helping me understand. I will keep working on the code. Thanks again @Avi-88

Haywayaheadshot:
@yashikakhurana Following your suggestion to select the divs with an id instead of by classNames that are bound to change, I added `id`'s to the `Select` components of the primary and secondary outcomes. Then I used them to select the respective divs to be updated. Below is a snippet of the pushed code of how the primary outcomes validation is now done.

Please let me know how else to improve on the code.

 ```
  if (primaryContainerDiv) {
      const valueContainer = primaryContainerDiv.querySelector(
        "#primary-outcomes > div > div",
      );
      const valueContainerDiv = primaryContainerDiv.querySelector(
        "#primary-outcomes > div",
      );
      if (valueContainer) {
        valueContainer.classList.toggle("form-control", valid.primary);
        valueContainer.classList.toggle("is-valid", valid.primary);
        valueContainer.classList.toggle("border-0", valid.primary);
        valueContainerDiv?.classList.toggle("border-success", valid.primary);
      }
    }
```

yashikakhurana:
@Haywayaheadshot the above approach looks good to me

Haywayaheadshot:
@yashikakhurana Thank you. I have pushed the updated code.

yashikakhurana (changes requested):
I reviewed your code I think something you are missing here, if you will visit the audience page and enter/change the expected no. of clients it will show you green tick 
image
and once you move to other page and visit audience page again, that green tick will be no longer visible, the idea of the green tick is to show the ticks when user is typing something to show them if its valid entered value or not, in your case green ticks are showing up but they are never going back even you move to different pages 
image

Haywayaheadshot:
> I reviewed your code I think something you are missing here, if you will visit the audience page and enter/change the expected no. of clients it will show you green tick
> image
> and once you move to other page and visit audience page again, that green tick will be no longer visible, the idea of the green tick is to show the ticks when user is typing something to show them if its valid entered value or not, in your case green ticks are showing up but they are never going back even you move to different pages
> image

Sure! Thank you for bringing it to my notice. I will work on that change asap.

Haywayaheadshot:
> @Haywayaheadshot do you want to close this PR and create a new one, once you finish woking on this?


@yashikakhurana. It looks like a good idea. Seeing that there are a lot of comments on this one. I will do that.

Issue #8559, opened 2023-03-22
Update diagrams for isRolloutDirty instead of dirty publish status


data-sync-user:
In [https://mozilla-hub.atlassian.net/browse/EXP-3256](https://mozilla-hub.atlassian.net/browse/EXP-3256|smart-link) we made a separate bool on the experiment (similar to isEnrollmentPaused) to track isRolloutDirty. 

This will also require our sequence diagrams to be updated to show this new logic.

[https://mozilla-hub.atlassian.net/browse/EXP-3317](https://mozilla-hub.atlassian.net/browse/EXP-3317|smart-link) handles changing the logic, and should be landed at the same time as this.

[https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3319)

Issue #8558, opened 2023-03-22
Use isRolloutDirty instead of dirty publish status


data-sync-user:
In [https://mozilla-hub.atlassian.net/browse/EXP-3256](https://mozilla-hub.atlassian.net/browse/EXP-3256|smart-link) we made a separate bool on the experiment (similar to isEnrollmentPaused) to track isRolloutDirty. 

This will also require our sequence diagrams to be updated to show this new logic

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3317)

Issue #8439, opened 2023-03-08
Use isRolloutDirty flag on page summary


data-sync-user:
In [https://mozilla-hub.atlassian.net/browse/EXP-3256](https://mozilla-hub.atlassian.net/browse/EXP-3256|smart-link) we made a separate bool on the experiment (similar to isEnrollmentPaused) to track isRolloutDirty. We only want to use this on the Unpublished changes status pill and the bucketing warning on PageSummary. Any other changes (kinto tasks, serializers for v5, and approval UI) will require our sequence diagrams to be updated, which can come later.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3257)

PR #8560, opened 2023-03-22
For #8439: use isRolloutDirty flag in addition to publishStatus

eliserichards:
Because

- We added `is_rollout_dirty` in #8850
- We are going to use this flag to determine whether a rollout is dirty **in addition to publish status** (for now.. publish status will be removed later, see https://github.com/mozilla/experimenter/issues/8559 and https://github.com/mozilla/experimenter/issues/8558)

This commit

- Both `publish_status=DIRTY` and `is_rollout_dirty=true` are being set to denote that the rollout has been edited (has unpublished updates)
- This is being used **only for** the "Unpublished changes" status pill and the "Request update" buttons

----
image

----

image

### DIFF
--- a/experimenter/experimenter/kinto/tasks.py
+++ b/experimenter/experimenter/kinto/tasks.py
@@ -188,7 +188,7 @@ def handle_updating_experiments(applications, records):
             experiment.publish_status = NimbusExperiment.PublishStatus.IDLE
             experiment.status_next = None
             experiment.published_dto = published_record
-            experiment.is_dirty = False
+            experiment.is_rollout_dirty = False
             experiment.save()
 
             generate_nimbus_changelog(
@@ -212,6 +212,7 @@ def handle_ending_experiments(applications, records):
             experiment.status = NimbusExperiment.Status.COMPLETE
             experiment.status_next = None
             experiment.publish_status = NimbusExperiment.PublishStatus.IDLE
+            experiment.is_rollout_dirty = False
             experiment.save()
 
             generate_nimbus_changelog(
--- a/experimenter/experimenter/kinto/tasks.py
+++ b/experimenter/experimenter/kinto/tasks.py
@@ -188,6 +188,7 @@ def handle_updating_experiments(applications, records):
             experiment.publish_status = NimbusExperiment.PublishStatus.IDLE
             experiment.status_next = None
             experiment.published_dto = published_record
+            experiment.is_dirty = False
             experiment.save()
 
             generate_nimbus_changelog(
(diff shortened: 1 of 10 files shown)


experimenter/experimenter/kinto/tasks.py
```diff
@@ -188,6 +188,7 @@ def handle_updating_experiments(applications, records):
             experiment.publish_status = NimbusExperiment.PublishStatus.IDLE
             experiment.status_next = None
             experiment.published_dto = published_record
+            experiment.is_rollout_dirty = False
```
jaredlockhart, line 191: I don't see any changes to the task tests to check for the cases where it goes from True > False?

jaredlockhart (dismissed):
Looks good, makes perfect sense, I think some tests are missing?

experimenter/experimenter/nimbus-ui/src/lib/experiment.ts
```diff
@@ -50,16 +53,19 @@ export function getStatus(
       statusNext === NimbusExperimentStatusEnum.COMPLETE,
     updateRequested:
       isRollout === true &&
+      isRolloutDirty === true &&
       status === NimbusExperimentStatusEnum.LIVE &&
       publishStatus === NimbusExperimentPublishStatusEnum.REVIEW &&
       statusNext === NimbusExperimentStatusEnum.LIVE,
     updateRequestedApproved:
       isRollout === true &&
+      isRolloutDirty === true &&
```
yashikakhurana, line 62: once request get approved, this should go back to false??

yashikakhurana (dismissed):
@eliserichards thank you, one suggestion if you could add a test like start with rollout, isRolloutDirty false, then you changed the population size, it should turn isRolloutDirty to true and after approval isRolloutDirty back to false and so on

experimenter/experimenter/nimbus-ui/src/lib/experiment.ts
eliserichards: this will go back to false once the update is fully approved in experimenter _and_ remote settings (which means it goes through `publish_status`=DIRTY -> REVIEW -> APPROVE -> WAITING -> IDLE). This means that the kinto tasks will set `publish_status` back to IDLE and `is_rollout_dirty` to False:

experimenter/experimenter/kinto/tasks.py
eliserichards, line 191: @yashikakhurana here

experimenter/experimenter/nimbus-ui/src/lib/experiment.ts
yashikakhurana: okay okay makes sense 💟 , thank you @eliserichards

experimenter/experimenter/kinto/tests/test_tasks.py
```diff
@@ -509,6 +511,7 @@ def test_check_with_approved_update_sets_experiment_to_idle_saves_published_dto(
                 new_publish_status=NimbusExperiment.PublishStatus.IDLE,
             ).exists()
         )
+        self.assertFalse(updated_experiment.is_rollout_dirty)
```
yashikakhurana, line 514: is_rollout_dirty is false again 🥳

yashikakhurana (approved):
thank you @eliserichards for adding more tests, looks 💯

PR #8564, opened 2023-03-22
fix(nimbus): Put pref exclusion targeting behind a sticky expression

### LINKED ISSUE #8563: Preference exclusion targeting causing unenrollments
An experiment targeting an early startup feature that sets prefs and uses the pref exclusion targeting ("prevent pref conflicts"), experiments will end up unenrolling themselves due to targeting mismatch. This is due to the `isEarlyStartup=true` features set prefs on the user branch and the pref exclusion targeting checks if any of the prefs controlled by the feature are set on the user branch.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3321)

brennie:
Because

- isEarlyStartup features cause setPref variables to set values on the user branch;
- the "prevent pref conflict" targeting checks for prefs set on the user branch;
- the "prevent pref conflict" targeting will be evaluated every time Nimbus syncs Remote Settings; and
- the above causes experiments to unenroll when targeting is re-evaluated

This commit

- puts the pref conflict targeting behind its own sticky expression, regardless of whether or not the experiment uses sticky targeting

Fixes #8563

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -369,26 +369,26 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             sticky_expressions.append(f"region in {countries}")
 
         if self.is_sticky and sticky_expressions:
-            sticky_clause = "is_already_enrolled"
-            if is_desktop:
-                if self.is_rollout:
-                    sticky_clause = "experiment.slug in activeRollouts"
-                else:
-                    sticky_clause = "experiment.slug in activeExperiments"
-            sticky_expressions_joined = " && ".join(
-                [f"({expression})" for expression in sticky_expressions]
+            expressions.append(
+                make_sticky_targeting_expression(
+                    is_desktop, self.is_rollout, sticky_expressions
+                )
             )
-            sticky_expression = f"({sticky_clause}) || ({sticky_expressions_joined})"
-            expressions.append(sticky_expression)
         else:
             expressions.extend(sticky_expressions)
 
         if prefs := self._get_targeting_pref_conflicts():
-            expressions.extend(f"!('{pref}'|preferenceIsUserSet)" for pref in prefs)
+            expressions.append(
+                make_sticky_targeting_expression(
+                    is_desktop,
+                    self.is_rollout,
+                    (f"!('{pref}'|preferenceIsUserSet)" for pref in prefs),
+                )
+            )
 
         #  If there is no targeting defined all clients should match, so we return "true"
         return (
-            " && ".join([f"({expression})" for expression in expressions])
+            " && ".join(f"({expression})" for expression in expressions)
             if expressions
             else "true"
         )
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -369,26 +369,26 @@ def targeting(self):
...
+                    is_desktop,
+                    self.is_rollout,
+                    (f"!('{pref}'|preferenceIsUserSet)" for pref in prefs),
+                )
+            )
 
         #  If there is no targeting defined all clients should match, so we return "true"
         return (
-            " && ".join([f"({expression})" for expression in expressions])
+            " && ".join(f"({expression})" for expression in expressions)
```
jaredlockhart, line 391: This might be causing that test failure?

jaredlockhart (changes requested):
Looks right to me, thank you, looks like a string is getting flattened somewhere and iterating by character in the tests?

experimenter/experimenter/experiments/models.py
```diff
@@ -369,26 +369,26 @@ def targeting(self):
...
+                    is_desktop, self.is_rollout, sticky_expressions
+                )
             )
-            sticky_expression = f"({sticky_clause}) || ({sticky_expressions_joined})"
-            expressions.append(sticky_expression)
         else:
             expressions.extend(sticky_expressions)
 
         if prefs := self._get_targeting_pref_conflicts():
-            expressions.extend(f"!('{pref}'|preferenceIsUserSet)" for pref in prefs)
```
Mardak, line 387: Would moving these 2 lines of original code above the `is_sticky` check and putting pref conflict checks as part of `sticky_expressions` do what we want and be a bit simpler?
```py
if countries…
  sticky_expressions.append(country stuff)

if pref_conflicts…
  sticky_expressions.append(prefs stuff)

if is_sticky…
```

brennie: No, the pref conflict check must always be sticky, even if the targeting is not sticky. Otherwise we will see unenrollments when experiments use an `isEarlyStartup=true` feature. Nimbus will flip the pref and when the targeting gets re-evaluated, the pref conflict targeting will fail due to thet pref being set.

experimenter/experimenter/experiments/models.py
brennie: No, thats a generator expression that yields strings. It should behave identitically but be faster.

The problem was:

```python
        if prefs := self._get_targeting_pref_conflicts():
            expressions.extend( # <-- calling extend with str instead of append
                make_sticky_targeting_expression(
```

jaredlockhart: Ah yep 👍

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -747,9 +747,10 @@ def test_targeting_with_prevent_pref_conflicts_set_prefs(self):
             experiment.targeting,
             (
                 '(browserSettings.update.channel == "release") && '
+                "((experiment.slug in activeExperiments) || "
                 "(!('nimbus.test.string'|preferenceIsUserSet)) && "
                 "(!('nimbus.test.int'|preferenceIsUserSet)) && "
-                "(!('nimbus.test.boolean'|preferenceIsUserSet))"
+                "(!('nimbus.test.boolean'|preferenceIsUserSet)))"
```
jaredlockhart, line 753: My favourite part of this PR is that all the existing tests remain unchanged and pass and this test changes to the new expected behaviour.  I love it.

jaredlockhart (approved):
Approved, looks fantastic, good idea pulling the sticky formatter out, thanks @brennie ! 🎉 🎉 🎉 🎉

PR #8570, opened 2023-03-22
fixes mozilla#8345 bug(nimbus): highlight fields on branches page

emmaz12:
Because:
- the branches page did not highlight when the save button is clicked

This Commit
- allows all the input fields to be highlighted when the save button is clicked

https://user-images.githubusercontent.com/117552772/227035003-b6d58ff2-63c0-4640-bd07-8750b0af97eb.mov


experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
```diff
@@ -227,6 +232,7 @@ export const FormBranches = ({
         className="my-3"
         noValidate
         onSubmit={handleSave}
+        validated={isValid && formSubmitted}
```
yashikakhurana, line 235: there is something wrong, I don't think we are using states to show green checks, I would suggest doing some research more on this without state, also if you will see on other page if we change let's say expected no. of clients and if its valid it only shows green on that part, but in your case, all the other checkboxes are becoming green too, it shouldn't be doing that
image

yashikakhurana (changes requested):
I would suggest to retry your approach, but thank you for looking into this @emmaz12

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
emmaz12: @yashikakhurana I will look into that! However, just to clarify, it seemed like the branches page already had the green checks implemented when you change an input field. As seen in the image you attached, I also attached an image from the branches page to simulate that. From what I could tell, the only difference between the audience and the branches page was that when the "Save" button was pressed, the audience page saved everything on the form. However, the branches page did not. Please correct me if I am wrong :) 
Screen Shot 2023-03-22 at 11 25 31 AM

https://user-images.githubusercontent.com/117552772/227320456-e3021d7b-480a-4eef-9c9a-bc5d82ec53b5.mov

emmaz12: @yashikakhurana Thanks for catching that! I believe it should be updated now :) Thanks again.

emmaz12:
@yashikakhurana Sounds good! My apologies for not catching that. Will fix this issue!

emmaz12:
@yashikakhurana How does this look now? Thanks for your patience!

emmaz12:
@yashikakhurana How does this look? Thanks again for you patience!

yashikakhurana (approved):
thank you @emmaz12 working 💯 , great job 🎉

PR #8575, opened 2023-03-24
fixes #8498 fix(nimbus): change search bar so that it checks if input…

### LINKED ISSUE #8498: Search space issue

Prithvi-Chanchal:
… contains space at the end or not, if it does not have space at end add space at end and also change test case according to that

Because

- getting different result when input string in search bar is with space and without space

This commit

- change the search bar such that it checks if input string contains space at the end or not, if input string does not contain any space at the end add space at end
- modify the test cases according to the change


Prithvi-Chanchal:
@yashika-khurana I have modified the test file in page home also, please check that and see if it will cause error anywhere'

experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx
```diff
@@ -58,6 +58,16 @@ const SearchBar: React.FunctionComponent<SearchBarProps> = ({
...
+    // Check if the input value already has a space at the end
+    if (
+      inputValue.length > 0 &&
+      inputValue.charAt(inputValue.length - 1) !== " "
+    ) {
+      // If not, add a space at the end
+      inputValue = inputValue + " ";
+    }
+
     setSearchTerms(event.target.value);
```
yashikakhurana, line 71: you can do `setSearchTerms(inputValue)`

experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx
```diff
@@ -66,7 +76,7 @@ const SearchBar: React.FunctionComponent<SearchBarProps> = ({
       setClearIcon(true);
 
       const newTimer = setTimeout(() => {
-        const results = fuse.search(searchTerms);
+        const results = fuse.search(inputValue);
```
yashikakhurana, line 79: then you don't have to do this

yashikakhurana (changes requested):
thank you @Prithvi-Chanchal, one suggestion to use that state itself, let me know if you have any questions thank you :)

experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.tsx
Prithvi-Chanchal: Yes I have already tried that, but the when you type anything it will add space after every character and show that in search bar and you could not delete anything in the search bar, you can only add new characters but they will also be added with space.
see this screen shot for reference:
Screenshot from 2023-03-25 00-20-59
if we enter 'lol' it will be take input as 'l o l ' like this that's why I am not modifying setSeachTerms.
@yashikakhurana

Prithvi-Chanchal:
@yashikakhurana ma'am could you please check, if I have done it right

yashikakhurana (approved):
thank you @Prithvi-Chanchal looks good now, thank you for updating the test cases too :)

PR #8586, opened 2023-03-27
fixes #8499 fix(nimbus) store and revisit search result with test

### LINKED ISSUE #8499: Store search results

KramStyles:
Because

- the filtered search results are not retained after navigating away from the search page, we need to modify the behavior of the "back to experiment" button to retain the search term and filtered results.

This commit

- Test to ensure new fix works properly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,9 @@ target/
 # VS Code
 **/.vscode/*
 
+# WebStorm
+**/.idea/*
+
 # Hosted assets
 experimenter/experimenter/served/
 experimenter/experimenter/legacy/legacy-ui/assets/
--- a/experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageHome/SearchBar/index.test.tsx
@@ -39,6 +39,30 @@ describe("SearchBar", () => {
       expect(searchInput).toHaveValue("");
     });
   });
+
+  it("updates the component when search values are saved in localStorage", async () => {
+    // Mock localStorage to return a particular value
+    jest.spyOn(Storage.prototype, "getItem").mockImplementation((key) => {
+      return JSON.stringify({
+        searchValue: "Experiment",
+        results: [{ item: "Saved experiment details" }],
+      });
+    });
+
+    const onChange = jest.fn();
+    render(<Subject experiments={[]} onChange={onChange} />);
+
+    // wait for useEffect to complete
+    await waitFor(() => {
+      // check that the component has updated with the values from localStorage
+      const searchInput = screen.getByTestId("SearchExperiments");
+      expect(searchInput).toBeInTheDocument();
+      expect(searchInput).toHaveValue("Experiment");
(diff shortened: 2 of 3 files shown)


yashikakhurana (changes requested):
@KramStyles thank you for working on this, can you explain your approach in more detail, like why have you added another timer we already have one timer why not having that single timer only

KramStyles:
> @KramStyles thank you for working on this, can you explain your approach in more detail, like why have you added another timer we already have one timer why not having that single timer only

I'm so sorry. I forgot to take out the timer. My initial approach was to refetch the data but since I'm saving the experiments in the localStorage, I believe there's no need for the timer as the information would be displayed immediately. 

Can I proceed to take out the timer in the useEffect since I won't be needing it? @yashikakhurana

yashikakhurana:
@KramStyles yes sure go ahead :)

KramStyles:
> @KramStyles yes sure go ahead :)

@yashikakhurana I took out the timer and worked with only this code within the useEffect but it wasn't updating the experiments when the user returns. 
```
      // get search term from localStorage
      let fetchedValues = localStorage.getItem("nimbus-ui-search");
      if (fetchedValues !== null) {
        fetchedValues = JSON.parse(fetchedValues);
        const searchValue = (fetchedValues as any)["searchValue"] as string;
        const results = (fetchedValues as any)["results"] as any[];
        const searchResults = results.map((character) => character.item);

        // Update components
        onChange(searchResults);
        setSearchTerms(searchValue);
        setClearIcon(true);
      }
  });
```

I decided to work with a setTimeout function without assigning it to a timer variable 
```
setTimeout(() => {
      // get search term from localStorage
      let fetchedValues = localStorage.getItem("nimbus-ui-search");
      if (fetchedValues !== null) {
        fetchedValues = JSON.parse(fetchedValues);
        const searchValue = (fetchedValues as any)["searchValue"] as string;
        const results = (fetchedValues as any)["results"] as any[];
        const searchResults = results.map((character) => character.item);

        // Update components
        onChange(searchResults);
        setSearchTerms(searchValue);
        setClearIcon(true);
      }
    }, 700);
    return () => {};
  });
```
and I was warned of a memory leak which was noticeable when I ran tests as I wasn't able to clear the timeout. I believe it doesn't work properly without the timer variable.

yashikakhurana:
@KramStyles I would suggest take some more time and think what best could be done :)

KramStyles:
> @KramStyles I would suggest take some more time and think what best could be done :)

@y

> @KramStyles I would suggest take some more time and think what best could be done :)

@yashikakhurana Thank you. You mentioned working with the already created timer but I don't know how I can do that. Is it possible to get some help on how I can achieve this?!

yashikakhurana:
@KramStyles I would suggest do your research what is the best way to tackle this issue and what are the alternatives, see pros and cons and come up with optimal solution for this

yashikakhurana:
@KramStyles I am closing this pr for now, you are welcome to come with a new approach, and playgaound and submit a new PR, let me know if you have any questions :)

Issue #8581, opened 2023-03-24
Cloning a dirty rollout should make publish status idle


data-sync-user:
**[Prerequisites]:**

* Have a new Nimbus rollout created and opened.

**[Steps to reproduce]:**

1. Click the “Audience” button from the navigation menu. 
1. Observe the value from the “Percent of clients” field.
1. Click the “Save and continue” button.
1. Observe the results on the “Summary page”

**[Expected result]:**

* “Unpublished changes" does not appear on the “Summary” page if no changes are made to the “Percent of Clients” field.

**[Actual result]:**

* “Unpublished changes” pill appears after pressing the “Save and Continue” button even if no changes are made. 

**[Notes]:** 

* Attached is a recording of the issue.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-3328)

PR #8588, opened 2023-03-27
For #8581 fix(nimbus) test(nimbus): Dirty rollouts should not be dirty when cloned

eliserichards:
Because

- A cloned rollout should not persist the `is_rollout_dirty` value

This commit

- Explicitly sets `is_rollout_dirty` back to `None` when a rollout is cloned
- Adds tests


yashikakhurana (approved):
oops we always miss out our clone method thank you for fixing that 🎉  @eliserichards

PR #8589, opened 2023-03-27
For #8574 fix(nimbus): Only updating population percent sets dirty publish status

eliserichards:
Because

- We only want to make our `publish_status=DIRTY` when a change has been made
- It was setting DIRTY every time `update` was called (when you click "Save and continue") even if there was no change

This commit

- Checks to see if the population percent value has changed on update
- Adds tests

### DIFF
--- a/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
@@ -1222,6 +1222,48 @@ class TestNimbusExperimentSerializer(TestCase):
         experiment = serializer.save()
         self.assertEqual(experiment.publish_status, NimbusExperiment.PublishStatus.IDLE)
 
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -1222,6 +1222,48 @@ def test_can_update_publish_status_for_non_approved_state(self):
...
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LAUNCH_APPROVE_APPROVE,
+            is_rollout=True,
+            population_percent=50,
+        )
+
+        serializer = NimbusExperimentSerializer(
+            experiment,
+            data={
+                "population_percent": 77,
```
yashikakhurana, line 1235: 🎉

yashikakhurana (approved):
yeah completely makes sense, thank you :)

PR #8600, opened 2023-03-28
fix #8531 feat(nimbus): add slug to result fetch logging

### LINKED ISSUE #8531: Log experiment slug in jetstream fetch task
It would be helpful to include the experiment slug in the jetstream fetch task logging so that we can inspect task logs in the Google Cloud Console for a given experiment.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3300)

mikewilli:
Because

- the `slug` is the common identifier for experiments

This commit

- adds experiment `slug` to logging during results fetch

### DIFF
--- a/experimenter/experimenter/jetstream/tasks.py
+++ b/experimenter/experimenter/jetstream/tasks.py
@@ -23,7 +23,12 @@ def fetch_experiment_data(experiment_id):
         metrics.incr("fetch_experiment_data.completed")
     except Exception as e:
         metrics.incr("fetch_experiment_data.failed")
-        logger.info(f"Fetching experiment data for {experiment_id} failed: {e}")
+        failure_message = "Fetching experiment data for {experiment_id} "
+        if experiment is not None and hasattr(experiment, "slug"):
+            failure_message += experiment.slug
+            failure_message += " "
+        failure_message += "failed: {e}"
+        logger.info(failure_message)
         raise e
 
 
@@ -47,13 +52,15 @@ def fetch_jetstream_data():
                     >= dt.date.today()
                 )
             ):
-                logger.info(f"Fetching Jetstream data for {experiment.name}")
+                logger.info(
+                    f"Fetching Jetstream data for {experiment.name} ({experiment.slug})"
+                )
                 fetch_experiment_data.delay(experiment.id)
                 metrics.incr("fetch_jetstream_data.completed")
             else:
                 metrics.incr("fetch_jetstream_data.skipped")
                 logger.info(
-                    f"Skipping cache refresh for old experiment {experiment.name}"
+                    f"Skipping cache refresh for old experiment {experiment.name} ({experiment.slug})"
                 )
 
     except Exception as e:
--- a/experimenter/experimenter/jetstream/tasks.py
+++ b/experimenter/experimenter/jetstream/tasks.py
@@ -16,6 +16,7 @@ metrics = markus.get_metrics("jetstream.tasks")
 @metrics.timer_decorator("fetch_experiment_data")
 def fetch_experiment_data(experiment_id):
     metrics.incr("fetch_experiment_data.started")
+    experiment = None
     try:
         experiment = NimbusExperiment.objects.get(id=experiment_id)
         experiment.results_data = get_experiment_data(experiment)
@@ -23,11 +24,10 @@ def fetch_experiment_data(experiment_id):
         metrics.incr("fetch_experiment_data.completed")
     except Exception as e:
         metrics.incr("fetch_experiment_data.failed")
-        failure_message = "Fetching experiment data for {experiment_id} "
+        failure_message = f"Fetching experiment data for {experiment_id} "
         if experiment is not None and hasattr(experiment, "slug"):
-            failure_message += experiment.slug
-            failure_message += " "
-        failure_message += "failed: {e}"
+            failure_message += f"{experiment.slug} "
+        failure_message += f"failed: {e}"
         logger.info(failure_message)
         raise e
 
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/jetstream/tasks.py
```diff
@@ -23,7 +23,12 @@ def fetch_experiment_data(experiment_id):
         metrics.incr("fetch_experiment_data.completed")
     except Exception as e:
         metrics.incr("fetch_experiment_data.failed")
-        logger.info(f"Fetching experiment data for {experiment_id} failed: {e}")
+        failure_message = "Fetching experiment data for {experiment_id} "
+        if experiment is not None and hasattr(experiment, "slug"):
+            failure_message += experiment.slug
+            failure_message += " "
+        failure_message += "failed: {e}"
+        logger.info(failure_message)
```
yashikakhurana, line 31: nothing wrong on this, just a suggestion 
```python
failure_message = f"Fetching experiment data for {experiment_id} "
if experiment and hasattr(experiment, "slug"):
    failure_message += f"{experiment.slug} "
failure_message += "failed: {e}"
logger.info(failure_message, e=exception)

```

eliserichards (approved):
🚀

yashikakhurana (changes requested):
one minor suggestion :) and ci is complaining for formatting, run `make code_format`

experimenter/experimenter/jetstream/tasks.py
mikewilli: Yea it seemed unnecessary to have an f-string just for a space, but I guess it's probably neater that way. I also missed some f-strings so fixed that up in latest commit. Thanks!

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -16,14 +16,19 @@
...
     try:
         experiment = NimbusExperiment.objects.get(id=experiment_id)
         experiment.results_data = get_experiment_data(experiment)
         experiment.save()
         metrics.incr("fetch_experiment_data.completed")
     except Exception as e:
         metrics.incr("fetch_experiment_data.failed")
-        logger.info(f"Fetching experiment data for {experiment_id} failed: {e}")
+        failure_message = f"Fetching experiment data for {experiment_id} "
+        if experiment is not None and hasattr(experiment, "slug"):
```
yashikakhurana, line 28: I think you can remove `is not None ` completely and can do just `if experiment and ....`

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -47,13 +52,16 @@ def fetch_jetstream_data():
...
+                logger.info(
+                    f"Fetching Jetstream data for {experiment.name} ({experiment.slug})"
+                )
                 fetch_experiment_data.delay(experiment.id)
                 metrics.incr("fetch_jetstream_data.completed")
             else:
                 metrics.incr("fetch_jetstream_data.skipped")
                 logger.info(
                     f"Skipping cache refresh for old experiment {experiment.name}"
+                    + f" ({experiment.slug})"
```
yashikakhurana, line 64: f string under f string 👀

mikewilli: I think this is correct because it is concatenating, and the second string does not know that the first one is an f-string.

experimenter/experimenter/jetstream/tasks.py
mikewilli: I have been bitten by using `if x` instead of `if x is not None` in the past and always default to the latter. Is there a reason to use `if experiment` here (and _not_ `if experiment is not None`)?

yashikakhurana: no major concerns, just a suggestion :)

yashikakhurana (approved):
thank you @mikewilli :)

experimenter/experimenter/jetstream/tasks.py
mikewilli: I saw that as an option, but wasn't sure what the advantage was over the `+`. Can you elaborate or is it just a style/preference thing?

jaredlockhart: Oh yeah, in general I prefer to avoid using the `+` operator when dealing with strings because it's possible to get into situations sometimes where you might have something like

```py
"this" + something + "that"
```

where you expect `something` to be a string, but it's Python, it could be anything, and then you could either get it cast into a string but formatted in some undesired way, or it could be something that refuses to cast into a string altogether and then you get a runtime error.

So in general it's better to always just wrap everything in an fstring, and then when you've explicitly wrapped everything in fstrings, you can use the multiline paren string join behaviour to make the string as long as you want, and then you don't need the `+` at all, and even though it's redundant and you've already declared everything as a string, you don't need to signal to the reader that there's even a chance something is being cast into a string at all.

So yeah it's mostly a stylistic thing just to make it very explicit that everything is being wrapped in an fstring, nothing is being concatted or cast through the `+` operator, everything is wrapped in an fstring.  Just consistency I guess.

mikewilli: Makes sense to me, thanks for the explanation!

mikewilli: I wonder if we could enforce this via one of our various linters...

jaredlockhart: There actually is [such a thing](https://github.com/MichaelKim0407/flake8-use-fstring) and we are using it but I don't think it specifically has a rule about preventing `+` concatenation.

Issue #8522, opened 2023-03-16
The prompt to open Remote Settings for rollout updates is missing some text


data-sync-user:
**[Prerequisites]:**

* Have a Live rollout open.

**[Steps to reproduce]:**

1. Have someone else save a change in the population %.
1. Have that person click “Request Update”.
1. Click “Approve and Update Rollout”.
1. Observe the prompt displayed.

**[Expected result]:**

* The prompt displayed is complete.

**[Actual result]:**

* The prompt displayed reads “*Action required —* Please review this change in Remote Settings to ”.

**[Notes]:**

* Attached is a recording of the issue.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-3293)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/35070">missing text.mov</a>

data-sync-user:
➤ Maria Heres commented:

I have verified that the issue is no longer reproducible. The prompt reads “*Action required —* Please review this change in Remote Settings to update this rollout” now.

PR #8614, opened 2023-03-30
For #8522 fix(nimbus): Check if approved or waiting for update constants

eliserichards:
Because

- We were only checking if we went from [Dirty -> Review](https://github.com/mozilla/experimenter/blob/77401cdbf9f12003c1bab39e16fb609604abd705/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx#L165), [FormRemoteSettingsPending](https://github.com/mozilla/experimenter/blob/77401cdbf9f12003c1bab39e16fb609604abd705/experimenter/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/FormRemoteSettingsPending.tsx#L25) had partial text

This commit

- Adds checks for Review -> Approved and Approved -> Waiting so that we are passing the right constants to the forms

----

before vs after:
image

image

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
@@ -162,7 +162,11 @@ const PageSummary = (props: RouteComponentProps) => {
         approveChange: onLaunchReviewApprovedClicked,
         ...LIFECYCLE_REVIEW_FLOWS.LAUNCH,
       };
-    } else if (status.updateRequested) {
+    } else if (
+      status.updateRequested ||
+      status.updateRequestedApproved ||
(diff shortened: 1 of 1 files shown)


eliserichards:
Since the only thing this is changing is the text, should it have a test? There don't seem to be any existing unit tests that are looking at that specific text 🤔

jaredlockhart:
> Since the only thing this is changing is the text, should it have a test? There don't seem to be any existing unit tests that are looking at that specific text 🤔

Well it looks like it's the text as well as which callbacks are attached to the buttons right?  But yeah I'd say just follow the testing patterns that are already there.  If there's no tests for the existing cases then we can just address it more broadly later.

jaredlockhart (approved):
👍

PR #8620, opened 2023-03-30
fix #8026 fix(nimbus): fix results link displaying erroneously

### LINKED ISSUE #8026: Disable results link until results are available
Jetstream will compute some early results while an experiment is live and we will enable the results link when those become available.  However recent changes to both the error reporting behaviour as well as changes to the jetstream data schema itself have now caused all live experiments to immediately enable the results link but show an error:

Screenshot 2022-12-09 at 13-09-41 Android Reengagment Notifications – Analysis Experimenter

In this case, the results link should be entirely disabled.  Let's see if we can update that check to disable the results link until actual results are available.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2977)

mikewilli:
Because

- schema changes broke our logic around whether to show the results link
- we added a field to the experiment that derives this logic

This commit

- adds this new field to the GQL API response
- uses this new field on the experiment model to determine whether to show the Results link on the UI
- updates related tests


jaredlockhart (approved):
Very nice 👏 👏 👏

PR #8626, opened 2023-03-31
fix #8624 fix(nimbus): clarify meaning of confusing control branch error

### LINKED ISSUE #8624: Clarify 'no control branch' Error on Results page
The error "Unknown error: No control branch found in analysis results." causes a lot of confusion. Most of this confusion should be removed when #8026 is resolved, but we can add clarification about the usual reason for this error to the message just in case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3340)

mikewilli:
Because

- this error message causes confusion

This commit

- explains the usual cause for the error


jaredlockhart (approved):
👍

PR #8630, opened 2023-03-31
fixes #8499 fix(nimbus) retain and display search result when `back to experiment` button is clicked

### LINKED ISSUE #8499: Store search results

KramStyles:
Because

- The present behavior of the "back to experiment" button does not display the search term and filtered results upon navigating back to the search page. Therefore, we need to modify the behavior to persistently retain the filtered search results, which will ensure that the search term and results are readily available when the user navigates back to the search page.

This commit

- Test to ensure new fix works properly


KramStyles:
Hello @yashika-khurana. I've taken out the duplicate timer and I've fixed the memory leak issue that occurred.

yashikakhurana (approved):
perfect, thank you @KramStyles

PR #8659, opened 2023-04-04

Haywayaheadshot:
Because

- The overview and audience pages highlight all the fields in the green border and checkmark after saving to show that they were validated and saved, but the metrics page does not. We should fix that so that we show the valid fields as green when saving, it's just a nice UI signal that your stuff was saved properly.

This commit

- I declared react functional component hooks (useRef and useState) and two functions (`toggleClasses`, `validatePrimary`, and `validateSecondary`).

- The `primaryContainerDivRef` and `secondaryContainerDivRef` are used to store references to the `primary` and `secondary` container div elements, respectively.

- The `toggleClasses` is a function that takes two arguments: `valueContainer`, which is an HTMLElement to toggle classes on, and `isValid`, which is a boolean indicating whether or not the `valueContainer` is valid. This function toggles the `.form-control`, `.is-valid`, and `.border-0` classes on `valueContainer` depending on the value of `isValid`.

- The `validatePrimary` and `validateSecondary` are functions that validate the content of `primaryContainerDivRef` and `secondaryContainerDivRef`, respectively. These functions first check if the container div element exists and then select the appropriate child div element based on its ID. If the child div element exists, `toggleClasses` is called with the appropriate arguments to toggle the classes on the child div. The `border-success` class is also toggled on the parent div element if the child div element exists.

- I called `validatePrimary` and `validateSecondary` in their respective `select` components using an `onBlur` method so the validation can only happen when the user is done selecting.

- Overall, the code validates the contents of two container div elements and toggles classes on their child div elements depending on whether or not the child div elements are valid.

### Video for;
- Before fix;

https://user-images.githubusercontent.com/105272189/229922631-cf793ec8-3a44-4f91-a751-57210caa2a68.mp4

- After fix;

https://user-images.githubusercontent.com/105272189/229923274-3673e5d8-5302-47a7-a5f9-32effc846fbf.mp4


yashikakhurana (changes requested):
thank you for working on this @Haywayaheadshot, if you will notice the behavior of the drop-down after selecting green border also gets green, I will suggest looking more into this, and observing the behavior of the existing dropdowns

Haywayaheadshot:
> thank you for working on this @Haywayaheadshot, if you will notice the behavior of the drop-down after selecting green border also gets green, I will suggest looking more into this, and observing the behavior of the existing dropdowns

Thank you for your review @yashikakhurana. I am working on it.

PR #8666, opened 2023-04-05
fix #8665 chore(nimbus): Add a wait to every element search command for the integration tests

### LINKED ISSUE #8665: Adjust waits for functions on summary page for the integration tests
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3374)

jrbenny35:
Because:
* Having to add waits to each method can be time consuming

This Commit:
* Adds a default wait for every time we need to search for an element.


jaredlockhart (changes requested):
Oh yeah this is a good idea, but instead of just overriding the existing method, can we create a new method called something like `wait_and_find(...)` and then replace all the existing explicit calls to wait and find with calls to that method?

jrbenny35:
> Oh yeah this is a good idea, but instead of just overriding the existing method, can we create a new method called something like `wait_and_find(...)` and then replace all the existing explicit calls to wait and find with calls to that method?

Not sure I understand why we should have 2 separate methods that do kind of the same thing. I think always waiting for the locator makes more sense since most of the time, the wait will be like Milliseconds at most, since the elements are usually already on the page and visible. It will help for those times when the computer is just a little too fast.

jrbenny35:
Looks like it is slowing the tests down quite a bit, I need to figure that out.

experimenter/tests/integration/nimbus/pages/base.py
```diff
@@ -25,3 +25,10 @@ def _wait_for_it(selenium):
                 return True
 
         self.wait.until(_wait_for_it, message=message)
+
+    def find_element(self, strategy, locator):
+        self.wait.until(
+            EC.presence_of_element_located((strategy, locator)),
+            message=f"Element {locator} not found within the timeout period.",
```
yashikakhurana, line 32: I noticed you are using `presence_of_element_located` which may be taking more time to run because it waits for the element to be present in the DOM, you can try using `visibility_of_element_located`, which waits for the element to be visible on the page rather than just present in the DOM, may be this will help cutting the time

jaredlockhart:
@jrbenny35 This change by itself will result in calling 'wait' twice, once explicitly before find_element is called, and once again inside the overriden method.  This was my point.  Overriding the method changes the behaviour of all subsequent callers, and since the method's default behaviour is ONLY to find and not to both wait and find, then you change the behaviour of the method at all the existing call sites.  By making it an entirely new method you preserve the behaviour of the existing methods where that's necessary, and force the caller to be the one to explicitly invoke the find and wait behaviour.  Regardless of the approach, you also need to update all the existing call sites.

jrbenny35:
I updated all the methods to use our find and wait method. We should still be able to use the normal Page Object Model styling we have now, but with waits built in. I also added a check within the ```selenium``` fixture that will not do an experiment cleanup step on CI, this should improve our run times since kinto seems to be slower than it used to be.

.circleci/config.yml
```diff
@@ -186,7 +186,7 @@ jobs:
           name: Run integration tests
           command: |
             cp .env.integration-tests .env
-            make refresh SKIP_DUMMY=1 PYTEST_SENTRY_DSN=$PYTEST_SENTRY_DSN up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
+            make refresh SKIP_DUMMY=1 PYTEST_SENTRY_DSN=$PYTEST_SENTRY_DSN CIRCLECI=$CIRCLECI up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
```
yashikakhurana, line 189: I understand the reason of adding`PYTEST_ARGS` but I am not about the reason behind adding `CIRCLECI`

yashikakhurana (commented):
Oh one of the check is failing and one question to understand the reasoning about it

experimenter/tests/integration/nimbus/pages/experimenter/metrics.py
```diff
@@ -23,22 +25,37 @@ class MetricsPage(ExperimenterBase):
...
+        self.wait_for_and_find_element(*self._primary_outcome_root_locator)
+        multifeature_el = self.wait_for_and_find_element(
+            *self._multifeature_element_locator
+        )
+        # self.wait.until(
+        #     EC.presence_of_all_elements_located(
+        #         self._multifeature_element_locator
+        #     )
+        # )
+        # multifeature_el = root_locator.find_element(self._multifeature_element_locator)
```
yashikakhurana, line 37: do you want to remove this?

.circleci/config.yml
jrbenny35: Yes, this was added to pass in to the ```selenium``` fixture, so that we can skip the experiment clean up step on CI

experimenter/tests/integration/nimbus/pages/experimenter/metrics.py
jrbenny35: Doh, good catch

experimenter/tests/integration/nimbus/pages/experimenter/metrics.py
```diff
@@ -23,22 +25,37 @@ class MetricsPage(ExperimenterBase):
...
         els[0].send_keys(values)
         els[0].send_keys(Keys.RETURN)
 
     @property
     def secondary_outcomes(self):
-        root_locator = self.find_element(*self._secondary_outcome_root_locator)
-        multifeature_el = root_locator.find_element(*self._multifeature_element_locator)
+        # root_locator = self.wait_for_and_find_element(
+        #     *self._secondary_outcome_root_locator
+        # )
```
yashikakhurana, line 49: same in here

experimenter/tests/integration/nimbus/conftest.py
```diff
@@ -98,17 +98,13 @@ def firefox_options(firefox_options):
...
-        kinto_client.approve()
-        summary = SummaryPage(selenium, urljoin(base_url, experiment_slug)).open()
-        summary.wait_for_complete_status()
-    except Exception:
-        pass
+    if os.getenv("CIRCLECI") is None:
+        experiment_slug = str(slugify(experiment_name))
+        try:
+            helpers.approve_and_launch_experiment(experiment_slug)
+            kinto_client.approve()
```
jaredlockhart, line 105: Wait why is this launching it?  I thought the helper here was to end experiments that are already live?

jaredlockhart (changes requested):
Everything is looking good, but I'm confused about that one thing and also looks like still some commented out code to remove?

jaredlockhart (approved):
W00t!  More lines removed than added!  Now that's a refactor I can get behind.  Perfect, thank you @jrbenny35 ! 🎉 🎉 🎉

PR #8667, opened 2023-04-06

Haywayaheadshot:
Because

- The overview and audience pages highlight all the fields in the green border and checkmark after saving to show that they were validated and saved, but the metrics page does not. We should fix that so that we show the valid fields as green when saving, it's just a nice UI signal that your stuff was saved properly.

This commit

- I declared React functional component hooks (`useRef` and `useState`) and two functions (`toggleClasses`, `validatePrimary`, and `validateSecondary`).

- The `primaryContainerDivRef` and `secondaryContainerDivRef` are used to store references to the `primary` and `secondary` container div elements, respectively.

- The `toggleClasses` is a function that takes two arguments: `valueContainer`, which is an `HTMLElement` to toggle classes on, and `isValid`, which is a boolean indicating whether or not the `valueContainer` is valid. This function toggles the `.form-control`, `.is-valid`, and `.border-0` classes on `valueContainer` depending on the value of `isValid`.

- The `validatePrimary` and `validateSecondary` are functions that validate the content of `primaryContainerDivRef` and `secondaryContainerDivRef`, respectively. These functions first check if the container div element exists and then select the appropriate child div element based on its ID. If the child div element exists, `toggleClasses` is called with the appropriate arguments to toggle the classes on the child div. The `border-success` class is also toggled on the parent div element if the child div element exists.

- I called `validatePrimary` and `validateSecondary` in their respective `select` components using an `onBlur` method so the validation can only happen when the user is done selecting.

- Additionally, I used two `useEffect` hooks to validate the `primary` and `secondary` containers if their respective `hasInteracted` values are `true`. This ensures that validation only occurs if the user has interacted with the form fields. The `useEffect` hooks also include dependencies for the `validatePrimary` and `validateSecondary` functions to ensure they are re-evaluated when changes occur.

### Video for;

- Before change;

https://user-images.githubusercontent.com/105272189/230250383-edfdff01-9c1f-484f-b684-680fc58b7b67.mp4

- After change;

https://user-images.githubusercontent.com/105272189/230250447-d13315bf-8a42-4591-b217-42341731008d.mp4


Haywayaheadshot:
@yashika-khurana The other PR had commits from other merged PRs as a part of my changes because I synced changes after rebasing. I closed it so it does not become confusing. I hope that is ok.

PR #8669, opened 2023-04-06
For #8559 docs(nimbus): Update state and sequence diagrams to use is_dirty flag

eliserichards:
~🚨 do not merge! waiting on #8558~ done

Because

- We want to use the `is_dirty` flag instead of Dirty publish status

This commit

- Updates the state diagrams to remove the Dirty publish status
- Removes Dirty publish status from sequence diagrams
- Adds `is_dirty` flag to sequence diagrams

image
 
state-diagrams-v3


docs/experimenter/README.md
```diff
@@ -471,7 +470,7 @@ A live rollout can have updates pushed to its state while remaining Live. These
     rect rgb(255,204,255) 
         Note right of Experiment Owner: Owner makes changes in Experimenter
         Experiment Owner->>Experimenter UI: Make updates to live rollout
-        Experimenter UI->>Experimenter Backend: Status: Live <br/> Publish status: Dirty <br/> Status next: <none> <br/> + changelog
+        Experimenter UI->>Experimenter Backend: Status: Live <br/> Publish status: Idle <br/> Status next: <none> <br/> is_dirty: True <br/> + changelog
```
jaredlockhart, line 473: This shows the UI setting the `is_dirty` flag, but why would we change this from the serializer setting it?  We shouldn't need to involve the frontend in it at all?

jaredlockhart (dismissed):
God it's so nice to have these mermaid diagrams now 🎉 🎉 🎉

docs/experimenter/README.md
eliserichards: Good catch. This is one of the things that is changing with the flag vs. publishstatus. I'll update these..

yashikakhurana (commented):
Overall it looks good, my suggestion would be to appropriately show the updates we should do for `is_dirty`something like this, thoughts?
image

eliserichards:
Right now it is looking like this: 
image

Do you have a suggestion on making this more clear?

yashikakhurana:
oh perfect, I thought description one is the most recent one

yashikakhurana (approved):

eliserichards:
Oh sorry, I thought you meant the lifecycle state diagram! I misled you 😅

eliserichards:
Adding "Blocked" until #8558 is reviewed so we can merge them together 🚀

docs/experimenter/README.md
```diff
@@ -471,7 +470,9 @@ A live rollout can have updates pushed to its state while remaining Live. These
     rect rgb(255,204,255) 
         Note right of Experiment Owner: Owner makes changes in Experimenter
         Experiment Owner->>Experimenter UI: Make updates to live rollout
-        Experimenter UI->>Experimenter Backend: Status: Live <br/> Publish status: Dirty <br/> Status next: <none> <br/> + changelog
+        Note over Experimenter Backend: The experiment is updated and is <br/> marked as dirty on the backend
+        Experimenter UI->>Experimenter Backend: Status: Live <br/> Publish status: Idle <br/> Status next: <none>  <br/> + changelog
+        Experimenter Backend->>Experimenter Backend: is_dirty: True
```
jaredlockhart, line 475: Oh yeah that's a great way to represent it 🎉

docs/experimenter/README.md
```diff
@@ -518,12 +519,13 @@ A live rollout can have updates pushed to its state while remaining Live. These
         Note over Experimenter Worker: Worker updates rollout
         Experimenter Worker->>Remote Settings Backend: Check collection (timeout)
         Experimenter Worker->>Experimenter Backend:  Status: Live <br/> Publish status: Idle <br/> Status next: <none> <br/> + changelog
+        Experimenter Backend->>Experimenter Backend: is_dirty: False
```
jaredlockhart, line 522: No, there's no need to explicitly set the value here.  It gets set when the write happens above and then stays there until it is successfully resolved back to False after the update in RS completes.

jaredlockhart (dismissed):
Yes, the little loopty doo is the perfect way to show the backend setting the value.  You just don't need the second one in all the failure cases, take that out and this should be good to go.  Let's land this first and the code second, no need to land them 'together'?

docs/experimenter/README.md
eliserichards: We are showing `is_paused` on these diagrams, which is set in the same places as `is_dirty`. Should we remove `is_paused` from the diagrams as well? (I would prefer to keep both of them)

eliserichards: oh wait I see what you mean. Misread this ;)

PR #8671, opened 2023-04-06
For #8558 feat(nimbus): Add dirty flag to experiment inputs

eliserichards:
Because

- We want to use the new `is_rollout_dirty` flag instead of dirty publish status

This commit

- Adds `is_rollout_dirty` to `ExperimentInput` so that it can be used [here](https://github.com/mozilla/experimenter/blob/8d46516b281ef1c9e352634c6ea9113eca0712be/experimenter/experimenter/nimbus-ui/src/components/Summary/index.tsx#L70)

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/inputs.py
+++ b/experimenter/experimenter/experiments/api/v5/inputs.py
@@ -57,6 +57,7 @@ class ExperimentInput(graphene.InputObjectType):
     is_enrollment_paused = graphene.Boolean()
     is_first_run = graphene.Boolean()
     is_rollout = graphene.Boolean()
+    is_rollout_dirty = graphene.Boolean()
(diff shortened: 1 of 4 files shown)


yashikakhurana:
@eliserichards are you going to add logic in different PR?

jaredlockhart (dismissed):
Oh let's update one of the mutation tests to test writing to this field and storing it on the model 🙏

yashikakhurana:
> Oh let's update one of the mutation tests to test writing to this field and storing it on the model 🙏

yeah, I was hoping that one of the tests would fail if we are getting extra field

eliserichards:
I was curious why there weren't any explicit tests for the inputs 🤔 I just added a test to `test_mutations`

yashikakhurana (approved):
🌹

PR #8698, opened 2023-04-17
DRAFT ONLY - attempt at adding parallel runs to jobs.

b4handjr:
Will adding parallelism to jobs make them run faster?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,10 +56,6 @@ jobs:
           name: Run integration tests
           command: |
             cp .env.integration-tests .env
-            export PYTEST_OLD_ARGS=$PYTEST_ARGS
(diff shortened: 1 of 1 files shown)


jrbenny35:
What do you think? I'd like to implement it and see how it works for future testing additions, and it does improve our times a bit. @jaredlockhart

jaredlockhart:
I'm not sure this is the right approach.  The ones in main are taking ~13.5m and this one comes in at ~12.5m, so that's a lot of changes, particularly around pulling individual test names out and then computing their splits, only to save a minute.  Looking at it now I'd say table this and if things get worse again we can resurrect it but in the mean time the times are actually looking pretty good.  But thank you for investigating 🙏

PR #8715, opened 2023-04-19
For #8558 refactor(nimbus): Use isRolloutDirty flag instead of dirty publishStatus

eliserichards:
Because

- Using the dirty publish status to denote a dirty (edited live) rollout did not allow us to tell if a rollout was dirty when the publish status moved into Review/Approved/Waiting (this happens for request live update and ending the rollout)

This commit

- Changes everything over to use isRolloutDirty instead of publishStatus=Dirty
   - constants
   - v5 serializer
   - models: `can_edit` and `Filters`
   - kinto tasks
   - PageHome rendering the "Unpublished Updates" column (`DirectoryColumnUnpublishedUpdates`)
   - PageSummary and Summary components (approve/reject workflow setting publishStatus)
   - `getStatus`

Related: 
* https://github.com/mozilla/experimenter/pull/8560
* https://github.com/mozilla/experimenter/pull/8671
* https://github.com/mozilla/experimenter/pull/8669


jaredlockhart:
Let's block this on https://github.com/mozilla/experimenter/pull/8669

eliserichards:
#8669 is approved, so this is ready for review

experimenter/experimenter/experiments/models.py
```diff
@@ -1131,8 +1133,9 @@ class Filters:
             new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
         )
         IS_UPDATE_REVIEW_REQUEST = Q(
-            old_publish_status=NimbusExperiment.PublishStatus.DIRTY,
+            old_publish_status=NimbusExperiment.PublishStatus.IDLE,
             new_publish_status=NimbusExperiment.PublishStatus.REVIEW,
+            experiment_data__is_rollout_dirty=True,
```
jaredlockhart, line 1138: God I love that we can define these filters here like this.

experimenter/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -145,18 +144,18 @@ export const DirectoryColumnEndDate: ColumnComponent = ({
...
+    {d ? (
       <>
         <Badge
           className={
             "ml-2 border rounded-pill px-2 bg-white font-weight-normal border-danger text-danger"
           }
           data-testid="directory-unpublished-updates"
         >
-          {p === NimbusExperimentPublishStatusEnum.DIRTY ? "YES" : ""}
+          {d === true ? "YES" : ""}
```
jaredlockhart, line 158: You can just get rid of `=== true` altogether.

experimenter/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
```diff
@@ -145,18 +144,18 @@ export const DirectoryColumnEndDate: ColumnComponent = ({
 );
 
 export const DirectoryColumnUnpublishedUpdates: ColumnComponent = ({
-  publishStatus: p,
+  isRolloutDirty: d,
```
jaredlockhart, line 147: Why are these single letter variables at all?

experimenter/experimenter/nimbus-ui/src/components/PageHome/DirectoryTable/index.tsx
jaredlockhart: Also we're checking `d` above so it will always be `true` by the time we get down here?

experimenter/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -61,14 +61,7 @@ const Summary = ({ experiment, refetch }: SummaryProps) => {
...
     },
     {
-      publishStatus:
-        experiment.isRollout &&
-        status.live &&
-        experiment.statusNext ===
-          (NimbusExperimentStatusEnum.LIVE ||
-            NimbusExperimentStatusEnum.COMPLETE)
-          ? NimbusExperimentPublishStatusEnum.DIRTY
-          : NimbusExperimentPublishStatusEnum.IDLE,
```
jaredlockhart, line 71: God it's nice to get rid of this 🙏

experimenter/experimenter/nimbus-ui/src/lib/experiment.ts
```diff
@@ -163,10 +161,7 @@ export const resultsReadySortSelector: ExperimentSortSelector = (experiment) =>
 
 export const unpublishedUpdatesSortSelector: ExperimentSortSelector = (
   experiment,
-) =>
-  experiment.publishStatus === NimbusExperimentPublishStatusEnum.DIRTY
-    ? "1"
-    : "0";
+) => (experiment.isRolloutDirty === true ? "1" : "0");
```
jaredlockhart, line 164: `=== true` is redundant

jaredlockhart (changes requested):
Generally looking great 🎉 

Couple nits but also one thing I noticed, there's no reason for `isRolloutDirty` to be `null | false | true` we can just have it be `false | true` so let's update the field to be default false in the db, amend the migration, then we don't need to check for the `null` case at all.

jaredlockhart (approved):
Awesome.  Tested locally, everything seems ship shape.  Code all looks correct.  This is great thank you @eliserichards 🎉 🎉 🎉 

Also let's make sure the state diagram is up to date as well.  Can do that in a followup if it needs it.

PR #8734, opened 2023-04-21
feat(nimbus): check Remote Settings connectivity in heartbeat

leplatrem:
Because

- We want to discover Remote Settings connectivity issue before the background tasks fail

This commit

- Registers a new check
- Checks heartbeat
- Checks credentials

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@ PAD = -------------------------------------------------\n
 COLOR_CHECK = && echo "${GREEN}${PAD}All Checks Passed\n${PAD}${NOCOLOR}" || (echo "${RED}${PAD}Some Checks Failed\n${PAD}${NOCOLOR}";exit 1)
 PYTHON_TEST = pytest --cov --cov-report term-missing
 PYTHON_TYPECHECK = pyright experimenter/
-PYTHON_CHECK_MIGRATIONS = python manage.py makemigrations --check --dry-run --noinput
-PYTHON_MIGRATE = python manage.py migrate
+PYTHON_CHECK_MIGRATIONS = python manage.py makemigrations --check --dry-run --noinput --skip-checks
+PYTHON_MIGRATE = python manage.py migrate --skip-checks
 ESLINT_LEGACY = yarn workspace @experimenter/core lint
 ESLINT_FIX_CORE = yarn workspace @experimenter/core lint-fix
 ESLINT_NIMBUS_UI = yarn workspace @experimenter/nimbus-ui lint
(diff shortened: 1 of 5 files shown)


ahoneiser (approved):

Makefile
```diff
@@ -34,13 +34,13 @@ RUFF_CHECK = ruff experimenter/ tests/
...
-LOAD_LANGUAGES = python manage.py loaddata ./experimenter/base/fixtures/languages.json
-LOAD_FEATURES = python manage.py load_feature_configs
-LOAD_DUMMY_EXPERIMENTS = [[ -z $$SKIP_DUMMY ]] && python manage.py load_dummy_experiments || python manage.py load_dummy_projects
+CHECK_DOCS = python manage.py generate_docs --check=true --skip-checks
+GENERATE_DOCS = python manage.py generate_docs --skip-checks
+LOAD_COUNTRIES = python manage.py loaddata ./experimenter/base/fixtures/countries.json --skip-checks
+LOAD_LOCALES = python manage.py loaddata ./experimenter/base/fixtures/locales.json --skip-checks
+LOAD_LANGUAGES = python manage.py loaddata ./experimenter/base/fixtures/languages.json --skip-checks
+LOAD_FEATURES = python manage.py load_feature_configs --skip-checks
+LOAD_DUMMY_EXPERIMENTS = [[ -z $$SKIP_DUMMY ]] && python manage.py load_dummy_experiments --skip-checks || python manage.py load_dummy_projects --skip-checks
```
yashikakhurana, line 43: I am just wondering why `--skip-checks` flag?

yashikakhurana (commented):
I am curious to know about the addition of `--skip-checks` 🤔

yashikakhurana:
> > I am curious to know about the addition of --skip-checks 🤔
> 
> Unfortunately, the checks are executed implicitly on most `manage.py` commands, including migrations, loaddata, etc.
> 
> Since python-dockerflow uses this check framework for the `__heartbeat__` endpoint, we have to adapt :/ I could not find any other way to register the check only once the application has truely started to listen (via runserver or wsgi)

Got your point, my only concern skipping checks can lead to unexpected behavior or errors 👀, lets see what @jaredlockhart thoughts are on this

jaredlockhart (approved):
Awesome thanks so much @leplatrem 🙏 This will be very handy.

As far as skipping the checks, yes @leplatrem is right, all management commands invoke them and it's not reasonable to expect an rs instance is always availble whenever they're run.  Those commands should still fail/succeed for the specific things they're doing/checking for regardless of whether the checks are in place. cc @yashikakhurana 

I'll merge this 👍

yashikakhurana:
@jaredlockhart perfect, makes sense, thank you :)

PR #8740, opened 2023-04-21
fix #8738 docs(nimbus): ADR for jetstream schema definition

### LINKED ISSUE #8738: ADR for Jetstream - Nimbus shared schema
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3422)

mikewilli:
Because

- Jetstream analysis results do not have a well-defined schema
- Jetstream (export) and Experimenter (ingest) need to be in sync for the data sent to avoid errors/bugs
- the current approach for shared schemas (typescript schema in nimbus-shared) perpetuates fragmentation in Nimbus ecosystem (and is not widely used)

This commit

- documents the decisions for a) defining the schema, and b) where to put it


docs/adrs/0010-jetstream-shared-schema.md
```diff
@@ -0,0 +1,155 @@
...
+* Autogenerate bindings and publish to, e.g., pypi/npm/cargo for various use-cases
+
+#### Pros
+
+* Aligns with FastAPI approach used for ongoing Cirrus work
+* Easily generate schema definitions for different scenarios
+  * Django, Typescript, JSON
+
+#### Cons
+
```
yashikakhurana, line 87: Another con I can think is if the consuming applications are not using the same version

yashikakhurana (commented):
Thank you @mikewilli for putting this together, overall the pros and cons are well weighted, and clear diagram deserves a 🎖️

docs/adrs/0010-jetstream-shared-schema.md
```diff
@@ -0,0 +1,155 @@
...
+### Schema Definition
+
+* [**Option 1**] Pydantic
+* [**Option 2**] Typescript
+
+### Schema Location
+
+* [**Option A**] `jetstream` repository
+* [**Option B**] `experimenter` (Nimbus) repository
+* [**Option C**] `nimbus-shared` repository
```
scholtzan, line 36: Wasn't there some talk around having a `nimbus` repository at some point where `experimenter` and the new nimbus web effort will live? Would it make sense to have schemas in there as well (instead of moving them into experimenter)?

docs/adrs/0010-jetstream-shared-schema.md
```diff
@@ -0,0 +1,155 @@
...
+Technical Story: https://mozilla-hub.atlassian.net/browse/EXP-2265
+
+## Context and Problem Statement
+
+Experiment analysis for the Nimbus experimentation platform occurs in a separate project called Jetstream. Jetstream delivers results to Experimenter, which must parse these results based on the expected format of Jetstream’s output. However, Experimenter is forced to make assumptions about what it expects in the results, increasing the potential for buggy behavior during both ingestion and on the Results UI for Experimenter. Currently, various schemas are defined in a few different places for different purposes, but one schema definition shared between Experimenter and Jetstream would allow for validation of outputs from Jetstream while removing the need for Experimenter to make assumptions about the data it is receiving from Jetstream. This document describes the decisions around how and where that schema will be defined.
+
+
+## Decision Drivers
+
+* Single schema definition shared by Experimenter and Jetstream
```
scholtzan, line 16: Will we only have the jetstream result schema in there?

Jetstream also consumes data from the Experimenter API and would also benefit from having a schema that represents the API

mikewilli: Yea I think that is a natural follow-on to this. It should be fairly simple to use the same decisions made here and the resulting setup for related schemas, and that one should be even easier because it is the same components consuming it.

docs/adrs/0010-jetstream-shared-schema.md
mikewilli: Yes, I originally had a note about this but it appears that I accidentally dropped it in editing. But the important point is that the `experimenter` repo _is_ the (planned) `nimbus` repo, so your thought here is both valid and in line with the ADR.

docs/adrs/0010-jetstream-shared-schema.md
mikewilli: Oh, I see where I put it -- it's mentioned in the Pros section for Option B towards the bottom. Let me know if you think it's worth clarifying up front though and I can add a note.

jaredlockhart: @scholtzan that schema already exists, it's published in the [nimbus-shared](https://github.com/mozilla/nimbus-shared) repo and consumable via [pypi](https://pypi.org/project/mozilla-nimbus-shared/).  The Experimenter V6 API that jetstream consumes validates against this schema, [example here](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/tests/api/v6/test_serializers.py#L112)

I hope that after @mikewilli gets through this body of work we can use this as a foundation to move and consolidate all of the schemas for Nimbus into the Experimenter repo and publish them together into a Nimbus schemas package with bindings for Python, TypeScript, and Rust that can be consumed by all the parts of the Nimbus system.

jaredlockhart (approved):
Awesome thank you @mikewilli this all looks right to me.  Thank you for taking the time to explore all these options and write them out.  Approved let's get some schemas! 🎉

PR #8743, opened 2023-04-21
feat(nimbus): add targeting to exclude import infrequent rollouts

Mardak:
Because we want to exclude any user in either the original https://experimenter.services.mozilla.com/nimbus/import-infrequent-rollout-make-yourself-at-home or relaunched https://experimenter.services.mozilla.com/nimbus/updated-import-infrequent-rollout-make-yourself-at-home-copy so that a new experiment https://experimenter.services.mozilla.com/nimbus/new-import-interface-firefox-114-launch for a separate feature doesn't interact poorly (import spotlight modal could be stuck when some new wizard branches opens about:preferences in a tab)

This commit adds targeting to exclude (and unenroll -- i.e., don't want sticky) from the experiment anybody in the rollout. It's insufficient to add targeting that is opposite of the rollouts (i.e., `!(infrequent 5 bookmarks)`) because the rollouts have sticky enrollment.

This also updates the tests to handle "intersect" binary operator in the jexl tests to match https://searchfox.org/mozilla-central/rev/26790fecfcda622dab234b28859da721b80f3a35/toolkit/components/utils/FilterExpressions.sys.mjs#35,84-96

@dmose @jwayn

### DIFF
--- a/experimenter/experimenter/experiments/tests/__init__.py
+++ b/experimenter/experimenter/experiments/tests/__init__.py
@@ -1,8 +1,15 @@
 from pyjexl.jexl import JEXLConfig
-from pyjexl.operators import default_binary_operators, default_unary_operators
+from pyjexl.operators import Operator, default_binary_operators, default_unary_operators
 from pyjexl.parser import Parser, jexl_grammar
 
-jexl_config = JEXLConfig({}, default_unary_operators, default_binary_operators)
+jexl_config = JEXLConfig(
+    {},
+    default_unary_operators,
+    {
+        **default_binary_operators,
+        "intersect": Operator("intersect", 40, lambda x, y: set(x).intersection(y)),
+    },
+)
 
 
 class JEXLParser(Parser):
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1156,6 +1156,22 @@ TRR_MODE_ZERO = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+NOT_IMPORT_INFREQUENT_ROLLOUT = NimbusTargetingConfig(
+    name="Not in Import Infrequent Rollouts",
+    slug="not_import_infrequent_rollout",
+    description="Exclude users in the import infrequent rollouts",
+    targeting=(
+        "(activeRollouts intersect ["
+        "   'import-infrequent-rollout-make-yourself-at-home',"
+        "   'updated-import-infrequent-rollout-make-yourself-at-home-copy'"
+        "])|length == 0"
+    ),
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
 
 class TargetingConstants:
     TARGETING_VERSION = "version|versionCompare('{version}') >= 0"


experimenter/experimenter/targeting/constants.py
```diff
@@ -1156,6 +1156,20 @@ def __post_init__(self):
...
 )
 
+NOT_IMPORT_INFREQUENT_ROLLOUT = NimbusTargetingConfig(
+    name="Not in Import Infrequent Rollouts",
+    slug="not_import_infrequent_rollout",
+    description="Exclude users in the import infrequent rollouts",
+    targeting="""(activeRollouts intersect [
+            'import-infrequent-rollout-make-yourself-at-home',
+            'updated-import-infrequent-rollout-make-yourself-at-home-copy'
+        ])|length == 0""",
```
Mardak, line 1166: Eh… the resulting json has `\n` which is fine but prevents easy copy/paste into jexl evaluators. Switching to plain strings instead of multiline.

Better?
intersect jexl

jrbenny35:
> This also updates the tests to handle "intersect" binary operator in the jexl tests to match

Why is this needed?

jrbenny35:
> Various targeting has made use of it since 2017 for normandy and other checking of arrays, or are you suggesting we should find a different way to do this without the operator?

Oh no, purely curious, as seeing the code you referenced was done quite a while ago.

Mardak:
Ah yeah, I would think normandy primarily used `intersect` for checking if an add-on was installed or not. Messaging has also used it for matching a list of top sites:
https://searchfox.org/mozilla-central/rev/1f4f99a8f331cce8467a50742178b6d46914ab89/browser/components/newtab/lib/CFRMessageProvider.jsm#140-147

Some existing advanced targeting could probably be slightly simplified:
https://github.com/mozilla/experimenter/blob/294f4560439464d2714234aee653ec00742e3a39/experimenter/experimenter/targeting/constants.py#L836-L842

But generally if we're comparing / looking for a single exact match, `in` is good enough. In this case, we're checking for 2, so having a pair `… in … || … in …` would probably be fine too.

experimenter/experimenter/experiments/tests/__init__.py
```diff
@@ -1,8 +1,15 @@
...
+from pyjexl.operators import Operator, default_binary_operators, default_unary_operators
 from pyjexl.parser import Parser, jexl_grammar
 
-jexl_config = JEXLConfig({}, default_unary_operators, default_binary_operators)
+jexl_config = JEXLConfig(
+    {},
+    default_unary_operators,
+    {
+        **default_binary_operators,
+        "intersect": Operator("intersect", 40, lambda x, y: set(x).intersection(y)),
```
jaredlockhart, line 10: Oh interesting! I didn't know this was a supported operator, thanks for adding!

jaredlockhart (approved):
Looks good, thanks @Mardak 🎉

jaredlockhart:
Hrmmm looks like that one integration test doesn't want to pass here.  I'll dig in and take a look and see if I can figure out why it's failing.

Mardak:
Might be something intermittent? I believe it was `ci/circleci: Test Desktop Enrollment (Release Firefox)` and I noticed on previous pushes/rebases that it would pass and not pass. With this latest rebase, it's green so far.

PR #8745, opened 2023-04-24
fixes #8744 chore(Nimbus): Add HTML report generation and gathering for integration tests

### LINKED ISSUE #8744: Add HTML report to circleci integration tests
Diagnosing errors found on CI can be tricky without seeing what is happening, lets enable HTML reports so we can see screenshots of the failures.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3434)

jrbenny35:
Because:

* Diagnosing errors on CI can be hard without seeing what the page looks like

This Commit:

* Adds HTML reporting which will include screenshots for debugging errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,8 @@ jobs:
           command: |
             cp .env.integration-tests .env
             make refresh SKIP_DUMMY=1 PYTEST_SENTRY_DSN=$PYTEST_SENTRY_DSN CIRCLECI=$CIRCLECI up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
+      - store_artifacts:
+          path: /experimenter/tests/integration/test-reports/report.htm
 
   integration_nimbus_desktop_beta_targeting:
     machine:
@@ -84,6 +86,8 @@ jobs:
           command: |
             cp .env.integration-tests .env
             make refresh SKIP_DUMMY=1 PYTEST_SENTRY_DSN=$PYTEST_SENTRY_DSN CIRCLECI=$CIRCLECI up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
+      - store_artifacts:
+          path: /experimenter/tests/integration/test-reports/report.htm
 
   integration_nimbus_desktop_nightly_targeting:
     machine:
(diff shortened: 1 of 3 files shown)


yashikakhurana (commented):
This is a fantastic thank you @jrbenny35 ⭐ , question- we are setting the path for all the checks as `experimenter/experimenter/tests/integration/test-reports/report.htm`, do we want to have a different path, I know the same path won't make a difference as it is running on CI, maybe then its worth considering having to define this in a single place then, thoughts?

.circleci/config.yml
```diff
@@ -56,6 +56,8 @@ jobs:
           command: |
             cp .env.integration-tests .env
             make refresh SKIP_DUMMY=1 PYTEST_SENTRY_DSN=$PYTEST_SENTRY_DSN CIRCLECI=$CIRCLECI up_prod_detached integration_test_nimbus PYTEST_ARGS="$PYTEST_ARGS"
+      - store_artifacts:
+          path: ~/experimenter/experimenter/tests/integration/test-reports/report.htm
```
eliserichards, line 60: Should these be `.html` instead of `.htm`?

jrbenny35:
> This is a fantastic thank you @jrbenny35 ⭐ , question- we are setting the path for all the checks as `experimenter/experimenter/tests/integration/test-reports/report.htm`, do we want to have a different path, I know the same path won't make a difference as it is running on CI, maybe then its worth considering having to define this in a single place then, thoughts?

Yes good point, I need to update our config to use circleci's pipeline formatting, where we define one single job and just give it variables. I'll have to file an issue for that

yashikakhurana:
> > This is a fantastic thank you @jrbenny35 ⭐ , question- we are setting the path for all the checks as `experimenter/experimenter/tests/integration/test-reports/report.htm`, do we want to have a different path, I know the same path won't make a difference as it is running on CI, maybe then its worth considering having to define this in a single place then, thoughts?
> 
> Yes good point, I need to update our config to use circleci's pipeline formatting, where we define one single job and just give it variables. I'll have to file an issue for that

oh perfect, so that will be done in separate PR, right?

yashikakhurana (approved):
perfect 🎖️

.circleci/config.yml
jaredlockhart: Wow that's some old school stuff.  File type suffixes were generally capped at 3 characters on old operating systems like Windows 95/98, so `.htm` and `.html` were interchangeable for a long time.  Hasn't been that way in decades but wow that takes me back.

Issue #8716, opened 2023-04-19
Create release date field for mobile


data-sync-user:
Create a release date field on the experimenter backend. This should be a date, and it should only be available for mobile.

See notes here: [https://docs.google.com/document/d/1nC_bcAQuGdo8T2ck3v-6HdT9olFffpFoct63CPzbU-w/edit](https://docs.google.com/document/d/1nC_bcAQuGdo8T2ck3v-6HdT9olFffpFoct63CPzbU-w/edit|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3403)

PR #8749, opened 2023-04-25
For #8716 feat(nimbus): Add proposed release date to backend

eliserichards:
Because

- We want to add a release date field to make our calculations more clear

This commit

- Adds the field `proposed_release_date`

Screen Shot 2023-04-24 at 5 07 04 PM


yashikakhurana (approved):
perfect, field added, migrations are there, new field presented in a test case, flowing from backend to frontend with correct type 💯

Issue #8733, opened 2023-04-21
Set up api to get enrollment status


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3417)

data-sync-user:
➤ Yashika Khurana commented:

Define an enrollment_statusapi which will accept various parameters and will be useful in computing enrollment/unenrollment

PR #8750, opened 2023-04-26
fix #8731 feat(cirrus): Cirrus api structure

### LINKED ISSUE #8731: Create structure of api computation
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3419)

yashikakhurana:
Because

- Cirrus Api will perform multiple tasks like fetching recipes from remote settings, storing features value, and calling SDK to compute the enrollment, etc

This commit

- Creates the structure for the API `/enrollment_status`, including various dummy components like manifest loader, fml, SDK, and experiment_recipes to mimic the flow of the api
- Add new tests
Please Note that currently API doesn't take parameters into consideration, params will be added later in this ticket https://github.com/mozilla/experimenter/issues/8733

fixes #8731


yashikakhurana:
FYI you will also notice some `__init__.py` files in the directories to make them importable as packages

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Dict, List
```
jaredlockhart, line 1: Oh this is why I was saying to set the typing to the strictest, so we wouldn't ever have `Any` be allowed.  Force us to used `TypedDict` or Pydantic or whatever to explicitly lay out the structures of data.

cirrus/server/cirrus/main.py
```diff
@@ -1,8 +1,31 @@
...
+    recipes = rs.get_recipes()
+    # will recieve as part of incoming request
+    targeting_context = {"client_id": "testid"}
+    enrolled_partial_configuration = sdk.compute_enrollments(recipes, targeting_context)
+    feature_configurations = manifest_loader.get_latest_feature_manifest()
+    client_feature_configuration = fml.compute_feature_configurations(
+        enrolled_partial_configuration, feature_configurations
+    )
+
+    return client_feature_configuration
```
jaredlockhart, line 31: So awesome to see the structure starting to take shape 🤩 🤩 🤩

cirrus/server/cirrus/main.py
```diff
@@ -1,8 +1,31 @@
-from fastapi import FastAPI
+from fastapi import FastAPI, status
+
+from .experiment_recipes import RemoteSetting
+from .fml import FML
+from .manifest_loader import ManifestLoader
+from .sdk import SDK
```
jaredlockhart, line 6: Any way we can use absolute imports like `from cirrus.whatever import X`?

cirrus/server/tests/conftest.py
```diff
@@ -1,7 +1,7 @@
 from fastapi.testclient import TestClient
 from pytest import fixture
 
-from cirrus.main import app
+from ..cirrus.main import app
```
jaredlockhart, line 4: Why `..`?

jaredlockhart (approved):
Awesome to see this start to take shape, and stubbing out the basic parts.  The sky is looking cloudy and I'm here for it ☁️ ☁️ ☁️ ☁️ 

Also I think while this is still in its early phase we can be pretty loose about like, what's included in a given pr, stylistic choices, etc, better to just get the basics going and then worry about stricter things, so I'll to just give feedback as it occurs to me but not block anything.  Feel free to organize changes or followups as you like 🙏

cirrus/server/cirrus/manifest_loader.py
```diff
@@ -0,0 +1,14 @@
...
+    feature_manifest: List[Dict[str, Any]] = []
+
+    @classmethod
+    def get_latest_feature_manifest(cls) -> List[Dict[str, Any]]:
+        return cls.feature_manifest
+
+    @classmethod
+    def update_feature_manifest(cls,
+                                updated_feature_manifest: List[Dict[str, Any]]) -> None:
+        cls.feature_manifest = updated_feature_manifest
```
jrbenny35, line 14: ```get_latest_feature_manifest```, does this imply there will be an "old" feature manifest?
```update_feature_manifest``` , does this imply you are updating the "latest" feature manifest or something else?

**Review by jrbenny35 — DISMISSED** [review]
Looking good, did we consider using Dataclasses?

File: cirrus/server/cirrus/experiment_recipes.py (hunk 1 above)
Comment (yashikakhurana) [reply cirrus/server/cirrus/experiment_recipes.py:1]: so right now pyright is in strict mode, for more context- Pyright differentiates between explicit and implicit forms of Any. The implicit form is referred to as [Unknown](https://github.com/microsoft/pyright/blob/main/docs/type-inference.md#unknown-type). For example, if a parameter is annotated as list[Any], that is a use of an explicit Any, but if a parameter is annotated as list, that is an implicit Any, so pyright refers to this type as list[Unknown]. Pyright implements several checks that are enabled in “strict” type-checking modes that report the use of an Unknown type. Such uses can mask type errors.

Mypy does not track the difference between explicit and implicit Any types, but it supports various checks that report the use of values whose type is Any: --warn-return-any and --disallow-any-*. For details, refer to [this documentation](https://mypy.readthedocs.io/en/stable/command_line.html#disallow-dynamic-typing).

Pyright’s approach gives developers more control. It provides a way to be explicit about Any where that is the intent. When an Any is implicitly produced due to an missing type argument or some other condition that produces an Any within the type checker logic, the developer is alerted to that condition.
So I will file a separate ticket and look into more how to do that, but right now `Any` is just a place holder which will be replaced by data classes

File: cirrus/server/cirrus/main.py (hunk 3 above)
Comment (yashikakhurana) [reply cirrus/server/cirrus/main.py:6]: right, this directory/module is behaving weird right now and only liking relative import

File: cirrus/server/tests/conftest.py (hunk 4 above)
Comment (yashikakhurana) [reply cirrus/server/tests/conftest.py:4]: added some `__init__.py` to fix module issue

File: cirrus/server/cirrus/fml.py [hunk 6]
```diff
@@ -0,0 +1,10 @@
+from typing import Any, Dict, List
+
+
+class FML:
```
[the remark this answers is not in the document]
yashikakhurana: Good point

cirrus/server/cirrus/main.py
```diff
@@ -1,8 +1,31 @@
...
 
 app = FastAPI()
+rs = RemoteSetting()
+sdk = SDK()
+manifest_loader = ManifestLoader()
+fml = FML()
```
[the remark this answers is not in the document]
yashikakhurana: I would say hard to say we create an object of rust, but for better readability, I will rename this to remote_setting

cirrus/server/cirrus/manifest_loader.py
yashikakhurana: latest will be an assumption that we are trying to get latest manifest copy which will be available as part of the container, and yes update_feature_manifest means we will be updating our copy of the latest feature manifest

cirrus/server/cirrus/experiment_recipes.py
jaredlockhart: Thanks for the excellent explanation @yashikakhurana 🙏

yashikakhurana: also follow up ticket- https://github.com/mozilla/experimenter/issues/8762

yashikakhurana:
I am merging this PR, if @jrbenny35 anything comes up I can also address in any other PR too :)

cirrus/server/cirrus/generate_docs.py
```diff
@@ -3,8 +3,16 @@
...
+sys.path.append(cirrus_path)
+
+
 from fastapi.openapi.utils import get_openapi
-from main import app
+
+from cirrus.main import app
 
 DOCS_DIR = "cirrus/server/cirrus/docs"
 OPENAPI_PATH = os.path.join(DOCS_DIR, "openapi.json")
```
jrbenny35, line 18: We should use pathlib here.

cirrus/server/cirrus/sdk.py
```diff
@@ -0,0 +1,8 @@
+from typing import Any, Dict, List
+
+
+class SDK:
+
+    def compute_enrollments(self, recipes: List[Dict[str, Any]],
+                            targeting_context: Dict[str, str]) -> List[Dict[str, str]]:
+        return []
```
jrbenny35, line 8: This doesn't look formatted correctly

cirrus/server/cirrus/generate_docs.py
yashikakhurana: is there a benefit to this?

cirrus/server/cirrus/manifest_loader.py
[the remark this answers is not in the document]
yashikakhurana: oh it looks like but its formatted correctly, visually it looks like this because # Same as Black.
line-length = 90

PR #8752, opened 2023-04-26
For #8721 feat(nimbus): Use release date to calculate start date

eliserichards:
Because

- We added a new field, `proposed_release_date`
- We want to use this field to calculate the start date

This commit

- Uses the release date to calculate start date (if it exists). If not, it falls back to the original start date calculation
- Adds tests ✨


jaredlockhart (commented):
Just to summarize what we discussed, let's:

1) Leave startdate untouched
2) Serialize the proposed release date out verbatim
3) Add the proposed release date to the timeline between start date and enrollment end date

PR #8777, opened 2023-05-03
feat(nimbus): Update models for localized experiments

brennie:
Because

- we want to support localized experiments in Nimbus

this commit

- updates our models to support localizations.

Co-authored-by: Jason Prickett <jprickett@mozilla.com>


experimenter/experimenter/experiments/models.py
```diff
@@ -209,6 +209,9 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
     prevent_pref_conflicts = models.BooleanField(blank=True, null=True, default=False)
     proposed_release_date = models.DateField(blank=True, null=True)
 
+    is_localized = models.BooleanField(default=False)
+    localizations = models.TextField(blank=True, null=True)
```
yashikakhurana, line 213: we can use JSONField here, thoughts 🤔 ?

yashikakhurana (commented):
Generally look good, just wondering why not use JSONField in place of the text field so that we can take advantage of json validation, serialization, querying etc etc

experimenter/experimenter/experiments/models.py
jaredlockhart: Exactly correct @brennie but a good thought @yashikakhurana 👍   We used to use JSON field for this but we had to move away from it.

jaredlockhart (approved):
👍 looks good thnx @brennie 🎉

PR #8787, opened 2023-05-08
fix #6740 feat(nimbus): switch nimbus ui to multifeature fields

### LINKED ISSUE #6740: Use feature_configs in nimbus-ui

jaredlockhart:
TODO


brennie:
Closed in favour of #8860

PR #8797, opened 2023-05-10
fix #8732 feat(cirrus): Sync Remote setting to get collection of experiments

### LINKED ISSUE #8732: Periodically call remote settings collections active experiments
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3418)

yashikakhurana:
Because

- We want to fetch experiments from RS and keep a copy of those experiments in memory

This commit

- Uses [Apscheduler library](https://apscheduler.readthedocs.io/en/3.x/userguide.html) to sync [RS collection](https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/nimbus-web-experiments/records)
- uses  `.env` values to configure remote setting URL and refresh rate 
- provide sample env file as `.env.sample` file inside cirrus/server directory to configure/set value
- Retry failed job up to 3 times
- test cases to verify changes
- Rename API to `v1/features/` and post request
Note: The JSON response will be validated in a different ticket, the scope of this ticket is to get the ability to sync it time to time and get list in memory, also logger will be improved in another ticket


cirrus/server/cirrus/main.py
```diff
@@ -27,5 +44,39 @@ async def compute_enrollment_status():
...
+            seconds=remote_setting_refresh_rate_in_seconds,
+            max_instances=1,
+            max_retries=3,
+        )
+
+
+# Schedule the initial job to fetch recipes
+scheduler.add_job(  # type: ignore
+    fetch_schedule_recipes,
+    "interval",
```
jeddai, line 79: Something I think we need to discuss at some point is intervals vs cron, and managing memory state across the containers. 

If all containers' intervals aren't lined up, then some might have more recent/older recipes than others. Similarly with cron though, while I'd argue we _can_ trust that all containers will have the same system time, if containers go down and they fetch recipes when they come online, we could still end up in that state albeit everything will re-sync at the next cron match.

Maybe it's not a huge concern, I just want to note that I think it is at a minimum a valid concern we should provide options for mitigating. Maybe it's something we work out in the production-hardening phase.

yashikakhurana: yes I also thought about that, I think eventually they will get updated copy as they are planning to send request to cirrus with every backend call 🤔

jeddai: Are the recipes going to be fetched on every call to the enrollment endpoint?

The concern I was trying to share is if two different services ping two different cirrus pods as a part of a single request flow, each pod could have two different sets of recipes because the intervals could be triggering at different times.

yashikakhurana: they will be fetched in the background, so whatever current copy of recipe it will have in the memory will use that, but I think in single request flow it will only be hitting one cirrus pod per request, isn't it? I agree two pod may have different copy, it also depends how often we get new recipes

jeddai: I'm not clear on their plans for deployment.

It's definitely something we can do in the production hardening phase, I just wanted to go ahead and note it.

yashikakhurana: makes sense, I have noted down this we can circle back :)

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -1,13 +1,32 @@
...
+    def update_recipes(self, new_recipes: List[Dict[str, Any]]) -> None:
+        self.recipes = new_recipes
 
-    @classmethod
-    def update_recipes(cls, new_recipes: List[Dict[str, Any]]) -> None:
-        cls.recipes = new_recipes
+    def fetch_recipes(self):
+        try:
+            response = requests.get(self.url)
+            if response.status_code == 200:
```
jaredlockhart, line 23: Do you have to check this?

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -1,13 +1,32 @@
...
+                data = response.json().get("data", [])
```
jaredlockhart, line 24: What happens if it's invalid json?

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -1,13 +1,32 @@
...
+                if data:
```
jaredlockhart, line 25: Could do a little walrus here

```py
if data := response.json().get('data')
```

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -1,13 +1,32 @@
...
+                    self.update_recipes(data)
+                    logger.info("Fetched resources")
+                else:
+                    logger.warning("No recipes found in the response")
+        except requests.exceptions.RequestException as e:
+            logger.error(f"Failed to fetch recipes: {e}")
+            raise e
```
jaredlockhart, line 32: I would move this up to just enclose the `requests.get` since that's the only line that might throw this exception

cirrus/server/cirrus/main.py
```diff
@@ -1,24 +1,41 @@
...
 app = FastAPI()
 remote_setting = RemoteSettings()
 sdk = SDK()
 manifest_loader = ManifestLoader()
 fml = FML()
 
 
+remote_setting_refresh_rate_in_seconds: int = int(
+    config("REMOTE_SETTING_REFRESH_RATE_IN_SECONDS", default=10)  # type: ignore
+)
```
jaredlockhart, line 23: Settings are a little littered around here.  Might be a good idea to have a `settings.py` at the root that does all the `config` loading and then import from settings everywhere else rather than having them littered throughout the project?

cirrus/server/cirrus/main.py
```diff
@@ -1,24 +1,41 @@
...
+
+
+# Set up scheduler with ThreadPoolExecutor and configure job options
+scheduler: AsyncIOScheduler = AsyncIOScheduler(
+    job_defaults={"coalesce": False, "max_instances": 3, "misfire_grace_time": 60},
+)
```
jaredlockhart, line 29: Nnnnnnneat!

cirrus/server/cirrus/main.py
```diff
@@ -27,5 +44,39 @@ async def compute_enrollment_status():
...
+
+@app.on_event("startup")
+async def start_scheduler():
+    scheduler.start()
+    app.state.scheduler = scheduler
+
+
+@app.on_event("shutdown")
+async def shutdown_scheduler():
+    scheduler.shutdown()
```
jaredlockhart, line 58: Interesting!

cirrus/server/cirrus/main.py
jaredlockhart: Yeah that's a super interesting point @jeddai becuase if you have N copies of the container and their RS data are out of sync, they'll give back different possibly conflicting feature configs throughout the consuming applications architecture.  Suuuuuper interesting.  Let's think about that some more!

cirrus/server/cirrus/main.py
```diff
@@ -27,5 +44,39 @@ async def compute_enrollment_status():
...
+    except Exception as e:
+        # If an exception is raised, log the error and schedule a retry
+        logger.error(f"Failed to fetch recipes: {e}")
+        scheduler.add_job(  # type: ignore
+            fetch_schedule_recipes,
+            "interval",
+            seconds=remote_setting_refresh_rate_in_seconds,
+            max_instances=1,
+            max_retries=3,
+        )
```
jaredlockhart, line 73: Why do you need to schedule a retry like this here?  Especially using the globally defined refresh rate?  Isn't this redundant with the interval scheduler below?

cirrus/server/tests/test_experiment_recipes.py
```diff
@@ -1,13 +1,87 @@
...
+        rs = self.get_remote_settings()
+        assert rs.get_recipes() == []
+
+    def test_update_recipes(self):
+        rs = self.get_remote_settings()
+        new_recipes = [{"experiment1": True}, {"experiment2": False}]
+        rs.update_recipes(new_recipes)
+        assert rs.get_recipes() == new_recipes
+
+    @patch("requests.get")
```
jaredlockhart, line 23: Oh I don't think this is the right pattern.  This is mocking out the `requests` object as its imported in this file, and then if this file is parsed by the Python runtime before it parses the Remote settings file, then that instance of the module in memory will be reused.  But for instance, if you moved 

```py
from ..cirrus.experiment_recipes import RemoteSettings
```

up above

```py
import requests
```

then `RemoteSettings` module would get initialized first, it would import its local copy of requests, then you'd import yours, then you'd be mocking the wrong one?

Generally you're supposed to mock the path to the object in the file that it's being called from, ie `cirrus.experiment_recipes.requests`

cirrus/server/tests/test_experiment_recipes.py
```diff
@@ -1,13 +1,87 @@
...
+
+        rs = RemoteSettings()
+        try:
+            rs.fetch_recipes()
+        except requests.exceptions.RequestException as e:
+            # Assert that the exception is the expected exception
+            self.assertEqual(str(e), "Failed request")
+        else:
+            # If no exception was raised, fail the test case
+            self.fail("Expected RequestException but no exception was raised")
```
jaredlockhart, line 72: There's also [assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises)

jaredlockhart (changes requested):
This is starting to get super cool!  All this async scheduling, I love it! 😻 

Bunch of thoughts but things look great.  Does the SDK/FML stuff at the bottom need to be in here?  I feel like we're starting to get to a point where PRs can start to be more focused, if they're not required for the RS pieces maybe pull them out into separate PRs? 🤔

cirrus/server/cirrus/main.py
yashikakhurana: they can have same copy if we schedule all jobs in sync, like crontab

yashikakhurana:
> This is starting to get super cool! All this async scheduling, I love it! 😻
> 
> Bunch of thoughts but things look great. Does the SDK/FML stuff at the bottom need to be in here? I feel like we're starting to get to a point where PRs can start to be more focused, if they're not required for the RS pieces maybe pull them out into separate PRs? 🤔

awesome, yeah I thought of doing that, but thought to let go with this PR as it is essentially adding test, but I can take that out and put separate PRs for that

cirrus/server/cirrus/experiment_recipes.py
yashikakhurana: the validation part will come in separate PR, I planned to break it down, first landing the functionality as suppose nothing can break, and remaining part everything can break :D

cirrus/server/cirrus/main.py
yashikakhurana: so this max retries/ scheduler job is optional, the idea is let's say global rate is 1 hour and so if the job fails for some reason it will only try after 1 hour (the outside scheduler will do that), but having this job schedule (inside a function when exception is raised) will help us to retry, technically the refresh rate of this should be different than the global refresh rate and will only try failed job 3 times

cirrus/server/cirrus/main.py
jeddai: Yep and `appscheduler` can use cron instead of an interval, so I really think it should be a configurable thing. 

If we want to get into making sure this _can't_ happen then we'll likely need to talk about some service discoverability stuff if we want to keep it ephemeral. But again all this can totally be figured out in the hardening phase, just wanted to go on record with it now

cirrus/server/cirrus/experiment_recipes.py
yashikakhurana: just making sure status is 200, but can also do something like this 
```python
try:
    response = requests.get(url)
    response.raise_for_status()  # raises an HTTPError for bad status codes
    data = response.json().get("data", [])
    if data:
        # process the data
    else:
        print("No recipes found in the response")
except requests.exceptions.RequestException as e:
    print(f"Failed to fetch recipes: {e}")

File: cirrus/server/cirrus/main.py (hunk 6 above)
Comment (yashikakhurana) [reply cirrus/server/cirrus/main.py:23]: good point 👍

File: cirrus/server/tests/test_experiment_recipes.py (hunk 10 above)
Comment (yashikakhurana) [reply cirrus/server/tests/test_experiment_recipes.py:23]: makes sense 👍

**Comment by yashikakhurana** [comment]
Okay so the fml sdk, coverage and docker part is taken out in this PR https://github.com/mozilla/experimenter/pull/8813, this is ready for review now

File: cirrus/server/cirrus/experiment_recipes.py [hunk 12]
```diff
@@ -1,13 +1,32 @@
...
+    def update_recipes(self, new_recipes: List[Dict[str, Any]]) -> None:
+        self.recipes = new_recipes
 
-    @classmethod
-    def update_recipes(cls, new_recipes: List[Dict[str, Any]]) -> None:
-        cls.recipes = new_recipes
+    def fetch_recipes(self):
+        try:
+            response = requests.get(self.url)
+            response.raise_for_status()  # raises an HTTPError for bad status codes
```
jaredlockhart, line 24: Nice

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -1,13 +1,32 @@
...
+        except requests.exceptions.RequestException as e:
+            logger.error(f"Failed to fetch recipes: {e}")
+            raise e
+        if data := response.json().get("data", []):
```
jaredlockhart, line 28: 50first-dates-jocko-the-walrus

jaredlockhart (approved):
Yep all very nice looking good! 🎉

PR #8804, opened 2023-05-10
fix #8803 feat(nimbus): Don't serialize invalid localization json

### LINKED ISSUE #8803: Error: Expecting property name enclosed in double quotes: line 4 column 1 (char 25)
Sentry Issue: [EXPERIMENTER-STAGE-42](https://mozilla.sentry.io/issues/4172322508/?referrer=jira_integration)

```
Error: Expecting property name enclosed in double quotes: line 4 column 1 (char 25)
  at t (../../../src/errors/index.ts:60:5)
  at a.getQueryResult (../../../../src/react/data/QueryData.ts:366:17)
  at t.prototype.getExecuteResult (../../../../src/react/data/QueryData.ts:151:25)
  at t.prototype.execute (../../../../src/react/data/QueryData.ts:76:47)
  at result (../../../../src/react/hooks/utils/useBaseQuery.ts:65:55)
...
(45 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3467)

brennie:
Because

- the localizations field can contain invalid JSON

this commit

- only serializes the localizations field when it is valid JSON.

### DIFF
--- a/experimenter/experimenter/experiments/tests/api/v6/test_serializers.py
+++ b/experimenter/experimenter/experiments/tests/api/v6/test_serializers.py
@@ -432,3 +432,25 @@ class TestNimbusExperimentSerializer(TestCase):
             serializer.data["localizations"], json.loads(TEST_LOCALIZATIONS)
         )
         check_schema("experiments/NimbusExperiment", serializer.data)
+
+    @parameterized.expand(
+        [
+            ("invalid json", None),
+            (json.dumps({}), {}),
+        ]
+    )
+    def test_localized_localizations_json(self, l10n_json, expected):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE,
+            application=NimbusExperiment.Application.DESKTOP,
+            is_localized=True,
+            localizations=l10n_json,
+        )
+
+        serializer = NimbusExperimentSerializer(experiment)
+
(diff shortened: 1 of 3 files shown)


jaredlockhart (changes requested):
Oh interesting.  Very nice.  So it just falls back to returning `None`?  Also can we add a test for this? 🙏

jaredlockhart (changes requested):
Oh the test looks right except it's testing against v5 but the logic is in v6, i'd expect the test to be against the v6 serializer?

brennie:
The test was added in the v5 serializer because that is where the bug was occurring (becuase the v5 serializer uses the v6 serializer for `recipeJson`). The test fails without the fix.

yashikakhurana (approved):
Perfect 🚀

brennie:
Added an explicit test against the v6 serializer.

experimenter/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -432,3 +432,16 @@ def test_localized_desktop(self):
             serializer.data["localizations"], json.loads(TEST_LOCALIZATIONS)
         )
         check_schema("experiments/NimbusExperiment", serializer.data)
+
+    def test_localized_invalid_json(self):
```
yashikakhurana, line 436: do you want to parameterize as you did for the other test?

PR #8806, opened 2023-05-10
fix #8798 feat(nimbus): fetch remote schemas in circle task

### LINKED ISSUE #8798: Fetch feature schemas during circle task
Presently, we fetch remote feature manifests in an [hourly circle task](https://github.com/mozilla/experimenter/blob/main/.circleci/config.yml#L330-L368).  However those feature manifests can refer to remote schemas, which we then remotely fetch during the [feature load management command](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/features/__init__.py#L60-L68) that runs at container startup time.  

As the number of features that contain remote schemas has increased, we've seen intermittent deploy failures.  Similarly, we've seen that the Mozilla Central repo can be intermittently unavailable, which can cause remote fetches to its resources to fail.  We suspect that fetching the remote schemas during container startup time can cause container start to fail which then causes the deploy to fail.

To mitigate this, let's move fetching the remote schemas out of the feature loading management command and into the hourly circle task, so that both feature manifests and their dependent schemas are fetched by circle and automatically pr'ed into the repo.  That way when the container is being deployed and ready to start, all of the feature manifests and schemas should already be available on disk and no remote calls should need to be made.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3464)

jaredlockhart:
Because

* We have been prefetching feature manifests in a circle task and storing them in the repo
* Feature manifests can refer to remote feature schemas
* We have been remotely fetching those schemas at container startup time
* As the number of remote feature schemas increases, this can delay container startup and introduces the risk of remote schemas failing to fetch during container start
* We have been seeing intermittent deploy failures which may be caused by slow/failing remote schema fetches

This commit

* Fetches remote schemas as part of the circle ci task
* Stores remote schemas in the repo
* Loads remote schemas from disk at container startup time

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,8 @@ PYTHON_PATH_SDK = PYTHONPATH=/application-services/components/nimbus/src
 
 
 JETSTREAM_CONFIG_URL = https://github.com/mozilla/metric-hub/archive/main.zip
-FEATURE_MANIFEST_DESKTOP_URL = https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/nimbus/FeatureManifest.yaml
+MOZILLA_CENTRAL_ROOT = https://hg.mozilla.org/mozilla-central/raw-file/tip
+FEATURE_MANIFEST_DESKTOP_URL = ${MOZILLA_CENTRAL_ROOT}/toolkit/components/nimbus/FeatureManifest.yaml
 FEATURE_MANIFEST_FENIX_URL = https://raw.githubusercontent.com/mozilla-mobile/firefox-android/main/fenix/app/.experimenter.yaml
 FEATURE_MANIFEST_FXIOS_URL = https://raw.githubusercontent.com/mozilla-mobile/firefox-ios/main/.experimenter.yaml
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -78,6 +78,14 @@ feature_manifests:
...
 	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/focus-android.yaml $(FEATURE_MANIFEST_FOCUS_ANDROID)
 	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/focus-ios.yaml $(FEATURE_MANIFEST_FOCUS_IOS)
+	cat experimenter/experimenter/features/manifests/firefox-desktop.yaml | grep path: | \
+	awk -F'"' '{print "https://hg.mozilla.org/mozilla-central/raw-file/tip/" $$2}' | sort -u | \
+	while read -r url; do \
+		file=$$(echo $$url | sed 's|https://hg.mozilla.org/mozilla-central/raw-file/tip/||'); \
+		file="experimenter/experimenter/features/manifests/schemas/$$file"; \
+		mkdir -p $$(dirname $$file); \
+		curl $$url -o $$file; \
+	done
```
brennie: Maybe this is complicated enough to be its own shell script?

brennie (approved):

Issue #8758, opened 2023-05-01
Make dirty flag not optional


eliserichards:
Let's change the `is_rollout_dirty` flag to be a boolean, and remove the optional nullability

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3438)

jaredlockhart:
Fixed in https://github.com/mozilla/experimenter/pull/8809

PR #8809, opened 2023-05-11
For #8758 refactor(nimbus): Make dirty flag non-optional

eliserichards:
Because

- We don't want to have the `is_rollout_dirty` flag be optional
- It will just be `false` if it is a non-rollout
- And we don't need it in `inputs` anymore

This commit

- Makes `is_rollout_dirty` non-optional, default to false
- Update tests to look for bool instead of null
- Removes `is_rollout_dirty` from inputs (no longer necessary)
- Removes tests for setting `is_rollout_dirty` in mutations (no longer necessary)


jaredlockhart (approved):
Looks good thank you 🙏

eliserichards:
Ah! You beat me to it! I was about to send this to you for review ;)

PR #8819, opened 2023-05-12
fix #8364 feat(nimbus): Add a draft experiments API

### LINKED ISSUE #8364: Expose experiments that are in preview or draft mode through the API
This ask is related to providing a mechanism to generate previews for experiment analyses: https://github.com/mozilla/jetstream/blob/main/docs/proposal-0004-jetstream-preview.md

Experimenter only has an API that surfaces experiments that are currently live. Sometimes users would like to write and validate analysis configurations for experiments before they go live. To enable this, changes would need to be made to surface these experiments through the API.

I could see two possible options for this:
1. provide separate API endpoints for draft and preview experiments
2. add a `status` flag indicating whether an experiment is in preview, live, completed, ... and use the existing API endpoints

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3202)

brennie:
Because

- we need to expose draft experiments for several reasons (generating previews for jetstream, integration with automation for localized experiments, ...)

this commit

- adds a new API endpoint to return draft experiments, optionally filtered by the `is_localized` field.

### DIFF
--- a/docs/experimenter/openapi-schema.json
+++ b/docs/experimenter/openapi-schema.json
@@ -1537,6 +1537,82 @@
         ]
       }
     },
+    "/api/v6/draft-experiments/": {
+      "get": {
+        "operationId": "listNimbusExperiments",
+        "description": "",
+        "parameters": [
+          {
+            "name": "is_localized",
+            "required": false,
+            "in": "query",
+            "description": "is_localized",
+            "schema": {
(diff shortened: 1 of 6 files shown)


jaredlockhart (approved):
Awesome looks good 👍

PR #8824, opened 2023-05-12
For #8823 feat(nimbus): Add migration from Dirty to Idle

eliserichards:
Because

- We want to change existing experiments from publish status `Dirty` to Idle, and set the new `is_rollout_dirty` flag to `True` (#8558)

This commit

- Adds a migration to fix this :)
- Amends the test_migration test


jaredlockhart (approved):
Boom! 💥

jaredlockhart (dismissed):
ACK no lol, we also have to set

`is_dirty=True` XDDDDD

jaredlockhart (approved):
🙏 🙏 🙏

PR #8838, opened 2023-05-16
fix #8707 bug(project): check not running on main

### LINKED ISSUE #8707: Checks not running on main branch
We have some file path checks to skip tests when no changes are made in a given path, however looking at the recent main branch runs I see that even though there are changes in a given path, the relevant tests are still being skipped.

For instance here we have some changes to Cirrus:

https://app.circleci.com/pipelines/github/mozilla/experimenter/23089/workflows/2c56e77f-3f2a-4c81-9ef3-334e9692c477

but in Circle the Cirrus tests were skipped:

https://app.circleci.com/pipelines/github/mozilla/experimenter/23089/workflows/2c56e77f-3f2a-4c81-9ef3-334e9692c477/jobs/77499

Let's double check that the path checks are working correctly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3397)

jaredlockhart:
Because

* We run make check on each branch but also on main before deploys
* This last check is to make sure that main is still in a healthy state before it is deployed
* We added path filters to prevent certain circle tasks from running depending on which files were changed
* This path filter was preventing make check from running on main

This commit

* Splits the check job into two, one for branches that includes the path check, and one for main that excludes it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -537,11 +537,13 @@ workflows:
   build:
     jobs:
       - check_branch:
+          name: Check Experimenter (Branch)
           filters:
             branches:
               ignore:
                 - main
       - check_main:
+          name: Check Experimenter (Main)
           filters:
             branches:
               only: main
@@ -550,7 +552,7 @@ workflows:
             branches:
               only: main
           requires:
-            - check_main
+            - Check Experimenter (Main)
       - integration_nimbus_desktop_release_targeting:
           name: Test Desktop Targeting (Release Firefox)
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -548,6 +548,7 @@ workflows:
             branches:
               only: main
       - deploy:
+          name: Deploy Experimenter
           filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,6 +29,27 @@ jobs:
                 circleci-agent step halt
             fi
 
+  check_main:
+    machine:
+      image: ubuntu-2004:202107-02 # Ubuntu 20.04, Docker v20.10.7, Docker Compose v1.29.2
```
jeddai, line 34: I see the blame for the other images is from a few years ago, would it make sense to open a ticket to look at moving to 22.04 at some point? 20.04 is LTS so it's definitely not pressing, just something I thought about.

jeddai (approved):

.circleci/config.yml
jaredlockhart: Oh yeah good idea 👍 https://github.com/mozilla/experimenter/issues/8839

PR #8840, opened 2023-05-16
fix #8827 chore(ci): Enable reruns for integration tests on CI

### LINKED ISSUE #8827: Re-enable reruns for integration tests on CircleCI
We need to add the rerun flag to the `PYTEST_ARGS` environment variable in the circleci config.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3487)

jrbenny35:
Because:
* We disabled reruns locally and on CI

This commit:
* Re-enables reruns on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,7 @@ jobs:
     working_directory: ~/experimenter
     environment:
       FIREFOX_VERSION: nimbus-firefox-release
-      PYTEST_ARGS: -k FIREFOX_DESKTOP -m run_targeting -n 4
+      PYTEST_ARGS: -k FIREFOX_DESKTOP -m run_targeting -n 4 --reruns 1
       MOZ_REMOTE_SETTINGS_DEVTOOLS: 1 # allows us to override and set the remote settings URL
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


yashikakhurana:
@jrbenny35 general question- what's the benefit of rerun? Just trying to understand the benefit and use case to learn 🤔

jrbenny35:
> @jrbenny35 general question- what's the benefit of rerun? Just trying to understand the benefit and use case to learn 🤔

Just to help with intermittency. These tests are quite involved and sometimes weird things happen while they run. We only use 1 rerun so that we aren't relying on reruns, and with the pytest-sentry plugin, if a test is rerun and passes, it is marked as intermittent, so we can investigate.

It's really because these tests are so complicated and there are a lot of moving parts. Computers have no patience either 😅

yashikakhurana:
> > @jrbenny35 general question- what's the benefit of rerun? Just trying to understand the benefit and use case to learn 🤔
> 
> Just to help with intermittency. These tests are quite involved and sometimes weird things happen while they run. We only use 1 rerun so that we aren't relying on reruns, and with the pytest-sentry plugin, if a test is rerun and passes, it is marked as intermittent, so we can investigate.
> 
> It's really because these tests are so complicated and there are a lot of moving parts. Computers have no patience either 😅

Thanks @jrbenny35 , good to know about pytest-sentry plugin, this is awesome 🎉

jaredlockhart (approved):
👏 👏 👏  awesome ty 🙏

PR #8841, opened 2023-05-16
fix #8837 feat(nimbus): Do localization validation only when localized

### LINKED ISSUE #8837: Localizations validation occurring for non localized experiment
An experiment is failing the review serializer with a warning that there's an error on the Overview page, but nothing is showing.  Looking at the gql response we see:

image

however there should be no validation for this field since the experiment is not localized:

image

Enabling the checkbox does indeed show the error:

image

This validation shouldn't kick in unless the experiment is localized.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3491)

brennie:
Becuase:
- the localizations field is only considered in the case of localized experiments

this commit:
- only validates the localizations field for localized experiments.

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1316,26 +1316,6 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError("Hypothesis cannot be the default value.")
         return value
 
-    def validate_localizations(self, value):
-        if value is not None:
-            try:
-                localizations = json.loads(value)
-            except Exception as e:
-                raise serializers.ValidationError(f"Invalid JSON: {e}")
-
-            schema = settings.EXPERIMENT_SCHEMA["definitions"]["NimbusExperiment"][
-                "properties"
-            ]["localizations"]
-
-            try:
-                jsonschema.validate(localizations, schema)
-            except Exception as e:
-                raise serializers.ValidationError(
-                    f"Localization schema validation error: {e}"
-                )
-
-        return value
-
     def _validate_feature_value_against_schema(
         self, feature_config, value, localizations
     ):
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1652,39 +1631,51 @@ def _validate_localizations(self, data):
...
                 {"locales": ["Locales must not be empty for a localized experiment."]}
             )
 
-        experiment_locale_codes = [locale.code for locale in experiment_locales]
-        localizations = json.loads(localizations_json)
+        localizations_json = data.get("localizations")
+        try:
+            localizations = json.loads(localizations_json)
+        except Exception as e:
+            raise serializers.ValidationError(f"Invalid JSON: {e}")
```
jrbenny35, line 1644: Can you catch a ValidationError only? If not, I'd suggest using ```from e``` so we don't lose the context of the actual error.

jrbenny35 (commented):
Just had a thought.

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -2113,6 +2111,27 @@ def test_localized_value_missing_keys(self, l10n_obj, error_msg):
             {"reference_branch": {"feature_value": [error_msg]}},
         )
 
+    def test_not_localized_with_localizations(self):
```
eliserichards, line 2114: Wonderful! 🎉

eliserichards (approved):

yashikakhurana (approved):
localization validated ❗ 🚀

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
jaredlockhart: Excellent 💯

PR #8846, opened 2023-05-17
Fixes #8834 feat(nimbus): Compute fields based on release date or start date

### LINKED ISSUE #8834: Use release date to calculate durations
The new release date field should be used to calculate the following fields:

* Launch month
* Proposed enrollment end date
* Proposed end date
* Computed enrollment days
* Computed enrollment end date
* Enrollment duration
* Computed duration days

!image-20230419-185320.png|width=1350,height=184!

See notes here: [https://docs.google.com/document/d/1nC_bcAQuGdo8T2ck3v-6HdT9olFffpFoct63CPzbU-w/edit](https://docs.google.com/document/d/1nC_bcAQuGdo8T2ck3v-6HdT9olFffpFoct63CPzbU-w/edit|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3488)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/37240">image-20230419-185320.png</a>

eliserichards:
Because

- If an experiment is first run, we might have a release date
- And the following fields on the model are calculated using start date:
   - Launch month
   - Proposed enrollment end date
   - Proposed end date
   - Computed enrollment days
   - Computed enrollment end date
   - Enrollment duration
   - Computed duration days

This commit

- Add `release_date` field, which checks if the experiment is first run (`is_first_run`) before checking if `proposed_release_date` exists
- Add `enrollment_start_date` field, which returns `proposed_release_date` if it exists on a first run experiment and `start_date` otherwise
- DOES NOT change the `start_date` calculations
- `start_date` will still be used in all calculations where the experiment is not first-run and has no proposed release date

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -439,18 +439,19 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                 self.save()
                 return self._start_date
 
-    def has_valid_release_date(self):
-        if self.proposed_release_date is not None and self.is_first_run:
-            return True
-        return False
+    @property
+    def release_date(self):
+        if self.is_first_run:
+            return self.proposed_release_date
+
+    @property
+    def enrollment_start_date(self):
+        return self.release_date or self.start_date
 
     @property
     def launch_month(self):
-        if self.has_valid_release_date():
-            assert self.proposed_release_date
-            return self.proposed_release_date.strftime("%B")
-        elif self.start_date:
-            return self.start_date.strftime("%B")
+        if self.enrollment_start_date is not None:
+            return self.enrollment_start_date.strftime("%B")
 
     @property
     def end_date(self):
@@ -470,38 +471,26 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def proposed_enrollment_end_date(self):
-        if self.proposed_enrollment is not None:
-            if self.has_valid_release_date():
-                assert self.proposed_release_date
-                return self.proposed_release_date + datetime.timedelta(
-                    days=self.proposed_enrollment
-                )
-            elif self.start_date:
-                return self.start_date + datetime.timedelta(days=self.proposed_enrollment)
+        if (
+            self.proposed_enrollment is not None
+            and self.enrollment_start_date is not None
+        ):
+            return self.enrollment_start_date + datetime.timedelta(
+                days=self.proposed_enrollment
+            )
 
     @property
     def proposed_end_date(self):
-        if self.proposed_duration is not None:
-            if self.has_valid_release_date():
-                assert self.proposed_release_date
-                return self.proposed_release_date + datetime.timedelta(
-                    days=self.proposed_duration
-                )
-            elif self.start_date:
-                return self.start_date + datetime.timedelta(days=self.proposed_duration)
+        if self.proposed_duration is not None and self.enrollment_start_date is not None:
+            return self.enrollment_start_date + datetime.timedelta(
+                days=self.proposed_duration
+            )
 
     @property
     def computed_enrollment_days(self):
-        begin_date = None
-        if self.has_valid_release_date():
-            assert self.proposed_release_date
-            begin_date = self.proposed_release_date
-        elif self.start_date is not None:
-            begin_date = self.start_date
-
-        if self._enrollment_end_date is not None:
-            if begin_date:
-                return (self._enrollment_end_date - begin_date).days
+        enrollment_start = self.enrollment_start_date
+        if self._enrollment_end_date is not None and enrollment_start is not None:
+            return (self._enrollment_end_date - enrollment_start).days
 
         if self.is_paused:
             if paused_changelogs := [
@@ -519,8 +508,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                 paused_change = sorted(paused_changelogs, key=lambda c: c.changed_on)[-1]
                 self._enrollment_end_date = paused_change.changed_on.date()
                 self.save()
-                if begin_date:
-                    return (paused_change.changed_on.date() - begin_date).days
+                return (paused_change.changed_on.date() - enrollment_start).days
 
         if self.end_date:
             return self.computed_duration_days
@@ -529,16 +517,13 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def computed_enrollment_end_date(self):
-        start_date = self.start_date
-        computed_enrollment_days = self.computed_enrollment_days
-        if computed_enrollment_days is not None:
-            if self.has_valid_release_date():
-                assert self.proposed_release_date
-                return self.proposed_release_date + datetime.timedelta(
-                    days=computed_enrollment_days
-                )
-            elif start_date is not None:
-                return start_date + datetime.timedelta(days=computed_enrollment_days)
+        if (
+            self.computed_enrollment_days is not None
+            and self.enrollment_start_date is not None
+        ):
+            return self.enrollment_start_date + datetime.timedelta(
+                days=self.computed_enrollment_days
+            )
 
     @property
     def computed_end_date(self):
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -745,6 +776,7 @@ def clone(self, name, user, rollout_branch_slug=None):
         cloned.is_paused = False
         cloned.is_rollout_dirty = False
         cloned.reference_branch = None
+        cloned.proposed_release_date = None
```
yashikakhurana, line 779: good point, we always miss the clone part

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -823,16 +823,49 @@ def test_start_date_returns_None_for_not_started_experiment(self):
...
+    def test_launch_month_returns_release_date_month_for_started_first_run_experiment(
+        self,
+    ):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.LIVE_ENROLLING,
+            proposed_release_date=datetime.date.today() + datetime.timedelta(days=1),
+            is_first_run=True,
+        )
+
+        assert experiment.proposed_release_date
```
yashikakhurana, line 847: why this assert? 🤔

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -917,6 +950,22 @@ def test_end_date_uses_cached_end_date(self):
     def test_enrollment_duration_for_ended_experiment(self):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.LIVE_ENROLLING,
+            is_first_run=True,
+            proposed_release_date=datetime.date.today() + datetime.timedelta(days=1),
+        )
+
+        assert experiment.proposed_release_date
+        assert experiment.computed_end_date
```
yashikakhurana, line 958: don't we want to assert with something? 🤔

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -996,10 +1055,24 @@ def test_should_end_enrollment_returns_False_before_proposed_enrollment_end_date
...
+        )
+        self.assertTrue(experiment.should_end_enrollment)
+
+    def test_end_enrollment_returns_True_after_proposed_enrollment_end_date(
+        self,
+    ):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.LAUNCH_APPROVE_APPROVE,
             start_date=datetime.date.today() - datetime.timedelta(days=10),
+            proposed_release_date=None,
```
yashikakhurana, line 1075: do we need to pass `None`, I think by default it should be None

yashikakhurana (commented):
In general, looks good, the only thing to reduce if I can see is to take out this common line `if self.is_first_run and self.proposed_release_date:`to have in a separate method, thoughts?

experimenter/experimenter/experiments/tests/test_models.py
eliserichards: These tests are all copy/pasted from the existing tests to extend to testing `proposed_release_date` in addition to `start_date` :)

experimenter/experimenter/experiments/tests/test_models.py
eliserichards: same ^ I'm happy to take these out, I was just following what was previously there

eliserichards:
I added a helper function for checking `is_first_run` and `proposed_release_date`, but the problem is that the type checking still thinks it's possible for `proposed_release_date` to be `None` even if the function returns `True`. I tried a few things (like [TypeGuarding](https://peps.python.org/pep-0647/#typeguard-type) the function) but just landed on `assert`ing `proposed_release_date`. I'm not sure if this makes it better or worse lol. What do you think @yashikakhurana?

experimenter/experimenter/experiments/models.py
```diff
@@ -439,9 +439,15 @@ def start_date(self):
...
 
+    def has_valid_release_date(self):
+        return self.is_first_run and self.proposed_release_date
+
     @property
     def launch_month(self):
-        if self.start_date:
+        if self.has_valid_release_date():
+            assert self.proposed_release_date
+            return self.proposed_release_date.strftime("%B")
```
eliserichards, line 449: @yashikakhurana if I don't have the `assert` here, line 449 will error: 
```
error: "strftime" is not a known member of "None" (reportOptionalMemberAccess)
```

eliserichards:
Ok, let me know what you think @yashikakhurana. It still requires the `assert`s, but the helper function will never return `None` now

experimenter/experimenter/experiments/models.py
```diff
@@ -462,18 +470,38 @@ def end_date(self):
...
+            elif self.start_date:
+                return self.start_date + datetime.timedelta(days=self.proposed_enrollment)
 
     @property
     def proposed_end_date(self):
-        if self.start_date and self.proposed_duration is not None:
-            return self.start_date + datetime.timedelta(days=self.proposed_duration)
+        if self.proposed_duration is not None:
+            if self.has_valid_release_date():
+                assert self.proposed_release_date
```
yashikakhurana, line 486: same here

experimenter/experimenter/experiments/tests/test_models.py
yashikakhurana: I think we should assert with some value as expected value

yashikakhurana (dismissed):
In general looks good, you can remove assert from the non test cases code and also it would be helpful if we can assert with some value in test cases

experimenter/experimenter/experiments/models.py
```diff
@@ -462,18 +470,38 @@ def end_date(self):
 
     @property
     def proposed_enrollment_end_date(self):
-        if self.start_date and self.proposed_enrollment is not None:
-            return self.start_date + datetime.timedelta(days=self.proposed_enrollment)
+        if self.proposed_enrollment is not None:
+            if self.has_valid_release_date():
+                assert self.proposed_release_date
```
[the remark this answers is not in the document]
eliserichards: You can't remove the asserts (or you need to replace them with self.proposed_release_date != None). The code does not know that it is not None, even with the helper function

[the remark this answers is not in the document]
jaredlockhart:
```py
if release_date := self..... is not None:
    ....
```

experimenter/experimenter/experiments/models.py
```diff
@@ -439,9 +439,17 @@ def start_date(self):
                 self.save()
                 return self._start_date
 
+    def has_valid_release_date(self):
+        if self.proposed_release_date is not None and self.is_first_run:
+            return True
+        return False
```
jaredlockhart, line 445:
```py
return bool(...)
```

jaredlockhart:
```py
def release_date(self):
    if self.is_first_run:
        return self.proposed_release_date
```

jaredlockhart:
```py
def enrollment_start_date(self):
    return self.release_date or self.start_date
```

experimenter/experimenter/experiments/models.py
```diff
@@ -439,10 +439,19 @@ def start_date(self):
...
 
+    @property
+    def release_date(self):
+        if self.is_first_run:
+            return self.proposed_release_date
+
+    @property
+    def enrollment_start_date(self):
+        return self.release_date or self.start_date
+
```
eliserichards, line 450: These two fields replace the need for a helper function, and make it clear that:
1. `self.proposed_release_date` is dependent on `is_first_run`
2. The start date for enrollment will either be `release_date` or `start_date` in all places

brennie (approved):
LGTM but you may want to wait on review from someone more familiar with this than I am.

yashikakhurana (approved):
perfect 💯

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2004,6 +2071,9 @@ def _clone_experiment_and_assert_common_expectations(
         self.assertEqual(child.is_archived, False)
         self.assertEqual(child.is_paused, False)
         self.assertEqual(child.is_rollout_dirty, False)
+        self.assertEqual(child.proposed_release_date, None)
+        self.assertEqual(child.release_date, None)
+        self.assertEqual(child.enrollment_start_date, None)
```
jaredlockhart, line 2076: 💯 💯 💯 💯 💯 💯 💯

PR #8847, opened 2023-05-17
fix #8842: add jetstream/analysis schemas

### LINKED ISSUE #8842: Add Jetstream schema to Nimbus repo
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3494)

mikewilli:
Because

- We want to have a better data contract between Jetstream and Experimenter

This commit

- Creates schemas for the data products that Jetstream produces
- Adds a general python package structure with basic schema tests


jaredlockhart (commented):
Whoaaaa neeeaaattt it's beautiful 🥹

Okay so we should add some stuff like formatting/linting/tests etc, do you want to do some/any of that in this pr or just get this landed and hook more of that stuff up later?

Also I guess we should setup infrastructure to publish this to pypi and then consume it in experimenter/jetstream?  That we can definitely do as followups.

But this is exactly what I was hoping for it looks great 🎉 🎉 🎉

mikewilli:
Yea, so some of that I was considering for future PRs (publish to pypi, consumption in jetstream/experimenter).

For formatting/linting, I guess I need to ensure that our existing stuff runs against this new dir, and that makes sense to setup here/now.

For tests, do you want to see some tests specific to the schemas themselves, or would it be better to wait until we're using the schemas and then write tests of the functionality they enable alongside that?

jaredlockhart:
I think a minimal test of just having an example json that stresses the whole schema and running them through it would be nice here.  But since this is living at the top level you'll probably need to setup something around it to get pytest going, maybe just a poetry project?  Maybe wrap it in a Docker container to ensure there's a stable Python environment for it? (which is what Docker was actually invented for, funnily enough)

schemas/jetstream/metadata.py
```diff
@@ -0,0 +1,37 @@
...
+from pydantic import BaseModel
+
+from .statistics import AnalysisBasis
+
+
+class Metric(BaseModel):
+    analysis_bases: list(AnalysisBasis)
+    bigger_is_better: bool
+    description: str
+    friendly_name: str
```
scholtzan, line 11: I think these two attributes can be null

scholtzan (approved):
There are some attributes that looking in Jetstream are set to `Optional`, so I think they should be optional here too.

Other than that, looks great.

scholtzan (approved):

schemas/mozilla_nimbus_schemas/jetstream/analysis_errors.py
```diff
@@ -0,0 +1,25 @@
+import datetime as dt
+from typing import Optional
+
+from pydantic import BaseModel
+
+from .statistics import AnalysisBasis
```
brennie, line 6: The rest of experimenter uses full import paths instead of relative.

schemas/mozilla_nimbus_schemas/tests/jetstream/test_analysis_errors.py
```diff
@@ -0,0 +1,212 @@
...
+from pydantic import ValidationError
+
+from mozilla_nimbus_schemas.jetstream import (
+    AnalysisBasis,
+    AnalysisError,
+    AnalysisErrors,
+)
+
+
+class TestAnalysisErrors:
```
brennie, line 13: You're not subclassing `unittest.TestCase` so these don't need to be in a class.

schemas/mozilla_nimbus_schemas/tests/jetstream/test_metadata.py
```diff
@@ -0,0 +1,105 @@
+import datetime as dt
+import json
+
+import pytest
+from pydantic import ValidationError
+
+from mozilla_nimbus_schemas.jetstream import ExternalConfig, Metadata, Metric, Outcome
+
+
+class TestMetadata:
```
brennie, line 10: Same here re: classes

schemas/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+    "RET503",
+    "RET504",
+    "RET505",
+    "SIM102",
+]
+
+[tool.flake8]
+max-line-length = 90
+[tool.black]
+line-length = 90
```
brennie, line 50: Missing newline

Makefile
```diff
@@ -244,3 +244,14 @@ cirrus_typecheck_createstub: cirrus_build
...
+	(cd schemas && poetry run black --check --diff .)
+
+schemas_ruff_check:
+	(cd schemas && poetry run ruff .)
+
+schemas_check: schemas_black_check schemas_ruff_check
+	(cd schemas && poetry run pytest)
+
+schemas_code_format:
+	(cd schemas && poetry run black . && poetry run ruff --fix .)
```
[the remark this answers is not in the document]
mikewilli: Pretty sure I fixed this in the latest commit, but test this out again if you get a chance.

schemas/mozilla_nimbus_schemas/tests/jetstream/test_analysis_errors.py
```diff
@@ -0,0 +1,217 @@
...
+        metric="test-metric",
+        segment="test-segment",
+        statistic="test-statistic",
+        timestamp="2023-05-17T06:42:31+00:00",
+    )
+    assert ae.segment == "test-segment"
+    assert ae.metric == "test-metric"
+    assert ae.analysis_basis == "enrollments"
+    ae_json = ae.json()
+    print(ae_json)
```
jrbenny35, line 44: Do you need this print?

mikewilli: Nope, good catch. Thanks!

jrbenny35 (commented):
I think the majority of the tests here aren't really needed. Pydantic has great coverage for most of what you are testing. I think the json parts of the tests would be good to keep, as you did write a custom ```json_loads``` function.

mikewilli:
> I think the majority of the tests here aren't really needed. Pydantic has great coverage for most of what you are testing. I think the json parts of the tests would be good to keep, as you did write a custom `json_loads` function.

Yea, makes sense to me. I wrote some of them partly to test how it worked, but you're right! I'll take a sweep and see what I can remove.

jaredlockhart (approved):
Reviewed again and everything looks great 🎉 

So followup are we gonna publish this to pypi? 😱

PR #8852, opened 2023-05-17
feat(nimbus): Enforce constraints on localization string IDs

brennie:
Because:
- we want users to write usable, informative string IDs in their
  experiments

this commit:
- adds validation logic that requires l10n string IDs are at least 9
  characters long and only contain alphanumeric characters and hyphens.

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1,4 +1,5 @@
 import json
+import re
 import typing
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


brennie:
Blocked on #8852

jaredlockhart (approved):
Yep very nice, clear logic, clear tests 🙏 J'approve 🙆‍♀️

yashikakhurana (approved):
Looks good, the only suggestion I have if we want to move value `9` in some variable, otherwise 💯

PR #8860, opened 2023-05-18
fix #6740 feat(nimbus): switch nimbus ui to multifeature fields

### LINKED ISSUE #6740: Use feature_configs in nimbus-ui

brennie:
Because:
- we are moving to a multi-feature world

This commit:
- adds a new per-feature input (`FormFeatureValue`);
- updates the frontend to use the multi-featured API.

Co-authored-by: Barret Rennie <barret@brennie.ca>


experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -206,11 +206,11 @@ class Meta:
...
-    def resolve_feature_value(self, info):
-        return (
-            self.feature_values.exists()
-            and self.feature_values.all().order_by("feature_config__slug").first().value
-        ) or ""
+    # def resolve_feature_value(self, info):
+    #     return (
+    #         self.feature_values.exists()
+    #         and self.feature_values.all().order_by("feature_config__slug").first().value
+    #     ) or ""
```
jrbenny35, line 213: Some commented code that could be removed

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -186,15 +186,15 @@ class Meta:
...
+    # feature_value = graphene.String(required=False)
     feature_values = graphene.List(NimbusBranchFeatureValueType)
     screenshots = DjangoListField(NimbusBranchScreenshotType)
 
     class Meta:
         model = NimbusBranch
         fields = (
             "description",
-            "feature_value",
+            # "feature_value",
```
jrbenny35, line 197: Some commented code that could be removed

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -186,15 +186,15 @@ class Meta:
 
 class NimbusBranchType(DjangoObjectType):
     id = graphene.Int(required=False)
-    feature_value = graphene.String(required=False)
+    # feature_value = graphene.String(required=False)
```
jrbenny35, line 189: Some commented code that could be removed

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -442,7 +442,7 @@ class NimbusExperimentType(DjangoObjectType):
     countries = graphene.List(graphene.NonNull(NimbusCountryType), required=True)
     documentation_links = DjangoListField(NimbusDocumentationLinkType)
     enrollment_end_date = graphene.DateTime()
-    feature_config = graphene.Field(NimbusFeatureConfigType)
+    # feature_config = graphene.Field(NimbusFeatureConfigType)
```
jrbenny35, line 445: Some commented code that could be removed

experimenter/experimenter/experiments/tests/api/v5/test_mutations.py
```diff
@@ -1209,8 +1219,8 @@ def test_update_experiment_branches_with_feature_configs(self):
             "description": "a control",
             "ratio": 1,
             "featureValues": [
-                {"featureConfig": feature1.id, "value": ""},
-                {"featureConfig": feature2.id, "value": ""},
+                {"featureConfig": str(feature1.id), "value": ""},
+                {"featureConfig": str(feature2.id), "value": ""},
```
jrbenny35, line 1223: Why didn't you use f-strings here? Is ```feature1.id``` an integer?

jrbenny35 (changes requested):
Thanks for updating those tests, just had some other thoughts.

experimenter/experimenter/experiments/tests/api/v5/test_mutations.py
brennie: I didn't think to use f-strings here. `feature1.id` is an integer and calling str() seem simple enough.

brennie:
> Thanks for updating those tests, just had some other thoughts.

Thanks for the review! I thought I had removed all of Jared's WIP comments but I guess I missed a few 😅

experimenter/experimenter/experiments/tests/api/v5/test_mutations.py
brennie: It is only a test, and we're using str() elsewhere for the same thing in that file, so I think I'll leave it be.

jaredlockhart: @jrbenny35 that's interesting about fstrings being faster, but I don't think it'll make a big difference here.  Plus typically fstrings are used for combining some variables with string constants, but here we're just straight up converting a value to a str, and I still think of `str(...)` as the most idiomatic way to do that.

jaredlockhart (approved):
Okay.  Finally got to review this and test it locally.  Thanks for extra eyes @jrbenny35 🎉 And thank you @brennie for picking this up and getting it wrapped up.  Everything looks right, tested locally couldn't break it. A+ 10/10 let's get this landed!! 🎉 🎉 🎉 🎉

PR #8870, opened 2023-05-23
fix #8865 feat(nimbus): add version 113.0.1 to version list

### LINKED ISSUE #8865: Target 113.0.1+ for set Fx as default pdf handler experiments
Because we have multiple upcoming experiments that will be using the set Fx as default PDF handler special message action which had an integral fix in 113.0.1, we would like to be able to target this release as a minimum version.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3511)

jwayn:
Because

- We want to target minimum versions of 113.0.1 for future experiments

This commit

- Adds 113.0.1 to the version list


jaredlockhart (approved):
Very nice thank you!

PR #8892, opened 2023-05-29
bug(nimbus): Correctly escape nested quotes in make integration_test_nimbus

brennie:
Because

- the integration_test_nimbus target was encountering an error if
  the UPDATE_FIREFOX_VERSION environment variable was not set

this commit:

- fixes the Makefile integration_test_nimbus target to correctly
  escape nested quotes.


jaredlockhart (approved):
Oh good thnx @brennie 🙏

jrbenny35 (approved):
My shell knowledge is showing...

PR #8913, opened 2023-06-05
fixes #8912 feat(nimbus): add expanded targeting for early day users

### LINKED ISSUE #8912: Add advanced targeting to support early day users who need to set default or pin
In support of FXE-280 Fox doodle experiments, we need targeting for users who have profiles < 28 days old and have not set Firefox as default, as well as users with profiles < 28 days old who have set Firefox as default but have not yet pinned.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3545)

jwayn:
Because

- we need to be able to target users with profiles aged less than 28 days on the status of Firefox being pinned or set to default

This commit

- add targeting for early day users who have not set Firefox as default, early day users who have not yet pinned Firefox, and early day users who have set Firefox as default but have not yet pinned

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1234,6 +1234,39 @@ EARLY_DAY_USER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
thank you @jwayn

PR #8919, opened 2023-06-05
feat(nimbus): add targeting opposite of new profile created

Mardak:
Because we want to roll out the winning branch from https://experimenter.services.mozilla.com/nimbus/new-import-interface-firefox-114-launch which implicitly relied on another concurrent experiment with targeting from #8470 to prevent new users.

This commit adds a negation of the targeting to explicitly exclude recent profile creation.

not new targeting


yashikakhurana (approved):
thank you @Mardak

PR #8927, opened 2023-06-07
fixes #8394 feat(nimbus):change adds views and urls needed for the changelogs app

### LINKED ISSUE #8394: Define the URL routes and views for the Changelogs app
* Define the URL routes for accessing the app's views
* Create the views for rendering the app's templates
* Add any necessary context to the template rendering process

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3224)

Avi-88:
Because

- to implement the changelogs page, django views and urls will be required

This commit

- adds the url  `history/{experiment-slug}/changelogs` which will render the `NimbusChangelogsView` required for the history page


Avi-88:
I've just implemented this URl as first iteration, if it requires to be changed to a more suitable one please let me know, thanks

yashikakhurana (commented):
okay, as we were discussing yesterday, Coverage went down, that means you need to add test cases too, ps: I haven't reviewed your pr yet, just notice ci complaining about it

Avi-88:
> okay, as we were discussing yesterday, Coverage went down, that means you need to add test cases too, ps: I haven't reviewed your pr yet, just notice ci complaining about it

sorry about that, I just noticed the ci tests failing. I'll write the necessary tests and update the PR

experimenter/experimenter/changelog/templates/changelog-base.html
```diff
@@ -0,0 +1,14 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>History</title>
+</head>
+<body>
+    {% block content %}
+
+    {% endblock %}
+</body>
+</html>
```
brennie, line 14: Missing newline at end of file. You will probably want to configure your editor to automatically add them.

experimenter/experimenter/changelog/templates/changelog-base.html
```diff
@@ -0,0 +1,14 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>History</title>
+</head>
+<body>
+    {% block content %}
+
+    {% endblock %}
```
brennie, line 12: We don't need the whitespace inbetween the start and end of the block.

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,31 @@
...
+        slug = self.kwargs.get("slug")
+        experiment = get_object_or_404(NimbusExperiment, slug=slug)
+        return experiment
+
+    def get_context_data(self, **kwargs):
+        slug = self.kwargs.get("slug")
+        context = super().get_context_data(**kwargs)
+        experiment = self.object
+
+        # Retrieve the changelogs for the experiment
```
brennie, line 26: This comment is unnecessary

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,31 @@
...
+        changelogs = experiment.changes.all()
```
brennie, line 27: I would make this `changelogs = list(experiment.changes.all())` because `.all()` returns a QuerySet and iterating over it does a DB query. If the template ever ended up iterating over `changelogs` twice, it would do two DB queries.

brennie (changes requested):
Some minor nits, otherwise this looks great 👍

experimenter/experimenter/changelog/views.py
Avi-88: Oh, I didn't know that. I'll update the PR accordingly , thanks :)

experimenter/experimenter/changelog/templates/changelog/test.html
```diff
@@ -0,0 +1,11 @@
+{% extends "changelog-base.html" %}
```
yashikakhurana, line 1: regarding the file name `test.html`, we generally give test file name to those files which have some test code in it, since this is not a test template too, it is part of the `templates/changelog` you can give the appropriate name now, and keep building that page, we don't want to change name in the future

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,30 @@
...
+
+class NimbusChangeLogsView(ListView):
+    model = NimbusChangeLog
+    template_name = "changelog/test.html"
+    context_object_name = "experiment"
+
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        return super().get(request, *args, **kwargs)
+
```
yashikakhurana, line 15: here- The `get` method is unnecessary in this case since it only calls the parent class's get method with the same arguments. You can remove this method altogether, and the parent class's get method will be called automatically.

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,30 @@
...
+    def get_object(self, queryset=None):
+        slug = self.kwargs.get("slug")
+        experiment = get_object_or_404(NimbusExperiment, slug=slug)
+        return experiment
```
yashikakhurana, line 19: here- You can eliminate the need for the `get_object` method by using the` get_queryse`t method provided by the ListView. By overriding the get_queryset method, you can filter the objects based on the slug and experiment, like 
```python
def get_queryset(self):
    slug = self.kwargs.get("slug")
    return NimbusChangeLog.objects.filter(experiment__slug=slug)

Comment (yashikakhurana) [reply experimenter/experimenter/changelog/views.py:19]: and this also eliminates the need to manually fetch the NimbusExperiment object and simplifies the code.

File: experimenter/experimenter/urls.py [hunk 8]
```diff
@@ -23,6 +23,7 @@
     re_path(r"^api/v5/", include("experimenter.experiments.api.v5.urls")),
     re_path(r"^api/v6/", include("experimenter.experiments.api.v6.urls")),
     re_path(r"^admin/", admin.site.urls),
+    re_path(r"^history/", include("experimenter.changelog.urls")),
```
yashikakhurana, line 26: keep it consistent, somewhere we are saying history, somewhere we are saying changelogs, as we defined in URLs, would suggest changing history to changelog

yashikakhurana (changes requested):
thank you @Avi-88 overall you are right on track, just suggested some few suggestions to be consistent

experimenter/experimenter/changelog/views.py
Avi-88: I did this so that we can add the experiment data to the context as well because it will be needed when implementing the timeline for the experiment right ? Please let me know if there's a better way of achieving this

experimenter/experimenter/urls.py
yashikakhurana: you can also keep this :)

Avi-88:
> few suggestions
> 
> you could have
> 
> `/history/some-experiment-slug/` and can use detailview instead of listview on model NimbusExperiment and then in your template it's `{% for change in experiment.changes.all %},` also it will require some preprocessing on the changelogs before sending that to template, for that you can add an extra method to do it on experiment and use that method in template, for example
> 
> ```python
> def changes_with_extras(self):
>     return [{
>         'change': change,
>         'this': 'that',
>         ...
>     } for change in self.changes.all().order_by('-changed_on')]
> ```
> 
> and then in template file
> 
> ```python
> {% for change in experiment.changes_with_extras %}
>     {{ change.this }}
> {% endfor %}
> and feel free to keep `history` keyword
> ```

Oh, I just noticed this comment. Currently the url is `changelog/{slug}` should I update it again to be `history/{slug}` ?.

Avi-88:
> few suggestions
> 
> you could have
> 
> `/history/some-experiment-slug/` and can use detailview instead of listview on model NimbusExperiment and then in your template it's `{% for change in experiment.changes.all %},` also it will require some preprocessing on the changelogs before sending that to template, for that you can add an extra method to do it on experiment and use that method in template, for example
> 
> ```python
> def changes_with_extras(self):
>     return [{
>         'change': change,
>         'this': 'that',
>         ...
>     } for change in self.changes.all().order_by('-changed_on')]
> ```
> 
> and then in template file
> 
> ```python
> {% for change in experiment.changes_with_extras %}
>     {{ change.this }}
> {% endfor %}
> and feel free to keep `history` keyword
> ```

Will it work if I define a method that will perform the preprocessing in a helpers.py file and use it in the get_context

> few suggestions
> 
> you could have
> 
> `/history/some-experiment-slug/` and can use detailview instead of listview on model NimbusExperiment and then in your template it's `{% for change in experiment.changes.all %},` also it will require some preprocessing on the changelogs before sending that to template, for that you can add an extra method to do it on experiment and use that method in template, for example
> 
> ```python
> def changes_with_extras(self):
>     return [{
>         'change': change,
>         'this': 'that',
>         ...
>     } for change in self.changes.all().order_by('-changed_on')]
> ```
> 
> and then in template file
> 
> ```python
> {% for change in experiment.changes_with_extras %}
>     {{ change.this }}
> {% endfor %}
> and feel free to keep `history` keyword
> ```

I was thinking  about defining a method that will perform the preprocessing on the changelogs in a helpers.py file and use it in the `get_context_data` method itself and then add the transformed data into the context as `changelogs`. Please let me know your thoughts on this approach

yashikakhurana:
> > few suggestions
> > you could have
> > `/history/some-experiment-slug/` and can use detailview instead of listview on model NimbusExperiment and then in your template it's `{% for change in experiment.changes.all %},` also it will require some preprocessing on the changelogs before sending that to template, for that you can add an extra method to do it on experiment and use that method in template, for example
> > ```python
> > def changes_with_extras(self):
> >     return [{
> >         'change': change,
> >         'this': 'that',
> >         ...
> >     } for change in self.changes.all().order_by('-changed_on')]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then in template file
> > ```python
> > {% for change in experiment.changes_with_extras %}
> >     {{ change.this }}
> > {% endfor %}
> > and feel free to keep `history` keyword
> > ```
> 
> Will it work if I define a method that will perform the preprocessing in a helpers.py file and use it in the get_context
> 
> > few suggestions
> > you could have
> > `/history/some-experiment-slug/` and can use detailview instead of listview on model NimbusExperiment and then in your template it's `{% for change in experiment.changes.all %},` also it will require some preprocessing on the changelogs before sending that to template, for that you can add an extra method to do it on experiment and use that method in template, for example
> > ```python
> > def changes_with_extras(self):
> >     return [{
> >         'change': change,
> >         'this': 'that',
> >         ...
> >     } for change in self.changes.all().order_by('-changed_on')]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then in template file
> > ```python
> > {% for change in experiment.changes_with_extras %}
> >     {{ change.this }}
> > {% endfor %}
> > and feel free to keep `history` keyword
> > ```
> 
> I was thinking about defining a method that will perform the preprocessing on the changelogs in a helpers.py file and use it in the `get_context_data` method itself and then add the transformed data into the context as `changelogs`. Please let me know your thoughts on this approach

@Avi-88 for sure we can, when the context will be more, we will write a method, will suggest keeping the logic intact in the experimenter itself, and can define some methods over there, as we currently all that there

experimenter/experimenter/changelog/templates/changelog-base.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
yashikakhurana, line 1: here since it is already in changelog/templates/, you can omit `changelog` prefix, and just keep `base.html`

yashikakhurana: so technically you can add a` template` folder inside django root project and each subfolder for the app in your case `changelog`, I am not sure which base.html you are referring to, but I am referring to have this structure
```general 
project_name/
    manage.py
    project_name/
        __init__.py
        settings.py
        urls.py
        ...
    templates/
        base.html
        app1/
            template1.html
            template2.html
        app2/
            template3.html
            template4.html
    app1/
        __init__.py
        models.py
        views.py
        ...
    app2/
        __init__.py
        models.py
        views.py
        ...
    ...

**Review by yashikakhurana — CHANGES_REQUESTED** [review]
thank you for making those changes, left one comment regarding the template directory structure :)

**Review by brennie — CHANGES_REQUESTED** [review]

File: experimenter/experimenter/changelog/templates/changelog-base.html (hunk 9 above)
Comment (Avi-88) [reply experimenter/experimenter/changelog/templates/changelog-base.html]: So I have updated the templates to be according to the above file structure. The only thing is that I named the base template as `experimenter_base.html` as the template structure that was being followed by the legacy ui earlier is having conflicts with the new `base.html` as the files are collected altogether in the template engine. The order in which the template dirs are set helps in using the name `base.html` but if done so few tests fail as those refer to a different `base.html` // @yashikakhurana

**Review by yashikakhurana — APPROVED** [review]
thank you @Avi-88 looks good to me, nice job maintaining the directory structure and adding test cases 👏

File: experimenter/templates/changelog/overview.html [hunk 10]
```diff
@@ -0,0 +1,14 @@
+{% extends "experimenter_base.html" %}
+
+{% block title %}
+<title>History</title>
+{% endblock %}
+
+{% block content %}
+    <h1>This is the changes overview template</h1>
```
brennie, line 8: How about just "Changes Overview"

experimenter/templates/experimenter_base.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    {% block title %}
+    {% endblock %}
```
brennie, line 8: Instead of this, lets do:

```django
<title>{% block title %}{% endblock %}</title>
```

and then we don't have to repeat `<title>` every time we inherit from this template.

experimenter/experimenter/settings.py
```diff
@@ -124,6 +126,7 @@
             os.path.join(BASE_DIR, "legacy", "legacy-ui", "templates"),
             os.path.join(BASE_DIR, "nimbus-ui", "templates"),
             os.path.join(BASE_DIR, "docs"),
+            os.path.join(PROJECT_DIR, "templates"),
```
brennie, line 129: The template directory should be in BASE_DIR/templates following [Django best practices](https://docs.djangoproject.com/en/4.2/howto/overriding-templates/)

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,27 @@
...
+    context_object_name = "experiment"
+
+    def get_object(self, queryset=None):
+        slug = self.kwargs.get("slug")
+        experiment = get_object_or_404(NimbusExperiment, slug=slug)
+        return experiment
+
+    def get_context_data(self, **kwargs):
+        slug = self.kwargs.get("slug")
+        context = super().get_context_data(**kwargs)
```
brennie, line 19: super() call should be first in the method.

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,27 @@
...
+        experiment = self.get_object()
+
+        changelogs = list(experiment.changes.all())
+
+        context["slug"] = slug
```
brennie, line 24: Unnecessary. We have the experiment, so we can use `experiment.slug` in the template (and by default `kwargs` gets inserted into the context [here](https://github.com/django/django/blob/ddb6506618ea52c6b20e97eefad03ed847a1e3de/django/views/generic/detail.py#L100)

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,27 @@
...
+from django.shortcuts import get_object_or_404
+from django.views.generic import DetailView
+
+from experimenter.experiments.models import NimbusExperiment
+
+
+class NimbusChangeLogsView(DetailView):
+    model = NimbusExperiment
+    template_name = "changelog/overview.html"
+    context_object_name = "experiment"
```
brennie, line 10: This should be unnecessary:

> Well, if you’re dealing with a model object, this is already done for you. When you are dealing with an object or queryset, Django is able to populate the context using the lowercased version of the model class’ name.

[ref](https://docs.djangoproject.com/en/4.2/topics/class-based-views/generic-display/#making-friendly-template-contexts)

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,27 @@
...
+
+    def get_object(self, queryset=None):
+        slug = self.kwargs.get("slug")
+        experiment = get_object_or_404(NimbusExperiment, slug=slug)
+        return experiment
+
+    def get_context_data(self, **kwargs):
+        slug = self.kwargs.get("slug")
+        context = super().get_context_data(**kwargs)
+        experiment = self.get_object()
```
brennie, line 20: This is unnecessary. The view will populate the context with `[context_object_name] = get_object()` [src](https://github.com/django/django/blob/ddb6506618ea52c6b20e97eefad03ed847a1e3de/django/views/generic/detail.py#L99)

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,27 @@
...
+
+class NimbusChangeLogsView(DetailView):
+    model = NimbusExperiment
+    template_name = "changelog/overview.html"
+    context_object_name = "experiment"
+
+    def get_object(self, queryset=None):
+        slug = self.kwargs.get("slug")
+        experiment = get_object_or_404(NimbusExperiment, slug=slug)
+        return experiment
```
brennie, line 15: This method is unnecessary. By default, `get_object()` will grab the `slug` from `self.kwargs` and look up the object that way.

[src](https://github.com/django/django/blob/ddb6506618ea52c6b20e97eefad03ed847a1e3de/django/views/generic/detail.py#L40)

[docs](https://docs.djangoproject.com/en/4.2/ref/class-based-views/mixins-single-object/#django.views.generic.detail.SingleObjectMixin.get_object)

brennie (changes requested):

jaredlockhart (commented):
@brennie Already covered everything with very helpful links to the docs, I just wanted to reinforce that we shouldn't need to declare anything on the view class other than the model and template name.  ChangeLogs can be looked up from the experiment instance right in the template, either by accessing the related manager directly as `experiment.changes.all` or by adding a method to the NimbusExperiment model to do any additional pre-processing.

Aside from that, this is all looking very good so far 🎉   Thank you @Avi-88 ! 🎉

experimenter/experimenter/settings.py
Avi-88: We have the other apps inside the experimenter directory contrary to what the default django project structure is where they are present at  the root, so the `BASE_DIR` refers to that path. I followed this structure which was suggested in an earlier feedback which will refer to the templates folder present at the root of the project like so. Should I move it back inside the experimenter directory ? // @brennie 
```project_name/
    manage.py
    project_name/
        __init__.py
        settings.py
        urls.py
        ...
    templates/
        base.html
        app1/
            template1.html
            template2.html
        app2/
            template3.html
            template4.html
    app1/
        __init__.py
        models.py
        views.py
        ...
    app2/
        __init__.py
        models.py
        views.py
        ...
    ...```

File: experimenter/experimenter/changelog/views.py (hunk 16 above)
Comment (Avi-88) [reply experimenter/experimenter/changelog/views.py:20]: So should I remove the `get_object` method from the view and have just the `get_context_data` method like this 
```
    context = super().get_context_data(**kwargs)
    context['changelogs'] = list(self.object.changes.all())
    return context
```
or just remove all the methods altogether and access the changelogs directly like 
```
for changelog in nimbusexperiment.changes.all
```
But I guess the latter will have issues as I'm directly iterating a queryset right ? // @brennie

**Comment by Avi-88** [comment]
> 

Thank you, I'm still trying to grasp how data is handled in django so there's too much unnecessary code , sorry about that. Also @brennie mentioned in a previous feedback that the `all()` method returns a queryset and iterating over it make a DB query which may have performance implications. So will updating the get_context_data method to be just this work ? // @jaredlockhart 
```
        context = super().get_context_data(**kwargs)
        context['changelogs'] = list(self.object.changes.all())
        return context
```

**Comment by brennie** [comment]
> > 
> 
> Thank you, I'm still trying to grasp how data is handled in django so there's too much unnecessary code , sorry about that. Also @brennie mentioned in a previous feedback that the `all()` method returns a queryset and iterating over it make a DB query which may have performance implications. So will updating the get_context_data method to be just this work ? // @jaredlockhart
> 
> ```
>     def get_context_data(self, **kwargs):
>         context = super().get_context_data(**kwargs)
>         context['changelogs'] = list(self.object.changes.all())
>         return context
> ```

Yep, that would be fine!

File: experimenter/experimenter/changelog/views.py (hunk 16 above)
Comment (brennie) [reply experimenter/experimenter/changelog/views.py:20]: Pre-fetching changelogs in get_context_data is fine

File: experimenter/experimenter/settings.py (hunk 12 above)
Comment (brennie) [reply experimenter/experimenter/settings.py:129]: `templates` should be located at `experimenter/experimenter/`. The first directory is just to separate the project from other projects. The second directory is the Python package directory.

**Comment by yashikakhurana** [comment]
> > > 
> > 
> > 
> > Thank you, I'm still trying to grasp how data is handled in django so there's too much unnecessary code , sorry about that. Also @brennie mentioned in a previous feedback that the `all()` method returns a queryset and iterating over it make a DB query which may have performance implications. So will updating the get_context_data method to be just this work ? // @jaredlockhart
> > ```
> >     def get_context_data(self, **kwargs):
> >         context = super().get_context_data(**kwargs)
> >         context['changelogs'] = list(self.object.changes.all())
> >         return context
> > ```
> 
> Yep, that would be fine!

@Avi-88 you can also add a method in experimenter like this 
```python
def get_changelogs(self):
        return list(self.changes.all())
```
by having this whatever processing we need to do in the future we can do here and then our view will just have template and model name

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,13 @@
...
+
+
+class NimbusChangeLogsView(DetailView):
+    model = NimbusExperiment
+    template_name = "changelog/overview.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["changelogs"] = list(self.object.changes.all())
+        return context
```
yashikakhurana, line 13: I would suggest creating a method for the experiment and removing `get_context_data`

yashikakhurana (changes requested):

experimenter/experimenter/changelog/views.py
```diff
@@ -1 +1,8 @@
-# Create your views here.
+from django.views.generic import DetailView
+
+from experimenter.experiments.models import NimbusExperiment
+
+
+class NimbusChangeLogsView(DetailView):
+    model = NimbusExperiment
+    template_name = "changelog/overview.html"
```
yashikakhurana, line 8: 🎉

yashikakhurana (approved):
okay everything looks good, view is perfect, method moved to model 🚀  and test coverage 💯 , thank you @Avi-88

experimenter/experimenter/experiments/models.py
```diff
@@ -812,6 +812,9 @@ def clone(self, name, user, rollout_branch_slug=None):
 
         return cloned
 
+    def get_changelogs(self):
+        return list(self.changes.all())    
+
```
brennie, line 817: This doesn't provide any benefit since we're not pulling it into a variable.

experimenter/experimenter/templates/changelog/overview.html
```diff
@@ -0,0 +1,17 @@
+{% extends "experimenter_base.html" %}
+
+{% block title %}
+  History
+{% endblock %}
+
+{% block content %}
+    <h1>Changes Overview</h1>
+    <p>Experiment Slug: {{ object.slug }}</p>
+    {% for changelog in object.get_changelogs %}
```
brennie, line 10: You can use `experiment` here instead of `object`

brennie, line 10: Just use `experiment.changelogs.all` -- the list optimization I suggested was premature.

brennie (changes requested):

experimenter/experimenter/experiments/models.py
yashikakhurana: oh yes this is just for now temporarily, to make sure we have a function that returns something related to changelogs, in the future PR he will do processing and everything

experimenter/experimenter/templates/changelog/overview.html
Avi-88: The model is named as NimbusExperiment and I'll need to use this in lowercase like this `nimbusexperiment` which doesn't look good or else give it a custom name like I did earlier that's why I just used the generic object here

Avi-88: I assume we won't remove the  `get_changelogs` method inside the model as it will be required for pre processing, I'll just return the changelogs directly for now without using the `list` method

brennie (approved):
Sorry about all the rounds of feedback but LGTM! Let's get this landed as soon as tests are passing!

r+

Avi-88:
> Sorry about all the rounds of feedback but LGTM! Let's get this landed as soon as tests are passing!
> 
> r+

Oh, I completely forgot to update the test after the recent changes, sorry about that I'll update it.

yashikakhurana (approved):

jaredlockhart:
Thank you @Avi-88 for working through this.  This is very nice and succinct, and excellent foundation to start building some UI on top of.  Great job! 🎉 🎉 🎉   And thank you @yashikakhurana and @brennie for working through all the feedback on this 🙏

PR #8937, opened 2023-06-09
fix #8874 feat(cirrus): Integrate nimbus fml ✍️ 

### LINKED ISSUE #8874: Integrate nimbus fml
Nimbus fml can be used to get the feature configuration, given fml and nimbus SDK configurations

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3517)

yashikakhurana:
Because

- To calculate the feature config based on the provided `enrolled partial configuration`, we need SDK to help us to evaluate `feature configurations` that can be used by the client to set features value

This commit

- Integrate `nimbus_fml` library
- Used to calculate the feature configurations
- test cases to support the changes


jaredlockhart (changes requested):
Absolutely incredible @yashikakhurana it's so good 🎉 🎉 🎉 🎉 🎉 

Let's make sure to check the case where a broken path is passed in and it can't load.

yashikakhurana:
> Absolutely incredible @yashikakhurana it's so good 🎉 🎉 🎉 🎉 🎉
> 
> Let's make sure to check the case where a broken path is passed in and it can't load.

yes yes good call

jaredlockhart:
And also worth checking that it's valid YAML

jeddai:
FWIW the FML client validates path+yaml and has specific errors returned for each, they're the same Python error class though — definitely worth having tests around error handling for constructing the FML client.

cirrus/server/cirrus/feature_manifest.py
```diff
@@ -1,13 +1,42 @@
...
+    ) -> Dict[str, Any]:
+        feature_configs = {
+            key: value["feature"]["value"]
+            for key, value in enrolled_partial_configuration[
+                "enrolledFeatureConfigMap"  # slug, featureid, value,
+            ].items()
+        }
+        merged_res: Dict[str, Any] = self.fml_client.merge(  # type: ignore
+            feature_configs
+        )
```
brennie, line 29: `merged_res` is not a Dict according to the [UDL](https://github.com/mozilla/application-services/blob/b8d3b225ee6e7069ff4e2c14e8099d9514c9f034/components/support/nimbus-fml/src/fml.udl#L11)

brennie (changes requested):

cirrus/server/cirrus/feature_manifest.py
jeddai: Yep, you should be able to import a `MergedJsonWithErrors` classref from `fml_sdk`

cirrus/server/cirrus/settings.py
```diff
@@ -19,3 +19,4 @@
         "channel": channel,
     }
 )
+fml_path: str = cast(str, config("CIRRUS_FML_PATH", default=""))
```
jaredlockhart, line 22: You could mock this in the test to override it in Python space to a test fixture path like here:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/features/tests/__init__.py#L6-L10

cirrus/server/tests/test_feature_manifest.py
```diff
@@ -1,23 +1,185 @@
...
+            "events": [
+                {
+                    "branch_slug": "treatment",
+                    "change": "Enrollment",
+                    "enrollment_id": "enrollment_id",
+                    "experiment_slug": "experiment_slug",
+                    "reason": None,
+                }
+            ],
+        }
```
jaredlockhart, line 75: We could use [factories](https://factoryboy.readthedocs.io/en/stable/) to generate these so we don't need to worry about their structure and values in every test.

cirrus/server/cirrus/feature_manifest.py
```diff
@@ -1,13 +1,42 @@
...
+
+from .settings import channel, fml_path
+
+logger = logging.getLogger(__name__)
+
 
 class FeatureManifestLanguage:
+    def __init__(self):
+        self.fml_client = FmlClient(fml_path, channel)
+        self.merge_errors: List[FmlError] = []
```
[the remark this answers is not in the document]
yashikakhurana: this is specifically for the testing since we return the attribute .json of the merge request response from nimbus_fml, I wasn't able to write a test, I could technically write the same test you wrote, but wanted to capture the error as part of fml in the API, I am happy to explain more

cirrus/server/cirrus/feature_manifest.py
brennie: You should be able to leave it without a type annotation and it should infer the type from the return of self.fml_client.merge

cirrus/server/cirrus/feature_manifest.py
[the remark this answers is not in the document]
jeddai: Fair enough

cirrus/server/cirrus/feature_manifest.py
yashikakhurana: it doesn't infer that, `strict` mode needs to be defined with the type

yashikakhurana: good call @jeddai and @brennie

yashikakhurana: changed with the correct type

PR #8939, opened 2023-06-09
fix #8848 feat(nimbus): add pypi publishing of schemas package

### LINKED ISSUE #8848: Publish Jetstream schemas
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3500)

mikewilli:
Because

- we want to use the new schemas in various places

This commit

- sets up automatic publishing to PyPI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -519,12 +519,38 @@ jobs:
             if git diff --name-only main HEAD | grep -E 'schemas/'
               then
                 echo "Changes detected in /schemas directory. Running Schemas tests and linting."
-                env
                 make schemas_check
               else
                 echo "No changes in /schemas directory. Skipping tests and linting."
                 circleci-agent step halt
             fi
+  schemas_deploy:
+    docker:
+      - image: cimg/python:3.10.8 # poetry 1.2.2
+    steps:
+      - checkout
+      - run:
+          name: Check for changes in schemas/ in last commit before proceeding.
+          command: |
+            if git diff --name-only main HEAD~1 | grep -E 'schemas/'
+              then
+                echo "Found changes to schemas dir, proceeding with deployment."
+              else
+                echo "No changes in /schemas directory. Skipping mozilla-nimbus-schemas deployment."
+                circleci-agent step halt
+            fi
+      - run:
+          name: Create the distribution files
+          command: |
+            poetry build
+      - run:
+          name: Upload to PyPI
+          command: |
+            # Relies on the TWINE_USERNAME and TWINE_PASSWORD environment variables configured at:
+            #   https://app.circleci.com/settings/project/github/mozilla/experimenter/environment-variables
+            # For more on twine, see:
+            #   https://twine.readthedocs.io/en/latest/
+            poetry run twine upload --skip-existing dist/*
 
 workflows:
   version: 2
@@ -651,3 +677,9 @@ workflows:
             branches:
               ignore:
                 - main
+  deploy_schemas:
+    jobs:
+      - schemas_deploy:
+          filters:
+            branches:
+              only: main
--- a/schemas/LICENSE
+++ b/schemas/LICENSE
@@ -1,0 +1,373 @@
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
+1.1. "Contributor"
+    means each individual or legal entity that creates, contributes to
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -519,12 +519,38 @@ jobs:
...
+      - run:
+          name: Check for changes in schemas/ in last commit before proceeding.
+          command: |
+            if git diff --name-only main HEAD~1 | grep -E 'schemas/'
+              then
+                echo "Found changes to schemas dir, proceeding with deployment."
+              else
+                echo "No changes in /schemas directory. Skipping mozilla-nimbus-schemas deployment."
+                circleci-agent step halt
+            fi
```
jrbenny35, line 541: This will deploy after every schema change? I think we should only deploy if the toml file changes, because we would potentially deploy a change without updating the version.

.circleci/config.yml
```diff
@@ -651,3 +677,9 @@ workflows:
             branches:
               ignore:
                 - main
+  deploy_schemas:
+    jobs:
+      - schemas_deploy:
+          filters:
+            branches:
+              only: main
```
jrbenny35, line 685: This deploy could rely on the tests you wrote passing.

jrbenny35 (changes requested):
Just had a few questions

.circleci/config.yml
mikewilli: Yea, that's a good idea. I was a little worried about that scenario, but didn't think to just check for an update in the `pyproject.toml`.

jaredlockhart: @jrbenny35 Yes totally.  I think we only want to deploy on version bumps.  I've been confused about how to deal with that in this repo since using repo tags might not make sense.  But checking on version bumps is a great idea.  And yeah exactly @mikewilli that's a good idea since the pyproject can change for unrelated reasons.  Let's try all that good thinking both 👍

mikewilli: I am regretting my suggestion. With the old `setup.py` project configuration, it was easy to use python to dynamically parse a `VERSION` file, but with TOML that isn't possible. I went with what seemed like the most sensible approach that preserves the current poetry setup and the requirement of a single version definition in `pyproject.toml`: I changed the git diff check to look only in the `schemas/pyproject.toml` path, and then grep for `version` in the changes. It's still possible that this could be tricked, but to me this seems good enough. Open to suggestions though.

.circleci/config.yml
mikewilli: Is it necessary to re-run the checks on main? We've already run them in the PR and merge queue, so not sure if this adds any value.

.circleci/config.yml
```diff
@@ -524,6 +524,33 @@ jobs:
...
 
+  schemas_deploy:
+    docker:
+      - image: cimg/python:3.10.8 # poetry 1.2.2
+    steps:
+      - checkout
+      - run:
+          name: Check for package version change in last commit before proceeding.
+          command: |
+            if git diff main HEAD~1 schemas/pyproject.toml | grep 'version'
```
jaredlockhart, line 535: Just curious why `HEAD~1` and not `HEAD`?

.circleci/config.yml
```diff
@@ -524,6 +524,33 @@ jobs:
...
+          command: |
+            poetry build
+      - run:
+          name: Upload to PyPI
+          command: |
+            # Relies on the TWINE_USERNAME and TWINE_PASSWORD environment variables configured at:
+            #   https://app.circleci.com/settings/project/github/mozilla/experimenter/environment-variables
+            # For more on twine, see:
+            #   https://twine.readthedocs.io/en/latest/
+            poetry run twine upload --skip-existing dist/*
```
jaredlockhart, line 553: 🎉 🎉 🎉 🎉

jaredlockhart (approved):
Awesome looks great thank you!  Let's gooooooooo 🎉 🎉 🎉 🎉 🎉 🎉 🎉

.circleci/config.yml
mikewilli: `HEAD` is the current commit so `HEAD~1` compares current to previous.

mikewilli:
@jrbenny35 merging is blocked because you requested changes. Can you re-review when you get a chance? Thanks!

PR #8949, opened 2023-06-12
fix #8306 chore(project): update circle file path checks

### LINKED ISSUE #8306: Review path checks in CI
We added the capability to skip some CI checks based on which paths are touched in a PR.  Now that we've reorganized all the CI tasks, let's double check those paths and make sure we're checking the right paths in the right tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3154)

jaredlockhart:
Becuase

* The experimenter repo now hosts many projects
* Not all tests will be related to changes in all of those folders
* We can conditionally check which tests to run based on the paths of changes in a PR

This commit

* Updates all of the file path checks on all of the circleci tasks so we don't run unnecessary tests in PRs


.circleci/config.yml
```diff
@@ -151,6 +151,16 @@ jobs:
...
+      - run:
+          name: Check file paths
+          command: |
+            if git diff --name-only main HEAD | grep -E 'experimenter/'
+              then
+                echo "Continuing"
+              else
+                echo "No targeting changes, skipping"
+                circleci-agent step halt
+            fi
```
mikewilli, line 163: You can define a top level `commands` section with reusable steps that you can then call in here (with parameters if, for example, you want to pass the repo path to check for differences). https://circleci.com/docs/reusing-config/#authoring-reusable-commands

mikewilli (changes requested):
I'd suggest looking into the reusable commands, but otherwise this is great. Much better to separate out the steps more discretely.

Issue #8869, opened 2023-05-23
Add release date to timeline on Summary page


data-sync-user:
Add the release date to the timeline on experiment summary once it's live

!image-20230523-150704.png|width=1399,height=531!

See notes here: [https://docs.google.com/document/d/1nC_bcAQuGdo8T2ck3v-6HdT9olFffpFoct63CPzbU-w/edit](https://docs.google.com/document/d/1nC_bcAQuGdo8T2ck3v-6HdT9olFffpFoct63CPzbU-w/edit|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3513)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/37535">image-20230523-150704.png</a>

PR #8950, opened 2023-06-12
Fixes #8933 feat(nimbus): Only show release date for mobile

### LINKED ISSUE #8933: Disable release date field if not mobile
We should only let users pick a release date if it could be first run – so this should only be allowed for mobile.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3553)

eliserichards:
Because

- We only want the proposed release date field to be used if it is a first run experiment

This commit

* Renames the field to "First Run Release Date"
* Disables the field entirely if first run is not checked
* Hides the field (with the first run checkbox) if application is not mobile
* Moves it down with the first run checkbox

----

Screen Shot 2023-06-12 at 1 10 00 PM

https://github.com/mozilla/experimenter/assets/43795363/a7c13716-c59c-4d21-a05b-201115199457

Screen Shot 2023-06-12 at 1 10 29 PM

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -904,7 +904,6 @@ describe("FormAudience", () => {
       totalEnrolledClients: MOCK_EXPERIMENT.totalEnrolledClients,
       proposedEnrollment: "" + MOCK_EXPERIMENT.proposedEnrollment,
       proposedDuration: "" + MOCK_EXPERIMENT.proposedDuration,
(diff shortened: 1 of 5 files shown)


eliserichards:
@jaredlockhart Yes! The timeline will be added in #8869 which will be put up right after this 😄 

> you gotta mouse all the way over to the right to click the calendar dealy

Agreed, I'll make this smaller 👍 

> maybe it should be optionally required if it's first run? I noticed it's not incorporated into the timeline yet but what happens if it's first run but no release date is set? Should we keep it optional and then fall back on start date if it's first run but no release date is set? 

This is exactly correct! It's not required, even if it is first run. It falls back to start date for all of the computations if there's nothing entered for release date. There's still a decision being made if we want to maybe allow edits to this field after it's live if they decide they don't want start date... 👀

PR #8978, opened 2023-06-15
fix #8940 feat(cirrus): Move objects into main file 🧹 

### LINKED ISSUE #8940: Move resources into main file
Currently all the objects are being getting created in the individual file, we should move into single file and have a method for the fast API app creation too

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3557)

yashikakhurana:
Because

- Python `parses` the file line by line and it may cause an issue when creating several objects like` fml, SDK,` etc, currently, we are creating objects into their individual class file and importing them into the `main`

This commit

- Deletes unnecessary files
- move objects into the main file and separate method for setup
- FML- Exception handling and exit the container` if the path or channel is invalid `
- SDK- Exception handling and exit the container if `context is missing any required argument`
- Removes file path from docker volumes for feature manifest

UPDATE: uses the concept of `lifespan` instead of using `startup and shutdown events`, thanks to @jaredlockhart 
- This PR also updates the version of Fast API to support `lifespan`
closes #8940


cirrus/server/tests/test_experiment_recipes.py
```diff
@@ -4,11 +4,16 @@
 import requests
 
 from cirrus.experiment_recipes import RemoteSettings
+from cirrus.sdk import SDK
+from cirrus.settings import context
 
 
 class TestFetchRecipes(unittest.TestCase):
```
jrbenny35, line 11: Why unittest and not pytest?

cirrus/server/tests/test_experiment_recipes.py
```diff
@@ -4,11 +4,16 @@
...
+    def setUp(self):
+        self.sdk = SDK(context)
```
jrbenny35, line 13: a wild Camel appears 
```suggestion
    def set_up(self):
        self.sdk = SDK(context)
```

cirrus/server/tests/test_main.py
```diff
@@ -1,6 +1,51 @@
...
+from fml_sdk import FmlError
+
+from cirrus.experiment_recipes import RemoteSettings
+from cirrus.feature_manifest import FeatureManifestLanguage as FML
+from cirrus.main import fetch_schedule_recipes, initialize_app, setup
+from cirrus.sdk import SDK
+
+
+class TestSetup(unittest.TestCase):
+    def test_setup_with_fml_error(self):
```
jrbenny35, line 18: These really should be using pytest, no reason to use unittest just for cirrus.

jrbenny35 (changes requested):
Curious as to why we are using unittest over pytest

cirrus/server/tests/test_experiment_recipes.py
jaredlockhart: Sadly, the builtin methods in the Python unittest framework use [camelcase](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp).

cirrus/server/tests/test_experiment_recipes.py
jaredlockhart: Usually it's prudent to avoid adding any additional dependencies until they're necessary.  Python has a robust builtin test framework that is used broadly across Nimbus, for instance by all of the Experimenter unit tests.  I wouldn't suggest incorporating pytest unless there's a specific need to.

cirrus/server/tests/test_feature_manifest.py
```diff
@@ -6,12 +6,23 @@
...
 
 
 class FeatureManifestLanguageTestCase(unittest.TestCase):
     def setUp(self):
-        self.fml = FeatureManifestLanguage()
-        self.sdk = SDK()
+        self.fml = FeatureManifestLanguage(fml_path, channel)
+        self.sdk = SDK(context=context)
+
+    def test_invalid_fml_path(self):
```
jaredlockhart, line 17: 👏 👏 👏

cirrus/server/tests/test_feature_manifest.py
```diff
@@ -6,12 +6,23 @@
...
+        with self.assertRaises(FmlError) as e:
+            FeatureManifestLanguage("invalid_fml_path.txt", channel)
+        self.assertTrue(str(e.exception).startswith("FmlError.InvalidPath"))
+
+    def test_invalid_channel(self):
```
jaredlockhart, line 22: 👍 👍 👍

cirrus/server/tests/test_sdk.py
```diff
@@ -1,13 +1,52 @@
+import json
 import unittest
 
 from cirrus_sdk import NimbusError
+from parameterized import parameterized
 
+from cirrus.main import sdk
```
jaredlockhart, line 7: This is the same pattern again.  Don't initialize stateful objects in the global namespace at all.  If you need an instance of `SDK` in the tests, initialize it inside the `setUp` method.

cirrus/server/cirrus/main.py
```diff
@@ -1,23 +1,63 @@
...
+def setup():
+    app = initialize_app()
+    sdk = create_sdk()
+    remote_setting = RemoteSettings(sdk)
+    fml = create_fml()
+    scheduler = create_scheduler()
+    return app, sdk, remote_setting, fml, scheduler
+
+
+app, sdk, remote_setting, fml, scheduler = setup()
```
jaredlockhart, line 60: This is the same problem.  Don't initialize stateful objects in the global namespace, it looks like FastAPI uses a [dependency injection](https://fastapi.tiangolo.com/tutorial/dependencies/#declare-the-dependency-in-the-dependant) style.  Give that a shot.

jaredlockhart (changes requested):
Getting there.  Left a couple pointers for main.py and tests.

cirrus/server/cirrus/sdk.py
```diff
@@ -4,14 +4,12 @@
...
-from .settings import context
-
 logger = logging.getLogger(__name__)
 
 
 class SDK:
-    def __init__(self):
-        self.client = CirrusClient(context)
+    def __init__(self, context: str):
+        self.client = CirrusClient(context)  # does not match struct- throw an exception
```
brennie, line 12: What does this comment mean?

brennie:
Could you clarify your summary/description with regards to "move objects into main file"? That isn't super clear -- what are the objects in question?

cirrus/server/tests/test_experiment_recipes.py
yashikakhurana: I completely understand @jrbenny35 how weird it looks in python code, but builtin methods :(

cirrus/server/tests/test_experiment_recipes.py
yashikakhurana: so currently unittest meets our testing requirements, and incorporating pytest would introduce an additional dependency without a specific need. However, we can keep pytest in mind for future considerations :)

cirrus/server/cirrus/main.py
yashikakhurana: good find, actually this got introduced in version `0.93.0` and I was using `0.92.0 `(that time it was current :D), so I updated the fast API to a latest version now- `0.97.0` in this PR, this is actually neat, no need to have a separate method for startup and shutdown, thanks @jaredlockhart 🌷

cirrus/server/cirrus/sdk.py
yashikakhurana: so CirrusClient is actually coming from nimbus_sdk, if the context we are providing doesn't match what it is expecting, it will throw an error

yashikakhurana:
> Could you clarify your summary/description with regards to "move objects into main file"? That isn't super clear -- what are the objects in question?

Absolutely- The reason behind consolidating the object creation into the `main.py` file was to address potential issues that can arise when creating multiple objects in separate files. Python parses files line by line, and if the objects are created in their individual class files, there is a possibility of encountering dependency or import-related problems. By moving the object creation process to the main file and importing the necessary classes, just making ensure that the objects are created in the correct order and any dependencies are properly handled. so that to avoid potential runtime errors, let me know if it helps :)

cirrus/server/tests/test_experiment_recipes.py
jrbenny35: > Usually it's prudent to avoid adding any additional dependencies until they're necessary. Python has a robust builtin test framework that is used broadly across Nimbus, for instance by all of the Experimenter unit tests. I wouldn't suggest incorporating pytest unless there's a specific need to.

There's pytest being declared wthin some of these tests already. Also, a lot of the tests redefine the same variables for each test, pytest would solve this easily.  We have pytest for the Rust integration tests, now unit test here, and we will probably use pytest for the integration tests here. That seems like a lot of fragmentation for testing, when we could just use pytest everywhere and have a matching framework for all the tests.

> Sadly, the builtin methods in the Python unittest framework use camelcase.

No need for setup functions either with pytest.

jaredlockhart: @jrbenny35 Yes super duper good points, we should absolutely be consistent in which patterns and frameworks we're using.  If we're already using some pytest constructs we should continue to, 100% agree 👍

jaredlockhart (approved):
Okay cool!  This is starting to look really good.  In terms of addressing the issues of stateful globals, I think this looks good.  @jrbenny35 raises some good points about fragmentation in testing patterns around unittest and pytest that I think would be good to think about but can probably be done as a followup rather than blowing up the scope of this pr.  R+ from me 👍

cirrus/server/tests/test_experiment_recipes.py
yashikakhurana: yeah, my mistake, I missed that, I think in starting tests I added, then started using unittest, this also requires some changes in other test files too- so added a ticket, will do follow up in here- https://github.com/mozilla/experimenter/issues/8993

yashikakhurana:
> Okay cool! This is starting to look really good. In terms of addressing the issues of stateful globals, I think this looks good. @jrbenny35 raises some good points about fragmentation in testing patterns around unittest and pytest that I think would be good to think about but can probably be done as a followup rather than blowing up the scope of this pr. R+ from me 👍

yup absolutely, @jrbenny35 thank you for noticing that, @jaredlockhart I added a follow up ticket- https://github.com/mozilla/experimenter/issues/8993

PR #8984, opened 2023-06-15
fix #8983 fix(schemas): make schemas deploy dependent on install and build

### LINKED ISSUE #8983: Schemas deploy needs install dependency
Twine isn't installed right now

mikewilli:
Because

- twine isn't installed until poetry installs dependencies
- `poetry build` doesn't install dependencies

This commit

- adds the `schemas_install` to the deploy's prereqs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -679,12 +679,6 @@ workflows:
                 - main
   deploy_schemas:
     jobs:
-      - schemas_check:
-          name: Check Schemas (main)
-          filters:
-            branches:
-              ignore:
-                - main
       - schemas_deploy:
           filters:
             branches:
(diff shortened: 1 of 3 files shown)


eliserichards (approved):
Second time's the charm! 💯

jaredlockhart (approved):
You can technically test it by changing the circle config to run the deploy command on branches instead of main and keep pushing to your branch to trigger a circle run until you're confident in it and then flip it to main only when you're ready to merge.  Also doing it using make command dependency is exactly the right way to do it.  If you want to test it in this pr try what I said above before you merge.

mikewilli:
That's true, and probably a good idea here. I'm a little wary of this as a general practice because it seems easy to make a mistake and commit the wrong thing, and I don't think it's a direct 1-for-1 comparison (the git diff is probably not right, for example), but at least would have found the issues I ran into here.

PR #8987, opened 2023-06-15
fix #8986 fix(schemas): update the make command to switch dirs properly

### LINKED ISSUE #8986: fix schemas deploy context
schemas deploy is in the wrong directory

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3588)

mikewilli:
Because

- the make command to deploy schemas uses the wrong syntax to change directories

This commit

- fixes the syntax in order to run the twine upload from the directory with the build is

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,6 +265,4 @@ schemas_build:
 	(cd schemas && poetry build)
 
 schemas_deploy_pypi: schemas_install schemas_build
-	cd schemas;
-	poetry run twine upload --skip-existing dist/*;
-	cd ..
+	cd schemas; poetry run twine upload --skip-existing dist/*;


jaredlockhart (approved):
I wonder if there's some more idiomatic way to do this?  Reading [this](https://stackoverflow.com/questions/1789594/how-do-i-write-the-cd-command-in-a-makefile) it seems like there's lots of patterns but nothing jumps out at me as The One Way to do it so let's just roll with this.

mikewilli:
Yep, I was looking around for the best to do it and landed on a very similar discussion (maybe even the same one).

PR #8991, opened 2023-06-16
fix #8889 feat(nimbus): adds sizing data fetch task 

### LINKED ISSUE #8889: Ingest population sizing data into Experimenter
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3534)

mikewilli:
Because

- we are generating population sizing data automatically
- experimenter can use this data to help users determine audience sizing for certain experiments

This commit

- adds a task to fetch the sizing data
- adds schemas for the sizing data
- puts the ingested/validated sizing data into django cache as JSON


experimenter/experimenter/jetstream/tasks.py
```diff
@@ -2,11 +2,12 @@
 
 import markus
 from celery.utils.log import get_task_logger
+from django.core.cache import cache
```
jaredlockhart, line 5: So, by default this will use the [in memory cache](https://docs.djangoproject.com/en/4.2/topics/cache/#local-memory-caching) which means this data won't be available in the request handlers.  So we'd either have to change the cache backend to point to redis, which we do have deployed and available

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/settings.py#L343-L345

but isn't currently [configured](https://docs.djangoproject.com/en/4.2/topics/cache/#redis) so if you wanna use that you'd have to either land that in here or as a separate PR I think?

jaredlockhart (changes requested):
Comments about caching, generally the structure looks right to me, and we'd have to add some tests.  Looking good, coming along! 🎉

experimenter/experimenter/jetstream/tasks.py
mikewilli: Ah great! This was something I was wondering, so good to know. This part is easy to change, is there a better solution to make the data available for the front end? I think you mentioned something last week but I forget what it was and cannot find any notes on the discussion.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -1108,3 +1113,147 @@ def test_exception_for_fetch_experiment_data(self, mock_get_experiment_data):
...
+                    }
+                }
+                """
+
+        def open_file(filename):
+            return File(filename)
+
+        mock_open.side_effect = open_file
+        mock_exists.return_value = True
+
```
jaredlockhart, line 1209: I would assert that it is None here before we fetch.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -1108,3 +1113,147 @@ def test_exception_for_fetch_experiment_data(self, mock_get_experiment_data):
...
+        tasks.fetch_population_sizing_data()
+        sizing_cache = caches["sizing"]
+        sizing_results = sizing_cache.get("population_sizing")
+        self.assertIsNotNone(sizing_results)
```
jaredlockhart, line 1213: Could we assert on the actual content here?

experimenter/experimenter/settings.py
```diff
@@ -344,6 +344,17 @@
 REDIS_PORT = config("REDIS_PORT")
 REDIS_DB = config("REDIS_DB")
 
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    },
```
jaredlockhart, line 350: We can probably leave 'default' out here altogether since we're not using caching anywhere and declaring it here could indicate that we expect it to be available when we don't.

jaredlockhart (changes requested):
Generally looks really good!  What's a good way to test this locally?

experimenter/experimenter/settings.py
mikewilli: So `default` is required, which means we can either use the Redis cache as the default, or do something like what I have here. My original thought was to leave the default as-is, but if we are intentionally not using it then I think it's fine to swap it for the Redis one (and remove the `sizing` cache).

mikewilli:
> Generally looks really good! What's a good way to test this locally?

You can run the task in a `make bash` --> `./manage.py shell` like so:
```python
from experimenter.jetstream.tasks import fetch_population_sizing_data
fetch_population_sizing_data()
```

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -1108,3 +1118,155 @@ def test_exception_for_fetch_experiment_data(self, mock_get_experiment_data):
...
+                return sizing_test_data
+
+        def open_file(filename):
+            return File(filename)
+
+        mock_open.side_effect = open_file
+        mock_exists.return_value = True
+
+        sizing_results = cache.get(SIZING_DATA_KEY)
+        assert sizing_results is None
```
jaredlockhart, line 1218: So this is the nittiest of all nits, but when you're in a `unittest` context you have all the [built in assertions](https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug) which will format the output nicely, whereas `assert` is really naive and will just throw an exception, so generally if you're in a `unittest` context it's nicer to use the builtin asserts.

experimenter/experimenter/settings.py
```diff
@@ -346,6 +346,15 @@
 REDIS_PORT = config("REDIS_PORT")
 REDIS_DB = config("REDIS_DB")
 
+CACHES = {
+    "default": {
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": f"redis://{REDIS_HOST}:{REDIS_PORT}/{REDIS_DB}",
+        "TIMEOUT": None,
+    },
+}
```
jaredlockhart, line 355: Oh I meant the other way around, like 

```py
CACHES = {
    "population_sizing": {
        ...
    }
}
```

and have no default that way 1) anybody trying to use the default will error out 2) people have to explicitly request the cache key they want to use

experimenter/experimenter/settings.py
```diff
@@ -346,6 +346,15 @@
...
+SIZING_DATA_KEY = "population_sizing"
```
jaredlockhart, line 356: 👍 🙏

jaredlockhart (approved):
Okay I didn't get a chance to test this locally but everything looks solid to me.  Couple small points but nothing blocking.  Let's get this landed and running in prod and if we missed anything sentry will tell us.  Great job @mikewilli can't wait to see this start filling in on the ui!  🎉 🎉 🎉

experimenter/experimenter/settings.py
mikewilli: Yea, tried to explain why I went with this in the other reply, but the important thing is that `default` is required, so our options are to either define `required` as something we won't use and `sizing` (or whatever) as redis, or just define `default` as redis.

I'll leave it as-is for now and we can modify as needed.

mikewilli: One thing I just noticed is that django provides a [DummyCache](https://docs.djangoproject.com/en/3.2/topics/cache/#dummy-caching-for-development), so maybe this would be worth looking into as an option for `default`. I think it might be better to use something that, instead of silently succeeding without caching like DummyCache, explicitly errors if you try to use it so that anyone who tries to use it knows it won't work.

Not a big concern for the moment, but let me know what you think.

experimenter/experimenter/jetstream/tests/test_tasks.py
mikewilli: I'm not sure why I didn't use `assertIsNone` here! Good catch.

PR #8995, opened 2023-06-17
fixes #8994 feat(nimbus): add a targeting config for Firefox Suggest data collection disabled

### LINKED ISSUE #8994: Add a targeting config for Firefox Suggest data collection disabled
The Firefox Suggest team will be running an experiment that needs to target users who have the pref `browser.urlbar.quicksuggest.dataCollection.enabled` set to false.

Note that there's already a targeting config for the opposite -- users who have this pref set to true. See [URLBAR_FIREFOX_SUGGEST_DATA_COLLECTION_ENABLED](https://github.com/mozilla/experimenter/blob/cfae9787d16bef74db54d4020a5da4b2cf8aef8c/experimenter/experimenter/targeting/constants.py#L449-L458).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3593)

0c0w3:


0c0w3:
Hi, I don't know if I did this right, so to be clear I want to target users who have the pref `browser.urlbar.quicksuggest.dataCollection.enabled` set to false. It's false by default. In case it matters, the pref is sticky, so if the user flips it on and then back off again, it will keep its `false` value on the user branch.

jaredlockhart (approved):
Yes this looks exactly right to me 👍

PR #8999, opened 2023-06-19
fix #8997 chore(project): remove unused containers

### LINKED ISSUE #8997: Clear docker images on refresh
The make refresh command will delete old unused images and volumes so that the disk allocation for docker doesn't fill up endlessly, but we missed one set of images that can grow.  Amend the makefile to include deleting these other images.

```
-volumes_rm:
+docker_prune:
+       docker image prune -f
        docker volume prune -f
 
-kill: compose_stop compose_rm volumes_rm
+kill: compose_stop compose_rm docker_prune
        echo "All containers removed!"
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3595)

jaredlockhart:
Because

* Starting, stopping, and rebuilding docker containers can result in a growing number of old images, containers, and volumes
* Eventually this can fill the disk allocated to docker and prevent new containers from starting
* We have commands built into the Makefile for removing old and unused images and containers but it doesn't catch everything

This commit

* Updates the Makefile to remove more old and unused containers and images without impacting the build cache

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,9 @@ compose_rm:
 	$(COMPOSE_INTEGRATION) rm -f -v || true
 	$(COMPOSE_PROD) rm -f -v || true
 
-volumes_rm:
+docker_prune:
+	docker container prune -f
+	docker image prune -f
 	docker volume prune -f
 
 static_rm:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -120,7 +120,9 @@ compose_rm:
 	$(COMPOSE_INTEGRATION) rm -f -v || true
 	$(COMPOSE_PROD) rm -f -v || true
 
-volumes_rm:
+docker_prune:
+	docker container prune -f
+	docker image prune -f
 	docker volume prune -f
```
yashikakhurana, line 126: 🎉

yashikakhurana (approved):
yeah after a certain amount of time I have to run it, good to have it here 💪 , thanks @jaredlockhart

PR #9005, opened 2023-06-21
fix #9001 feat(nimbus): Add support for targeting firefox version 114.3

### LINKED ISSUE #9001: Added version 114.3 as a target criteria
I would like to target iOS version 114.3 with a feature rollout.

OrlaM:
Because

    We want to target minimum versions of 114.3 for future experiments

This commit

    Adds 114.3 to the version list


experimenter/experimenter/experiments/constants.py
```diff
@@ -329,6 +329,7 @@ def parse(version_str):
         FIREFOX_113 = "113.!"
         FIREFOX_113_0_1 = "113.0.1"
         FIREFOX_114 = "114.!"
+        FIREFOX_114_3 = "114.3"
```
jaredlockhart, line 332: Ah generally it's better to specify the whole semver so `114.3.0` or `114.3.!`

jaredlockhart (changes requested):

jaredlockhart (approved):
Perfect thank you!

PR #9006, opened 2023-06-21
fix #8733 feat(cirrus): Accept request data for cirrus API

### LINKED ISSUE #8733: Set up api to get enrollment status

yashikakhurana:
Because

- We want to receive data as part of the user `post` request - `/v1/features/` to evaluate the feature configuration

This commit

- Accept the params in following structure 
```json
{
  "client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449",
  "context": {
    "key1": "value1",
    "key2": {
      "key2.1": "value2",
      "key2.2": "value3"
    }
  }
}
```
ps: If the client id or client key, context key or context value is missing it will still return the default configuration
- Updates the openapi schema

closes #8733


cirrus/server/cirrus/main.py
```diff
@@ -16,6 +17,11 @@
 logger = logging.getLogger(__name__)
 
 
+class FeatureRequest(BaseModel):
+    client_id: Optional[str]
+    context: Optional[Dict[str, Any]]
```
jaredlockhart, line 22: Should these be `Optional`?

cirrus/server/cirrus/main.py
```diff
@@ -76,9 +82,11 @@ def read_root():
 
 
 @app.post("/v1/features/", status_code=status.HTTP_200_OK)
-async def compute_features():
-    # # will recieve as part of incoming request
-    targeting_context: Dict[str, Any] = {"clientId": "test", "requestContext": {}}
+async def compute_features(request_data: FeatureRequest):
+    targeting_context: Dict[str, Any] = {
```
jaredlockhart, line 86: Could we not reuse the type structure we declared above here?

cirrus/server/cirrus/main.py
```diff
@@ -76,9 +82,11 @@ def read_root():
...
+        "clientId": request_data.client_id or "",
+        "requestContext": request_data.context or {},
```
jaredlockhart, line 88: Should we not fail some validation step and abort here if either the `client_id` or `context` are missing?

jaredlockhart (changes requested):

cirrus/server/cirrus/main.py
yashikakhurana: request data expect` client_id`, sdk need `clientId`, I can either change the request parameter key and then use this as a type too, or make another structure for this, what do you think?

jaredlockhart: Oh I see.  Oh well anyways, we know we need to at some point declare actual types for all these dicts, including other ones in other parts of the code.  I guess we can do that later as part of hardening.

cirrus/server/cirrus/main.py
yashikakhurana: my idea was whatever happens with a request if the client id is provided or not provided we return the default config, definitely it will be logged whenever that happen, or we can return 400 with the below changes
```python
 if not request_data.client_id:
        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Missing client ID")
 if not request_data.context:
        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Missing context")

File: cirrus/server/cirrus/main.py (hunk 1 above)
Comment (yashikakhurana) [reply cirrus/server/cirrus/main.py:22]: okay so if we decide we want to return default config whether key is present or not, then yes it should be optional, if we don't want to continue and required these two keys then we can remove this optional

File: cirrus/server/cirrus/main.py (hunk 3 above)
Comment (jaredlockhart) [reply cirrus/server/cirrus/main.py:88]: Yeah my point about returning the defaults was in the case where we're unable to sync with remote settings.  A malformed request should always return an error to the caller, otherwise something could break on their end and they'd have no way to know.

Comment (jaredlockhart) [reply cirrus/server/cirrus/main.py:88]: Yeah I'd expect it to return a 400.  You could do it that way but at some point we'll want to have a proper structure for deserializing and validating the data.

Comment (yashikakhurana) [reply cirrus/server/cirrus/main.py:88]: okay so will be using default 422 which fastapi provides to validate

File: cirrus/server/cirrus/main.py [hunk 4]
```diff
@@ -76,9 +82,22 @@ def read_root():
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Client ID value is missing or empty",
+        )
+
+    if not request_data.context:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Context value is missing or empty",
+        )
```
jaredlockhart, line 96: I see, so if the key is missing you get 422 but if the key is present but null or context is `{}` or anything falsy, you get 400.  Interesting.  I wonder if these shoudl all be 422s just to be consistent?  Just a thought, no need to change it right now.

jaredlockhart (approved):
10/10 A+ looks great thank you @yashikakhurana ! 🎉 🎉 🎉 🎉

cirrus/server/cirrus/main.py
yashikakhurana: yeah I was debating myself on this should it be 400 vs 422, but good point to be consistent, I will keep a note of these and can change it later 💪

PR #9007, opened 2023-06-21
Add targeting needed for the Easy Setup pt.4 experiment

AllegroFox:
This commit adds targeting for users that opt out of pinning Firefox during installation (or otherwise are not pinned), for use by the Easy setup pt.4 experiment.

First start-up users on Windows 10 1903 (build 18362) or newer, with a new profile, needing default needing pin

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -236,7 +236,7 @@ FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903_PREFER_MOTION = NimbusTargetingC
     application_choice_names=(Application.DESKTOP.name,),
 )
 
-FIRST_RUN_NEED_DEFAULT_NEED_PIN_WINDOWS_1903 = NimbusTargetingConfig(
+FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_NEED_PIN_WINDOWS_1903 = NimbusTargetingConfig(
     name=(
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
LGTM 👍

experimenter/experimenter/targeting/constants.py
```diff
@@ -211,6 +211,33 @@ def __post_init__(self):
...
+        "First start-up users on Windows 10 1903 (build 18362) or newer, with a "
+        "new profile, needing default & pin"
+    ),
+    slug="first_run_need_default_need_pin",
+    description=(
+        "First start-up users (e.g. for about:welcome) on Windows 1903+, "
+        "with a new profile, needing default & pin"
+    ),
+    targeting=(
+        "{first_run} && os.windowsBuildNumber >= 18362 && {new_profile} && "
```
punamdahiya, line 225: Should we reuse `
FIRST_RUN_NEW_PROFILE_NEED_DEFAULT_WINDOWS_1903` similar to how its done in `FIRST_RUN_NEW_PROFILE_HAS_PIN_NEED_DEFAULT_WINDOWS_1903`, in this case we can use doesAppNeedPin directly

punamdahiya (changes requested):

experimenter/experimenter/targeting/constants.py
AllegroFox: Done! Also move the targeting down so that it's next to the related ones

punamdahiya (approved):
LGTM! Thanks

PR #9015, opened 2023-06-22
fix #9012 fix(schemas): update schemas to work against valid data

### LINKED ISSUE #9012: Update schemas after further testing
Testing validation of the ingest produced some errors in the schema that we should not be seeing. Schema needs some minor modifications to account for these errors so they do not appear with valid data.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3612)

mikewilli:
Because

- adding validation on ingest using the schemas raised validation errors on valid data

This commit

- updates the schemas to work with valid data


brennie (approved):
LGTM! r+

PR #9016, opened 2023-06-22
fix #9013 feat(nimbus): Require Firefox 96 as minimum version

### LINKED ISSUE #9013: Set minimum supported version for all experiments/rollouts to 96
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3611)

brennie:
Because
- we want to simplify the v6 API to serialize the features field unconditionally; and
- Firefox 96 is the maximum minimum supported version for multi-feature experiments

this commit
- sets the minimum required Firefox version for experiments and rollouts to 96.

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -411,6 +411,7 @@ Optional - We believe this outcome will <describe impact> on <core metric>
     ERROR_FIREFOX_VERSION_MIN = (
         "Ensure this value is less than or equal to the maximum version"
     )
+    ERROR_FIREFOX_VERSION_MIN_96 = "The minimum targetable Firefox version is 96"
     ERROR_FIREFOX_VERSION_MAX = (
         "Ensure this value is greater than or equal to the minimum version"
     )
@@ -451,3 +452,5 @@ Optional - We believe this outcome will <describe impact> on <core metric>
     DEFAULT_TREATMENT_BRANCH_NAME = "Treatment A"
 
     L10N_MIN_STRING_ID_LEN = 9
+
+    MIN_REQUIRED_VERSION = Version.FIREFOX_96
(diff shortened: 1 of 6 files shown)


jrbenny35:
The UI test failed, looks like there's a field missing or something? https://output.circle-artifacts.com/output/job/7c8e65c8-f24b-45f8-b281-a7db7d4901bd/artifacts/0/~/experimenter/experimenter/tests/integration/test-reports/report.htm

experimenter/experimenter/experiments/constants.py
```diff
@@ -411,6 +411,8 @@ class EmailType(models.TextChoices):
     ERROR_FIREFOX_VERSION_MIN = (
         "Ensure this value is less than or equal to the maximum version"
     )
+    ERROR_FIREFOX_VERSION_MIN_96 = "The minimum targetable Firefox version is 96"
+    ERROR_FIREFOX_VERSION_MIN_REQUIRED = "Experiments must set a minimum Firefox version"
```
jaredlockhart, line 415: Where is this used?

brennie: `ERROR_FIREFOX_VERSION_MIN_REQUIRED` isnt used anymore.

jaredlockhart: But why is it in the PR?

jaredlockhart (approved):
Fantastic 👍 ty 🙏

experimenter/experimenter/experiments/constants.py
brennie: I left it in by accident. A previous version had different messages when the version was too low vs not present at all. I replaced it with a single message so that people wouldnt hit both errors if they tried to set No version -> a version too low.

jaredlockhart: Yep makes sense 👍

PR #9029, opened 2023-06-23
fix #8501 docs(cirrus): Update readme for cirrus

### LINKED ISSUE #8501: Update Readme for Cirrus
We are adding Cirrus and adding make commands related to it, we should also explain and list down in readme, like how to start service, how to run test cases etc.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3279)

yashikakhurana:
Because

- We use various configs and commands for the Cirrus

This commit

- Update readme to help users to understand what configs value is needed to set up cirrus
- Explains make commands related to Cirrus

Also removes the path from docker , can be configured in `.env` itself
closes #8501

### DIFF
--- a/cirrus/README.md
+++ b/cirrus/README.md
@@ -1,9 +1,87 @@
-Create `.env` file inside directory `cirrus/server` and copy
-`cp .env.example .env`
+# Cirrus
+
+Cirrus is a feature configuration server that allows clients to obtain a set of features based on their provided `client_id` and `context` information.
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Excellent fantastic I love it exactly what I was hoping for 🎉 🎉 🎉 🎉 🎉  ty @yashikakhurana !!

PR #9044, opened 2023-06-27
fix #9096 feat(nimbus): Add inclusion and exclusion targeting APIs

### LINKED ISSUE #9043: Inclusion/Exclusion Targeting UI
Let’s add some UI to the Audience page in Experimenter that has two drop downs, one called ‘Include Experiments’ and one called ‘Exclude Experiments’. Each drop down should be a multiselect that lists all NimbusExperiments in all statuses. A user can select any number of other experiments in each list, but no experiment should be selectable in both. The selected included and excluded experiments should appear as part of the targeting JEXL string in the recipe as either being present or not present in the list of enrolled experiment slugs in the client.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3643)

### LINKED ISSUE #9096: API changes for inclusion/exclusion targeting
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3673)

brennie:
Because:
- we want to target against past and present enrollments in other
  experiments and rollouts

this commit:
- updates the API with `excluded_experiments` and `required_experiments`
  fields; and
- adds a new gql query to list experiments filtered by application
  to allow the front-end to populate these fields in an upcoming change.


jaredlockhart:
We should either block this PR or at least version gate it on the versions this will be available in the client, which might not be clear since the patches aren't landed yet.

jaredlockhart:
Oh yeah you've got a version gate in there, but we should at least wait until we know what the version is before we land this.  

Also sidebar, I would've split this apart into at least backend and frontend changes rather than lumping it all together.  We should always expect that a PR is revertible in prod, with the caveat that migrations aren't easily reverted, so isolating changes into smaller prs as much as possible should be preferred.

But everything in here looks generally correct to me 👍

brennie:
Blocked waiting for mobile to support `enrollments` targeting variable

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1477,6 +1542,26 @@ def _validate_countries_versions(self, data):
                 )
         return data
 
+    def _validate_enrollment_targeting(self, data):
+        excluded_experiments = set(data.get("excluded_experiments", []))
+        required_experiments = set(data.get("required_experiments", []))
+
+        common = excluded_experiments & required_experiments
```
jaredlockhart, line 1549: I love sets so much 🙏

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -662,3 +669,9 @@ def resolve_languages(self, info):
 
     def resolve_changes(self, info):
         return self.changes.all().order_by("changed_on")
+
+    def resolve_excluded_experiments(self, info):
+        return self.excluded_experiments.only("id", "slug")
+
+    def resolve_required_experiments(self, info):
+        return self.required_experiments.only("id", "slug")
```
jaredlockhart, line 677: Slug and not name?  Name might be more humane?

jaredlockhart: Also above it says it's a list of ExperimentType but here we're only providing id and slug, would it make sense to add a type that's just those two fields?

experimenter/experimenter/experiments/api/v5/queries.py
```diff
@@ -32,5 +38,8 @@ def resolve_experiment_by_slug(self, info, slug):
         except NimbusExperiment.DoesNotExist:
             return None
 
+    def resolve_experiment_meta_by_application(self, info, application):
```
jaredlockhart, line 41: Why `meta` in the name here?  I would expect this to be called something like `experiments_by_application`?

jaredlockhart (commented):
Looking good.  I would still split this by backend/frontend changes to be smaller to review/land.

experimenter/experimenter/experiments/api/v5/types.py
brennie: The .only() is just a DB optimization because we know in practice the query is just ehse fields. If the query asks for more fields, the model will query for them on-the-fly.

brennie: Also turns out the query was asking for name and publicDescription already, so I've added those to the query and will update this soon.

experimenter/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -2138,6 +2225,102 @@ def test_query_feature_order(self):
         )
 
 
+class TestNimbusAllExperimentsMetaQuery(GraphQLTestCase):
```
jaredlockhart, line 2228: Probably `TestNimbusExperimentsByApplicationQuery` now?

experimenter/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -2138,6 +2225,102 @@ def test_query_feature_order(self):
...
+    GRAPHQL_URL = reverse("nimbus-api-graphql")
+
+    @property
+    def default_headers(self):
+        return {settings.OPENIDC_EMAIL_HEADER: "user@example.com"}
+
+    def test_query(self):
```
jaredlockhart, line 2235: You can probably get rid of this test altogether, the other one tests both the inclusion and exclusion cases?

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -1491,3 +1491,34 @@ def test_preview_to_draft_sets_published_dto_to_None(self):
...
         experiment = serializer.save()
         self.assertIsNone(experiment.published_dto)
+        self.assertEquals(experiment.proposed_release_date, None)
+
+    @parameterized.expand(
+        [
+            ("excluded_experiments",),
+            ("required_experiments",),
+        ]
+    )
```
jaredlockhart, line 1501: I don't think that's necessary, just explicitly set both fields in the test.

jaredlockhart (approved):
Yep looks great, thnx for splitting it up.  I don't think you need to parameterize the tests across both fields, I'd just hardcode them, but up to you.  Thnx @brennie ! 🎉

PR #9050, opened 2023-06-28
feat(nimbus): Models and migrations for inclusion and exclusion targeting

brennie:
Because:
- we want to target experiments based on whether or not the user has been in other experiments

this commit:
- updates the db models to allow for such targeting.


yashikakhurana (dismissed):
In general looks good, can we add a test where `required_experiments` and `excluded_experiments` is not empty 🤔

yashikakhurana (approved):
thank you for referencing to that PR, 💯

jaredlockhart:
Perfect thank you 🙏 🎉   And yeah to @yashikakhurana 's point, I don't think a model change PR like this actually needs any tests.

PR #9067, opened 2023-07-04
Add ADR for exposure events and coenrolling features

jhugman:
Because

- Coenrolling features are about to exist
- Exposure events rely on the default behavior of feature exclusivity

This commit

- Proposes several ways forward in the form of an Architectural Decision Record.

[Rendered version](https://github.com/mozilla/experimenter/blob/jhugman/exp-3630-adr-for-exposure-events-of-coenrolling-features/docs/adrs/0012-exposure-events-and-coenrolling-features.md) for easier reading.

My strong recommendation is for **Option 3: Annotate the message with the experiment slug**

Fixes [EXP-3630](https://mozilla-hub.atlassian.net/browse/EXP-3630).

### DIFF
--- a/docs/adrs/0012-exposure-events-and-coenrolling-features.md
+++ b/docs/adrs/0012-exposure-events-and-coenrolling-features.md
@@ -64,19 +64,19 @@ The [`is-control` message property](https://experimenter.info/mobile-messaging/#
 
 ## Decision Outcome
 
-Under consideration. Please submit comments before 2023-07-11.
+Chosen option: "Option 3: Add the experiment slug to the feature schema, so the experiment slug travels with the message", because it is the only option that meets the criteria set by the decision drivers.
 
-<!-- Chosen option: "[option 1]", because [justification. e.g., only option, which meets k.o. criterion decision driver | which resolves force force | … | comes out best (see below)]. -->
+### Positive Consequences
 
-<!-- ### Positive Consequences
-
-* [e.g., improvement of quality attribute satisfaction, follow-up decisions required, …]
-* …
+* Good, because implementation is relatively simple, and restricted to the client side.
+* Good, because the correct number of exposure events are recorded.
+* Good, because no extra steps needed in post experiment data-analysis.
 
 ### Negative Consequences
 
-* [e.g., compromising quality attribute, follow-up decisions required, …]
-* … -->
+* Bad, because it involves a change to the feature JSON, and the user to change their behavior.
+
+This has consequences for the implementation of `messaging` in Firefox for iOS and Firefox for Android, and some work in the Nimbus SDK, and documentation in experimenter.info.
 
 ## Pros and Cons of the Options
 
--- a/docs/adrs/0012-exposure-events-and-coenrolling-features.md
+++ b/docs/adrs/0012-exposure-events-and-coenrolling-features.md
@@ -8,11 +8,11 @@ Technical Story: [Supporting exposure events for coenrolled features](https://mo
 
 ## Context and Problem Statement
 
-Application feature code can be configured from and by Nimbus. In the case of `messaging` and `onboarding`, the complete set of messages and all their associated data is implemented by this feature configuration.
+Application feature code can be configured from and by Nimbus. In the case of `messaging`, the complete set of messages and all their associated data is implemented by this feature configuration.
 
 The feature configuration is composed by merging together the default configuration, any configuration from a rollout, and any configuration from experiments.
 
-In the next release, features can be configured to enroll in more than one experiment on the same client device. Such features, known as coenrolling features, serve multiple treatments to the user (e.g. a `messaging` message, or an `onboarding` card) from different parts of the complete feature configuration.
+In the next release, features can be configured to enroll in more than one experiment on the same client device. Such features, known as coenrolling features, serve multiple treatments to the user (e.g. a `messaging` message) from different parts of the complete feature configuration.
 
 It is not _currently_ possible to work from the part of the feature configuration backwards to the experiment to record an exposure event, giving credit for experiment that provided this treatment.
 
--- a/docs/adrs/0012-exposure-events-and-coenrolling-features.md
+++ b/docs/adrs/0012-exposure-events-and-coenrolling-features.md
@@ -8,11 +8,17 @@ Technical Story: [Supporting exposure events for coenrolled features](https://mo
 
 ## Context and Problem Statement
 
+Application feature code can be configured from and by Nimbus. In the case of `messaging` and `onboarding`, the complete set of messages and all their associated data is implemented by this feature configuration.
+
 The feature configuration is composed by merging together the default configuration, any configuration from a rollout, and any configuration from experiments.
 
-Features can now be configured to enroll in more than one experiment on the same client device. Such features, known as coenrolling features, serve multiple treatments to the user (e.g. a `mmessaging` message, or an `onboarding` card) from different parts of the complete feature configuration.
+In the next release, features can be configured to enroll in more than one experiment on the same client device. Such features, known as coenrolling features, serve multiple treatments to the user (e.g. a `messaging` message, or an `onboarding` card) from different parts of the complete feature configuration.
 
-It is not currently possible to work from the part of the feature configuration backkwards to the experiment to record an exposure event, giving credit for experiment that provided this treatment.
+It is not _currently_ possible to work from the part of the feature configuration backwards to the experiment to record an exposure event, giving credit for experiment that provided this treatment.
+
+This ADR is deciding what to do— in the case of messaging— about the attribution of which experiment a particular message came from.
+
+### Current status: no coenrollments
 
 Currently, for [experimenting with mobile `messaging`](https://experimenter.info/mobile-messaging/#experimenting-with-messages), the message sender (the user) must identify which message or messages are under experiment, then the system works out the experiment. For example, the feature JSON for annotating a message as being under experiment is:
 
@@ -74,9 +80,9 @@ Under consideration. Please submit comments before 2023-07-11.
 
 ## Pros and Cons of the Options
 
-### Record exposure events only when we're sure of which experiment is involved
+### Option 1: Record exposure events only when we're sure of which experiment is involved
 
-In practice, this is might be for when exactly one experiment is using the `messaging` feature.
+In practice, this might be only for when exactly one experiment is using the `messaging` feature.
 
 The message sending experiment owner should continue to use the `message-under-experiment` key.
 
@@ -89,7 +95,7 @@ Fewer exposure events will be emitted than should be.
 
 This is the current implementation.
 
-### Record the exposure event with the message key as well as the experiment slug
+### Option 2: Record the exposure event with the message key
 
 The Glean event schema for exposure events would have an additional `part-id` extra, or even another type of exposure event created.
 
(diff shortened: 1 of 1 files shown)


docs/adrs/0012-exposure-events-and-coenrolling-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+When the experiment slug is known, e.g. at enrollment, the feature config is checked for strings containing `{experiment}`, and the slug used to replace it: i.e. the user should not need to type the actual slug.
+
+When the message is displayed, the `experiment` property is the experiment slug; this can be used to look up the branch, and an exposure event then recorded.
+
+* Good, because implementation is relatively simple, and restricted to the client side.
+* Good, because the correct number of exposure events are recorded.
+* Good, because no extra steps needed in post experiment data-analysis.
+* Bad, because it involves a change to the feature JSON, and the user to change their behavior.
+
```
danielkberry, line 138: Is there an Option 4, which is basically option 3 but we automate annotating the message with the slug so that the user doesn't need to do it each time they create an experiment? I'm worried about the manual effort on the user's part, particularly for something like surveys where they'll copy and relaunch surveys periodically and it would be pretty easy for them to forget to update the slug.

shell1: May need to talk to Jared on https://mozilla-hub.atlassian.net/browse/EXP-3602 - which reads like co-enrollment should support other features.  There’s too many options already - with sticky and first run and now multi-feature and rollouts.  I don’t think we’re ready to expose this new highly complex pattern beyond messaging.

The change to allow co-enrollment should be the Messaging feature only.  We should not change onboarding to support co-enrollment - or expose this to other features.   This is a common pattern for messages only.  It’s an exception and one I think we want to avoid for others in the foreseeable future.  There is the one possible case for attribution based onboarding - but that is the exception and may be put on hold depending on [these discussions](https://docs.google.com/document/d/13ZFY7Dhprl5LE7tscdl1x9rAu8IPDJ3MN7mIQJraZBQ/edit#heading=h.5ipiksf3te2q).

shell1: Agree that option 3 looks good, with the caveat of what Daniel commented.  The training and trying to capture human errors is getting overly complex.  The chance for easily done errors is too high - and it would invalidate the experiment learnings if the information isn’t added.

docs/adrs/0012-exposure-events-and-coenrolling-features.md
```diff
@@ -0,0 +1,142 @@
...
+* Status: proposed
+* Deciders: dberry@mozilla.com, jhugman@mozilla.com, jlockhart@mozilla.com, shell@mozilla.com
+* Date: 2023-07-04, making the decision on or before 2023-07-11.
+
+Technical Story: [Supporting exposure events for coenrolled features](https://mozilla-hub.atlassian.net/browse/EXP-3602)
+
+## Context and Problem Statement
+
+Application feature code can be configured from and by Nimbus. In the case of `messaging` and `onboarding`, the complete set of messages and all their associated data is implemented by this feature configuration.
+
```
shell1, line 12: I thought that:
If the default configuration for a feature and no experiment = default config
if a user is enrolled in an experiment - it uses the experiment values + default values for any not defined.  
If a user is enrolled in a rollout - it uses the rollout values + default values for any not defined.
If a user is enrolled in an experiment AND a rollout - it applies all the experiment values + default values for any not applied.

docs/adrs/0012-exposure-events-and-coenrolling-features.md
```diff
@@ -0,0 +1,142 @@
...
+* Status: proposed
+* Deciders: dberry@mozilla.com, jhugman@mozilla.com, jlockhart@mozilla.com, shell@mozilla.com
+* Date: 2023-07-04, making the decision on or before 2023-07-11.
+
+Technical Story: [Supporting exposure events for coenrolled features](https://mozilla-hub.atlassian.net/browse/EXP-3602)
+
+## Context and Problem Statement
+
+Application feature code can be configured from and by Nimbus. In the case of `messaging` and `onboarding`, the complete set of messages and all their associated data is implemented by this feature configuration.
```
shell1, line 11: May need to talk to Jared on https://mozilla-hub.atlassian.net/browse/EXP-3602 - which reads like co-enrollment should support other features. There’s too many options already - with sticky and first run and now multi-feature and rollouts. I don’t think we’re ready to expose this new highly complex pattern beyond messaging.

The change to allow co-enrollment should be the Messaging feature only. We should not change onboarding to support co-enrollment - or expose this to other features. This is a common pattern for messages only. It’s an exception and one I think we want to avoid for others in the foreseeable future. There is the one possible case for attribution based onboarding - but that is the exception and may be put on hold depending on [these discussions](https://docs.google.com/document/d/13ZFY7Dhprl5LE7tscdl1x9rAu8IPDJ3MN7mIQJraZBQ/edit#heading=h.5ipiksf3te2q).

docs/adrs/0012-exposure-events-and-coenrolling-features.md
```diff
@@ -0,0 +1,142 @@
...
+
+The feature configuration is composed by merging together the default configuration, any configuration from a rollout, and any configuration from experiments.
```
shell1, line 13: I thought that:
If the default configuration for a feature and no experiment = default config
if a user is enrolled in an experiment - it uses the experiment values + default values for any not defined.
If a user is enrolled in a rollout - it uses the rollout values + default values for any not defined.
If a user is enrolled in an experiment AND a rollout - it applies all the experiment values + default values for any not applied.

docs/adrs/0012-exposure-events-and-coenrolling-features.md
shell1: @jaredlockhart ^

jhugman: TBC: `allow-coenrollment` isn't covered in this ADR; however: as implemented in EXP-3602: it's in the feature manifest, and I expect will only be used under guidance.

docs/adrs/0012-exposure-events-and-coenrolling-features.md
jhugman: The user adds to the message.

```json
"experiment": "{experiment}"
```

I've proposed that the SDK (or experimenter!) does the replacement of the `{experiment}` with the actual slug, so users don't need to manage slugs themselves, and can add it to their own templates.

docs/adrs/0012-exposure-events-and-coenrolling-features.md
shell1: I'm on board so long as we are aligned that being in the feature manifest, but not exposing in Experimenter.  We need no-decision solution - set default on for Message Feature and default off for all others.

If it requires change - "using under guidance" has surpassed it's limits.  The one off support if we add another decision point and new scenario to other features is too expensive to make them ask me and then I ask eng :).  We need to learn the edges of co-enrollment with these 3 surfaces.  I cannot support onboarding even having it as an option until we have run coenrollment for a while and learned the use patterns and implications.

If we say "onboarding" in this document - it will become an assumed capability to use in the near term.

docs/adrs/0012-exposure-events-and-coenrolling-features.md
danielkberry: But users still need to remember to add that string template to every message? That's certainly better than adding the exact slug, but is there a way for the SDK or experimenter to do the whole process for the user? If we're going to replace the template with the slug, can we just automate the whole thing?

docs/adrs/0012-exposure-events-and-coenrolling-features.md
danielkberry: Does coenrollment apply to both messaging experiments and rollouts?

jhugman: You're right with the rules you've given. The `merge` operation is defined by (pseudo code):

```
fn merge(override: JSONValue, fallback: JSONValue) -> JSONValue:
	when override is None => return fallback,
	when override is JSONObject, fallback is JSONObject =>
			result = clone(fallback)
			foreach key in override.keys():
				result[key] = merge(override[key], fallback[key])
			return result
	otherwise => return override
```

Multiple merges happens, depending on the feature and the experiments currently in play.

For completeness/documentation:

If the default configuration for a feature is:
```json
{
  "a": "default",
  "b": "default",
  "c": "default",
  "M": {}
}
```

and a experiment varies: `"a"`, and adds an `exp` to the `M` object:

```json
{
  "a": "experiment",
  "M": {
	  "exp": 1
  }
}
```

and a rolliout varies "a" and "b", and adds `ro` to the `M` object:

```json
{
   "a": "rollout",
   "b": "rollout",
   "M": {
	  "ro": 2
   }
}
```

Then a user in the experiment gets: 
- Experiment + Defaults, i.e. `merge(experiment, defaults)`
```json
{
  "a": "experiment",
  "b": "default",
  "c": "default",
  "M": {
    "exp": 1
  }
}
```


For a user in the rollout, but not the experiment:
- Rollout + Defaults, i.e. `merge(rollout, defaults)`
```json
{
  "a": "rollout",
  "b": "rollout",
  "c": "default",
  "M": {
    "ro": 2
  }
}
```

And a user in both rollout and experiment gets:
Experiment + Rollout + Defaults, i.e. `merge(experiment, merge(rollout, defaults))`
```json
{
  "a": "experiment",
  "b": "rollout",
  "c": "default",
  "M": {
    "exp": 1,
    "ro": 2,
  }
}
```

For coenrolling features, a feature can be involved in more than one experiment, so the feature config can be:

Experiment 1 + Experiment2 + Rollout1 + Rollout2 + Defaults i.e. 

So if each of the experiments and rollouts contribute something with a different key in `M` , for example:

experiment1
```json
{ "M": { "exp1": 1 } }
```

experiment2
```json
{ "M": { "exp2": 2 } }
```

rollout1
```json
{ "M": { "ro1": 3 } }
```

rollout2
```json
{ "M": { "ro1": 4 } }
```

`merge(merge(merge(experiment1, experiment2), merge(rollout1, rollout2), defaults)` gives:
```json
{
  "a": "default",
  "b": "default",
  "c": "default",
  "M": {
    "exp1": 1,
    "exp2": 2,
    "ro1": 3,
    "ro2": 4
  }
}
```

The ADR asks how do we unpick where does the `2` come from? For `messaging`, the `M` is `messages`, and the integers which are values are objects.

So for `experiment1` may be:
```json
{
	"messages": {
		"my-message-1": {
			"experiment": "{experiment}"
		}
	}
}
```

and `experiment2` may be:
```json
{
	"messages": {
		"my-message-2": {
			"experiment": "{experiment}"
		}
	}
}
```

and it the `{experiment}` that the user typed gets replaced, and merged into:

```json
{
	"messages": {
		"my-message-1": {
			"experiment": "experiment1",
		},
		"my-message-2": {
			"experiment": "experiment2",
		},
	}
}
```

When `my-message-1` is displayed, then we can get the `experiment` property, and find out that it was the experiment with slug `experiment1`  that contributed this message.

We can go further, and replace `{experiment}` in the message key, so the user types in *both* `experiment1` and `experiment2`:

```json
{
	"messages": {
		"{experiment}-en": {
			"experiment": "{experiment}"
		},
		"{experiment}-es": {
			"experiment": "{experiment}"
		},
		"{experiment}-fr": {
			"experiment": "{experiment}"
		}
	}
}
```
and the replace/merge yields:

```json
{
	"messages": {
		"experiment1-en": {
			"experiment": "experiment1"
		},
		"experiment1-es": {
			"experiment": "experiment1"
		},
		"experiment1-fr": {
			"experiment": "experiment1"
		},
		"experiment2-en": {
			"experiment": "experiment2"
		},
		"experiment2-es": {
			"experiment": "experiment2"
		},
		"experiment2-fr": {
			"experiment": "experiment2"
		}
	}
}
```

docs/adrs/0012-exposure-events-and-coenrolling-features.md
eliserichards: I agree. I don't think `experiment` should be optional

eliserichards (commented):
This looks great. I agree that **Option 3** is the best way to go, with the caveat that we should make the addition of `experiment` required to reduce confusion on the users' part. If that can be streamlined, this seems perfect 👍

jaredlockhart (approved):
Okay it sounds like we have consensus on option 3, let's go with that.  Thank you for putting this together @jhugman and @shell1 and @danielkberry and @eliserichards for feeding back 🙏

docs/adrs/0012-exposure-events-and-coenrolling-features.md
jhugman: >  we are aligned that being in the feature manifest

Yes.

> If we say "onboarding" in this document - it will become an assumed capability to use in the near term.

I have removed mention of `onboarding` in this document.

danielkberry:
Based on conversations in Slack, I now understand that my hypothesized alternate would break the current ownership boundary (where experiment teams are solely responsible for their feature schemas) and mean stronger coupling between experimenter (or the SDK) and the messaging feature schema. Experimenter (or the SDK) would need to know and keep up to date with the messaging schema. We don't want that coupling, so cannot implement this option. 

Based on that, I switch my recommendation to Option 3

PR #9070, opened 2023-07-04
Fixes #9056 feat(nimbus): change implements differentiating of changelog values from old ones

### LINKED ISSUE #9056: Implement filtering/diff of changelog events from old values
* Define the functionality for filtering and diffing changelog events
* Implement the functionality in the app's views
* Ensure the filtering and diffing works as expected
* Which value changed from `what` to `To`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3652)

Avi-88:
Because

- The default changelogs aren't specific on what change was made to the experiment

This commit

- Introduces a new method to the NimbusExperiment model which runs a transformation logic on the changelogs to differentiate what field was changed along with old and new values


experimenter/experimenter/experiments/models.py
```diff
@@ -820,7 +821,69 @@ def clone(self, name, user, rollout_branch_slug=None):
...
+            }
+            changes_by_date[first_changelog.changed_on.date()].append(change)
+
+        transformed_changelogs = [
+            {"date": date, "changes": changes}
+            for date, changes in changes_by_date.items()
+        ]
+        sorted_changelogs = sorted(
+            transformed_changelogs, key=lambda x: x["date"], reverse=True
+        )
```
jaredlockhart, line 884: So here we're sorting by the outer key `date`, but not sorting the inner changes, so that list can be in any order, which means two things

1) They could appear in any order when rendering the page so they could jump around
2) We can't make any strong assertions about the ordering in the tests

We do know the exact times these changes were made so we can explicitly order them by `changed_on`, and it might also be nice to show the time in the UI, so it might be worth including it in the dict and then sorting them here.

experimenter/experimenter/templates/changelog/overview.html
```diff
@@ -9,9 +9,13 @@ <h1>Changes Overview</h1>
     <p>Experiment Slug: {{ experiment.slug }}</p>
     {% for changelog in experiment.get_changelogs %}
```
jaredlockhart, line 10: So up here, this isn't really a changelog anymore, now it's a group of changelogs and a date?

jaredlockhart (changes requested):
Looking really great @Avi-88 this is all on the right track.  I put a bunch of thoughts but overall I'd say this is exactly how I'd expect it to look.  Great job 🎉

experimenter/experimenter/experiments/models.py
Avi-88: Oh right,  I'll update it accordingly

experimenter/experimenter/templates/changelog/overview.html
Avi-88: Yes, should I reference it differently ?

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2180,10 +2180,43 @@ def _clone_experiment_and_assert_common_expectations(
...
+
+        generate_nimbus_changelog(experiment, user, "created")
+
+        experiment.status = NimbusExperiment.Status.DRAFT
+        experiment.status_next = NimbusExperiment.Status.LIVE
+        experiment.publish_status = NimbusExperiment.PublishStatus.REVIEW
+        experiment.published_dto = {"id": "experiment", "test": True}
+        experiment.save()
+
+        generate_nimbus_changelog(experiment, user, "test message")
```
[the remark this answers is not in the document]
Avi-88: Yeah, I was a bit confused on how to generate the changelogs and make them so that they are reported on different date but couldn't really figure out how I could achieve that. I'll look into it more and try and make the test more robust

experimenter/experimenter/templates/changelog/overview.html
jaredlockhart: Yeah I would update the name to reflect what is being looped over.

experimenter/experimenter/experiments/models.py
Avi-88: So basically with the updated logic , the changes will be stored by latest first as well. I've updated the test case accordingly. It required making a small change to the existing `generate_changelog` method, could you please check if its correct ?

experimenter/experimenter/experiments/changelog_utils.py
```diff
@@ -64,16 +64,32 @@ def generate_nimbus_changelog(experiment, changed_by, message):
...
+            old_status_next=old_status_next,
+            old_publish_status=old_publish_status,
+            new_status=experiment.status,
+            new_status_next=experiment.status_next,
+            new_publish_status=experiment.publish_status,
+            changed_by=changed_by,
+            experiment_data=experiment_data,
+            published_dto_changed=published_dto_changed,
+            message=message,
+        )
```
yashikakhurana, line 95: why if else?

yashikakhurana (changes requested):

experimenter/experimenter/experiments/changelog_utils.py
Avi-88: The `changed_on` parameter was just added for a single test and I didn't want to break the existing test cases. Please do let me know if there was a better way of handling this and I'll update it accordingly.

brennie: You shouldn't need the if-else since the value is `None`.

experimenter/experimenter/experiments/models.py
```diff
@@ -820,7 +821,58 @@ def clone(self, name, user, rollout_branch_slug=None):
...
+
+                diff_fields = {
+                    field: {
+                        "old_value": previous_data.get(field),
+                        "new_value": current_data.get(field),
+                    }
+                    for field in current_data
+                    if field != "_updated_date_time"
+                    and field != "published_dto"
+                    and current_data[field] != previous_data.get(field)
```
brennie, line 843: Could you format this as:

```
if (
    ... and
    ... and
    ...
)

File: experimenter/experimenter/experiments/models.py [hunk 6]
```diff
@@ -820,7 +821,58 @@ def clone(self, name, user, rollout_branch_slug=None):
...
 
     def get_changelogs(self):
-        return self.changes.all()
+        changes_by_date = defaultdict(list)
+        date_option = "%I:%M:%S %p"
+        changelogs = self.changes.order_by("-changed_on").prefetch_related("changed_by")
+        size = len(changelogs) - 1
+
+        for index, changelog in enumerate(changelogs):
+            if index < size:
```
brennie, line 830: You can use slice notation to omit the last changelog, and then you don't need `size`:

```python
for index, changelog in enumerate(changelogs[:-1]):
```

experimenter/experimenter/experiments/models.py
```diff
@@ -820,7 +821,58 @@ def clone(self, name, user, rollout_branch_slug=None):
         return cloned
 
     def get_changelogs(self):
```
brennie, line 823: Can we rename this to `get_changes_by_date` ?

experimenter/experimenter/templates/changelog/overview.html
```diff
@@ -9,9 +9,17 @@ <h1>Changes Overview</h1>
...
     <div>
-        <h2>{{ changelog.changed_on }}</h2>
-        <p><i class="fa-regular fa-pen-to-square"></i> : {{ changelog.message }}</p>  
-        <p><i class="fa-regular fa-user"></i> : {{ changelog.changed_by }}</p>
+        <h2>{{ changelog.date }}</h2>
+        {% for change in changelog.changes %}
+        <div>
+          {% if change.event == "GENERAL" %}
+            <p>Change Log: {{ change.changed_by}} changed the value of {{ change.field }} from {{ change.old_value }} to {{ change.new_value}}</p>
```
brennie, line 16: Space after the variable name inside {{}}

brennie (changes requested):

experimenter/experimenter/experiments/models.py
Avi-88: The negative indexing isn't supported in `queryset` so had to put it in a list before slicing, please let me know If i missed anything

experimenter/experimenter/experiments/changelog_utils.py
```diff
@@ -48,10 +49,13 @@ class Meta:
...
 
 
-def generate_nimbus_changelog(experiment, changed_by, message):
+def generate_nimbus_changelog(experiment, changed_by, message, changed_on=None):
     latest_change = experiment.changes.latest_change()
     experiment_data = dict(NimbusExperimentChangeLogSerializer(experiment).data)
 
+    if not changed_on:
+        changed_on = timezone.now()
+
```
yashikakhurana, line 58: 👍

experimenter/experimenter/experiments/models.py
```diff
@@ -836,7 +837,61 @@ def clone(self, name, user, rollout_branch_slug=None):
...
-        return self.changes.all()
+        changes_by_date = defaultdict(list)
+        date_option = "%I:%M:%S %p"
+        changelogs = list(
+            self.changes.order_by("-changed_on").prefetch_related("changed_by")
+        )
+        len(changelogs) - 1
```
yashikakhurana, line 845: we are not storing it in a variable or something, if you don't want to use it we can remove this line

experimenter/experimenter/templates/changelog/overview.html
```diff
@@ -9,9 +9,17 @@ <h1>Changes Overview</h1>
...
+        <h2>{{ changelog.date }}</h2>
+        {% for change in changelog.changes %}
+        <div>
+          {% if change.event == "GENERAL" %}
+            <p>Change Log: {{ change.changed_by }} changed the value of {{ change.field }} from {{ change.old_value }} to {{ change.new_value }}</p>  
+          {% elif change.event == "CREATION" %}
+            <p>Change Log: {{ change.changed_by }} created this experiment</p>  
+          {% endif %}
+          <p>Changed at: {{ change.timestamp }}</p>
+        </div>
```
yashikakhurana, line 21: you can avoid putting logic here and can put a key like `message` and build message there because later we will be having more events like review etc etc, so it doesn't make sense to have all logic processing here

yashikakhurana (dismissed):
thank you @Avi-88 it is coming together, few minor suggestions :)

experimenter/experimenter/templates/changelog/overview.html
yashikakhurana: I noticed we already use `message` so you can use like `event_message`

Avi-88: So the reason I made them into separate fields is to make the implementation more flexible, like different styling for the values, the user or the timestamp  so that they are more distinguished

experimenter/experimenter/experiments/models.py
```diff
@@ -835,8 +836,61 @@ def clone(self, name, user, rollout_branch_slug=None):
...
+                    "event": "GENERAL",
+                    "old_value": field_diff["old_value"],
+                    "new_value": field_diff["new_value"],
+                    "changed_by": changelog.changed_by.get_full_name(),
+                    "timestamp": timestamp,
+                }
+
+                changes_by_date[changelog.changed_on.date()].append(change)
+
+        if creation_log := changelogs[-1] if changelogs else None:
```
brennie, line 876: Having a Python ternary inside an if is weird, I've never seen that before.

I would probably format this as:

```python
if changelogs:
  creation_log = changelogs[-1]
  # ...
```

I'm assuming that `changelogs[-1]` won't be falsy -- it shouldn't need a second if to check it.

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2230,10 +2230,73 @@ def _clone_experiment_and_assert_common_expectations(
...
-        generate_nimbus_changelog(experiment, experiment.owner, "Test message")
-        experiment_changelogs = experiment.get_changelogs()
-        self.assertEqual(experiment_changelogs[0].message, "Test message")
+        experiment = NimbusExperimentFactory.create(
+            slug="experiment-1",
+            published_dto={"id": "experiment", "test": False},
+        )
+        user = UserFactory.create()
+        time_format = "%I:%M:%S %p"
+        current_date = datetime.date.today()
```
brennie, line 2239: All datetimes that end up in the databse should use `django.utils.timezone` so that they are time-zone aware. datetime.datetime and datetime.date will be created without tzinfo.

If you insert a datetime.datetime into the DB without creating it via `django.utils.timezone` and then get the value out of the db, the values will be different because the one retrieved from the DB will be timezone-aware.

This means that tests that create a date and then check if its equal when it comes out of the database will sometimes fail when running them later in the day.

[docs](https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/)

Avi-88: I saw the test cases using datetime so I just followed that, I've updated it to use timezone now :)

brennie (approved):
Excellent! This looks great!

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2230,10 +2231,79 @@ def _clone_experiment_and_assert_common_expectations(
         return child
 
     def test_get_changelogs(self):
-        experiment = NimbusExperimentFactory.create(slug="experiment-1")
-        generate_nimbus_changelog(experiment, experiment.owner, "Test message")
-        experiment_changelogs = experiment.get_changelogs()
-        self.assertEqual(experiment_changelogs[0].message, "Test message")
+        experiment = NimbusExperimentFactory.create(
```
eliserichards, line 2234: I think I would separate this into a few tests:
1. Get changelogs when none exist
2. Get changelogs returns as expected
3. Get changelogs returns multiple (which is this existing test)

eliserichards (changes requested):

yashikakhurana (approved):
thank you @Avi-88 looks good to me, feel free to add test cases as @eliserichards suggested

experimenter/experimenter/experiments/tests/test_models.py
Avi-88: I can add for the none test case, but I'm a bit confused between 2 and 3. So for the expected one should I just have it return a single change and test its format ?

eliserichards: The difference between 2 and 3 is that test 2 checks that "this should return something" (it is the opposite of test 1) but it doesn't care **what** is returned. Test 3 checks **what is returned** (it is formatted correctly? how many items are returned? etc). You don't HAVE to have test 2, it's just a good excuse to write more tests ;)

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2229,11 +2230,111 @@ def _clone_experiment_and_assert_common_expectations(
                 )
         return child
 
+    def test_get_changelogs_without_prior_change(self):
```
eliserichards, line 2233: ✨ perfect! ✨

eliserichards (approved):

PR #9079, opened 2023-07-06
Add targeting for the Challenge the Default experiments

AllegroFox:
Add targeting for the [Challenge the Default](https://mozilla-hub.atlassian.net/browse/FXE-295) experiment that includes the paidsearch attribution.


eliserichards (approved):
🚀

PR #9085, opened 2023-07-07
fix #9084 bug(nimbus): set published_dto to None in v5 serializer

### LINKED ISSUE #9084: Set published_dto to None when moving back to draft
When an experiment is sent back to Draft from Preview, we need to clear out the `published_dto` so that certain methods don't display it on the summary page in case anything about the experiment is later modified.  Presently, we set `published_dto = None` inside the preview sync task, however, that task can fail if Remote Settings becomes unavailable and then published_dto will never be set back to `None`.  We should instead set it to None right in the V5 serializer at save time so that it is set back to None reliably.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3666)

jaredlockhart:
Because

* We cache the published dto on the experiment model on a field called published_dto
* This field is set when we move from Draft to Preview or Live
* When we move from Preview back to Draft, we want to set that field back to None
* Right now that takes place in the kinto preview sync task
* That task can fail for different reasons, but then we can wind up with an experiment in Draft with a published_dto
* The summary page uses the published_dto and so it can appear as though the recipe is out of sync if changes are made to the experiment

This commit

* Sets published_dto to None in the V5 serializer at status change time
* Removes the published_dto change from the kinto task
* Adds a data migration to set published_dto to None for any experiment in Draft


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -877,17 +877,18 @@ class Meta:
...
+            and data
+            and (data.get("status") == NimbusExperiment.Status.PREVIEW)
+        )
+        self.is_preview_to_draft = instance and (
+            instance.status == NimbusExperiment.Status.PREVIEW
+            and data
+            and (data.get("status") == NimbusExperiment.Status.DRAFT)
+        )
+        self.should_call_preview_task = (
+            self.is_draft_to_preview or self.is_preview_to_draft
```
yashikakhurana, line 891: 🎉

yashikakhurana (approved):
oh I see, this was the reason regarding - we found an issue with experimenter this week https://mozilla-hub.atlassian.net/browse/EXP-3657. On stage at least we aren't able to make any changes to the audience section after an experiment is moved back to draft after being in preview. Or rather, we can make changes, but the JSON recipe is not updated and pushing to Preview again does not trigger an update.?

PR #9090, opened 2023-07-10
feat(nimbus): Enable multi-feature UI

brennie:
Because
- multi-feature support in clients has been QA'd

this commit
- enables the multi-feature UI in the console.

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
@@ -205,9 +205,6 @@ export const FormBranches = ({
     setIsSelectValid(newValue.length > 0);
     setSelectDirty(true);
 
-    // Limit to a single value until multi-feature support has been QA'd.
-    newValue = newValue.length > 0 ? [newValue[newValue.length - 1]] : newValue;
-
     return handleFeatureConfigsChange(
       newValue.map((value) => parseInt(value.value, 10)),
     );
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
Do you want a test to ensure that duplicate features cannot be selected? Not sure what the UI looks like or if this is even possible, but that's all I can think of. Otherwise looks good to me!

brennie:
> Do you want a test to ensure that duplicate features cannot be selected? Not sure what the UI looks like or if this is even possible, but that's all I can think of. Otherwise looks good to me!

We have tests for this already in the backend: https://github.com/mozilla/experimenter/blob/4966ec833253ea1d2878e646622f5af24aa354f7/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_branch_serializer.py#L170

The frontend component doesn't let you select the same object twice (it removes the DOM element once its selected)

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/index.tsx
```diff
@@ -205,9 +205,6 @@ export const FormBranches = ({
     setIsSelectValid(newValue.length > 0);
     setSelectDirty(true);
 
-    // Limit to a single value until multi-feature support has been QA'd.
-    newValue = newValue.length > 0 ? [newValue[newValue.length - 1]] : newValue;
-
```
jaredlockhart, line 210: BYYYYEEEEE 👋 👋 👋 👋 👋 👋 👋

PR #9091, opened 2023-07-10
fix #9082 bug(nimbus): validate no floats in feature values

### LINKED ISSUE #9082: Kinto does not accept float values
Here on stage we see that kinto will reject a record that includes a float value:

https://mozilla.sentry.io/issues/4264507499/?project=6377194&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

We do not currently validate that feature values do not contain any float values, and so Experimenter will allow a float value but then publishing to kinto will fail and break.  We should validate that feature values never contain floats.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3661)

jaredlockhart:
Because

* We recently discovered that kinto does not support float values in json objects
* This is causing the preview sync task on stage to get stuck
* Luckily we have not seen it in production

This commit

* Adds a validation step to ensure there are no floats in any feature values before launching

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1169,11 +1169,28 @@ class NimbusBranchFeatureValueReviewSerializer(NimbusBranchFeatureValueSerialize
         list_serializer_class = NimbusBranchFeatureValueListSerializer
 
     def validate_value(self, value):
+        data = None
         if value:
             try:
-                json.loads(value)
+                data = json.loads(value)
             except Exception as e:
                 raise serializers.ValidationError(f"Invalid JSON: {e.msg}") from e
+
+        def throw_on_float(item):
+            if isinstance(item, (list, tuple)):
+                for i in item:
+                    throw_on_float(i)
+            elif isinstance(item, dict):
+                for i in item.values():
+                    throw_on_float(i)
+            elif isinstance(item, float):
+                raise serializers.ValidationError(
+                    NimbusExperiment.ERROR_NO_FLOATS_IN_FEATURE_VALUE
+                )
+
+        if data is not None:
+            throw_on_float(data)
+
         return value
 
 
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1174,6 +1174,23 @@ def validate_value(self, value):
...
+
+        def find_values(item):
+            values = []
+            if isinstance(item, (list, tuple)):
+                [values.extend(find_values(i)) for i in item]
+            elif isinstance(item, dict):
+                [values.extend(find_values(i)) for i in item.values()]
+            else:
+                values = [item]
+            return values
```
brennie, line 1186: Why build up this whole list? Why not:

```
        def throw_on_float(item):
            if isinstance(item, (list, tuple)):
                find_values(i)) for i in item
            elif isinstance(item, dict):
                values.extend(find_values(i)) for i in item.values()
            elif isinstance(item, float):
                     NimbusExperiment.ERROR_NO_FLOATS_IN_FEATURE_VALUE
                )

        throw_on_float(json.loads(data))
```

jaredlockhart: Oh yeah could do that too.

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1174,6 +1174,19 @@ def validate_value(self, value):
                 json.loads(value)
             except Exception as e:
                 raise serializers.ValidationError(f"Invalid JSON: {e.msg}") from e
+
+        def throw_on_float(item):
+            if isinstance(item, (list, tuple)):
+                [throw_on_float(i) for i in item]
+            elif isinstance(item, dict):
+                [throw_on_float(i) for i in item.values()]
```
brennie, line 1182: You're still making lists (my original comment was wrong, I updated it to use loops)

jaredlockhart: Technically yes but it's so minor 🤷‍♀️

brennie (approved):

PR #9097, opened 2023-07-10
fix #9043 feat(nimbus): Add inclusion and exclusion targeting UI

### LINKED ISSUE #9043: Inclusion/Exclusion Targeting UI

brennie:
Because:
- we want to target against past and present enrollments in other
  experiments and rollouts

this commit:
- adds UI to select multiple experiments to include and exclude in the
  generated targeting expression.

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
@@ -10,7 +10,10 @@ import { displayConfigLabelOrNotSet } from "src/components/Summary";
 import { useConfig } from "src/hooks";
 import { ReactComponent as CollapseMinus } from "src/images/minus.svg";
 import { ReactComponent as ExpandPlus } from "src/images/plus.svg";
-import { getExperiment_experimentBySlug } from "src/types/getExperiment";
+import {
+  getExperiment_experimentBySlug,
+  getExperiment_experimentBySlug_excludedExperiments,
+} from "src/types/getExperiment";
 import { NimbusExperimentApplicationEnum } from "src/types/globalTypes";
 
 type TableAudienceProps = {
@@ -160,6 +163,16 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
                 </td>
               </tr>
             )}
+            <tr>
+              <th>Required Experiments</th>
+              <td>
+                <ExperimentList experiments={experiment.requiredExperiments} />
+              </td>
+              <th>Excluded Experiments</th>
+              <td>
+                <ExperimentList experiments={experiment.excludedExperiments} />
+              </td>
+            </tr>
 
             {experiment.jexlTargetingExpression &&
             experiment.jexlTargetingExpression !== "" ? (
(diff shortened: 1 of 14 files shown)


brennie:
Blocked on #9044

jaredlockhart (changes requested):
Looking good.  The layout of the drop downs is especially nice 🙏 

Couple things
1) We need to add the required/excluded experiments to the audience section of the summary page
2) We should remove selected options from the other list so the same experiment can not be selected for both required and excluded (same pattern we do on the metrics page between primary and secondary)
3) We should also validate on the backend that the same experiment can not be both required and excluded filed https://github.com/mozilla/experimenter/issues/9123 to address separately

brennie:
> Looking good. The layout of the drop downs is especially nice 🙏
> 
> Couple things
> 
>     1. We need to add the required/excluded experiments to the audience section of the summary page

Done

>     2. We should remove selected options from the other list so the same experiment can not be selected for both required and excluded (same pattern we do on the metrics page between primary and secondary)

Done

>     3. We should also validate on the backend that the same experiment can not be both required and excluded filed [Validate an experiment can not be both required and excluded #9123](https://github.com/mozilla/experimenter/issues/9123) to address separately

As I noted in #9123, this is already done in the previous PR.

experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
```diff
@@ -230,4 +243,23 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
...
+  if (experiments.length === 0) {
+    return <span>None</span>;
+  }
+
+  return (
+    <ul className="m-0">
+      {experiments.map((experiment) => (
+        <li key={experiment.slug}>
+          <a href={`/nimbus/${experiment.slug}/summary`}>{experiment.slug}</a>
+        </li>
```
jaredlockhart, line 259: Oh good yes this is helpful, but the spacing and dot with the `<li>`s takes so much space, can we put these in `<p>`s instead?  And use the name instead of the slug? 🙏

jaredlockhart (changes requested):
Oh good yes, the filtering to prevent using both works perfectly and thank you for adding to summary page 🙏

experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
brennie: Done!

jaredlockhart (approved):
YESS!!!  Wow looks fantastic, works fantastic, the multiselect entries look beautiful (I wish all our multiselects looked this good), the cross field filtering works perfectly, table summary looks great, targeting string looks great, version error works perfectly.  This is great this is one of those smallish but huge impact type features, this unlocks so much capability we should've had for so long.  Thank you @brennie this is awesome 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #9110, opened 2023-07-12
fix #9109 feat(nimbus): Model changes for feature versioning

### LINKED ISSUE #9109: Model changes for feature versioning
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3647)

brennie:
Because:
- we want to support feature configurations across multiple versions of
  Firefox; and
- `NimbusFeatureConfig` objects were conditionally read-only in the
  database based on the `read_only` field

this commit:
- separates the schema out of NimbusFeatureConfig into its own model:
  NimbusVersionedSchema; and
- removes the `read_only` field and makes `NimbusFeatureConfig`s
  unconditionally read-only.


experimenter/experimenter/experiments/admin.py
```diff
@@ -328,14 +329,20 @@ class NimbusExperimentAdmin(
...
     exclude = ("read_only",)
 
     def has_change_permission(self, request, obj: Optional[NimbusFeatureConfig] = None):
         return obj and not obj.read_only
 
 
+class NimbusVersionedFeatureConfigAdmin(
+    NoDeleteAdminMixin, admin.ModelAdmin[NimbusVersionedSchema]
+):
+    pass
```
eliserichards, line 343: (I'm not super familiar with the admin stuff) what is this class doing? Why do we need a second one for the versioned configs?

experimenter/experimenter/experiments/changelog_utils.py
```diff
@@ -10,10 +10,15 @@
...
 
 class NimbusFeatureConfigChangeLogSerializer(serializers.ModelSerializer):
+    schema = serializers.SerializerMethodField()
+
     class Meta:
         model = NimbusFeatureConfig
         exclude = ("id",)
 
+    def get_schema(self, obj):
+        return obj.schemas.get(version=None).schema
```
eliserichards, line 20: Can we have the schema's `version` default to `None` so we don't have to explicitly set this everywhere?

experimenter/experimenter/experiments/admin.py
brennie: The `NoDeleteAdminMixin` prevents model instances from being deleted in the admin panel. 
The `NimbusVersionedFeatureConfig` the `ModelAdmin` class for `NimbusVersionedSchema`s (the name of the class is wrong -- search and replace didnt get this one) that tells the django admin panel about this model and how to display its models and forms, etc.

experimenter/experimenter/experiments/changelog_utils.py
brennie: We can't, django adds the .get method for one-to-many relationships on the field.

Also, a lot of this code that is querying for the `None`-versioned schema will be going away once we actually have versioned schemas in the DB. Right now, all schemas are being imported without a version, but that will change in future patches.

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -20,39 +23,39 @@ def handle(self, *args, **options):
...
 
         for feature in Features.all():
-            feature_config, created = NimbusFeatureConfig.objects.get_or_create(
+            feature_config, created = NimbusFeatureConfig.objects.update_or_create(
                 slug=feature.slug,
                 application=feature.applicationSlug,
+                defaults={
+                    "name": feature.slug,
+                    "description": feature.description,
+                },
```
jaredlockhart, line 32: Neat!

experimenter/experimenter/experiments/admin.py
```diff
@@ -326,16 +327,22 @@ class NimbusExperimentAdmin(
...
     exclude = ("read_only",)
 
     def has_change_permission(self, request, obj: Optional[NimbusFeatureConfig] = None):
         return obj and not obj.read_only
 
 
+class NimbusVersionedSchemaAdmin(
+    ReadOnlyAdminMixin, admin.ModelAdmin[NimbusVersionedSchema]
+):
+    pass
```
jaredlockhart, line 343: Oh we should add it as an inline into the feature config admin so we can still see it all in one place.

jaredlockhart (changes requested):
Wow, this touched a lot of things.  Good thing making a change like this is so much saner in the backend than the frontend.  Good job keeping the API interface the same.  Migration and test look good.  Changes to the feature loader look good.

I try not to be a stickler about naming in general, but using the past tense in a class name stands out to me.  Can we use something in the present tense like `NimbusSchema` or `NimbusFeatureConfigSchema` or `NimbusSchemaVersion`? 🙏 

But yeah this all looks great thank you 🎉 🎉 🎉 🎉

brennie:
> Wow, this touched a lot of things. Good thing making a change like this is so much saner in the backend than the frontend. Good job keeping the API interface the same. Migration and test look good. Changes to the feature loader look good.
> 
> I try not to be a stickler about naming in general, but using the past tense in a class name stands out to me. Can we use something in the present tense like `NimbusSchema` or `NimbusFeatureConfigSchema` or `NimbusSchemaVersion`? 🙏

Here "versioned" is an adjective, not really a past-tense verb. Consider a "versioned schema" vs an "unversioned schema". A "schema with a version" is "versioned".

> 
> But yeah this all looks great thank you 🎉 🎉 🎉 🎉

jaredlockhart (approved):
Something still feels off about using 'versioned' in a class name, but yep everything looks good thnx @brennie ! 🎉

PR #9115, opened 2023-07-13
fix #9073 feat(cirrus): Set value using glean metrics

### LINKED ISSUE #9073: Record Enrollment metrics
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3572)

yashikakhurana:
Because

- We want to set metrics and send pings

This commit

- Set metrics and sends pings using the defined `metrics.yaml` and `pings.yaml`
- Test cases to verify the changes


cirrus/server/tests/conftest.py
```diff
@@ -51,3 +52,114 @@ def fml():
...
+
+
+@fixture(name="reset_glean", scope="function", autouse=True)
+def fixture_reset_glean():
+    testing.reset_glean(application_id="cirrus-test", application_version="0.1.0")
+
+
+@fixture
+def get_telemetry():
+    return initialize_glean()
```
travis79, line 64: Resetting glean via the reset API also initializes Glean, so initializing it before a test could be accomplished by just calling that API. There is also an [optional parameter](https://github.com/mozilla/glean/blob/4198786204fa19c041bdefe7abb91ddc193747db/glean-core/python/glean/testing/__init__.py#L27) to the reset function if you need to re-initialize Glean without clearing the stored information.

travis79 (approved):
This looks good to me. I did give one suggestion about just using `reset_glean` for initializing during tests but no need to block on that.

cirrus/server/cirrus/main.py
```diff
@@ -109,6 +110,41 @@ def initialize_glean():
...
+        else ""
+    )
+    events = enrollment_id = enrolled_partial_configuration.get("events", [])
+    enrollment_id = events[0].get("enrollment_id") if events else ""
+    experiment_type = (
+        app.state.remote_setting.get_recipe_type(experiment_slug)
+        if experiment_slug
+        else ""
+    )
+    return experiment_slug, branch_slug, enrollment_id, experiment_type
```
jeddai, line 128: To collate the enrollment data into the data that should be sent through Glean you'll actually want it to iterate across all the enrollments, filter them down to only those with the `Enrolled` status, and return a list of enrollment extra dicts (or a specific type if you'd prefer).

```suggestion
def get_metrics_data(enrolled_partial_configuration: Dict[str, Any]):
    result = []
    enrollments = enrolled_partial_configuration.get("enrollments", [])
    
    for enrollment in enrollments:
        # Check if status is `Enrolled`
        # Find enrollment event in `events` to get enrollment uuid
        # Put together dict/object with values noted above
        # Append dict to `result`
        pass
        
    return result
```

jeddai (changes requested):
This is looking great from a Glean SDK perspective! I have some comments about how it is set up to collate the Enrollment metrics from the partial config data but this is shaping up very nicely!

yashikakhurana:
closed in favor of https://github.com/mozilla/experimenter/pull/9140

PR #9120, opened 2023-07-13
fix #9119 feat(nimbus): Update minimum version warning to new version and add warning to mobile apps

### LINKED ISSUE #9119: Update rollout minimum version warning
Update the rollout minimum version warning to show for versions under 115 for Desktop, and versions under 116 for mobile apps.

https://mozilla-hub.atlassian.net/browse/EXP-3485

jeddai:
Because

- Rollout un-enrollment and re-enrollment issues were recently fixed in the desktop and mobile versions of Firefox

This commit

- Updates the rollout min-version warning to show for mobile applications with a version lower than fx116
- Updates the rollout min-version warning to show for fx desktop versions lower than fx115

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1482,17 +1482,20 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
                 }
             )
 
-        if (
-            application == NimbusExperiment.Application.DESKTOP
-            and is_rollout
-            and parsed_min_version
-            < NimbusExperiment.Version.parse(
-                NimbusConstants.DESKTOP_ROLLOUT_MIN_SUPPORTED_VERSION
+        rollout_live_resize_min_app_version = (
+            NimbusConstants.ROLLOUT_LIVE_RESIZE_MIN_SUPPORTED_VERSION.get(application)
+        )
+        if rollout_live_resize_min_app_version:
+            parsed_min_app_version = NimbusExperiment.Version.parse(
+                rollout_live_resize_min_app_version
             )
-        ):
-            self.warnings["firefox_min_version"] = [
-                NimbusConstants.ERROR_DESKTOP_ROLLOUT_VERSION
-            ]
+            if is_rollout and parsed_min_version < parsed_min_app_version:
+                self.warnings["firefox_min_version"] = [
+                    NimbusConstants.ERROR_ROLLOUT_VERSION.format(
+                        application=NimbusExperiment.Application(application).label,
+                        version=parsed_min_app_version,
+                    )
+                ]
 
         excluded_experiments = data.get("excluded_experiments")
         required_experiments = data.get("required_experiments")
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -371,7 +371,13 @@ class NimbusConstants(object):
     }
 
     FEATURE_ENABLED_MIN_UNSUPPORTED_VERSION = Version.FIREFOX_104
-    DESKTOP_ROLLOUT_MIN_SUPPORTED_VERSION = Version.FIREFOX_114
+    ROLLOUT_LIVE_RESIZE_MIN_SUPPORTED_VERSION = {
+        Application.DESKTOP: Version.FIREFOX_115,
+        Application.FENIX: Version.FIREFOX_116,
+        Application.FOCUS_ANDROID: Version.FIREFOX_116,
+        Application.IOS: Version.FIREFOX_116,
+        Application.FOCUS_IOS: Version.FIREFOX_116,
+    }
 
     ROLLOUT_SUPPORT_VERSION = {
         Application.DESKTOP: Version.FIREFOX_105,
@@ -437,9 +443,10 @@ Optional - We believe this outcome will <describe impact> on <core metric>
         will be enrolled in one and not the other and \
         you will not be able to adjust the sizing for this rollout."
 
-    ERROR_DESKTOP_ROLLOUT_VERSION = "WARNING: Decreasing the population size while the \
-        rollout is live is not supported for Desktop versions under 114. You will still \
-        be able to increase the population size."
+    ERROR_ROLLOUT_VERSION = (
+        "WARNING: Adjusting the population size while the"
+        "rollout is live is not supported for {application} versions under {version}."
+    )
 
     ERROR_DESKTOP_LOCALIZATION_VERSION = (
         "Firefox version must be at least 113 for localized experiments."
--- a/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -232,10 +232,20 @@ class TestNimbusReviewSerializerSingleFeature(TestCase):
             },
         )
 
-    def test_desktop_min_version_under_113_shows_warning(self):
+    @parameterized.expand(
+        [
+            NimbusExperiment.Application.DESKTOP,
+            NimbusExperiment.Application.FENIX,
+            NimbusExperiment.Application.FOCUS_ANDROID,
+            NimbusExperiment.Application.IOS,
+            NimbusExperiment.Application.FOCUS_IOS,
+        ]
+    )
+    def test_rollout_min_version_under_115_shows_warning(self, application):
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.CREATED,
-            firefox_min_version=NimbusExperiment.Version.FIREFOX_106,
+            firefox_min_version=NimbusExperiment.Version.FIREFOX_114,
+            application=application,
             is_rollout=True,
             is_sticky=True,
         )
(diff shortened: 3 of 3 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1483,15 +1483,23 @@ def _validate_versions(self, data):
...
+                NimbusConstants.ROLLOUT_MIN_SUPPORTED_VERSION[application]
             )
         ):
             self.warnings["firefox_min_version"] = [
-                NimbusConstants.ERROR_DESKTOP_ROLLOUT_VERSION
+                NimbusConstants.ERROR_ROLLOUT_VERSION.format(
+                    application=NimbusExperiment.Application(application).label,
+                    version=NimbusExperiment.Version.parse(
+                        NimbusConstants.ROLLOUT_MIN_SUPPORTED_VERSION[application]
+                    ),
```
jeddai, line 1501: Thinking I may refactor the parsed min-supported version out into a variable too since it's used twice now

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -232,75 +232,79 @@ def test_invalid_experiment_requires_min_version_less_than_max_version(self):
...
-            is_rollout=True,
-            is_sticky=True,
-        )
-        for branch in experiment.treatment_branches:
-            branch.delete()
-        experiment.save()
-        serializer = NimbusReviewSerializer(
-            experiment,
-            data=NimbusReviewSerializer(
+    def test_rollout_min_version_under_115_shows_warning(self):
```
jeddai, line 235: Would love some feedback on this test, context in code comments

experimenter/experimenter/experiments/constants.py
```diff
@@ -371,7 +371,13 @@ class EmailType(models.TextChoices):
...
-    DESKTOP_ROLLOUT_MIN_SUPPORTED_VERSION = Version.FIREFOX_114
+    ROLLOUT_MIN_SUPPORTED_VERSION = {
+        Application.DESKTOP: Version.FIREFOX_115,
+        Application.FENIX: Version.FIREFOX_116,
+        Application.FOCUS_ANDROID: Version.FIREFOX_116,
+        Application.IOS: Version.FIREFOX_116,
+        Application.FOCUS_IOS: Version.FIREFOX_116,
+    }
 
     ROLLOUT_SUPPORT_VERSION = {
```
eliserichards, line 382: I think we can get rid of this variable now if we're defining the min for all applications in `ROLLOUT_MIN_SUPPORTED_VERSION`

eliserichards: Note to self: we'll need to update the docs for this https://experimenter.info/deep-dives/experimenter/rollouts#supported-platforms-and-minimum-version-targeting

experimenter/experimenter/experiments/constants.py
```diff
@@ -437,9 +443,8 @@ class EmailType(models.TextChoices):
         will be enrolled in one and not the other and \
         you will not be able to adjust the sizing for this rollout."
 
-    ERROR_DESKTOP_ROLLOUT_VERSION = "WARNING: Decreasing the population size while the \
-        rollout is live is not supported for Desktop versions under 114. You will still \
-        be able to increase the population size."
+    ERROR_ROLLOUT_VERSION = "WARNING: Adjusting the population size while the \
+        rollout is live is not supported for {application} versions under {version}."
```
eliserichards, line 447: _chefs kiss_

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
eliserichards: Instead of doing the `for` loop for the applications, you can pass them in using `parameterize.expand()` ([example](https://github.com/mozilla/experimenter/blob/41fe48a884c338ae073711d7f3e62b0a6767d848/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py#L385))

eliserichards:
```py
@parameterized.expand(
        [
            (NimbusExperiment.Application.DESKTOP,),
            (NimbusExperiment.Application.FENIX,),
            (NimbusExperiment.Application.FOCUS_ANDROID,),
            (NimbusExperiment.Application.IOS,),
            (NimbusExperiment.Application.FOCUS_IOS,),
        ]
 )
def test_rollout_min_version_under_115_shows_warning(self, application):
     experiment = ...
         application=application,
         ...
```

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -232,75 +232,79 @@ def test_invalid_experiment_requires_min_version_less_than_max_version(self):
...
+            self.assertTrue(serializer.is_valid())
+            self.assertEqual(
+                serializer.warnings,
+                {
+                    "firefox_min_version": [
+                        # Ehhhhhh on this whole thing here? It's basically the same logic that exists in the actual code
+                        # I'm not a fan of it but idk how else to do it because I don't want to write the entire text out
+                        # Should I just write a separate test for each app?
+                        NimbusExperiment.ERROR_ROLLOUT_VERSION.format(
+                            application=NimbusExperiment.Application(application).label,
```
eliserichards, line 271: yeahhh matching this stuff is a pita. This looks fine to me; maybe Yashika/Jared will have some better suggestions 😬

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1483,15 +1483,23 @@ def _validate_versions(self, data):
...
             and is_rollout
             and parsed_min_version
             < NimbusExperiment.Version.parse(
-                NimbusConstants.DESKTOP_ROLLOUT_MIN_SUPPORTED_VERSION
+                NimbusConstants.ROLLOUT_MIN_SUPPORTED_VERSION[application]
             )
         ):
             self.warnings["firefox_min_version"] = [
-                NimbusConstants.ERROR_DESKTOP_ROLLOUT_VERSION
+                NimbusConstants.ERROR_ROLLOUT_VERSION.format(
```
eliserichards, line 1497: Do we want this to be an error or a warning? I'm thinking that we might want to change this to--
```py
                {
                    "firefox_min_version": [
                        NimbusExperiment.ERROR_ROLLOUT_VERSION.format(...)
                    ],
                }
            )
```

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
jeddai: Magical, thank you!

experimenter/experimenter/experiments/constants.py
jeddai: I don't think we actually want to remove that as afaik that's referring to when the overall concept of rollouts existed for those apps. That check actually raises a ValidationError. I think what might make more sense is to change the name of `ROLLOUT_MIN_SUPPORTED_VERSION` to something like `ROLLOUT_LIVE_RESIZE_MIN_SUPPORTED_VERSION`?

experimenter/experimenter/experiments/constants.py
brennie, line 447: Please reformat without `\`. You can continue strings across multiple lines as so:

```python
ERROR_ROLLOUT_VERSION = (
    "WARNING: ..."
    "...."
)
```

brennie (changes requested):

experimenter/experimenter/experiments/constants.py
jeddai: Ah that specific message was previously using the backslash so I didn't change it, yeah I'll switch it to the parens

experimenter/experimenter/experiments/constants.py
jaredlockhart: @eliserichards oh yeah it's easy to think of that case, ie someone just goes right ahead and launches a 100% rollout and never intends to change it.  I think that pattern is somewhat common.  I think we should stick with the pattern we had, where if you launch a rollout before the re-enrollment supported version we just warn that they won't be able to effectively resize it.

jeddai: Yeah and I think the big thing is they can resize _up_ but if they resize down and then back up weird things happen. Is that correct?

jaredlockhart: Yeah that's the problem but I don't even think we need to be that specific with the warning, we should just be like 'resizing won't work right below this version'

jaredlockhart: Like it won't be long before we're healthily several versions past that cutoff date and this becomes moot.

jeddai: > we should just be like 'resizing won't work right below this version'

luke-skywalker

jaredlockhart (approved):
Perfect LGTM thanks @jeddai 🎉

brennie (approved):
R+!

PR #9129, opened 2023-07-14
fix 9128 feat(nimbus): add pop sizing data to graphql fetch

mikewilli:
Because

- we want to display pre-computed population sizing data on the Audience config page

This commit

- gets the pop sizing data from graphql

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 import jsonschema
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.core.cache import cache
 from django.db import models, transaction
 from django.utils.text import slugify
 from rest_framework import serializers
@@ -35,6 +36,7 @@ from experimenter.kinto.tasks import (
 )
 from experimenter.outcomes import Outcomes
 from experimenter.projects.models import Project
+from experimenter.settings import SIZING_DATA_KEY
 
 
 class TransitionConstants:
@@ -207,6 +209,7 @@ class NimbusConfigurationDataClass:
     types: typing.List[LabelValueDataClass]
     hypothesisDefault: str = NimbusExperiment.HYPOTHESIS_DEFAULT
     maxPrimaryOutcomes: int = NimbusExperiment.MAX_PRIMARY_OUTCOMES
+    population_sizing_data: str = ""
 
     def __init__(self):
         self.applications = self._enum_to_label_value(NimbusExperiment.Application)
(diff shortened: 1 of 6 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -295,6 +299,10 @@ def _get_feature_configs(self):
             .order_by("name")
         ]
 
+    def _get_population_sizing_data(self):
+        sizing_data = cache.get(SIZING_DATA_KEY)
+        return sizing_data.json() if sizing_data else "Sizing data not available."
```
jaredlockhart, line 304: Oh I really gotta rip this thing out.  This serializer isn't used at all, it was added as part of an effort to migrate off of GQL and onto DRF but was never finished and this whole class is unused.

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -346,6 +349,10 @@ def resolve_conclusion_recommendations(self, info):
             NimbusExperiment.ConclusionRecommendation
         )
 
+    def resolve_population_sizing_data(self, info):
+        sizing_data = cache.get(SIZING_DATA_KEY)
+        return sizing_data.json() if sizing_data else "Sizing data not available."
```
jaredlockhart, line 354: Wait if we return a non JSON parseable string here then the frontend has to distinguish this as a constant? Why not just return null here?

experimenter/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -2298,9 +2302,22 @@ def test_query_excludes_other_applications(self):
...
+    },
+)
 class TestNimbusConfigQuery(GraphQLTestCase):
     GRAPHQL_URL = reverse("nimbus-api-graphql")
 
+    def setUp(self):
+        super().setUp()
+        cache.delete(SIZING_DATA_KEY)
+        sizing_test_data = SampleSizesFactory.build()
+        cache.set(SIZING_DATA_KEY, sizing_test_data)
```
jaredlockhart, line 2319: Wait if we're just setting the same data (random here but looks indistinguishable from the test perspective?) then shouldn't we only need to do this once, ie in `setUpClass`?

jaredlockhart: And then if you have to do the same override and setup in multiple test cases, I'd pull it out into a single mixin class and just mix that in so you don't have to repeat it verbatim every time?

jaredlockhart (changes requested):
Generally looks good, couple thoughts.

jaredlockhart (approved):
Ya much nicer great job 🎉 🎉 🎉 🎉 🎉

PR #9154, opened 2023-07-21
fixes #9153 feat(nimbus): Add rich JSON editing

### LINKED ISSUE #9153: Rich JSON editing
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3708)

brennie:
Because:
- editing JSON in a plain textarea is not a great user experience; and
- CodeMirror cannot exist in a JSDOM environment

this commit:
- adds a rich text editor for feature configurations, which supports
  limited auto-complete and linting; and
- uses this editor *only* in the real environment and falls back to a
  regular text input in tests.


experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/FeatureValueEditor.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+let FeatureValueEditor: (props: FeatureValueEditorProps) => JSX.Element;
+
+if (process.env.JEST_WORKER_ID !== undefined) {
+  FeatureValueEditor = function FeatureValueEditor({
+    formControlAttrs,
+  }: FeatureValueEditorProps) {
+    return <Form.Control {...{ ...formControlAttrs }} as="textarea" rows={4} />;
+  };
```
jeddai, line 22: This is falling back to the existing input for the unit test issue yeah?

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
```diff
@@ -0,0 +1,413 @@
...
+  formControlAttrs: ReturnType<
+    ReturnType<typeof useCommonNestedForm>["formControlAttrs"]
+  >;
+}
+
+export default function RichFeatureValueEditor({
+  featureConfig,
+  field,
+  formControlAttrs,
+}: RichFeatureValueEditorProps) {
```
jeddai, line 31: I love destructuring

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
```diff
@@ -0,0 +1,413 @@
...
+      value={formControlAttrs.defaultValue ?? ""}
+      ref={editor}
+      extensions={extensions}
+      className={classNames(
+        formControlAttrs.className,
+        "feature-value-editor",
+        {
+          "is-valid": formControlAttrs.isValid,
+          "is-invalid": formControlAttrs.isInvalid,
+        },
```
jeddai, line 96: Are these values being set? I don't see the valid or invalid classes getting added to the element at any point.

jeddai (commented):
This is looking great, I am so excited for it! It's unfortunate we label nesting feature value types as just "json" so we don't get much in the way of nested autocompletion/validation.

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/FeatureValueEditor.tsx
brennie: Yes

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
brennie: This broke at some point during refactoring, good catch.

eliserichards (commented):
This looks really good! I love the error messages and formatting that you've got going on. I have a few small q's:

experimenter/tests/integration/nimbus/pages/experimenter/branches.py
```diff
@@ -153,6 +164,23 @@ def add_screenshot_buttons(self):
...
+            for line in editor_element.find_elements(By.CSS_SELECTOR, ".cm-line")
+        )
+
+    def _set_feature_value(self, editor_element, text):
+        line = editor_element.find_elements(By.CSS_SELECTOR, ".cm-line")[-1]
+        line.click()
+
+        actions = ActionChains(self.driver)
+        actions.send_keys(text)
+        actions.perform()
```
jrbenny35, line 179: Do you need to use ActionChains here? Will sending the keys to the element not work without an action?

jrbenny35 (changes requested):
Just had 1 question

jaredlockhart (changes requested):
Interesting, I was just playing around locally and I started with 0 features selected, added one, filled in its value, then added another and the new input field for the second selected feature was auto populated with the value of the other feature:

image

Is this caused by some state management in the rich editor?

Without it I can't replicate that behaviour:

image


Ah I think I see another example,

Start with this

image

Then remove 'cfr' and wind up like this

image


So I think something about the state management of the rich editor is going sideways.

Other than that it looks and feels great 🎉 🎉 🎉 🎉

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/FeatureValueEditor.tsx
```diff
@@ -0,0 +1,46 @@
...
+export interface FeatureValueEditorProps {
+  featureConfig: getConfig_nimbusConfig_allFeatureConfigs;
+  formControlAttrs: ReturnType<
+    ReturnType<typeof useCommonNestedForm>["formControlAttrs"]
+  >;
+}
+
+let FeatureValueEditor: (props: FeatureValueEditorProps) => JSX.Element;
+
+if (process.env.JEST_WORKER_ID !== undefined) {
```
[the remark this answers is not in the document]
brennie: JEST_WORKER_ID is only defined when running tests. So if its undefined, we're in prod and we can use the rich text component.

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
```diff
@@ -0,0 +1,413 @@
...
+        (state: EditorState, focusing: boolean) => {
+          if (!focusing) {
+            const value = state.doc.toString();
+
+            // We cannot trigger a change while a React update is on the stack.
+            setTimeout(() => {
+              field.onChange(value);
+            }, 0);
+
+            hideSubmitErrors?.();
```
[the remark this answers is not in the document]
brennie: It won't call the function if it is nullish

experimenter/tests/integration/nimbus/pages/experimenter/branches.py
brennie: I couldn't call send_keys on the element because it was a regular div, not an input element.

brennie:
> Interesting, I was just playing around locally and I started with 0 features selected, added one, filled in its value, then added another and the new input field for the second selected feature was auto populated with the value of the other feature:
> image
> 
> Is this caused by some state management in the rich editor?
> 
> Without it I can't replicate that behaviour:
> image
> 
> Ah I think I see another example,
> 
> Start with this
> image
> 
> Then remove 'cfr' and wind up like this
> image
> 
> So I think something about the state management of the rich editor is going sideways.
> 
> Other than that it looks and feels great 🎉 🎉 🎉 🎉

I think this is because we set the value on blur, which can race with feature values being added or removed. I was trying to avoid using an onChange handler because the view has to serialize its internal data structure into a string every key press, but that was likely premature.

jaredlockhart:
I did `make refresh up` on a clean checkout of the branch and then my git showed this:

image

not sure if we need to `.gitignore` that or something?

brennie:
> I did `make refresh up` on a clean checkout of the branch and then my git showed this:
> image
> 
> not sure if we need to `.gitignore` that or something?

My bad, I accidentally committed that file when switching branches.

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
```diff
@@ -0,0 +1,435 @@
...
+        z.union([
+          z.literal("integer"),
+          z.literal("boolean"),
+          z.literal("string"),
+          z.literal("json"),
+        ]),
+      ),
+    }),
+  ),
+});
```
jaredlockhart, line 142: Wait why can't we use the jsonschemas that are already present in the backend?

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
```diff
@@ -0,0 +1,435 @@
...
+        from,
+        to,
+        message,
+        severity: "error",
+      });
+    }
+
+    reportFloatValues(view.state.doc, rootNode, diagnostics);
+
+    return diagnostics;
```
jaredlockhart, line 317: I didn't think we'd have to do all this ourselves, I was hoping we could just naively pass the jsonschemas we have straight to this thing and it'd start validating.  But in any case, would it be worth pulling this out of here and into some kind of utility file and testing it separately rather than as part of the component?

jaredlockhart (approved):
Tested locally, everything works perfectly, behaves beautifully, this is so cool and amazing.  I dreamed we'd have this some day it's wild to see it finally be real.  Thanks for taking this on, I think this will be a huge workflow improvement.  Great job @brennie 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
brennie: This is the Zod definition of a JSON schema "object" type with simple properties

experimenter/experimenter/nimbus-ui/src/components/PageEditBranches/FormBranches/FormFeatureValue/RichFeatureValueEditor.tsx
brennie: That is what it does for the most part, with some additional filtering so that we don't get a bunch of repeated error messages.

PR #9157, opened 2023-07-24
fix #9156 feat(schemas): typescript schema generation and npm publish

### LINKED ISSUE #9156: Publish nimbus-schemas package to NPM
Build typescript types and publish to NPM

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3711)

mikewilli:
Because

- we want to use the population sizing schemas for graphql and UI

This commit

- sets up generation of typescript schemas from the pydantic schemas
- sets up publishing the typescript schemas to npm


jaredlockhart (changes requested):
Damn this is awesome.  Everything looks good but one thing stands out to me.  Here there's two parts

1) We have the generated TS schema tracked in the repo as index.d.ts
2) In circle we run the command to generate the TS schema and then publish

So I think what could happen is

1) Someone modifies the Python schema
2) Creates a PR that includes changes to the Python schema but fails to regenerate the TS interface
3) That PR passes and lands, and then the deploy job runs the generation step, updates the TS bindings, and publishes

Now the published package has a TS schema that doesn't match the one in the repo.

The way we get around this in other code generation parts of Experimenter is to add a check that looks to see if re running the generation step would result in a change and then block the PR if it does, example:

https://github.com/mozilla/experimenter/blob/main/Makefile#L31

So I would say we should either

1) Not track the TS schema in the repo at all and make it entirely an artifact of the deploy process (less a fan of this approach)
2) Adopt a similar pattern to above and fail a PR if it would cause a change to the TS schema but doesn't include it in the PR

But wow this looks great.  Can't wait to move more schemas in here, and get rust bindings! 🎉

mikewilli:
Good call. I added a similar check to the Makefile as part of the `schemas_check` target.

Makefile
```diff
@@ -273,12 +275,27 @@ schemas_test:
 	(cd schemas && poetry run pytest)
 
 schemas_check: schemas_install schemas_black schemas_ruff schemas_test
+	(cd schemas && poetry run pydantic2ts --module mozilla_nimbus_schemas.jetstream --output ./test_index.d.ts --json2ts-cmd "yarn json2ts")
+	diff schemas/test_index.d.ts schemas/index.d.ts || (echo nimbus-schemas typescript package is out of sync please run make schemas_build;rm schemas/test_index.d.ts;exit 1)
+	rm schemas/test_index.d.ts && echo "Done. No problems with schemas."
```
jaredlockhart, line 280: Oh right we're running these in the host env.  We should put all this in a container at some point, prolly easier for people.  Also instead of doing this write file then rm thing, maybe better to just write to `/tmp`.  Trying to rm a file like this always gets spotty.  In the other ones we write and then forget about it because when the container exits the file goes with it.

.circleci/config.yml
```diff
@@ -564,6 +564,19 @@ jobs:
             # For more on twine, see:
             #   https://twine.readthedocs.io/en/latest/
             make schemas_deploy_pypi
+      - run:
+          name: Create the NPM distribution files
+          command: |
+            make schemas_build_npm
```
jaredlockhart, line 570: So then we can take this out here since by the time it merges the file should be up to date?

jaredlockhart (changes requested):

Makefile
mikewilli: Yea, I was thinking of this and I like the `/tmp` folder option as a temporary (pun intended) solution. Added that for now and I'll create a ticket to migrate this to docker.

jaredlockhart (approved):
Awesome fantastic let's gooooooo 🎉 🎉 🎉 🎉  thnx @mikewilli 🙌 🙌 🙌 🙌

PR #9190, opened 2023-07-26
fix #9189 chore(project): use latest ruff

### LINKED ISSUE #9189: Use latest ruff rules
We recently switched to using [ruff](https://github.com/astral-sh/ruff) for Python linting, and it has since added many new helpful rules.  Let's update to the latest version of ruff and apply as many rules as we can.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3716)

jaredlockhart:
Because

* We are using ruff for linting and autoformatting Python
* There are fancy new ruff rules available

This commit

* Updates to the latest ruff
* Applies as many of the new rules as possible


jaredlockhart:
Latest ruff wants to use python 3.11 specific syntax, but the integration tests are stuck on python 3.8.  Rather than risking diverging our linting and syntax across the application and integration test code, I will just close this for now.

PR #9192, opened 2023-07-27
Add target for users with no autofill addresses

groovecoder:
Because

- Form Autofill addresses degrade the Relay integration UX, we want to release the Relay integration only to FXA users who do NOT have Relay and do NOT have any autofill addresses saved, so ...

This commit

- adds `NO_AUTOFILL_ADDRESSES` and `NOT_RELAY_USER_WITH_NO_AUTOFILL_ADDRESSES` targeting configs.


jaredlockhart (approved):
LGTM thnx @groovecoder 👍

PR #9194, opened 2023-07-27
fix #9193 chore(project): use py3.11 in integration tests

### LINKED ISSUE #9193: Update integration test images to use python 3.11
We run our integration tests in containers provided by selenium, but they use ubuntu 20.04 which ships with python 3.8.  Experimenter uses python 3.11 now, and to keep our linting and formatting and syntax consistent across the application code and integraiton tests, it's better for everything to use 3.11 so the same syntax is available.  Let's bump our selenium containers to python 3.11.

jaredlockhart:
Because

* Experimenter now uses py3.11 for application code
* To keep our formatting and linting rules consistent for the application and integration test code, we should use py3.11 everywhere
* Integration tests run in a container provided by selenium which uses py3.8

This commit

* Installs py3.11 in the selenium containers used for integration tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,17 @@ commands:
                 echo "No changes in << parameters.paths >>. Skipping tests and linting."
                 circleci-agent step halt
             fi
+  pull_firefox_images:
+    description: "Pull the latest Firefox Selenium images"
+    parameters:
+      firefox_version:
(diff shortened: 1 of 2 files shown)


jaredlockhart:
Tried a lot of things to get this to work but couldn't make any progress.  Closing this for now.

PR #9200, opened 2023-08-02
fix #9199 feat(schemas): remove min_version from sizing schemas

### LINKED ISSUE #9199: Remove min_version from sizing schemas
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3729)

mikewilli:
Because

- the population sizing tool removed min_version from calculations

This commit

- removes min_version from the schemas

### DIFF
--- a/schemas/mozilla_nimbus_schemas/tests/jetstream/test_population_sizing.py
+++ b/schemas/mozilla_nimbus_schemas/tests/jetstream/test_population_sizing.py
@@ -13,15 +13,14 @@ class TestPopulationSizing(TestCase):
         """Test against known good data."""
         sizing_test_data = """
         {
-            "firefox_desktop:release:['EN-US']:US:108": {
+            "firefox_desktop:release:['EN-US']:US": {
                 "new": {
                     "target_recipe": {
                         "app_id": "firefox_desktop",
                         "channel": "release",
                         "locale": "('EN-US')",
                         "country": "US",
-                        "new_or_existing": "new",
-                        "minimum_version": "108"
+                        "new_or_existing": "new"
                     },
                     "sample_sizes": {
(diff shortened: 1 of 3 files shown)


mikewilli:
With thanks to @alekhyamoz for contributing!

schemas/mozilla_nimbus_schemas/tests/jetstream/test_population_sizing.py
```diff
@@ -13,15 +13,14 @@ def test_parse_population_sizing(self):
         """Test against known good data."""
         sizing_test_data = """
         {
-            "firefox_desktop:release:['EN-US']:US:108": {
+            "firefox_desktop:release:['EN-US']:US": {
```
jrbenny35, line 16: Curious as to why you removed this?

mikewilli: This is part of what was removed from the formatting of the output that we're ingesting, so I wanted to represent it in the test data. It's hard to represent in the schema itself because it's a concatenation of values, and having an enum of all possible combinations would be huge.

brennie: theoretically could describe the format with a regex pattern (like https://json-schema.org/understanding-json-schema/reference/regular_expressions.html) but I dont know how to express that in pydandic and its probably overkill

mikewilli: Hm that's an interesting thought, but yea I would agree that it's probably overkill here.

brennie (approved):
r+

PR #9204, opened 2023-08-02
fixes #9201 bug(nimbus): Do not use negation in targeting expressions

### LINKED ISSUE #9201: Mobile does not support negation in targeting expressions
jexl-rs does not support negation (see [this open PR](https://github.com/mozilla/jexl-rs/pull/30)) and the recently added inclusion/exclusion targeting uses negation.

See-also https://github.com/mozilla-mobile/firefox-ios/issues/15843
See-also https://bugzilla.mozilla.org/show_bug.cgi?id=1846471

brennie:
Because:
- the JEXL evaluator in mobile does not support negation (`!expr`)

this commit:
- updates the generated exclusion targeting not to use negation and use comparison instead.


jeddai (approved):
This looks good, should be good to merge assuming all the tests pass

PR #9211, opened 2023-08-04
fix #9210 feat(cirrus): Add application support for cirrus

### LINKED ISSUE #9210: Add “user_id” to BucketRandomizationUnit
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3735)

yashikakhurana:
Because

- SDK (Cirrus) uses user id for bucketing
- Need to push Monitor web experiments to [web RS](https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/nimbus-web-experiments/records)
- a new application option for Monitor

This commit

- Add user id option for bucketing
- Push Monitor experiments to RS web URL
- Option to choose Monitor as an application
- Migration to support changes


experimenter/experimenter/experiments/migrations/0243_no_feature_config_monitor.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 3.1.8 on 2021-04-29 17:22
```
brennie, line 1: We do not need a tombstone feature for web applications.

We only need tombstone features for applications that did not always support multi-feature experiments. Since Cirrus will starts with supporting multi-feature experiments, we have no reason to serialize a payload with a tombstone feature.

experimenter/experimenter/experiments/constants.py
```diff
@@ -124,6 +125,19 @@ class ApplicationConfig:
     randomization_unit=BucketRandomizationUnit.NIMBUS,
 )
 
+APPLICATION_CONFIG_MONITOR_WEB = ApplicationConfig(
```
brennie, line 128: This is going to be our first application that isnt available in multiple versions. Right now the version field really only applies to desktop and mobile browsers.

We should add a flag to ApplicationConfig that indicates whether or not it is versioned. We will also want to omit the min_firefox_version and max_firefox_version fields on the Audience page.

The v5 serializer should be updated to make sure that you cannot set the min_firefox_version or max_firefox_version for unversioned apps.

cc @jaredlockhart does this all sound correct?

brennie (changes requested):

experimenter/experimenter/kinto/tests/test_tasks.py
```diff
@@ -568,7 +569,9 @@ def test_check_with_dirty_rollout_rejected_end_rolls_back_and_pushes(self):
             ).exists()
         )
 
-    def test_check_with_approved_update_sets_experiment_to_idle_saves_published_dto(self):
+    def test_check_with_approved_update_sets_experiment_to_idle_saves_published_dto(
+        self,
+    ):
```
jrbenny35, line 574: Is this change due to a formatter?

experimenter/experimenter/experiments/constants.py
yashikakhurana: yes that all will be done in a separate issue- making sure all the fields applicable to monitor renders that only

experimenter/experimenter/kinto/tests/test_tasks.py
jaredlockhart: Huh weird

jaredlockhart (dismissed):
Generally looks good, please fold those two migrations together into one.  @brennie 's right that now that we have multifeature we don't need the 'No Feature' feature for every application, but let's deal with that in another PR.  If you want to remove that test in another PR first, and then land this after and just not include the migration that adds the feature, you can do that?

Edit: Actually I take that back, maybe we can just nuke that test in this PR and remove the migration and just forget about 'No Feature' altogether.

yashikakhurana:
okey dokey, the test removed and migration too @jaredlockhart 🖖

jaredlockhart (approved):
Looks good thank you 🎉

PR #9238, opened 2023-08-08
fix #9235 bug(nimbus): fix pop sizing test with factory

### LINKED ISSUE #9235: fix pop sizing task test
There's a test of the fetch pop sizing task that is failing due to some static test data that is no longer valid (after a schema update). We should refactor this so that it generates valid test data so we don't need to change the test whenever the schema changes.

Related to this: https://app.circleci.com/pipelines/github/mozilla/experimenter/27502/workflows/5d5274e1-3d1f-401c-be05-107bae393b3e/jobs/101716

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3740)

mikewilli:
Because

- a sizing schemas change had an incompatible change that led to a test failing
- schemas tests should live in the package and not be concerned with the specific structure of the data
- we still want to test the task itself

This commit

- changes the static test data in experimenter for pop sizing to a factory


jaredlockhart (approved):
Heck yeah!!

PR #9240, opened 2023-08-09
fix #9239 feat(schemas): add factories to generate test data

### LINKED ISSUE #9239: add factories to schemas package to generate test data
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3744)

mikewilli:
Because

- we don't want to have to update static test data whenever schemas change
- we do still want tests on schemas package consumers to test overall functionality

This commit

- adds some test data generation factories to the schemas package for consumers to use within tests


schemas/mozilla_nimbus_schemas/tests/jetstream/test_population_sizing.py
```diff
@@ -87,3 +87,4 @@ def test_parse_population_sizing(self):
         """
         sample_sizes = SampleSizes.parse_raw(sizing_test_data)
         self.assertEqual(json.dumps(json.loads(sizing_test_data)), sample_sizes.json())
+s
```
jaredlockhart, line 90: `s`

schemas/mozilla_nimbus_schemas/tests/jetstream/test_population_sizing.py
```diff
@@ -1,7 +1,7 @@
 import json
 from unittest import TestCase
 
-from mozilla_nimbus_schemas.jetstream import SampleSizes
+from mozilla_nimbus_schemas.jetstream import SampleSizes, SampleSizesFactory
```
jaredlockhart, line 4: I don't see where this is used in the tests?

jaredlockhart (changes requested):

schemas/mozilla_nimbus_schemas/tests/jetstream/test_population_sizing.py
mikewilli: Sorry thought I removed that. I added it to make sure it worked, but for the package I thought I'd keep using the already-written static test data.

jaredlockhart: Oh, if the factories aren't used in the tests then there's no way to assert that they're working?

jaredlockhart (approved):
Awesome yes heck yeah.  So then we have both hard coded examples to stress the schema, and we test that the factory passes the same thing.  Love it 10/10 no notes, thnx @mikewilli 🎉 🎉 🎉

PR #9249, opened 2023-08-09
Fixes #9231 feat(nimbus): Add Gain Amount text field to Takeaways

### LINKED ISSUE #9231: Gain Amount - Add new fields in Takeaways section
Expose the following fields in Takeaways section -  “Gain Amount”.  

"Gain Amount" should be an open text field - so we can see how these wins get measured.  But I suspect it will be a sentence like “.5% gain in retention” or “.5% gain in days of use”..

!image-20230203-203812.png|width=1130,height=281!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3738)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/40653">image-20230203-203812.png</a>

eliserichards:
Because

- We want to give users the ability to specify what kind of gains that their experiment made

This commit

- Adds a new field, `takeaways_gain_amount`
- Adds this field as a freeform text box to the Takeaways section on the Summary page (only visible when an experiment is Complete)
- Let's you edit and save the text
- Gives you some example text

----

https://github.com/mozilla/experimenter/assets/43795363/1e7870e4-c185-40f1-8f28-9fc2fdd1dee4

----

Screen Shot 2023-08-09 at 4 30 31 PM


yashikakhurana:
quick review- can we do Gain amount and the value in the same line, currently there is one line break

eliserichards:
It's got a line break right now to match the Summary formatting: 
image

I definitely want to change this whole section's UI to be cleaner, but I think we need to get Shell's input before we do that. WDYT @yashikakhurana? 🤔

yashikakhurana:
@eliserichards gotcha, if its a paragraph then its fine, I thought it will be one liners, my bad

experimenter/experimenter/nimbus-ui/src/components/PageSummary/Takeaways/index.test.tsx
```diff
@@ -259,32 +294,6 @@ describe("TakeawaysEditor", () => {
     expect(result2.takeawaysMetricGain).toEqual(true);
   });
 
-  it("submits takeaway checkbox values even if value not changed", async () => {
```
eliserichards, line 262: This case is already covered by "submits form data when save is clicked"

eliserichards:
^ Adding a small change to this from @shell1 -- changing the wording for "promising metric gain" to "statistically significant DaU gain".

experimenter/experimenter/nimbus-ui/src/components/PageSummary/Takeaways/index.test.tsx
jaredlockhart: I love removing duplicate tests 🎉

jaredlockhart (approved):
Yep all looks right, tested locally works perfect.  This whole takeaways section is starting to grow at some point we should think about a nicer way to display that information in the non editable state, but we can deal with that later once we get all the new fields in and things settle down a bit.

Thanks @eliserichards this is great 🎉

PR #9260, opened 2023-08-11
Fixes #9259 feat(nimbus): updated timestamp format and fixed the changed_by bug

### LINKED ISSUE #9259: Update timestamp format for changes on history page and fix the username not being rendered
**Approach**
- update the timestamp format to show the hours minutes and the timezone 
- update the `changed_by` field in the changes to be the user's email instead of name as it currently returns empty for user with no name

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3758)

Avi-88:
Because

- the current timestamp doesn't specify timezone 
- some changes dont render the `changed_by` field as some users dont have their full name defined

This commit

- updates the timestamp format 
- updates the `changed_by` field to be the email of the user  instead of their full name

**Screenshot**
image


experimenter/experimenter/experiments/models.py
```diff
@@ -875,26 +876,26 @@ def get_changelogs_by_date(self):
...
             creation_log = changelogs[-1]
-            first_timestamp = creation_log.changed_on.strftime(date_option)
+            first_local_timestamp = timezone.localtime(creation_log.changed_on)
+            first_timestamp = first_local_timestamp.strftime(date_option)
             change = {
                 "event": "CREATION",
                 "event_message": (
-                    f"{creation_log.changed_by.get_full_name()} "
-                    f"created this experiment"
+                    f"{creation_log.changed_by} " f"created this experiment"
```
brennie, line 896: Two strings got moved onto the same line, so you can make this a single string:

```suggestion
                    f"{creation_log.changed_by} created this experiment"
```

brennie (dismissed):

experimenter/experimenter/experiments/models.py
Avi-88: Oh I think it changed after the `make code_format`. I'll update it :+1:

yashikakhurana (approved):
thank you @Avi-88 looks good to me 👏

PR #9270, opened 2023-08-11
fix #9269 feat(nimbus): add developer channel for ios

### LINKED ISSUE #9269: Add "developer" channel for iOS experiments/rollouts
Both mobile platforms have an additional update channel named developer. For iOS, whenever we build locally and run on the simulator the build can’t be changed (at least by QA) and will default to “developer” channel.

This is not an issue for Android as we can build and sign locally with any “channel” that we would want. 

Adding this option in the Audience section of Nimbus will enable QA to create valid experiments that would naturally enroll on dev builds.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3733)

jaredlockhart:
Because

* Local builds of IOS will report the channel as developer
* QA often uses local builds for testing
* They have been forced to manually alter recipes to channel developer for testing

This commit

* Adds developer as a supported channel for IOS to facilitate QA testing


yashikakhurana (approved):
perfect ⭐

brennie (approved):

PR #9280, opened 2023-08-15
fix #9267 feat(nimbus): convert monitor web fml to experimenter format

### LINKED ISSUE #9267: Process Monitor FML file into format consumable by Experimenter
The format that FML files appear in is not parseable by Experimenter, and existing applications like Firefox desktop and mobile will use the FML CLI to preprocess their FML files into a format Experimenter can consume.  We'll have to do the same thing for the Monitor FML file, but rather than expect the Monitor team to do it as part of their development process, we can just run it locally in the Circle task ourselves.

jaredlockhart:
Becuase
    
* The FML format that lives in an applications repo is not parseable by Experimenter
* The FML command line tool will convert it into the format expected by Experimenter
* Currently supported applications do this as part of their build chain
* New applications, particularly web, may not do this so we should do this conversion on our end as part of the external resource fetch step
    
This commit
    
* Adds a step to the fetch_external_resources make command to download and build the FML tool and convert the FML files into a format Experimenter can parse

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ feature_manifests:
 	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/ios.yaml $(FEATURE_MANIFEST_FXIOS_URL)
 	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/focus-android.yaml $(FEATURE_MANIFEST_FOCUS_ANDROID)
 	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/focus-ios.yaml $(FEATURE_MANIFEST_FOCUS_IOS)
-	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/monitor-web.yaml $(FEATURE_MANIFEST_MONITOR)
+	curl -LJ --create-dirs -o experimenter/experimenter/features/manifests/monitor-web.fml.yaml $(FEATURE_MANIFEST_MONITOR)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -90,8 +90,19 @@ feature_manifests:
...
+	cd experimenter/experimenter/features/manifests/;\
+	curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh -s -- -y;\
+	source "$HOME/.cargo/env";\
+	git clone git@github.com:mozilla/application-services.git;\
+	cd application-services;\
+	git submodule init;\
+	git submodule update --recursive;\
+	cd components/support/nimbus-fml;\
+	cargo build;\
+	cargo run "../../../../monitor-web.fml.yaml" experimenter --output "../../../../monitor-web.yaml" --channel developer;
```
jeddai, line 103: You can actually eliminate the `cargo build` step if you want — `cargo run` does the build if it's not been built yet.

Also, you can get around needing to `cd` into `nimbus-fml` using `--manifest-path`
```
cargo run --manifest-path components/support/nimbus-fml/Cargo.toml <rest-of-command>
```
Just thought that might cut down on the `../`s needed

brennie (approved):

Makefile
brennie: Do we care if we do a release build or not?

jeddai: Not for this particular use case imo

Issue #9268, opened 2023-08-11
Add filters for new Takeaways


data-sync-user:
[https://mozilla-hub.atlassian.net/browse/EXP-3113](https://mozilla-hub.atlassian.net/browse/EXP-3113|smart-link) , [https://mozilla-hub.atlassian.net/browse/EXP-3737](https://mozilla-hub.atlassian.net/browse/EXP-3737|smart-link) , and [https://mozilla-hub.atlassian.net/browse/EXP-3738](https://mozilla-hub.atlassian.net/browse/EXP-3738|smart-link) added “QBR notable learning”, “Statistically Significant DaU Gain”, and “Gain Amount” to the Takeaways section.  

We want to add the first 2 (“QBR notable learning” and “Statistically Significant DaU Gain”) as filters to the main screen (at the bottom - as these will be less used - but critical to be able to use).  So it would be a show all “QBR notable learning” and/or “Statistically Significant DaU Gain” filters.  Similar to what was in the older experimenter for other items.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3710)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/40063">image-20230203-203812.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/40064">image-20230203-222813.png</a>

PR #9285, opened 2023-08-16
For #9268 fix(nimbus): Update filter name to DaU gain

eliserichards:
Because

- We want to update the filter name from "Metric Gain" to "DAU Gain"

This commit

- Updates the name!

image


yashikakhurana (approved):
💯 wonderful

PR #9328, opened 2023-08-30
fix #9189 chore(project): use latest ruff rules

### LINKED ISSUE #9189: Use latest ruff rules

jaredlockhart:
Because

* The latest version of ruff has many more linting and auto formatting rules

This commit

* Updates to the latest ruff and applies many of its new linting and auto formatting rules


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -407,7 +407,6 @@ def to_representation(self, data):
 
         This enforces that the serialized data is ordered by the feature config ID.
         """
-        super().to_representation
```
mikewilli, line 410: Is this not important?

mikewilli (approved):
Neat! Separate PRs for Cirrus and Schemas directories?

experimenter/experimenter/experiments/api/v5/serializers.py
mikewilli: Ah yea fair enough, good job ruff.

jaredlockhart:
> 

Oh yeah I got started but there's a bunch of other infrastructural work that has to happen first so I'll just do it separately.

PR #9330, opened 2023-08-31
Fixes #9057 feat(nimbus): added the new history page link to summary page

### LINKED ISSUE #9057: Add links to the summary page
* Define the links that should be added to the summary page
* Create the links using Django's url template tag
* Ensure the links are properly styled and formatted

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3654)

Avi-88:
Because

- there was no way to navigate to the new history page form an experiment summary page

This commit

- adds a link to the new history page 

**Screenshots**

image

image


experimenter/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
```diff
@@ -236,6 +237,22 @@ export const AppLayoutSidebarLaunched = ({
...
+              <p className="edit-divider position-relative small my-2">
+                <span className="position-relative bg-light pl-1 pr-2 text-muted">
+                  History
+                </span>
+              </p>
+
+              <LinkNav
+                route={`/history/${slug}`}
+                testid={`history-page-${slug}`}
+                isExternalRoute={true}
```
mikewilli, line 250: No need to add `={true]` for boolean props.
```suggestion
                isExternalRoute
```

experimenter/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.tsx
```diff
@@ -127,6 +128,21 @@ export const AppLayoutWithSidebar = ({
...
               ))}
 
+              <p className="edit-divider position-relative small my-2">
+                <span className="position-relative bg-light pl-1 pr-2 text-muted">
+                  History
+                </span>
+              </p>
+
+              <LinkNav
+                route={`/history/${slug}`}
```
mikewilli, line 138: For consistency, could we serve this page at `/nimbus/{slug}/history`?

experimenter/experimenter/nimbus-ui/src/components/AppLayoutWithSidebar/index.tsx
```diff
@@ -127,6 +128,21 @@ export const AppLayoutWithSidebar = ({
...
+                testid={`history-page-${slug}`}
+                isExternalRoute={true}
```
mikewilli, line 140: My preference is to use a more consistent URL (see previous comment), but if there is a good reason to keep the current URL, then I'd suggest renaming this prop. To me, "external" implies "outside of Experimenter". Maybe something like `relativeToRoot`?

experimenter/experimenter/nimbus-ui/src/components/LinkNav/index.tsx
```diff
@@ -64,6 +70,20 @@ export const LinkNav = ({
...
+          href={to}
+          className={classNames(
+            textColor,
+            "d-flex font-weight-semibold m-0 p-0 b-0 align-items-center",
+          )}
+          data-testid={testid}
+          onClick={onClick}
+        >
+          {children}
+        </Button>
```
mikewilli, line 86: To reduce code duplication, could you add `href={isExternalRoute ? to : ""}` to the above block? This prop appears to be the only difference between the two Buttons.

mikewilli (commented):
It seems to me that serving the changelogs out of `/nimbus/{slug}/history` would simplify this and be more consistent with other experiment pages. Is there a reason not to do this?

jaredlockhart:
We can't say 'Changelogs' as the link label, we should call it 'History'

mikewilli:
>We can't say 'Changelogs' as the link label, we should call it 'History'

I had the same thought. Is it weird to have a History section that only has a History link in it? Maybe we can put this under Links instead?

yashikakhurana (changes requested):
I would say to add a test to check if history button is getting rendered, otherwise 💯

yashikakhurana (approved):
🚢

PR #9331, opened 2023-08-31
Use pre-built nimbus-cli for FML transformations

jhugman:
Because

- prebuilt versions of nimbus-cli include FML are available for linux-musl

This commit

- Installs the nimbus-cli as part of deployment
- Uses this to fetch and transform FML files from the Firefox Mobile, Focus and Monitor apps

This means:

- Faster deploy times
- We can remove the .experimenter.yaml files from the apps' repos and build scripts.


jaredlockhart (approved):
Okay tested locally and in circle, everything looks right, this is awesome thnx @jhugman 🎉 🎉 🎉 🎉 🎉 🎉

PR #9340, opened 2023-09-01
fix #9339 bug(schemas): include proper types filename in npm package

### LINKED ISSUE #9339: Fix schemas npm project config

mikewilli:
Because

- I changed the name of the generated types file to be consistent with normal TS type filenames (`*.d.ts`)
- I forgot to reference the new filename in the package.json

This commit

- uses the new filename in the package.json
- bumps schemas version so it will be published


jaredlockhart (approved):
👍

PR #9353, opened 2023-09-05
fixes #9352 feat(nimbus): add targeting for core users with 28+ days of activity

### LINKED ISSUE #9352: Add targeting for users active every day in the past 28 days
Because

    we would like to display a message to core users that have been active every day for the past 28 days

This commit

    adds advanced targeting to target users that have profiles older than 28 days, and have been active every day for the last 28 days

jwayn:
Because

- we would like to target users with profiles > 28 days that have been active every day for the last 28 days

This commit

- add targeting for users core users who have been active for 28 days or more in the last month


experimenter/experimenter/targeting/constants.py
```diff
@@ -1454,6 +1454,17 @@ def __post_init__(self):
     ),
 )
 
+CORE_USER_FULLY_ACTIVE = NimbusTargetingConfig(
+    name="Core user (Active Every Day)",
+    slug="core_user_active_every_day",
+    description="Active every day in the past 28 days",
+    targeting=f"{PROFILE28DAYS} && userMonthlyActivity|length >= 28",
```
jaredlockhart, line 1461: This should be

```py
    targeting=f"{PROFILE28DAYS.targeting} && userMonthlyActivity|length >= 28",
```

jaredlockhart (changes requested):

experimenter/experimenter/targeting/constants.py
jwayn: Thanks for catching this!

jaredlockhart:
Oh sorry no you were right the first time 😅 , I thought it was referencing a different targeting parameter.  Set it back to where it was and I'll merge 👍

jaredlockhart (approved):
👍

jwayn:
Thanks, Jared!

PR #9361, opened 2023-09-05
fix #9359 feat(schemas): add experiment schema to nimbus schemas

### LINKED ISSUE #9359: Move recipe schema to mozilla-nimbus-schemas
Now that we have mozilla-nimbus-schemas setup to publish to pypi and npm, let's move the recipe schema out of mozilla-nimbus-shared and into mozilla-nimbus-schemas and then we can migrate all the consumers over and deprecate mozilla-nimbus-shared.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3800)

jaredlockhart:
Because

* We now have the mozilla-nimbus-schemas package for storing all Nimbus related schemas
* We should add the experiment schema there and update all components to use it

This commit

* Adds the Nimbus experiment schema to the nimbus schemas package
* Adds all existing test fixtures from shared


schemas/mozilla_nimbus_schemas/experiments/experiments.py
```diff
@@ -0,0 +1,70 @@
+from typing import List, Optional, Union
+
+from pydantic import BaseModel
+
+
+class BucketConfig(BaseModel):
+    randomizationUnit: str
```
mikewilli, line 7: Guessing the camelCase is because these schemas were originally defined in typescript, but just commenting in case you want to change.

mikewilli (changes requested):
Awesome! Excited to be expanding the scope of this package. One main thing:

This is missing the TS types, which I realized is the main thing that should probably be updated with this PR. The library that converts pydantic to json doesn't accept the package as an input and requires individual modules, so right now the make command only builds TS types for jetstream. We'll need to fix that and then regenerate the TS types, and it might be worth a quick discussion about how to go about it.

schemas/mozilla_nimbus_schemas/experiments/experiments.py
jaredlockhart: No no yeah we have to keep it in camel case 🐫 🐍

jaredlockhart:
@mikewilli Oh! I thought it would auto generate the TS bindings as part of the deploy.  Okay I'll see if I can add them in here.

mikewilli:
Yea, we decided in the PR that added TS types (#9157) to do it this way. Unfortunately, because of the restriction I mentioned above where we need to generate a file for each module, the CI check that should fail here didn't fail (it only checks the jetstream module because that is the only currently generating TS types).

schemas/index.d.ts
```diff
@@ -60,6 +60,68 @@ export interface Outcome {
...
+  locales?: string[];
+}
+export interface BucketConfig {
+  randomizationUnit: string;
+  namespace: string;
+  start: number;
+  count: number;
+  total: number;
+}
+export interface Outcome1 {
```
mikewilli, line 98: Uh oh -- name conflict. It added a `1` to the end because `jetstream` has its own `Outcome`.

Makefile
```diff
@@ -292,7 +292,7 @@ schemas_test:
 	(cd schemas && poetry run pytest)
 
 schemas_check: schemas_install schemas_black schemas_ruff schemas_test
-	(cd schemas && poetry run pydantic2ts --module mozilla_nimbus_schemas.jetstream --output /tmp/test_index.d.ts --json2ts-cmd "yarn json2ts")
+	(cd schemas && poetry run pydantic2ts --module mozilla_nimbus_schemas.__init__ --output /tmp/test_index.d.ts --json2ts-cmd "yarn json2ts")
```
mikewilli, line 295: Nice, obvious in retrospect! Whoops. Can you just pass `mozilla_nimbus_schemas` as the module now that there's an `__init__` file?

mikewilli (changes requested):
Requesting changes because of the typescript name conflict across modules that I mentioned in a comment. 

The best solution I can think of would be to keep separate `.d.ts` files for each module and then aggregate them in `index.d.ts`, similar to the hierarchy of `__init__.py` files. Not sure how easy that is to do generically in the Makefile though.

Let me know how you want to approach it or if you want to discuss.

schemas/index.d.ts
jaredlockhart: I'll just rename it

Makefile
jaredlockhart: Sadly no becuase `mozilla_nimbus_schemas` is both the name of the folder and the name of the Python import path, and [here](https://github.com/phillipdupuis/pydantic-to-typescript/blob/master/pydantic2ts/cli/script.py#L34-L42) they check whether it's a valid path FIRST and try to do the path loading mechanism which isn't working for whatever reason, so explicitly pointing it to `mozilla_nimbus_schema.__init__` makes it not a valid file path but a valid Python import path so it forces the check into the Python import flow which does work.  Silly nuance of the module.

mikewilli: Ah, oh well. At least this catches everything now. Thanks!

schemas/index.d.ts
mikewilli: Alright, that's fine for now but we may want to think about an approach going forward. I don't like how this seemed like a silent error that would be very easy to miss and confusing. Maybe we can at least error/warn on name conflicts via CI as opposed to the TS conversion just renaming without telling us. Maybe a strict linting that doesn't allow TS types to end with a number?

jaredlockhart: No I think the strategy here is to write tests that hit the ts generated schemas.  If the renaming had broken the interface the tests would have failed.

mikewilli (approved):
Heck yea schemas++ :tada:

schemas/index.d.ts
mikewilli: Oh yea that's better, I'll add a ticket.

jaredlockhart: Already filed ;) https://github.com/mozilla/experimenter/issues/9368

PR #9366, opened 2023-09-05
fixes #9365 feat(nimbus): add targeting for early day users with default bookmarks toolbar 

### LINKED ISSUE #9365: Add targeting for early day users with bookmarks toolbar pref unchanged
Because we have upcoming experiments that will be changing the behavior of the bookmarks toolbar, we'd like to be able to target a group of users who have not altered the default visibility behavior of the bookmarks toolbar.

jwayn:
Because

    - we would like to target early day users with bookmarks toolbar default behavior

This commit

    - add targeting for early day users who have not changed the default value of the bookmarks toolbar visibility


jaredlockhart (approved):
Looks good thnx @jwayn 🎉

PR #9370, opened 2023-09-06
fix #9288 feat(cirrus): Editable client id and context

### LINKED ISSUE #9288: Configure option for client and context
Once we have the demo app integrated with Cirrus and fml, experimenter, to test for various inputs we want to configure client id or context from the frontend, if its provided we will be using that else we will provide the default value to test

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3771)

yashikakhurana:
Because

- Currently, to test the demo app with Cirrus, we have a static client ID and context

This commit

- Provide a way so that we can dynamically pass the client ID and context to the demo app frontend, which will help us in the integration tests to test with different scenarios
- The above code is modified in such a way that if client id and context is provided, then it will override the default values
Screenshot 2023-09-06 at 1 00 50 AM

fixes #9288

### DIFF
--- a/demo-app/frontend/src/index.js
+++ b/demo-app/frontend/src/index.js
@@ -10,15 +10,32 @@ root.render(
 
 function App() {
   const [message, setMessage] = useState({});
+  const [clientId, setClientId] = useState('');
+  const [context, setContext] = useState('');
+  const [apiCallTriggered, setApiCallTriggered] = useState(false);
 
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
AWESOME 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

Issue #9342, opened 2023-09-04
Where/when does experimenter get the FML location from?


data-sync-user:
Somewhere in experimenter, we should have a configuration file for each app. This task is adding to this configuration:

* the Github repository for the app
* the path from the root of the repo to the FML file
* the mapping from app version to git branch name or tag.

The version that the mapping function takes should be of the form that is used by the audience configuration page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3794)

Issue #9286, opened 2023-08-16
Move FML fetching/parsing into management command


jaredlockhart:
We will be adding more applications in the future, and so we should consolidate all of the configuration to add a new app into one place.  We have entries in experiments/constants.py, but the only thing missing there is the path to the manifest files.  If we move the path to the manifest files into the application constants, then we can move the fetching and conversion of the manifest files out of bash scripts in the makefile into a management command in the Experimenter container.

PR #9389, opened 2023-09-11
For #9342 feat(project): Add configuration folder with yaml files for each mobile application

eliserichards:
Because

- We want to specify the following for each of the mobile apps:
   - repo link
   - path to the fml file for the repo
   - a mapping for the version tags (so that we can find the appropriate FML for a specific version)

This commit

- Adds a new folder, `/experimenter/features/configs` that contains a file for each mobile application
   - This locations--`/features`--also contains the feature definitions that map to the `NimbusFeature` API
- Adds files for: 
   - ios
   - focus ios
   - fenix
   - focus android

----
We don't need anything for Klar, do we?

### DIFF
--- a/experimenter/experimenter/features/configs/apps.yaml
+++ b/experimenter/experimenter/features/configs/apps.yaml
@@ -1,0 +1,20 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+---
+fenix:
+  repo: @mozilla-mobile/firefox-android
+  fml_path: fenix/app/nimbus.fml.yaml
+  branch_mapping: "v{version}"
+ios:
+  repo: @mozilla-mobile/firefox-ios
+  fml_path: nimbus.fml.yaml
+  branch_mapping: "release/v{version}"
+focus-android:
+  repo: @mozilla-mobile/firefox-android
+  fml_path: focus-android/app/nimbus.fml.yaml
+  branch_mapping: "v{version}"
+focus-ios:
+  repo: @mozilla-mobile/focus-ios
+  fml_path: nimbus.fml.yaml
+  branch_mapping: "v{version}"


jaredlockhart (commented):
Should these be in separate yamls?  Or folded into the [application config constants](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/constants.py#L60-L175)?  Where will these be consumed?

jaredlockhart:
Also I should've pointed out, right now these paths are [defined](https://github.com/mozilla/experimenter/blob/main/Makefile#L56-L60) and [consumed](https://github.com/mozilla/experimenter/blob/main/Makefile#L91-L95) directly in the Makefile, but as we add (many!) more applications, we'll want a more streamlined way to add and configure applications, which I believe is what this PR is driving towards, and to that effect one option is to move it all into Python space and then move the FML integration commands also into Python space as a management command filed here https://github.com/mozilla/experimenter/issues/9286, though I pre emptively filed that and it may not be the correct approach and I defer to to @brennie to decide on an effective strategy.

experimenter/experimenter/features/configs/fenix.yaml
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+---
+fenix:
+  repo: @mozilla-mobile/firefox-android/tree/main/fenix
+  fmlPath: app/nimbus.fml.yaml
```
brennie, line 7:
```suggestion
  repo: @mozilla-mobile/firefox-android
  fmlPath: fenix/app/nimbus.fml.yaml
```

experimenter/experimenter/features/configs/focus-android.yaml
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+---
+focus-android:
+  repo: @mozilla-mobile/firefox-android/tree/main/focus-android
+  fmlPath: app/nimbus.fml.yaml
```
brennie, line 7:
```suggestion
  repo: @mozilla-mobile/firefox-android
  fmlPath: focus-android/app/nimbus.fml.yaml
```

brennie (changes requested):

brennie (approved):

PR #9392, opened 2023-09-12
fix #9391 chore(project): use docker hydrobuild

### LINKED ISSUE #9391: Use docker hydrobuild
Now that we have beta access to [Docker Hydrobuild](https://docs.docker.com/build/hydrobuild/), let's update the Makefile to explicitly use `docker buildx build` instead of `DOCKERBUILD_KIT=1 docker build` so that we can then set hydrobuild as the default global builder and the rest of our tooling should then start using the clouder builder.  We can update our circle config to use the latest docker client and set the cloud builder as the default and the rest of our CI suite should then also be using it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3816)

jaredlockhart:
Because

* We now have access to Docker Hydrobuild
* This allows us to use a shared cloud builder for our docker images so we can share a build cache between CI and local dev

This commit

* Updates the makefile to explicitly use docker buildx
* Updates the circle config to install the latest docker client and setup a cloud builder

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,23 +19,45 @@ commands:
                 echo "No changes in << parameters.paths >>. Skipping tests and linting."
                 circleci-agent step halt
             fi
+  setup_docker:
+    description: "Setup Docker with Hydrobuild"
+    steps:
+      - run: |
+          mkdir -vp ~/.docker/cli-plugins/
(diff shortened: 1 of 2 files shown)


brennie (approved):
Let's goooooo!

PR #9407, opened 2023-09-15
fix #9405 feat(cirrus): Update channels

### LINKED ISSUE #9405: Update channels for monitor
Set new channels `staging` and `production` for the monitor app

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3831)

yashikakhurana:
Because

- Monitor uses staging and production channels

This commit

- Add new channels - `staging` and `production`
- Update the monitor app to use those channels

fixes #9405


jaredlockhart (changes requested):
Let's block this until the [Monitor FML](https://github.com/mozilla/blurts-server/blob/main/config/nimbus.yaml#L4-L5) shows the correct channels since that is the canonical definition of channels.

yashikakhurana:
> Let's block this until the [Monitor FML](https://github.com/mozilla/blurts-server/blob/main/config/nimbus.yaml#L4-L5) shows the correct channels since that is the canonical definition of channels.

here is the ticket- https://github.com/mozilla/blurts-server/issues/3416

jaredlockhart (approved):
Awesome thank you!  🎉

PR #9439, opened 2023-09-19
fix #9436 fix(cirrus): Update channels for monitor in make command

### LINKED ISSUE #9436: Cannot fetch manifests due to misconfigured channels in monitor
As of https://github.com/mozilla/blurts-server/pull/3420, the monitor channels are now staging and production, not beta and release. The Makefile needs to be updated.

See-also #9294.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3847)

yashikakhurana:
Because

- As the monitor channels are now staging and production, our cron job is failing.

This commit

- Modify the make command with the appropriate channel option `production`


jhugman (approved):
🚢 :shipit: ✅ 🥅

yashikakhurana:
> Oh I thought you'd just fold it into #9439 since that's touching all the other places but yep this works too and then we can merge the other one right?

yes right, I thought it would be easier

PR #9444, opened 2023-09-19
fix #9443 bug(project): disable hydrobuild for prs

### LINKED ISSUE #9443: Disable hydrobuild for PRs
Using [Docker Hydrobuild](https://docs.docker.com/build/hydrobuild/) requires access to environment secrets in Circle to login to Dockerhub.  Previously the only place we used those secrets to login to Dockerhub was after a merge to main to push the images, which takes place directly against the mozilla/experimenter repo and so secrets are available.  Now that we've enabled Hydrobuild for PRs, we would need access to those secrets, but that would prevent PRs from forks from building, and since Experimenter is an open source project we need support for forked PRs.  Let's diable Hydrobuild for all PRs and re enable the Circle Docker Layer caching, but keep Hydrobuild enabled for the post merge build step where the secrets are available.  That way we can still pre-hydrate the Hydrobuild cache so that pulling from main should result in a local cache hit, but doesn't require that PRs use it and need access to secrets.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3849)

jaredlockhart:
Because

* Docker Hydrobuild requires access to environment secrets
* PRs from forks will not have access to environment secrets

This commit

* Disables hydrobuild for PRs that do not have access to secrets
* Re enables Circle docker layer caching


experimenter/experimenter/test.txt
```diff
@@ -0,0 +1 @@
+test
```
yashikakhurana, line 1: do we want this file?

yashikakhurana (changes requested):

experimenter/experimenter/test.txt
jaredlockhart: That's just to force it to run the entire suite which is gated on a path check I'll remove it before merging

jaredlockhart:
Actually I'll have to submit this from a fork to test it.

PR #9446, opened 2023-09-19
fix #9443 bug(project): disable hydrobuild for prs

### LINKED ISSUE #9443: Disable hydrobuild for PRs

jaredlockhart:
Because

* Docker Hydrobuild requires access to environment secrets
* PRs from forks will not have access to environment secrets

This commit

* Disables hydrobuild for PRs that do not have access to secrets
* Re enables Circle docker layer caching

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,27 +30,18 @@ commands:
           curl --silent -L --output ~/.docker/cli-plugins/docker-compose https://github.com/docker/compose-desktop/releases/download/v2.21.0-desktop.1/docker-compose-linux-x86_64
           chmod a+x ~/.docker/cli-plugins/docker-buildx
           chmod a+x ~/.docker/cli-plugins/docker-compose
-          echo "$DOCKER_PASS" | docker login --username $DOCKER_USER --password-stdin
-          docker buildx create --use --driver cloud "mozilla/default"
+          if [ -n "$DOCKER_USER" ]; then
+            echo "$DOCKER_PASS" | docker login --username $DOCKER_USER --password-stdin
+            docker buildx create --use --driver cloud "mozilla/default"
+          else
+            echo "DOCKER_USER is empty, skipping docker login"
+          fi
+
 jobs:
-  build_docker_images:
-    machine:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -30,27 +30,18 @@ commands:
           curl --silent -L --output ~/.docker/cli-plugins/docker-compose https://github.com/docker/compose-desktop/releases/download/v2.21.0-desktop.1/docker-compose-linux-x86_64
           chmod a+x ~/.docker/cli-plugins/docker-buildx
           chmod a+x ~/.docker/cli-plugins/docker-compose
-          echo "$DOCKER_PASS" | docker login --username $DOCKER_USER --password-stdin
-          docker buildx create --use --driver cloud "mozilla/default"
+          if [ -n "$DOCKER_USER" ]; then
```
mikewilli, line 33: Any harm in testing for both user and pass?

```suggestion
          if [ -n "$DOCKER_USER" -a -n "$DOCKER_PASS" ]; then
```

mikewilli (approved):

.circleci/config.yml
jaredlockhart: nope no harm 👍

yashikakhurana (approved):
🎉

PR #9452, opened 2023-09-20
fix #9451 chore(project): build multiple docker platforms in circle

### LINKED ISSUE #9451: Build multiple platforms with Docker in CI
We have to specify the target platforms when we build in CI so let's explicitly annotate our build commands with

```
--platform linux/amd64,linux/arm64
```

but only in CI, locally we should only build on the local platform.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3856)

jaredlockhart:
Because

* We can now target multiple platforms in our Docker builds in Circle

This commit

* Enables multiplatform builds for Docker in Circle

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ COMPOSE_LEGACY = ${COMPOSE} -f docker-compose-legacy.yml
 COMPOSE_TEST = docker compose -f docker-compose-test.yml
 COMPOSE_PROD = docker compose -f docker-compose-prod.yml
 COMPOSE_INTEGRATION = ${COMPOSE_PROD} -f docker-compose-integration-test.yml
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -119,16 +120,16 @@ compose_build:
 	$(COMPOSE) build
 
 build_dev: ssl
-	docker buildx build --target dev -f experimenter/Dockerfile -t experimenter:dev experimenter/
+	docker buildx build $(BUILD_PLATFORMS) --target dev -f experimenter/Dockerfile -t experimenter:dev experimenter/
```
brennie, line 123: Do you want to pull this whole thing into a `$(DOCKER_BUILD)` variable?

brennie (approved):

Makefile
jaredlockhart: You mean the `docker buildx build` part?

Issue #9388, opened 2023-09-11
Get paths and refs for FmlClient


data-sync-user:
~This may require some moving of the order of the audience selection and branch configuration pages around.~ (That will be done in #9432)

This task is creating a functionality that takes the app and the app versions that the experiment defines. It then

* uses [https://mozilla-hub.atlassian.net/browse/EXP-3794](https://mozilla-hub.atlassian.net/browse/EXP-3794|smart-link) to find the locations of the feature manifest for each version of the app.
* loads the version of the feature manifest into memory.
* iteration over each of the feature manifests, getting the feature ids available for that manifest
* finds the set of feature ids that are in all of the versions of the app.

~It may be beneficial to start with a class called `FeatureManifestForExperiment` or `FeatureManifestLoader` which takes an `app_name` and a list of `app_versions`, and loads each version of the manifest into memory for methods to use.~ (This is done in #9429)

Then have the `get_features` do the iterating over the feature manifests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3795)

data-sync-user:
➤ Elise Richards commented:

This is going to overlap a bit with the work that Barret Rennie is doing for feature versioning: ticket #…

PR #9454, opened 2023-09-20
For #9388 feat(project): Get manifest paths and release versions in fml loader

eliserichards:
Because

- We want to get the repos, fml paths, and formatted versions from the `apps.yaml` file

This commit

- Adds methods to the loader that:
   - Gets the repo paths from the yaml file
   - Gets the manifest path from the yaml file
   - Formats the versions according to the format in the yaml file
   - Sets up the loader methods to be hooked up to the FML

### DIFF
--- a/experimenter/experimenter/features/manifests/nimbus_fml_loader.py
+++ b/experimenter/experimenter/features/manifests/nimbus_fml_loader.py
@@ -1,78 +1,61 @@
 import os
 
 import yaml
-from packaging import version
+from packaging import version as Version
+from rust_fml import FmlClient
 
 from experimenter.settings import BASE_DIR
-
-from rust_fml import FmlClient
 
 
 class NimbusFmlLoader:
     __base_path = os.path.join(BASE_DIR, "features", "manifests", "apps.yaml")
-    __major_release = "major_release_branch"
-    __minor_release = "minor_release_tag"
 
-    def __init__(self, application: str, channel: str):
+    def __init__(self, application: str, channel: str, file_location=__base_path):
         self.application: str = application
         self.channel: str = channel
+        self.application_data = self.get_application_data(application, file_location)
+        # Todo: Connect FML paths and channels https://mozilla-hub.atlassian.net/browse/EXP-3845
+        self.fml_client = self.create("", "")
 
     @staticmethod
-    def parse_version(version_str: str):
-        return version.parse(version_str)
-
-    def create(self, path: str, channel: str) -> FmlClient:
-        return FmlClient.new_with_ref(
-            "@mozilla-mobile/firefox-android/fenix/app/nimbus.fml.yaml", "release", "main"
-        )
-
-    @staticmethod
-    def load_yaml(file):
-        return yaml.load(file.read(), Loader=yaml.Loader)
-
-    def get_repo_location(self, file_location=__base_path):
+    def get_application_data(app, file_location=__base_path):
         if os.path.exists(file_location):
             with open(file_location) as application_yaml_file:
-                application_data = self.load_yaml(application_yaml_file)
-                for feature_slug in application_data:
-                    if feature_slug in self.application:
-                        return application_data[feature_slug]["repo"]
-
-    def get_fm_path(self, file_location=__base_path):
-        if os.path.exists(file_location):
-            with open(file_location) as application_yaml_file:
-                application_data = self.load_yaml(application_yaml_file)
-                for feature_slug in application_data:
-                    if feature_slug in self.application:
-                        return application_data[feature_slug]["fml_path"]
+                file = yaml.load(application_yaml_file.read(), Loader=yaml.Loader)
+                for app_name in file:
+                    if app_name in app:
+                        return file[app_name]
 
     def get_manifest_paths(self, versions: list[str]):
         manifest_paths = []
         for v in versions:
-            version = self.parse_version(v)
-            if os.path.exists(self.__base_path):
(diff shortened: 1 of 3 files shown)


jhugman (commented):
This is a great start! I love the testing!

I've added a few comments and suggestions. 😄

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -1,40 +1,58 @@
...
+        self.application_data = self.get_application_data(application, file_location)
+
+    @staticmethod
+    def get_application_data(app, file_location=__base_path):
+        if os.path.exists(file_location):
+            with open(file_location) as application_yaml_file:
+                file = yaml.load(application_yaml_file.read(), Loader=yaml.Loader)
+                for app_name in file:
+                    if app_name in app:
+                        return file[app_name]
```
jaredlockhart, line 27: Just a thought, rather than touch disk lazily like this, we could just load all these files into memory at startup and then do dict lookups when data is requested.

eliserichards: I think that's the way we want to go eventually

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -1,37 +1,60 @@
...
 from rust_fml import FmlClient
 
 from experimenter.settings import BASE_DIR
 
-logger = logging.getLogger()
-
 
 class NimbusFmlLoader:
-    def __init__(self, application: str, channel: str):
+    __base_path = os.path.join(BASE_DIR, "features", "manifests", "apps.yaml")
```
brennie, line 11: Don't use `os.path`. Use `pathlib.Path`:

```python
from pathlib import Path

_base_path = Path(BASE_DIR) / "features" / "manifests" / "apps.yaml"

# ...

if file_location.exists():
  with file_location.open() as yaml_file:
    # ...
```

brennie, line 11: Does this need to use a double underscore? `__symbol` is only necessary to mangle it for subclass inheritance (e.g., this symbol will get mangled to `_NimbusFmlLoader__base_path`).

brennie, line 11: This is a constants and should be UPPER_CASE

experimenter/experimenter/features/tests/test.yaml
```diff
@@ -0,0 +1,9 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+---
+fenix:
+  repo: \@mozilla-mobile/firefox-android
```
brennie, line 6: I think we should just not include the @ in these and add the @ to the CLI invocations etc when necessary.

experimenter/experimenter/features/tests/test_nimbus_fml_loader.py
```diff
@@ -1,46 +1,129 @@
...
 from django.test import TestCase
+from packaging import version
 
 from experimenter.features.manifests.nimbus_fml_loader import NimbusFmlLoader
+from experimenter.settings import BASE_DIR
 
 
 class TestNimbusFmlLoader(TestCase):
-    def test_intiate_new_fml_client_no_error(self):
+    __base_path = os.path.join(BASE_DIR, "features", "manifests", "apps.yaml")
```
brennie, line 11: This definitely doesn't need to be __. Also, UPPER_CASE here too.

brennie (changes requested):

experimenter/experimenter/features/tests/test.yaml
eliserichards: agreed, let's keep it for now until we can get that into the CLI. I confirmed with James.

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
eliserichards: ahh that makes sense. Uppercase it is! 🙌

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -1,37 +1,60 @@
...
         self.channel: str = channel
+        self.application_data = self.get_application_data(application, file_location)
+        # Todo: Connect FML paths and channels https://mozilla-hub.atlassian.net/browse/EXP-3845
+        self.fml_client = self.create("", "")
+
+    @staticmethod
+    def get_application_data(app, file_location=__base_path):
+        if os.path.exists(file_location):
+            with open(file_location) as application_yaml_file:
+                file = yaml.load(application_yaml_file.read(), Loader=yaml.Loader)
```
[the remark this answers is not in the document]
eliserichards: Oooo good to know! I was using the same thing that was used in features `__init__.py` (https://github.com/mozilla/experimenter/blob/b2bc1a8e4725a6c5057434fba8305abbe9d4c455/experimenter/experimenter/features/__init__.py#L115) so I'll file something to change that one too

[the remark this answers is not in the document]
eliserichards: #9488

jhugman (approved):
This looks fine to me. Keep going!

PR #9456, opened 2023-09-20
fix #9453 feat(nimbus): ignore unset fields during sizing data serialization

### LINKED ISSUE #9453: Serialize sizing data without unset fields
So that the UI doesn't have to parse out `null` values when showing parameter values.

mikewilli:
Because

- the pop sizing UI shows the configured parameters that affect the precomputed sizing data
- mobile and desktop experiments have different parameters (locale vs language is what matters here)
- it would not be helpful to show, e.g., `locale: null, language: "EN"`

This commit

- ignores unset fields in the sizing data model during serialization so that the UI doesn't have to filter them out


jaredlockhart (approved):
Nice 👍

PR #9462, opened 2023-09-20
fix #9461 chore(project): parameterize setup docker in circle

### LINKED ISSUE #9461: Parameterize docker setup step in Circle file
We added a step to our Circle config file to login to Docker and setup the remote builder.  We need to use different accounts for each of Experimenter and Cirrus though, so rather than hard code the Experimenter credentials and override them with the Cirrus ones, let's just parameterize the Docker setup step to take in a username and password and explicitly use the correct ones for each step.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3862)

jaredlockhart:
Because

* We have many tasks in Circle that depend on a Docker account
* The docker account depends on which module is being tested ie Experimenter, Cirrus
* We should explicitly parameterize the Docker setup step with the credentials needed for that module

This commit

* Parameterizes the setup_docker step in circle
* Uses the appropriate credentials for each of Experimenter and Cirrus in each circle task

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,11 @@ commands:
             fi
   setup_docker:
     description: "Setup Docker with Hydrobuild"
+    parameters:
+      username:
+        type: string
+      password:
+        type: string
     steps:
       - run: |
           mkdir -vp ~/.docker/cli-plugins/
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
fantastic 👏 👏

PR #9468, opened 2023-09-21
fix #9466 chore(project): add top level dockerignore

### LINKED ISSUE #9466: Ignore node_modules in dockerignore
We need to ignore all node_modules paths in our dockerignore file.  We have a line for it but it's not qualified for arbitrary depth ie `**/`.  Let's add that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3865)

jaredlockhart:
Because

* We are now dockerizing new Nimbus components like schemas
* We need a top level docker ignore to prevent the build context from including things like node_modules

This commit

* Adds a top level dockerignore file

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+**/__pycache__
+**/.cache
+**/.coverage
+**/.DS_Store
+**/.mypy
+**/.pycache
+**/.pytest_cache
+**/.vscode
+**/.vscode/
+**/coverage/
+**/node_modules
+**/yarn-error.log


yashikakhurana (approved):
🚀

.dockerignore
```diff
@@ -0,0 +1,12 @@
+**/__pycache__
+**/.cache
+**/.coverage
+**/.DS_Store
+**/.mypy
+**/.pycache
+**/.pytest_cache
+**/.vscode
+**/.vscode/
```
brennie, line 9: Dupe?

PR #9473, opened 2023-09-21
fix #9472 chore(project): add arm64 support in circle

### LINKED ISSUE #9472: Add arm64 circle machines for make check
Now that we have binary assets as part of our CI flow, and those binary assets must be built for both amd64 and aarch64 (ie arm), we need our CI to validate both platforms.  Let's add an arm instance to our circle config for just `make check` to ensure that the containers will at least build and start.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3868)

jaredlockhart:
Because
    
* We now include binary assets that must target both amd64 and arm64
* We must verify that these are available and functioning in CI
    
This commit
    
* Adds an arm64 instance for cirrus and experimenter check to verify that our containers will build and run on both platforms
* Disables path checks for changes in the root project dir or .circleci folder or on main branch so that we run all tests if anything project wide changes
* Removes now redundant branch/main copies of experimenter since path checks will be disabled on main

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,13 +12,24 @@ commands:
       - run:
           name: Check file paths
           command: |
-            if git diff --name-only main HEAD | grep -E '<< parameters.paths >>'
+            # Check if CIRCLE_BRANCH is "main" and skip the check
+            if [ "$CIRCLE_BRANCH" == "main" ]; then
+                echo "On main branch. Skipping path checks."
+                exit 0
+            fi
+
+            diff_output=$(git diff --name-only main HEAD)
+            echo "Changed files:"
+            echo "$diff_output"
+
+            if echo "$diff_output" | grep -E '<< parameters.paths >>|^[^/]+$|^.circleci/'
               then
-                echo "Changes detected in << parameters.paths >> directory. Running tests and linting."
+                echo "Changes detected in << parameters.paths >> or .circleci or root directory. Running tests and linting."
               else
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -62,6 +62,25 @@ jobs:
             cp .env.sample .env
             make check
 
+  check_experimenter_branch_arm64:
```
brennie, line 65: can we call this _aarch64 so its more visually distinct?

.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ commands:
           fi
 
 jobs:
-  check_experimenter_branch:
+  check_experimenter_branch_amd64:
```
brennie, line 46: and maybe call this _x86_64 too?

.circleci/config.yml
```diff
@@ -643,8 +662,14 @@ workflows:
...
-      - check_experimenter_branch:
-          name: Check Experimenter (Branch)
+      - check_experimenter_branch_amd64:
+          name: Check Experimenter amd64 (Branch)
+          filters:
+            branches:
+              ignore:
+                - main
+      - check_experimenter_branch_arm64:
+          name: Check Experimenter arm64 (Branch)
```
brennie, line 672: here too

brennie (approved):
r+ with some suggestions

brennie (approved):

PR #9498, opened 2023-09-25
fix #9487 chore(project): use faker.unique to prevent uniqueness errors

### LINKED ISSUE #9487: Use faker.unique for slugs
We use the Python library faker to generate slugs and other data in our test runs, but occasionally it will generate the same slug twice and tests will fail becuase of a uniqueness violation.  Faker has a [uniqueness](https://faker.readthedocs.io/en/stable/fakerclass.html#unique-values) mechanism to prevent this case.  Let's update our factories to use that and prevent generating duplicate slugs and failing tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3881)

jaredlockhart:
Because

* We use Faker in our unit tests to generate slugs for test factories
* Sometimes it will generate the same value twice and cause a database uniqueness violation
* Faker includes a unique feature that will prevent this

This commit

* Updates all factories to use faker.unique for all unique fields like names/slugs

### DIFF
--- a/experimenter/experimenter/base/tests/factories.py
+++ b/experimenter/experimenter/base/tests/factories.py
@@ -1,9 +1,9 @@
 import factory
-from faker import Factory as FakerFactory
+from faker import Faker
 
 from experimenter.base.models import Country, Language, Locale
 
-faker = FakerFactory.create()
+faker = Faker()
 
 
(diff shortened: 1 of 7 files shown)


yashikakhurana (approved):
great improvement 👏

PR #9504, opened 2023-09-26
fix #9501 feat(cirrus): update variables prefix

### LINKED ISSUE #9501: Update cirrus env to start with prefix CIRRUS_
Env variables starts with different name and its hard to check value in the implementing application, so we should rename and starts all cirrus env with CIRRUS_ prefix

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3888)

yashikakhurana:
Because

- ENV variables are being used in cirrus, It is easy to check the set values if all start with Prefix `CIRRUS_`
- sets the app id and channel from the config for glean initialization

This commit

- Adds a prefix to all .env variables
- Update glean config to use app id and channel correctly

fixes #9501


jaredlockhart (approved):
Fantastic 🎉

PR #9506, opened 2023-09-27
add support to target 118.0.1 users

cbellini:
Because

- we will be updating a frecency experiment to target users with 118.0.1

This commit

- adds 118.0.1 as a selectable version target


jaredlockhart (approved):
Thanks @cbellini ! 🎉

PR #9516, opened 2023-09-28
add experiment support for targeting users with 118.0.2

cbellini:
Because

- adding support to target Frecency experiment improvements to 118.0.2

This commit

- adds 118.0.2 as a target


jaredlockhart (approved):
Thnx @cbellini ! 🎉

PR #9520, opened 2023-09-29
fix #9419 feat(nimbus): use qcdou instead of dou for desktop guardrail results

mikewilli:
Because

- DOU causes confusion vs DAU
- QCDOU is already calculated for desktop experiments

This commit

- puts QCDOU in the highlights/guardrails sections of the Results page for Desktop experiments


mikewilli:
cc @danielkberry

jaredlockhart (approved):
LGTM thnx @mikewilli 🎉

PR #9546, opened 2023-10-03
fix #9545 bug(project): docker login necessary for deploy

### LINKED ISSUE #9545: Docker must be logged in to deploy
After disabling hydrobuild, we neglected to make sure the docker client is logged in to push images during deploy.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3914)

jaredlockhart:
Because

* When we disabled hydrobuild we also disabled logging into docker
* Logging into docker is necessary at the deploy step

This commit

* Logs into docker if possible but disables hydrobuild


yashikakhurana (approved):
ahaa, got it

PR #9548, opened 2023-10-04
fix #9539 feat(nimbus): Add custom targeting for at risk users background task notification.

### LINKED ISSUE #9539: Add custom targeting for at risk users background task notification (PiP notification experiment)
This tracks adding custom targeting for background task messaging.  The definition is for "at risk users" and is intended for use in the Picutre-in-picture (PiP) notification experiment.

* Windows 10+
* running background task
* >= 20 days of use in the last month (28 days)
* last active between [7-28) days ago

ncalexan:
Because

- the PiP notification has particular targeting needs

This commit

- adds custom targeting for this notification

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1127,6 +1127,42 @@ WINDOWS_10_PLUS_BACKGROUND_TASK_NOTIFICATION_ = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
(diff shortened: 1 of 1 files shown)


ncalexan:
I've tested this locally by a complicated method: I stitch this `targeting` into a JSON file and use special purpose `firefox.exe --backgroundtask message ...` invocations to witness the various boolean states as I modify parts of the default profile targeting snapshot.  I did catch an error with this approach: I originally had `.length`, which is accepted but produces an incorrect value; that is updated now to `|length`, which does the correct thing.

jaredlockhart (approved):
Cool thanks @ncalexan and good to test.  Your point about `.length` vs `|length` is tricky, I've experimented with validating JEXL subclauses but it's still difficult to catch everything so it's good to still manually verify.

PR #9558, opened 2023-10-05
Adds advanced targeting for Windows 10 users

danielkberry:
Because

- We'd like to be able to target any windows 10 user

This commit

- Adds the advanced targeting constant


jaredlockhart (approved):
LGTM thnx @danielkberry 👍

PR #9565, opened 2023-10-06
fix #9562 feat(nimbus): Add custom targeting for new non-default user…

### LINKED ISSUE #9562: Add custom targeting for new non-default users for background task notification (EU notification experiment)
This tracks adding custom targeting for background task messaging. The definition is for "new non-default users" and is intended for use in the EU (formerly Challenge the Default) notification experiment.

* Windows 10+
* running background task
* this Firefox is not the user's default browser
* profile creation date exactly 7 days ago

Note to future me: the exact date comparison is not the most robust, but if these users are already using Firefox enough for the precise days to cause them to not be targeted, that's a good problem to have.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3943)

ncalexan:
…s for background task notification (EU notification experiment).

Because

- We need new targeting for the EU notification experiment.

This commit

- Adds the targeting.

The unusual construction is because the math is floating point, and I don't see a JS-y way (`Math.floor`, `| 0`, `~~`) to convert to an integer.  (Aside: this very restricted JEXL environment is frustrating.)

Heavily tested manually.

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1163,6 +1163,39 @@ WINDOWS_10_PLUS_BACKGROUND_TASK_NOTIFICATION_AT_RISK_USER = NimbusTargetingConfi
     application_choice_names=(Application.DESKTOP.name,),
 )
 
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Complex but seems well formed 👍

PR #9598, opened 2023-10-17
bug(nimbus): Refactor FeatureManifest schema

brennie:
Because:
- the feature schema was causing superfluous errors that weren't really
  errors (e.g., complaining that has_exposure must be true/false) when
  running into real errors; and
- the hasExposure field is required by desktop;

This commit:
- adds a new Feature type which is a discriminated union, which will
  prevent these errors; and
- marks the hasExposure field as a required field.


schemas/mozilla_nimbus_schemas/experiments/feature_manifests.py
```diff
@@ -50,13 +50,19 @@ class FeatureWithExposure(BaseFeature):
...
+    has_exposure: Literal[False] = Field(alias="hasExposure")
 
     @property
     def exposure_description(self):
         return None
 
 
+class Feature(BaseModel):
+    __root__: Union[FeatureWithExposure, FeatureWithoutExposure] = Field(
+        discriminator="has_exposure"
```
mikewilli, line 62: Neat!

mikewilli (approved):
LGTM

PR #9611, opened 2023-10-18
fix #9610 bug(nimbus): metrics key may be missing

### LINKED ISSUE #9610: 'metrics' key may be absent in outcomes
An outcome has recently landed [here](https://github.com/mozilla/experimenter/pull/9607/files) that is missing the `metrics` key.  Let's update our outcome loader to check whether that key is present.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3957)

jaredlockhart:
Because

* The metrics key may be absent in an outcome file

This commit

* Checks for the presence of the metrics key before attempting to parse metrics

### DIFF
--- a/experimenter/experimenter/outcomes/tests/test_outcomes.py
+++ b/experimenter/experimenter/outcomes/tests/test_outcomes.py
@@ -15,7 +15,7 @@ class TestOutcomes(TestCase):
 
     def test_load_all_outcomes_and_ignore_examples(self):
         outcomes = Outcomes.all()
-        self.assertEqual(len(outcomes), 4)
+        self.assertEqual(len(outcomes), 5)
         self.assertIn(
             Outcome(
                 application=NimbusExperiment.Application.FENIX,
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/outcomes/tests/test_outcomes.py
```diff
@@ -83,6 +94,17 @@ def test_load_outcomes_by_application(self):
...
                 desktop_outcomes,
             )
+        self.assertIn(
+            Outcome(
+                application=NimbusExperiment.Application.DESKTOP,
+                description="a test outcome missing metrics",
+                friendly_name="test outcome",
+                slug="missing_metrics",
+                is_default=False,
+                metrics=[],
```
yashikakhurana, line 104: do we have a test where metrics is not default `[]`?

yashikakhurana (commented):
Overall looks good, just one question about test case where metrics key is present and non default value

experimenter/experimenter/outcomes/tests/test_outcomes.py
jaredlockhart: The other tests and fixtures cover the cases where metrics are present.

yashikakhurana (approved):
:shipit:

PR #9616, opened 2023-10-19
fix #9271 chore(project): dockerize schemas

### LINKED ISSUE #9271: Migrate schemas make commands to docker
Related to this ticket: https://github.com/mozilla/experimenter/pull/9157#discussion_r1291691246, but in general this follows the pattern used elsewhere in the repo and is safer practice.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3762)

jaredlockhart:
Because

* As we add a wider variety of elements to the Nimbus project including binary packages it becomes increasingly important to have a consistent environment for each of our modules
* Using Docker allows us to control the environment each module runs in and use a consistent base layer
* Dockerizing the schemas package will allow us to use identical Python, Node, and Rust environments as the other modules in Nimbus

This commit

* Wraps the schemas module in a Dockerfile that is consistent with the other modules in Nimbus


Makefile
```diff
@@ -246,43 +246,54 @@ build_demo_app:
 
 
 # nimbus schemas package
-SCHEMAS_VERSION_FILE := schemas/VERSION
-SCHEMAS_VERSION := $(shell cat ${SCHEMAS_VERSION_FILE})
+SCHEMAS_ENV ?=  # This is empty by default
+SCHEMAS_VERSION = \$$(cat VERSION)
+SCHEMAS_RUN = docker run -ti $(SCHEMAS_ENV) -v ./schemas:/schemas -v /schemas/node_modules schemas:dev sh -c
+SCHEMAS_BLACK = poetry run black --check --diff .
```
mikewilli, line 252: Do we still need `poetry run` for all these commands? I was originally using this so that things would run in the poetry-managed venv, but with docker I think we can just run the commands directly (as long as we install them in the container).

Makefile
```diff
@@ -246,43 +246,54 @@ build_demo_app:
...
+SCHEMAS_RUFF = poetry run ruff .
+SCHEMAS_DIFF_PYDANTIC = \
+	poetry run pydantic2ts --module mozilla_nimbus_schemas.__init__ --output /tmp/test_index.d.ts --json2ts-cmd 'yarn json2ts' &&\
+	diff /tmp/test_index.d.ts index.d.ts || (echo nimbus-schemas typescript package is out of sync please run make schemas_build;exit 1) &&\
+	echo 'Done. No problems found in schemas.'
+SCHEMAS_TEST = pytest
```
mikewilli, line 258: If we're using `poetry run` elsewhere, why not here?

schemas/VERSION
```diff
@@ -1 +1 @@
-2023.10.2
+2023.10.3
```
mikewilli, line 1: Is this version bump just so that we test that it still works when it merges to main?

schemas/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+# Credit: https://gist.github.com/kekru/8ac61cd87536a4355220b56ae2f4b0a9
+COPY . /schemas/
+RUN chmod -R 555 /schemas \
+    && chown -R root:root /schemas \
+    && find /schemas -exec touch -a -m -t 201512180130.09 {} \;
+
+
+#-------------------------
+# System packages
+FROM python:3.11.1 AS system-builder
```
mikewilli, line 16: Ooh we're on python 3.11 now, neat!

schemas/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update
+RUN apt-get --no-install-recommends install -y apt-utils ca-certificates postgresql-client yarn parallel
```
mikewilli, line 21: `postgresql-client`?

schemas/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+
+# Install nvm with node and npm
+ENV NODE_VERSION=16.19.0
```
mikewilli, line 24: This is EOL as of August. Maybe not worth worrying about on this PR but maybe worth a separate ticket (across the whole repo if needed).

mikewilli (approved):
r+wc Can you just update the schemas/README to match the new make commands?

I have a few questions, but overall looks good and worked well in testing :tada:

schemas/Dockerfile
jaredlockhart: https://github.com/mozilla/experimenter/issues/9635

schemas/VERSION
jaredlockhart: Yeah I had to trigger a deploy to test it out.

PR #9620, opened 2023-10-19
fixes #9619 feat(nimbus): soft fail when fetches fail

### LINKED ISSUE #9619: Soft fail when fetches fail
We should not fail the entire fetch task if a single fetch fails. This is especially important as we start fetching versioned releases, which will result in a lot of fetches that could potentially fail.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3959)

brennie:
Because

- a single fetch failure currently breaks the entire fetch task; and
- we are about to drastically increase the number of fetches per task

This commit

- makes fetch failures not cause the tool to exit; and
- reports a summary of all successes and failures at the end of the task.

### DIFF
--- a/experimenter/manifesttool/cli.py
+++ b/experimenter/manifesttool/cli.py
@@ -1,13 +1,10 @@
-import sys
 from dataclasses import dataclass
 from pathlib import Path
 
 import click
-import yaml
-from mozilla_nimbus_schemas import FeatureManifest
 
-from manifesttool import github_api, hgmo_api, nimbus_cli
-from manifesttool.appconfig import AppConfig, AppConfigs, RepositoryType
+from manifesttool.appconfig import AppConfigs
+from manifesttool.fetch import fetch_fml_app, fetch_legacy_app
 
 MANIFESTS_DIR = Path(__file__).parent.parent / "experimenter" / "features" / "manifests"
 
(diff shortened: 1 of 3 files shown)


experimenter/manifesttool/fetch.py
```diff
@@ -0,0 +1,156 @@
...
+
+        print(f"fetch-latest: {app_name}: generate experimenter.yaml")
+        # The single-file fml file for each channel will generate the same
+        # experimenter.yaml, so we can pick any here.
+        nimbus_cli.generate_experimenter_yaml(
+            app_config,
+            channels[0],
+            manifest_dir,
+        )
+    except Exception as e:
```
jaredlockhart, line 65: image

experimenter/manifesttool/fetch.py
```diff
@@ -0,0 +1,156 @@
...
+        for result in results:
+            if result.exc is None:
+                print(f"{result.app_name} at {result.ref_name} ({result.ref})")
+
+    print("")
+    if any(result.exc is not None for result in results):
+        print("FAILURES:\n")
+        for result in results:
+            if result.exc is not None:
+                print(f"{result.app_name} at {result.ref_name} ({result.ref})")
```
jaredlockhart, line 156: Couldn't you just write this as

```py
        for result in results:
            if result.exc is None:
                print(f"SUCCESS: {result.app_name} at {result.ref_name} ({result.ref})")
            else:
                print(f"FAILED: {result.app_name} at {result.ref_name} ({result.ref})
```

experimenter/manifesttool/cli.py
```diff
@@ -42,112 +39,16 @@ def fetch_latest(ctx: click.Context):
...
         context.manifest_dir.joinpath(app_config.slug).mkdir(exist_ok=True)
 
         if app_config.fml_path is not None:
-            fetch_fml_app(context, app_name, app_config)
+            results.append(fetch_fml_app(context.manifest_dir, app_name, app_config))
         elif app_config.experimenter_yaml_path is not None:
-            fetch_legacy_app(context, app_name, app_config)
+            results.append(fetch_legacy_app(context.manifest_dir, app_name, app_config))
         else:  # pragma: no cover
             assert False, "unreachable"
```
jaredlockhart, line 52: Wait why is this here?

jaredlockhart (approved):
Cool yeah good to handle this case now 👍

experimenter/manifesttool/cli.py
brennie: We don't actually enforce that one of fml_path or experimenter_yaml_path is non none but in practice there is one or the other. Ill follow up to make it required in the model

experimenter/manifesttool/fetch.py
brennie: This groups all the successes and all the failures

PR #9623, opened 2023-10-19
feat(nimbus): Explicitly require a manifest path in AppConfig

brennie:
Because

- we want to ensure that apps.yaml is always valid; and
- we want to ensure that each app specifies exactly one manifest file

This commit

- adds a test that loads the production apps.yaml;
- updates the AppConfig model with a validator to ensure only one
  manifest path can be specified

### DIFF
--- a/experimenter/manifesttool/tests/test_appconfig.py
+++ b/experimenter/manifesttool/tests/test_appconfig.py
@@ -1,0 +1,41 @@
+from unittest import TestCase
+
+from manifesttool.cli import MANIFESTS_DIR
+from manifesttool.appconfig import AppConfigs
+
+
+class AppConfigTests(TestCase):
+    def test_parse_apps_yaml(self):
+        """Testing that we can parse apps.yaml."""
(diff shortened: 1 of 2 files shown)


experimenter/manifesttool/tests/test_appconfig.py
```diff
@@ -0,0 +1,11 @@
...
+
+from manifesttool.cli import MANIFESTS_DIR
+from manifesttool.appconfig import AppConfigs
+
+
+class AppConfigTests(TestCase):
+    def test_parse_apps_yaml(self):
+        """Testing that we can parse apps.yaml."""
+        apps_yaml_path = MANIFESTS_DIR / "apps.yaml"
+        AppConfigs.load_from_file(apps_yaml_path)
```
jaredlockhart, line 11: What about tests for the two failure cases? 🤔

jaredlockhart (changes requested):
Good idea 👍

jaredlockhart (approved):
Yep looks good thnx @brennie 🎉

PR #9634, opened 2023-10-24
fix #9633 feat(project): push docker images to gcr

### LINKED ISSUE #9633: Push to Google Container Registry
We've been seeing a lot of intermittent failures with image pulls from Dockerhub, so let's start pushing to Google Container Registry and if the intermittency doesn't resolve we can switch Jenkins to pull from GCR instead of Dockerhub.  We can replicate the push patterns [here](https://github.com/mozilla/taar_gcp_etl/blob/master/.circleci/config.yml#L32-L51).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3962)

jaredlockhart:
Because

* We've been seeing consistent intermittent failures with pulling images from dockerhub
* An alternative is to use the Google Container Registry instead of Dockerhub
* If we have better luck with Google Container Registry then we can use that and deprecate Dockerhub

This commit

* Pushes to Google Container Registry in addition to Dockerhub

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 orbs:
-  gh: circleci/github-cli@2.0
+  gh: circleci/github-cli@2.2
+  gcp-cli: circleci/gcp-cli@3.1.1
 commands:
   check_file_paths:
     description: "Check file paths"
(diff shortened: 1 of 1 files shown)


brennie (approved):

haroldwoo (approved):

haroldwoo:
if we decide to deprecate dockerhub sooner than gcp v2 migration, we need a better story around spoorfing a webhook to dockerhub, and probably use circleci contexts with a new SA to push images to a better location in GCR.

PR #9639, opened 2023-10-25
Fixes #9638 tests(nimbus): Set env variable precedence in Docker files

### LINKED ISSUE #9638: Point integration test Docker files to correct rust location
Breaking this out of [https://mozilla-hub.atlassian.net/browse/EXP-3876](https://mozilla-hub.atlassian.net/browse/EXP-3876|smart-link). Context [here](https://github.com/mozilla/experimenter/pull/9618#issuecomment-1778211333)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3968)

eliserichards:
Because

* We were referencing the megazord correctly for the top level dockerfile but not for integration tests

~This commit~
* ~`tox.ini` (for integration tests) was referencing the wrong place for AS: `application-services/components/nimbus/src` instead of `application-services/megazord`~
* ~in the top level dockerfile, we needed to copy the AS megazord `--from=dev`~
* ~[according to this](https://docs.docker.com/compose/environment-variables/envvars-precedence/), we weren't correctly setting the precedence for the env variable `PYTHONPATH` -- added `PYTHONPATH` to the `.env` files and the rust image dockerfile for integration tests~
* ~the rust image dockerfile was also referencing `PYTHON_PATH` instead of `PYTHONPATH`~

See comment below.

### DIFF
--- a/experimenter/Dockerfile
+++ b/experimenter/Dockerfile
@@ -179,6 +179,7 @@ RUN apt-get --no-install-recommends install -y apt-utils ca-certificates postgre
 COPY --from=dev /usr/local/bin/ /usr/local/bin/
 COPY --from=dev /usr/local/lib/python3.11/site-packages/ /usr/local/lib/python3.11/site-packages/
 COPY --from=dev /experimenter/bin/ /experimenter/bin/
+COPY --from=dev /application-services/megazord/ /application-services/megazord/
 COPY --from=file-loader /experimenter/manage.py /experimenter/manage.py
 COPY --from=file-loader /experimenter/experimenter/ /experimenter/experimenter/
 COPY --from=ui /experimenter/experimenter/legacy/legacy-ui/assets/ /experimenter/experimenter/legacy/legacy-ui/assets/
--- a/experimenter/Dockerfile
+++ b/experimenter/Dockerfile
@@ -186,3 +186,4 @@ COPY --from=ui /experimenter/experimenter/legacy/legacy-ui/assets/ /experimenter
 COPY --from=ui /experimenter/experimenter/nimbus-ui/build/ /experimenter/experimenter/nimbus-ui/build/
 COPY --from=ui /experimenter/experimenter/theme/static/ /experimenter/experimenter/theme/static/
 COPY --from=ui /experimenter/experimenter/static/scripts/ /experimenter/experimenter/static/scripts/
+ENV PYTHONPATH=$PYTHONPATH:/application-services/megazord
--- a/experimenter/tests/integration/tox.ini
+++ b/experimenter/tests/integration/tox.ini
@@ -22,7 +22,7 @@ commands =
     pip install -r ../requirements.txt
     pytest --verify-base-url --base-url https://nginx/nimbus/ --html=test-reports/report.htm --self-contained-html --driver Firefox nimbus/test_mobile_targeting.py {posargs} -vvv
 setenv =
-    PYTHONPATH=/application-services/megazord
+    PYTHONPATH=/application-services/components/nimbus/src
 
 [pytest]
 addopts = -p no:warnings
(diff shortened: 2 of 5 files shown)


jrbenny35 (approved):
LGTM

jrbenny35:
I think you need to change this, https://github.com/mozilla/experimenter/blob/main/experimenter/tests/integration/nimbus/utils/Dockerfile-rust-image#L50

eliserichards:
> I think you need to change this, https://github.com/mozilla/experimenter/blob/main/experimenter/tests/integration/nimbus/utils/Dockerfile-rust-image#L50

What do you mean? I changed it from `PYTHON_PATH` -> `PYTHONPATH` in `Dockerfile-rust-image`

jrbenny35:
> > I think you need to change this, https://github.com/mozilla/experimenter/blob/main/experimenter/tests/integration/nimbus/utils/Dockerfile-rust-image#L50
> 
> @jrbenny35 What do you mean? I changed it from `PYTHON_PATH` -> `PYTHONPATH` in `Dockerfile-rust-image`

Oh I missed that, sorry, I shouldn't code review after hours 😅

experimenter/tests/integration/tox.ini
```diff
@@ -22,7 +22,7 @@ commands =
     pip install -r ../requirements.txt
     pytest --verify-base-url --base-url https://nginx/nimbus/ --html=test-reports/report.htm --self-contained-html --driver Firefox nimbus/test_mobile_targeting.py {posargs} -vvv
 setenv =
-    PYTHONPATH=/application-services/components/nimbus/src
```
jrbenny35, line 25: I suspect this might be the problem. Does the new path point to where the nimbus.py file is created?

jrbenny35 (changes requested):

experimenter/tests/integration/tox.ini
eliserichards: I thought the same thing, but it should be the correct place: 
```COPY --from=rust-builder /application-services/megazord/nimbus.py /application-services/megazord/rust_sdk.py```

**Comment by eliserichards** [comment]
Let's try this again. 

The top-level `Dockerfile` is setting the `PYTHONPATH` when it makes the `dev` image, but then it needs to do the following on the `deploy` image:
* copy the AS megazord from the `dev` image
* set the `PYTHONPATH`  to point to the megazord

That way, when the `Dockerfile-rust-image` is built on top of the `deploy` image, it has the correct location for the megazord and can set the `PYTHONPATH` correctly.

**Comment by eliserichards** [comment]
giphy

**Review by jrbenny35 — APPROVED** [review]

File: experimenter/Dockerfile [hunk 2]
```diff
@@ -179,9 +179,11 @@ RUN apt-get --no-install-recommends install -y apt-utils ca-certificates postgre
...
 COPY --from=dev /usr/local/lib/python3.11/site-packages/ /usr/local/lib/python3.11/site-packages/
 COPY --from=dev /experimenter/bin/ /experimenter/bin/
+COPY --from=dev /application-services/megazord/ /application-services/megazord/
 COPY --from=file-loader /experimenter/manage.py /experimenter/manage.py
 COPY --from=file-loader /experimenter/experimenter/ /experimenter/experimenter/
 COPY --from=ui /experimenter/experimenter/legacy/legacy-ui/assets/ /experimenter/experimenter/legacy/legacy-ui/assets/
 COPY --from=ui /experimenter/experimenter/nimbus-ui/build/ /experimenter/experimenter/nimbus-ui/build/
 COPY --from=ui /experimenter/experimenter/theme/static/ /experimenter/experimenter/theme/static/
 COPY --from=ui /experimenter/experimenter/static/scripts/ /experimenter/experimenter/static/scripts/
+ENV PYTHONPATH=$PYTHONPATH:/application-services/megazord
```
jaredlockhart, line 189: Heyyyyyy look at that.  Yeah that's a tricky little bit of trickiness, but this looks right to me.  Good job @eliserichards .  You know what would've helped with this is having an integration test that stresses this.  Then we would've found it before we made the changes.  We should make sure we get an integration test that hits this soon 🙏 thnx @eliserichards this is great 🎉 🎉 🎉

Issue #9640, opened 2023-10-26
Android for fakespot - early and later day user targeting


shell1:
or the real fakespot stat sig review checker experiment (not the beta channel smoketest experiment) we want to add criteria for early days or not. This is so we can have 2 experiments and size for early users (fewer) and later (more).

Later days users: PROFILE28DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 >= 28"
Early Days user: PROFILELESSTHAN28DAYS = "(currentDate|date - profileAgeCreated|date) / 86400000 < 28"

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3969)

shell1:
PR up https://github.com/mozilla/experimenter/pull/9642

cpeterso:
PR https://github.com/mozilla/experimenter/pull/9642 has been merged, so I think we can close this issue.

PR #9642, opened 2023-10-27
fix #9564, #9640 feature(nimbus): add targeting for new and later users for Firefox iOS and Android

### LINKED ISSUE #9564: Custom targeting for iPhone
For the Fakespot 120 experiment, we need to be able to target only iPhone users (ie no iPad users). We can use the `deviceModel` field.

We will need 2 advanced targeting criteria with this iPhone users only:

1. iphone only + new users only
1. iphone only + existing (or not new) users only.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3944)

jeddai:
Because

- New advanced targeting configurations are needed for Fakespot experiments

This commit

- Adds the four advanced targeting options required

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1669,6 +1669,54 @@ WINDOWS_10_PLUS = NimbusTargetingConfig(
 )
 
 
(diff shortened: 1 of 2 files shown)


jeddai:
@adudenamedruby @thatswinnie Here's the PR

brennie (approved):

PR #9646, opened 2023-10-28
add support for targeting 121

cbellini:
Because

- we have experiments that will target features that will be available in Fx 121

This commit

- adds Fx 121 as a target version


experimenter/experimenter/experiments/constants.py
```diff
@@ -212,7 +212,8 @@ class Application(models.TextChoices):
         APPLICATION_CONFIG_MONITOR_WEB.slug,
         APPLICATION_CONFIG_MONITOR_WEB.name,
     )
-    DEMO_APP = (APPLICATION_CONFIG_DEMO_APP.slug, APPLICATION_CONFIG_DEMO_APP.name)
+    DEMO_APP = (APPLICATION_CONFIG_DEMO_APP.slug,
+                APPLICATION_CONFIG_DEMO_APP.name)
```
jrbenny35, line 216: Was this change needed? Maybe you were using a formatter or something

jrbenny35 (changes requested):
Just 1 question

experimenter/experimenter/experiments/constants.py
cbellini: I was using a formatter, yes, and it aligned it to the format used with other lines, but I can revert it to the original formatting.

brennie (approved):

PR #9647, opened 2023-10-28
fixes #9027 feat(nimbus): hide enrollment duration for rollouts

### LINKED ISSUE #9027: Hide enrollment duration for rollouts
Because we disabled the ability to pause rollouts, the enrollment duration field has no meaning and so we should hide this field altogether for rollouts.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3635)

tj-26:
Because

- Because the ability to pause rollouts was disabled so, the enrollment-duration field has no meaning and it should not be visible for rollouts.

This commit

- Hides the enrolment duration if the experiment is a rollout.

Screenshot 2023-10-28 at 3 37 12 PM

fixes #9027


experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -2076,6 +2076,44 @@ it("enables save buttons when not archived", async () => {
...
+        channel: NimbusExperimentChannelEnum.NIGHTLY,
+        status: NimbusExperimentStatusEnum.DRAFT,
+        isRollout: false,
+        firefoxMinVersion: NimbusExperimentFirefoxVersionEnum.FIREFOX_107,
+      }}
+    />,
+  );
+
+  await waitFor(() => {
+    expect(screen.queryByTestId("proposedEnrollment")).toBeInTheDocument();
```
brennie, line 2113:
```suggestion
    screen.getByTestId("proposedEnrollment");
```

`getByTestId` will throw if the element is not found.

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
```diff
@@ -2076,6 +2076,44 @@ it("enables save buttons when not archived", async () => {
...
+        channel: NimbusExperimentChannelEnum.NIGHTLY,
+        status: NimbusExperimentStatusEnum.DRAFT,
+        isRollout: true,
+        firefoxMinVersion: NimbusExperimentFirefoxVersionEnum.FIREFOX_107,
+      }}
+    />,
+  );
+
+  await waitFor(() => {
+    expect(screen.queryByTestId("proposedEnrollment")).not.toBeInTheDocument();
```
brennie, line 2094:
```suggestion
    expect(screen.queryByTestId("proposedEnrollment")).toBeNull();
```

queryByTestId returns null if not found, so we're testing if `null` will be in the document. It will never be true, but semantically it makes more sense to check for null in the return value.

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -801,6 +773,38 @@ export const FormAudience = ({
               <FormErrors name="proposedDuration" />
             </InputGroup>
           </Form.Group>
+
+          {experiment.isRollout ? (
```
eliserichards, line 777: This change also switches the order of "Experiment duration" and "Enrollment period", right? Can we keep the order the same (e.g. for non-rollouts, "Enrollment period" is on the left and "Experiment duration" is on the right):
image

eliserichards (changes requested):
Thanks for the contribution @tj-26! I have one suggestion:

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
eliserichards: (For context, if we change the order of these fields it will definitely confuse our users and they may enter incorrect values)

eliserichards: I think we would actually like to have the field greyed out instead of totally hidden (see comment [here](https://github.com/mozilla/experimenter/issues/9027#issuecomment-1789616457)). Sorry to change the requirements on you mid-PR 😅

tj-26: Hi @eliserichards. No Worries. Thanks for letting me know. I'll make the change.

tj-26:
Updated Behaviour: Now on rollouts the enrolment period field will be greyed out(disabled) instead of hidden completely.
Also, because the disabled field visibility is not great I added a tooltip that only shows on rollouts to provide more information.
Screenshot 2023-11-06 at 10 54 44 PM

yashikakhurana (approved):
Thank you @tj-26 for contribution,  looks perfect 🎉 🎉

PR #9656, opened 2023-11-01
fix #9651 feat(nimbus): api v6 uses actual enrollment end date or null

### LINKED ISSUE #9651: API (v6) should report the official enrollment end date
The API currently reports the enrollment end date regardless of experiment status. While enrollment is ongoing, a proposed enrollment end date is calculated based on experiment start date and proposed enrollment duration. After enrollment finishes, the enrollment end date becomes final and is reported. However, consumers (like Jetstream) have no way to know when this date is proposed vs final.

We should add a new field on the Experiment model that is the official enrollment end date and is null until enrollment ends, at which point it will be the date. This is what should be reported by the API.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3974)

mikewilli:
Because

- Cirrus uses continuous enrollment (i.e., enrollment can't end until the experiment does)
- Jetstream requires an end for enrollment before the experiment ends for analysis
- We can't set `isEnrollmentPaused` to true without prematurely ending the experiment in Cirrus context

This commit

- Adds an `actual_enrollment_end_date` field to the NimbusExperiment model
- uses `actual_enrollment_end_date` for the serialized `enrollmentEndDate` in API v6

### DIFF
--- a/experimenter/experimenter/experiments/api/v6/serializers.py
+++ b/experimenter/experimenter/experiments/api/v6/serializers.py
@@ -100,7 +100,7 @@ class NimbusExperimentSerializer(serializers.ModelSerializer):
     probeSets = serializers.ReadOnlyField(default=[])
     outcomes = serializers.SerializerMethodField()
     startDate = serializers.DateField(source="start_date")
-    enrollmentEndDate = serializers.DateField(source="computed_enrollment_end_date")
+    enrollmentEndDate = serializers.DateField(source="actual_enrollment_end_date")
     endDate = serializers.DateField(source="end_date")
     proposedDuration = serializers.ReadOnlyField(source="proposed_duration")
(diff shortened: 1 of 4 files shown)


jaredlockhart (approved):
Oh yeah, thank you @mikewilli 👍 Can we safely land this first and then make changes to jetstream or do we need to do it the other way around?

Also there's a better pattern for doing this but we'd need to move a bunch of methods around so we can just do it as a followup.

mikewilli:
Yep, it will be safe for Jetstream this way -- the only thing Jetstream will need to change after this is to remove the `isEnrollmentPaused` check that we added last half, since it's no longer necessary and will break things for Cirrus experiments.

PR #9684, opened 2023-11-07
fix #9678 feat(nimbus): Add is_web flag for web apps

### LINKED ISSUE #9678: Add is_web to App config
We have introduced experimenter support for the web applications and it requires some changes on the console app to hide/show certain fields, let’s add `is_web` property to the applications which are supported by Cirrus

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3994)

yashikakhurana:
Because

- We want to make changes on UI based on if its a web app or not (cirrus)

This commit

- Add a new property `is_web` to the application config
Screenshot 2023-11-06 at 6 00 03 PM
Screenshot 2023-11-06 at 5 56 57 PM

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/types.py
+++ b/experimenter/experimenter/experiments/api/v5/types.py
@@ -468,6 +468,7 @@ class NimbusExperimentType(DjangoObjectType):
     is_localized = graphene.Boolean()
     is_rollout = graphene.Boolean()
     is_sticky = graphene.Boolean()
+    is_web = graphene.Boolean()
     jexl_targeting_expression = graphene.String()
     languages = graphene.List(graphene.NonNull(NimbusLanguageType), required=True)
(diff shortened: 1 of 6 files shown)


experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -468,6 +468,7 @@ class NimbusExperimentType(DjangoObjectType):
     is_localized = graphene.Boolean()
     is_rollout = graphene.Boolean()
     is_sticky = graphene.Boolean()
+    is_web = graphene.Boolean()
```
brennie, line 471: This should use `NonNull` because every application has `is_web` defined.

```suggestion
    is_web = graphene.NonNull(graphene.Boolean())
```

brennie (changes requested):

jaredlockhart (approved):
🙏🙏🙏🙏 💯 💯 💯 Web applications!!!

PR #9690, opened 2023-11-07
fix #9687 chore(project): lock application services version

### LINKED ISSUE #9687: Capture Application Services version hash in Dockerfile
As part of the Dockerfiles for Experimenter and Cirrus, we [fetch Application Services](https://github.com/mozilla/experimenter/blob/main/experimenter/Dockerfile#L86).  However we presently fetch the latest `main` branch which may introduce breaking changes to Cirrus and Experimenter.  To prevent break changes appearing, let's hard code a commit hash as part of the reference here, and then we can increment that reference as part of PRs that will consume and adapt to changes in Application Services.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4003)

jaredlockhart:
Because

* Changes to Application Services can break consuming code in Cirrus/Experimenter
* We currently consume the main branch of Application Services
* This means that changes can land in main in Application Services and break main in Cirrus/Experimenter
* We also neglected to explicitly rebuild the Cirrus test container when running Cirrus tests

This commit

* Explicitly locks Application Services to a specific hash when consumed in Cirrus/Experimenter
* Explicitly builds the test container for Cirrus when running Cirrus tests

### DIFF
--- a/cirrus/server/Dockerfile
+++ b/cirrus/server/Dockerfile
@@ -1,8 +1,6 @@
 # First stage: Build the Rust components
 FROM rust:buster AS rust-builder
 
-ARG as_version
-
 RUN update-ca-certificates
 
 RUN apt-get update -qqy \
@@ -15,12 +13,9 @@ RUN apt-get update -qqy \
 
 WORKDIR /application-services
 
+# Checkout application services at a locked commit
 RUN git clone https://github.com/mozilla/application-services.git .
-RUN if [[ -z "${as_version}" ]] ; then \
-    echo "No application-services version supplied, using latest commit." ; \
-    else \
-    git fetch ; \
-    fi
+RUN git fetch && git checkout da9301f829de67149d50dce5b709aed42f1bfe1b
 
 RUN git submodule init
 RUN git submodule update --recursive
--- a/experimenter/Dockerfile
+++ b/experimenter/Dockerfile
@@ -80,15 +80,9 @@ FROM system-builder AS rust-builder
 WORKDIR /application-services
 
 # Rust packages
-# This is temporary while megazord packaging is pending.
-ARG as_version
-
+# Checkout application services at a locked commit
 RUN git clone https://github.com/mozilla/application-services.git .
-RUN if [[ -z "${as_version}" ]] ; then \
-    echo "No application-services version supplied, using latest commit." ; \
-    else \
-    git fetch ; \
-    fi
+RUN git fetch && git checkout da9301f829de67149d50dce5b709aed42f1bfe1b
 
 RUN git submodule init
 RUN git submodule update --recursive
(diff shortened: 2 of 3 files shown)


experimenter/Dockerfile
```diff
@@ -80,15 +80,9 @@ FROM system-builder AS rust-builder
...
-ARG as_version
-
+# Checkout application services at a locked commit
 RUN git clone https://github.com/mozilla/application-services.git .
-RUN if [[ -z "${as_version}" ]] ; then \
-    echo "No application-services version supplied, using latest commit." ; \
-    else \
-    git fetch ; \
-    fi
+RUN git checkout da9301f829de67149d50dce5b709aed42f1bfe1b
```
jhugman, line 85: Can we specify this `ref` once, and only once?

cirrus/server/Dockerfile
```diff
@@ -15,12 +13,9 @@ RUN apt-get update -qqy \
...
 WORKDIR /application-services
 
+# Checkout application services at a locked commit
 RUN git clone https://github.com/mozilla/application-services.git .
-RUN if [[ -z "${as_version}" ]] ; then \
-    echo "No application-services version supplied, using latest commit." ; \
-    else \
-    git fetch ; \
-    fi
+RUN git checkout da9301f829de67149d50dce5b709aed42f1bfe1b
```
jeddai, line 18: It may need a fetch before the checkout?
```suggestion
RUN git fetch && git checkout da9301f829de67149d50dce5b709aed42f1bfe1b
```

experimenter/Dockerfile
jaredlockhart: @jhugman Exactly yeah but I'm going to pull it out into a shared base layer as a followup 👍

cirrus/server/Dockerfile
jaredlockhart: No it doesn't, it includes the history as part of the clone 👍

jrbenny35:
Could we use an ```ENV``` variable for the hash so we can just change it via the circleci file or something?

jaredlockhart:
@jrbenny35 It has to be included somewhere in the repo so that it can be locked within a single change

cirrus/server/Dockerfile
jaredlockhart: Oh maybe you're right, why does that work locally and not in CI....

jaredlockhart: Yeah you're right @jeddai but that passed on my local but not in CI, they're on the same Docker version, even after a local prune.  I wonder why that would be.... 🤔

jaredlockhart:
@jrbenny35 Oh sorry I misread your comment, yeah it can definitely be pulled out and then referenced in a single place like the Makefile or Circle config, I'm gonna reorganize all this as a followup.

jeddai (approved):

eliserichards (approved):

PR #9691, opened 2023-11-07
feat(nimbus): Support fetching refs in manifest-tool

brennie:
Because

- we want to keep track of features at different versions of products;
  and
- each version will be at a different ref

This commit

- updates the fetch_fml_app and fetch_legacy_app to support fetching
  from arbitrary refs, not just main/tip


jaredlockhart (approved):
Interesting 👍 I like using more static type annotations here, good to see this pattern grow.  Nice simplification of some of the test cases, and using a dataclass to wrap a Ref.  Looking great excited to see this come along 🎉 🎉 🎉 🎉

PR #9692, opened 2023-11-07
feat(nimbus): List GitHub branches and tags in manifesttool

brennie:
Because

- we need to discover versions based on GitHub branches and tags

This commit

- adds methods to manifesttool's github_api to find all branches and
  tags.

### DIFF
--- a/experimenter/manifesttool/github_api.py
+++ b/experimenter/manifesttool/github_api.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any
+from typing import Any, Generator
 
 import requests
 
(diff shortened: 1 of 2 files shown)


experimenter/manifesttool/github_api.py
```diff
@@ -17,9 +17,33 @@
...
+def paginated_api_request(
+    path: str, per_page: int = 100
+) -> Generator[dict[str, Any], None, None]:
+    """Make several reqeusts to a paginated API resource and yield each page of results."""
+    page = 1
+
+    while True:
+        rsp = api_request(path, params={"page": page, "per_page": per_page})
+
+        if rsp:
```
jaredlockhart, line 41: This is checking for the `{}` empty json response?

jaredlockhart: Is there a way to make the conditions for the `break` case clearer?

jaredlockhart (changes requested):
Neat!

experimenter/manifesttool/github_api.py
brennie: I added a comment.

jaredlockhart (approved):
Cool looks good 👍 🎉

PR #9693, opened 2023-11-07
bug(nimbus): don't use cloud builder in CI

brennie:
Because

- we just disabled HydroBuild in CI (#9680)

This commit

- configures our CI to not use cloud builders.


eliserichards (approved):
🚀

PR #9697, opened 2023-11-08
bug(nimbus): don't use multiplatform builds in CI

brennie:
Because

- the default docker builder does not support multi-platform builds

This commit

- removes multi-platform build support from our Makefile.


jaredlockhart (approved):
Missed one!! Thnx @brennie 🙏

PR #9698, opened 2023-11-08
fixes #9688 feat(nimbus): auto-update application-services

### LINKED ISSUE #9688: Setup Circle job to auto bump Application Services hash
After we have added an explicit reference to an Application Services SHA in our Dockerfiles, let's setup a Circle job that periodically checks whether that SHA has incremented and updates the Dockerfiles accordingly by creating a new PR.  That way we can stay up to date with the latest Application Services but still enforce that no changes will break Experimenter or Cirrus on main.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4004)

brennie:
Because

- we want to automatically update the version of application-services in
  cirrus and experimenter

This commit

- updates the application-services build ID from a CI task

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -520,6 +520,43 @@ jobs:
                 echo "No config changes, skipping"
             fi
 
+  update_application_services:
+    machine:
+      image: ubuntu-2004:2023.10.1
+      docker_layer_caching: true
+    steps:
+      - add_ssh_keys:
+          fingerprints:
+            - "32:8e:72:0b:9a:a1:1c:b8:7e:90:e1:53:a3:73:68:47" # for git pushes from circleci, since relies on ssh
+      - checkout
+      - gh/setup:
+          token: GH_TOKEN # for gh commands from circleci, since relies on user token, since por que no los dos?
+      - run:
+          name: Setup Git
+          command: |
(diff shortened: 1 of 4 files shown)


brennie:
We probably actually don't want to have this in the fetch_external_resources task since breaking A-S changes will break the PRs that update manifests etc. I'll update this to use a different task and separate PRs.

.circleci/config.yml
```diff
@@ -520,6 +520,43 @@ jobs:
...
+              git commit -m "chore(nimbus): Update application-services"
+              if (($((git diff update-application-services origin/update-application-services || git diff HEAD~1) | wc -c) > 0)); then
+                git push origin update-application-services -f
+                gh pr create -t "chore(nimbus): Update application-services" -b "" --base main --head update-application-services --repo mozilla/experimenter || echo "PR already exists, skipping"
+              else
+                echo "Changes already committed, skipping"
+              fi
+            else
+              echo "No config changes, skipping"
+            fi
```
brennie, line 558: @jaredlockhart  Can we refactor this into some sort of named step between the two update jobs? Or should this be made into a shell script and re-used that way?

jaredlockhart:
Let's close this for now until we can revisit how we want to build these in the new year.

PR #9709, opened 2023-11-09
fix #9701 feat(nimbus): Fetch feature id by passing the slug

### LINKED ISSUE #9701: find feature by slug instead of id
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4009)

yashikakhurana:
Because

- We noticed that we are passing feature id to find the feature for the integrations, this approach was working fine as previous apps had `no-feature`  feature's ID which are pretty stable, but for apps like the demo app, the feature ID frequently changes as the feature gets updated

This commit

- Finds the feature id by passing the slug and app name

fixes #9701


experimenter/tests/integration/nimbus/conftest.py
```diff
@@ -26,14 +26,27 @@
...
+    ),
+    BaseExperimentApplications.FOCUS_ANDROID: helpers.get_feature_id_as_string(
+        "no-feature-focus-android", BaseExperimentApplications.FOCUS_ANDROID.value
+    ),
+    BaseExperimentApplications.FOCUS_IOS: helpers.get_feature_id_as_string(
+        "no-feature-focus-ios", BaseExperimentApplications.FOCUS_IOS.value
+    ),
+    BaseExperimentApplications.DEMO_APP: helpers.get_feature_id_as_string(
+        "example-feature", BaseExperimentApplications.DEMO_APP.value
+    ),
```
jaredlockhart, line 46: Yes fantastic.  This way we get both, consistency from using the slugs but mapping them back to IDs so we don't have to change anything else.  Very clever, great job 👍

jaredlockhart (approved):

jrbenny35 (approved):
Nice addtion, great work @yashikakhurana !

PR #9710, opened 2023-11-09
Add support for targeting experiments based on pointer size and RAM

PaulBone:
Because

- We want this targeting support for a PHC guardrail experiment

This commit

- Adds a targeting constant for 64bit Firefox builds running on systems with at least 8GB of RAM.

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -4,6 +4,11 @@ from django.db import models
 
 from experimenter.experiments.constants import Application, NimbusConstants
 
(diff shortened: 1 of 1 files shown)


brennie:
The corresponding changes need to land in desktop first, see
https://phabricator.services.mozilla.com/D193159
https://phabricator.services.mozilla.com/D193160

eliserichards:
@PaulBone @brennie is this still blocked?

eliserichards:
Yes it's still blocked

brennie (approved):

PR #9713, opened 2023-11-09
Allow enrollment when other browser is PDF handler

bytesized:
We would like to be able to enroll browsers in Nimbus experiments based on whether or not the default PDF handler is currently a browser other than the current Firefox installation. This will add that as an Advanced Targeting option.

Because

- We would like to be able to target [this experiment](https://experimenter.services.mozilla.com/nimbus/set-default-pdf-handler-rollout/summary) at users that currently have a browser other than the current Firefox installation as their default PDF handler.

This commit

- Adds that as an Advanced Targeting option

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1717,6 +1717,21 @@ ANDROID_LATER_DAY_USERS_ONLY = NimbusTargetingConfig(
 )
 
 
+DEFAULT_PDF_IS_DIFFERENT_BROWSER = NimbusTargetingConfig(
+    name="Default PDF handler is a different browser",
+    slug="default_pdf_is_different_browser",
+    description=(
+        "Targeting users that have their default PDF handler set to a browser other "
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/targeting/constants.py
```diff
@@ -1717,6 +1717,24 @@ def __post_init__(self):
...
 
+DEFAULT_PDF_IS_DIFFERENT_BROWSER = NimbusTargetingConfig(
+    name="Default PDF handler is a different browser",
+    slug="default_pdf_is_different_browser",
+    description=(
+        "Targeting users that have their default PDF handler set to a browser other "
+        "than the current Firefox installation"
+    ),
+    targeting=(
+        "'pdf' in defaultHandler && !defaultHandler['pdf'] && "
```
brennie, line 1728: `defaultHandler` should be `isDefaultHandler`

brennie (changes requested):

brennie (approved):

PR #9734, opened 2023-11-13
Fixes #9481 feat(nimbus): Add caching to fml loader and remove unused code

### LINKED ISSUE #9481: Add caching
Add caching to the fml error calls. We are creating a `NimbusFmlLoader` and getting clients/inspectors every single time we run the serialization.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3877)

eliserichards:
Because

- We don't want to remake the whole loader, client, and filepaths to the local manifests every time serialization is called

This commit

- Adds caching to
   - the creation of the `NimbusFmlLoader` (we only need a single loader)
   - creating `FmlClient`s (these will always be the same for application/channel -- only need to re-create if they switch channels)
- Create a single loader to be used in the branch validation instead of being created every time the validation for fml is run
- Removes `application_data` and other stale code that is no longer needed for fetching because we have local manifests

Fixes #9481


experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -14,26 +14,18 @@ class NimbusFmlLoader:
...
-        """
-        if Path.exists(file_location):
-            with open(file_location) as application_yaml_file:
-                file = yaml.safe_load(application_yaml_file.read())
-                if application_name in file:
-                    return file[application_name]
-                else:
-                    return None
+    @cache
+    def create_loader(application: str, channel: str):
```
brennie, line 25: Since this is caching, we should probably name it something like `get_loader_for`.

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -14,26 +14,18 @@ class NimbusFmlLoader:
...
-        repos, fml paths within the repos, and release versions.
-        """
-        if Path.exists(file_location):
-            with open(file_location) as application_yaml_file:
-                file = yaml.safe_load(application_yaml_file.read())
-                if application_name in file:
-                    return file[application_name]
-                else:
-                    return None
+    @cache
```
brennie, line 24: This cache will grow forever. We probably want to impose a max size on it using [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache), which is what backs `functools.cache`

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -14,26 +14,18 @@ class NimbusFmlLoader:
...
+    def create_loader(application: str, channel: str):
+        return NimbusFmlLoader(application, channel)
 
+    @cache  # noqa: B019
     def _get_local_file_path(self):
```
brennie, line 29: We should used [`cached_property`](https://docs.python.org/3/library/functools.html#functools.lru_cache) here because there is only ever one value.

If we switch it to a property, it should be named as such (i.e., no `get_`).

```suggestion
    @cached_property
    def file_path(self):
```

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -62,6 +54,7 @@ def _get_fml_clients(self, versions: list[str]) -> list[FmlClient]:
                 )
         return clients
 
+    @cache  # noqa: B019
     def _create_client(self, path: str, channel: str) -> FmlClient:
```
brennie, line 58: `@cached_property` here too

brennie (changes requested):

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
eliserichards: It might not be a single value, because the user might change the `Channel` (since the Audience page comes after the Branches page). Does `cached_property` still make sense if it can change like that?

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
eliserichards: Same as above

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
eliserichards: It's really just initializing a new loader here, so I'd like to keep it as create rather than get

jaredlockhart: Yeah `get` is usually what methods like this are called.  Django calls it `get_or_create` to be explicit.  It's not that big a deal.

experimenter/experimenter/features/tests/test_nimbus_fml_loader.py
```diff
@@ -319,11 +284,53 @@ def test_return_no_errors_for_invalid_application(
...
+        mock_create_client.return_value = None
+        channel = "release"
+        loader = self.create_loader()
+
+        client = loader._create_client(self.TEST_MANIFEST_PATH, channel)
+        mock_create_client.assert_called_once()
+
+        cached_client = loader._create_client(self.TEST_MANIFEST_PATH, channel)
+        mock_create_client.assert_called_once()  # not called a second time
+        self.assertEqual(client, cached_client)
```
jaredlockhart, line 336: @eliserichards and I chatted about this being generally unnecessary since it's really more testing the cache decorator than it is testing our logic, so you don't really need to test this here.

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -14,26 +14,18 @@ class NimbusFmlLoader:
     BASE_PATH = Path(BASE_DIR, "features", "manifests", "apps.yaml")
     MANIFEST_PATH = Path(BASE_DIR, "features", "manifests")
```
jaredlockhart, line 15: These you'll want to move into `settings.py`

eliserichards: `BASE_PATH` is removed here entirely since it's no longer needed now that @brennie has landed the local manifests.

`MANIFEST_PATH` is staying here for now in this PR, and the work to override the settings that we discussed will be in #9768

experimenter/experimenter/features/manifests/nimbus_fml_loader.py
eliserichards: I decided to remove the caching on this altogether, since it really won't change 👍

yashikakhurana (approved):
Perfect, thank you @eliserichards 🎉 🎉

PR #9735, opened 2023-11-13
chore(project): update commit message guide

### LINKED ISSUE #9749: "Fixes" in PR titles not linking or closing issues
Our commit guidelines say to write commit message summaries like "fixes #xyz feat(nimbus): foo bar baz". However, with the introduction of the merge queue, having "fixes" in the title no longer links PRs with issues.

Moving the "Fixes #xyz" text into the description has the desired effect so we should start doing that and update our docs to reflect this process change.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4030)

brennie:
Because:

- the GitHub merge queue broke our workflow of including "fixes" in PR titles; and
- we now have more projects in this repo (Cirrus and schemas)

this commit:

- updates contributing.md to instruct users to put the "fixes" in the description; and
- updates the scope to include Cirrus and schemas.

Fixes #9749

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -67,7 +67,7 @@ Please note we have a [Code of Conduct](https://www.mozilla.org/en-US/about/gove
 
 The subject should follow the this pattern:
 
-`fixes #github_issue_number type(scope): Description`
+`type(scope): Description`
 
 #### Type
 
@@ -90,6 +90,8 @@ One of the following:
 
 - **project**: Anything that affects the entire project
 - **nimbus**: Anything scoped only to Nimbus experiment frontend
+- **cirrus**: Anything scoped to Cirrus.
+- **schemas**: Anything scoped to Nimbus schemas.
 - **legacy**: Anything scoped only to Legacy experiment frontend
 - **visualization**: Anything scoped only to Analysis Visualization
 
@@ -113,6 +115,8 @@ Because
 This commit
 
 * Change
+
+Fixes #github_issue_number
 ```
 
 Just as in the **subject**, use the imperative, present tense: "change" not "changed" nor "changes". Commits are expected to follow this format.


contributing.md
```diff
@@ -67,7 +67,7 @@ Please note we have a [Code of Conduct](https://www.mozilla.org/en-US/about/gove
 
 The subject should follow the this pattern:
 
-`fixes #github_issue_number type(scope): Description`
+`type(scope): Description`
```
mikewilli, line 70: It seems useful to continue putting the issue number in the title so that the list of commits has quick links to the fixed issue. Any particular reason not to continue with this?

yashikakhurana: +1 to Mike's comment

jaredlockhart: Oh funnily enough, there's no issue linked to this pr XD  It should still link to the issue even if it's in the commit body instead of the title, @brennie can you file an issue and then format this pr according to the updated formatting? 🙏  That should show that the pr links to the issue the same way.

brennie: I'm personally not a fan of having fixes in the PR title, just due to the fact that it takes up room and isn't immediately helpful. If I want more information about a commit that has landed, I'm probably going to look at the entire commit message in `git log` or similar. The landed commit also already has the PR number in its summary.

If the above argument doesn't sway folks, I don't feel strongly enough about this issue to debate it further.

jaredlockhart: Oh yeah like, it's not clickable in the actual title, it adds the link elsewhere, and this is clickable in the PR description so you still get the same effect.  I think this is a good change since it accomplishes the effect of linking prs to issues clickably, and also closes the issue which has been broken.

jaredlockhart (approved):
Thanks for figuring this out and updating the docs @brennie !

contributing.md
brennie: @mikewilli are you ok with that?

PR #9744, opened 2023-11-14
feat(nimbus): Resolve branches to versions based on files

### LINKED ISSUE #9736: Parse versions from files in repositories
When we look at a branch for a version, often the branch name will only give us information about the major version. We need to know the specific version so that we can import features to the correct version, instead of always importing over top of the major version.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4023)

brennie:
Because

- versioned branches (e.g., release/v121) do not reveal their full
  version information in the branch name; and
- products store version information in plain text and plist files

this commit

- defines where version information is in each app's repository;
- adds `version.parse_version_file` to parse version from these files;
  and
- adds `version.resolve_ref_versions` to resolve repository refs to
  versions based on these files.

Fixes #9736

### DIFF
--- a/experimenter/manifesttool/github_api.py
+++ b/experimenter/manifesttool/github_api.py
@@ -1,11 +1,14 @@
 import os
-from typing import Any, Generator
+from pathlib import Path
+from typing import Any, Generator, Optional, overload
 
 import requests
 
+from manifesttool import download
 from manifesttool.repository import Ref
 
 GITHUB_API_URL = "https://api.github.com"
+GITHUB_RAW_URL = "https://raw.githubusercontent.com"
 GITHUB_API_HEADERS = {
     "Accept": "application/vnd.github+json",
     "X-GitHub-Api-Version": "2022-11-28",
@@ -81,3 +84,47 @@ def _get_refs(repo: str, kind: str) -> list[Ref]:
         for page in paginated_api_request(f"repos/{repo}/{kind}")
         for ref in page
     ]
+
+
+@overload
+def fetch_file(repo: str, file_path: str, rev: str) -> str:
+    ...  # pragma: no cover
+
+
+@overload
+def fetch_file(repo: str, file_path: str, rev: str, download_path: Path) -> None:
+    ...  # pragma: no cover
+
+
+def fetch_file(
+    repo: str,
+    file_path: str,
+    rev: str,
+    download_path: Optional[Path] = None,
+):
+    """Fetch the file path at the given revision from the repository.
+
+    Args:
+        repo:
+            The name of the repository.
+
+        file_path:
+            The path to the file in the repository.
+
+        rev:
+            The revision at which the file is to be fetched.
+
+        download_path:
+            If provided, the file will be written to disk at this location.
+
+    Returns:
+        If ``download_path`` is ``None``, the file contents are returned as a
+        ``str``. Otherwise, ``None`` is returned.
+    """
+    url = f"{GITHUB_RAW_URL}/{repo}/{rev}/{file_path}"
+
+    if download_path is None:
+        return download.as_text(url)
+
+    download.to_path(url, download_path)
+    return None
--- a/experimenter/manifesttool/tests/test_github_api.py
+++ b/experimenter/manifesttool/tests/test_github_api.py
@@ -1,10 +1,13 @@
-from unittest import TestCase
+from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import Any
+from unittest import TestCase
 
 import responses
 from responses import matchers
 
 from manifesttool import github_api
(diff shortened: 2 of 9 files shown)


experimenter/manifesttool/github_api.py
```diff
@@ -81,3 +84,47 @@ def _get_refs(repo: str, kind: str) -> list[Ref]:
...
+
+
+@overload
+def fetch_file(repo: str, file_path: str, rev: str) -> str:
+    ...  # pragma: no cover
+
+
+@overload
+def fetch_file(repo: str, file_path: str, rev: str, download_path: Path) -> None:
+    ...  # pragma: no cover
```
jaredlockhart, line 96: What's this for?

experimenter/manifesttool/tests/test_github_api.py
```diff
@@ -158,3 +161,23 @@ def test_get_tags(self):
...
+
+            self.assertTrue(tmp_filename.exists())
+
+            with tmp_filename.open("rb") as f:
+                contents = f.read()
+
+            self.assertEqual(contents, b"hello, world\n")
+
+        contents = github_api.fetch_file("repo", "file/path.txt", "ref")
+        self.assertEqual(contents, "hello, world\n")
```
jaredlockhart, line 183: Why do we do these calls/assertions again outside of the `with` block?

jaredlockhart (commented):
Yeah I see why we have to do this, how scalable do you think this will be as we add more applications?

experimenter/manifesttool/github_api.py
brennie: Its just type hints because fetch_file has two modes: download to a path or return a string.

experimenter/manifesttool/tests/test_github_api.py
brennie: The calls are different. In the `with` block we are downloding a file, reading it, and comparing it

outside the with block, we are calling it without a download path and it returns the contents as as string.

brennie:
> Yeah I see why we have to do this, how scalable do you think this will be as we add more applications?

We have a GH api key in CI, so it should be fine?

For each github hosted application we need to do the following:

1. List all branches. Right now no app has > 100 branches so its 1 API call per.
2. Filter branches down to the most recent 5 major revisions. This cuts out the majority of the branches down to ~10.
3. Resolve all branches to their proper versions, 1 request per.
4. List all tags. Fenix has 340 right now which is 4 requests.
5. We also limit the tags to the most recent 5 major revisions, which removes the majority.
6. for FML apps, download-single-file makes non API requests to githubuserdata.com. It depends on how each app is structured because of FML inclusion

That puts us in the ballpark of ~20 API requests (rounding up) per GitHub-hosted app.
According to the [docs](https://docs.github.com/en/rest/overview/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github-app-installations) we should have limit of 5,000 requests/hour for our token, so theoretically we could support ~250 apps?

If we grow that large, we probably want to do something else.

experimenter/manifesttool/github_api.py
jaredlockhart: But it can't infer that from the optional arg in the actual function implementation?  Like will type checking fail if you pull these out?

jaredlockhart (approved):
Annoying we have to deal with these nuances.  What's your sense of how this will scale as we add more applications?  Will we have to keep adding version fetching/parsing special cases for different apps that use different tags?

Also yep this looks good, thnx @brennie

brennie:
> Annoying we have to deal with these nuances. What's your sense of how this will scale as we add more applications? Will we have to keep adding version fetching/parsing special cases for different apps that use different tags?
> 
> Also yep this looks good, thnx @brennie

I suggested earlier that we should keep an index of each repository's branches/tags and how they map to SHAs and versions. That way, we can skip a lot of steps if every time we fetch branches and tags we filter out anything that hasn't changed. That will remove the majority of the API calls and fetches (and will speed up the application a lot).

Version parsing from branches is based on a regexes and has to be added in a per repository manner. Parsing from version files is kind of a hack, but we could potentially require something in each app, perhaps in the FML manifest? that way it would be in a known location for every app.

The idea of a per-app "experimenter.manifest" like we discuessed earleir that defines a lot of these things would be really nice.

experimenter/manifesttool/github_api.py
brennie: The inferred type signature would otherwise be:

```
def fetch_file(
    repo: str,
    file_path: str,
    rev: str,
    download_path: Optional[Path] = None,
) -> Optional[str]
```

which does not capture the link between download_path and the return type.

PR #9747, opened 2023-11-15
fix #9660 feat(nimbus): Update end enrollment and end experiment for web applications

### LINKED ISSUE #9660: Update End Enrollment behavior for Cirrus
Experimenter's End Enrollment flow needs to change for Cirrus experiments -- `isEnrollmentPaused` should remain false (otherwise the experiment will effectively end for Cirrus), but the enrollment end date needs to be set on the experiment (so that Jetstream knows to start analysis).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3979)

yashikakhurana:
Because

- For the web applications that are using `Cirrus`, end enrollment is a little bit different as enrollment is continuous and it only ends once we end the experiment

This commit

- Changes end experiment to send explicitly `true` value
- Changes end enrollment to only send `true` value if it is not a web experiment, otherwise it will remain false

fixes #9660


jaredlockhart (approved):
Very nice succinct change @yashikakhurana thank you!

PR #9756, opened 2023-11-15
fix #9755 feat(cirrus): Make record metrics function async

### LINKED ISSUE #9755: Make record metric method async
We want to continue serving requests to the users when we are recording metrics and flushing the events, there is no way to manually flush the events, if filling it triggers the buffer full the it submits the pings. So to make it async we can make `record_metric` method async.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3908)

yashikakhurana:
Because

- We want to continue serving requests to the users when we are recording metrics and flushing the events, there is no way to manually flush the events, if filling it triggers the buffer full then it submits the pings. 

This commit

- Make the record metrics function async so that it can do tasks in the background as a nonblocking task

### DIFF
--- a/cirrus/server/cirrus/main.py
+++ b/cirrus/server/cirrus/main.py
@@ -152,7 +152,7 @@ def collate_enrollment_metric_data(
     return data
 
 
-def record_metrics(enrolled_partial_configuration: dict[str, Any], client_id: str):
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Neat!!

PR #9761, opened 2023-11-16
Fix #9757 fix(nimbus): Remove prefetched changelogs for csv reports

### LINKED ISSUE #9757: Can't pull reports
Shell is trying to get Sept and Oct numbers - but it keeps failing to run report.  eventually give me a 503

!image-20231108-190926.png|width=775,height=234!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4012)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/52765">image-20231108-190926.png</a>

eliserichards:
Because

- Our reports are taking too long to generate, and nginx is timing them out
- We're seeing that we're prefetching changelogs, but we only have to use the changelogs to compute the start date (find the changelog where the experiment goes from DRAFT->LIVE)

This commit

- Removes the prefetching of the changelogs, which should cut down the time needed to generate the reports

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/views.py
+++ b/experimenter/experimenter/experiments/api/v5/views.py
@@ -18,7 +18,7 @@ class NimbusExperimentCsvListView(ListAPIView):
 
     queryset = (
         NimbusExperiment.objects.select_related("owner")
-        .prefetch_related("feature_configs", "changes")
+        .prefetch_related("feature_configs")
         .filter(is_archived=False)
     )
 


jaredlockhart (approved):
🙏 🎉

yashikakhurana (approved):

yashikakhurana:
good catch @eliserichards 🎉 , Also its worth looking into N+1 Queries that may be happening

PR #9762, opened 2023-11-16
chore(nimbus): only run integration tests for a single app

### LINKED ISSUE #9718: Run iOS integration tests against only iOS
Here we define which applications the ios/remote settings integration tests should run against:

https://github.com/mozilla/experimenter/blob/main/.circleci/config.yml#L238

but now that we have new applications, this is catching more than just ios it's also running against monitor so it's taking twice as long.  Let's explicitly filter that to ios only.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4018)

jaredlockhart:
Because

* We run many integration tests against many applications for coverage
* As we add more applications, we don't want to disrupt the existing tests
* Some integration tests were specifying applications by excluding other applications becuase IOS and FOCUS_IOS would both match the same string IOS
* This was resulting in some test runs running against multiple applications which is slowing down CI and PRs

This commit

* Renames applications to have unique names
* Updates all the config references to use those names

Fixes #9718

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -214,7 +214,7 @@ jobs:
     working_directory: ~/experimenter
     environment:
       FIREFOX_VERSION: nimbus-firefox-release
-      PYTEST_ARGS: -k FENIX -m remote_settings --reruns 1
+      PYTEST_ARGS: -k FIREFOX_FENIX -m remote_settings --reruns 1
     steps:
       - checkout
       - check_file_paths:
@@ -235,7 +235,7 @@ jobs:
     working_directory: ~/experimenter
     environment:
       FIREFOX_VERSION: nimbus-firefox-release
-      PYTEST_ARGS: -k 'not (FOCUS_IOS or DESKTOP or FENIX or FOCUS_ANDROID)' -m remote_settings --reruns 1
+      PYTEST_ARGS: -k FIREFOX_IOS -m remote_settings --reruns 1
     steps:
       - checkout
       - check_file_paths:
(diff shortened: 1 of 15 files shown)


brennie (approved):
Let's gooooo 🚀

.circleci/config.yml
```diff
@@ -361,7 +361,7 @@ jobs:
     working_directory: ~/experimenter
     environment:
       FIREFOX_VERSION: nimbus-firefox-release
-      PYTEST_ARGS: -k 'not (FOCUS_IOS or DESKTOP or FENIX or FOCUS_ANDROID or IOS)'  -m demo_app --reruns 1
+      PYTEST_ARGS: -k DEMO_APP  -m demo_app --reruns 1
     steps:
```
yashikakhurana, line 365: I tried this, but for some reason, it was running for all, that's why I explicitly said not these apps

jaredlockhart: Yeah I'm updating the names so that we should be able to just specify the one app.

experimenter/tests/integration/nimbus/test_remote_settings.py
```diff
@@ -257,30 +256,32 @@ def test_rollout_live_update_reject_on_experimenter(
...
+#     summary.launch_and_approve()
+
+#     kinto_client.approve()
+
+#     summary = SummaryPage(selenium, experiment_url).open()
+#     summary.wait_for_live_status()
+
+#     summary.wait_for_timeline_visible()
+#     summary.wait_until_timeline_release_date()
+#     summary.wait_until_audience_section_release_date()
```
jrbenny35, line 287: Something wrong with this test?

jrbenny35 (changes requested):
Looks good, thanks for this. Just had 1 question.

experimenter/tests/integration/nimbus/test_remote_settings.py
jaredlockhart: Oh yeah because of the way it's reading from os.environ I have to change that part, it'll be fixed before I move this out of draft 👍

yashikakhurana (approved):
perfect and good catch 👏

PR #9776, opened 2023-11-17
feat(nimbus): get analysis bases from weekly results as fallback

### LINKED ISSUE #9775: Show analysis basis radio buttons based on overall or weekly results
Currently we only get the list of analysis bases from overall, we should fall back to weekly so that exposures are available for preliminary results as well.

cc @danielkberry

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4042)

mikewilli:
Because

- weekly results are available before overall
- experiments with multiple analysis bases may want to seem them in preliminary weekly results

This commit

- gets the analysis basis list from weekly as a fallback if overall results are not available

fixes #9775


jaredlockhart (approved):
Neat! 🎉

Issue #9781, opened 2023-11-18
Dont fetch the world every time


data-sync-user:
Right now, `manifest-tool fetch`will always attempt to fetch everything, even things that have not been updated since its last fetch, because it doesnt know what has or has not been fetched. We should keep an index mapping ref names to refs and versions so we can compare new fetches against that cache and prevent extra fetches.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4041)

brennie:
Fixed in #9785

PR #9785, opened 2023-11-20
feat(nimbus): Cache refs from fetched releases in manifest-tool

brennie:
Because

- we currently fetch every matching release for every app;
- that many fetches requires a lot of GitHub API calls;
- all these fetches take time to effectively do nothing; and
- we expect most releases (e.g., tags) not to change between fetches

This commit

- adds a new per-app cache mapping ref names to SHAs;
- compares the results of release discovery to this cache; and
- excludes un-updated releases from the fetch pipeline.

Fxies #9781


jaredlockhart (approved):
Yeah that makes sense, but on the flip side... the time spent in the circle jobs doesn't really matter.  The rate limiting matters if it means the update tasks become fairly intermittent.  That might become more of an issue as we scale up the number of apps, so dealing with this now might prevent errors like that in the future, so that's probably worthwhile.

Altogether there's a lot of machinery in the update configs task now so we should probably make sure to wrap it all in Sentry so if update tasks fail we get notified.

A nice side benefit of this is it lets us clearly track exactly which ref we're up to date with which might be handy in the future if people are trying to reason about which changes are where and whether they're reflected in Experimenter or not.

Anyways this is really neat, very simple and straight forward, nice and readable, storing them as YAMLs is a nice idea, very clear to grok.

Thanks @brennie 🙏 👍

Issue #9025, opened 2023-06-23
Migrate Experiment Schema to New Schemas Package


jaredlockhart:
Now that we have a new schemas package 🎉 we can migrate the experiment schema out of the shared repo and into the schemas package.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3633)

PR #9787, opened 2023-11-20
fix #9786 feat(nimbus): add published_date field to schema

### LINKED ISSUE #9786: Add published_date field to experiment/rollout remote settings objects
https://mozilla-hub.atlassian.net/browse/EXP-4002

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4058)

jeddai:
Because

- a `published_date` is needed for holdback support

This commit

- adds the `published_date` field to the experiment schema


schemas/mozilla_nimbus_schemas/experiments/experiments.py
```diff
@@ -72,3 +72,4 @@ class NimbusExperiment(BaseModel):
     featureValidationOptOut: Optional[bool]
     localizations: Optional[dict]
     locales: Optional[List[str]]
+    publishedDate: Optional[str]
```
mikewilli, line 75: Should this be a datetime object instead of string? Pydantic will accept and parse it for you if it's a string in ISO format. https://docs.pydantic.dev/1.10/usage/types/#datetime-types

jeddai: The other dates in here use string as well, I was mostly going off of that

mikewilli: Works for me! Not worth changing them all for this PR so we can address it separately if we want.

jaredlockhart (changes requested):
We have not yet [migrated all consumers](https://github.com/mozilla/experimenter/issues/9025) over to using the new schemas package, so we may have to also add the new field in nimbus-shared as well.  We'll discuss with everybody and figure out which parts of the system are consuming which schemas and then figure out where we need to make the change.

jaredlockhart:
And another sidebar, even in a world where we were solely consuming the new schemas package, we'd want to land the new field there alone, publish it, and then consume it in experimenter by bumping the schemas package version and then proceeding with the other changes included here.

Also we shouldn't need to include this field in V5 at all since it'll be captured on the model at publish time and included in V6, I don't think it's needed in the UI anywhere.

eliserichards (approved):
Yay unblocking this!

jaredlockhart (changes requested):

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -1544,6 +1544,29 @@ def test_can_set_excluded_required_experiments(self):
...
+            experiment,
+            {
+                "status": NimbusExperiment.Status.DRAFT,
+                "publish_status": NimbusExperiment.PublishStatus.APPROVED,
+                "changelog_message": "Test changelog",
+            },
+            context={"user": self.user},
+        )
+
+        print(serializer.instance)
```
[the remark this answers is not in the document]
jeddai: 😬

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1175,6 +1177,9 @@ def save(self):
                 if self.is_preview_to_draft:
                     self.instance.published_dto = None
 
+                if self.should_call_push_task and self.instance.published_date is None:
+                    self.instance.published_date = timezone.now()
```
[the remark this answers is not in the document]
jeddai: Ah cool! Got that updated!

experimenter/experimenter/kinto/tests/test_tasks.py
```diff
@@ -768,7 +768,7 @@ def test_updating_experiment_with_published_dto_none_is_skipped(self):
 
 
 class TestNimbusPushExperimentToKintoTask(MockKintoClientMixin, TestCase):
-    def test_push_experiment_to_kinto_sends_desktop_experiment_data_and_sets_accepted(
+    def test_push_experiment_to_kinto_sends_desktop_experiment_data_and_sets_accepted_and_published_date(  # noqa: E501
```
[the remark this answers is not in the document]
jeddai: Fixed!

experimenter/experimenter/kinto/tasks.py
```diff
@@ -264,6 +265,9 @@ def nimbus_push_experiment_to_kinto(collection, experiment_id):
 
         kinto_client = KintoClient(collection)
 
+        if experiment.published_date is None:
+            experiment.published_date = timezone.now()
```
jaredlockhart, line 269: Actually we could always force overwrite this, if for whatever reason it's still set when we get here and we're publishing to kinto, then the most recent publish timestamp would be the canonical one wouldn't it?

jaredlockhart (changes requested):

experimenter/experimenter/kinto/tasks.py
jeddai: I don't necessarily know that we'd want that? With rollouts that increment their pop size they would be moving their published_date pretty frequently and I imagine we'd want that to stay static, wouldn't we?

jaredlockhart: Oh yes very good pointtttttt......... Then it's more about state ie if it's not live we'd always want to overwrite?

jeddai: Perhaps? My understanding was it wasn't meant to roll but if we want it to roll then yes what you've described makes sense

jaredlockhart: Yeah like I'd imagine that date is the most recent date that a launching recipe is pushed to kinto.  We can unset it on timeout/rejection but those tasks aren't gauranteed to run/complete so if we got here, we're not live, and for whatever reason publish date is set, wouldn't we always want the most recent date?

jeddai: Fair enough — okay so switching it to
- set if not set
- set if currently live

jaredlockhart: * if currently *not* live, right?

jeddai: Ah yes sorry

jaredlockhart:
Closing and moving over to https://github.com/mozilla/experimenter/pull/9787

PR #9794, opened 2023-11-21
feat(nimbus): Land new advanced targeting with cor…

### LINKED ISSUE #9783: Land new advanced targeting with correct Early Day User definition
[Track this issue on Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1857305)

We have a few "Early Day User ..." definitions targeting users with profiles <= 28 days. This is not consistent with DS definitions and overlaps with "Existing User" targeting (profiles >= 28 days). Since we don't want to modify existing targeting(to avoid affecting current/past experiments) This task is to add new targeting that corrects all targeting that references "Early user" to only target profile age < 28 days (NON INCLUSIVE) so that it's mutually exclusive with Existing user definitions. 

Current Early day user definitions with <=28 days:-
* Early day user
* Early day user has default needs pin
* Early day user need default
* Early day user need pin

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4055)

halemu:
We have a few "Early Day User ..." definitions targeting users with profiles <= 28 days. This is not consistent with DS definitions and overlaps with "Existing User" targeting (profiles >= 28 days). Since we don't want to modify existing targeting(to avoid affecting current/past experiments) This pull request is to add new targeting that corrects all targeting that references "Early user" to only target profile age < 28 days (NON INCLUSIVE) so that it's mutually exclusive with Existing user definitions.

Current Early day user definitions with <=28 days:-

    Early day user
    Early day user has default needs pin
    Early day user need default
    Early day user need pin

Fixes #9783

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1526,11 +1526,33 @@ EARLY_DAY_USER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+EARLY_DAY_USER_V2 = NimbusTargetingConfig(
+    name="Early day user (less than 28 days)",
+    slug="early_day_user_v2",
+    description="Users with profiles that are less than 28 days old",
+    targeting=f"{PROFILELESSTHAN28DAYS}",
+    desktop_telemetry="",
+    sticky_required=True,
+    is_first_run_required=False,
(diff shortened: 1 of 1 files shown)


brennie:
As a heads up, we updated our commit format to have the fixes #xyz in the commit body, so that PRs correctly link to issues when merged via the merge queue. I went ahead and updated your title and description for you.

PR #9807, opened 2023-11-21
feat(nimbus): don't open PRs for just ref-cache changes

### LINKED ISSUE #9806: dont open PRs of just ref-cache changes
Now that we are caching refs in manifest-tool, we are seeing PRs opened to update feauture configs that are just bumping SHAs of main, etc. without any associated feature config changes.

We should check if the PR we’re about to push is just .ref-cache changes and, if so, not open a pr / push changes to the PR branch.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4064)

brennie:
Because

- we only want to merge PRs that have feature configuration changes

this commit

- adds a script to check if there are only `.ref-cache` changes;
- updates the `update_external_configs` CI job to call that script and;
- only push to the `external-config` branch and open PRs if that script reports changes.

Fixes #9806

### DIFF
--- a/experimenter/bin/should-pr.py
+++ b/experimenter/bin/should-pr.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+
+import subprocess
+import sys
+from pathlib import Path
+
+MANIFEST_DIR = (
+    Path(__file__)
+    .parent.parent.joinpath("experimenter", "features", "manifests")
+    .resolve()
+)
+
+
+def changed_files() -> list[Path]:
+    output = subprocess.check_output(["git", "status", "--porcelain"], encoding="utf-8")
+    files = [Path(line[3:]) for line in output.splitlines()]
+    return files
+
+
+def is_ref_cache(path: Path):
+    return path.name == ".ref-cache.yaml" and path.resolve().is_relative_to(MANIFEST_DIR)
+
+
+def main() -> int:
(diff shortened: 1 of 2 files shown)


experimenter/bin/should-pr.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def changed_files() -> list[Path]:
+    output = subprocess.check_output(["git", "status", "--porcelain"], encoding="utf-8")
+    files = [Path(line[3:]) for line in output.splitlines()]
+    return files
+
+
+def is_ref_cache(path: Path):
+    return path.name == ".ref-cache.yaml" and path.parent.parent.resolve() == MANIFEST_DIR
```
jaredlockhart, line 21: If we end up moving paths around that `.parent.parent` might break, would it be better just do do like `MANIFEST_DIR in path` or something?

experimenter/bin/should-pr.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def main() -> int:
+    files = [f for f in changed_files() if not is_ref_cache(f)]
+
+    if len(files):
+        return 0
+    else:
+        return 1
```
jaredlockhart, line 30: `return not len(files)`?

jaredlockhart (approved):
Nnnnnnnnneat! Thnx @brennie 🙏

PR #9814, opened 2023-11-23
feat(nimbus): Load versioned manifests in load_feature_configs

### LINKED ISSUE #9809: Import versioned features from vendored manifests
Now that we’re vendoring manifests per app version, we need to import these into Experimenter so we can start using them.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4066)

brennie:
Because

- we have versioned feature manifests; and
- we want to use these in Experimenter

This commit

- updates load_feature_configs to import versioned manifests.

Fixes #9809

### DIFF
--- a/experimenter/experimenter/features/__init__.py
+++ b/experimenter/experimenter/features/__init__.py
@@ -87,12 +87,10 @@ class Features:
         features = []
 
         for application in NimbusConstants.APPLICATION_CONFIGS.values():
-            application_yaml_path = os.path.join(
-                settings.FEATURE_MANIFESTS_PATH, application.slug, "experimenter.yaml"
-            )
+            application_yaml_path = settings.FEATURE_MANIFESTS_PATH / application.slug / "experimenter.yaml"
 
-            if os.path.exists(application_yaml_path):
-                with open(application_yaml_path) as application_yaml_file:
+            if application_yaml_path.exists():
+                with application_yaml_path.open() as application_yaml_file:
                     application_data = yaml.safe_load(application_yaml_file)
                     manifest = FeatureManifest.parse_obj(application_data)
 
--- a/experimenter/experimenter/features/__init__.py
+++ b/experimenter/experimenter/features/__init__.py
@@ -1,8 +1,8 @@
 import json
-import os
+import re
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Union
+from typing import Iterable, Optional, Union
 
 import yaml
 from django.conf import settings
@@ -14,7 +14,8 @@ from mozilla_nimbus_schemas.experiments.feature_manifests import (
     FeatureWithoutExposure,
 )
 
-from experimenter.experiments.constants import NimbusConstants
+from experimenter.experiments.constants import ApplicationConfig, NimbusConstants
+from manifesttool.version import Version
 
 FEATURE_SCHEMA_TYPES = {
     FeatureVariableType.INT: "integer",
@@ -34,6 +35,7 @@ class Feature:
     slug: str
     application_slug: str
     model: Union[FeatureWithExposure, FeatureWithoutExposure]
+    version: Optional[Version] = None
 
     def load_remote_jsonschema(self):
         if self.model.json_schema is not None:
@@ -80,28 +82,53 @@ class Feature:
 
 
 class Features:
-    _features = None
+    _features: Optional[list[Feature]] = None
+
+    @classmethod
+    def _read_manifest(
+        cls,
+        application: ApplicationConfig,
+        manifest_path: Path,
+        version: Version = None,
+    ):
+        with manifest_path.open() as manifest_file:
+            application_data = yaml.safe_load(manifest_file)
+            manifest = FeatureManifest.parse_obj(application_data)
+
+            for feature_slug, feature_model in manifest.__root__.items():
+                yield Feature(
+                    slug=feature_slug,
+                    application_slug=application.slug,
+                    model=feature_model.__root__,
+                    version=version,
+                )
 
     @classmethod
     def _load_features(cls):
         features = []
+        version_re = re.compile(r"^v(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)")
 
         for application in NimbusConstants.APPLICATION_CONFIGS.values():
-            application_yaml_path = settings.FEATURE_MANIFESTS_PATH / application.slug / "experimenter.yaml"
+            application_dir: Path = settings.FEATURE_MANIFESTS_PATH / application.slug
+            application_yaml_path = application_dir / "experimenter.yaml"
 
             if application_yaml_path.exists():
-                with application_yaml_path.open() as application_yaml_file:
-                    application_data = yaml.safe_load(application_yaml_file)
-                    manifest = FeatureManifest.parse_obj(application_data)
+                features.extend(cls._read_manifest(application, application_yaml_path))
 
-                    for feature_slug, feature_model in manifest.__root__.items():
-                        features.append(
-                            Feature(
-                                slug=feature_slug,
-                                application_slug=application.slug,
-                                model=feature_model.__root__,
+                for child in application_dir.iterdir():
+                    if not child.is_dir():
+                        continue
+
+                    if m := version_re.match(child.name):
+                        version = Version.from_match(m.groupdict())
+
+                        application_yaml_path = child / "experimenter.yaml"
+                        if application_yaml_path.exists():
+                            features.extend(
+                                cls._read_manifest(
+                                    application, application_yaml_path, version
+                                )
                             )
-                        )
 
         return features
 
@@ -110,15 +137,23 @@ class Features:
         cls._features = None
 
     @classmethod
-    def all(cls):
+    def all(cls) -> list[Feature]:
         if cls._features is None:
             cls._features = cls._load_features()
 
         return cls._features
 
     @classmethod
-    def by_application(cls, application):
+    def by_application(cls, application) -> list[Feature]:
         return [f for f in cls.all() if f.application_slug == application]
+
+    @classmethod
+    def unversioned(cls) -> Iterable[Feature]:
+        return (f for f in cls.all() if f.version is None)
+
+    @classmethod
+    def versioned(cls) -> Iterable[Feature]:
+        return (f for f in cls.all() if f.version is not None)
 
 
 @register()
--- a/experimenter/experimenter/features/management/commands/load_feature_configs.py
+++ b/experimenter/experimenter/features/management/commands/load_feature_configs.py
@@ -7,6 +7,7 @@ from django.core.management.base import BaseCommand
 from experimenter.experiments.constants import NO_FEATURE_SLUG
 from experimenter.experiments.models import (
     NimbusFeatureConfig,
+    NimbusFeatureVersion,
     NimbusVersionedSchema,
 )
 from experimenter.features import Features
--- a/experimenter/experimenter/features/management/commands/load_feature_configs.py
+++ b/experimenter/experimenter/features/management/commands/load_feature_configs.py
@@ -1,4 +1,6 @@
+import itertools
 import logging
+from typing import Optional
 
 from django.core.management.base import BaseCommand
 
@@ -8,6 +10,7 @@ from experimenter.experiments.models import (
     NimbusVersionedSchema,
 )
 from experimenter.features import Features
+from manifesttool.version import Version
 
 logger = logging.getLogger()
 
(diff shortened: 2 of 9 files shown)


experimenter/experimenter/features/__init__.py
```diff
@@ -80,30 +82,53 @@ def get_jsonschema(self):
 
 
 class Features:
-    _features = None
+    _features: Optional[list[Feature]] = None
```
eliserichards, line 85: Do we want this to be `list[Feature]` instead of Optional?

experimenter/experimenter/features/__init__.py
```diff
@@ -80,30 +82,53 @@ def get_jsonschema(self):
...
+                    slug=feature_slug,
+                    application_slug=application.slug,
+                    model=feature_model.__root__,
+                    version=version,
+                )
 
     @classmethod
     def _load_features(cls):
         features = []
+        version_re = re.compile(r"^v(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)")
```
eliserichards, line 109: 😰 👏

experimenter/experimenter/features/__init__.py
```diff
@@ -80,30 +82,53 @@ def get_jsonschema(self):
...
 
         for application in NimbusConstants.APPLICATION_CONFIGS.values():
-            application_yaml_path = os.path.join(
-                settings.FEATURE_MANIFESTS_PATH, application.slug, "experimenter.yaml"
-            )
+            application_dir: Path = settings.FEATURE_MANIFESTS_PATH / application.slug
+            application_yaml_path = application_dir / "experimenter.yaml"
```
eliserichards, line 113:
```suggestion
            application_yaml_path = Path(application_dir, "experimenter.yaml")
```

experimenter/experimenter/features/__init__.py
```diff
@@ -80,30 +82,53 @@ def get_jsonschema(self):
...
-                    for feature_slug, feature_model in manifest.__root__.items():
-                        features.append(
-                            Feature(
-                                slug=feature_slug,
-                                application_slug=application.slug,
-                                model=feature_model.__root__,
+                    if m := version_re.match(child.name):
+                        version = Version.from_match(m.groupdict())
+
+                        application_yaml_path = child / "experimenter.yaml"
```
eliserichards, line 125:
```suggestion
                        application_yaml_path = Path(child, "experimenter.yaml")
```

experimenter/experimenter/features/__init__.py
```diff
@@ -112,16 +137,24 @@ def clear_cache(cls):
...
 
         return cls._features
 
     @classmethod
-    def by_application(cls, application):
+    def by_application(cls, application) -> list[Feature]:
         return [f for f in cls.all() if f.application_slug == application]
 
+    @classmethod
+    def unversioned(cls) -> Iterable[Feature]:
```
eliserichards, line 151: ohhh this is great!

experimenter/experimenter/features/tests/__init__.py
```diff
@@ -1,37 +1,25 @@
...
 
 from django.test import override_settings
 
+FIXTURE_DIR = Path(__file__).parent.absolute() / "fixtures"
+
 mock_valid_features = override_settings(
-    FEATURE_MANIFESTS_PATH=os.path.join(
-        pathlib.Path(__file__).parent.absolute(), "fixtures", "valid_features"
-    )
+    FEATURE_MANIFESTS_PATH=FIXTURE_DIR / "valid_features"
```
eliserichards, line 8: Can we update these to be using the `Path(x, x, x, ...)` syntax?

eliserichards (approved):
🥇 with a few nits, but feel free to ignore

experimenter/experimenter/features/tests/__init__.py
jaredlockhart: I think if we wanted to set some consistency around that we'd want a lint rule, otherwise there's no good way to enforce it.  Maybe ruff will add rules around it, but barring that we're already using this syntax in other places so it's probably not worth fretting about it here.

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -22,7 +26,31 @@ def handle(self, *args, **options):
             NimbusFeatureConfig.objects.values_list("application", "slug")
         )
 
-        for feature in Features.all():
+        # A mapping of (application, slug) to feature configs.
```
jaredlockhart, line 29: It's application and feature slug?

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -22,7 +26,31 @@ def handle(self, *args, **options):
...
+        # Features will include multiples of the same feature, each with a
+        # different version. We only need to create a single NimbusFeatureConfig
+        # for each of those, so if we keep track of them we can skip a bunch of
+        # update_or_create() calls.
+        feature_configs: dict[tuple[str, str], NimbusFeatureConfig] = {}
+
+        # Iterate through the unversioned features first for the initial
+        # update_or_create() calls. The unversioned features are the most
+        # up-to-date (since they come from the main branch (or equivalent)), so
+        # lets use the descriptions from there.
```
jaredlockhart, line 40: Mmmm interesting.  I wonder if we want to make some kind of wrapper around a feature that lets us manage which fields come from where, ie descriptions from latest main, a main schema, or versioned schemas, etc?  So we can just do that kind of lookup in one place and then everywhere we consume them we don't have to reason about looking through unversioned/versioned to find some particular field.  Really depends on where/how we consume those fields.  Just musing.

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -22,7 +26,31 @@ def handle(self, *args, **options):
...
+        # aren't in the unversioned manifests.
+        #
+        # When we are ingesting versioned Features, we want to update the
+        # NimbusFeatureConfig objects with the most up-to-date description.
+        for feature in itertools.chain(
+            Features.unversioned(),
+            sorted(Features.versioned(), key=lambda f: f.version, reverse=True),
+        ):
+            if (feature.application_slug, feature.slug) in feature_configs:
+                continue
```
jaredlockhart, line 52: So is the idea here that we go over the unversioned ones first, and if we hit every feature then that's sufficient, but there may be some feature for which there is no unversioned schema but there may be a versioned schema, and so in that case we fall back to that?

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -33,9 +61,38 @@ def handle(self, *args, **options):
...
+        versions: dict[Version, NimbusFeatureVersion] = {}
+        for feature in Features.versioned():
+            if feature.version in versions:
+                continue
+
+            versions[feature.version] = NimbusFeatureVersion.objects.get_or_create(
+                major=feature.version.major,
+                minor=feature.version.minor,
+                patch=feature.version.patch,
+            )[0]
```
jaredlockhart, line 88: Oh becuase it returns the tuple of `(instance, created_bool)`?  I always hate that XD  Another way to write this is

```py
versions[feature.version], _ = NimbusFeatureVersion.objects.get_or_create(.....)
```

instead of tacking `[0]` at the end.  Not really much different but makes it clear it's a 2tuple that we only care about the first field, whereas `[0]` could be read as 'there's some arbitrary list here and we're just pulling the first instance out'.  But doesn't really matter either way.

experimenter/experimenter/features/__init__.py
jaredlockhart: We'll have to be careful about future apps that don't conform to the full semver, this is another case where we'd wanna make sure if we really fail to match we throw to Sentry, I imagine.

jaredlockhart (approved):
Wow, wow wow wow.  This is some goody goodness right here:

```py
In [14]: feature.schemas.filter(version__major=115)
Out[14]: <QuerySet [<NimbusVersionedSchema(feature_config_id=53, version=115.6.0)>, <NimbusVersionedSchema(feature_config_id=53, version=115.5.0)>]>

In [15]: feature.schemas.filter(version__major__gte=115)
Out[15]: <QuerySet [<NimbusVersionedSchema(feature_config_id=53, version=115.6.0)>, <NimbusVersionedSchema(feature_config_id=53, version=122.0.0)>, <NimbusVersionedSchema(feature_config_id=53, version=115.5.0)>, <NimbusVersionedSchema(feature_config_id=53, version=120.0.1)>, <NimbusVersionedSchema(feature_config_id=53, version=121.0.0)>, <NimbusVersionedSchema(feature_config_id=53, version=120.0.0)>]>

In [16]: feature.schemas.filter(version__major__gte=115, version__major__lte=120)
Out[16]: <QuerySet [<NimbusVersionedSchema(feature_config_id=53, version=115.6.0)>, <NimbusVersionedSchema(feature_config_id=53, version=115.5.0)>, <NimbusVersionedSchema(feature_config_id=53, version=120.0.1)>, <NimbusVersionedSchema(feature_config_id=53, version=120.0.0)>]>
```

That is so clean and succinct and easy.  Hats off to you, this is great 🎉 🎉 🎉 🎉

experimenter/experimenter/features/__init__.py
brennie: manifest-tool explicitly generates major.minor.patch versions, ignoring all the other gunk (like 0.0.0a0, 0.0.0rc1, etc.)

experimenter/experimenter/features/__init__.py
brennie: using / is idiomatic when joining `Path`s.

experimenter/experimenter/features/management/commands/load_feature_configs.py
brennie: I'm personally not a fan of destructing with `_` in Python because its an actual variable and not some placeholder. Plus there is a common pattern in Django of:

```
from django.utils.translation import gettext as _
```

which conflicts with using _. I know we don't support i10n, but maybe one day we do and this would be a footgun.

experimenter/experimenter/features/__init__.py
jaredlockhart: Oohhhh neat 👍

experimenter/experimenter/features/tests/__init__.py
brennie: Using `/` is idiomatic when the left-hand side is a pth.

If `FIXTURE_DIR` were a string, then we would have to do `Path(FIXTURE_DIR, "valid_features")`

experimenter/experimenter/features/__init__.py
brennie: We want it to be Optional because we're assigning it to None right there :)

experimenter/experimenter/features/management/commands/load_feature_configs.py
brennie: Yeah I'll update this comment to mention which slug.

experimenter/experimenter/features/management/commands/load_feature_configs.py
brennie: Right. Say Firefox 121 removes feature A. Then when we are on 121 nightly, the unversioned experimenter.yaml won't have feature A, but the versioned experimenter.yaml for 120, 120.0.1, etc. will have it listed.

experimenter/experimenter/features/management/commands/load_feature_configs.py
jaredlockhart: Yeah I know, if you use an actual variable there the linter would complain it wasn't being used but knows to ignore `_`.  I wish there was a better pattern for this in Python.

experimenter/experimenter/features/management/commands/load_feature_configs.py
brennie: Yeah i think in the future we want to move the description into the `NimbusVersionedSchema` because we could end up in a case where two features with the same name N versions apart aren't actually the same feature any more.

I'm not sure of a good way to present all this to the user right now, though, so I left the description on the `NimbusFeatureConfig` model, since that is what we are using to generate the UI on the branch page. Moving it into `NimbusVersionedSchema` would require a large refactoring of the front-end to support that.

jaredlockhart: Totally, I mean you don't need to do any of that refactoring really, just change description to a property that manages that lookup.  This is a general pattern we've used in lots of places, provide a facade of a field to the APIs/frontend that actually reasons about doing lookups, similar pattern to how we manage the dates.

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -1,64 +1,193 @@
...
+            if feature.version in versions:
+                continue
+
+            version = NimbusFeatureVersion(
+                major=feature.version.major,
+                minor=feature.version.minor,
+                patch=feature.version.patch,
+            )
+            versions[feature.version] = version
+            version.save()
```
jaredlockhart, line 131: Don't wanna roll these into a bulk create?

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -1,64 +1,193 @@
...
+                # the main branch (or equivalent)), so lets use the descriptions from
+                # there, if they exist.
+                continue
+
+            # Django doesn't keep track of whether or not fields were updated
+            # when you call save(). By default, it will update every field in
+            # the object, whether or not it was dirtied. However, if we are
+            # creating the object, we have to save every field anyway.
+            created = False
+            dirty_fields = []
```
jaredlockhart, line 66: I swear I thought Django models had a dirty tracking mechanism built in....

jaredlockhart (commented):
Yeah too bad we have to deal with this explicitly but everything looks right to me.  It'll be good to invest in a bit of performance tuning for this now since our number of applications and versions is going to continue to grow and we don't want to blowup start/load/deploy times.  Thnx for digging into the performance 👍

jaredlockhart:
If the number of cases to test gets too big it's probably sufficient to remove all the specific field checks and just roll up the bulk creates and leave the updates as naive.

Issue #9804, opened 2023-11-21
focus-ios FML fetching is broken on main


data-sync-user:
The Focus iOS and Firefox iOS are merging into a monorepo. In preparation for that, [https://github.com/mozilla-mobile/focus-ios/pull/3917](https://github.com/mozilla-mobile/focus-ios/pull/3917|smart-link)  landed in focus-ios earlier today which moves all the focus-specific stuff, including its FML, into a subdirectory. This breaks our fetch logic, which will need to be updating to support fetching different paths in different branches.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4063)

PR #9816, opened 2023-11-23
feat(nimbus): Include fetch summary in update_external_resources PR

### LINKED ISSUE #9815: Report fetch summary in PRs
The update external configs task now is importing a lot more manifests at once and its hard to keep track of whats going on. Additionally, sometimes individual fetches fail, which doesnt bring down the whole job (see [https://github.com/mozilla/experimenter/issues/9619](https://github.com/mozilla/experimenter/issues/9619|smart-link)).

The fetch command generates a summary, which we should include in the generated PR so we can tell what is going on in the fetch task. If a PR isnt generated (e.g., because all the fetches fail), we should instead hard fail the CI task so we get an alert and can fix it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4069)

brennie:
Because

- fetching feature manifests can fail (see #9804 for an example);
- the fetch command does not hard fail when an individual fetch fails; and
- we are generating a summary of all the fetches

This commit

- updates the fetch command to hard fail if there are fetch failures and no
  successes (i.e., in the case where it would not open or update a PR);
- adds a flag to the fetch command to write the summary to a file; and
- opens and updates PRs with a summary of the fetches in the PR description

Fixes #9815

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -508,7 +508,8 @@ jobs:
             git checkout main
             git pull origin main
             cp .env.sample .env
-            env GITHUB_BEARER_TOKEN="${GH_TOKEN}" make fetch_external_resources
+            env GITHUB_BEARER_TOKEN="${GH_TOKEN}" make fetch_external_resources FETCH_ARGS="--summary fetch-summary.txt"
+            mv ./experimenter/fetch-summary.txt /tmp/fetch-summary.txt
             if python3 ./experimenter/bin/should-pr.py
               then
                 git checkout -B external-config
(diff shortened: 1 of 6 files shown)


jaredlockhart (commented):
Oh heck yeah.  Can you think of way to force this so we can test it before we land it?  Like remove the main filter and temporarily futz the diff check to force it to create a PR just so we can see what it'll spit out before we land it?

brennie:
See https://github.com/mozilla/experimenter/pull/9821

jaredlockhart (approved):
So cool!!  We should later extend to include more information about the rich diff, like fields types etc.  Thnx @brennie 🎉 🎉 🎉 🎉

brennie:
Blocked on #9826

PR #9824, opened 2023-11-24
bug(nimbus): Disable features that appear only in versioned manifests

### LINKED ISSUE #9823: Correctly disable features that only exist in versioned manifests
[https://github.com/mozilla/experimenter/issues/9809](https://github.com/mozilla/experimenter/issues/9809|smart-link)  updated the import logic, but it made it so that a feature in _any_ manifest (unversioned or versioned) will cause it to be marked enabled. This is fine now, since we only have versioned features going back 3-5 versions (depending on the application), but soon we will stop disabling feature configs altogether.

The logic should be updated so that feature configs are only marked enabled if they’re present in an unversioned manifest.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4073)

brennie:
Because:

- we still want to mark old features in Nimbus as disabled; and
- #9809 updated the import logic to mistakenly mark every feature in every
  manifest as enabled;

This commit:

- updates importing to only consider unversioned features when disabling.

Fixes #9823

### DIFF
--- a/experimenter/experimenter/features/management/commands/load_feature_configs.py
+++ b/experimenter/experimenter/features/management/commands/load_feature_configs.py
@@ -58,50 +58,51 @@ class Command(BaseCommand):
                 # there, if they exist.
                 continue
 
-            # Django doesn't keep track of whether or not fields were updated
-            # when you call save(). By default, it will update every field in
-            # the object, whether or not it was dirtied. However, if we are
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Good catch 👍

PR #9826, opened 2023-11-24
bug(nimbus): Run should-pr.py with Python 3

### LINKED ISSUE #9825: update_external_config task not succeding due to syntax error
should-pr.py is being run with Python 2 insetad of Python 3, which is causing it to fail with syntax errors.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-4074)

brennie:
Because:

- should-pr.py is written in Python 3; and
- the default Python on Ubuntu 20.04 is Python 2.7

This commit:

- explicitly runs should-pr.py with Python 3 in CI.

Fixes #9825


jaredlockhart (approved):
🙏

PR #9846, opened 2023-11-28
Win11 pin targeting, with `isDefaultBrowser` check

AllegroFox:
Adds targeting for Win 11 build number 22621, for first startup users with a new profile, needing pin, and needing default.

This is for use with the [Pinning adaptation experiment](https://docs.google.com/document/d/1oivKNuH7CT_HcQkMiEm1vCOke65VpgRT6BFlVYllBok/edit#heading=h.uoblsnu302hk), in response to API changes to pinning in Win 11.


jaredlockhart (approved):
LGTM 👍

PR #9849, opened 2023-11-28
feat(nimbus): add note about preview delay

### LINKED ISSUE #9848: Add a note to preview button about CDN 1hr delay
After someone marks an experiment for preview, it can take up to an hour for the json to be replicated out to the Remote Settings CDNs which is confusing to users.  Let's add a little info note next to the button that indicates as much so people have a heads up.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4083)

jaredlockhart:
Because

* After an experiment is marked for preview, it can take up to an hour for it to be replicated into the Remote Settings CDN
* This often confuses users since they expect it to happen immediately

This commit

* Adds a note to the preview guidance about the delay

fixes #9848


eliserichards (approved):
Thank you SO much for adding this! 👏

PR #9853, opened 2023-11-28
chore(cirrus): Update sample env

### LINKED ISSUE #9852: Update sample.env
Update sample.env to only use the required cirrus variables

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4084)

yashikakhurana:
Because

- There are some fields in the sample env that we don't need it

This commit

- Remove extra fields from `sample.env`
Fixes #9852


jaredlockhart (changes requested):
Should we also update [the docs](https://github.com/mozilla/experimenter/tree/main/cirrus#environment-setup)?

yashikakhurana:
> Should we also update [the docs](https://github.com/mozilla/experimenter/tree/main/cirrus#environment-setup)?
good point, done!

jaredlockhart (commented):
Aren't these settings still being [used](https://github.com/mozilla/experimenter/blob/34dd9eb1a5dab7651645828629955267fe8525b0/cirrus/server/cirrus/settings.py#L32-L54)?

yashikakhurana:
> Aren't these settings still being [used](https://github.com/mozilla/experimenter/blob/34dd9eb1a5dab7651645828629955267fe8525b0/cirrus/server/cirrus/settings.py#L32-L54)?

these are defined though, but I don't see it getting used in the near future. I kept it in the settings but to avoid confusion just removed it from the docs like sample env or cirrus docs. In the future if we feel we want to edit something we can use these params

.env.sample
```diff
@@ -56,11 +56,4 @@ CIRRUS_APP_ID=demo-app-beta
...
-CIRRUS_SENTRY_DSN=
-CIRRUS_METRICS_BUILD=
-CIRRUS_METRICS_DATA_DIR=/var/glean
-CIRRUS_METRICS_LOG_LEVEL=WARNING
-CIRRUS_METRICS_MAX_EVENTS_BUFFER=500
-CIRRUS_SERVER_ENDPOINT=
-CIRRUS_METRICS_UPLOAD_ENABLED=True
-CIRRUS_METRICS_VERSION=1.0
\ No newline at end of file
+CIRRUS_SENTRY_DSN=
```
brennie, line 59: Missing newline at end of file.

brennie:
> > Aren't these settings still being [used](https://github.com/mozilla/experimenter/blob/34dd9eb1a5dab7651645828629955267fe8525b0/cirrus/server/cirrus/settings.py#L32-L54)?
> 
> these are defined though, but I don't see it getting used in the near future. I kept it in the settings but to avoid confusion just removed it from the docs like sample env or cirrus docs. In the future if we feel we want to edit something we can use these params

If these entries in settings.py are just placeholders for something we *might* need in the future, we should remove them. If they *were* being used, I'd say that we need to keep them in the docs.

We definitely don't want to end up in a place where these settings *are* doing something but aren't documented, leading to undocumented secret behaviour.

jaredlockhart:
I'm inclined to agree with @brennie , either they are useful settings in which case they should be documented, otherwise they're useless and we should remove them entirely until we have a use for them and then add them back.

yashikakhurana:
I think that's a good point, we really don't want to be in a state where we have something but it is not documented, I will remove them from the settings too, and will see in the future when we actually want to override any particular settings we will do it then.

yashikakhurana:
okay @jaredlockhart @brennie refactored the settings too 🙏 🙏 and thank you both for the feedback

jaredlockhart (approved):
Very nice 🙏 Thnx @yashikakhurana !  Deleting a bunch of unnecessary code, docs earns you a 🧁

PR #9863, opened 2023-11-29
bug(nimbus): fix export of qa_status enum

### LINKED ISSUE #9862: Handle new nullable enum during experiment export
We added a new `qa_status` field of type `NimbusExperimentQAStatusEnum`, which is a nullable enum on the NimbusExperiment model. This breaks the export/import without special handling, so we should make sure handle this case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4098)

mikewilli:
Because

- we added `qa_status` enum field to the experiment model
- the export/import converts a null value for an enum to empty string instead of `None`
- type checking doesn't allow empty string as enum value but `None` is allowed

This commit

- converts a null enum value to `None` explicitly during export/import

Fixes #9862


eliserichards (approved):
💯

PR #9867, opened 2023-11-30
chore(nimbus): use gql api in remote settings tests

### LINKED ISSUE #9841: Use GQL to create experiments in remote settings integration tests
There's two ways to create an experiment in the integration tests, either using the V5 GraphQL API, or using Selenium to drive the forms.  We should definitely use the UI for some integration tests but it's not necessary to use them for all of them.  Presently we use the forms to create experiments for the Experimenter UI integration tests, and so we ensure that all the forms are intact in that suite.  The longest running suite is the Remote Settings tests where we want to ensure that all the interactions between Experimenter and Remote Settings remain intact, but it's not necessary to also check the Experimenter forms in that suite since that's already stressed in the Experimenter UI tests.  Let's refactor the Remote Settings integration tests to use the GraphQL API to save time in CI.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4081)

jaredlockhart:
Because

* We should test the Nimbus UI somewhere in the integration tests
* Using the forms to create experiments is time consuming
* We use the Nimbus UI forms in the Experimenter UI integration tests
* We don't also need to test them in the Remote Settings tests
* The Remote Settings tests are already slow and so we can speed them up by using the V5 GQL API to create experiments in those tests

This commit

* Changes the Remote Settings tests to use the GQL API to create experiments
* Refactors the GQL API client in the integration tests to use the same set of default data as the UI creation data to be more reusable in the future

fixes #9841

### DIFF
--- a/experimenter/tests/integration/nimbus/conftest.py
+++ b/experimenter/tests/integration/nimbus/conftest.py
@@ -336,48 +336,6 @@ def trigger_experiment_loader(selenium):
     return _trigger_experiment_loader
 
 
-@pytest.fixture(name="experiment_default_data")
-def fixture_experiment_default_data():
-    return {
-        "hypothesis": "Test Hypothesis",
-        "application": BaseExperimentApplications.FIREFOX_DESKTOP.value,
-        "changelogMessage": "test updates",
-        "targetingConfigSlug": "no_targeting",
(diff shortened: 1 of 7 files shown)


jrbenny35 (approved):
Looks good, thanks for taking sometime won this!

PR #9873, opened 2023-12-01
feat(cirrus): Api docs for Cirrus

### LINKED ISSUE #8389: Generate static docs for the api
FastApi autogenerates the docs for us (swagger) /docs and /redocs, we want to store these docs statically. Add a make command too that if there are any updates or differences, generate the docs again,

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3218)

yashikakhurana:
Because

- We want to have static docs for the Cirrus's Api

This commit

- Generates api doc in `html` format whenever `openapi.json` gets updated
- Screenshot
Screenshot 2023-12-01 at 10 13 29 AM

Fixes #8389


jaredlockhart (approved):
Awesome!!!  Some followups to consider:

* Linking to this from the README in Cirrus or maybe even the top level one? (Although I think we should reorganize this at some point so the top level one is more about Nimbus and move the Experimenter specific docs down a level, maybe we can do that as part of moving experimenter-docs in?)
* Set up the tasks we have for the other doc generator that checks for differences in PRs [like this](https://github.com/mozilla/experimenter/blob/main/Makefile#L38-L39)?

But yeah this is awesome thnx @yashikakhurana 🎉

PR #9894, opened 2023-12-04
chore(Cirrus): pin rust version in cirrus dockerfile

### LINKED ISSUE #9893: Hard code rust version in Cirrus Dockerfile
We presently declare the `rust:buster` as the base image for the Cirrus image, however this will always take the latest version of Rust which may introduce unanticipated bugs with no code changes.  We should explicitly declare a version and then we can explicitly bump it when we need to and address any possible bugs there.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4121)

jaredlockhart:
Because

* We inherit from a rust base image in the Cirrus docker image
* We did not specify a pinned version
* This can cause updates to the base image to break our build without any code changes

This commit

* Pins the rust base image to 1.74.0

fixes #9893

### DIFF
--- a/cirrus/server/Dockerfile
+++ b/cirrus/server/Dockerfile
@@ -1,5 +1,5 @@
 # First stage: Build the Rust components
-FROM rust:buster AS rust-builder
+FROM rust:1.74.0-buster AS rust-builder
 
 RUN update-ca-certificates
 


yashikakhurana (approved):
good point, thanks @jaredlockhart

Issue #9857, opened 2023-11-29
Validate FML-based apps with versioned features in the API


data-sync-user:
Once Feature Versioning ([https://github.com/mozilla/experimenter/issues/9810](https://github.com/mozilla/experimenter/issues/9810|smart-link) ) is complete, we can use the same machinery to determine the versions we should use to validate FML-based apps (all non-desktop apps; android, ios, cirrus).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4089)

data-sync-user:
➤ Elise Richards commented:

https://github.com/mozilla/experimenter/issues/9810 ( https://github.com/mozilla/experimenter/issues/9810|smart-link )  is done, so this is now unblocked 👍

data-sync-user:
➤ Elise Richards commented:

NimbusFmlLoader logic changes: https://github.com/mozilla/experimenter/pull/9896 ( https://github.com/mozilla/experimenter/pull/9896|smart-link )

data-sync-user:
➤ Elise Richards commented:

Refactor for NimbusFmlLoader: https://github.com/mozilla/experimenter/pull/9897 ( https://github.com/mozilla/experimenter/pull/9897|smart-link )

data-sync-user:
➤ Elise Richards commented:

Serializer changes: https://github.com/mozilla/experimenter/pull/9912 ( https://github.com/mozilla/experimenter/pull/9912|smart-link )

PR #9896, opened 2023-12-05
For #9857 feat(nimbus): Add feature versioning to NimbusFmlLoader

eliserichards:
According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_n_) change.

Because...

- We want to fetch FML errors for the different versions of the FML that are being targeted

This commit...

- Changes the `NimbusFmlLoader` so that it accepts a version that it can use to check for errors
- Sets it up so that we can loop through versions in the serialization instead of looping through the versions in the loader itself (this is why the `_get_fml_clients` method is removed -- there is a client for each version, but if we only do one version at a time this method is unnecessary)

**This commit does not**

- Do anything with serialization (see PR #TBD)
- Make any refactorings that would interfere with the ease of reviewing (see PR #TBD)

For #9857


experimenter/experimenter/features/manifests/nimbus_fml_loader.py
```diff
@@ -24,55 +26,49 @@ def __init__(self, application: str, channel: str):
...
+                return None
         else:
             logger.error(
                 "Nimbus FML Loader: Invalid application. Failed to get local "
                 "feature manifest."
             )
         return None
 
-    # Todo: Add versioning https://mozilla-hub.atlassian.net/browse/EXP-3875
-    def _get_fml_clients(self) -> list[FmlClient]:
```
eliserichards, line 41: No longer necessary since we will handle only one version at a time in the loader.

brennie (approved):

jaredlockhart:
@eliserichards +95 −143 🧁 🧁 🧁 🧁 🧁

eliserichards:
It's failing on `DEMO_APP` tests since there are no `fml.yaml` files. Adding a check to prevent this 👍

PR #9908, opened 2023-12-06
Remove property checks from PDF targeting

bytesized:
Although checking for the presence of properties using "in" doesn't seem to cause any kind of parsing error, it also never seems to evaluate to true.

Because

- "Default PDF handler is a different browser" Advanced Targeting always seems to evaluate to `false`

This commit

- Removes property checks from the PDF targeting definition.

Fixes [Bug 1868553](https://bugzilla.mozilla.org/show_bug.cgi?id=1868553)


bytesized:
I did a bit more testing this time around, and it seems that this returns `null` on non-Windows. Is that okay? Will that be equivalent to `false`?

brennie:
> I did a bit more testing this time around, and it seems that this returns `null` on non-Windows. Is that okay? Will that be equivalent to `false`?

On desktop, [we coerce everything to a boolean](https://searchfox.org/mozilla-central/rev/16526d690cccc9c60cacf09cc08e2c3041ef884e/toolkit/components/nimbus/lib/RemoteSettingsExperimentLoader.sys.mjs#561-573), so returning `null` will be fine :)

brennie (approved):

Issue #9860, opened 2023-11-29
Ingest Pairwise Branch Comparisons Results into Experimenter


data-sync-user:
Ingest results with pairwise branch comparisons into Experimenter. We may need to change how ingestion works depending on where assumptions are made about a fixed reference branch. We will also likely need to change the data format so that the UI can more easily switch between branches without doing filtering on deeply nested results.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4093)

PR #9921, opened 2023-12-07
chore(nimbus): refactor jetstream client with type hints etc

mikewilli:
related to #9860 

O(1) or O(n)?

Because

- I wanted to more easily see what objects were being passed around during results ingestions

This commit

- Adds more type hints
- refactors schema objects to leverage the schemas package

no fixes yet


jaredlockhart:
Hey neat!!  And pyright is actually kicking into action!!

```py
  /experimenter/experimenter/jetstream/models.py:229:69 - error: "slug" is not a known member of "None" (reportOptionalMemberAccess)
```

🎉 🎉 🎉 🎉 

Pyright can't really help us with Django code but we can still use it in other parts of the code that are more isolated, like the Jetstream handling, the manifesttool, etc.  So I like this trend!

jaredlockhart:
And yeah exactly, the types represent the state all the way from when the experiment is created, at which time `reference_branch` CAN be `None` but not by the time we get here.  I've imagined ways we could represent that more clearly, ie having different types depending on what state it's in, but that's not very well representable using Python typing.  Rust on the other hand....

experimenter/experimenter/jetstream/models.py
```diff
@@ -70,17 +80,12 @@ class Group:
...
-    metric: str = None
-    branch: str = None
-    statistic: str = None
-    window_index: str = None
-    comparison: str = None
-    segment: str = Segment.ALL
-    analysis_basis: str = AnalysisBasis.ENROLLMENTS
+class JetstreamDataPoint(JetstreamStatisticResult):
+    """
+    Same as the mozilla-nimbus-schemas `Statistic` but sets a default analysis_basis.
```
brennie, line 85: Python docstrings should begin on the `"""` line.

mikewilli: Yea, but then it goes over the 90-character line length. I guess I could edit the sentence to make it a few characters shorter.

experimenter/experimenter/jetstream/models.py
```diff
@@ -1,7 +1,17 @@
 from typing import Any, Dict, List
 
+from mozilla_nimbus_schemas.jetstream import AnalysisBasis
+from mozilla_nimbus_schemas.jetstream import Statistic as JetstreamStatisticResult
 from pydantic import BaseModel, create_model
 
+from experimenter.experiments.models import NimbusExperiment
+
+
+class AnalysisWindow:
```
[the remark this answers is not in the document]
mikewilli: Yea good call out. The goal of the PR was not a comprehensive refactor, but merely to check in the code that I had changed while figuring out what needed to be changed for pairwise branch comparisons. However, I appreciate the suggestions and think that at least some of these (like this one!) are worth addressing now.

PR #9933, opened 2023-12-12
fix(nimbus): use StrEnum for string conversion

### LINKED ISSUE #9932: Experiment results ingestion building path to results wrong
`AnalysisWindow` enum not being converted to string properly

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4147)

mikewilli:
Because

- #9921 broke analysis results ingestion by including the enum member in the string path to results instead of the enum value
- the fix was obvious enough that it was worth fixing instead of reverting

This commit

- uses `StrEnum` as the base class for enums in `experimenter.jetstream.models`

Fixes #9932

### DIFF
--- a/experimenter/experimenter/jetstream/models.py
+++ b/experimenter/experimenter/jetstream/models.py
@@ -1,4 +1,4 @@
-from enum import Enum
+from enum import StrEnum
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Oh interesting I haven't used that one before, looks good thnx @mikewilli 👍

PR #9936, opened 2023-12-12
chore(nimbus): fetch results every 6 hours on 12/6/12/6 (UTC)

### LINKED ISSUE #9935: Investigate configuring results fetch on regular schedule
Currently this runs every 8 hours, but it starts 8 hours after a deploy. This makes it hard to know when results were last fetched, and might delay fetching during times with frequent deploys.

It would be nice to do something like a cron schedule where we just set the times we want to fetch. Can we do that with celery tasks in django? And if we're updating the schedule, do we want to keep the same 8-hour increments?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4149)

mikewilli:
Because

- Experimenter fetches results every 8 hours, starting 8 hours after a deploy
- This makes it hard to know when results were last fetched, and might delay fetching during times with frequent deploys

This commit

- changes to a crontab style scheduling pattern, fetching at the same times every day
- increases the frequency of fetches to every 6 hours (at 12/6/12/6 UTC)

Fixes #9935

### DIFF
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -15,6 +15,7 @@ from importlib import resources
 from pathlib import Path
 from urllib.parse import urljoin
 
+from celery.schedules import crontab
 from decouple import config
 from django.contrib.admin import ModelAdmin, StackedInline, TabularInline
 from django.db.models import DecimalField, ForeignKey, JSONField, ManyToManyField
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Ah interesting!  Thnx @mikewilli 🙏

PR #9939, opened 2023-12-13
feat(nimbus): Add VPN application support

### LINKED ISSUE #9937: VPN application option in experimenter
Add new application option for VPN

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4113)

yashikakhurana:
Because

- We are integrating VPN with the experimenter using Cirrus

This commit

- Add new VPN application support

Fixes #9937


yashikakhurana:
Waiting on this discussion to finalized- https://bugzilla.mozilla.org/show_bug.cgi?id=1867770

yashikakhurana:
Ready to review now- this PR has been merged now https://github.com/mozilla/probe-scraper/pull/670

experimenter/experimenter/experiments/constants.py
```diff
@@ -19,6 +19,7 @@ class Channel(models.TextChoices):
     DEVELOPER = "developer"
     STAGING = "staging"
     PRODUCTION = "production"
+    VPN = "vpn"
```
jaredlockhart, line 22: Wait why do we need another channel here?

yashikakhurana: oh, you remember we discussed that we don't need to create a channel, as the VPN stage will always point to Remote stage and the Prod will always point to the prod remote, so we don't need to divide the channel essentially.

yashikakhurana: We can just use single channel, this is still pending to update for Monitor, but for VPN- starting with the single channel

eliserichards (approved):
🎉

experimenter/experimenter/experiments/constants.py
jaredlockhart: No but I mean why can't we just use one of the existing channels like `production`? Why do we need to make a new one just for VPN?

yashikakhurana: oh, I see what you mean, because I was thinking that they can get confused with `production` as a channel option on `stage experimenter console`, what do you think @jaredlockhart?

jaredlockhart: No I'd say just leave it as 'production'.  We should try to avoid blowing up the list of channels as much as possible I think.  If people want to change it later we can but I think 'production' still has meaning here.  And I mean that's exactly what we did for monitor too.

yashikakhurana: for monitor they have two options- stage and production

yashikakhurana:
@jaredlockhart changes done and now only channel is available as `production`

PR #9941, opened 2023-12-13
chore(nimbus): remove feature config schema from getAllExperiments gql call

### LINKED ISSUE #9940: Remove feature config schemas from getAllExperiments in GQL
The call to getAllExperiments that's used to populate the landing page has been increasing in size and is now 17mb, which is quite a lot.  It seems the largest offender here is including the feature config schema in that call which is not actually used anywhere on the landing page and only needs to be included in the getConfigs and getExperiment calls that are used for the forms and summary pages.  By removing that field we may be able to shrink the call on the landing page by up to 97% from local testing of an extreme case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4150)

jaredlockhart:
Because

* We included the feature config schema in the getAllExperiments call that's used to populate the landing page
* Those schemas can get quite large
* That field is not used by any callers of that query
* It can be safely removed
* This should reduce the size of the payload and time to load that page

This commit

* Removes feature config schema from the getAllExperiments gql query

fixes #9940

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/gql/experiments.ts
+++ b/experimenter/experimenter/nimbus-ui/src/gql/experiments.ts
@@ -251,7 +251,6 @@ export const GET_EXPERIMENTS_QUERY = gql`
         description
         application
         ownerEmail
-        schema
       }
       targetingConfig {
         label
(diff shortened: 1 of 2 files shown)


eliserichards (approved):
Are there any tests for `getAllExperiments`?

jaredlockhart:
@eliserichards Oh many.  Static type checking should infer whether anybody used that field, the frontend unit tests, and the integration tests will all cover it.

yashikakhurana (approved):
good catch, thanks @jaredlockhart

PR #9947, opened 2023-12-15
feat(nimbus): use the GitHub contents API to find download URLs

### LINKED ISSUE #9946: Get download URLs via the contents API
If we want to download files from private repos, we need to first hit the file contents API to get a download URL with a token.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4153)

brennie:
Because: 

- we can't directly access raw.githubusercontent.com paths for private
  repositories, even with a GitHub API key

This commit:

- updates fetch logic to use the file contents API, which will return a
  download URL that includes a token; and
- downloads the file from the returned URL.

Fixes #9946

### DIFF
--- a/experimenter/manifesttool/github_api.py
+++ b/experimenter/manifesttool/github_api.py
@@ -8,7 +8,6 @@ from manifesttool import download
 from manifesttool.repository import Ref
 
 GITHUB_API_URL = "https://api.github.com"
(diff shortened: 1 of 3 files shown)


jaredlockhart (approved):
Oh crazy, good catch, nice simple fix though, very nice thnx @brennie 🎉 🎉 🎉

PR #9948, opened 2023-12-15
feat(nimbus): ingestion updates to support pairwise branch comparisons

### LINKED ISSUE #9860: Ingest Pairwise Branch Comparisons Results into Experimenter

mikewilli:
Because

- Jetstream will soon be able to send pairwise branch comparison results

This commit

- Ingests results for each branch as reference instead of a statically defined reference from the experiment

Fixes #9860


mikewilli:
Some notes now that this is out of draft:
- Vast majority of the changes were for new or updated test data
- In addition to the existing and new tests, I tested this against a bunch of real experiments in my local environment by manually importing them and then triggering a results fetch
- The changes to the transformed results are all backwards-compatible (i.e., no changes needed to the UI and no new schema version needed)
- I removed the `"daily"` results from the transformed data because they are not used in the UI (to keep it backwards compatible, I left the empty structure) and this substantially reduces the payload size
- Each metric now contains some additional branch comparison data
  - We can come back later and take out some of the old objects once the UI is updated

The new transformed results look like the following (with comment annotations):
```
    [window]: {
        [analysis_basis]: {
            [segment]: {
                [branch]: {
                    "is_control":  true/false,
                    "branch_data": { // uses the experiment-defined reference_branch as control
                        [metrics_group]: {
                            [metric_name]: {
                                "absolute": { // ALL DATA
                                    "all": [...],
                                    "first": {...}
                                },
                                "difference": {
                                    // all and first are compared to defined reference
                                    "all": [...],
                                    "first": {...},
                                    // ADDITION (1 for each other branch)
                                    [comparison_branch]: {
                                        "all": [...],
                                        "first": {...},
                                    },
                                },
                                "significance": {
                                    "weekly": {},
                                    "overall: {}
                                },
                                "relative_uplift": {
                                    // all and first are compared to defined reference
                                    "all": [...],
                                    "first": {...},
                                    // ADDITION (1 for each other branch)
                                    [comparison_branch]: { // compared to `comparison_branch`
                                        "all": [...],
                                        "first": {...},
                                    },
                                }
                            }
                        }
                    }
                }
            }
        }
    }
```

jaredlockhart (approved):
The tests all look good in here, this is the kind of thing it would be nice to have some kind of end to end integration test with the visualization UI to make sure it doesn't choke on this change but without that we'll just have to merge it and wait and see.  Thnx @mikewilli this is great 🎉

PR #9967, opened 2023-12-19
feat(nimbus): Support multiple locations for version files

### LINKED ISSUE #9966: Support Firefox iOS Monorepo: support multiple locations for version files
focus-iOS has moved the file we are using for versioning on main and releases_v122 branches, so we can no longer find it. We need to support looking up multiple version files, in the case where they get moved across branches.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4161)

brennie:
Because

- firefox-ios and focus-ios are merging into a single mono-repo;
- focus-ios relocated the file we use to determine versions; and
- firefox-ios will soon be doing the same

This commit

- updates AppConfig to support multiple version files;
- updates the resolving logic to try each file in turn; and
- updates the version file entries for firefox-ios and focus-ios.

Fixes #9966

### DIFF
--- a/experimenter/experimenter/features/manifests/apps.yaml
+++ b/experimenter/experimenter/features/manifests/apps.yaml
@@ -29,7 +29,9 @@ firefox_ios:
   release_discovery:
     version_file:
       type: "plist"
-      path: "Client/Info.plist"
+      path:
+        - "firefox-ios/Client/Info.plist"
+        - "Client/Info.plist"
       key: "CFBundleShortVersionString"
     strategies:
       - type: "tagged"
@@ -64,7 +66,9 @@ focus_ios:
   release_discovery:
     version_file:
       type: "plist"
-      path: "Blockzilla/Info.plist"
+      path:
+        - "focus-ios/Blockzilla/Info.plist"
+        - "Blockzilla/Info.plist"
       key: "CFBundleShortVersionString"
     strategies:
       # We don't use a branched strategy for focus_ios because it is on v9000.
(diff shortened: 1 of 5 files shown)


experimenter/experimenter/features/manifests/apps.yaml
```diff
@@ -29,7 +29,9 @@ firefox_ios:
   release_discovery:
     version_file:
       type: "plist"
-      path: "Client/Info.plist"
+      path:
```
jaredlockhart, line 32: This is an example where we could've made it a new field called 'paths' that only takes a list rather than folding two types into one field?  We control both sides so we don't really need to reason about backwards compatibility, just musing.

jaredlockhart (approved):
Awesome, good test coverage 🎉 🎉 🎉 🎉

experimenter/experimenter/features/manifests/apps.yaml
brennie: We could, but then the pydantic schema needs code to validate that one of `paths` or `path` is specified, which is more complicated IMO

experimenter/manifesttool/tests/test_version.py
```diff
@@ -411,9 +411,9 @@ def test_parse_version_file_plist(self, version_str, expected):
             (
                 app_config,
                 [
-                    (Ref("v120", "r0"), "120.0a1\n"),
-                    (Ref("v120.1", "r1"), "120.1a1\n"),
-                    (Ref("v121", "r2"), "121.0a1\n"),
+                    (Ref("v120", "r0"), b"120.0a1\n"),
+                    (Ref("v120.1", "r1"), b"120.1a1\n"),
+                    (Ref("v121", "r2"), b"121.0a1\n"),
```
eliserichards, line 416: Why do we need these newlines?

brennie: Those are the bodies of the mock HTTP responses. We don't technically need them, but version files in practice all end with a newline (e.g., https://github.com/mozilla-mobile/firefox-android/blob/main/version.txt)

jrbenny35 (approved):
Thanks for the quick fix

experimenter/experimenter/features/manifests/apps.yaml
```diff
@@ -76,37 +80,37 @@ focus_ios:
...
-  repo:
-    type: "github"
-    name: "mozilla/blurts-server"
-  fml_path: "config/nimbus.yaml"
+# monitor_cirrus:
+#   slug: "monitor-web"
+#   repo:
+#     type: "github"
+#     name: "mozilla/blurts-server"
+#   fml_path: "config/nimbus.yaml"
```
yashikakhurana, line 88: @brennie I am just wondering any reason to comment out monitor?

yashikakhurana: and same for firefox desktop, both are comment out

experimenter/experimenter/features/manifests/apps.yaml
```diff
@@ -76,37 +80,37 @@ focus_ios:
...
+#           # both of those will have the same version number for a period of
+#           # time.
+#           #
+#           # In this order, if two consecutive branches have the same version,
+#           # then the latter will take precedence. e.g., if both beta and release
+#           # report the same version, we will use the feature manifest from release.
+#           - "central"
+#           - "beta"
+#           - "release"
+#           - "esr115"
```
yashikakhurana, line 116: this too 👓

PR #10003, opened 2024-01-05
feat(schemas): add results ingestion enums to schemas package

### LINKED ISSUE #10001: Migrate results ingestion enums to schemas package
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4176)

mikewilli:
Because

- Some of these enums could be useful in the UI as typescript types

This commit

- Copies the enums defined in Experimenter's Jetstream results ingestion client to the schemas package

Fixes #10001


mikewilli:
So I realized that unfortunately this doesn't actually help the problem, at least not with the way we're generating typescript types from Pydantic. Because these are just enums (`StrEnum`) and don't inherit from pydantic's `BaseModel`, they don't get picked up by the pydantic-to-typescript converter, which is why the `index.d.ts` file is unchanged in this PR. 

I think it's still worth merging this PR so that they're available to the python Jetstream client, but we should look into alternate methods for converting everything over to typescript.

jaredlockhart (approved):
Yeah absolutely putting them in here is a good move.  We can worry about how to get them to the frontend later since a lot of that will probably be moving around this half anyways.

PR #10007, opened 2024-01-08
Update statistics.py

danielkberry:
Adds field to statistics schema

Because

- It would be useful to data scientists if Jetstream results contained p-values

This commit

- Adds p-values to the shared schemas

Fixes #github_issue_number


mikewilli (approved):
:tada:

Issue #9902, opened 2023-12-05
Add field for qa status comment


data-sync-user:
We want to add a field on the backend so that we can add a comment along with the QA status so that we can add context when the status is set

!image-20231205-171306.png|width=1142,height=500!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4126)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/53990">image-20231205-171306.png</a>

PR #10012, opened 2024-01-08
Fixes #9903 feat(nimbus): Add UI for QA comments

### LINKED ISSUE #9903: Add textfield UI for qa status comment
We want to add a textfield on the Summary page in the QA Status editor so that we can add a comment along with the QA status so that we can add context when the status is set

!image-20231205-171306.png|width=50%!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4127)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/53991">image-20231205-171306.png</a>

eliserichards:
Because

- #9902 added a backend field for qa comments

This commit

- Adds the UI for qa comments:
   - "Comments" section does not show up if qa_comment is null/empty
   - Newline are honored 
   - You can edit the comment as many times as you want (these will show up in the history page)
   - You can edit comment and qa status independently (you don't have to add a comment every time you edit the qa status)
   - It wordwraps correctly when you're not in edit mode
   - It shows an error if the input is longer than 4096 chars

Fixes #9903 

----

Screenshot 2024-01-08 at 12 40 35 PM
Screenshot 2024-01-08 at 12 41 32 PM
Screenshot 2024-01-08 at 1 50 04 PM

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -899,7 +899,11 @@ class NimbusExperimentSerializer(
         required=False,
     )
     qa_comment = serializers.CharField(
-        min_length=0, max_length=4096, required=False, allow_blank=True
+        min_length=0,
+        max_length=4096,
+        required=False,
+        allow_blank=True,
+        allow_null=True,
     )
 
     class Meta:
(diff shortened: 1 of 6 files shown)


jaredlockhart (changes requested):
Yep all looks right and works right, one thing missing, doesn't display error if input is too long:

image

eliserichards:
Gah! good catch 😛

eliserichards:
Done! and it also wordwraps now (I forgot that too)

Screenshot 2024-01-08 at 2 05 14 PM

Screenshot 2024-01-08 at 2 21 58 PM

jaredlockhart:
Ah another one, should be able to save the flag without a comment I think?

image

eliserichards:
Bah! Needed `allow_null=True` in addition to `allow_blank=True`

jaredlockhart (approved):
Yep all seems good, this is awesome thnx @eliserichards 🎉 🎉 🎉

PR #10015, opened 2024-01-09
chore(schemas): bump version to 2024.1.2

mikewilli:
Because

- #10007 updated the schemas package but not the versions

This commit

- updates the version in the config files so that it will be deployed by CI


danielkberry (approved):
LGTM

PR #10035, opened 2024-01-11
bug(nimbus): dedupe results during ingestion

### LINKED ISSUE #10034: Results page fails to load "E[t] is undefined"
Seeing this error again for [this experiment](https://experimenter.services.mozilla.com/nimbus/set-default-pdf-handler-rollout/results). We should add some resiliency against duplicate data in the results on the ingestion side, even if the error itself is on the Jetstream side.

cc @danielkberry

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4199)

mikewilli:
Because

- Experimenter cannot control Jetstream pushing duplicate results
- duplicate results breaks the Results page

This commit

- makes Experimenter more resilient to dupes from Jetstream by deduping them during ingestion

Fixes #10034


jaredlockhart (approved):
Oh cool good idea.  Tests? 🤔

mikewilli:
Yea sorry this should've been opened in draft. Just updated the tests so the existing ones pass and added a dupe to one of them.

PR #10045, opened 2024-01-12
feat(cirrus): Configure sentry for different projects

### LINKED ISSUE #9999: Add sentry tags
Since all the web applications will be using sentry, it will be hard to identify which container failed, so we should include sentry tag, we can use the value from .env `CIRRUS_INSTANCE_NAME=test_instance`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4175)

yashikakhurana:
Because

- We are using Sentry for all the Cirrus implementing applications such as Monitor, VPN, and so on...... Currently, it is hard to track where the error is coming from until we see the issue in detail. 

This commit

- Sentry supports the concept of environment, so in the same project you can configure different environment
- By default, Sentry uses a `production` environment, so the default value used in initialization is `production`
- Add support for using environment so that the `cirrus-stage` project can have `monitor-stage` and `vpn-stage` and similarly `cirrus-prod` can have `monitor-prod` and `vpn-prod` environments.

Note: Sentry [environment](https://docs.sentry.io/product/sentry-basics/concepts/environments/) selection also makes it easy to filter the issues, for example
image

Fixes #9999

### DIFF
--- a/.env.integration-tests
+++ b/.env.integration-tests
@@ -52,3 +52,4 @@ CIRRUS_APP_NAME=demo_app
 CIRRUS_CHANNEL=release
 CIRRUS_FML_PATH=./feature_manifest/sample.yml
 CIRRUS_SENTRY_DSN=
+CIRRUS_ENV_NAME=test_instance_stage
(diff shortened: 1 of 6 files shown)


jaredlockhart (approved):
Fantastic!!  This is a great idea 👏 👏 👏 👏

yashikakhurana:
Okay both monitor and VPN SRE has been informed about the new variable

PR #10055, opened 2024-01-16
feat(nimbus): Update QA status to not be nullable

### LINKED ISSUE #10036: Add option for Not Set to QA status enum
Let’s add a fourth option for “Not Set” instead of having the not set case be `null`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4200)

eliserichards:
Because

- We don't want to rely on `null` to represent an unset QA status

This commit

- Changes the Enum values of `QAStatus` to include `NOT_SET`, and removes the nullability on the backend field
- The migration sets all previous `null` values to `NOT_SET`

Fixes #10036

----

Screenshot 2024-01-12 at 1 26 43 PM
Screenshot 2024-01-12 at 1 26 54 PM
Screenshot 2024-01-12 at 1 26 30 PM


experimenter/experimenter/experiments/admin.py
```diff
@@ -109,9 +109,9 @@ def dehydrate_status_next(self, experiment):
         return experiment.status_next
 
     def dehydrate_qa_status(self, experiment):
-        """Return None instead of empty string for nullable enums"""
+        """Return NOT_SET instead of empty string for nullable enums"""
         if experiment.qa_status not in dict(NimbusConstants.QAStatus.choices):
-            return None
```
eliserichards, line 114: @mikewilli I'm not sure if this is valid or not 🤔

mikewilli: Looks fine to me. It's basically making sure the QA Status is valid so as long as what you return is valid that should work. My only concern is that it may be the string value that is needed, but you can add or update a test (and test locally) to figure that out. Probably worth exporting and importing a couple experiments locally too.

experimenter/experimenter/experiments/migrations/0257_alter_nimbusexperiment_qa_status.py
```diff
@@ -0,0 +1,39 @@
...
+
+from experimenter.experiments.constants import NimbusConstants
+
+
+def set_default_qa_status(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+    for experiment in NimbusExperiment.objects.all():
+        if experiment.qa_status is None or experiment.qa_status == "":
+            experiment.qa_status = NimbusConstants.QAStatus.NOT_SET
+            experiment.save()
```
eliserichards, line 13: I added this directly into the generated migration. Should I have separated this out into a separate manual one?

experimenter/experimenter/experiments/admin.py
mikewilli: Also this function may not even be necessary anymore because it is for nullable enums, and it looks like you changed the field to not be nullable anymore. If you want to remove it, you could test it out by exporting some experiments from Prod and importing them into your local instance, and then also export/import from/to local. I think it would break with the former but work with the latter (assuming migrations have been run locally but not yet on prod).

eliserichards: awesome, I'll test that!

experimenter/experimenter/experiments/migrations/0257_alter_nimbusexperiment_qa_status.py
jaredlockhart: Actually you can just write this as 

```py
NimbusExperiment.objects.filter(qa_status=None).update(qa_status=NimbusConstants.QAStatus.NOT_SET)
```

jaredlockhart: And no that's exactly the right place to put it

experimenter/experimenter/experiments/admin.py
```diff
@@ -280,6 +280,9 @@ class NimbusExperimentAdminForm(forms.ModelForm):
     conclusion_recommendation = forms.ChoiceField(
         choices=NimbusExperiment.ConclusionRecommendation.choices, required=False
     )
+    qa_status = forms.ChoiceField(
+        choices=NimbusExperiment.QAStatus.choices, required=False
```
yashikakhurana, line 284: I think we shoud set a default here as `Not Set`

yashikakhurana: because your migration only sets this value for the existing experiments in the DB, for the future experiments, if they don't select the value, we can set `Not set` as a default value

experimenter/experimenter/experiments/models.py
```diff
@@ -902,7 +901,7 @@ def clone(self, name, user, rollout_branch_slug=None, changed_on=None):
         cloned._start_date = None
         cloned._end_date = None
         cloned._enrollment_end_date = None
-        cloned.qa_status = None
+        cloned.qa_status = NimbusExperiment.QAStatus.NOT_SET
```
yashikakhurana, line 904: good point, we always forget to make changes on the clone 💪

yashikakhurana (commented):
Few minor suggestions, otherwise looks good to me 🌟

experimenter/experimenter/experiments/admin.py
brennie: I think it also might have to be requried=True

eliserichards: It should be `required=False` for the admin, I'm pretty sure. All of the fields are not required here, and I'm pretty sure you don't want to have to set `qa_status` every time you make a change. I also don't think the admin has to set a default, since that is set in the [model](https://github.com/mozilla/experimenter/pull/10055/files#diff-181caffbf9f9e1237da4cac08b35076e64e1321a12ee1935c61c7af2c155bf25) 🤔  I'm also not super familiar with the admin, so pls correct me if I'm wrong

experimenter/experimenter/experiments/admin.py
eliserichards: Looks like it's not needed anymore! Both the import and export are working perfectly without the qa_status dehydration 💪

experimenter/experimenter/experiments/admin.py
yashikakhurana: I am looking at the django docs if set a default on model, will be used in admin as well or not

yashikakhurana: yup just checked the docs, if we have the value default defined on the model, then admin will recreate and use those models default and pre-populate the default value, @eliserichards also have verified this locally, so all set 💯

yashikakhurana (approved):
looks good, thanks @eliserichards for referring to docs as well 👏

PR #10056, opened 2024-01-16
tests(nimbus): Use a random qa status in factories

### LINKED ISSUE #10038: Include QA status in factories
This will make it easier for testing

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4201)

eliserichards:
Because

- We want to include qa_status in our factories 

This commit

- Uses a random choice of qa status

Fixes #10038

### DIFF
--- a/experimenter/experimenter/experiments/tests/api/v5/test_mutations.py
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_mutations.py
@@ -1507,8 +1507,8 @@ class TestUpdateExperimentMutationMultiFeature(GraphQLTestCase):
         user_email = "user@example.com"
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.LAUNCH_APPROVE_APPROVE,
+            qa_status=NimbusExperiment.QAStatus.NOT_SET,
         )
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/tests/test_admin.py
```diff
@@ -208,6 +208,7 @@ def test_resource_dehydrate_none(self):
         experiment.reference_branch = None
         experiment.status_next = ""
         experiment.conclusion_recommendation = ""
+        experiment.qa_status = None
```
yashikakhurana, line 211: I think from your previous changes, it should be set to `Not set`

yashikakhurana (commented):
one minor note, otherwise looks healthy

experimenter/experimenter/experiments/tests/test_admin.py
jaredlockhart: Oh yeah that's a good point, that field should be not null now and fall back to the 'not set' state

eliserichards: That will all be updated when the other one lands 👍 Whichever one lands first, this will be fine

brennie (approved):

PR #10064, opened 2024-01-18
[WIP] feat(nimbus): Group errors for desktop and FML based on version

### LINKED ISSUE #9917: Group FML errors based on similarities
We want to group the versioned FML errors a little bit better so that it doesn’t look like this:

!Screenshot 2023-12-06 at 4.57.18 PM.png|width=1569,height=762!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4141)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/54162">Screenshot 2023-12-06 at 4.57.18 PM.png</a>

eliserichards:
Because

- We want a more user-friendly way to surface errors that exist in multiple versions

This commit

- Groups the errors by version for both FML errors and desktop schema
   - For a range of versions, it now looks like: "{err} between versions {min_version} and {max_version}"
   - For a single version: "{err} at version {version}"

Fixes #9917

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1,3 +1,4 @@
+from distutils.version import Version
 import json
 import logging
 import re
@@ -16,7 +17,7 @@ from django.db.models import Prefetch
 from django.utils.text import slugify
 from rest_framework import serializers
 from rest_framework_dataclasses.serializers import DataclassSerializer
-
+from manifesttool import version as manifesttool_version
 from experimenter.base.models import (
     Country,
     Language,
@@ -1607,7 +1608,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
                     feature_config=feature_config,
                 )
             ]
-        errors = defaultdict(list)
+        errors = []
 
         for version in schemas_in_range.unsupported_versions:
             errors.append(
@@ -1617,30 +1618,43 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
                 )
             )
 
+        errs_with_version: dict[str, list[NimbusFeatureVersion]] = dict()
         for schema in schemas_in_range.schemas:
             version = schema.version
             if fml_errors := loader.get_fml_errors(blob, feature_config.slug, version):
-                for e in fml_errors:
-                    errors[e].append(version)
-                # errors.extend(
-                #     [
-                #         f"{NimbusExperiment.ERROR_FML_VALIDATION}: {e.message} at line "
-                #         f"{e.line+1} column {e.col} at version {version}"
-                #         for e in fml_errors
-                #     ]
-                # )
+                for e in fml_errors: 
+                    if e.message in errs_with_version:
+                        errs_with_version[e.message].append(version)
+                    else:
+                        errs_with_version[e.message] = [version]
+        
 
+        formatted_errs = self._format_grouped_errors(errs_with_version)
+        for e in formatted_errs:
+            errors.append(e)
         return errors
 
     def _format_grouped_errors(self, errs: dict[str, list[NimbusFeatureVersion]]):
         formatted = []
         for err, versions in errs.items():
             if len(versions) > 1:
-                formatted.append(f"{err} between versions {min(versions)} and {max(versions)}")
+                formatted.append(
+                    NimbusConstants.ERROR_FEATURE_CONFIG_UNSUPPORTED_IN_VERSION_RANGE.format(
+                        err=err,
+                        min_version=min(versions),
+                        max_version=max(versions),
+                    )
+                )
             elif len(versions) == 1:
-                formatted.append(f"{err} at version {versions[0]}")
+                formatted.append(
+                    NimbusConstants.ERROR_FEATURE_CONFIG_UNSUPPORTED_IN_VERSION_RANGE.format(
+                        err=err,
+                        version=versions[0],
+                    )
+                )
             else:
                 formatted.append(err)
+        return formatted
 
     def _validate_schema(
         self, obj: Any, schema: dict[str, Any], version: Optional[NimbusFeatureVersion]
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1596,7 +1596,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         blob: str,
         min_version: packaging.version.Version,
         max_version: Optional[packaging.version.Version],
-    ):
+    ) -> dict[str, list[NimbusFeatureVersion]]:
         schemas_in_range = feature_config.get_versioned_schema_range(
             min_version, max_version
         )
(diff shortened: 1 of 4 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1521,6 +1521,20 @@ def _validate_required_or_excluded_experiments(self, value):
...
+        existing_errs_with_version: dict[str, list[NimbusFeatureVersion]],
+    ) -> dict[str, list[NimbusFeatureVersion]]:
+        errs_with_version = existing_errs_with_version
+        for e in errs:
+            if e in errs_with_version:
+                errs_with_version[e].append(version)
+            else:
+                errs_with_version[e] = [version]
+        return errs_with_version
+
```
brennie, line 1537: Python has something for this exact use case: [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict). It is a dict subclass that accepts a factory that will be used to generate values for keys not present in the dict. It allows you to do:

```python
>>> from collections import defaultdict
>>> xs = defaultdict(list)
>>> xs[0].append("a")
>>> xs[1].append("b")
>>> xs
defaultdict(<class 'list'>, {0: ['a'], 1: ['b']})
```

Additionally, you can also use [`list.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) if you only need that behaviour sometimes:

```python
>>> ys = {}
>>> ys.setdefault(0, []).append("a")
>>> ys
{0: ['a']}
```

I would reccomend using `defaultdict` in this instance and then return it with `dict(xs)` where xs is the defaultdict instance. That way the returned value will not have the defaultdict behaviour.

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1571,19 +1590,28 @@ def _validate_feature_value_against_schema(
...
 
-                        if schema.version is not None:
-                            err_msg += f" at version {schema.version}"
-
-                        errors.append(err_msg)
-                        errors.extend(schema_errors)
-
+                        if version is not None:
+                            errs_with_version = self.add_to_versioned_errs(
+                                # covered by test_validate_feature_versioned_localized
```
brennie, line 1602: Unnecessary. That should be evident by the fact that the coverage passes.

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1571,19 +1590,28 @@ def _validate_feature_value_against_schema(
...
+                                [*schema_errors, err_msg], # pragma: no cover
```
brennie, line 1603: If this is actually covered, you wouldn't need the pragma.

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1571,19 +1590,28 @@ def _validate_feature_value_against_schema(
...
+                                errs_with_version,
+                            )
+                        else:
+                            errors.append(err_msg)
+                            errors.extend(schema_errors)
+
+        if len(errs_with_version.keys()) > 0:
+            formatted_errs = self._format_grouped_errors(errs_with_version)
+            for e in formatted_errs:
+                errors.append(e)
```
brennie, line 1614: You can use [`extend()`](https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types) here:

```suggestion
            errors.extend(self._format_grouped_errors(errs_with_version)
```

However, I think we should return `dict[str, list[NimbusFeatureversion]]` from both the FML and Schema validators, and then we can call the format function in a single place, e.g.


```python
def _validate_feature_configs(self, data):
    # ...
    for feature_value_data in data.get("reference_branch", {}).get("feature_values", []):
        if application == NimbusExperiment.Application.DESKTOP:
            if schema_errors := self._validate_feature_value_against_schema(...):
                reference_branch_errors.append({"value": self._format_grouped_errors(schema_errors) })
        elif NimbusExperiment.Application.is_mobile(application): # NB <- This condition is wrong now! we aren't validating cirrus with fml!
            if fml_errors := self._validate_with_fml(...):
                reference_branch_errors.append({"value": self._format_grouped_errors(fml_errors) })     
```

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1596,7 +1624,7 @@ def _validate_with_fml(
         blob: str,
         min_version: packaging.version.Version,
         max_version: Optional[packaging.version.Version],
-    ):
+    ) -> dict[str, list[NimbusFeatureVersion]]:
```
brennie, line 1627: This return type is wrong because we return `errors` not `errs_with_version`.

experimenter/experimenter/experiments/models.py
```diff
@@ -1451,12 +1452,41 @@ class Meta:
...
         unique_together = ("major", "minor", "patch")
 
+    @classmethod
+    def parse(cls, s: str) -> "NimbusFeatureVersion":
+        VERSION_RE = re.compile(
+            r"^(?P<major>\d+)(?:\.(?P<minor>\d+)(?:\.(?P<patch>\d+))?)?"
+        )
+        if m := VERSION_RE.match(s):
+            return cls.from_match(m.groupdict())
+        return "0.0.0"
```
brennie, line 1462: This is unused. Remove it.

experimenter/experimenter/experiments/models.py
```diff
@@ -1451,12 +1452,41 @@ class Meta:
...
+        return (self.major, self.minor, self.patch)
+
     def __repr__(self):  # pragma: no cover
         return f"<NimbusFeatureVersion({self.major}, {self.minor}, {self.patch})>"
 
     def __str__(self):  # pragma: no cover
         return f"{self.major}.{self.minor}.{self.patch}"
 
+    def __hash__(self) -> int:
+        return hash(self.as_tuple())
```
brennie, line 1476: You shouldn't need this, I'm pretty sure django base models implement hash

experimenter/experimenter/experiments/models.py
```diff
@@ -1451,12 +1452,41 @@ class Meta:
...
+        return self.as_tuple() < other.as_tuple()
+
+    def __le__(self, other: "NimbusFeatureVersion") -> bool:
+        return self.as_tuple() <= other.as_tuple()
+
+    def __gt__(self, other: "NimbusFeatureVersion") -> bool:
+        return self.as_tuple() > other.as_tuple()
+
+    def __ge__(self, other: "NimbusFeatureVersion") -> bool:
+        return self.as_tuple() >= other.as_tuple()
```
brennie, line 1488: You technically only need to implement `__lt__` to sort in Python.

eliserichards:
Closing this to make these edits :D

experimenter/experimenter/experiments/api/v5/serializers.py
eliserichards: That's what I was thinking too (re: moving these into the top level "if desktop, format errors/ if not, format errors")

PR #10070, opened 2024-01-18
feat(cirrus): Set glean max event buffer value

### LINKED ISSUE #9534: Update glean maxEvents value on the cirrus side
When setting up glean sdk on cirrus side, we set the maxEvents to 500, that means it waits on the client side to gather all 500 events and once it gets 500 then it sends the batch of events, we want to change the value to 1, so that we can see the immediate results. Lets set maxEvents value to 1

[https://mozilla.github.io/glean/book/reference/general/initializing.html#configuration](https://mozilla.github.io/glean/book/reference/general/initializing.html#configuration|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3907)

yashikakhurana:
Because

- When setting up glean sdk on cirrus side, we set the maxEvents to 500, which means it waits on the client side to gather all 500 events, and once it gets 500 then it sends the batch of events. Implementing applications should be able to adjust the value that works best for them.

This commit

- Add the option to set no. of events to store in buffer through `.env`, by default it uses `10` now

Fixes #9534

### DIFF
--- a/cirrus/server/cirrus/settings.py
+++ b/cirrus/server/cirrus/settings.py
@@ -30,7 +30,9 @@ instance_name: str = cast(
     str, config("CIRRUS_INSTANCE_NAME", default="instance name not defined")
 )
 env_name = cast(str, config("CIRRUS_ENV_NAME", default="production"))
-glean_max_events_buffer = config("CIRRUS_GLEAN_MAX_EVENTS_BUFFER", default=10, cast=int)
+glean_max_events_buffer: int = int(
+    config("CIRRUS_GLEAN_MAX_EVENTS_BUFFER", default=10)  # type: ignore
+)
 
 
 @dataclass


.env.sample
```diff
@@ -58,3 +58,4 @@ CIRRUS_FML_PATH=./feature_manifest/sample.yml
 CIRRUS_SENTRY_DSN=
 CIRRUS_INSTANCE_NAME=
 CIRRUS_ENV_NAME=test_instance_stage
+CIRRUS_GLEAN_MAX_EVENTS_BUFFER=20
```
eliserichards, line 61: If we're setting it to 10 for prod, do we also want to set it to 10 here? (or 1, like test and integration-tests?)

cirrus/server/cirrus/settings.py
```diff
@@ -39,7 +40,7 @@ class MetricsConfiguration:
     channel: str = channel
     data_dir: str = "/var/glean"
     log_level: Union[str, int] = logging.getLevelName("WARNING")
-    max_events_buffer: int = 500
+    max_events_buffer: int = int(glean_max_events_buffer)
```
brennie, line 43: On line 33 you do `cast(int, ...)` so this should already be an int, no?

.env.sample
yashikakhurana: oh its just a random value, to show that we can adjust the values, we are not using this in our tests

cirrus/server/cirrus/settings.py
yashikakhurana: just fixed it

jaredlockhart (approved):
LGTM thnx @yashikakhurana !

PR #10083, opened 2024-01-20
Indicate that results are not available for rollouts

### LINKED ISSUE #9766: Indicate that results are not available for rollouts
Presently we disable the results button for rollouts but the help text says `Experiment analysis not ready yet.`  We should change that text to something like `Analysis results are not available for rollouts.` for rollouts so it's clear that no results will ever appear.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4037)

nox1134:
Because

-  The results button for rollouts is disabled while the help text incorrectly states "Experiment analysis not ready yet."

This commit

- Changes the help text to "Analysis results are not available for rollouts." clarifying that no results will ever appear.

Fixes #9766


experimenter/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
```diff
@@ -235,7 +235,7 @@ export const AppLayoutSidebarLaunched = ({
                     "Experiment analysis was skipped"
                   ) : (
                     <>
-                      Experiment analysis not ready yet.
+                      Analysis results are not available for rollouts.
```
jaredlockhart, line 238: Oh you'll want to actually check if it's a rollout here.  It's an experiment then `Experiment analysis not ready yet` is correct, but if we land here and it's a rollout then the new message is correct.

jaredlockhart (changes requested):
Thanks for taking this on @nox1134 !  You'll want to add a check whether to show the rollout message, but otherwise this should be good!

experimenter/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.test.tsx
```diff
@@ -105,7 +105,9 @@ describe("AppLayoutSidebarLaunched", () => {
         `${BASE_PATH}/my-special-slug`,
       );
 
-      screen.getByText("Experiment analysis not ready yet.", { exact: false });
+      screen.getByText("Analysis results are not available for rollouts.", {
+        exact: false,
+      });
```
jaredlockhart, line 110: And then you'd want two versions of this test, one for the rollout case, one for the experiment case.

jaredlockhart (approved):
Oh fantastic, thank you for wrapping this up @nox1134 🎉 🎉 🎉 🎉  Tested locally, works perfectly 🙏

nox1134:
Thank you @jaredlockhart for your help!

PR #10102, opened 2024-01-22
feat(nimbus): Ignore specific versions in manifesttool

### LINKED ISSUE #10100: Ignore specific versions in release discovery
It is possible for a race to occur when someone pushes a branch to focus-ios withour first changing the version in that branch from the main version (v9000). We are actively ignoring main as v9000 because it breaks our Last N version logic in manifesttool. We need a way to ignore specific versions in discovery so we can avoid this issue.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4251)

brennie:
Because

- we ran into a case where focus-ios reported v9000 for a release branch; and
- we don't want to add a v9000 for focus-ios because it is used a placeholder version

This commit:

- adds support for a "ignored_versions" field on tagged discovery strategies, which will omit any matching releases from fetches.

Fixes #10100

### DIFF
--- a/experimenter/manifesttool/releases.py
+++ b/experimenter/manifesttool/releases.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from manifesttool import github_api, hgmo_api
 from manifesttool.appconfig import (
     AppConfig,
@@ -12,6 +14,20 @@ from manifesttool.version import (
     find_versioned_refs,
     resolve_ref_versions,
 )
+
+
+def _filter_ignored_versions(
+    app_name: str,
+    versions: dict[Version, Ref],
+    ignored_versions: Optional[list[Version]],
+):
+    if ignored_versions:
+        for ignored_version in ignored_versions:
+            try:
+                ref = versions.pop(ignored_version)
+                print(f"fetch: {app_name}: Ignored version {ignored_version} at {ref}")
+            except KeyError:
+                pass
 
 
 def discover_tagged_releases(
@@ -30,6 +46,8 @@ def discover_tagged_releases(
         strategy.branch_re,
         strategy.ignored_branches,
     )
+
+    _filter_ignored_versions(app_name, versions, strategy.ignored_versions)
 
     # Limit to the last 5 major versions.
     #
@@ -53,6 +71,8 @@ def discover_tagged_releases(
         tags = github_api.get_tags(app_config.repo.name)
         tag_versions = find_versioned_refs(tags, strategy.tag_re, strategy.ignored_tags)
         tag_versions = filter_versioned_refs(tag_versions, min_version)
+
+        _filter_ignored_versions(app_name, tag_versions, strategy.ignored_versions)
 
         versions.update(tag_versions)
 
--- a/experimenter/manifesttool/version.py
+++ b/experimenter/manifesttool/version.py
@@ -1,13 +1,17 @@
+from __future__ import annotations
+
 import plistlib
 import re
 from dataclasses import dataclass
-from typing import Iterable, Optional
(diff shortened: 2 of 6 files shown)


experimenter/manifesttool/version.py
```diff
@@ -1,14 +1,18 @@
...
+from typing import Any, Iterable, Optional, TYPE_CHECKING
 
 from requests import HTTPError
 
 from manifesttool import github_api, hgmo_api
-from manifesttool.appconfig import AppConfig, RepositoryType, VersionFile, VersionFileType
 from manifesttool.repository import Ref
 
+if TYPE_CHECKING:
+    from manifesttool.appconfig import AppConfig, VersionFile
```
eliserichards, line 14: Why not just import these normally? I'm not a fan of conditional imports 🤔

experimenter/manifesttool/version.py
```diff
@@ -145,6 +167,8 @@ def parse_version_file(f: VersionFile, contents: str) -> Optional[Version]:
     Returns:
         The parsed Version.
     """
+    from manifesttool.appconfig import VersionFileType
```
eliserichards, line 170: Same here

experimenter/manifesttool/version.py
```diff
@@ -175,6 +199,8 @@ def resolve_ref_versions(
     Returns:
         A mapping of Versions to the Refs.
     """
+    from manifesttool.appconfig import RepositoryType
```
eliserichards, line 202: Same here

experimenter/manifesttool/releases.py
```diff
@@ -14,6 +16,20 @@
...
+def _filter_ignored_versions(
+    app_name: str,
+    versions: dict[Version, Ref],
+    ignored_versions: Optional[list[Version]],
+):
+    if ignored_versions:
+        for ignored_version in ignored_versions:
+            try:
+                ref = versions.pop(ignored_version)
+                print(f"fetch: {app_name}: Ignored version {ignored_version} at {ref}")
```
yashikakhurana, line 28: oh you might want to remove this print

yashikakhurana (commented):
also, does it make sense to refer to ignore versions somewhere in constants?

experimenter/manifesttool/version.py
jaredlockhart: Usually something like this is to avoid circular imports, see here:

https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles

experimenter/manifesttool/version.py
jaredlockhart: This is also a common pattern for avoiding circular imports

jaredlockhart (approved):
Yep all looks right to me, nice simple change, tests are clear 👍

experimenter/manifesttool/releases.py
brennie: That print is intentional

brennie:
> also, does it make sense to refer to ignore versions somewhere in constants?

I'm not sure where we'd do that or why we'd put it in a constants file. Its defined in the config file, which cant reference any outside files. We also don't really need this information anywhere else, so at this point I'm fine sticking this just in the config file.

experimenter/manifesttool/version.py
brennie: Yep, the import doesnt actually happen until the function executes, which lets us have the app config module import this module for the version definitions. Good question :)

experimenter/manifesttool/version.py
eliserichards: That makes sense 👍

PR #10082, opened 2024-01-19
feat(nimbus): select required/excluded branches in audience page

jaredlockhart:
Because

* We can now require/exclude specific branches of experiments/rollouts

This commit

* Adds entries in the require/exclude drop downs for specific branches of each experiment/rollout

fixes #10029

image

PR #10113, opened 2024-01-22
feat(nimbus): filter require/exclude options from each others fields

### LINKED ISSUE #10090: Remove required/excluded experiments/rollouts from other field
It doesn’t make sense to require a branch but exclude another branch from the same experiment/rollout, so if a branch is selected to be required/excluded, we should remove all other branches from the other field for the same experiment/rollout.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4243)

jaredlockhart:
Because
    
* If someone wants to require/exclude an experiment by branch, they can not also do the opposite with the remaining branches since that will already be handled by the branch allocator
    
This commit
    
* Hides all branches of any experiment that is selected to be required/excluded from the opposite field
    
fixes #10090

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -464,66 +464,75 @@ describe("FormAudience", () => {
       });
     });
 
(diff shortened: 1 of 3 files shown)


jaredlockhart:
Blocked on https://github.com/mozilla/experimenter/pull/10082

brennie (approved):

PR #10117, opened 2024-01-23
feat(nimbus): only show one require/exclude option if only one branch

jaredlockhart:
Because

* We don't need to show the 'All branches' option if there's only a single branch in an experiment/rollout

This commit

* Only shows a single entry in require/exclude for experiments/rollouts with one branch

fixes #10082


jeddai (approved):
This looks good!

PR #10123, opened 2024-01-23
feat(nimbus): update proposed duration

### LINKED ISSUE #10119: Rollout should not accept enrollment days greater than experiment duration
We have seen that if set enrollment days lets say 10 days and experiment duration 5 days, and then mark it as a Rollout, it will not pop the error that `the enrollment duration must be less than or equal to the experiment duration`, it will pop out once the user will try to increase rollout percentage. for example https://experimenter.services.mozilla.com/nimbus/backgroundupdate-enable-unelevated-installations-rollout-3-release/audience 
image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4255)

yashikakhurana:
Because

- Users were able to set the proposed duration to greater than the proposed enrollment days, previously the enrollment period validation wasn't occurring for rollouts because the field was disabled and that is now fixed so there are experiments with invalid pairs of enrollment/duration that should not have been launchable but are live now.

This commit

- Migrate the invalid ones in the DB (by making sure if the proposed duration is less than the proposed enrollment, in that case, sets proposed duration days equal to the proposed enrollment and ensures this same invalid configuration can't be launched.

Fixes #10119


eliserichards (approved):
Perfect! 🥇

jaredlockhart (approved):
Yep fantastic thnx @yashikakhurana 🎉 🎉 🎉

experimenter/experimenter/experiments/api/v5/mutations.py
```diff
@@ -70,6 +70,9 @@ def mutate(cls, root, info, input: ExperimentInput):  # noqa: A002
         if "feature_config_ids" in input:
             input["feature_configs"] = input.pop("feature_config_ids", None)
 
+        if "subscribed" in input:
+            input["subscribed"] = input.pop("subscribed")
+
```
brennie, line 75: This shouldn't be here?

brennie (changes requested):

experimenter/experimenter/experiments/api/v5/mutations.py
yashikakhurana: Oops sorry my bad

brennie (approved):

PR #10129, opened 2024-01-24
feat(nimbus): add published_date field to schema

### LINKED ISSUE #9786: Add published_date field to experiment/rollout remote settings objects

jaredlockhart:
Because

* We need a way to track the published date of a recipe so that clients can use it as a strict ordering for evaluating recipes
* We want to set it when recipes are published to remote settings
* It should always be the most recent publish date for launching experiments/rollouts, but not overwritten after launch

This commit

* Adds a publish date field to NimbusExperiment
* Adds it to the V6 serializer
* Sets it in the kinto tasks

fixes #9786


experimenter/experimenter/experiments/tests/test_migrations.py
```diff
@@ -1,8 +1,5 @@
 from django_test_migrations.contrib.unittest_case import MigratorTestCase
 
-from experimenter.experiments.constants import NimbusConstants
```
brennie, line 3: This doesn't seem right

jaredlockhart: Oh yeah it's safe to empty out the test migration file, sometimes they'll fail when new migrations come in, something about the way the migration test framework is setup, but the migration tests are only necessary for the pr that introduces the data migration and test.  After they land it's safe to empty this file out until it's needed again.

jaredlockhart: Oh I see what happened here, we can't use the imported model in the migration tests since it can be out of sync with the state of the model at the time the test is testing (that's awkward to say).  We always have to use the shimmed models provided by the tests.  That's why it was failing.  This should keep the test intact though what I said above is also true, after a migration lands it's no longer necessary to keep its data migration test intact.

jaredlockhart: cc @yashika-khurana @eliserichards @mikewilli for visibility

brennie: sorry what I meant was it doesnt seem right that most of the file was gone, but its back now so

jaredlockhart: Yeah that's what I was explaining above, if some PR introduces a new migration, sometimes that'll break a past migration test (in this case it was just because it was using the wrong instance of NimbusExperiment, but I've seen it be more obtuse in the past), and it's okay to empty that file out in those cases.

PR #10134, opened 2024-01-24
chore(nimbus): remove old required/excluded api fields

### LINKED ISSUE #10047: Remove old required/exclude V5 API fields
After we migrate Nimbus UI to use the new required/excluded fields, we can safely remove the old ones.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4209)

jaredlockhart:
Because

* We've now migrated the UI over to use the new required/excluded fields
* It is now safe to remove the old required/excluded fields

This commit

* Removes the old required/excluded fields from the V5 API

fixes #10047


yashikakhurana (approved):
Bye to the old code piece 🎉

PR #10140, opened 2024-01-25
feat(nimbus): Fetch specific apps with --app flag

### LINKED ISSUE #10061: Add an --app flag to manifesttool
It would be nice to add a flag to manifesttool to only fetch a specific app or apps, instead of editing the config file each time you want to test app-specific changes.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4218)

brennie:
Because:

- often in manual testing we only want to fetch a single app;
- the previous way we tested this was commenting out entire apps in the apps.yaml file; and
- we ended up accidentally committing the commented out apps in a previous PR

This commit:

- adds a `--app` option to manifest-tool to fetch only the specified apps.

Fixes #10061

### DIFF
--- a/experimenter/manifesttool/cli.py
+++ b/experimenter/manifesttool/cli.py
@@ -45,13 +45,30 @@ def main(ctx: click.Context, *, manifest_dir: Path):
 @click.option(
     "--summary", "summary_filename", type=Path, help="Write a summary to this file."
 )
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Oh interesting, good idea 👍

PR #10142, opened 2024-01-25
feat(nimbus): unset published date on timeout/reject

### LINKED ISSUE #10138: Unset published date on timeout/rejection
If something is publishing to RS for the first time and gets either rejected or times out, then we should unset the publish_date so we don’t wind up with things in the database that were never published but indicate that they have a published date.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4254)

jaredlockhart:
Because

* We now set the published_date when an experiment/rollout is published to kinto
* If a launching experiment/rollout either times out or is rejected, we should unset the published_date so we don't wind up with experiments/rollouts in the database that haven't launched but do have a published_date

This commit

* Sets published_date to None on either reject or timeout

fixes #10138


yashikakhurana (approved):
perfect 💯

PR #10144, opened 2024-01-25
bug(nimbus): handle risk types properly in nimbus ui

### LINKED ISSUE #10139: Missing attestations on overview page incorrectly validated as green but changes fail to save
When attempting to save the overview page without checking one of the attestation radio buttons ("If the public, users or press,..."), the page refuses to save with a generic "Sorry, an error occurred while submitting. Please try again." error message and the radials are highlighted in green. This is a confusing UX. 

Expected Behavior: 
The error message should indicate the failure condition (that one of the attestation radio buttons is required to be selected but isn't selected) and the radio buttons should be highlighted in red to show the location of the failure. 

Screenshot 2024-01-25 at 9 31 44 AM

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4285)

jaredlockhart:
Because

* If graphql receives the wrong types for any input field it will refuse to submit the data
* The frontend types do not effectively use the ExperimentInput type to ensure the types are correct
* This can result in attempting to save a page with valid inputs but receiving a graphql error

This commit

* Updates the FormOverview to use the correct ExperimentInput type
* Updates the risk fields to convert from string/null/undefined into a boolean correctly
* Updates/adds tests

fixes #10139


brennie (approved):
Love to see a `Record<string, any>` go away.

PR #10148, opened 2024-01-26
feat(nimbus): Upgrade sdk integration tests to modern application-services

### LINKED ISSUE #9820: Build application services for integration tests
When we first started using the Nimbus SDK in Application Services as part of the integration tests, we setup a periodic Circle task to build it and push it to Dockerhub as an image we could then depend on in the tests.  However this allows breaking changes in application services to work their way into the integration tests without making any code changes in the Experimenter repo.  Now that Experimenter more closely depends on the Nimbus SDK, we have set it up to checkout and build Application Services at a locked commit as part of its build phase so that code changes in Application Services and Experimenter can remain in sync.  Let's do the same thing for the integration tests and explicitly checkout and build Application Services as part of starting the integration tests so that we can explicitly lock it to a commit and keep Application Services and the tests in sync.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4072)

brennie:
Because

- The sdk integration test image was stuck on application-services
  v97.5.1 due to build system changes, which prevented landing any new
  targeting using new targeting attributes from v98 onwards.
- The current system of periodically pushing a integration test image
  and pulling that pre-built image in tests lead to headaches and
  issues.
- I discovered several oddities in the current
  integration-test-nimbus-rust setup (specifying a webdriver driver,
  sudo commands that never ran because sudo was not installed).
- The new application services has a very different NimbusClient
  interface from the one currently used in tests.

This commit

- Adds a `experimenter:megazords` container, which is responsible for
  downloading a tagged nightly of the Cirrus and Nimbus-Experimenter
  megazords.
- Adds a new `experimenter:integration-tests` container, which takes the
  place of the old "nimbus-rust-image" container. This container is very
  barebones, and only includes the application services megazord (copied
  from the other container) and tox.
- Removes the old `Dockerfile-rust-image` and its publishing steps from
  CI.
- Configures our integration tests to build the image on the fly instead
  of pulling a pre-built image from a registry.
- Cleans up a few things in the `test_mobile_targeting.py` suite and
  updates the code to match the new interfaces.
- Updates the `integration_test_nimbus_rust` task to run tox with `-vv`,
  which includes the output of any commands issued.

Fixes #9820


Makefile
```diff
@@ -102,9 +102,15 @@ update_kinto:  ## Update latest Kinto/Remote Settings container
...
 	$(COMPOSE) build
 
+build_megazords:
+	$(DOCKER_BUILD) -f application-services/Dockerfile -t experimenter:megazords application-services/
+
 build_dev: ssl
 	$(DOCKER_BUILD) --target dev -f experimenter/Dockerfile -t experimenter:dev experimenter/
 
+build_integration_test: ssl build_megazords
+	$(DOCKER_BUILD) -f experimenter/tests/integration/Dockerfile -t experimenter:integration-tests experimenter/
```
eliserichards, line 112: 👏

application-services/application-services.env
```diff
@@ -0,0 +1 @@
+APPLICATION_SERVICES_BUILD_ID=124.20240126050308
```
eliserichards, line 1: Yes yes yes this is so much easier than tracking it down in two different files 🙌

eliserichards (changes requested):
This looks great 👏 I think you're missing removing the old AS hash from the Cirrus dockerfile

brennie:
> This looks great 👏 I think you're missing removing the old AS hash from the Cirrus dockerfile

This does not affect the cirrus build at all. It only affects the integration tests

eliserichards (approved):
Chatted with @brennie and the cirrus stuff will be done in another patch 👍 🚀 r+

yashikakhurana (commented):
great job @brennie 🎉 , just a reminder we will need to update Readme as well, specifically for integration tests one

yashikakhurana (approved):
Awesome, thanks @brennie for adding in the docs as well 👏

experimenter/tests/integration/nimbus/test_mobile_targeting.py
```diff
@@ -8,7 +8,24 @@
...
+        pass
+
+    def record_feature_activation(self, *args, **kwargs):
+        pass
+
+    def record_feature_exposure(self, *args, **kwargs):
+        pass
+
+    def record_malformed_feature_config(self, *args, **kwargs):
+        pass
```
jrbenny35, line 28: This is neat

jrbenny35:
I'm late, but nice job. Good seeing these tests are still quite useful

PR #10173, opened 2024-01-31
bug(cirrus): Fix cirrus builds

### LINKED ISSUE #10172: Cirrus Builds are busted
When #10151 landed it subtly broke the build for Cirrus in two ways:

1. The `cirrus_build_test` task did not depend on `build_megazords`, so `cirrus_check` et al. were failing.
2. rustc was mistakenly removed from the python-builder step. As it turns out, it is not required on x86_64 but is required on arm64. This is due to some packages not being on pypi for the correct architecture.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4300)

brennie:
Because:

- the cirrus test jobs require the megazords; and
- the cirrus build requires rustc on arm64

This commit:

- adds `build_megazords` as a dependency of `cirrus_build_test`; and
- installs `rustc` inside the `python-builder` phase of the Cirrus build.

Fixes #10172

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ CIRRUS_GENERATE_DOCS = python cirrus/generate_docs.py
 cirrus_build: build_megazords
 	$(CIRRUS_ENABLE) $(DOCKER_BUILD) --target deploy -f cirrus/server/Dockerfile -t cirrus:deploy cirrus/server/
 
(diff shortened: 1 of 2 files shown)


eliserichards (approved):
Works for me 👍

PR #10184, opened 2024-01-31
chore(nimbus): remove unused fields from get all experiments query

### LINKED ISSUE #10040: Improve getAllExperiments query
And in theory as part of that fix we could do a more exhaustive check on every field that's included in the `getAllExperiments` query and remove all the fields that aren't being used.  Theoretically typescript should help us here by just commenting out all the fields on [the auto generated type](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/nimbus-ui/src/types/getAllExperiments.ts#L43-L77) and then running and rerunning the typechecker until it passes with the minimum set of fields necessary for that page to work.  There may be some fields used that typescript won't catch so you'll probably also have to manually verify on your local that it's intact, but paring out as many of those fields as we can should be an easy way to speed up the initial load.

_Originally posted by @jaredlockhart in https://github.com/mozilla/experimenter/issues/10032#issuecomment-1887783611_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4202)

jaredlockhart:
Because

* Every field that's queried in gql increases the time of the call
* There are fields in the getAllExperiments query that are not used by the list page

This commit

* Removes all unused fields from the getAllExperiments query

fixes #10040


yashikakhurana (approved):
💯

eliserichards (approved):
💯 🧽 🧹

PR #10190, opened 2024-01-31
feat(nimbus): add backoff to polling

### LINKED ISSUE #10186: Backoff summary page refreshes
Presently, we refresh the data on the summary page every 30s.  This is handy right after you launch so it updates to the current status after Remote Settings updates.  But after that nothing is going to change, and if someone leaves a tab open it'll sit there refreshing every 30s and hitting the backend with potentially slow calls.  We could implement a backoff so the timer doubles in length after each refresh up to a max of maybe 1h.  That should alleviate some of the pressure on the servers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4311)

jaredlockhart:
Because

* We have a polling timer on the summary page that auto refreshes periodically
* Setting this to a constant 30s will add unnecessary load to the servers
* Immediately after launching it is handy for the page to update quickly to show its new live status
* After that there should not be any significant changes so we can safely back off that timer to 1h

This commit

* Adds a backoff to the polling timer logic that doubles from 30s up to 1h

fixes #10186


brennie (approved):

yashikakhurana (approved):
perfect 💯  hope it reduces noise on slack now

PR #10191, opened 2024-02-01
refactor(nimbus): Reusable warning on PageSummary

### LINKED ISSUE #10096: Add warning to Summary page
Add a warning to the Summary page that can be reused for all of our existing warnings.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4249)

eliserichards:
Because

- We're going to need some easily reusable warnings on the summary page

This commit

- Creates a new warning `AlertWarning` (bad name) and replaces the three existing warnings with it
- Gives it variants, nice testids, etc

Fixes #10096

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
@@ -471,9 +471,7 @@ describe("PageSummary", () => {
     mutationMock.result.errors = [new Error("Boo")];
     render(<Subject mocks={[mock, mutationMock]} />);
     await launchFromDraftToReview();
-    await waitFor(() =>
-      expect(screen.getByTestId("submit-error")).toBeInTheDocument(),
-    );
+    await waitFor(() => screen.getByTestId("submit-error-warning"));
   });
 
   // TODO: #6802
@@ -912,7 +910,7 @@ describe("PageSummary", () => {
       targetingConfigSlug: "OH_NO",
     });
     render(<Subject mocks={[mock]} />);
-    expect(screen.queryByTestId("bucketing-warning")).toBeInTheDocument();
+    screen.getByTestId("bucketing-warning");
   });
 
   it("displays no duplicate rollout warning for experiments", async () => {
@@ -934,7 +932,7 @@ describe("PageSummary", () => {
       targetingConfigSlug: "OH_NO",
     });
     render(<Subject mocks={[mock]} />);
-    expect(screen.queryByTestId("bucketing-warning")).not.toBeInTheDocument();
+    expect(screen.queryByTestId("bucketing-warning")).toBeNull();
   });
 
   it("displays a warning for desktop rollouts under v114", async () => {
@@ -954,11 +952,7 @@ describe("PageSummary", () => {
       targetingConfigSlug: "OH_NO",
     });
     render(<Subject mocks={[mockRollout]} />);
-    await waitFor(() =>
-      expect(
-        screen.queryByTestId("desktop-min-version-warning"),
-      ).toBeInTheDocument(),
-    );
+    await waitFor(() => screen.queryByTestId("desktop-min-version-warning"));
   });
 
   it("displays no warning for desktop rollouts above v113", async () => {
@@ -976,9 +970,7 @@ describe("PageSummary", () => {
       targetingConfigSlug: "OH_NO",
     });
     render(<Subject mocks={[mockRollout]} />);
-    expect(
-      screen.queryByTestId("desktop-min-version-warning"),
-    ).not.toBeInTheDocument();
+    expect(screen.queryByTestId("desktop-min-version-warning")).toBeNull();
   });
 
   it("displays no min version warning for non-desktop rollouts", async () => {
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
```diff
@@ -373,3 +381,27 @@ const StatusPill = ({
...
+}: {
+  text: string | SerializerMessage;
+  testId: string;
+  learnMoreLink?: string;
+  learnMoreText?: string;
+  variant?: string;
+}) => (
+  <Alert data-testid={`warning-${testId}`} variant={variant}>
+    {text}
+    {learnMoreLink && (
```
brennie, line 400: Include a space so the learn more link is separated from the error text. We don't do this currently, but we should.

```suggestion
    {text}
    {" "}
    {learnMoreLink && (
```

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
```diff
@@ -373,3 +381,27 @@ const StatusPill = ({
     {label}
   </Badge>
 );
+
+const AlertWarning = ({
```
brennie, line 385: These warnings are all about launching the experiment, so can we call this `LaunchWarning` ? or maybe just `Warning` ?

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -956,7 +956,7 @@ describe("PageSummary", () => {
     render(<Subject mocks={[mockRollout]} />);
     await waitFor(() =>
       expect(
-        screen.queryByTestId("desktop-min-version-warning"),
+        screen.queryByTestId("warning-desktop-min-version"),
       ).toBeInTheDocument(),
```
brennie, line 960: nit: replace the `expect(...).toBeInTheDocument()` with `screen.getByTestId(...)`

`getByTestId` will throw if it does not find the selector.

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -912,7 +912,7 @@ describe("PageSummary", () => {
       targetingConfigSlug: "OH_NO",
     });
     render(<Subject mocks={[mock]} />);
-    expect(screen.queryByTestId("bucketing-warning")).toBeInTheDocument();
+    expect(screen.queryByTestId("warning-bucketing")).toBeInTheDocument();
```
brennie, line 915: nit: replace the `expect(...).toBeInTheDocument()` with `screen.getByTestId(...)`

`getByTestId` will throw if it does not find the selector.

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -472,7 +472,7 @@ describe("PageSummary", () => {
     render(<Subject mocks={[mock, mutationMock]} />);
     await launchFromDraftToReview();
     await waitFor(() =>
-      expect(screen.getByTestId("submit-error")).toBeInTheDocument(),
+      expect(screen.getByTestId("warning-submit-error")).toBeInTheDocument(),
```
brennie, line 475: nit: replace the `expect(...).toBeInTheDocument()` with `screen.getByTestId(...)`

`getByTestId` will throw if it does not find the selector.

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
```diff
@@ -977,7 +977,7 @@ describe("PageSummary", () => {
     });
     render(<Subject mocks={[mockRollout]} />);
     expect(
-      screen.queryByTestId("desktop-min-version-warning"),
+      screen.queryByTestId("warning-desktop-min-version"),
     ).not.toBeInTheDocument();
```
brennie, line 981: Semantically, this should be `expect(...).toBeNull()`. `expect(node).not.toBeInTheDocument()` should be used if `node` is an actual DOM element. ie, you would use this with an element you grabbed earlier to check if it got removed.

querySelector et al return `null` when they dont find an element, so expecting `null` not to be in the document is semantically wrong because it never will be. We should expect the method to not find anything instead.

brennie (approved):

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
eliserichards: done! image

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.test.tsx
eliserichards: We should definitely file some cleanup tickets, because `expect(...).toBeInTheDocument()` is all over in our tests 😖 I'll fix these for now since this is a refactor PR, but it would be great to give them all some love

experimenter/experimenter/nimbus-ui/src/components/PageSummary/index.tsx
eliserichards: `Warning` sounds better to me 👍

PR #10201, opened 2024-02-01
feat(nimbus): Add subscribe button to TableOverview

### LINKED ISSUE #10091: Add UI on Summary for adding subscriber
This should be an email address field, most likely with a valid mozilla email address

* Check what we allow for the owner field

Open questions:

* Where do we want this subscriber UI to go? We want it on the Summary page, but should it be on the sidebar? Up at the top of the page? 
* Should it be a popup, or inline “Edit/Save” like Takeaways/QA sections?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4245)

eliserichards:
Because …
- We want a button to add/remove a subscriber 

This commit…
- Adds the current user to the config so that it can be evaluated on the frontend
- Adds a button with “+Subscribe”/”- Unsubscribe” to `TableOverview` on the Summary page
- The button triggers the update mutation and passes the user’s email address along with the updated subscription status
	- For example, if a user is already subscribed to the experiment, they will see their email address on the list of subscribers and the “- Unsubscribe” button
	- When the user clicks the button, the following update will be sent to unsubscribe them from the experiment:
```
…
subscribers = {
   [
      “email”: user,
      “subscribed”: false,
   ]
}
```
- Successful subscriptions/unsubscriptions trigger a page refresh so you can see the latest list with your email address added/removed

Fixes #10091

----
Screenshot 2024-02-01 at 2 59 17 PM
Screenshot 2024-01-31 at 6 39 13 PM
Screenshot 2024-01-31 at 6 39 32 PM

https://github.com/mozilla/experimenter/assets/43795363/3c2ce5d8-bca4-4a3c-a41a-2a9bf7c74389


yashikakhurana (approved):
Looks perfect, tested locally 🎉 . Was able to subscribe, we will be able to test it on stage when real users will add or remove. My only suggestion to move this `subscriber` option on left side of the panel or on top of `summary page`, but @eliserichards did mentioned shell wanted here, but I will try to convince @shell1  and on follow up we can move button :D

PR #10210, opened 2024-02-03
For #10094 feat(nimbus): Check if there are other live multifeature experiments for targeted features

### LINKED ISSUE #10094: Calculate when live multi-feature experiments exist for this feature
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4247)

eliserichards:
Because

- Audience overlap can happen if there are other live multifeature experiments for features being targeted by your experiment

This commit

- Adds a query to the model to check for other live multifeature experiments for your features
- This query returns a list of the slugs of any matching experiments, so that we will be able to say "these are the experiments that might cause overlap with yours" once we connect this to the frontend

Fixes #10094

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/types.py
+++ b/experimenter/experimenter/experiments/api/v5/types.py
@@ -494,6 +494,9 @@ class NimbusExperimentType(DjangoObjectType):
         lambda: graphene.List(graphene.NonNull(NimbusExperimentBranchThroughExcludedType))
     )
     feature_configs = DjangoListField(NimbusFeatureConfigType)
+    feature_has_live_multifeature_experiments = graphene.NonNull(
+        lambda: graphene.List(graphene.NonNull(graphene.String))
+    )
     firefox_max_version = NimbusExperimentFirefoxVersionEnum()
     firefox_min_version = NimbusExperimentFirefoxVersionEnum()
     hypothesis = graphene.String()
@@ -571,6 +574,7 @@ class NimbusExperimentType(DjangoObjectType):
             "documentation_links",
             "enrollment_end_date",
             "feature_configs",
+            "feature_has_live_multifeature_experiments",
             "firefox_max_version",
             "firefox_min_version",
             "hypothesis",
@@ -754,3 +758,6 @@ class NimbusExperimentType(DjangoObjectType):
 
     def resolve_subscribers(self, info):
         return self.subscribers.all().order_by("username")
+
+    def resolve_feature_has_live_multifeature_experiments(self, info):
+        return self.feature_has_live_multifeature_experiments
--- a/experimenter/experimenter/experiments/api/v5/types.py
+++ b/experimenter/experimenter/experiments/api/v5/types.py
@@ -758,6 +758,3 @@ class NimbusExperimentType(DjangoObjectType):
 
     def resolve_subscribers(self, info):
         return self.subscribers.all().order_by("username")
-
-    def resolve_feature_has_live_multifeature_experiments(self, info):
-        return self.feature_has_live_multifeature_experiments
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -17,7 +17,7 @@ from django.core.files.base import ContentFile
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import MaxValueValidator
 from django.db import models
-from django.db.models import F, Q, QuerySet
+from django.db.models import Count, F, Q, QuerySet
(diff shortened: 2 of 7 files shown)


experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -754,3 +758,6 @@ def resolve_required_experiments_branches(self, info):
 
     def resolve_subscribers(self, info):
         return self.subscribers.all().order_by("username")
+
+    def resolve_feature_has_live_multifeature_experiments(self, info):
+        return self.feature_has_live_multifeature_experiments
```
jaredlockhart, line 763: I think if the field name matches a field right on the model it'll do this automatically?

experimenter/experimenter/experiments/models.py
```diff
@@ -798,6 +798,29 @@ def allocate_bucket_range(self):
...
+        live_experiments = NimbusExperiment.objects.filter(
+            status=self.Status.LIVE,
+            application=self.application,
+        )
+        if len(live_experiments.all()) >= 0:
+            for feature in self.feature_configs.all().values_list("slug", flat=True):
+                experiments = (
+                    live_experiments.annotate(n_feature_configs=Count("feature_configs"))
+                    .filter(n_feature_configs__gt=1)
+                    .filter(feature_configs__slug=feature)
```
jaredlockhart, line 818: Oh you can get rid of the loop if you do

```py
feature_slugs = self.feature_configs.all().values_list("slug", flat=True)
...
    .filter(feature_configs__slug__in=feature_slugs)
```

experimenter/experimenter/experiments/models.py
```diff
@@ -798,6 +798,29 @@ def allocate_bucket_range(self):
...
+        matching = set()
+        live_experiments = NimbusExperiment.objects.filter(
+            status=self.Status.LIVE,
+            application=self.application,
+        )
+        if len(live_experiments.all()) >= 0:
+            for feature in self.feature_configs.all().values_list("slug", flat=True):
+                experiments = (
+                    live_experiments.annotate(n_feature_configs=Count("feature_configs"))
+                    .filter(n_feature_configs__gt=1)
```
jaredlockhart, line 817: Excellent, this is exactly how to express this 🙏 🎉

experimenter/experimenter/experiments/models.py
```diff
@@ -798,6 +798,29 @@ def allocate_bucket_range(self):
...
+                    .filter(feature_configs__slug=feature)
+                    .all()
+                )
+                matching.update([e.slug for e in experiments])
```
jaredlockhart, line 821: And then you can just do the same thing you did above to pull the slugs out and then you don't need the set either 

```py
.values_list("slug", flat=True).distinct()
```

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2539,6 +2539,99 @@ def test_get_changelogs(self):
             ],
         )
 
+    @parameterized.expand(
+        [
+            (NimbusExperiment.Application.FENIX, NimbusExperiment.Application.FENIX, 3),
+            (NimbusExperiment.Application.FENIX, NimbusExperiment.Application.DESKTOP, 0),
+        ]
+    )
+    def test_get_live_multifeature_experiments_for_feature(
```
jaredlockhart, line 2548: So the method is defined on the model, so I would put the exhaustive tests about all the logical combinations right on the model and then you only need a single integration test on the API.

jaredlockhart (changes requested):
Yes looking great, very clear method, very clear tests, simple field on the API all looking great, couple ideas but looking great 👏 👏 👏 👏

experimenter/experimenter/experiments/models.py
jaredlockhart: Oh actually and prolly wanna throw a `.order_by('slug')` in here too so the order doesn't jump around in the UI, since it's doing a join and the order is just 🤷‍♀️ 🤷‍♀️ 🤷‍♀️

experimenter/experimenter/experiments/models.py
eliserichards: `__slug__in` -- that's exactly what I was looking for!

experimenter/experimenter/experiments/tests/test_models.py
jaredlockhart: Oh wow I completely misread this, yes this is already exactly correct.  Nevermind me 🤣

jaredlockhart (approved):
Beautiful, great job 🎉 🎉 🎉 🎉 🎉

PR #10215, opened 2024-02-05
feat(nimbus): Add model query for getting a list of experiments that are live and excluded

### LINKED ISSUE #10095: Calculate when an experiment excludes other live deliveries
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4248)

eliserichards:
Because

- Audience overlap can happen when there are Live experiments that you are explicitly excluding from your experiment

This commit

- Adds a query to the model to find any of these matching experiments
- This query returns a list of the slugs of any matching experiments, so that we will be able to say "these are the experiments that might cause overlap with yours" once we connect this to the frontend

Fixes #10095

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -824,11 +824,6 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def excluded_live_deliveries(self):
-        """Live experiments that are excluded by the current experiment.
-
-        Returns:
-            A list of experiment slugs.
-        """
         matching = []
         if self.excluded_experiments.exists():
             matching = (
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -823,6 +823,26 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         return matching
 
     @property
+    def excluded_live_deliveries(self):
+        """Live experiments that are excluded by the current experiment.
+
+        Returns:
+            A list of experiment slugs.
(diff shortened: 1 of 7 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -822,6 +822,26 @@ def feature_has_live_multifeature_experiments(self):
...
 
+    @property
+    def excluded_live_deliveries(self):
+        """Live experiments that are excluded by the current experiment.
+
+        Returns:
+            A list of experiment slugs.
+        """
+        matching = []
+        if self.excluded_experiments.exists():
```
yashikakhurana, line 833: we don't include the docs, may you can add a line comment, as in other properties we don't define input and returns

yashikakhurana (dismissed):
looks good to me, just a minor suggestion

experimenter/experimenter/experiments/models.py
eliserichards: Sounds good! The doc for [live multifeature](https://github.com/mozilla/experimenter/pull/10210) matches already landed, but I will take it out of this one since we don't maintain those 👍

yashikakhurana (approved):
perfect, thanks @eliserichards 🚢

experimenter/experimenter/experiments/models.py
brennie: Just because we don't have a lot of in-code documentation doesn't mean we shouldn't add it for new code.

If you put this information in docstrings, then IDE integrations will be able to show you it when you hover over a method. If you put that information in comments, that information is only visible if you go to look at the actual method source, which makes it less discoverable and less useful overall.

PR #10233, opened 2024-02-06
feat(nimbus): Add a warning for rollouts that will re-enroll and clobber prefs

### LINKED ISSUE #10137: Warn that pref-setting rollouts may re-enroll and override user prefs
Normally if a user sets a pref controlled by a nimbus experiment or rollout, they will be unenrolled from the experiment or rollout. However, rollouts automatically re-enroll if they were not opted out of (via about:studies, pref flips dont count). If a rollout does not use the "Prevent enrollment if users have changed any prefs set by this experiment" targeting option, then it will always re-enroll when a user changes a pref controlled by the experiment, which will result in the rollout continuously overriding the user’s prefs. We should add a warning about this case to make this behaviour clear.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4283)

### LINKED ISSUE #10232: Warn about setPref and rollout interactions
Rollouts with setPref features can result in unexpected constant re-enrollment and clobbering of user prefs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4345)

brennie:
Because:

- rollouts always re-enroll when possible, unless the user opts out;
- unenrollments from changing prefs are not marked as opt-outs; and
- this can result in constant re-enrollment of rollouts that overwrite user pref choices

This commit:

- adds a warning when this situtation could occur; and
- also improves the performance of feature value validation by refactoring the
  validation logic to cache the `VersionedSchemaRange`s per feature ID, saving
  O(F*B) queries (where F is is the number of features and B is the number of
  treatment branches)

Fixes #10137

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -7,7 +7,6 @@ from dataclasses import dataclass
 from typing import Any, Optional
 
 import jsonschema
-import packaging
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.cache import cache
@@ -1475,6 +1474,10 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         super().__init__(*args, **kwargs)
         self.warnings = defaultdict(list)
 
+        self.schemas_by_feature_id: dict[
+            str, NimbusFeatureConfig.VersionedSchemaRange
+        ] = {}
+
     def to_representation(self, instance):
         data = super().to_representation(instance)
         data["feature_config"] = None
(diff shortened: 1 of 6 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1548,39 +1551,10 @@ def _validate_feature_value(
         application: str,
         feature_config: NimbusFeatureConfig,
         value: str,
-        min_version: packaging.version.Version,
-        max_version: Optional[packaging.version.Version],
+        schemas_in_range: NimbusFeatureConfig.VersionedSchemaRange,
```
eliserichards, line 1554: 👏

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1548,39 +1551,10 @@ def _validate_feature_value(
...
-        feature_config: NimbusFeatureConfig,
-        value: str,
-        localizations: Optional[dict[str, Any]],
-        min_version: packaging.version.Version,
-        max_version: Optional[packaging.version.Version],
-    ):
-        schemas_in_range = feature_config.get_versioned_schema_range(
-            min_version, max_version
-        )
+    ) -> list[str]:
```
eliserichards, line 1557: YES removing the optional 🙌

experimenter/experimenter/nimbus-ui/src/lib/constants.ts
```diff
@@ -50,6 +50,8 @@ export const EXTERNAL_URLS = {
     "https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#PrefFlipandAddOnExperiments-Isthisstudypartnerrelated?riskPARTNER",
   RISK_REVENUE:
     "https://mana.mozilla.org/wiki/display/FIREFOX/Pref-Flip+and+Add-On+Experiments#PrefFlipandAddOnExperiments-riskREV",
+  ROLLOUT_SETPREF_REENROLL_EXPLANATION:
+    "https://experimenter.info/rollouts/setpref-reenroll-warning",
```
eliserichards, line 54:
```suggestion
  ROLLOUT_SETPREF_REENROLL_EXPLANATION:
    "https://experimenter.info/faq/rollouts/rollout-setpref-reenroll",
```

eliserichards (approved):
This looks great! The URL needs to be updated, but otherwise this is good to go

experimenter/experimenter/nimbus-ui/src/lib/constants.ts
brennie: nice catch

PR #10234, opened 2024-02-06
feat(nimbus): Record variable -> pref mapping for setPref features

### LINKED ISSUE #10075: Record the names of setPref variables in NimbusVersionedSchema
NimbusVersionedSchema has the list of prefs the feature sets, but does not track the actual variables that set those prefs. We need that information so we can properly warn/error on the case of very large values that would be written as prefs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4234)

brennie:
Because:

- we currently don't keep a mapping of variables to the prefs they set in the database; and
- we need this information in order to provide warnings about large pref writes

This commit:

- adds a `set_pref_vars` field to the `NimbusVersionedSchema` model; and
- updates the `load_feature_configs` management command to record this information.

Fixes #10075


yashikakhurana (approved):
perfect :shipit:

experimenter/experimenter/features/management/commands/load_feature_configs.py
```diff
@@ -207,3 +214,10 @@ def set_pref_name(v: Union[SetPref, str]) -> str:
         NimbusVersionedSchema.objects.bulk_create(schemas_to_create)
 
         logger.info("Features Updated")
+
+
+def _set_pref_name(v: Union[SetPref, str]) -> str:
```
eliserichards, line 219: 🤌

experimenter/experimenter/features/tests/test_load_feature_configs.py
```diff
@@ -75,12 +75,29 @@ def test_loads_new_feature_configs(self):
...
 
+        self.assertEqual(
+            schema.set_pref_vars,
+            {
+                "string": "nimbus.test.string",
+                "int": "nimbus.test.int",
+                "boolean": "nimbus.test.boolean",
+            },
+        )
+
```
eliserichards, line 86: Is there any way this can fail on us if we somehow pass in the wrong things to `set_pref_vars`? (I see we're checking if it's `None` in `load_feature_configs.py`)

eliserichards (approved):
r+ with one (1) question ☝️

experimenter/experimenter/features/tests/test_load_feature_configs.py
brennie: If its not JSON serializable it should throw and fail to import features.

The thing we're checking for none is here:

```python
sets_prefs = [
    _set_pref_name(v.set_pref)
    for v in feature.model.variables.values()
    if v.set_pref is not None  # <--
] 
```

Here if `v.set_pref` is None, it just means there is no `setPref:` entry for the variable, e.g.


```yaml
fooFeature:
  variables:
    bar:
      type: string
      setPref:
        branch: user
        pref: 'foo.bar.baz'
    qux:
      type: int
```

Here `bar` will have `set_pref = SetPref(branch="user", pref="foo.bar.baz")` and `qux` will have `set_pref = None`

PR #10236, opened 2024-02-06
feat(nimbus): Use set_pref_vars in place of sets_prefs

### LINKED ISSUE #10076: Switch from NimbusVersionedSchema.set_prefs to NimbusVersionedSchema.set_pref_vars
Once we add the new set_pref_vars field in #10075 we can switch all the uses of the old set_prefs field to the new field.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4235)

brennie:
Because:

- we now have the `set_pref_vars` field, which includes both the prefs and the variables that set them

This commit:

- migrates all logic to use the new field.

Fixes #10076

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1857,7 +1857,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
     def _validate_desktop_pref_rollouts(self, data):
         if self.instance.is_rollout:
             any_feature_sets_prefs = any(
-                schema.sets_prefs
+                bool(schema.set_pref_vars)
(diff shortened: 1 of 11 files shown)


brennie:
Blocked on #10233, #10234

jaredlockhart (commented):
Looks good, we don't have to do a data migration to populate the field?

brennie:
> Looks good, we don't have to do a data migration to populate the field?

No because we are generating the field from the feature config loading. If there is feature that has been removed from the manifest, we wont have the prefs but also we cant launch any experiments for it. For all the versioned manifests, we will have perfect information.

jaredlockhart (approved):
Yep all looks good to me thnx @brennie 👍

PR #10258, opened 2024-02-08
chore(nimbus): remove v5 config serializer

### LINKED ISSUE #10257: Delete V5 configuration serializer
At one point we intended to replace GraphQL with Django Rest Framework and so we started reimplementing the GQL API using DRF, starting with the configuration serializer.  However we've since abandoned this strategy and so the V5 config serializer can be safely deleted.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4357)

jaredlockhart:
Because

* We had intended to replace graphql with django rest framework
* We started by reimplementing the config endpoint in rest framework
* We abandoned the rest framework strategy to focus instead on htmx
* The v5 config serializer still exists but was never hooked up to anything
* It can be safely removed

This commit

* Removes the v5 config serializer

fixes #10257


brennie (approved):
🔥 🔥 🔥

PR #10286, opened 2024-02-15
feat(nimbus): add v3 ingestion for pairwise results

### LINKED ISSUE #10188: restructure analysis results for simpler UI
The results after ingestion from Jetstream should be in a slightly different format to work better with typescript in the UI. Instead of dynamic branch keys at the same level as static first/all keys, we should nest the dynamic keys in a static field.

The resulting object should look something like this:
```
interface AnalysisPointData {
  first: FormattedAnalysisPoint;
  all: FormattedAnalysisPoint[];
  comparison_data?: {
    [comparison_branch: string]: AnalysisPointData | undefined;
  }
}
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4313)

mikewilli:
Because

- we need to change the output schema of results ingestion for the UI
- these changes are not backwards compatible

This commit

- adds the new output schema under a `v3` key on the main `results_data` object
- updates tests to reflect the `v2` and `v3` schemas living side-by-side

Fixes #10188


experimenter/experimenter/jetstream/tests/constants.py
```diff
@@ -22,12 +22,12 @@
 
 class JetstreamTestData:
     @classmethod
-    def get_absolute_metric_data(cls, DATA_POINT):
-        return MetricData(
+    def get_absolute_metric_data(cls, DATA_POINT, is_v2=False):
```
brennie, line 25: I would prefer something like

```
class WhateverVersion(Enum): #IngestionVersion,MetricVersion,???
   V2 = ...
   V3 = ...

# ...

def get_absolute_metric_data(cls, DATA_POINT, version: WhateverVersion):
   ...
```

experimenter/experimenter/jetstream/client.py
```diff
@@ -203,9 +204,17 @@ def get_experiment_data(experiment: NimbusExperiment):
         "show_analysis": settings.FEATURE_ANALYSIS,
         "metadata": experiment_metadata,
     }
+    experiment_data_old = experiment_data = {
+        # DAILY included for backwards compatibility with UI
+        # TODO: remove DAILY after updating UI
```
brennie, line 209: Is there a bug filed for this? If so put it here.

experimenter/experimenter/jetstream/models.py
```diff
@@ -201,6 +201,164 @@ def compute_significance(data_point: DataPoint) -> Significance:
...
+                    else jetstream_data_point.comparison
+                )
+                data_point = DataPoint(
+                    lower=jetstream_data_point.lower,
+                    upper=jetstream_data_point.upper,
+                    point=jetstream_data_point.point,
+                )
+
+                # Set window_index (always 1 for OVERALL). Used by weekly DataPoint
+                # objects and for storing significance for each window.
```
brennie, line 234:
```suggestion
                # Used by weekly DataPoint
                # objects and for storing significance for each window.
```

The comment doesnt need to say "Setting X to Y", we can see that in the code.

experimenter/experimenter/jetstream/models.py
```diff
@@ -201,6 +201,164 @@ def compute_significance(data_point: DataPoint) -> Significance:
...
+                                if p.window_index == window_index
+                            ]
+                        )
+                        == 0
+                    ):
+                        pairwise_comparison_data.all.append(data_point)
+
+    def append_conversion_count(self, primary_metrics_set):
+        for branch_name in self.dict():
+            branch = getattr(self, branch_name)
```
brennie, line 328:
```suggestion
        for branch_name, branch in self.dict().items():
```

Is this the preferred way to iterate over fields in Pydantic? It seems expensive to construct a dict from the object, potentially recursively, just to get field names.

It looks like there might be a `__fields__` member that you can iterate over (self.__fields__.keys()) to get the field names

mikewilli: I agree that it's less than ideal, or at least looks odd, but there is a reason for this construction -- we want `branch` to be a `Branch` object (and `branch_data` to be a `BranchData`), and calling `self.dict()` converts the entire model to dictionaries. Doing `getattr(self, branch_name)` for each of the keys leave us with the `Branch` model instead of a dict.

That said, we could get the keys more efficiently, as you mentioned. I'll update to iterate `self.__fields__` instead of `self.dict()` (there's also a `__fields_set__`, which only includes fields set at instantiation, but given the custom constructor and dynamic model, I think it's safer to use `__fields__` in this case).

experimenter/experimenter/jetstream/client.py
```diff
@@ -197,15 +199,24 @@ def get_experiment_data(experiment: NimbusExperiment):
...
+    #     - https://github.com/mozilla/experimenter/issues/9861
+    experiment_data_old = {
         AnalysisWindow.DAILY: {},
         "show_analysis": settings.FEATURE_ANALYSIS,
         "metadata": experiment_metadata,
     }
 
+    # import ipdb
+
+    # ipdb.set_trace()
```
jaredlockhart, line 215: 💀

experimenter/experimenter/jetstream/models.py
```diff
@@ -201,6 +201,128 @@ def compute_significance(data_point: DataPoint) -> Significance:
...
+
+                branch = jetstream_data_point.branch
+                branch_obj = getattr(self, branch)
+                branch_obj.is_control = experiment.reference_branch.slug == branch
+
+                group_obj = getattr(
+                    branch_obj.branch_data, METRIC_GROUP.get(metric, Group.OTHER)
+                )
+
+                if metric == Metric.USER_COUNT and statistic == Statistic.PERCENT:
```
jaredlockhart, line 250: Not that you need to change it here but in general, if you need to do multiple comparisons like this you can also write it as

```py
(metric, statistic) == (Metric.USER_COUNT, Statistic.PERCENT)
```

and that makes it easier to write for more than 2 comparisons

jaredlockhart (approved):
So it's an interesting pattern here to try to support both flows simultaneously, ie having `_old` `_new` fields, calling out TODOs where it'll be safe to remove them later, adding flags about whether to do the old or new version.  We're gonna wind up having to do this every so often so we should pay attention to how we want to manage transitions like this, ie maybe instead of `_old` `_new` `is_old_version` we can use some identifier like `_v{N}` or something that we can grep for later.  Something to think about.

Anyways nothing jumps out at me as particularly crazy here (other than the sheer volume of test data we need).

Thanks @mikewilli stoked to see this PR finally land 🎉 🎉 🎉 🎉

experimenter/experimenter/jetstream/client.py
mikewilli: :skull:

experimenter/experimenter/jetstream/models.py
mikewilli: Is that the preferred way to do multiple comparisons in python? It seems less intuitive to me, though not terribly so.

jaredlockhart: Like using tuples/lists for comparisons is very common ie I would write

```py
if x == A or x == B or ....:
```

as 

```py
if x in [A, B, ...]:
```

or the example above of using tuples  to do multiple comparisons, or using sets, so yeah being able to use collection literals in comparisons is very Pythonic.

PR #10304, opened 2024-02-22
feat(experimenter): When cancel the review, change status next to nul…

### LINKED ISSUE #10303: Not able to update audience
User wanted to [change this rollout](https://experimenter.services.mozilla.com/nimbus/backgroundupdate-enable-unelevated-installations-rollout-3-release) from 50% to 80%, but clicked on end rollout before changing the percentage under "population". User saved the page, noticed that mistake, and cancelled the ending of the experiment. With that, the "Request Update" button is disabled even after the user changes and save other values under "popluation".

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4377)

yashikakhurana:
Because

- Accidentally user clicked on `end rollout` and then canceled the review, and then tried to change the population size from 50% to 80%, although the population was getting updated, they are not able to request an update- https://experimenter.services.mozilla.com/nimbus/backgroundupdate-enable-unelevated-installations-rollout-3-release 

This commit

- The problem is with the flow when we cancel the review, when we request for review we change value of `statusNext` from `null` to `Live`, but when we cancel the review, we keep `statusNext` as Live
- Adds the condition so that if the user cancels the review for rollout, should go back to same state as it was before i.e. `statusNext=null`

Fixes #10303


experimenter/experimenter/nimbus-ui/src/components/Summary/index.tsx
```diff
@@ -72,7 +72,9 @@ const Summary = ({ experiment, refetch }: SummaryProps) => {
       changelogMessage: CHANGELOG_MESSAGES.CANCEL_REVIEW,
       statusNext:
         experiment.status === NimbusExperimentStatusEnum.LIVE
-          ? NimbusExperimentStatusEnum.LIVE
+          ? experiment.isRollout
+            ? null
+            : NimbusExperimentStatusEnum.LIVE
           : null,
```
jaredlockhart, line 78: Shouldn't `statusNext` always go to `null` when canceling a review, regardless of whether it's a rollout or an experiment?

image
image

jaredlockhart (changes requested):

experimenter/experimenter/nimbus-ui/src/components/Summary/index.tsx
yashikakhurana: okay actually yes if its rollout/experiment is going from draft, we launch and we cancel it then it should go back to statusNext = null, but lets say if rollout/experiment is live, then it should go to statusNext as live, but setting it to live for rollout, doesn't let us make changes 🤔

yashikakhurana:
okay, I found the other place, actually serializer was not marking up rollout dirty, frontend logic is correct, so modified the update method and tested that locally 💪

jaredlockhart (changes requested):
I still don't think that's right, `statusNext` should be `null/None` when a live update request is canceled?

image

jaredlockhart (approved):
Awesome ty @yashikakhurana 🙏

PR #10308, opened 2024-02-22
feat(nimbus): add my experiments tab

### LINKED ISSUE #10259: Add My Experiments checkbox
Related to #10250

Add a checkbox up at the top of the filters that shows "My experiments", and filters by experiments that the current User owns AND experiments that the current user is subscribed to.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4358)

jaredlockhart:
Because

* Now that users can subscribe to experiments it can be a useful way to tag experiments you want to quickly get back to
* We can add a section to the summary page for experiments that the current user either owns or is subscribed to

This commit

* Adds a tab to the directory table for experiments that the current user either owns or is subscribed to

fixes #10259
image

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageHome/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageHome/index.test.tsx
@@ -49,6 +49,7 @@ describe("PageHome", () => {
       ["drafts", screen.getByText("Draft (3)")],
       ["live", screen.getByText("Live (3)")],
       ["archived", screen.getByText("Archived (1)")],
+      ["owned", screen.getByText("My Experiments (1)")],
     ] as const;
 
   const findSearchTabs = () =>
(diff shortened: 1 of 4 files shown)


jrbenny35 (approved):

experimenter/experimenter/nimbus-ui/src/components/PageHome/index.test.tsx
```diff
@@ -49,6 +49,7 @@ describe("PageHome", () => {
       ["drafts", screen.getByText("Draft (3)")],
       ["live", screen.getByText("Live (3)")],
       ["archived", screen.getByText("Archived (1)")],
+      ["owned", screen.getByText("My Experiments (1)")],
```
mikewilli, line 52: Is it potentially confusing to call this "owned" when it's owned and subscribed?

jaredlockhart: It isssssss but trying to use more words there or different words there didn't feel any better, but I'm open to suggestions.

mikewilli: Yea fair enough, and obviously not a big deal. I was trying to think of a suggestion when I asked but didn't come up with anything obvious. Would just "subscribed" be better if we want one word? I think it's more implicit that "subscribed" includes "owned" but not necessarily the other way around.

PR #10337, opened 2024-02-29
feat(nimbus): add warnings and errors for large pref writes

### LINKED ISSUE #10340: Add errors and warnings about potential large pref writes
We recently had a case where an experiment launched with a 25kb JSON payload, which caused a 25kb pref to be written. We really want to avoid doing these things, especially if the payload reaches the hard limit of prefs, which would cause the nimbus desktop client to break.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4198)

brennie:
Because

- it is possible to cause large pref writes on desktop;
- prefs over 4kB will cause warnings on desktop; and
- prefs over 1mB will cause errors on desktop and cause client errors

This commit

- adds dismissable warnings at the 4kB-1mB boundary; and
- adds undimissable errors at the 1mB+ boundary.

Fixes #10340

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -7,7 +7,7 @@ import re
 import typing
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Any, NotRequired, Optional, TypedDict
+from typing import Any, NotRequired, Optional, Self, TypedDict
 
 import jsonschema
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


brennie:
I'll replace the ticket with a GH issue when the data sync bot is working again.

jaredlockhart (approved):
Tested locally, works good 👍  It was a lot of refactors to get to erroring/warning on len > max.  I wonder if there's somethign we can do to make that less gnarly.  But looks good works good thnx @brennie 🎉

PR #10384, opened 2024-03-08

mikewilli:
Because

- jetstream computes pairwise branch comparison data
- we are ingesting pairwise data into a new structure stored under `v3` in each experiment's `results_data` field

This commit

- adds a data migration to add a `v3` pairwise structure under experiments that have `results_data` but not `v3`


jaredlockhart (approved):
I have to basically trust that the tests are rigorous enough here to cover things.  It is good that we're versioning the data so we'll still have the v2 data if we need to roll back the ui changes or regenerate the v3 data, but we should find a better pattern for changes like this becuase this is too gnarly to really review effectively.  But let's just keep this moving forward for now, thnx @mikewilli 🙏

PR #10404, opened 2024-03-13
Revert "feat(nimbus): add data migration for v3 pairwise branch comparison data"

mikewilli:
Reverts mozilla/experimenter#10384

There was an error during execution of the data migration, so in order to unblock deploys this PR reverts the data migration.


jaredlockhart (approved):
👍

PR #10405, opened 2024-03-13

mikewilli:
Because

- #7821 changed the schema of analysis results parsed from Jetstream
- old experiments were not updated for this schema
- AND the last PR (#7846) with this migration was missing an important check which caused it to fail

This commit

- adds a data migration to update old experiments to the new schema
- AND updates the migration to ensure the type assumptions are correct
- AND updates the migration test to cover this other case


jaredlockhart (approved):
Second times the charm! 🍀

PR #10409, opened 2024-03-14

mikewilli:
Because

- #7821 changed the schema of analysis results parsed from Jetstream

This commit

- adds a data migration to update old experiments to the new schema


experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
```diff
@@ -0,0 +1,181 @@
...
+                                        },
+                                        "significance": {
+                                            // optional daily/weekly/overall moved from here into [comparison_branch]
+                                            [comparison_branch]: { // this level is new
+                                                "weekly": {},
+                                                "overall: {}
+                                            }
+                                        },
+                                        "percent": 0.0
+            """
```
brennie, line 51: Multiline comments should use `#`

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
```diff
@@ -0,0 +1,181 @@
...
+
+            comparison_default = {"first": {}, "all": []}
+            significance_default = {"weekly": {}, "overall": []}
+
+            for key, value in data["v3"].items():
+                if value is not None and key in ["weekly", "overall"]:
+                    for basis, basis_data in value.items():
```
brennie, line 58: Might I suggest inverting this condition and adding `continue` so we don't nest so deeply?

```suggestion
                if value is None or key in ["weekly", "overall"]:
                    continue
                    
                for basis, basis_data in value.items():
```

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
```diff
@@ -0,0 +1,181 @@
...
+                                                "overall: {}
+                                            }
+                                        },
+                                        "percent": 0.0
+            """
+
+            comparison_default = {"first": {}, "all": []}
+            significance_default = {"weekly": {}, "overall": []}
+
+            for key, value in data["v3"].items():
```
brennie, line 56: Please pull some of this out into helper functions. The nesting makes it unreadable.

experimenter/experimenter/experiments/tests/test_migrations.py
```diff
@@ -19,24 +20,296 @@ def prepare(self):
...
+
+        prechange_data = NimbusExperimentOld.objects.get(
+            slug="test-experiment"
+        ).results_data
+        changed_data = NimbusExperiment.objects.get(slug="test-experiment").results_data
+        self.assertDictEqual(changed_data["v2"], prechange_data["v2"])
+        changes = changed_data["v3"]["weekly"]["enrollments"]["all"]
+
+        changes_control = changes["control"]["branch_data"]["other_metrics"]["identity"]
+        self.assertTrue("percent" in changes_control)
```
brennie, line 211: Here and elsewhere

```suggestion
        self.assertIn("percent", changes_control)
```

experimenter/experimenter/experiments/tests/test_migrations.py
```diff
@@ -19,24 +20,296 @@ def prepare(self):
...
+
+        self.assertTrue("control" in changes_control["difference"])
+        self.assertTrue("treatment" in changes_control["difference"])
         self.assertEqual(
-            experiment.proposed_duration, 20
-        )  # Expecting the proposed_duration to be updated
+            ["control", "treatment"], list(changes_control["difference"].keys())
+        )
+        self.assertFalse("all" in changes_control["difference"])
```
brennie, line 218: Here and elsewhere

```suggestion
        self.assertNotIn("all", changes_control["difference"])
```

experimenter/experimenter/experiments/tests/test_migrations.py
```diff
@@ -19,24 +20,296 @@ def prepare(self):
...
+        self.assertTrue("control" in nonchanges["significance"])
+        self.assertTrue("treatment" in nonchanges["significance"])
+        self.assertFalse("weekly" in nonchanges["significance"])
+        self.assertFalse("overall" in nonchanges["significance"])
+        self.assertTrue("1" in nonchanges["significance"]["treatment"]["weekly"])
+
+        prechange_data = NimbusExperimentOld.objects.get(
+            slug="another-experiment"
+        ).results_data
+        self.assertDictEqual(unchanged_data, prechange_data)
```
brennie, line 307: Just use assertEqual. It will dispatch to assertDictEqual.

brennie (changes requested):
Check out https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug for all the `assert` methods that Testcase offers

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
mikewilli: I too hate this function. I started trying to do this recursively, but the dynamic nature of all of the nested keys made it extremely difficult to follow, and I was unsure if it would even work. @jaredlockhart's original suggestion was to not spend a lot of time on it because a migration only runs once. In light of the errors merging this the past few times I can take another pass at reorganizing it, but I'm not sure the best way to go about it.

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
```diff
@@ -0,0 +1,165 @@
...
+                                            )
+                                            if comparison in [
+                                                "difference",
+                                                "relative_uplift",
+                                            ]:
+                                                data["v3"][key][basis][segment][
+                                                    cur_branch
+                                                ]["branch_data"][metrics_group][metric][
+                                                    comparison
+                                                ] = {}
```
brennie, line 94: We do this deeply nested indexing repeatedly. Lets pull some of these out into vars so its more readable.

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
brennie: Pulling out the contents of this loop into a function would be a good start.

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
```diff
@@ -0,0 +1,165 @@
...
+
+def update_results_data_schema(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+
+    for experiment in NimbusExperiment.objects.all():
+        branches = [b.slug for b in experiment.branches.all()]
+        reference_branch = experiment.reference_branch
+        data = experiment.results_data
+        if data is not None and "v3" not in data and "v2" in data:
+            data["v3"] = data["v2"]
```
brennie, line 15: We can do the not-continue thing here too to de-nest one level.

```suggestion
        if data is None or "v3" in data or "v2" not in data:
            continue
            
        data["v3"] = data["v2"]
```

brennie (changes requested):

experimenter/experimenter/experiments/migrations/0266_nimbusexperiment_results_pairwise_data.py
jaredlockhart: Well the flip side of this is, we tested this locally against the prod data in its current form and validated the output.  Applying a bunch of structural refactoring now could potentially introduce new bugs.  This code only runs once, we don't need to worry about the maintainability of it.  Yes the structure is unreadable and difficult to reason about, but I'd rather focus on finding ways to improve this pattern in general for future code/migrations than spend time trying to apply them here.

brennie (approved):
alright, yolo

PR #10412, opened 2024-03-15
feat(nimbus): Collapsible locales, languages and countries

### LINKED ISSUE #10312: Optionally collapse the regions and languages field in the summary page
We recently had an experiment that targeted 248/249 regions, which resulted in an extremely long summary page. We should conditionally show a collapse/show button when there is say, more than 10 countries in the list.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4379)

yashikakhurana:
Because

- Sometimes locales, languages, and countries lists are so extensive that we have to scroll a page. For example https://experimenter.services.mozilla.com/nimbus/private-window-visual-refresh/summary

This commit

- Collapsible locales, languages and countries option if user have selected more than 10

Fixes #10312 

Example 
- Less than 10
Screenshot 2024-03-15 at 9 29 34 AM

- More than 10
Screenshot 2024-03-15 at 9 32 24 AM
Screenshot 2024-03-15 at 9 32 34 AM

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
@@ -33,6 +33,9 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
     MOBILE_APPLICATIONS.includes(experiment.application);
 
   const [expand, setExpand] = useState(false);
+  const [expandLocale, setExpandLocale] = useState(false);
+  const [expandCountry, setExpandCountry] = useState(false);
+  const [expandLanguage, setExpandLanguage] = useState(false);
 
   return (
     <Card className="border-left-0 border-right-0 border-bottom-0">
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
```diff
@@ -80,11 +83,67 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
...
+                            </Accordion.Collapse>
+                          </Accordion>
+                        </>
+                      ) : (
+                        <ul className="list-unstyled mb-0">
+                          {experiment.locales.map((l) => (
+                            <li key={l.id}>{l.name}</li>
+                          ))}
+                        </ul>
+                      )
```
mikewilli, line 146: Could this be its own component since it's now a non-trivial amount of code that is basically copied three times in this file?

yashikakhurana: good point, refactored it

PR #10417, opened 2024-03-15
bug(nimbus): fix broken django import

### LINKED ISSUE #10415: Fix django experiment import
The django import no longer works. It was working sometime in the last month or two, so something recent must have broken it -- it looks like there was a new foreign key field added that needs to be handled explicitly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4416)

mikewilli:
Because

- the django experiment import stopped working
- the error is in a function that is not necessary for imports

This commit

- fixes the error in the import flow by bypassing the unnecessary field export logic
- also upgrades django-import-export to the latest version
- also sets the import/export allowed types to just JSON (the other ones didn't work in my testing)

Fixes #10415


mikewilli:
Any objections to excluding this code from the coverage report? The idea is that this should be a temporary fix until the new version of the library is out (which will have the fix built-in).

PR #10424, opened 2024-03-18
feat(nimbus): cache v6 api

### LINKED ISSUE #10423: Improve V6 API performance
We now have more frequent callers for the V6 API, and it's now taking ~15s per call, so let's see what we can do to bring that down to something more reasonable.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4418)

jaredlockhart:
Because

* There are now many callers for the V6 API
* It can be very slow when fetching and serializing all experiments

This commit

* Caches the V6 API for 60 minutes

fixes #10423


yashikakhurana (approved):
perferct @jaredlockhart 👏

PR #10439, opened 2024-03-19
feat(nimbus): use v3 results for pairwise data

### LINKED ISSUE #10436: Support v3 results (pairwise data) in UI
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4422)

mikewilli:
Because

- we added a `v3` to experiments' `results_data` containing pairwise branch comparison data

This commit

- switches the v3 API to use this `v3` data
- updates the UI to work with `v3` format of data

Fixes #10436

**Note** that this PR should not change anything visible to users. There will be a follow-up PR containing the UI changes that will allow users to select different branches as reference.

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -914,8 +914,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
     @property
     def has_displayable_results(self):
         # True if self.results_data has weekly or overall results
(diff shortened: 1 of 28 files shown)


jaredlockhart (approved):
Funny how we now have data-v3 and api-v3.  So anyways yeah this now makes the reference branch a parameter so that we can then select it later.  Neat.  Looks good 👍

PR #10449, opened 2024-03-21
bug(nimbus): fix outcome slug lookup to filter by app

### LINKED ISSUE #9505: Validate that outcomes are for the correct application
There was a [recent experiment](https://experimenter.services.mozilla.com/nimbus/ios-felt-privacy/summary) on Firefox for iOS that appears to be using a Fenix outcome. I'm not sure if this is accurate or just a UI issue, but we should investigate and confirm that experiments can't be configured with outcomes that don't match the application of the experiment.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3889)

mikewilli:
Because

- there can be overlaps in outcome slugs across applications
- we saw some experiment summary pages linking to the wrong application's outcome definitions

This commit

- ensures that the outcomes hook filters to the experiment's application

Fixes #9505

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
@@ -18,11 +18,6 @@ export function useOutcomes(experiment: getExperiment_experimentBySlug): {
       return [];
     }
 
-    if (experiment.application === null) {
-      return slugs
-        .map((slug) => outcomes!.find((outcome) => outcome!.slug === slug))
-        .filter((outcome) => outcome != null);
-    }
     return slugs
       .map((slug) =>
         outcomes!.find(
--- a/experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
@@ -24,7 +24,13 @@ export function useOutcomes(experiment: getExperiment_experimentBySlug): {
         .filter((outcome) => outcome != null);
     }
     return slugs
-      .map((slug) => outcomes!.find((outcome) => outcome!.slug === slug && outcome!.application === experiment.application))
+      .map((slug) =>
+        outcomes!.find(
+          (outcome) =>
+            outcome!.slug === slug &&
+            outcome!.application === experiment.application,
+        ),
+      )
       .filter((outcome) => outcome != null);
   };
 
--- a/experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
@@ -18,8 +18,13 @@ export function useOutcomes(experiment: getExperiment_experimentBySlug): {
       return [];
     }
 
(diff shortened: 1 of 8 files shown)


experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
```diff
@@ -18,8 +18,19 @@ export function useOutcomes(experiment: getExperiment_experimentBySlug): {
       return [];
     }
 
+    if (experiment.application === null) {
+      return slugs
+        .map((slug) => outcomes!.find((outcome) => outcome!.slug === slug))
+        .filter((outcome) => outcome != null);
+    }
```
jaredlockhart, line 25: Application can never be null, is this here because typescript forces you to deal with the null case?

brennie: its because the graphql is incorrect

brennie: https://github.com/mozilla/experimenter/blob/a518f6feda664a67c1e2a97a12a59713f371187c/experimenter/experimenter/experiments/api/v5/types.py#L335

this should be `graphene.NonNull(graphene.List(graphene.NonNull(NimbusOutcomeType))` (or something similar)

brennie: Oops I relized i commented about outcome, not application. Thats https://github.com/mozilla/experimenter/blob/a518f6feda664a67c1e2a97a12a59713f371187c/experimenter/experimenter/experiments/api/v5/types.py#L136 and it should be `graphene.NonNull(...)`

mikewilli: Updated these and regenerated types.

experimenter/experimenter/nimbus-ui/src/lib/mocks.tsx
```diff
@@ -1141,8 +1141,19 @@ export function mockOutcomeSets(experiment: getExperiment_experimentBySlug): {
       return [];
     }
 
+    if (experiment.application === null) {
+      return slugs
+        .map((slug) => outcomes!.find((outcome) => outcome!.slug === slug))
+        .filter((outcome) => outcome != null);
+    }
```
jaredlockhart, line 1148: So you should be able to get rid of this case right?

experimenter/experimenter/nimbus-ui/src/hooks/useOutcomes.tsx
jaredlockhart: Oh yeah I wasn't gonna say we should change all the types, probably just a `!` would've been sufficient.

experimenter/experimenter/nimbus-ui/src/lib/mocks.tsx
mikewilli: Yep, good catch! I forgot that this function was copied into the tests.

jaredlockhart (approved):
Much nicer, thnx @mikewilli 🙏

PR #10494, opened 2024-03-28
feat(cirrus): Support non-root user

### LINKED ISSUE #10459: Add non-root user to Cirrus image
In order to run the Cirrus image in a container in the GCP v2 environment, we need to run it as a non-root user due to security policies. Running the image locally, I don't see a non-root user for this purpose. Please add a non-root user to the Cirrus image.

yashikakhurana:
Because

- Fxa can run the cirrus image in a container in the GCP v2 environment as a non-root user only due to security policies.

This commit

- Support cirrus images to run as non-root users too.

Fixes #10459

### DIFF
--- a/cirrus/README.md
+++ b/cirrus/README.md
@@ -49,19 +49,15 @@ By following these steps, you will create the `.env` file and configure the nece
 
 ## Running as Non-Root User
 
-By default, the Cirrus Docker image runs the application as the root user. However, if you prefer to run the application as a non-root user for security reasons, you can build the Docker image with additional parameters.
+By default, the Cirrus Docker image runs the application as cirrus/1000/1000. However, if you prefer to run the application as a different user for security reasons, you can build the Docker image with additional parameters.
 
-To run the Cirrus application as a non-root user, follow these steps:
-
-- Build the Docker image while specifying the desired username, user ID, and group ID for the non-root user. For example:
+- Build the Docker image while specifying the desired username, user ID, and group ID. For example:
 
 ```bash
 docker build --build-arg USERNAME=myuser --build-arg USER_UID=1000 --build-arg USER_GID=1000 -t your_image_name:tag .
 ```
 
 Replace myuser with the desired username and 1000 with the desired user ID and group ID.
-
-- After building the image, you can run the Cirrus container as usual, and it will execute as the non-root user specified during the build process.
 
 ## Commands
 
--- a/cirrus/README.md
+++ b/cirrus/README.md
@@ -47,6 +47,22 @@ To set up the Cirrus environment, follow these steps:
 
 By following these steps, you will create the `.env` file and configure the necessary environment variables for the Cirrus application.
 
+## Running as Non-Root User
+
+By default, the Cirrus Docker image runs the application as the root user. However, if you prefer to run the application as a non-root user for security reasons, you can build the Docker image with additional parameters.
+
+To run the Cirrus application as a non-root user, follow these steps:
+
+- Build the Docker image while specifying the desired username, user ID, and group ID for the non-root user. For example:
+
+```bash
+docker build --build-arg USERNAME=myuser --build-arg USER_UID=1000 --build-arg USER_GID=1000 -t your_image_name:tag .
+```
+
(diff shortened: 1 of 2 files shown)


cirrus/README.md
```diff
@@ -47,6 +47,22 @@ To set up the Cirrus environment, follow these steps:
 
 By following these steps, you will create the `.env` file and configure the necessary environment variables for the Cirrus application.
 
+## Running as Non-Root User
+
+By default, the Cirrus Docker image runs the application as the root user. However, if you prefer to run the application as a non-root user for security reasons, you can build the Docker image with additional parameters.
```
jaredlockhart, line 52: Is this true now?  Like if you don't pass in those parameters, it'll use the new 'myuser/1000/1000' user defined below, so will it ever run as root?

cirrus/server/Dockerfile
```diff
@@ -30,6 +30,14 @@ RUN poetry config virtualenvs.create false && \
 # Second stage: Deploy stage
 FROM python:3.11-slim-bullseye as deploy
 
+# Create a non-root user
+ARG USERNAME=myuser
```
jaredlockhart, line 34: Let's use `cirrus` for the username?

jaredlockhart (dismissed):
Okay cool so it's a pretty simple change, and thank you for updating the docs as well, just a couple questions 🙏

cirrus/server/Dockerfile
```diff
@@ -46,5 +54,11 @@ ENV PYTHONPATH=$PYTHONPATH:/application-services
 # Copy all other files into the container's working directory
 COPY . .
 
+# Change ownership of the working directory to the non-root user
+RUN chown -R $USERNAME:$USERNAME /cirrus
+
+# Switch to non-root user
+USER $USERNAME
+
```
jrbenny35, line 62: Should you switch to the non-root user? This would make this user the default

jrbenny35 (commented):
Had a question

cirrus/README.md
yashikakhurana: I have tried it locally to build, by default the root user uses `1000` as a first user id, without any changes in other places,I mean without providing build arguments it worked, can you double check in your local to make sure? I want to make sure it doesn't impact monitor deployment.

cirrus/server/Dockerfile
yashikakhurana: sorry I am not sure :(, do you mean why we need to switch?

cirrus/README.md
jaredlockhart: So on my local, on main I get

```
[app-sbAM2aaX-py3.11]jaredlockhart@Jareds-MacBook-Pro~/Documents/experimenter(main)$ docker run -ti cirrus:deploy bash
root@3a1326278208:/cirrus# id
uid=0(root) gid=0(root) groups=0(root)
```

and then in your branch i get

```
[app-sbAM2aaX-py3.11]jaredlockhart@Jareds-MacBook-Pro~/Documents/experimenter(10459/add_non_root_user)$ docker run -ti cirrus:deploy bash
cirrus@92a4a09a84dc:/cirrus$ id
uid=1000(cirrus) gid=1000(cirrus) groups=1000(cirrus)
```

so without any args it's using the new user defined in the args, which is different than the default behaviour now.  

The good news here is this is all taking place inside the container so in theory it shouldn't matter to the external consumer, and using non root inside the container is probably a good security precaution, so it should be okay to just default to always using the new non-root user inside the container, I don't think that should break existing deployments, and even if it does it's just a one time fix.

So it's probably fine to leave this as it is and then just sayin the docs 'by default it runs as cirrus/1000/1000 and if you want to override it to something else just pass these args'.

yashikakhurana: awesome, thank you for testing that, yeah if we don't provide it uses 1000 default value,  I don't think it should break the existing deployment, but I will update the docs too, and will keep an eye on monitor logs

cirrus/server/Dockerfile
jrbenny35: Oh I see, I missed the variable name!

PR #10504, opened 2024-04-01
feat(nimbus): make side navbars sticky

### LINKED ISSUE #8016: Scroll main content independent of sidebar when viewing an experiment
This is a small UX improvement, but it would be nice to scroll the main content of the page without scrolling the left-hand sidebar. This would be particularly helpful on the Results page where some sidebar links help to quickly navigate to different metrics.

Of note, the sidebar should also be able to scroll independently, in case the content overflows the window.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2970)

mikewilli:
Because

- some pages are long enough that scrolling down page moves the nav bar out of view
- having navigation controls in view is nice

This commit

- makes the nav bar sticky to the top of the page so it comes with you when you scroll
- makes the nav bar contents scrollable in case the window isn't large enough to show all the content on one screen

Fixes #8016


yashikakhurana (approved):
🚢

PR #10544, opened 2024-04-09
chore(nimbus): move new ui into nimbus_ui_new

### LINKED ISSUE #10542: Move new Nimbus UI packages into nimbus_ui_new folder
Let's move all of the new Nimbus UI packages, ie templates, static assets, views, urls, etc, into a folder called `nimbus_ui_new` so that we can keep everything together.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4464)

jaredlockhart:
Because

* We're starting a new UI for Nimbus using Django Views/Forms/Templates
* Let's collect all the templates, static assets, views, tests in one folder so it's easier to collect and find things

This commit

* Moves all the new UI folders into nimbus_ui_new

fixes #10542


mikewilli:
The idea is that this is a temporary folder until it reaches parity with `nimbus_ui` and then we rename that one? Or does this live potentially indefinitely alongside `nimbus_ui`? Thinking about things like the Results page that may not fit as neatly into HTMX and so might take a while to migrate, and how a longterm `_new` folder might start to look odd.

Also, heck yea :tada: /gif IT'S HAPPENING

jaredlockhart:
@mikewilli Yeah after we reach parity in `_new` we can drop `nimbus-ui` and rename this one.

docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
       - media_volume:/experimenter/experimenter/media
       - /experimenter/experimenter/legacy/legacy-ui/core/.cache/
       - /experimenter/experimenter/legacy/legacy-ui/core/node_modules/
-      - /experimenter/experimenter/theme/static_src/node_modules/
+      - /experimenter/experimenter/nimbus_ui_new/theme/static_src/node_modules/
       - /experimenter/experimenter/nimbus-ui/node_modules/
```
brennie, line 22: nimbus_ui_new vs nimbus-ui ?

jaredlockhart: I know it's because `-` is not valid for Python package names, which didn't matter for `nimbus-ui` but does for `nimbus_ui_new` since it has views, tests, etc in it.  Annoying.

yashikakhurana (approved):
Tested that locally, works absolutely fine 💯 , thanks @jaredlockhart

PR #10545, opened 2024-04-09
Trigger CI docker builds so we can update to the newest docker selenium images.

jrbenny35:
This should fix the macos integration tests


jrbenny35:
Fixed in #10547

PR #10553, opened 2024-04-10
FXIOS-8918 Add iOS iPhone only target

adudenamedruby:
Because

- We need to be able to target iPhone or iPad users only...

This commit

- Adds a target to give the ability to do so.

Fixes [this issue](https://github.com/mozilla-mobile/firefox-ios/issues/19706)


experimenter/experimenter/targeting/constants.py
```diff
@@ -1764,6 +1764,17 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+IOS_IPHONE_USERS_ONLY = NimbusTargetingConfig(
+    name="iPhone users only",
+    slug="ios_iphone_users_only",
+    description="Targeting iPhone users",
+    targeting="is_phone",
+    desktop_telemetry="",
+    sticky_required=True,
```
jaredlockhart, line 1773: You shouldn't need sticky here since a profile on an iphone can't be migrated onto an ipad or other ios device, I don't think?

jaredlockhart (approved):
Very nice @adudenamedruby 🙏

PR #10556, opened 2024-04-11
bug(nimbus): fetch legacy Fenix manifests first

### LINKED ISSUE #10555: Imported feature with version is disabled
I just noticed that [this feature](https://localhost/admin/experiments/nimbusfeatureconfig/151/change/?_changelist_filters=application__exact%3Dfenix) is imported, has a version, but is marked as disabled, which should not be the case.  This feature is available in current versions of the application and so should be enabled, so it we must not be handling some case correctly.  This could be related to the fact that we recently had to [add two entries for fenix](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/features/manifests/apps.yaml#L5-L43) in the application configs to deal with the migration to MC.

brennie:
Because:

- we have two entries for Fenix in apps.yml (one for @mozilla-mobile/firefox-android and one for @mozilla/gecko-dev); and
- the order of these entries resulted (gecko-dev before firefox-android) resulted in the unversioned manifests for Fenix from mozilla-central being overwritten by the unversioned manifests from firefox-android

This commit:

- re-orders the entries so that the version from mozilla-central is the one that is used.

Fixes #10555


jaredlockhart (approved):
J'approve, merci @brennie 🙏 🎉

PR #10560, opened 2024-04-11
feat(nimbus): results_ready should be based off of the computed enrollment

### LINKED ISSUE #10559: Results ready date should use the actual enrollment end date if available
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4474)

mikewilli:
Because

- enrollment can end before proposed enrollment duration is over
- results link is not available until `results_ready_date` + 8 days
- `results_ready_date` is (was) based off of the proposed enrollment
- proposed enrollment does not change to reflect actual enrollment

This commit

- updates the `results_ready_date` to use the actual (if available) or computed enrollment

Fixes #10559


jaredlockhart (approved):
Ah yep totally, good catch 👍

PR #10578, opened 2024-04-15
adding support for targeting 125.0.1

cbellini:
Because

- we have an experiment that will target 125.0.1 and higher

This commit

- adds 125.0.1 as a select-able target for experiments


yashikakhurana (approved):
Thank you @cbellini for adding this 🎉

PR #10582, opened 2024-04-16
feat(nimbus): Add first run iOS iPhone only target

adudenamedruby:
Because

- We need to be able to target first run iPhone users only...

This commit

- Adds a target to give the ability to do so... I think.

Fixes [FXIOS-8918](https://github.com/mozilla-mobile/firefox-ios/issues/19706)


jaredlockhart (approved):
Yep looks good, thnx @adudenamedruby 🎉

PR #10587, opened 2024-04-16
adding support for targeting 125.0.2

cbellini:
Because

- we need to target an experiment for 125.0.2 and higher (formerly 125.0.1)

This commit

- adds 125.0.2 as a target

Fixes #github_issue_number


yashikakhurana (approved):
Thank you @cbellini

PR #10602, opened 2024-04-22
feat(nimbus): Fetch focus iOS manifests from the firefox-ios repo

### LINKED ISSUE #10601: Support Firefox iOS Monorepo: Focus iOS
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4159)

brennie:
Because:

- in 124, focus-ios merged into the @mozilla-mobile/firefox-ios repository;

This commit:

- adds support for specifying minimum versions in apps.yaml;
- updates the focus-ios entry to reference the firefox-ios repository; and
- deletes the focus-ios ref cache because there will be no overlapping refs between the old and new repositories.

Fixes #10601


experimenter/manifesttool/releases.py
```diff
@@ -60,6 +60,9 @@ def discover_tagged_releases(
 
     min_version = Version(max_major_version - 4, 0, 0)
 
+    if strategy.minimum_version is not None:
+        min_version = max(min_version, strategy.minimum_version)
+
```
jrbenny35, line 65: Could you do
```suggestion
    if strategy.minimum_version:
        min_version = max(min_version, strategy.minimum_version)

```

jrbenny35 (approved):

experimenter/manifesttool/releases.py
brennie: Yes, but I prefer the explicit None check, which makes it clear its either something or None. Also

> Explicit is better than implicit.
according to the zen of python (https://peps.python.org/pep-0020/) :P

jaredlockhart: Oh yeah explicitly checking for `None` has semantic meaning, good to keep here 👍

jaredlockhart:
@brennie Is out sick, so I'll just merge this now and if we need to revert it we can 👍

PR #10628, opened 2024-04-29
chore(nimbus): show all branches option for single branch experiments

### LINKED ISSUE #10609: Required/excluded experiments not being shown in edit
Experiment [(TEST) Messaging Version](https://experimenter.services.mozilla.com/nimbus/test-messaging-version/edit/audience?show-errors) excludes "[Viewpoint iOS July 2023 (All branches)](https://experimenter.services.mozilla.com/nimbus/viewpoint-ios-july-2023/summary)" but it does not appear in the audience page.

image

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4488)

jaredlockhart:
Because

* We recently added the ability to require/exclude experiments by branch
* We added multiple entries to the require/exclude drop downs for either the "All Branches" option, or individual branches
* For experiments with a single branch, we filtered out the "All Branches" option since it's essentially a duplicate of the single branch option
* However some experiments may get into a state where they point to the "All Branches" option for a single branch experiment, but then show no selection in the drop down
* We can revisit this behaviour later when we finish HTMXing the UI, let's just find a simple solution for now

This commit

* Removes the filter that was previously added
* Always shows the "All Branches" option

fixes #10609

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -130,7 +130,7 @@ describe("FormAudience", () => {
       const excluded = screen.getByLabelText(/Exclude users enrolled/);
       await selectEvent.openMenu(excluded);
       let options = container.querySelectorAll(query("excludedExperiments"));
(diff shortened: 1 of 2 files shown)


yashikakhurana (approved):
Okay it makes sense to do this way, neat fix

PR #10653, opened 2024-05-06
chore(nimbus): replace tailwind with bootstrap

### LINKED ISSUE #10543: Replace tailwind with bootstrap in new Nimbus UI
After some initial prototyping with Tailwind, we've decided to just stick with bootstrap, so let's migrate the existing history page over to bootstrap and then we can continue to build the new Nimbus UI on top of that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4465)

jaredlockhart:
Because

* We investigated using tailwind for the new nimbus ui
* It turns out tailwinds utility class pattern is more verbose than expected
* We'll just continue to use bootstrap since it's worked well for previous uis in experimenter

This commit

* Removes tailwind
* Adds bootstrap5
* Reimplements the new history page in bootstrap5

fixes #10543

image

image


experimenter/experimenter/experiments/changelog_utils.py
```diff
@@ -167,6 +168,7 @@ def get_formatted_change_object(field_name, field_diff, changelog, timestamp):
         )
 
     return {
+        "id": str(uuid.uuid4()),
```
brennie, line 171: Why this change? And why is it a different UUID for each call?

experimenter/experimenter/experiments/models.py
```diff
@@ -3,6 +3,7 @@
 from collections import defaultdict
 from dataclasses import dataclass
 from decimal import Decimal
+import json
```
brennie, line 6: Unused?

experimenter/experimenter/nimbus_ui_new/templates/changelog/overview.html
```diff
@@ -1,42 +1,90 @@
...
-        {% for changelog in experiment.get_changelogs_by_date  %}
-          <li class="mb-8">
-            <div class="flex-start flex items-center">
-              <div class="-ml-[16px] -mt-2 mr-3 flex  items-center justify-center rounded-full p-2 floating bg-white">
-                <i class="fa-regular  fa-calendar  text-azure"></i>
-              </div>
-              <h4 class="-mt-2 text-lg sm:text-xl font-semibold text-gray-600">{{ changelog.date }}</h4>
+    <div class="col-11">
+      <span class="fs-5 fw-semibold">
+        {{ changelog.date }}</span>
```
brennie, line 26: Why is this formatted like this?

Can we get an auto-formatter pass for django templates?

experimenter/experimenter/experiments/models.py
jaredlockhart: oops 🙊

experimenter/experimenter/experiments/changelog_utils.py
jaredlockhart: Yeah so I needed to have some unique identifier for each record to use as the html ID for the 'show more' id reference, and I couldn't use the db id since a single changelog appears as multiple entries in the history page.  I'm open to suggestions about some other way to uniquely identify each row?

mikewilli: I'm probably missing something important here, but if I'm correctly understanding what this is doing, could you do something like `${changelog.id}_${field_name}`?

experimenter/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.6.0 and should not be changed by hand.
```
mikewilli, line 1: So this is unrelated to this specific PR, but maybe worth discussing... This seems to go back and forth between <poetry version in repo> and <some local poetry version someone has> depending on whether the lockfile is generated in docker or locally. Is it worth trying to validate this in some way?

experimenter/experimenter/experiments/changelog_utils.py
jaredlockhart: Yeah I was thinking that too but then we're placing constraints on those fields, when really we just want something that says 'this is an arbitrary unique identifier for this row.'  Another way to do it is to hash its contents.  I'm not sure any one is preferable over another?

experimenter/poetry.lock
jaredlockhart: Oh funny, look at that.  I think we can just rely on poetrys own enforcement of its lockfile.  There are cases where you can update the lockfile outside of the container in a way that will cause the container side to error out entirely.  I'd say if poetry is happy with the change inside the container then it's fine.

experimenter/experimenter/nimbus_ui_new/templates/changelog/overview.html
jaredlockhart: Huzzah! https://github.com/mozilla/experimenter/issues/10656

experimenter/poetry.lock
mikewilli: Yea that's true -- the changes should be caught by existing testing so fair enough!

experimenter/experimenter/nimbus_ui_new/templates/common/footer.html
```diff
@@ -1,5 +1,6 @@
-<footer class="w-full mt-2 flex justify-center items-center">
-    <div class="flex flex-col justify-center text-spaceCadet items-center w-5/6 p-4 border-t border-azure">
-      © Mozilla Corporation 2023
-    </div>
+<footer class="text-center my-4">
+  <hr />
```
brennie, line 2: You don't need the `/`. Thats a JSX / XHTML thing.

experimenter/experimenter/nimbus_ui_new/templates/experimenter_base.html
```diff
@@ -1,25 +1,29 @@
...
+
 <head>
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>{% block title %}{% endblock %}</title>
-    {% tailwind_css %}
     <script src="{% static 'fontawesomefree/js/all.min.js' %}"></script>
-    <script defer src="{% static 'scripts/bundle.js' %}"></script>
+    <link rel="icon" href="/static/imgs/favicon.ico" />
```
brennie, line 15: Again no need for `/` here.

brennie (changes requested):

experimenter/experimenter/nimbus_ui_new/templates/common/footer.html
jaredlockhart: Ooo let's see if we can enforce that with the djlinter.

PR #10666, opened 2024-05-07
feat(nimbus): Multi choice recommendations

### LINKED ISSUE #10650: Support multi recommendation backend
Add a new field to support multi recommendations field

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4501)

yashikakhurana:
Because

- Users wanted to do multi-select for conclusion recommendations in a takeaway section

This commit

- Adds new field to support multi-support for conclusion recommendations

Note: In the next PR we will update frontend and migration, will remove the old field then

Fixes #10650
Api response

Screenshot 2024-05-07 at 11 56 49 AM

Admin panel 

Screenshot 2024-05-07 at 11 50 46 AM


experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -502,6 +502,9 @@ class NimbusExperimentType(DjangoObjectType):
     conclusion_recommendation = graphene.Field(
         NimbusExperimentConclusionRecommendationEnum
     )
+    conclusion_recommendations = graphene.List(
+        NimbusExperimentConclusionRecommendationEnum
+    )
```
brennie, line 507:
```
conclusion_recommendations = graphene.List(
    NimbusExperimentConclusionRecommendationEnum
)
```

generates 
```
conclusionRecommendations: (NimbusExperimentConclusionRecommendationEnum | null)[] | null;
```

We definitely dont want to accept `[null]` here, so we need to wrap the `NimbusExperimentConclusionRecommendationEnum` in `graphene.NonNull`

Do we want the field to accept null?  (probably not) If no, we need to wrap it in a nonnull.

```
graphene.NonNull(graphene.List(graphene.NonNull(NimbusExperimentConclusionRecommendationEnum)))
```
This will generate

```
conclusionRecommendations: NimbusExperimentConclusionRecommendationEnum[];
```

brennie (changes requested):

experimenter/experimenter/experiments/api/v5/types.py
yashikakhurana: Since in my model I have defined 
` conclusion_recommendations = models.JSONField(
        verbose_name="Conclusion Recommendations", blank=True, null=True
    )`
I think it should be ready to accept null, for example when we just create the experiment at that time it won't have anything

yashikakhurana: or I need to change the model to provide default value `[]`

brennie: yeah change the default to be [] and we dont have to worry about nulls anywhere.

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -502,6 +502,9 @@ class NimbusExperimentType(DjangoObjectType):
     conclusion_recommendation = graphene.Field(
         NimbusExperimentConclusionRecommendationEnum
     )
+    conclusion_recommendations = graphene.NonNull(
+        graphene.List(NimbusExperimentConclusionRecommendationEnum)
```
brennie, line 506: You need nonnull here too

graphene.List(graphene.NonNull(...))

brennie (approved):

PR #10707, opened 2024-05-15
bug(nimbus): Correctly remove fetch-summary.txt

### LINKED ISSUE #10706: update_external_configs still broken
We're attempting to remove fetch-summary.txt instead of experimenter/fetch-summary.txt

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4537)

brennie:
Because:

- we are attempting to remove the wrong file, causing the update task to fail

this commit:

- creates the experimenter/fetch-summary.txt file before calling
  `make fetch_external_resources` so that it will have read and write
  permissions; and
- moves the file afterwards so it will not be included in the commit.

Fixes #10706

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -504,9 +504,9 @@ jobs:
             git checkout main
             git pull origin main
             cp .env.sample .env
+            touch ./experimenter/fetch-summary.txt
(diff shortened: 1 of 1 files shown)


brennie:
See #10709

jaredlockhart (approved):
Oh I don't miss having to deal with cases like this, yep let's give it a go 👍 thnx @brennie 🙏

PR #10715, opened 2024-05-15
chore(nimbus): Update application-services and address breaking change

jeddai:
Because

- A breaking change was introduced to application services for generated Python code

This commit

- Updates the version of application services being used
- Addresses the breaking change noted here #10705


jaredlockhart (approved):
Awesome thank you @jeddai 🎉 🙏

PR #10747, opened 2024-05-22
chore(nimbus): add search rollout targeting

### LINKED ISSUE #10746: Create search rollout advanced targeting for explicit namespacing
We have expected that at some point users will want/need to be able to explicitly create/control namespaces to either explicitly keep multiple experiments/rollouts in the same namespace or in separate parallel namespaces.  We've now encountered the first use case of that with two search rollouts that would otherwise share the same namespace but want to be able to scale up/down independently without stepping on each other.  

To accomplish this, we can use two advanced targetings with empty targeting strings but different slugs.  By selecting each of the two advanced targetings, the advanced targeting slugs will be included in the namespace and thus we can use advanced targeting slug as a way to manually control namespacing.

Obviously we shouldn't use this as a long term solution, but this will serve as a nice prototype of a first example of explicitly controlling namespacing.  As this use case continues to grow we should design and implement a more robust solution.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4566)

jaredlockhart:
Because

* We need to be able to put two rollouts for the same feature into two separate namespaces
* The simplest way to do that right now is by using two advanced targetings with different slugs
* This can serve as a prototype of a user definable namespace workflow that we can develop further in the future

This commit

* Adds two advanced targetings for desktop for Search to use as a way to control their rollout namespaces

fixes #10746

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1944,6 +1944,28 @@ NEW_PROFILE_MAC_ONLY = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+SEARCH_ROLLOUT_1 = NimbusTargetingConfig(
+    name="Search Rollout 1",
+    slug="search_rollout_1",
+    description="Search Rollout 1 Namespace",
+    targeting="",
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+SEARCH_ROLLOUT_2 = NimbusTargetingConfig(
+    name="Search Rollout 2",
+    slug="search_rollout_2",
+    description="Search Rollout 2 Namespace",
+    targeting="",
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
 
 class TargetingConstants:
     TARGETING_VERSION = "version|versionCompare('{version}') >= 0"


experimenter/experimenter/targeting/constants.py
```diff
@@ -1944,6 +1944,28 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+SEARCH_ROLLOUT_1 = NimbusTargetingConfig(
+    name="Search Rollout 1",
+    slug="search_rollout_1",
+    description="Search Rollout 1 Namespace",
+    targeting="",
```
brennie, line 1951: This should probably be `true` because I can see this easily getting goofed up when we mash together targeting stuff (like sticky)

experimenter/experimenter/targeting/constants.py
```diff
@@ -1944,6 +1944,28 @@ def __post_init__(self):
...
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
+
+SEARCH_ROLLOUT_2 = NimbusTargetingConfig(
+    name="Search Rollout 2",
+    slug="search_rollout_2",
+    description="Search Rollout 2 Namespace",
+    targeting="",
```
brennie, line 1962: This too

brennie (changes requested):

experimenter/experimenter/targeting/constants.py
jaredlockhart: There are tests that create a JEXL out of each advanced targeting and validate it against both the python and js JEXL runtimes so if this generates something invalid I'd expect those tests to fail.

brennie (approved):

experimenter/experimenter/targeting/constants.py
jaredlockhart: Ah also, empty targeting will not be included in the targeting string at all:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/models.py#L474-L475

but the slug will be included in the namespace:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/models.py#L797-L798

which I think is waht we want here

PR #10749, opened 2024-05-22
bug(nimbus): count statuses in db instead of in memory

### LINKED ISSUE #10748: Count experiments in the db instead of in memory
We recently deployed the new htmx list page, but it is taking huge amounts of time and memory.  One possible cause for this is loading all ~1k experiments into memory to count them.  Instead, let's try issuing individual db calls to count each status.  This will add more db round trips but should lower the memory cost and prevent loading a large number of experiments out of the db and into memory unnecessarily.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4567)

jaredlockhart:
Because

* We had added logic to count the number of experiments in each status for the new list page
* We did this by looping over them and counting them in memory
* This unnecessarily loads a large number of experiments into memory which blows up the db query and memory cost per request

This commit

* Counts the experiments in the db instead of in memory

fixes #10748

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui_new/tests/test_views.py
@@ -87,8 +87,8 @@ class NimbusExperimentsListViewTest(TestCase):
         )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(
-            [e.slug for e in response.context["experiments"]],
-            expected_slugs,
+            {e.slug for e in response.context["experiments"]},
+            set(expected_slugs),
         )
 
     @patch(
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus_ui_new/tests/test_views.py
```diff
@@ -60,7 +60,7 @@ def test_render_to_response(self):
             dict(response.context["status_counts"]),
             {
                 NimbusExperiment.Status.COMPLETE.value: 1,
-                NimbusExperiment.Status.DRAFT.value: 2,
+                NimbusExperiment.Status.DRAFT.value: 3,
```
yashikakhurana, line 63: why does the count increase? Isn't that should remain the same? just curious

yashikakhurana (approved):
perfect, thanks @jaredlockhart  ⭐

PR #10753, opened 2024-05-23
feat(nimbus): Support multiple RS collections per app

### LINKED ISSUE #10744: Support multiple RS collections per app
The current ApplicationConfig class only supports a single RS collection. However, now we are publishing Firefox Desktop recipes to multiple collections so that needs to change.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4565)

brennie:
Because:

- Experimenter will soon be publishing Firefox Desktop recipes to different collections depending on recipe contents

this commit:

- adds scaffolding to support multiple collections per application.

Fixes #10744.


experimenter/experimenter/experiments/constants.py
```diff
@@ -54,9 +59,17 @@ class ApplicationConfig:
     slug: str
     app_name: str
     channel_app_id: Dict[str, str]
-    kinto_collection: str
+    default_kinto_collection: str
     randomization_unit: str
     is_web: bool
+    additional_kinto_collections: list[str] = field(default_factory=list)
```
jaredlockhart, line 65: Won't we want this to be a dict of feature_config slugs to collection ids?

jaredlockhart: Also if we're gonna introduce this field here in this pr, do we wanna also setup the logic that populates and consumes it?

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_experiment_serializer.py
```diff
@@ -953,13 +955,23 @@ def test_update_publish_status_to_approved_invokes_push_task(self, application):
...
-        experiment = serializer.save()
+        expected_collection = experiment.application_config.get_kinto_collection_for(experiment)
+
+        with patch.object(
+            experiment.application_config,
+            "get_kinto_collection_for",
+            wraps=experiment.application_config.get_kinto_collection_for,
+        ) as get_kinto_collection_for:
+            experiment = serializer.save()
+            get_kinto_collection_for.assert_called_once_with(experiment)
```
jaredlockhart, line 966: Wait why do we have to mock and assert on this call here?  Isn't it sufficient just to check which collection is called by the kinto client later?

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1709,6 +1719,33 @@ def _validate_feature_configs(self, data):
...
+        application_config = self.instance.application_config
+        collections_by_feature_id = application_config.kinto_collections_by_feature_id
+
+        if collections_by_feature_id is not None:
+            target_collections = {
+                feature_config.slug: collections_by_feature_id.get(
+                    feature_config.slug, application_config.default_kinto_collection
+                )
+                for feature_config in feature_configs
+            }
```
jaredlockhart, line 1734: chefs-kiss-french-chef

experimenter/experimenter/experiments/constants.py
```diff
@@ -54,9 +59,38 @@ class ApplicationConfig:
...
+                    feature.slug, self.default_kinto_collection
+                )
+                for feature in experiment.feature_configs.all()
+            }
+
+            if len(target_collections) == 1:
+                return next(iter(target_collections))
+
+            elif len(target_collections) > 1:
+                raise AssertionError("Experiment targets multiple collections")
```
jaredlockhart, line 83: Ah yeah I didn't think of it here, this should be handled in the review serializer.

PR #10786, opened 2024-06-03
Revert "feat(nimbus): Update conclusion recommendation in changelogs"

yashikakhurana:
Because
- We are trying to update changelogs, whereas we should make the history page more subtle to changes when we update the field.

This Commit 
- Reverts mozilla/experimenter#10782


brennie (approved):
rs+

PR #10789, opened 2024-06-03
feat(nimbus): add sorting to new list page

### LINKED ISSUE #10682: List Page Sorting
Let's add the ability to sort by each of the table headers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4525)

jaredlockhart:
Because

* We are building a new list page in django templates and htmx
* We want to be able to sort the table by the column headers

This commit

* Adds sorting by each of the column headers
* Defaults to sorting by most recently updated (similar to previous list pages)

fixes #10682

https://github.com/mozilla/experimenter/assets/119884/7253c112-ed66-44a5-9bcc-6f501c57734b

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/filtersets.py
+++ b/experimenter/experimenter/nimbus_ui_new/filtersets.py
@@ -26,6 +26,25 @@ class TypeChoices(models.TextChoices):
     EXPERIMENT = "Experiment"
 
 
+class SortChoices(models.TextChoices):
+    NAME_UP = "name"
+    NAME_DOWN = "-name"
+    QA_UP = "qa_status"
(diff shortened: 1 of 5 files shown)


brennie (approved):

yashikakhurana:
the only feedback I have is that it is not clearly visible that you can sort the fields unless we click on that

jaredlockhart:
Yeah, I'm a bit confused about the visual pattern around black text vs blue text for clickable things.  I keep finding conflicting examples in the bootstrap docs and also just in examples.  Github does it where things are black unless you mouse over and then it becomes blue, that indicates that something is clickable.  Maybe we should do something like that?

PR #10802, opened 2024-06-05
chore(nimbus): prefetch feature configs on new nimbus list page

### LINKED ISSUE #10713: List page db query optimization
After we get the new list page setup, let's make sure we're prefetching the right fields and not anything we don't need to.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4540)

jaredlockhart:
Because

* Now that all the fields are in place for the new nimbus list page we can clean up the db queries
* Looks like the only n+1 related call is to feature configs

This commit

* Prefetches feature configs on the new nimbus list page

fixes #10713


yashikakhurana (approved):
Improvements 💪 , thanks @jaredlockhart

PR #10804, opened 2024-06-05
feat(nimbus): Handle the changelogs if field does not exist

### LINKED ISSUE #10780: NimbusExperiment has no field named 'conclusion_recommendation'
Sentry Issue: [EXPERIMENTER-PROD-PZ](https://mozilla.sentry.io/issues/5434439427/?referrer=github_integration)

```
NimbusExperiment has no field named 'conclusion_recommendation'
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4576)

yashikakhurana:
Because

-  In the changelogs we capture all the data, so if we have removed the field from the model it will still refer to the removed field which can lead to an error like [this](https://mozilla.sentry.io/issues/5434439456/?project=6377199&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=4)

This commit

- Add exception handling so that if the field doesn't exist, it can skip that changelog

Fixes #10780


jaredlockhart (approved):
Yeah exactly, great job thnx @yashikakhurana 🙏

experimenter/experimenter/experiments/tests/test_changelog_utils.py
```diff
@@ -374,3 +375,21 @@ def test_generates_changelog_with_out_of_date_latest_change(self):
...
+
+        try:
+            change_object = get_formatted_change_object(
+                non_existent_field_name,
+                field_diff,
+                changelog,
+                timestamp,
+            )
+        except Exception as e:
+            self.fail(f"Exception raised: {e}")
```
brennie, line 393: TIL self.fail, but we dont need the try/except because if it throws the test will fail

PR #10808, opened 2024-06-06
feat(nimbus): separate view for experiment table in new nimbus ui

### LINKED ISSUE #10803: Separate table and sidebar rendering in new list page
After looking at the flame graph in Sentry profiling for the new nimbus list page here:

image

I can see that the bulk of the time spent to render the page is in the sidebar, where each filter is rendering hundreds of entries (for features, locales, countries, users, etc).  This is what's taking up about 80% of the time to render the page.  But we only need to render the sidebar once on the initial page load, after that we only need to re render the table of experiments.  

So let's split the view in two, one to render the entire page, and one to render just the table.  Then we can set all the htmx calls to update just the table.

One slight hiccup is figuring out how to update the url with the query args without swapping to the table url.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4587)

jaredlockhart:
Because

* After profiling the new nimbus ui list view we see that the bulk of time rendering is in the sidebar
* We only need to render the sidebar once on initial page load
* We can update only the table after filters, tabs, sorts change
* This should significantly improve the responsiveness of the page

This commit

* Splits the table out into its own template
* Adds a new view that renders only the table
* Updates the htmx directives to call the table renderer

fixes #10803

Before:

https://github.com/mozilla/experimenter/assets/119884/7807c453-ad72-4a75-a8e3-0416b027b7da

After:

https://github.com/mozilla/experimenter/assets/119884/653c0d60-302f-43a4-95e7-1dc6bcb4c0cc

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/list.html
+++ b/experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/list.html
@@ -7,11 +7,10 @@
 {% endblock title %}
 
 {% block sidebar %}
-  <form hx-get="{% url "nimbus-new-list" %}"
+  <form id="filter-form"
+        hx-get="{% url "nimbus-new-table" %}"
         hx-trigger="keyup,mouseup delay:100ms"
-        hx-select="#experiment-list"
         hx-target="#experiment-list"
-        hx-swap="outerHTML"
         hx-push-url="true">
     {% for field in filter.form %}{{ field }}{% endfor %}
   </form>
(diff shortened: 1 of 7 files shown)


experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/list.html
```diff
@@ -7,140 +7,30 @@
...
-        </div>
-      </div>
-    {% endif %}
   </div>
 {% endblock main_content %}
 
 {% block extrascripts %}
   {{ block.super }}
   document.body.addEventListener('htmx:afterSwap', function(evt) {
   $('.selectpicker').selectpicker()
```
brennie, line 29: Missing semicolon :)

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/list.html
```diff
@@ -7,140 +7,30 @@
...
+  const urlParams = new URLSearchParams(window.location.search);
+  const sortValue = urlParams.get('sort');
+  const statusValue = urlParams.get('status');
+  $('#filter-form input[name="status"]').each((i, e) => e.value = statusValue);
+  $('#filter-form input[name="sort"]').each((i, e) => e.value = sortValue);
```
brennie, line 34: as we discussed, this deserves a comment

PR #10811, opened 2024-06-06
bug(nimbus): push query args to list page url properly

### LINKED ISSUE #10810: Persist search terms in url after htmx get in new nimbus list page
Now that we're fetching the table separately on the new nimbus list page, the push url will replace the pages url with the tables url.  We need a way to persist the query args in the url without changing the path.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4590)

jaredlockhart:
Becuase

* When you fetch a url with htmx it will override the current url in the urlbar and history
* When we separated the experiment table into its own template/view/url, then the url after fetching becomes the url of the table and not the list page
* We need a way to push only the query args to the url without changing the path
* HTMX allows you to do this by setting a header in the response

This commit

* Sets the HX-Push header in the response from the table view to only include the query args so that the path will remain the same but the query args will be updated with the new values

fixes #10810


yashikakhurana (approved):
:shipit:

PR #10817, opened 2024-06-07
chore(nimbus): disable intermittently failing nimbus ui tests

### LINKED ISSUE #10812: Disable intermittently failing nimbus ui test
There is a frequently intermittently failing test in Nimbus UI:

  ● PageHome › supports updating search params when tabs are clicked

That is frequently blocking prs/merges/deploys.  Since we are in the process of migrating away from Nimbus UI to Nimbus UI New, let's disable this test now so that we don't have frequent interruptions in merges and deploys.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4591)

jaredlockhart:
Because

* We are in the process of migrating our frontend from react to django templates/htmx
* There are some intermittently failing tests in the react frontend
* We are not making any significant changes to the react frontend
* We can safely disable these tests until we deprecate the react frontend

This commit

* Skips two intermittently failing tests
* Skips coverage for the affected lines

fixes #10812

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageHome/FilterBar/index.tsx
@@ -183,6 +183,7 @@ export const FilterSelect = <
             )!,
           options: filterOptions as Options<OptionType>,
           onChange: (fieldValue: Options<OptionType>) => {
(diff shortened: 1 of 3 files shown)


yashikakhurana (approved):
fair point

PR #10820, opened 2024-06-10
feat(nimbus): Filter with conclusion recommendations

### LINKED ISSUE #10813: add recommendation options to the dropdown filter for "take-aways"
Challenge:  People don’t fill out take-aways today - and who can blame them?  There’s not much in it for them unless they need to go back and look at a specific experiment link.  Even then it can be hard to find the specific experiment with the right verbose results if they launched a series across different Geographies or a series of experiments over time (or relaunched an experiment so have the old link).

**What would be the benefit if they fill out take-aways?**

* Leadership can find the experiments that moved DaU meaningfully or had QBR valuable learnings (exists today)
* Onboarding, messaging, and Comms team want to be able to see all experiments they’ve had to re-run to see if there are common errors that could be avoided.
* Nimbus team asked “so how many experiments helped us decide what to do/changed our minds” - can filter for changed course
* 【 highlighted #97a0af 】_Future: Product can find which experiment have any verbose takeaways (could skip the “reruns” as those likely aren’t the final results) - to more easily find tl;dr learnings and which experiments to look at the experiment brief details._【 end highlighted 】

Current Takeaway filter is multi-select with the bolded options.  proposal is to add the regular text options.  Future could be to add the 【 highlighted #97a0af 】_italic options_【 end highlighted 】.  

【 highlighted #bf2600 】**QUESTION: should these all be in one filter or should we create a filter for recommendations?**【 end highlighted 】

* **QBR Learning**
* **DaU Gain**
* rerun
* graduate
* change course
* stop
* run follow ups
* 【 highlighted #97a0af 】_Has takeaway (has filled in any words in the open text box)_【 end highlighted 】
* 【 highlighted #97a0af 】_Has no takeaway (has not filled in words in the open text box)_【 end highlighted 】

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4462)

yashikakhurana:
Because

- Users want to find the experiments with the concluded recommendations like Rerun, Grduate etc.

This commit

- Enable Conclusion recommendation filtering in the takeaways filtering section

Fixes #10813 
Screenshot 2024-06-10 at 9 35 27 AM


experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -202,7 +202,8 @@ class NimbusExperimentFilter(django_filters.FilterSet):
     )
     takeaways = django_filters.MultipleChoiceFilter(
         method="filter_takeaways",
-        choices=NimbusExperiment.Takeaways.choices,
+        choices=NimbusExperiment.Takeaways.choices
+        + NimbusExperiment.ConclusionRecommendation.choices,
```
brennie, line 206:
```suggestion
        choices=(
            NimbusExperiment.Takeaways.choices
            + NimbusExperiment.ConclusionRecommendation.choices
        ),
```

experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -302,8 +303,12 @@ def filter_type(self, queryset, name, value):
 
     def filter_takeaways(self, queryset, name, value):
         query = Q()
-        if NimbusExperiment.Takeaways.DAU_GAIN in value:
-            query |= Q(takeaways_metric_gain=True)
-        if NimbusExperiment.Takeaways.QBR_LEARNING in value:
-            query |= Q(takeaways_qbr_learning=True)
+        for val in value:
+            if val in dict(NimbusExperiment.Takeaways.choices):
```
brennie, line 307: do we really need to call dict() here? I'm sure theres something available on Takeaways

experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -302,8 +303,12 @@ def filter_type(self, queryset, name, value):
...
+                if val == NimbusExperiment.Takeaways.QBR_LEARNING:
+                    query |= Q(takeaways_qbr_learning=True)
+                elif val == NimbusExperiment.Takeaways.DAU_GAIN:
+                    query |= Q(takeaways_metric_gain=True)
+            elif val in dict(NimbusExperiment.ConclusionRecommendation.choices):
```
brennie, line 312: same here

experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -302,8 +303,12 @@ def filter_type(self, queryset, name, value):
 
     def filter_takeaways(self, queryset, name, value):
         query = Q()
-        if NimbusExperiment.Takeaways.DAU_GAIN in value:
-            query |= Q(takeaways_metric_gain=True)
-        if NimbusExperiment.Takeaways.QBR_LEARNING in value:
-            query |= Q(takeaways_qbr_learning=True)
+        for val in value:
```
brennie, line 306: Rename `value` to `values` and do `for value in values`

brennie (dismissed):

experimenter/experimenter/nimbus_ui_new/filtersets.py
yashikakhurana: so in all the methods like this for filtering we are using `def filter_takeaways(self, queryset, name, value):` as parameter name, I really don't want to change all the method name params to `values`, so that's why I did `val in value`

experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -202,7 +202,10 @@ class NimbusExperimentFilter(django_filters.FilterSet):
     )
     takeaways = django_filters.MultipleChoiceFilter(
         method="filter_takeaways",
-        choices=NimbusExperiment.Takeaways.choices,
+        choices=(
+            NimbusExperiment.Takeaways.choices
+            + NimbusExperiment.ConclusionRecommendation.choices
+        ),
```
jaredlockhart, line 208: Oh yes, but also I was gonna say there's a nicer way to say this now

```py
[
    *NimbusExperiment.Takeaways.choices,
    *NimbusExperiment.ConclusionRecommendation.choices,
]
```

since using math operands on collections technically functions but is a bit of an old/awkward way to say that.

experimenter/experimenter/nimbus_ui_new/filtersets.py
jaredlockhart: I know I was thinking the same thing.  Both things are true.  But the argument is positional so we should be able to rename it to `values` without breaking the call interface.

jaredlockhart (approved):
Okay tested locally, works perfectly, thanks @yashikakhurana! 🎉

experimenter/experimenter/nimbus_ui_new/filtersets.py
yashikakhurana: yeah then let me change other filters too from `value` to `values`

experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -300,10 +303,14 @@ def filter_type(self, queryset, name, value):
...
-            query |= Q(takeaways_metric_gain=True)
-        if NimbusExperiment.Takeaways.QBR_LEARNING in value:
-            query |= Q(takeaways_qbr_learning=True)
+        for value in values:
+            if value in NimbusExperiment.Takeaways:
+                if value == NimbusExperiment.Takeaways.QBR_LEARNING:
+                    query |= Q(takeaways_qbr_learning=True)
+                elif value == NimbusExperiment.Takeaways.DAU_GAIN:
+                    query |= Q(takeaways_metric_gain=True)
+            elif value in NimbusExperiment.ConclusionRecommendation:
```
brennie, line 314: We don't need the outside if, its redundant.

```suggestion
            if value == NimbusExperiment.Takeaways.QBR_LEARNING:
                query |= Q(takeaways_qbr_learning=True)
            elif value == NimbusExperiment.Takeaways.DAU_GAIN:
                query |= Q(takeaways_metric_gain=True)
            elif value in NimbusExperiment.ConclusionRecommendation:
```

yashikakhurana: the only reason was to group takeaways together and conclusion recommendations separately

PR #10822, opened 2024-06-10
feat(nimbus): update nimbus list page counts with filters

### LINKED ISSUE #10796: Update counts to reflect filters on new nimbus list page
When we apply filters, we need to update the counts on each tab to reflect the number of filtered experiments.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4584)

jaredlockhart:
Because

* Now that filters are hooked up on the new list page, we need to update the counts in the tabs with the applied statuses

This commit

* Counts statuses with all filters except the status filter on the new list page

fixes #10796

https://github.com/mozilla/experimenter/assets/119884/3dc006b1-2315-4dda-a46d-ae16752dff36

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/views.py
+++ b/experimenter/experimenter/nimbus_ui_new/views.py
@@ -38,10 +38,22 @@ class NimbusExperimentsListView(FilterView):
 
         return kwargs
 
+    def get_filterset(self, filterset_class):
+        kwargs = self.get_filterset_kwargs(filterset_class)
+
+        # Create a second filterset with the status removed from data
+        # to count experiments with all other filters applied except status
+        kwargs_no_status = kwargs.copy()
+        kwargs_no_status["data"] = kwargs_no_status["data"].copy()
+        kwargs_no_status["data"].pop("status")
+        self.filterset_no_status = filterset_class(**kwargs_no_status)
+
+        return filterset_class(**kwargs)
+
     def get_context_data(self, **kwargs):
-        queryset = self.get_queryset()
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -38,10 +38,22 @@ def get_filterset_kwargs(self, filterset_class):
...
 
+    def get_filterset(self, filterset_class):
+        kwargs = self.get_filterset_kwargs(filterset_class)
+
+        # Create a second filterset with the status removed from data
+        # to count experiments with all other filters applied except status
+        kwargs_no_status = kwargs.copy()
+        kwargs_no_status["data"] = kwargs_no_status["data"].copy()
+        kwargs_no_status["data"].pop("status")
+        self.filterset_no_status = filterset_class(**kwargs_no_status)
```
brennie, line 49: Can we just move this into get_context_data? I dont see the point of caching this on the view instance since its only going to be used once.

brennie (changes requested):

experimenter/experimenter/nimbus_ui_new/views.py
jaredlockhart: Oh yeah we could but then we'd have to pull a bunch of args/methods into the context method that aren't otherwise there.  I don't think caching it on the instance is a problem?

brennie (approved):

yashikakhurana (approved):
Tried locally, worked absolutely perfect 🎉  thanks @jaredlockhart

PR #10844, opened 2024-06-13
test(nimbus): Add android e2e test config.

### LINKED ISSUE #10843: Add android e2e tests to experimenter CI
Using some of the steps that we do in Klaatu, we can create a simple end to end test that checks enrollment within firefox on android. We should build it from source using a docker image.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4614)

jrbenny35:
Because

It would be awesome to extend our coverage to android devices to make sure those stay healthy.

This commit

* Adds a job to build the fenix test APKs from mozilla central.
* Adds a job to run a small suite of tests on fenix

Fixes #10843


experimenter/tests/integration/android/build_android.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd mobile/android/fenix \
+&& ./gradlew clean app:assembleFenixDebug \
+&& mv app/build/outputs/apk/fenix/debug/app-fenix-x86_64-debug.apk ./ \
+&& ./gradlew clean app:assembleFenixDebugAndroidTest \
+&& mv app/build/outputs/apk/androidTest/fenix/debug/app-fenix-debug-androidTest.apk ./
```
brennie, line 9: Several files missing newline at eof.

if youre using vscode, set

```
"files.insertFinalNewline": true,
```

brennie, line 9: You don't need `&&` here.

Since you have `set -e`, any command that fails will cause the script to exit.
```suggestion
cd mobile/android/fenix
./gradlew clean app:assembleFenixDebug
mv app/build/outputs/apk/fenix/debug/app-fenix-x86_64-debug.apk ./
./gradlew clean app:assembleFenixDebugAndroidTest
mv app/build/outputs/apk/androidTest/fenix/debug/app-fenix-debug-androidTest.apk ./
```

brennie (approved):
r+ but sad we cant piggy back off other build infrstructure for this. Do we know how long this build takes to run?

.circleci/config.yml
```diff
@@ -584,6 +621,27 @@ jobs:
...
+            cd experimenter/tests/integration/android
+            docker build -t moz-central-builder -f moz-central.test-container.Dockerfile .
+            docker run --name fenix-builder moz-central-builder bash build_android.sh
+            docker cp fenix-builder:mozilla-central/mobile/android/fenix/app-fenix-debug-androidTest.apk ./
+            docker cp fenix-builder:mozilla-central/mobile/android/fenix/app-fenix-x86_64-debug.apk ./
+      - save_cache:
+          key: experimenter-firefox-fenix
+          paths:
+            - experimenter/tests/integration/android/app-fenix-debug-androidTest.apk
+            - experimenter/tests/integration/android/app-fenix-x86-debug.apk
```
jaredlockhart, line 643: Oh you know one thing we learned from the other integration tests that built a firefox was that building them in a separate task and periodically pushing them can make it harder to keep things in sync.  Can we do these build steps as part of the test run above so that the containers are built on each test run?  The docker layer caching should help keep that quick if nothing has changed.

.circleci/config.yml
```diff
@@ -396,6 +397,42 @@ jobs:
...
+            adb install app-fenix-debug-androidTest.apk
+            adb install app-fenix-x86_74-debug.apk
+      - run:
+          name: Download test files and run
+          command: |
+            git clone https://github.com/mozilla/gecko-dev.git --depth 1
+            cd gecko-dev/mobile/android/fenix/app/src/androidTest/java/org/mozilla/fenix/experimentintegration
+            pip install pipenv
+            pipenv install
+            pipenv run pytest --experiment new-user-study-android-april-2024 --experiment-branch treatment --experiment-feature generic_test
```
jaredlockhart, line 430: Awesome 🤩 🤩 🤩 🤩 🤩 🤩 🤩

jrbenny35:
So I totally forgot I wanted to set up experimenter locally and have the test use that as the experiment it enrolls in. I might close this once I get the docker build uploaded and reopen.

.circleci/config.yml
jrbenny35: I think I can do this if I can just do a simple `hg pull && hg update` once I get the initial image set up. Right now, it will take forever and I don't see much value in having the absolute latest fenix changes. Weekly seems fine but if it doesn't take long with the previously mentioned changes, we could do it that way.

jaredlockhart: Ya if it's too slow then leaving it weekly is fine, but if we can do it right inside the test run that'll make it easier to keep everything in sync, and then we can do what we did for other repo dependencies and hardcode a hash and setup a task to bump it and that way we'll know if Fenix changes, and that breaks the Nimbus integration, we'll be able to identify where the breakage occurred, like we did recently with application services :D

jaredlockhart:
If you need to add a 'developer' or 'debug' channel to get this to pass, you can just do it right here:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/constants.py#L122-L124

.circleci/config.yml
brennie: Building Firefox can take upwards of 40 minutes on relatively modern hardware, so I would be wary of building Firefox in CI on every push.

If we build on a cron job, we should set up some sort of reporting so that we know immediately if the build starts to fail.

The tests that use the pre-built firefox should also check that it is a recent build (eg last two weeks?)

jaredlockhart: Hrmm yeah good point.  We should find some way that we can replicate the workflow we have with application services though, where we build against a known hash/version, and only bump that in a PR and then know if that creates any regressions.

experimenter/tests/integration/nimbus/android/local.properties
```diff
@@ -0,0 +1,2 @@
+nimbus.remote-settings.url=http://10.0.2.2:8888/
+glean.custom.server.url="http://10.0.2.2:5000/"
```
brennie, line 2: missing trailing newline

experimenter/tests/integration/nimbus/android/adbrun.py
```diff
@@ -0,0 +1,19 @@
...
+    logger = logging.getLogger()
+    here = Path()
+
+    def launch(self):
+        # First close sim if any then launch
+        os.system(
+            "~/Library/Android/sdk/platform-tools/adb devices | grep emulator | cut -f1 | while read line; do ~/Library/Android/sdk/platform-tools/adb -s $line emu kill; done"  #  NOQA
+        )
+        # Then launch sim
+        os.system("sh launchSimScript.sh")
```
brennie, line 19: This should probably be a bash script that we invoke with subprocess.check_result. Remember to `set -euo pipefail`

brennie, line 19: I don't see launchsimscript.sh?

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+import time
+from pathlib import Path
+
+import pytest
+import requests
+
+from .gradlewbuild import GradlewBuild
+from .models.models import TelemetryModel
+
+here = Path().cwd()
```
brennie, line 17: This gets the current working directory. Do you want the cwd or the directory of the file? (if the latter, `Path(__file__).parent()`)

Also, `cwd` is a classmethod, so you can do `Path.cwd()`.

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def start_process(path, command):
```
brennie, line 20: type annotations? :)

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+            encoding="utf-8",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+            cwd=module_path.absolute(),
+        )
+        stdout, stderr = process.communicate(timeout=5)
+
+        if process.returncode != 0:
+            raise Exception(stderr)
```
brennie, line 35: we seem to be throwing away the IO, so we should use `subprocess.check_call()` with `stdout` and `stderr` set to `subprocess.DEVNULL`

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+        finally:
+            with Path.open(gradlewbuild_log, "w") as f:
+                f.write(f"{out}")
+
+    return _run_nimbus_cli_command
+
+
+@pytest.fixture(name="open_app")
+def fixture_open_app(run_nimbus_cli_command):
+    def _():
```
brennie, line 59: Please give this a name so any stack traces are nicer.

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+def fixture_set_experiment_test_name():
+    os.environ["EXP_NAME"] = "fenix"
+
+
+@pytest.fixture(name="setup_experiment")
+def fixture_setup_experiment(
+    run_nimbus_cli_command,
+    delete_telemetry_pings,
+):
+    def _():
```
brennie, line 144: Here too

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+        experiment_json = next(iter(here.glob("**/fixtures/experiment.json")))
+        delete_telemetry_pings()
+        logging.info("Testing fenix e2e experiment")
+        command = f"nimbus-cli --app fenix --channel developer enroll firefox-fenix-test-experiment --branch control --file {experiment_json} --reset-app"  #  NOQA
```
brennie, line 148: You should prefer passing a list of strings instead of a single string so we dont have to worry about shell quoting.

experimenter/tests/integration/nimbus/conftest.py
```diff
@@ -84,6 +62,30 @@ def slugify(name):
     return name.lower().replace(" ", "-").replace("[", "").replace("]", "")
 
 
+@pytest.fixture(name="application_feature_ids")
```
brennie, line 65: Why this change? I'm not a fan of pytest fixtures tbh, they're pretty magical

experimenter/tests/integration/nimbus/android/test_fenix_integration.py
```diff
@@ -0,0 +1,46 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+
+@pytest.fixture
+def experiment_slug():
+    return "firefox-fenix-test-experiment"
```
brennie, line 10: Why not `EXPERIMENT_SLUG = "firefox-fenix-test-experiment"`? See my other comment about fixture magic.

experimenter/tests/integration/nimbus/android/ping_server/ping_server.py
```diff
@@ -0,0 +1,48 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
```
brennie, line 5: Can we add a module docstring explaining what this is? I think its a dummy telemetry ingestion service?

jaredlockhart (commented):
This is awesome @jrbenny35 🎉 🎉 🎉 🎉

experimenter/tests/integration/nimbus/conftest.py
jaredlockhart: @brennie That's definitely true but they're a well established pattern in our integration tests and there's no reason to change that now.

experimenter/tests/integration/nimbus/android/conftest.py
jaredlockhart: I don't think we've used type annotations in any of our existing integration tests?

experimenter/tests/integration/nimbus/android/moz-central.test-container.Dockerfile
```diff
@@ -0,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
jrbenny35: It's being used in the current weekly job to build the fenix test APKs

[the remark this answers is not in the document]
jaredlockhart: Right, you mean this?

https://github.com/mozilla/experimenter/pull/10844/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R637

That runs from a container that's already built and pushed to Dockerhub, but where do we run `docker build` from that Dockerfile?

experimenter/tests/integration/nimbus/android/adbrun.py
jrbenny35: This is actually not being used so ill delete it. More of a carry-over from the mozilla central workflow.

experimenter/tests/integration/nimbus/android/conftest.py
jrbenny35: Soon ™️

.circleci/config.yml
```diff
@@ -584,6 +620,34 @@ jobs:
...
+            docker cp fenix-builder:mozilla-central/mobile/android/fenix/app-fenix-x86_64-debug.apk ./
+            docker commit fenix-builder moz-central-builder
+            docker tag moz-central-builder ${DOCKERHUB_REPO}:experimenter-moz-central
+            docker push ${DOCKERHUB_REPO}:experimenter-moz-central
+          no_output_timeout: 1h
+      - save_cache:
+          key: experimenter-firefox-fenix-apks-{{ checksum "experimenter/tests/integration/nimbus/android/app-fenix-x86_64-debug.apk" }}
+          paths:
+            - experimenter/tests/integration/nimbus/android/app-fenix-debug-androidTest.apk
+            - experimenter/tests/integration/nimbus/android/app-fenix-x86_64-debug.apk
```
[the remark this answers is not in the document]
jrbenny35: Push to dockerhub is just mozzila-central with a `./mach build` completed, so that next time we just do a `hg pull` and `hg update` and we don't have to clone the entire repo. Caching the APKs is for the actual test job.

experimenter/tests/integration/nimbus/android/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+    os.environ["EXP_NAME"] = "fenix"
+
+
+@pytest.fixture(name="setup_experiment")
+def fixture_setup_experiment(
+    run_nimbus_cli_command,
+    delete_telemetry_pings,
+):
+    def _():
+        experiment_json = next(iter(here.glob("**/fixtures/experiment.json")))
```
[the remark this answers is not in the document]
jrbenny35: The problem is starting the fenix app with automation. Espresso won't allow you to just start the app cleanly. I'd have to use Appium, and with Appium it takes over the running adb instance, so I'd have to do some weird juggling of ADB and who controls the emulator at a certain time. Using `nimbus-cli`, it still gives us the telemetry we need in the end.

experimenter/tests/integration/nimbus/conftest.py
jrbenny35: These are pretty simple fixtures really. This change is because the old dict would get setup once pytest loads the `conftest.py` and for these fenix specific tests, it wasn't needed and causing some failures. This allows us to just call this fixture when we need it.

experimenter/tests/integration/nimbus/android/conftest.py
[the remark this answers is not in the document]
jaredlockhart: Okay I see, still another way to do this is to stand up experimenter, create experiment, fetch the json, and inject it with the CLI the way you're doing.  If we hardcode the recipe in here, then we could break something in Experimenter and this test would pass.  We should use the fact that experimenter is present here to ensure that Experimenter and Fenix integration stays intact.  We can do it as a followup since this is already passing self contained.

experimenter/tests/integration/nimbus/android/test_fenix_integration.py
jrbenny35: This is used in another fixture. I was trying to keep the methods and calls similar to how we do it on desktop. No magic here, just returning a string 😃

experimenter/tests/integration/nimbus/android/ping_server/ping_server.py
jrbenny35: Correct, I can do that.

experimenter/tests/integration/nimbus/android/conftest.py
jrbenny35: Would this block the python script from continuing? The ping server just needs to start, I remember trying `check_output` and I couldn't use it because it was blocking and the server has to stay running, but I couldn't start the test suite.

experimenter/tests/integration/nimbus/android/conftest.py
jrbenny35: This is so nice, I did not know how to do this. Thanks for the tip!

experimenter/tests/integration/nimbus/android/test_fenix_integration.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@pytest.mark.generic_test
+def test_experiment_unenrolls_via_studies_toggle(
```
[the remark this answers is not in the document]
jrbenny35: Come on now Jared. its me!

experimenter/tests/integration/nimbus/kinto/test-reports/report.htm
```diff
@@ -0,0 +1,478 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
jrbenny35: Dang, it snuck in, yes.

experimenter/tests/integration/nimbus/android/moz-central.test-container.Dockerfile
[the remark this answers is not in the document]
jrbenny35: We don't. It needed to be run once and pushed as a docker image. Should I delete the dockerfile from the repo?

PR #10853, opened 2024-06-14
chore(nimbus): sort nimbus list filter drop downs

### LINKED ISSUE #10850: Nimbus list order features by (application, slug)
We should order the features drop down by (application, slug).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4623)

jaredlockhart:
Because

* It's easier to scan through drop down lists when their entries are sorted

This commit

* Sorts some of the filter drop downs on the new Nimbus list page
* Some drop downs with a small number of enum choices can be left alone since the enum ordering is often semantic

fixes #10850


brennie (approved):
🚢 it

PR #10854, opened 2024-06-14
bug(nimbus): status counts and filters should match

### LINKED ISSUE #10851: Nimbus list shows 1 experiment in review but the table is empty
The new list page shows Review(1) in the tab headers but when you click the tab there is no experiment.  Probably just filtering the queryset wrong.

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4624)

jaredlockhart:
Because

* We discovered a case where the Review tab counted 1 experiment but there were no experiments in the table
* This is becuase the definition of review for the counts did not match the definition of the review filter
* This uncovered a broader inconsistency around how we handle the review publish statuses in the list page
* By having anything in review appear only in the review tab, the user would experience their live experiment disappearing from the live tab and moving into the review tab, and then moving back to the live tab, which is probably less desirable than just having all live experiments always appear in the live tab and also in the review tab when they're in review

This commit

* Simplifies the status counts and filters by having the draft and live tabs always show all draft and live experiments, including ones in review
* The review tab will show experiments in any publish status other than idle, so that they're still marked as in review while they're being reviewed in remote settings

fixes #10851

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/filtersets.py
+++ b/experimenter/experimenter/nimbus_ui_new/filtersets.py
@@ -266,9 +266,8 @@ class NimbusExperimentFilter(django_filters.FilterSet):
     def filter_status(self, queryset, name, value):
         match value:
             case StatusChoices.REVIEW:
-                return queryset.filter(
-                    status=NimbusExperiment.Status.DRAFT,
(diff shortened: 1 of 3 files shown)


jrbenny35 (approved):
Nice find

PR #10855, opened 2024-06-14
bug(nimbus): Don't provide a default value for KINTO_ADMIN_URL

### LINKED ISSUE #10852: KINTO_ADMIN_URL uses wrong default value
If you set KINTO_HOST but not KINTO_ADMIN_URL, the resulting computed KINTO_ADMIN_URL will strip off the `/v1` and result in the wrong url.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4625)

brennie:
Because:
- We were providing a computed default value for KINTO_ADMIN_URL based on KINTO_HOST, but it was incorrectly using urljoin

this commit:
- removes the computed default, making KINTO_ADMIN_URL a required setting.

Fixes #10852.


jaredlockhart (approved):
Great catch, thnx @brennie 🎉 🎉 🎉 🎉

yashikakhurana (approved):

PR #10934, opened 2024-07-02
chore(nimbus): Fix fenix integration tests to not use circleci cache.

### LINKED ISSUE #10933: Store our fenix APKs in a docker image to allow for builds on forks
Using the circleci on forks in our fenix integration tests doesn't work due to our security policy, we should just use a small docker image to store these for now. We need to look into using taskcluster to build our images.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4656)

jrbenny35:
Because

- Using the circleci cache on forks in our fenix integration tests doesn't work due to our security policy

This commit

- Fixes that by storing the APKs in a small docker image.

Fixes #10933

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -673,7 +673,6 @@ jobs:
           name: Build and upload images
           command: |
             cd experimenter/tests/integration/nimbus/android
-            docker build -t moz-central-builder -f moz-central.test-container.Dockerfile . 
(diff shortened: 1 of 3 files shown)


jrbenny35:
Going to have to add a follow up change to revert the firefox fenix builds from running as well.

jaredlockhart (approved):
Yep, if it works, it works, g/j thnx @jrbenny35 🎉

PR #10952, opened 2024-07-04
feat(nimbus): New QA form on summary page

### LINKED ISSUE #10913: Summary page QA form
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4603)

yashikakhurana:
Because

- QA should be able to edit its status and comment on new summary page in HTMX

This commit

- New QA form on the new summary page

Fixes #10913 
Screenshot 2024-07-04 at 11 24 49 AM
Screenshot 2024-07-04 at 11 24 59 AM
Screenshot 2024-07-04 at 11 25 15 AM


experimenter/experimenter/experiments/forms.py
```diff
@@ -0,0 +1,14 @@
+# forms.py
+
+from django import forms
+
+from .models import NimbusExperiment
```
jaredlockhart, line 5: Ah, we usually use absolute imports everywhere.  I wonder if there's a lint rule I can add for that?

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -114,4 +116,21 @@ def get_context_data(self, **kwargs):
...
+        if form.is_valid():
+            form.save()
+            context = self.get_context_data(object=self.object)
+            return render(request, self.template_name, context)
+        else:
+            # if form is not valid, still prepare context data
+            context = self.get_context_data(object=self.object)
+            context["form"] = form
+            context["qa_edit_mode"] = True
+            return render(request, self.template_name, context)
```
jaredlockhart, line 136: Oh instead of doing this, I'd suggest either

1) Using the [FormMixin](https://github.com/django/django/blob/main/django/views/generic/edit.py#L13)
2) Make it a totally separate generic view for POSTing too

jaredlockhart (changes requested):
Looking good so far but we can improve the view 🙏

experimenter/experimenter/nimbus_ui_new/views.py
yashikakhurana: I think I will perfer approach 2 as it will be easy to maintain for each forms

experimenter/experimenter/experiments/forms.py
yashikakhurana: yeah let me check that too

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -94,24 +97,61 @@ def get(self, *args, **kwargs):
...
+        return self.render_to_response(context)
+
+    def get_success_url(self):
+        return reverse("nimbus-new-detail", kwargs={"slug": self.kwargs["slug"]})
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        experiment = NimbusExperiment.objects.get(slug=self.kwargs["slug"])
+        kwargs["instance"] = experiment
+        return kwargs
```
jaredlockhart, line 157: This is what [UpdateView](https://docs.djangoproject.com/en/5.0/ref/class-based-views/generic-editing/#django.views.generic.edit.UpdateView) is for.  You should be able to just subclass that, override a couple class properties and that's it.  Everything you're doing here is built in.

jaredlockhart (dismissed):

experimenter/experimenter/nimbus_ui_new/views.py
yashikakhurana: my bad

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -94,24 +97,51 @@ def get(self, *args, **kwargs):
...
+
+    def form_invalid(self, form):
+        context = self.get_context_data(form=form)
+        experiment_context = build_experiment_context(self.object)
+        context.update(experiment_context)
+        context["qa_edit_mode"] = True
+        return self.render_to_response(context)
+
+    def get_success_url(self):
+        return reverse("nimbus-new-detail", kwargs={"slug": self.object.slug})
```
jaredlockhart, line 147: Yesssss this is the way 🙏

jaredlockhart (approved):
Okay tested locally, everything looks and works great.  I see what you're doing with fetching the entire summary and swapping out the body, not clear to me if we want to do that or just swap out just the qa status section, but we can just wait and see what it's like when we start adding more functionality to the page.  Anyways this is a great start, thanks @yashikakhurana 🎉 🎉 🎉

PR #10959, opened 2024-07-08
bug(project): install poetry with install script

### LINKED ISSUE #10957: CI error: no such option: --break-system-packages
Looks like our CI is all choked up on this:

```
Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --break-system-packages
```

Let's take that out and see if that clears it up.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4670)

jaredlockhart:
Because

* We recently switched our integration tests to use poetry instead of tox
* We also recently discovered that the most recent version of selenium docker images are buggy on mac
* We rolled back the selenium docker image to a version that works on mac
* This version comes with an old version of pip that's incompatible with the install script we setup to install poetry

This commit

* Uses the official poetry installer script locked to the same version we use in the Experimenter image

fixes #10957

### DIFF
--- a/experimenter/tests/experimenter_legacy_tests.sh
+++ b/experimenter/tests/experimenter_legacy_tests.sh
@@ -5,12 +5,7 @@ set +x
 
 export PATH=$PATH:/home/seluser/.local/bin
 
-sudo apt-get -qqy update && sudo apt-get -qqy install python3-venv python3-pip
-
-# https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-break-system-packages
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
Makes sense to me

PR #10996, opened 2024-07-11
feat(nimbus): add link to matching audiences

### LINKED ISSUE #10816: Add link to audience/summary page with GET parameters for matching targeting
Once we have the new list page, we should add links to the audience page and summary pages that link back to the list page with filters populated with the targeting of the experiment you're on so you can find other experiments that match that targeting.

Why?  If we can reference a past experiment that had the same criteria (except specific version number) - we can:

* what % and expected enrolled was versus how many people enrolled

Why is this useful?

1. DS can use it to see if they are "in the ball park" or reason why calculations don't match
1. We can size the ones that are OK with rougher numbers without needing DS.  Estimate we can save 10-25% of time in sizing for ones that have past similar targeting

Why exclude specific release number from the matching?

1. because something similar a few versions ago is still relevant if we use human logic for release/timing/update throttling
1. human logic needed sometimes to see how relevant the amount enrolled is - between the min version targeted by an experiment, when that version released, when the experiment launched, throttling, and update rate on different platforms.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4593)

jaredlockhart:
Because

* Before we fully automate population sizing, we can use previously launched experiments as a way to estimate population sizes
* To find matching audiences, we can add a link that filters the new list page to experiments that match another experiments audience parameters

This commit

* Adds helpful links to matching audiences to the summary and edit audience pages

fixes #10816
Screenshot 2024-07-11 at 15 31 53
Screenshot 2024-07-11 at 15 31 31

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/types.py
+++ b/experimenter/experimenter/experiments/api/v5/types.py
@@ -490,6 +490,7 @@ class NimbusConfigurationType(graphene.ObjectType):
 
 class NimbusExperimentType(DjangoObjectType):
     application = graphene.NonNull(NimbusExperimentApplicationEnum)
+    audience_url = graphene.NonNull(graphene.String)
     can_archive = graphene.Boolean()
     can_edit = graphene.Boolean()
     can_review = graphene.Boolean()
@@ -584,6 +585,7 @@ class NimbusExperimentType(DjangoObjectType):
         model = NimbusExperiment
         fields = (
             "application",
+            "audience_url",
             "can_archive",
             "can_edit",
             "can_review",
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass
 from decimal import Decimal
 from pathlib import Path
 from typing import Any, Dict, Optional
-from urllib.parse import urljoin
+from urllib.parse import urlencode, urljoin
 from uuid import uuid4
 
 import packaging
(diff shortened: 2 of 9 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -790,6 +790,38 @@ def review_url(self):
...
             )
 
+    @property
+    def audience_url(self):
+        filters = [
+            ("application", self.application),
+        ]
+        if self.channel:
+            filters.append(("channel", self.channel))
+        if self.firefox_min_version:
```
yashikakhurana, line 800: do we want to check max version too?

jaredlockhart: I don't think so, since we just want to know the most recent matching number, and either some version is live and above the minimum so clients are still enrolled and that's the number we care about, or they've passed the max and gone to zero in which case we don't care about those numbers.  If we decide we do want to filter on it we can add it later, but I think this is all we need for now.

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -490,6 +490,7 @@ def resolve_status_update_exempt_fields(self, info):
 
 class NimbusExperimentType(DjangoObjectType):
     application = graphene.NonNull(NimbusExperimentApplicationEnum)
+    audience_url = graphene.String()
```
brennie, line 493: This should be non-null, then we dont need the `audienceUrl!` in TS.
```suggestion
    audience_url = graphene.NonNull(graphene.String)
```

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/index.tsx
```diff
@@ -114,6 +115,13 @@ const PageEditAudience: React.FunctionComponent<RouteComponentProps> = () => {
 
   return (
     <AppLayoutWithExperiment title="Audience" testId="PageEditAudience">
+      <Row className="mb-4">
+        <Col>
+          <a target="_blank" href={experiment.audienceUrl!} rel="noreferrer">
```
brennie, line 120: With the gql type change, dont need the !

experimenter/experimenter/nimbus-ui/src/components/Summary/TableAudience/index.tsx
```diff
@@ -112,7 +112,18 @@ const TableAudience = ({ experiment }: TableAudienceProps) => {
...
   return (
     <Card className="border-left-0 border-right-0 border-bottom-0">
-      <Card.Header as="h5">Audience</Card.Header>
+      <Card.Header>
+        <Row>
+          <Col>
+            <h5>Audience</h5>
+          </Col>
+          <Col className="text-right">
+            <a target="_blank" href={experiment.audienceUrl!} rel="noreferrer">
```
brennie, line 121: Same here.

brennie (approved):

PR #10997, opened 2024-07-11
bug(nimbus): Don't throw in NimbusExperiment.review_url

### LINKED ISSUE #10949: Handle multiple collections as a review error
We want to prevent an experiment from launching that would deploy to multiple collections in Remote Settings.  We check for that [here](https://github.com/mozilla/experimenter/blob/77b403096e0812dababccc1d28d35f58b4c22f8a/experimenter/experimenter/experiments/constants.py#L82-L83) and throw an `AssertionError`.  However this will simply cause all UI interactions to break on a hard GraphQL error like so:

image

Which prevents the user from correcting the error.  This should be displayed as a Review error preventing launch.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4668)

brennie:
Because:

- we serialize the review_url in the graphQL API;
- the review_url isn't always available (e.g., when a multi-feature experiment targets multiple kinto collections for publishing)

this commit:

- adds a new exception type for the multiple kinto collection case; and
- catches that exception in `NimbusExperiment.review_url`.

Fixes #10949.

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -53,6 +53,13 @@ class BucketRandomizationUnit(models.TextChoices):
     USER_ID = "user_id"
 
 
+class TargetingMultipleKintoCollectionsError(Exception):
+    """An error that is raised when an experiment contains multiple features
+    that each would require it to be published to different kinto
+    collections.
+    """
+
+
 @dataclass
 class ApplicationConfig:
     name: str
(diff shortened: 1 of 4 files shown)


jaredlockhart (approved):
Awesome, tested locally, everything works perfectly.  Thanks @brennie 🎉

yashikakhurana (approved):
:shipit:

PR #10998, opened 2024-07-11
feat(cirrus): Nimbus preview flag to return features based on preview experiments

### LINKED ISSUE #10953: Update cirrus to support qa flag
If Cirrus receives the request param 

`?nimbus_preview=true`

then it should fetch the features from the new nimbus web preview collections

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4662)

yashikakhurana:
Because

- QA should be able to test web experiments on the prod of the implementing application

This commit

- Supports Cirrus to handle the `nimbus_preview` flag so that it should return a response based on the preview records

Fixes #10953


Makefile
```diff
@@ -238,7 +238,7 @@ CIRRUS_ENABLE = export CIRRUS=1 &&
 CIRRUS_BLACK_CHECK = black -l 90 --check --diff .
 CIRRUS_BLACK_FIX = black -l 90 .
 CIRRUS_RUFF_CHECK = ruff check .
-CIRRUS_RUFF_FIX = ruff --fix .
+CIRRUS_RUFF_FIX = ruff check --fix .
```
jaredlockhart, line 241: Oh dang I forgot to update this for Cirrus too XDDD whoops sorry

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -18,38 +18,62 @@ class RecipeType(Enum):
 
 
 class RemoteSettings:
-    def __init__(self, sdk: SDK):
+    def __init__(self, sdk: SDK, preview_sdk: SDK):
```
jaredlockhart, line 21: Okay so the first thing I'll say is rather than having `sdk` and `preview_sdk` I'd suggest we prefix both names so it's more clear which is which, maybe something like `live_sdk` and `preview_sdk`?

Makefile
yashikakhurana: no worries :)

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -18,38 +18,62 @@ class RecipeType(Enum):
...
-            logger.info(f"Fetched resources: {data}")
-        else:
-            logger.warning("No recipes found in the response")
+
+    def update_recipes(self, new_recipes: dict[str, list[Any]]) -> None:
+        self.recipes = new_recipes
+        self.sdk_live.set_experiments(json.dumps(self.recipes))
+
+    def update_preview_recipes(self, new_recipes: dict[str, list[Any]]) -> None:
+        self.preview_recipes = new_recipes
```
jaredlockhart, line 78: Oh same thing here, if we're going to store the recipes right on the object for both the live and preview collections we should probably follow teh same naming convention we do for the sdks.

jaredlockhart, line 78: Isn't there some more generic way to structure this rather than having two duplicate hard coded methods?  I know it's only ever going to be the two but it's still a bit clunky.

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -18,38 +18,62 @@ class RecipeType(Enum):
...
-        self.recipes = new_recipes
-        self.sdk.set_experiments(json.dumps(self.recipes))
+    def fetch_recipes(self) -> None:
+        self._fetch_and_update(self.url, self.update_live_recipes)
+        self._fetch_and_update(self.preview_url, self.update_preview_recipes)
 
-    def fetch_recipes(self):
+    def _fetch_and_update(
+        self, url: str, update_method: Callable[[dict[str, Any]], None]
+    ) -> None:
```
yashikakhurana, line 59: @jaredlockhart yeah like i am doing it here

cirrus/server/cirrus/experiment_recipes.py
yashikakhurana: can look into more, or may be can do refactoring later on, what do you think

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -18,38 +18,62 @@ class RecipeType(Enum):
 
 
 class RemoteSettings:
-    def __init__(self, sdk: SDK):
-        self.recipes: dict[str, list[Any]] = {"data": []}
+    def __init__(self, sdk_live: SDK, sdk_preview: SDK):
```
jaredlockhart, line 21: Okay actually, instead of having one RemoteSettings class that has parameters and attributes and methods for handling two collections, can't we just have two instances of RemoteSettings each of which only handles a single url, sdk, recipes, etc?

cirrus/server/cirrus/main.py
```diff
@@ -42,11 +42,15 @@ async def lifespan(app: FastAPI):
...
+    app.state.sdk_live = create_sdk(
         app.state.fml.get_coenrolling_feature_ids(),
         CirrusMetricsHandler(app.state.metrics, app.state.pings),
     )
-    app.state.remote_setting = RemoteSettings(app.state.sdk)
+    app.state.sdk_preview = create_sdk(
+        app.state.fml.get_coenrolling_feature_ids(),
+        CirrusMetricsHandler(app.state.metrics, app.state.pings),
+    )
+    app.state.remote_setting = RemoteSettings(app.state.sdk_live, app.state.sdk_preview)
```
jaredlockhart, line 53: And then here we'll just have two remote settings

cirrus/server/cirrus/main.py
```diff
@@ -149,15 +153,23 @@ class EnrollmentMetricData(NamedTuple):
...
-            experiment_type = app.state.remote_setting.get_recipe_type(experiment_slug)
+            experiment_type = None
+            if nimbus_preview_flag:
+                experiment_type = app.state.remote_setting.get_preview_recipe_type(
+                    experiment_slug
+                )
+            else:
+                experiment_type = app.state.remote_setting.get_live_recipe_type(
+                    experiment_slug
+                )
```
jaredlockhart, line 172: And then this stays the same just calls into two different remote settings instances instead of the same one with two methods.

cirrus/server/cirrus/main.py
```diff
@@ -210,9 +230,14 @@ async def compute_features(request_data: FeatureRequest):
...
-        targeting_context
-    )
+    if nimbus_preview:
+        enrolled_partial_configuration: dict[str, Any] = (
+            app.state.sdk_preview.compute_enrollments(targeting_context)
+        )
+    else:
+        enrolled_partial_configuration: dict[str, Any] = (
+            app.state.sdk_live.compute_enrollments(targeting_context)
+        )
```
jaredlockhart, line 240: Another way to write this is

```py
sdk = app.state.sdk_live
if nimbus_preview:
    sdk = app.state.sdk_preview
enrolled_partial_configuration = sdk.compute_enrollments(targetring_context)
```

cirrus/server/cirrus/main.py
```diff
@@ -194,7 +211,10 @@ def read_root():
 
 
 @app.post("/v1/features/", status_code=status.HTTP_200_OK)
-async def compute_features(request_data: FeatureRequest):
+async def compute_features(
+    request_data: FeatureRequest,
+    nimbus_preview: Optional[bool] = Query(False, alias="nimbus_preview"),
```
jaredlockhart, line 216: Does it have to be optional? Can't it be default `False`?

jaredlockhart (changes requested):

cirrus/README.md
```diff
@@ -187,6 +189,21 @@ Example input:
   }
 }
 ```
+## Optional Query Parameter
+
+`nimbus_preview (boolean)`: Pass this as a query parameter to enable preview mode. When set to true, the endpoint will use the preview experiments to compute enrollments.
+
+Example usage with nimbus_preview query parameter:
+
+```json
```
brennie, line 198: this isnt json, its shell

cirrus/server/cirrus/main.py
```diff
@@ -149,15 +160,19 @@ class EnrollmentMetricData(NamedTuple):
...
 ) -> list[EnrollmentMetricData]:
     events: list[dict[str, Any]] = enrolled_partial_configuration.get("events", [])
     data: list[EnrollmentMetricData] = []
     for event in events:
         if event.get("change") == "Enrollment":
             experiment_slug = event.get("experiment_slug", "")
             branch_slug = event.get("branch_slug", "")
-            experiment_type = app.state.remote_setting.get_recipe_type(experiment_slug)
+            experiment_type = None
+            remote_setting = app.state.remote_setting_live
```
jaredlockhart, line 172: I think `remote_settings` should always be pluralized?

cirrus/server/cirrus/main.py
```diff
@@ -149,15 +160,19 @@ class EnrollmentMetricData(NamedTuple):
...
+            if nimbus_preview_flag:
+                remote_setting = app.state.remote_setting_preview
```
jaredlockhart, line 174: Ya exactly 👍

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -18,38 +17,38 @@ class RecipeType(Enum):
...
-                if experiment.get("isRollout", False):
-                    return RecipeType.ROLLOUT.value
-                else:
-                    return RecipeType.EXPERIMENT.value
-
+                return (
+                    RecipeType.ROLLOUT.value
+                    if experiment.get("isRollout", False)
+                    else RecipeType.EXPERIMENT.value
+                )
```
jaredlockhart, line 35: If you're trying to find the recipe out of the list in recipes you can also just do it like

```py
experiment = {r['slug']: r for r in self.recipes['data']}.get(experiment_slug)
if experiment is not None:
    return ....
```

yashikakhurana:
tested with preview 
Screenshot 2024-07-18 at 8 46 21 PM
Screenshot 2024-07-18 at 8 45 56 PM

cirrus/server/cirrus/main.py
```diff
@@ -210,7 +233,10 @@ async def compute_features(request_data: FeatureRequest):
         "clientId": request_data.client_id,
         "requestContext": request_data.context,
     }
-    enrolled_partial_configuration: dict[str, Any] = app.state.sdk.compute_enrollments(
+    sdk = app.state.sdk_live
+    if nimbus_preview:
+        sdk = app.state.sdk_preview
```
jaredlockhart, line 238: Nice that the whole change just boils down to this.

cirrus/server/cirrus/main.py
```diff
@@ -221,14 +247,16 @@ async def compute_features(request_data: FeatureRequest):
...
     )
 
     return client_feature_configuration
 
 
 async def fetch_schedule_recipes() -> None:
     try:
-        app.state.remote_setting.fetch_recipes()
+        app.state.remote_setting_live.fetch_recipes()
+        app.state.remote_setting_preview.fetch_recipes()
```
jaredlockhart, line 259: Ah you know what I just realized, if we put these in the same exception block, then if live throws, preview will never be refreshed.  Might wanna split these into separate try blocks and add tests to ensure that if one throws the other still fetches.

jaredlockhart (dismissed):
This all looks right, ty for testing the local case and putting the screenshots.  I think everything is looking good except I just realized about putting both refreshes in the same try block.  Split those, add tests, and this should be good to go.

cirrus/server/cirrus/main.py
yashikakhurana: yup good point

yashikakhurana:
> This all looks right, ty for testing the local case and putting the screenshots. I think everything is looking good except I just realized about putting both refreshes in the same try block. Split those, add tests, and this should be good to go.

done!

jaredlockhart (approved):
Okay great job @yashikakhurana let's get this going!! 🎉 🎉 🎉 🎉

PR #10999, opened 2024-07-11
chore(nimbus): keep create experiment button right on small displays

### LINKED ISSUE #10977: Create button folds under tabs on new list page
The create button is folding under the tabs when the width decreases on the new list page, let's have it go above the tabs so they stay connected to the table.

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4673)

jaredlockhart:
Because

* It would be nice if the Create Experiment on the new list page stayed separate from the tabs when the viewport goes below a certain size

This commit

* Uses bootstrap classes to keep the Create Experiment button separate from tabs on the new list page when the viewport shrinks

fixes #10977

image
image


yashikakhurana (approved):
:shipit:

PR #11012, opened 2024-07-15
feat(nimbus): remove feature config and version from audience url

### LINKED ISSUE #11011: Remove feature and version from matching population links
After using matching population link in production, it's clear we don't need to include the features at all since they shouldn't affect the enrolled population.  We should probably also remove version since the likelihood of matching that is small.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4698)

jaredlockhart:
Because

* After using the new matching audience link it's clear we don't need feature config or version in there since neither will affect the sizing

This commit

* Removes feature config and version from the audience urls

fixes #11011

### DIFF
--- a/experimenter/experimenter/experiments/tests/test_models.py
+++ b/experimenter/experimenter/experiments/tests/test_models.py
@@ -1868,12 +1868,6 @@ class TestNimbusExperiment(TestCase):
             )
 
     def test_audience_url(self):
-        feature1 = NimbusFeatureConfigFactory.create(
-            application=NimbusExperiment.Application.DESKTOP, slug="a"
-        )
-        feature2 = NimbusFeatureConfigFactory.create(
-            application=NimbusExperiment.Application.DESKTOP, slug="b"
-        )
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -1893,8 +1885,7 @@ def test_audience_url(self):
         self.assertEqual(
             experiment.audience_url,
             (
-                "/nimbus/?application=firefox-desktop&channel=release&firefox_min_version=100.%21"
-                f"&feature_configs={feature1.id}&feature_configs={feature2.id}"
+                "/nimbus/?application=firefox-desktop&channel=release"
```
brennie, line 1888: I guess dict insertion order = iteration order makes this a stable URL?

jaredlockhart: Oh it's not a dict, it's a list of tuples, so you can have keys repeated.

PR #11031, opened 2024-07-16
Fix fenix apk update script

### LINKED ISSUE #11030: Update fenix apk decision script
Looks like we are using the wrong taskcluster path for current fenix builds. This needs to be updated as we aren't triggering new fenix apk builds based on the current changes.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4700)

jrbenny35:
Because

- Our current script uses the wrong API path for querying current fenix beta builds

This commit

- Changes to the correct path that is currently being used

Fixes #11030


jaredlockhart (approved):
LGTM 👍

PR #11037, opened 2024-07-16
chore(nimbus): Fix fenix apk updater circleci logic.

### LINKED ISSUE #11036: Fenix apk PR creation job has incorrect logic
The logic in our circleci job to create a PR to update the fenix apks is wrong. We should fix this!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4702)

jrbenny35:
Because

- The logic in the circleci job to create a PR updating our fenix APKs was incorrect. The recent changes triggered correctly but the if statement was wrong and didn't allow the PR to be created

This commit

- Fixes this logic by just having 1 condition when it checks if there is a new build on taskcluster,

Fixes #11036


jaredlockhart (approved):
Aha!  Let's try it!  Thnx @jrbenny35 🙏

PR #11043, opened 2024-07-17
bug(nimbus): Rename ja-JP-mac locale to ja-JP-macos

### LINKED ISSUE #11041: ja-JP-mac locale should be ja-JP-macos
The `ja-JP-mac` locale code is incorrect and no longer used by Firefox. Instead it uses `ja-JP-macos`. The locale was removed in #6265. It should not have been removed, becuase the maOSc-specific Japanese locale is different from regular ja / ja-JP because of OS-specific terminology.

* A migration should be added to rename ja-JP-mac to ja-JP-macos.
* The ja-JP-macos locale should be added to the locale fixtures in place of ja-JP-mac

brennie:
Because:

- Firefox reports its macOS-specific locale as `ja-JP-macos`;
- Experimenter exposes a `ja-JP-mac` locale code in stage and prod; and
- the `ja-JP-mac` locale fixture was removed, but the Japanese macOS locale is a specific variant from a targeting point of view;

this commit:

- adds a locale fixture for `ja-JP-macos` named "Japanese (macOS)" so it is clear that it is different from the regular `ja` locale; and
- adds a migration to rename the `ja-JP-mac` locale to `ja-JP-macos` and change its name to "Japanese (macOS)".

Fixes #11041.

### DIFF
--- a/experimenter/experimenter/base/migrations/0005_jajpmacos.py
+++ b/experimenter/experimenter/base/migrations/0005_jajpmacos.py
@@ -1,0 +1,19 @@
+from django.db import migrations
+
+
+def update_ja_jp_mac_locale(apps, schema_editor):
+    Locale = apps.get_model("base", "Locale")
+
+    Locale.objects.filter(code="ja-JP-mac").update(
+        code="ja-JP-macos", name="Japanese (macOS)"
+    )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("base", "0004_language"),
+    ]
+
+    operations = [
+        migrations.RunPython(update_ja_jp_mac_locale),
+    ]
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/base/migrations/0005_jajpmacos.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+
+def update_ja_jp_mac_locale(apps, schema_editor):
+    Locale = apps.get_model("base", "Locale")
+
+    Locale.objects.filter(code="ja-JP-mac").update(
+        code="ja-JP-macos",
+        name="Japanese (macOS)"
```
jaredlockhart, line 9: Ah good yes ty 🙏

jaredlockhart (approved):
The triumphant return of Japanese mac.  Thnx @brennie 🙏 🇯🇵 🍎

PR #11046, opened 2024-07-17
chore(nimbus): Hard code experimenter docker repo in fenix integration tests.

### LINKED ISSUE #11045: Hard code experimenter dockerhub repo name in fenix integration test job.
We want to support forks but environment variables aren't shared with forks due to security polices. Because of this, the fenix integration test job was failing due to not having access to an environment variable. We should change this and just supply the repo which is `mozilla/experimenter`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4706)

jrbenny35:
Because

- We want to support forks but environment variables aren't shared with forks due to security polices. Because of this, the fenix integration test job was failing due to not having access to an environment variable.

This commit

- Hard code's experimenter docker hub repo `mozilla/experimenter` within the fenix test job.

Fixes #11045

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -415,7 +415,7 @@ jobs:
       - run:
           name: Install APKs
           command: |
-            docker run -d --name fenix-files ${DOCKERHUB_REPO}:fenix-apk-store
+            docker run -d --name fenix-files mozilla/experimenter:fenix-apk-store
             docker cp fenix-files:/usr/src/app/files /home/circleci/project
             adb install /home/circleci/project/files/app-fenix-x86_64-debug.apk
             adb install /home/circleci/project/files/app-fenix-debug-androidTest.apk


.circleci/config.yml
```diff
@@ -415,7 +415,7 @@ jobs:
       - run:
           name: Install APKs
           command: |
-            docker run -d --name fenix-files ${DOCKERHUB_REPO}:fenix-apk-store
+            docker run -d --name fenix-files mozilla/experimenter:fenix-apk-store
```
jaredlockhart, line 418: Oh it doens't look like we even need to `run` here since I think that's basically just triggering a pull and then exiting, you could just write `docker pull` here, I think.

jaredlockhart (approved):

.circleci/config.yml
jrbenny35: Will the `cp` do the run? I think I need to run it before I copy from it.

jaredlockhart: `cp` might only work if the image is already available locally, so you'll prolly need to do SOMETHING to get the image locally available, whether it's run with implicit fetch or explicit fetch.

jrbenny35: What would you like here, I am not help to either solution

jaredlockhart: Just leave it alone for now, if it works it's fine, we can change it later if we need to 👍

PR #11048, opened 2024-07-17
chore(nimbus): reduce number of remote settings integration tests

### LINKED ISSUE #11047: Reduce number of remote settings integration tests in CI
We have a series of integration tests for remote settings that test each sequence of launching, rejecting, etc.  Early in the project we only ran this suite against Firefox Desktop and discovered that a bug was preventing launching an experiment to Fenix that hadn't been caught because we weren't running the RS integration suite for all applications.  Since then we've run the entire RS suite for all applications to make sure every application is launchable.  However as the number of applications grows, the CI cost grows significantly to.  We only really need to test a single RS integration case for each application to ensure the configuration between Experimenter and RS is intact for each application/collection, so let's collapse our RS CI runs down to two tasks:
* Test launching for every application
* Run the full RS suite for a single application

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4707)

jaredlockhart:
Becuase

* We want to run at least one remote settings integration test for every app to make sure all the configuration and code is intact for every application
* We don't need to check every remote settings case for every application
* Reducing the number of remote settings tests will free up some CI space to add fun new tests

This commit

* Collapses the remote settings integration tests down to two tasks
* One for launching every application
* One for the entire suite for a single application

fixes #11047

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,7 +186,7 @@ jobs:
       - store_artifacts:
           path: ~/experimenter/experimenter/tests/integration/test-reports/report.htm
 
-  integration_nimbus_desktop_remote_settings:
+  integration_nimbus_remote_settings_launch:
     machine:
       image: ubuntu-2004:2023.10.1
       docker_layer_caching: true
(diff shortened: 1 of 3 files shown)


brennie (approved):
Nice cleanup!

jrbenny35 (approved):
Woooooooo 🎉 🥳

PR #11093, opened 2024-07-26
chore(nimbus): exclude MSIX targeting

### LINKED ISSUE #11092: Add an all-but-MSIX targeting
The default agent can currently set OS values, but there are circumstances under which it cannot determine what is currently configured on the OS level for MSIX. While this is not fully understood or implemented we want a way to start experiments without users with such installations.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4725)

MaxAtMozilla:
Because

* We need a way to experiment with non-MSIX installations

This commit

* Basically clones the WINDOWS_10_PLUS and excludes MSIX

Fixes #11092


jaredlockhart (approved):
LGTM thanks @MaxAtMozilla 🎉

PR #11096, opened 2024-07-26
feat(nimbus): Only allow prefFlips feature on 128 ESR

### LINKED ISSUE #11086: Prevent launching prefFlips on 128 release
We want to prevent accidental launching of prefFlips to 128 release because they will not function, even though the feature is technically available in v128. Different channels can have different manifests. The prefFlips feature will be in 128ESR but it is not present in 128 release outside of a placeholder feature. When we start syncing 128esr manifests, those will overwrite the 128 release manifests. Thankfully, the manifest are identical at this point, but that could change with uplifts.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4723)

brennie:
Because:

- the prefFlips feature will be uplifted to esr128; and
- it is not implemented on v128 of other Firefox channels

this commit:

- adds a check in the serializer to limit the prefFlips feature to v129+ on all Desktop channels and v128 of ESR.

Fixes #11086.

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -2102,6 +2102,41 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
 
         return data
 
+    def _validate_desktop_pref_flips(self, data):
+        if not any(
+            fc.slug == NimbusConstants.DESKTOP_PREFFLIPS_SLUG
+            for fc in data.get("feature_configs", [])
+        ):
+            return data
+
+        min_version = NimbusExperiment.Version.parse(data["firefox_min_version"])
+        channel = data.get("channel")
+
+        if not channel:
+            raise serializers.ValidationError(
+                {
+                    "channel": NimbusConstants.ERROR_DESKTOP_PREFFLIPS_CHANNEL_REQUIRED,
+                }
+            )
+
+        # We have already validated that the feature is supported by the current
+        # version. However, we don't keep track of per-channel manifests. The
+        # prefFlips feature landed in 128 Nightly as a placeholder and was only
+        # implemented in 129 Nightly+, but it was uplifted to 128 ESR.
+        if channel == NimbusExperiment.Channel.ESR:
+            return data
+
+        elif min_version.major == 128:
+            raise serializers.ValidationError(
+                {
+                    "firefox_min_version": (
+                        NimbusConstants.ERROR_DESKTOP_PREFFLIPS_128_ESR_ONLY
+                    )
+                }
+            )
+
+        return data
+
     def validate(self, data):
         if self.instance.status == self.instance.Status.DRAFT:
             application = data.get("application")
@@ -2121,6 +2156,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
             data = self._validate_proposed_release_date(data)
             if application == NimbusExperiment.Application.DESKTOP:
                 data = self._validate_desktop_pref_rollouts(data)
+                data = self._validate_desktop_pref_flips(data)
             else:
                 data = self._validate_languages_versions(data)
                 data = self._validate_countries_versions(data)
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -2102,6 +2102,41 @@ def _validate_desktop_pref_rollouts(self, data):
...
+                    "channel": NimbusConstants.ERROR_DESKTOP_PREFFLIPS_CHANNEL_REQUIRED,
+                }
+            )
+
+        # We have already validated that the feature is supported by the current
+        # version. However, we don't keep track of per-channel manifests. The
+        # prefFlips feature landed in 128 Nightly as a placeholder and was only
+        # implemented in 129 Nightly+, but it was uplifted to 128 ESR.
+        if channel == NimbusExperiment.Channel.ESR:
+            return data
```
jaredlockhart, line 2127: Like we allow any version on channel ESR?

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -2957,6 +2957,104 @@ def test_desktop_large_prefs(
...
+            NimbusExperiment.Channel.ESR,
+            NimbusExperiment.Channel.AURORA,
+        )
+    )
+    def test_desktop_prefflips_129(self, channel):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            targeting_config_slug=NimbusExperiment.TargetingConfig.NO_TARGETING,
+            application=NimbusExperiment.Application.DESKTOP,
+            firefox_min_version=NimbusExperiment.Version.FIREFOX_128,
```
jaredlockhart, line 3042: We should try to have the test names be almost like sentences describing waht they're testing.  Like this is testing 'test that desktop pref flips are valid for 129'?  It just says 'test desktop pref flips 129' but what about them? That they're valid? Invalid?  The test name here refers to 129, but the version in the test itself is 128, and then at the bottom asserts it's valid, for all channels, so I'm not sure what's expected here?  The ticket says it should only be valid for 128 in ESR, but that's not what this test is saying?

experimenter/experimenter/experiments/api/v5/serializers.py
brennie: If we're hiting this point, we've already validated that the feature is available for this version. I'll add a comment.

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
brennie: Yeah that was supposed to be v129, which is why all the tests are failing :) bady copy+paste

experimenter/experimenter/experiments/api/v5/serializers.py
brennie: Oh. there already is a comment there.

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -2102,6 +2102,41 @@ def _validate_desktop_pref_rollouts(self, data):
...
+
+        elif min_version.major == 128:
```
jaredlockhart, line 2129: And then we hit here and know that channel is ESR *and* version is 128 then we bail out.  It was just a bit tricky to follow cuz there's three things going on here

1) The feature manifest validation might be sufficient
2) ESR for 128 is allowed
3) Any other channel for 129 is allowed

And it just wasn't clear to me the combinations and relationships of all of those.

jaredlockhart (approved):
Okay cool thank you, the tests are much clearer now.  Looks good let's ship it 🎉 thnx @brennie

PR #11115, opened 2024-07-29
chore(nimbus): Update fenix integration test.

### LINKED ISSUE #11114: XFail android studies toggle integration test
We should xfail the test `test_experiment_unenrolls_via_studies_toggle` due to there being some lag within the emulator. I will look into this and try another method of refreshing the app.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4730)

jrbenny35:
Because

- The `test_experiment_unenrolls_via_studies_toggle` test can sometimes fail due to the emulator lagging and not updating as quickly as it should

This commit

- xfails this test so that it doesn't cause too many false failures. I will look into this test and make some changes but for now this will work.

Fixes #11114

### DIFF
--- a/experimenter/tests/integration/nimbus/android/test_fenix_integration.py
+++ b/experimenter/tests/integration/nimbus/android/test_fenix_integration.py
@@ -11,6 +11,7 @@ def experiment_slug():
 
 
 @pytest.mark.generic_test
+@pytest.mark.xfail(reason="Timeouts due to system lag")
 def test_experiment_unenrolls_via_studies_toggle(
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Interesting.  What does `xfail` do?

jrbenny35:
> Interesting. What does `xfail` do?

If it fails the test it wont fail the entire test suite, so it will report as green but will notify us of the failure.

https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

PR #11129, opened 2024-07-31
feat(cirrus): Fetch resources if url is provided

### LINKED ISSUE #11123: Cirrus preview url fetch errors
Currently if preview URL is empty, it is giving us error in SENTRY, instead if no preview URL is provided then we should not try to fetch and if someone calls with preview flag enable we should tell them that this Cirrus doesn’t support preview mode. Also if someone tries to start server with remote settings URL it should fail and let the user know that it needs that url

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4734)

yashikakhurana:
Because

- Sentry reports many errors if the remote settings preview URL is not set.

This commit

- Will not fetch from preview URL if it's not provided
- Will exit the server if remote setting live URL is not provided

Fixes #11123

### DIFF
--- a/cirrus/server/cirrus/main.py
+++ b/cirrus/server/cirrus/main.py
@@ -91,7 +91,7 @@ def initialize_sentry():
 
 
 def verify_settings():
-    if remote_setting_url == "":
+    if not remote_setting_url:
         logger.error("Remote setting URL is required but not provided.")
         sys.exit(1)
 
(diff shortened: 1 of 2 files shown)


cirrus/server/cirrus/main.py
```diff
@@ -254,20 +266,18 @@ async def compute_features(
 
 
 async def fetch_schedule_recipes() -> None:
-    live_failed = False
-    preview_failed = False
+    live_failed, preview_failed = False, False
```
brennie, line 269: There's no reason to use tuple-assignment here.

brennie (approved):

PR #11188, opened 2024-08-15
chore(ci): Add proper branch filtering for recipe generator.

### LINKED ISSUE #11157: Mobile integration tests run when they shouldn't on circleci
I noticed that for a couple PRs the mobile integration tests ran even though there was no change within the `experimenter/` directory. We should fix that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4745)

### LINKED ISSUE #11158: Firefox for Android and iOS recipe creation job on circleci runs on main after merging
We need to add a filter [here](https://github.com/mozilla/experimenter/blob/main/.circleci/config.yml#L725-L726) so this job doesn't run on main but only on PRs.

Example: https://app.circleci.com/pipelines/github/mozilla/experimenter/42606

b4handjr:
Because

- We are running an integration test on main that doesn't need to be run

This commit

- Adds the proper filtering to that job in the circleci config

Fixes #11158

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -723,6 +723,10 @@ workflows:
                 - main
       - create_fenix_fennec_recipes:
           name: Create fenix and fennec recipes
+          filters:
+            branches:
+              ignore:
+                - main
       - integration_nimbus_desktop_release_targeting:
           name: Test Desktop Targeting (Release Firefox)
           filters:


jaredlockhart (approved):
👍

PR #11193, opened 2024-08-19
chore(project): switch to STORAGES setting for storage providers

### LINKED ISSUE #11192: Switch to STORAGES in settings
We were using `get_storage_class` to switch between different storage providers, but it was [removed in 5.1](https://docs.djangoproject.com/en/5.1/releases/5.1/#features-removed-in-5-1).  So we need to use the new [STORAGES](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-STORAGES) setting to manage multiple storage providers.  Let's migrate to that so we can update to Django 5.1.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4751)

jaredlockhart:
Because

* We need to use different storage providers for local dev, tests, and in prod
* We used to configure this by using a custom function that called get_storage_class
* However get_storage_class is depcated in Django 5.1
* We need to switch to using the STORAGES setting

This commit

* Removes the old storage setting code
* Adds STORAGES to settings.py
* Updates tests as necessary

fixes #11192


experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -598,12 +594,16 @@ def read(self):
...
             return File(filename)
 
-        mock_open.side_effect = open_file
-        mock_exists.return_value = True
+        with (
+            patch("experimenter.jetstream.client.default_storage.open") as mock_open,
+            patch("experimenter.jetstream.client.default_storage.exists") as mock_exists,
+        ):
+            mock_open.side_effect = open_file
+            mock_exists.return_value = True
```
brennie, line 602: You can provide patch() with the function to replace it with / the return value

```suggestion
         with (
            patch("experimenter.jetstream.client.default_storage.open", open_file),
            patch("experimenter.jetstream.client.default_storage.exists", side_effect=True)
        ):
```

experimenter/experimenter/settings.py
```diff
@@ -487,15 +487,31 @@
 KINTO_REVIEW_TIMEOUT = config("KINTO_REVIEW_TIMEOUT", cast=int)
 
 # Jetstream GCS Bucket data
-DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
+# DEFAULT_FILE_STORAGE = "storages.backends.gcloud.GoogleCloudStorage"
```
brennie, line 490: Should we remove this line or leave it commented?

brennie (approved):

experimenter/experimenter/settings.py
jaredlockhart: Oops!

experimenter/experimenter/jetstream/tests/test_tasks.py
jaredlockhart: Oh yeah good suggestions.  I'll just leave it as it is for now, but this is good to know in future.

PR #11203, opened 2024-08-20
chore(ci): Add file path check to mobile recipe generator job.

### LINKED ISSUE #11157: Mobile integration tests run when they shouldn't on circleci

b4handjr:
Because

- This job could run on changes to other parts of the repo when it shouldn't

This commit

- Fixes it to run on the correct change paths. This will be fixed soon with the change to how we test external dependencies.

Fixes #11157

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -413,6 +413,8 @@ jobs:
       docker_layer_caching: true
     steps:
       - checkout
+      - check_file_paths:
+          paths: "experimenter/"
       - run:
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Very nice 👍

PR #11208, opened 2024-08-20
chore(nimbus): Remove file caching from Dockerfile

### LINKED ISSUE #11207: Remove file caching from experimenter Dockerfile
The file caching was added to share layers with builds done in the cloud across potentially different operating systems. We are no longer doing that and this step may be a cause of unnecessary rebuilds of containers when unrelated files change. We should remove it and simplify our build.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4752)

brennie:
Because:

- file caching was added in #5948 (see #5946) to enable remote caching;
- we are no longer using remote caching;
- this step adds complexity to our build process; and
- this step may be triggering unnecessary container rebuilds

this commit:

- removes file caching.

Fixes #11207.

### DIFF
--- a/experimenter/Dockerfile
+++ b/experimenter/Dockerfile
@@ -1,17 +1,4 @@
 #-------------------------
-FROM alpine:3.12.0 as file-loader
-
-# To preserve layer caching across machines which may have different local file properties
-# such as permissions, timestamps, etc, all files are copied into a container and their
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Bye bye file loader 👋 thnx @brennie

PR #11224, opened 2024-08-23
feat(nimbus): add group_id randomization unit

### LINKED ISSUE #11220: Add group ID to recipe in experimenter
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4675)

jaredlockhart:
Because

* Desktop will be migrating towards using group_id as the randomization unit
* Before all clients are migrated over, we can run some tests on a small number of experiments using group_id

This commit

* Adds a flag to experiment to use group_id as the randomization unit
* Adds group_id to the bucket namespace
* Sets the randomization unit of an isolation group to group_id if it includes experiments set to use it

fixes #11220


yashikakhurana (approved):
It looks good, question-will you add a flag in UI in the upcoming PR?

jaredlockhart:
@yashikakhurana No we won't add it to the UI at all, we'll just set it from the admin.

experimenter/experimenter/experiments/constants.py
```diff
@@ -51,6 +51,7 @@ class BucketRandomizationUnit(models.TextChoices):
     NORMANDY = "normandy_id"
     NIMBUS = "nimbus_id"
     USER_ID = "user_id"
+    GROUP_ID = "group_id"
```
mikewilli, line 54: I know this was merged already, but any chance we could call this `profile_group_id` to match what the column is called? I'm not sure if Nimbus cares if it matches an exact field somewhere else, but it would allow me to match against the [AnalysisUnit enum](https://github.com/mozilla/metric-hub/blob/main/lib/metric-config-parser/metric_config_parser/__init__.py#L4) we have (something like 
```py
analysis_unit = AnalysisUnit.PROFILE_GROUP_ID if bucket_config.randomization_unit == AnalysisUnit.PROFILE_GROUP_ID.value else AnalysisUnit.CLIENT_ID
```
).

jaredlockhart: Do you mean the enum key or the string constant value?  The value is determined by the [client here](https://searchfox.org/mozilla-central/source/toolkit/components/nimbus/lib/ExperimentManager.sys.mjs#445) so it has to be `group_id`.

mikewilli: I was hoping to match up the string constant with the enum (and for consistency across everything from client to tables), even though I recognize they are different things. Not a big deal thoug, I'll just look for `group_id` instead, and maybe I can get (or create?) a different enum with the allowed randomization unit values.

Thanks!

PR #11225, opened 2024-08-24
adding new advanced targeting: WINDOWS_10_MSIX_ONLY

erchen03:
Because a msix-only targeting does not exists right now,
This commit adds the relevant targeting to be used for the msix start menu pinning experiment


jaredlockhart (approved):
Looks good thnx @erchen03 👍

PR #11233, opened 2024-08-26
chore(project): Change fenix integration tests to run off of upstream trigger.

### LINKED ISSUE #11230: Trigger Fenix (Android) enrollment and UI tests based on fenix beta builds
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4763)

b4handjr:
Because

- We want to run certain jobs based on external dependency changes as they happen.

This commit

- Changes the fenix integration and enrollment tests to run when a change has been detected within fenix.

Fixes #11230


b4handjr:
Here is an example build before I changed the branches, https://app.circleci.com/pipelines/github/mozilla/experimenter/43412/workflows/21b04f64-6864-4f92-a62b-d4cc2bd85e6e

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
```diff
@@ -5,21 +5,25 @@
...
-@pytest.fixture
-def default_data_api(default_data_api):
-    feature_config_id = helpers.get_feature_id_as_string(
-        "messaging", BaseExperimentApplications.FIREFOX_IOS.value
-    )
+def test_create_mobile_experiment_for_integration_test(
+    selenium, experiment_url, kinto_client, default_data_api, experiment_slug, application
+):
+    """Create a mobile experiment for device integration tests"""
+    apps = ["IOS", "FENIX"]
```
jaredlockhart, line 22: Should this be a 'params' fixture here?

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
```diff
@@ -5,21 +5,25 @@
...
+
 from nimbus.pages.experimenter.summary import SummaryPage
 from nimbus.utils import helpers
 
 
 @pytest.fixture
-def experiment_slug():
-    return "firefox-ios-integration-test"
+def experiment_slug(application):
+    return f"firefox-{application.lower()}-integration-test-experiment"
```
jaredlockhart, line 15: How come we hard code the app specific slugs above, but then also do it here as parameterized?

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
```diff
@@ -5,21 +5,25 @@
...
-        "messaging", BaseExperimentApplications.FIREFOX_IOS.value
-    )
+def test_create_mobile_experiment_for_integration_test(
+    selenium, experiment_url, kinto_client, default_data_api, experiment_slug, application
+):
+    """Create a mobile experiment for device integration tests"""
+    apps = ["IOS", "FENIX"]
+    logging.info(application)
+    if str(application) not in apps:
+        pytest.skip()
```
jaredlockhart, line 25: Is this just for debugging?

experimenter/tests/integration/nimbus/utils/create_experiment_via_api.py
```diff
@@ -0,0 +1,60 @@
+import sys
```
jaredlockhart, line 1: Don't we already have all this, the default data and the feature config maps, in conftest?

jaredlockhart (changes requested):

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
b4handjr: No this is used for the `pytest.skip()` in the following lines.

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
b4handjr: Mainly because this "test" runs in a different scope with a different pytest `conftest.py`. So when the actual test for say ios runs in `tests/integration/nimbus/ios`, it doesn't have knowledge of the slug for use with `nimbus-cli` (we have to provide the slug to enroll). With this way its just a set slug. 

If you think we should keep the parametrized slug, I could set it as an env variable and just load that into the `nimbus-cli` call. That would give us the ability to rerun on CI.

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
b4handjr: No, we don't need to create recipes for anything other than ios, or fenix. So this just skips the test if the `application` is neither one of them.

experimenter/tests/integration/nimbus/utils/create_experiment_via_api.py
b4handjr: Oh this shouldn't have been added, whoops!

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
jaredlockhart: Oh like, require the parameterized application fixture, which creates an instance for every application, but skip all but those two.  I had to deal with this recently, there was a way to do it with just the decorators.  I can't remember.  Anyways that makes sense.

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
jaredlockhart: How come it runs in a different scope?  There isn't a way to share the same conftest?

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
b4handjr: Yeah I could do it with decorators

experimenter/tests/integration/nimbus/test_create_mobile_experiment.py
b4handjr: For android, it needs a bunch of android specific files, `grandle.py`, being one of them, and I wanted to keep them out of the normal `desktop` test configs. I guess I could put all of the conftest stuff in the test file itself so as to not have an additional `conftest.py`

b4handjr:
Okay I removed the downstream conftest in the `/android` folder, so now we are just using 1 conftest and any overrides are within the test file itself.

jaredlockhart (approved):
Alright! Let's give it a shot!!  Thnx @b4handjr you earned a 🧁 +281 −1,608

b4handjr:
> Alright! Let's give it a shot!! Thnx @b4handjr you earned a 🧁 +281 −1,608

I love cupcakes! Once everything passes ill make the final changes to the circleci config

PR #11243, opened 2024-08-27
feat(project): Retrieve metric-hub segments in Experimenter

### LINKED ISSUE #11242: Retrieve metric-hub segments in Experimenter
Experimenter needs to retrieve the available segments from metric-hub as they are currently not available there.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4748)

RJAK11:
Because

* Experimenter needs to retrieve the segments from metric-hub as they are currently not available there

This commit

* allows Experimenter to retrieve segments from metric-hub

Fixes #11242


experimenter/experimenter/settings.py
```diff
@@ -530,6 +530,14 @@
     BASE_DIR / "outcomes" / "metric-hub-main" / "jetstream" / "outcomes"
 )
 
+JETSTREAM_CONFIG_SEGMENTS_PATH = (
```
mikewilli, line 533: So the `JETSTREAM_CONFIG` part of the other name is actually a relic of pre-metric-hub config repo, which was called jetstream-config. Maybe it's time to update these settings to say `METRIC_HUB_` instead of `JETSTREAM_CONFIG_`? This would make all three of these related settings consistent, and update them for the new repo.

Something like:
- `METRIC_HUB_OUTCOMES_PATH`
- `METRIC_HUB_SEGMENTS_PATH_DEFAULT`
- `METRIC_HUB_SEGMENTS_PATH_JETSTREAM`

experimenter/experimenter/segments/tests/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+
+from django.test import override_settings
+
+mock_valid_segments = override_settings(
+    JETSTREAM_CONFIG_SEGMENTS_PATH=(
+        Path(__file__).parent.absolute() / "fixtures" / "valid_segments"
+    ),
+    DEFINITIONS_CONFIG_SEGMENTS_PATH=(
+        Path(__file__).parent.absolute() / "fixtures" / "valid_segments_definitions"
+    ),
```
mikewilli, line 11: Along the same lines as my suggestion to rename these settings, I'd suggest changing this path to `valid_segments_jetstream` as well.

experimenter/experimenter/segments/tests/fixtures/valid_segments_definitions/firefox_desktop.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[segments.desktop_segment_2]
+friendly_name = "Desktop Segment 2"
+description = "Firefox desktop segment used for testing"
+select_expression="country_code in ('IN', 'US')"
+
+[segments.desktop_segment_3]
+friendly_name = "Desktop Segment 3"
+description = "Firefox desktop segment used for testing"
+select_expression="country_code in ('IN', 'US')"
```
mikewilli, line 14: It might be more valuable to configure these slightly differently so that the test covers more cases. For example, if some fields aren't required, you can test that by leaving some out in segment 2 and others in segment 3 (or add a couple new ones) and make sure the tests still pass. It looks like you'll need to tweak the tests to handle this change too.

experimenter/experimenter/segments/tests/test_segments.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@mock_valid_segments
+class TestCheckSegmentTOMLs(TestCase):
+    def setUp(self):
+        Segments.clear_cache()
+
+    def test_valid_segments_do_not_trigger_check_error(self):
+        errors = check_segment_tomls(None)
+        self.assertEqual(errors, [])
```
mikewilli, line 90: Consider combining these into a single class with two functions. The setup is the same for each so splitting them isn't necessary (and duplicates code), and I'd say the name `TestCheckSegmentTOMLs` describes the behavior of both tests pretty well. You can move the decorator to the functions instead of the classes.

mikewilli (changes requested):
This is great!! The comments I left are all pretty minor -- take a look and let me know if you have any questions or concerns with what I've suggested.

experimenter/experimenter/segments/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+from experimenter.experiments.constants import NimbusConstants
+
+
+@dataclass
+class Segment:
+    slug: str
+    friendly_name: str
+    application: str
+    description: str
+    select_expression: str
```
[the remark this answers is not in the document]
mikewilli: Filed tickets from these two comments, resolving...

mikewilli (approved):
LGTM great work!! :tada: :shipit:

PR #11245, opened 2024-08-27
chore(schemas): create RandomizationUnit enum

### LINKED ISSUE #11244: Add Randomization Units Enum to schemas
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4764)

mikewilli:
Because

- randomization unit is configurable
- new profile mgmt feature requires jetstream to behave differently depending on the randomization unit
- randomization unit options are defined in Experimenter where Jetstream can't use them
- we don't want to redefined the options in multiple places

This commit

- adds randomization unit enum to schemas package so it can be used by both Experimenter and Jetstream

Fixes #11244


jaredlockhart (approved):
Perf thnx @mikewilli 👍

mikewilli:
Added the `use_enum_values` config so re-requested review. I don't know if it's actually necessary (depends on how we use the BucketConfig model), but I think it should be safer given that we changed it from a `str` (so if it's serialized it will still be a string).

PR #11267, opened 2024-08-30
chore(schemas): use python 3.10 in schemas package

### LINKED ISSUE #11266: Use Python 3.10 in schemas for compatibility with analysis
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4774)

mikewilli:
Because

- analysis tooling is using python 3.10
- schemas package is using 3.11
- analysis tooling relies on schemas package

This commit

- downgrades schemas package to bring them in sync on python 3.10

Fixes #11266


schemas/Dockerfile
```diff
@@ -13,17 +13,17 @@ RUN chmod -R 555 /schemas \
 
 #-------------------------
 # System packages
-FROM python:3.11.1 AS system-builder
+FROM python:3.10.14 AS system-builder
```
jaredlockhart, line 16: 😭 😭 😭 😭 😭 😭 😭 😭

jaredlockhart (approved):
Yep looks okay to me, thnx @mikewilli 🙏

schemas/Dockerfile
mikewilli: 14 is bigger than 1 so it's an upgrade.

PR #11283, opened 2024-09-04
feat(project): Add segments to Experiment model

### LINKED ISSUE #11276: Add segments to Experiment model
We need to adjust the Experiment model by adding support for segments.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4775)

RJAK11:
Because

* We need to adjust the Experiment model to add support for segments.

This commit

* Adds support for segments in the Experiment model.

Fixes #11276


experimenter/experimenter/experiments/tests/test_change_formatting_utils.py
```diff
@@ -284,6 +284,16 @@ def test_formatting_for_array_fields(self):
             experiment, self.user, "primary outcomes were added", timestamp_2
         )
 
+        old_value_segments = []
+        new_value_segments = ["test-segment"]
```
mikewilli, line 288: I can't comment on outcome lines, but can you update the outcome vars on lines 277-278 to make them more clear now that there are segments in here too? `old_value` --> `old_value_outcomes` for example

mikewilli: And same for the other vars that are currently generic but should be distinguished from the segments vars.

experimenter/experimenter/experiments/tests/test_change_formatting_utils.py
```diff
@@ -315,6 +325,42 @@ def test_formatting_for_array_fields(self):
 
         self.assertDictEqual(change, expected_change)
 
+        field_name_segments = "segments"
+        field_diff_segments = {
+            "old_value": old_value_segments,
+            "new_value": new_value_segments,
+        }
+        change_timestamp_segments = self._create_formatted_timestamp(
+            comparison_log.changed_on
```
mikewilli, line 334: I'm surprised that this works for both segments and outcomes. Maybe it's a strange side effect of using the same timestamp for both changelogs, that both changes get lumped together into one changelog? Is that what's happening here? I would have expected `comparison_log` to contain the info about the outcomes change but not the segments.

mikewilli (changes requested):
Nice work! And sorry about the advice on ignoring the admin form for now, I assume django needs a field on the admin page for each field on the models. Now we know!

I just have a couple questions and minor changes to the one test, otherwise looks great!

mikewilli (approved):
LGTM! :shipit:

PR #11284, opened 2024-09-04
Adding msix first run advanced targeting

erchen03:
Because 
Requiring a targeting specifically for MSIX Firefox users on their first launch of the application

This commit
Adds that targeting


jaredlockhart (approved):
Looks good, thanks @erchen03 👍

PR #11289, opened 2024-09-05
feat(nimbus): add filter reset link to filters on new list page

### LINKED ISSUE #11248: Add filter clear button on list page
We need a button to clear all filters on the list page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4766)

jaredlockhart:
Because

* We need a way to clear the filters on the list page

This commit

* Adds a filter reset link to the bottom of the filters on the new list page

fixes #11248

image

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/list.html
+++ b/experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/list.html
@@ -20,6 +20,11 @@
         hx-target="#experiment-list"
         hx-push-url="true">
     {% for field in filter.form %}{{ field }}{% endfor %}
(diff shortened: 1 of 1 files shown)


mikewilli (commented):
It looks like this just navigates back to the base homepage URL? Is there a way to reset the filters instead so that other things on the page stay fixed (e.g., if you're on the Completed tab, refreshing will reset that).

jaredlockhart:
@mikewilli In theory yes but that's more complicated and unless people explicitly ask for it I don't think it's worth the time investment.

mikewilli (approved):

mikewilli:
Fair enough, looks good to me then!

PR #11300, opened 2024-09-06
feat(project): create v8 api for segments support

### LINKED ISSUE #11290: Create v8 api for segments support
Create a new v8 api based on the v6 api to support adding segments to the existing experiment model.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4784)

RJAK11:
Because

* We need to extend experiments to support segment functionality
* We want to be able to update the API for Jetstream without affecting the clients that consume v6

This commit

* Creates a new v8 api based on the existing v6 api to support adding segments

Fixes #11290

### DIFF
--- a/experimenter/experimenter/experiments/api/v8/views.py
+++ b/experimenter/experimenter/experiments/api/v8/views.py
@@ -52,7 +52,7 @@ class NimbusExperimentViewSet(
     filter_backends = [DjangoFilterBackend]
     filterset_class = NimbusExperimentFilterSet
 
-    @method_decorator(cache_page(settings.V8_API_CACHE_DURATION))
+    @method_decorator(cache_page(settings.API_CACHE_DURATION))
     def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)
 
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -371,8 +371,7 @@ CACHES = {
         "TIMEOUT": None,
     },
 }
-V6_API_CACHE_DURATION = 60 * 60
-V8_API_CACHE_DURATION = 60 * 60
+API_CACHE_DURATION = 60 * 60
 SIZING_DATA_KEY = "population_sizing"
 
 # Celery
--- a/experimenter/experimenter/experiments/api/v6/views.py
+++ b/experimenter/experimenter/experiments/api/v6/views.py
@@ -52,7 +52,7 @@ class NimbusExperimentViewSet(
     filter_backends = [DjangoFilterBackend]
     filterset_class = NimbusExperimentFilterSet
 
-    @method_decorator(cache_page(settings.V6_API_CACHE_DURATION))
+    @method_decorator(cache_page(settings.API_CACHE_DURATION))
     def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)
 
--- a/experimenter/experimenter/experiments/api/v7/views.py
+++ b/experimenter/experimenter/experiments/api/v7/views.py
@@ -46,6 +46,6 @@ class NimbusExperimentViewSet(
     filter_backends = [DjangoFilterBackend]
     filterset_class = NimbusExperimentFilterSet
 
-    @method_decorator(cache_page(settings.V6_API_CACHE_DURATION))
+    @method_decorator(cache_page(settings.API_CACHE_DURATION))
     def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)


mikewilli:
Another thing for the "Because" list:
- we want to be able to update the API for Jetstream without affecting the clients that consume v6

experimenter/experimenter/experiments/api/v8/urls.py
```diff
@@ -0,0 +1,21 @@
...
+    NimbusExperimentViewSet,
+)
+
+router = SimpleRouter()
+router.register(r"experiments", NimbusExperimentViewSet, "nimbus-experiment-rest-v8")
+router.register(
+    r"experiments-first-run",
+    NimbusExperimentFirstRunViewSet,
+    "nimbus-experiment-rest-v8-first-run",
+)
```
mikewilli, line 15: I think we can safely omit this from v8, first-run isn't used by jetstream as far as I know.
```suggestion
```

experimenter/experimenter/experiments/tests/api/v8/test_views.py
```diff
@@ -0,0 +1,306 @@
...
+            )
+
+            if experiment.status == NimbusExperiment.Status.DRAFT:
+                expected_slugs.append(experiment.slug)
+
+        response = self.client.get(reverse(self.LIST_VIEW))
+        self.assert_returned_slugs(response, expected_slugs)
+
+
+class TestNimbusExperimentFirstRunViewSet(NimbusExperimentFilterMixin, CachedViewSetTest):
```
mikewilli, line 279: If we delete the first-run endpoint we can also remove this test class.

experimenter/experimenter/settings.py
```diff
@@ -203,6 +203,11 @@
     "nimbus-experiment-rest-v6-draft-detail",
     "nimbus-experiment-rest-v7-list",
     "nimbus-experiment-rest-v7-detail",
+    "nimbus-experiment-rest-v8-list",
+    "nimbus-experiment-rest-v8-detail",
+    "nimbus-experiment-rest-v8-first-run-list",
```
mikewilli, line 208: This can also be deleted along with other first-run endpoint stuff.

experimenter/experimenter/settings.py
```diff
@@ -203,6 +203,11 @@
...
+    "nimbus-experiment-rest-v8-draft-list",
+    "nimbus-experiment-rest-v8-draft-detail",
```
jaredlockhart, line 210: We can probably scrap these in V8 since these are for the mobile first run flow, which jetstream shouldn't care about @mikewilli ?

mikewilli (commented):
I'll get back to finish the review later, but wanted to leave you some feedback to get started on.

- the first-run endpoints aren't relevant for Jetstream so that seems like a safe removal for this setup
- it'd be cool if the v8 endpoint could be in a section of the swagger docs that indicates it's public and for analysis tooling (Nimbus: Public Analysis maybe?). Take a look and let me know if you'd like some assistance

experimenter/experimenter/experiments/api/v8/views.py
```diff
@@ -0,0 +1,78 @@
...
+    queryset = (
+        NimbusExperiment.objects.with_related()
+        .exclude(status__in=[NimbusExperiment.Status.DRAFT])
+        .order_by("slug")
+    )
+    serializer_class = NimbusExperimentSerializer
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = NimbusExperimentFilterSet
+
+    @method_decorator(cache_page(settings.V6_API_CACHE_DURATION))
```
mikewilli, line 55: Probably want to create a new setting to use here for V8.

experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
```diff
@@ -0,0 +1,372 @@
...
+                    )
+                ),
+                "endDate": experiment.end_date.isoformat().replace("+00:00", "Z"),
+                "id": experiment.slug,
+                "isEnrollmentPaused": True,
+                "isRollout": False,
+                "proposedDuration": experiment.proposed_duration,
+                "proposedEnrollment": experiment.proposed_enrollment,
+                "referenceBranch": experiment.reference_branch.slug,
+                "schemaVersion": settings.NIMBUS_SCHEMA_VERSION,
```
mikewilli, line 74: We should probably do something different with this setting as well. Maybe create a NIMBUS_ANALYSIS_SCHEMA_VERSION and just set it to 1.0.0 for now. @jaredlockhart any suggestions here?

experimenter/experimenter/experiments/api/v8/serializers.py
```diff
@@ -0,0 +1,194 @@
...
+    class Meta:
+        model = NimbusBranch
+        fields = ("slug", "ratio", "feature", "features")
+
+    def get_feature(self, obj):
+        return {
+            "featureId": "this-is-included-for-mobile-pre-96-support",
+            "enabled": False,  # TODO: Remove after Desktop 104 is no longer supported
+            "value": {},
+        }
```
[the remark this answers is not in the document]
mikewilli: I'm going to slightly revise this -- we should do that eventually, but don't worry about it for now. I think the approach we should take is something like:
1. get new v8 API up basically identical to v6
2. switch jetstream over to v8
3. add segments to v8 API
4. consume segments in jetstream

And then if we have time and see it as a priority relative to other possibilities, we can worry about trimming down v8 API to remove things jetstream doesn't care about.

mikewilli (approved):
r+wc 

This looks great! I'd request that you update the settings for the few things I commented on before merging, but go ahead and merge after that.

experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
jaredlockhart: Oh yeah that's there solely for client behaviour, probably has no meaning here.  Having a locked schema like that is only useful for supporting legacy clients to prevent breakage.  I don't expect we'd need anything similar for jetstream since we control both sides and there's no case where we'd have 'legacy jetstreams' running we're not able to update.

mikewilli: Ok, I'd say we just leave the existing setting alone then for now. We can rip it out entirely later, no reason to create another useless setting in the meantime.

experimenter/experimenter/experiments/api/v8/views.py
jaredlockhart: Either a new setting or just take v6 out of the name and use it for both ¯\_(ツ)_/¯

PR #11304, opened 2024-09-06
chore(schemas): update to pydantic v2

### LINKED ISSUE #11214: Update schemas pydantic to v2
We're on a now-quite-old version of pydantic, and some external dependencies on the data side are requesting that we update

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4755)

mikewilli:
Because

- the schemas package is used by a lot of data tooling
- the data tooling is starting to complain about schemas requiring pydantic v1 due to other dependencies
- the pydantic-to-typescript package is outdated and does not support pydantic v2

This commit

- updates the schemas package to v2
- changes the typescript generation to a script internal to experimenter

Fixes #11214


schemas/generate_json_schema.py
```diff
@@ -0,0 +1,134 @@
+"""
```
jaredlockhart, line 1: Eeeeehhhhhh I wish we didn't have to do this.  There's not something else that works off the shelf?

mikewilli: So yea, I agree. I think there may be some options, but nothing that jumped out at me when doing some initial searching. I also know Beth was interested in pursuing something like Zod and I don't really have the time to evaluate things like that at the moment. This is becoming more of an issue for data so I saw this as a bit of a stopgap, but taking inspiration from tooling we already use.

jaredlockhart: Okey doke fair enough.  Let's do it for now to unblock things and maybe we can find a better solution in the future.  Thanks for taking this on @mikewilli 🙏

jaredlockhart (approved):
Awesome ty @mikewilli 🙏

PR #11329, opened 2024-09-09
feat(nimbus): add no sizing data message to audience page

### LINKED ISSUE #10126: Improvements to the pre-computed pop sizing UI on Audience page
Right now, it's mostly invisible (somewhat intentionally) unless you know the targeting combinations that will make it appear. We should work to make this more visible. A few options (worth thinking if there are other/better options too):
- don't hide the population sizing behind the current parameters
  - allow separate selection of sizing info (instead of selecting all the parameters via audience/targeting, select the specific sizing combo from the list of pre-computed options)
  - collapse all sizing info by default (mostly because it takes up a lot of the screen)
- show a note indicating why sizing info is not displayed
  - something like "Sorry, no pre-computed sizing info is available for the selected targeting."
  - possible including info about what targeting has pre-computed sizing available

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4259)

mikewilli:
Because

- we don't always have pre-computed sizing information, and we only show it when we have it
- inconsistent UI can be confusing
- low number of pre-computed targets mean the feature may not be well-known

This commit

- adds a note about how to find pre-computed sizing information if we don't have it available for the selected targeting

Fixes #10126 

**Screenshots**
image

No sizing data available
image

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -280,29 +280,6 @@ NO_FEATURE_SLUG = [
 ]
 
 
-class ApplicationNameMap(models.TextChoices):
-    DESKTOP = (APPLICATION_CONFIG_DESKTOP.slug, APPLICATION_CONFIG_DESKTOP.app_name)
-    FENIX = (APPLICATION_CONFIG_FENIX.slug, APPLICATION_CONFIG_FENIX.app_name)
-    IOS = (APPLICATION_CONFIG_IOS.slug, APPLICATION_CONFIG_IOS.app_name)
-    FOCUS_ANDROID = (
-        APPLICATION_CONFIG_FOCUS_ANDROID.slug,
-        APPLICATION_CONFIG_FOCUS_ANDROID.app_name,
-    )
-    KLAR_ANDROID = (
-        APPLICATION_CONFIG_KLAR_ANDROID.slug,
-        APPLICATION_CONFIG_KLAR_ANDROID.app_name,
-    )
-    FOCUS_IOS = (APPLICATION_CONFIG_FOCUS_IOS.slug, APPLICATION_CONFIG_FOCUS_IOS.app_name)
-    KLAR_IOS = (APPLICATION_CONFIG_KLAR_IOS.slug, APPLICATION_CONFIG_KLAR_IOS.app_name)
-    MONITOR = (
-        APPLICATION_CONFIG_MONITOR_WEB.slug,
-        APPLICATION_CONFIG_MONITOR_WEB.app_name,
-    )
-    VPN = (APPLICATION_CONFIG_VPN_WEB.slug, APPLICATION_CONFIG_VPN_WEB.app_name)
-    FXA = (APPLICATION_CONFIG_FXA_WEB.slug, APPLICATION_CONFIG_FXA_WEB.app_name)
-    DEMO_APP = (APPLICATION_CONFIG_DEMO_APP.slug, APPLICATION_CONFIG_DEMO_APP.app_name)
-
-
 class Application(models.TextChoices):
     DESKTOP = (APPLICATION_CONFIG_DESKTOP.slug, APPLICATION_CONFIG_DESKTOP.name)
     FENIX = (APPLICATION_CONFIG_FENIX.slug, APPLICATION_CONFIG_FENIX.name)
@@ -381,8 +358,6 @@ class NimbusConstants:
 
     Application = Application
 
-    ApplicationNameMap = ApplicationNameMap
-
     class Type(models.TextChoices):
         EXPERIMENT = "Experiment"
         ROLLOUT = "Rollout"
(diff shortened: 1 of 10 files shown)


yashikakhurana:
do you want to add docs on experimenter.info and link that here as well?

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -86,6 +88,20 @@ export const MOBILE_APPLICATIONS = [
...
+  IOS: "firefox_ios",
+  DEMO_APP: "NOT_SUPPORTED",
+  FOCUS_ANDROID: "focus_android",
+  FOCUS_IOS: "focus_ios",
+  FXA: "NOT_SUPPORTED",
+  KLAR_ANDROID: "klar_android",
+  KLAR_IOS: "klar_ios",
+  MONITOR: "monitor_cirrus",
+  VPN: "NOT_SUPPORTED",
+};
```
jaredlockhart, line 103: This kind of thing should really come from the backend instead of being hard coded here.  I'd say it doesn't matter if we rip out the old frontend soon anyways but I suspect we will keep some of the data specific parts for at least some period of time so we should do these parts properly.  This information can all be gathered in the backend as constants and shared through the config endpoint.

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
```diff
@@ -982,6 +1015,33 @@ export const FormAudience = ({
...
+              criteria. See below for target combinations with sizing available
+              for the current application (
+              {APP_NAME_TO_SLUG_MAP[experiment.application]}):
+            </p>
+            <p>
+              {getSizingAvailableTargets.length > 0 ? (
+                getSizingAvailableTargets.map((target) => (
+                  <Code
+                    codeString={JSON.stringify(target, (k, v) =>
+                      k === "new_or_existing" || k === "app_id" || v === null
```
jaredlockhart, line 1033:
```js
["new_or_existing", "app_id", null].includes(k)
```

jaredlockhart (changes requested):

experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.tsx
mikewilli: Neat! Is it still worth doing this with it being `v === null` and not `k`? i.e., 
```ts
["new_or_existing", "app_id"].includes(k) || v === null)
```

mikewilli:
>do you want to add docs on experimenter.info and link that here as well?

@yashikakhurana Great idea! I added a link to the docs page about the tool (https://experimenter.info/auto-sizing-cli/) in a tooltip.

experimenter/experimenter/experiments/constants.py
```diff
@@ -280,6 +280,29 @@ def kinto_collections(self) -> set[str]:
...
+    )
+    FOCUS_IOS = (APPLICATION_CONFIG_FOCUS_IOS.slug, APPLICATION_CONFIG_FOCUS_IOS.app_name)
+    KLAR_IOS = (APPLICATION_CONFIG_KLAR_IOS.slug, APPLICATION_CONFIG_KLAR_IOS.app_name)
+    MONITOR = (
+        APPLICATION_CONFIG_MONITOR_WEB.slug,
+        APPLICATION_CONFIG_MONITOR_WEB.app_name,
+    )
+    VPN = (APPLICATION_CONFIG_VPN_WEB.slug, APPLICATION_CONFIG_VPN_WEB.app_name)
+    FXA = (APPLICATION_CONFIG_FXA_WEB.slug, APPLICATION_CONFIG_FXA_WEB.app_name)
+    DEMO_APP = (APPLICATION_CONFIG_DEMO_APP.slug, APPLICATION_CONFIG_DEMO_APP.app_name)
```
jaredlockhart, line 303: Can we not just dynamically derive this?

```
ApplicationNameMap = models.TextChoices([(a.slug, a.app_name) for a in .....])
```

jaredlockhart (changes requested):

experimenter/experimenter/experiments/constants.py
mikewilli: Sure, I was just following the example of the Application class. I'll update this though.

mikewilli: Oh wait, the issue here is that each one is derived from a different ApplicationConfig. There is no list of all of them, so either I can keep it the way it is, or create a new list of all of them. Preference?

jaredlockhart: Ya it's here:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/constants.py#L382-L394

jaredlockhart (approved):
Okay very nice looks good 👍 Thnx @mikewilli 🙏

PR #11333, opened 2024-09-09
chore(schemas): upgrade twine to fix publishing error

### LINKED ISSUE #11331: Fix schemas PyPI publishing
Getting a CI error (KeyError 'license') that appears to be due to a combination of `twine` removing `license` from the metadata, and `importlib-metadata` moving to treating implicit Nones as KeyErrors. This is apparently fixed in twine 5.1.1 so we should upgrade.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4791)

mikewilli:
Because

- last publish of schemas package to PyPI failed
- failure appears to be related to a bug documented in twine that was fixed in the latest version

This commit

- upgrades to the latest 5.1.1 version
- bumps version of schemas so it will be published

Fixes #11331


mikewilli:
cc @scholtzan

jaredlockhart (approved):
Good catch 👍

PR #11358, opened 2024-09-12
Advanced targeting config: Set PDF handler eligibility

### LINKED ISSUE #11357: Advanced targeting config: Set PDF handler eligibility
We need to add two advanced targeting configs to support the upcoming [PDF Annotations experiment](https://docs.google.com/document/d/1KZObLp8oRLuJYJXZKciF6kgw_32R9lcRmWX6EAwcjHI/edit#heading=h.tagczcke62ij). The targeting parameters are like this:

1. Users that are eligible to set Firefox as the default PDF handler and have yet to do so. This excludes users with enterprise policies, as well as users that have set their default PDF handler to an app other than a browser (like a PDF editor).
2. Everybody else.

Two experiments will be made, one for each cohort, and the messaging will differ for each cohort. We will offer to set Firefox as the default PDF handler, but only to the first cohort that's eligible. That offer will be omitted for the second cohort.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4800)

aminomancer:
Fixes #11357

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1943,6 +1943,46 @@ DEFAULT_PDF_IS_DIFFERENT_BROWSER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+ELIGIBLE_FOR_DEFAULT_PDF_HANDLER = NimbusTargetingConfig(
+    name="Eligible for default PDF handler",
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Very nice, looks good to me, thanks @aminomancer 🎉

aminomancer:
Thank you!

PR #11382, opened 2024-09-17
fix(dockerfile): Run as non-root.

mikaeld:
Because

- Running the application as root inside the container can pose security risks, and it's best practice to run containers as a non-root user. GCPv2 security requirements proscribes running containers as root.

This commit

- Modifies the Dockerfile to create and use a non-root user with UID 10001.
- Updates ownership of necessary directories and files to ensure the non-root user can access and modify them.
- Switches the running user to the non-root user `appuser` after environment setup.

# Related PR
https://github.com/mozilla-it/webservices-infra/pull/2741


experimenter/Dockerfile
```diff
@@ -116,14 +115,18 @@ RUN yarn workspace @experimenter/nimbus_ui_new build
...
+
 WORKDIR /experimenter
 EXPOSE 7001
 
 # Disable python pyc files
 ENV PYTHONDONTWRITEBYTECODE 1
 
 # Add poetry to path
-ENV PATH "/root/.poetry/bin:${PATH}"
+ENV PATH "/home/appuser/.poetry/bin:${PATH}"
```
mikaeld, line 129: Is poetry actually required in the final stage?

jaredlockhart: Hrmm good question.  It shouldn't be, let me confirm.

jaredlockhart:
@mikaeld Looks like we have the same hard coded celery paths in the prod infra as well: 

https://github.com/mozilla-services/cloudops-infra/blob/ed5ac24f6c744c3d5f9b941bf58217f1f606a833/projects/experimenter/k8s/charts/experimenter/templates/deployment-beat.yaml#L47-L52

We'll have to remove those before we merge this so celery can start after we merge this.  Do you want to update that there?

mikaeld:
@jaredlockhart  Thanks for helping on this! I will file a PR to change the hard coded path in `cloudops-infra` and we should be good to deploy tomorrow!

yashikakhurana (approved):
Looks good

PR #11387, opened 2024-09-18
chore(ci): Fix branch name in external integration script.

### LINKED ISSUE #11363: Incorrect branch name in external integration updater script
The branch that gets created within this script is wrong and there are circleci jobs using it to trigger their runs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4801)

b4handjr:
Because

- The branch that was being created did not match the branch being pushed within our external integration script

This commit

- Fixes the branch names

Fixes #11363

### DIFF
--- a/scripts/external_integration_updater_script.sh
+++ b/scripts/external_integration_updater_script.sh
@@ -18,7 +18,7 @@ if (($(git status --porcelain | wc -c) > 0)); then
     git add .
     git commit -m "chore(nimbus): Check external firefox integrations and update keys"
     git push origin -f check_external_firefox_integrations
-    gh pr create -t "chore(nimbus): Check external firefox integrations and update keys" -b "" --base main --head check-external-firefox-integrations --repo mozilla/experimenter || echo "PR already exists, skipping"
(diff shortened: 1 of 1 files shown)


jaredlockhart (changes requested):
Here too?

https://github.com/mozilla/experimenter/pull/11387/files#diff-db4680c904a681460c8d4f067239f6db8c61db304a3de8fc7bc574065cd04249R5

Also, whenever I need to rename something like this I use the global search replace in my editor so it replaces it everywhere in the entire project, maybe that would help?

b4handjr:
> Here too?
> 
> https://github.com/mozilla/experimenter/pull/11387/files#diff-db4680c904a681460c8d4f067239f6db8c61db304a3de8fc7bc574065cd04249R5
> 
> Also, whenever I need to rename something like this I use the global search replace in my editor so it replaces it everywhere in the entire project, maybe that would help?

Yeah I work with 3 split windows on vscode and sometimes I just miss seeing stuff... but ill keep this in mind. Thank you.

jaredlockhart (approved):
👍

PR #11410, opened 2024-09-20
feat(project): Add segments to v8 api

### LINKED ISSUE #11401: Add segments to v8 api
The current v8 api does not support the functionality to handle segments so it needs to be adjusted.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4814)

RJAK11:
Because
* The current v8 API does not support segments

This commit
* Adds segments to the v8 API for consumption by jetstream

Fixes #11401

### DIFF
--- a/experimenter/experimenter/experiments/api/v8/serializers.py
+++ b/experimenter/experimenter/experiments/api/v8/serializers.py
@@ -99,6 +99,7 @@ class NimbusExperimentSerializer(serializers.ModelSerializer):
     featureIds = serializers.SerializerMethodField()
     probeSets = serializers.ReadOnlyField(default=[])
     outcomes = serializers.SerializerMethodField()
+    segments = serializers.SerializerMethodField()
     startDate = serializers.DateField(source="start_date")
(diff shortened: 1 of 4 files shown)


experimenter/experimenter/experiments/api/v8/serializers.py
```diff
@@ -179,6 +181,9 @@ def get_outcomes(self, obj):
             for slug in outcomes
         ]
 
+    def get_segments(self, obj):
+        return [{"slug": slug} for slug in obj.segments]
```
jaredlockhart, line 185: I was going to say we don't need to wrap it in a `{'slug': ....}` object since it's already only a slug, but then I noticed it's just mirroring the same pattern/interface for outcomes above, and theoretically this lets us add additional keys if we need to later so actually probably just fine to leave it like this.

jaredlockhart (approved):
Nice clean simple change, very nice 10/10 thnx @rahulsainani 🎉

PR #11412, opened 2024-09-20
chore(ci): Modify integration test script to install firefox desktop from stored key.

### LINKED ISSUE #11411: Add steps to install firefox release using the stored version key
Follows on #11402, we just need to modify the `nimbus_integration_tests.sh` script to use the version key stored.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4818)

b4handjr:
Because

- We added the capability to track and store what version of firefox release is current for use in our integration tests

This commit

- Follows on from that to modify the script so that it will download and install that specific version

Fixes #11411


jaredlockhart (approved):
👍

PR #11464, opened 2024-10-01
fix(nimbus): don't log erroneous error

### LINKED ISSUE #11463: Erroneous analysis error appears on results page before experiment is ended
Seeing an error message like:
```Could not find data in analysis bucket at path ....overall.json```

This is a scary looking error that doesn't actually tell us anything because it is happening for experiments which are not ended and so we don't expect overall results yet. We should filter these out somehow so they don't appear on the UI.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4836)

### PR DESCRIPTION
**Pull request by mikewilli**
Because

- we added error logging to catch cases of misconfigured results bucket
- the error logging shows up on the UI in cases where it's not a real error (e.g., the experiment doesn't have final results, but the experiment hasn't ended yet)

This commit

- filters out errors about overall results if the experiment hasn't yet ended

Fixes #11463

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -1,6 +1,6 @@
 import json
 from collections import defaultdict
-from datetime import datetime
+from datetime import date, datetime
 from itertools import chain
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: experimenter/experimenter/jetstream/client.py [hunk 1]
```diff
@@ -335,15 +335,21 @@ def get_experiment_data(experiment: NimbusExperiment):
...
-        analysis_error = AnalysisError(
-            experiment=experiment.slug,
-            filename="experimenter/jetstream/client.py",
-            func_name="load_data_from_gcs",
-            log_level="ERROR",
-            message=e,
-            timestamp=datetime.now(),
-        )
-        errors_experiment_overall.append(analysis_error.dict())
+        # only store runtime errors for overall window if we expect those results
```
jaredlockhart, line 338: No test? 🤔

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -2168,29 +2168,32 @@ def test_results_data_null(self, lifecycle):
...
-                },
             ]
+            if lifecycle == NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE:
+                experiment_errors.append(
+                    {
+                        "analysis_basis": None,
+                        "source": None,
+                        "exception": None,
+                        "exception_type": None,
+                        "experiment": f"{experiment.slug}",
```
brennie, line 2186: nit: you don't need to fstring here, or in any of ther other locations

```suggestion
                        "experiment": experiment.slug,
```

brennie (approved):

experimenter/experimenter/jetstream/tests/test_tasks.py
mikewilli: Ah yea good point. Feels like a copy paste remnant.. Thanks!

PR #11515, opened 2024-10-08
feat(schemas): Align the experiments schemas with nimbus-shared

### LINKED ISSUE #11481: Parity between new schemas package and nimbus-shared
We should update our new schemas package to have parity with the types in nimbus-shared so that we can deprecate nimbus-shared.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4852)

brennie:
Because:
- the generated TypeScript interfaces did not match the TypeScript interfaces from nimbus-shared; and
- all fields and types were missing descriptions

this commit:
- adds descriptions to almost every experiment schema (copied from nimbus-shared and updated as appropriate); and
- aligns the types so that the generated interfaces more closely match the existing interfaces from nimbus-shared.

Fixes #11481

### DIFF
--- a/schemas/index.d.ts
+++ b/schemas/index.d.ts
@@ -5,6 +5,9 @@
 /* Do not modify by hand - update the pydantic models and re-run the script
  */
 
+/**
+ * A unique, stable indentifier for the user used as an input to bucket hashing.
+ */
 export type RandomizationUnit = "normandy_id" | "nimbus_id" | "user_id" | "group_id";
 export type Feature = FeatureWithExposure | FeatureWithoutExposure;
 export type FeatureVariableType = "int" | "string" | "boolean" | "json";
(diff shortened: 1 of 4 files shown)


mikewilli (approved):
Nice! Want to bump the version so it gets published? 

```sh
./script/set_schema_version.sh 2024.10.1
make schemas_version
```

jaredlockhart (changes requested):
Okay so if we're ready to do this now, then in theory we could replicate the test recipes here:

https://github.com/mozilla/nimbus-shared/tree/main/data/experiment-recipe-samples

into this repo, and use those as the baseline to ensure the schema defined here is correct?

brennie:
> Nice! Want to bump the version so it gets published?
> 
> ```shell
> ./script/set_schema_version.sh 2024.10.1
> make schemas_version
> ```

I have another change I want to bundle into this (having the .json.schema files in the generated NPM package for mozilla-central)

brennie:
> Okay so if we're ready to do this now, then in theory we could replicate the test recipes here:
> 
> https://github.com/mozilla/nimbus-shared/tree/main/data/experiment-recipe-samples
> 
> into this repo, and use those as the baseline to ensure the schema defined here is correct?

see-also https://github.com/mozilla/experimenter/tree/main/schemas/mozilla_nimbus_schemas/tests/experiments/fixtures/experiments and https://github.com/mozilla/experimenter/blob/main/schemas/mozilla_nimbus_schemas/tests/experiments/test_experiments.py

but ill update them to add the new publishedDate field

schemas/index.d.ts
```diff
@@ -28,71 +31,279 @@ export type SizingMetricName = "active_hours" | "search_count" | "days_of_use" |
...
+   * If null, it has not yet been published.
+   */
+  publishedDate?: string;
 }
+/**
+ * Bucketing configuration.
+ */
 export interface ExperimentBucketConfig {
   randomizationUnit: RandomizationUnit;
+  /**
```
brennie, line 178: The old schema had randomizationUnit: string https://github.com/mozilla/nimbus-shared/blob/main/types/experiments.ts

That means new older clients would accept things with invalid randomizationUnits up to the point of bucketing

New clients would reject them at recipe validation time.

Before we publish these as schemas for desktop, do we want to change this @jaredlockhart ?

schemas/index.d.ts
```diff
@@ -28,71 +31,279 @@ export type SizingMetricName = "active_hours" | "search_count" | "days_of_use" |
...
+   *
+   * e.g., if branch A=1 and branch B=3, then branch A would get 25% of the population.
+   */
   ratio: number;
-  feature: ExperimentFeatureConfig;
-  features: ExperimentFeatureConfig[];
+  /**
+   * An array of feature configurations.
+   */
+  features: ExperimentFeatureConfig1[];
```
brennie, line 260: Ugh again with the type aliases.

schemas/mozilla_nimbus_schemas/experiments/experiments.py
```diff
@@ -1,84 +1,272 @@
...
+    referenceBranch: str | None = Field(
+        description=(
+            'The slug of the reference branch (i.e., the branch we consider "control").'
+        )
+    )
+    featureValidationOptOut: bool | SkipJsonSchema[None] = Field(
+        description="Opt out of feature schema validation. Only supported on desktop.",
+        default=None,
+    )
+    localizations: dict[str, dict[str, str]] | None = Field(
```
jaredlockhart, line 245: In theory we'd use like TypedDict here to actually specify the shape/keys, if they're known.

jaredlockhart (approved):
Awesome this looks really fantastic.  Thanks @brennie 🎉 🎉 🎉 🎉 🎉 🎉

schemas/index.d.ts
jaredlockhart: Like as an enum with an exact set of keys?

jaredlockhart: Oh I see, old schema it was just string, new schema it is an exact set of constants.  In theory I would've said not to introduce that as a change ie have the new schema mirror as closely as possible the behaviour of the existing schema, and then handle that change separately.  

Here it's already part of the new schema, even though it's not in use or hooked up to anything yet.

It should be fine, we know what the expected values are, we can see them hard coded in MC.  So it should be fine.

brennie: Yes the difference is 
```diff
-randomizationUnit: string;
+randomizationUnit: "normandy_id" | "nimbus_id" | "user_id" | "group_id";
```

brennie: (this will slightly change the behaviour of desktop clients as to when we would reject something with an unsupported randomization unit)

schemas/mozilla_nimbus_schemas/experiments/experiments.py
brennie: The data looks like
```ts
{
  // ...
  localizations: {
    "en-US": { // limited to known locale codes that Firefox uses
      "foo": "foo-text", // these keys are arbitrary
      "bar": "bar-text"
    },
    "fr": {
      "foo": "foo-text en francais",
      "bar": "bar-text en francais",
    }
 }
}
```

so a more strongly typed version would be

```ts
type LocaleKey = "en-US" | "en-CA" | "fr" | ...;
type Localizations = Record<LocaleKey, Record<string, string>>;
```

but I dont think its worth enforcing the locale key correctness at the schema level since it is enforced in the client already

PR #11519, opened 2024-10-08
feat(schemas): generate .schema.json files for Experimenter and Feature Manifest schemas

### LINKED ISSUE #11518: Add JSON schemas to NPM package
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4854)

brennie:
Because:

- we need the .schema.json files in Desktop;
- they are not currently packaged into either the NPM or PyPI packages;
  and
- the generated schemas do not include the custom validation logic
  expressed by Pydantic validators

This commit:

- adds JSON schema validation via Pydantic's json_schema_extra hook to
  give the JSON schemas validation parity with the Pydantic models and
  existing JSON schemas in Desktop;
- generates the JSON schemas and commits them (in schemas/schemas);
- packages the produced schemas into the NPM package;
- packages the produced schemas into the PyPI package (but uncommitted,
  since they are duplicates);
- revises the Makefile build steps for schema package generation; and
- bumps the schema package version.

Fixes #11518


brennie:
Blocked on #11515.

brennie:
I still need to go through this and check for additionalProperties parity with the existing schemas, but its my EOD and I wanted to get this up for posterity.

jaredlockhart (dismissed):
I see what you've done here by having a strict schema just for desktop and then a broader schema to match both desktop and SDK but it seems to me that having two schemas for desktop, ie one stricter and one looser, may allow them to drift in the future where something validates in one or not the other. 

Rather than having a looser schema that handles both desktop and SDK, can we have two strict schemas, one only for desktop, and one only for sdk, and then when we're parsing in experimenter we'll know which one to use because we'll know the path/application for each schema file at parse time?

That way there's only ever one schema used to validate each type of manifest and we can use it in both the integration applications repo/ci/etc, and in experimenter?

brennie:
This needs more unit test coverage

Makefile
```diff
@@ -312,20 +313,17 @@ schemas_format: schemas_docker_build  ## Format schemas source tree
...
+	$(SCHEMAS_RUN) "$(SCHEMAS_GENERATE)"
 
-schemas_dist: schemas_docker_build schemas_dist_pypi schemas_dist_npm
 
-schemas_deploy_pypi: schemas_docker_build
+schemas_deploy_pypi: schemas_build
 	$(SCHEMAS_RUN) "$(SCHEMAS_DEPLOY_PYPI)"
 
-schemas_deploy_npm: schemas_docker_build
+schemas_deploy_npm: schemas_build
```
mikewilli, line 326: Is it always desirable to build before deploy? And if so, is there any particular advantage to enforcing it?

Most importantly, if we do this, we need to update the CI config -- currently it runs `make schemas_dist` and then the deploy steps.

schemas/.gitignore
mikewilli, line 1: We currently put everything into the top level .gitignore (it includes some `experimenter/` and `cirrus/` paths), so maybe worth continuing that trend for visibility? I'm OK with this though if you'd rather this be the start of a new trend.

schemas/pyproject.toml
```diff
@@ -19,6 +22,7 @@ pytest = "^7.3.1"
 twine = "^5.1.1"
```
mikewilli, line 22: Out of scope for this PR, but maybe we should consider using `poetry publish` instead of `twine`?

mikewilli (changes requested):
:tada: LGTM I have many thoughts, most of them are along the lines of "thank you I'm sorry you had to do this". That said, all of the complicated stuff seems to be out of necessity and is well-documented with comments, so appreciate that.

Just need to make sure the Makefile commands are how we want them, and in sync with the CI config and should be good to :shipit:

schemas/pyproject.toml
brennie: https://github.com/mozilla/experimenter/issues/11574

Makefile
brennie: Yes it is desired. We *need* the build step to run to copy the schemas into the python package.

brennie:
@mikewilli https://app.circleci.com/pipelines/github/mozilla/experimenter/45884/workflows/9baddcdd-58c4-4c63-bcfd-c46c176cd147/jobs/246412 Looks good ?

.circleci/config.yml
```diff
@@ -518,26 +518,34 @@ jobs:
...
-          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc
-      - run:
-          name: Upload to NPM
-          command: |
-            # Relies on the NPM_TOKEN environment variable configured at:
-            #   https://app.circleci.com/settings/project/github/mozilla/experimenter/environment-variables
-            make schemas_deploy_npm SCHEMAS_ENV="-e NPM_TOKEN=${NPM_TOKEN}"
+          name: Build the distribution files
+          command: |
+            make schemas_build
```
mikewilli, line 523: Do we still need this as a separate step since it's a dep to the dist now?

mikewilli (approved):
Nice, thanks! Just fix up the CI config be back to non-testing mode and we should be good to go.

.circleci/config.yml
brennie: No probably not :)

PR #11548, opened 2024-10-15
chore(project): Update firefox iOS github branch ID tracking script.

### LINKED ISSUE #11479: External integration updater script is sometimes deleting fennec beta versions
Currently the github api is not returning the fennec (firefox-ios) next beta branch `release/v132` and the script is actually deleting the tracking env id. It looks like the new branch doesn't have protections enabled, and the script is only checking for branches with protections. We might need to adjust this to check both types of branches for the new branch name.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4850)

b4handjr:
Because

- The tracking id Firefox for iOS beta were being deleted due to how they operate their repo. The script was looking under protected branches and a branch is only marked as protected when it is a Release Candidate.

This commit

- Checks both types of branches so we can track beta before it is a Release Candidate.

Fixes #11479


jaredlockhart (approved):
Awesome 🎉 👍

PR #11549, opened 2024-10-15
feat(nimbus): Update mozilla-nimbus-schemas and pydantic

### LINKED ISSUE #11547: Update Experimenter to Pydantic v2
To use the new version of mozilla-nimbus-schemas, we need to also update Experimenter to Pydantic v2.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4875)

brennie:
Because:

- mozilla-nimbus-schemas is on pydantic v2; and
- experimenter is not

this commit:

- updates experimenter to use pydantic v2;
- updates experimenter to use the new version of mozilla-nimbus-schemas.

Fixes #11547


brennie:
This will fail tests because it is not pointing to the updated version of mozilla-nimbus-schemas or pydantic v2 in pyproject.toml, since mozilla-nimbus-schemas wont be published until #11519 lands

experimenter/experimenter/jetstream/models.py
```diff
@@ -329,71 +334,71 @@ def append_conversion_count(self, primary_metrics_set):
...
 
 
+# XXX: does this need @lru_cache ?
 def create_results_object_model(data: JetstreamData):
-    branches = {}
-    metrics = {}
-
-    for jetstream_data_point in data:
-        branches[jetstream_data_point.branch] = {}
+    branches = [data_point.branch for data_point in data]
```
mikewilli, line 339: Should this be a set? I don't think there's any reason to iterate over a list of the full length of `data` twice (which we'd be doing here and then below if `branches` is a list of the same length as `data`).

experimenter/experimenter/jetstream/models.py
```diff
@@ -329,71 +334,71 @@ def append_conversion_count(self, primary_metrics_set):
...
 
     # create a dynamic model with all branches, leveraging BranchComparisonData
-    branches_data = {b: BranchComparisonData() for b in branches}
     PairwiseBranchComparisonData = create_model(
         "PairwiseBranchComparisonData",
-        **branches_data,
+        **{
+            branch: (BranchComparisonData, Field(default_factory=BranchComparisonData))
```
mikewilli, line 345: What's the advantage of declaring a `Field` with `default_factory` vs just setting it to an instance?

experimenter/experimenter/jetstream/models.py
```diff
@@ -159,34 +159,39 @@ def append_retention_data(self, weekly_data):
...
 
 class MetricData(BaseModel):
-    absolute: BranchComparisonData
-    difference: BranchComparisonData
-    relative_uplift: BranchComparisonData
-    significance: SignificanceData
+    absolute: BranchComparisonData = Field(default_factory=BranchComparisonData)
+    difference: BranchComparisonData = Field(default_factory=BranchComparisonData)
+    relative_uplift: BranchComparisonData = Field(default_factory=BranchComparisonData)
+    significance: SignificanceData = Field(default_factory=BranchComparisonData)
```
mikewilli, line 194: Why do these have default values now? Should they?

experimenter/experimenter/jetstream/models.py
```diff
@@ -329,71 +334,71 @@ def append_conversion_count(self, primary_metrics_set):
...
     # Dynamically create our grouped models which are dependent on metrics
     # available for a given experiment
-    SearchData = create_model("SearchData", **search_data)
-    UsageData = create_model("UsageData", **usage_data)
-    OtherData = create_model("OtherData", **other_data)
+    SearchData = create_model("SearchData", **{
+        metric_name: (PairwiseMetricData, Field(default_factory=PairwiseMetricData))
+        for metric_name in metrics
+        if metric_name in SEARCH_METRICS
+    })
```
mikewilli, line 372: I think these are missing the `absolute=BranchComparisonData()` part? (from old line 360)

experimenter/experimenter/jetstream/tests/constants.py
```diff
@@ -1,6 +1,7 @@
 from copy import deepcopy
+from functools import lru_cache
```
mikewilli, line 2: I mentioned this before, but I think `cache` is preferred over `lru_cache` ([link](https://docs.python.org/3/library/functools.html#functools.cache)).

mikewilli (commented):
Thanks for doing this! I left a bunch of questions, but my main concern is about the potentially missing `absolute` field during the results object model creation.

Also just fyi I focused on the changes in the jetstream folder because that's what I'm more familiar with.

experimenter/experimenter/jetstream/models.py
brennie: I think pydantic will magically transform stuff like `x = F()` to `x = Field(default_factory=F)` if F isnt hashable but i think this is more clear as to whats going on.

experimenter/experimenter/jetstream/models.py
brennie: tests were failing because these were not provided.

experimenter/experimenter/jetstream/models.py
brennie: nope, it has a default_factory in the MetricsData class

mikewilli: Ah I see, thanks!

experimenter/experimenter/jetstream/models.py
```diff
@@ -159,34 +159,39 @@ def append_retention_data(self, weekly_data):
...
+    @classmethod
+    def coerce_window_index(cls, v: Any):
+        if isinstance(v, int):
+            return str(v)
+
+        return v
 
     def has_bounds(self) -> bool:
-        return self.lower and self.upper
+        return self.lower and self.upper # TODO: should these be `is not None` checks?
```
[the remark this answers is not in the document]
mikewilli: So thinking about this a bit more.. I'd say yes we should change to `is not None` checks.

Looks like the current evaluation is wrong if `lower` or `upper` is 0, which would be falsey and so the function would erroneously return `False`. Because we have default values of `None` for these fields, the value should either be explicitly `None` or a valid `float`, so checking `is not None` is correct here.

experimenter/experimenter/jetstream/models.py
brennie: You're right, this mistakenly changed set-like behaviour to a list.

mikewilli (approved):
One question about a RootModel subclassing, but otherwise LGTM!

experimenter/experimenter/jetstream/models.py
```diff
@@ -90,35 +90,35 @@ class Config:
         use_enum_values = True
 
 
-class JetstreamData(BaseModel):
+class JetstreamData(RootModel[JetstreamDataPoint]):
```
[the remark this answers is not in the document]
brennie: It was complaining in the type check.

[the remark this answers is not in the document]
mikewilli: Huh, strange. I guess it's working so ¯\_(ツ)_/¯

experimenter/experimenter/jetstream/models.py
```diff
@@ -159,34 +159,39 @@ def append_retention_data(self, weekly_data):
...
 
 class MetricData(BaseModel):
-    absolute: BranchComparisonData
-    difference: BranchComparisonData
-    relative_uplift: BranchComparisonData
-    significance: SignificanceData
+    absolute: BranchComparisonData = Field(default_factory=BranchComparisonData)
+    difference: BranchComparisonData = Field(default_factory=BranchComparisonData)
+    relative_uplift: BranchComparisonData = Field(default_factory=BranchComparisonData)
+    significance: SignificanceData = Field(default_factory=BranchComparisonData)
```
mikewilli, line 194: Looking through this PR to find out why there was an error when deployed and found this bug:

The default for `significance` should be a `SignificanceData` not a `BranchComparisonData`.

PR #11563, opened 2024-10-17
feat(nimbus): add segments to new Nimbus UI summary page

### LINKED ISSUE #11533: Add segments to new Nimbus UI summary page
We need to add the selected segments with links to their definitions to the new summary page for the user to view.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4873)

RJAK11:
Because

* We want to allow the user to view the selected segments on the new Nimbus UI summary page

This commit

* Updates the new Nimbus UI summary page to display the selected segments with links to their definitions

Fixes #11533

image
image

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/views.py
+++ b/experimenter/experimenter/nimbus_ui_new/views.py
@@ -112,6 +112,7 @@ def build_experiment_context(experiment):
     segment_links = [
         (
             segment,
+            # ruff prefers this implicit syntax for concatenating strings
             f"{segment_doc_base_url}"
             f"{experiment.application.replace('-', '_')}/"
             f"#{segment}",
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -98,19 +98,31 @@ def get(self, *args, **kwargs):
...
         for outcome in experiment.secondary_outcomes
     ]
+
+    segment_doc_base_url = "https://mozilla.github.io/metric-hub/segments/"
+    segment_links = [
+        (
+            segment,
+            f"{segment_doc_base_url}"
+            f"{experiment.application.replace('-', '_')}/"
+            f"#{segment}",
```
mikewilli, line 117: Was this too long to fit on one line? I was confused at first about why we had a 4-tuple here (because I didn't notice the missing commas). What do you think about using `\` and `+` at the end/beginning of each line so it's clearer that it's a single string?

mikewilli (approved):
Excellent work! I left the smallest of comments, but not too concerned about it -- up to you what you think is clearer syntax.

PR #11608, opened 2024-10-22
Add a debug targeting config for bug 1920698

### LINKED ISSUE #11607: Add a debug targeting config for bug 1920698
See [bug 1920698, comment 13](https://bugzilla.mozilla.org/show_bug.cgi?id=1920698#c13) - we need to revive the old targeting config from before safe navigators were added in #11467 so we can reproduce the original bug. But we don't want to revert the changes, since we have experiments relying on the safe targeting configs. So we'll add a new targeting config exclusively for debugging.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4901)

aminomancer:
Fixes #11607

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -2005,6 +2005,26 @@ INELIGIBLE_FOR_DEFAULT_PDF_HANDLER = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+DEBUG_INELIGIBLE_FOR_DEFAULT_PDF_HANDLER = NimbusTargetingConfig(
+    name="DEBUG Ineligible for default PDF handler",
+    slug="debug_ineligible_for_default_pdf_handler",
+    description=(
+        "Targeting users that are ineligible to set Firefox as the default PDF "
+        "handler or have already done so. This omits safe navigators so bugs "
+        "in the targeting getters can be debugged."
+    ),
+    targeting=(
+        "(!os.isWindows || os.windowsVersion < 10) "
+        "|| hasActiveEnterprisePolicies "
+        "|| isDefaultHandler.pdf "
+        "|| (defaultPDFHandler.registered && !defaultPDFHandler.knownBrowser)"
+    ),
+    desktop_telemetry="",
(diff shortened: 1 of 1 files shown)


brennie (approved):

experimenter/experimenter/targeting/constants.py
```diff
@@ -2005,6 +2005,26 @@ def __post_init__(self):
...
+DEBUG_INELIGIBLE_FOR_DEFAULT_PDF_HANDLER = NimbusTargetingConfig(
+    name="DEBUG Ineligible for default PDF handler",
+    slug="debug_ineligible_for_default_pdf_handler",
+    description=(
+        "Targeting users that are ineligible to set Firefox as the default PDF "
+        "handler or have already done so. This omits safe navigators so bugs "
+        "in the targeting getters can be debugged."
+    ),
+    targeting=(
+        "(!os.isWindows || os.windowsVersion < 10) "
```
punamdahiya, line 2017: @aminomancer defaultPDFHandler on non-windows is easy to replicate. It will help if we can land 

`ELIGIBLE_FOR_DEFAULT_PDF_HANDLER` thats known to return `defaultPDFHandler` value duting early startup

punamdahiya (changes requested):

experimenter/experimenter/targeting/constants.py
punamdahiya: Thinking more, I believe JEXL will still evaluate rest of the expression on windows even if `(!os.isWindows || os.windowsVersion < 10) ` evaluates `false` and should see `defaultPDFHandler` exception to debug. 

But if we have to land one targeting for debugging will encourage `DEBUG_ELIGIBLE_FOR_DEFAULT_PDF_HANDLER` which should evaluate `true` with no exception irrespective of JEXL order

punamdahiya (approved):

experimenter/experimenter/targeting/constants.py
aminomancer: I think defaultPDFHandler only works on windows. I figured I'd just add the one that Marius used for the bug last time since at least we know for sure it throws

PR #11615, opened 2024-10-24
chore(nimbus): Update pyyaml and watchdog dependencies

### LINKED ISSUE #11601: Some dependencies don't build with Poetry 1.7.x
Some of our dependencies won't build with a more modern poetry (1.7.0 and higher) that require [PEP 517](https://peps.python.org/pep-0517/) builds.

These include:

- pyyaml (a transitive dependency pinned at 6.0.0, fixed in 6.0.1)
- watchdog (also a transitive dependency, pinned at 2.2.1, fixed in 4.0.0)

See-also: https://github.com/python-poetry/poetry/issues/8287

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4897)

brennie:
Because:

- pyyaml 6.0.0 and watchdog 2.21.0 don't build with PEP 517

This commit:

- updates our pinned versions to versions that do build

Fixes #11601


mikewilli (commented):
Should we also update the pyproject.toml to reflect this (I see `PyYAML = "^6.0"`)?

ETA: it's also in the schemas pyproject.toml (though the lockfile seems correct there at least)

b4handjr:
Does this not need the `pyproject.toml` file to reflect these changes?

brennie:
This was accomplished by:

- adding `pyyaml = "^6.0.1"` and `watchdog = "^4.0.0"` to the depenencies
- running `poetry lock --no-update`
- removing `pyyaml` and `watchdog` from the dependencies
- running `poetry lock --no-update` again

We shouldn't run into any trouble because `poetry lock` should never downgrade a package.

b4handjr:
> This was accomplished by:
> 
>     * adding `pyyaml = "^6.0.1"` and `watchdog = "^4.0.0"` to the depenencies
> 
>     * running `poetry lock --no-update`
> 
>     * removing `pyyaml` and `watchdog` from the dependencies
> 
>     * running `poetry lock --no-update` again
> 
> 
> We shouldn't run into any trouble because `poetry lock` should never downgrade a package.

So we don't want to track `pyyaml` and `watchdog` explicitly?

mikewilli:
That makes sense and I agree that our use of poetry.lock should prevent any issues, but PyYAML is in the pyproject (for experimenter and schemas) so I'd suggest at least updating to reflect our minimum required version.

That said, I don't think we need to block on it if you feel strongly otherwise, so I'll go ahead and approve.

mikewilli (approved):

brennie:
Oh sorry I misunderstood. I didn't realize you were saying experimenter is using `import yaml` (... even though, looking at git log, I wrote some of those lines ...)

I'll update this to depend on pyyaml explicitly.

brennie:
> > This was accomplished by:
> > ```
> > * adding `pyyaml = "^6.0.1"` and `watchdog = "^4.0.0"` to the depenencies
> > 
> > * running `poetry lock --no-update`
> > 
> > * removing `pyyaml` and `watchdog` from the dependencies
> > 
> > * running `poetry lock --no-update` again
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We shouldn't run into any trouble because `poetry lock` should never downgrade a package.
> 
> So we don't want to track `pyyaml` and `watchdog` explicitly?

re: mike's comment, we should track pyyaml. However, watchdog is just a transitive dep so I think it's fine to just bump it so that it builds.

yashikakhurana (approved):

PR #11616, opened 2024-10-24
chore(nimbus): Update to poetry 1.8.4

### LINKED ISSUE #11602: Upgrade poetry
We're using an old version of poetry and our installation might not be working on newer poetry (cf https://github.com/mozilla/experimenter/issues/11601).

brennie:
Because:

- We are on an older version of poetry

This commit:

- Updates to the latest version of peotry

Fixes #11602


brennie:
Blocked on #11615

mikewilli (changes requested):
Want to update schemas and cirrus Dockerfiles here as well? I also noticed a few shell scripts in experimenter/tests that might be worth updating so everything's in sync.

Or maybe there's a way we could have all of these inherit the poetry version from the same spot?

brennie:
> Want to update schemas and cirrus Dockerfiles here as well? I also noticed a few shell scripts in experimenter/tests that might be worth updating so everything's in sync.
> 
> Or maybe there's a way we could have all of these inherit the poetry version from the same spot?

I only want to change one at a time in case we break one thing we dont have to back it out for everything else.

yashikakhurana (approved):

yashikakhurana:
can we unblock this now? @brennie

jaredlockhart (approved):
Very nice 👍 ty 🙏

mikewilli (approved):

PR #11639, opened 2024-10-29
feat(nimbus): Subscribe/Unsubscribe to experiment on new summary page

### LINKED ISSUE #11611: Summary page subscribers
Enable subscriber functionality on the new summary page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4884)

yashikakhurana:
Because

- We want to add functionality on the new summary page so that we can subscribe to the experiment

This commit

- Subscribe/Unsubscribe button option on the summary page

Fixes #11611
Screenshot 2024-10-29 at 1 10 13 PM
Screenshot 2024-10-29 at 1 10 07 PM
Screenshot 2024-10-29 at 1 04 16 PM


experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -215,3 +216,25 @@ class MetricsUpdateView(NimbusExperimentViewMixin, RequestFormMixin, UpdateView)
...
+
+        context = {
+            "experiment": experiment,
+        }
+        html = render_to_string(
+            "nimbus_experiments/subscribers_list.html",
+            context,
+            request=request,
+        )
+        return HttpResponse(html)
```
jaredlockhart, line 240: Yeah you can technically do it this way but it makes it inflexible for moving other handling logic around in your views, do you wanna try doing the pattern I showed you where you hand this off to the form?

jaredlockhart: Also if you split this into two separate views, one just for subscribing, one for unsubscribing, then you don't need to do the check here at all.

yashikakhurana: let me try that

brennie: see-also https://docs.djangoproject.com/en/5.1/topics/http/shortcuts/#django.shortcuts.render which is equivalent to HttpResonse(render_to_string(...))

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/subscribers_list.html
```diff
@@ -0,0 +1,27 @@
+<tr class="subscribers-section">
+  <th>Subscribers</th>
+  <td id="subscribers-list" colspan="2">
+    {% for subscriber in experiment.subscribers.all %}
+      <p>{{ subscriber.email }}</p>
+    {% empty %}
+      <span class="text-danger">Not set</span>
+    {% endfor %}
+  </td>
```
mikewilli, line 9: Probably a concern that can be dealt with later if it becomes a problem, but perhaps this should get a max height and overflow auto or some other way to ensure the list of subscribers isn't huge?

Or, alternatively, is there a good reason to show every user the list of every subscriber? Could probably just show the button so they know whether they are subscribed or not.

yashikakhurana: Good point, @mikewilli, for longer list support, I have added the CSS. For the second point, I am not sure design-wise. I am taking note of this and will ask Shell, too, so we can think more about this when we start redesigning.

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -191,3 +191,35 @@ def __init__(self, *args, **kwargs):
...
+        fields = []
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super().__init__(*args, **kwargs)
+
+    def save(self, commit=True):
+        self.instance.subscribers.add(self.user)
+        if commit:
+            self.instance.save()
```
jaredlockhart, line 208: Do you need this here?  If you do I'd do it as

```py
experiment = super().save(commit=commit)
```

But if the only thing this form is doing is adding the subscriber, this shouldn't be necessary?

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -215,3 +218,27 @@ class MetricsUpdateView(NimbusExperimentViewMixin, RequestFormMixin, UpdateView)
...
+class SubscribeView(NimbusExperimentViewMixin, RequestFormMixin, UpdateView):
+    form_class = SubscribeForm
+    template_name = "nimbus_experiments/subscribers_list.html"
+
+    def form_valid(self, form):
+        experiment = get_object_or_404(NimbusExperiment, slug=self.kwargs["slug"])
+        form.instance = experiment
+        form.user = self.request.user
+        form.save()
+        return render(self.request, self.template_name, {"experiment": experiment})
```
jaredlockhart, line 232: Shouldn't you just be able to write this similar to what is above, ie

```

    def form_valid(self, form):
        super().form_valid(form)
        return self.render_to_response(self.get_context_data(form=form))
```

Since all the 404, save, etc logic is already in the super call of that method?

jaredlockhart (changes requested):

experimenter/experimenter/nimbus_ui_new/views.py
yashikakhurana: done!

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -215,3 +217,23 @@ class MetricsUpdateView(NimbusExperimentViewMixin, RequestFormMixin, UpdateView)
...
         super().form_valid(form)
         return self.render_to_response(self.get_context_data(form=form))
+
+
+class SubscribeView(NimbusExperimentViewMixin, RequestFormMixin, UpdateView):
+    form_class = SubscribeForm
+    template_name = "nimbus_experiments/subscribers_list.html"
+
+    def form_valid(self, form):
+        form.user = self.request.user
```
jaredlockhart, line 227: This should be handled by the `RequestFormMixin` that's already included in this view?

jaredlockhart (changes requested):

experimenter/experimenter/nimbus_ui_new/views.py
yashikakhurana: getting error without it

yashikakhurana: I tried without this too

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -191,3 +191,39 @@ def __init__(self, *args, **kwargs):
...
+        model = NimbusExperiment
+        fields = []
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super().__init__(*args, **kwargs)
+
+    def save(self, commit=True):
+        experiment = super().save(commit=commit)
+        experiment.subscribers.remove(self.user)
```
jaredlockhart, line 225: Ah this should be 

```
experiment.subscribers.remove(self.request.user)
```

experimenter/experimenter/nimbus_ui_new/urls.py
```diff
@@ -52,4 +54,10 @@
         NimbusExperimentsCreateView.as_view(),
         name="nimbus-new-create",
     ),
+    re_path(r"^(?P<slug>[\w-]+)/subscribe/", SubscribeView.as_view(), name="subscribe"),
+    re_path(
+        r"^(?P<slug>[\w-]+)/unsubscribe/",
+        UnsubscribeView.as_view(),
+        name="unsubscribe",
+    ),
```
jaredlockhart, line 62: Oh these url names should be

```
nimbus-new-subscribe
nimbus-new-unsubscribe
```

so we dont' accidentally get collissions with other views 🙏

jaredlockhart (dismissed):
Okay tested locally, everything looks and works great, just noticed the url names are missing a prefix but other than that I think it all looks great!

PR #11640, opened 2024-10-29
feat(nimbus): Store sizing data as JSON in the cache and validate

### LINKED ISSUE #11638: Do not store pickled sizing data in redis cache
Deploying [https://github.com/mozilla/experimenter/pull/11549](https://github.com/mozilla/experimenter/pull/11549|smart-link)  broke the v5 API (see[https://github.com/mozilla/experimenter/issues/11618](https://github.com/mozilla/experimenter/issues/11618|smart-link)). This was due to the following sequence of events:

1. the fetch_population_sizing_data() task ([https://github.com/mozilla/experimenter/blob/015e24b0be7e074a2eb5bae1cdc4218479a48145/experimenter/experimenter/jetstream/tasks.py#L77-L83](https://github.com/mozilla/experimenter/blob/015e24b0be7e074a2eb5bae1cdc4218479a48145/experimenter/experimenter/jetstream/tasks.py#L77-L83|smart-link) ) saves a pydantic v1 model to the cache (which, in the background, uses pickle)
1. we update experimenter to use pydantic v2
1. the population_sizing field in the v5 API tries to read the data from the cache ([https://github.com/mozilla/experimenter/blob/015e24b0be7e074a2eb5bae1cdc4218479a48145/experimenter/experimenter/experiments/api/v5/types.py#L398-L400](https://github.com/mozilla/experimenter/blob/015e24b0be7e074a2eb5bae1cdc4218479a48145/experimenter/experimenter/experiments/api/v5/types.py#L398-L400|smart-link) ) which attempts to unpickle the v1 model into the v2 class, which fails

This was reproduced locally by running the population sizing task on a pydantic v1 revision and persisting the redis cache, then updating to the pydantic v2 revision and running the following in `./manage.py shell`:

    In [1]: from django.core.cache import cache
    
    In [2]: cache.get("population_sizing")
    ValueError                                Traceback (most recent call last)
    File /usr/local/lib/python3.11/site-packages/django_redis/client/default.py:445, in DefaultClient.decode(self, value)
        444 try:
    --> 445     value = int(value)
        446 except (ValueError, TypeError):
    
    ValueError: invalid literal for int() with base 10: b"\x80\x04\x95`\xb8\x00\x00\x00\x00\x00\x00\x8c2mozilla_nimbus_schemas.jetstream.population_sizing\x94\x8c\x0bSampleSizes\x94\x93\x94)\x81\x94}\x94(\x8c\x08__dict__\x94}\x94\x8c\x08__root__\x94}\x94(\
    
    During handling of the above exception, another exception occurred:
    
    KeyError                                  Traceback (most recent call last)
    Cell In[2], line 1
    ----> 1 cache.get("population_sizing")
    
    File /usr/local/lib/python3.11/site-packages/django_redis/cache.py:92, in RedisCache.get(self, key, default, version, client)
         91 def get(self, key, default=None, version=None, client=None):
    ---> 92     value = self._get(key, default, version, client)
         93     if value is CONNECTION_INTERRUPTED:
         94         value = default
    
    File /usr/local/lib/python3.11/site-packages/django_redis/cache.py:29, in omit_exception.<locals>._decorator(self, *args, **kwargs)
         26 @functools.wraps(method)
         27 def _decorator(self, *args, **kwargs):
         28     try:
    ---> 29         return method(self, *args, **kwargs)
         30     except ConnectionInterrupted as e:
         31         if self._ignore_exceptions:
    
    File /usr/local/lib/python3.11/site-packages/django_redis/cache.py:99, in RedisCache._get(self, key, default, version, client)
         97 @omit_exception(return_value=CONNECTION_INTERRUPTED)
         98 def _get(self, key, default, version, client):
    ---> 99     return self.client.get(key, default=default, version=version, client=client)
    
    File /usr/local/lib/python3.11/site-packages/django_redis/client/default.py:265, in DefaultClient.get(self, key, default, version, client)
        262 if value is None:
        263     return default
    --> 265 return self.decode(value)
    
    File /usr/local/lib/python3.11/site-packages/django_redis/client/default.py:452, in DefaultClient.decode(self, value)
        449     except CompressorError:
        450         # Handle little values, chosen to be not compressed
        451         pass
    --> 452     value = self._serializer.loads(value)
        453 return value
    
    File /usr/local/lib/python3.11/site-packages/django_redis/serializers/pickle.py:32, in PickleSerializer.loads(self, value)
         31 def loads(self, value: bytes) -> Any:
    ---> 32     return pickle.loads(value)
    
    File /usr/local/lib/python3.11/site-packages/pydantic/root_model.py:97, in RootModel.__setstate__(self, state)
         96 def __setstate__(self, state: dict[Any, Any]) -> None:
    ---> 97     _object_setattr(self, '__pydantic_fields_set__', state['__pydantic_fields_set__'])
         98     _object_setattr(self, '__dict__', state['__dict__'])
    
    KeyError: '__pydantic_fields_set__'

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4907)

brennie:
Because:

- by default, django-redis will pickle objects to cache them;
- we were storing pickled SampleSizes instances in the cache; and
- updating to mozilla-nimbus-schemas v2024.10.1 (and pydantic v2) broke the unpickling

This commit:

- changes our django-redis configuration to use a JSON serializer;
- updates the usage of the cache to always store a JSON representation of the sizing data; and
- evicts invalid sample size cache entries if they cannot be parsed

Fixes #11638

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/types.py
+++ b/experimenter/experimenter/experiments/api/v5/types.py
@@ -1,10 +1,13 @@
 import json
+import logging
 
 import graphene
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.cache import cache
 from graphene_django import DjangoListField
 from graphene_django.types import DjangoObjectType
+from mozilla_nimbus_schemas.jetstream import SampleSizes
 
 from experimenter.base.models import Country, Language, Locale
 from experimenter.experiments.api.v5.serializers import (
@@ -26,7 +29,8 @@ from experimenter.experiments.models import (
 )
 from experimenter.outcomes import Outcomes
 from experimenter.projects.models import Project
-from experimenter.settings import SIZING_DATA_KEY
+
+logger = logging.getLogger(__name__)
 
 
 class NimbusExperimentStatusEnum(graphene.Enum):
@@ -396,8 +400,20 @@ class NimbusConfigurationType(graphene.ObjectType):
         )
 
     def resolve_population_sizing_data(self, info):
-        sizing_data = cache.get(SIZING_DATA_KEY)
-        return sizing_data.json() if sizing_data else "{}"
+        try:
+            sizing_data = cache.get(settings.SIZING_DATA_KEY)
+            if not isinstance(sizing_data, str):
+                raise TypeError(f"expected str, got {type(sizing_data).__name__}")
+
+            return SampleSizes.parse_raw(sizing_data).json()
+        except Exception as e:
+            logger.error(
+                f"Could not parse sizing data in cache key '{settings.SIZING_DATA_KEY}': "
+                f"{e}; evicting cache entry"
+            )
+            cache.delete(settings.SIZING_DATA_KEY)
+
+        return "{}"
 
     @staticmethod
     def sort_version_choices(choices):
--- a/experimenter/experimenter/experiments/tests/api/v5/test_queries.py
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_queries.py
@@ -5,6 +5,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.urls import reverse
 from graphene_django.utils.testing import GraphQLTestCase
+from mozilla_nimbus_schemas.jetstream import SampleSizesFactory
 from parameterized import parameterized
 
 from experimenter.base.models import Country, Language, Locale
@@ -2438,9 +2439,13 @@ class TestNimbusExperimentsByApplicationMetaQuery(GraphQLTestCase):
 class TestNimbusConfigQuery(MockSizingDataMixin, GraphQLTestCase):
     GRAPHQL_URL = reverse("nimbus-api-graphql")
 
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.user_email = "user@example.com"
+
     def test_nimbus_config(self):
         self.setup_cached_sizing_data()
-        user_email = "user@example.com"
         feature_configs = NimbusFeatureConfigFactory.create_batch(10)
         application = NimbusExperiment.Application.DESKTOP
         feature_config = NimbusFeatureConfigFactory.create(application=application)
@@ -2568,7 +2573,7 @@ class TestNimbusConfigQuery(MockSizingDataMixin, GraphQLTestCase):
                 }
             }
             """,
-            headers={settings.OPENIDC_EMAIL_HEADER: user_email},
+            headers={settings.OPENIDC_EMAIL_HEADER: self.user_email},
         )
         self.assertEqual(response.status_code, 200, response.content)
         content = json.loads(response.content)
@@ -2593,7 +2598,8 @@ class TestNimbusConfigQuery(MockSizingDataMixin, GraphQLTestCase):
 
         pop_sizing_data = self.get_cached_sizing_data()
         self.assertEqual(
-            config["populationSizingData"], pop_sizing_data.json(exclude_unset=True)
+            config["populationSizingData"],
+            pop_sizing_data,
         )
 
         self.assertEqual(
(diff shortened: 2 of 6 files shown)


experimenter/experimenter/jetstream/client.py
```diff
@@ -365,3 +369,21 @@ def get_population_sizing_data():
...
     return {"v1": sizing}
+
+
+def get_cached_population_sizing_data() -> SampleSizes | None:
+    try:
+        sizing_data = cache.get(settings.SIZING_DATA_KEY)
+        if not isinstance(sizing_data, str):
+            raise TypeError(f"expected str, got {type(sizing_data)}")
+
+        sizing_data = SampleSizes.parse_raw(sizing_data)
```
mikewilli, line 380: This PR is pre-pydantic-v2 migration, yes? Otherwise this should be `model_validate_json`.

experimenter/experimenter/jetstream/client.py
```diff
@@ -365,3 +369,21 @@ def get_population_sizing_data():
...
+    except Exception as e:
+        logger.warning(
```
mikewilli, line 382: Seems like an error since it indicates bad sizing data, no?
```suggestion
        logger.error(
```

experimenter/experimenter/experiments/tests/api/v5/test_queries.py
```diff
@@ -2722,4 +2727,35 @@ def assertChoices(data, text_choices):
...
+        [
+            object(),
+            123,
+            json.dumps(
+                {
+                    "bogus": "data",
+                }
+            ),
+        ]
+    )
```
mikewilli, line 2742: Probably unnecessary, but could also include a `SampleSizes` instance here since that's what caused the original issue.

experimenter/experimenter/jetstream/client.py
```diff
@@ -365,3 +369,21 @@ def get_population_sizing_data():
     sizing_data = get_sizing_data(suffix="latest")
     sizing = SampleSizes.parse_obj(sizing_data) if sizing_data is not None else {}
     return {"v1": sizing}
+
+
+def get_cached_population_sizing_data() -> SampleSizes | None:
```
mikewilli, line 374: I'm not sure this should live here since it doesn't deal with Jetstream output or the celery tasks to retrieve it, and it's only used in `api/v5/types.py`. Any reason not to just put it in there?

mikewilli (changes requested):
My opinion would be to put the retrieve-from-cache functionality into `api/v5/types.py` (or somewhere else within the API or app side of things). Looks good other than that!

brennie:
> My opinion would be to put the retrieve-from-cache functionality into `api/v5/types.py` (or somewhere else within the API or app side of things). Looks good other than that!

The code that writes to the cache is in jetstream.tasks so I would expect the cache read to also live in jetstream somewhere.

experimenter/experimenter/jetstream/client.py
brennie: yep this lands before the v2 migration

experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -24,9 +24,11 @@
     NimbusExperimentBranchThroughRequired,
     NimbusFeatureConfig,
 )
+from experimenter.jetstream.client import get_cached_population_sizing_data
```
mikewilli, line 27: This can be removed now yea?

```suggestion
```

mikewilli: And also need settings here:
```py
from django.conf import settings
```

experimenter/experimenter/settings.py
```diff
@@ -369,6 +369,10 @@
         "BACKEND": "django_redis.cache.RedisCache",
         "LOCATION": f"redis://{REDIS_HOST}:{REDIS_PORT}/{REDIS_DB}",
         "TIMEOUT": None,
+        "OPTIONS": {
+            "CLIENT_CLASS": "django_redis.client.DefaultClient",
+            "SERIALIZER": "django_redis.serializers.json.JSONSerializer",
```
mikewilli, line 374: Neat!

experimenter/experimenter/jetstream/client.py
```diff
@@ -5,6 +5,7 @@
 from pathlib import Path
 
 from django.conf import settings
+from django.core.cache import cache
```
mikewilli, line 8:
```suggestion
```

mikewilli (approved):
I think there's a few import issues but looks good thanks!

experimenter/experimenter/experiments/api/v5/types.py
brennie: not enough coffee

PR #11646, opened 2024-10-30
chore(schemas): upgrade to Poetry 1.8.4

### LINKED ISSUE #11645: Update schemas to Poetry 1.8.4
We're using poetry 1.8.4 everywhere else now as of #11616, so lets be consistent here too.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4910)

brennie:
chore(schemas): upgrade to Poetry 1.8.4

Because:

- everything else is using Poetry 1.8.4; and
- the README says we're using Python 3.11 but actually we're using 3.10

This commit:

- updates mozilla_nimbus_schemas to use Poetry 1.8.4; and
- updates the Python version in the README.

Fixes #11645

### DIFF
--- a/schemas/Dockerfile
+++ b/schemas/Dockerfile
@@ -33,7 +33,7 @@ RUN . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
 ENV PYTHONDONTWRITEBYTECODE 1
 
 # Python packages
-RUN curl -sSL https://install.python-poetry.org | python3 - --version 1.6.0
+RUN curl -sSL https://install.python-poetry.org | python3 - --version 1.8.4
 ENV PATH "/root/.local/bin:$PATH"
 RUN poetry config virtualenvs.create false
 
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
Looks good!

Related: we should probably remove the versions for all the prereqs in the schemas README so that we don't have to constantly update them with changes like this. I just looked to see where else we might be saying a poetry version and found that:
- python says 3.11 which is wrong
- poetry says 1.2.2 which is OLD
- yarn says 1.22 but seems like the Dockerfile just grabs the latest?

Anyway feel free to ignore for this PR but just a thought. I can file a follow-up issue from this comment if you'd rather not combine scope.

brennie:
> Looks good!
> 
> Related: we should probably remove the versions for all the prereqs in the schemas README so that we don't have to constantly update them with changes like this. I just looked to see where else we might be saying a poetry version and found that:
> 
>     * python says 3.11 which is wrong
> 
>     * poetry says 1.2.2 which is OLD
> 
>     * yarn says 1.22 but seems like the Dockerfile just grabs the latest?
> 
> 
> Anyway feel free to ignore for this PR but just a thought. I can file a follow-up issue from this comment if you'd rather not combine scope.

Ill update the schemas/README in this change, but the yarn stuff should be a follow-upl. Can you file that @mikewilli ?

PR #11648, opened 2024-10-30
chore(ci): Update tracking namespaces for firefox fenix builds.

### LINKED ISSUE #11647: Update fenix task ID tracking config
It looks like our fenix tracking keys for the integration tests haven't been updated in a while. Digging into this a bit more I found out that there are different taskcluster namespaces I need to track, lets update to those.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4911)

b4handjr:
Because

- The namespaces we were using to track fenix builds in Mozilla central were incorrect

This commit

- Updates the namespaces and the files with the correct task IDs

Fixes #11647


jaredlockhart (approved):
Cool, nice that it's so easy to update, thnx @b4handjr 👍

PR #11655, opened 2024-10-30
bug(nimbus): dont throw when cached population sizing data is None

### LINKED ISSUE #11654: Could not parse sizing data in cache key 'population_sizing': expected str, got NoneType; evictin...
Sentry Issue: [EXPERIMENTER-PROD-SV](https://mozilla.sentry.io/issues/6031775758/?referrer=github_integration)

```
Could not parse sizing data in cache key 'population_sizing': expected str, got NoneType; evicting cache entry
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4916)

brennie:
Because:

- we now throw when we get an unexpected data type from the cache when reading the population sizing data; and
- we are not doing a check for None, which is causing an exception to be logged for every graphQL query for nimbus configuration

This commit:

- adds an explicit `is not None` check to the cache read.

Fixes #11654

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/types.py
+++ b/experimenter/experimenter/experiments/api/v5/types.py
@@ -402,10 +402,11 @@ class NimbusConfigurationType(graphene.ObjectType):
     def resolve_population_sizing_data(self, info):
         try:
             sizing_data = cache.get(settings.SIZING_DATA_KEY)
-            if not isinstance(sizing_data, str):
-                raise TypeError(f"expected str, got {type(sizing_data).__name__}")
+            if sizing_data is not None:
+                if not isinstance(sizing_data, str):
+                    raise TypeError(f"expected str, got {type(sizing_data).__name__}")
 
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/experiments/api/v5/types.py
```diff
@@ -402,10 +402,11 @@ def resolve_conclusion_recommendations_choices(self, info):
     def resolve_population_sizing_data(self, info):
         try:
             sizing_data = cache.get(settings.SIZING_DATA_KEY)
-            if not isinstance(sizing_data, str):
-                raise TypeError(f"expected str, got {type(sizing_data).__name__}")
+            if sizing_data is not None:
```
jaredlockhart, line 405: Omg I was literally thinking this all day yesterday I can't believe I forgot when reviewing XDDD

jaredlockhart (approved):
🙏

PR #11663, opened 2024-10-31
feat(project): add a make refresh_db target

### LINKED ISSUE #11661: Add a separate Makefile command to rebuild the experimenter DB
I often find that I run `make up` without running `make refresh`, which means I will build all the containers but there won't be a valid database. However, `make refresh` is defined as:

```
refresh: kill build_dev compose_build  ## Rebuild all containers
	$(COMPOSE) run -e SKIP_DUMMY=$$SKIP_DUMMY experimenter bash -c '$(WAIT_FOR_DB) $(PYTHON_MIGRATE)&&$(LOAD_LOCALES)&&$(LOAD_COUNTRIES)&&$(LOAD_LANGUAGES)&&$(LOAD_FEATURES)&&$(LOAD_DUMMY_EXPERIMENTS)'
```

which means running `make up refresh` will build all the containers and then blow them all away to rebuild them again.

We should refactor the db changes out of refresh into a separate target (`make refresh_db` ?) that refresh depends on so behaviour doesn't change, but allowing us to recreate the DB without blowing away all the containers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4919)

brennie:
Because:

- make refresh does rebuild the DB but also blows away all containers; and
- I just want to rebuild the DB

This commit:

- adds a make refresh_db (which refresh depends on) that will just recreate the database.

Fixes #11661

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,9 @@ migrate: build_dev  ## Run database migrations
 bash: build_dev
 	$(COMPOSE) run experimenter bash
 
-refresh: kill build_dev compose_build  ## Rebuild all containers
+refresh: kill build_dev compose_build refresh_db  ## Rebuild all containers and the database
+
+refresh_db:  # Rebuild the database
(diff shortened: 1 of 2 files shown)


jaredlockhart (changes requested):
Ah handy!  But don't forget to update README 🙏

jaredlockhart (approved):
Awesome ty! 🎉

PR #11664, opened 2024-10-31
chore(project): docker run with --rm

### LINKED ISSUE #11662: Run one off `docker compose run` commands with --rm
I run `make bash` a lot and therefore have a lot of containers laying around that aren't automatically deleted. If we run with `--rm`, the containers will be deleted after exiting.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4920)

brennie:
Because:

- we are creating many short-lived containers for commands; and
- we are not cleaning them up except in `make kill`

This commit:

- updates all docker run commands to use the `--rm` flag so that they are automatically deleted after the container exits.

Fixes #11662

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,13 @@ WAIT_FOR_RUNSERVER = /experimenter/bin/wait-for-it.sh -t 30 localhost:7001 &&
 
 COMPOSE_CIRRUS = [[ -n $$CIRRUS ]] && echo "-f docker-compose-cirrus.yml"
 COMPOSE = docker compose -f docker-compose.yml $$(${COMPOSE_CIRRUS})
+COMPOSE_RUN = ${COMPOSE} run --rm
 COMPOSE_LEGACY = ${COMPOSE} -f docker-compose-legacy.yml
 COMPOSE_TEST = docker compose -f docker-compose-test.yml
+COMPOSE_TEST_RUN = ${COMPOSE_TEST} run --rm
 COMPOSE_PROD = docker compose -f docker-compose-prod.yml $$(${COMPOSE_CIRRUS})
 COMPOSE_INTEGRATION = ${COMPOSE_PROD} -f docker-compose-integration-test.yml $$(${COMPOSE_CIRRUS})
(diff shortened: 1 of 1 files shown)


brennie:
Blocked on #11663

Makefile
```diff
@@ -85,7 +85,10 @@ jetstream_config:
 	rm -Rf experimenter/experimenter/segments/metric-hub-main/.script/
 
 feature_manifests: build_dev
-	$(COMPOSE) run -e GITHUB_BEARER_TOKEN=$(GITHUB_BEARER_TOKEN) experimenter /experimenter/bin/manifest-tool.py fetch $(FETCH_ARGS)
+	$(COMPOSE) run \
+		--rm \
```
jaredlockhart, line 89: If we're doing this pattern of

```
$(COMPOSE) run \
		--rm \
```

everywhere, could we just add it as another alias at the top of the file so we don't have to repeat it everywhere? 🤔

Makefile
```diff
@@ -5,10 +5,13 @@ WAIT_FOR_RUNSERVER = /experimenter/bin/wait-for-it.sh -t 30 localhost:7001 &&
 
 COMPOSE_CIRRUS = [[ -n $$CIRRUS ]] && echo "-f docker-compose-cirrus.yml"
 COMPOSE = docker compose -f docker-compose.yml $$(${COMPOSE_CIRRUS})
+COMPOSE_RUN = ${COMPOSE} run --rm
 COMPOSE_LEGACY = ${COMPOSE} -f docker-compose-legacy.yml
 COMPOSE_TEST = docker compose -f docker-compose-test.yml
+COMPOSE_TEST_RUN = ${COMPOSE} run --rm
```
jaredlockhart, line 11: Should this not be

```
COMPOSE_TEST_RUN = ${COMPOSE_TEST} run --rm
```

?

jaredlockhart (changes requested):

Makefile
brennie: Yes :)

PR #11674, opened 2024-11-01
feat(nimbus): Integrate segment retrieval function from metric-config-parser

### LINKED ISSUE #11623: Integrate segment retrieval function in Experimenter
Update Experimenter to import and utilize the new segment retrieval function from `metric-config-parser`, replacing manual segment parsing.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4893)

RJAK11:
Because

* experimenter was manually parsing segments when it could just import them from metric-config-parser

This commit

* replaces manual segment parsing with metric-config-parser’s retrieval function

Fixes #11623


experimenter/experimenter/segments/__init__.py
```diff
@@ -68,22 +68,22 @@ def clear_cache(cls):
         cls._segments = None
 
     @classmethod
-    def all(cls):
+    def all(cls, segment_data=None):
```
mikewilli, line 71: I'd suggest adding type hints to segment_data since it may not be immediately obvious what it is intended to be (here and in `by_application`).

experimenter/pyproject.toml
```diff
@@ -64,7 +64,9 @@ djangorestframework-dataclasses = "^1.3.0"
 graphene-django = "^3.2.0"
 pyright = "^1.1.291"
 django-types = "^0.19.1"
+mozilla-nimbus-shared = "^2.5.2"
```
mikewilli, line 67: Beth just removed this dependency in the [latest commit](https://github.com/mozilla/experimenter/commit/ad9134c36fdbadefd92a9a4f1b5493c967f60fa2#diff-86347d9aaa24ca8462da5352e18d80569f750e036764c45a5d3ef6e7c0fa33d3), so I'm guessing this was accidentally added back in resolving a conflict? In any case, we definitely shouldn't be adding it back in here.

mikewilli (changes requested):
Nice! I think there was an mistakenly re-added dependency, and I had one other small suggestion. Other than that, great work and thanks for figuring out the confusing CI issues!

experimenter/experimenter/segments/__init__.py
```diff
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from typing import Dict, List, Optional
```
mikewilli, line 2: Not a big deal, but in the interest of staying up-to-date with Python changes, I'd suggest forgoing these in favor of the newer python type hints:
```py
list[...]  # vs List[...]
dict[...]  # vs Dict[...]
str | None  # vs Optional[str]
```
As a bonus, don't need any imports!

mikewilli (approved):
r+wc

Suggested using the more modern type aliases, but otherwise looks good. Thanks! :tada:

PR #11677, opened 2024-11-01
feat(schemas): pull out NimbusExperiment.localizations into a schema model

### LINKED ISSUE #11673: Pull out NimbusExperiment.localizations into a model
We’re currently doing type validation on the localizations field by pulling out the “localizations” definition from the generated [mozilla_nimbus_shared JSON schema][1]:

```
        schema = settings.EXPERIMENT_SCHEMA["definitions"]["NimbusExperiment"][
            "properties"
        ]["localizations"]

        try:
            jsonschema.validate(localizations, schema)
        except Exception as e:
                {"localizations": [f"Localization schema validation error: {e}"]}
            ) from e
```

since we’re using pydanic models now, we can pull this out into a separate model and use that for validation

[1]: [https://github.com/mozilla/experimenter/blob/6664e457795d9e1ca04545523eebca52ddbf5654/experimenter/experimenter/experiments/api/v5/serializers.py#L2057](https://github.com/mozilla/experimenter/blob/6664e457795d9e1ca04545523eebca52ddbf5654/experimenter/experimenter/experiments/api/v5/serializers.py#L2057|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4929)

brennie:
Because:

- we previously were using the generated JSONSchema definition of the localizations field to validate it in Experimenter; and
- we now have pydantic v2 in Experimenter

This commit:

- pulls out the localizations field into a separate model so that we can use that for valdiation directly in Experimenter; and
- bumps the version to 2024.11.1

Fixes #11673


schemas/mozilla_nimbus_schemas/experiments/experiments.py
```diff
@@ -109,6 +109,16 @@ class ExperimentMultiFeatureMobileBranch(BaseExperimentMultiFeatureBranch):
     """
 
 
+class ExperimentLocalizations(RootModel[dict[str, dict[str, str]]]):
```
jaredlockhart, line 112: Oh wild so you can declare the actual type using the generic type syntax right at the class definition level.  That's crazy I didn't know that.

jaredlockhart (approved):
Looks good, also using the word 'model' here is confusing to me, I know it's what pydantic calls it but maybe we should preface it with 'schema model' or something to distinguish it from the django db model case.  But yep a nice succinct change, so good that we can use the schemas for more and more now 🙏

schemas/mozilla_nimbus_schemas/experiments/experiments.py
brennie: new in pydantic v2! I think you could also do `ExperimentLocalizations = RootModel[...]` but then you dont have the docstring attached

PR #11679, opened 2024-11-01
chore(schemas): Re-order Dockerfile to avoid rebuilding the world

### LINKED ISSUE #11676: Reorder schemas Dockerfile to prevent layer invalidation
We're presently doing:

```
COPY ./mozilla_nimbus_schemas /schemas/mozilla_nimbus_schemas
COPY ./README.md /schemas/README.md
RUN poetry install
```

but really we should be running `poetry install` after only copying the pyproject.toml and poetry.lock files and *then* copying everything else over

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4931)

brennie:
Because:

- we were running poetry install *after* copying files, which was causing unnecessary layer invalidation

This commit:

- re-orders the dependency installtion before file copying, so that file edits do not require you to reinstall dependencies.

Fixes #11676

### DIFF
--- a/schemas/Dockerfile
+++ b/schemas/Dockerfile
@@ -30,16 +30,21 @@ RUN poetry config virtualenvs.create false
 FROM system-builder AS python-builder
 WORKDIR /schemas
 
+# Poetry needs a package to install in package-mode.
+RUN mkdir mozilla_nimbus_schemas && \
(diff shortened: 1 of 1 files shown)


b4handjr (approved):
Nice one

jaredlockhart (approved):
🙏 👍

PR #11680, opened 2024-11-01
chore(nimbus): Use new ENV KEY=value syntax in all Dockerfiles

### LINKED ISSUE #11678: LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format
We're doing this in all our dockerfiles and now docker is complaining all the time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4932)

brennie:
Because:

- Docker has deprecated `ENV KEY value` syntax and is now complaining about our usage

This commit:

- updates our usage to the new syntax.

Fixes #11678


jaredlockhart (approved):
Awesome ty 🙏

PR #11709, opened 2024-11-06
feat(nimbus): Enrollment data on timeline

yashikakhurana:
Because

- We are not showing when the enrollment ended on the new summary page timeline

This commit

- Add a new entry for the enrollment end date if it's not a rollout.

Fixes #11621 
Screenshot 2024-11-06 at 11 57 33 AM
Screenshot 2024-11-06 at 11 52 36 AM

PR #11711, opened 2024-11-06
feat(nimbus): Enrollment and complete days

### LINKED ISSUE #1: Add Circle CI
Let's add circles to the things!

yashikakhurana:
Because

- We are not showing days (enrollment days and total duration of the experiment) on the new summary page

This commit

- Add enrollment days
- Add total duration of the experiment

Fixes #1
Screenshot 2024-11-06 at 12 45 37 PM
1710


yashikakhurana:
Blocked on https://github.com/mozilla/experimenter/pull/11709

experimenter/experimenter/experiments/models.py
```diff
@@ -793,33 +793,54 @@ def computed_duration_days(self):
...
+                4,
+                {
+                    "label": NimbusConstants.ENROLLMENT_END,
+                    "date": self._enrollment_end_date,
+                    "is_active": self._enrollment_end_date is not None,
+                    "days": (
+                        self.computed_enrollment_days
+                        if self._enrollment_end_date
+                        else None
+                    ),
```
mikewilli, line 839: This has a couple things I'm unsure about (as is often the case for experiment dates...):

1. I'm not quite sure if the logic is correct here: `self.computed_enrollment_days` is a computed property, and if `self._enrollment_end_date` is not set but `self.is_paused == True`, then it goes and computes and sets `self._enrollment_end_date` based on the changelogs. Maybe we should change the `if` here to include `or self.is_paused`? Not sure if that's only a relic of Normandy or if it's still used.
2. I think the `if` should include `is not None` like you have above for `is_active` (though not sure if this is likely to have an effect in practice).

experimenter/experimenter/experiments/models.py
```diff
@@ -793,33 +793,54 @@ def computed_duration_days(self):
...
                 "label": self.Status.LIVE,
                 "date": self.start_date,
                 "is_active": self.is_live,
+                "days": None,
             },
             {
                 "label": self.Status.COMPLETE,
                 "date": self.computed_end_date,
                 "is_active": self.is_complete,
+                "days": self.computed_duration_days if self.computed_end_date else None,
```
mikewilli, line 825: Related to my other comment about dates -- why do we use the computed duration here, but the actual enrollment for that one?

It also just occurred to me that the answer to all of my questions might be "because this is how we have it on the current pages", which is a perfectly valid answer!

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -1710,34 +1710,62 @@ def test_timeline_dates_includes_correct_status_dates_and_flags(self):
...
+
+        # Calculate expected days
+        enrollment_days = (
+            experiment.computed_enrollment_days
+            if experiment._enrollment_end_date
+            else None
+        )
+        duration_days = (
+            experiment.computed_duration_days if experiment.computed_end_date else None
+        )
```
mikewilli, line 1722: These should be static values, otherwise we're just computing it the same way twice and comparing.

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/timeline.html
```diff
@@ -4,6 +4,7 @@
       <li class="list-group-item flex-fill text-center d-flex flex-column justify-content-center {% if status.is_active %}bg-primary text-white{% endif %}">
         <strong>{{ status.label }}</strong>
         <small>{{ status.date|default:'---' }}</small>
+        {% if status.days is not None %}<small>{{ status.days }} days</small>{% endif %}
```
mikewilli, line 7: The current page has a little info tooltip about the experiment duration, not sure if it's really necessary but just wanted to flag the difference.

mikewilli: Also, it looks like there's a built-in django template function for `pluralize` if you want to make `day` or `days` based on whether the value is plural.

experimenter/experimenter/experiments/models.py
yashikakhurana: yup I referred to the current pages and translated that

experimenter/experimenter/experiments/tests/test_models.py
yashikakhurana: good point

experimenter/experimenter/experiments/models.py
mikewilli: Ok fair enough. Still seems a bit strange to me, but maybe this was discussed and agreed upon already?

jaredlockhart (changes requested):
Oh, I think we should only ever have one of the status boxes highlighted at a time?

image

Also we could change 'live' to 'enrolling' and 'enrollment end' to 'observing' to make the titles shorter and maybe a little clearer?

Also I like adding teh 'X days' to them, we should put that on all of them, but we can do that seprately.

experimenter/experimenter/experiments/models.py
yashikakhurana: I changed the logic now, would you like to take another look into that?

experimenter/experimenter/experiments/models.py
yashikakhurana: if its not ended we use computed one. but if its ended computed one returns the actual one

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/timeline.html
yashikakhurana: for all the tooltips- I have added a ticket to cover all the tooltips here is the ticket- https://github.com/mozilla/experimenter/issues/11748 and for the `pluralize` I have made that changes, good point though

yashikakhurana:
> Oh, I think we should only ever have one of the status boxes highlighted at a time?
> image
> 
> Also we could change 'live' to 'enrolling' and 'enrollment end' to 'observing' to make the titles shorter and maybe a little clearer?
> 
> Also I like adding the 'X days' to them, we should put that on all of them, but we can do that seprately.

made the changes to show only one box active at a time and changed the name 
Screenshot 2024-11-12 at 10 22 15 AM
Screenshot 2024-11-12 at 10 21 02 AM
Screenshot 2024-11-12 at 10 19 37 AM
Screenshot 2024-11-12 at 10 13 59 AM
Screenshot 2024-11-12 at 10 13 43 AM
Screenshot 2024-11-12 at 10 13 19 AM
 
To add days in each section- I filled the ticket for that https://github.com/mozilla/experimenter/issues/11747

experimenter/experimenter/experiments/models.py
```diff
@@ -812,35 +822,41 @@ def timeline(self):
...
         if not self.is_rollout:
             timeline_entries.insert(
                 4,
                 {
-                    "label": NimbusConstants.ENROLLMENT_END,
+                    "label": NimbusConstants.OBSERVING,
                     "date": self._enrollment_end_date,
-                    "is_active": self._enrollment_end_date is not None,
+                    "is_active": self.is_observing,
+                    "days": self.computed_enrollment_days,
```
mikewilli, line 859: I think this should be a new computed duration (something like `computed_observation_days = end_date - enrollment_end_date`), and `computed_enrollment_days` should go with the "Enrolling" section (formerly "Live").

experimenter/experimenter/experiments/constants.py
```diff
@@ -813,7 +813,8 @@ class EmailType(models.TextChoices):
     WARNING_PREF_FLIPS_PREF_CONTROLLED_BY_FEATURE = (
         "Pref '{pref}' is controlled by a variable in feature {feature_config_slug}'"
     )
-    ENROLLMENT_END = "Enrollment End"
+    OBSERVING = "Observing"
+    ENROLLING = "Enrolling"
```
mikewilli, line 817: I'd suggest that "Enrolling" and "Observing" be changed to "Enrollment" and "Observation", respectively, to make them seem more like states that are not necessarily current. This one's definitely a personal preference though so if others feel differently that's fine too.

mikewilli (dismissed):
I think with the change from "Live" to "Enrolling" and "Enrollment End" to "Observing" it's muddled the "# days" indicators. To me it's a bit confusing that "Observing" shows the # of days that were spent enrolling, and this should instead go under "Enrolling", with a separate computed duration for the observation (`end_date - enrollment_end_date`).

experimenter/experimenter/experiments/models.py
yashikakhurana: I believe this is what `computend_enrollment_days` are doing- https://github.com/mozilla/experimenter/blob/54ad0a197b2d3db07ed89e173e89f0342860f5bb/experimenter/experimenter/experiments/models.py#L747

experimenter/experimenter/experiments/constants.py
yashikakhurana: @jaredlockhart what do you think?

experimenter/experimenter/experiments/models.py
mikewilli: `computed_enrollment_days` is slightly different -- it's `enrollment_end_date - enrollment_start_date`, or the period between when experiment became live (enrollment start date) and end of enrollment (enrollment end date). I'm talking about having a different duration that is the number of days between the end of enrollment and the end of the experiment.

yashikakhurana:
@mikewilli I changed and moved the days around, do you want to take another look? Also thank you for bringing that up

experimenter/experimenter/experiments/constants.py
jaredlockhart: Oh yeah that's not a bad point.

Draft
Preview
Review
Enrollment
Observation
Complete


Yeah I think that makes sense 👍

mikewilli (approved):
LGTM thanks!

PR #11716, opened 2024-11-07
feat(schemas): Split experiment schema into Firefox Desktop and SDK-specific versions

### LINKED ISSUE #11704: Invalid desktop schemas can technically validate as mobile schemas
If we add desktop-specific things to the branch schemas (e.g., required fields), recipes can still pass validation *without those required fields* because the mobile branch schema will match with the field missing.

We need to refactor how our schemas are defined to either (a) a mobile specific experiment JSON schema and desktop specific experiment JSON schema or (b) use condition validation on the app name.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4936)

brennie:
Because:

- our current pan-application schema currently allows SDK experiments to validate on Firefox Desktop (because there is no conditional validation enforcing a relationship between the application field and the branches field);
- we no longer publish single feature experiments; and
- we no longer support applications that do not support multi-feature experiments.

This commit:

- splits our NimbusExperiment schema into a DesktopNimbusExperiment and an SdkNimbusExperiment;
- removes support for mono-feature experiments from both Desktop and SDK experiment schemas;
- removes some old, mono-feature fixtures that do not validate with our new schemas; and
- bumps the schemas package version to 2024.11.3 to publish these changes.

Fixes #11704


jaredlockhart (approved):
So this is a breaking change to the schema here, funny how using the date based versioning doesn't reflect that.  But anyways yes, makes good sense, very clear test structure.  I notice the schema docstring says it's used by jetstream which isn't true anymore, can prolly take that out at some point.  Also I didn't know that by Microsoft camel convention you only capitalize the first letter of an initialization.  Anyways thanks for this @brennie looks great 🙏

PR #11718, opened 2024-11-07
bug(schemas): Match the placeholder feature ID to Experimenter

### LINKED ISSUE #11717: Experimenter uses a different tombstone feature ID than specified in schemas
The v6 and v8 APIs use the following placeholders for the feature ID:

```
"this-is-included-for-desktop-pre-95-support",
"this-is-included-for-mobile-pre-96-support",
```

The desktop feature ID in the schema is `'this-is-included-for-desktop-pre-95-support'` however. We need to update the schemas to match this (and we should maybe add the tombstone to the mobile schema).

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-4943)

brennie:
Because:

- Experimenter uses a different placeholder feature ID than the specified in the schemas; and
- using the new split schema in Firefox Desktop would break the world

This commit:

- splits the Firefox Desktop schema into two versions: a strict, backwards-compatible version (one that enforces the branches[].feature field is present), and a more lax client version (that does not require the branches[].feature field);
- updates the strict schema to use the correct placeholder feature ID;
- updates all the fixtures to use the correct placeholder;
- updates all the tests to test against both the strict and lax schemas; and
- updates the schemas package to 2024.11.4.

Fixes #11717


jaredlockhart (approved):
👍 🙏

jaredlockhart (changes requested):
Oh if we were able to land the wrong tombstone then we must be missing a test or example recipe somewhere right?

jaredlockhart (approved):
Oh funny the fixtures had the wrong value I wonder how that happened?

Anyways good catch 👍

brennie:
I have validated the new schema works on desktop. I'm gonna have a lot of tests to fix :)

jaredlockhart (changes requested):
Okay ya I like this structure, but wrt names:

I'd call `Tombstone` as `Pre95` just to make it clear since really it'sa bout versions and we dont' use 'tombstone' as a name generally in other places and

`DesktopClient` as `DesktopAllVersions` since these schemas are all for clients so 'client' is a bit redundant here?

jaredlockhart (approved):
Ya reads much better, ty @brennie 🙏

PR #11772, opened 2024-11-14
feat(nimbus): switch segment retrieval to use local directories

### LINKED ISSUE #11730: Switch segment retrieval to use local directories
Currently, the `ConfigCollection.from_github_repos` function is fetching segment definitions directly from a GitHub repo, which leads to potential latency issues. We should modify the `ConfigCollection` setup to source segment definition files from the specified local directories instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4957)

RJAK11:
Because

* fetching segment definitions from the metric-hub github repository introduces extra network calls

This commit

* modifies the ConfigCollection segment retrieval setup to source segment definition files from the specified local directories rather than directly from github

Fixes #11730


experimenter/experimenter/segments/__init__.py
```diff
@@ -7,6 +9,22 @@
...
+
+@dataclass
+class MockRepository:
+    git_dir: Path
+
+    def iter_commits(self, *args, **kwargs):
+        mock_commit = MockCommit(
+            committed_date=int(datetime.now().timestamp()),
+        )
+        yield mock_commit
```
mikewilli, line 25: I wish this wasn't necessary, but I do think this is a reasonable approach. A couple suggestions here though:

- can you rename the classes to indicate that they're private by prefixing with `_`?
- might be worth a comment to indicate how they're used? i.e., that they're used exclusively to allow the ConfigCollection to treat the local copy of metric-hub as a git repo, which is needed due to how `from_local_repo` works?

experimenter/experimenter/segments/__init__.py
```diff
@@ -31,10 +49,13 @@ def _load_segments(
 
         config_collection = None
         if segment_data is None:
-            config_collection = ConfigCollection.from_github_repos(
-                [
-                    "https://github.com/mozilla/metric-hub",
-                ]
+            repo_git_dir = Path("/experimenter/experimenter/segments/metric-hub-main")
```
mikewilli, line 52: @jaredlockhart This path should be valid in all environments, right?

mikewilli (commented):
Looks good overall, just a couple comments and want to confirm that the absolute path is valid everywhere.

experimenter/experimenter/segments/__init__.py
brennie: We don't do `class _PrivateCls` anywhere else in the tree. I don't think we should start now.

```
~/W/s/g/m/experimenter $ rg "class _"
experimenter/typings/graphene/pyutils/signature.pyi
20:class _void:
25:class _empty: ...
27:class _ParameterKind(int):

experimenter/typings/graphene/utils/props.pyi
5:class _OldClass: ...
6:class _NewClass: ...

experimenter/typings/factory/declarations.pyi
63:class _UNSPECIFIED: ...
203:class _FactoryWrapper:

experimenter/typings/factory/base.pyi
98:class _Counter:

experimenter/typings/celery/_state.pyi
36:class _TLS(threading.local):

experimenter/typings/import_export/forms.pyi
31:    class _ExportActionForm(ActionForm):

experimenter/typings/celery/beat.pyi
218:class _Threaded(Thread):

experimenter/typings/celery/canvas.pyi
248:class _chain(Signature):
354:class _basemap(Signature):
462:class _chord(Signature):

experimenter/typings/celery/utils/functional.pyi
150:class _regen(UserList, list):

experimenter/typings/celery/utils/threads.pyi
69:class _LocalStack:
153:class _FastLocalStack(threading.local):

experimenter/typings/celery/utils/abstract.pyi
11:class _AbstractClass(metaclass=ABCMeta):

experimenter/typings/celery/bin/logtool.pyi
17:class _task_counts(list):

experimenter/typings/celery/utils/time.pyi
53:class _Zone:

experimenter/typings/celery/utils/deprecated.pyi
39:class _deprecated_property:
````

File: experimenter/experimenter/segments/__init__.py [hunk 3]
```diff
@@ -7,6 +9,22 @@
 from experimenter.experiments.constants import NimbusConstants
 
 
+@dataclass
+class MockCommit:
+    committed_date: int
```
brennie, line 14: Looking at metric_config_parser, this doesnt look like it has enough information:


```python
        for outcome_file in files_path.glob(f"{OUTCOMES_DIR}/*/*.toml"):
            commit_hash = next(repo.iter_commits("HEAD", paths=outcome_file)).hexsha
```

[source](
https://github.com/mozilla/metric-hub/blob/7321591a45793f61f5a63d301eb24688374dfb9a/lib/metric-config-parser/metric_config_parser/config.py#L424)

experimenter/experimenter/segments/__init__.py
brennie: We probably want this to be a setting in settings.py so that it can be configurable.

experimenter/experimenter/segments/__init__.py
brennie: Can I suggest we add something to metric_config_parser so that this isn't necessary? This seems like the wrong layer of abstraction.

experimenter/experimenter/segments/__init__.py
mikewilli: Oh yea good point. It is in settings already (`METRIC_HUB_SEGMENTS_PATH_DEFAULT` though value is slightly different now), but I think this got a bit jumbled when we moved things around in the last PR. Pretty sure we can just modify that setting and use it here.

experimenter/experimenter/segments/__init__.py
mikewilli: `commit_hash` isn't required for `Outcome`, but I've come around to the "just do this in metric-config-parser" now, so no longer important.

experimenter/experimenter/segments/__init__.py
mikewilli: Yea.. I was willing to accept this as an alternative to creating new functionality just for this one use-case, but I think you're right. I'll go ahead and add something on the m-c-p side that we can call here without these mocks.

mikewilli: I referenced this PR in the metric-config-parser PR, but just for completeness: https://github.com/mozilla/metric-hub/pull/683

mikewilli:
Oh another thought that just occurred to me -- we can remove the `git` installation from the Dockerfile now, right?

mikewilli (approved):
Woo LGTM

PR #11773, opened 2024-11-14
bug(nimbus): Prevent mismatches between NimbusReviewSerializer and the front-end allowing invalid experiments

### LINKED ISSUE #11767: NimbusReviewSerializer is busted, letting experiments be published without full validation
#11283 added the `segments` field to the experiments model:

```python
    segments = ArrayField(
        models.CharField(max_length=255),
        default=list,
        verbose_name="Segments",
    )
```

but did not add this field to the `NimbusReviewSerializer`. Thus, when saving the front-end, we end up with the following response in the graphQL payload:

```json
      "readyForReview": {
        "ready": false,
        "message": {
          "segments": [
            "This list may not be empty."
          ]
        },
        "warnings": {},
        "__typename": "NimbusReviewType"
      },
```

However, our [useReviewCheck hook](https://github.com/mozilla/experimenter/blob/41804a1566451fba8b21cdd99823881ddd35c6c2/experimenter/experimenter/nimbus-ui/src/hooks/useReviewCheck.tsx#L13-L42) does not map this field to a specific page because this field is not editable in the current UI, only the new UI. PageSummary then passes the invalidPages prop from useReviewCheck down to ChangeApprovalOperations which uses [that prop](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/nimbus-ui/src/components/ChangeApprovalOperations/index.tsx#L69-L71) to determine whether or not we should be able to launch experiments.

This was not caught by our API unit tests because in our factories for generating test data, we always [set segments to a list](https://github.com/mozilla/experimenter/blob/41804a1566451fba8b21cdd99823881ddd35c6c2/experimenter/experimenter/experiments/tests/factories.py#L462-L464), which prevents the error from triggering. Changing this attribute to be an empty list results in 585 test failures and 12 errors.

The fix for this is as follows:

- Add segments to the NimbusReviewSerializer with allow_empty=True
- Change the factory to default to an empty list of segments
- Change the frontend to prevent publishing if readyForReview.ready !== true

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4973)

brennie:
Because:

- #112832 added the `segments` field to the `NimbusExperiment` model.
- It did not add it to the `NimbusReviewSerializer`, which caused the serializer to infer a default serializer for the field which marks it as a required field.
- Therefore, the `NimbusReviewSerializer` was always throwing **before** we hit its `validate()` method (due to the invalid `segments` field), which prevented the majority of the validation from running.
- And the front-end was not checking the `ready` property from the `useReviewCheck()` hook, which allowed launching invalid experiments.
- Our tests were always creating `NimbusExperiment`s with non-empty `segments` fields.

This commit:

- Adds a `segments` field to the `NimbusReviewSerializer` which makes the field optional.
- Adds a test that explicitly creates a `NimbusExperiment` with an empty `segments` field and verifies it does not raise a `ValidationError`. (This test fails without the changes to the serializer.)
- Changes our `NimbusExperimentFactory` to randomly use between 0 and 2 values for `segments`, `primary_outcomes`, and `secondary_outcomes` so that we have broader test coverage over all possible inputs.
- Updates our front-end to use the `ready` key instead of `invalidPages` to prevent launches.
- Updates the front-end to instruct users to come to #ask-experimenter if `ready` is false but there are no invalid pages (i.e., they've hit a programming error).

Fixes #11767

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1206,6 +1206,10 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         allow_null=False,
         error_messages={"null": NimbusConstants.ERROR_REQUIRED_QUESTION},
     )
+    segments = serializers.ListField(
+        child=serializers.CharField(),
+        required=False,
+    )
     is_localized = serializers.BooleanField(required=False)
     localizations = serializers.CharField(
         required=False, allow_blank=True, allow_null=True
--- a/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -3158,6 +3158,23 @@ class TestNimbusReviewSerializerSingleFeature(MockFmlErrorMixin, TestCase):
                 },
             )
 
+    def test_empty_segments(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            targeting_config_slug=NimbusExperiment.TargetingConfig.NO_TARGETING,
+            application=NimbusExperiment.Application.DESKTOP,
+            firefox_min_version=NimbusConstants.MIN_REQUIRED_VERSION,
+            segments=[],
+        )
+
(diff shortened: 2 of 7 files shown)


jaredlockhart (approved):
Okay I tested everything locally, all looks and works good!

Thanks for catching and picking this up so quickly @brennie 🙏 🙏 🙏

PR #11774, opened 2024-11-14
chore(tests): Add error catch for SDK Targeting helper methods on startup.

### LINKED ISSUE #11766: Intermittent integration test failure: Test SDK Targeting (Release Firefox)
Seems we're getting a lot of intermittent failures in the SDK targeting tests.  It looks like it's attempting to fetch the configs from graphql before the python process is finished starting which returns invalid JSON.  We should probably add more retries and delays so that it's resilient to the python process taking varying amounts of time to start.

Examples:

https://app.circleci.com/pipelines/github/mozilla/experimenter/47181/workflows/3ae867d8-6abd-4184-855b-a0d7cb972373/jobs/256872

https://app.circleci.com/pipelines/github/mozilla/experimenter/47180/workflows/9ee9d702-b062-4bf8-aba7-319756c300ce/jobs/256861

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4972)

b4handjr:
Because

- There have been intermittent errors on CI within the SDK targeting tests. These are due to the test making a query to the graphQL API before it is ready.

This commit

- Adds a retry within the called function to attempt to query the API again if it fails.

Fixes #11766

### DIFF
--- a/experimenter/tests/integration/nimbus/utils/helpers.py
+++ b/experimenter/tests/integration/nimbus/utils/helpers.py
@@ -9,7 +9,7 @@ from nimbus.models.base_dataclass import (
     BaseExperimentApplications,
 )
 
-LOAD_DATA_RETRIES = 10
+LOAD_DATA_RETRIES = 30
 LOAD_DATA_RETRY_DELAY = 1.0
 
 


experimenter/tests/integration/nimbus/utils/helpers.py
```diff
@@ -142,7 +142,16 @@ def load_config_data():
...
-    config_data = load_config_data()
+    config_data = []
+
+    for retry in range(LOAD_DATA_RETRIES):
+        try:
+            config_data = load_config_data()
+        except requests.exceptions.JSONDecodeError:
+            if retry + 1 >= LOAD_DATA_RETRIES:
+                raise
+            time.sleep(LOAD_DATA_RETRY_DELAY)
```
jaredlockhart, line 153: But we're already retrying in the `load_graphql_data` function, which this calls, shouldn't we just increase the number of retries?

b4handjr: Oh yeah, you're right. For some reason I thought this was failing instantly but I see now it's about 10 seconds which is the length of the retries we have. Ill just increase that.

jaredlockhart (approved):
Perf ty 🙏

PR #11800, opened 2024-11-19
fix(nimbus): change jetstream fetch to integer scheduling

### LINKED ISSUE #11799: fetch_jetstream_data task not firing on prod
All the other scheduled tasks are working fine, so maybe related to the crontab-style scheduling. Although it does work on a local deployment with both crontab and integer scheduling.

cc @jaredlockhart

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5000)

mikewilli:
Because

- jetstream fetch isn't firing
- jetstream fetch is the only scheduled task using crontab style schedule

This commit

- changes the jetstream fetch to integer-based schedule
- decreases the scheduled time from 6 to 4 hours to account for the fact that the integer-based schedule is interrupted by each deploy (i.e., first task fires 4 hours after deployment)

Fixes #11799

### DIFF
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -13,7 +13,6 @@ https://docs.djangoproject.com/en/1.9/ref/settings/
 from pathlib import Path
 from urllib.parse import urljoin
 
-from celery.schedules import crontab
 from decouple import config
 from django.contrib.admin import ModelAdmin, StackedInline, TabularInline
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
👍 Thnx @mikewilli fingers crossed this does it 🤞 

Also looking at the sentry logs for the popsizing task, since it'll only fire after 24 hours with no deploys looks like it's firing roughly once a week.  If that's okay then let's leave it, but if we want it updated more often we should think about bringing that down.

PR #11828, opened 2024-11-25
Revert "fix(nimbus): change jetstream fetch to integer scheduling"

mikewilli:
Reverts mozilla/experimenter#11800

This didn't resolve the issue, and we've determined that this was not actually the cause of the problem, so let's go back to the crontab-style scheduling.


jaredlockhart (approved):
👍

PR #11852, opened 2024-11-26
chore(nimbus): change pop sizing task to crontab scheduling

### LINKED ISSUE #11851: Switch popsizing celery fetch task to cron scheduling
Let's switch the `fetch_population_sizing_data` task to using cron style scheduling like the `fetch_jetstream_data` task so that it fires exactly once a week.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5020)

mikewilli:
Because

- Tasks with integer-based scheduling first fire <integer> seconds after the latest deploy
- Pop sizing fetch task is scheduled for 24 hours
- We often have multiple deployments per day, so 24 hours after the latest deploy will usually not occur until the weekend

This commit

- changes the scheduling to use crontab so that the task will fire at 06:00 UTC every day regardless of deployments

Fixes #11851

### DIFF
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -395,7 +395,7 @@ CELERY_BEAT_SCHEDULE = {
     },
     "fetch_population_sizing_data": {
         "task": "experimenter.jetstream.tasks.fetch_population_sizing_data",
-        "schedule": 86400,
+        "schedule": crontab(minute=0, hour=6, day_of_week=1),
     },
 }
 


experimenter/experimenter/settings.py
```diff
@@ -395,7 +395,7 @@
     },
     "fetch_population_sizing_data": {
         "task": "experimenter.jetstream.tasks.fetch_population_sizing_data",
-        "schedule": 86400,
+        "schedule": crontab(minute=0, hour=6),
```
jaredlockhart, line 398: I htought we wanted this to be weekly?

mikewilli: That's fine too, I was debating. I'll update to weekly.

mikewilli: Ok done. I changed to only execute on Mondays now (the job runs Sunday).

jaredlockhart (approved):
Awesome ty 🙏

PR #11819, opened 2024-11-21
feat(cirrus): Update cirrus to include features

yashikakhurana:
Because

- We want to return features under a single key called `Features`

This commit

- Update the return format to include new key in the response

Fixes #11810

PR #11860, opened 2024-11-26
feat(cirrus): Send enrollment key in a response

### LINKED ISSUE #11814: Cirrus returns the enrollment data in the response
update cirrus to return this as a format [https://gist.github.com/yashikakhurana/f2fec74b0b6182d03adf957ceedd41b3](https://gist.github.com/yashikakhurana/f2fec74b0b6182d03adf957ceedd41b3|smart-link) for enrollment data

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5005)

yashikakhurana:
Because

- We don't send back enrollment responses to the calling application and sometimes its hard to find if they are in the experiment or not

This commit

- Return the enrollment data along with the features

Fixes #11814

### DIFF
--- a/cirrus/server/cirrus/main.py
+++ b/cirrus/server/cirrus/main.py
@@ -161,53 +161,55 @@ def initialize_glean():
 
 
 class EnrollmentMetricData(NamedTuple):
+    nimbus_user_id: str
+    app_id: str
     experiment_slug: str
     branch_slug: str
     experiment_type: str
+    is_preview: bool
 
 
 def collate_enrollment_metric_data(
-    enrolled_partial_configuration: dict[str, Any], nimbus_preview_flag: bool
+    enrolled_partial_configuration: dict[str, Any],
+    client_id: str,
+    nimbus_preview_flag: bool,
 ) -> list[EnrollmentMetricData]:
     events: list[dict[str, Any]] = enrolled_partial_configuration.get("events", [])
+    remote_settings = (
+        app.state.remote_setting_preview
+        if nimbus_preview_flag
+        else app.state.remote_setting_live
+    )
+
     data: list[EnrollmentMetricData] = []
     for event in events:
         if event.get("change") == "Enrollment":
             experiment_slug = event.get("experiment_slug", "")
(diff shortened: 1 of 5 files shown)


yashikakhurana:
Blocked on https://github.com/mozilla/experimenter/pull/11819

cirrus/server/tests/test_main.py
```diff
@@ -508,18 +522,67 @@ def test_get_features_with_and_without_nimbus_preview(
             ],
         }
 
-        # Without nimbus_preview
```
jaredlockhart, line 511: Each of these cases should really be their own individual test case rather than folded into one.

jaredlockhart (approved):
This looks really great!  Thanks for getting through all these changes 🙏

cirrus/server/tests/test_main.py
yashikakhurana: good idea, will split this one

cirrus/server/cirrus/main.py
```diff
@@ -256,14 +258,32 @@ async def compute_features(
...
+        nimbus_preview_flag=nimbus_preview,
     )
 
-    return client_feature_configuration
+    # Record metrics
+    await record_metrics(enrollment_data)
+
+    response = {
+        "Features": client_feature_configuration,
+        "Enrollments": [
```
brennie, line 273: why capital keys?

yashikakhurana: no specific reason, its something we agreed upon as part of the plan

jaredlockhart (approved):
Oh let's move this to a `/v2/` endpoint and leave the existing structure intact in `/v1/` that way we can land this without breaking downstream consumers 👍

jaredlockhart (changes requested):
Oh let's move this to a /v2/ endpoint and leave the existing structure intact in /v1/ that way we can land this without breaking downstream consumers 👍

yashikakhurana:
closing this now, will add other work

PR #11931, opened 2024-12-11
chore(project): Add ci workflow rfc to repo.

### LINKED ISSUE #11930: Add accepted proposal for ci workflow changes to experimenter repo as an ADR
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4762)

b4handjr:
Because

- We are keeping track of various proposals/adrs/rfc/etc in the repo

This commit

- Adds the CI workflow update RFC.

Fixes #11930


jaredlockhart (approved):
Thanks for writing this up @b4handjr 🙏 It might make more sense as a README in the integration test folder rather than an ADR?  But either way good to keep a record of the thinking that went into it 👍

b4handjr:
> Thanks for writing this up @b4handjr 🙏 It might make more sense as a README in the integration test folder rather than an ADR? But either way good to keep a record of the thinking that went into it 👍

Fair, but I'd like it to be visible with all of the other nimbus related stuff too. I can move it if need be

PR #11932, opened 2024-12-11
feat(nimbus): Draft/Preview/Review workflow

### LINKED ISSUE #11757: Draft to preview and preview to draft
On new summary page lets add the option to send it to preview and from preview to draft back

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4964)

yashikakhurana:
Because

We want to include workflow on the new summary page: 
- Send draft to preview
-  preview to review
- preview to draft
- review to draft 
- cancel review

This commit

- Adds the preview, review workflow

Fixes #11757 

https://github.com/user-attachments/assets/cb7978af-aa85-45a3-a40d-0d890c94bb54


yashikakhurana:
wait for review, fixing some layouts

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -335,6 +335,9 @@ class Meta:
     def save(self, commit=True):
         experiment = super().save(commit=commit)
         experiment.subscribers.add(self.request.user)
+        generate_nimbus_changelog(
+            experiment, self.request.user, self.get_changelog_message()
+        )
```
jaredlockhart, line 340: Why add this here?  This is already done by `NimbusChangeLogFormMixin`?

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/launch_with_preview_controls.html
```diff
@@ -0,0 +1,59 @@
...
+<div class="alert alert-success bg-transparent text-success">
+  <p class="my-1">
+    <svg class="align-top"
+         xmlns="http://www.w3.org/2000/svg"
+         fill="currentColor"
+         viewBox="0 0 16 16"
+         width="16"
+         height="16">
+      <path d="M13.485 1.4a.5.5 0 0 1 .035.71L6.543 9.793 3.379 6.93a.5.5 0 1 1 .708-.707l2.1 2.1 6.514-7.1a.5.5 0 0 1 .707-.035z" />
+    </svg>
```
jaredlockhart, line 11: We should be able to use fontawesome here?

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -264,3 +271,93 @@ class UnsubscribeView(
 ):
     form_class = UnsubscribeForm
     template_name = "nimbus_experiments/subscribers_list.html"
+
+class TimelineAndControlsMixin:
```
jaredlockhart, line 275: So I think a different way to do this is to move all this logic directly into the summary template, and then use hx-select and hx-target to just update the controls/timeline, rather than doing logic in the views here.

jaredlockhart (changes requested):
Cool!  I tried locally and it works great!  Needs to be updated to latest main and I think we can do the views/controls differently, but this is coming along great 🎉 🎉 🎉 🎉

experimenter/experimenter/nimbus_ui_new/forms.py
yashikakhurana: oh if I am overriding the save method, I noticed that it is not generating the changelogs, so for that, I needed to explicitly called the generate changelogs method

experimenter/experimenter/nimbus_ui_new/static/js/control.js
```diff
@@ -0,0 +1,14 @@
+window.showRecommendation = function () {
```
jaredlockhart, line 1: Yep, nice simple little js script, but 'control.js' isn't a super descriptive name, maybe something more explicit like 'review_controls.js' or something? 🙏

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -272,3 +279,27 @@ class UnsubscribeView(
 ):
     form_class = UnsubscribeForm
     template_name = "nimbus_experiments/subscribers_list.html"
+
+
+class StatusUpdateView(RequestFormMixin, RenderResponseMixin, NimbusExperimentDetailView):
+    fields = None
```
jaredlockhart, line 285: Why do we need `fields=None` here?

experimenter/experimenter/nimbus_ui_new/tests/test_views.py
```diff
@@ -1288,6 +1288,80 @@ def test_post_updates_metrics_and_segments(self):
         self.assertEqual(experiment.segments, [segment1.slug, segment2.slug])
 
 
+class TestLaunchViews(AuthTestCase):
+    def setUp(self):
+        super().setUp()
+        self.experiment = NimbusExperimentFactory.create()
+
+    def test_draft_to_preview(self):
```
jaredlockhart, line 1296: Yep nice succinct little tests, but we should explicitly set the three status fields values before we call into the form, ie we're asserting on their values after we call but not before so the experiment might've already had that value and the form could not write it and the test would still pass 🙏

jaredlockhart (changes requested):
Ya this is looking great.  Very nice and succinct changes, great tests, works great locally.  Couple last points about the tests but this is pretty much there, amazing thank you @yashikakhurana 🙏 🎉

jaredlockhart:
Oh also there's a little gap on the right of the timeline here, see if you can get rid of that:

image

yashikakhurana:
> Oh also there's a little gap on the right of the timeline here, see if you can get rid of that:
> image

Fixed it 
Screenshot 2025-01-14 at 3 57 50 PM

experimenter/experimenter/nimbus_ui_new/tests/test_forms.py
```diff
@@ -322,11 +327,93 @@ def test_unsubscribe_form_removes_subscriber(self):
...
         self.assertEqual(changelog.changed_by, self.user)
         self.assertIn("dev@example.com removed subscriber", changelog.message)
 
 
+class TestLaunchForms(RequestFormTestCase):
+    def setUp(self):
+        super().setUp()
+        self.experiment = NimbusExperimentFactory.create()
+
+    def test_draft_to_preview_form(self):
```
jaredlockhart, line 340: Oh all these tests need to setup the status/status_next/publish_status fields to some known values before we update them and assert on them, otherwise they could already be in those states, the form could do nothing, and they'd still pass 🙏

jaredlockhart (changes requested):
Yep this is all looking and working great, one last pass on the tests and this should be good to go 🎉

jaredlockhart (approved):
Perfect ty for updating the tests, this is great, let's gooooo 🎉 🎉 🎉 🎉

PR #11965, opened 2024-12-17
bug(nimbus): Filter versioned schemas inclusively

### LINKED ISSUE #11002: No feature config schemas found for min==max version
If you set the min and max versions to the same value, the review serializer will complain it can't find any feature configs within that version range.  However setting min and max is valid from targeting since the targeting expression is `min <= version <= max` so we should support finding feature config schemas within a min==max range.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4692)

brennie:
Because:

- the firefox_max_version logic uses inclusive ranges; and
- filtering schema ranges was using exclusive ranges, which is not correct behaviour

This commit

- updates schema ranges to be inclusive

Fixes #11002

### DIFF
--- a/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
@@ -2466,7 +2466,7 @@ class TestNimbusReviewSerializerSingleFeature(MockFmlErrorMixin, TestCase):
                 experiment,
                 context={"user": self.user},
             ).data,
-            context={"user", self.user},
+            context={"user": self.user},
             partial=True,
         )
 
@@ -3074,7 +3074,7 @@ class TestNimbusReviewSerializerSingleFeature(MockFmlErrorMixin, TestCase):
         serializer = NimbusReviewSerializer(
             experiment,
             data=NimbusReviewSerializer(experiment, context={"user": self.user}).data,
-            context={"user", self.user},
+            context={"user": self.user},
         )
 
         self.assertTrue(serializer.is_valid(), serializer.errors)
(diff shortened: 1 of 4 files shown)


jaredlockhart (commented):
Can you highlight which test now covers the case where min==max and it successfully validates?

brennie:
> Can you highlight which test now covers the case where min==max and it successfully validates?

I added a new test that covers min == max, min <= max (bounded), and min <= max (unbounded). It turns out we really weren't covering the success cases that well before.

https://github.com/mozilla/experimenter/pull/11965/files#diff-6f56746f1db5e4c83a518bc30ced77793968dcd2925454bc521e372e6aa689a7R3215

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -2466,7 +2466,7 @@ def test_not_localized_with_localizations(self):
                 experiment,
                 context={"user": self.user},
             ).data,
-            context={"user", self.user},
+            context={"user": self.user},
```
jaredlockhart, line 2469: Oh dang.  Annoying that `{a, b}` is valid syntax now.  Surprising nothing complained about not being able to find the user.... Oh you know why, cuz that's used to populate the user in the changelog generator and the review serializer doesn't save.

jaredlockhart (approved):
Okay perfect thank you for the additional tests, looks good, tested locally everything looks great, thnx @brennie 🎉 

image

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
brennie: I filed a bug about us refactoring this common pattern a while ago so we cant make this mistake again.

PR #11973, opened 2024-12-19
chore(ci): Fix ios log path for artifact fetching on CI.

b4handjr:
Because

- The log path changed recently on our iOS CI job (maybe due to new macos images on circleci).

This commit

- Changes the path to be a bit more generic.

Fixes #11973


jaredlockhart (approved):
Ah good catch, also missing the github bug number 🙏  Thanks @b4handjr

PR #11985, opened 2024-12-28
stop application edit after creation

### LINKED ISSUE #7083: Application should not be editable after creation
I just noticed a test was updating application for a created experiment but that shouldn't be possible since that can invalidate other fields.  Application should only be saveable on creation and our logic should prevent it from being changed, even in tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-2215)

olamidepeterojo:
Because

- it shouldn't be possible to edit applications after creation since that can invalidate other fields.

This commit

- will stop application from being edited after creation

Fixes #7083

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -234,6 +234,11 @@ class NimbusBranchSerializer(serializers.ModelSerializer):
 
     def validate(self, data):
         data = super().validate(data)
+        
+        if self.instance and "application" in data:
+            raise serializers.ValidationError(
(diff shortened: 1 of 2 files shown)


b4handjr:
Hello, thanks for this PR! Could you check the failing tests runs and get those fixed up? Thanks!

jaredlockhart:
Hi @olamidepeterojo ! Thanks for the submission.  Looks like there's some unresolved failures, so I'll close this for now, feel free to pick it back up any time and reopen it.  Looks like the failures are just formatting which you can resolve with `make code_format` and resubmit 👍 🙏

PR #11989, opened 2025-01-03
feat: split start/end dates into separate columns on main list

### LINKED ISSUE #12007: Allow sort by end date on main experiments list
Probably by splitting out the "Dates" column into start and end date.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5086)

mikewilli:
Because

- start and end date are different fields
- it would be nice to sort on start/end date separately

This commit

- splits the start and end dates into separate columns
- indicates proposed vs actual end date with italics for proposed
- annotates the NimbusExperiment queryset with a sort_end_date so that we can sort on proposed end date if experiment is live

Fixes #12007 

image
image


experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/table.html
```diff
@@ -109,11 +110,16 @@ <h1 class="modal-title fs-5" id="createModalLabel">
...
+              {{ experiment.start_date|date:"Y-m-d" }}
+            {% else %}
+              N/A
+            {% endif %}
+          </td>
+          <td>
+            {% if experiment.end_date %}
+              {{ experiment.end_date|date:"Y-m-d" }} ({{ experiment.computed_duration_days }} days)
+            {% elif experiment.proposed_end_date %}
+              <i>{{ experiment.proposed_end_date|date:"Y-m-d" }} ({{ experiment.computed_duration_days }} days)</i>
```
jaredlockhart, line 122: Oh no, we should just display `computed_end_date` here.  If we start splitting up different logic then we'll wind up with different dates displayed in different places/pages and it gets confusing.

jaredlockhart: Also I get your point about using italics here but we don't use that anywhere else to mean that, so I'd take that out for consistency.

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -62,7 +64,17 @@ class NimbusChangeLogsView(NimbusExperimentViewMixin, DetailView):
...
+        NimbusExperiment.objects.annotate(
+            sort_end_date=Coalesce(
+                F("_end_date"),
+                ExpressionWrapper(
+                    Coalesce(F("proposed_release_date"), F("_start_date"))
+                    + F("proposed_duration"),
+                    output_field=DateField(),
+                ),
+            )
+        )
```
jaredlockhart, line 76: Yeah also here we should just be sorting in python space on teh `computed_end_date` field so that we don't have multiple parallel definitions of what 'computed_end_date' means.

jaredlockhart (changes requested):
Yep good start, couple thoughts, thnx @mikewilli 🙏

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/table.html
mikewilli: Yep I think I did this because I wasn't able to do sorting in the same way using computed properties, so I made the display consistent with the sorting. But if we can just sort in python where the `computed_end_date` is available then that's definitely better!

Also yes totally fine re: italics. Even less important if we land https://github.com/mozilla/experimenter/pull/11988

experimenter/experimenter/nimbus_ui_new/filtersets.py
```diff
@@ -258,6 +262,33 @@ class Meta:
...
+                reverse=reverse,
+            )
+
+            preserved_order = Case(
+                *[
+                    When(pk=exp.id, then=Value(pos))
+                    for pos, exp in enumerate(queryset_sorted)
+                ],
+                output_field=IntegerField()
+            )
```
jaredlockhart, line 286: Huh!  That's really interesting!  I've always wondered how to do that, I've never seen this pattern before, but it makes perfect sense.  Great job!

jaredlockhart (approved):
Okay this all looks right to me and tested locally works great.  Logic is clear, UI is clear, test is well written, great change great job @RJAK11 🎉 🎉 🎉  and thnx @mikewilli for kicking this off 🙏

jaredlockhart:
Blocked on https://github.com/mozilla/experimenter/issues/12105

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/table.html
```diff
@@ -110,11 +111,14 @@ <h1 class="modal-title fs-5" id="createModalLabel">
...
-              <br>
-              ({{ experiment.computed_duration_days }} days)
+              {{ experiment.start_date|date:"Y-m-d" }}
+            {% else %}
+              N/A
+            {% endif %}
+          </td>
+          <td>
+            {% if experiment.computed_end_date %}
+              {{ experiment.computed_end_date|date:"Y-m-d" }}({{ experiment.computed_duration_days }} days)
```
mikewilli, line 121: Just noticed this typo:
```suggestion
              {{ experiment.computed_end_date|date:"Y-m-d" }} ({{ experiment.computed_duration_days }} days)
```

jaredlockhart: Hrmmm I can't seem to apply this change, probably because it's on @RJAK11 's fork.  Let's just land this and we can do that as a followup when @RJAK11 is back 🙏

mikewilli: It's my fork so I could probably do it, but that's fine too -- it's just a space!

PR #11990, opened 2025-01-03
feat: add results link to experiments table

### LINKED ISSUE #10946: Add results link to experiment row when results are available
For the new HTMX UI, it would be nice to continue showing the results link when results are available for each experiment row (or at least some sort of indication that results are available).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4666)

mikewilli:
Because

- it is impossible to tell if an experiment has results without clicking into it
- it can be useful to see a list of experiments with/without results
- it can be useful to navigate directly to the experiment's results

This commit

- adds a results link to the experiments table if results exist

Fixes #10946 
image


experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/table.html
```diff
@@ -118,6 +119,11 @@ <h1 class="modal-title fs-5" id="createModalLabel">
               N/A
             {% endif %}
           </td>
+          <td>
+            {% if experiment.show_results_url %}
+              <a href="{% url "nimbus-results" slug=experiment.slug %}">Results</a>
+            {% endif %}
```
yashikakhurana, line 125: nice 👍 , just in case if results are not available do you want to say not available or may be some other placeholder?

mikewilli: If I remember right, we previously left the links column blank if there were no links to display, but I'm happy to put "N/A" or something else here if you think that would be better. What do you think?

yashikakhurana: I would just change the text from `Result` to `View Results` and if they are not available link won't be active, but I think what you have clearly tells if the results are available or not, or just simply put N/A if its not available

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/table.html
```diff
@@ -118,6 +119,13 @@ <h1 class="modal-title fs-5" id="createModalLabel">
               N/A
             {% endif %}
           </td>
+          <td>
+            {% if experiment.show_results_url %}
+              <a href="{% url "nimbus-results" slug=experiment.slug %}">View Results</a>
+            {% else %}
+              <a>View Results</a>
```
jaredlockhart, line 126: Oh maybe just changed this to `N/A` 🙏

PR #11999, opened 2025-01-07
chore(project): Pin poetry version to 1.8.4

### LINKED ISSUE #11998: Pin poetry install command to 1.8.4 for now
Poetry recently introduced some breaking changes to our workflow with version 2.0.0. Let's pin to version 1.8.4, and update when we need to.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5079)

b4handjr:
Because

- Poetry introduced some breaking changes with their v2 release

This commit

- Pins all poetry installs to 1.8.4 for now.

Fixes #11998

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -304,7 +304,7 @@ jobs:
           name: Download test files and run
           command: |
             curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/mozilla/application-services/main/install-nimbus-cli.sh | bash
-            pip install poetry
+            pip install poetry=1.8.4
             make integration_test_nimbus_fenix PYTEST_ARGS="$PYTEST_ARGS"
       - android/save-gradle-cache:
           cache-prefix: v1a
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
LGTM thanks!

I'd suggest 1.8.4 for this change because we use that in all of the other Dockerfiles (except Cirrus which is on 1.6.0?), or updating those to match, but don't think that's a big deal.

We may want to consider using `POETRY_VERSION` env var as well in the future.

Want to file a ticket for whichever of the above you don't include here?

b4handjr:
> LGTM thanks!
> 
> I'd suggest 1.8.4 for this change because we use that in all of the other Dockerfiles (except Cirrus which is on 1.6.0?), or updating those to match, but don't think that's a big deal.
> 
> We may want to consider using `POETRY_VERSION` env var as well in the future.
> 
> Want to file a ticket for whichever of the above you don't include here?

Good catch, ill do 1.8.4 then. Not a big difference between the 2

PR #12008, opened 2025-01-08
feat(cirrus): V2 api

### LINKED ISSUE #12000: v2 api-Cirrus returns the enrollment data in the response
update cirrus to return this as a format [https://gist.github.com/yashikakhurana/f2fec74b0b6182d03adf957ceedd41b3](https://gist.github.com/yashikakhurana/f2fec74b0b6182d03adf957ceedd41b3|smart-link) for enrollment data

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5081)

yashikakhurana:
Because

- We want to return features under the key `Features`
- We don't send back enrollment responses to the calling application and sometimes its hard to find if they are in the experiment or not

This commit

- Returns feature and enrollments as part of the new `v2 api`

Fixes #12000 #12001


cirrus/server/tests/test_main.py
```diff
@@ -49,7 +50,26 @@ def test_read_root(client):
...
+        (
+            "/v2/features/",
+            {
+                "Features": {
+                    "example-feature": {"enabled": False, "something": "wicked"},
+                    # return default features
+                },
+                "Enrollments": [],
+            },
+        ),
```
jaredlockhart, line 69: I get why you'd fold them together like this but I think we should keep the v1 and v2 tests completely isolated from one another as a way to say 'these changes do not impact the design or behaviour of the v1 endpoints' so we know that by making these changes we aren't impacting the v1 endpoint in any way.

cirrus/server/tests/test_main.py
```diff
@@ -523,7 +714,165 @@ def test_get_features_with_and_without_nimbus_preview(
...
+                {
+                    "nimbus_user_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449",
+                    "app_id": "test_app_id",
+                    "experiment": "experiment_slug_2",
+                    "branch": "treatment",
+                    "experiment_type": "experiment",
+                    "is_preview": True,
+                }
+            ],
+        }
```
jaredlockhart, line 869: In general we should try to keep tests to testing a single case in each one test.  This is testing two cases, we should split that into two separate tests.

jaredlockhart (dismissed):
Yep logic all looks good, we should try to split the tests up to more clearly separate the cases.  Thanks @yashikakhurana !

demo-app/server/index.js
```diff
@@ -29,7 +29,7 @@ const server = http.createServer(async (req, res) => {
       nimbus_preview: nimbusPreview,
     };
 
-    let pathName = '/v1/features/'
```
jaredlockhart, line 32: Actually just out of an abundance of caution, can we keep the test app on v1 so we ensure that v1 completely remains intact while landing this and we can bump it to v2 later? 🙏

jaredlockhart (dismissed):
Okay, everything looks great, thank you for the test changes, they're very clear and read nicely.  Just one last thing to make sure we keep v1 intact and then I think this is good to go 🙏

jaredlockhart (approved):
Let's goooooo, this is fantastic thanks so much @yashikakhurana 🥳 🎉 🎉 🎉 🎉

PR #12015, opened 2025-01-09
feat(nimbus): hide archived experiments from require/exclude fields

### LINKED ISSUE #12013: Need to remove an experiment from include/exclude options
we archived one long term holdback experiment and recreated with a very similar name (just no space between 2025H1).

This isn't uncommon - so is there a way to make sure we do not show archived experiments in the include/exclude?

We need to NOT see the archived name in the drop down - as for the whole year people could easily select the Archived holdback and have their experiments fail to enroll anyone.  

Archived (has a space) Long term holdback 2025 H1 Growth Desktop[https://experimenter.services.mozilla.com/nimbus/long-term-holdback-2025-h1-growth-desktop/](https://experimenter.services.mozilla.com/nimbus/long-term-holdback-2025-h1-growth-desktop/)

Live (no space 2025H1) Long term holdback 2025H1 Growth Desktop [https://experimenter.services.mozilla.com/nimbus/long-term-holdback-2025h1-growth-desktop/](https://experimenter.services.mozilla.com/nimbus/long-term-holdback-2025h1-growth-desktop/)

~~It looks like Shell can edit the Name and Slug~~  but I'm asking permission before doing so as I've never changed those and aside from making sure they aren't replicated... I don't know if I could break things.-  NO - bad shell - never do this!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5083)

jaredlockhart:
Because

* We need a way to hide some experiments from the require/exclude field so users don't accidentally select the wrong one when they have nearly identical names
* The simplest way to do that for now is to just use archiving as a way to hide them from the drop downs

This commit

* Hides archived experiments from the require/exclude drop downs in both the old and new Nimbus UI

fixes #12013

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageEditAudience/FormAudience/index.test.tsx
@@ -150,6 +150,61 @@ describe("FormAudience", () => {
       ).toBeUndefined();
     });
 
+    it("only hides archived experiments from required and excluded drop downs", async () => {
+      const experiment = {
+        ...MOCK_EXPERIMENT,
+        name: MOCK_EXPERIMENTS_BY_APPLICATION[0].name,
+        slug: MOCK_EXPERIMENTS_BY_APPLICATION[0].slug,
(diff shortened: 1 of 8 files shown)


brennie (approved):
Shame isArchived is typed as boolean | null

r+

PR #12020, opened 2025-01-10
fix(nimbus): Make name field in new UI required

### LINKED ISSUE #12016: Empty name in new nimbus ui raises 500
@RJAK11 just noticed this:

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5088)

RJAK11:
Because

- Submitting a form with an empty name field in the new Nimbus UI causes a 500 error.

This commit

- Adds validation to ensure the name field is required and prevents submission with an empty value

Fixes #12016

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/tests/test_forms.py
+++ b/experimenter/experimenter/nimbus_ui_new/tests/test_forms.py
@@ -378,6 +378,31 @@ class TestOverviewForm(RequestFormTestCase):
         )
         self.assertEqual(documentation_link.link, "https://www.example.com")
 
+    def test_name_field_is_required(self):
+        project = ProjectFactory.create()
+        documentation_link = NimbusDocumentationLinkFactory.create()
+
+        form_data = {
+            "hypothesis": "new hypothesis",
+            "risk_brand": True,
+            "risk_message": True,
+            "projects": [project.id],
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus_ui_new/tests/test_forms.py
```diff
@@ -378,6 +378,31 @@ def test_valid_form_saves(self):
...
+            "documentation_links-0-id": documentation_link.id,
+            "documentation_links-0-title": (
+                NimbusExperiment.DocumentationLink.DESIGN_DOC.value
+            ),
+            "documentation_links-0-link": "https://www.example.com",
+        }
+
+        form = OverviewForm(data=form_data)
+
+        self.assertFalse(form.is_valid())
```
jaredlockhart, line 404: Cool, couple things here:

Instead of removing the 'name' key from the data altogether, try having it and setting it to an empty string. 

Also should explicitly check that the 'name' field is in`form.errors`

jaredlockhart (changes requested):
Good start!  Couple ideas for the test 🙏

jaredlockhart (approved):
Okay fantastic, tested locally, works perfectly, thnx @RJAK11 🎉

PR #12035, opened 2025-01-14
Application

### LINKED ISSUE #7083: Application should not be editable after creation

olamidepeterojo:
Because

- it shouldn't be possible to edit applications after creation since that can invalidate other fields.

This commit

- will stop application from being edited after creation

Fixes #7083


brennie (changes requested):
Hi! Thanks for taking this on. This is unfortunately not the correct way to go about this change. I've left you feedback on the correct way to implement this.

Additionally, this commit message does not meet our standards. Take a look at other commits in our history and look specifically at [the commit guidelines section of our contributing guide](https://github.com/mozilla/experimenter/blob/248a5c026942c07349a1a026b001b8daeaf80a0d/contributing.md?plain=1#L64).

yashikakhurana:
Hi @olamidepeterojo ! Thanks for the submission. We can't take this submission as this is on our old UI and we are working on transitioning it to the new UI, so I'll close this for now. Please feel free to work on non-UI task at the moment, we will be closing all the react UI changes tickets soon

PR #12038, opened 2025-01-14
feat(nimbus): Add custom targetting for address autofill

### LINKED ISSUE #12037: Add advanced targetting for address autofill rollout
For https://mozilla-hub.atlassian.net/browse/FXCM-1351 and, in the future, other experiments in other locales, we want to be able to target users who have not changed the preference 'extensions.formautofill.addresses.supported'. Valid values of this preference are 'on', 'detect' and 'off'.

- either the preference has not been changed from the default value
- or, the preference has been changed to a value other than 'on'
 
This mirrors the equivalent change made for credit cards done in https://github.com/mozilla/experimenter/issues/11702

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5095)

Enndeakin:
As with credit cards, target users that have not changed the value of the pref extensions.formautofill.addresses.supported or changed it to something other than 'on'

Fixes #12037


experimenter/experimenter/targeting/constants.py
```diff
@@ -2270,6 +2270,22 @@ def __post_init__(self):
...
+    ),
+    targeting=(
+        "!('extensions.formautofill.addresses.supported'|preferenceIsUserSet) || "
+        "'extensions.formautofill.addresses.supported'|preferenceValue != 'on'"
+    ),
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+)
```
brennie, line 2288: I think you just need to add a blank line after this to make the formatter happy. You can also run `make code_format` but that will take longer :P

jaredlockhart (approved):
Okay this should be good to go, let's get it landed 👍

PR #12039, opened 2025-01-14
feat(nimbus): Targeting for users not in HTTPS-Only mode

simon-friedberger:
Because

- this is needed for a timeout which is better tested in HTTPS-First

This commit

- adds an option to exclude HTTPS-Only users

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -2270,6 +2270,23 @@ DEFAULT_AUTOFILL_CREDIT_CARDS_SUPPORTED = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+NO_HTTPS_ONLY_DESKTOP = NimbusTargetingConfig(
+    name="Users who are not in HTTPS-Only Mode",
+    slug="no_https_only_desktop",
+    description=(
+        "Targets users who do not have HTTPS-Only Mode enabled. Neither with "
+        "'dom.security.https_only_mode' nor 'dom.security.https_only_mode_pbm'."
+    ),
+    targeting=(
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/targeting/constants.py
```diff
@@ -2270,6 +2270,23 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
+NO_HTTPS_ONLY = NimbusTargetingConfig(
```
brennie, line 2273: You might want to rename this to `NO_HTTPS_ONLY_DESKTOP` since youre adding a separate targeting for Fenix.

brennie (approved):

experimenter/experimenter/targeting/constants.py
```diff
@@ -2270,6 +2270,23 @@ def __post_init__(self):
...
+        "'dom.security.https_only_mode' nor 'dom.security.https_only_mode_pbm'."
+    ),
+    targeting=(
+        "!('dom.security.https_only_mode'|preferenceValue || "
+        "'dom.security.https_only_mode_pbm'|preferenceValue)"
+    ),
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name),
```
jaredlockhart, line 2287: This is missing a comma

```py
```

brennie:
CI failure seems spurious

brennie:
CI failure is not spurious -- please rebase on latest main and push again. the CI depends on TC artifacts, and it looks like your patch is trying to use an expired artifact

brennie:
(looks like I can rebase your PR automatically, so lets see if that fixed it)

PR #12053, opened 2025-01-15
feat(nimbus): push cirrus image to commit hash tag

### LINKED ISSUE #12052: Push cirrus container builds to tags with commit hash
Let's push Cirrus docker builds to tags with individual commit hashes the way we do for other containers, so that downstream consumers can hard code specific hashes rather than always using 'latest' in case they want to revert to some last known good build.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5104)

jaredlockhart:
Because

* In addition to pushing cirrus docker images to a latest tag, we should also push to a commit hash tag
* This will allow downstream consumers to hardcode a known working tag and either bump or revert as necessary

This commit

* Pushes cirrus docker image to a git commit hash tag in addition to latest

fixes #12052

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -495,6 +495,10 @@ jobs:
             docker tag cirrus:deploy ${DOCKERHUB_CIRRUS_REPO}:latest
             docker push ${DOCKERHUB_CIRRUS_REPO}:latest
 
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
Nice 🎉 , thanks @jaredlockhart

PR #12054, opened 2025-01-15
bug(nimbus): fix monitoring dashboard url timestamps

### LINKED ISSUE #12050: Live Monitoring Dashboard URL uses wrong time filter name
The link to the Live Monitoring Dashboard in the sidebar has a time range in the URL, so the dashboard should load pre-filtered to the date range of the experiment/rollout. However, the parameter is wrong, so instead it falls back to the default range of Last 28 Days.

To fix this, we need to change the URL params from `?Experiment=experiment-slug&Time+Range=2025-01-07+to+2025-01-15` to `?Experiment=experiment-slug&Time+Range+[UTC]=2025-01-07+to+2025-01-15` (add `+[UTC]`).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5102)

jaredlockhart:
Because

* We were passing the incorrect timestamp parameter to the monitoring dashboard url
* This resulted in the dashboard defaulting to the most recent 28 days instead of the proper time range

This commit

* Updates the monitoring dashboard url to use the correct timestamp parameter

fixes #12050

### DIFF
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -406,7 +406,7 @@ RECIPE_SLUG_MAX_LEN = 80
 MONITORING_URL = (
     # from_date and to_date format is YYYY-mm-dd
     "https://mozilla.cloud.looker.com/dashboards-next/216?"
-    "Experiment={slug}&Time+Range={from_date}+to+{to_date}"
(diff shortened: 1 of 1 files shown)


mikewilli (approved):
Nice thanks for the quick fix!

Issue #12065, opened 2025-01-17
Use DAU metric as guardrail for Firefox apps


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4413)

data-sync-user:
➤ Daniel Berry commented:

Mike Williams With https://github.com/mozilla/jetstream/pull/2047 ( https://github.com/mozilla/jetstream/pull/2047|smart-link )  merged, I believe we have what we need to change Experimenter to use the client_level_daily_active_users metrics as the guardrail.

PR #12066, opened 2025-01-17
feat(nimbus): ingest DAU and its new statistic

mikewilli:
Because

- DAU has replaced DOU as the metric we want to track
- DAU uses a new statistic that the jetstream client doesn't handle

This commit

- adds DAU and its new statistic to the ingestion

This is step 1 of 2 for #12065 
Step 2 will be to update the UI.

/cc @danielkberry


jaredlockhart (commented):
No test changes? 🤔

jaredlockhart (approved):
Very nice 👍

PR #12067, opened 2025-01-17
feat(nimbus): change DOU to DAU in Results UI

### LINKED ISSUE #12065: Use DAU metric as guardrail for Firefox apps

mikewilli:
Because

- with #12066 we will be ingesting DAU properly alongside DOU
- we want to switch from DOU to DAU as the default guardrail in the UI
- we still want to show DOU for older experiments that do not have DAU
- we no longer care about supporting QCDOU as a highlighted metric for desktop experiments

This commit

- adds DAU and uses it instead of DOU for highlighted tables
- but falls back to using DOU if DAU doesn't exist
- removes special-case code for QCDOU in desktop that was added in #9520 

Fixes #12065

Note: per @danielkberry, we don't want to display DOU as a secondary metric at all, so we don't. We show either DAU (preferred) or DOU (fallback) in the highlighted metrics tables, but if both are available we only show DAU.


experimenter/experimenter/nimbus-ui/src/lib/visualization/utils.tsx
```diff
@@ -97,3 +99,48 @@ export const getExtremeBounds = (
...
   return extreme;
 };
+
+/**
+ * Determine whether UI should fallback to using DOU instead of DAU.
+ *
+ * @param results BranchDescription to search for DOU and DAU
+ * @returns true if `results` contains DOU but not DAU results, else false
+ */
+export const shouldUseDou = (results: BranchDescription | undefined) => {
```
jaredlockhart, line 109: Ya nice to pull this out into a clean little check.

experimenter/experimenter/nimbus-ui/src/components/PageResults/index.tsx
```diff
@@ -463,10 +462,6 @@ const PageResults: React.FunctionComponent<RouteComponentProps> = () => {
                     className="rounded-bottom mb-3 border-top-0"
                     analysisBasis={selectedAnalysisBasis}
                     segment={selectedSegment}
-                    isDesktop={
-                      experiment.application ===
-                      NimbusExperimentApplicationEnum.DESKTOP
-                    }
```
jaredlockhart, line 469: Nice that we can get rid of this.  So now we can abstract this way into just whether DAU exists or not, regardless of app, and that covers this case eh?

jaredlockhart (approved):
Okay very nice, very good clear test coverage.  Thanks @mikewilli 🙏 🎉

experimenter/experimenter/nimbus-ui/src/components/PageResults/index.tsx
mikewilli: Yea exactly. We had put in a special-case check for desktop because there was an improved metric specifically for desktop (QCDOU), but DAU supercedes that and is available for all apps.

PR #12090, opened 2025-01-22
feat(nimbus): Approval and rejection flow

### LINKED ISSUE #12072: Experiment approval/rejection process UI
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5100)

yashikakhurana:
Because

- We want to add approval and rejection flow on the new summary page

This commit

- Add option to approve and Launch
- Add the option to reject the experiment with the message
- Shows the timeout message 

Fixes #12072 
Note: This currently does not show the rejection reason-which I will add in the separate PR

Approval flow

https://github.com/user-attachments/assets/84731825-e241-43fd-b9d0-f9dc2922c85c

Rejection flow

https://github.com/user-attachments/assets/9d201e88-cfc7-49f3-b6ba-3fd110488237

Timeout message
Screenshot 2025-01-22 at 10 27 54 AM


yashikakhurana:
Blocked on https://github.com/mozilla/experimenter/pull/11932

jaredlockhart:
Okay!  This is already looking great 🎉 🎉 🎉 

Couple things I notice from local testing:

- doesn't display rejection note
- doesn't immediately call remote settings sync

Look at what the V5 serializer does here on save:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/api/v5/serializers.py#L1044-L1055

We need to do these same things.

yashikakhurana:
> Okay! This is already looking great 🎉 🎉 🎉
> 
> Couple things I notice from local testing:
> 
>     * doesn't display rejection note
> 
>     * doesn't immediately call remote settings sync
> 
> 
> Look at what the V5 serializer does here on save:
> 
> https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/api/v5/serializers.py#L1044-L1055
> 
> We need to do these same things.

@jaredlockhart rejection message not showing up is related to changelog issue and that's why it's not included. I have added all the sync issues which was happening on the serializer to the forms, it should be called immediately now. I will investigate the changelog issue and made a point to implement this rejection message functionality with it.

experimenter/experimenter/experiments/models.py
```diff
@@ -699,6 +699,28 @@ def can_preview_to_draft(self):
...
+    def can_show_remote_settings_pending(self, reviewer):
+
+        return self.publish_status in (
+            self.PublishStatus.APPROVED,
+            self.PublishStatus.WAITING,
+        ) and self.can_review(reviewer)
+
+    @property
+    def can_show_timeout_message(self):
+        return self.changes.latest_timeout()
```
jaredlockhart, line 711: These are the right methods to do, i think I'd preface them with `should_` like the others.

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -535,6 +539,19 @@ def save(self, commit=True):
...
+            experiment.allocate_bucket_range()
+
+        if experiment.should_call_preview_task:
+            nimbus_synchronize_preview_experiments_in_kinto.apply_async(countdown=5)
+
+        if experiment.should_call_push_task:
+            collection = experiment.kinto_collection
+            nimbus_check_kinto_push_queue_by_collection.apply_async(
+                countdown=5, args=[collection]
+            )
```
jaredlockhart, line 553: Oh so yes this mirrors the way we did it in the serializer where it's a single save method handling all possible state changes.  But I think we can simplify this even more since we have specific state transition forms below, so rather than checking whether we're in some state transition and whether we should do any of these, we can just hard code them into the save methods of the appropriate forms below, and then we don't need to check which state transition we're in.

jaredlockhart (dismissed):
Okay tested locally, everything's look good, try the things I said below and then I think we can land.  Sorry for all the delay in reviewing! 🙏

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -535,8 +539,12 @@ def save(self, commit=True):
         experiment.status_next = self.status_next
         experiment.publish_status = self.publish_status
         experiment.save()
+        self.post_save_actions(experiment)
         return experiment
 
+    def post_save_actions(self, experiment):
+        pass
```
jaredlockhart, line 546: Oh yeah, the method override pattern totally works here, but also you can just do this with overriding save

```py
        experiment = super().save(*args, **kwargs)
        <some post save actions here>
        return experiment
```

jaredlockhart (changes requested):
Okay tested locally, everything is looking and working good, code generally looks good, couple things I noticed
 
* Rejection messages are not shown, is this a followup?

* No state is highlighted in the timeline during RS review, probably need to update the methods that include which states we're in

image

Moving the tasks into the individual forms is looking great but I don't think you need an extra method there.  Anyways getting close to wrapped! 🎉 🎉 🎉

jaredlockhart (approved):
Okay thank you for those last changes, this is all great, I think we're getting to a really nice clear set of patterns, thanks @yashikakhurana 🎉 🎉 🎉 🎉

PR #12098, opened 2025-01-23
fix(nimbus): respect boolean instead of string

adudenamedruby:
Because

- I used a string instead of a boolean, things are borked when we try to run the experiment.

This commit

- Uses a boolean, instead of a string, thus, unborking things.

Fixes mozilla-mobile/firefox-ios#24037

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -1886,8 +1886,8 @@ WINDOWS_10_MSIX_ONLY = NimbusTargetingConfig(
     application_choice_names=(Application.DESKTOP.name,),
 )
 
-IOS_DEFAULT_BROWSER_USER = NimbusTargetingConfig(
-    name="Default Browser FXiOS Users",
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/targeting/constants.py
```diff
@@ -1890,7 +1890,7 @@ def __post_init__(self):
     name="Default Browser FXiOS Users",
     slug="ios_default_browser_user",
     description="Users that already have FXiOS set as the default browser",
-    targeting="is_default_browser == 'true'",
+    targeting="is_default_browser == true && is_first_run",
```
brennie, line 1893: Since your targeting requires is_first_run, I would change the name to include `(first run)` or similar

brennie (approved):

adudenamedruby:
can haz landing? 🛩️

experimenter/experimenter/targeting/constants.py
jaredlockhart: Oh just highlighting that `== true` is superfluous here, but shouldn't harm anything.

PR #12101, opened 2025-01-23
bug(nimbus): handle None in results data

### LINKED ISSUE #12094: 500 on landing page
Looks like splitting the date columns into separate start and end dates has introduced a 500 with the following query parameters:

https://experimenter.services.mozilla.com/nimbus/?sort=_start_date&status=Complete

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5114)

jaredlockhart:
Becuase

* We recently added the show results url link on the landing page
* That involves checking the contents of the results data payload
* In old experiments, some keys unexpected contain null/None values
* The method to check did not handle this case and was raising 500

This commit

* Adds a test case for null values in the results data payload
* Handles the null case in the show results data method

fixes #12094


mikewilli (approved):
LGTM thanks!

PR #12163, opened 2025-02-10
feat(nimbus): Show invalid fields errors

### LINKED ISSUE #12009: Show missing fields errors on the summary page
If mandatory fields are not filled, show the error on the summary page. for example

!Screenshot 2024-12-02 at 11.17.07 AM-20241202-191713.png|width=1630,height=674,alt="Screenshot 2024-12-02 at 11.17.07 AM-20241202-191713.png"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5026)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/73783">Screenshot 2024-12-02 at 11.17.07 AM-20241202-191713.png</a>

RJAK11:
Because

- There is no warning message displayed on the summary page for pages with invalid fields
- The user has no way of knowing which fields are invalid and why

This commit

- Adds a warning message on the summary page if there are pages with invalid fields
- Includes links to the pages with invalid fields in the warning message
- Highlights the invalid fields and displays the reasons for the errors on the form when the link in the warning message is clicked

Fixes #12009

https://github.com/user-attachments/assets/36807404-aade-45d7-ae3d-150b593b2e64


experimenter/experimenter/nimbus_ui_new/static/js/show_errors.js
```diff
@@ -0,0 +1,13 @@
+document.addEventListener("DOMContentLoaded", function () {
+  const urlParams = new URLSearchParams(window.location.search);
+  if (urlParams.get("show_errors")) {
+    const formIds = ["audience-form", "metrics-form"];
+
+    formIds.forEach(function (formId) {
+      const form = document.getElementById(formId);
+      if (form) {
+        form.requestSubmit();
```
yashikakhurana, line 9: Consider checking form.checkValidity() before submitting.

experimenter/experimenter/experiments/models.py
```diff
@@ -1329,6 +1329,47 @@ def audience_overlap_warnings(self):
...
+        field_errors = self.get_invalid_fields_errors()
+
+        if not field_errors:
+            return []
+
+        error_fields = [field for field, _ in field_errors]
+        pages_with_errors = []
+
+        for page, fields in NimbusUIConstants.FIELD_PAGE_MAP.items():
+            if any(field in error_fields for field in fields):
```
yashikakhurana, line 1368: set(error_fields) might be better for performance but nothing major, just a suggestion

yashikakhurana (changes requested):
Overall looks good, just some minor changes. This is really looking good @RJAK11 great job 🎉

experimenter/experimenter/experiments/models.py
```diff
@@ -1329,6 +1329,47 @@ def audience_overlap_warnings(self):
 
         return warnings
 
+    def get_invalid_fields_errors(self, data=None):
+        from experimenter.experiments.api.v5.serializers import (
+            NimbusReviewSerializer,
+        )
+
+        temp_data = NimbusReviewSerializer(self).data
```
jaredlockhart, line 1337: `temp_data` isn't a very descriptive variable name.  I might've called this `experiment_data` or `serializer_data` or something.  In general think about what the contents of the variable are, where it comes from, how it's used, etc to think of names.

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -242,6 +277,11 @@ class Meta:
         ]
 
     def __init__(self, *args, **kwargs):
+        request = kwargs.get("request")
+        if request:
+            self.show_validation_errors = request.GET.get("show_errors", "") == "true"
+        else:
+            self.show_validation_errors = False
```
jaredlockhart, line 284: Ah we can pull this out into a mixin as well 🙏

jaredlockhart (changes requested):
Okay, in general, this is fantastic.  It looks right, it works right, all the patterns are all in the right places and correct, defining the mapping, checking which pages by which fields, wrapping the review serializer, all of it, great job 🎉 🎉 🎉 

The most important thing here is, we want to display all the warnings exactly as we're doing, BUT, we can't actually prevent saving.  We must preserve the behaviour that an incomplete/invalid form can still be saved.  So we have to decouple displaying these errors from actually marking fields as invalid and preventing saving, does that make sense?  Display the errors, still allow saving.

Let me know if you need any help 🙏  Great job!

experimenter/experimenter/experiments/models.py
brennie: any will short circuit and this is a generator expression, so it should be pretty good perf wise.

brennie: (`set(error_fields)` will actually construct the entire set and then evaluate `set.__bool__()`, which won't short circuit)

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,40 @@ def save(self, *args, **kwargs):
...
+                "excluded_experiments_branches",
+                "required_experiments_branches",
+            ]:
+                if field_name in cleaned_data:
+                    slugs = cleaned_data[field_name]
+                    pks = []
+                    for slug in slugs:
+                        experiment_slug = slug.split(":")[0]
+                        experiment = NimbusExperiment.objects.get(slug=experiment_slug)
+                        pks.append(experiment.pk)
```
brennie, line 58: This loop is gonna do a bunch of single get queries. We should optimize it to do a single query only returning the primary keys:

```python
slugs = [
    slug_and_branch.split(":")[0]
    for slug_and_branch in cleaned_data[field_name]
]
pks = list(NimbusExperiment.objects.filter(slug__in=slugs).values_list("pk"))
```

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,40 @@ def save(self, *args, **kwargs):
...
+                    cleaned_data[field_name.replace("branches", "")] = pks
```
brennie, line 59: Should this be `_branches` ?

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,40 @@ def save(self, *args, **kwargs):
...
+
+            errors = self.instance.get_invalid_fields_errors(data=cleaned_data)
+
+            if errors:
+                for field, error in errors:
+                    if field in self.fields:
+                        self.add_error(field, error)
```
brennie, line 66: No need for the `if`. If errors is empty, we will do zero loop iterations.

experimenter/experimenter/nimbus_ui_new/static/js/show_errors.js
brennie: Unnecessary, requestSubmit() does that.

experimenter/experimenter/experiments/models.py
brennie: Even better, lets do `error_fields = set(...)` above on line 1364.

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,40 @@ def save(self, *args, **kwargs):
         return experiment
 
 
+class NimbusSerializerErrorMixin:
+    def clean(self):
+        if self.show_validation_errors:
+            cleaned_data = super().clean()
```
brennie, line 42: we can rewrite this a bit to be more clear:

```
def clean(self):

    if self.show_validation_errors:
        # ...

    return cleaned_data
```

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -454,6 +494,11 @@ class Meta:
         ]
 
     def __init__(self, *args, **kwargs):
+        request = kwargs.get("request")
+        if request:
```
brennie, line 498:
```suggestion
        if request := kwargs.get("request"):
```

yashikakhurana (approved):
thank you @RJAK11 looks good to me :shipit:

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/detail.html
```diff
@@ -6,6 +6,40 @@
...
+    <div class="alert alert-danger" role="alert">
+      <p class="mb-1">
+        Before this experiment can be reviewed or launched, all required fields must be completed. Fields on the
+        <strong>
+          {% for page in experiment.get_invalid_pages %}
+            {% if page == "overview" %}
+              <a href="{{ experiment.get_update_overview_url }}?show_errors=true"
+                 rel="noopener noreferrer">Overview</a>
+            {% elif page == "branches" %}
+              <a href="{{ experiment.get_update_overview_url }}?show_errors=true"
```
jaredlockhart, line 20: Oh yeah branches doesn't exist yet, but instead of linking to overview, just leave the href empty or something so we don't accidentally forget and leave it pointing to overview 🙏

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,58 @@ def save(self, *args, **kwargs):
...
+                    slugs = [
+                        slug_and_branch.split(":")[0]
+                        for slug_and_branch in cleaned_data[field_name]
+                    ]
+                    pks = list(
+                        NimbusExperiment.objects.filter(slug__in=slugs).values_list(
+                            "pk", flat=True
+                        )
+                    )
+                    cleaned_data[field_name.replace("_branches", "")] = pks
```
jaredlockhart, line 65: What is this for?

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,58 @@ def save(self, *args, **kwargs):
...
+
+        return cleaned_data
+
+
+class ShowValidationErrorsMixin:
+    def __init__(self, *args, **kwargs):
+        if self.request:
+            self.show_validation_errors = (
+                self.request.GET.get("show_errors", "") == "true"
+            )
```
jaredlockhart, line 85: So yes technically the request is available on the form and you can query it like this, but this kind of request unpacking really should happen on the view in the `get_form_kwargs` method, and then just be a kwarg on init like

```py
    def __init__(self, *args, show_errors=False, **kwargs):
        super().__init__(*args, **kwargs)
        self.show_errors = show_errors
```

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -35,6 +36,58 @@ def save(self, *args, **kwargs):
...
 
 
+class NimbusSerializerErrorMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.validation_errors = {}
+
+    def clean(self):
+        cleaned_data = super().clean()
+        if self.show_validation_errors:
```
jaredlockhart, line 46: See here how we rely on `self.show_validation_errors`, but it's not set/defined on this mixin, but on the other one?  That suggests to me that these two mixins must always be used together, in which case they should probably just be a single mixin.

experimenter/experimenter/nimbus_ui_new/static/js/show_errors.js
jaredlockhart, line 9: Oh I see, this is a way to force it into the validation state so it shows the errors?  Is there not a way we can check this in the template and add a CSS class that puts it into the validation state rather than doing it in JS this way?

jaredlockhart (changes requested):
Okay!  Tested this locally, everything looks and works great.  Some thoughts about the code, but overall I think this is great 🎉 🎉 🎉 🎉

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -46,6 +46,26 @@ def get_form_kwargs(self):
...
+        if form and show_errors:
+            for field, error in errors:
+                if field in form.fields:
+                    if field not in validation_errors:
+                        validation_errors[field] = []
+                    validation_errors[field].append(error)
+
+        context["validation_errors"] = validation_errors
+
+        return context
```
jaredlockhart, line 67: We should put this into its own mixin and only explicitly include it on the pages that need it.  If we include it in this mixin, then there's a bunch of pages that'll be doing the review serializer work that might not need it.  Also I see your point here about checking against the form, but since we're actually looking the individual fields up by name in the templates, we don't really need to do this filtering at all.  I'd just generate the invalid field errors, dump them right into the context, the template can pull out the fields it cares about, and that'll be it.

jaredlockhart (changes requested):
Tested locally, everything works great, one little simplification and then we're good to go 🎉 🎉 🎉 🎉

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -45,6 +45,26 @@ def get_form_kwargs(self):
...
+        show_errors = self.request.GET.get("show_errors", "") == "true"
+        errors = self.get_object().get_invalid_fields_errors()
+
+        validation_errors = {}
+
+        if show_errors:
+            for field, error in errors:
+                if field not in validation_errors:
+                    validation_errors[field] = []
+                validation_errors[field].append(error)
```
jaredlockhart, line 61: Looks like we start in the `{field: [errors}}` structure coming out of the serializer, then convert to `[(field, error)]` coming out of `get_invalid_fields_errors`, I guess to make the page lookup a little easier, and then convert BACK to `{field: [errors]}` to actually show in the template.  

Seems more like we should just return the original serializer structure out of `get_invalid_fields_errors`, look up the fields with errors to map to pages from that, and then use that structure directly to pass into the template since that's already the structure we'd want?

jaredlockhart (changes requested):

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/detail.html
```diff
@@ -6,6 +6,39 @@
 {% block title %}{{ experiment.name }}{% endblock %}
 
 {% block main_content %}
+  <!-- Invalid Pages Warnings Card -->
+  {% if experiment.get_invalid_pages %}
+    <div class="alert alert-danger" role="alert">
+      <p class="mb-1">
+        Before this experiment can be reviewed or launched, all required fields must be completed. Fields on the
+        <strong>
+          {% for page in experiment.get_invalid_pages %}
```
jaredlockhart, line 15: Ah this will compute it twice, I think you want something like

```py
{% with invalid_pages=experiment.get_invalid_pages %}
  {% if invalid_pages %}
    {% for page in invalid_pages %}
      ...
    {% endfor %}
  {% endif %}
{% endwith %}

File: experimenter/experimenter/nimbus_ui_new/views.py [hunk 17]
```diff
@@ -45,6 +45,23 @@ def get_form_kwargs(self):
...
+        field_errors = self.get_object().get_invalid_fields_errors()
+
+        validation_errors = {}
+
+        if show_errors:
+            validation_errors = field_errors
+
+        context["validation_errors"] = validation_errors
+
+        return context
```
jaredlockhart, line 62: Yep this is perfect, there's also another way to write this like

```py
        show_errors = self.request.GET.get("show_errors", "") == "true"
        field_errors = self.get_object().get_invalid_fields_errors()

        validation_errors = {}
        if show_errors:
            validation_errors = field_errors

        return super().get_context_data(validation_errors=validation_errors, **kwargs)
```

Becuase `get_context_data` turns its `kwargs` into the context.  Just a little simplification.

jaredlockhart (approved):
Okay.  Fantastic.  Great job on this.  Looks great, works great, well written, very clear, great test coverage.  I added two little notes for improvements you can add if you like before landing, but this is great, thank you for getting this done 🙏 🎉 🎉 🎉

PR #12170, opened 2025-02-11
feat(nimbus): V7 schema

### LINKED ISSUE #12060: Add a schema for the v7 api
Skylight is currently using the v7 api to get NimbusExperiment data and was using the `NimbusExperiment` type from the `@mozilla/nimbus-schemas` package. However, the current schemas in the package don't apply to the v7 api so we would like to request for a v7 schema for type checking purposes.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5109)

yashikakhurana:
Because

- We don't have schema available for v7 API

This commit

- Add experimenter v7 schema api

Fixes #12060


yashikakhurana:
for some reason command `make schemas_generate` is only changing the .ts file but not generating the json file

schemas/mozilla_nimbus_schemas/experiments_v7/experiments_v7.py
```diff
@@ -0,0 +1,127 @@
+import datetime
+from enum import Enum
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic.json_schema import SkipJsonSchema
+from typing_extensions import Self
+
+
+class V7RandomizationUnit(str, Enum):
```
brennie, line 10: Why not use the RandomizationUnit from the other module?

schemas/mozilla_nimbus_schemas/experiments_v7/experiments_v7.py
```diff
@@ -0,0 +1,127 @@
...
+    """A unique, stable identifier for the user used as an input to bucket hashing."""
+
+    NORMANDY = "normandy_id"
+    NIMBUS = "nimbus_id"
+    USER_ID = "user_id"
+    GROUP_ID = "group_id"
+
+
+class V7ExperimentBucketConfig(BaseModel):
```
brennie, line 19: Likewise for this, it isn't changing.

brennie (changes requested):
I don't think we should add a whole 'nother module for this. Instead, can we define v7 api mixins and make a class that inherits from the v6 api schemas and the mixins?

yashikakhurana:
> I don't think we should add a whole new module for this. Instead, can we define v7 API mixins and make a class that inherits from the v6 API schemas and the mixins?

yeah @mikewilli and I were discussing this too, for the first approach we kept out separately

mikewilli:
>>I don't think we should add a whole new module for this. Instead, can we define v7 API mixins and make a class that inherits from the v6 API schemas and the mixins?

>yeah @mikewilli and I were discussing this too, for the first approach we kept out separately

I don't feel strongly either way, but it seems like there's a lot of duplication here so @brennie's suggestion makes sense to me.

yashikakhurana:
thank you both @mikewilli @brennie , I was thinking of keeping v7 separately but reusing the v6 code, does it seem to be the right approach?

schemas/generate_json_schema.py
```diff
@@ -15,7 +15,10 @@
 from polyfactory.factories.pydantic_factory import ModelFactory
 from pydantic import BaseModel, create_model
 
-from mozilla_nimbus_schemas import experiments, experiments_v7, jetstream
+from mozilla_nimbus_schemas import (
+    jetstream,
+)
+from mozilla_nimbus_schemas.experimenter_apis import experiments, experiments_v7
```
mikewilli, line 21: I'd suggest exposing `experiments` and `experiments_v7` at the top level so that these can be imported from the package directly without needing to know the package structure.

brennie (approved):

schemas/generate_json_schema.py
```diff
@@ -15,14 +15,17 @@
 from polyfactory.factories.pydantic_factory import ModelFactory
 from pydantic import BaseModel, create_model
 
-from mozilla_nimbus_schemas import experiments, jetstream
+from mozilla_nimbus_schemas import (
+    jetstream,
+)
```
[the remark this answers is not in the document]
jaredlockhart: I think that happens becuase of the trailing comma, if you get rid of that black should put it back on one line

schemas/generate_json_schema.py
yashikakhurana: moved at the top level

mikewilli: Oh sorry I realize this may not have been clear. What I was meaning to suggest was just to include experiment and experiments_v7 in the appropriate `__init__.py` files so that you could import them directly from the top level, while keeping the same structure. I thought the general structure you had was nice. That said, if you prefer this (or if what I'm saying is actually impossible for some reason I'm not thinking of...) then that's fine too.

mikewilli: Actually yea now that I look at it, I think with the `common` stuff only being common to Experimenter API schemas, it makes sense to go back to the structure you had. Let me know if what I'm suggesting doesn't make sense or seems to be impossible and we can discuss or ignore it, it's not that big of a deal.

yashikakhurana: oh sorry I thought you were indicating to move top, let me roll back the changes

mikewilli (dismissed):
Looking good, left some suggestions and especially want to make sure that the interface exposes the new sub-packages.

schemas/mozilla_nimbus_schemas/experimenter_apis/experiments_v7/experiments_v7.py
```diff
@@ -0,0 +1,122 @@
...
+
+from mozilla_nimbus_schemas.experimenter_apis.common import (
+    ExperimentBucketConfig,
+    ExperimentFeatureConfig,
+    ExperimentLocalizations,
+    ExperimentOutcome,
+)
+
+
+class V7BaseExperimentBranch(BaseModel):
```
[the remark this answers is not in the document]
brennie: i agree about moving v7 to the end

schemas/mozilla_nimbus_schemas/__init__.py
```diff
@@ -1,2 +1,2 @@
-from mozilla_nimbus_schemas.experiments import *
+from mozilla_nimbus_schemas.experimenter_apis.experiments import *
 from mozilla_nimbus_schemas.jetstream import *
```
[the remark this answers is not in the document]
yashikakhurana: gotcha just did the modifications

schemas/mozilla_nimbus_schemas/experimenter_apis/experiments/experiments.py
```diff
@@ -0,0 +1,269 @@
...
+    2. Jetstream via the Experimenter API
+    """
+
+    schemaVersion: str = Field(
+        description="Version of the NimbusExperiment schema this experiment refers to"
+    )
+    slug: str = Field(description="Unique identifier for the experiment")
+    id: str = Field(
+        description="Unique identifier for the experiment. This is a duplicate of slug."
+    )
```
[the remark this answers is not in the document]
yashikakhurana: oops thank you for noticing that

schemas/mozilla_nimbus_schemas/tests/experiments_v7/test_experiments_v7.py
```diff
@@ -0,0 +1,141 @@
...
+            "firefoxLabsDescription": "test-desc",
+            "firefoxLabsDescriptionLinks": None,
+            "firefoxLabsGroup": "test-group",
+        }
+    )
+    V7DesktopNimbusExperiment.model_validate(experiment_json)
+    v7_desktop_nimbus_experiment_schema_validator.validate(experiment_json)
+
+
+def _v7_desktop_nimbus_experiment(isRollout: bool) -> dict[str, Any]:
```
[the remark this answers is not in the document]
yashikakhurana: I just used the same format of the existing test cases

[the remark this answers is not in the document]
mikewilli: Gotcha, works for me!

mikewilli (approved):
Not worth blocking approval, but did you consider the V7-at-the-end naming?

Otherwise LGTM!

yashikakhurana:
> Not worth blocking approval, but did you consider the V7-at-the-end naming?
> 
> Otherwise LGTM!

done, v7 at the end now

jaredlockhart:
Thanks everyone for hopping on this, I thin it's all looking good, but I'm confused about one thing.  I see that there's a folder for v7 fixtures that includes desktop recipe examples, but v7 isn't used in the client and doesn't need to maintain any client compatibility or backwards compatibility, so I'm confused about why that's there?

yashikakhurana:
> Thanks everyone for hopping on this, I think it's all looking good, but I'm confused about one thing. I see that there's a folder for v7 fixtures that includes desktop recipe examples, but v7 isn't used in the client and doesn't need to maintain any client compatibility or backwards compatibility, so I'm confused about why that's there?

I wasn't sure about the scope of the v7 usage, just added those in case. do you think I should take it out @jaredlockhart?

jaredlockhart (changes requested):
Since V7 isn't consumed by clients
- We don't need special cases for desktop/sdk
- We don't need to test against client recipes

And let's keep naming consistent so if we're going to identify things v7 we should also identify things as v6.

freshstrangemusic:
> Since V7 isn't consumed by clients
> 
>     * We don't need special cases for desktop/sdk
> 
>     * We don't need to test against client recipes
> 
>     * Let's have a sample fixture for v7 as well, but it's actually separate from the existing client fixtures, just generate a factory locally, run it through v7, snapshot that.
> 
> 
> And let's keep naming consistent so if we're going to identify things v7 we should also identify things as v6.

I'm against renaming the v6 to have "V6" since they're the canonical representations (and also means we'd have to update our clients to look for new schemas)

I like the idea of separating the schemas into experiments schemas and API schemas (i.e., `schemas/experiments/`, `schemas/experimenter_api/v7.py` or similar).

yashikakhurana:
@jaredlockhart @freshstrangemusic I removed the client from v7, I haven't changed the old one to v6 as Beth mentioned that clients need to make changes too, let me know if its necessary I am happy to rename that too v6

schemas/mozilla_nimbus_schemas/experimenter_apis/experiments_v7/experiments_v7.py
```diff
@@ -0,0 +1,88 @@
...
+        description="First published date to Remote Settings.", default=None
+    )
+    localizations: ExperimentLocalizations | None = Field(
+        description="Per-locale localization substitutions.", default=None
+    )
+
+    model_config = ConfigDict(use_enum_values=True)
+
+
+class NimbusExperimentV7(BaseExperimentV7):
```
freshstrangemusic, line 83: Splitting tihs up into multiple classes is unnecessary. Please join these two into a `NimbusExperimentV7`

schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
```diff
@@ -0,0 +1,88 @@
...
+    """Per-locale localization substitutions.
+
+    The top level key is the locale (e.g., "en-US" or "fr"). Each entry is a mapping of
+    string IDs to their localized equivalents.
+    """
+
+    model_config = ConfigDict(title="ExperimentLocalizations")
+
+
+class _CommonBaseExperimentBranch(BaseModel):
```
freshstrangemusic, line 59: This class should not have a leading underscore. Please rename it back to `BaseExperimentBranch`

schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
```diff
@@ -0,0 +1,88 @@
...
+            "e.g., if branch A=1 and branch B=3, then branch A \
+                would get 25% of the population."
+        )
+    )
+    features: list[ExperimentFeatureConfig] = Field(
+        description="An array of feature configurations."
+    )
+
+
+class _CommonDesktopExperimentBranch(_CommonBaseExperimentBranch):
```
freshstrangemusic, line 73: This class is unnecessary. Please move this back into DesktopExperimentBranch (it only has one subclass).

schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Firefox Desktop-specific fields should be added to *this* schema. They will be
+    # inherited by the stricter DesktopAllVersionsExperimentBranch schema.
+
+    firefoxLabsTitle: str | None = Field(
+        description="The branch title shown in Firefox Labs (Fluent ID)", default=None
+    )
+
+
+class _CommonSdkExperimentBranch(_CommonBaseExperimentBranch):
```
freshstrangemusic, line 84: This class is unnecesssary. Please move this back into SdkExperimentBranch as it only has one subclass.

schemas/mozilla_nimbus_schemas/experimenter_apis/experiments_v7/experiments_v7.py
```diff
@@ -0,0 +1,88 @@
...
+    ExperimentLocalizations,
+    ExperimentOutcome,
+    _CommonBaseExperimentBranch,
+)
+
+
+class BaseExperimentV7(BaseModel):
+    """The base experiment definition for V7."""
+
+    schemaVersion: str = Field(description="Version of the NimbusExperiment schema.")
```
freshstrangemusic, line 17: We previously discussed moving all the comomn fields between the v6 and v7 api into a separate model inheriting from that. What happened?

freshstrangemusic (changes requested):

jaredlockhart (approved):
Okay yes I think this all looks good, thanks everyone for hopping in for reviews and @yashikakhurana for getting this done! 🎉 🎉 🎉 🎉

PR #12179, opened 2025-02-12
chore(nimbus): add locale, language, country to admin

### LINKED ISSUE #12150: Bengali (bn) missing from locale list
Hi Experimenter Team!

Would it be possible to add Bengali (bn) as a locale for desktop?

jaredlockhart:
Becuase

* We need to add some missing locales, languages, countries to the db
* We could do this through a data migration but it ends up conflicting with the test factories
* Let's add a create only interface for those models to the admin for special cases

This commit

* Adds a create only interface for locale, language, country in the admin

fixes #12150

### DIFF
--- a/experimenter/experimenter/base/admin.py
+++ b/experimenter/experimenter/base/admin.py
@@ -1,7 +1,24 @@
 from django import forms
 from django.contrib import admin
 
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
nice and simple 🚀

PR #12207, opened 2025-02-18
chore(project): replace black with ruff format

### LINKED ISSUE #12164: Replace black with ruff formatter
Looks like ruff now has black built in so we should be able to ditch black entirely and just use ruff for everything 🎉 

https://docs.astral.sh/ruff/formatter/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5151)

jaredlockhart:
Because

* Ruff is amazing
* Ruff now supports formatting
* We can consolidate all our linting and formatting on only Ruff

This commit

* Replaces black with ruff

fixes #12164


yashikakhurana (approved):
🎉 one less dependency, I will do the same for cirrus

b4handjr (approved):
Very nice

freshstrangemusic (approved):
🚀

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -814,11 +814,7 @@ def test_targeting_with_locales_and_countries_desktop(self):
         )
         self.assertEqual(
             experiment.targeting,
-            (
-                "(os.isMac) "
-                "&& (locale in ['en-CA', 'en-US']) "
-                "&& (region in ['CA', 'US'])"
-            ),
+            ("(os.isMac) && (locale in ['en-CA', 'en-US']) && (region in ['CA', 'US'])"),
```
freshstrangemusic, line 817: Not a fan of it reflowing strings across multiple lines :(

experimenter/typings/django_filters/utils.pyi
```diff
@@ -49,7 +49,7 @@ def get_model_field(model, field_name):  # -> None:
 
     ex::
 
-        f = get_model_field(Book, 'author__first_name')
```
freshstrangemusic, line 52: Should it be modifying these .pyi files?

jaredlockhart: I don't see why not?

experimenter/experimenter/experiments/tests/test_models.py
jaredlockhart: Yeah it's definitely more aggressive about strings than black is

PR #12231, opened 2025-02-24
feat(nimbus): Update targeting for non_sidebar_users

### LINKED ISSUE #12229: Update targeting for NON_SIDEBAR_USERS
Per discussion on slack on friday and the updated condition tracked in this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1943195 ) we want to exclude users with small screens/resolution.

```targeting=( "!('sidebar.revamp'|preferenceValue) && " "'sidebar.backupState'|preferenceValue == '{}'" && primaryResolution.width > 1366 && primaryResolution.height > 768 )```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5230)

We want to get this change merged for the 136 release experiment.

freshstrangemusic:
Because:

- We want to target users with specific screen resolutions

This commit:

- Updates the targeting to include those resolutions.

Fixes #12229

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -2342,7 +2342,9 @@ NON_SIDEBAR_USERS = NimbusTargetingConfig(
     description="Target users who have never used the new or old sidebar",
     targeting=(
         "!('sidebar.revamp'|preferenceValue) && "
-        "'sidebar.backupState'|preferenceValue == '{}'"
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
I was wondering if we should update this in place or make a new one, looking at prod doens't look like anyones used it yet so probably fine like this.  Thnx @freshstrangemusic 🙏

Issue #12216, opened 2025-02-20
Split start/end dates in table view with python performance improvements


jaredlockhart:
The first time we tried to split the start/end date columns, the performance was too slow to load.  Then we did some db optimization, which helped, but it was still too slow.  From sentry we see time being lost in Python space, so let's use the profiler to figure out what it's losing time on.  Also let's make sure to fix the intermittent test failure.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5210)

jaredlockhart:
https://github.com/mozilla/experimenter/issues/12206

Issue #12206, opened 2025-02-18
Intermittent test failure: test_sort_by_end_date


jaredlockhart:
Looks like we've introduced an intermittent test failure with the end date changes, seen here:

https://app.circleci.com/pipelines/github/mozilla/experimenter/51266/workflows/83ef9a4f-ac98-47cc-964e-e86dc8b9d5fa/jobs/285705

```py
_____________ NimbusExperimentsListViewTest.test_sort_by_end_date ______________
[gw0] linux -- Python 3.11.11 /usr/local/bin/python3.11

self = <experimenter.nimbus_ui_new.tests.test_views.NimbusExperimentsListViewTest testMethod=test_sort_by_end_date>

    def test_sort_by_end_date(self):
        experiment1 = NimbusExperimentFactory.create_with_lifecycle(
            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE,
            start_date=datetime.date(2024, 1, 1),
            end_date=datetime.date(2024, 2, 1),
        )
        experiment2 = NimbusExperimentFactory.create_with_lifecycle(
            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE,
            start_date=datetime.date(2024, 1, 2),
            end_date=datetime.date(2024, 2, 2),
        )
        experiment3 = NimbusExperimentFactory.create_with_lifecycle(
            NimbusExperimentFactory.Lifecycles.LIVE_ENROLLING,
            start_date=datetime.date(2024, 1, 2),
            proposed_duration=10,
        )
        experiment4 = NimbusExperimentFactory.create(
            status=NimbusExperiment.Status.LIVE,
        )
    
        response = self.client.get(
            reverse("nimbus-list"),
            {
                "sort": SortChoices.END_DATE_UP,
            },
        )
    
        expected_order_up = sorted(
            [experiment1, experiment2, experiment3], key=lambda exp: exp.computed_end_date
        )
    
>       self.assertEqual(
            [e.slug for e in response.context["experiments"]],
            ([exp.slug for exp in expected_order_up] + [experiment4.slug]),
        )
E       AssertionError: Lists differ: ['mon[25 chars]t', 'up-sized-client-driven-functionalities', [74 chars]uct'] != ['mon[25 chars]t', 'de-engineered-3rdgeneration-flexibility',[74 chars]uct']
E       
E       First differing element 1:
E       'up-sized-client-driven-functionalities'
E       'de-engineered-3rdgeneration-flexibility'
E       
[...]

PR #12238, opened 2025-02-25
feat(nimbus): Improve python performance for splitting start/end dates and fix intermittent test failure

RJAK11:
Because
* Previous attempts to split start/end date columns were too slow, even after DB optimization.
* Sentry logs showed time loss in Python space.

This commit
* Implements caching of computed_end_date property to improve performance.
* Fixes intermittent test failure.

Fixes https://github.com/mozilla/experimenter/issues/12216 and https://github.com/mozilla/experimenter/issues/12206


experimenter/experimenter/experiments/migrations/0279_merge_20250225_2006.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 5.1.5 on 2025-02-25 20:06
```
freshstrangemusic, line 1: This migration seems like a whoopsie

freshstrangemusic:
PS I reccomend a rebase based workflow for dealing with PRs as main moves. You end up with a cleaner local history

b4handjr:
Do you have details (numbers) on the performance increases from this?

RJAK11:
> Do you have details (numbers) on the performance increases from this?

Not sure if I did this correctly but I created 1000 experiments using the factory locally and the response time before these changes hovered around 195 ms and after these changes it went down to around 58 ms which seems close to how long sorting based on the other fields takes (around 50-55 ms)

jaredlockhart (approved):
Okay I've tested this locally, both for the full lifecycle of a single experiment, as well as for a large number of factory generated experiments.  I think it all looks and behaves correctly.  I made one small addition to the methods so that we can initially populate the cache for existing experiments.  Thanks for fleshing this out @RJAK11 🙏 🎉

PR #12243, opened 2025-02-26
feat(nimbus): Update targeting for non_sidebar_users to include addons and button usage

nikk15:
Because:

-  We want to target users who have not used the sidebar. By excluding users with sidebar add ons we can get quite close to the achieved targeting. By excluding users who haven't used the sidebar button we increase the chance of targeting non-sidebar users.

This commit:

- Updates the targeting to include users who haven't used the sidebar button or installed the listed sidebar addons.

Fixes:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1943195

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -2342,9 +2342,19 @@ NON_SIDEBAR_USERS = NimbusTargetingConfig(
     description="Target users who have never used the new or old sidebar",
     targeting=(
         "!('sidebar.revamp'|preferenceValue) && "
-        "'sidebar.backupState'|preferenceValue == '{}' &&"
-        "primaryResolution.width > 1366 &&"
-        "primaryResolution.height > 768"
+        "!('browser.engagement.sidebar-button.has-used'|preferenceValue) && "
+        "primaryResolution.width > 1366 && "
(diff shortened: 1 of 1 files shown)


freshstrangemusic (approved):

experimenter/experimenter/targeting/constants.py
```diff
@@ -2342,9 +2342,19 @@ def __post_init__(self):
     description="Target users who have never used the new or old sidebar",
     targeting=(
         "!('sidebar.revamp'|preferenceValue) && "
-        "'sidebar.backupState'|preferenceValue == '{}' &&"
+        "!('browser.engagement.sidebar-button.has-used'|preferenceValue) && "
         "primaryResolution.width > 1366 &&"
-        "primaryResolution.height > 768"
+        "primaryResolution.height > 768 &&"
```
muffinresearch, line 2347: With this implicit concatenation, presumably each line needs a space before the closing quote on each line except the last one to avoid a result like `&&addonsInfo.addons` etc?

freshstrangemusic (approved):

experimenter/experimenter/targeting/constants.py
nikk15: Ah, good catch!

PR #12253, opened 2025-02-28
chore(nimbus): change to correct monitoring url

### LINKED ISSUE #12252: Experimenter Enrollments Live Monitoring link should point to correct dashboard
Right now we link to `/216` dashboard, but the one we actually want is `/experimentation::experiment_enrollments` so that it is the one managed in code here: https://github.com/mozilla/looker-spoke-default/blob/main/experimentation/dashboards/experiment_enrollments.dashboard.lookml

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5260)

mikewilli:
Because

- the current link to `/216` is not the one managed by the lookml we maintain
- this is confusing when making changes because they don't appear to get deployed
- changes to `/216` must be made in the Looker UI which is less sustainable

This commit

- changes the URL to the correct managed dashboard

Fixes #12252


yashikakhurana (approved):
neat :shipit:

PR #12264, opened 2025-03-03
feat(nimbus): add tooltip with un-rounded result values

### LINKED ISSUE #12263: Show result values without rounding
Rounding the result values is helpful in most cases, but it would be nice to have a way to display the un-rounded values as well.

Screenshot shows an example of what the rounded values look like.
Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5263)

mikewilli:
Because

- rounded values can cause confusion when differences are small
- we have the un-rounded values in the database already

This commit

- shows the un-rounded values in a tooltip when hovering on the actual values

Fixes #12263 

/cc @danielkberry I believe we've talked about doing this, let me know what you think.

Screenshot from 2025-03-03 17-23-42

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/PageResults/ConfidenceInterval/index.tsx
@@ -6,6 +6,7 @@
 // TODO: EXP-638 remove this ^
 
 import React from "react";
+import ReactTooltip from "react-tooltip";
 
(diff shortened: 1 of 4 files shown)


danielkberry:
I'm not qualified to review the PR, but I think this is a great idea. Maybe we should limit to something like 6 decimal places in the tooltip though.

jaredlockhart (approved):
Great idea thanks @mikewilli 🙏 🎉

PR #12265, opened 2025-03-03
feat(nimbus): more helpful message when observation period is too short

### LINKED ISSUE #11901: Alternative Sidebar Results availbility message for too-short observation window
If the observation window is less than 7 days, Jetstream can't compute weekly results and so none will be available for Experimenter. However, Experimenter currently just sees this as "no results yet" and provides the typical message about when results are expected. In this case, we should change the message to something like:
>Experiment could not be analyzed: observation window too short.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5033)

mikewilli:
Because

- we display a message indicating that results are expected even when we know the observation period is too short for results to be computed

This commit

- displays a more helpful message when the observation period is too short
- links to the docs on observation window https://experimenter.info/configuring/#the-observation-window

Fixes #11901 

image

### DIFF
--- a/experimenter/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
+++ b/experimenter/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
@@ -166,6 +166,20 @@ export const AppLayoutSidebarLaunched = ({
     </>
   );
 
+  // compute the observation period in days (or default to a value > 7)
+  const observationPeriodDays =
+    experiment.enrollmentEndDate && experiment.computedEndDate
+      ? Math.round(
+          (new Date(experiment.computedEndDate).getTime() -
+            new Date(experiment.enrollmentEndDate).getTime()) /
+            (1000 * 60 * 60 * 24),
+        )
+      : 8;
+
+  const observationTooShort =
+    status.complete &&
+    (observationPeriodDays < 7 || !!!experiment.isEnrollmentPaused);
+
   return (
     <Container fluid className="h-100vh" data-testid={testid}>
       <Row className="h-md-100">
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus-ui/src/components/AppLayoutSidebarLaunched/index.tsx
```diff
@@ -233,6 +247,14 @@ export const AppLayoutSidebarLaunched = ({
                     </>
                   ) : analysis?.metadata?.external_config?.skip ? (
                     "Experiment analysis was skipped"
+                  ) : observationTooShort ? (
+                    <>
+                      Experiment could not be analyzed:{" "}
+                      <LinkExternal href="https://experimenter.info/configuring/#the-observation-window">
+                        observation period
+                      </LinkExternal>
+                      {" "}too short.
```
yashikakhurana, line 256: small nit `observation period is too short`

mikewilli: Sure, I was trying to make it concise and direct, but I can add "is" np.

mikewilli: OK, I went to change it and realized I could make it more helpful and state the minimum observation period as well. What do you think?

>Experiment could not be analyzed: observation period must last at least 1 week.

jaredlockhart (approved):
Good idea, thanks @mikewilli 🙏  We'll have to replicate this in the new ui as well, can you please file a followup ticket to do that so we don't lose it 🙏

yashikakhurana (approved):
perfect, thanks @mikewilli

mikewilli:
@jaredlockhart issue created https://github.com/mozilla/experimenter/issues/12269

PR #12280, opened 2025-03-05
bug(project): find bash via env in Makefile

### LINKED ISSUE #12278: /bin/bash doesn't exist by default on all systems
Our Makefile uses /bin/bash instead of /bin/sh, but /bin/bash is not guaranteed to exist on all systems at that location.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5269)

freshstrangemusic:
Because:

- bash isn't at /bin/bash on all systems (NixOS, some BSDs)

This commit:

- uses env (which is a part of POSIX) to invoke bash

Fixes #12278

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SHELL = /bin/bash
+SHELL = /usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


b4handjr (approved):
Beth with the Unix knowledge

PR #12290, opened 2025-03-06
chore(schemas): Release v3000

### LINKED ISSUE #12289: Switch from calver to semver in schemas package
dependabot doesnt understand calver and isnt opening prs when we update schemas

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5278)

freshstrangemusic:
Because:

- We have been using calender versioning for schemas
- Dependabot and other tooling does not understand calendar versioning

This commit:

- Updates the schemas package to v3000

Fixes #12289


jaredlockhart:
Wait actually I'm confused, calver follows 'x.y.z', why wouldn't dependabot support it?  It's updated it in the past:

https://github.com/mozilla/experimenter/pull/11914
https://github.com/mozilla/experimenter/pull/11735

mikewilli (changes requested):
Should update the README section that talks about CalVer too: https://github.com/mozilla/experimenter/blob/main/schemas/README.md?plain=1#L58-L59

mikewilli:
I think the problem with dependabot is that it assumes the version scheme is semver and so ignores non-minor (or however dependabot is configured) updates. Though we/other consumers could make an exception in the dependabot config for this package to include major versions (not just minor/patch).

jaredlockhart:
Oh yes we also have patch updates disabled for all packages in dependabot otherwise it would be too noisy all the time, but even if we switched schemas to semver that would still apply equally.  I don't think switching which versioning scheme we're using will make a difference here?

mikewilli:
It's less patch vs minor (since we already have minor enabled) and more minor vs major. As configured now, we'd need to manually update whenever a new year's version is published because 2025.1.1 is a major update from 2024.12.1, at least for semver and dependabot.

jaredlockhart:
No we have patch disabled, so only major and minor are caught.  But regardless, dependabot can detect and automatically bump changes regardless of whether it's semver or calvar, and may not becuase it's capped to 5 a week, in which case we'd have to bump manually, which is also the same flow regardless of whether it's calver or semver, so I don't see what changing from calver to semver solves here?

mikewilli:
Looks like we ignore both major and patch (https://github.com/mozilla/experimenter/blob/main/.github/dependabot.yml), which I think is likely pretty normal for most dependabot configs.

jaredlockhart:
Oh you're right, but I think we can explicitly call out specific packages with different rules so we could set it always update a specific package: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-disabling-version-updates-for-some-dependencies

jaredlockhart (approved):
Bye bye calver 👋

mikewilli (approved):
:shipit: :rocket:

PR #12311, opened 2025-03-12
SNG-2084 - Segment Ad Blocker Traffic

mandysGit:
Because
- We need to be able to segment the users who have an ad blocker installed from users who do not have an ad blocker installed

This commit
- Adds a target for users who have installed an ad blocker

Fixes [SNG-2084](https://mozilla-hub.atlassian.net/browse/SNG-2084)


jaredlockhart (approved):
Looks good thanks @mandysGit 🙏

PR #12350, opened 2025-03-19
feat(schemas): Add documentationLinks to v7 experiment schema

### LINKED ISSUE #12349: Add documentationLinks from the v7 api into the v7 schema
[documentationLinks](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/api/v7/serializers.py#L76-L78) are exposed in the v7 api and so I would like to request that they are added to the [v7 schema](https://github.com/mozilla/experimenter/blob/main/schemas/mozilla_nimbus_schemas/experimenter_apis/experiments_v7/experiments_v7.py), which is being used by Skylight.

freshstrangemusic:
Because:

- The v7 Experiment contains a documentationLinks field that does not appear in the schema; and
- The localizations field is included in all Experiment API responses

This commit:

- Adds the documentationLinks field to the v7 API; and
- Moves the localizations field into the common schema.

Fixes #12349

### DIFF
--- a/schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
+++ b/schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
@@ -1,4 +1,3 @@
-# experimenter_apis/common.py
 import datetime
 from enum import Enum
 from typing import Any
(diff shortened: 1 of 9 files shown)


schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
```diff
@@ -1,4 +1,3 @@
-# experimenter_apis/common.py
```
mikewilli, line 1: ??? Weird that this was there

mikewilli (approved):
LGTM

schemas/mozilla_nimbus_schemas/experimenter_apis/common.py
freshstrangemusic: Yeah i never noticed it before

PR #12356, opened 2025-03-20
added the 10 min to 2 day old profile targeting

raymak:
This targeting can be used for in-product surveys that want to talk to new profiles but also not bother them in the first few minutes. 

The targeting will be QA'ed along with the experimenter recipe that uses it.


jaredlockhart (approved):
Interesting, looks good 👍

PR #12359, opened 2025-03-21
feat(nimbus): Add archive/clone functionality to new UI

### LINKED ISSUE #12247: Add actions in sidebar-clone and archieve
!Screenshot 2025-01-13 at 11.08.21 AM-20250113-190828.png|width=278,height=598,alt="Screenshot 2025-01-13 at 11.08.21 AM-20250113-190828.png"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5094)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/75880">Screenshot 2025-01-13 at 11.08.21 AM-20250113-190828.png</a>

RJAK11:
Because  
* Users need the ability to archive and clone experiments through the new UI

This commit  
* adds functionality to archive experiments  
* adds functionality to clone experiments through sidebar
* implements the functionality for the promote to rollout button 

Fixes #12247

Archiving demo:

https://github.com/user-attachments/assets/123e2179-2f17-41a9-9c18-49be47aacd0e

Clone and promote to rollout demo:

https://github.com/user-attachments/assets/4b164a07-f846-4013-a095-bf0949a8ea8c

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/forms.py
+++ b/experimenter/experimenter/nimbus_ui_new/forms.py
@@ -144,8 +144,7 @@ class NimbusExperimentSidebarCloneForm(NimbusChangeLogFormMixin, forms.ModelForm
         return f"{self.request.user} cloned this experiment from {self.instance.name}"
 
     def save(self):
-        clone = self.instance.clone(self.cleaned_data["name"], self.cleaned_data["owner"])
-        return clone
+        return self.instance.clone(self.cleaned_data["name"], self.cleaned_data["owner"])
 
 
 class NimbusExperimentPromoteToRolloutForm(
@@ -172,12 +171,11 @@ class NimbusExperimentPromoteToRolloutForm(
         fields = ["owner", "name", "slug"]
 
     def save(self):
-        clone = self.instance.clone(
+        return self.instance.clone(
             self.cleaned_data["name"],
             self.cleaned_data["owner"],
             self.cleaned_data["branch_slug"],
         )
-        return clone
 
 
 class ToggleArchiveForm(NimbusChangeLogFormMixin, forms.ModelForm):
--- a/experimenter/experimenter/nimbus_ui_new/views.py
+++ b/experimenter/experimenter/nimbus_ui_new/views.py
@@ -277,10 +277,9 @@ class NimbusExperimentsCloneView(NimbusExperimentViewMixin, RequestFormMixin, Up
         kwargs["data"]["owner"] = self.request.user
         return kwargs
 
+    # TODO: https://github.com/mozilla/experimenter/issues/12432
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        context["experiment"] = self.get_object()
-        return context
+        return super().get_context_data(experiment=self.get_object(), **kwargs)
 
     def post(self, *args, **kwargs):
         response = super().post(*args, **kwargs)
@@ -302,7 +301,7 @@ class NimbusExperimentsPromoteToRolloutView(NimbusExperimentsCloneView):
     template_name = "nimbus_experiments/clone.html"
 
 
-class UpdateCloneSlugView(NimbusExperimentViewMixin, RenderResponseMixin, UpdateView):
+class UpdateCloneSlugView(UpdateView):
     template_name = "nimbus_experiments/clone_slug_field.html"
 
     def post(self, request, *args, **kwargs):
@@ -323,10 +322,10 @@ class ToggleArchiveView(
     def form_valid(self, form):
         form.save()
 
+        response = HttpResponse()
         if self.request.headers.get("HX-Request"):
-            response = HttpResponse()
             response.headers["HX-Refresh"] = "true"
-            return response
+        return response
 
 
 class OverviewUpdateView(
--- a/experimenter/experimenter/nimbus_ui_new/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui_new/tests/test_views.py
@@ -1241,6 +1241,19 @@ class TestNimbusExperimentsSidebarCloneView(AuthTestCase):
 
         self.assertEqual(response.context["experiment"], self.experiment)
 
+    def test_form_invalid_renders_with_experiment_context(self):
+        response = self.client.post(
+            reverse("nimbus-new-clone", kwargs={"slug": self.experiment.slug}),
+            {"owner": self.user, "name": "$."},
+        )
+
+        self.assertEqual(response.status_code, 200)
+
+        self.assertTrue(response.context["form"].errors)
+
+        self.assertIn("experiment", response.context)
+        self.assertEqual(response.context["experiment"], self.experiment)
+
 
 class TestNimbusExperimentPromoteToRolloutView(AuthTestCase):
     def setUp(self):
@@ -1275,6 +1288,19 @@ class TestNimbusExperimentPromoteToRolloutView(AuthTestCase):
 
         self.assertEqual(response.status_code, 200)
 
+        self.assertEqual(response.context["experiment"], self.experiment)
+
+    def test_form_invalid_renders_with_experiment_context(self):
+        response = self.client.post(
+            reverse("nimbus-new-clone", kwargs={"slug": self.experiment.slug}),
+            {"owner": self.user, "name": "$."},
+        )
+
+        self.assertEqual(response.status_code, 200)
+
+        self.assertTrue(response.context["form"].errors)
+
+        self.assertIn("experiment", response.context)
         self.assertEqual(response.context["experiment"], self.experiment)
 
 


experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -106,6 +106,109 @@ def clean(self):
         return cleaned_data
 
 
+class NimbusExperimentCloneFormMixin:
```
jaredlockhart, line 109: Yeah so technically this works but actually you probably don't even need to use a mixin here.  You can just fold this into `NimbusExperimentSidebarCloneForm` and then subclass that form for `NimbusExperimentPromoteToRolloutForm` and just override the save and changelog methods.

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -106,6 +106,109 @@ def clean(self):
...
+    def __init__(self, *args, parent_slug=None, branch_slug=None, **kwargs):
+        self.parent_slug = parent_slug
+        self.branch_slug = branch_slug
```
jaredlockhart, line 112: Is there a reason you did this this way instead of just making these form fields that are passed in through the actual form data?

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -106,6 +106,109 @@ def clean(self):
...
+            self.cleaned_data["name"], self.cleaned_data["owner"], self.branch_slug
+        )
+        return clone
+
+
+class ToggleArchiveForm(NimbusChangeLogFormMixin, forms.ModelForm):
+    owner = forms.ModelChoiceField(
+        User.objects.all(),
+        widget=forms.widgets.HiddenInput(),
+    )
```
jaredlockhart, line 195: Do we need owner to archive? 🤔

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/detail.html
```diff
@@ -393,6 +395,33 @@ <h6>
...
+            <div class="modal-header">
+              <h1 class="modal-title fs-5">Promote {{ branch.slug }} to Rollout</h1>
+              <button type="button"
+                      class="btn-close"
+                      data-bs-dismiss="modal"
+                      aria-label="Close"></button>
+            </div>
+            <div class="modal-body bg-body-tertiary">
+              <form id="promoteToRolloutForm-{{ branch.slug }}"
+                    hx-post="{% url 'nimbus-new-promote-to-rollout' experiment.slug branch.slug %}"
```
jaredlockhart, line 414: Oh I see, you put them in the url, and then unpack them and pass them straight to the form.  Yeah that totally works.  I think you could also make them actual form fields and then use [`hx-vals`](https://htmx.org/attributes/hx-vals/) to pass them along, so that way everything is just automatically passed in to the form through `data` and you don't need to explicitly unpack them and pass them in.

I wouldn't say it's strictly necessary, just might remove some special cases.  I think you could keep the experiment slug in the url so that the object gets looked up and passed as `instance`, and pass the branch slug as a form field, and that way both are looked up automatically and passed along.

Just a thought, not strictly necessary.

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -252,11 +270,71 @@ def post(self, *args, **kwargs):
         return response
 
 
+class NimbusExperimentsCloneView(NimbusExperimentViewMixin, RequestFormMixin, CreateView):
```
jaredlockhart, line 273: Oh okay, so what you could do here is instead of using `CreateView` you could use `UpdateView` which would then include the parent slug in the url and you could access the parent via `self.object`

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -252,11 +270,71 @@ def post(self, *args, **kwargs):
...
+        kwargs["data"] = kwargs["data"].copy()
+        kwargs["data"]["owner"] = self.request.user
+        kwargs["parent_slug"] = self.kwargs.get("slug")
+        kwargs["branch_slug"] = self.kwargs.get("branch")
+        return kwargs
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["experiment"] = self.get_object()
+        return context
```
jaredlockhart, line 285: Why is this necessary?  Shouldn't this happen automatically?

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -252,11 +270,71 @@ def post(self, *args, **kwargs):
...
+    template_name = "nimbus_experiments/clone.html"
+
+
+class UpdateCloneSlugView(NimbusExperimentViewMixin, RenderResponseMixin, UpdateView):
+    template_name = "nimbus_experiments/clone_slug_field.html"
+
+    def post(self, request, *args, **kwargs):
+        name = self.request.POST.get("name", "")
+        slug = slugify(name)
+        return self.render_to_response({"slug": slug})
```
jaredlockhart, line 313: Oh I didn't realize you were doing this via HTMX too!  Ha!!  Instead of a tiny snippet of JS.  That's so funny.  So in a maximally HTMX world, yes, exactly, all logic lives on the backend, any dynamic updates take place through a full request cycle like this.  Like this is how you'd do it in Phoenix LiveView.  I'd say let's keep it like this and see if this works well for people, or if the latency ends up being too slow in which case we can switch it to some JS, but yeah, this is so fun.  Really good instincts having this take place through a full HTMX cycle.

experimenter/experimenter/nimbus_ui_new/forms.py
jaredlockhart: Oohhhh, down below you set `owner` on the actual experiment to the archiving user.  Oh I don't think we should do that.  Where did you get that idea?

jaredlockhart (changes requested):
Okay!  Tested locally.  Everything worked PERFECTLY.  Like not a single issue.  

Actually the only thing I could think of is we should try to put the 'Archived' pill next to the name at the top of the page.  Then you'll need to change the hx-select/target to be like, teh whole page to capture both the change to the button on teh sidebar and the title.

I put a bunch of thoughts on the code, but overall like, everything's completely idiomatic, really excellent use of Django template facilities, Django form/view patterns, HTMX patterns, like other than some thoughts I put on a couple things, I don't think it could've been written better.  Really excellent work.

Take a look at my comments and see if you wanna move anything around, but ya great work, thnx @RJAK11 🙏 🎉

experimenter/experimenter/nimbus_ui_new/views.py
freshstrangemusic: :+1: because of `NimbusExperimentViewMixin` settings `model` and `context_object_name`, this should be unnecessary.

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -252,11 +270,71 @@ def post(self, *args, **kwargs):
...
+    form_class = ToggleArchiveForm
+    template_name = "nimbus_experiments/archive_button.html"
+
+    def form_valid(self, form):
+        form.save()
+
+        if self.request.headers.get("HX-Request"):
+            response = HttpResponse()
+            response.headers["HX-Refresh"] = "true"
+            return response
```
freshstrangemusic, line 329: This should return an HttpResponse regardless.
```suggestion
    def form_valid(self, form):
        form.save()
        
        response = HttpResponse()
        if self.request.headers.get("HX-Request"):
            response.headers["HX-Refresh"] = "true"

        return response
```

freshstrangemusic (approved):
r+ with nits

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -106,6 +106,96 @@ def clean(self):
...
+        if "name" in cleaned_data:
+            cleaned_data["slug"] = slugify(cleaned_data["name"])
+        return cleaned_data
+
+    def get_changelog_message(self):
+        return f"{self.request.user} cloned this experiment from {self.instance.name}"
+
+    def save(self):
+        clone = self.instance.clone(self.cleaned_data["name"], self.cleaned_data["owner"])
+        return clone
```
jaredlockhart, line 148:
```py
return self.instance.clone(....)
```

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -106,6 +106,96 @@ def clean(self):
...
+        widget=forms.HiddenInput(),
+        required=False,
+    )
+
+    class Meta:
+        model = NimbusExperiment
+        fields = ["owner", "name", "slug"]
+
+    def save(self):
+        clone = self.instance.clone(
```
jaredlockhart, line 175:
```py
return self.instance.clone(....)
```

experimenter/experimenter/nimbus_ui_new/views.py
jaredlockhart: Yeah I tried it too, we coudln't figure out why it's necessary but no need to block here if it works, filed as a followup

https://github.com/mozilla/experimenter/issues/12432

jaredlockhart (approved):
Okay!! Tested locally, everything runs smoothly, looks great, code is excellent, great test coverage, great use of Django patterns and HTMX patterns.  Thanks for all your work on this @RJAK11 let's ship it! 🎉 🎉 🎉 🎉 🎉 🎉

PR #12437, opened 2025-04-03
fix(nimbus): ensure analysis results branches are in experiment config

### LINKED ISSUE #12433: Results page error branch_data not found
Full error: `can't access property "branch_data", w[e] is undefined`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5370)

mikewilli:
Because

- erroneous branches can end up in analysis results data
- this is at best confusing/annoying
- this can crash the UI if the results are only in parts of the data

This commit

- cross references the analysis branches with the actual experiment config

Fixes #12433


jaredlockhart (approved):
Oh good, I was wondering what that was, thanks for hopping on that @mikewilli 🙏

PR #12494, opened 2025-04-16
feat(nimbus): add htmx branch forms

### LINKED ISSUE #10838: Branches Form
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4618)

jaredlockhart:
Because

* We need to add the branch forms to the new HTMX UI

This commit

* Adds forms for managing branches and feature values
* Adds views for editing, adding, deleting branches
* Adds a template for the new branch editing page
* Adds tests

fixes #10838

https://github.com/user-attachments/assets/6a11aaa0-b28b-4a53-b64c-3c48224dc216


experimenter/experimenter/nimbus_ui_new/static/js/index.js
```diff
@@ -70,4 +70,14 @@ $(() => {
...
+
+  let isInternalChange = false;
+  $(".selectpicker").on("changed.bs.select", function () {
+    if (isInternalChange) {
+      return;
+    }
+    isInternalChange = true;
+    this.dispatchEvent(new Event("change", { bubbles: true }));
+    isInternalChange = false;
+  });
```
freshstrangemusic, line 82: Please add a comment explaining why we're doing this because it is not immediately clear to me why this is necessary.

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/edit_branches.html
```diff
@@ -0,0 +1,190 @@
...
+                  </div>
+                </div>
+                {% if not forloop.first %}
+                  <a class="text-primary position-absolute top-0 end-0 m-2"
+                     hx-post="{% url 'nimbus-new-delete-branch' slug=experiment.slug %}"
+                     hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'
+                     hx-params="branch_id"
+                     hx-vals='{"branch_id": {{ branch_form.instance.id }} }'
+                     hx-select="#branches"
+                     hx-target="#branches">
```
freshstrangemusic, line 110: We should consider adding a templatetag that generates all these htmx attributes based on Python objects instead of having to stringify them in the HTML.

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -389,6 +393,258 @@ def get_changelog_message(self):
...
+class NimbusBranchDeleteForm(NimbusChangeLogFormMixin, forms.ModelForm):
+    branch_id = forms.ModelChoiceField(queryset=NimbusBranch.objects.all())
+
+    class Meta:
+        model = NimbusExperiment
+        fields = ["branch_id"]
+
+    def clean_branch_id(self):
+        branch_id = self.cleaned_data["branch_id"]
+        if branch_id == self.instance.reference_branch:
```
freshstrangemusic, line 634: can we compare against `reference_branch.id` because its weird that Django models implement equality against int

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -389,6 +393,258 @@ def get_changelog_message(self):
...
+
+    def save(self, *args, **kwargs):
+        experiment = super().save(*args, **kwargs)
+        if not experiment.reference_branch:
+            name = "Control"
+        else:
+            branch_names_all = [f"Treatment {chr(i)}" for i in range(65, 91)]
+            branch_names_used = experiment.branches.values_list("name", flat=True)
+            branch_names_free = sorted(set(branch_names_all) - set(branch_names_used))
+            name = branch_names_free[0]
```
freshstrangemusic, line 607: We are generating the entire list of 26 branch names each time. What about a generator instead?

```suggestion
            branch_names_all = (f"Treatment {chr(i)}" for i in range(65, 91))
            branch_names_used = experiment.branches.values_list("name", flat=True)
            name = next(
                 branch_name
                 for branch_name in branch_names_all
                 if branch_name not in branch_names_used
            )
```

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -389,6 +393,258 @@ def get_changelog_message(self):
...
+            if experiment.reference_branch:
+                branches = branches.exclude(id=experiment.reference_branch.id)
+
+            branches.delete()
+
+        saved_experiment_feature_configs = set(experiment.feature_configs.all())
+        saved_branch_feature_configs = {
+            fv.feature_config
+            for b in experiment.branches.all()
+            for fv in b.feature_values.all()
```
freshstrangemusic, line 568: Lets use `branch` and `feature_value`

freshstrangemusic, line 568: This is an n*m query. We should turn this into two explicit queries.

experimenter/experimenter/nimbus_ui_new/forms.py
jaredlockhart: Thankfully, no Django doesn't implement equality between an integer and a model instance:

```py
In [4]: experiment.id
Out[4]: 34

In [5]: 34 == experiment
Out[5]: False

In [6]: 34 == experiment.id
Out[6]: True
```


But unfortunately, the way the Django model form field works is the data it takes in from the form is a string of an integer id, but after it passes through the 'clean' phase, it hydrates it back to the model instance.  So then you have to choose whether to call the field 'branch', in which case its' not obvious in the form that it's the id, or 'branch_id' so it's clear in the form that the value it tracks in the client and takes in is the integer id, and it's just misleading to use as the actual instance.  This always bugs me, but I think generally I've seen the forms refer to the field as `field_id`.  We can call the variable 'branch' here to make it clearer that it's the model instance and not the id.

experimenter/experimenter/nimbus_ui_new/forms.py
jaredlockhart: Yeah theoretically, but in general I wouldn't use a generator unless it's either

1) Some very large dataset where precomputing it is actually costly
2) Involves IO

Here it's a fixed list of 26 strings which is so insignificantly small I don't think it makes a difference.

experimenter/experimenter/nimbus_ui_new/forms.py
jaredlockhart: It's funny, generally I think there's no execuse ever to use a single letter variable name, EXCEPT within list comprehensions becuase they're so narrowly scoped and then you can fit it on a single line, except here it's broken onto separate lines anyways so we can use the full varname anyways ;D

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/edit_branches.html
jaredlockhart: Ya that's a good idea.  I'm hoping we do a big ol pass on this after we get the basics in place and figure out ways to make the HTMX patterns more consistent and reusable, so I'd say we can find patterns to extract after we get all the basic mechanics in place.

experimenter/experimenter/nimbus_ui_new/forms.py
freshstrangemusic: Except it's not a fixed list of 26 strings, it's a list comprehension.

jaredlockhart: Yeah but the resulting list is of fixed size is my point

experimenter/experimenter/nimbus_ui_new/forms.py
freshstrangemusic: Sorry I meant its n+1 instead of n*m:

- `experiment.branches.all()` is one query
- `b.feature_values.all()`

we should do two explicit queries:

```python
bs = list(experiment.branches.all())
fvs = FeatureValue.objects.filter(experiment=experiment, branch__in=bs)
```

freshstrangemusic: even better!

freshstrangemusic (approved):

PR #12508, opened 2025-04-21
feat(nimbus): Implement initial Klaatu Client Prototype

### LINKED ISSUE #12415: Initial Klaatu Client Prototype
Let's make a prototype of a Klaatu client that can do the following

```py
run_test(experiment_slug, branch_slug, targets)  -> job id

example

run_test('my-experiment', 'control', ['latest', 'latest-beta', '137']) -> 12345



is_job_complete(job_id) -> bool

is_job_complete(12345) -> False    # still running
is_job_complete(12345) -> True # finished


fetch_results(job_id) -> text

fetch_results(12345) -> """<html>Test Runs.....</html>"""
```

https://github.com/mozilla/klaatu/actions/workflows/windows_manual.yml
https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5350)

RJAK11:
Because

* we need to be able to trigger GitHub workflows to run tests on an experiment
* we need to check for workflow completion and get the generated html reports

This commit

* adds methods to dispatch workflow, poll for run IDs, and check for completion
* adds method to download artifacts and extract them into a local path

Fixes #12415


experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,98 @@
...
+                f"Failed to trigger workflow: {response.status_code}, {response.text}"
+            )
+
+        runs_url = (
+            f"{self.base_url}/actions/workflows/{self.workflow_name}/runs"
+            f"?event=workflow_dispatch&created=>={before_dispatch_time}"
+        )
+
+        timeout = 30
+        elapsed = 0
```
RJAK11, line 47: I wasn't sure if polling here is the best idea since it might be cleaner to have a worker periodically check for workflows created after a given time and get the run ID separately. This kind of introduces blocking behaviour and makes assumptions about timing but let me know what y'all think is better!

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,98 @@
...
+        while elapsed < timeout:
+            runs_response = requests.get(runs_url, headers=self.headers)
+            if runs_response.status_code == 200:
+                workflow_runs = runs_response.json().get("workflow_runs", [])
+                for run in workflow_runs:
+                    if experiment_slug in run.get("display_title", ""):
+                        return run["id"]
+            time.sleep(3)
+            elapsed += 3
```
b4handjr, line 56: Do we know if this works when a job is queued and not started?

RJAK11: @b4handjr Based on the second response example on the right side [here](https://docs.github.com/en/rest/actions/workflow-runs?utm_source=chatgpt.com&apiVersion=2022-11-28#), I'm assuming it works since it shows the run has a status of queued but still has an id but I might be wrong

b4handjr: Ah yeah, I see it.

experimenter/experimenter/klaatu/client.py
b4handjr: I’m thinking we may not need to block while waiting for the `run_id`. Since we only need it later for checking job status and fetching artifacts, maybe we can just assume the tests are running if `run_tests()` completes without throwing an exception. Then, when we actually need the status, a worker could wake up and query the API for the `run_id`. We could put that worker on a schedule too, like once an hour.

What do you think?

RJAK11: Yup that makes sense!

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
+import io
+import json
+import zipfile
+
+import requests
+from django.conf import settings
+
+
+class KlaatuClient:
```
jaredlockhart, line 9: Let's add an enum of workflow options.

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+            "inputs": {
+                "slug": experiment_slug,
+                "branch": json.dumps([branch_slug]),
+                "firefox-version": json.dumps(targets),
+            },
+        }
+
+        response = requests.post(url, headers=self.headers, json=data)
+        if response.status_code != 204:
+            raise Exception(
```
jaredlockhart, line 34: Let's make a KlaatuException type to wrap these so that a caller knows its Klaatu throwing and not something else.

```py
class KlaatuException(Exception):
    pass


raise KlaatuException(.....)
```

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+        self.auth_token = settings.GITHUB_AUTH_TOKEN
+        self.base_url = "https://api.github.com/repos/mozilla/klaatu"
+        self.headers = {
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {self.auth_token}",
+            "Content-Type": "application/json",
+        }
+
+    def run_test(self, experiment_slug, branch_slug, targets):
+        url = f"{self.base_url}/actions/workflows/{self.workflow_name}/dispatches"
```
jaredlockhart, line 21: Ah let's use [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) here so we don't accidentally get double slashes and etc.

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+        response = requests.post(url, headers=self.headers, json=data)
+        if response.status_code != 204:
+            raise Exception(
+                f"Failed to trigger workflow: {response.status_code}, {response.text}"
+            )
+
+    def find_run_id(self, experiment_slug, dispatched_at=None):
+        created_filter = ""
+        if dispatched_at:
+            created_filter = f"&created=>={dispatched_at.isoformat()}Z"
```
jaredlockhart, line 41: Here's an example where we would use [urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+
+        url = (
+            f"{self.base_url}/actions/workflows/{self.workflow_name}/runs"
+            f"?event=workflow_dispatch{created_filter}"
```
jaredlockhart, line 45: [urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
+import io
+import json
+import zipfile
+
+import requests
+from django.conf import settings
+
+
+class KlaatuClient:
+    def __init__(self, workflow_name):
```
jaredlockhart, line 10: Add type annotations to the methods, that'll help document the expected args and responses.

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+        for run in workflow_runs:
+            if experiment_slug in run.get("display_title", ""):
+                return run["id"]
+
+    def is_job_complete(self, job_id):
+        url = f"{self.base_url}/actions/runs/{job_id}"
+        response = requests.get(url, headers=self.headers)
+        if response.status_code == 200:
+            status = response.json()["status"]
+            return status == "completed"
```
jaredlockhart, line 64: Move string constants up to the top of the file, here you could treat it as a single string constant like

```py
STATUS_COMPLETE = 'completed'
```

Or you could treat it as an enum even though it's only one value like

```py
class KlaatuStatus(Enum):
    COMPLETE = 'completed'
```

experimenter/experimenter/klaatu/tests/test_client.py
```diff
@@ -0,0 +1,180 @@
...
+
+        called_url = mock_get.call_args[0][0]
+        self.assertIn("created=>=2025-04-20T15:30:00", called_url)
+
+        self.assertEqual(run_id, "12345")
+
+    @mock.patch("requests.get")
+    def test_is_job_complete_returns_true(self, mock_get):
+        mock_get.return_value.status_code = 200
+        mock_get.return_value.json.return_value = {"status": "completed"}
```
jaredlockhart, line 88: Here we can refer to the constant

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+            if download_response.status_code != 200:
+                raise Exception(
+                    f"Failed to download artifact: {download_response.status_code}, "
+                    f"{download_response.text}"
+                )
+
+            with zipfile.ZipFile(io.BytesIO(download_response.content)) as zip_file:
+                zip_file.extractall(
+                    f"experimenter/experimenter/klaatu/reports/{job_id}/{artifact['name']}.html"
+                )
```
jaredlockhart, line 95: Instead of writing the artifacts out to disk, let's just read them into memory as strings and return them to the caller.  Then we can store them somewhere, like the database as a textblob, and render it out or add a download link or whatever.

experimenter/experimenter/klaatu/client.py
```diff
@@ -0,0 +1,95 @@
...
+        self.workflow_name = workflow_name
+        self.auth_token = settings.GITHUB_AUTH_TOKEN
+        self.base_url = "https://api.github.com/repos/mozilla/klaatu"
+        self.headers = {
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {self.auth_token}",
+            "Content-Type": "application/json",
+        }
+
+    def run_test(self, experiment_slug, branch_slug, targets):
```
jaredlockhart, line 20: Let's also add an enum for targets 🙏

experimenter/experimenter/klaatu/tests/test_client.py
```diff
@@ -0,0 +1,180 @@
...
+from unittest import mock
+
+from experimenter.klaatu.client import KlaatuClient
+
+
+class TestKlaatuClient(unittest.TestCase):
+    def setUp(self):
+        self.client = KlaatuClient(workflow_name="windows_manual.yml")
+
+    @mock.patch("requests.post")
```
jaredlockhart, line 13: Usually you want to specify the full path at the CALL site, ie inside of 

```py
experimenter.klaatu.client
```

is where we want to mock, so here we would say

```py
@mock.patch('experimenter.klaatu.client.requests')
```

experimenter/experimenter/klaatu/client.py
```diff
@@ -31,65 +88,83 @@ def run_test(self, experiment_slug, branch_slug, targets):
...
-            created_filter = f"&created=>={dispatched_at.isoformat()}Z"
+            query["created"] = f">={dispatched_at.isoformat()}Z"
 
-        url = (
-            f"{self.base_url}/actions/workflows/{self.workflow_name}/runs"
-            f"?event=workflow_dispatch{created_filter}"
-        )
+        path = KlaatuEndpoints.RUNS.format(workflow=self.workflow_name)
+        base_url = urljoin(self.base_url, path)
+        url = base_url + "?" + urlencode(query)
```
b4handjr, line 106: Could we do this as an f-string?

experimenter/experimenter/klaatu/client.py
```diff
@@ -1,24 +1,81 @@
...
+    RUN_STATUS = "actions/runs/{run_id}"
+    ARTIFACTS = "actions/runs/{run_id}/artifacts"
+
+
+class KlaatuTargets(str, Enum):
+    LATEST_DEVEDITION = "latest-devedition"
+    LATEST_NIGHTLY = "latest-nightly"
+    LATEST_BETA = "latest-beta"
+    LATEST_ESR = "latest-esr"
+    LATEST = "latest"
```
b4handjr, line 36: Could we add `RELEASE` to the `LATEST` here? 

So `LATEST_RELEASE`

b4handjr (changes requested):
Just had a few comments, this is looking good.

experimenter/experimenter/klaatu/tests/test_client.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @mock.patch("experimenter.klaatu.client.requests.post")
+    def test_run_test_raises_on_failed_post(self, mock_post):
+        mock_post.return_value.status_code = 400
+        mock_post.return_value.text = "Bad Request"
+
+        with self.assertRaises(KlaatuError) as error:
+            self.client.run_test("slug", "branch", [KlaatuTargets.LATEST_RELEASE])
+
+        self.assertIn("Failed to trigger workflow", str(error.exception))
```
jaredlockhart, line 38: Oh funny, I would've thought the assert would need to be idented in the `with` block, but I guess its variable persists outside the block.

jaredlockhart (approved):
Really clean, really well laid out and written, very clear comprehensive testing.  Can't imagine how this could be written better.  Great job @RJAK11 thank you 🙏 🎉 🎉 🎉 🎉 🎉

experimenter/experimenter/klaatu/client.py
jaredlockhart: Oh in general yes, always f string, but I was saying to @RJAK11 the other day that specifically when joining urls or file paths, it's better to use the 'urljoin' thingy cuz it'll catch edge cases with trailing/leading slahes, etc

b4handjr: Oh yeah, didn't even see this was a URL.

PR #12520, opened 2025-04-23
feat(nimbus): lag one day before reporting results fetch error

### LINKED ISSUE #12379: Failed results fetch error should have a 1 day lag before reporting
We have a generic error that shows in the Results UI when we fail to fetch results when we expect them to be available (e.g., experiment ended yesterday, `overall` results JSON is not available in GCS). This regularly results in questions in #ask-experimenter, when sometimes the only problem is that the automated fetch occurred before Jetstream finished analysis for the day. We should add some clarity to the message by adding something like the following to the current message:

>This error may occur if automated analysis is not yet complete. We fetch new results every two hours, so this error may resolve itself when automated analysis completes. If the error persists, please contact #ask-experimenter.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5338)

mikewilli:
Because

- Jetstream runs at approx 4am UTC daily
- Experimenter's results fetch occurs every 2 hours
- experiment results are expected based on dates (not times)
- results fetch expects results the day after the experiment ends, but results are never available the first few fetches due to the time between 12am and whenever jetstream finishes analysis
- it is confusing to see an error on the Results UI when you expect to see results

This commit

- waits a day before reporting any errors in fetching results to allow for jetstream to complete

Fixes #12379

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -1,6 +1,6 @@
 import json
 from collections import defaultdict
-from datetime import date, datetime
+from datetime import date, datetime, timedelta
 from itertools import chain
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Very nice 🙏 thnx @mikewilli 🎉

PR #12527, opened 2025-04-24
feat(nimbus): Update targeting for non_sidebar_users to include "side…

nikk15:
…bar.old-sidebar.has-used" pref

Because

- This pref was introduced in 137 after the initial targeting patch. Whilst it won't help with pre 137 sidebar users, it will help weed out sidebar user 137 onwards.

This commit

- Adds this pref to the targeting

Fixes [Bug 1943195](https://bugzilla.mozilla.org/show_bug.cgi?id=1943195)


jaredlockhart (approved):
Oh good idea!  Yeah I think this will be helpful for future experiments/rollouts.  Thnx @nikk15 🙏 🎉

PR #12540, opened 2025-04-28
feat(nimbus): default to group_id for desktop

### LINKED ISSUE #12537: Change default randomization unit for desktop to group_id in Experimenter
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5450)

jaredlockhart:
Because

* In anticipation of the release of multiprofile support in desktop, it is time to switch the default randomization unit for desktop to group_id for all new experiments
* We need to support both normandy_id and group_id for a period while currently live experiments still exist using normandy_id because when they pause they need to be reserialized and we need to preserve the normandy_id path
* When there are no more live experiments using normandy_id we can remove the code for it entirely

This commit

* Changes use_group_id to default to True
* Filters the cases where group_id is used to only include desktop
* Updates tests

fixes #12537


freshstrangemusic (approved):

jaredlockhart:
Filed followup for later date https://github.com/mozilla/experimenter/issues/12541

PR #12556, opened 2025-04-29
feat(nimbus): enforce same feature variables in all branches

### LINKED ISSUE #12414: Validate that every branch feature value contains the same set of keys
We have recently seen many examples where each branch's feature value specifies disjoint sets of keys such as

```
Branch1:
{'key1': true}

Branch2:
{'key2: false}
```

The current behaviour is for absent keys to fall back either to default values (in the SDK) or to undefined (in desktop).  However, when keys set prefs on desktop, this has the effect that a pref can be set in some branches and nother others, which can result in asymmetric unenrollment when prefs are touched by other parts of the browser.

To prevent cases like this, let's add an additional validation step that requires that every branch specify the same set of keys.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5349)

jaredlockhart:
Because

* We previously allowed branches to set different variables for a given feature
* In the pref feature case this means different branches can unenroll if a pref is touched causing asymmetric branch sizes which can invalidate an analysis
* To prevent this case we can enforce that all branches set the same variables for a given feature

This commit

* Adds validation to the review serializer to enforce that the same variables are set for a feature in all branches
* Degrades from error to warning if 'warn only' is selected
* Skips this check altogether for features with remote schemas
* Adds an additional field to mark whether a feature uses remote schemas
* Updates tests

fixes #12414
Screenshot 2025-04-29 at 11-43-24 asdf – Branches Experimenter


experimenter/experimenter/experiments/models.py
```diff
@@ -2116,6 +2116,7 @@ class NimbusVersionedSchema(models.Model):
     # Desktop-only
     set_pref_vars = models.JSONField[dict[str, str]](null=False, default=dict)
     is_early_startup = models.BooleanField(null=False, default=False)
+    has_remote_schema = models.BooleanField(null=False, default=False)
```
freshstrangemusic, line 2119: Can we rename this something like `is_trivial_schema`, `is_autogenerated_schema` or something? It's not immediately clear what "remote" means.

jaredlockhart: Oh I tried to match the naming of the other places it was referred to:

https://github.com/mozilla/experimenter/blob/511530effcb23e03948d1ed746d16e7dd5454b45/experimenter/experimenter/features/__init__.py#L43

freshstrangemusic (approved):

PR #12590, opened 2025-05-06
feat(nimbus): add save and continue redirects to new nimbus forms

### LINKED ISSUE #11400: Setup Save and Continue buttons in new Nimbus UI Forms
After all the new Nimbus UI forms are finished, we can hook up the 'Save and Continue' buttons to navigate to each successive form.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4813)

jaredlockhart:
Because

* We want the Save and Continue buttons on all the new Nimbus UI forms to save and redirect to the next form

This commit

* Adds a mixin to handle redirecting to the next form or the summary page after the audience form
* Updates the save buttons in the templates
* Adds tests

fixes #11400

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/views.py
+++ b/experimenter/experimenter/nimbus_ui_new/views.py
@@ -343,7 +343,22 @@ class ToggleArchiveView(
         return response
 
 
+class SaveAndContinueMixin:
+    def form_valid(self, form):
+        response = super().form_valid(form)
+        if (
+            self.request.headers.get("HX-Request")
+            and self.request.POST.get("save_action") == "continue"
+        ):
+            response = HttpResponse()
+            response.headers["HX-Redirect"] = reverse(
+                self.continue_url_name, kwargs={"slug": self.object.slug}
+            )
+        return response
+
+
 class OverviewUpdateView(
+    SaveAndContinueMixin,
     NimbusExperimentViewMixin,
     RequestFormMixin,
     RenderResponseMixin,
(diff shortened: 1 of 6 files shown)


freshstrangemusic (approved):

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -343,7 +343,22 @@ def form_valid(self, form):
...
+    def form_valid(self, form):
+        response = super().form_valid(form)
+        if (
+            self.request.headers.get("HX-Request")
+            and self.request.POST.get("save_action") == "continue"
+        ):
+            response = HttpResponse()
+            response.headers["HX-Redirect"] = reverse(
+                self.continue_url_name, kwargs={"slug": self.object.slug}
+            )
```
relud, line 356: everything looks sane, I see where these values come from, but I don't understand what's happening on this line.

relud (approved):
r+wc

experimenter/experimenter/nimbus_ui_new/views.py
jaredlockhart: Ya so we're rewriting the Experimenter UI in a combination of Django templates and [HTMX](https://htmx.org/) which is a handy dandy little library that lets you add some dynamic behaviours without needing to write a ton of JS.  So in this case, we set the [HX-Redirect header](https://htmx.org/headers/hx-redirect/) and HTMX picks it up and redirects the client to the destination URL.

HTMX is a bit weird to get used to at first if you haven't used it, I can give you a more detailed tour if it would help.

relud: that clears it up, tyty

PR #12592, opened 2025-05-06
chore(project): Update makefile to remove docker container before tests.

b4handjr:
Because

- We made a change recently to capture test results for a test metric dashboard. We needed to get the logs from the container before it is removed.

This commit

- Changes when the image is removed. Instead of it being removed after the tests (which may not happen if the command is interrupted or if a test fails), we now do it before each test.

Fixes #12859


jaredlockhart (approved):
Yep tested locally works perfectly, fixes the problem, thnx @b4handjr 🙏 🎉

PR #12596, opened 2025-05-06
feat(nimbus): Advanced targeting for users who have not accepted ToU yet, have not disabled ads, and are on Mac or Win

### LINKED ISSUE #12594: Add advanced targeting to support the existing user infobar baseline experiment
- User has not disabled ads or related features
- AND user has _not_ accepted the new TOU yet
- AND user is on Mac or Windows

See [experiment brief ](https://docs.google.com/document/d/11g9PiGuust9WDfXQWBKlQblODBz5JGaU6YkmoelblWU/edit?tab=t.0)advanced targeting section for more context on requirements.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5471)

lmegviar:
Because

- We want to target users who have not accepted the Terms of Use yet, have not disabled ads, and are on Windows or Mac for an upcoming experiment

This commit

- Adds advanced targeting described above (see [experiment brief ](https://docs.google.com/document/d/11g9PiGuust9WDfXQWBKlQblODBz5JGaU6YkmoelblWU/edit?tab=t.0) advanced targeting section for more context)

Fixes #12594


jaredlockhart:
@lmegviar Yep this looks right to me, you'll have to split that long line and then run `make code_format` or we can run it for you when you're ready 👍

cpeterso:
@lmegviar : Will we need to update this advanced targeting when the ToU prefs are renamed in https://bugzilla.mozilla.org/show_bug.cgi?id=1964544? Can we make this advanced targeting future proof, checking both the old `datareporting.policy.dataSubmissionPolicyAcceptedVersion` and new `termsofuse.*` prefs?

experimenter/experimenter/targeting/constants.py
```diff
@@ -2561,6 +2574,21 @@ def __post_init__(self):
...
+TOU_NOT_ACCEPTED_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
+    name="TOU not accepted yet, ads enabled, Mac or Win",
+    slug="tou_not_accepted_ads_enabled_mac_win",
+    description=(
+        "Users who have not accepted the terms of use yet, "
+        "have not disabled ads, "
+        "and are on Mac or Windows"
+    ),
+    targeting=f"((os.isWindows || os.isMac) && !({ACCEPTED_TOU_IN_NIMBUS_EXPERIMENT} || {ACCEPTED_TOU_IN_ON_TRAIN_ROLLOUT} || {ACCEPTED_TOU_V1}) && !{ADS_DISABLED})"
+    """,
```
dmose, line 2586: Just executing `python constants.py` blows up starting with a problem around here - presumably it's the " followed by the """.  It's not entirely obvious to me what's intended here.

experimenter/experimenter/targeting/constants.py
```diff
@@ -2561,6 +2574,21 @@ def __post_init__(self):
...
 
+TOU_NOT_ACCEPTED_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
+    name="TOU not accepted yet, ads enabled, Mac or Win",
+    slug="tou_not_accepted_ads_enabled_mac_win",
+    description=(
+        "Users who have not accepted the terms of use yet, "
+        "have not disabled ads, "
+        "and are on Mac or Windows"
+    ),
+    targeting=f"((os.isWindows || os.isMac) && !({ACCEPTED_TOU_IN_NIMBUS_EXPERIMENT} || {ACCEPTED_TOU_IN_ON_TRAIN_ROLLOUT} || {ACCEPTED_TOU_V1}) && !{ADS_DISABLED})"
```
dmose, line 2585: I'd like to suggest that some (many? most?) of the long f string lines added in this patch be switched to multi-line f-strings, as it will make them easier to read, review, and edit in ways where errors become moderately obvious.  As an example:

```
    targeting=f"""
        (
            (os.isWindows || os.isMac) &&
            !(
                {ACCEPTED_TOU_IN_NIMBUS_EXPERIMENT} ||
                {ACCEPTED_TOU_IN_ON_TRAIN_ROLLOUT} ||
                {ACCEPTED_TOU_V1}
            ) &&
            !{ADS_DISABLED}
        )
        """,
```

Now in this case, I suspect if you also went one step further & pulled out the `!( `THREE ACCEPTED SUBCLAUSES `)` into their own variable, that would help even more, to the point where it should be fairly easy to spot issues in each variable and their compositions.

experimenter/experimenter/targeting/constants.py
dmose, line 2586: This """ appears to be extraneous.

experimenter/experimenter/targeting/constants.py
```diff
@@ -38,6 +38,19 @@ def __post_init__(self):
 CORE_ACTIVE_USERS_TARGETING = "'{event}'|eventCountNonZero('Days', 28, 0) >= 21"
 RECENTLY_LOGGED_IN_USERS_TARGETING = "'{event}'|eventCountNonZero('Weeks', 12, 0) >= 1"
 
+# The following indicate whether the user accepted the terms of use in one of our initial phases of experimentation/rollout
+ACCEPTED_TOU_IN_NIMBUS_EXPERIMENT = "'datareporting.policy.dataSubmissionPolicyAcceptedVersion'|preferenceValue == 3"
+# Accepted in 100% on-train rollout in 139 release (after May 27, 6am PT timestamp), or in earlier partial rollout indicated by presence of on-train rollout population value
+ACCEPTED_TOU_IN_ON_TRAIN_ROLLOUT = "(('browser.preonboarding.enrolledInOnTrainRollout'|preferenceValue && 'datareporting.policy.dataSubmissionPolicyAcceptedVersion'|preferenceValue == 2) && ('datareporting.policy.dataSubmissionPolicyNotifiedTime'|preferenceValue >= "1748350800000" || 'browser.preonboarding.onTrainRolloutPopulation'|preferenceValue))"
```
dmose, line 44: The current thing fails to parse because of the "174..." inside the larger quoted string.  Switching the inner `"` symbols to `"\` should fix this.

dmose (changes requested):
The splitting you've done so far is already a big improvement in readability. I have a few tweaks and a proposal to make it even easier to gain confidence in the sub-expressions and overall expression by making it easier avoid & spot errors.

experimenter/experimenter/targeting/constants.py
```diff
@@ -38,6 +38,19 @@ def __post_init__(self):
...
+ACCEPTED_TOU_V1 = "'termsofuse.acceptedVersion'|preferenceValue == 1"
+
+# The following indicate whether the user has changed prefs suggesting they prefer not to see ads or ad-like features
+NEW_TAB_AND_HOMEPAGE_NOT_DEFAULT = "(!homePageSettings.isDefault && !newtabSettings.isDefault)"
+SEARCH_SUGGESTIONS_DISABLED = "!'browser.urlbar.suggest.quicksuggest.sponsored'|preferenceValue"
+TOPSITES_DISABLED = "(!'browser.newtabpage.activity-stream.feeds.topsites'|preferenceValue || !'browser.newtabpage.activity-stream.showSponsoredTopSites'|preferenceValue)"
+STORIES_DISABLED = "(!'browser.newtabpage.activity-stream.feeds.section.topstories'|preferenceValue || !'browser.newtabpage.activity-stream.showSponsored'|preferenceValue)"
```
mikeconley, line 51: To make this clearer, I wonder if this should be `SPONSORED_TOPSITES_DISABLED` and `SPONSORED_STORIES_DISABLED`, as that's the only thing guaranteed by this being `true`. Top Sites is not guaranteed to be disabled if, for example, only `browser.newtabpage.activity-stream.showSponsoredTopSites'` is `false`.

mikeconley (commented):
Beyond the one thing I noticed about the labelling of TOPSITES_DISABLED and STORIES_DISABLED, I think this makes sense to me. Thanks!

experimenter/experimenter/targeting/constants.py
```diff
@@ -38,6 +38,70 @@ def __post_init__(self):
...
+    ||
+    !'browser.newtabpage.activity-stream.showSponsored'|preferenceValue
+)
+"""
+ADS_DISABLED = f"""
+(
+    (
+        {NEW_TAB_AND_HOMEPAGE_NOT_DEFAULT}
+        &&
+        {SEARCH_SUGGESTIONS_DISABLED}
```
dmose, line 92: Add the SPONSORED_ onto the name of the constant here, since without that it means something confusingly different?

experimenter/experimenter/targeting/constants.py
```diff
@@ -38,6 +38,70 @@ def __post_init__(self):
...
+    {ACCEPTED_TOU_IN_ON_TRAIN_ROLLOUT}
+    ||
+    {ACCEPTED_TOU_V1}
+)
+"""
+
+# The following indicate whether the user has changed prefs suggesting they prefer not to see ads or ad-like features
+NEW_TAB_AND_HOMEPAGE_NOT_DEFAULT = "(!homePageSettings.isDefault && !newtabSettings.isDefault)"
+SEARCH_SUGGESTIONS_DISABLED = "!'browser.urlbar.suggest.quicksuggest.sponsored'|preferenceValue"
+SPONSORED_TOPSITES_DISABLED = """
```
dmose, line 73: Since we're trying to match Leif's EXCLUDE THESE USERS spec in the gdoc, I think we want this to be "TOPSITES OR_SPONSORED_TOPSITES_DISABLED".

dmose (changes requested):
OK, I think we're almost there.  Once the various comments here and in Slack/the doc are addressed, I think we'll be ready to go.

experimenter/experimenter/targeting/constants.py
```diff
@@ -38,6 +38,70 @@ def __post_init__(self):
...
+    ||
+    (
+        {SPONSORED_TOPSITES_DISABLED}
+        &&
+        {SPONSORED_STORIES_DISABLED}
+        &&
+        {SEARCH_SUGGESTIONS_DISABLED}
+    )
+)
+"""
```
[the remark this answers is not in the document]
mikeconley: Yes, I think that's a good idea. I just tested it, and it doesn't appear that that state gets reflected in newtabSettings. :/ So we should check that pref too.

dmose:
OK, just looked over those; thanks for the fixes!

jaredlockhart (approved):
Alright looks good let's go! 🎉  Thanks everyone 🙏

PR #12723, opened 2025-06-03
Add 4GB of RAM or more target

cbellini:
adds targeting for devices with 4GB or more of system memory

Because

- to use in an experiment to target users meeting the minimum requirements of a new feature

This commit

- adds a targeting rule for 4GB of system memory or more

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -2318,6 +2318,7 @@ IS_4GB_RAM = NimbusTargetingConfig(
     sticky_required=False,
     is_first_run_required=False,
     application_choice_names=(Application.DESKTOP.name,),
+)
 
 IS_64BIT_WITH_8GB_RAM = NimbusTargetingConfig(
     name="64bit Firefox build running on a computer with at least 8GB of RAM",


experimenter/experimenter/targeting/constants.py
```diff
@@ -2309,6 +2309,16 @@ def __post_init__(self):
...
+IS_4GB_RAM = NimbusTargetingConfig(
+    name="Firefox build running on a computer with at least 4GB of RAM",
+    slug="is_atleast_4gb_ram",
+    description="Target computers with at least 4GB of RAM.",
+    targeting="memoryMB >= 4000",
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
+
```
b4handjr, line 2322: I think you forgot a closing parenthesis here.

b4handjr (changes requested):

experimenter/experimenter/targeting/constants.py
cbellini: Ooops - good catch 🙂

b4handjr (approved):
Looks good and the tests pass!

PR #12729, opened 2025-06-05
feat(nimbus): Rollout review cycle

### LINKED ISSUE #11726: Rollout review cycle
Rollouts have different review cycle, like they don’t have end enrollment option and as well as they allow to update the audience as needed.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4953)

yashikakhurana:
Because

- We want to update and end the rollout on the new summary page

This commit

- Adds the option to update the rollout, it will show up the tile if there are unpublished changes, and will disable the update rollout button if there are no updates to make
- Adds the option to end the rollout

Fixes #11726 

Update rollout flow

https://github.com/user-attachments/assets/94ed5fa2-21a0-4ef1-8cfa-65a3042a85e4

end rollout flow

https://github.com/user-attachments/assets/c6482254-48da-4f04-ab19-3edc029b7081

### DIFF
--- a/experimenter/experimenter/experiments/tests/test_models.py
+++ b/experimenter/experimenter/experiments/tests/test_models.py
@@ -3840,7 +3840,7 @@ class TestNimbusExperiment(TestCase):
         block = experiment.rejection_block
         self.assertIsNotNone(block)
         self.assertEqual(
-            block["action"].value,
+            block["action"],
             NimbusUIConstants.ReviewRequestMessages[expected_flow_key].value,
         )
         self.assertEqual(


jaredlockhart:
Still says experiment here:

image

experimenter/experimenter/nimbus_ui_new/constants.py
```diff
@@ -61,11 +64,12 @@ class NimbusUIConstants:
...
-        "END_EXPERIMENT": "end this experiment",
-        "END_ENROLLMENT": "end enrollment for this experiment",
-        "LAUNCH_EXPERIMENT": "launch this experiment",
-    }
+    class ReviewRequestMessages(Enum):
+        END_EXPERIMENT = "end this experiment"
+        END_ENROLLMENT = "end enrollment for this experiment"
+        LAUNCH_EXPERIMENT = "launch this experiment"
+        UPDATE_ROLLOUT = "update this rollout"
+        END_ROLLOUT = "end this rollout"
```
jaredlockhart: 🎉 🎉 🎉 🎉 🎉

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -1109,3 +1137,102 @@ def get_changelog_message(self):
...
+        return experiment
+
+
+class LiveToUpdateRolloutForm(UpdateStatusForm):
+    status = NimbusExperiment.Status.LIVE
+    status_next = NimbusExperiment.Status.LIVE
+    publish_status = NimbusExperiment.PublishStatus.REVIEW
+
+    def get_changelog_message(self):
+        return f"{self.request.user} requested review to update Audience"
```
jaredlockhart, line 1199: Can't we just set `is_rollout_dirty=True` in here instead of checking the logic above?

yashikakhurana: so we need to check certain logic, the logic above is on the audience page and it triggers when someone changes the audience population and then it verifies all the status and checks if the percentage is updated, this allows us to not to mark is_rollout_dirty in every state, that's why kept the audience population percentage logic separate and controlling the is_rollout_dirty from there

yashikakhurana: so we will only give the option to update, if they have changed the audience, this form will only be trigger when is_rollout_dirty is already true and someone requested the review on that

yashikakhurana:
> Still says experiment here:
> image

good catch let me fix this

experimenter/experimenter/nimbus_ui_new/forms.py
jaredlockhart: Oh right okay.  How about instead of checking the specific field like 'population_percent', we just check the statuses, and we can disable all the other fields?

yashikakhurana: yes exactly, currently all the forms are editable even when its live, as a next step we will disable the fields and will only population field to be editable

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -1109,3 +1137,102 @@ def get_changelog_message(self):
                 f"{self.cleaned_data['changelog_message']}"
             )
         return f"{self.request.user} {self.cleaned_data['cancel_message']}"
+
+
+class LiveToCompleteRolloutForm(UpdateStatusForm):
```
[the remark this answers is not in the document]
jaredlockhart: Pffft, no there isn't copilot ;P

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -861,14 +861,42 @@ def save_experiments_branches(self, field_name, model):
...
+            return
+
+        population_changed = (
+            "population_percent" in self.changed_data
+            and self.cleaned_data.get("population_percent")
+            != self.initial.get("population_percent")
+        )
+
+        publish_status = (
+            self.cleaned_data.get("publish_status") or instance.publish_status
```
jaredlockhart, line 875: Oh my point was more about like, not hardcoding specific fields here but just looking at the statuses to know if we're dirty, and then using the 'disabled' flag on the fields to determine which ones are editable, becuase that's going to change over time.  But we can just leave it like this and change it later.

jaredlockhart (approved):
Okay anyways this is all great, I'm so happy we're getting cleaner with these patterns, it makes it so much easier to make changes like this.  Thanks @yashikakhurana 🙏 🎉

yashikakhurana:
@jaredlockhart  fixed Screenshot 2025-06-05 at 3 58 36 PM

PR #12733, opened 2025-06-05
Add rule for Sponsored Suggestion Enabled and 4GB minimum of RAM

cbellini:
Because

- Required for experiment targeting

This commit

- Adds a rule for targeting the with Sponsored Firefox Suggestions enabled and whose device has a 4GB minimum of RAM

Fixes #github_issue_number


jaredlockhart (approved):
LGTM! 👍

PR #12767, opened 2025-06-11
Fix case issue and add minimum profile age to TOU targeting expression

lmegviar:
Because

- When an in-progress patch to update the TOU pref names lands, we expect the `termsofuse` prefix to be all lower case. Without this change, we wouldn't be respecting the bypass notification pref.
- Additionally, QA noted that the infobar could trigger for new users. As we only want to target existing users.

This commit

- Change `termsOfUse.bypassNotification` to `termsofuse.bypassNotification`
- Updates the targeting expression for this experiment to include a minimum profile age of 28 days.

Fixes ##12766

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -83,7 +83,7 @@ TOU_NOTIFICATION_BYPASS_ENABLED = """
 (
     'datareporting.policy.dataSubmissionPolicyBypassNotification'|preferenceValue
     ||
-    'termsOfUse.bypassNotification'|preferenceValue
+    'termsofuse.bypassNotification'|preferenceValue
 )
 """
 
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/targeting/constants.py
```diff
@@ -83,7 +83,7 @@ def __post_init__(self):
 (
     'datareporting.policy.dataSubmissionPolicyBypassNotification'|preferenceValue
     ||
-    'termsOfUse.bypassNotification'|preferenceValue
+    'termofuse.bypassNotification'|preferenceValue
```
dmose:
```suggestion
    'termsofuse.bypassNotification'|preferenceValue
```

experimenter/experimenter/targeting/constants.py
```diff
@@ -2681,7 +2681,7 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
-TOU_NOT_ACCEPTED_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
+TOU_NOT_ACCEPTED_EXISTING_USER_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
     name="TOU not accepted yet, ads enabled, Mac or Win",
     slug="tou_not_accepted_ads_enabled_mac_win",
     description=(
```
dmose, line 2687: I like that you've updated the constant name so it still reflects what the constant does.  I'd suggest also updating the `name`, `slug`, and `description`. 

If we do this, we'll also need to make sure that the change of the `name` and `slug` don't confuse experimenter.

If we don't do it, it seems like it's a footgun that has some (somewhat small) chance of confusing us in (eg) debug messages, review, etc for this experiment. And a larger chance if for some reason we need to change or reuse the targeting in some other experiment along this whole path.

Either way, I think we would do well to have another round of QA.

dmose (changes requested):

experimenter/experimenter/targeting/constants.py
lmegviar: Cosmin is planning on it and knows this patch is coming. I've updated the name and slug.

experimenter/experimenter/targeting/constants.py
lmegviar: Ah, thanks! Not a typo fixing the typo 🙈.

experimenter/experimenter/targeting/constants.py
lmegviar: Note this slug has not been used in any live experiments, so should be fine to change.

PR #12786, opened 2025-06-16
chore(nimbus): Update klaatu client to accept a list of branches.

### LINKED ISSUE #12784: Adjust Klaatu client to take a list of branches instead of just 1
Our klaatu client currently accepts a single branch per test run, we should change this to accept a list branches so that a test run on github includes all branches and firefox versions in one single job.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5542)

b4handjr:
Because

- We want all available branches to be tested at once when a klaatu job triggers

This commit

- Updates the klaatu client to send a list of branches instead of just 1.

Fixes #12784

### DIFF
--- a/experimenter/experimenter/klaatu/client.py
+++ b/experimenter/experimenter/klaatu/client.py
@@ -71,7 +71,7 @@ class KlaatuClient:
     def run_test(
         self,
         experiment_slug: str,
-        branch_slug: str,
+        branch_slugs: list[str],
         targets: list[Union[KlaatuTargets, str]],
     ) -> None:
         path = KlaatuEndpoints.DISPATCH.format(workflow=self.workflow_name)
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/klaatu/client.py
```diff
@@ -81,7 +81,7 @@ def run_test(
             "ref": "main",
             "inputs": {
                 "slug": experiment_slug,
-                "branch": json.dumps([branch_slug]),
+                "branch": json.dumps([branch_slugs]),
```
relud:
```suggestion
                "branch": json.dumps(branch_slugs),
```
seems like this shouldn't become double nested?

relud (approved):
r+wc

experimenter/experimenter/klaatu/client.py
b4handjr: Ah yes, good catch.

PR #12807, opened 2025-06-17
fix(nimbus): if available, show results regardless of date

### LINKED ISSUE #12806: Check for results in addition to dates when showing/hiding Results link
It is possible for an experiment that Experimenter thinks cannot have results to actually have results, due to custom configs. For example, an experiment that ends enrollment and then ends on the same day has a 0-day observation period, which means no analysis can be performed (at least as far as Experimenter knows). However, a custom config can be added later that changes the enrollment period and creates an observation period. Jetstream picks this up and computes results, but Experimenter (rightly) does not update its own dates based on the custom config and so does not show the Results link in the sidebar.

If possible, we should check for the existence of results (in addition to the current date checks) before determining that the sidebar link should be disabled.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5547)

mikewilli:
Because

- a custom config can override dates without Experimenter's knowledge
- Jetstream uses the custom config dates to compute results 

This commit

- ignores dates when determining whether the Results link should be shown

Fixes #12806

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1384,15 +1384,18 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             results_data = self.results_data["v3"]
             for window in ["overall", "weekly"]:
                 if results_data.get(window):
-                    enrollments = results_data[window].get("enrollments", {}).get("all")
-                    if enrollments is not None:
-                        return True
+                    for base in ["enrollments", "exposures"]:
+                        base_results = results_data[window].get(base, {}).get("all")
+                        if base_results is not None:
+                            return True
 
         return False
 
     @property
     def show_results_url(self):
-        return self.has_displayable_results and self.results_ready and not self.is_rollout
+        # if there are results, show them! even if the dates are wrong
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -1384,15 +1384,18 @@ def has_displayable_results(self):
...
+                            return True
 
         return False
 
     @property
     def show_results_url(self):
-        return self.has_displayable_results and self.results_ready and not self.is_rollout
+        # if there are results, show them! even if the dates are wrong
+        # (the dates may have been overridden in metric-hub)
+        return self.has_displayable_results
```
yashikakhurana: we should keep `not self.is_rollout` check

yashikakhurana (commented):
overall looks good, just one minor thing

experimenter/experimenter/experiments/models.py
mikewilli: Yea I actually had that in there still just forgot to hit save oops. I'm fairly certain this will never matter, but it doesn't hurt.

yashikakhurana (approved):
:shipit:

PR #12820, opened 2025-06-19
feat(nimbus): Disable/enable forms

### LINKED ISSUE #12781: Disbale forms when experiment is live
We allow users to change data until the experiment is live; once it is live and it is a rollout, then we only allow editing the percentage of the audience. So make all the forms disable if the experiment/rollout is live, and just allow audience page to be enabled.

When experiment/rollout is drive then only we allow all forms to be editable

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5541)

yashikakhurana:
Because

- We only want users to edit the forms if they're in a draft state, and the only exception we have is that the audience page should be allowed to edit if it's live and a rollout.

This commit

- Enable/disable forms
- Enable audience page if it's a rollout and in a live state

Note: Currently, it allows the whole audience page enabled to edit that. In a separate ticket, it will only restrict to allow only audience percentage to change.

Fixes #12781 
Screenshot 2025-06-18 at 3 06 52 PM
Screenshot 2025-06-18 at 3 06 41 PM
Screenshot 2025-06-18 at 3 06 30 PM
Screenshot 2025-06-18 at 3 06 11 PM

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1033,7 +1033,6 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         return None
 
     def sidebar_links(self, current_path):
-        # Determine editability based on status
         is_draft = self.is_draft
         is_live_rollout = self.is_rollout and self.is_enrolling
 
@@ -1059,7 +1058,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             },
             {
                 "title": "Results",
-                "link": "",  # Real URL later
+                "link": "",
                 "icon": "fa-solid fa-chart-column",
                 "active": False,
                 "disabled": True,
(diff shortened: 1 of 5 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -1032,6 +1032,67 @@ def computed_observations_days(self):
             return (self.computed_end_date - enrollment_end_date).days
         return None
 
+    def sidebar_links(self, current_path):
+        is_live_rollout = self.is_rollout and self.is_enrolling
+
+        def is_edit_enabled(section):
+            if self.is_draft:
+                return True
+            return bool(section == "Audience" and is_live_rollout)
```
jaredlockhart: Oh yeah, I was gonna say we should have this method right on the actual experiment, and since ther'es only four sections you can just hardcode four different methods, that's what we've done in the past

def can_edit_overview
def can_edit_branches
def can_edit_metrics
def can_edit_audience

and then inside each one you do like, status, type, etc kinds of checks

jaredlockhart (dismissed):
Yeah this looks great I'd just break the edit check methods out so we can modify them later more discretely 🙏

jaredlockhart (approved):
R+WC 🙏 🎉 thnx @yashikakhurana 🎉 🎉 🎉

PR #12843, opened 2025-06-24
chore(cirrus): include fml files in docker image

relud:
Because

- cirrus web needs access to fml files, and this is easier than mounting the files in the helm chart and needing to synchronize them across separate repos

This commit

- adds the experimenter feature manifests directory to /fml in the cirrus image
- adds fml files for experimenter from https://github.com/mozilla/webservices-infra/tree/2745554/cirrus/k8s/cirrus/fml/experimenter

fixes [EXP-5553](https://mozilla-hub.atlassian.net/browse/EXP-5553) along with https://github.com/mozilla/webservices-infra/pull/6211


jeddai (commented):
You'll want to define the base FML yaml and then have build steps that generate the individual environment yamls

jaredlockhart (approved):
I was confused why this included the experimenter FMLs until I read the other side removing it from the ops repo, but yep this all looks good.  We should be able to just copy all the manifests in at container build time, pass in the application id, and then load the relevant FMLs at startup.

RIght now we load the FMLs deep into the experimenter tree, but since we're gonna depend on them in both Experimenter and Cirrus, maybe we should move that up to the top of the repo so that both can more clearly pull them in.  We can do that as a followup though.

Anyways this looks good, let's land this, onward and upward, thanks @relud 🎉 🎉 🎉

PR #12849, opened 2025-06-25
feature(nimbus): Add klaatu celery tasks to trigger validation workfl…

### LINKED ISSUE #12785: Add celery tasks to call github actions workflows
We want to be able to trigger klaatu runs via tasks

1. when an experiment gets moved into preview. 
1. every hour to add test reports to experiments

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5528)

b4handjr:
Because

- We need a way to call klaatu validation workflows from experimenter via celery tasks.

This commit

- Adds the necessary tasks and tests utilizing the klaatu client.
- Adds a field on the experiment model to track the recent run id
- Updates the experiment model tests
- Added the macos workflow name to the Klaatu client

Fixes #12785


yashikakhurana:
@b4handjr can you share how we can test this locally please

b4handjr:
> @b4handjr can you share how we can test this locally please

Which part would you like to test?

yashikakhurana:
how to test its calling the task and validates that it is updating it correctly, also any work flow you want to test with this PR I can test that too

experimenter/experimenter/klaatu/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+    KlaatuTargets,
+    KlaatuWorkflows,
+)
+
+logger = get_task_logger(__name__)
+metrics = markus.get_metrics("klaatu.nimbus_tasks")
+
+
+def get_release_version() -> int:
+    versions = requests.get("https://whattrainisitnow.com/api/firefox/releases/").json()
```
yashikakhurana: can you move this url to constants file, I think it would be easy to manage

experimenter/experimenter/klaatu/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+def get_firefox_targets(experiment: NimbusExperiment) -> list[str]:
+    versions = []
+    release_version = get_release_version()
+    latest_per_major = {}
+    final_versions = []
+    max_version = None
+
+    # get max version
+    try:
+        max_version = int(float(experiment.firefox_max_version.replace("!", "0"))) + 1
```
yashikakhurana: we have a parse method https://github.com/mozilla/experimenter/blob/2e444d2fea66b6aa862cb9b6f24db90f1840e768/experimenter/experimenter/experiments/constants.py#L420 that you can use, so you wont need to do replace

experimenter/experimenter/klaatu/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # if no max version they want to test against release, so include nightly/beta
+    if "desktop" in experiment.application and experiment.firefox_max_version == "":
+        final_versions.extend(
+            [KlaatuTargets.LATEST_NIGHTLY.value, KlaatuTargets.LATEST_BETA.value]
+        )
+
+    # fill versions up to max version
+    for _version in NimbusConstants.Version:
+        _version = _version.value.replace("!", "0")
```
yashikakhurana, line 65: same you can get rid of this replace

experimenter/experimenter/klaatu/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+    experiment.save()
+
+
+@app.task
+def klaatu_check_job_complete(experiment: NimbusExperiment, application: str) -> None:
+    client = create_klaatu_client(experiment, application)
+    client.is_job_complete(job_id=experiment.klaatu_recent_run_id)
+    if client.is_job_complete(job_id=experiment.klaatu_recent_run_id):
+        experiment.klaatu_status = True
+        experiment.save()
```
yashikakhurana, line 121: Do you think having the klaatu id history would be helpful? Because currently it will replace that, if you want to maintain the history somewhere, will suggest to add a changelog so if in future we need to show somewhere status based on the ids we can track from there

b4handjr: Not really as I think most will care about the most recent run.. and github will delete the old runs. We will keep the test results within experimenter as well as the test report.

PR #12853, opened 2025-06-25
Add advance targeting to ensure FxA is enabled for Account Adoption e…

AllegroFox:
Adds advance targeting to be used in upcoming Account Adoption messaging experiments; ensuring that FxA is available and enabled.


AllegroFox:
Closing in favour of a direct push for test reasons

PR #12863, opened 2025-06-27
feat(nimbus): Update advanced targeting expression for TOU_NOT_ACCEPT…

### LINKED ISSUE #12858: Update advanced targeting to support the existing user Latin American infobar baseline experiment
We'd like to update the advanced targeting `TOU_NOT_ACCEPTED_EXISTING_USER_ADS_ENABLED_MAC_OR_WIN`, with some country specific changes to the definition of "ads disabled" to reflect defaults in Brazil and Mexico. This will remain backwards compatible with the one current use of this targeting for [a US, Canada, and India only experiment.](https://experimenter.services.mozilla.com/nimbus/tos-existing-users-infobar-baseline-test/summary)

For the users in Brazil and Mexico, they are considered to have ads disabled if:
- New Tab sponsored shortcuts (topsites) are disabled

[Experiment Brief](https://docs.google.com/document/d/1u_JhgHYFVs3tOCY2MbUIwZhL0t7pE8iG7bsE7QoHzJg/edit?tab=t.0#heading=h.fcnbbybcfpfg)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5572)

lmegviar:
…ED_ADS_ENABLED_MAC_OR_WIN to handle BR and MX specific ad defaults

Because

- We want to run an infobar baseline experiment in Mexico and Brazil

This commit

- Updates the advanced targeting `TOU_NOT_ACCEPTED_EXISTING_USER_ADS_ENABLED_MAC_OR_WIN` with changes to the definition of "ads disabled" to reflect defaults in Brazil and Mexico. This will remain backwards compatible with the one current use of this targeting for [a US, Canada, and India only experiment.](https://experimenter.services.mozilla.com/nimbus/tos-existing-users-infobar-baseline-test/summary)

Users in Brazil and Mexico  are considered to have ads disabled if New Tab sponsored shortcuts (topsites) are disabled (see [this Slack discussion](https://mozilla.slack.com/archives/C08FQ6FQ9RS/p1750883714703869) for additional context).

Fixes #12858

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -127,6 +127,9 @@ ADS_DISABLED = f"""
     {SPONSORED_SEARCH_SUGGESTIONS_DISABLED}
 )
 """
+# Most sponsored content is off by default in Brazil and Mexico, last updated
+# June 26, 2025.
+ADS_DISABLED_BR_MX_2025_06_26 = TOPSITES_OR_SPONSORED_TOPSITES_DISABLED
 
 NO_TARGETING = NimbusTargetingConfig(
     name="No Targeting",
@@ -2787,6 +2790,10 @@ TOU_NOT_ACCEPTED_EXISTING_USER_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
         "have not disabled ads, "
         "and are on Mac or Windows"
     ),
+    # For MX/BR users, respect the region-specific ADS_DISABLED_BR_MX_2025_06_26
+    # constant. Ffor all other regions, fall back to the global ADS_DISABLED.
+    # This ensures we never enroll users whose ads have been disabled regionally
+    # due to a mismatch between the two ads disabled constants.
     targeting=f"""
     (
         (
@@ -2801,7 +2808,17 @@ TOU_NOT_ACCEPTED_EXISTING_USER_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
         &&
         !{TOU_NOTIFICATION_BYPASS_ENABLED}
         &&
-        !{ADS_DISABLED}
+        (
+            (
+                (region == "MX" || region == "BR")
+                && !{ADS_DISABLED_BR_MX_2025_06_26}
+            )
+            ||
+            (
+                (region != "MX" && region != "BR")
+                && !{ADS_DISABLED}
+            )
+        )
     )
     """,
     desktop_telemetry="",
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -127,7 +127,7 @@ ADS_DISABLED = f"""
     {SPONSORED_SEARCH_SUGGESTIONS_DISABLED}
 )
 """
-# Most sponsored content is off by default in Brazil and Mexico, last updated
+# Most sponsored content is off by default in Brazil and Mexico, as of
 # June 26, 2025.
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/targeting/constants.py
```diff
@@ -2801,7 +2804,17 @@ def __post_init__(self):
...
         &&
-        !{ADS_DISABLED}
+        (
+            (
+                (region == "MX" || region == "BR")
+                && !{ADS_DISABLED_BR_MX_2025_06_26}
+            )
+            ||
+            (
+                (region != "MX" && region != "BR")
```
dmose, line 2818: Do we need the above line?  It seems like just an extra thing to manually keep in sync with line 2809...

r=dmose with or without that change

lmegviar: I'm tempted to leave it to make it super explicit, though I hear what you're saying re: keeping in sync. While I think we can expect that if `ADS_DISABLED_BR_MX_2025_06_26` is true for a user in Mexico and Brazil and we fall through to the second case, then `ADS_DISABLED` would _also_ be true, but I worry that future changes to how we determine the general ads disabled case may alter that assumption.

dmose: But why does it matter whether the second clause is true?  If the first clause is true, we're going to fire whatever the second clause evaluates to...

lmegviar: My point is we assume `ADS_DISABLED` would be true for BR/MX if `ADS_DISABLED_BR_MX_2025_06_26` is true. If `ADS_DISABLED_BR_MX_2025_06_26` is true for users in those regions, we fall through to second clause. So long as `ADS_DISABLED` is also true for them, that's fine, they won't be enrolled. But what if some iteration of `ADS_DISABLED` returns _false_ for BR/MX users while `ADS_DISABLED_BR_MX_2025_06_26` returns true? Then we might target them when we don't intend to.

dmose: Good catch; let's leave it as is.   Maybe copy-paste a version of the paragraph I'm responding to into that you just wrote into the code as a comment so our future selves (and adders of other regions) understand what's going on and why?

lmegviar: Added, thanks @dmose!

experimenter/experimenter/targeting/constants.py
```diff
@@ -127,6 +127,9 @@ def __post_init__(self):
     {SPONSORED_SEARCH_SUGGESTIONS_DISABLED}
 )
 """
+# Most sponsored content is off by default in Brazil and Mexico, last updated
```
dmose:
```suggestion
# Most sponsored content is off by default in Brazil and Mexico, as of
```

experimenter/experimenter/targeting/constants.py
```diff
@@ -2787,6 +2790,10 @@ def __post_init__(self):
         "have not disabled ads, "
         "and are on Mac or Windows"
     ),
+    # For MX/BR users, respect the region-specific ADS_DISABLED_BR_MX_2025_06_26
+    # constant. Ffor all other regions, fall back to the global ADS_DISABLED.
```
dmose:
```suggestion
    # constant. For all other regions, fall back to the global ADS_DISABLED.
```

dmose (approved):
Looks good to me; added a couple of comment tweak suggestions.  r=dmose

PR #12900, opened 2025-07-03
feat(nimbus): Enable population precentage only for live rollout

### LINKED ISSUE #12821: Allow only the audience population to edit when its a live rollout
We should only allow the population percentage to edit when its a live rollout, check is already done when validating the updated population that only it is getting changed in a correct status here- [https://github.com/mozilla/experimenter/blob/881ee3ff68aa95f58d81797901898ec77c896a02/experimenter/experimenter/nimbus_ui_new/forms.py#L864](https://github.com/mozilla/experimenter/blob/881ee3ff68aa95f58d81797901898ec77c896a02/experimenter/experimenter/nimbus_ui_new/forms.py#L864|smart-link)  we need to update the UI and disable all the others field

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5561)

yashikakhurana:
Because

- We only want to allow the population percentage to be editable on the Audience page if it's a live rollout. Currently, all fields are editable on the new Audience page.

This commit

- Only allows audience population percentage to be editable (in case its a live rollout only)

Fixes #12821 
Screenshot 2025-07-03 at 9 38 56 AM

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1038,6 +1038,10 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             return (self.computed_end_date - enrollment_end_date).days
         return None
 
+    @property
+    def is_live_rollout(self):
+        return self.is_rollout and self.is_enrolling
+
     def can_edit_overview(self):
         return self.is_draft
 
@@ -1048,7 +1052,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         return self.is_draft
 
     def can_edit_audience(self):
-        return self.is_draft or (self.is_rollout and self.is_enrolling)
+        return self.is_draft or self.is_live_rollout
 
     def sidebar_links(self, current_path):
         return [
--- a/experimenter/experimenter/nimbus_ui_new/forms.py
+++ b/experimenter/experimenter/nimbus_ui_new/forms.py
@@ -839,6 +839,11 @@ class AudienceForm(NimbusChangeLogFormMixin, forms.ModelForm):
                 "hx-target": "#first-run-fields",
             }
         )
+        # If this is a live rollout, restrict edits to only population_percent
+        if self.instance.is_live_rollout:
+            for field_name in self.fields:
+                if field_name != "population_percent":
+                    self.fields[field_name].disabled = True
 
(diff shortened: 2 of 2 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -1048,7 +1052,7 @@ def can_edit_metrics(self):
         return self.is_draft
 
     def can_edit_audience(self):
-        return self.is_draft or (self.is_rollout and self.is_enrolling)
+        return self.is_draft or self.is_live_rollout
```
jaredlockhart, line 1055: EXACTLY 👏 👏 👏 👏 👏 

It's so nice to be able clearly define states and capabilities this way 🙏 🙏 🙏 🙏

experimenter/experimenter/nimbus_ui_new/forms.py
```diff
@@ -839,6 +839,11 @@ def __init__(self, *args, **kwargs):
                 "hx-target": "#first-run-fields",
             }
         )
+        # If this is a live rollout, restrict edits to only population_percent
+        if self.instance.is_live_rollout:
+            for field_name in self.fields:
+                if field_name != "population_percent":
+                    self.fields[field_name].disabled = True
```
jaredlockhart, line 846: AHA!  I was gonna say, see if you can do it in the form like this, and you did!  Brilliant 🙏 

We should probably add a test against the form for this though 🙏

jaredlockhart (dismissed):

jaredlockhart:
Also it's annoying that the bootstrap-selects don't grey out when they're disabled.  Maybe we need to pass some additional data parameter to them? They must have a disabled state.

experimenter/experimenter/nimbus_ui_new/forms.py
yashikakhurana: added test case

yashikakhurana:
> Also it's annoying that the bootstrap-selects don't grey out when they're disabled. Maybe we need to pass some additional data parameter to them? They must have a disabled state.

@jaredlockhart they ARE kind of grade out, for example this is when it is not graded out vs graded out 
Screenshot 2025-07-03 at 12 46 57 PM
Screenshot 2025-07-03 at 9 38 56 AM

jaredlockhart (approved):
Perfect ty 🙏 🙏 🙏 🙏 🙏

PR #12917, opened 2025-07-07
chore(nimbus): add four dot releases to all versions 140+

### LINKED ISSUE #12916: Add four dot releases for every version
It seems dot releases are becoming more and more frequent and regular, so let's make sure there are four dot releases available for every future version, which should cover most of the cases that people want to target against specific ones.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5621)

jaredlockhart:
Becuase

* Dot releases are now becoming more frequent
* Let's prefill four dot releases for every version after 140 so we don't have to add them in manually

This commit

* Adds four dot releases for desktop and mobile for every version after 140

fixes #12916


yashikakhurana (approved):
:shipit:

PR #12921, opened 2025-07-07
chore(nimbus): improve audience page performance

### LINKED ISSUE #12903: HTMX Edit Audience is slow/hangs
Looks like the new edit audience page is slow/hanging, probably loading too many choices for the drop downs.  Let's see if we can profile that and get it faster.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5611)

jaredlockhart:
Because

* The edit audience page is loading very slowly
* We are doing N+1 queries to populate the require/exclude branch fields

This commit

* Prefetches branches before rendering the require/exclude fields
* Only does a single query for both fields
* Filters require/exclude options to same application

fixes #12903


yashikakhurana (approved):
oh n+1 queries, good catch, this will definitely improve the time, Thank you @jaredlockhart :tada

PR #12924, opened 2025-07-07
chore(cirrus): publish docker image to GAR

relud:
Because

- Cirrus is currently deployed using the docker hub mirror, but ArgoCD's image updater can't use that for automatic deploys, so we need to publish this image to GAR like we do for experimenter

This commit

- Uploads cirrus docker images to GAR in addition to existing docker hub image

Fixes [EXP-5623](https://mozilla-hub.atlassian.net/browse/EXP-5623)
Requires https://github.com/mozilla/webservices-infra/pull/6431

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -882,147 +882,147 @@ workflows:
 
   build:
     jobs:
-      - check_experimenter_x86_64:
-          name: Check Experimenter x86_64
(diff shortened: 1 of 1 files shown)


relud:
to verify this works, I added 01febb467 to temporarily modify the CI job to run on this branch, and after testing reverted in 5d62db1ad

PR #12938, opened 2025-07-07
chore(nimbus): further audience page performance improvements

### LINKED ISSUE #12925: HTMX Edit Audience is still slow/hangs
Following from https://github.com/mozilla/experimenter/issues/12903 we can now see the full trace in Sentry for this page and it is still hanging on setting up the required/excluded branch choices.  Instead of querying the entire object, let's only fetch the fields we need into a tuple to build the choices that way.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5626)

jaredlockhart:
Becuase

* In #12903 we made some changes to improve the load times of the new edit audience page
* However once on prod we see that it is still quite slow but now have more information about why
* We removed an N+1 query, but it is still loading unnecessary information out of the database which is taking time

This commit

* Uses values_list to query only the necessary fields
* Removes model methods and querysets entirely to build the require/exclude choices
* This should reduce unnecessary construction of model instances

fixes #12925


yashikakhurana (approved):
ahh I see, good catch @jaredlockhart 🎉

PR #12972, opened 2025-07-11
feat(experimenter): integrate with cirrus

relud:
Note: needs data review for glean data collection

Because

- Reason

This commit

- Change

Fixes #github_issue_number


experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,26 @@
+from hashlib import sha256
+
+
+class CirrusMiddleware:
```
jaredlockhart, line 28: Oh awesome! You went ahead and did it!  This looks fantastic 🎉 🎉 🎉 🎉

jaredlockhart:
This is brilliant @relud 🎉 

Couple other things:
* Should this also add a Experimenter feature manifest? 🤔 
* We can also use the local Cirrus to test/use this locally if we set it up in the compose files 🎉

relud:
>     * Should this also add a Experimenter feature manifest? 🤔

Initial experimenter feature manifests already exists! 🎉 https://github.com/mozilla/experimenter/tree/main/experimenter/experimenter/features/manifests/experimenter

>     * We can also use the local Cirrus to test/use this locally if we set it up in the compose files 🎉

will do 🫡

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    A middleware that provides experiment enrollment information via nimbus.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.cirrus_url = settings.CIRRUS_URL
+
+    def __call__(self, request):
+        request.cirrus = {}
```
jaredlockhart: I would actually put `None` here so we can clearly dilineate whether a Cirrus response was received.

experimenter/experimenter/results/templates/nimbus/index.html
```diff
@@ -12,7 +12,11 @@
   <meta name="robots" content="noindex,nofollow" />
   <meta name="viewport"
     content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes" />
-  <title>Mozilla Experimenter</title>
+  <title>Mozilla Experimenter
+    {% if request.cirrus.get("Features", {}).get("example-feature", {}).get("enabled", False) %}
```
jaredlockhart: {% if request.cirrus and ..... %}

jaredlockhart: Also you can't invoke functions like this in Django templates

All you can do in Django templates is

obj.property

So a common pattern is to do something like

```py
class CirrusFeatures
    def __init__(self, cirrus_data):
        self.cirrus_data = cirrus_data

    @property
    def is_some_feature_enabled(self):
        return self.cirrus_data.get('someFeature').get('enabled')
```

Then we can use a template context processor to inject it into teh templates like this:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/base/context_processors.py#L10-L24

Then in the templates we can access it like this 

```html
{% if cirrus.is_some_feature_enabled %}
    <h1>Enabled!</h1>
{% endif %}
```

experimenter/experimenter/glean/middleware.py
```diff
@@ -0,0 +1,24 @@
...
+            settings.BASE_DIR / "telemetry" / "metrics.yaml"
+        )
+        Glean.set_collection_enabled(settings.ENABLE_GLEAN)
+
+    def __call__(self, request):
+        self.metrics.page.view.record(
+            self.metrics.page.ViewExtra(
+                path=request.path,
+                nimbus_user_id=(request.user.id if hasattr(request, "user") else None),
+            )
```
jaredlockhart: In here is where we'll want to use the 'enrollments' key from the cirrus response as view_extra details so we can track the enrolled slugs in the telemetry like this:

https://github.com/mozilla/experimenter/blob/main/cirrus/server/cirrus/main.py#L210-L217

experimenter/experimenter/cirrus/tests/test_middleware.py
```diff
@@ -0,0 +1,39 @@
+from unittest import mock
+
+from django.test import TestCase
+from pytest_httpserver import HTTPServer
```
jaredlockhart: Oh yeah I wouldn't do this I would just mock it out

jaredlockhart:
Dont' forget to add CIRRUS_URL to the `.env.sample`

jaredlockhart:
Oh and we have to add Experimenter as an app

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/constants.py#L260-L272

So if you do that, load the Experimenter FMLs in (which should happen at startup), then you should be able to create an experiment for Experimenter locally, publish it to local RS, local cirrus should read it, then it should show in the UI if you add it to the template

https://github.com/mozilla/experimenter?tab=readme-ov-file#accessing-remote-settings-locally

jaredlockhart:
As a followup to this we'll want to pull `?use_preview=True` out of the URL and pass it into Cirrus

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,20 @@
...
+        if hasattr(request, "user") and request.user.id and self.cirrus_url:
+            try:
+                cirrus_response = requests.get(
+                    cirrus_url, json={"client_id": request.user.id, "context": {}}
+                )
+                request.cirrus = cirrus_response.json()
+            except requests.exceptions.RequestException as e:
+                pass  # FIXME: log cirrus failure
```
jaredlockhart: Probably forward to sentry but don't block completing the request

experimenter/experimenter/glean/middleware.py
relud: I ran into a problem here, enrollments is a list, and event extras don't support nested types according to the docs, should I record the first enrollment? one page view per enrollment? turn enrollments into a string and put it all in a single field? each enrollment field is a comma-separated list of values? I'm unclear on what the intent is here

relud: @yashikakhurana perhaps you can answer this?

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,55 @@
...
+
+class CirrusData:
+    def __init__(self, data: dict):
+        self.enrollments = data["Enrollments"]
+        self.features = CirrusFeatures(data["Features"])
+
+
+class CirrusMiddleware:
+    """
+    A middleware that provides experiment enrollment information via nimbus.
```
freshstrangemusic: nit: `Nimbus`

experimenter/experimenter/glean/middleware.py
yashikakhurana: I agree and I noticed that it doesn't support nesting as list or array, okay the intent is that if you are on experimenter app lets say, right now the only way is to check if you are sending the enrollment value and what values you are sending is to check using the queries, we want to make it little bit more easier, such that we can see in a network call and see what values we are sending so that it is easy to see which experiment enrolled, what id we are sending, also for testing it is easier to validate we are enrolled in a correct experiment or rollout. I am not sure what's the string limit is if we decide to do that. may be send using the loops until we reach to the end of the enrollment data

freshstrangemusic: According to https://mozilla.github.io/glean/book/reference/metrics/event.html#limits:

> * The keys in the extra_keys list must be written using printable ASCII characters, with a maximum length of 40 bytes, when encoded as UTF-8.
> * The values in the extras object have a maximum length of 500 bytes when serialized and encoded as UTF-8. Longer values are truncated, and an invalid_overflow error is recorded.

Using a comma-separated list is probably fine -- We're not going to be launching tons of experiments or rollouts against Experimenter.

experimenter/experimenter/telemetry/metrics.yaml
```diff
@@ -0,0 +1,51 @@
...
+    expires: never
+    send_in_pings:
+      - events
+    extra_keys:
+      path:
+        description: The path of the page.
+        type: string
+      nimbus_user_id:
+        description: Nimbus user ID
+        type: string
```
freshstrangemusic: I may be wrong but this makes the event *not* interaction but *sensitive* because this is considered PII.

I don't know how much that matters for internal only telemetry, though, since you need ldap to use Experimenter.

This would be a question for a data steward.

experimenter/experimenter/cirrus/tests/test_middleware.py
```diff
@@ -0,0 +1,109 @@
...
+    def test_without_user(self):
+        del self.request.user
+
+        with self.settings(CIRRUS_URL="cirrus"):
+            middleware = CirrusMiddleware(lambda request: self.response)
+
+        with mock.patch(
+            "experimenter.cirrus.middleware.requests.get"
+        ) as mock_requests_get:
+            response = middleware(self.request)
```
freshstrangemusic: Any reason not to use RequestFactory and the associated Django machinery?

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,55 @@
...
+        return self._data.get("example-feature", {}).get("emoji")
+
+
+class CirrusData:
+    def __init__(self, data: dict):
+        self.enrollments = data["Enrollments"]
+        self.features = CirrusFeatures(data["Features"])
+
+
+class CirrusMiddleware:
```
freshstrangemusic, line 28: I think this middleware can be a plain old function returning a function?

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    def is_example_feature_enabled(self):
+        return self._data.get("example-feature", {}).get("enabled")
+
+    @property
+    def example_feature_emoji(self):
+        return self._data.get("example-feature", {}).get("emoji")
+
+
+class CirrusData:
```
freshstrangemusic, line 23: Should this be a dataclass?

experimenter/experimenter/glean/middleware.py
yashikakhurana: yeah if the string length is not the issue, we can do that too

experimenter/experimenter/cirrus/tests/test_middleware.py
relud: I copied and modified openidc's tests: https://github.com/mozilla/experimenter/blob/5a52e8d904cbbff548a6ae3571120e9689218e1f/experimenter/experimenter/openidc/tests/test_middleware.py#L30

experimenter/experimenter/cirrus/middleware.py
relud: it can, but i was under the impression a class was more idiomatic for django? I've never made a middleware before, but aside from one example pointing out that a function returning a function works, all of the examples I saw were classes.

experimenter/experimenter/cirrus/tests/test_middleware.py
relud: I have no idea how to use a request factory, would it cause an issue that i'm instantiating the middleware differently for each test here?

experimenter/experimenter/glean/middleware.py
relud: for now I've implemented it as multiple events, one for the page view, one for each enrollment, @jaredlockhart what do you think?

experimenter/experimenter/cirrus/middleware.py
relud: sure

experimenter/experimenter/glean/middleware.py
relud: to be clear, I think we were under the impression when we discussed this last week that I could add nimbus enrollments as an `object` type event extra, and I have confirmed that does not work, if I try i get this error indicating event extras are limited to `string`, `boolean` or `quantity`: 

```
ValueError: /experimenter/experimenter/telemetry/metrics.yaml:
    ```
    page:
      view:
        extra_keys:
          nimbus_enrollments:
            type: object
    ```

    'object' is not one of ['string', 'boolean', 'quantity']
```

sending a separate event for each enrollment is my proposed workaround

experimenter/experimenter/cirrus/tests/test_middleware.py
freshstrangemusic: Upon further investigation, RequestFactory does not actually instantiate middleware. womp womp

experimenter/experimenter/glean/middleware.py
yashikakhurana: I think yes we can do that only for now till the glean team supports that- sending a separate event for each enrollment

experimenter/experimenter/cirrus/tests/test_middleware.py
freshstrangemusic: self.client.get(url, **kwargs) might add middleware to the request but im not sure

experimenter/experimenter/glean/middleware.py
relud: I have changed this to submit a page-view ping with enrollments as an object-type metric, so that page views and enrollments are recorded together.

experimenter/experimenter/telemetry/metrics.yaml
relud: from data-review thread in slack:

> in the past we have treated UUIDs as "technical" unless they were hashed from some user or hardware specific information. [...] I think, combined with the url_path that's also being recorded we might go ahead and put this through the escalated review process just so everything is okay'd by T&L.

relud:
> Oh and we have to add Experimenter as an app
> 
> https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/constants.py#L260-L272
> 
> So if you do that, load the Experimenter FMLs in (which should happen at startup), then you should be able to create an experiment for Experimenter locally, publish it to local RS, local cirrus should read it, then it should show in the UI if you add it to the template
> 
> https://github.com/mozilla/experimenter?tab=readme-ov-file#accessing-remote-settings-locally

confirmed that experimenter can now locally experiment on itself, adding a string prefix to the title of `localhost/nimbus/` as defined by the `emoji` variable that defaults to ❤️, and that `?nimbus_preview=true` works to preview the experiment.

experimenter/experimenter/cirrus/middleware.py
jaredlockhart: Yeah replicating the structure and style of the existing middlewares I think is a very sensible approach.

jaredlockhart:
@relud Looking great!  Can you provide some more detailed instructions on how to get this going locally?

relud:
to test:

- `cp .env.sample .env` or add `CIRRUS_URL=http://cirrus-experimenter:8001/v2/features/` to `.env`
- `make refresh up`
- open `localhost` in your browser
- create an experiment for experimenter with `example-feature` config `{"enabled":true}`, channel `developer`, percent of clients 100, and expected clients 100, fill out required fields as needed
- click **subscribe** and then "Preview for Testing"
- open `localhost/nimbus/?nimbus_preview=true` confirm that the page title starts with ❤️
- open `localhost/nimbus/` force-refresh, confirm that it doesn't have ❤️ in the title
- publish experiment and review it
- force refresh `localhost/nimbus/` and it should have ❤️ in the title
- end rollout and review, force refresh, no ❤️ in title

yashikakhurana:
@relud thank you for providing these instructions, I am so excited to test the experimentation on experimenter :D I will test this locally, just one question from where it is picking up channel ("staging"), I mean should we create one dev environment for local and keep staging for stage experimenter?

docker-compose.yml
```diff
@@ -144,6 +145,24 @@ services:
...
+    build:
+      additional_contexts:
+        - fml=experimenter/experimenter/features/manifests/
+      context: cirrus/server/
+      dockerfile: Dockerfile
+    env_file: .env
+    environment:
+      CIRRUS_APP_ID: experimenter.cirrus
+      CIRRUS_APP_NAME: experimenter
+      CIRRUS_CHANNEL: staging
```
yashikakhurana: ahh I see here the channel is defined

yashikakhurana (commented):
Overall the code looks good, haven't tested yet locally, but were you able to test and verify the telemetry part that if we are sending the right data?

docker-compose.yml
relud: channel for local testing is set here, i am somewhat unclear on the implications of having a dev value vs using staging, i went with staging because `CIRRUS_ENV_NAME=test_instance_stage` in `.env.sample` and didn't see any other app had a dev value in `experimenter/experimenter/experiments/constants.py` for `channel_app_id`, but maybe i misread there

experimenter/experimenter/experiments/constants.py
```diff
@@ -271,6 +271,20 @@ def kinto_collections(self) -> set[str]:
...
 )
 
+APPLICATION_CONFIG_EXPERIMENTER = ApplicationConfig(
+    name="Experimenter",
+    slug="experimenter",
+    app_name="experimenter",
+    channel_app_id={
+        Channel.STAGING: "experimenter.cirrus",
+        Channel.PRODUCTION: "experimenter.cirrus",
+    },
```
yashikakhurana: yeah here we define the channels that we want to use, we have one developer here https://github.com/mozilla/experimenter/blob/c710eb2ddef883be9492c5e3db00e31b006953e2/experimenter/experimenter/experiments/constants.py#L26 I think we can use that, if you want to add that here and then use that in config

docker-compose.yml
yashikakhurana: https://github.com/mozilla/experimenter/blob/c710eb2ddef883be9492c5e3db00e31b006953e2/experimenter/experimenter/experiments/constants.py#L26

relud:
i haven't verified that the data collection is working yet, because we don't have data review yet, and also i figured verifying that requires probe scraper to be configured to collect our metrics.yaml

is that something we can verify in our local dev environment? i was expecting to verify that in stage

yashikakhurana:
> i haven't verified that the data collection is working yet, because we don't have data review yet, and also i figured verifying that requires probe scraper to be configured to collect our metrics.yaml
> 
> is that something we can verify in our local dev environment? i was expecting to verify that in stage

Right, I mean can we verify the events if they are getting sent (may not be collected now) but if we verify in networks tab

relud:
glean was added to the django middleware, not the frontend, so we shouldn't see any new requests in the network tab

yashikakhurana:
sorry I was busy with the migration getting landed, I will test this today

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+    def __call__(self, request):
+        request.cirrus = None
+        if hasattr(request, "user") and request.user.id and self.cirrus_url:
+            params = {}
+            if (nimbus_preview := request.GET.get("nimbus_preview")) is not None:
+                params["nimbus_preview"] = nimbus_preview
+            try:
+                cirrus_response = requests.post(
+                    self.cirrus_url,
+                    # NOTE(relud): context must not be empty
```
jaredlockhart: :O Oh what really why?

jaredlockhart (approved):
Okay wow fantastic, tested locally, looks great, works great 🎉 🎉 🎉 🎉 

Great job @relud 👏 👏 👏 👏

experimenter/experimenter/cirrus/middleware.py
relud: if it's empty cirrus returns HTTP 400 `{"detail":"Context value is missing or empty"}`

yashikakhurana: but I like it how cute little note

yashikakhurana (approved):
@relud tested locally, works perfectly fine, just as a follow up if you can add these instructions in the doc to do it locally for other users

jaredlockhart:
Good point @yashikakhurana .  Also if this is blocked on data review, we should mark the pr as `blocked` and add an explanation/link so we can keep on top of it.  If it comes to pass that the data review will take too long we should move the glean parts out into a separate PR so we can land this now without blocking on that.

experimenter/experimenter/cirrus/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+                    json={"client_id": str(request.user.id), "context": {"": ""}},
```
relud: after https://github.com/mozilla/experimenter/pull/13102 context may be empty:
```suggestion
                    json={"client_id": str(request.user.id), "context": {}},
```

relud:
> @relud tested locally, works perfectly fine, just as a follow up if you can add these instructions in the doc to do it locally for other users

this was done in https://github.com/mozilla/experimenter/pull/13110

PR #12983, opened 2025-07-11
feat(nimbus): Show sticky is required message

### LINKED ISSUE #12980: Review serializer errors are not being displayed
I just noticed we can get the new frontend into a state where the review serializer is failing and yet no errors are displayed in the UI and launching is not blocked.

Example:

```
{'is_sticky': [ErrorDetail(string='Selected targeting expression requires sticky                    enrollment to function correctly', code='invalid')]}
```

and yet the UI shows it's fully launchable:

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5668)

yashikakhurana:
Because

- If the targeting selected requires is sticky to be true, then the new UI summary page doesn't show the error
- If there is error where it is not mapped to any field, show the error to the user and let them know about the issue

This commit

- Adds the fields to show the `is_sticky` error
- Adds the regular check if serializer is failing, don't let the users launch the experiment and show them the error with the message

Fixes #12980 
Screenshot 2025-07-11 at 10 30 13 AM
Screenshot 2025-07-11 at 10 29 57 AM
Screenshot 2025-07-14 at 9 52 08 AM

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1617,6 +1617,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
         serializer_data = NimbusReviewSerializer(self).data
         serializer = NimbusReviewSerializer(self, data=serializer_data)
+        self.is_ready_to_launch = serializer.is_valid()
 
         if serializer.is_valid():
             return {}
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1617,7 +1617,6 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
         serializer_data = NimbusReviewSerializer(self).data
         serializer = NimbusReviewSerializer(self, data=serializer_data)
-        self.is_ready_to_launch = serializer.is_valid()
 
         if serializer.is_valid():
             return {}
--- a/experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/launch_controls.html
+++ b/experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/launch_controls.html
@@ -1,231 +1,7 @@
 {% load nimbus_extras %}
 
 <div id="launch-controls">
-  {% if is_ready_to_launch %}
-    {% with rejection=experiment.rejection_block %}
-      {% if rejection %}
-        <div class="alert alert-warning"
-             id="rejection-reason"
-             data-testid="rejection-notice">
-          <div class="text-body">
-            <p class="mb-2">
-              The request to {{ rejection.action }} was <strong>Rejected</strong> due to:
-            </p>
-            <p class="mb-2">{{ rejection.email }} on {{ rejection.date|date:"F j, Y" }}:</p>
-            <p class="bg-light text-dark rounded border p-2 mb-0">{{ rejection.message }}</p>
-          </div>
-        </div>
-      {% endif %}
-    {% endwith %}
-    <form>
-      {% csrf_token %}
-      <!-- Draft Mode Controls -->
-      {% if experiment.is_draft %}
-        <div id="default-controls" class="alert alert-secondary">
-          <p>
-            Do you want to test this experiment before launching to production?
-            <a href="{{ EXTERNAL_URLS.PREVIEW_LAUNCH_DOC }}"
-               target="_blank"
-               class="mr-1">Learn more</a>
-          </p>
-          {% if experiment.can_draft_to_preview %}
-            <button type="button"
-                    class="btn btn-primary"
-                    hx-post="{% url 'nimbus-new-draft-to-preview' slug=experiment.slug %}"
-                    hx-select="#content"
-                    hx-target="#content"
-                    hx-swap="outerHTML">Preview for Testing</button>
-          {% endif %}
-          {% if experiment.can_draft_to_review %}
-            <button type="button"
-                    class="btn btn-secondary"
-                    onclick="showRecommendation()">Request Launch without Preview</button>
-          {% endif %}
-        </div>
-        <!-- Recommendation Message -->
-        <div id="recommendation-message" class="d-none">
-          <div class="alert alert-warning">
-            <p>
-              <strong>We recommend previewing before launch</strong>
-              <button type="button"
-                      class="btn btn-primary"
-                      hx-post="{% url 'nimbus-new-draft-to-preview' slug=experiment.slug %}"
-                      hx-select="#content"
-                      hx-target="#content"
-                      hx-swap="outerHTML">Preview Now</button>
-            </p>
-            <div class="form-check">
-              <input type="checkbox"
-                     class="form-check-input"
-                     id="checkbox-1"
-                     onchange="toggleSubmitButton()">
-              <label class="form-check-label" for="checkbox-1">I understand the risks associated with launching an experiment</label>
-            </div>
-            <div class="form-check">
-              <input type="checkbox"
-                     class="form-check-input"
-                     id="checkbox-2"
-                     onchange="toggleSubmitButton()">
-              <label class="form-check-label" for="checkbox-2">
-                I have gone through the <a href="{{ EXTERNAL_URLS.TRAINING_AND_PLANNING_DOC }}" target="_blank">experiment onboarding program</a>
-              </label>
-            </div>
-            <button type="button"
-                    class="btn btn-primary"
-                    id="request-launch-button"
-                    hx-post="{% url 'nimbus-new-draft-to-review' slug=experiment.slug %}"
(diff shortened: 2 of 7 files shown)


freshstrangemusic (approved):

jaredlockhart (dismissed):
Sorry no I don't mean add `is_sticky` as a single change, I mean we should never allow launching if the review serializer is invalid.  There are many fields missing.

freshstrangemusic (changes requested):
I didn't realize the ticket was for handling the generic case, I thought it was specifically for this one attribute.

We should do this + handle the generic case of serializer not being valid preventing launching

yashikakhurana:
@jaredlockhart @freshstrangemusic okay, so adding an extra to check to review the serializer, and if that is the case, then show the message
Screenshot 2025-07-11 at 2 12 44 PM

experimenter/experimenter/nimbus_ui_new/views.py
```diff
@@ -240,6 +241,11 @@ def get_context_data(self, **kwargs):
         if context["takeaways_edit_mode"]:
             context["takeaways_form"] = TakeawaysForm(instance=self.object)
 
+        experiment = self.object
+        review_serializer = NimbusReviewSerializer(experiment)
+        serializer = NimbusReviewSerializer(experiment, data=review_serializer.data)
+        context["ready"] = serializer.is_valid()
```
jaredlockhart: Yeah kinda except now we're doing this work twice, once in the `get_invalid_pages` and once here again, and the review serializer isn't fast, so maybe there's some way we can do this work once?

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/launch_controls.html
```diff
@@ -69,159 +113,122 @@
...
       {% endif %}
-    {% endif %}
-  </form>
+    </form>
+  {% else %}
+    <div class="alert alert-danger" role="alert">
+      This experiment cannot be launched due to an internal Experimenter error.
+      Please ask in <code>#ask-experimenter</code>.
+    </div>
+  {% endif %}
```
jaredlockhart: Should we show the actual errors even if it's not pretty?

jaredlockhart (changes requested):

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/launch_controls.html
yashikakhurana: Screenshot 2025-07-14 at 9 52 08 AM
here just for example 
Although we have mapped the is sticky field, this is just an example

experimenter/experimenter/nimbus_ui_new/views.py
yashikakhurana: yup removed that and using this method `get_invalid_fields_errors`

yashikakhurana:
Added some test cases because coverage was missing

experimenter/experimenter/nimbus_ui_new/templates/nimbus_experiments/launch_controls.html
jaredlockhart: YES!!! perfect thank you 🙏 🙏 🙏 🙏 🙏

experimenter/experimenter/experiments/models.py
```diff
@@ -1617,6 +1617,7 @@ def get_invalid_fields_errors(self):
 
         serializer_data = NimbusReviewSerializer(self).data
         serializer = NimbusReviewSerializer(self, data=serializer_data)
+        self.is_ready_to_launch = serializer.is_valid()
```
jaredlockhart: Oh I'm not sure this is exactly the right pattern, becuase then you need to call `get_invalid_field_errors` before you can access `is_ready_to_launch`, which isn't obvious, maybe another way to do it is something like

```py

@cachedproperty
def review_serializer(self):
    from experimenter.experiments.api.v5.serializers import NimbusReviewSerializer

    serializer_data = NimbusReviewSerializer(self).data
    return NimbusReviewSerializer(self, data=serializer_data)   

def get_invalid_fields_errors(self):
    if self.review_serializer.is_valid():
        return {}
    else:
        ...
```

and then in teh template like

```
{% if not experiment.review_serializer.is_valid %}
   {% for error in experiment.get_invalid_field_errors %}
```

or something like that?

jaredlockhart: Or actually maybe another way to do it is to do it something like

```py
class NimbusReviewSerializer:
    def get_invalid_field_errors(self):
        return {.....}

class NimbusExperiment:
    def get_review_serializer(self):
        from experimenter.experiments.api.v5.serializers import NimbusReviewSerializer

        serializer_data = NimbusReviewSerializer(self).data
        return NimbusReviewSerializer(self, data=serializer_data)   
```

and then in the template

```html
{% with review_serializer = experiment.get_review_serializer %}
  {% if not review_serializer.is_valid %}
    {% for field, errors in review_serializer.get_invalid_field_errors %}
```

or something?

yashikakhurana: okay just realized we don't need the review serializer again, we are calling the fields error in the validation error mixin, we can just call that and mark the status is ready to launch or not by just seeing the fields error, also extended the validation error mixin to the summary page so we can show the errors there

jaredlockhart:
Okay yes looking good, but local testing, you can have a validation error, be in a launch state, and there's no 'cancel review' button

image

jaredlockhart:
Like maybe instead of doing someting like

```html
{% if is_ready_to_launch %}
  {{ show launch controls }}
{% else %}
  {{ show warnings }}
{% endif %}
```

it should be something more like

```html
{% with review_errors=..... %}
  {% if review_errors %}
    {{ show review errors }}
  {% endif %}
  {% if should_show_review_controls %}
    {{ show reiew controls }}
  {% endif %}
{% endwith %}
```

see in the old one

image


vs the new one

image

jaredlockhart:
also notice `firefox_min_version` is another field that isn't linked to a page

jaredlockhart:
Looks like there's a lot of fields missing:

Missing from FIELD_PAGE_MAP["overview"]:
"name"
"hypothesis"

Missing from FIELD_PAGE_MAP["branches"]:
"equal_branch_ratio"
"is_localized"
"prevent_pref_conflicts"
"warn_feature_schema"

Missing from FIELD_PAGE_MAP["audience"]:
"firefox_max_version"
"is_first_run"
"proposed_release_date"

Metrics page
"primary_outcomes"
"secondary_outcomes"
"segments"


I'll do it separately https://github.com/mozilla/experimenter/issues/13000

yashikakhurana:
> Okay yes looking good, but local testing, you can have a validation error, be in a launch state, and there's no 'cancel review' button
> image

okay yeah I noticed that too, since this is getting created from the factory and leading us to this state, in practical we will not have this state as we will not launch it if there are errors

jaredlockhart:
@yashikakhurana No I know, but we can also change the review serializer over time so something that's in draft and valid today could become invalid tomrrow if we change the rules, and then it would get stuck in a state where you can't back it out of review.  The existing UI lets you cancel review even if the review serializer is invalid, we should also have that in the new UI.

yashikakhurana:
> Looks like there's a lot of fields missing:
> 
> Missing from FIELD_PAGE_MAP["overview"]: "name" "hypothesis"
> 
> Missing from FIELD_PAGE_MAP["branches"]: "equal_branch_ratio" "is_localized" "prevent_pref_conflicts" "warn_feature_schema"
> 
> Missing from FIELD_PAGE_MAP["audience"]: "firefox_max_version" "is_first_run" "proposed_release_date"
> 
> Metrics page "primary_outcomes" "secondary_outcomes" "segments"
> 
> I'll do it separately #13000

thank you for finding those

yashikakhurana:
> Like maybe instead of doing someting like
> 
> ```
> {% if is_ready_to_launch %}
>   {{ show launch controls }}
> {% else %}
>   {{ show warnings }}
> {% endif %}
> ```
> 
> it should be something more like
> 
> ```
> {% with review_errors=..... %}
>   {% if review_errors %}
>     {{ show review errors }}
>   {% endif %}
>   {% if should_show_review_controls %}
>     {{ show reiew controls }}
>   {% endif %}
> {% endwith %}
> ```
> 
> see in the old one
> image
> 
> vs the new one
> image

yeah I agree, if I do this now, I need to move and change logic at lot of the places, can I do this as part of improvement in post launch epic?

yashikakhurana:
@jaredlockhart valid point, yes if we make changes in the future, this can go in a state where user won't be able to do anything about it. Did the changes
Screenshot 2025-07-15 at 9 58 31 AM

jaredlockhart (approved):
Okay yes fantastic thank you!! 🎉 🎉 🎉 🎉

PR #12991, opened 2025-07-14
chore(nimbus): update stage link

relud:
Because

- we were moved from dataops to webservices

This commit

- Changes `stage.experimenter.nonprod.dataops.mozgcp.net` to `stage.experimenter.nonprod.webservices.mozgcp.net`


jaredlockhart (approved):
Tytytyty 🙏

PR #13003, opened 2025-07-15
chore(nimbus): map validation errors to form fields

### LINKED ISSUE #13000: Add missing fields to field page map
Looks like we're missing a lot of fields that are in the forms from the field pages map:

Looks like there's a lot of fields missing:

Missing from FIELD_PAGE_MAP["overview"]:
"name"
"hypothesis"

Missing from FIELD_PAGE_MAP["branches"]:
"equal_branch_ratio"
"is_localized"
"prevent_pref_conflicts"
"warn_feature_schema"

Missing from FIELD_PAGE_MAP["audience"]:
"firefox_max_version"
"is_first_run"
"proposed_release_date"

Metrics page
"primary_outcomes"
"secondary_outcomes"
"segments"

Let's add all those.

jaredlockhart:
Becuase

* When the review serializer is invalid, we want to direct the user to the page with the errors
* If we have an explicit map of fields to pages, we can forget to add a field to that list
* Instead we should use the list of fields defined on the forms
* That way if we add/change a field on a form, it will automatically be mapped correctly

This commit

* Removes the explicit map of fields to pages
* Moves the page mapping logic to the ValidationMixin
* Uses the form field lists to map to pages

fixes #13000

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1635,22 +1635,6 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
             return errors
 
-    @property
-    def get_invalid_pages(self):
-        field_errors = self.get_invalid_fields_errors()
-
-        if not field_errors:
-            return []
-
-        error_fields = field_errors.keys()
-        pages_with_errors = []
-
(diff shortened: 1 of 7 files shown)


freshstrangemusic:
This is much nicer than having to manually correlate everything :)

PR #13004, opened 2025-07-15
feat(nimbus): validate primary secondary outcome intersection

### LINKED ISSUE #11399: Validate primary and secondary metrics overlap in review serializer
In the old Nimbus UI, we could use react to dynamically remove options from either the primary or secondary outcomes fields when they're selected so that no outcome is marked as both a primary and secondary outcome.  In theory we could do something like this with HTMX but it might be a bit hairy, so instead let's just move this into the review validation serializer.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4812)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/79346">image-20250303-210230.png</a>

jaredlockhart:
Because
    
* The old Nimbus UI would dynamically remove selected primary/secondary outcomes from the drop downs
* We didn't do the same thing in the new HTMX UI
* This means a user could select an outcome for both primary and secondary
* We should prevent that with the review serializer
    
This commit
    
* Adds a new check to the review serializer to prevent an outcome from being selected as both primary and secondary
* Hooks up the review serializer to the metrics page
    
fixes #11399

### DIFF
--- a/experimenter/experimenter/nimbus_ui_new/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui_new/tests/test_views.py
@@ -1226,6 +1226,31 @@ class NimbusExperimentDetailViewTest(AuthTestCase):
             "`is_ready_to_launch` should be False when the review serializer is invalid",
         )
 
+    def test_ready_is_false_if_review_serializer_invalid_for_metrics(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.CREATED,
+            targeting_config_slug=NimbusExperiment.TargetingConfig.NO_TARGETING,
+            firefox_min_version=NimbusExperiment.Version.FIREFOX_120,
+            primary_outcomes=["outcome"],
+            secondary_outcomes=["outcome"],
+        )
+
+        response = self.client.get(
+            reverse("nimbus-new-detail", kwargs={"slug": experiment.slug})
+        )
+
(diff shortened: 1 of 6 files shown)


jaredlockhart:
~Blocked on https://github.com/mozilla/experimenter/pull/13003~

experimenter/experimenter/nimbus_ui_new/tests/test_views.py
```diff
@@ -1218,6 +1218,9 @@ def test_ready_is_false_if_review_serializer_invalid(self):
 
         self.assertEqual(response.status_code, 200)
         self.assertIn("validation_errors", response.context)
+        self.assertEqual(
+            set(response.context["invalid_pages"]), {"overview", "branches", "audience"}
+        )
         self.assertFalse(
             response.context["is_ready_to_launch"],
             "`is_ready_to_launch` should be False when the review serializer is invalid",
```
yashikakhurana: lets add a test for metrics page too

yashikakhurana (changes requested):
Overall looks good, just if we can add a test for invalid page for the metrics page, thank you @jaredlockhart

experimenter/experimenter/nimbus_ui_new/tests/test_views.py
jaredlockhart: Oh good idea

yashikakhurana (approved):
:shipit:

PR #13066, opened 2025-07-24
chore(nimbus): remove slug preview from clone dialog

### LINKED ISSUE #13041: Don't fetch clone slug on page load
Right now we're triggering a load to the clone slug HTMX endpoint on every page load to pre populate the slug field in the clone form.  But this is causing an unnecessary HTMX GET on every page load, even if the user never interacts with the clone form.  Instead let's pre-populate the slug using the [slugify](https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#slugify) template filter on the initial page load, and then we can use the HTMX swap endpoint for subsequent edits to the name if the user engages with the clone form.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5698)

jaredlockhart:
Becuase

* We had added a readonly slug preview field to the clone dialog in the new ui becuase it was in the old ui
* However to adapt this to HTMX we had to add some additional server side mechanics to use the django slugify function dynamically
* This is a lot of machinery for a small feature
* After polling users it seems this feature is unnecessary

This commit

* Removes the slug preview from the clone dialog altogether

fixes #13041

image


yashikakhurana (approved):
simple and clean, -58 🎉 🎉 , good catch @jaredlockhart

PR #13084, opened 2025-07-25
bug(nimbus): redirect form pages to summary when not editable

### LINKED ISSUE #13082: Form pages should redirect to summary for non editable experiments
Once an experiment is in a non editable state if the user navigates directly to a form page we should redirect back to summary.  Note that the audience page for live rollouts is editable.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5741)

jaredlockhart:
Because

* The forms shouldn't be accessible for experiments that aren't editable, ie preview/live/complete
* The audience form should be accessible for live rollouts

This commit

* Adds redirects for each of the form views when that form is not editable

fixes #13082


yashikakhurana (approved):
I like the use of mixing here, and then re-using those functions,thanks @jaredlockhart

PR #13094, opened 2025-07-25
chore(cirrus) explicitly set uvicorn workers to 1

relud:
Because

- Just in case so that it never tries to set it based on the number of available CPUs or anything

This commit

- add `--workers 1` to the cirrus docker image default command


jaredlockhart (approved):
1️⃣ 🙏

PR #13095, opened 2025-07-25
chore(cirrus) use uid and gid 10001

relud:
Because

- @jaredlockhart pointed out in https://github.com/mozilla/webservices-infra/pull/6682#pullrequestreview-3056779940 that 10001 is the value recommended by dockerflow

This commit

- Changes cirrus uid and gid to 10001


jaredlockhart (approved):
🙏

PR #13102, opened 2025-07-28
fix(cirrus): don't require context

### LINKED ISSUE #13105: Allow cirrus to take empty context
Cirrus is not configured to take empty context, it returns 400, we should allow Cirrus to take empty context as the input

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5748)

relud:
Because

- context should be optional

This commit

- changes context from required to using a default value of `{}`

Fixes #13105

### DIFF
--- a/cirrus/server/tests/test_main.py
+++ b/cirrus/server/tests/test_main.py
@@ -89,7 +89,7 @@ def test_get_features_v2_with_required_field(client):
 @pytest.mark.parametrize(
     "request_data, expected_status, expected_message",
     [
-        (
+        pytest.param(
             {
                 "client_id": "",
                 "context": {
@@ -99,8 +99,9 @@ def test_get_features_v2_with_required_field(client):
             },
             status.HTTP_400_BAD_REQUEST,
             "Client ID value is missing or empty",
+            id="empty_client_id",
         ),
-        (
+        pytest.param(
             {
                 "context": {
                     "key1": "value1",
@@ -121,13 +122,9 @@ def test_get_features_v2_with_required_field(client):
                     },
                 }
             ],
+            id="missing_client_id",
         ),
-        (
-            {"client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449", "context": {}},
-            status.HTTP_400_BAD_REQUEST,
-            "Context value is missing or empty",
-        ),
-        (
+        pytest.param(
             {"client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449"},
             status.HTTP_422_UNPROCESSABLE_ENTITY,
             [
@@ -138,8 +135,9 @@ def test_get_features_v2_with_required_field(client):
                     "input": {"client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449"},
                 }
             ],
+            id="missing_client_id_and_context",
         ),
-        (
+        pytest.param(
             {},
             status.HTTP_422_UNPROCESSABLE_ENTITY,
             [
@@ -156,6 +154,7 @@ def test_get_features_v2_with_required_field(client):
                     "input": {},
                 },
             ],
+            id="missing_context",
         ),
     ],
 )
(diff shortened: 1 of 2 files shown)


jaredlockhart (commented):
Oh I'm not sure we should make the context completely optional entirely, I only meant that it could be empty.  I dunno is there a big difference between the two cases? 🤔 Should we keep it in but support an empty object?  I worry about making it optional entirely and some consumers baking that assumption into their code and then later we find some other reason to depend on the object in some case?

relud:
> Oh I'm not sure we should make the context completely optional entirely, I only meant that it could be empty. I dunno is there a big difference between the two cases?

I don't think so

> 🤔 Should we keep it in but support an empty object?

That is a trivial change to make

> I worry about making it optional entirely and some consumers baking that assumption into their code and then later we find some other reason to depend on the object in some case?

i don't think so, but i don't feel strongly

cirrus/server/tests/test_main.py
```diff
@@ -122,23 +122,6 @@ def test_get_features_v2_with_required_field(client):
...
-                {
-                    "type": "missing",
-                    "loc": ["body", "context"],
-                    "msg": "Field required",
-                    "input": {"client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449"},
-                }
-            ],
-        ),
         (
             {},
```
yashikakhurana: I will not remove this test case and instead this should return valid default feature now

cirrus/server/tests/test_main.py
```diff
@@ -149,12 +132,6 @@ def test_get_features_v2_with_required_field(client):
                     "msg": "Field required",
                     "input": {},
                 },
-                {
-                    "type": "missing",
-                    "loc": ["body", "context"],
-                    "msg": "Field required",
-                    "input": {},
-                },
```
yashikakhurana: same here

cirrus/server/tests/test_main.py
```diff
@@ -444,13 +392,7 @@ def test_get_features_v1_with_nimbus_preview_flag(client):
...
 
 def test_get_features_v2_with_nimbus_preview(client):
-    request_data = {
-        "client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449",
-        "context": {
-            "key1": "value1",
-            "key2": {"key2.1": "value2", "key2.2": "value3"},
-        },
-    }
+    request_data = {"client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449"}
```
yashikakhurana, line 472: split this into two one with context with key value and one with `{}`

cirrus/server/tests/test_main.py
relud: these tests are called like `test_get_features_v1_missing_required_field`, so I think it's misleading for future developers to have cases on it that aren't missing a required field.

Instead I modified `test_get_features_v1_with_nimbus_preview_flag` and `test_get_features_v2_with_nimbus_preview` to cover testing empty context (and updated to parametrize for both empty and nonempty per your review).

cirrus/server/tests/test_main.py
relud: this is no longer touched, so it's no longer relevant, but also this isn't a test case that was removed, it was expecting a different return value

cirrus/server/tests/test_main.py
```diff
@@ -99,8 +99,9 @@ def test_get_features_v2_with_required_field(client):
             },
             status.HTTP_400_BAD_REQUEST,
             "Client ID value is missing or empty",
+            id="empty_client_id",
```
relud, line 102: I went through all the parametrized tests in this file and added ids because i couldn't tell them apart when debugging

cirrus/server/tests/test_main.py
```diff
@@ -443,13 +451,23 @@ def test_get_features_v1_with_nimbus_preview_flag(client):
     }
 
 
-def test_get_features_v2_with_nimbus_preview(client):
+@pytest.mark.parametrize(
```
yashikakhurana, line 454: oh thank you, yeah perfect, v1 and v2 both with nimbus preview flag, can you also do the same for without preview flag

yashikakhurana (approved):
Perfect, all in one place, thank you for adding those params 🙏

PR #13107, opened 2025-07-28
feat(nimbus): test PR to observe CI

cyndichin:
Because

- Discussed with @freshstrangemusic on testing the CI to see what it looks like if we set advanced targeting here without making modifications to iOS code

This commit

- Should not be merged in


cyndichin:
Seems like we do fail because we are missing some fields in https://github.com/mozilla/experimenter/blob/main/experimenter/tests/integration/nimbus/test_mobile_targeting.py#L92-L105
Screenshot 2025-07-28 at 12 26 52 PM

Need to check if we fail after we add this for tests.

cyndichin:
@freshstrangemusic were you able to confirm if tests without the iOS changes? I assume so based on this [commit](https://github.com/mozilla/experimenter/pull/13107/commits/9d2c6c46cb685c77afa04d1a5c60475eb324a718)?

PR #13109, opened 2025-07-28
fix(nimbus): Handle features without schemas in new UI

### LINKED ISSUE #13076: [HTMX UI] Using the "No Feature" feature for Android and iOS results in an error on the Branches page.
When using the feature “No Feature Android” or “No Feature iOS”, the feature configuration field on the branches page does not load, but the page reports an error: “This field may not be blank.” There is also no way to edit this field.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-5738)

freshstrangemusic:
Because:

- some features do not have schemas (e.g., "No Feature iOS" and "No Feature Fenix")

this commit:

- only serializes the schema to the dataset API if the schema exists; and
- guards against schemas not existing in the frontend.

Fixes #13076


jaredlockhart (approved):
Okay tested locally, looks good works good thnx @freshstrangemusic 🙏 

I filed https://github.com/mozilla/experimenter/issues/13113 as a followup 👍

yashikakhurana (approved):
Thank you @freshstrangemusic looking good 💯

PR #13129, opened 2025-07-29
feat(nimbus): Add advanced targeting for tou users

t-p-white:
Because

- We want to be able to target users that have accepted or not accepted the Terms of Use (ToU). See associated https://bugzilla.mozilla.org/show_bug.cgi?id=1970135 for more details.

This commit

- Adds the advanced targeting for both user groups (accted and not accepted)

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1970135
Fixes https://mozilla-hub.atlassian.net/browse/FXDROID-4666


jaredlockhart (approved):
Looks good, I updated the test harness 👍

PR #13131, opened 2025-07-29
 feat(nimbus): Removed the redundant review checker advanced targeting from Android

t-p-white:
Because

- Feature is now redundant on Android

This commit

- Removes the targeting option.

Fixes #github_issue_number


t-p-white:
~~Please **hold off merging** until we have clarification that this is okay to remove.~~

This is good to marge if the PR is accepted

jaredlockhart (approved):
Okay cool it looks like this wasn't used at all, safe to remove.  Thanks @t-p-white 🙏

PR #13156, opened 2025-07-30
chore(nimbus): remove duplicate end rollout logic

### LINKED ISSUE #13153: Remove duplicate experiment/rollout views/forms
We had originally build separate paralle sets of forms and views for ending experiments and rollouts, however the diagrams show that these should be the same flow, and some differences they originally contained were removed, so now they are identical.  It should be safe now to remove the duplicates.

jaredlockhart:
Becuase

* We had originally built separate flows for ending experiments and rollouts
* However this was based on an incorrect set of assumptions about differences in their flows
* Actually ending experiments and rollouts should have identical flows

This commit

* Removes duplicate set of end rollout forms/views
* Removes the assumption that all experiments must pause before ending
* Adds additional tests that were missing for some methods

fixes #13153


experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -2397,108 +2396,6 @@ def test_cancel_end_experiment_view(self):
...
-        self.experiment.is_paused = True
-        self.experiment.is_rollout = True
-        self.experiment.save()
-
-        response = self.client.post(
-            reverse(
-                "nimbus-ui-cancel-end-rollout", kwargs={"slug": self.experiment.slug}
-            ),
-            data={
-                "changelog_message": "Rollout not complete yet.",
```
yashikakhurana, line 2435: I know we already have those tests for ending experiments, but can we add or parameterize them to work with rollouts too?

yashikakhurana:
okay I tested on main, the error exist on stage and prod already

jaredlockhart:
@yashikakhurana Ah great catch!!!! Okay I fixed that as well.

PR #13212, opened 2025-08-06
add win10 eos toast targeting config

aminomancer:


experimenter/experimenter/targeting/constants.py
```diff
@@ -1007,6 +1007,20 @@ def __post_init__(self):
...
+    slug="win10_eos_sync_toast_eligible",
+    description="Users in the Windows 10 EOS Sync treatment branch",
+    targeting=(
+        "defaultProfile.enrollmentsMap['optin-windows-10-eos-sync-messaging'] "
+        "== 'treatment-a'"
+    ),
+    desktop_telemetry="",
+    sticky_required=False,
+    is_first_run_required=False,
+    application_choice_names=(Application.DESKTOP.name,),
```
jaredlockhart: Would this targeting be for the updater?

aminomancer: Yeah. The updater's targeting context gets a `defaultProfile` property, which is a snapshot of the default firefox profile's targeting context. So this lets us see if the real profile is enrolled in the branch we want.

jaredlockhart: Ah I see, to force this to only run in the background task you'll want to include 

`isBackgroundTaskMode`

PR #13221, opened 2025-08-07
feat(nimbus): add channels field to nimbus experiment

### LINKED ISSUE #13203: Support multiple channels in NimbusExperiment model
Let’s add a new field to NimbusExperiment called channels that can store multiple slugs, similar to what we do now for the outcomes fields.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5730)

jaredlockhart:
Becuase

* On desktop we want to be able to target multiple channels with a single recipe
* Unfortunately we can not do the same for any other application for now
* We will add an arrayfield called channels and then use it for desktop and continue to use channel for all other applications

This commit

* Adds a field called channels to NimbusExperiment which is an arrayfield of charfields

fixes #13203


yashikakhurana (approved):
multiple channels 🎉 , nice step by step changes, thank you @jaredlockhart

PR #13228, opened 2025-08-07
feat(nimbus): add channels to jexl targeting

### LINKED ISSUE #13222: Serialize multiple channels into JEXL targeting
Once we have the channels field on NimbusExperiment, we can use it to generate the multi channel targeting expression.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5731)

jaredlockhart:
Becuase
    
* We added a new field called channels for desktop to target multiple channels
    
This commit
    
* Adds a channels clause to the jexl targeting for desktop
* Preserves the existing single channel case while we migrate the forms over to use the new field
    
fixes #13222

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -572,8 +572,12 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         if self.targeting_config and self.targeting_config.targeting:
             sticky_expressions.append(self.targeting_config.targeting)
 
-        if self.is_desktop and self.channel:
-            expressions.append(f'browserSettings.update.channel == "{self.channel}"')
+        if self.is_desktop:
+            if self.channels:
+                channels = json.dumps(sorted(self.channels))
+                expressions.append(f"browserSettings.update.channel in {channels}")
+            elif self.channel:  # TODO remove in #13224
+                expressions.append(f'browserSettings.update.channel == "{self.channel}"')
 
         sticky_expressions.extend(self._get_targeting_min_version())
         expressions.extend(self._get_targeting_max_version())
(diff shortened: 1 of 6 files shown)


jaredlockhart:
~Blocked on https://github.com/mozilla/experimenter/pull/13221~

experimenter/experimenter/experiments/models.py
```diff
@@ -572,8 +572,12 @@ def targeting(self):
         if self.targeting_config and self.targeting_config.targeting:
             sticky_expressions.append(self.targeting_config.targeting)
 
-        if self.is_desktop and self.channel:
-            expressions.append(f'browserSettings.update.channel == "{self.channel}"')
+        if self.is_desktop:
+            if self.channels:
+                channels = json.dumps(sorted(self.channels))
```
yashikakhurana, line 577: why are we doing sorting? just curious to know

experimenter/experimenter/experiments/tests/api/v6/test_serializers.py
```diff
@@ -60,9 +61,9 @@ def test_expected_schema_with_desktop(self):
         branches_data = [dict(b) for b in experiment_data.pop("branches")]
         feature_ids_data = experiment_data.pop("featureIds")
 
-        assert experiment.start_date
-        assert experiment.actual_enrollment_end_date
-        assert experiment.end_date
+        self.assertIsNotNone(experiment.start_date)
+        self.assertIsNotNone(experiment.actual_enrollment_end_date)
+        self.assertIsNotNone(experiment.end_date)
```
yashikakhurana, line 66: self asserts 🎉

experimenter/experimenter/experiments/models.py
jaredlockhart: Oh it's so that you don't get the situation where the values jump around in the db (the order isn't guaranteed) and then later you regenerate the recipe, and just the channels have jumped around, and then you're trying to compare it to another copy of the recipe to look for a difference and that's different, so you just force it to always be ordered.

experimenter/experimenter/experiments/tests/api/v6/test_serializers.py
jaredlockhart: Yeah I just noticed that and thought I'd sneak in a little cleanup 🙈

yashikakhurana (approved):
looks good, thank you for explaining the changes as well

PR #13232, opened 2025-08-07
    feat(nimbus): use channels in nimbus ui

### LINKED ISSUE #13230: Update all templates to read both channel/channels
Once we have the channels field, we’ll need to update the templates to display either channel or channels depending on which app it is.  We can make a nice little wrapper field and then use that everywhere in the templates.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5788)

jaredlockhart:
Becuase
    
* We added a new channels field for desktop experiments
* We kept the existing channel field for all non-desktop experiments
* We only want to show channel(s) in one place so we need to handle both cases in each place a channel would be shown
    
This commit
    
* Overrides the get_channel_display method on NimbusExperiment to handle both the channels and channel case
* Updates the templates to always call get_channel_display where ever a channel is shown
    
fixes #13230

image
image
image

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1822,6 +1822,13 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         return transformed_changelogs
 
     @property
+    def get_channel_display(self):
+        if self.is_desktop and self.channels:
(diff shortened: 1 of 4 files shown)


jaredlockhart:
~Blocked on https://github.com/mozilla/experimenter/pull/13228~

yashikakhurana (approved):
tested localy, it works great, thank you @jaredlockhart

PR #13244, opened 2025-08-08
feat(nimbus): use channels in table filters

### LINKED ISSUE #13239: Support channel and channels in table filters
We need to be able to filter against channel and channels in the table filters on the list page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5793)

jaredlockhart:
Becuase

* We want to be able to filter on both the channel and channels fields
* This should be transparent to the user so there should only be one channel filter

This commit

* Searches both the channel and channels fields in the channel filter on the table page

fixes #13239


yashikakhurana (approved):
:shipit: nice simple change

PR #13256, opened 2025-08-11
bug(nimbus): handle 404 from kinto delete

### LINKED ISSUE #13254: Handle Kinto 404 in ending experiment task
I just noticed a case where we request end for an experiment, the record is successfully removed in RS, and yet experimenter mistakenly interpreted the state in RS as the request timed out.  In that case all you can do is request end a second time, the delete record call will 404 and the task will halt there, and the experiment is now stuck in ending approved state but will never move to end becuase we only move to COMPLETE/IDLE from LIVE/WAITING.  

So we should handle the 404 in the delete task and continue to move on to LIVE/WAITING so that the ending task will pick it up and check that it's removed from RS and move it to COMPLETE.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5805)

jaredlockhart:
Because

* We recently saw an experiment where experimenter mistakenly interpreted a timed out review for an ending experiment from remote settings
* This puts the experiment back in approved state waiting for the remote settings review to take place
* However the record was successfully removed from remote settings
* The experiment should now be in waiting state to move on to the next step in the flow of moving to complete
* Having the experiment stuck in approved means it can not move forward to waiting and then to complete
* Re-requesting ending will trigger the delete task which will fail for 404 becuase the record is already moved
* In this situation we should catch the 404 and proceed to marking the experiment as waiting since it is already successfully removed from the collection

This commit

* Catches 404 from kinto delete and proceeds to mark the experiment as waiting
* Reraises any other error code from kinto

fixes #13254

### DIFF
--- a/experimenter/experimenter/kinto/tasks.py
+++ b/experimenter/experimenter/kinto/tasks.py
@@ -3,6 +3,7 @@ from celery.utils.log import get_task_logger
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.utils import timezone
+from kinto_http import KintoException
 
 from experimenter.celery import app
 from experimenter.experiments.api.v6.serializers import NimbusExperimentSerializer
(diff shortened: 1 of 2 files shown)


yashikakhurana (approved):
oh nice job catching this case, thanks @jaredlockhart

PR #13260, opened 2025-08-11
bug(nimbus): Prevent publishing recipes containing secure features to…

### LINKED ISSUE #13251: recipes with prefFlips and newtabTrainhopAddon are previewable
The prefFlips and newtabTrainhopAddon features not indended to be previewable because they require publishing to a specific collection and we reject those features from all other collections.

This was true in the old UI but we must have missed it when porting to HTMX.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5802)

freshstrangemusic:
… preview

Because:

- the old UI prevented publishing secure features to the preview collection and that was missed in the HTMX rewrite; and
- Firefox clients will reject recipes in the preview collection

this commit:

- prevents publishing of recipes containing secure features to preview.

Fixes #13251

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -732,7 +732,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def can_draft_to_preview(self):
-        return self.is_draft and not self.is_review
+        return self.is_draft and not self.is_review and self.can_publish_to_preview
 
     @property
     def can_draft_to_review(self):
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/launch_controls.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/launch_controls.html
@@ -38,41 +38,56 @@
     {% csrf_token %}
     <!-- Draft Mode Controls -->
     {% if experiment.is_draft %}
-      <div id="default-controls" class="alert alert-secondary">
-        <p>
-          Do you want to test this experiment before launching to production?
-          <a href="{{ EXTERNAL_URLS.PREVIEW_LAUNCH_DOC }}"
-             target="_blank"
-             class="mr-1">Learn more</a>
-        </p>
-        {% if experiment.can_draft_to_preview %}
-          <button type="button"
-                  id="draft-to-preview-button"
-                  class="btn btn-primary"
-                  hx-post="{% url 'nimbus-ui-draft-to-preview' slug=experiment.slug %}"
-                  hx-select="#content"
-                  hx-target="#content"
-                  hx-swap="outerHTML">Preview for Testing</button>
-        {% endif %}
-        {% if experiment.can_draft_to_review %}
-          <button type="button"
-                  id="draft-to-review-button"
-                  class="btn btn-secondary"
-                  onclick="showRecommendation()">Request Launch without Preview</button>
-        {% endif %}
-      </div>
-      <!-- Recommendation Message -->
-      <div id="recommendation-message" class="d-none">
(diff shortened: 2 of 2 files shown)


relud (approved):

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/launch_controls.html
```diff
@@ -38,41 +38,55 @@
...
+          {% if experiment.can_draft_to_review %}
+            <button type="button"
+                    id="draft-to-review-button"
+                    class="btn btn-secondary"
+                    onclick="showRecommendation()">Request Launch without Preview</button>
+          {% endif %}
+        </div>
+      {% endif %}
+      <!-- Recommendation Message -->
+      <div id="recommendation-message"{% if experiment.can_publish_to_preview %}class="d-none"{% endif %}>
```
relud: pretty sure this is missing a space:
```suggestion
      <div id="recommendation-message"{% if experiment.can_publish_to_preview %} class="d-none"{% endif %}>
```

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -1195,6 +1195,10 @@ class DraftToPreviewForm(UpdateStatusForm):
     status_next = NimbusExperiment.Status.PREVIEW
     publish_status = NimbusExperiment.PublishStatus.IDLE
 
+    def clean(self):
+        if not self.instance.can_publish_to_preview:
+            raise forms.ValidationError("This experiment cannot be previewed")
```
jaredlockhart: Yeah interesting idea, I don't think we do this anywhere else now, since validation failures of this kind aren't caught and displayed in the UI so I'm not sure sticking this here is the right way to catch cases like this.  I think this would just fail silently, possibly break the flow?

experimenter/experimenter/experiments/models.py
```diff
@@ -732,7 +732,7 @@ def is_started(self):
 
     @property
     def can_draft_to_preview(self):
-        return self.is_draft and not self.is_review
+        return self.is_draft and not self.is_review and self.can_publish_to_preview
```
jaredlockhart, line 735: Also very concise way to represent this 🙏 🙏 🙏

jaredlockhart:
A simpler way to manage the UI here might be to just conditionally disable the preview button with a tool tip rather than making a second copy of the entire launch/review flow?

yashikakhurana:
+1 to conditionally disable the preview flow from users point of view too so that they are aware why it is disable, instead of removing it completely

experimenter/experimenter/nimbus_ui/forms.py
freshstrangemusic: you should never be able to submit this form because you should never be able to render it -- i can remove this if you want

freshstrangemusic:
> +1 to conditionally disable the preview flow from users point of view too so that they are aware why it is disable, instead of removing it completely

It is very clear /why/ it is disabled. There is a big warning message.

> A simpler way to manage the UI here might be to just conditionally disable the preview button with a tool tip rather than making a second copy of the entire launch/review flow?

We dont show the preview button at the same time as the cancel/publish button (only one set of controls are displayed at a time)

jaredlockhart (approved):
We should probably just pull the preview flow out into a separate set of controls so it's not so mixed in with the launch controls, prolly simpler.  But anyways tested all the combinations I could think of, everything looked right.

Thnx @freshstrangemusic 🙏 🎉

experimenter/experimenter/nimbus_ui/forms.py
jaredlockhart: Right exactly, that's how we handle all the other cases, just don't make the button available.

PR #13277, opened 2025-08-12
feat(nimbus): Sort my deliveries columns

### LINKED ISSUE #13263: My Stuff IV- sorting on all columns
On My stuff table, we want to sort all the columns that we will add. 

Below are the columns

Name (name should be clickable and leading to the experimenter summary page)
status
QA
Application
Type
Channel
Size
Dates/Days remaining
Versions
Feature
Results

We should be able to sort this and wants this to be consistent with the current home page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5718)

yashikakhurana:
Because

- We want to add the ability to sort the columns of the My Deliveries section

This commit

- Allow users to sort all columns of the my deliveries table
- Persist the filtering of owned/subscribed or pagination with the sorting

Fixes #13263


yashikakhurana:
Screenshot 2025-08-11 at 2 50 05 PM
Screenshot 2025-08-11 at 2 49 25 PM
Screenshot 2025-08-11 at 2 49 07 PM
Screenshot 2025-08-11 at 2 48 48 PM
Screenshot 2025-08-11 at 2 48 42 PM
Screenshot 2025-08-11 at 2 48 15 PM

experimenter/experimenter/nimbus_ui/templates/common/home_sortable_header.html
```diff
@@ -0,0 +1,34 @@
...
+        hx-trigger="submit"
+        hx-target="{{ hx_target }}"
+        hx-select="{{ hx_select }}"
+        class="d-inline-block m-0">
+    {# Preserve all current filters except "sort" #}
+    {% for field in form %}
+      {% if field.name != "sort" %}
+        {% if field.value %}{{ field.as_hidden }}{% endif %}
+      {% endif %}
+    {% endfor %}
```
freshstrangemusic, line 13: `field` is already defined in this template. We should use a different variable name here.

freshstrangemusic (dismissed):

experimenter/experimenter/nimbus_ui/filtersets.py
```diff
@@ -316,6 +316,41 @@ class MyDeliveriesChoices(models.TextChoices):
...
+
+
+HOME_SORTABLE_HEADERS = []
+seen = set()
+
+for choice in HomeSortChoices:
+    field = choice.value.lstrip("-")
+    if field not in seen:
+        HOME_SORTABLE_HEADERS.append((field, choice.label))
+        seen.add(field)
```
jaredlockhart: Wait waht's happening here?  Why are there global variables in this file?

yashikakhurana: sending this in UI so that we don't have to hard-code in templates

jaredlockhart: Sure but can't we do that in the context method of the relevant view?  In general I'd avoid this kind of bare logic at the top level of a module.  It's stateful, and will get executed at import time.  That can cause confusing/unexpected behaviour later.

yashikakhurana: sure I can move this to the view

experimenter/experimenter/nimbus_ui/templates/common/home_sortable_header.html
yashikakhurana: umm its only defined once, I don't see it defined again, can you share where do you see it

experimenter/experimenter/nimbus_ui/tests/test_filters.py
```diff
@@ -153,3 +154,15 @@ def test_filter_returns_all_deliveries_experiments(self):
...
+        for i, sort_choice in enumerate(HomeSortChoices):
+            with self.subTest(sort=sort_choice.value):
+                NimbusExperimentFactory.create(
+                    owner=self.user,
+                    name=f"Sort Test {i}",
+                )
+                response = self.client.get(
+                    f"{reverse('nimbus-ui-home')}?sort={sort_choice.value}"
+                )
+                self.assertEqual(response.status_code, 200)
```
jaredlockhart: Yeah so, this is okay to check that we don't 400/500 here, but we probably should have something that actually tests the sorting for each field.  Looking at the landing page we do have tests that explicitly test sorting by each field.  There's probably a way to do it for each column as a parameterized test.

Also wow I didn't know about `self.subTest`! You learn something new every day.  But also I think this is another way to do what we already do with `@parameterized.expand`, so I wonder if it's not better to keep things consistent by using parameterized here?

jaredlockhart (changes requested):
Okay tested locally, looks and works great 🎉 🎉 🎉 

But we probably should have tests that actually assert on sorting by each sortable field 🙏

experimenter/experimenter/nimbus_ui/tests/test_filters.py
yashikakhurana: added test for each fields

experimenter/experimenter/nimbus_ui/tests/test_filters.py
```diff
@@ -153,3 +159,98 @@ def test_filter_returns_all_deliveries_experiments(self):
...
+            firefox_min_version=120,
+        )
+        high = NimbusExperimentFactory.create(
+            owner=self.user,
+            name="Z High",
+            application=NimbusExperiment.Application.DESKTOP,
+            is_rollout=True,
+            channel=NimbusExperiment.Channel.RELEASE,
+            population_percent=50,
+            firefox_min_version=130,
```
b4handjr, line 196: What does `low` and `high` mean here?

experimenter/experimenter/nimbus_ui/tests/test_filters.py
```diff
@@ -153,3 +159,98 @@ def test_filter_returns_all_deliveries_experiments(self):
...
+        [
+            (HomeSortChoices.NAME_UP, HomeSortChoices.NAME_DOWN),
+            (HomeSortChoices.APPLICATION_UP, HomeSortChoices.APPLICATION_DOWN),
+            (HomeSortChoices.TYPE_UP, HomeSortChoices.TYPE_DOWN),
+            (HomeSortChoices.CHANNEL_UP, HomeSortChoices.CHANNEL_DOWN),
+            (HomeSortChoices.SIZE_UP, HomeSortChoices.SIZE_DOWN),
+            (HomeSortChoices.VERSIONS_UP, HomeSortChoices.VERSIONS_DOWN),
+        ]
+    )
+    def test_sorting_changes_first_row_for_choice(self, sort_up, sort_down):
```
jaredlockhart, line 179: Yep good perf ty 🙏

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -669,4 +670,13 @@ def get_context_data(self, **kwargs):
...
         context["links"] = NimbusUIConstants.HOME_PAGE_LINKS
         context["tooltips"] = NimbusUIConstants.HOME_PAGE_TOOLTIPS
+
+        seen = set()
+        headers = []
+        for choice in HomeSortChoices:
+            field = choice.value.lstrip("-")
+            if field not in seen:
+                headers.append((field, choice.label))
+                seen.add(field)
```
jaredlockhart: Oh I was also thinking this could be a staticmethod on the enum itself, doesn't make a huge difference, but ty for moving out of global context 🙏

experimenter/experimenter/nimbus_ui/tests/test_filters.py
yashikakhurana: for the sorting, to make it clear the order

experimenter/experimenter/nimbus_ui/views.py
yashikakhurana: sure I can move that

PR #13281, opened 2025-08-13
fix(experimenter): added missing package: write permission for mozcloud-publish action

dlactin:
Because

- Ran into a few issues sharing the Docker context between the reusable build-push action and the images created via make 

This commit

- Opted to reuse the existing `make build_prod` commands to build the experimenter images and use the re-usable `docker-push` action to push the resulting images to GAR. 
- Copied the `Generate MozCloud Tag` step from the re-usable `docker-build` action 
- Specifying the PR head ref in the checkout step to ensure we're building an image with the same tag Argo expects. (https://github.com/orgs/community/discussions/26325)

### DIFF
--- a/.github/workflows/experimenter-mozcloud-publish.yaml
+++ b/.github/workflows/experimenter-mozcloud-publish.yaml
@@ -12,8 +12,12 @@ on:
       - 'experimenter/**'
   workflow_dispatch: {}
 
+env:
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
🎉 🎉 🎉 🎉 🎉

PR #13291, opened 2025-08-15
chore(experimenter): require experimenter FML generated updates to be committed

### LINKED ISSUE #13171: Require experimenter FML to be locally generated instead of by manifest tool
when [~accountid:5b33f755731cfc3ac2027747] first added the FML for experimenter, they configured it in manifest tool like we do for FML in other repos.

[~accountid:6047ca43b2ea12011598cfed] has pointed out that this is inconvenient because it causes a delay between modifying the experimenter fml config and generating the experimenter fml files from it, so we should instead add a make command that generates experimenter’s FML, and add a CI job to require that generated FML files are up to date when the config changes.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5679)

relud:
Because

- Changes to experimenter's nimbus.fml.yaml should be included in PRs rather than deferred to the next automated manifest tool run.

This commit

- Changes manifest tool to use the local version of experimenter's nimbus.fml.yaml
- Adds a circleci job to check that generated fml for experimenter is up to date

Fixes #13171

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -232,6 +232,29 @@ jobs:
           destination: gs://ecosystem-test-eng-metrics/experimenter/coverage
           extension: json
 
+  check_local_feature_manifests:
+    machine:
+      image: ubuntu-2204:2024.11.1
+      docker_layer_caching: true
+    resource_class: xlarge
+    working_directory: ~/experimenter
+    steps:
+      - checkout
+      - check_file_paths:
+          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/"
+      - run:
+          name: Check local feature manifests
+          command: |
+            cp .env.sample .env
+            make feature_manifests FETCH_ARGS="--local-apps"
+            if ! git diff --quiet experimenter/experimenter/features/manifests/; then
+              echo 'Please commit generated updates to local feature manifests:'
+              echo
+              echo '    make feature_manifests FETCH_ARGS="--local-apps"'
+              echo
+              exit 1
+            fi
+
   check_schemas:
     machine:
       image: ubuntu-2204:2024.11.1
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -232,6 +232,29 @@ jobs:
           destination: gs://ecosystem-test-eng-metrics/experimenter/coverage
           extension: json
 
+  check_local_feature_manifests:
+    machine:
+      image: ubuntu-2204:2024.11.1
+      docker_layer_caching: true
+    resource_class: xlarge
```
b4handjr: Do we need `xlarge` for this?

experimenter/manifesttool/appconfig.py
```diff
@@ -28,6 +29,16 @@ def validate_default_branch(cls, values: dict[str, Any]):
 
         return values
 
+    @model_validator(mode="before")
+    def validate_name(cls, values: dict[str, Any]):
+        ty = values.get("type")
```
b4handjr: Could we do something like `_type`?

experimenter/manifesttool/cli.py
```diff
@@ -52,19 +52,38 @@ def main(ctx: click.Context, *, manifest_dir: Path):
...
+            if app_name not in context.app_configs.root:
                 print(f"fetch: unknown app {app_name}", file=sys.stderr)
                 sys.exit(1)
     else:
-        app_names = context.app_configs.root.keys()
+        app_names = [
+            app_name
+            for app_name, app_config in context.app_configs.root.items()
+            # use ^ (XOR) to choose exclusively locally or remotely configured apps
+            if local_apps ^ (app_config.repo.type != RepositoryType.LOCAL)
```
b4handjr, line 85: neat!

b4handjr (changes requested):
Just had a few thoughts.

experimenter/manifesttool/appconfig.py
relud: I copied `ty` from the method right above this, happy to use `_type` instead

.circleci/config.yml
relud: we do not, that was copypasta from `check_experimenter_x86_64`, removed

b4handjr (approved):
🚀

PR #13313, opened 2025-08-20
feat(nimbus): Add validation for Firefox Labs fields

### LINKED ISSUE #13304: Firefox Labs fields are not validated
Its possible to launch a labs delivery with `firefoxLabsDescriptionLinks = ""`.
If you set the firefoxLabsDescriptionLinks field to a JSON value, it actually gets set to a string of that value instead of the object.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5847)

freshstrangemusic:
Because:

- Firefox Labs fields were originally behind the Django admin panel and not user-editable and thus we decided not to add validation to the `NimbusReviewSerializer` at implementation time;
- we have completed the UI migration to htmx and we now have Firefox Labs fields that are user-editable;
- these fields were added without adding validation to the `NimbusReviewSerializer`, thus making it possible to create and launch invalid Firefox Labs rollouts;
- the `firefox_labs_description_links` field was defined as a JSON field in the model, but the UI was defined as a text field, leading to differences between the serialized and saved data; and
- we were saving the NimbusExperiment multiple times in `NimbusBranchesForm.save()`

this commit:

- adds validation to the `NimbusReviewSerializer` for Firefox Labs fields;
- migrates the `firefox_labs_description_links` field to a `TextField`; and
- adds a `NimbusBranchesForm.clean()` to enforce that all Firefox Labs deliveries are rollouts.

Fixes #13304


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -2316,6 +2329,80 @@ def _validate_primary_secondary_outcomes(self, data):
...
+
+                elif (
+                    not isinstance(description_links_obj, dict)
+                    and description_links_obj is not None
+                ):
+                    errors["firefox_labs_description_links"] = [
+                        NimbusExperiment.ERROR_FIREFOX_LABS_DESCRIPTION_LINKS_JSON,
+                    ]
+
+            except Exception:
```
jaredlockhart: Maybe just JSONException here instead of all exceptions? 🤔

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -2316,6 +2329,80 @@ def _validate_primary_secondary_outcomes(self, data):
...
+            for field in (
+                "firefox_labs_title",
+                "firefox_labs_description",
+                "firefox_labs_group",
+            )
+            if not len((data.get(field) or "").strip())
+        }
+
+        if description_links := (
+            data.get("firefox_labs_description_links") or ""
```
jaredlockhart, line 2383: dict.get(key, fallback) you can use the fallback instead of `or`ing, not a huge difference

experimenter/experimenter/experiments/migrations/0288_nimbusexperiment_firefox_labs_description_links_text_field.py
```diff
@@ -0,0 +1,41 @@
...
+import json
+
+from django.db import migrations, models
+from django.db.models import Q
+
+
+def normalize_description_links(apps, schema_editor):
+    NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
+
+    experiments = list(NimbusExperiment.objects.filter(is_firefox_labs_opt_in=True))
```
jaredlockhart, line 11: Should this only be drafts?

experimenter/experimenter/experiments/migrations/0288_nimbusexperiment_firefox_labs_description_links_text_field.py
```diff
@@ -0,0 +1,41 @@
...
+    for experiment in experiments:
+        value = experiment.firefox_labs_description_links
+
+        if isinstance(value, str):
+            if value == '""':
+                value = None
```
jaredlockhart: Maybe instead of reusing the variable here, explicitly have variables for the old and new values?

experimenter/experimenter/experiments/models.py
```diff
@@ -392,7 +392,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         blank=True,
         null=True,
     )
-    firefox_labs_description_links = models.JSONField[dict[str, str]](
+    firefox_labs_description_links = models.TextField(
```
jaredlockhart, line 395: Wait what's the motivation for switching to textfield?  We wanna put JSON in it and validate that it's JSON but keep it as str?  We use JSONField elsewhere?

experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -677,7 +677,9 @@ def create(
             experiment=experiment, name="Control"
         )
         experiment.save()
-        NimbusBranchFactory.create(experiment=experiment, name="Treatment")
+
+        if not experiment.is_rollout:
+            NimbusBranchFactory.create(experiment=experiment, name="Treatment")
```
jaredlockhart, line 682: 😱  Oh good idea 👍

experimenter/experimenter/nimbus_ui/tests/test_forms.py
```diff
@@ -2287,6 +2287,9 @@ def test_branches_form_saves_firefox_labs_fields(self):
         self.assertTrue(experiment.is_firefox_labs_opt_in)
         self.assertEqual(experiment.firefox_labs_title, labs_title)
         self.assertEqual(experiment.firefox_labs_description, labs_description)
+        import pdb
+
+        pdb.set_trace()
```
jaredlockhart: Oops! 🙈

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -701,18 +700,27 @@ def errors(self):
             errors["branches"] = self.branches.errors
         return errors
 
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if not self.was_labs_opt_in and cleaned_data["is_firefox_labs_opt_in"]:
```
jaredlockhart, line 706: Ah okay cool, but also missing tests for this?

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -123,6 +124,18 @@ class TransitionConstants:
     }
 
 
+def is_valid_http_url(s: Any):
```
jaredlockhart, line 127: There's also this if you wanna use it:

https://github.com/django/django/blob/7063d31cc37028e0f945faa43030ce23dc1a5c23/django/core/validators.py#L130

jaredlockhart (changes requested):
Looking good!  Thnx I don't know how I missed adding the validation around this 🙈 

Some questions and thoughts but generally looks right 🙏

experimenter/experimenter/experiments/api/v5/serializers.py
freshstrangemusic: this handles the case where the key is set but is set to None (which happened in the tests because the field can be None)

experimenter/experimenter/experiments/models.py
freshstrangemusic: JSONField /must/ be valid JSON when saving. We're currently attempting to save arbitrary text to it and its becoming `json.dumps({ whatever })` instead of `{ whatever }`.

If we want users to be able to save their progress always, we must allow them to save invalid JSON.

experimenter/experimenter/nimbus_ui/forms.py
freshstrangemusic: This is covered by the tests you added. This is just a refactor to manipulate the form data instead of the experiment fields

experimenter/experimenter/experiments/migrations/0288_nimbusexperiment_firefox_labs_description_links_text_field.py
freshstrangemusic: The migration is thoroughly tested, i think its fine.

experimenter/experimenter/experiments/api/v5/serializers.py
freshstrangemusic: Ive restructured this so only json.loads is in the try.

experimenter/experimenter/nimbus_ui/forms.py
freshstrangemusic: specifically https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/nimbus_ui/tests/test_forms.py#L2296-L2396

experimenter/experimenter/experiments/api/v5/serializers.py
jaredlockhart: Oh yeah I was thinking that, it's funny cuz irl it will always be a string when it comes back from the request, None only occurs in the artificial test case, but yeah same same

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -2322,6 +2335,82 @@ def _validate_primary_secondary_outcomes(self, data):
...
+            )
+            if not len((data.get(field) or "").strip())
+        }
+
+        if description_links := (
+            data.get("firefox_labs_description_links") or ""
+        ).strip():
+            try:
+                description_links_obj = json.loads(description_links)
+            except Exception:
```
jaredlockhart, line 2387: Oh yeah I just meant, if this throws for some other random reason other than JSON parsing we'd want that to bail and throw to sentry.

freshstrangemusic: The only reason it can now throw is because of a JSON error (TypeError) -- what else are we gonna hit?

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -123,6 +124,18 @@ class TransitionConstants:
...
 
 
+def is_valid_http_url(s: Any):
+    if not isinstance(s, str):
+        return False
+
+    try:
+        url = urlparse(s)
+    except Exception:  # pragma: nocover
+        return False
```
freshstrangemusic: The docs say this can throw if the netloc is invalid but I couldn't make it throw :/

jaredlockhart: Oh if you want to force it to throw to handle the case you can just mock.patch

Issue #13318, opened 2025-08-21
Intermittent failures for labs experiments created without a group specified


freshstrangemusic:
The `firefox_labs_group` arg randomly picks from the available choices, but some of those choices have higher version restrictions than others as of #13314.

PR #13333, opened 2025-08-21
bug(nimbus): Require Firefox Labs fields in factory

freshstrangemusic:
Because:

- we now have different Firefox Labs groups based on version;
- these fields are conditionally required only for Firefox Labs fields;
- we don't want to tie the NimbusExperiment factory too tightly to the
  NimbusReviewSerializer logic; and
- the validity of some fields depends on the version of the experiment

this commit:

- makes the `firefox_labs_title`, `firefox_labs_description`, and
  `firefox_labs_group` required fields when `is_firefox_labs=True` is
  passed to `NimbusExperimentFactory.create()` et al.

Fixes https://github.com/mozilla/experimenter/issues/13318

### DIFF
--- a/experimenter/experimenter/experiments/tests/factories.py
+++ b/experimenter/experimenter/experiments/tests/factories.py
@@ -491,20 +491,7 @@ class NimbusExperimentFactory(factory.django.DjangoModelFactory):
         lambda o: random.choice(list(NimbusExperiment.QAStatus)).value
     )
     is_firefox_labs_opt_in = factory.LazyAttribute(lambda o: False)
-    firefox_labs_title = factory.LazyAttribute(
-        lambda o: faker.catch_phrase() if o.is_firefox_labs_opt_in else None
-    )
-    firefox_labs_description = factory.LazyAttribute(
-        lambda o: faker.catch_phrase() if o.is_firefox_labs_opt_in else None
-    )
     firefox_labs_description_links = factory.LazyAttribute(lambda o: "null")
-    firefox_labs_group = factory.LazyAttribute(
-        lambda o: (
-            random.choice(NimbusExperiment.FirefoxLabsGroups.choices)[0]
-            if o.is_firefox_labs_opt_in
-            else None
-        )
-    )
     requires_restart = factory.LazyAttribute(
         lambda o: (random.choice([True, False]) if o.is_firefox_labs_opt_in else False)
     )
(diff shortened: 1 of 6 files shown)


jaredlockhart:
Here we locked the group in the tests so they pass consistently:

https://github.com/mozilla/experimenter/pull/13332

I'm not sure we want to bake logic into the factories for them to pass specific validation criteria in the review serializer?  

In general I think we take a strategy of tests must specify their static preconditions to pass, and if they don't, the point of having the review serializer generating random values is to act as a bit of a fuzzer to stress test edge cases.

Another thought is that this somewhat ties the logic of the factory to the specific behaviour of the review serializer, such that if we change the review serializer in the future we may also have to change this logic, when in general I would imagine the factories should be quite agnostic to that behaviour?

experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -424,6 +424,18 @@ def random_slice(lst, *, min_items=0, max_items=2):
...
+    def factory(experiment):
+        if experiment.is_firefox_labs_opt_in:
+            raise Exception(
+                f"The field {field} is required when is_firefox_labs_opt_in is True"
+            )
+
+        return
+
+    return factory
+
```
jaredlockhart: Oh we already have other checks like this at teh top of the `create` method that raise FactoryError, I'd probably put a check like this there so they're all in the same place and raise the same kind of exception 🙏

experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -648,15 +647,26 @@ def create(
...
+            for field in (
+                "firefox_labs_title",
+                "firefox_labs_description",
+                "firefox_labs_group",
+            ):
+                if kwargs.get(field) is None:
+                    raise factory.FactoryError(
+                        f"The field {field} is required when is_firefox_labs_opt_in=True"
+                    )
+
```
jaredlockhart, line 655: Yeah exactly 👍

experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -424,6 +424,18 @@ def random_slice(lst, *, min_items=0, max_items=2):
...
+def required_for_firefox_labs(field):
+    def factory(experiment):
+        if experiment.is_firefox_labs_opt_in:
+            raise Exception(
+                f"The field {field} is required when is_firefox_labs_opt_in is True"
+            )
+
+        return
+
+    return factory
```
jaredlockhart: Prolly don't need this then?

jaredlockhart (approved):
Yep looks great thnx @freshstrangemusic 🎉 🎉 🎉 🎉

Issue #13098, opened 2025-07-25
Randomize the cirrus remote settings sync time 


data-sync-user:
Because currently all cirrus containers can sync at the same time and put stress while serving the request, we should randomize the sync time from 1 to 20 minutes for each run

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5749)

data-sync-user:
➤ Daniel Thorn commented:

prod currently syncs every 180 seconds, and non-prod every 100 seconds, so i went with a jitter of 60 seconds in prod and 10 seconds for non-prod

Issue #11487, opened 2024-10-07
cirrus: IntervalTrigger.__init__() got an unexpected keyword argument 'max_retries'


freshstrangemusic:
Source location: https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/cirrus/main.py#L291

It does not seem like this kwarg is supported:

https://apscheduler.readthedocs.io/en/3.x/modules/schedulers/base.html#apscheduler.schedulers.base.BaseScheduler.add_job

https://apscheduler.readthedocs.io/en/3.x/modules/triggers/interval.html#apscheduler.triggers.interval.IntervalTrigger

and we mock this out in unit tests so that all of these pass, when they should fail ue to this exception:

https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/tests/test_main.py#L183-L189
https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/tests/test_main.py#L205-L211
https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/tests/test_main.py#L230-L236
https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/tests/test_main.py#L257-L263
https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/tests/test_main.py#L284-L290
https://github.com/mozilla/experimenter/blob/9f01a5da0be16da80fb184e51eeb9c463eaf07c0/cirrus/server/tests/test_main.py#L307-L313

We should have *at least one test* that actually calls the function to make sure it works.

Sentry issue: [CIRRUS-PROD-V](https://mozilla.sentry.io/issues/5640026987/?referrer=github_integration)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4856)

PR #13346, opened 2025-08-22
chore(cirrus): add jitter to remote settings sync time

relud:
Because

- cirrus instances that were all started at the same time shouldn't continuously sync remote settings at the same time, as that could negatively impact the availability of the service.

This commit 

- Adds a setting `CIRRUS_REMOTE_SETTING_REFRESH_RATE_IN_SECONDS` that randomizes the timing of remote settings sync jobs by up to that many seconds.

For #13098


cirrus/server/cirrus/main.py
```diff
@@ -133,6 +134,7 @@ def start_and_set_initial_job():
         fetch_schedule_recipes,
         "interval",
         seconds=remote_setting_refresh_rate_in_seconds,
+        jitter=remote_setting_refresh_jitter_in_seconds,
```
jaredlockhart: Oh it just has a jitter parameter eh? God that's handy.

jaredlockhart: Do we not have tests that mock and assert on the call parameters? 🤔

jaredlockhart (approved):
Well that's handy!  Looks good thnx @relud 🙏

yashikakhurana (changes requested):
Thank you @relud easy solution, can you also update the docs to reflect the same please

cirrus/server/cirrus/main.py
relud: not yet, but I'm about to post a fix for https://github.com/mozilla/experimenter/issues/11487 which will add that

jaredlockhart:
Oh good point @yashikakhurana !

yashikakhurana (approved):
perfect, looks good thank you @relud

PR #13365, opened 2025-08-26
feat(nimbus): sort by merged channel and channels in list views

### LINKED ISSUE #13357: Sort by channels in list views
We added the ability to display and filter by channel/channels on the list and detail views, but we missed sorting by channel/channels in the table view (both the current landing page and the new home page).  We’ll need to combine channel and channels into a single field using Django annotate and then sort on that field to keep it consistent across applications.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5875)

jaredlockhart:
Becuase

* We now have two fields for storing the channel(s) of an experiment
* To be able to sort by channel we need to combine those into a single field that can be sorted on

This commit

* Adds a query annotation called merged_channel that combines channel and channels
* Updates the landing and new home page sort fields to sort on that field
* Updates tests

fixes #13357

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -18,7 +18,7 @@ from django.core.files.base import ContentFile
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import MaxValueValidator
 from django.db import models
-from django.db.models import Count, F, Q, QuerySet
+from django.db.models import Case, Count, F, Q, QuerySet, When
 from django.db.models.constraints import UniqueConstraint
 from django.urls import reverse
 from django.utils import timezone
@@ -131,6 +131,15 @@ class NimbusExperimentManager(models.Manager["NimbusExperiment"]):
         return self.for_collection(
             self.filter(NimbusExperiment.Filters.IS_ENDING, application__in=applications),
             collection,
+        )
+
+    def with_merged_channel(self):
+        return self.get_queryset().annotate(
+            merged_channel=Case(
+                When(Q(channel__isnull=False) & ~Q(channel=""), then=F("channel")),
+                default=F("channels__0"),
+                output_field=models.CharField(),
+            )
         )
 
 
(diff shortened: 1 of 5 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -133,6 +133,15 @@ def waiting_to_end_queue(self, applications, collection):
...
 
+    def with_merged_channel(self):
+        return self.get_queryset().annotate(
+            merged_channel=Case(
+                When(Q(channel__isnull=False) & ~Q(channel=""), then=F("channel")),
+                default=F("channels__0"),
+                output_field=models.CharField(),
+            )
+        )
+
```
yashikakhurana, line 145: 🎉

experimenter/experimenter/nimbus_ui/filtersets.py
```diff
@@ -328,8 +328,8 @@ class HomeSortChoices(models.TextChoices):
     APPLICATION_DOWN = "-application", "Application"
     TYPE_UP = "is_rollout", "Type"
     TYPE_DOWN = "-is_rollout", "Type"
-    CHANNEL_UP = "channel", "Channel"
-    CHANNEL_DOWN = "-channel", "Channel"
+    CHANNEL_UP = "merged_channel", "Channel"
+    CHANNEL_DOWN = "-merged_channel", "Channel"
     SIZE_UP = "population_percent", "Size"
     SIZE_DOWN = "-population_percent", "Size"
```
yashikakhurana, line 334: thank you for update for the home sort too

experimenter/experimenter/nimbus_ui/filtersets.py
```diff
@@ -406,7 +406,7 @@ def filter_my_deliveries(self, queryset, name, value):
                 return queryset  # Default = All Deliveries
 
     def filter_sort(self, queryset, name, value):
-        return queryset.order_by(value)
+        return queryset.order_by(value, "slug")
```
yashikakhurana, line 409: why are we adding slug?

jaredlockhart: Yeah it's becuase if you have two experiments with the same channel and that's teh only thing you're sorting on then there's nothing guaranteeing the order of those two, but now in the tests we have multiple experiments with the same channel and we do want to assert on the specific order, so we need a second field to gaurantee that order, so we can use slug.

PR #13377, opened 2025-08-27
chore(nimbus): prevent wrong channel(s) from being set in NimbusExperimentFactory

### LINKED ISSUE #13224: Remove temporary single channel case for desktop
While we are migrating desktop over to multiple channels we’ll need to support both the single and multiple channel cases for experiments that are in flight.  Once we have moved all desktop experiments over to the multi channel case, we can drop the single channel case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5789)

### LINKED ISSUE #13240: Ensure only channel or channels are set in factories/tests
We’re introducing a new constraint that desktop experiments can only set the `channels` field and non desktop experiments can only set the `channel` field.  We’ll have to update the factories and a lot of tests to match this new constraint.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5794)

jaredlockhart:
Becuase
    
* Now that we have a channels field for desktop and still use the channel field for everything else
* We should enforce in the NimbusExperimentFactory that channels is only used for desktop and channel is only used for non-desktop
* This forces test setups to be explicit and correct about which field is being used
* Which can that expose potential downstream bugs or incorrect logic around handling channels
    
This commit
    
* Prevents using channels for non desktop experiments
* Prevents using channel for desktop experiments
* Updates tests as necessary
    
fixes #13240
fixes #13224


experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -657,6 +661,16 @@ def create(
                 "new ones."
             )
 
+        if experiment.is_desktop and experiment.channel:
+            raise factory.FactoryError(
+                "A desktop experiment must not use channel.  Use channels instead."
+            )
+
+        if not experiment.is_desktop and experiment.channels:
+            raise factory.FactoryError(
```
yashikakhurana, line 668: thank you for fixing the factories too, it will be helpful, I noticed the same things while I was working on the channels filter

yashikakhurana (approved):
Thank you @jaredlockhart this will keep tests intact

experimenter/experimenter/experiments/tests/factories.py
freshstrangemusic, line 668: You'll have to rebase this off my change for factories regardless, but IMO we should do these checks before constructing the NimbusExperiment.

jaredlockhart: Actually, I have two thoughts on this:

1) Regardless of whether it happens before or after it's inserted into the db it'll still throw and fail the test which is the point, so I'm not sure it makes a difference

2) Won't only fields that are explicitly passed to the method appear in kwargs, but not the fields that are set by the actual factory, so if you want to reason about all those fields combined, is there a way to do it before calling super().create()?

freshstrangemusic: > Won't only fields that are explicitly passed to the method appear in kwargs, but not the fields that are set by the actual factory, so if you want to reason about all those fields combined, is there a way to do it before calling super().create()?

This is literally about forbidding the use of a kwarg so we should just do that.

But it doesnt matter either way.

PR #13447, opened 2025-09-04
feat(nimbus): Filter by feature

### LINKED ISSUE #13424: My stuff*-Add filter for feature name
we should be able to filterby feature name
this could be harder to display with multi-feature experiments… so plan on it taking more room.
BONUS POINTS:  Limit the filter to only features that are in my stuff if possible - not the whole list.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5717)

yashikakhurana:
Because

- We want to allow users to filter by feature on my deliveries section

This commit

- Shows by default all the features if you no experiment subscribed or owned and which is not archieved
- If you have deliveries, then only show filter for the feature which are linked to the deliveries instead of showing it all

Fixes #13424

### DIFF
--- a/experimenter/experimenter/nimbus_ui/filtersets.py
+++ b/experimenter/experimenter/nimbus_ui/filtersets.py
@@ -416,6 +416,14 @@ class NimbusExperimentsHomeFilter(django_filters.FilterSet):
             attrs={"title": "All Applications"},
         ),
     )
+    feature_configs = django_filters.ModelMultipleChoiceFilter(
+        method="filter_feature_configs",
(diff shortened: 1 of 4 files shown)


yashikakhurana:
Screenshot 2025-09-04 at 10 14 01 AM
Screenshot 2025-09-04 at 10 13 55 AM

b4handjr (approved):
Are you not including web experiments for a specific reason?

yashikakhurana:
> Are you not including web experiments for a specific reason?
 this should include web experiments

PR #13454, opened 2025-09-08
bug(nimbus): run local fml ci check on change to files in experimeter/

### LINKED ISSUE #13453: Local FML CI check needs to run on changes to files in experimenter/
dependencies were updated and it slightly changed the yaml format for generated fml, and it was merged because the check didn’t run due to no changes to the listed relevant paths, which caused CI to start failing on unrelated PRs.

the check should run on changes to files like experimenter/pyproject.toml and experimenter/Dockerfile and experimenter/poetry.lock

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-5918)

relud:
Because

- Dependency updates and config files can change generated FML

This commit

- Adds `^experimenter/[^/]+$` to the check file paths for running check_local_feature_manifests

Fixes #13453

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,7 +240,7 @@ jobs:
     steps:
       - checkout
       - check_file_paths:
-          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/"
+          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/|^experimenter/[^/]+$|application-services/"
       - run:
           name: Check local feature manifests
           command: |


.circleci/config.yml
```diff
@@ -240,7 +240,7 @@ jobs:
     steps:
       - checkout
       - check_file_paths:
-          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/"
+          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/|^experimenter/[^/]+$"
```
freshstrangemusic: swap this to `application-services/` because its a change there that caused this.

relud: I'll add application-services. I think we still also need to be including files experimenter/, so i intend to leave that in

.circleci/config.yml
```diff
@@ -240,7 +240,7 @@ jobs:
     steps:
       - checkout
       - check_file_paths:
-          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/"
+          paths: "experimenter/experimenter/features/manifests/|experimenter/manifesttool/|^experimenter/[^/]+$|application-services/"
```
freshstrangemusic, line 243: if we're keeping experimenter/ cant this just be `experimenter/|application-services/` ?

relud: we don't need all of experimenter, this is only adding files directly in experimenter, not any files in any subdirectories

I think it's worth the complexity here because experimenter/ changes much more often than this subset, but if you think including all of experimenter is better i'm fine with that too

relud: if you want all of experimenter, i'll do that in a follow-up PR

PR #13486, opened 2025-09-11
feat(nimbus): add clear visual indication of future states while organizing experiment

### LINKED ISSUE #13135: Add clearer indication of past, current, future states
We currently show the timeline like this:

Image

But users are getting confused about what the current state is, so let's add some additional visual cues to show that the blue state is the current state, states to the left are completed, and states to the right have not yet occurred.  Maybe something like

Image

Or maybe checkmarks on the completed ones, and greyed out or italics on the not yet completed ones?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5760)

moibra05:
Because

- Users found it unclear about the current state while managing experiments

This commit

- Adds visual indication for future states

Fixes #13135

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1192,7 +1192,6 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         for item in timeline:
             if item["is_active"]:
                 return item["step"]
-        return NimbusUIConstants.EXPERIMENT_ORDERING["Draft"]
 
     @property
     def should_end(self):


moibra05:
image
image
image

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -287,12 +287,20 @@ def get_context_data(self, **kwargs):
...
         )
         context["qa_edit_mode"] = self.request.GET.get("edit_qa_status") == "true"
         context["takeaways_edit_mode"] = self.request.GET.get("edit_takeaways") == "true"
+        context["experiment_order"] = NimbusUIConstants.EXPERIMENT_ORDERING
+
+        for item in experiment.timeline():
+            if item["is_active"]:
+                context["experiment_step"] = item["step"]
+                break
+
```
yashikakhurana: Instead of doing it here, add that as a property in a model that returns the active timeline status. You can name it like` experiment_active_status` and use that in the template directly

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -293,6 +293,7 @@ def get_context_data(self, **kwargs):
         )
         context["qa_edit_mode"] = self.request.GET.get("edit_qa_status") == "true"
         context["takeaways_edit_mode"] = self.request.GET.get("edit_takeaways") == "true"
+
```
yashikakhurana: can you remove this extra line please 🙏

experimenter/experimenter/experiments/models.py
```diff
@@ -1184,6 +1190,14 @@ def timeline(self):
 
         return timeline_entries
 
+    @property
+    def experiment_active_status(self):
+        timeline = self.timeline()
+        for item in timeline:
+            if item["is_active"]:
+                return item["step"]
+        return None
```
yashikakhurana: so they will always be a state where its draft/review etc etc, we can skip return None, or can add return as draft

yashikakhurana:
@moibra05 you would need to add a test, that's why check is failing

moibra05:
> @moibra05 you would need to add a test, that's why check is failing

ah, what would I write a test for and where would it go?

yashikakhurana:
> > @moibra05 you would need to add a test, that's why check is failing
> 
> ah, what would I write a test for and where would it go?

first step check the output of make check command, it will tell you which line is missing coverage, and then add a test for that, we have all the tests in the test folder, and each one is mapped with the corresponding test file

moibra05:
> > > @moibra05 you would need to add a test, that's why check is failing
> > 
> > 
> > ah, what would I write a test for and where would it go?
> 
> first step check the output of make check command, it will tell you which line is missing coverage, and then add a test for that, we have all the tests in the test folder, and each one is mapped with the corresponding test file

figured it'd just be easier to remove the line causing the problem :)

yashikakhurana:
@moibra05 we have a test where it is testing the timeline, can you please add a test to see if its returning the correct value please 🙏

PR #13489, opened 2025-09-11
feat(nimbus): Add home page on Header

### LINKED ISSUE #13482: Add home page on the main header
Once QA is green, we should add a button for all deliveries on the header and that should link to the existing landing page, and add home page to renders by default, and home page icon should also appear to go to the home page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5938)

yashikakhurana:
Because

- We want to make home page the default and want to add the option to go to the current landing page i.e. All deliveries 

This commit

- Update the url `/`-> home page
- url `/all`-> all deliveries page (currently localhost/nimbus)
- Adjusting the space to be consistent between the cards

Fixes #13482 

Note: We will merge this once we get green from QA


yashikakhurana:
Screenshot 2025-09-11 at 1 31 14 PM

yashikakhurana:
oh I just realized we need to move create experiment on the header too, let me do that

moibra05:
image
image

experimenter/experimenter/urls.py
```diff
@@ -26,14 +29,9 @@
     re_path(r"^api/v8/", include("experimenter.experiments.api.v8.urls")),
     re_path(r"^admin/", admin.site.urls),
     re_path(r"^experiments/", include("experimenter.legacy.legacy_experiments.urls")),
-    re_path(r"^nimbus/", include("experimenter.nimbus_ui.urls")),
-    re_path(r"^nimbus/$", NimbusExperimentsListView.as_view(), name="nimbus-list"),
+    re_path(r"^all/", include("experimenter.nimbus_ui.urls")),
+    re_path(r"^all/$", NimbusExperimentsListView.as_view(), name="nimbus-list"),
```
relud: i think it would be good to redirect /nimbus/* to /all/* so that old urls and bookmarks aren't broken for now

experimenter/experimenter/urls.py
```diff
@@ -26,14 +30,15 @@
     re_path(r"^api/v8/", include("experimenter.experiments.api.v8.urls")),
     re_path(r"^admin/", admin.site.urls),
     re_path(r"^experiments/", include("experimenter.legacy.legacy_experiments.urls")),
-    re_path(r"^nimbus/", include("experimenter.nimbus_ui.urls")),
-    re_path(r"^nimbus/$", NimbusExperimentsListView.as_view(), name="nimbus-list"),
-    re_path(r"^legacy/$", ExperimentListView.as_view(), name="home"),
+    re_path(r"^all/", include("experimenter.nimbus_ui.urls")),
```
jaredlockhart: Even if we want to move everything down to `/` we'd still want to keep `/nimbus/<slug>/` becuase there will be so many links like that that we should keep intact?

yashikakhurana: okay so we can do `/` to home page and keep rest as it is?

jaredlockhart: Hrmmm okay I tested locally, I see what you mean about the redirects, it does work fine, but I don't think `/all/<slug/` is quite right?  Usually it'd be more like 

* `/<type>/` for a list view
* `/<type>/<slug>/` for a detail view

In the past we would've used like `/experiments/<slug>/` like we did in legacy, but now it's probably more like `deliveries` since that's what we're calling them now?

jaredlockhart (changes requested):
Actually I think it's probably best if we move the url swap out into its own PR, that way we can land the layout/link changes here, and just do the URL swap tomorrow, and that way if we need to revert the URL swap we can do that quickly without also losing the rest of the changes 🙏

jaredlockhart (approved):
Very good 👍

PR #13504, opened 2025-09-15
feature(nimbus): Add deliveries table to feature health page.

### LINKED ISSUE #13485: Create a table that shows the Experiments, Rollouts and Firefox Lab usage of the feature
This table should have 6 columns. There should be a counter that shows the total amount of times this feature has been used in an experiment.

#### Table 3:  All Affiliated Nimbus Deliveries | [#] Total Deliveries [?] 
? info message: _This shows all Nimbus experiments, rollouts, Labs experiences, etc. associated with your selected feature._

|_Recipe Name_
_(Sort: A<>Z)_|_Launch Date_
_(Sort: Oldest <> Newest)_|_Type of Delivery_
_(Sort: A<>Z)_|_Channel(s)_
_(Sort: A<>Z)_|_Minimum Version_|_Population Size_
_(Sort: Low<>High)_|_Delivery Brief_|
|----|----|----|----|----|----|----|
|﹍[Recipe link]﹍  | |Experiment|N| |5,000,000|[﹍Brief link]﹍|
| | | | | | | |
| | | | | | | |

Default pagination will be 5.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5799)

b4handjr:
Because

- We want to display data about recent deliveries based on a feature/application combo on the feature health page

This commit

- Adds a deliveries table to the feature health page

Fixes #13485 #13610

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -1480,6 +1480,7 @@ class ApproveUpdateRolloutForm(UpdateStatusForm):
 
 class FeaturesForm(forms.ModelForm):
     application = forms.ChoiceField(
+        required=False,
         label="",
         choices=NimbusExperiment.Application.choices,
         widget=forms.widgets.Select(
@@ -1490,6 +1491,7 @@ class FeaturesForm(forms.ModelForm):
         initial=NimbusExperiment.Application.DESKTOP.value,
     )
     feature_configs = forms.ChoiceField(
+        required=True,
         label="",
         choices=[],
         widget=SingleSelectWidget(),
@@ -1497,14 +1499,18 @@ class FeaturesForm(forms.ModelForm):
     update_on_change_fields = ("application", "feature_configs")
 
     def get_feature_config_choices(self, application, qs):
-        return sorted(
-            [
-                (application.pk, f"{application.name} - {application.description}")
-                for application in NimbusFeatureConfig.objects.all()
-                if application in qs
-            ],
-            key=lambda choice: choice[1].lower(),
+        choices = [("", "Nothing Selected")]  # Add a default blank field.
+        choices.extend(
+            sorted(
+                [
+                    (application.pk, f"{application.name} - {application.description}")
+                    for application in NimbusFeatureConfig.objects.all()
+                    if application in qs
+                ],
+                key=lambda choice: choice[1].lower(),
+            )
         )
+        return choices
 
     class Meta:
         model = NimbusFeatureConfig
@@ -1533,4 +1539,9 @@ class FeaturesForm(forms.ModelForm):
             "hx-swap": "outerHTML",
         }
         self.fields["application"].widget.attrs.update(htmx_attrs)
+        htmx_attrs.update(
+            {
+                "hx-select-oob": "#deliveries-table",
+            }
+        )
         self.fields["feature_configs"].widget.attrs.update(htmx_attrs)
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -1491,7 +1491,6 @@ class FeaturesForm(forms.ModelForm):
         initial=NimbusExperiment.Application.DESKTOP.value,
     )
     feature_configs = forms.ChoiceField(
-        required=True,
         label="",
         choices=[],
         widget=SingleSelectWidget(),
@@ -1499,7 +1498,7 @@ class FeaturesForm(forms.ModelForm):
     update_on_change_fields = ("application", "feature_configs")
 
     def get_feature_config_choices(self, application, qs):
-        choices = [("", "Nothing Selected")]  # Add a default blank field.
+        choices = []  # Add a default blank field.
         choices.extend(
             sorted(
                 [
@@ -1537,11 +1536,7 @@ class FeaturesForm(forms.ModelForm):
             "hx-select": "#features-form",
             "hx-target": "#features-form",
             "hx-swap": "outerHTML",
+            "hx-select-oob": "#deliveries-table",
         }
         self.fields["application"].widget.attrs.update(htmx_attrs)
-        htmx_attrs.update(
-            {
-                "hx-select-oob": "#deliveries-table",
-            }
-        )
         self.fields["feature_configs"].widget.attrs.update(htmx_attrs)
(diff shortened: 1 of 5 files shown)


b4handjr:
View with features
image

b4handjr:
View no features
image

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -30,11 +30,47 @@
...
+                <tr scope="row">
+                  <td id="delivery-name" class="p-2 border-1">
+                    <a target="_blank" href="{{ experiment.experiment_url }}">{{ experiment.name }}</a>
+                  </td>
+                  <td id="delivery-published-date" class="p-2 border-1">{{ experiment.published_date|format_not_set }}</td>
+                  {% if experiment.is_rollout %}
+                    <td id="delivery-type-rollout" class="p-2 border-1">Rollout</td>
+                  {% else %}
+                    <td id="delivery-type-experiment" class="p-2 border-1">Experiment</td>
+                  {% endif %}
```
yashikakhurana: we have this now and we can re use this https://github.com/mozilla/experimenter/blob/88faa237d441eedfb3a2ad8562a5eeb6c46f7509/experimenter/experimenter/experiments/models.py#L1539

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -30,11 +30,47 @@
...
+                  <td id="delivery-channel" class="p-2 border-1">{{ experiment.get_channel_display }}</td>
+                  <td id="delivery-min-version" class="p-2 border-1">{{ experiment.firefox_min_version|parse_version }}</td>
+                  <td id="delivery-population-percent" class="p-2 border-1">{{ experiment.population_percent|floatformat:"-1" }}</td>
+                  <td id="delivery-brief" class="p-2 border-1">{{ experiment.delivery_brief|format_not_set }}</td>
```
yashikakhurana: also I see that you are applying the same class `p-2 border 1` if we move that class up then we wont need to define for each column

b4handjr: So you just want the row defined with a border?

yashikakhurana: I believe if you will move to the row that css, this should apply to all

b4handjr: Like this? `<tr scope="row" class="p-2 border-1">`, this seems to just do the row, but not each column in that row.

yashikakhurana: I think we can skip the borders, if you will notice the other table, we don't use the borders, I would suggest to refer to home page table to be consistent with the design between the pages

yashikakhurana: Our plan is to use this table on the landing page as well in the future

yashikakhurana (changes requested):
I think we need to remove borders, it is showing 2 borders now, lets say consistent with the design of tables with the home page. Also noticed 2 bugs
1: Look how it is showing the selected feature, if I select that feature again it is not showing any deliveries

https://github.com/user-attachments/assets/7f9610ea-f93a-4b4f-a508-bceaab0fc7a1

2: If you select the different application, deliveries data is not getting changed, it is still showing the old data

https://github.com/user-attachments/assets/bf46e67c-f4f3-4050-9776-6938aad6bcc1

b4handjr:
How does this look
image

I also fixed the mix of deliveries that was showing up when you first navigated to the page

yashikakhurana:
@b4handjr It is looking really amazing, some minor suggestions would be like to add the same kind of background where we are choosing application and features too

b4handjr:
Done
image

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -1464,21 +1465,26 @@ class FeaturesForm(forms.ModelForm):
...
     def get_feature_config_choices(self, application, qs):
-        return sorted(
-            [
-                (application.pk, f"{application.name} - {application.description}")
-                for application in NimbusFeatureConfig.objects.all()
-                if application in qs
-            ],
-            key=lambda choice: choice[1].lower(),
+        choices = [("", "Nothing Selected")]  # Add a default blank field.
+        choices.extend(
```
yashikakhurana: I think you can skip this setting manually `Nothing selected` option, you can refer to locales, countries and channels etc fields they show nothing selected option by default

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -17,24 +17,66 @@
...
+                         class="text-decoration-none fw-medium">{{ experiment.name }}</a>
+                    </td>
+                    <td id="delivery-published-date">{{ experiment.published_date|format_not_set }}</td>
+                    <td id="delivery-type-{{ experiment.home_type_choice }}">{{ experiment.home_type_choice }}</td>
+                    <td id="delivery-channel">{{ experiment.get_channel_display }}</td>
+                    <td id="delivery-min-version">{{ experiment.firefox_min_version|parse_version }}</td>
+                    <td id="delivery-population-percent">{{ experiment.population_percent|floatformat:"-1" }}</td>
+                    <td id="delivery-brief">{{ experiment.delivery_brief|format_not_set }}</td>
+                  {% endfor %}
+                </tr>
```
yashikakhurana, line 81: Here can you add a check if no experiment is found matching the desktop and feature configuration to show No deliveries found message

yashikakhurana (changes requested):
Thank you @b4handjr it is all coming up together, UI is looking really great, found 2 issues, first if I select the application and feature which have experiments available, it shows in the table, but if I change the application it is not updating the deliveries table, you would need to use `htmx-target` and such as to achieve that, second thing if we dont have experiments available having a message will help such as `No deliveries found` 

https://github.com/user-attachments/assets/abb3b75f-845b-494b-b853-606b7415a8e2

b4handjr:
Added "No Deliveries" text, I added some padding so it stands out more
image

yashikakhurana (changes requested):
ok

okay tested locally and it seems like we cant set ultil we manually set Nothing selected, @b4handjr can you add back nothing selected option and the bug that I see that because it is already selected and not showing the table accordingly until I select it again https://github.com/user-attachments/assets/49b931cc-be8c-4b28-b72c-86720edd489f
option 1- add nothing selected manually
option 2- dont even select the application, then it will not show the features default selected

yashikakhurana (approved):
perfect, thank you @b4handjr

PR #13508, opened 2025-09-16
feat(nimbus): Tags model

### LINKED ISSUE #13503: Add model for Tags
To support the tagging, 

* Add `Tag` model with fields: `name`, `color`
* Add many-to-many relation between `Tag` and `NimbusExperiment`
* Only allow unique name of tags

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5944)

yashikakhurana:
Because

- We want to allow users to tag deliveries

This commit

- Adds the new tag model with a many-to-many relationship with the experiment so that users can select multiple tags

Fixes #13503


experimenter/experimenter/experiments/models.py
```diff
@@ -430,6 +438,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         blank=True,
         default=list,
     )
+    tags = models.ManyToManyField(Tag, blank=True, related_name="tags")
```
jaredlockhart: Oh I don't think this is the right `related_name`.  That would be the field that appears on a tag instance to point back to experiments, I think you'd want something like `related_name="experiments"` here?

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -3209,6 +3210,15 @@ def test_can_archive(self, expected_can_archive, lifecycle):
         experiment = NimbusExperimentFactory.create_with_lifecycle(lifecycle)
         self.assertEqual(experiment.can_archive, expected_can_archive)
 
+    def test_tag_experiment_relationship(self):
+        tag = Tag.objects.create(name="TestTag", color="#123456")
+        experiment = NimbusExperimentFactory.create()
+        experiment.tags.add(tag)
+        self.assertIn(tag, experiment.tags.all())
+        self.assertIn(experiment, tag.tags.all())
```
jaredlockhart: Yeah see, I would expect this to be `tag.experiments.all()`?

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -3209,6 +3210,15 @@ def test_can_archive(self, expected_can_archive, lifecycle):
...
+        self.assertEqual(experiment.tags.count(), 1)
+        self.assertEqual(str(tag), str(experiment.tags.first()))
```
jaredlockhart: This is mostly testing Django internals, which are already covered by Djangos tests.  We'll have more tests when we start adding the forms and stuff around it, so I'm not sure this is needed here?

jaredlockhart (dismissed):
Looking good!

experimenter/experimenter/experiments/tests/test_models.py
yashikakhurana: added this because was getting str method is not covered in test coverage, may be when we will have more test we can remove this one

jaredlockhart: Oh coverage for the string method right, kk not a big deal to have it 👍

Issue #13379, opened 2025-08-27
The Include logic does not work as users expect/need (AND vs OR)


data-sync-user:
The understanding of how the include/exclude logic should work is different from how it actually works (at least for Include).

If you list 2 experiments in the include (or exclude) boxes… you want to filter for users that meet ANY of the experiments listed.  So believe the wanted logic is OR

Currently the logic is for ALL of the experiments listed.  

Changing to ANY wouldn’t change how the Exclude works, but would change how the include works.

* INCLUDE:  the logic would now be if you are in ANY of experiment 1 OR experiment 2 - you meet the targeting.
* EXCLUDE:  ALL currently works the same way ANY will... because excluding ALL says you need to be in NONE of the experiments listed.  So if you were in any of the experiments listed you would fail to meet the targeting. 

We should also update the Experimenter web page’s descriptions of the include/exclude options, changing:

{quote}Exclude users enrolled in 【 highlighted #ff5630 】**_all_**【 end highlighted 】 of these experiments/rollouts (past or present)

Require users to be enrolled in 【 highlighted #ff5630 】**_all_**【 end highlighted 】 of these experiments/rollouts (past or present){quote}

to:

{quote}Exclude users enrolled in 【 highlighted #36b37e 】**_any_**【 end highlighted 】 of these experiments/rollouts (past or present)

Require users to be enrolled in 【 highlighted #36b37e 】**_any_**【 end highlighted 】 of these experiments/rollouts (past or present){quote}

[https://mozilla.slack.com/archives/CF94YGE03/p1755121541516719](https://mozilla.slack.com/archives/CF94YGE03/p1755121541516719|smart-link)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-5820)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/92496">image-20250813-230831.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/92495">image-20250813-230945.png</a>

data-sync-user:
➤ Shell Escalante commented:

Looking at exclude logic for

!image-20250813-230831.png|width=730,height=72,alt="image-20250813-230831.png"!

&& (('1-callout-contextual-chatbot-suggestion' in enrollments) == false) && (('1-click-pin-during-onboarding' in enrollments) == false) && ((enrollmentsMap['1234'] == 'treatment-a') == false)

include logic for same

[...]

PR #13522, opened 2025-09-17
Update web page description of Include and Exclude targeting

cpeterso:
Because

- the web page description of the Include and Exclude logic is confusing and now out of date after #13379 was fixed.

This commit

- makes the web page description of the Include and Exclude logic match the implementation: Nimbus includes or excludes users who were enrolled in ANY of the listed experiments, not ALL of them.

This fix is follow-up work for #13379.


jaredlockhart (approved):
Oh yes thank you @cpeterso I kept meaning to do this and forgetting 🙏

PR #13572, opened 2025-09-23
feat(nimbus): Add advanced targeting for TOU experiences in users who…

### LINKED ISSUE #13571: [Targeting] Targeting for Three Different TOU Experiences
### What is the criteria you need to target based on?

There will be three possible TOU experiences defined, based on the number of criteria users meet from a defined list. See [this section of the PRD](https://docs.google.com/document/d/16ODO3B0_lx7eq665gFQBaEv3j1CjKElTVbyad_iqQK4/edit?tab=t.0#heading=h.ecl5dqo18xu9) for details.

### Should the targeting include or exclude users?

It will include users based on the criteria.

### Additional Notes (optional)

_No response_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5972)

lmegviar:
… have not accepted TOU

Because

- We wish to show different TOU experiences to users based on a list of set criteria (see linked issue for more details)

This commit

- Adds three new targeting constants for the three TOU experiences
- Makes existing TOU related variables more robust across regions by checking to see if system prefs are enabled before checking the related non-system pref is truthy or falsey.
- Removes outdated variables/TOU version references that no longer apply now that we've migrated to new TOU preference names.

Fixes #13571

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -3208,8 +3208,7 @@ TOU_EXPERIENCE_0 = NimbusTargetingConfig(
     name="TOU Experience 0",
     slug="tou_experience_0",
     description=(
-        "User has not accepted TOU V4 or higher",
-        "and should see TOU experience 0",
+        "User has not accepted TOU V4 or higher and should see TOU experience 0"
     ),
     targeting=f"""
     (
@@ -3230,8 +3229,7 @@ TOU_EXPERIENCE_1 = NimbusTargetingConfig(
     name="TOU Experience 1",
     slug="tou_experience_1",
     description=(
-        "User has not accepted TOU V4 or higher",
-        "and should see TOU experience 1",
+        "User has not accepted TOU V4 or higher and should see TOU experience 1"
     ),
     targeting=f"""
     (
@@ -3252,8 +3250,7 @@ TOU_EXPERIENCE_2 = NimbusTargetingConfig(
     name="TOU Experience 2",
     slug="tou_experience_2",
     description=(
-        "User has not accepted TOU V4 or higher",
-        "and should see TOU experience 2",
+        "User has not accepted TOU V4 or higher and should see TOU experience 2"
     ),
     targeting=f"""
     (


experimenter/experimenter/targeting/constants.py
```diff
@@ -102,16 +56,28 @@ def __post_init__(self):
 HOMEPAGE_NOT_DEFAULT = "!homePageSettings.isDefault"
 TOPSITES_OR_SPONSORED_TOPSITES_DISABLED = """
 (
-    !'browser.newtabpage.activity-stream.feeds.topsites'|preferenceValue
+    (
+        'browser.newtabpage.activity-stream.feeds.system.topsites'|preferenceValue
```
lmegviar, line 60: Ensure this feature is actually enabled for the user's region before checking to see if it's enabled or disabled in the non-system pref.

experimenter/experimenter/targeting/constants.py
```diff
@@ -3104,51 +3126,6 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
 
-TOU_NOT_ACCEPTED_EXISTING_USER_ADS_ENABLED_MAC_OR_WIN = NimbusTargetingConfig(
```
lmegviar, line 3107: The only experiments that used this targeting have ended or are no longer enrolling. We will not use this targeting in the future in favor of the newly added experience constants.

jaredlockhart:
Okay I cleaned up the autoformatting so we'll see what CI says and then we can land this.  Thnx for confirming about past experiments no longer depending on the existing implementations.

jaredlockhart:
Okay lemme see if I can clear up the test failure

jaredlockhart:
Ah I see what happened

```py
     description=(
-        "User has not accepted TOU V4 or higher",
-        "and should see TOU experience 0",
+        "User has not accepted TOU V4 or higher and should see TOU experience 0"
     ),
     targeting=f"""
     (
@@ -3230,8 +3229,7 @@ TOU_EXPERIENCE_1 = NimbusTargetingConfig(
     name="TOU Experience 1",
     slug="tou_experience_1",
     description=(
-        "User has not accepted TOU V4 or higher",
-        "and should see TOU experience 1",
+        "User has not accepted TOU V4 or higher and should see TOU experience 1"
     ),
     targeting=f"""
     (
@@ -3252,8 +3250,7 @@ TOU_EXPERIENCE_2 = NimbusTargetingConfig(
     name="TOU Experience 2",
     slug="tou_experience_2",
     description=(
-        "User has not accepted TOU V4 or higher",
-        "and should see TOU experience 2",
+        "User has not accepted TOU V4 or higher and should see TOU experience 2"
     ),
     targ
```

Putting commas inside the multiline string like that turns it from a string into a tuple which is a quirky Python quirk that's easy to do by accident and hard to find.

The multiline Python syntax is

```py
something = (
  "This"
  "That"
)
```

with no commas, and Python will join it into a single string.

Very close is

```py
something = (
  "This",
  "That",
)
```

which will become a tuple, and because Python is so dynamically typed it'll just happily be a tuple all over the place.

Maybe we should use a dataclass there that has stricter rules around the types of the fields 🤔 

Anyways I cleared it up so this should pass now 🙏

jaredlockhart:
Oh it is already a dataclass and still didn't catch that, great job Python 😅

experimenter/experimenter/targeting/constants.py
```diff
@@ -38,57 +38,11 @@ def __post_init__(self):
...
-)
-"""
-TOU_NOTIFICATION_BYPASS_ENABLED = """
-(
-    'datareporting.policy.dataSubmissionPolicyBypassNotification'|preferenceValue
-    ||
-    'termsofuse.bypassNotification'|preferenceValue
-)
-"""
+TOU_NOTIFICATION_BYPASS_ENABLED = "'termsofuse.bypassNotification'|preferenceValue"
```
dmose, line 45: I assume we'll be using this for experiments that have 142+ minimum versions?

experimenter/experimenter/targeting/constants.py
```diff
@@ -134,6 +100,67 @@ def __post_init__(self):
...
+                        HAS_PRIVACY_SETTING,
+                        HAS_AD_BLOCKER
+                    ]
+            ) | regExpMatch('true','g')
+        )
+        ||
+        []
+    ) | length
+)
+"""
```
dmose, line 162: This trick with regexp capture groups is clever; it took me a bit to figure out what was going on, so I'm a bit concerned about future maintenance here without comments.

dmose (commented):
In general, this looks good, though I've added a comment and a question.  Which is to say, r=dmose.

If it hasn't already been tested by QA, I'd strongly suggest trying to get them to QA it before using it.

My two concerns:
* typo somewhere in the JEXL, pref names or add-on IDs
* I've never seen any targeting like this regexp trick before, and it'd be nice to get QA to poke around the edges a bit

experimenter/experimenter/targeting/constants.py
lmegviar: Yes, Fx144+

PR #13581, opened 2025-09-24
feat(nimbus): make sidebar stick to top when scrolling

### LINKED ISSUE #13219: Sidebar shouldn't scroll with page content
It would be nice if all the links on the sidebar stayed in view even if the other page content scrolled.

moibra05:
Because

- It's annoying always needing to scroll up the entire page to access the sidebar 

This commit

- Gives the sidebar sticky positioning so that it's always on the user's screen

Fixes #13219


moibra05:
image

jaredlockhart (approved):
Okay ya looks great works great!  Only thing is if the links flow past the bottom of the viewport you gotta scroll all the way to the bottom of the page to see them, but we'll see if users complain and then figure out a better solution.  Thnx @moibra05 🙏

jaredlockhart:
Hrmm looking at circle it seems like the UI tests don't like this change, we may need to update them to account for this.

moibra05:
The reason circle ci was failing was b/c the sticky positioning of the sidebar was being inherited by the clone form which was preventing it from being interacted with since it would appear behind the modal overlay

PR #13592, opened 2025-09-25
fix(nimbus): fix recipe dropdown not able to close after opening

### LINKED ISSUE #13562: Collapsible menus cant be closed after opening
Recipe JSON dropdown cannot be closed after opening.

Reproducible on any experiment page.

Evidence:
https://github.com/user-attachments/assets/6e67d912-9240-4e67-a957-a4e310b2678e

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5970)

moibra05:
Because

- Collapsible menus such as the JSON recipes could not be closed after they were open

This commit

- FIxes this issue

Fixes #13562

### DIFF
--- a/experimenter/experimenter/nimbus_ui/static/js/experiment_detail.js
+++ b/experimenter/experimenter/nimbus_ui/static/js/experiment_detail.js
@@ -1,5 +1,3 @@
-import * as bootstrap from "bootstrap";
-
 const setupCollapseRecipe = () => {
   const collapseRecipe = document.getElementById("collapse-recipe");
   if (!collapseRecipe) return;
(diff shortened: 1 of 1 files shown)


yashikakhurana (commented):
+1 to Jared's comment, I think if we can keep in experiment_detail.js that would be a better place

moibra05:
the ease of this fix did not match the headache it gave me, but i've come out of it better 💔

jaredlockhart (approved):
Ooohhhh right right, that whole script was about the navigation link that should trigger it to expand, and here it was colliding with another instance of bootstrap rather than reusing the one that's already initialized.  Yeah yep you're right here, I can see why this was a hassle to find 🙈 

But yes testing locally I can expand and collapse it with the button, and the 'Preview Recipe JSON' does navigate to and expand it correctly.

Good job finding it, tricky!

PR #13596, opened 2025-09-25
fix(nimbus): make boolean text more visible in dark mode in codemirror windows

### LINKED ISSUE #13594: Boolean variables in configs are difficult to see in dark mode
Boolean variables are colored in a dark blue which is very difficult to read against the dark background (see green circled region in screenshot). Can we change this to a color that's easier to read?

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5977)

moibra05:
Because

- Some text (particularily boolean fields) wasn't visible in dark mode in any codemirror sections

This commit

- Improves the visibility of the text in dark mode

Fixes #13594


moibra05:
image
 
image

moibra05:
> Can't we just use some of the out of the box themes that specifically are for light/dark modes like
> Image Image
> 
> ?

I don't believe there are any default themes included in the core cm6 library but it is possible to install and integrate third party ones

jaredlockhart:
Yeah I'm just curious if that's a cleaner solution than managing the colours explicitly?  @moibra05 @yashikakhurana what do you think?

moibra05:
Incorporating themes would be a more robust solution but based on some research i did, it looks like having it dynamically switch between dark/light modes would add a bit of complexity

jaredlockhart (approved):
Okay well better to go with the simple solution first to get something out the door and then we if we want/need to we can spend more time on a broader solution later.  Thnx for hopping on this quick 🙏

experimenter/experimenter/nimbus_ui/static/js/branch_detail.js
```diff
@@ -1,20 +1,40 @@
...
 import { EditorState } from "@codemirror/state";
 import { json, jsonParseLinter } from "@codemirror/lang-json";
 import { linter } from "@codemirror/lint";
+import {
+  HighlightStyle,
+  syntaxHighlighting,
+  defaultHighlightStyle,
+  foldGutter,
+} from "@codemirror/language";
+import { tags } from "@lezer/highlight";
```
freshstrangemusic, line 16: What library is this? Are we transitively depending on it? If so, we should make it an explicit dependency.

I'm also not sure this is the correct approach. I'm looking at a [thread on codemirror.net](https://discuss.codemirror.net/t/dynamic-light-mode-dark-mode-how/4709/6) and this highlighted post seems to solve exactly what we're trying to do.

freshstrangemusic: See also this last post https://discuss.codemirror.net/t/dynamic-light-mode-dark-mode-how/4709/8 which has refactored it into a view plugin.

freshstrangemusic (changes requested):

experimenter/experimenter/nimbus_ui/static/js/branch_detail.js
moibra05: I got the "insipiration" from [this](https://codemirror.net/examples/styling/#highlighting) source. i thought since the issue with readibility was only with the boolean text in dark mode that that'd be a decent solution but i could totally try implementing dedicated themes for dark/light mode instead if that'd be the better approach.

experimenter/experimenter/nimbus_ui/static/js/edit_branches.js
```diff
@@ -12,8 +33,26 @@ const setupCodemirror = (selector, textarea, extraExtensions) => {
...
+    bracketMatching(),
+    closeBrackets(),
+    highlightSelectionMatches(),
+    indentOnInput(),
+    highlightActiveLine(),
+    syntaxHighlighting(highlightStyle),
+    syntaxHighlighting(defaultHighlightStyle),
+    history(),
+    search(),
+    keymap.of([...defaultKeymap, ...searchKeymap, ...historyKeymap]),
```
jaredlockhart, line 55: Yeah this is a bit hairy.  If using dedicated light/dark themes is too complicated let's just go back to your original approach of just overriding the colour of the bools.  Simplest, smallest solution is usually best 🙏

jaredlockhart (dismissed):

experimenter/experimenter/nimbus_ui/static/js/edit_branches.js
moibra05: The approach i have here is only changing the colour of the boolean text. `basicSetup` [packages all of these extensions](https://codemirror.net/docs/ref/#codemirror.basicSetup) without any option to exclude specific ones, which is an issue because [this](https://codemirror.net/docs/ref/#language.defaultHighlightStyle) included extension prevents me from setting a custom `HighlightStyle` that overrides boolean text colour. Annoyingly, the only way to fix this is to manually include all the extensions except `defaultHighlightStyle`

experimenter/experimenter/nimbus_ui/static/js/branch_detail.js
freshstrangemusic: `@lezer/highlight` is coming from a transitive dependency on codemirror, but it is not in our [package.json](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/nimbus_ui/static/package.json).

If we are going to depend on this package it should become a regular dependency.

freshstrangemusic (approved):

PR #13598, opened 2025-09-25
fix(nimbus): fix json fields being duplicated on unrelated htmx updates

### LINKED ISSUE #13595: Duplicate JSON fields created on HTMX updates
On the summary page, if you click the 'Edit' button the qa section, a duplicate JSON field on the feature values will be created.

We had the same issue on other pages, it's because we swap out just the QA section, so the rest of the DOM remains unaffected, but we run the 'afterSwap' event every time.

Simple thing to do here would be to add a flag to the JSON textareas that indicates whether codemirror was initialized and gate on that.

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5978)

moibra05:
Because

- Branch value json fields were being duplicated when "Edit" button in qa section was pressed

This commit

- Adds a flag to text area components that checks if it has been rendered to prevent duplicates

Fixes #13595

### DIFF
--- a/experimenter/experimenter/nimbus_ui/static/js/branch_detail.js
+++ b/experimenter/experimenter/nimbus_ui/static/js/branch_detail.js
@@ -11,6 +11,8 @@ const setupCodemirroReadOnlyJSON = () => {
   const textareas = document.querySelectorAll(selector);
 
   textareas.forEach((textarea) => {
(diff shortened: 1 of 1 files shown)


moibra05:
image

PR #13602, opened 2025-09-26
feat(nimbus): made branch values on overview page collapsible

### LINKED ISSUE #12988: Make branch values collapsible on Summary page
The JSON branch value can be quite large, it'd be nice if it was collapsible.

Example experiment https://experimenter.services.mozilla.com/nimbus/accounts-vpn-promotion/summary

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5675)

moibra05:
Because

- Long branch values would take up large amounts of space on the screen

This commit

- Allows branch values to be collapsed similar to recipe json

Fixes #12988


moibra05:
blocked as i wait to test if fixes made in #13592 apply

jaredlockhart (changes requested):
Yep that works, can you see if you can find some way to combine [overflow](https://getbootstrap.com/docs/5.0/utilities/overflow/) and [max height](https://getbootstrap.com/docs/5.0/utilities/sizing/) to see if you can do something like 

```
1: ....
2: ...
3: ...
4: ...
5: ...
<Show All>
```

Ie show the first 5 lines, hide the rest, and then have the collapse button trigger showing the rest? 🙏 

If it's too much of a hassle we can just leave it as is, but give it a shot.

moibra05:
ooo i'm definitely going to try that first suggestion

moibra05:
i'm a little iffy about using innerHTML cuz ik it isn't best practice to use so lmk if thats something I should tweak

image
image

jaredlockhart (changes requested):
Okay so, yes this does work exactly correctly, but futzing about with like, pixel calculations and hard coding html snippets in js isn't really a healthy pattern, becuase later when you're moving things around in the html this will break and it won't be obvious why.

But I think there's another way to go about it.  Try something like this

```html
<div class="collapsable-json">
  <div class="json-preview-collapsed">
    <div class="max-height-whatever">
      <textarea>{{ featurevalue }}</textarea>
    </div>
    <button class="expand-button">Expand</button>
  </div>
  <div class="json-preview-expanded d-none">
    <textarea>{{featurevalue }}</textarea>
    <button class="hide-button">Hide</button>
  </div>
</div>
```

and then in the js it's more like

```js
$('expand-button').click((evt) => {
  evt.target.parent('collapsable-json').find('json-preview-collapse').addClass('d-none');  # roughly, futz with finding the parent 
  evt.target.parent('collapsable-json').find('json-preview-expanded').removeClass('d-none');
});

# inverse for the hide button
```

so basically there's two, one that's only ever max-heighted, one that's only ever full height, and you just swap the visibility of the two.

Play around with that and see if you can get it 🙏

moibra05:
From what I can tell and if i understood ur feedback correctly, there isnt a way to set a fixed max height value using bs classes so i didnt see any other way to do this w/o manually setting styles

yashikakhurana:
@moibra05 this might help-https://getbootstrap.com/docs/5.0/utilities/sizing/#relative-to-the-viewport

jaredlockhart (approved):
Ya this is great.  Much clearer, works perfectly.  Great job @moibra05 ty 🙏 🎉

PR #13711, opened 2025-10-17
feat(nimbus): Add sorting to deliveries table on featues page.

### LINKED ISSUE #13707: Add sorting to deliveries table
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6028)

b4handjr:
Because

- We want to add sorting to the deliveries table

This commit

- Adds sorting to the deliveries table
- Fixes formatting issues

Fixes #13707


b4handjr:
> found one bug on sorting on dates tooltip is stuck

Removed the tooltip as it isn't a part of the spec. Since I am using the `common/home_sortable_header.html` for the header, it was being generated because I had the Launch Date field named as "Dates".

PR #13718, opened 2025-10-20
feat(nimbus): setup results page with required data

### LINKED ISSUE #13682: Prepare Django results view context for migration from React
Migrate the experiment results page from React to Django by preparing the Django view to fetch, format, and provide all the required analysis and experiment data to the template for rendering

{color:#97a0af}

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6021)

moibra05:
Because

- Results page view context requires specific data to render contents 

This commit

- Passes the required context data through the view

Fixes #13682

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1546,6 +1546,19 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         return False
 
     @property
+    def has_exposures(self):
+        # True if there are any exposures in the results data
+        if self.results_data and "v3" in self.results_data:
+            results_data = self.results_data["v3"]
+            for window in ["overall", "weekly"]:
+                if results_data.get(window):
+                    exposure_data = results_data[window].get("exposures", {}).get("all")
+                    if exposure_data is not None:
+                        return True
+
+        return False
+
+    @property
     def show_results_url(self):
         # if there are results, show them! even if the dates are wrong
         # (the dates may have been overridden in metric-hub)
--- a/experimenter/experimenter/experiments/tests/test_models.py
+++ b/experimenter/experimenter/experiments/tests/test_models.py
@@ -3259,6 +3259,22 @@ class TestNimbusExperiment(TestCase):
 
         self.assertTrue(experiment.has_displayable_results)
 
+    def test_has_exposures_results_true(self):
+        results_data = {"v3": {"overall": {"exposures": {"all": {}}}}}
+        experiment = NimbusExperimentFactory.create()
+        experiment.results_data = results_data
+        experiment.save()
+
+        self.assertTrue(experiment.has_exposures)
+
+    def test_has_exposures_results_false(self):
+        results_data = {"v3": {"overall": {"enrollments": {"all": {}}}}}
+        experiment = NimbusExperimentFactory.create()
+        experiment.results_data = results_data
+        experiment.save()
+
+        self.assertFalse(experiment.has_exposures)
+
     @parameterized.expand(
         [
             ({},),
--- a/experimenter/experimenter/nimbus_ui/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui/tests/test_views.py
@@ -2988,6 +2988,102 @@ class TestResultsView(AuthTestCase):
         self.assertEqual(response.context["experiment"], experiment)
         self.assertTemplateUsed(response, "nimbus_experiments/results.html")
 
+    @parameterized.expand(
+        [
+            (
+                {
+                    "v3": {
+                        "other_metrics": {"group": {"metricA": "Metric A"}},
+                        "metadata": {
+                            "metrics": {"metricA": {"friendlyName": "Friendly Metric A"}},
+                        },
+                        "overall": {
+                            "enrollments": {"all": {}},
+                            "exposures": {"all": {}},
+                        },
+                        "weekly": {
+                            "enrollments": {"all": {}},
+                            "exposures": {"all": {}},
+                        },
+                    }
+                },
+                "exposures",
+                {"metricA": "Friendly Metric A"},
+            ),
+            (
+                {
+                    "v3": {
+                        "other_metrics": {"group": {"metricA": "Metric A"}},
+                        "metadata": {
+                            "metrics": {"metricA": {"friendlyName": "Friendly Metric A"}},
+                        },
+                        "overall": {"enrollments": {"all": {}}},
+                    }
+                },
+                "enrollments",
+                {"metricA": "Friendly Metric A"},
+            ),
+        ]
+    )
+    def test_results_view_context_and_defaults(
+        self, results_data, selected_analysis_basis, default_metrics
+    ):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE,
+        )
+
+        experiment.results_data = results_data
+        experiment.save()
+
+        response = self.client.get(
+            reverse("nimbus-ui-results", kwargs={"slug": experiment.slug}),
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.context["results_data"], experiment.results_data.get("v3")
+        )
+        self.assertEqual(response.context["default_metrics"], default_metrics)
+        self.assertEqual(
+            response.context["selected_reference_branch"],
+            experiment.reference_branch.name,
+        )
+        self.assertEqual(response.context["selected_segment"], "all")
+        self.assertEqual(
+            response.context["selected_analysis_basis"], selected_analysis_basis
+        )
+
+    def test_results_view_query_param_overrides(self):
+        experiment = NimbusExperimentFactory.create_with_lifecycle(
+            NimbusExperimentFactory.Lifecycles.ENDING_APPROVE_APPROVE,
+        )
+
+        experiment.results_data = {
+            "v3": {
+                "other_metrics": {},
+                "metadata": {"metrics": {}},
+                "overall": {"enrollments": {"all": {}}, "exposures": {"all": {}}},
+            }
+        }
+        experiment.save()
+
+        response = self.client.get(
+            reverse(
+                "nimbus-ui-results",
+                kwargs={"slug": experiment.slug},
+                query={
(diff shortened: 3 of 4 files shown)


schemas/mozilla_nimbus_schemas/jetstream/metadata.py
```diff
@@ -50,6 +50,7 @@ class Metadata(BaseModel):
     version_info: Optional[ConfigVersions] = None
     version_date: Optional[dt.datetime] = None
     schema_version: int = SCHEMA_VERSION
+    has_exposures: bool = False
```
mikewilli: My preference would be to make this a little more generic, but if we want to start with this and iterate later that's OK too.

mikewilli: Thinking some more about this, I don't know if this is the approach we want to take here. This schema defines the contract for metadata coming from Jetstream, which does not have this value. When we set the default value to `False` that will be incorrect for many experiments. So I think we want to at least change the default value to `None`, but we could also consider putting this information inside the `results_data` alongside the results instead of the metadata.

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -634,6 +634,45 @@ class ApproveUpdateRolloutView(StatusUpdateView):
...
+
+        selected_reference_branch = self.request.GET.get(
+            "reference_branch", experiment.reference_branch.name
+        )
+        context["selected_reference_branch"] = selected_reference_branch
+
+        selected_segments = self.request.GET.get("segment", "all")
+        context["selected_segments"] = selected_segments
+
+        has_exposures = analysis_data.get("has_exposures", False)
```
mikewilli: This won't be correct for past experiments with exposures because the metadata value won't be there and so it will default to `False`. We'll need to add a database migration to retroactively populate.

experimenter/experimenter/jetstream/client.py
```diff
@@ -244,6 +244,7 @@ def get_experiment_data(experiment: NimbusExperiment):
                 experiment_data[window][AnalysisBasis.ENROLLMENTS] = {}
                 raw_data[window][AnalysisBasis.ENROLLMENTS] = {}
             elif point["analysis_basis"] == AnalysisBasis.EXPOSURES:
+                experiment_data["has_exposures"] = point.get("point", 0) > 0
```
mikewilli: I don't think this solves the issue I mentioned previously. It will still set this value based on the last data point in the list, and it will be the same for all windows.

experimenter/experimenter/jetstream/tests/test_tasks.py
mikewilli, line 1: We should be sure to test a case where it's `False`, whether that's within an existing test or a new one.

moibra05: I have some tests in tests_views.py with `has_exposures` set to`False` unless you specifically mean we need some within this file as well

experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -2978,6 +2978,87 @@ def test_render_to_response(self):
...
+                "other_metrics": {},
+                "metadata": {"metrics": {}},
+                "has_exposures": True,
+            }
+        }
+        experiment.save()
+
+        response = self.client.get(
+            reverse("nimbus-ui-results", kwargs={"slug": experiment.slug})
+            + "?reference_branch=treatment-a&segment=foo&analysis_basis=exposures"
```
freshstrangemusic:
```suggestion
            reverse("nimbus-ui-results", kwargs={"slug": experiment.slug}, query={"reference_branch": "treatment-a", "segment": "foo", "analysis_basis": "exposures"})
```

`reverse()` supports a query= argument for providing a querystring [docs](https://docs.djangoproject.com/en/5.2/ref/urlresolvers/)

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -650,6 +650,45 @@ class ApproveUpdateRolloutView(StatusUpdateView):
...
+
+        analysis_data = experiment.results_data.get("v3", {})
+        other_metrics = analysis_data.get("other_metrics", {})
+        metrics_metadata = analysis_data.get("metadata", {}).get("metrics", {})
+        default_metrics = {}
+
+        for value in other_metrics.values():
+            for metricKey, metricValue in value.items():
+                default_metrics[metricKey] = (
+                    metrics_metadata.get(metricKey, {}).get("friendlyName") or metricValue
```
freshstrangemusic: Why not `.get("friendlyName", metricValue)` ? Are we intentionally covering the case where `friendlyName` is present but an empty string?

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -650,6 +650,45 @@ class ApproveUpdateRolloutView(StatusUpdateView):
...
+        has_exposures = analysis_data.get("has_exposures", False)
+
+        if has_exposures:
+            context["selected_analysis_basis"] = self.request.GET.get(
+                "analysis_basis", "exposures"
+            )
+        else:
+            context["selected_analysis_basis"] = self.request.GET.get(
+                "analysis_basis", "enrollments"
+            )
```
freshstrangemusic: We can simplify this a bit:

```suggestion
        default_analysis_basis = "exposures" if analysis_data.get("has_exposures", False) else "enrollments"
        context["selected_analysis_basis"] = self.request.GET.get("analysis_basis", default_analysis_basis)
```

experimenter/experimenter/nimbus_ui/views.py
moibra05: empty strings are falsy in python so it'd still fall back on `metricValue`. i think here i was just translating js code and forgot i could do that

experimenter/experimenter/experiments/models.py
```diff
@@ -1523,6 +1523,17 @@ def has_displayable_results(self):
 
         return False
 
+    @property
+    def has_exposures(self):
+        # True if there are any exposures in the results data
+        if self.results_data and "v3" in self.results_data:
+            overall_results = self.results_data.get("v3", {}).get("overall", {})
```
mikewilli: We should check `weekly` as well as `overall`, otherwise this will never evaluate to true until the experiment ends. You could also consider whether it's worth having both weekly and overall checked distinctly. I'm not sure what the UI will look like to say whether that's worthwhile or not, and it can also be something that changes later, especially with this property-based approach.

mikewilli: It's probably reasonable to just take the approach in `has_displayable_results` for consistency.

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -3194,6 +3194,22 @@ def test_has_displayable_results_true(self, results_data):
 
         self.assertTrue(experiment.has_displayable_results)
 
+    def test_has_exposures_results_true(self):
+        results_data = {"v3": {"overall": {"exposures": {"all": [1]}}}}
```
mikewilli: This should at least match the type of data inside `all` (i.e., it's an object not a list). As I mentioned above, I think it's reasonable to accept `{}` as valid here, but if you want to ensure there is actual data you can also descend further into the object to check the branch_data. Either way, the test should reflect what you're actually looking for instead of `[1]`.

mikewilli: I see you're using `[1]` in the view tests below as well, so for brevity I'll just suggest that those match the expected data too.

experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -2979,6 +2979,94 @@ def test_render_to_response(self):
...
+            response.context["default_metrics"]["metricA"], "Friendly Metric A"
+        )
+        self.assertEqual(
+            response.context["selected_reference_branch"],
+            experiment.reference_branch.name,
+        )
+        self.assertEqual(response.context["selected_segments"], "all")
+        self.assertEqual(response.context["selected_analysis_basis"], "exposures")
+
+    def test_results_view_no_exposures(self):
```
mikewilli: Consider integrating this test into the one above using pytest.parametrize to pass dynamic values into parts of the `results_data` object.

This approach should also help if you add `weekly` checks as I suggested above.

experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -2979,6 +2979,94 @@ def test_render_to_response(self):
...
+
+        response = self.client.get(
+            reverse(
+                "nimbus-ui-results",
+                kwargs={"slug": experiment.slug},
+                query={
+                    "reference_branch": "treatment-a",
+                    "segment": "foo",
+                    "analysis_basis": "exposures",
+                },
```
mikewilli, line 3078: This is probably just a contrived test, but it seems like there shouldn't even be an option to select things that aren't in the data, no?

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -651,6 +651,38 @@ class ApproveUpdateRolloutView(StatusUpdateView):
...
+
+        context["default_metrics"] = default_metrics
+
+        selected_reference_branch = self.request.GET.get(
+            "reference_branch", experiment.reference_branch.name
+        )
+        context["selected_reference_branch"] = selected_reference_branch
+
+        selected_segments = self.request.GET.get("segment", "all")
+        context["selected_segments"] = selected_segments
```
mikewilli: Nit: typo (also in the tests)
```suggestion
        selected_segment = self.request.GET.get("segment", "all")
        context["selected_segment"] = selected_segment
```

experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -2988,6 +2988,102 @@ def test_render_to_response(self):
...
+                        "overall": {
+                            "enrollments": {"all": {}},
+                            "exposures": {"all": {}},
+                        },
+                        "weekly": {
+                            "enrollments": {"all": {}},
+                            "exposures": {"all": {}},
+                        },
+                    }
+                },
```
mikewilli, line 3009: If you wanted to be more concise, you could keep the static parts in the function and only put the dynamic stuff (the parts that go in `weekly` and `overall`) in here. This is OK for now though.

PR #13777, opened 2025-10-24
feat(nimbus): create new results page overview section

### LINKED ISSUE #13734: Add a new results page view and template
Add a barebone results page that includes all the static data (i.e., the overview section containing the hypothesis)

This new page should have an updated sidebar that references the new results page and its sections (e.g., overview, KPI metrics, etc)

The page should be available at a new url, /nimbus/<slug>/results-new/, which will initially only be available by editing the URL. Eventually we will replace the old results view with this new one and remove the temporary URL.

The page does not need interactivity at this point (toggleable sections, dropdowns for enrollment vs exposure basis, segments, etc)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6041)

moibra05:
Because

- Results overview section requires redesign

This commit

- Creates a new route at 'nimbus/[slug]/results-new' hosting redesigned results page
- Contains the redesigned overview section

Fixes #13734

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -361,7 +361,7 @@ jobs:
     resource_class: large
     working_directory: ~/experimenter
     environment:
-      FIREFOX_CHANNEL: nightly
+      FIREFOX_CHANNEL: release
       PYTEST_ARGS: -k FIREFOX_DESKTOP -m desktop_enrollment --reruns 1 --base-url https://nginx/nimbus/
       PYTEST_BASE_URL: https://nginx/nimbus/
     steps:
--- a/cirrus/server/poetry.lock
+++ b/cirrus/server/poetry.lock
@@ -1384,14 +1384,14 @@ files = [
 
 [[package]]
 name = "starlette"
-version = "0.47.2"
+version = "0.49.1"
 description = "The little ASGI library that shines."
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "starlette-0.47.2-py3-none-any.whl", hash = "sha256:c5847e96134e5c5371ee9fac6fdf1a67336d5815e09eb2a01fdb57a351ef915b"},
-    {file = "starlette-0.47.2.tar.gz", hash = "sha256:6ae9aa5db235e4846decc1e7b79c4f346adf41e9777aebeb49dfd09bbd7023d8"},
+    {file = "starlette-0.49.1-py3-none-any.whl", hash = "sha256:d92ce9f07e4a3caa3ac13a79523bd18e3bc0042bb8ff2d759a8e7dd0e1859875"},
+    {file = "starlette-0.49.1.tar.gz", hash = "sha256:481a43b71e24ed8c43b11ea02f5353d77840e01480881b8cb5a26b8cae64a8cb"},
 ]
 
 [package.dependencies]


moibra05:
image

yashikakhurana:
@moibra05 I will review it once it is ready, but the screenshot you provided, results page is looking good, can't wait to see once this is finished

jaredlockhart:
Wow wow wow!  Look at it go!

experimenter/experimenter/experiments/models.py
```diff
@@ -1568,6 +1568,19 @@ def has_displayable_results(self):
...
+        # True if there are any exposures in the results data
+        if self.results_data and "v3" in self.results_data:
+            results_data = self.results_data["v3"]
+            for window in ["overall", "weekly"]:
+                if results_data.get(window):
+                    exposure_data = results_data[window].get("exposures", {}).get("all")
+                    if exposure_data is not None:
+                        return True
+
+        return False
```
jaredlockhart, line 1639: Exactly 🎉 🎉 🎉 🎉

jaredlockhart:
This is all coming along great, looks great.  I'd say this could've been done as several PRs, one just to land the skeleton of the page, and then the parts about looking up exposures, the parts about number formatting, etc could've all been their own little PRs.  No reason to change this, but as you keep working on it think of things like that all as their own littler changes, easier to review, faster to land.

Edit: I didn't realize this included another PR in it, but still good to think about when things can be split up even more 👍

experimenter/experimenter/nimbus_ui/templatetags/nimbus_extras.py
```diff
@@ -303,3 +303,25 @@ def experiment_date_progress(experiment):
...
+        return f"{n}"
+    if n < 1_000_000:
+        v = n / 1000.0
+        if v.is_integer():
+            return f"{int(v)}k"
+        return f"{v:.{int(precision)}f}k"
+    v = n / 1_000_000.0
+    if v.is_integer():
+        return f"{int(v)}M"
+    return f"{v:.{int(precision)}f}M"
```
b4handjr: There's gotta be a library or something for this.. right??

moibra05: i found [this](https://pypi.org/project/numerize/) although i think it might be a bit overkill

jaredlockhart: What do you mean by overkill?  Usually better to use a well supported library than do stuff manually when possible.

moibra05: I've always thought that if something seems simple enough to write myself, relying on a library would be excessive. Another learning opportunity for me lol

b4handjr: Check out https://pypi.org/project/humanize/

jaredlockhart: @moibra05 Every line of code written is a line that has to be maintained.  Every line has a cost.  Less line better 🙏

freshstrangemusic: There is always a trade off because on the other end of the spectrum you end up in NPM left-pad hell and you pull in 10s to 100s of packages for what is a 10 line function. But on the other end you have problems that appear easy (like dates and times) that should really be outsourced to people who think about that stuff all day ([relevant xkcd](https://xkcd.com/2867/))

experimenter/experimenter/experiments/models.py
```diff
@@ -1258,6 +1258,65 @@ def timeline(self):
...
+        )
+
+        branch_data = []
+        branches = []
+
+        if self.reference_branch:
+            branches.append(self.reference_branch)
+        branches.extend(self.treatment_branches)
+
+        for branch in branches:
```
freshstrangemusic: branch iteration order doesn't seem to matter here so we can:


```suggestion
        for branch in self.branches.all():
```

We should probably prefetch the screenshot FK relationship:

```
self.branches.all().prefetch_related("screenshots")
```

otherwise we do N + 1 queries instead of 2

experimenter/experimenter/nimbus_ui/templates/common/branch_card.html
```diff
@@ -0,0 +1,30 @@
+{% load nimbus_extras %}
+
+<div class="d-flex gap-4">
+  {% with shot=branch.screenshots.first %}
+    {% if shot and shot.image %}
+      <img src="{{ shot.image.url }}"
+           alt="{{ shot.description|default:branch.name }}"
+           class="border border-secondary-subtle rounded w-50"
+           style="height: 120px;
+                  object-fit: cover" />
```
freshstrangemusic, line 10: cc @jaredlockhart but stuff like this feels like it should be in a CSS file 

dont spin too long on this, i havent actually worked on the new UI myself so I dont know if we have infra for that

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -675,12 +666,27 @@ def get_context_data(self, **kwargs):
...
+
+        return context
+
+
+class ResultsView(NimbusExperimentViewMixin, DetailView):
+    template_name = "nimbus_experiments/results.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
         return context
```
freshstrangemusic, line 690: This is equivalent to not providing the method.

However, I'm confused. Why are we removing all the context from the old results view? Does that page still work with this patch applied?

moibra05: i believe that's how it was setup before i started working on the results page so i thought to just revert it back to how it was originally

experimenter/experimenter/experiments/models.py
moibra05: the order is important here since it defines how the branch cards are displayed which make the most sense having the control branch be first

experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -3069,7 +3069,7 @@ def test_results_view_query_param_overrides(self):
 
         response = self.client.get(
             reverse(
-                "nimbus-ui-results",
+                "nimbus-ui-new-results",
```
yashikakhurana, line 3072: I would suggest to keep the existing test intact for now until we switch the page

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -648,25 +648,16 @@ class ApproveUpdateRolloutView(StatusUpdateView):
     form_class = ApproveUpdateRolloutForm
 
 
-class ResultsView(NimbusExperimentViewMixin, DetailView):
-    template_name = "nimbus_experiments/results.html"
+class NewResultsView(NimbusExperimentViewMixin, DetailView):
+    template_name = "nimbus_experiments/results-new.html"
```
yashikakhurana, line 653: oh I am not sure, are we launching this page? If we are planning to launch, my suggestion will be to do in a separate PR so that we can revert back if anything goes wrong, if we are not planning to launch, lets keep the current pages intact

experimenter/experimenter/nimbus_ui/templates/common/branch_card.html
freshstrangemusic: You can ignore this :)

experimenter/experimenter/nimbus_ui/views.py
freshstrangemusic: Oh I didn't realize :)

PR #13780, opened 2025-10-24
chore(tests): Add publish/unpublish to preview test.

### LINKED ISSUE #8436: Add integration test for publishing and unpublishing to preview
Let's add an integration test that creates an experiment, publishes it to preview, asserts it's in the preview collection, moves it back to draft, and then asserts it's removed from the preview collection.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-3255)

b4handjr:
Because

- We want to make sure we can publish and unpublish to preview correctly

This commit

- Adds a test to ensure we can publish and unpublish from preview
- Adds the ability to unpublish from preview for local environments including tests

Fixes #8436

### DIFF
--- a/experimenter/experimenter/kinto/client.py
+++ b/experimenter/experimenter/kinto/client.py
@@ -16,6 +16,9 @@ class KintoClient:
         )
         self.review = review
         self.collection_data = None
+        self.read_bucket = (
+            settings.KINTO_BUCKET_WORKSPACE if not review else settings.KINTO_BUCKET_MAIN
+        )
 
     def _fetch_collection_data(self):
         self.collection_data = (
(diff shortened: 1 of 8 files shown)


experimenter/experimenter/kinto/client.py
```diff
@@ -94,6 +97,6 @@ def get_main_records(self):
         return {
             r["id"]: r
             for r in self.kinto_http_client.get_records(
-                bucket=settings.KINTO_BUCKET_MAIN, collection=self.collection
+                bucket=self.read_bucket, collection=self.collection
```
yashikakhurana, line 100: oh @b4handjr I think should make changes for local environment only so that we don't break the prod and stage, you can add a check and change the bucket accordingly

b4handjr: Well I technically didn't break any tests... 🤐

yashikakhurana: but test doesn't include our stage and prod, I am just worried about that

jaredlockhart: Oh @yashikakhurana I don't think this changes which environment it goes to, it's just whether it's in the `main` or `main-workspace` collections in the local test kinto 🙏

PR #13787, opened 2025-10-27
feat(nimbus): Add subscribers field to feature config model.

### LINKED ISSUE #13783: Add subscribers field to FeatureConfig
We can just copy the way we do this on NimbusExperiment:

[https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/models.py#L406-L411](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/models.py#L406-L411|smart-link) 

on to the NimbusFeatureConfig model

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6061)

b4handjr:
Because

- We want to allow a user to subscribe to a feature and get notifications when that feature is used

This commit

- Adds the subscribers field to the Feature Config class.

Fixes #13783

### DIFF
--- a/experimenter/experimenter/experiments/migrations/0297_nimbusfeatureconfig_subscribers.py
+++ b/experimenter/experimenter/experiments/migrations/0297_nimbusfeatureconfig_subscribers.py
@@ -5,16 +5,20 @@ from django.db import migrations, models
 
 
 class Migration(migrations.Migration):
-
     dependencies = [
-        ('experiments', '0296_nimbusexperiment_qa_run_date_and_more'),
+        ("experiments", "0296_nimbusexperiment_qa_run_date_and_more"),
         migrations.swappable_dependency(settings.AUTH_USER_MODEL),
     ]
 
     operations = [
         migrations.AddField(
-            model_name='nimbusfeatureconfig',
-            name='subscribers',
-            field=models.ManyToManyField(blank=True, related_name='subscribed_nimbusfeatures', to=settings.AUTH_USER_MODEL, verbose_name='Subscribers'),
+            model_name="nimbusfeatureconfig",
+            name="subscribers",
+            field=models.ManyToManyField(
+                blank=True,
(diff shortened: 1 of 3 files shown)


freshstrangemusic:
> ahaa I see adding a subscriber for that, just a suggestion the name `subscriber` is little bit confusing for me, as we have subscriber for the experiments too, seeing the same field might be little confusing 🤔

Its the same name on two different models. And the concept is the same -- a subscription.

We have types, I don't see this being a problem.

experimenter/experimenter/experiments/migrations/0297_nimbusfeatureconfig_subscribers.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 5.2.7 on 2025-10-27 15:00
```
freshstrangemusic, line 1: This needs a ruff format.

experimenter/experimenter/experiments/models.py
```diff
@@ -2291,6 +2291,12 @@ class NimbusFeatureConfig(models.Model):
     )
     owner_email = models.EmailField(blank=True, null=True)
     enabled = models.BooleanField(default=True)
+    subscribers = models.ManyToManyField(
+        User,
+        related_name="subscribed_nimbusfeatures",
```
freshstrangemusic: can we make the related name `subscribed_nimbus_features` ?

experimenter/experimenter/experiments/migrations/0297_nimbusfeatureconfig_subscribers.py
b4handjr: We must have a bug on our `make lint` step as it sometimes doesn't report these.

yashikakhurana (approved):
yeah I understand the concept is same, I said it is confusing for me :D, lets :shipit:

PR #13789, opened 2025-10-27
bug(experimenter): give glean a unique data dir per pid

### LINKED ISSUE #13788: Glean requires a different data dir per pid
When experimenter glean support was merged, we started getting glean [errors in sentry](https://mozilla.sentry.io/issues/6963802494/?project=6377199), even though data uploads are disabled. Initially I talked to Travis and we thought this was glean issue, but then when I was explaining the issue to Jared, I realized there were multiple processes initializing glean with the same data dir, which was an issue we also saw in cirrus. We should fix this by adding process id as a subdir to the end of the glean data dir, just like we did for cirrus.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6093)

relud:
Because

- glean must have a unique data dir per pid

This commit

- Adds a process id subdir to the glean data dir path

Fixes #13788

### DIFF
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -10,6 +10,7 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/1.9/ref/settings/
 """
 
+import os
 from pathlib import Path
 from urllib.parse import urljoin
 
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/settings.py
```diff
@@ -74,7 +75,7 @@
 
 GLEAN_UPLOAD_ENABLED = config("GLEAN_UPLOAD_ENABLED", default=False, cast=bool)
 
-GLEAN_DATA_DIR = config("GLEAN_DATA_DIR", default="/var/glean")
+GLEAN_DATA_DIR = Path(config("GLEAN_DATA_DIR", default="/var/glean")) / str(os.getpid())
```
relud, line 78: this matches what was done for the same issue in cirrus: https://github.com/mozilla/experimenter/blob/2598eb66840092b6ed8394844c6f98054aac6512/cirrus/server/cirrus/settings.py#L64

Issue #13715, opened 2025-10-17
Add basic feature changes table


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6030)

PR #13799, opened 2025-10-29
feat(nimbus): Add feature schema data to feature changes table.

b4handjr:
Because

- We want to be able to show a diff view against the previous version of a feature schema in the Feature Changes table.

This commit

- Adds the capability to show a diff view of each schema using codemirror
- Add DeepDiff to get  specifics on the differences between the schemas

Extends #13715 

I needed to specify the direct paths for webpack due to an issue where it will find and load 2 codemirror state modules causing an error.

### DIFF
--- a/experimenter/experimenter/nimbus_ui/views.py
+++ b/experimenter/experimenter/nimbus_ui/views.py
@@ -782,15 +782,15 @@ class NimbusFeaturesView(TemplateView):
         # Get feature schema versions with their diffs
         feature_schemas = []
         feature_id = self.request.GET.get("feature_configs")
+        thresholds = SCHEMA_DIFF_SIZE_CONFIG["thresholds"]
+        labels = SCHEMA_DIFF_SIZE_CONFIG["labels"]
+
         if feature_id:
             schemas = list(
                 NimbusVersionedSchema.objects.filter(feature_config_id=feature_id)
                 .select_related("version")
                 .order_by("-version__major", "-version__minor", "-version__patch")
             )
-
-            thresholds = SCHEMA_DIFF_SIZE_CONFIG["thresholds"]
-            labels = SCHEMA_DIFF_SIZE_CONFIG["labels"]
 
             schema_cache = []
             for schema in schemas:
@@ -858,7 +858,6 @@ class NimbusFeaturesView(TemplateView):
             "qa_runs_non_sortable_header": qa_runs_non_sortable_fields,
             "feature_schemas": feature_schemas,
             "schemas_with_changes": schemas_with_changes,
-            "schema_diff_labels": SCHEMA_DIFF_SIZE_CONFIG["labels"],
         }
         return context
 


b4handjr:
image

Diff View
image

experimenter/experimenter/nimbus_ui/static/webpack.config.js
```diff
@@ -10,11 +11,24 @@ module.exports = {
...
+      "@codemirror/state": path.resolve(nodeModulesPath, "@codemirror/state"),
+      "@codemirror/view": path.resolve(nodeModulesPath, "@codemirror/view"),
+      "@codemirror/merge": path.resolve(nodeModulesPath, "@codemirror/merge"),
+      "@codemirror/lang-json": path.resolve(
+        nodeModulesPath,
+        "@codemirror/lang-json",
+      ),
+      codemirror: path.resolve(nodeModulesPath, "codemirror"),
+    },
+  },
```
freshstrangemusic, line 31: Can you remove these aliases and try adding direct dependencies for `@codemirror/state`, `@codemirror/view`, and `@codemirror-lang-json` in our package.json ?

freshstrangemusic: (at the same version as code mirror)

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -342,36 +342,73 @@ <h5 class="fw-semibold">
...
+                    <span>Change Diff from Prev. (Previous | Current)</span>
                   </th>
                 </tr>
               </thead>
               <tbody>
-                {% for change in feature_changes %}
+                {% for item in feature_schemas %}
                   <tr scope="row">
-                    <td id="feature-version">{{ change.change_version|format_not_set }}</td>
-                    <td id="feature-change-date">{{ change.change_date|date:"m/d/Y @ H:i"|format_not_set }}</td>
```
yashikakhurana, line 371: You asked “will we keep the date?” — yes, I’d suggest keeping it.
It helps QA and users see the timeline of schema evolution.
If it feels visually heavy, consider:

moving it under the version number in smaller muted text (text-muted fs-sm), or

showing only the date (not time) to reduce clutter.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -342,36 +342,73 @@ <h5 class="fw-semibold">
...
+                      {{ item.schema.version|default:"current" }}
+                    </td>
+                    <td scope="col" class="col-4">
+                      {% if item.size_label == "Small" %}
+                        <span class="badge bg-primary">{{ item.size_label }}</span>
+                      {% elif item.size_label == "Medium" %}
+                        <span class="badge bg-warning text-dark">{{ item.size_label }}</span>
+                      {% elif item.size_label == "Large" %}
+                        <span class="badge bg-danger">{{ item.size_label }}</span>
+                      {% elif item.size_label is None %}
```
yashikakhurana: this we can move here for the mapping, like this- https://github.com/mozilla/experimenter/blob/3b47b59f73e6c48af58786975f231b38d53fab62/experimenter/experimenter/nimbus_ui/constants.py#L199

yashikakhurana:
```
FEATURE_CHANGE_SIZE_LABELS = {
    "none": "No Changes",
    "small": "Small",
    "medium": "Medium",
    "large": "Large",
}
```

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -761,6 +767,63 @@ def get_context_data(self, **kwargs):
...
+            )
+
+            # Pair each schema with its previous version
+            for i, schema in enumerate(schemas):
+                current_json = schema.schema
+                previous_json = schemas[i + 1].schema if i + 1 < len(schemas) else None
+
+                if previous_json is not None:
+                    current_obj = json.loads(current_json)
+                    previous_obj = json.loads(previous_json)
```
yashikakhurana: suggestion, this we can improve this by not computing json twice, I mean current can be previous of someother and hence we are computing it twice

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -761,6 +767,63 @@ def get_context_data(self, **kwargs):
...
+                        size_label = "No Changes"
+                    elif total_changes <= 3:
+                        size_label = "Small"
+                    elif total_changes <= 10:
+                        size_label = "Medium"
+                    else:
+                        size_label = "Large"
+
+                    if total_changes > 0:
+                        schemas_with_changes += 1
```
yashikakhurana, line 830: we can store the boundaries as in constant too 
```
FEATURE_CHANGE_THRESHOLDS = (0, 3, 10)  # (none, small, medium) upper bounds
```

yashikakhurana (changes requested):
Thank you @b4handjr tested locally overall looks good, nice job showing the diff, I am glad diff merged was simple and easy to show the changes, couple of thoughts

I think by using fixed height somewhere and hence it is showing the blank empty space, can you fix this to be dynamic
Screenshot 2025-10-29 at 10 26 58 AM

Second when we are saying first version- can we show the initial version only, i know we can see by going to the newer version and see comparison as previous version, but from the users point of view if we can show the first version that would be better

Screenshot 2025-10-29 at 10 26 48 AM
 Third- some schemas have lot of history and hence we are keep scrolling, can we add pagination to it
https://github.com/user-attachments/assets/b1265fe4-36f6-4f8a-bd46-003300ca4c33

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
b4handjr: Well we have no dates in the database. These features are all bulk loaded when the container is created so there's no way to get a date from them.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
b4handjr: Is this needed for things that are used in 1 place? I could see your case with the example you linked but there won't be any other pages that use this mapping.

yashikakhurana: It's better practice to keep things in one place, such as constants, so that in future if we want to use somewhere else we can refer to them :)

b4handjr:
Instead of doing the Pagination, I am using a feature in MergeView to just show the lines changed:
image

When there are no lines changed:
image-1

With the dynamic height, it fits well.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -342,36 +342,73 @@ <h5 class="fw-semibold">
...
+                    </td>
+                    <td scope="col" class="col-4">
+                      {% if item.size_label == schema_diff_labels.small %}
+                        <span class="badge bg-primary">{{ item.size_label }}</span>
+                      {% elif item.size_label == schema_diff_labels.medium %}
+                        <span class="badge bg-warning text-dark">{{ item.size_label }}</span>
+                      {% elif item.size_label == schema_diff_labels.large %}
+                        <span class="badge bg-danger">{{ item.size_label }}</span>
+                      {% elif item.size_label is None %}
+                        <span class="text-muted">First version</span>
```
yashikakhurana: the one of the reason was to move to the constant so that we dont need the multiple ifs for the right classes, for example- https://github.com/mozilla/experimenter/blob/8eb2efc2b99d917f9edce595f894ef00637c5ea6/experimenter/experimenter/nimbus_ui/constants.py#L172

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -769,6 +779,67 @@ def get_context_data(self, **kwargs):
...
+
+                    if total_changes == 0:
+                        size_label = labels["no_changes"]
+                    elif total_changes <= thresholds["small"]:
+                        size_label = labels["small"]
+                    elif total_changes <= thresholds["medium"]:
+                        size_label = labels["medium"]
+                    else:
+                        size_label = labels["large"]
+
```
yashikakhurana, line 828: ⭐

b4handjr:
I'll be adding pagination and sorting in a following PR

b4handjr:
I had to move the test files around so that my tests for this were in the correct class.

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -784,6 +856,9 @@ def get_context_data(self, **kwargs):
             "deliveries_non_sortable_header": deliveries_non_sortable_fields,
             "qa_runs_sortable_header": qa_runs_sortable_header,
             "qa_runs_non_sortable_header": qa_runs_non_sortable_fields,
+            "feature_schemas": feature_schemas,
+            "schemas_with_changes": schemas_with_changes,
+            "schema_diff_labels": SCHEMA_DIFF_SIZE_CONFIG["labels"],
```
yashikakhurana: I think you are not using this anymore, I think your above dict will have the labels now 🤔

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -769,6 +779,68 @@ def get_context_data(self, **kwargs):
...
+
+                feature_schemas.append(
+                    {
+                        "schema": schema,
+                        "current_json": current_json,
+                        "previous_json": previous_json,
+                        "size_label": size_label.get("text"),
+                        "size_badge": size_label.get("badge_class"),
+                    }
+                )
```
yashikakhurana, line 842: what I mean was that your dict contains the labels already here, I am not sure you are using this variable anywhere else ` "schema_diff_labels": labels,` I mean your feature schema dict have all the labels information now

yashikakhurana (changes requested):

experimenter/experimenter/nimbus_ui/views.py
b4handjr: Oh okay, I didn't know exactly what you meant

b4handjr:
> Thanks @b4handjr tested locally, looks good now, Also as a followup can you add a ticket for what Beth suggested, just don't want to loose the improvement suggestion and can tackle later

Sure, this is a node issue though and it will probably be solved when we update our node version.

PR #13819, opened 2025-10-29
bug(nimbus): display branch name validation errors

### LINKED ISSUE #13603: Duplicate branch slug error not shown
We should be showing an error here that the branch slugs can't be duplicated

Image

jaredlockhart:
Becuase

* Branch names and slugs are editable after experiment creation time
* We need to enforce slug uniqueness for branches
* We missed adding a validator that checked this to the branch form

This commit

* Adds a validator for unique name/slug pairs for branches
* Adds tests

fixes #13603


jaredlockhart:
image

PR #13826, opened 2025-10-30
feat(nimbus): use fml linter in branches form

### LINKED ISSUE #12968: Fix and implement FML linting in new Nimbus UI
It seems at some point our FML validation endpoint broke and is no longer working.  Let's fix that and then implement it in the new Nimbus UI.

jaredlockhart:
Becuase

* In the old nimbus-ui, we called out to an API to do FML validation for non desktop feature values
* This was lost when we setup the new nimbus-ui

This commit

* Adds a linter to the new feature values form that calls the FML linting API

fixes #12968


jaredlockhart:
Screenshot 2025-10-30 at 11 44 36

PR #13827, opened 2025-10-30
feat(nimbus): add localizations to summary page

### LINKED ISSUE #13070: Localizations for the multi-locale experiments are not shown in the summary page
Shane and Kamyar reported that the substitutions are no longer visible when in preview mode or launched

[shane] it was easier before the new UI, since I could see the l10n substitutions at the bottom

the substitutions are no longer visible when in preview mode or launched.  This makes it tricky to fully review the config without putting it back to draft mode.

here is an example experiment now where you can see that [https://experimenter.services.mozilla.com/nimbus/windows-10-eos-sync-messaging-background-message/summary/](https://experimenter.services.mozilla.com/nimbus/windows-10-eos-sync-messaging-background-message/summary/)

[shane] they do show up if I expand the raw recipe json. hard to locate them there though, with such a giant experiment. it would be more straightforward if there was a l10n section I could search for with ctrl+f

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5734)

jaredlockhart:
Becuase

* We should display the localizations info on the summary page because the branches page is no longer available after launch

This commit

* Adds localization info to the summary page

fixes #13070

### DIFF
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail.html
@@ -381,6 +381,30 @@
 
       {% endfor %}
     </div>
+    <!-- Localizations Card -->
+    {% if experiment.is_localized %}
+      <div class="card mb-3">
+        <div class="card-header">
+          <h4>Localizations</h4>
+        </div>
+        <div class="card-body">
+          <table class="table table-striped">
+            <tbody>
+              <tr>
+                <th>Is Localized</th>
+                <td colspan="3">{{ experiment.is_localized }}</td>
+              </tr>
+              <tr>
+                <th>Localizations JSON</th>
+                <td colspan="3">
+                  <textarea class="readonly-json">{{ experiment.localizations|default:"null" }}</textarea>
+                </td>
+              </tr>
+            </tbody>
+          </table>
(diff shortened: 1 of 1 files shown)


jaredlockhart:
Screenshot 2025-10-30 at 12 27 53

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail.html
```diff
@@ -381,6 +381,30 @@ <h4>Branches ({{ experiment.branches.all.count }})</h4>
...
+            <tbody>
+              <tr>
+                <th>Is Localized</th>
+                <td colspan="3">{{ experiment.is_localized }}</td>
+              </tr>
+              <tr>
+                <th>Localizations JSON</th>
+                <td colspan="3">
+                  <textarea class="readonly-json">{{ experiment.localizations|default:"null" }}</textarea>
+                </td>
```
yashikakhurana, line 401: In future if json is too big, then we can add codemirror too 💪

jaredlockhart: It is using codemirror already thanks to the super useful classes added by @moibra05 🎉 🎉 🎉

PR #13829, opened 2025-10-30
bug(experimenter): send glean opt outs in custom ping

### LINKED ISSUE #13828: Glean does not submit events pings
we need to send data collection opt outs in a custom ping because with the python glean sdk the events ping only gets submitted when there are 500 outstanding events or when [glean.Glean.handle_client_inactive](https://mozilla.github.io/glean/python/glean/index.html#glean.Glean.handle_client_inactive) is called, and I don’t think we should depend on either of those for this.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6125)

relud:
Because

- python glean doesn't send the events ping unless there's 500 outstanding events or we use [glean.Glean.handle_client_inactive](https://mozilla.github.io/glean/python/glean/index.html#glean.Glean.handle_client_inactive)

This commit

- Removes the data collection opt out event
- Adds a data collection opt out ping

Fixes #13828

### DIFF
--- a/experimenter/experimenter/glean/views.py
+++ b/experimenter/experimenter/glean/views.py
@@ -11,6 +11,7 @@ class OptOutView(UpdateView):
     template_name = "glean/opt_out_button.html"
 
     metrics = glean.load_metrics(settings.BASE_DIR / "telemetry" / "metrics.yaml")
+    pings = glean.load_pings(settings.BASE_DIR / "telemetry" / "pings.yaml")
 
     def get_object(self, queryset=None):
         if not hasattr(self.request.user, "glean_prefs"):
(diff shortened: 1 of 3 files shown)


relud:
> `500 outstanding events` I think we can configure this like I mean we can adjust the value when to send it

that makes sense and i didn't think of that, but going from an event to a ping also makes it so shredder doesn't need a derived table in https://github.com/mozilla/bigquery-etl/pull/8370, so i'm going to stick with the custom ping

yashikakhurana:
> > `500 outstanding events` I think we can configure this like I mean we can adjust the value when to send it
> 
> that makes sense and i didn't think of that, but going from an event to a ping also makes it so shredder doesn't need a derived table in [mozilla/bigquery-etl#8370](https://github.com/mozilla/bigquery-etl/pull/8370), so i'm going to stick with the custom ping

absolutely +1 on using custom ping in this scenario

PR #13850, opened 2025-10-31
feat(nimbus): Add sorting to feature changes table.

### LINKED ISSUE #13813: Add sorting to feartue changes table
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6032)

b4handjr:
Because

- We need to be able to sort the feature changes table

This commit

- Adds sorting to the feature changes table.

Fixes #13813

### DIFF
--- a/experimenter/experimenter/nimbus_ui/views.py
+++ b/experimenter/experimenter/nimbus_ui/views.py
@@ -805,30 +805,17 @@ class NimbusFeaturesView(TemplateView):
         if feature_id:
             sort = self.request.GET.get("sort", "")
 
-            schema_order_by = [
-                "-version__major",
-                "-version__minor",
-                "-version__patch",
-            ]
-
-            # Map version sort
             if sort == "change_version":
-                schema_order_by = [
-                    "version__major",
-                    "version__minor",
-                    "version__patch",
-                ]
-            elif sort == "-change_version":
-                schema_order_by = [
-                    "-version__major",
-                    "-version__minor",
-                    "-version__patch",
-                ]
+                queryset = NimbusVersionedSchema.objects.with_version_ordering(
+                    descending=False
+                )
+            else:
+                queryset = NimbusVersionedSchema.objects.with_version_ordering(
+                    descending=True
+                )
 
             schemas = list(
-                NimbusVersionedSchema.objects.filter(feature_config_id=feature_id)
-                .select_related("version")
-                .order_by(*schema_order_by)
+                queryset.filter(feature_config_id=feature_id).select_related("version")
             )
 
             schema_cache = []
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -2573,6 +2573,14 @@ class NimbusFeatureVersion(models.Model):
         return packaging.version.parse(str(self))
 
 
+class NimbusVersionedSchemaManager(models.Manager["NimbusVersionedSchema"]):
+    def with_version_ordering(self, descending=False):
+        """Order schemas by semantic version (major.minor.patch)."""
+        if descending:
+            return self.order_by("-version__major", "-version__minor", "-version__patch")
+        return self.order_by("version__major", "version__minor", "version__patch")
+
+
 class NimbusVersionedSchema(models.Model):
     feature_config = models.ForeignKey(
         NimbusFeatureConfig,
@@ -2591,6 +2599,8 @@ class NimbusVersionedSchema(models.Model):
     set_pref_vars = models.JSONField[dict[str, str]](null=False, default=dict)
     is_early_startup = models.BooleanField(null=False, default=False)
     has_remote_schema = models.BooleanField(null=False, default=False)
+
+    objects = NimbusVersionedSchemaManager()
 
     class Meta:
         verbose_name = "Nimbus Versioned Schema"


experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -779,17 +784,51 @@ def get_context_data(self, **kwargs):
...
+                    "version__minor",
+                    "version__patch",
+                ]
+            elif sort == "-change_version":
+                schema_order_by = [
+                    "-version__major",
+                    "-version__minor",
+                    "-version__patch",
+                ]
+
```
yashikakhurana: same here

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -838,9 +878,15 @@ def get_context_data(self, **kwargs):
...
                         "size_badge": size_label.get("badge_class"),
+                        "total_changes": total_changes,
                     }
                 )
 
+            if sort == "change_size":
+                feature_schemas.sort(key=lambda x: x["total_changes"])
+            elif sort == "-change_size":
+                feature_schemas.sort(key=lambda x: x["total_changes"], reverse=True)
+
```
yashikakhurana, line 876: same here

yashikakhurana:
Also @b4handjr one more task is left to add the features button on the header so that users can access it

experimenter/experimenter/nimbus_ui/views.py
b4handjr: So we can do versions, which is what I am doing, but I am just using a different keyword instead of sorting by "-version__major, -version__minor, -version__patch", because it makes the URL for the sort quite long. And the "changes" aren't in the database so they have to be sorted externally from using a filterset on the data.

yashikakhurana: umm this doesn't seems to be the right approach, may be try annotation or if its a computed property you can try making it a property

b4handjr: Can you explain with an example?

yashikakhurana:
> Also @b4handjr one more task is left to add the features button on the header so that users can access it

oh also noticed its in the requirement doc too, it will help users to access the new page, otherwise only we know the new url to access it
Screenshot 2025-10-31 at 1 25 36 PM

experimenter/experimenter/nimbus_ui/views.py
yashikakhurana: one way annotation
```
   # Add annotation for change_size
    qs = qs.annotate(
        change_size=Case(
            When(schema__isnull=True, then=0),  # No schema
            When(previous_schema__isnull=True, then=1),  # First version
            default=F('computed_diff_size'),  # Use computed field
            output_field=IntegerField()
        )
    )
```
second way property
```
class NimbusVersionedSchema(models.Model):
    # ... existing fields
    
    @property
    def computed_diff_size(self):
        # Your existing diff calculation logic
        return self.calculate_diff_size()
```
third way manager
```
class NimbusVersionedSchemaManager(models.Manager):
    def with_change_size_ordering(self, descending=False):
        order = "-diff_size" if descending else "diff_size"
        return self.annotate(
            computed_change_size=F('diff_size')
        ).order_by(order)
    
    def with_version_ordering(self, descending=False):
        if descending:
            return self.order_by("-version__major", "-version__minor", "-version__patch")
        return self.order_by("version__major", "version__minor", "version__patch")
```

b4handjr:
I was able add a manager for the versions but I think adding something similar for the "Size of Changes" field will be a bit overkill. This would require a migration and to populate that field on the model before sorting. This field is only used for this tabel and doing some research it seems sorting data after it has been queried from the db with normal python sort methods isn't technically wrong.

yashikakhurana:
> I was able add a manager for the versions but I think adding something similar for the "Size of Changes" field will be a bit overkill. This would require a migration and to populate that field on the model before sorting. This field is only used for this tabel and doing some research it seems sorting data after it has been queried from the db with normal python sort methods isn't technically wrong.

@b4handjr I believe we can annotate or may be can add a property without requiring to need the another field in db

yashikakhurana:
then     
``` 
@property
    def change_size_value(self):
        if hasattr(self, 'sortable_change_size'):
            return self.sortable_change_size
        
        # Fallback calculation based on version
        if self.version:
            if self.version.major < 121:
                return 0
            elif self.version.major == 121:
                return 1
            elif self.version.major == 122:
                return 2
            else:
                return 3
        return 0
```
and in view 
```
    # ... existing code ...
    
    if feature_id:
        sort = self.request.GET.get("sort", "")
        
        # Get base queryset
        if sort == "change_version":
            schemas = list(
                NimbusVersionedSchema.objects.filter(feature_config_id=feature_id)
                .with_version_ordering(descending=False)
                .select_related("version")
            )
        elif sort == "-change_version":
            schemas = list(
                NimbusVersionedSchema.objects.filter(feature_config_id=feature_id)
                .with_version_ordering(descending=True)
                .select_related("version")
            )
        elif sort in ["change_size", "-change_size"]:
            schemas = list(
                NimbusVersionedSchema.objects.filter(feature_config_id=feature_id)
                .with_change_size_annotation()
                .order_by("sortable_change_size" if sort == "change_size" else "-sortable_change_size")
                .select_related("version")
            )
        else:
            schemas = list(
                NimbusVersionedSchema.objects.filter(feature_config_id=feature_id)
                .select_related("version")
                .order_by("-version__major", "-version__minor", "-version__patch")
            )
        
        for i, schema in enumerate(schemas):
            # ... existing diff calculation ...
            feature_schemas.append({
                "schema": schema,
                "current_json": current_json,
                "previous_json": previous_json,
                "size_label": size_label.get("text"),
                "size_badge": size_label.get("badge_class"),
            })

b4handjr:
> you can annotate same size change like you did for the version in
> 
> ```
> class NimbusVersionedSchemaManager
>     def with_change_size_annotation(self):
>         
>         return self.annotate(
>             sortable_change_size=Case(
>                 # You can add more sophisticated logic here
>                 When(version__major__lt=121, then=0),  # First versions
>                 When(version__major=121, then=1),      # Small changes
>                 When(version__major=122, then=2),      # Medium changes  
>                 When(version__major__gt=122, then=3),  # Large changes
>                 default=0,
>                 output_field=IntegerField()
>             )
>         )
> ```

I don't understand this. What does the version number have to do with the size of changes? The changes are computed using DeepDiff and comparing the schemas.

yashikakhurana:
> > you can annotate same size change like you did for the version in
> > ```
> > class NimbusVersionedSchemaManager
> >     def with_change_size_annotation(self):
> >         
> >         return self.annotate(
> >             sortable_change_size=Case(
> >                 # You can add more sophisticated logic here
> >                 When(version__major__lt=121, then=0),  # First versions
> >                 When(version__major=121, then=1),      # Small changes
> >                 When(version__major=122, then=2),      # Medium changes  
> >                 When(version__major__gt=122, then=3),  # Large changes
> >                 default=0,
> >                 output_field=IntegerField()
> >             )
> >         )
> > ```
> 
> I don't understand this. What does the version number have to do with the size of changes? The changes are computed using DeepDiff and comparing the schemas.

@b4handjr this is just the example to help you.  you can structure according to your requirements, Also since it is a version schema manger, logically it can contain size of the version too

b4handjr:
> @b4handjr this is just the example to help you. you can structure according to your requirements, Also since it is a version schema manger, logically it can contain size of the version too

Wouldn't this mean I would need to add the diff data to the database?

yashikakhurana:
> > @b4handjr this is just the example to help you. you can structure according to your requirements, Also since it is a version schema manger, logically it can contain size of the version too
> 
> Wouldn't this mean I would need to add the diff data to the database?

not if you try to do with annotation or compute property

b4handjr:
> we can land this with the current state, thank you for adding the improvement for the versions, for the change diff we don't need to add db field but can be done as compute property or like annotation, can you add a followup ticket to improve that? Atleast we will be able to land this and then your button pr too and then can do the followup, @b4handjr what do you think?

I can do that. We will have to pair on zoom and you will have to walk me through this as I am still new to this.

b4handjr:
Filed: https://github.com/mozilla/experimenter/issues/13872

yashikakhurana:
> > we can land this with the current state, thank you for adding the improvement for the versions, for the change diff we don't need to add db field but can be done as compute property or like annotation, can you add a followup ticket to improve that? Atleast we will be able to land this and then your button pr too and then can do the followup, @b4handjr what do you think?
> 
> I can do that. We will have to pair on zoom and you will have to walk me through this as I am still new to this.

sure can do that, can you file a ticket please so that we don't loose this improvement, I will approve it then

yashikakhurana:
> Thanks @b4handjr, unrelated but found another bug but we should definitely fix this now as this will not go through QA process, on changing the application, feature version schema table is not getting updated
> Screen.Recording.2025-11-03.at.12.07.36.PM.mov

@b4handjr can you file ticket for that and fix this 🙏

PR #13859, opened 2025-10-31
chore(experimenter): use server-side glean

### LINKED ISSUE #13858: Use server-side glean in experimenter
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6138)

relud:
Because

- We want to limit our glean pings to what comes from our gcpv2 environments

This commit

- Swaps out python glean sdk for glean-parser generated `server_events.py`

Fixes #13858


relud:
note: this requires glean-parser >= `18.1.0`, which is not yet released, and until then the new CI job `Check Glean Generated Files` should not pass

experimenter/experimenter/glean/server_events.py
```diff
@@ -0,0 +1,222 @@
+"""
+This Source Code Form is subject to the terms of the Mozilla Public
+License, v. 2.0. If a copy of the MPL was not distributed with this
+file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+AUTOGENERATED BY glean_parser v18.0.6.post1.dev0+gb423bb9f7.d20251031. DO NOT EDIT. DO NOT COMMIT.
```
jaredlockhart: If this can be auto generated, does it need to be committed?  Can we gitignore it and do it as part of the container build step so it winds up in teh container but doesn't appear in the repo?

experimenter/experimenter/glean/middleware.py
```diff
@@ -29,17 +25,21 @@ def __call__(self, request):
...
+                enrollments = [
+                    {
+                        **enrollment,
+                        "is_preview": enrollment.get("is_preview") == "true",
+                    }
+                    for enrollment in request.cirrus.enrollments
+                ]
+            self.page_view_ping.record(
+                user_agent=request.META.get("HTTP_USER_AGENT"),
+                ip_address=get_request_ip(request),
```
jaredlockhart, line 41: Do we need to track the client IP?

Makefile
```diff
@@ -216,6 +216,9 @@ format: build_dev  ## Format source tree
 	$(COMPOSE_RUN) experimenter sh -c '${PARALLEL} "$(RUFF_FIX);$(DJLINT_FIX);$(RUFF_FORMAT_FIX)" "$(ESLINT_FIX_CORE)" "$(ESLINT_FIX_RESULTS)" "$(ESLINT_FIX_NIMBUS_UI)"'
 code_format: format
 
+glean_generated_files: build_dev
```
jaredlockhart: In other cases where we auto generate files we have two make commands, one to generate, and one to check whether the generation step needs to be run, by running it and diffing against teh committed version.  Should we also be doing that here?

experimenter/experimenter/glean/middleware.py
relud: if we report i believe the telemetry pipeline will remove it after enriching the pings with geoip fields, which is what we do for client-side glean applications, but we can leave it empty

experimenter/experimenter/glean/server_events.py
relud: local dev environments mount the git root which would hide any generated files in that tree, so we need to commit it for dev to work

Makefile
relud: sure

experimenter/experimenter/glean/server_events.py
jaredlockhart: Oh you can block out specific paths from the volume mount if that's the issue:

https://github.com/mozilla/experimenter/blob/main/docker-compose.yml#L17-L25

Makefile
```diff
@@ -55,6 +55,11 @@ RUFF_FORMAT_CHECK = ruff format --check --diff .
 RUFF_FORMAT_FIX = ruff format .
 CHECK_DOCS = python manage.py generate_docs --check=true
 GENERATE_DOCS = python manage.py generate_docs
+GLEAN_PARSER_TRANSLATE = glean_parser translate experimenter/telemetry/pings.yaml experimenter/telemetry/metrics.yaml -f python_server -o experimenter/glean
+GLEAN_PARSER_TRANSLATE_CHECK = \
+  glean_parser translate experimenter/telemetry/pings.yaml experimenter/telemetry/metrics.yaml -f python_server -o /tmp/glean &&\
+	diff /tmp/glean/server_events.py experimenter/glean/server_events.py || (echo "glean_parser translate is out of sync please run make glean_parser_translate"; exit 1) &&\
+	echo "Done. No problems found in glean_parser translate."
```
jaredlockhart: Perfect ya exactly

experimenter/experimenter/glean/server_events.py
relud: til, i'll do that instead of committing the file

relud: fixed

jaredlockhart (approved):
Ya good, looks good, thnx @relud 🙏 🎉

PR #13870, opened 2025-11-03
feat(nimbus): create filters for updated results page

### LINKED ISSUE #13720: Support analysis basis, segments, and reference branch on new results page
We need to be able to swap what our analysis basis is (enrollment vs exposure), as well as what segments and reference branch we are using for analysis.

The exposure basis option should only be present if there are exposure events present.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6044)

moibra05:
Because

- Users need a way to change the analysis basis, reference branch, and segment view for result data

This commit

- Creates drop down filters for each of the above listed fields
- Analysis basis drop down only shows exposures if the experiment has them

Fixes #13720


moibra05:
image

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new.html
```diff
@@ -6,113 +6,157 @@
...
-              </div>
-              <div class="col">
-                <p class="text-muted mb-0">Application</p>
-                <p>{{ experiment.application|default:"Unknown" }}</p>
-              </div>
+  <div>
+    <form hx-get="{{ request.path }}"
+          hx-push-url="true"
+          hx-trigger="change"
+          hx-target="body"
```
yashikakhurana: you can also target specific div by using the ids for example here- https://github.com/mozilla/experimenter/blob/709f0f962b8392a6fbe080b57ab4776890bc491f/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/home.html#L78

yashikakhurana (commented):
Thanks @moibra05 in general looks good, I have one suggestion if you can try to target specifics id instead of body

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new.html
```diff
@@ -6,168 +6,6 @@
...
-              </div>
-            </div>
-          </div>
-        </div>
-        <div class="card p-5">
-          <div class="row row-cols-2 row-cols-xxl-3 g-4">
-            {% for branch in branch_data %}
-              <div class="col">
-                {% include "common/branch_card.html" with branch=branch %}
+  {% include "nimbus_experiments/results-new-fragment.html" %}
```
yashikakhurana, line 9: I believe this file `results-new.html` is only containing `results-new-fragment` right? if that is the case I dont think we need separate file for this?

moibra05: `results-new.html` also extends `nimbus_experiments/experiment_base.html` which contains other experimenter page elements such as the header and sidebar. when i make htmx get requests back to the new results page to reload content on filter changes i only want the results page content to be returned which is what `results-new-fragment` contains. separating these files gives me control in the exact content returned by the view depending on if the `HX-Request` header is provided

moibra05: also yes, filters should work and appear to be with the experiment you linked

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new-inner.html
```diff
@@ -0,0 +1,157 @@
...
+              {% include "common/metric_card.html" with slug=branch.slug reference_branch=selected_reference_branch %}
+              {% include "common/metric_card.html" with slug=branch.slug reference_branch=selected_reference_branch %}
+              {% include "common/metric_card.html" with slug=branch.slug reference_branch=selected_reference_branch %}
+              {% include "common/metric_card.html" with slug=branch.slug reference_branch=selected_reference_branch %}
+              {% include "common/metric_card.html" with slug=branch.slug reference_branch=selected_reference_branch %}
+              {% include "common/metric_card.html" with slug=branch.slug reference_branch=selected_reference_branch %}
+
+            </div>
+          </div>
+        {% endfor %}
```
yashikakhurana, line 153: I wish we can find better solution for this

PR #13879, opened 2025-11-03
fix(nimbus): Fix feature changes table data persisting on app change.

### LINKED ISSUE #13877: Feature changes table does not reset when application is changed
The schema table isn't resetting to empty when we change an application. It looks like the `feature_config` variable is being kept even though it may not exist with the new application.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6145)

b4handjr:
Because

- We uncovered a bug in which the feature changes table on the feature health page would persist data after loading and the user selecting another application.

This commit

- Only loads the feature schemas table if there is a valid and existing queryset.

Fixes #13877

### DIFF
--- a/experimenter/experimenter/nimbus_ui/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui/tests/test_views.py
@@ -4462,6 +4462,66 @@ class TestNimbusFeaturesView(AuthTestCase):
         self.assertIn(self.user, feature.subscribers.all())
         self.assertEqual(response.status_code, 200)
 
+    def test_features_view_tables_reset_on_new_request_after_loading(self):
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -804,7 +804,7 @@ def get_context_data(self, **kwargs):
         labels = SCHEMA_DIFF_SIZE_CONFIG["labels"]
         total_changes = 0
 
-        if feature_id:
+        if feature_id and qs:
```
yashikakhurana, line 823: can you add a test related to this scenario 🙏

PR #13854, opened 2025-10-31
feat(nimbus): display feature coenrollment

jaredlockhart:
Because
    
* We now support coenrolling features in all Nimbus clients
* It would be handy to know if a feature is coenrolling in the UI
    
This commit
    
* Adds a new field to store allow_coenrollment on features
* Displays a message on the branch page if a feature supports coenrollment
* Displays a message on the summary page if a feature supports coenrollment
    
fixes #12914

PR #13888, opened 2025-11-04
chore(nimbus): add nimbusuiconstants context processor

### LINKED ISSUE #13878: Use a context processor for NimbusUIConstants
Rather than having to specifically pass individual NimbusUIConstant properties into views, we can just setup a context processor to add the whole class to the template context so those contants are always available.  Should simplify all the invidual view context overrides.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6146)

jaredlockhart:
Because
    
* We refer to NimbusUIConstants in many templates
* Rather than explicitly passing it in in multiple views, we can just pass it in once in a context processor
    
This commit
    
* Adds a context processor for NimbusUIConstants
* Updates all references to use that
* Removes unnecessary context variables in individual views
    
Fixes #13878


jaredlockhart:
Blocked on https://github.com/mozilla/experimenter/pull/13854

PR #13902, opened 2025-11-04
bug(nimbus): strictly order branches on edit branches page

### LINKED ISSUE #13760: Nimbus branches change their order after the experiment is saved
**[Steps to reproduce]:**

1. Create a new delivery and navigate to the branches page.
1. Fill in the page with data and save it.
1. Navigate to the Summary page.
1. Return to the branches page and observe the branches section.

**[Expected result]:**

* The branches remain saved in the order they were created.

**[Actual result]:**

* The branches switch positions.

**[Notes]:**

* Attached a screen recording of the issue:

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6072)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/99316">branches_switch.gif</a>

jaredlockhart:
Becuase

* The NimbusBranch model uses slug as its default ordering
* No stable ordering was defined on the edit branches page
* This means that if a branch is renamed, it will get reordered on the page
* We also did not force the control branch to appear first
* The control branch cannot be deleted
* By renaming branches the control branch could be moved out of the first position and have a delete button even though it cannot be deleted

This commit

* Adds a stable ordering to the branches forms that forces the control branch to be first and the remaining branches to be ordered by id so they remain in the same order they are created

fixes #13760

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -5,7 +5,8 @@ from datetime import UTC, datetime
 import markus
 from django import forms
 from django.contrib.auth.models import User
-from django.forms import BaseModelFormSet, inlineformset_factory
+from django.db.models import Case, When
+from django.forms import BaseInlineFormSet, BaseModelFormSet, inlineformset_factory
 from django.http import HttpRequest
 from django.urls import reverse
 from django.utils.text import slugify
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -680,21 +698,22 @@ class Meta:
...
         }
         if self.files:
             branches_formset_kwargs["files"] = self.files
 
-        self.branches = self.NimbusBranchFormSet(**branches_formset_kwargs)
+        NimbusBranchFormSet = inlineformset_factory(
+            NimbusExperiment,
+            NimbusBranch,
+            form=NimbusBranchForm,
+            formset=OrderedBranchFormSet,
```
yashikakhurana, line 712: 🎉

PR #13905, opened 2025-11-04
bug(experimenter): make glean writes to stdout more atomic

### LINKED ISSUE #13893: server-side glean pings aren't being delivered
┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6158)

relud:
Because

- In prod glean pings are coming out as `{ping}{ping}\n\n` instead of `{ping}\n{ping}\n` and markh suggested this might fix it

This commit

- Overrides the glean generated emit_record methods to use `stdout.write(f"{ping}\n")` instead of `print(ping)`

Fixes #13893

### DIFF
--- a/experimenter/experimenter/glean/utils.py
+++ b/experimenter/experimenter/glean/utils.py
@@ -1,6 +1,25 @@
+import json
+from datetime import datetime
+from sys import stdout
+from typing import Any
+
+from experimenter.glean.generated.server_events import GLEAN_EVENT_MOZLOG_TYPE
+
+
 def get_request_ip(request) -> None | str:
     if xff := request.META.get("HTTP_X_FORWARDED_FOR"):
         # Only trust the last 3 values in XFF because they are added by the Google Cloud
         # Load Balancer and nginx, and use the least recent of those values.
         return xff.rsplit(",", 4)[-3:][0]
     return request.META.get("REMOTE_ADDR")
+
+
+def emit_record(now: datetime, ping: dict[str, Any]):
+    ping_envelope = {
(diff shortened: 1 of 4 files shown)


relud:
if this works, we should move the change upstream to glean-parser, and revert it here when we pick up the change

experimenter/experimenter/glean/utils.py
```diff
@@ -1,6 +1,25 @@
...
+
+def emit_record(now: datetime, ping: dict[str, Any]):
+    ping_envelope = {
+        "Timestamp": now.isoformat(),
+        "Logger": "glean",
+        "Type": GLEAN_EVENT_MOZLOG_TYPE,
+        "Fields": ping,
+    }
+    ping_envelope_serialized = f"{json.dumps(ping_envelope)}\n"
+    stdout.write(ping_envelope_serialized)
```
relud, line 25: the original method is:
```py
 def emit_record(self, now: datetime, ping:dict[str, Any]) -> None:
        """Log the ping to STDOUT.
        Applications might want to override this method to use their own logging.
        If doing so, make sure to log the ping as JSON, and to include the
        `Type: GLEAN_EVENT_MOZLOG_TYPE`."""
        ping_envelope = {
            "Timestamp": now.isoformat(),
            "Logger": "glean",
            "Type": GLEAN_EVENT_MOZLOG_TYPE,
            "Fields": ping,
        }
        ping_envelope_serialized = json.dumps(ping_envelope)

        print(ping_envelope_serialized)
```

yashikakhurana (approved):
Thanks @relud

PR #13931, opened 2025-11-10
bug(nimbus): fix version filtering on list page

### LINKED ISSUE #13904: Filtering by multiple versions returns empty list
If you select more than one version in the version filter on the list page, it'll return an empty list rather than all the experiments that match any of those versions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6168)

jaredlockhart:
Becuase

* We added version filtering on the list page
* It worked if you select a single version, but not if you select multiple
* Turns out using reversed() in choices is silently buggy because it returns an iterator which only works once but needs to be accessed multiple times
* Wrapping it in list() solves the issue
* This wasn't caught becuase the existing test only filtered on a single version

This commit

* Updates the test to filter on multiple versions
* Wraps the reversed choices in a list so it can be accessed multiple times

fixes #13904

### DIFF
--- a/experimenter/experimenter/nimbus_ui/filtersets.py
+++ b/experimenter/experimenter/nimbus_ui/filtersets.py
@@ -155,7 +155,7 @@ class NimbusExperimentFilter(django_filters.FilterSet):
         ),
     )
     firefox_min_version = django_filters.MultipleChoiceFilter(
-        choices=reversed(NimbusExperiment.Version.choices),
+        choices=list(reversed(NimbusExperiment.Version.choices)),
         widget=IconMultiSelectWidget(
             icon="fa-solid fa-code-branch",
             attrs={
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/nimbus_ui/filtersets.py
```diff
@@ -155,7 +155,7 @@ class NimbusExperimentFilter(django_filters.FilterSet):
         ),
     )
     firefox_min_version = django_filters.MultipleChoiceFilter(
-        choices=reversed(NimbusExperiment.Version.choices),
+        choices=list(reversed(NimbusExperiment.Version.choices)),
```
yashikakhurana, line 158: oh sweet

yashikakhurana (approved):
Nice change, thanks @jaredlockhart

PR #13932, opened 2025-11-10
feat(nimbus): update metric data view to use real experiment data

### LINKED ISSUE #13871: Display real metric data in cards
Currently the metric cards are static and do not actually use the obtained metric data. Update the experiment model so that metric data can be easily used by the view and template and is reflected in the “change cards”

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6143)

moibra05:
Because

- Results card was not displaying real experiment data

This commit

- Updates metric cards to use the actual data obtained by experiments

Fixes #13871

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -994,6 +994,45 @@ We believe this because we have observed <this> via <data source, UR, survey>.
 Optional - We believe this outcome will <describe impact> on <core metric>
     """  # noqa
 
+    KPI_AREA = "KPI Metrics"
+    NOTABLE_CHANGES_AREA = "Notable Changes"
+    DEFAULT_METRIC_AREAS = [NOTABLE_CHANGES_AREA, KPI_AREA]
+
+    DAILY_ACTIVE_USERS = "client_level_daily_active_users_v2"
+    DAYS_OF_USE = "days_of_use"
+    RETENTION = "retained"
+    SEARCH_COUNT = "search_count"
+
+    DAU_METRIC = {
+        "group": "other_metrics",
+        "friendly_name": "Daily Active Users",
+        "slug": DAILY_ACTIVE_USERS,
+        "description": "Average number of client that sent a main ping per day.",
+    }
+
+    DOU_METRIC = {
+        "group": "other_metrics",
+        "friendly_name": "Days of Use",
+        "slug": DAYS_OF_USE,
+        "description": "Average number of days each client sent a main ping.",
+    }
+
+    KPI_METRICS = [
+        {
+            "group": "other_metrics",
+            "friendly_name": "Retention",
+            "slug": RETENTION,
+            "description": "Percentage of users who returned to Firefox two weeks later.",
+            "display_type": "percentage",
+        },
+        {
+            "group": "search_metrics",
+            "friendly_name": "Search Count",
+            "slug": SEARCH_COUNT,
+            "description": "Daily mean number of searches per user.",
+        },
+    ]
+
     MAX_PRIMARY_OUTCOMES = 2
     DEFAULT_PROPOSED_DURATION = 28
     DEFAULT_PROPOSED_ENROLLMENT = 7
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1286,6 +1286,15 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         return timeline_entries
 
     @property
+    def metric_areas(self):
+        metric_areas = NimbusConstants.DEFAULT_METRIC_AREAS.copy()
+
+        # TODO(13721): check all metrics with data and add their areas to metric_areas
+        # if not already present
+
+        return metric_areas
+
+    @property
     def default_metrics(self):
         analysis_data = self.results_data.get("v3", {}) if self.results_data else {}
         other_metrics = analysis_data.get("other_metrics", {})
@@ -1301,40 +1310,28 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
         return default_metrics
 
-    def get_branch_data(self, analysis_basis, selected_segment):
-        overall_results = (
-            (
-                self.results_data.get("v3", {})
-                .get("overall", {})
-                .get(analysis_basis, {})
-                .get(selected_segment, {})
-            )
-            if self.results_data
-            else {}
-        )
+    def get_branch_data(self, analysis_basis, selected_segment, window="overall"):
+        window_results = self.get_window_results(analysis_basis, selected_segment, window)
 
         branch_data = []
 
-        for branch in self.branches.all().prefetch_related("screenshots"):
+        for branch in self.get_sorted_branches():
             slug = branch.slug
             participant_metrics = (
                 (
-                    overall_results.get(slug, {})
+                    window_results.get(slug, {})
                     .get("branch_data", {})
                     .get("other_metrics", {})
                     .get("identity", {})
                 )
-                if isinstance(overall_results, dict)
+                if isinstance(window_results, dict)
                 else {}
             )
             num_participants = (
                 participant_metrics.get("absolute", {}).get("first", {}).get("point", 0)
             )
 
-            branch_data.insert(
-                0
-                if self.reference_branch and slug == self.reference_branch.slug
-                else len(branch_data),
+            branch_data.append(
                 {
                     "slug": slug,
                     "name": branch.name,
@@ -1346,6 +1343,170 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             )
 
         return branch_data
+
+    def get_metric_area_data(
+        self, metrics, analysis_basis, segment, reference_branch, window="overall"
+    ):
+        metric_data = {}
+        metric_area_data = {"metrics": metrics, "data": metric_data}
+
+        window_results = self.get_window_results(analysis_basis, segment, window)
+
+        for branch in self.get_sorted_branches():
+            branch_results = (
+                window_results.get(branch.slug, {}).get("branch_data", {})
+                if isinstance(window_results, dict)
+                else {}
+            )
+            branch_metrics = {}
+
+            for metric in metrics:
+                slug = metric.get("slug")
+                group = metric.get("group")
+
+                metric_src = branch_results.get(group, {}).get(slug, {})
+
+                absolute_data_list = metric_src.get("absolute", {}).get("all", [])
+                relative_data_list = (
+                    metric_src.get("relative_uplift", {})
+                    .get(reference_branch, {})
+                    .get("all", [])
+                )
+
+                def formatted_analysis_point_comparator(point):
+                    wi = point.get("window_index")
+                    return int(wi) if wi is not None else 0
+
+                absolute_data_list.sort(key=formatted_analysis_point_comparator)
+                relative_data_list.sort(key=formatted_analysis_point_comparator)
+
+                significance_map = (
+                    metric_src.get("significance", {})
+                    .get(reference_branch, {})
+                    .get(window, {})
+                )
+
+                abs_entries = []
+                for i, data_point in enumerate(absolute_data_list):
+                    lower = round(data_point.get("lower"), 2)
+                    upper = round(data_point.get("upper"), 2)
+
+                    if metric.get("display_type") == "percentage":
+                        lower = f"{lower * 100}%"
+                        upper = f"{upper * 100}%"
+
+                    significance = significance_map.get(str(i + 1), "neutral")
+                    abs_entries.append(
+                        {"lower": lower, "upper": upper, "significance": significance}
+                    )
+
+                rel_entries = []
+                for i, data_point in enumerate(relative_data_list):
+                    lower = data_point.get("lower")
+                    upper = data_point.get("upper")
+                    avg_rel_change = abs(data_point.get("point") * 100)
+                    significance = significance_map.get(str(i + 1), "neutral")
+                    rel_entries.append(
+                        {
+                            "lower": lower,
+                            "upper": upper,
+                            "significance": significance,
+                            "avg_rel_change": avg_rel_change,
+                        }
+                    )
+
+                branch_metrics[slug] = {"absolute": abs_entries, "relative": rel_entries}
+
+            metric_data[branch.slug] = branch_metrics
+
+        return metric_area_data
+
+    def get_kpi_metrics(
+        self, analysis_basis, segment, reference_branch, window="overall"
+    ):
+        kpi_metrics = NimbusConstants.KPI_METRICS.copy()
+        window_results = self.get_window_results(analysis_basis, segment, window)
+        other_metrics = (
+            (
+                window_results.get(reference_branch, {})
+                .get("branch_data", {})
+                .get("other_metrics", {})
+            )
+            if isinstance(window_results, dict)
+            else {}
+        )
+
+        if NimbusConstants.DAILY_ACTIVE_USERS in other_metrics:
+            diff_metrics = other_metrics.get(NimbusConstants.DAILY_ACTIVE_USERS, {}).get(
+                "difference", {}
+            )
+            for branch in diff_metrics:
+                if len(diff_metrics.get(branch, {}).get("all", [])) > 0:
+                    kpi_metrics.append(NimbusConstants.DAU_METRIC)
+                    break
+        elif NimbusConstants.DAYS_OF_USE in other_metrics:
+            if (
+                len(
+                    other_metrics.get(NimbusConstants.DAYS_OF_USE, {})
+                    .get("absolute", {})
+                    .get("all", [])
+                )
+                > 0
+            ):
+                kpi_metrics.append(NimbusConstants.DOU_METRIC)
+
+        return kpi_metrics
+
+    def get_metric_data(
+        self, analysis_basis, segment, reference_branch, window="overall"
+    ):
+        metric_areas = self.metric_areas
+        metric_data = {}
+
+        for area in metric_areas:
+            match area:
+                case NimbusConstants.KPI_AREA:
+                    kpi_metric_list = self.get_metric_area_data(
+                        self.get_kpi_metrics(
+                            analysis_basis,
+                            segment,
+                            reference_branch,
+                            window,
+                        ),
+                        analysis_basis,
+                        segment,
+                        reference_branch,
+                        window,
+                    )
+                    metric_data[area] = kpi_metric_list
+                case _:
+                    continue
+
+        return metric_data
+
+    def get_window_results(self, analysis_basis, segment, window="overall"):
+        return (
+            (
+                self.results_data.get("v3", {})
+                .get(window, {})
+                .get(analysis_basis, {})
+                .get(segment, {})
+            )
+            if self.results_data
+            else {}
+        )
+
+    def get_sorted_branches(self):
+        return (
+            [
+                self.reference_branch,
+                *self.branches.exclude(id=self.reference_branch.id),
+            ]
+            if self.reference_branch
+            else []
+            # reference branch is always defined for created experiments, the empty
+            # list is just a fallback for a specific test
+        )
 
     @property
     def experiment_active_status(self):
--- a/experimenter/experimenter/experiments/tests/test_models.py
+++ b/experimenter/experimenter/experiments/tests/test_models.py
@@ -2351,6 +2351,270 @@ class TestNimbusExperiment(TestCase):
         self.assertEqual(second["num_participants"], 75)
         self.assertEqual(second["description"], branch_b.description)
 
+    def test_get_metric_data_returns_correct_data(self):
+        experiment = NimbusExperimentFactory.create()
+        branch_a = NimbusBranchFactory.create(
+            experiment=experiment, name="Branch A", slug="branch-a"
+        )
+        branch_b = NimbusBranchFactory.create(
+            experiment=experiment, name="Branch B", slug="branch-b"
+        )
+
+        experiment.results_data = {
+            "v3": {
+                "overall": {
+                    "enrollments": {
+                        "all": {
+                            "branch-a": {
+                                "branch_data": {
+                                    "other_metrics": {
+                                        "retained": {
+                                            "absolute": {
+                                                "all": [
+                                                    {
+                                                        "lower": 1.49,
+                                                        "upper": 1.74,
+                                                        "point": 1.62,
+                                                    }
+                                                ]
+                                            },
+                                            "relative_uplift": {
+                                                "branch-b": {
+                                                    "all": [
+                                                        {
+                                                            "lower": -0.12,
+                                                            "upper": 0.15,
+                                                            "point": 0.02,
+                                                        }
+                                                    ]
+                                                }
+                                            },
+                                            "significance": {
+                                                "branch-b": {"overall": {"1": "neutral"}}
+                                            },
+                                            "difference": {
+                                                "branch-b": {
+                                                    "all": [
+                                                        {
+                                                            "lower": 0.01,
+                                                            "upper": 0.03,
+                                                            "point": 0.02,
+                                                        }
+                                                    ]
+                                                }
+                                            },
+                                        }
+                                    }
+                                }
+                            },
+                            "branch-b": {
+                                "branch_data": {
+                                    "other_metrics": {
+                                        "retained": {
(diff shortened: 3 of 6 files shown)


moibra05:
image

experimenter/experimenter/experiments/models.py
```diff
@@ -1285,6 +1285,15 @@ def timeline(self):
 
         return timeline_entries
 
+    @property
+    def metric_areas(self):
+        metric_areas = NimbusConstants.DEFAULT_METRIC_AREAS.copy()
+
+        # TODO: check all metrics with data and add their areas to metric_areas if not
+        # already present
```
jaredlockhart: Oh yep good to leave notes like this but make sure to prefile the followup ticket and then include the gh issue number here 🙏

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+                .get("overall", {})
+                .get(analysis_basis, {})
+                .get(segment, {})
+            )
+            if self.results_data
+            else {}
+        )
+
+        for branch in branches:
+            branch_slug = branch.slug
```
jaredlockhart: Lookups like this cost nothing so storing it locally like this doesn't really help anything

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+        wi = point.get("window_index")
+        return int(wi) if wi is not None else 0
+
+    def get_metric_area_data(
+        self, metrics, analysis_basis, segment, reference_branch, window="overall"
+    ):
+        metric_data = {}
+        metric_area_data = {"metrics": metrics, "data": metric_data}
+
+        branches = self.sorted_branches
```
jaredlockhart: I'm guessing `sorted_branches` here is a property method so storing it like this is to prevent having to rerun the computation multiple times?

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+
+    def get_metric_data(
+        self, analysis_basis, segment, reference_branch, window="overall"
+    ):
+        metric_areas = self.metric_areas
+        metric_data = {}
+
+        for area in metric_areas:
+            match area:
+                case "KPI Metrics":
```
jaredlockhart: String constants like this should always be declared somewhere as a constant in a consistent place, otherwise you have to go hunting around for random possibly duplicated strings.

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+
+        return metric_data
+
+    @property
+    def sorted_branches(self):
+        branches = list(self.branches.all().order_by("slug"))
+        if self.reference_branch:
+            branches.remove(self.reference_branch)
+            branches.insert(0, self.reference_branch)
+        return branches
```
jaredlockhart: Ah yeah there it is. 

Branches are default sorted by slug:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/models.py#L2142

And reference_branch will always be not-None by the time the experiment is created.

So you could just do something like

```py
return [self.reference_branch, *self.branches.exclude(id=self.reference_branch.id)]
```

instead of mutating a list.

Also it's mostly just a stylistic choice wehther to make something a property or not, but in the way it's used above, it's confusing to read becuase it's really a method call, which can be expensive/side-effectful, but it looks like a property access, which generally isn't.  So maybe this doesn't need to be a property?  Not a big deal, just something to think about.

experimenter/experimenter/outcomes/__init__.py
```diff
@@ -91,6 +91,13 @@ def all(cls):
     def by_application(cls, application):
         return [o for o in cls.all() if o.application == application]
 
+    @classmethod
+    def get_by_slug(cls, slug):
+        for o in cls.all():
+            if o.slug == slug:
+                return o
```
jaredlockhart: A small thing but pretty much always avoid using single letter variable names, the only time i would use them is within a single comprehension ie `[o for o in outcomes]` but here it should be `outcome`

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+
+                absolute_data_list = metric_src.get("absolute", {}).get("all", [])
+                relative_data_list = (
+                    metric_src.get("relative_uplift", {})
+                    .get(reference_branch, {})
+                    .get("all", [])
+                )
+
+                absolute_data_list.sort(key=self.formattedAnalysisPointComparator)
+                relative_data_list.sort(key=self.formattedAnalysisPointComparator)
```
jaredlockhart: Yeah so this pattern here, that comparator doesn't really need to be a method on the class becuase it doens't use `self`.  It's pretty common to define a small inner function for key/comparator lookups right above where it's used, so I'd expect that function to go just above here.

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
 
         return branch_data
 
+    def formattedAnalysisPointComparator(self, point):
+        wi = point.get("window_index")
+        return int(wi) if wi is not None else 0
+
+    def get_metric_area_data(
```
jaredlockhart, line 1347: Not something to do in this PR but it might make it cleaner/easier to read if we move all these methods up above the class into a little mixin of just the methods/properties for the results, so they're all in one place.  Can do as a followup later.

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+                .get("overall", {})
+                .get(analysis_basis, {})
+                .get(segment, {})
+            )
+            if self.results_data
+            else {}
+        )
+
+        for branch in self.get_sorted_branches():
+            branch_results = (
```
yashikakhurana: +1 to Jared's comment, may be can do optimisation after this like for the above method `get_overall_results`,`get_branch_results`, `get_metrics_result`

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+        return metric_data
+
+    def get_sorted_branches(self):
+        return (
+            [
+                self.reference_branch,
+                *self.branches.exclude(id=self.reference_branch.id),
+            ]
+            if self.reference_branch
+            else []
```
yashikakhurana, line 1506: is this intentional? get_sorted_branches() returns [] if there’s no reference_branch. That will make branch_data empty and all metric areas empty

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+                significance_map = (
+                    metric_src.get("significance", {})
+                    .get(reference_branch, {})
+                    .get(window, {})
+                )
+
+                abs_entries = []
+                for i, data_point in enumerate(absolute_data_list):
+                    lower = round(data_point.get("lower"), 2)
+                    upper = round(data_point.get("upper"), 2)
```
yashikakhurana, line 1392: can lower upper be None ever, if yes then round(None, 2) will raise if lower/upper are missing in result points.

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+                metric_src = branch_results.get(group, {}).get(slug, {})
+
+                absolute_data_list = metric_src.get("absolute", {}).get("all", [])
+                relative_data_list = (
+                    metric_src.get("relative_uplift", {})
+                    .get(reference_branch, {})
+                    .get("all", [])
+                )
+
+                def formattedAnalysisPointComparator(point):
```
yashikakhurana: can we use snake cashing here?

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+                        analysis_basis,
+                        segment,
+                        reference_branch,
+                        window,
+                    )
+                    metric_data[area] = kpi_metric_list
+                case _:
+                    continue
+
+        return metric_data
```
mikewilli, line 1485: It's late so entirely possible I'm missing something obvious, but it seems like this could be simplified to the following:
```suggestion
        metric_data = {}

        metric_data[NimbusConstants.KPI_AREA] = self.get_metric_area_data(
            NimbusConstants.KPI_METRICS,
            analysis_basis,
            segment,
            reference_branch,
            window,
        )

        return metric_data
```

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1353,124 @@ def get_branch_data(self, analysis_basis, selected_segment):
 
         return branch_data
 
+    def get_metric_area_data(
+        self, metrics, analysis_basis, segment, reference_branch, window="overall"
```
mikewilli, line 1348: I'd expect `window` to indicate what the caller wants to retrieve, but it seems like we're retrieving `overall` results regardless of what it's set to. The key into `significance` for the data we want to retrieve will always match the current window since that's what's being compared, so we should use this parameter to determine both the data to retrieve and the key in the significance map.

experimenter/experimenter/experiments/models.py
mikewilli, line 1392: We got a lot of feedback about showing more decimal points, so we added the ability to see more than 2 decimals on hover (we show 6). Can we keep that functionality? Not a blocker, maybe a follow-up PR?

experimenter/experimenter/experiments/constants.py
```diff
@@ -994,6 +994,38 @@ class FirefoxLabsGroups(models.TextChoices):
...
+            "friendly_name": "Daily Active Users",
+            "slug": "client_level_daily_active_users_v2",
+            "description": "Daily active users per client.",
+        },
+        {
+            "group": "other_metrics",
+            "friendly_name": "Days of Use",
+            "slug": "days_of_use",
+            "description": "Number of days the browser was used.",
+        },
```
mikewilli: `days_of_use` (DOU) isn't a KPI anymore, but there are older experiments that don't have data for the replacement (`client_level_daily_active_users_v2` or DAU). If I remember right, we kept it in the list so we could dynamically choose to fall back to it for these older experiment. I also think we decided we should not be showing DOU if we have DAU so as to not cause confusion. 

Ideally we'd do the same here, but not sure what the complexity of something like that would be. If that's not reasonable, a potential fallback option for now could be to show DAU in the highlighted/KPI metrics and then DOU alongside other non-KPI metrics.

experimenter/experimenter/experiments/constants.py
```diff
@@ -994,6 +994,38 @@ class FirefoxLabsGroups(models.TextChoices):
...
+        {
+            "group": "other_metrics",
+            "friendly_name": "Retention",
+            "slug": "retained",
+            "description": "Whether the user is retained over time.",
```
mikewilli: We should be clear that this is 2-week retention in both the `friendly_name` and `description` (see my other comment about where we're getting this info from).

experimenter/experimenter/experiments/constants.py
```diff
@@ -994,6 +994,38 @@ class FirefoxLabsGroups(models.TextChoices):
...
+    KPI_AREA = "KPI Metrics"
+    NOTABLE_CHANGES_AREA = "Notable Changes"
+    DEFAULT_METRIC_AREAS = [NOTABLE_CHANGES_AREA, KPI_AREA]
+
+    KPI_METRICS = [
+        {
+            "group": "other_metrics",
+            "friendly_name": "Daily Active Users",
+            "slug": "client_level_daily_active_users_v2",
+            "description": "Daily active users per client.",
```
mikewilli: What are these descriptions being used for, and where are they coming from? Can we re-use the constants we already have defined in experimenter/experimenter/results/src/lib/visualization/constants.ts? I think the descriptions could be the ones we have in `METRICS_TIPS` for example.

mikewilli: Even better would be to pull descriptions directly from the metadata we get from metric-hub via Jetstream's metadata output, but that's definitely a nice-to-have we can worry about later.

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2351,6 +2351,92 @@ def test_get_branch_data_returns_correct_data(self):
...
+    def test_get_metric_data_returns_correct_data(self):
+        experiment = NimbusExperimentFactory.create()
+        branch_a = NimbusBranchFactory.create(
+            experiment=experiment, name="Branch A", slug="branch-a"
+        )
+        branch_b = NimbusBranchFactory.create(
+            experiment=experiment, name="Branch B", slug="branch-b"
+        )
+
+        experiment.results_data = {
```
mikewilli, line 2363: It looks like we're changing a bit of what we care about in the UI vs what the client code is getting, which is fine (good, even!). However, I think the test data here should match the format we expect from the jetstream client code (e.g., includes `difference` in the metric data, each point includes `point`, extra keys that the client includes for schema flexibility, etc.). Then we can check more completely whether the data getter is doing what we expect.

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2351,6 +2351,92 @@ def test_get_branch_data_returns_correct_data(self):
...
+                        "all": {
+                            "branch-a": {
+                                "branch_data": {
+                                    "other_metrics": {
+                                        "retained": {
+                                            "absolute": {
+                                                "all": [{"lower": 1.49, "upper": 1.74}]
+                                            },
+                                            "relative_uplift": {
+                                                "control": {
```
mikewilli: The key here represents which branch is being compared to the current branch (`"branch-a"`), so based on the branches in the experiment, this should be `"branch-b"`.
```suggestion
                                                "branch-b": {
```

mikewilli: Same goes for the other places where you have `control`.

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2351,6 +2351,92 @@ def test_get_branch_data_returns_correct_data(self):
...
+            [{"lower": -0.12, "upper": 0.15, "significance": "neutral", "interval": 2}],
+        )
+
+        self.assertEqual(
+            second_data.get("retained").get("absolute"),
+            [{"lower": "124.0%", "upper": "163.0%", "significance": "positive"}],
+        )
+        self.assertEqual(
+            second_data.get("retained").get("relative"),
+            [{"lower": -0.3, "upper": 0.55, "significance": "positive", "interval": 13}],
```
mikewilli: It's not clear to me what the `interval` field represents. I can see how it's calculated but why is it needed?

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2351,6 +2351,92 @@ def test_get_branch_data_returns_correct_data(self):
...
+        first_data = data.get(branch_a.slug, {})
+        second_data = data.get(branch_b.slug, {})
+
+        self.assertEqual(
+            first_data.get("retained").get("absolute"),
+            [{"lower": "149.0%", "upper": "174.0%", "significance": "neutral"}],
+        )
+        self.assertEqual(
+            first_data.get("retained").get("relative"),
+            [{"lower": -0.12, "upper": 0.15, "significance": "neutral", "interval": 2}],
```
mikewilli: `overall` data always has a single window, but how do we handle weekly and other data that might have multiple windows with a significance that doesn't include that info? Are we retrieving a single window index at a time via the UI? If so, shouldn't that be a parameter to `get_metric_data`?

experimenter/experimenter/experiments/models.py
moibra05: yes, that check is only there to satisfy a test case where `get_sorted_branches` is called on an undefined experiment  I've learned from jared that reference_branch always exists once an experiment is created so in prod I dont think `get_sorted_branches` should ever return an empty list

experimenter/experimenter/experiments/models.py
moibra05: it looks like they can be None according to [this schema](https://github.com/mozilla/experimenter/blob/main/schemas/mozilla_nimbus_schemas/jetstream/statistics.py#L24-L25) but in the [old react code](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/results/src/components/PageResults/TableVisualizationRow/index.tsx#L278-L331) i based mine on, `lower` and `upper` are guaranteed to be non-null. I'm not sure what ensures this, maybe @mikewilli or @jaredlockhart might know?

experimenter/experimenter/experiments/models.py
mikewilli: Since this is sort of odd, special-case behavior, it's probably worth a comment referencing the test.

experimenter/experimenter/experiments/models.py
mikewilli: The schema allows for null values because the population count metric is only a `point` value and doesn't have confidence bounds. I believe that in all other cases they should be non-null.

experimenter/experimenter/outcomes/__init__.py
```diff
@@ -91,6 +91,13 @@ def all(cls):
     def by_application(cls, application):
         return [o for o in cls.all() if o.application == application]
 
+    @classmethod
+    def get_by_slug(cls, slug):
```
mikewilli: This (and the related test) seem unrelated to the rest of the PR, should they be left for a future PR?

mikewilli (changes requested):
This is a great start. I tried to be thorough with my comments so that we don't miss anything important, but I also want to clarify that many of them can be addressed in the future. Obviously the new Results page won't get done in a single PR!

On a related note, many of my comments are related to the cases where we're very narrow in what we're looking at here (e.g., `overall` only), so please feel free to ignore those if the plan is to start narrow and go broader/more generic in upcoming work.

experimenter/experimenter/experiments/models.py
moibra05: `metric_areas` is meant to be a list of all the groups metrics associated with an experiment fall under. at the moment the only metric area defined are KPIs since they're constant for every experiment. the intention with this is that in the future once i figure out a way to group other metrics by areas they may require some specific prerequisite steps before i can just call `self.get_metric_area_data`

experimenter/experimenter/experiments/models.py
moibra05: yea, i wanted to leave rounding numbers as a responsibility for the frontend but I couldnt get it to work nicely with metrics that had a `display_type` of percentage. i'll make a note to fix this in #13943

experimenter/experimenter/experiments/constants.py
moibra05: metrics can be clicked where a popout appears and shows the metric description 

image

moibra05: also if i remember correctly i took them from the outcomes config

experimenter/experimenter/experiments/tests/test_models.py
moibra05: interval probably isnt the best name but its the average percent of the relevant changes shown here

image

experimenter/experimenter/outcomes/__init__.py
moibra05: i actually deleted the code that was using that method but forgot to got rid of it as well, ty :)

experimenter/experimenter/experiments/models.py
mikewilli: Makes sense, thanks!

experimenter/experimenter/experiments/constants.py
mikewilli: Thanks for the info. Can we switch to using the ones from `METRICS_TIPS`?

experimenter/experimenter/experiments/tests/test_models.py
mikewilli: Why are we computing an average increase here instead of showing the range? I'm sure you're just following the design so maybe this is a better discussion to be had elsewhere. I'm just concerned that it could be misleading, especially for very broad ranges.

experimenter/experimenter/experiments/models.py
mikewilli: Alright sounds good, as long as we have it on the todo list then no need to worry about it immediately.

experimenter/experimenter/experiments/tests/test_models.py
moibra05: the object returned by `get_metric_area_data` contains a list of objects with a lower, upper, significance, and avg_rel_change field. for overall results this list contains only 1 item but when weekly metrics are requested then there are just multiple of these objects for each available week in the list

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1344,170 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+                    significance = significance_map.get(str(i + 1), "neutral")
+                    abs_entries.append(
+                        {"lower": lower, "upper": upper, "significance": significance}
+                    )
+
+                rel_entries = []
+                for i, data_point in enumerate(relative_data_list):
+                    lower = data_point.get("lower")
+                    upper = data_point.get("upper")
+                    avg_rel_change = abs(round((lower + upper) / 2 * 100))
```
mikewilli: I think we can just get the `point` value from the results here, no? And then multiply by 100 if you want a percentage.

experimenter/experimenter/experiments/constants.py
```diff
@@ -994,6 +994,45 @@ class FirefoxLabsGroups(models.TextChoices):
...
     """  # noqa
 
+    KPI_AREA = "KPI Metrics"
+    NOTABLE_CHANGES_AREA = "Notable Changes"
+    DEFAULT_METRIC_AREAS = [NOTABLE_CHANGES_AREA, KPI_AREA]
+
+    DAILY_ACTIVE_USERS = "client_level_daily_active_users_v2"
+    DAYS_OF_USE = "days_of_use"
+    RETENTION = "retained"
+    SEARCH_COUNT = "search_count"
```
mikewilli, line 1004: These are defined in an enum in the Jetstream schemas already if you want to pull them from there: https://github.com/mozilla/experimenter/blob/80fef74d645aa670678956d6ee4127f16f561072/experimenter/experimenter/jetstream/models.py#L29-L34

experimenter/experimenter/experiments/models.py
```diff
@@ -1347,6 +1344,170 @@ def get_branch_data(self, analysis_basis, selected_segment):
...
+    def get_sorted_branches(self):
+        return (
+            [
+                self.reference_branch,
+                *self.branches.exclude(id=self.reference_branch.id),
+            ]
+            if self.reference_branch
+            else []
+            # reference branch is always defined for created experiments, the empty
+            # list is just a fallback for a specific test
```
mikewilli, line 1508: Can you reference the test by path/name here so a future person can know if this special case continues to be relevant?

experimenter/experimenter/experiments/models.py
yashikakhurana: is there any default value we can use then?

mikewilli: I think I'd probably set it to `None` and log an error if it's a metric where we expect non-null bounds.

experimenter/experimenter/experiments/tests/test_models.py
mikewilli: A couple other things to expect in the `results_data` are all of the branches appearing inside `relative_uplift`, `significance`, and `difference`, and `weekly` appearing inside significance. You can worry about the `weekly` significance when you add other analysis periods if you want, but the other branches should probably be in here. They'll be empty for the wrong combinations (branch-a won't have relative_uplift/difference/significance data comparing to branch-a), but it seems like a possible point of failure if something changes in the parsing.

mikewilli (approved):
I left a few comments, but I don't want to block you on anything I've suggested since this is meant to be iterative and it's in preview. I'll leave it up to you whether you want to address things here or in future PRs.

PR #13933, opened 2025-11-10
chore(nimbus): style bootstrap-select disabled consistently

### LINKED ISSUE #12908: [HTMX UI] Add CSS to grey out disabled bootstrap-select
Looks like when you disable a bootstrap-select it greys out the text but not the background which makes it look inconsistent with the other bootstrap disabled fields:

Image

Let's see if we can add a CSS rule to grey them out consistently.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5618)

jaredlockhart:
Because

* bootstrap-select has its own styling that does not always match bootstrap5
* The disabled state did not match

This commit

* Overrides bootstrap-select disabled styling to match the default bootstrap5 disabled select styling

fixes #12908

### DIFF
--- a/experimenter/experimenter/nimbus_ui/static/css/style.scss
+++ b/experimenter/experimenter/nimbus_ui/static/css/style.scss
@@ -50,6 +50,24 @@
     min-height: 1.5em;
     line-height: 1.3;
   }
+
+  &.disabled {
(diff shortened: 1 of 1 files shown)


jaredlockhart:
image
image

PR #13936, opened 2025-11-10
feat(nimbus): sort versions semantically

### LINKED ISSUE #8030: Versions on list page are sorted lexicographically
The versions columns on the list page are not being sorted in verison order but rather [lexicographically](https://en.wikipedia.org/wiki/Lexicographic_order) which doesn't work well for version sorting.  Let's try to figure out a way to actually get them sorted properly by version number.  May try to parse them into numbers?

jaredlockhart:
Becuase

* We have version sorting on the home and list pages
* We've been sorting them lexigraphically which does not work with semantic versioning
* We should sort them semantically

This commit

* Updates the home and list filters to sort versions semantically

fixes #8030

### DIFF
--- a/experimenter/experimenter/nimbus_ui/filtersets.py
+++ b/experimenter/experimenter/nimbus_ui/filtersets.py
@@ -14,6 +14,22 @@ from experimenter.nimbus_ui.forms import MultiSelectWidget
 from experimenter.targeting.constants import TargetingConstants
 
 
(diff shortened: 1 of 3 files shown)


yashikakhurana (approved):
Tested locally, was able to sort semantically, I always like the good use of mixin. Thanks @jaredlockhart

PR #13957, opened 2025-11-13
chore(nimbus): use rust jexl evaluator in all tests

### LINKED ISSUE #13956: Validate all JEXL with Rust evaluator
We recently saw a case where we introduced a JEXL expression like `!(expr)` and all tests passed but we discovered in the client that the Rust evaluator failed to evaluate.  We currently validate all JEXL expressions using the Python evaluator, but let's also use the Rust evaluator in every JEXL validation so we don't accidentally introduce a regression like this again.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6182)

jaredlockhart:
Becuase

* We recently saw a case where we added a jexl expression that was valid in desktop but not in the sdk
* The tests used the python jexl evaluator which does not exactly match the behaviur of the rust one used in the sdk
* We have access to the rust evaluator so we can use it to validate jexl expressions in all the tests

This commit

* Adds a validate jexl function that uses both the python evaluator and optionally the rust evaluator for non desktop expressions
* Updates all the jexl validation tests to use it
* This was checked and would have caught the recent regression

fixes #13956

### DIFF
--- a/experimenter/experimenter/experiments/tests/jexl_utils.py
+++ b/experimenter/experimenter/experiments/tests/jexl_utils.py
@@ -1,0 +1,79 @@
+import json
+from pathlib import Path
+from unittest import mock
+
+from nimbus_megazord import nimbus as nimbus_rust
+
+from experimenter.experiments.constants import Application
+from experimenter.experiments.tests import JEXLParser
+
+
+def validate_jexl_expr(expression, application):
+    JEXLParser().parse(expression)
+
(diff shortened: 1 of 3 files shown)


jaredlockhart:
Blocked on https://github.com/mozilla/experimenter/pull/13948

yashikakhurana:
@jaredlockhart oh you got that 💯 , I started looking at this morning, Thank you 🙏

PR #13969, opened 2025-11-17
feat(nimbus): create popout for metrics

### LINKED ISSUE #13943: Add metric popouts (deep dives and weekly results) to the new results page
Users need to be able to dive into metrics and see the absolute values as well as the relative changes. This includes a weekly breakdown

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6045)

moibra05:
Because

- Metric have in depth information that can be accessed by clicking their cards

This commit

- Creates the metric popout view displaying overview data

Fixes #13943

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1339,14 +1339,10 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         for branch in self.get_sorted_branches():
             slug = branch.slug
             participant_metrics = (
-                (
-                    window_results.get(slug, {})
-                    .get("branch_data", {})
-                    .get("other_metrics", {})
-                    .get("identity", {})
-                )
-                if isinstance(window_results, dict)
-                else {}
+                window_results.get(slug, {})
+                .get("branch_data", {})
+                .get("other_metrics", {})
+                .get("identity", {})
             )
             num_participants = (
                 participant_metrics.get("absolute", {}).get("first", {}).get("point", 0)
@@ -1373,20 +1369,17 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
         window_results = self.get_window_results(analysis_basis, segment, window)
 
-        for branch in self.get_sorted_branches():
-            branch_results = (
-                window_results.get(branch.slug, {}).get("branch_data", {})
-                if isinstance(window_results, dict)
-                else {}
-            )
+        for metric in metrics:
+            slug = metric.get("slug")
+            group = metric.get("group")
             branch_metrics = {}
 
-            for metric in metrics:
-                slug = metric.get("slug")
-                group = metric.get("group")
+            for branch in self.get_sorted_branches():
+                branch_results = window_results.get(branch.slug, {}).get(
+                    "branch_data", {}
+                )
 
                 metric_src = branch_results.get(group, {}).get(slug, {})
-
                 absolute_data_list = metric_src.get("absolute", {}).get("all", [])
                 relative_data_list = (
                     metric_src.get("relative_uplift", {})
@@ -1409,12 +1402,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
                 abs_entries = []
                 for i, data_point in enumerate(absolute_data_list):
-                    lower = round(data_point.get("lower"), 2)
-                    upper = round(data_point.get("upper"), 2)
-
-                    if metric.get("display_type") == "percentage":
-                        lower = f"{lower * 100}%"
-                        upper = f"{upper * 100}%"
+                    lower = data_point.get("lower")
+                    upper = data_point.get("upper")
 
                     significance = significance_map.get(str(i + 1), "neutral")
                     abs_entries.append(
@@ -1425,7 +1414,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                 for i, data_point in enumerate(relative_data_list):
                     lower = data_point.get("lower")
                     upper = data_point.get("upper")
-                    avg_rel_change = abs(data_point.get("point") * 100)
+                    avg_rel_change = abs(data_point.get("point"))
                     significance = significance_map.get(str(i + 1), "neutral")
                     rel_entries.append(
                         {
@@ -1436,9 +1425,12 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                         }
                     )
 
-                branch_metrics[slug] = {"absolute": abs_entries, "relative": rel_entries}
+                branch_metrics[branch.slug] = {
+                    "absolute": abs_entries,
+                    "relative": rel_entries,
+                }
 
-            metric_data[branch.slug] = branch_metrics
+            metric_data[slug] = branch_metrics
 
         return metric_area_data
 
@@ -1516,6 +1508,35 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             if self.results_data
             else {}
         )
+
+    def get_max_metric_value(
+        self,
+        analysis_basis,
+        segment,
+        reference_branch,
+        outcome_group,
+        outcome_slug,
+    ):
+        overall_results = self.get_window_results(analysis_basis, segment, "overall")
+        max_value = 0
+
+        for branch in self.get_sorted_branches():
+            if overall_results:
+                for data_point in (
+                    overall_results.get(branch.slug, {})
+                    .get("branch_data", {})
+                    .get(outcome_group, {})
+                    .get(outcome_slug, {})
+                    .get("relative_uplift", {})
+                    .get(reference_branch, {})
+                    .get("all", [])
+                ):
+                    lower = data_point.get("lower")
+                    upper = data_point.get("upper")
+
+                    max_value = max(max_value, abs(lower), abs(upper))
+
+        return max_value
 
     def get_sorted_branches(self):
         return (
--- a/experimenter/experimenter/nimbus_ui/constants.py
+++ b/experimenter/experimenter/nimbus_ui/constants.py
@@ -394,3 +394,7 @@ QA_ICON_FILTER_TYPE = "qa_icon_info"
 CHANNEL_ICON_FILTER_TYPE = "channel_icon_info"
 APPLICATION_ICON_FILTER_TYPE = "application_icon_info"
 STATUS_ICON_FILTER_TYPE = "status_icon_info"
+
+# Represents the minimum width percentage for the bar in metric popout cards to
+# adjust number bounds and prevent overlapping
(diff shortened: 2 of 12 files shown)


moibra05:
image
image

experimenter/experimenter/nimbus_ui/templates/common/metric_popout_card.html
```diff
@@ -0,0 +1,60 @@
...
+      {% endif %}
+    </div>
+  {% elif type == "relative" %}
+    <div class="card pt-3 mx-5 mb-3 border-0 position-relative overflow-visible"
+         style="height: 75px">
+      <div class="w-100 border-bottom my-auto "></div>
+      {% if branch_relative_ui_properties %}
+        <div class="d-flex flex-row flex-nowrap justify-content-between position-absolute {% if sig == "positive" %}text-success{% elif sig == "negative" %}text-danger{% else %}text-body-tertiary{% endif %}"
+             style="width: {{ branch_relative_ui_properties.boundsWidth }}%;
+                    left: {{ branch_relative_ui_properties.leftBoundsPercent }}%">
```
freshstrangemusic: The little bubble is offset from the actual axis. It would be nice to have it on-axis (the css `transform: translateY(-50%)`) should do it)

Image

experimenter/experimenter/experiments/models.py
```diff
@@ -1496,6 +1498,41 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
             else {}
         )
 
+    def get_extreme_bounds(
```
freshstrangemusic: The name of this function is `get_extreme_bounds` but it only returns a single value.

And specifically the extreme here is just the maximum, so it could be renamed to reference maximum.

Finally, this is on the general Experiment model, so the name should maybe make this clear that this is about metrics/analysis. `get_max_metric_value` ?

experimenter/experimenter/experiments/models.py
```diff
@@ -1496,6 +1498,41 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+        self,
+        analysis_basis,
+        segment,
+        reference_branch,
+        outcome_slug,
+        outcome_group,
```
freshstrangemusic: nit: group should come first because slugs are organized into groups.

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -726,6 +726,62 @@ def get_context_data(self, **kwargs):
...
+                        continue
+                    lower = branch_data.get("relative", [{}])[0].get("lower")
+                    upper = branch_data.get("relative", [{}])[0].get("upper")
+                    confidence_range = round(extreme_bound * 1000) / 10
+                    fullWidth = confidence_range * 2
+                    barWidth = ((upper - lower) / fullWidth) * 100
+                    leftPercent = (abs(lower - confidence_range * -1) / fullWidth) * 100
+                    leftBoundsPercent = leftPercent
+
+                    print(barWidth)
```
freshstrangemusic: leftover debugging?

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -726,6 +726,62 @@ def get_context_data(self, **kwargs):
...
+                        "leftBoundsPercent": leftBoundsPercent - numDigits * 1.5,
+                        "boundsWidth": max(barWidth, NimbusUIConstants.MIN_BOUNDS_WIDTH)
+                        + numDigits * 3,
+                    }
+
+                    relative_change_ui_properties[metric_metadata["slug"]] = (
+                        metric_ui_properties
+                    )
+
+        context["relative_change_ui_properties"] = relative_change_ui_properties
```
freshstrangemusic: I think this could just be `"relative_changes"` or `"relative_metric_changes"` -- the fact that its UI stuff is kinda implicit because this is in nimbus_ui

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -726,6 +726,62 @@ def get_context_data(self, **kwargs):
...
+                    selected_reference_branch,
+                    metric_metadata["slug"],
+                    metric_metadata["group"],
+                )
+
+                for branch_slug, branch_data in data.items():
+                    if len(branch_data.get("relative", [])) == 0:
+                        continue
+                    lower = branch_data.get("relative", [{}])[0].get("lower")
+                    upper = branch_data.get("relative", [{}])[0].get("upper")
```
freshstrangemusic: We can simplify this significantly based on a few things:

1. empty lists coerce to a false value so `if len(x) == 0` is equivalent to `if not x`
2. if we make it past the `continue` statement, then we know len(branch_data["relative"]) >= 1, so we dont have to call .get()

combining these:

```suggestion
                    if not (relative := not branch_data.get("relative")):
                        continue
                        
                    lower = relative[0].get("lower")
                    upper = relative[0].get("upper")
```

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -726,6 +726,62 @@ def get_context_data(self, **kwargs):
...
+                    confidence_range = round(extreme_bound * 1000) / 10
+                    fullWidth = confidence_range * 2
+                    barWidth = ((upper - lower) / fullWidth) * 100
+                    leftPercent = (abs(lower - confidence_range * -1) / fullWidth) * 100
+                    leftBoundsPercent = leftPercent
+
+                    print(barWidth)
+
+                    numDigits = len(str(round(lower, 1)).replace(".", "")) + len(
```
freshstrangemusic: python variables should be `snake_case`

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -726,6 +726,62 @@ def get_context_data(self, **kwargs):
...
+                        str(round(upper, 1)).replace(".", "")
+                    )
+
+                    if barWidth < NimbusUIConstants.MIN_BOUNDS_WIDTH:
+                        leftBoundsPercent -= (
+                            NimbusUIConstants.MIN_BOUNDS_WIDTH - barWidth
+                        ) / 2
+
+                    metric_ui_properties[branch_slug] = {
+                        "barWidth": barWidth,
```
freshstrangemusic: these should all be snake case since they will be accessed through dot notation in templates.

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -180,6 +180,7 @@ class NimbusUIConstants:
         "Note: This feature supports co-enrollment with other experiments/rollouts "
         "for the selected versions."
     )
+    MIN_BOUNDS_WIDTH = 22
```
freshstrangemusic: can you add a comment explaining what this number means and its units?

experimenter/experimenter/experiments/models.py
```diff
@@ -1496,6 +1498,41 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+            if branch_outcome_data:
+                for data_point in (
+                    branch_outcome_data.get("relative_uplift", {})
+                    .get(reference_branch, {})
+                    .get("all", [])
+                ):
+                    lower = data_point.get("lower")
+                    upper = data_point.get("upper")
+
+                    extreme = max(extreme, abs(lower), abs(upper))
```
freshstrangemusic: What are the other types for `overall_results` -- is it just `None` ? if so we can do actually be simplified into a single loop:

```suggestion
            if overall_results:
                for data_point in (
                    overall_results.get(branch.slug, {})
                    .get("branch_data", {})
                    .get(outcome_group, {})
                    .get(outcome_slug, {})
                    .get("relative_uplift", {})
                    .get(reference_branch, {})
                    .get("all", [])
                ):
                    lower = data_point.get("lower")
                    upper = data_point.get("upper")

                    extreme = max(extreme, abs(lower), abs(upper))
```

moibra05: I cant remember where this is defined but on local, experiment results data is a list like this `[1]` which breaks my code if i dont check that its a dictionary first

freshstrangemusic: @mikewilli Why would this ever be just `[1]` instead of `[{...}]` ?

mikewilli: Looks like a bug in the factory that generates test data for results: https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/tests/factories.py#L515

experimenter/experimenter/experiments/models.py
```diff
@@ -1496,6 +1488,35 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+    def get_max_metric_value(
+        self,
+        analysis_basis,
+        segment,
+        reference_branch,
+        outcome_group,
+        outcome_slug,
+    ):
+        overall_results = self.get_window_results(analysis_basis, segment, "overall")
+        extreme = 0
```
freshstrangemusic:
```suggestion
        max_value = 0
```

lets be specific

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -394,3 +394,7 @@ def _get_schema_diff_size_config():
 CHANNEL_ICON_FILTER_TYPE = "channel_icon_info"
 APPLICATION_ICON_FILTER_TYPE = "application_icon_info"
 STATUS_ICON_FILTER_TYPE = "status_icon_info"
+
+# represents the minimum width percentage for the bar in metric popout cards to
+# adjust number bounds and prevent overlapping
```
freshstrangemusic: Sentence casing.

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -394,3 +394,7 @@ def _get_schema_diff_size_config():
...
+MIN_BOUNDS_WIDTH = 22
```
freshstrangemusic:
```suggestion
METRICS_MIN_BOUNDS_WIDTH = 22
```

This is only for metrics so lets make it clear what its for by name.

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -703,6 +704,59 @@ def get_context_data(self, **kwargs):
...
+
+                    if bar_width < METRICS_MIN_BOUNDS_WIDTH:
+                        left_bounds_percent -= (METRICS_MIN_BOUNDS_WIDTH - bar_width) / 2
+
+                    metric_ui_properties[branch_slug] = {
+                        "bar_width": bar_width,
+                        "left_percent": left_percent,
+                        "left_bounds_percent": left_bounds_percent - num_digits * 1.5,
+                        "bounds_width": max(bar_width, METRICS_MIN_BOUNDS_WIDTH)
+                        + num_digits * 3,
```
freshstrangemusic:
```suggestion
                        "bounds_width": (
                            max(bar_width, METRICS_MIN_BOUNDS_WIDTH)
                            + num_digits * 3
                        ),
```

or pull this out to a previous line.

freshstrangemusic (approved):

PR #13979, opened 2025-11-17
feat(nimbus): Improve qa and takeaways views

### LINKED ISSUE #13085: Summary page refactor
Now that we have clearer patterns for using HTMX in general, let’s refactor the summary page forms (takeaways, qa) to more closely resemble the patterns we’ve adopted in the other pages.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-4607)

yashikakhurana:
Because

- We have discovered the pattern where we don't need to pass the flag externally to set the values and enable forms like we do for qa and takeaways form

This commit

- Removes the external flag passing in the views for QA and Takeaways card
- Uses htmx-swap-oob to update the content at two places on change

Fixes #13085

### DIFF
--- a/experimenter/experimenter/nimbus_ui/views.py
+++ b/experimenter/experimenter/nimbus_ui/views.py
@@ -343,13 +343,6 @@ class QAStatusUpdateView(NimbusExperimentViewMixin, RequestFormMixin, UpdateView
     form_class = QAStatusForm
     template_name = "nimbus_experiments/qa_edit_form.html"
 
-    def get(self, request, *args, **kwargs):
-        self.object = self.get_object()
-        form = self.get_form()
-        return render(
-            request, self.template_name, {"form": form, "experiment": self.object}
-        )
-
     def form_valid(self, form):
         super().form_valid(form)
         return render(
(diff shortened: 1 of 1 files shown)


jaredlockhart (dismissed):
Yep overall looks good, looks like there's a select/target mismatch in the qa form:

Image

yashikakhurana:
@jaredlockhart thank you for noticing that, here fixed it 
https://github.com/user-attachments/assets/b50d4252-88ba-4389-a273-951ac465b711

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -347,33 +341,42 @@ def get_context_data(self, **kwargs):
...
-        context.update(experiment_context)
-        context["takeaways_edit_mode"] = True
-        context["takeaways_form"] = form
-        return self.render_to_response(context)
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        form = self.get_form()
+        return render(
+            request, self.template_name, {"form": form, "experiment": self.object}
+        )
```
jaredlockhart: Wait why do we have to do this?

jaredlockhart: If the idea is to get both the object and the form, we should override get_context_data?

jaredlockhart: But shouldn't UpdateView already do that for us?

yashikakhurana: you are 💯 correct, we definitely don't need `get` and as well as `get_context_data` as we can get the data from the mixin we are already inheriting

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -347,33 +341,28 @@ def get_context_data(self, **kwargs):
...
-
-    def get_success_url(self):
-        return reverse("nimbus-ui-detail", kwargs={"slug": self.object.slug})
+    def form_valid(self, form):
+        super().form_valid(form)
+        return render(
+            self.request,
+            "nimbus_experiments/takeaways_card.html",
+            {"experiment": self.object},
+        )
```
jaredlockhart, line 365: Ya exactly, looks great, great job 🎉 🎉 🎉 🎉

PR #13989, opened 2025-11-18
feat(nimbus): add loading spinner to list page

### LINKED ISSUE #13944: Show loading overlay on list page
Let’s add a loading overlay to the list page when filters/sorts are changed

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6178)

jaredlockhart:
Becuase

* Updates on the list page can be very slow
* It's not clear to a user if a request is in flight
* On other pages we put a loading overlay overtop the loading elements
* We don't need to put an overlay on top of the list becuase we don't need to prevent user input
* We can just use a spinner

This commit

* Adds a loading spinner to the list page

fixes #13944


jaredlockhart:
https://github.com/user-attachments/assets/62c93793-2cd3-4733-a991-7a7af812981e

relud (approved):

PR #13991, opened 2025-11-18
feat(nimbus): cache users in NimbusExperimentChangeLogInline

### LINKED ISSUE #13990: Django Experiment Admin has N queries
The changelog inline does `SELECT * FROM USERS` query for each changelog entry. Now that we are generating many more changelogs (one per fetch), the admin page is timing out and becoming unusable.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6192)

freshstrangemusic:
Because:

- the changelog admin inline queries all users for each changelog; and
- we now have many more changelogs than we previously had, which has resulted in the admin page for experiments becoming unusable

this commit:

- computes the list of users once and uses the same list for all
  changelog inlines, reducing the number of User model queries from O(N)
  to O(1).

Fixes #13990

### DIFF
--- a/experimenter/experimenter/experiments/admin.py
+++ b/experimenter/experimenter/experiments/admin.py
@@ -1,4 +1,5 @@
 from typing import Any
+from weakref import WeakKeyDictionary
 
 from django import forms
 from django.contrib import admin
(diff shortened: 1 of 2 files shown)


relud (approved):
should this have tests?

freshstrangemusic:
> should this have tests?

yeah probably.

relud (approved):

PR #14000, opened 2025-11-19
feat(nimbus): show feature schemas on branch page

### LINKED ISSUE #7217: Display feature definition for selected feature on branch page
When a user selects a feature on the branch page we should also display the feature configuration information so users have guidance about what parameters are available.  This would involve exposing the feature manifests through the config API and then displaying them on the branch page.

jaredlockhart:
Becuase

* While editing a feature value, it would be handy to have the schema for reference
* We can put a readonly codemirror of the schema next to each feature value input

This commit

* Adds a showable/hideable feature schema codemirror under every feature value input

fixes #7217


jaredlockhart:
Screenshot 2025-11-19 at 16 18 11
Screenshot 2025-11-19 at 16 18 21

PR #14001, opened 2025-11-19
feat(nimbus): create cards for other experiment metrics

### LINKED ISSUE #13721: Organize metrics into individual cards
Once we have the metric card view locked down, we can split up the metrics up into individual cards grouped by metric area to show related metrics together, as well as pull out notable results into a single card.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6043)

moibra05:
Because

- Results page was only showing kpi metrics
- Cards were not collapsible

This commit

- Displays all the metrics associated with an experiment (kpi, outcomes, defaults)
- Groups metrics by outcomes (for now) and alphabetically
- Adds a "Notable Changes" section where only metrics that have at least once branch with a stat significance change are shown

Fixes #13721

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -4,6 +4,7 @@ import json
 from collections import defaultdict
 from dataclasses import dataclass
 from decimal import Decimal
+from itertools import chain
 from pathlib import Path
 from typing import Any, Optional
 from urllib.parse import urlencode, urljoin
@@ -32,6 +33,7 @@ from experimenter.experiments.constants import (
     TargetingMultipleKintoCollectionsError,
 )
 from experimenter.nimbus_ui.constants import NimbusUIConstants
+from experimenter.outcomes import Outcomes
 from experimenter.projects.models import Project
 from experimenter.targeting.constants import TargetingConstants
 
@@ -1245,7 +1247,8 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             {
                 "title": "All metrics",
                 "subitems": [
-                    {"title": metric} for metric in self.default_metrics.values()
+                    {"title": metric["friendly_name"]}
+                    for metric in self.get_remaining_metrics_metadata()
                 ],
             },
         ]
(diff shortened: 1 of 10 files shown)


moibra05:
image
image

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2615,6 +2596,39 @@ def test_get_kpi_metrics_returns_correct_metrics(
 
         self.assertListEqual(kpi_metrics, expected_kpi_metrics)
 
+    def test_get_defaults_metrics_with_exclusions(self):
+        experiment = NimbusExperimentFactory.create()
+        branch_a = NimbusBranchFactory.create(
+            experiment=experiment, name="Branch A", slug="branch-a"
+        )
```
mikewilli: Is this needed for the test to work? Not a problem exactly just seems unnecessary for what's being tested here, unless it's necessary for the experiment to be valid, or something similar.

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -2615,6 +2596,39 @@ def test_get_kpi_metrics_returns_correct_metrics(
...
+                        "retained": "2 Week Retention",
+                        "search_count": "Search Count",
+                    }
+                },
+            }
+        }
+        branch_a.save()
+        experiment.save()
+
+        default_metrics = experiment.get_default_metrics(["search_count"])
```
mikewilli: It was not intuitive to me that the parameter here is _excluding_ the specified list. Perhaps passing by keyword would be better?

experimenter/experimenter/experiments/models.py
```diff
@@ -1285,30 +1287,101 @@ def timeline(self):
...
+                    and metric not in notable_metrics
+                ):
+                    notable_metrics.append(metric)
 
+        notable_metrics.sort(key=lambda m: m["friendly_name"])
         return metric_areas
 
-    @property
-    def default_metrics(self):
+    def get_default_metrics(self, exclude_slugs=None):
```
mikewilli: Without passing any "exclude" slugs, this will return more than just the default metrics, no? Should it be called something like `get_other_metrics`? Or, depending on how the new UI groups metrics, maybe this could take the UI's group name (maybe an enum in `NimbusUIConstants` defines the possible areas?) and return the metrics for that group?

experimenter/experimenter/experiments/models.py
```diff
@@ -1285,30 +1287,101 @@ def timeline(self):
...
         return timeline_entries
 
-    @property
-    def metric_areas(self):
-        metric_areas = NimbusConstants.DEFAULT_METRIC_AREAS.copy()
+    def metric_areas(self, analysis_basis, segment, reference_branch, window="overall"):
+        metric_areas = {}
+
+        notable_metrics = []
+        metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA] = notable_metrics
```
mikewilli: TODO?

experimenter/experimenter/experiments/models.py
```diff
@@ -1285,30 +1287,101 @@ def timeline(self):
...
+                formatted_metric = {
+                    "slug": metric.slug,
+                    "description": metric.description
+                    if metric.description
+                    else metrics_metadata.get(metric.slug, {}).get("description", ""),
+                    "group": "other_metrics",
+                    "friendly_name": metric.friendly_name
+                    if metric.friendly_name
+                    else metrics_metadata.get(metric.slug, {}).get(
+                        "friendly_name", metric.slug
```
mikewilli: I think there is (was?) a util function that would convert something like `metric_slug` to `Metric Slug` that you could use here? Not a big deal though if you don't see it.

experimenter/experimenter/experiments/models.py
```diff
@@ -1285,30 +1287,101 @@ def timeline(self):
...
+        metrics_metadata = (
+            self.results_data.get("v3", {}).get("metadata", {}).get("metrics")
+            if self.results_data
+            else {}
+        ) or {}
+        all_outcome_metric_slugs = []
+        for slug in self.primary_outcomes + self.secondary_outcomes:
+            outcome = Outcomes.get_by_slug_and_application(slug, self.application)
+            metrics = outcome.metrics if outcome else []
+            outcomes = []
```
mikewilli: I think this should be
```suggestion
            outcome_metrics = []
```

experimenter/experimenter/experiments/models.py
```diff
@@ -1285,30 +1287,101 @@ def timeline(self):
...
-        # if not already present
+        for metrics in metric_areas.values():
+            for metric in metrics:
+                if (
+                    metric_is_notable(metric["slug"], metric["group"])
+                    and metric not in notable_metrics
+                ):
+                    notable_metrics.append(metric)
 
+        notable_metrics.sort(key=lambda m: m["friendly_name"])
```
mikewilli: Do we need `notable_metrics` for something? It seems like we determine what it should contain, sort it, and then return without using it.

experimenter/experimenter/experiments/models.py
mikewilli: Also consider naming this something that is clear that it returns metadata and not results.

experimenter/experimenter/experiments/models.py
moibra05: No, the reason why I have the notable changes section defined earlier than when its actually initialized is because the order in which i insert entries into `metric_areas` is the same order that the metric areas are rendered on the ui. since notable metrics is meant to be the first section i insert it into `metric_areas` before anything else, but b/c its dependent on metrics in the following areas i can only start to append elements to `notable_metrics` after doing so for every other area

experimenter/experimenter/experiments/models.py
moibra05: this is the same `notable_metrics` that was defined earlier and attached to `metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA]`. i left my justifications for this decision in a previous comment

experimenter/experimenter/experiments/models.py
mikewilli: It seems a bit confusing to me to rely on the fact that python doesn't make a copy of `notable_metrics` and so mutating it later mutates it for this assignment as well. Wouldn't it be clearer to initialize the section here (`metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA] = []`) and then later append directly to this? You could also just initialize the whole dict with empty sections in the order you want and then append to them later.

As an alternative, you could consider defining the section order separately (maybe alongside the other constants) so that you don't have to rely on the insertion order here.

experimenter/experimenter/experiments/models.py
mikewilli: Gotcha, I'll mark this resolved and we can continue discussions there as needed.

experimenter/experimenter/experiments/models.py
moibra05: i like the idea of initializing the whole dict with empty sections and referencing them explicitly later - will do that instead, ty 🙏

--EDIT
after looking again, since the outcome areas are initialized dynamically, it wouldnt make much sense to get all the outcomes once to initialize empty lists then a second time to actually get the metrics; so, i've instead opted to just initialize the notable metrics as you suggested

experimenter/experimenter/experiments/models.py
```diff
@@ -1306,30 +1308,108 @@ def timeline(self):
...
+            analysis_basis, segment, reference_branch, window
+        )
+
+        metrics_metadata = (
+            self.results_data.get("v3", {}).get("metadata", {}).get("metrics")
+            if self.results_data
+            else {}
+        ) or {}
+        all_outcome_metric_slugs = []
+        for slug in self.primary_outcomes + self.secondary_outcomes:
```
freshstrangemusic:
```suggestion
        for slug in itertools.chain(self.primary_outcomes, self.secondary_outcomes):
```

if we use `itertools.chain()` we don't actually construct a new list.

experimenter/experimenter/experiments/models.py
```diff
@@ -1306,30 +1308,108 @@ def timeline(self):
...
+        metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA] = []
+        metric_areas[NimbusUIConstants.KPI_AREA] = self.get_kpi_metrics(
+            analysis_basis, segment, reference_branch, window
+        )
+
+        metrics_metadata = (
+            self.results_data.get("v3", {}).get("metadata", {}).get("metrics")
+            if self.results_data
+            else {}
+        ) or {}
```
freshstrangemusic: this is complicated enough that a ternary is hard to read.

```suggestion
        metrics_metadata = {}
        if self.results_data:
            metrics_metadata = (
                self.results_data.get("v3", {})
                .get("metadata", {})
                .get("metrics", {})
            )
```

experimenter/experimenter/experiments/models.py
```diff
@@ -1306,30 +1308,108 @@ def timeline(self):
...
+
+        metric_areas[NimbusUIConstants.OTHER_METRICS_AREA] = (
+            self.get_remaining_metrics_metadata(exclude_slugs=all_outcome_metric_slugs)
+        )
 
-        # TODO(13721): check all metrics with data and add their areas to metric_areas
-        # if not already present
+        window_results = self.get_window_results(analysis_basis, segment, window)
+
+        def metric_is_notable(slug, group):
```
freshstrangemusic:
```suggestion
        def is_metric_notable(slug, group) -> bool:
```

boolean returning functions should be named like questions

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -145,6 +145,12 @@ class NimbusUIConstants:
         "Next steps",
         "Project impact",
     ]
+    KPI_AREA = "KPI Metrics"
+    OTHER_METRICS_AREA = "Other Metrics"
+    NOTABLE_METRIC_AREA = "Notable Changes"
+    NOTABLE_METRIC_AREA_SUBHEADER = (
+        "All statistically significant changes that have occurred in the experiment"
+    )
```
freshstrangemusic: This doesn't need to be a constant. This can be in the template.

experimenter/experimenter/experiments/models.py
```diff
@@ -1308,30 +1311,109 @@ def timeline(self):
...
-        metric_areas = NimbusConstants.DEFAULT_METRIC_AREAS.copy()
+    def get_metric_areas(
+        self, analysis_basis, segment, reference_branch, window="overall"
+    ):
+        metric_areas = {}
+
+        metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA] = []
+        metric_areas[NimbusUIConstants.KPI_AREA] = self.get_kpi_metrics(
+            analysis_basis, segment, reference_branch, window
+        )
```
freshstrangemusic:
```suggestion
        metric_areas = {
            NimbusUIConstants.NOTABLE_METRIC_AREA: [],
            NimbusUIConstants.KPI_AREA: self.get_kpi_metrics(
                analysis_basis, segment, reference_branch, window
            )
        }
```

PR #14012, opened 2025-11-20
feat(nimbus): Update advanced targeting for OLD_SIDEBAR_USERS to exclude users without a sidebar-button

nikk15:
Because

- We need to exclude users without a sidebar button in their toolbar as it is needed to anchor the callout.

This commit

- Updates advanced targeting for OLD_SIDEBAR_USERS_V2


freshstrangemusic:
Please `make code_format` or, if you have pipx installed, `pipx run ruff -- format ./experimenter/experimenter/targeting/constants.py`

PR #14016, opened 2025-11-21
bug(nimbus): version sorting performance

### LINKED ISSUE #14011: Bug: Version sort hanging
We recently updated the version sorting to account for semantic version sorting, but it's taking too long and just hanging.  So let's figure out a better way to do it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6202)

jaredlockhart:
Because

* We changed the version sorting on the home and list pages to do semantic version sorting
* We did it in memory but it's too slow for production and might have been causing OOM exits
* We can do it in the db if we store a parsed tuple of the major/minor/patch values

This commit

* Adds a new field to store the parsed version
* Migrates existing experiments to populate the new field
* Updates it on save
* Changes the filter sorts to use the parsed field

fixes #14011


yashikakhurana (approved):
Tested locally, nice approach as adding the field and parsing it. Thanks @jaredlockhart

PR #14081, opened 2025-11-27
feat(nimbus): Improve desktop targeting tests

### LINKED ISSUE #14073: Improve Desktop targeting test runtime
There are several things that could be improved with the desktop targeting tests:

* The filter_expression.js file is opened and read from disk in each test (targeting slug) instead of being cached ([https://github.com/mozilla/experimenter/blob/1e74d9ecf77a83b7f173423bff94111a3eb905c9/experimenter/tests/integration/nimbus/test_desktop_targeting.py#L17-L20](https://github.com/mozilla/experimenter/blob/1e74d9ecf77a83b7f173423bff94111a3eb905c9/experimenter/tests/integration/nimbus/test_desktop_targeting.py#L17-L20)) This can be solved by using a `@functools.cache` decorator on the pixture after the `@pytest.fixture` decorator (e.g., [https://github.com/mozilla/experimenter/blob/1e74d9ecf77a83b7f173423bff94111a3eb905c9/schemas/mozilla_nimbus_schemas/tests/experiments_v7/test_experiments_v7.py#L17-L20](https://github.com/mozilla/experimenter/blob/1e74d9ecf77a83b7f173423bff94111a3eb905c9/schemas/mozilla_nimbus_schemas/tests/experiments_v7/test_experiments_v7.py#L17-L20|smart-link) )
* `test_check_audience_targeting` and `test_check_advanced_targeting` may be opening an `about:blank` tab for each test (`selenium.get("about:blank")`). We can open a single page at the start of the test by using an autouse fixture, e.g.,

```
@pytest.fixture(autouse=True)
def open_about_blank():
    selenium.get("about:blank")
    # other set up code
    yield
    # tear down code
```

* `filter_expression.js` is awaiting `TelemetryEnvironment.onInitialized()` and `ExperimentAPI.ready()` when evaluating each slug. We can do this in a separate script executed once at the beginning of the test, e.g., in the `open_about_blank()` proposed above.
* We are serializing async execution of all of these targeting expressions. I can imagine a way where we pass all of the targeting expressions at once and do `const results = await Promise.all(targetingSlugs.map(t => targetingContext.evalWithDefault(t)))` and report the result of each member of `results` as a single test.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6231)

yashikakhurana:
## Because

The desktop targeting tests were slow due to:
- Creating experiments via GraphQL for each targeting configuration
- Reading filter_expression.js from disk for each test
- Opening about:blank for each parameterized test iteration
- Awaiting TelemetryEnvironment and ExperimentAPI initialization for each evaluation
- Running targeting evaluations sequentially

## This commit

- **Eliminates experiment creation** - Tests targeting expressions directly without creating experiments via GraphQL
- **Exposes targeting expressions via GraphQL** - Added `targeting` field to the existing `nimbusConfig.targetingConfigs` query 
- **Parallelizes targeting evaluation** - Uses `Promise.all` to evaluate all targeting expressions concurrently in a single browser script execution
- **Caches filter_expression.js path** - Uses `@functools.cache` decorator to avoid repeated disk reads
- **Opens about:blank once per module** - Changed to `scope="module"` autouse fixture instead of per-test
- **Initializes modules once** - TelemetryEnvironment, ExperimentAPI, and other Firefox modules are imported and initialized only once at the script level
- **Consolidates test runs** - Changed from parameterized tests (running sequentially) to batch evaluation (running in parallel)

Fixes #14073

### DIFF
--- a/experimenter/tests/integration/nimbus/test_desktop_targeting.py
+++ b/experimenter/tests/integration/nimbus/test_desktop_targeting.py
@@ -6,7 +6,6 @@ from pathlib import Path
 import pytest
 
 from nimbus.models.base_dataclass import BaseExperimentApplications
-from nimbus.pages.browser import Browser
 from nimbus.utils import helpers
 
 
@@ -86,15 +85,22 @@ def test_check_advanced_targeting(
 
     # Evaluate all targeting expressions in parallel
     with filter_expression_path.open() as js:
-        results = Browser.execute_async_script(
-            selenium,
-            json.dumps(targeting_tests),
-            script=js.read(),
-            context="chrome",
-        )
+        with selenium.context(selenium.CONTEXT_CHROME):
+            results = selenium.execute_async_script(
+                js.read(),
+                json.dumps(targeting_tests),
+            )
 
     # Validate results
-    assert results is not None, "Failed to evaluate targeting expressions"
+    if results is None:
+        pytest.fail("Failed to evaluate targeting expressions - script returned None")
+
+    # Check if we got an error object instead of results array
+    if isinstance(results, dict) and "error" in results:
+        pytest.fail(
+            f"JavaScript error: {results.get('error')}\nStack: {results.get('stack')}"
+        )
+
     assert len(results) == len(targeting_configs), (
         f"Expected {len(targeting_configs)} results, got {len(results)}"
     )
@@ -181,15 +187,22 @@ def test_check_audience_targeting(
 
     # Evaluate all in parallel
     with filter_expression_path.open() as js:
-        results = Browser.execute_async_script(
-            selenium,
-            json.dumps(targeting_tests),
-            script=js.read(),
-            context="chrome",
-        )
+        with selenium.context(selenium.CONTEXT_CHROME):
+            results = selenium.execute_async_script(
+                js.read(),
+                json.dumps(targeting_tests),
+            )
 
     # Validate results
-    assert results is not None, "Failed to evaluate audience targeting"
+    if results is None:
+        pytest.fail("Failed to evaluate audience targeting - script returned None")
+
+    # Check if we got an error object instead of results array
+    if isinstance(results, dict) and "error" in results:
+        pytest.fail(
+            f"JavaScript error: {results.get('error')}\nStack: {results.get('stack')}"
+        )
+
     assert len(results) == len(audience_fields), (
         f"Expected {len(audience_fields)} results, got {len(results)}"
     )
--- a/experimenter/tests/integration/nimbus/utils/filter_expression.js
+++ b/experimenter/tests/integration/nimbus/utils/filter_expression.js
@@ -29,29 +29,31 @@ async function evaluateTargeting(targetingString, recipe) {
 }
 
 async function main() {
-    // Initialize once
-    await TelemetryEnvironment.onInitialized();
-    await ExperimentAPI.ready();
+    try {
+        // Initialize once
+        await TelemetryEnvironment.onInitialized();
+        await ExperimentAPI.ready();
 
-    const [testsJson, callback] = arguments;
-    const tests = JSON.parse(testsJson);
+        const [testsJson, callback] = arguments;
+        const tests = JSON.parse(testsJson);
 
-    // Evaluate all targeting expressions in parallel
-    const results = await Promise.all(
-        tests.map(async (test) => {
-            try {
-                const result = await evaluateTargeting(test.targeting, test.recipe);
-                return { slug: test.slug, result: result, error: null };
-            } catch (err) {
-                return { slug: test.slug, result: null, error: err.message };
-            }
-        })
-    );
+        // Evaluate all targeting expressions in parallel
+        const results = await Promise.all(
+            tests.map(async (test) => {
+                try {
+                    const result = await evaluateTargeting(test.targeting, test.recipe);
+                    return { slug: test.slug, result: result, error: null };
+                } catch (err) {
+                    return { slug: test.slug, result: null, error: err.message };
+                }
+            })
+        );
 
-    callback(results);
+        callback(results);
+    } catch (err) {
+        const [, callback] = arguments;
+        callback({ error: err.message, stack: err.stack });
+    }
 }
 
-main().catch(() => {
-    const [, callback] = arguments;
-    callback(null);
-});
+main();


experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -31,17 +38,28 @@ async function remoteSettings(targetingString, recipe) {
...
-  })
-  .catch(err => {
-    callback(null);
-  });
+
+if (arguments.length === 1) {
+    // Mode 1: Initialization - run once at test setup
+    const [callback] = arguments;
+    initialize().then(callback).catch(() => callback(false));
+} else {
```
freshstrangemusic: We should keep it simple and just have one mode of operation.

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -9,50 +11,101 @@
...
-
-
 @pytest.fixture
+@cache
 def filter_expression_path():
     path = Path(__file__).parent / "utils" / "filter_expression.js"
     return path.absolute()
 
 
+@pytest.fixture(scope="function", autouse=True)
```
freshstrangemusic, line 26: we want `scope="module"` to run this code once.

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -9,50 +11,101 @@
...
+        targeting = experiment_data["data"]["experimentBySlug"]["jexlTargetingExpression"]
+        recipe = experiment_data["data"]["experimentBySlug"]["recipeJson"]
+
+        targeting_tests.append(
+            {
+                "slug": slug,
+                "targeting": targeting,
+                "recipe": json.dumps({"experiment": recipe}),
+            }
+        )
```
freshstrangemusic: We should not create any experiments and instead use a mapping of targeting slug to targeting expression. We may be able to just import `nimbus.targeting.constants` directly here 

This is the real bottleneck of this test.

freshstrangemusic (changes requested):

experimenter/tests/integration/nimbus/test_desktop_targeting.py
yashikakhurana: actually just realised tha the selenium fixture is function-scoped, so we can't use module scope for setup_browser

experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -1,13 +1,11 @@
-async function remoteSettings(targetingString, recipe) {
-    const { TelemetryEnvironment } = ChromeUtils.importESModule("resource://gre/modules/TelemetryEnvironment.sys.mjs");
-    await TelemetryEnvironment.onInitialized();
-
-    const { ASRouterTargeting } = ChromeUtils.importESModule("resource:///modules/asrouter/ASRouterTargeting.sys.mjs");
-    const { ExperimentAPI } = ChromeUtils.importESModule("resource://nimbus/ExperimentAPI.sys.mjs");
-    const { TargetingContext } = ChromeUtils.importESModule("resource://messaging-system/targeting/Targeting.sys.mjs");
+// Initialize modules once
```
freshstrangemusic, line 1: Remove this comment. These modules are not being initialized, they are being imported. (a la regular `import` in js)

experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -26,22 +24,34 @@ async function remoteSettings(targetingString, recipe) {
...
-arguments[1] - the experiment recipe
-arguments[3] - the callback from selenium
-*/
-const [targetingString, recipe, callback] = arguments;
+async function main() {
+    // Initialize once
+    await TelemetryEnvironment.onInitialized();
+    await ExperimentAPI.ready();
+
+    const [testsJson, callback] = arguments;
```
freshstrangemusic: You have to set this outside the `main` function because `arguments` is a special global that refers to the arguments passed to the current function. ie so in this function it will be `[]`

experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -26,22 +24,34 @@ async function remoteSettings(targetingString, recipe) {
...
+
+    callback(results);
+}
 
-remoteSettings(targetingString, recipe)
-  .then(result => {
-    callback(result);
-  })
-  .catch(err => {
+main().catch(() => {
```
freshstrangemusic:
```suggestion
const [targeting, callback] = arguments;
main(targeting, callback).catch(() => callback(null));
```

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -9,103 +10,205 @@
...
 def filter_expression_path():
     path = Path(__file__).parent / "utils" / "filter_expression.js"
     return path.absolute()
 
 
+@pytest.fixture(scope="function", autouse=True)
+def setup_browser(selenium):
+    """Open about:blank once per test function."""
+    selenium.get("about:blank")
+    yield
```
freshstrangemusic, line 30: I'm actually curious if this test works without this function.

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -9,103 +10,205 @@
...
+
+@pytest.mark.run_targeting
+def test_check_audience_targeting(
+    selenium,
+    filter_expression_path,
+):
+    """Test that audience fields generate valid targeting expressions."""
+    # These tests verify the targeting expression generation logic
+    # without needing to create actual experiments
+    recipe = {
```
freshstrangemusic, line 131: This test actually doesn't work like this. The old test was testing that Experimenter was generating valid targeting expressions based on recipe configurations. This updated test does not do that since it does not actually pass any targeting.

We should revert most of this test to the old method so we're actually testing what comes out of experimenter. `O(1)` experiments should be fine -- the real bottleneck was the `O(n)` experiments made because of `n` advanced targetings.

freshstrangemusic (changes requested):

yashikakhurana:
There is issue that one needs the actual creation of experiments, and for the other we can skip it, going to test it more and will open up again @freshstrangemusic I will address the feedback you shared

PR #14084, opened 2025-11-28
feat(nimbus): Improve desktop targeting tests

### LINKED ISSUE #14083: Improve filter_expression.js for desktop targeting tests
There are several things that could be improved with the desktop targeting tests:

* The filter_expression.js file is opened and read from disk in each test (targeting slug) instead of being cached ([https://github.com/mozilla/experimenter/blob/1e74d9ecf77a83b7f173423bff94111a3eb905c9/experimenter/tests/integration/nimbus/test_desktop_targeting.py#L17-L20](https://github.com/mozilla/experimenter/blob/1e74d9ecf77a83b7f173423bff94111a3eb905c9/experimenter/tests/integration/nimbus/test_desktop_targeting.py#L17-L20|smart-link) ) This can be solved by using a `@functools.cache` decorator on the pixture after the `@pytest.fixture` decorator (e.g., ​ )
* `test_check_audience_targeting` and `test_check_advanced_targeting` may be opening an `about:blank` tab for each test (`selenium.get("about:blank")`). We can open a single page at the start of the test by using an autouse fixture, e.g.,

```
@pytest.fixture(autouse=True)

def open_about_blank():

 selenium.get("about:blank")

 # other set up code

 yield

 # tear down code
```

​

* `filter_expression.js` is awaiting `TelemetryEnvironment.onInitialized()` and `ExperimentAPI.ready()` when evaluating each slug. We can do this in a separate script executed once at the beginning of the test, e.g., in the `open_about_blank()` proposed above.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6237)

yashikakhurana:
Because
- Desktop targeting tests were performing redundant file I/O operations by reading filter_expression.js from disk on every test iteration
- Each test was opening about:blank individually, leading to duplicate browser navigation
- TelemetryEnvironment.onInitialized() and ExperimentAPI.ready() were being called repeatedly inside the evaluation function, causing unnecessary initialization overhead

This commit
- Adds cached helper function _get_filter_expression_path() with @cache decorator to compute and cache the file path once across all test invocations
- Moves Firefox module imports (TelemetryEnvironment, ASRouterTargeting, ExperimentAPI, TargetingContext) to top-level in filter_expression.js
- Refactors script execution into async main() function that initializes TelemetryEnvironment and ExperimentAPI once per script execution before evaluating targeting expressions
- Properly awaits initialization and evaluation to ensure callback is invoked only after completion

fixes #14083

### DIFF
--- a/experimenter/tests/integration/nimbus/test_desktop_targeting.py
+++ b/experimenter/tests/integration/nimbus/test_desktop_targeting.py
@@ -17,23 +17,10 @@ def targeting_config_slug(request):
 
 @pytest.fixture
 @cache
-def filter_expression_path():
+def targeting_script():
     path = Path(__file__).parent / "utils" / "filter_expression.js"
     with path.open() as f:
         return f.read()
-
-
-@pytest.fixture(scope="module")
-def browser_initialized():
-    return {"initialized": False}
-
-
-@pytest.fixture(autouse=True, scope="function")
-def setup_browser(selenium, browser_initialized):
-    if not browser_initialized["initialized"]:
-        selenium.get("about:blank")
-        browser_initialized["initialized"] = True
-    yield
 
 
 @pytest.mark.run_targeting
@@ -42,7 +29,7 @@ def test_check_advanced_targeting(
     targeting_config_slug,
     experiment_slug,
     default_data_api,
-    filter_expression_path,
+    targeting_script,
 ):
     default_data_api["targetingConfigSlug"] = targeting_config_slug
     experiment = helpers.create_experiment(
@@ -62,7 +49,7 @@ def test_check_advanced_targeting(
         selenium,
         targeting,
         json.dumps({"experiment": recipe}),
-        script=filter_expression_path,
+        script=targeting_script,
         context="chrome",
     )
     assert result is not None, "Invalid Targeting, or bad recipe"
@@ -95,7 +82,7 @@ def test_check_audience_targeting(
     audience_field,
     experiment_slug,
     default_data_api,
-    filter_expression_path,
+    targeting_script,
 ):
     default_data_api.update(audience_field)
     experiment = helpers.create_experiment(
@@ -115,7 +102,7 @@ def test_check_audience_targeting(
         selenium,
         targeting,
         json.dumps({"experiment": recipe}),
-        script=filter_expression_path,
+        script=targeting_script,
         context="chrome",
     )
     assert result is not None, "Invalid Targeting, or bad recipe"


experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -1,13 +1,10 @@
...
-
-    const { ASRouterTargeting } = ChromeUtils.importESModule("resource:///modules/asrouter/ASRouterTargeting.sys.mjs");
-    const { ExperimentAPI } = ChromeUtils.importESModule("resource://nimbus/ExperimentAPI.sys.mjs");
-    const { TargetingContext } = ChromeUtils.importESModule("resource://messaging-system/targeting/Targeting.sys.mjs");
+const { TelemetryEnvironment } = ChromeUtils.importESModule("resource://gre/modules/TelemetryEnvironment.sys.mjs");
+const { ASRouterTargeting } = ChromeUtils.importESModule("resource:///modules/asrouter/ASRouterTargeting.sys.mjs");
+const { ExperimentAPI } = ChromeUtils.importESModule("resource://nimbus/ExperimentAPI.sys.mjs");
+const { TargetingContext } = ChromeUtils.importESModule("resource://messaging-system/targeting/Targeting.sys.mjs");
 
+async function remoteSettings(targetingString, recipe) {
```
freshstrangemusic: Can we please rename this function to something else? It has nothing to do with remote settings

experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -34,14 +31,20 @@ async function remoteSettings(targetingString, recipe) {
...
 arguments[1] - the experiment recipe
-arguments[3] - the callback from selenium
+arguments[2] - the callback from selenium
 */
 const [targetingString, recipe, callback] = arguments;
 
-remoteSettings(targetingString, recipe)
-  .then(result => {
+async function main() {
+    // Initialize once
```
freshstrangemusic: Unnecessary comment.

experimenter/tests/integration/nimbus/utils/filter_expression.js
```diff
@@ -34,14 +31,20 @@ async function remoteSettings(targetingString, recipe) {
...
+    await TelemetryEnvironment.onInitialized();
+    await ExperimentAPI.ready();
+
+    // Evaluate targeting
```
freshstrangemusic: Unnecessary comment

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -14,12 +15,25 @@ def targeting_config_slug(request):
...
+@cache
+def _get_filter_expression_path():
+    """Cached helper to get the filter expression path."""
     path = Path(__file__).parent / "utils" / "filter_expression.js"
     return path.absolute()
 
 
+@pytest.fixture
+def filter_expression_path():
+    return _get_filter_expression_path()
```
freshstrangemusic: Caching the path is unnecessary -- this is just string concatenation.
Additionally, we do not need a separate function call for the cache decorator.

What we want to cache is the result of the I/O, e.g.

```suggestion
@pytest.fixture
@cache
def test_js():
    path = Path(__file__).parent / "utils" / "filter_expression.js"
    with path.open() as f:
        return f.read()
```

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -14,12 +15,25 @@ def targeting_config_slug(request):
...
+
+
+@pytest.fixture(autouse=True, scope="function")
```
freshstrangemusic: I mentioned last time that this should be `scope="module"` so that we only do it once.

freshstrangemusic (dismissed):

experimenter/tests/integration/nimbus/utils/filter_expression.js
yashikakhurana: I agree, I just made the function async `remoteSettings(targetingString, recipe) `and moved up with the same name, I can change the name

experimenter/tests/integration/nimbus/test_desktop_targeting.py
yashikakhurana: since `selenium` is function-scoped, so to work around with this I did this 
- Created browser_initialized module-scoped fixture to track state
- Modified setup_browser to only call selenium.get("about:blank") once per module
- Uses lazy initialization pattern to work around selenium's function scope

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -15,9 +16,24 @@ def targeting_config_slug(request):
...
+def browser_initialized():
+    return {"initialized": False}
+
+
+@pytest.fixture(autouse=True, scope="function")
+def setup_browser(selenium, browser_initialized):
+    if not browser_initialized["initialized"]:
+        selenium.get("about:blank")
+        browser_initialized["initialized"] = True
+    yield
```
freshstrangemusic: Why are we doing this extra layer of indirection?

```suggestion
@pytest.fixture(autouse=True, scope="module")
def setup_browser(selenium):
    selenium.get("about:blank")
    yield
```

yashikakhurana: so changing it to module CI fails https://app.circleci.com/pipelines/github/mozilla/experimenter/65876/workflows/6adb259c-6aec-4d7c-8f2e-d68915ff4edd/jobs/404467 
https://github.com/mozilla/experimenter/pull/14084#discussion_r2572301731

yashikakhurana: I have reverted to use the workaround

yashikakhurana: since selenium is function-scoped, so to work around with this I did this

    Created browser_initialized module-scoped fixture to track state
    Modified setup_browser to only call selenium.get("about:blank") once per module
    Uses lazy initialization pattern to work around selenium's function scope

freshstrangemusic: Oh this selenium helper is also doing whack stuff and making e.g., API calls and kinto calls, which it really doesn't need to be doing in most cases.

freshstrangemusic: Lets undo these changes since they're not having any effect whatsoever.

experimenter/tests/integration/nimbus/test_desktop_targeting.py
```diff
@@ -15,9 +16,24 @@ def targeting_config_slug(request):
 
 
 @pytest.fixture
+@cache
 def filter_expression_path():
```
freshstrangemusic: This is not the `filter_expression_path` any more.

PR #14121, opened 2025-12-05
feat(nimbus): Set QA date

### LINKED ISSUE #14023: Use date of qa field changelog for feature test date
On the feature page under test runs, let's use teh most recent changelog where the qa status last changed as the test date

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6209)

yashikakhurana:
Because

- We want to show users last QA date on the new feature page, which is empty right now

This commit

- Updates all the experiment to use the last QA change date as the QA run date using the migration
- Update QA form to update the qa run date when we update the QA field

Fixes #14023

### DIFF
--- a/experimenter/experimenter/experiments/migrations/0303_backfill_qa_run_date_from_changelog.py
+++ b/experimenter/experimenter/experiments/migrations/0303_backfill_qa_run_date_from_changelog.py
@@ -5,7 +5,6 @@ from django.db import migrations
 
 def backfill_qa_run_date_from_changelog(apps, schema_editor):
     NimbusExperiment = apps.get_model("experiments", "NimbusExperiment")
-    NimbusChangeLog = apps.get_model("experiments", "NimbusChangeLog")
 
     experiments_to_update = NimbusExperiment.objects.filter(
         qa_run_date__isnull=True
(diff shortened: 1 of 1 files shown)


yashikakhurana:
Screenshot 2025-12-05 at 9 04 01 AM

experimenter/experimenter/experiments/migrations/0303_backfill_qa_run_date_from_changelog.py
```diff
@@ -0,0 +1,55 @@
...
+        qa_status="NOT_SET"
+    )
+
+    for experiment in experiments_to_update:
+        # Get changelogs for this experiment ordered by most recent first
+        changelogs = list(
+            NimbusChangeLog.objects.filter(experiment=experiment)
+            .order_by("-changed_on")
+            .values("experiment_data", "changed_on")
+        )
```
jaredlockhart: Why don't we do something like

```py
experiment.changes.all().exclude(experiment_data__qa_status='NOT SET').order_by('changed_on').first().changed_on
```

Just get the first one where it goes from NOT SET to something?  Faster to do in the db than in memory.

jaredlockhart (changes requested):
Ya generally looks good, but we have to watch out for migrations that take too long, we had one recently and it blew argo up pretty good.  See if doing the query like this helps.

experimenter/experimenter/experiments/migrations/0303_backfill_qa_run_date_from_changelog.py
yashikakhurana: good point, thanks @jaredlockhart

experimenter/experimenter/experiments/migrations/0303_backfill_qa_run_date_from_changelog.py
```diff
@@ -0,0 +1,40 @@
...
+        latest_qa_changelog = (
+            experiment.changes.all()
+            .filter(experiment_data__qa_status=experiment.qa_status)
+            .order_by("-changed_on")
+            .first()
+        )
+
+        if latest_qa_changelog:
+            experiment.qa_run_date = latest_qa_changelog.changed_on.date()
+            experiment.save(update_fields=["qa_run_date"])
```
jaredlockhart, line 27: 🎉 🎉 🎉

jaredlockhart (approved):
Yes very nice, great test coverage, nice little form mixin, great job @yashikakhurana thank you 🙏

PR #14147, opened 2025-12-08
feat(nimbus): Split tables on feature page

### LINKED ISSUE #14025: Split feature page into separate cards
Rather than a single card with separators let's just make each table its own card to be consistent with the home page design.

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6211)

yashikakhurana:
Because

- Feature page has three tables in a single card which is not consistent with the existing design on the home page we have

This commit

- Splits the feature page tables into separate cards

Fixes #14025


yashikakhurana:
Before
Image

After
Screenshot 2025-12-08 at 3 41 54 PM

PR #14149, opened 2025-12-09
feat(nimbus): add targeting for accepting tou before and after partic…

### LINKED ISSUE #14148: [Targeting] add targeting for accepting tou before and after particular dates
### What is the criteria you need to target based on?

- Users who accepted TOU before Dec 17, 2025
- Users who accepted TOU after Dec 15, 2025

### Should the targeting include or exclude users?

Targeting should include these users.

### Additional Notes (optional)

_No response_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6254)

jwayn:
Because

- we want to target users who have accepted TOU before and after particular dates

This commit

- adds targeting for users who have accepted TOU before and after particular dates

Fixes #14148

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -53,6 +53,9 @@ ACCEPTED_TOU_V4_OR_HIGHER = "'termsofuse.acceptedVersion'|preferenceValue >= 4"
 # 23:59 UTC on Dec 9, 2025 when an updated version of the privacy
 # notice was published.
 DEC_9_2025 = 1765324740000
+# Privacy Notification Published date of 12:00 PM UTC on Dec 15, 2025
+DEC_15_2025 = 1765800000000
+DEC_17_2025 = 1765972800000
 
 TOU_NOTIFICATION_BYPASS_ENABLED = "'termsofuse.bypassNotification'|preferenceValue"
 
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/targeting/constants.py
```diff
@@ -53,6 +53,10 @@ def __post_init__(self):
 # 23:59 UTC on Dec 9, 2025 when an updated version of the privacy
 # notice was published.
 DEC_9_2025 = 1765324740000
```
jwayn, line 55: This constant is no longer being used, but I was unsure if it'd be ok to delete it. Happy to do so if that wouldn't cause any issues.

experimenter/experimenter/targeting/constants.py
```diff
@@ -3666,6 +3670,42 @@ def __post_init__(self):
     application_choice_names=(Application.DESKTOP.name,),
 )
```
jwayn, line 3670: Likewise, these two `ACCEPTED_TOU_BEFORE_DEC_9_2025` and `ACCEPTED_TOU_AFTER_DEC_9_2025` are no longer used. I left them in because I'm unsure if there would be any ill effects of removing them. I can remove these as well if it won't cause issues!

cpeterso: The ToU team will clean out unused TOU constants and targets after our rollouts in 2026.

PR #14171, opened 2025-12-10
feat(nimbus): only log jetstream fetches on new data

### LINKED ISSUE #14163: Too many jetstream changelogs
Recently we added changelogs for jetstream fetches.  However we are creating one on every ATTEMPTED fetch, regardless of whether there were any changes, and the large number of changelogs can then prevent the admin or summary pages from loading at all.

Let's delete all these spurious changelogs, and only create one if the data *changed*.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6265)

jaredlockhart:
Becuase

* We recently added changelogs on jetstream fetches to improve visibility on successful or failed jetstream fetches
* We logged on every fetch attempt, so many unnecessary changelogs were created
* Large numbers of changelogs can cause pages to load slowly or not at all
* We only need to log if the data has changed

This commit

* Deletes all existing jetstream fetch changelogs
* Updates the logic to only create a changelog if the results data changed

fixes #14163


experimenter/experimenter/experiments/migrations/0305_delete_jetstream_changelogs.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 5.2.9 on 2025-12-09 21:24
+
+from django.db import migrations
+
+
+def delete_results_fetched_changelogs(apps, _schema_editor):
+    NimbusChangeLog = apps.get_model("experiments", "NimbusChangeLog")
+    deleted_count, _ = NimbusChangeLog.objects.filter(
+        message="Experiment results fetched"
```
mikewilli, line 9: Should this be "updated" instead of "fetched" now?

experimenter/experimenter/experiments/migrations/0305_delete_jetstream_changelogs.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 5.2.9 on 2025-12-09 21:24
+
+from django.db import migrations
+
+
+def delete_results_fetched_changelogs(apps, _schema_editor):
```
mikewilli, line 6: This isn't going to clear out all the changelogs whenever the migration gets applied, right?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -23,13 +23,18 @@ def fetch_experiment_data(experiment_id):
     experiment = None
     try:
         experiment = NimbusExperiment.objects.get(id=experiment_id)
-        experiment.results_data = get_experiment_data(experiment)
+        old_results_data = experiment.results_data
+        new_results_data = get_experiment_data(experiment)
+        experiment.results_data = new_results_data
         experiment.save()
```
mikewilli: This could probably go inside the `if` block, though I imagine the performance difference is negligible..

experimenter/experimenter/experiments/migrations/0305_delete_jetstream_changelogs.py
jaredlockhart: Migrations run only once, so it'll happen once, we'll lose all the jetstream changelogs from the last few weeks, and then start getting filled in from land time forward.

mikewilli (approved):

PR #14179, opened 2025-12-11
feat(nimbus): Send channel notifications in slack DMs

### LINKED ISSUE #14173: Send slack notifications in users DM
We noticed that if the users are not in the slack channel they are not getting notified until they join the channel, we can also send the DM of the message to them, lets add to send the DM of the notification or the slack channel message reference link

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6271)

yashikakhurana:
Because

- We noticed that if the user is not in the slack channel, tagging user won't notify them

This commit

- Sends DMs to the users as well referencing the slack channel message
- In case it's not able to reference the channel message, it will still send the notification

Fixes #14173


experimenter/experimenter/slack/notification.py
```diff
@@ -15,18 +15,39 @@ def _get_slack_client():
...
             response = client.users_lookupByEmail(email=email)
-            mentions.append(f"<@{response['user']['id']}>")
+            user_id = response["user"]["id"]
+            mentions.append(f"<@{user_id}>")
+            user_ids.append(user_id)
         except SlackApiError as e:
             logger.warning(f"Could not find Slack user for {email}: {e}")
             continue
-    return " ".join(mentions)
+    return " ".join(mentions), user_ids
```
jaredlockhart: I mean I'd probably keep this function just to lookup and return the user ids and move constructing the mentions elsewhere rather than fold them both into one func

yashikakhurana: sure i can do that change

yashikakhurana: okay done!! 🎉

yashikakhurana:
I am blocking this one as pending on security review (to use updated permissions), once we get the approval will merge this

yashikakhurana:
okay another RRA done, got approval from both security and slack app manager, lets merge this now

PR #14189, opened 2025-12-12
chore(nimbus): always use transaction.atomic for form saves

### LINKED ISSUE #14188: Use transaction.atomic for all form saves
Let's make sure every form save is wrapped in transaction.atomic so we don't get races between the celery tasks and the UI changes.  Let's also make sure every save generates a changelog so we never wind up with changed data in the db without a record of when it was changed and by whom.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6277)

jaredlockhart:
Becuase

* We recently saw races and collisions between forms and celery tasks saving the same instances
* We can wrap both in transaction.atomic to ensure they occur separately

This commit

* Wraps all form saves in transaction.atomic

fixes #14188

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -5,6 +5,7 @@ from datetime import UTC, datetime
 import markus
 from django import forms
 from django.contrib.auth.models import User
+from django.db import transaction
 from django.db.models import Case, When
 from django.forms import BaseInlineFormSet, BaseModelFormSet, inlineformset_factory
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
Totally, I am just wondering if there might be some easy way instead of adding it on all save 🤔, but totally works. Thank you @jaredlockhart

jaredlockhart:
Yeah @yashikakhurana afaik this is the only way.  I mean without doing some kind of metaclass hackery, but I generally try to avoid that.

PR #14197, opened 2025-12-15
chore(nimbus): readonly changelogs in admin

### LINKED ISSUE #14164: NimbusExperiment admin page too slow
The form detail admin page for NimbusExperiment is loading too slowly and will just 502 in a lot of cases.  Biggest problem is serializing all the changelogs into readonly inlines.  Let's see if we can remove those and just dump them out as strings or something simpler and see if that helps load times.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6266)

jaredlockhart:
Becuase

* We had inline formsets for changelogs in the admin
* These are very slow to render especially for experiments with lots of changes
* We don't actually need to be able to edit changelogs in the admin, they really should be read only
* We can just display their values for reference purpose in the admin, this should be much faster to render

This commit

* Replaces NimbusExperimentAdmin changelog inline formsets with a read only table

fixes #14164

### DIFF
--- a/experimenter/experimenter/experiments/admin.py
+++ b/experimenter/experimenter/experiments/admin.py
@@ -1,12 +1,13 @@
 from typing import Any
-from weakref import WeakKeyDictionary
 
 from django import forms
 from django.contrib import admin
 from django.contrib.auth.models import User
 from django.contrib.postgres import forms as pgforms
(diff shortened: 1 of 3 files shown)


jaredlockhart:
image

PR #14199, opened 2025-12-15
feat(nimbus): wrap long segment names

### LINKED ISSUE #14170: Segment names overflow
Looks like the segment names are overflowing past the width of the sidebar because they're snake cased so they won't line break.

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6270)

jaredlockhart:
Becuase

* Segments can have long names
* They use snake case so they will not line break
* This can cause long segment names to overflow the sidebar

This commit

* Replaces underscores with spaces so long segment names can wrap

fixes #14170


jaredlockhart:
image

PR #14213, opened 2025-12-16
chore(nimbus): Remove Normandy Edit/Launch

### LINKED ISSUE #14200: Delete legacy forms
Now that Normandy is fully deprecated we can remove the forms in the legacy UI.  We should keep the list and detail pages so we can go back and reference old experiments if necessary, but we don't need to keep the forms around, which will let us delete all their tests which should speed up `make check`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6280)

jaredlockhart:
Becuase

* Normandy is now deprecated and the server is decommissed
* We no longer need the ability to launch or edit Normandy experiments
* We can safely disable all create/edit capabilities in the legacy UI

This commit

* Removes all Legacy forms/UI
* Removes bugzilla/email integrations for Legacy experiments
* Removes V1/V2 APIs which are no longer called
* Preserves a readonly list and detail page for all past Normandy experiments

fixes #14200


yashikakhurana (approved):
-29054 🎉 This is so great!! Thank you @jaredlockhart 🎉 🎉

PR #14218, opened 2025-12-17
feat(nimbus): Add prefix channel name in the slack DM message

### LINKED ISSUE #14215: Add prefix as the channel name in the DM message
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6284)

yashikakhurana:
Because

- We are sending DMs to the user if they are not in the channel about the slack message, but we are not including the channel name to promote them to join the channel

This commit

- Adds the channel name in the message

Fixes #14215


jaredlockhart (dismissed):
Aren't there tests?

yashikakhurana:
> Aren't there tests?
yup there are, my bad, skipped to commit that file

PR #14225, opened 2025-12-18
feat(nimbus): sync published_dto when null

### LINKED ISSUE #14223: Periodically sync published_dto
We have recently seen cases where an experiment can launch with `published_dto=null`.  This breaks a bunch of the workflow.  We've made changes to prevent this from occurring but it's not foolproof, so let's just add a period task that looks for any live experiment without a published dto and just syncs it with what's in remote settings.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6286)

jaredlockhart:
Because

* We recently saw cases where published_dto can be null for live experiments
* We added fixes to prevent this from happening but there may still be some live experiments missing published_dto
* We can add a task to fill those in if it happens again

This commit

* Adds a periodic task that looks for live experiments with published_dto=null and fetches them from kinto

fixes #14223

### DIFF
--- a/experimenter/experimenter/kinto/tasks.py
+++ b/experimenter/experimenter/kinto/tasks.py
@@ -535,3 +535,52 @@ def nimbus_send_emails():
             logger.info(f"{experiment} end email sent")
 
     metrics.incr("nimbus_send_emails.completed")
+
+
+@app.task
+@metrics.timer_decorator("nimbus_sync_published_dto")
+def nimbus_sync_published_dto():
+    """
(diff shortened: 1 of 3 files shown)


jaredlockhart:
@yashikakhurana the fix helps it happen LESS but I don't think there's any way to prevent it from happening EVER, so I think having this to catch and clean up cases is still a good idea.  Doesn't hurt anything to have this task.

yashikakhurana (approved):
Thank you for explaining @jaredlockhart :shipit:

PR #14253, opened 2025-12-29
feat(nimbus): update metrics to be grouped by new metric area field

### LINKED ISSUE #14239: Update metric groups to be by metric area
Metrics are currently grouped by outcomes. The final design requires metrics to be together based on their “metric area” which is a new field that must be added to metric hub

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6285)

moibra05:
Because

- The final design requires metrics to be group by area where possible
- Area is a new field that must be appended to the existing metric definitions with a fallback value when not provided

This commit

- Adds "metric_area" field to specific metric definitions according to [this](https://docs.google.com/spreadsheets/d/1CBntDFdz4n0oYVbvlzrA8SXWZ9O09WOegSY6-QNeXs8/edit?gid=0#gid=0) spreadsheet
- Fixes issue where `abs` function was being called on None when initializing relative points data

Fixes #14239

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -32,6 +32,7 @@ from experimenter.experiments.constants import (
     NimbusConstants,
     TargetingMultipleKintoCollectionsError,
 )
+from experimenter.metrics import MetricAreas
 from experimenter.nimbus_ui.constants import NimbusUIConstants
 from experimenter.outcomes import Outcomes
 from experimenter.projects.models import Project
@@ -1269,8 +1270,6 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
         ]
 
     def results_sidebar_sections(self):
-        # TODO: show metrics by grouped categories based on metric area
-
         return [
             {
                 "title": "Overview",
@@ -1282,13 +1281,18 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                     {"title": "Project Impact"},
                 ],
             },
-            {
-                "title": "All metrics",
-                "subitems": [
-                    {"title": metric["friendly_name"]}
-                    for metric in self.get_remaining_metrics_metadata()
-                ],
-            },
+            *[
+                {
+                    "title": area,
+                    "subitems": [
+                        {"title": metric["friendly_name"], "slug": metric["slug"]}
+                        for metric in metrics
+                    ],
+                }
+                for area, metrics in self.get_metric_areas(
+                    "enrollments", "all", self.reference_branch
+                ).items()
+            ],
         ]
 
     def timeline(self):
@@ -1365,12 +1369,11 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
     def get_metric_areas(
         self, analysis_basis, segment, reference_branch, window="overall"
     ):
-        metric_areas = {
-            NimbusUIConstants.NOTABLE_METRIC_AREA: [],
-            NimbusUIConstants.KPI_AREA: self.get_kpi_metrics(
-                analysis_basis, segment, reference_branch, window
-            ),
-        }
+        metric_areas = defaultdict(list)
+        metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA] = []
+        metric_areas[NimbusUIConstants.KPI_AREA] = self.get_kpi_metrics(
+            analysis_basis, segment, reference_branch, window
+        )
 
         metrics_metadata = {}
         if self.results_data:
@@ -1411,13 +1414,19 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             outcome_metrics.sort(key=lambda m: m["friendly_name"])
             metric_areas[outcome.friendly_name if outcome else slug] = outcome_metrics
 
-        metric_areas[NimbusUIConstants.OTHER_METRICS_AREA] = (
-            self.get_remaining_metrics_metadata(
-                exclude_slugs=all_outcome_metric_slugs,
-                analysis_basis=analysis_basis,
-                segment=segment,
-            )
+        remaining_metrics = self.get_remaining_metrics_metadata(
+            exclude_slugs=all_outcome_metric_slugs
         )
+        grouped_metrics = []
+        for metric in remaining_metrics:
+            area = MetricAreas.get(self.application, metric["slug"])
+
+            metric_areas[area].append(metric)
+            grouped_metrics.append(metric)
+
+        metric_areas[NimbusUIConstants.OTHER_METRICS_AREA] = [
+            m for m in remaining_metrics if m not in grouped_metrics
+        ]
 
         window_results = self.get_window_results(analysis_basis, segment, window)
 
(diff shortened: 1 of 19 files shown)


moibra05:
I should also mention that metrics are still grouped by outcomes as well where applicable. The PRD I received says that all other metrics without the area field should be thrown in the "Other Metrics" section but through testing I've found that this makes that section disproportionately larger than the others and kind of defeats the purpose of having this grouping functionality at all. Curious to know what others think and whether it'd be best to just stick to whats been outlined?

experimenter/experimenter/settings.py
```diff
@@ -550,7 +550,7 @@
 )
 
 METRIC_HUB_SEGMENTS_PATH_DEFAULT = BASE_DIR / "segments" / "metric-hub-main"
-
+METRIC_HUB_METRICS_PATH_DEFAULT = BASE_DIR / "metrics" / "metric-hub-main" / "definitions"
```
mikewilli: Unless I'm missing something, these are pulling in the same files -- the segments are defined within the definitions files. I haven't looked at the whole PR yet so I don't know what we need from the metric definitions, but we should be able to get the info from what we're already pulling in for segments.

mikewilli:
I know a big reason why this is 6000+ lines is because you're adding all the metric definitions, but consider splitting this up into 2-3 PRs. Maybe something like:
- PR 1: get metric area info from metric-hub (see my comment above about consolidating what you pull in)
- PR 2: add model info (maybe this makes sense in PR 1, maybe not)
- PR 3: update UI

experimenter/experimenter/segments/metric-hub-main/definitions/firefox_desktop.toml
```diff
@@ -1576,71 +1589,6 @@ type = "scalar"
...
-    COALESCE(SUM(mozfun.map.get_key(metrics.labeled_counter.pdfjs_editing_highlight_kind, "highlight")) > 0, FALSE)
-)"""
-data_source = "metrics"
-
-[metrics.pdf_free_highlight]
-select_expression = """(
-    COALESCE(SUM(mozfun.map.get_key(metrics.labeled_counter.pdfjs_editing_highlight_kind, "free_highlight")) > 0, FALSE)
-)"""
-data_source = "metrics"
-
```
moibra05, line 1643: Im not sure if metrics require any fields at all but I found these to be the only few that were missing the `friendly_name` field. I removed them b/c I need there to be at least 1 required field so that I am able to write tests that can catch metrics that are missing fields and throw the appropriate error. I made this decision b/c I don't know of any other way to trigger an error especially if every metric field is optional

experimenter/experimenter/segments/metric-hub-main/definitions/fenix.toml
```diff
@@ -5,19 +5,21 @@ data_source = "baseline"
 select_expression = '{{agg_sum("metrics.counter.events_normal_and_private_uri_count")}}'
 friendly_name = "URIs visited"
 description = "Counts the number of URIs each client visited"
+metric_area = "Engagement"
```
mikewilli: I don't think you want to be editing these files. They get pulled in to the Experimenter repo from metric-hub, so if you want to make changes to the metric definitions you'd want them to be in metric-hub. See my overall comment though for thoughts on this approach generally.

experimenter/experimenter/segments/metric-hub-main/definitions/firefox_desktop.toml
mikewilli: The answer to your question is that a metric should require either a) a `select_expression` and a `data_source`, or b) a `depends_on` parameter. So no single field is required for all metrics.

However, it seems safe to assume that metrics in metric-hub are valid. If there are any cases where metric-hub definitions are lacking in terms of what Experimenter needs, then we probably want to define a contract and/or tests for metric-hub to follow. Open to discussion though if you think otherwise.

experimenter/experimenter/experiments/models.py
```diff
@@ -1358,9 +1362,25 @@ def get_metric_areas(
...
-        metric_areas[NimbusUIConstants.OTHER_METRICS_AREA] = (
-            self.get_remaining_metrics_metadata(exclude_slugs=all_outcome_metric_slugs)
+        remaining_metrics = self.get_remaining_metrics_metadata(
+            exclude_slugs=all_outcome_metric_slugs
         )
+        grouped_metrics = []
+        for metric in remaining_metrics:
+            metric_details = Metrics.get_by_slug_and_application(
+                metric["slug"], self.application.replace("-", "_")
+            )
```
freshstrangemusic: This function should maybe just accept the Application instance and pull out the slug itself so that callers dont have to do the replace() call.

experimenter/experimenter/experiments/models.py
```diff
@@ -1358,9 +1362,25 @@ def get_metric_areas(
...
+            metric_area = metric_details.metric_area if metric_details else None
+
+            if metric_area:
+                if metric_area not in metric_areas:
+                    metric_areas[metric_area] = []
+                metric_areas[metric_area].append(metric)
```
freshstrangemusic: you can use a `defaultdict(list)`:

```py
from collections import defaultdict

metric_areas = defaultdict(list)
metric_areas[metric_area].append(metric)
```

[docs](https://docs.python.org/3/library/collections.html#collections.defaultdict)

experimenter/experimenter/experiments/models.py
```diff
@@ -1391,9 +1362,24 @@ def get_metric_areas(
...
         )
+        grouped_metrics = []
+        for metric in remaining_metrics:
+            area = MetricAreas.get_by_slug_and_application(
+                metric["slug"], self.application.replace("-", "_")
+            )
+
+            if area:
+                if area not in metric_areas:
+                    metric_areas[area] = []
```
freshstrangemusic: This seems like a good place for defaultdict(list)

experimenter/experimenter/metrics/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        metrics_path = settings.METRIC_HUB_METRICS_PATH_DEFAULT
+        metrics_dir = Path(metrics_path)
+
+        for app_name in metrics_dir.iterdir():
+            if app_name.suffix != ".toml":
+                continue
+
+            with app_name.open() as metric_file:
+                metric_toml = metric_file.read()
+                metric_data = toml.loads(metric_toml).get("metrics", {}).get("areas")
```
freshstrangemusic:
```suggestion
                metric_data = toml.load(metric_file).get("metrics", {}).get("areas")
```

experimenter/experimenter/metrics/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    @classmethod
+    def all(cls):
+        if cls._metric_areas is None:
+            cls._metric_areas = cls._load_metrics_areas()
+
+        return cls._metric_areas
+
+    @classmethod
+    def get_by_slug_and_application(cls, slug, application):
+        all_metrics = cls.all().get(application)
```
freshstrangemusic: IMO this should be `get_by_application_and_slug(cls, application, slug)` because the application narrows the metrics before the metric slug 

Alternatively it could just be `get(cls, application, slug)`

experimenter/experimenter/experiments/models.py
freshstrangemusic: Whoops github wasnt showing me pr comments i wrote the other day while i was reviewing today.

PR #14259, opened 2025-12-31
fix(nimbus): fix results page throwing error for delivery w/o results

### LINKED ISSUE #14111: Visiting results-new for a delivery without results causes an exception
We need to guard against `experiment.results_data` being `None` here:

https://github.com/mozilla/experimenter/blob/d18e13463287dfa0d1c5d8ad8a25970d9715bcfe/experimenter/experimenter/nimbus_ui/views.py#L686

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6245)

moibra05:
Because

- Accessing the new results page for an experiment w/o results_data throws an exception 

This commit

- Adds a safe fallback value for when results_data is missing

Fixes #14111

### DIFF
--- a/experimenter/experimenter/nimbus_ui/views.py
+++ b/experimenter/experimenter/nimbus_ui/views.py
@@ -728,7 +728,7 @@ class NewResultsView(NimbusExperimentViewMixin, DetailView):
         context = super().get_context_data(**kwargs)
         experiment = self.get_object()
 
-        analysis_data = experiment.results_data.get("v3", {})
+        analysis_data = (experiment.results_data or {}).get("v3", {})
 
         selected_reference_branch = self.request.GET.get(
             "reference_branch", experiment.reference_branch.slug


experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -728,7 +728,9 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         experiment = self.get_object()
 
-        analysis_data = experiment.results_data.get("v3", {})
```
yashikakhurana, line 753: wasn't this was returning {} too for the fallback? not sure the advantage of the check here

moibra05: `experiment.results_data` can be null so calling `.get()` on it causes errors

yashikakhurana: I see in that case you can do this `analysis_data = (experiment.results_data or {}).get("v3", {})`

PR #14261, opened 2025-12-31
fix(nimbus): hide weekly metrics from displaying on results page popout card if not available

### LINKED ISSUE #14257: Weekly metrics card appearing when empty
Not all metrics have weekly data. The current new results page implementation does not consider this and always renders the “Weekly breakdown” popout card even when it contains no data. Fix this to hide this card if a metric does not contain any weekly data

!image-20251231-173734.png|width=700,alt="image-20251231-173734.png"!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6298)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/106578">image-20251231-173734.png</a>

moibra05:
Because

- The current new results page ui assumes that all metrics contain weekly data which is not true. This results in empty "Weekly breakdown" sections that serve no purpose.

This commit

- Conditionally renders the "Weekly breakdown" component if the specific metric supports weekly results

Fixes #14257

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1659,11 +1659,21 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
                     abs_list = branch_data.get("absolute") or []
                     rel_list = branch_data.get("relative") or []
 
-                    weekly_data[branch_slug] = list(
-                        zip_longest(abs_list, rel_list, fillvalue=None)
-                    )
+                    if abs_list or rel_list:
+                        weekly_data[branch_slug] = list(
+                            zip_longest(abs_list, rel_list, fillvalue=None)
+                        )
 
-                weekly_metric_data[metric_metadata["slug"]] = weekly_data
+                if weekly_data:
+                    weekly_metric_data[metric_metadata["slug"]] = {
+                        "has_weekly_data": True,
+                        "data": weekly_data,
+                    }
+                else:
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -1692,11 +1659,13 @@ def get_weekly_metric_data(self, analysis_basis, segment, reference_branch):
...
-                        zip_longest(abs_list, rel_list, fillvalue=None)
-                    )
+                    if abs_list or rel_list:
+                        weekly_data[branch_slug] = list(
+                            zip_longest(abs_list, rel_list, fillvalue=None)
+                        )
 
-                weekly_metric_data[metric_metadata["slug"]] = weekly_data
+                if weekly_data:
+                    weekly_metric_data[metric_metadata["slug"]] = weekly_data
```
yashikakhurana, line 1677: you know what you can also add a property like is weekly_data_available and then renders accordingly, just a suggestion

moibra05: yes i originally thought to do something like this but figured that since id only be using that property in a single area it might be simpler to just not include weekly data if it exists so that i can just use a simple `if weekly_metric_data` check and not have to worry about additional flags or default values

yashikakhurana: for the maintainability purpose, would suggest to have the property as it helps other to easily work with it in the future, otherwise have to dig into templates and logic to understand it

PR #14262, opened 2026-01-02
fix(nimbus): fix absolute value being called on None

### LINKED ISSUE #14240: Absolute value calculated on nonexistent points
Some of the data points are missing the "point" value that the code calls abs() on to compute the relative metric change; because that lookup returns None, abs(None) raises a TypeError and breaks rendering. Need to add a guard against missing/null "point" values before computing

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6291)

moibra05:
Because

- Some metric points do not contain the `point` field and are throwing errors b/c the current code assumes otherwise
- The issue here is that `abs()` is attempting to be called on these values which throw errors when `point` is None

This commit

- Adds a guard to check if the accessed point is None before calling `abs`

Fixes #14240

### DIFF
--- a/experimenter/experimenter/nimbus_ui/templatetags/nimbus_extras.py
+++ b/experimenter/experimenter/nimbus_ui/templatetags/nimbus_extras.py
@@ -335,6 +335,8 @@ def format_string(value, arg):
 
 @register.filter
 def to_percentage(value, precision=None):
+    if value is None or type(value) not in (int, float):
+        return "N/A"
     percentage_value = value * 100
 
     if precision is None:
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/nimbus_ui/templatetags/nimbus_extras.py
```diff
@@ -335,6 +335,8 @@ def format_string(value, arg):
 
 @register.filter
 def to_percentage(value, precision=None):
+    if value is None or type(value) not in [int, float]:
```
freshstrangemusic:
```suggestion
    if value is None or type(value) not in (int, float):
```

tuples are cheaper than lists because they are not mutable.

PR #14266, opened 2026-01-05
feat(nimbus): stricter status transitions

### LINKED ISSUE #14227: Stricter state transitions
We've recently seen cases where multiple people interacting quickly with the UI can create duplicate state transitions, ie two users very quickly hitting the launch buttons at the same time.  We should make sure the state transition mixin strongly enforces which required states an experiment can be in before transitioning to the next state so we only make one correct state transition.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6287)

jaredlockhart:
Becuase

* With more experiments comes more collisions with multiple users pressing buttons quickly
* We do not have strict requirements for each status transition form
* This means things can be approved multiple times by multiple users concurrently
* This can leave experiments in broken states if users collide
* We should add strict conditions to every state transition form so a state transition can only occur once

This commit

* Adds required entry states for every status transition form based on the states defined in the diagrams
* Updates the UI to show more detailed information if a status collision occurs
* Updates tests

fixes #14227

### DIFF
--- a/experimenter/experimenter/nimbus_ui/constants.py
+++ b/experimenter/experimenter/nimbus_ui/constants.py
@@ -22,6 +22,10 @@ Optional - We believe this outcome will <describe impact> on <core metric>
         "Name maps to a pre-existing slug, please choose another name."
     )
     ERROR_TAG_DUPLICATE_NAME = "Tag with this Name already exists."
+    ERROR_INVALID_STATE_TRANSITION = (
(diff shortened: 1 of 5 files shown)


yashikakhurana (approved):
Thank you @jaredlockhart for adding this, I tried couple of scenario and seems that it is not allowing invalid states now 🎉

PR #14270, opened 2026-01-05
chore(cirrus): use cirrus image built by make in docker compose

### LINKED ISSUE #14269: make cirrus_build isn't used by docker compose
@freshstrangemusic and @relud both had the following error, due to cirrus image being outdated:
```
Traceback (most recent call last):
  File "/cirrus/cirrus/generate_docs.py", line 15, in <module>
    from cirrus.main import app
  File "/cirrus/cirrus/main.py", line 15, in <module>
    from .glean.server_events import (
ModuleNotFoundError: No module named 'cirrus.glean.server_events'
```

right now the only way to fix this is to run `docker compose build cirrus-experimenter`, but it should be fixable by running `make cirrus_build` or `make cirrus_build_dev`, and the appropriate target should be a dependency of `make up` and `make up_detached`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6301)

relud:
Because

- outdated cirrus docker images should be automatically built by make

This commit

- Changes make targets to depend on `cirrus_build`
- Changes the docker compose files to use the image generated from `make cirrus_build`, which matches how experimenter does it
- Changes make targets that don't need dependencies use `--no-deps` so they don't need to depend on `cirrus_build` because cirrus won't be started
- Adds make target `cirrus_up_detached`

Fixes #14269

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ build_prod: ssl build_megazords
 	$(DOCKER_BUILD) --target deploy -f experimenter/Dockerfile -t experimenter:deploy experimenter/
 
 compose_stop:
-	$(COMPOSE) kill || true
+	$(CIRRUS_ENABLE) $(COMPOSE) kill || true
 	$(COMPOSE_INTEGRATION) kill || true
 	$(COMPOSE_PROD) kill || true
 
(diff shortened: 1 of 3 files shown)


relud:
> if we're gonna run the experimenter-cirrus as part of `up` then we gotta also update `kill` to kill/rm it so it doesn't linger around between runs

fixed, but I think that container is from `cirrus_up` not `up`, either way it's useful to include it in `kill`

PR #14271, opened 2026-01-05
chore(nimbus): bump minimum version to 105

### LINKED ISSUE #13267: Bump minimum version to 105
We have a bunch of logic that checks the minimum version a bunch of different capabilities were added to the system, but now that certs have rotated for clients above 102, rollouts are only available after 105, we should bump the system wide minimum supported version to 105.

jaredlockhart:
Becuase

* Remote Settings Certs were rotated and collectsion are now only readable by clients >=105
* We can bump the minimum supported version to 105
* We can remove any logic which verified versions <=105

This commit

* Bumps the minimum supported version to 105
* Removes any review logic that checked for versions <=105
* Updates tests

fixes #13267

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1877,7 +1877,7 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError(
                 {
                     "firefox_min_version": [
-                        NimbusExperiment.ERROR_FIREFOX_VERSION_MIN_96
+                        NimbusExperiment.ERROR_FIREFOX_VERSION_MIN_SUPPORTED
                     ],
                 }
             )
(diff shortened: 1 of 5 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1938,46 +1938,6 @@ def _validate_versions(self, data):
...
-
-    def _validate_countries_versions(self, data):
-        application = data.get("application")
-        min_version = data.get("firefox_min_version", "")
-
-        if data.get("countries", []):
-            min_supported_version = (
-                NimbusConstants.COUNTRIES_APPLICATION_SUPPORTED_VERSION[application]
-            )
-            if NimbusExperiment.Version.parse(
```
yashikakhurana, line 1969: dont we want to keep this validation check @jaredlockhart 🤔

jaredlockhart: No these checks all checked for things below version 105, which is now the minimum, so they're not necessary anymore.

yashikakhurana (approved):
perfect, thank you @jaredlockhart for explaining it that too

PR #14278, opened 2026-01-06
feat(nimbus): Add collaborators management to feature pages

### LINKED ISSUE #14030: Add collaborators to feature page
We need a drop down of all users that lets person A add person B as a subscriber to a feature on a feature page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6216)

yashikakhurana:
Because

- Users needed a way to add other people as subscribers to feature configurations
- Previously, users could only subscribe/unsubscribe themselves, and no one could see who else was subscribed to the feature

This commit

- Adds a collaborators dropdown on the feature page, allowing users to add/remove any user as a subscriber
- Single card section to subscribe/unsubscribe and add collaborators

Fixes #14030

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -1929,8 +1929,8 @@ class CollaboratorsForm(NimbusChangeLogFormMixin, forms.ModelForm):
         return f"{self.request.user} updated collaborators"
 
 
-class FeatureCollaboratorsForm(forms.ModelForm):
-    collaborators = forms.ModelMultipleChoiceField(
+class FeatureSubscribersForm(forms.ModelForm):
+    subscribers = forms.ModelMultipleChoiceField(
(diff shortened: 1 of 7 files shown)


yashikakhurana:
https://github.com/user-attachments/assets/01e00c76-6ff8-45b5-8e67-aa2b851976c4

yashikakhurana:
> Okay tested locally, works perfectly. Couple thoughts
> 
>     1. Maybe call it 'Subscribers'?  It's not really the same as 'collaborators' on an experiment since you can't actually make changes to it you're just subscribing to the notifications?
> 
>     2. Having the 'subscribe' button next to the drop down, where your own email appears so you can subscribe yourself either way is a bit redundant/confusing.  If we're gonna put the users own email in the drop down and show it in the last then probably take away the subscribe button?


makes sense, here is the new design 
Screenshot 2026-01-07 at 10 28 25 AM

PR #14280, opened 2026-01-06
bug(cirrus): add nimbus user id as separate metric for shredder

### LINKED ISSUE #14279: Add nimbus user id as non-event-extra metric to cirrus pings
shredder can't use event extras as the id to delete on because accessing extras causes bigquery to throw:
```
google.api_core.exceptions.BadRequest: 400 [...]: Correlated subqueries that reference other tables are not supported unless they can be de-correlated, such as by transforming them into an efficient JOIN.
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6305)

relud:
Because

- shredder needs nimbus user id to be accessible in a non-map field (i.e. not in event extras)

This commit

- Changes cirrus to send nimbus user id metric on enrollment and enrollment-status pings
- Adds a new record_enrollment_statuses_v2 method on CirrusMetricsHandler so that rust can pass nimbus user id in when there's no events

Fixes #14279

### DIFF
--- a/cirrus/server/telemetry/metrics.yaml
+++ b/cirrus/server/telemetry/metrics.yaml
@@ -104,3 +104,21 @@ cirrus_events:
     send_in_pings:
       - startup
 
+nimbus:
+  nimbus_user_id:
+    description: The ID of the user requesting enrollment
+    type: string
+    bugs:
+      - https://mozilla-hub.atlassian.net/browse/EXP-6305
+      - https://github.com/mozilla/experimenter/issues/14279
+    data_reviews:
+      - https://github.com/mozilla/experimenter/pull/9081#issuecomment-1625751843
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - jlockhart@mozilla.com
+      - project-nimbus@mozilla.com
+    expires: never
+    send_in_pings:
+      - enrollment
+      - enrollment-status
--- a/cirrus/server/cirrus/main.py
+++ b/cirrus/server/cirrus/main.py
@@ -200,10 +200,13 @@ def collate_enrollment_metric_data(
     return data
 
 
-async def record_metrics(enrollment_data: list[EnrollmentMetricData]):
+async def record_metrics(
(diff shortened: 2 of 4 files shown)


cirrus/server/telemetry/metrics.yaml
```diff
@@ -104,3 +104,21 @@ cirrus_events:
     send_in_pings:
       - startup
 
+nimbus:
+  nimbus_user_id:
```
relud, line 108: this causes the metric to be named `nimbus_nimbus_user_id` in python kwargs and bigquery, which is unfortunately necessary because otherwise it would conflict with the python kwarg for the `nimbus_user_id` event extra in one of the generated glean apis we don't use and the python parser would reject the whole file

Makefile
```diff
@@ -308,7 +308,7 @@ cirrus_check_and_report: cirrus_lint
 	cp workspace/test-results/cirrus_coverage.json $(CIRRUS_COVERAGE_JSON)
 
 cirrus_code_format: cirrus_build
-	$(CIRRUS_ENABLE) $(COMPOSE_RUN) cirrus sh -c '$(CIRRUS_RUFF_FIX) && $(CIRRUS_BLACK_FIX)'
+	$(CIRRUS_ENABLE) $(COMPOSE_RUN) cirrus sh -c '$(CIRRUS_BLACK_FIX) && $(CIRRUS_RUFF_FIX)'
```
relud: without this ruff will prevent black from fixing things like line too long

jaredlockhart: Oh!  Black should be deprecated everywhere now in favour of ruff, this is probably just lingering.  I'd move this into a separate PR that just swaps all black invocations for cirrus to ruff.

It's becuase there was a period where ruff didn't handle all of blacks formatting rules so we had both, but I think now that's no longer true and I guess we dropped black for ruff in experimenter but not cirrus.

relud: sounds good, removed from this PR

jaredlockhart (approved):
Anyways yes looks good ty @relud 🙏 🎉

PR #14287, opened 2026-01-07
chore(cirrus): switch from black to ruff format

### LINKED ISSUE #14286: Switch cirrus from ruff to black
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6307)

relud:
Because

- Experimenter has dropped black in favor of ruff format, and cirrus should match that

This commit

- Changes cirrus to use ruff format instead of black, and removes black dependency
- Moves select and ignore from tool.ruff section of pyproject.toml to tool.ruff.lint section to fix warning
- Changes make target `cirrus_bash` to specify glean generated code directory as a volume so it doesn't get hidden by the other volume mounts

Fixes #14286


cirrus/server/pyproject.toml
```diff
@@ -44,6 +40,9 @@ pythonVersion = "3.10"
 
 [tool.ruff]
 exclude = ["cirrus/glean"]
+line-length = 90
+
+[tool.ruff.lint]
```
yashikakhurana, line 50: can we apply the same rules as we have for experimenter https://github.com/mozilla/experimenter/blob/1ac5eda18d71ec1854089e1056544c3be603a792/experimenter/pyproject.toml#L152

jaredlockhart: Oh good point @yashikakhurana 👍

yashikakhurana: https://github.com/mozilla/experimenter/blob/1ac5eda18d71ec1854089e1056544c3be603a792/experimenter/pyproject.toml#L157

relud: sure, fixed

experimenter/pyproject.toml
```diff
@@ -204,7 +204,6 @@ ignore = [
     "SIM102",
     "UP006",
     "UP007",
-    "UP038",
```
relud, line 207: removed because warning says this check no longer exists

jaredlockhart (approved):
Fantastic 🎉 🎉 🎉 🎉

PR #14300, opened 2026-01-08
feat(nimbus): Show test url even if QA status is not set

### LINKED ISSUE #14235: Setting testrail urls do not appear in feature page without qa status
If you set the new test rail url fields but leave QA Status as 'not set', no urls appear in in the 'qa runs' table on the feature page.  We should always show any data we have.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6290)

yashikakhurana:
Because

- If you set the new test rail url fields but leave QA Status as 'not set', no urls appear in in the 'qa runs' table on the feature page.

This commit

- Enables the entry of QA section if we have the urls available

Fixes #14235


yashikakhurana:
Screenshot 2026-01-07 at 1 51 01 PM

PR #14320, opened 2026-01-10
feat(nimbus): update Firefox Suggest advanced targeting

0c0w3:
Because

* Online Firefox Suggest has been redesigned on the desktop client
* We need to target users who have (1) online Firefox Suggest enabled and (2) accepted the ToU relevant to Firefox Suggest
* We need to target users who have (1) online Firefox Suggest enabled, (2) Firefox Suggest sponsored suggestions enabled, and (3) accepted the ToU relevant to Firefox Suggest

This commit

* Removes outdated targeting involving "data collection enabled" except for one filter that is currently used by a live experiment
* Removes targeting involving only sponsored suggestions
* Adds two new target filters for the points above

Please note that the ToU parts are copied from the existing `ACCEPTED_TOU_ON_OR_AFTER_DEC_15_2025` filter.

Fixes https://mozilla-hub.atlassian.net/browse/SNG-3164


0c0w3:
@ncloudioj If you're able to double-check me here it would be appreciated!

PR #14359, opened 2026-01-14
bug(nimbus): fix show more/less buttons on readonly json

### LINKED ISSUE #14358: Read only json show more button not visible
I just noticed the 'Show More' button isn't visible on the collapsed read only JSON fields

Image

Probably got lost when adding the copy contents buttons.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6340)

jaredlockhart:
Becuase

* We recently added some content copy buttons to the read only json
* We accidentally broke the show more/less buttons

This commit

* Restores the show/more less buttons
* Fixes the width of the readonly json fields on the detail page so they line up more nicely

fixes #14358


jaredlockhart:
image

relud (approved):

PR #14361, opened 2026-01-14
feat(cirrus): Add type hints for SDK.compute_enrollments

### LINKED ISSUE #13717: Use new participation API type in Cirrus
Becuase Cirrus also uses the Nimbus SDK and the Nimbus SDK is now using a participation enum instead of just a boolean, we need to update cirrus to use this new type

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6039)

freshstrangemusic:
Because:

- The Cirrus API is stringly-typed;
- we had a case where the types were out of sync; and
- there were some misnamed variables, implying they were of a different
  type

this commit:

- updates poetry in the Cirrus docker image to 2.2.1;
- updates Cirrus to require Python 3.11 (which mostly just affects
  pyright);
- updates pyright to use Python 3.11;
- updates to the latest pyright; and
- adds type hints to the `SDK.compute_enrollments` method.

Fixes #13717

### DIFF
--- a/cirrus/server/cirrus/sdk.py
+++ b/cirrus/server/cirrus/sdk.py
@@ -1,6 +1,7 @@
 import json
 import logging
-from typing import Any
+from enum import StrEnum
+from typing import Any, NotRequired, TypedDict
 
 from cirrus_sdk import (  # type: ignore
     CirrusClient,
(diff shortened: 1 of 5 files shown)


cirrus/server/cirrus/sdk.py
```diff
@@ -12,6 +13,50 @@
 logger = logging.getLogger(__name__)
 
 
+class ExperimentEnrollment(TypedDict):
+    slug: str
+    status: str
```
mikewilli, line 18: Should these live in the schemas package?

freshstrangemusic: Yeah that might be for the best going forward, but lets just sneak these in for now and file a follow-up.

PR #14362, opened 2026-01-14
chore(cirrus): use remote settings changeset api

relud:
Because

- RS team advises readers are expected to use changeset api not records api

This commit

- Changes cirrus to expect a remote settings response with records under "changes" and temporarily fall back to "data" for backwards compatibility

For #14363

### DIFF
--- a/cirrus/server/cirrus/experiment_recipes.py
+++ b/cirrus/server/cirrus/experiment_recipes.py
@@ -45,7 +45,9 @@ class RemoteSettings:
         try:
             response = requests.get(self.url)
             response.raise_for_status()
-            data = response.json().get("data")
+            response_json = response.json()
+            # While moving from records api to changeset api, support both
+            data = response_json.get("changes", response_json.get("data"))
             if data is not None:
                 self.update_recipes({"data": data})
                 logger.info(f"Fetched resources: {data}")
--- a/cirrus/server/tests/test_experiment_recipes.py
+++ b/cirrus/server/tests/test_experiment_recipes.py
@@ -32,10 +32,14 @@ def test_update_recipes(remote_settings):
     ["remote_settings_live", "remote_settings_preview"],
     indirect=True,
 )
-def test_empty_data_key(mock_get, remote_settings):
+@pytest.mark.parametrize(
+    "rs_response_key",
+    ["data", "changes"],
+)
+def test_empty_data_key(mock_get, remote_settings, rs_response_key):
     mock_response = MagicMock()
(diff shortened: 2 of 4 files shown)


relud:
this assumes that `?_expected=0` is set as part of the remote settings url so that it returns changes since the epoch, which minimizes the code changes needed here

cirrus/server/cirrus/experiment_recipes.py
```diff
@@ -45,7 +45,9 @@ def fetch_recipes(self) -> None:
         try:
             response = requests.get(self.url)
             response.raise_for_status()
-            data = response.json().get("data")
+            response_json = response.json()
+            # While moving from records api to changeset api, support both
+            data = response_json.get("changes", response_json.get("data"))
```
relud, line 50: after updating stage and prod urls this can be reduced to `data = response.json().get("changes")`

cirrus/server/tests/test_experiment_recipes.py
```diff
@@ -35,7 +35,7 @@ def test_update_recipes(remote_settings):
 def test_empty_data_key(mock_get, remote_settings):
     mock_response = MagicMock()
     mock_response.status_code = 200
-    mock_response.json.return_value = {"data": []}
+    mock_response.json.return_value = {"changes": []}
```
jaredlockhart: Well shouldn't we have tests for both?  Ie like, duplicate set of tests, or parameterize them?

jaredlockhart (changes requested):

cirrus/server/tests/test_experiment_recipes.py
relud: seemed like overkill if i'm planning to remove the old one shortly after this deploys

relud: but sure i can parametrize

cirrus/server/tests/test_experiment_recipes.py
```diff
@@ -32,10 +32,14 @@ def test_update_recipes(remote_settings):
     ["remote_settings_live", "remote_settings_preview"],
     indirect=True,
 )
-def test_empty_data_key(mock_get, remote_settings):
+@pytest.mark.parametrize(
+    "rs_response_key",
+    ["data", "changes"],
```
jaredlockhart, line 37: Better 🙏

jaredlockhart (approved):
Okay cool thank you, clean simple update thnx @relud 🙏

PR #14384, opened 2026-01-16
feat(nimbus): link contact experimenter support button to ask-experimenter slack channel

### LINKED ISSUE #14373: Implement "Contact Experimenter Support" functionality in error cards
When experiments encounter errors the new results ui has a card with a button to contact experimenter support. As of now, this button does nothing. Update it to ping the #ask-experimenter slack channel

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6304)

moibra05:
Because

- "Contact Experimenter Support" button which appears when an experiment has errors wasn't functional

This commit

- Makes clicking on the "Contact Experimenter Support" button take you to the ask-experimenter slack channel

Fixes #14373

### DIFF
--- a/experimenter/experimenter/settings.py
+++ b/experimenter/experimenter/settings.py
@@ -582,3 +582,8 @@ SUMMERNOTE_CONFIG = {
         ],
     },
 }
+
+ASK_EXPERIMENTER_SLACK_LINK = config(
+    "ASK_EXPERIMENTER_SLACK_LINK",
+    default="https://experimenter.info/",
+)
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new-fragment.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new-fragment.html
@@ -10,12 +10,12 @@
           <p class="text-muted">
             Your experiment is still running. Results for some metrics aren't available, but others are unaffected.
           </p>
-          <button class="btn btn-secondary bg-secondary-subtle bg-opacity-25 border-0 text-body fw-semibold">
-            Contact Experimenter Support
-          </button>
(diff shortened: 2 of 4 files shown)


experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -206,6 +206,10 @@ class NimbusUIConstants:
         ),
     }
 
+    ASK_EXPERIMENTER_SLACK_LINK = (
+        "https://mozilla.enterprise.slack.com/archives/CF94YGE03"
+    )
+
```
yashikakhurana: I would say we should move this to env

yashikakhurana: and then read for the env

experimenter/experimenter/settings.py
```diff
@@ -582,3 +582,8 @@
         ],
     },
 }
+
+ASK_EXPERIMENTER_SLACK_LINK = config(
+    "ASK_EXPERIMENTER_SLACK_LINK",
+    default="https://mozilla.enterprise.slack.com/archives/CF94YGE03",
```
yashikakhurana: oh to be clear, I intended that we should not add a link to the slack channel in the public repository, we may also need to update the env value for stage and prod

moibra05: ah i see, where exactly should i be looking to make such a a change?

yashikakhurana: so exactly how you are reading the config here, just dont add default or add default may be to experimenter.info something, then we can update the env on stage and prod

yashikakhurana (approved):
perfect, lets add a config for both stage and prod after merging of this

PR #14388, opened 2026-01-19
bug(nimbus): Update the advanced targeting config for ANDROID_EXISTING_USERS_NOT_ACCEPTED_TERMS_OF_USE_ONE_POINT

t-p-white:
Because

- The `ANDROID_EXISTING_USERS_NOT_ACCEPTED_TERMS_OF_USE_ONE_POINT` targeting does not **exclude users with more than 1 point**, which prevents users eligible for `ANDROID_EXISTING_USERS_NOT_ACCEPTED_TERMS_OF_USE_OVER_ONE_POINT` from being enrolled.

This commit

- Move the 'has ad-blocker' condition to a variable.
- Adds another condition to exclude users with more than 1 point. 

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=2010568


jaredlockhart (approved):

t-p-white:
Thanks @jaredlockhart 🚀

PR #14392, opened 2026-01-19
feat(nimbus): add risk ai field

### LINKED ISSUE #14390: Add a radio button in Experimenter that says, this delivery labs involves AI.
Let’s add a radio in the overview section that says

`Does this experiment use any AI features/functionality? O Yes O No`

If the user selects Yes then we can add the following to the targeting:

`'browser.ai.control.default'|preferenceValue == 'available'`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6360)

jaredlockhart:
Because

* We want to mark every experiment/rollout if it involves AI features
* We can add a targeting clause that excludes users that have opted out of AI features for desktop v148+

This commit

* Adds a risk field to mark experiments/rollouts that involve AI features
* Adds a pref targeting clause for desktop experiments that involve AI for v148+
* Moves all risk questions together with the AI question first
* Unsets risk questions after cloning so they must always be re evaluated before launching

fixes #14390


jaredlockhart:
image
image

PR #14393, opened 2026-01-19
    bug(nimbus): set broken preview experiments back to draft

### LINKED ISSUE #14381: Stage experiment in broken preview state
Looks like there's an experiment on stage in a broken preview state:

https://stage.experimenter.nonprod.webservices.mozgcp.net/nimbus/ppop-multichannel-adventure/summary/

```
Cannot perform this action: experiment must be in state (NimbusConstants.Status.DRAFT, None, NimbusConstants.PublishStatus.IDLE, False), but is currently in state ('Draft', 'Preview', 'Idle', False).
```

There could be more lingering on stage/prod so let's write a data migration to recover those.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6353)

jaredlockhart:
Becuase
    
* At some point there must have been a bug in the UI that was mistakenly setting publish_status to Preview which is not a valid state
* Experiments in that state can now not be modified because of the new strict transition checks
* We can recover those experiments back to Draft with a data migration
    
This commit
    
* Creates a data migration to set experiments with publish_status=Preview back to draft/idle
    
fixes #14381


jaredlockhart:
Blocked on https://github.com/mozilla/experimenter/pull/14392/changes which also introduces a migration

PR #14420, opened 2026-01-21
fix(nimbus): fix exposures being default for experiments without correct data

### LINKED ISSUE #14418: Experiments still defaulting to exposures when no data present
┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6376)

moibra05:
Because

- Previously exposure data was simply checked for by verifying whether the results data object had a non-null value for its exposure field
- This check was not enough as some experiments contain invalid exposure data in the form of missing metrics and containing only values of `{ "point": 0 }`

This commit

- Updates the `has_exposures` method to perform a deeper check of exposure data validity by validating that the client count for each exposure branch is at least 10
- Shows an error if experiments have exposures incorrectly setup by only enrolling clients into a proper subset of branches

Fixes #14418

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1710,16 +1710,45 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
     @property
     def has_exposures(self):
-        # True if there are any exposures in the results data
+        # Returns the enum corresponding to exposures state:
+        # - VALID when exposures exist for all branches
+        # - INVALID when exposures exist for some but not all branches
+        # - NO_EXPOSURES when no exposures data is present
+
         if self.results_data and "v3" in self.results_data:
             results_data = self.results_data["v3"]
             for window in ["overall", "weekly"]:
                 if results_data.get(window):
-                    exposure_data = results_data[window].get("exposures", {}).get("all")
-                    if exposure_data is not None:
-                        return True
+                    exposures_branch_data = (
+                        results_data[window].get("exposures", {}).get("all", {})
+                    )
+                    branches_with_exposures = 0
 
-        return False
+                    for branch_data in exposures_branch_data.values():
+                        client_count = (
+                            branch_data.get("branch_data", {})
+                            .get("other_metrics", {})
+                            .get("identity", {})
+                            .get("absolute", {})
+                            .get("first", {})
+                            .get("point", 0)
+                        )
+
+                        # Treat a branch as having meaningful exposures only when the
+                        # client_count exceeds a set cutoff. Manual testing can produce
+                        # tiny non-zero counts, so we use EXPOSURE_CLIENT_CUTOFF as an
+                        # arbitrary threshold to avoid false positives.
+                        if client_count > NimbusConstants.EXPOSURE_CLIENT_CUTOFF:
+                            branches_with_exposures += 1
+
+                    if branches_with_exposures == self.branches.all().count():
+                        return NimbusUIConstants.ExposuresStatus.VALID
+                    # Handles the case where exposures were only implemented for some but
+                    # not all branches.
+                    elif branches_with_exposures > 0:
+                        return NimbusUIConstants.ExposuresStatus.INVALID
+
+        return NimbusUIConstants.ExposuresStatus.NO_EXPOSURES
 
     @property
     def show_results_url(self):
--- a/experimenter/experimenter/nimbus_ui/constants.py
+++ b/experimenter/experimenter/nimbus_ui/constants.py
@@ -1,4 +1,4 @@
-from enum import Enum
+from enum import Enum, IntEnum
 
 from experimenter.experiments.constants import NimbusConstants
 
@@ -213,6 +213,11 @@ Optional - We believe this outcome will <describe impact> on <core metric>
         LAUNCH_ROLLOUT = "launch this rollout"
         UPDATE_ROLLOUT = "update this rollout"
         END_ROLLOUT = "end this rollout"
+
+    class ExposuresStatus(IntEnum):
+        NO_EXPOSURES = 0
+        VALID = 1
+        INVALID = 2
 
 
 def _get_qa_status_icon_map():
--- a/experimenter/experimenter/nimbus_ui/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui/tests/test_views.py
@@ -3192,11 +3192,36 @@ class TestResultsView(AuthTestCase):
                         },
                         "overall": {
                             "enrollments": {"all": {}},
-                            "exposures": {"all": {}},
-                        },
-                        "weekly": {
-                            "enrollments": {"all": {}},
-                            "exposures": {"all": {}},
+                            "exposures": {
+                                "all": {
+                                    "control": {
+                                        "branch_data": {
+                                            "other_metrics": {
+                                                "identity": {
+                                                    "absolute": {
+                                                        "first": {
+                                                            "point": 210,
+                                                        }
+                                                    }
+                                                }
+                                            }
+                                        }
+                                    },
+                                    "treatment": {
+                                        "branch_data": {
+                                            "other_metrics": {
+                                                "identity": {
+                                                    "absolute": {
+                                                        "first": {
+                                                            "point": 210,
+                                                        }
+                                                    }
+                                                }
+                                            }
+                                        }
+                                    },
+                                }
+                            },
                         },
                     }
                 },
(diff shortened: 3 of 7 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -2121,10 +2121,21 @@ def has_exposures(self):
...
+                        .get(
+                            self.reference_branch.slug if self.reference_branch else "",
+                            {},
+                        )
+                        .get("branch_data", {})
+                    )
+                    for metrics in exposure_data.values():
+                        for data in metrics.values():
+                            for data_point in data.get("absolute", {}).get("all", {}):
+                                if data_point != {"point": 0}:
```
mikewilli: I think you want to get `"point"` out of here and check its value instead of checking this object. If I'm not mistaken, there could be `"lower"` and `"upper"` values here too that might all be 0 and so `{"point": 0} != {"lower": 0, "point": 0, "upper": 0}` will evaluate to `True`, but that's not really what you want.

experimenter/experimenter/experiments/models.py
```diff
@@ -2121,10 +2121,21 @@ def has_exposures(self):
...
+                        .get("all", {})
+                        .get(
+                            self.reference_branch.slug if self.reference_branch else "",
+                            {},
+                        )
+                        .get("branch_data", {})
+                    )
+                    for metrics in exposure_data.values():
+                        for data in metrics.values():
+                            for data_point in data.get("absolute", {}).get("all", {}):
```
mikewilli:
```suggestion
                            for data_point in data.get("absolute", {}).get("all", []):
```

experimenter/experimenter/nimbus_ui/tests/test_views.py
mikewilli, line 3263: Since this test didn't catch the problem before, you will want to update the parametrized results_data list to include options that would have displayed incorrect behavior before but now are correct. The update you made fixes the true positive case, but you should also keep the old version and ensure that the corresponding analysis basis is `enrollments`.

mikewilli (changes requested):

experimenter/experimenter/experiments/models.py
moibra05: im not sure if there’s a consistent structure among the experiments that incorrectly default to exposures. I checked 3 experiments that were doing this, and they all seem to be filled with {"point": 0} objects

mikewilli: Ok, if this is a very specific case then that's fine. You might want to put a comment in here then to explain why you're specifically checking for this, since it's not intuitive (at least to me).

mikewilli:
@jaredlockhart Yea, that'd be another way to check. I think either is effective, but checking client counts on all branches (or perhaps all non-control branches since teams sometimes forget to send exposure events for control) would probably be more efficient. You'd probably want to check something greater than 0 though because there are plenty of examples where the client counts are very low but non-zero due to manual testing.

Something like this: `[analysis period].exposures.all.[branch name].branch_data.other_metrics.identity.absolute.first.point > 10`

jaredlockhart:
@mikewilli Yeah exactly, also I have seen the case where control has 0 exposed clients whereas treatments will, in that case is the exposure basis analysis correct/valid?  Can it compare without any exposures in control?

moibra05:
> @mikewilli Yeah exactly, also I have seen the case where control has 0 exposed clients whereas treatments will, in that case is the exposure basis analysis correct/valid? Can it compare without any exposures in control?

+1 to this question. found an [experiment](https://experimenter.services.mozilla.com/nimbus/new-tab-automated-personalization-v3/results-new/) that has this exact thing happening to but not sure whether to consider it valid or not

mikewilli:
>@mikewilli Yeah exactly, also I have seen the case where control has 0 exposed clients whereas treatments will, in that case is the exposure basis analysis correct/valid? Can it compare without any exposures in control?

Right, that's what I meant by "teams sometimes forget to send exposure events for control". I think handling this is a bit nuanced because it means there are incomplete results, but they could still be useful (so they shouldn't be hidden entirely). These results could still be used to compare different treatment branches (if there are multiple non-control branches), or just to see absolute numbers for exposure metrics.

Maybe the best way to handle this is to show an error in the top box that shows potential issues, and default to Enrollments in the dropdown, but make sure that Exposures is still selectable in the dropdown.

moibra05:
image

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -214,6 +214,11 @@ class ReviewRequestMessages(Enum):
         UPDATE_ROLLOUT = "update this rollout"
         END_ROLLOUT = "end this rollout"
 
+    class ExposuresStatus(Enum):
+        NONE = 0
+        VALID = 1
+        INVALID = 2
```
mikewilli: You should be able to use `IntEnum` as the base class and then not worry about having to do `.value` everywhere, it should automatically act like an integer value when you access the members. See https://docs.python.org/3/library/enum.html#enum.IntEnum
```suggestion
    class ExposuresStatus(IntEnum):
        NONE = 0
        VALID = 1
        INVALID = 2
```

experimenter/experimenter/experiments/models.py
```diff
@@ -1710,11 +1715,30 @@ def has_exposures(self):
...
+                        client_count = (
+                            branch_data.get("branch_data", {})
+                            .get("other_metrics", {})
+                            .get("identity", {})
+                            .get("absolute", {})
+                            .get("first", {})
+                            .get("point", 0)
+                        )
+
+                        if client_count > 10:
```
mikewilli: I think a comment here is warranted since it might be a bit confusing to someone why we chose 10 here instead of 0. Something to the effect of "make sure there are some exposed clients, and >10 to hopefully ensure there are real clients and not just a few from pre-launch testing".

Probably a good idea to make `10` a constant as well.

experimenter/experimenter/experiments/models.py
```diff
@@ -1710,11 +1715,30 @@ def has_exposures(self):
...
+                            branches_with_exposures += 1
+
+                    if branches_with_exposures == self.branches.all().count():
+                        return NimbusUIConstants.ExposuresStatus.VALID.value
+                    elif branches_with_exposures > 0:
+                        return NimbusUIConstants.ExposuresStatus.INVALID.value
```
mikewilli: Another spot where it's worth having a comment about why this might occur: "handles the not-uncommon case where exposures were only implemented for treatment branch(es)"

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -214,6 +214,11 @@ class ReviewRequestMessages(Enum):
         UPDATE_ROLLOUT = "update this rollout"
         END_ROLLOUT = "end this rollout"
 
+    class ExposuresStatus(Enum):
+        NONE = 0
```
mikewilli: Since `None` is a pretty common word in python, maybe something more explicit like `NO_EXPOSURES`?

moibra05: good point ty

experimenter/experimenter/experiments/models.py
```diff
@@ -1705,16 +1710,41 @@ def has_displayable_results(self):
 
     @property
     def has_exposures(self):
-        # True if there are any exposures in the results data
+        # True if there are any meaningful exposures in the results data
```
mikewilli: This is incorrect, no? It's truthy if there are _any_ branches with exposures, falsy if none of the branches have exposures. I guess while you're there, you could update it to say it returns the enum value corresponding to whether exposures exist and are valid, exist but partially (invalid), or do not exist.

mikewilli: Or just say that it returns `ExposuresStatus` and put the comments there on what each member means.

experimenter/experimenter/experiments/models.py
```diff
@@ -1705,16 +1710,41 @@ def has_displayable_results(self):
...
+                            .get("other_metrics", {})
+                            .get("identity", {})
+                            .get("absolute", {})
+                            .get("first", {})
+                            .get("point", 0)
+                        )
+
+                        # Treat a branch as having meaningful exposures only when the
+                        # client_count exceeds a set cutoff. Manual testing can produce
+                        # tiny non-zero counts, so we use 10 as an arbitrary threshold
```
mikewilli:
```suggestion
                        # tiny non-zero counts, so we use this value as an arbitrary threshold
```

mikewilli: I'd suggest this change just in case the constant gets changed. You could reference the constant by name here if you'd rather.

PR #14427, opened 2026-01-21
feat(nimbus): use readonly codemirror for jexl targeting

### LINKED ISSUE #14391: Use readonly codemirror for JEXL targeting string
Let's make the JEXL targeting field on teh detail page into a readonly codemirror with a copy button.  Maybe we can use the Javascript syntax highlighting.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6364)

jaredlockhart:
Because

* It is very hard to read long JEXL targeting expressions without proper formatting
* We have the codemirror utility for displaying code in the UI
* We have the PyJEXL library that has an AST parser
* We can combine those to parse and format JEXL targeting more nicely

This commit

* Adds a JEXL formatter using the PyJEXL parser library similar to how we used it in the integration tests
* Adds a readonly codemirror to display the formatted JEXL targeting
* Adds tests

fixes #14391

### DIFF
--- a/experimenter/experimenter/experiments/jexl_utils.py
+++ b/experimenter/experimenter/experiments/jexl_utils.py
@@ -1,0 +1,88 @@
+import json
+
+from pyjexl.jexl import JEXLConfig
+from pyjexl.operators import default_binary_operators, default_unary_operators
+from pyjexl.parser import (
+    ArrayLiteral,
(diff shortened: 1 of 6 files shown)


jaredlockhart:
image

PR #14437, opened 2026-01-22
feat(nimbus): require min 148+ for ai experiments

### LINKED ISSUE #14436: Require minimum 148 for AI experiments
We added a risk question about whether an experiment/rollout involves AI that includes a filter that respects the AI killswitch, but the pref to target is only available in 148+.  Rather than support launching AI experiments to <148 that would not include the pref filter, let's just force a minimum required version of 148 for any experiment/rollout that answers yes to the AI risk question.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6385)

jaredlockhart:
Becuase

* We added a risk question for whether an experiment involves AI
* We added a targeting filter against the AI killswitch pref that's available in 148+
* This leaves open the possibility for an AI experiment to launch against 147 without the targeting filter and continue enrolling clients in 148 regardless of their AI killswitch setting
* We can require AI experiments to target minimum 148 and then always include the pref filter and just preclude launching AI experiments to 147

This commit

* Removes the version check on the targeting filter for AI
* Adds a review check that requires minimum 148+ for AI experiments

fixes #14436

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1944,6 +1944,27 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
             )
         return data
 
+    def _validate_risk_ai_version(self, data):
+        risk_ai = data.get("risk_ai")
+        firefox_min_version = data.get("firefox_min_version")
+        application = data.get("application")
+
+        if (
+            risk_ai
+            and application == NimbusExperiment.Application.DESKTOP
+            and firefox_min_version
+            and NimbusExperiment.Version.parse(firefox_min_version)
+            < NimbusExperiment.Version.parse(NimbusExperiment.AI_RISK_MIN_VERSION)
+        ):
+            raise serializers.ValidationError(
+                {
+                    "firefox_min_version": [
+                        NimbusConstants.ERROR_FIREFOX_VERSION_MIN_148_FOR_AI_RISK
+                    ],
+                }
+            )
+        return data
+
     def _validate_enrollment_targeting(self, data):
         excluded_experiments = set(data.get("excluded_experiments", []))
         required_experiments = set(data.get("required_experiments", []))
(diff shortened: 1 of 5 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1944,6 +1944,27 @@ def _validate_versions(self, data):
...
+        risk_ai = data.get("risk_ai")
+        firefox_min_version = data.get("firefox_min_version")
+        application = data.get("application")
+
+        if (
+            risk_ai
+            and application == NimbusExperiment.Application.DESKTOP
+            and firefox_min_version
+            and NimbusExperiment.Version.parse(firefox_min_version)
+            < NimbusExperiment.Version.parse(NimbusExperiment.Version.FIREFOX_148)
```
yashikakhurana: I believe we used to have a place where we define the version constant like min_ai_risk_version and then we used to refer here 🤔

jaredlockhart: I don't think so?  We had the 148 hard coded in the targeting check, but we're moving it here.  I don't see any other reference to a constant?

yashikakhurana: like this- https://github.com/mozilla/experimenter/blob/ff3fff064fb889d7f1b0d93a20e2b50525b4e838/experimenter/experimenter/experiments/constants.py#L867

yashikakhurana: https://github.com/mozilla/experimenter/blob/ff3fff064fb889d7f1b0d93a20e2b50525b4e838/experimenter/experimenter/experiments/constants.py#L1029

jaredlockhart: Oh you mean we should add one, not that one exists, I misunderstood 👍

yashikakhurana: yup yup, sorry I was not clear enough

yashikakhurana (approved):
perf, thanks @jaredlockhart

PR #14439, opened 2026-01-22
fix(nimbus): fix dbz errors on results page

### LINKED ISSUE #14438: ZeroDivisionError: float division by zero
Sentry Issue: [EXPERIMENTER-PROD-17H](https://mozilla.sentry.io/issues/7208016672/?referrer=github_integration)

```
ZeroDivisionError: float division by zero
(1 additional frame(s) were not displayed)
...
  File "experimenter/nimbus_ui/views.py", line 818, in get_context_data
    bar_width = ((upper - lower) / full_width) * 100
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6386)

moibra05:
Because

- `get_max_metric_value` can return 0, which results in a confidence_range of 0

This commit

- Adds a defensive fallback value in case `confidence_range` is 0 preventing divide by zero or invalid confidence computations

Fixes #14438

### DIFF
--- a/experimenter/experimenter/nimbus_ui/views.py
+++ b/experimenter/experimenter/nimbus_ui/views.py
@@ -814,7 +814,7 @@ class NewResultsView(NimbusExperimentViewMixin, DetailView):
                     lower = relative[0].get("lower") * 100
                     upper = relative[0].get("upper") * 100
                     confidence_range = round(extreme_bound * 1000) / 10
-                    full_width = confidence_range * 2
+                    full_width = confidence_range * 2 if confidence_range != 0 else 1
                     bar_width = ((upper - lower) / full_width) * 100
                     left_percent = (abs(lower - confidence_range * -1) / full_width) * 100
                     left_bounds_percent = left_percent


experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -814,7 +814,7 @@ def get_context_data(self, **kwargs):
                     lower = relative[0].get("lower") * 100
                     upper = relative[0].get("upper") * 100
                     confidence_range = round(extreme_bound * 1000) / 10
-                    full_width = confidence_range * 2
+                    full_width = confidence_range * 2 if confidence_range != 0 else 1
```
yashikakhurana, line 817: in case its not 0, it should be 1? 🤔

moibra05: yes, it just means that `full_width` won't have an effect on the size of "relative change bar" which if `confidence_range` is ever 0  it means all the points for that specific metric are 0 so either way they wont be displayed

image

yashikakhurana (approved):
I see, thank you for explaining that, thanks @moibra05

PR #14448, opened 2026-01-23
fix(nimbus): fix null window access for older experiments

### LINKED ISSUE #14447: AttributeError: 'NoneType' object has no attribute 'get'
Sentry Issue: [EXPERIMENTER-PROD-17M](https://mozilla.sentry.io/issues/7210598252/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'get'
(4 additional frame(s) were not displayed)
...
  File "experimenter/nimbus_ui/views.py", line 782, in get_context_data
    all_metrics = results_manager.get_metric_data(
  File "experimenter/jetstream/results_manager.py", line 412, in get_metric_data
    metric_data[area] = self.get_metric_area_data(
  File "experimenter/jetstream/results_manager.py", line 394, in get_metric_area_data
    self.get_window_results(analysis_basis, segment, "weekly"),
  File "experimenter/jetstream/results_manager.py", line 426, in get_window_results
    .get(analysis_basis, {})
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6389)

moibra05:
Because

- Some older experiments have their weekly results_data fields set to None
- Currently the code for the new results page does not have a way to handle this and ends up raises errors when this null field has any attempts to access its value

This commit

- Adds a proper fallback after accessing the window field in results data in case it ever returns None
- Adds tests that mimic this edge case found in older experiments

Fixes #14447

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -420,12 +420,9 @@ class ExperimentResultsManager:
 
     def get_window_results(self, analysis_basis, segment, window="overall"):
         return (
-            (
-                self.experiment.results_data.get("v3", {})
-                .get(window, {})
-                .get(analysis_basis, {})
-                .get(segment, {})
-            )
+            (self.experiment.results_data.get("v3", {}).get(window, {}) or {})
+            .get(analysis_basis, {})
+            .get(segment, {})
             if self.experiment.results_data
             else {}
         )
--- a/experimenter/experimenter/jetstream/tests/test_results_manager.py
+++ b/experimenter/experimenter/jetstream/tests/test_results_manager.py
@@ -765,3 +765,183 @@ class TestExperimentResultsManager(TestCase):
             },
             kpi_metrics,
         )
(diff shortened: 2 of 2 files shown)


experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -420,12 +420,9 @@ def get_metric_data(
...
-                .get(window, {})
-                .get(analysis_basis, {})
-                .get(segment, {})
-            )
+            (self.experiment.results_data.get("v3", {}).get(window, {}) or {})
+            .get(analysis_basis, {})
+            .get(segment, {})
             if self.experiment.results_data
             else {}
         )
```
mikewilli, line 428: Not opposed to this solution, but had an idea. You could probably proactively cover other potential cases by getting the result instead of returning directly, and then check if the thing you retrieved is `None` and return `{}` if so. Maybe log a warning if we hit a case like this so we can figure out if something is wrong and worth fixing.

PR #14453, opened 2026-01-23
chore(nimbus): use readonly json for labs fields

### LINKED ISSUE #14395: Use readonly codemirror for localization field
The localization json field on the summage page is still just a naive textbox, it should be a readonly codemirror to be consistent with the rest of the page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6366)

jaredlockhart:
Becuase

* We're displaying some JSON in the labs description links fields
* We can use the same readonly json template we use everywhere else

This commit

* Uses the readonly json template for the labs description links fields
* Pulls the shared logic for collapsing the readonly json field into the base template

fixes #14395


jaredlockhart:
image

PR #14454, opened 2026-01-23
feat(nimbus): Manage alerts model

### LINKED ISSUE #14406: Create database models for alert state tracking and duplication
Create new Django models to track alert history and prevent duplicate notifications.

**Technical Approach**:

* Add new models to `experimenter/experiments/models.py`:
  * `NimbusAlertType` enum for alert categories
  * `NimbusAlert` model to store alert state
* Follow existing patterns from `NimbusEmail` and `NimbusChangeLog` models
* Store Slack thread message IDs for threading support

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6369)

yashikakhurana:
Because

-We want to manage the alerts and wants to record the alerts sent to avoid duplication sending of the notifications

This commit

- Adds the nimbus alert model to manage the alerts

Fixes #14406

### DIFF
--- a/experimenter/experimenter/experiments/tests/test_models.py
+++ b/experimenter/experimenter/experiments/tests/test_models.py
@@ -5927,7 +5927,6 @@ class TestNimbusAlert(TestCase):
         expected = f"{experiment.slug} - {alert.alert_type} - {alert.sent_on}"
         self.assertEqual(str(alert), expected)
 
-
     def test_was_sent_recently_returns_true_for_recent_alert(self):
         experiment = NimbusExperimentFactory.create()
         NimbusAlert.objects.create(


experimenter/experimenter/experiments/models.py
```diff
@@ -2906,6 +2906,38 @@ def __str__(self):  # pragma: no cover
...
+    sent_on = models.DateTimeField(auto_now_add=True, help_text="When the alert was sent")
+    slack_thread_id = models.CharField(
+        max_length=255,
+        null=True,
+        blank=True,
+        help_text="Slack message timestamp for threading",
+    )
+
+    class Meta:
+        unique_together = [["experiment", "alert_type"]]
```
jaredlockhart: But theoretically we could send the same alert several times? Like

Day N: load data from jetstream, find some error, send alert, type: jetstream error, message: metric A broken

Day N+1: load data from jetstream, find some other error, send alert, type: jetstream error, message: metric B broken

couldn't we?

experimenter/experimenter/experiments/constants.py
```diff
@@ -822,6 +822,16 @@ class EmailType(models.TextChoices):
         EXPERIMENT_END = "experiment end"
         ENROLLMENT_END = "enrollment end"
 
+    class AlertType(models.TextChoices):
+        ANALYSIS_ERROR_DAILY = "analysis_error_daily", "Daily Analysis Error"
+        ANALYSIS_ERROR_WEEKLY = "analysis_error_weekly", "Weekly Analysis Error"
+        ANALYSIS_ERROR_OVERALL = "analysis_error_overall", "Overall Analysis Error"
+        DAILY_RESULTS_READY = "daily_results_ready", "Daily Results Ready"
+        WEEKLY_RESULTS_READY = "weekly_results_ready", "Weekly Results Ready"
+        OVERALL_RESULTS_READY = "overall_results_ready", "Overall Results Ready"
```
jaredlockhart: Wait what's the difference between analysis and results? 🤔

experimenter/experimenter/experiments/models.py
yashikakhurana: good point let me remove the unique constraint

experimenter/experimenter/experiments/constants.py
yashikakhurana: one is for ready and one is for error case

mikewilli: Maybe `daily_results_ready` and `daily_results_error` for consistency? Or whatever order for these words but consistent and just swap out the analysis period and "error" vs "ready".

yashikakhurana: good idea

experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -5913,3 +5914,101 @@ def test_targeting_formatted_with_complex_expression(self):
...
+        )
+        alert2 = NimbusAlert.objects.create(
+            experiment=experiment2,
+            alert_type=NimbusConstants.AlertType.DAILY_RESULTS_READY,
+            message="Alert for exp2",
+        )
+
+        self.assertIsNotNone(alert1)
+        self.assertIsNotNone(alert2)
+        self.assertNotEqual(alert1.id, alert2.id)
```
jaredlockhart: I think this falls under the category of 'testing django built ins' which is already covered by Django's test suite.  We only need to test things where we implement our own logic.

experimenter/experimenter/experiments/constants.py
jaredlockhart: Oh yeah I see, error and ready, good to make distinction, but yeah as @mikewilli says better to format them the same.  I like to name things like this where the shared part goes first so when you sort alphabetically it groups things semantically so maybe something like

```
ANALYSIS_ERROR_DAILY
ANALYSIS_ERROR_WEEKLY
ANALYSIS_ERROR_OVERALL
ANALYSIS_READY_DAILY
ANALYSIS_READY_WEEKLY
ANALYSIS_READY_OVERALL
```

So the shared parts are

1) Analysis - applies to all
2) Error/Ready - splits into two groups
3) Daily/Weekly/Overall - most specific

PR #14492, opened 2026-01-27
feat(nimbus): Create scheduled Celery task to monitor experiments for alert conditions such as results and analysis error

### LINKED ISSUE #14407: Create scheduled Celery task to monitor experiments for alert conditions
Set up a new Celery Beat task that runs hourly to check all LIVE and COMPLETE experiments for alert conditions. Follows existing pattern from `fetch_jetstream_data` task.

**Technical Approach**:

* Create new `experimenter/alerts/` module
* Create scheduled task that queries LIVE/COMPLETE experiments
* Spawn child tasks for each alert type check
* Add to `CELERY_BEAT_SCHEDULE` in settings
* Use existing Celery infrastructure (Redis broker)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6370)

yashikakhurana:
Because

- We need to alert users if they have some analysis error or the results are available for their experiments

This commit

- Runs a celery task each day at 8am pst/ 11am est to check the results or analysis error for the live and complete experiments

Fixes #14407 

Note: Will implement the actual logic for the results and analysis error in the next tickets, created the dummy method to fit the celery tasks in it.


experimenter/experimenter/slack/tasks.py
```diff
@@ -43,3 +44,80 @@ def nimbus_send_slack_notification(
...
+                NimbusExperiment.Status.LIVE,
+                NimbusExperiment.Status.COMPLETE,
+            ]
+        )
+
+        experiment_count = experiments.count()
+        logger.info(f"Checking {experiment_count} experiments for alerts")
+
+        for experiment in experiments:
+            check_single_experiment_alerts.delay(experiment.id)
```
jaredlockhart: This is gonna kick off a task for every ended experiment, which is like 2k and growing, but we only fetch data for 3 days after an experiment ends, so probably we wanna filter this down to just experiments that ended in the last 3 days.  Can prolly use `_computed_end_date` since that should be db queryable and populated.

experimenter/experimenter/slack/tasks.py
```diff
@@ -43,3 +44,80 @@ def nimbus_send_slack_notification(
...
+def check_single_experiment_alerts(experiment_id):
+    try:
+        experiment = NimbusExperiment.objects.get(id=experiment_id)
+        logger.debug(f"Checking alerts for experiment: {experiment.slug}")
+
+        # Check for analysis errors (daily, weekly, overall)
+        check_analysis_errors.delay(experiment_id)
+
+        # Check if results became ready (daily, weekly, overall)
+        check_analysis_ready.delay(experiment_id)
```
jaredlockhart: Why do we need to dispatch further into individual tasks from here?

experimenter/experimenter/settings.py
```diff
@@ -420,6 +420,10 @@
         "task": "experimenter.kinto.tasks.nimbus_sync_published_dto",
         "schedule": crontab(minute=0, hour="*"),
     },
+    "check_experiment_alerts": {
+        "task": "experimenter.slack.tasks.check_experiment_alerts",
+        "schedule": crontab(minute=0, hour=16),
```
jaredlockhart: Why once a day at 4? I guess that's the once a day after jetstream finishes running?  I think jetstream runs at midnight so maybe doing this earlier in the day so people get the alerts at the start of their day would be helpful?  cc @mikewilli ?

experimenter/experimenter/slack/tasks.py
yashikakhurana: to send different notifications, but this can be combined

experimenter/experimenter/settings.py
yashikakhurana: right now this does this- Runs a celery task each day at 8am pst/ 11am est

yashikakhurana: I can change time too

experimenter/experimenter/slack/tasks.py
yashikakhurana: good point

experimenter/experimenter/slack/tasks.py
```diff
@@ -43,3 +46,59 @@ def nimbus_send_slack_notification(
...
         raise e
+
+
+@shared_task
+@metrics.timer_decorator("check_experiment_alerts")
+def check_experiment_alerts():
+    metrics.incr("check_experiment_alerts.started")
+
+    try:
+        # Get the cutoff date for COMPLETE experiments (3 days ago)
```
jaredlockhart, line 57: Perfect ty 🙏

PR #14499, opened 2026-01-28
feat(nimbus): add results indicator next to metric slug in metric area headers

### LINKED ISSUE #14460: Add metric result indicators next to subtitles
* Add coloured arrows next to each metric to indicate
  * Up (green)
    * All neutral or green
  * Neutral (grey)
    * All neutral
  * Down (red)
    * All neutral or red
  * Mixed (yellow question?)
    * Green or red?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6398)

moibra05:
Because

- Upon first visiting the new results page it is unclear the results of the experiment and whether there were any significant changes

This commit

- Adds an arrow indicator next to the metric names to indicate how that experiment affected that particular metric
- Fixed an issue where metrics would incorrectly be placed in the "Notable Changes" section

Fixes #14460

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from enum import StrEnum
 from itertools import chain, zip_longest
 
 from experimenter.experiments.constants import (
@@ -7,6 +8,13 @@ from experimenter.experiments.constants import (
 from experimenter.metrics import MetricAreas
 from experimenter.nimbus_ui.constants import NimbusUIConstants
 from experimenter.outcomes import Outcomes
+
+
+class MetricSignificance(StrEnum):
+    POSITIVE = "positive"
+    NEGATIVE = "negative"
+    MIXED = "mixed"
+    NEUTRAL = "neutral"
 
 
 class ExperimentResultsManager:
@@ -127,7 +135,7 @@ class ExperimentResultsManager:
         for i, data_point in enumerate(absolute_data_list):
             lower = data_point.get("lower")
             upper = data_point.get("upper")
-            significance = significance_map.get(str(i + 1), "neutral")
+            significance = significance_map.get(str(i + 1), MetricSignificance.NEUTRAL)
             abs_entries.append(
                 {"lower": lower, "upper": upper, "significance": significance}
             )
@@ -148,7 +156,7 @@ class ExperimentResultsManager:
             avg_rel_change = (
                 abs(data_point.get("point")) if data_point.get("point") else None
             )
-            significance = significance_map.get(str(i + 1), "neutral")
+            significance = significance_map.get(str(i + 1), MetricSignificance.NEUTRAL)
             rel_entries.append(
                 {
                     "lower": lower,
@@ -236,10 +244,18 @@ class ExperimentResultsManager:
             if self.metric_has_errors(kpi["slug"], analysis_basis, segment):
                 kpi["has_errors"] = True
 
+            kpi["overall_change"] = self.get_overall_change(
+                kpi["group"],
+                kpi["slug"],
+                analysis_basis,
+                segment,
+                reference_branch,
+            )
+
         return kpi_metrics
 
     def get_remaining_metrics_metadata(
-        self, exclude_slugs=None, analysis_basis=None, segment=None
+        self, exclude_slugs=None, analysis_basis=None, segment=None, reference_branch=None
     ):
         analysis_data = (
             self.experiment.results_data.get("v3", {})
@@ -265,6 +281,13 @@ class ExperimentResultsManager:
                         "friendly_name": metric_friendly_name,
                         "has_errors": self.metric_has_errors(
                             slug, analysis_basis, segment
+                        ),
+                        "overall_change": self.get_overall_change(
+                            "other_metrics",
+                            slug,
+                            analysis_basis,
+                            segment,
+                            reference_branch,
                         ),
                     }
                 )
@@ -317,6 +340,13 @@ class ExperimentResultsManager:
                     "has_errors": self.metric_has_errors(
                         metric.slug, analysis_basis, segment
                     ),
+                    "overall_change": self.get_overall_change(
+                        "other_metrics",
+                        metric.slug,
+                        analysis_basis,
+                        segment,
+                        reference_branch,
+                    ),
                 }
                 if formatted_metric not in outcome_metrics:
                     outcome_metrics.append(formatted_metric)
@@ -326,8 +356,12 @@ class ExperimentResultsManager:
             metric_areas[outcome.friendly_name if outcome else slug] = outcome_metrics
 
         remaining_metrics = self.get_remaining_metrics_metadata(
-            exclude_slugs=all_outcome_metric_slugs
+            exclude_slugs=all_outcome_metric_slugs,
+            analysis_basis=analysis_basis,
+            segment=segment,
+            reference_branch=reference_branch,
         )
+
         grouped_metrics = []
         for metric in remaining_metrics:
             area = MetricAreas.get(self.experiment.application, metric["slug"])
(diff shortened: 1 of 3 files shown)


moibra05:
image
image

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -427,3 +459,36 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+                continue
+
+            metric_significance = (
+                branch_data.get("branch_data", {})
+                .get(group, {})
+                .get(metric_slug, {})
+                .get("significance", {})
+                .get(reference_branch, {})
+                .get("overall", {})
+                .get("1", "neutral")
```
mikewilli: I don't know if the default should be `neutral`. I think it'd be better to show nothing here if we can't determine the significance (or a ❓ icon maybe if you want something for everything? or error icon if there's an error for this metric).

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -427,3 +459,36 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+            if metric_significance == "positive":
+                if overall_change == "negative":
+                    overall_change = "mixed"
+                    break
+                overall_change = "positive"
+            elif metric_significance == "negative":
+                if overall_change == "positive":
+                    overall_change = "mixed"
+                    break
+                overall_change = "negative"
```
mikewilli: Are we intending to treat a mix of `neutral` and `positive`/`negative` as only `positive`/`negative`? That seems like mixed results to me too, no? Though maybe that makes this icon less helpful since it'd probably be `mixed` almost all the time.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new-inner.html
```diff
@@ -219,7 +219,10 @@ <h4 class="mb-0">{{ area }}</h4>
                 {% else %}
                   {% for metric in metric_metadata %}
                     <div>
-                      <small class="text-muted me-2">{{ metric.friendly_name }}</small>
+                      <small class="text-muted">{{ metric.friendly_name }}</small>
+                      <small>
+                        <i class="fa-solid ms-1 me-2 {% if metric.overall_change == "positive" %}fa-up-long text-success{% elif metric.overall_change == "negative" %}fa-down-long text-danger{% elif metric.overall_change == "mixed" %}fa-right-left text-warning{% else %}fa-left-long text-muted{% endif %}"></i>
+                      </small>
```
mikewilli: Forgot one other thing -- these icons should have some explanatory hover text as well.

jaredlockhart:
Oh yeah I think I suggested four possible states

1) All neutral
2) Neutral or positive
3) Neutral or negative 
4) Mixed (neutral, positive, negative)

But you're right @mikewilli there's also a case for 

5) Missing data

I think that should cover all the states.

Yeah theoretically we could fold any positive/negative together into a 'notable results' I just think it would be great to see all the directionality in one zoomed out view before you zoom in, which we don't have right now.  If we find it's confusing or misleading we can always change it later.

moibra05:
> Oh yeah I think I suggested four possible states
> 
> 
> 
> 1) All neutral
> 
> 2) Neutral or positive
> 
> 3) Neutral or negative 
> 
> 4) Mixed (neutral, positive, negative)
> 
> 
> 
> But you're right @mikewilli there's also a case for 
> 
> 
> 
> 5) Missing data
> 
> 
> 
> I think that should cover all the states.
> 
> 
> 
> Yeah theoretically we could fold any positive/negative together into a 'notable results' I just think it would be great to see all the directionality in one zoomed out view before you zoom in, which we don't have right now.  If we find it's confusing or misleading we can always change it later.

Yep, I planned to address better ui for metrics that are missing data in a different pr

experimenter/experimenter/jetstream/results_manager.py
jaredlockhart: Oh let's move all these string constants into an enum and refer to them by reference 🙏

jaredlockhart:
Oh the double arrows means mixed!  That's clever, but actually I'd use up/down together since we're using up and down already for each of positive and negative.

experimenter/experimenter/jetstream/results_manager.py
moibra05: i originally thought the same but i figured the arrows wouldn't be all that useful if they showed mixed results 90% of the time

experimenter/experimenter/jetstream/results_manager.py
moibra05: will fix this in this issue #14501

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -9,6 +10,13 @@
 from experimenter.outcomes import Outcomes
 
 
+class MetricSignificance(StrEnum):
```
jaredlockhart, line 13: Perfect 🙏

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -427,3 +467,36 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+            if metric_significance == MetricSignificance.POSITIVE:
+                if overall_change == MetricSignificance.NEGATIVE:
+                    overall_change = MetricSignificance.MIXED
+                    break
+                overall_change = MetricSignificance.POSITIVE
+            elif metric_significance == MetricSignificance.NEGATIVE:
+                if overall_change == MetricSignificance.POSITIVE:
+                    overall_change = MetricSignificance.MIXED
+                    break
+                overall_change = MetricSignificance.NEGATIVE
```
jaredlockhart: This is okay, but another way to do this is using the magic of `set`!

Something like

```py
significances = {metric_significance, overall_change}

if significances == {POSITIVE, NEGATIVE}:
    overall = MIXED
elif significances.intersection({POSITIVE}):
    overall = POSITIVE
....etcetcetc
```    

Play around with it, it's useful when you want to compare things where you don't care about the order, and when you just want the unique list of things,  it lets you express things more succinctly.

moibra05: oo thats smart

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -427,3 +467,37 @@ def get_window_results(self, analysis_basis, segment, window="overall"):
...
+                .get(metric_slug, {})
+                .get("significance", {})
+                .get(reference_branch, {})
+                .get("overall", {})
+                .get("1", MetricSignificance.NEUTRAL)
+            )
+
+            significances = {metric_significance, overall_change}
+
+            if significances == {
```
jaredlockhart, line 493: Nice 🙏

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-new-inner.html
```diff
@@ -342,3 +348,17 @@ <h4 class="modal-title" id="edit-summary-{{ experiment.slug }}-label">Edit outco
...
+            var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
+            tooltipTriggerList.forEach(function (el) {
+              // eslint-disable-next-line no-undef
+              new bootstrap.Tooltip(el);
+            });
+          } catch (e) {
+            // If bootstrap isn't available or initialization fails, silently ignore
+          }
+        })();
+</script>
```
jaredlockhart: Oh I see, yeah this kind of thing should go in the JS file for the results page.  We should already have one shouldn't we?

moibra05: oops, that was some old code i had while i was trying to test something with the tooltips and didnt want to make a new file for it

PR #14509, opened 2026-01-29
feat(nimbus): add loading state when changing filters on new results page

### LINKED ISSUE #14462: Add spinner state after changing filter options
After a one of the filter option are change there is a slight delay as htmx needs to make a request to update the displayed information. To improve user experience we should add a spinner while the displayed data is being updated

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6403)

moibra05:
Because

- Changing filters from the new results page ui makes htmx network requests each time that have small delays before being fulfilled
-  There isn't any feedback from the ui letting the user know whether or not the request they made is being processed or not

This commit

- Adds a loading overlay in between filter changes

Fixes #14462


jaredlockhart (approved):
Oh sweet cool yeah just reusing the existing loading overlay pattern, perfect, tested locally works great thnx @moibra05 🙏

PR #14549, opened 2026-02-04
feat(nimbus): disable slack notification toggle for complete experiments

### LINKED ISSUE #14548: Grey out slack notifications toggle for complete experiments
Once an experiment is complete, there won't be any more slack review notifications, so we should disable that toggle.  We could hide it entirely, but we could also try to generally adopt a pattern of keeping things in place and disabling them rather than hiding them entirely so ui elements don't keep randomly appearing/disappearing.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6426)

jaredlockhart:
Becuase

* Once an experiment completes there will be no more slack review notifications
* We should disable the toggle for it once an experiment completes
* We could hide it entirely but perhaps it's a better UI pattern to leave it in place and disable it with a tooltip explaining why

This commit

* Disables the slack review notification toggle for complete experiments
* Adds an explanatory tooltip

fixes #14548

### DIFF
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/experiment_base.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/experiment_base.html
@@ -45,10 +45,12 @@
             class="d-flex align-items-center">
         {% csrf_token %}
         <input type="hidden" name="enable_review_slack_notifications" value="false">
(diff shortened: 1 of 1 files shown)


jaredlockhart:
image

PR #14559, opened 2026-02-05
feat(nimbus): improve ui for metrics with missing data

### LINKED ISSUE #14501: Metrics with missing data are not clear
It is unclear when identifying metrics with missing data as their values are currently using fallbacks for neutral data. This makes it difficult to point out metrics that may have issues.

Improve the ui to make a clear distinction between metrics that are missing data and others with neutral data.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6411)

moibra05:
Because

- It is unclear when a metric is missing data

This commit

- Adds error text when a metric is missing data
- Updates metric tooltip in header to show error icon if the metric is unavailable
- Disable the popout view for unavailable metrics

Fixes #14501

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -204,6 +204,43 @@ class ExperimentResultsManager:
                     return True
         return False
 
+    def metric_has_data(
+        self, metric_slug, group, analysis_basis, segment, reference_branch=None
+    ):
+        if self.experiment.results_data:
+            window_results = (
+                self.get_window_results(analysis_basis, segment, "overall")
+                or self.get_window_results(analysis_basis, segment, "weekly")
+                or self.get_window_results(analysis_basis, segment, "daily")
+            )
+
+            for branch in self.experiment.get_sorted_branches():
+                if branch.slug == reference_branch:
+                    continue
+
+                metric_data = (
+                    window_results.get(branch.slug, {})
+                    .get("branch_data", {})
+                    .get(group, {})
+                    .get(metric_slug, {})
+                )
+
+                abs_point_data = metric_data.get("absolute", {}).get("first", {})
+                rel_point_data = (
+                    metric_data.get("relative_uplift", {})
+                    .get(reference_branch, {})
+                    .get("first", {})
+                )
+
+                def check_valid_point(point):
+                    return any(value != 0 for value in point.values())
+
+                if (not abs_point_data or not check_valid_point(abs_point_data)) and (
+                    not rel_point_data or not check_valid_point(rel_point_data)
+                ):
+                    return False
+        return True
+
     def get_kpi_metrics(
         self, analysis_basis, segment, reference_branch, window="overall"
     ):
@@ -252,6 +289,10 @@ class ExperimentResultsManager:
                 reference_branch,
             )
 
+            kpi["has_data"] = self.metric_has_data(
+                kpi["slug"], kpi["group"], analysis_basis, segment, reference_branch
+            )
+
         return kpi_metrics
 
     def get_remaining_metrics_metadata(
@@ -288,6 +329,9 @@ class ExperimentResultsManager:
                             analysis_basis,
                             segment,
                             reference_branch,
+                        ),
+                        "has_data": self.metric_has_data(
+                            slug, group, analysis_basis, segment
                         ),
                     }
                 )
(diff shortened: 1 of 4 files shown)


moibra05:
image

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-inner.html
```diff
@@ -253,12 +253,15 @@ <h4 class="mb-0">{{ area }}</h4>
...
                             type="button"
                             data-bs-toggle="modal"
                             data-bs-target="#{{ experiment.slug }}-{{ metric.slug }}-{{ area|slugify }}"
-                            style="height: 100px">
-                      <i class="fa-solid fa-arrow-up-right-from-square text-secondary fa-sm position-absolute top-0 end-0 p-3 px-2"></i>
+                            style="height: 100px;
+                                   {% if not metric.has_data %}cursor: default;{% endif %}">
+                      {% if metric.has_data %}
+                        <i class="fa-solid fa-arrow-up-right-from-square text-secondary fa-sm position-absolute top-0 end-0 p-3 px-2"></i>
+                      {% endif %}
```
mikewilli, line 264: Might be nice to have an `else` condition here and show the warning icon in place of the click/link icon if the metric doesn't have results.

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -204,6 +204,40 @@ def metric_has_errors(self, metric_slug, analysis_basis, segment):
...
+                )
+
+                point_data = metric_data.get("absolute", {}).get("first", {})
+
+                def check_valid_point(point):
+                    return any(value != 0 for value in point.values())
+
+                # DAU is the only metric that does not include absolute counts; for this
+                # metric, validate the presence of relative uplift data instead.
+                if metric_slug == "client_level_daily_active_users_v2":
```
mikewilli: Good comment, but would this work for every metric? It's possible that we change this decision and add absolute data for DAU, or that we change the DAU metric name (v3 for example), and so it'd be nice to avoid a special case and do the same thing for every metric.

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -204,6 +204,40 @@ def metric_has_errors(self, metric_slug, analysis_basis, segment):
...
+    def metric_has_data(
+        self, metric_slug, group, analysis_basis, segment, reference_branch=None
+    ):
+        if self.experiment.results_data:
+            for branch in self.experiment.get_sorted_branches():
+                if branch.slug == reference_branch:
+                    continue
+
+                metric_data = (
+                    self.get_window_results(analysis_basis, segment)
```
mikewilli: The default behavior (which we are using here) is to get `overall` results. What happens if there are `weekly` or `daily` results though?

mikewilli:
Oh one other general comment about something I just noticed in the screenshot: do we want to say "other metrics are unaffected"? That might get annoying if it shows up for every metric... Perhaps "other metrics may not be affected" or even just leave it out entirely.

experimenter/experimenter/jetstream/results_manager.py
moibra05: can an experiment have weekly and/or daily results without overall results? if so, can the data be considered valid in that case?

mikewilli: Yep, daily are computed each day, and weekly each week during the observation period, but overall isn't computed until it ends. That's expected and valid behavior.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-inner.html
moibra05: im not the biggest fan of how this looks. what do you think?
image

jaredlockhart: Oh we can just remove the little modal icon and grey out the metric card?

moibra05: its already slightly greyed out but i can make it more grey if you think that looks better?

image

mikewilli: Yea agreed that's a lot of icons and redundant. I was thinking a gray icon like the modal icon, but even that is probably going to be too much. Thanks for trying it out though!

PR #14569, opened 2026-02-06
feat(nimbus): add back ingestion for daily metrics in experimenter

### LINKED ISSUE #14568: Add ingestion for daily metrics in experimenter code base
* Update the task that periodically pulls data into the experimenter database. Includes updating relevant tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6378)

moibra05:
Because

- Daily metrics are returning for certain guardrail metrics
- They're already being exported by Jetstream but need to be ingested into the results data objects

This commit

- Adds the daily results data back into the results data object

Fixes #14568

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -197,9 +197,9 @@ def get_other_metrics_names_and_map(
 
 def get_experiment_data(experiment: NimbusExperiment):
     recipe_slug = experiment.slug.replace("-", "_")
-    # we don't use DAILY results in Experimenter, so only get WEEKLY/OVERALL
-    windows = [AnalysisWindow.WEEKLY, AnalysisWindow.OVERALL]
+    windows = [AnalysisWindow.DAILY, AnalysisWindow.WEEKLY, AnalysisWindow.OVERALL]
     raw_data = {
+        AnalysisWindow.DAILY: {},
         AnalysisWindow.WEEKLY: {},
         AnalysisWindow.OVERALL: {},
     }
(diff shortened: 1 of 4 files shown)


moibra05:
> This will restore what we had before, but I'm now remembering that what we had for daily results was just a copy of the incoming data format without any transformation. Is that what we want here, or do we want the daily results to be in the same format as the weekly/overall in the `results_data` object?

there were tentative plans to eventually reformat a sizeable chunk of the results data object to eliminate some of the workload placed on the `results_manager`. i think this bit of work might make the most sense to be done alongside the rest of the reformatting plans in a separate pr?

mikewilli:
I think that's a great idea -- the existing `results_data` object was built for the old UI after all, so why not reconsider that structure for the new UI -- but in the meantime, is there any value in having the daily results in this object in their unmodified form? The `results_manager` is built for the weekly/overall structure, and so wouldn't be able to do anything with the daily results. Are you planning to use the daily results in this format in some way?

moibra05:
i originally thought that all the handling of the daily metric data would be isolated from the rest of the existing code, so there’d be no real benefit in refactoring the structure right now. thinking about it more, that’s not really true b/c this pr (#14559) operates on the assumption that all daily/weekly/overall data uses the same structure, so i'll go ahead and make that update

experimenter/experimenter/jetstream/client.py
```diff
@@ -280,6 +280,12 @@ def get_experiment_data(experiment: NimbusExperiment):
                 # Create the output object (weekly data)
                 ResultsObjectModel = create_results_object_model(data)
 
+                data = ResultsObjectModel(result_metrics, data, experiment, window)
+            elif data and window == AnalysisWindow.DAILY:
```
mikewilli: This could be a combined `elif WEEKLY or DAILY` like you have for exposures, no?

PR #14591, opened 2026-02-09
chore(nimbus): Add GHA workflow for Experimenter lint and tests

### LINKED ISSUE #14578: Migrate Experimenter Check to GHA (x86_64 + aarch64 + reporting)
Migrate the check_experimenter_x86_64, check_experimenter_aarch64, and check_experimenter_and_report CircleCI jobs to GitHub Actions. These run 'make check' (lint + unit tests) on both architectures, with the reporting variant uploading JUnit XML and coverage JSON to GCS. The reporting job only runs on main/update branches.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6434)

jaredlockhart:
Because

* We are migrating CI from CircleCI to GitHub Actions (EXP-6320)
* The Experimenter check jobs (`make check`) need to run on PRs and pushes
* Docker layer caching is needed to keep build times reasonable

This commit

* Adds a GitHub Actions workflow (`.github/workflows/check-experimenter.yml`) that runs `make check` on both x86_64 and aarch64 architectures
* Adds a reusable composite action (`.github/actions/setup-cached-build/action.yml`) that handles GCP workload identity auth, Docker Buildx setup, and registry-backed layer caching via Google Artifact Registry
* Adds `MEGAZORD_BUILD_FLAGS` and `EXPERIMENTER_BUILD_FLAGS` override points to the Makefile so CI can inject cache and build-context flags without duplicating build commands
* Triggers on changes to `experimenter/`, `application-services/`, and the workflow file itself
* Runs on PRs, pushes to main/update branches, and merge queue

fixes #14578

### DIFF
--- a/.github/workflows/check-experimenter.yml
+++ b/.github/workflows/check-experimenter.yml
@@ -1,0 +1,47 @@
+name: Experimenter Tests
+
+on:
+  push:
+    branches:
+      - main
+      - update_firefox_versions
+      - update-application-services
+    paths:
+      - "experimenter/**"
+      - "application-services/**"
+      - ".github/workflows/check-experimenter.yml"
+  pull_request:
+    paths:
+      - "experimenter/**"
+      - "application-services/**"
+      - ".github/workflows/check-experimenter.yml"
+  merge_group:
+    types: [checks_requested]
+
+jobs:
+  check:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-24.04
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+    name: "${{ matrix.arch }}"
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: ./.github/actions/setup-cached-build
+        with:
+          arch: ${{ matrix.arch }}
+          gcp-project-number: ${{ vars.GCPV2_WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER }}
+
+      - name: Run tests and linting
+        run: |
+          cp .env.sample .env
+          make check
--- a/.github/actions/setup-cached-build/action.yml
+++ b/.github/actions/setup-cached-build/action.yml
@@ -1,0 +1,52 @@
+name: "Setup Cached Docker Build"
+description: "Authenticate to GAR, set up Buildx, and build megazords with registry cache"
+
+inputs:
+  gar-cache:
+    description: "GAR cache path prefix"
+    default: "us-docker.pkg.dev/moz-fx-experimenter-prod-6cd5/experimenter-prod/cache"
+  arch:
+    description: "Architecture tag for cache (e.g. x86_64, aarch64)"
+    default: "x86_64"
+  gcp-project-number:
+    description: "GCP project number for workload identity"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Authenticate to Google Cloud
+      uses: google-github-actions/auth@v2
+      with:
+        workload_identity_provider: projects/${{ inputs.gcp-project-number }}/locations/global/workloadIdentityPools/github-actions/providers/github-actions
+        service_account: artifact-writer@moz-fx-experimenter-prod-6cd5.iam.gserviceaccount.com
+
+    - name: Login to Artifact Registry
+      shell: bash
+      run: gcloud auth configure-docker us-docker.pkg.dev --quiet
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Build megazords with cache
+      shell: bash
+      env:
+        MEGAZORD_BUILD_FLAGS: >-
+          --cache-from type=registry,ref=${{ inputs.gar-cache }}/megazords-cache:${{ inputs.arch }}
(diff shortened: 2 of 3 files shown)


.github/workflows/check-experimenter.yml
```diff
@@ -0,0 +1,49 @@
...
+  merge_group:
+
+jobs:
+  check:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-22.04
```
mikewilli: Why not `ubuntu-latest` (or `ubuntu-24.04` if you want to pin a version)?

.github/workflows/check-experimenter.yml
```diff
@@ -0,0 +1,49 @@
...
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
```
mikewilli, line 30: I don't see this option [here](https://github.com/actions/runner-images?tab=readme-ov-file#available-images), is it actually available? Do you want to use one of the macos images listed there instead?

mikewilli, line 30: I think you could write this as
```suggestion
      matrix:
        arch: [x86_64, aarch64]
        runner: [ubuntu-22.04, ubuntu-24.04-arm]
```
if you want to be more concise.

.github/workflows/check-experimenter.yml
```diff
@@ -0,0 +1,49 @@
...
+    paths:
+      - "experimenter/**"
+      - "application-services/**"
+      - ".github/workflows/check-experimenter.yml"
+  pull_request:
+    paths:
+      - "experimenter/**"
+      - "application-services/**"
+      - ".github/workflows/check-experimenter.yml"
+  merge_group:
```
mikewilli, line 18: It's [suggested](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#merge_group) to include the `types` in case GH adds new ones in the future that you may not want to trigger on (`checks_requested` is the only one right now though).
```suggestion
  merge_group:
    types: [checks_requested]
```

.github/workflows/check-experimenter.yml
```diff
@@ -0,0 +1,49 @@
...
+  check:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-22.04
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+    name: "Check Experimenter (${{ matrix.arch }})"
```
mikewilli: Minor suggestion based on prior experience: you might consider naming this without "Experimenter" since it's in the workflow name as well. I've noticed that with how I've named things for GHA so far, I end up with a lot of redundant info because GHA concatenates names like `Workflow Name / Job ID / Job Name` or something similar (not sure whether Job ID is always included). So in this case I think you'd end up with `Experimenter Test / check / Check Experimenter (x86_64)` whereas it's probably sufficient to have `Experimenter Test / Check / x86_64` or similar.

.github/workflows/check-experimenter.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
```
mikewilli: `v6` is the latest version here:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/check-experimenter.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: projects/${{ vars.GCPV2_WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER }}/locations/global/workloadIdentityPools/github-actions/providers/github-actions
+          service_account: artifact-writer@moz-fx-experimenter-prod-6cd5.iam.gserviceaccount.com
+
+      - name: Login to Artifact Registry
+        run: gcloud auth configure-docker ${{ env.GAR_LOCATION }} --quiet
```
mikewilli: There's a moz action `mozilla-it/deploy-actions/docker-push` to auth and deploy docker to GAR. You can see it being used in [jetstream](https://github.com/mozilla/jetstream/blob/main/.github/workflows/build-and-deploy.yml#L112-L119) as an example.

.github/workflows/check-experimenter.yml
relud: wouldn't that produce 4 jobs with 2 of them invalid?

mikewilli: Oops yep you're right. Ignore me!

jaredlockhart: Listed [here](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

.github/workflows/check-experimenter.yml
jaredlockhart: Ya pinned version better, don't want it to bump and break spontaneously on us

.github/workflows/check-experimenter.yml
jaredlockhart: Oh yeah we'll need that when we setup teh deploy tasks, so far we're just using GAR to do the caching which i don't think that action supports

.github/workflows/check-experimenter.yml
mikewilli: Oh weird, guess their docs are out of sync. Oh well, as long as it works!

PR #14616, opened 2026-02-10
feat(nimbus): Send results available notification

### LINKED ISSUE #14409: Alert when daily/weekly/overall Jetstream results become available
Detect when Jetstream results transition from empty to available and send one-time Slack alerts for daily, weekly, and overall results.

**Technical Approach**:

* Implement `check_results_ready(experiment_id)` task
* Check `experiment.results_data['v3']['daily']`, `['weekly']`, `['overall']`
* Verify actual data exists (not just empty containers)
* Send alert only once per result type using `NimbusAlert` deduplication
* Use existing Slack notification system

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6372)

yashikakhurana:
Because

- We want to notify users on slack when the weekly and overall results are available for the experiments

This commit

- Check whether the results are available, and if available, send the notification
- Also makes sure to send the alert once per window such as overall and weekly

Fixes #14409

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -813,6 +813,10 @@ class NimbusConstants:
         EXPERIMENT_LAUNCHED = "experiment_launched", "Experiment Launched"
         ENROLLMENT_HEALTHY = "enrollment_healthy", "Enrollment Healthy"
 
+    class AnalysisWindow(models.TextChoices):
+        WEEKLY = "weekly", "Weekly"
+        OVERALL = "overall", "Overall"
+
     class FirefoxLabsGroups(models.TextChoices):
         CUSTOMIZE_BROWSING = "experimental-features-group-customize-browsing"
         WEBPAGE_DISPLAY = "experimental-features-group-webpage-display"
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1703,6 +1703,21 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
         return False
 
+    def has_results_for_window(self, window):
+        if not self.results_data or "v3" not in self.results_data:
+            return False
+
+        window_data = self.results_data["v3"].get(window)
+        if not window_data:
+            return False
+
+        for base in ["enrollments", "exposures"]:
+            base_results = window_data.get(base, {}).get("all")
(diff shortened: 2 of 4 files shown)


experimenter/experimenter/experiments/constants.py
```diff
@@ -813,6 +813,10 @@ class AlertType(models.TextChoices):
         EXPERIMENT_LAUNCHED = "experiment_launched", "Experiment Launched"
         ENROLLMENT_HEALTHY = "enrollment_healthy", "Enrollment Healthy"
 
+    class AnalysisWindow(models.TextChoices):
+        WEEKLY = "weekly", "Weekly"
+        OVERALL = "overall", "Overall"
```
mikewilli, line 818: Could we generate this from the existing enum definition (either in [schemas](https://github.com/mozilla/experimenter/blob/12cd177ef95e2f849db95121e4f73fb47cee90ea/schemas/mozilla_nimbus_schemas/jetstream/misc.py#L24) or [jetstream client](https://github.com/mozilla/experimenter/blob/12cd177ef95e2f849db95121e4f73fb47cee90ea/experimenter/experimenter/jetstream/models.py#L11)?

experimenter/experimenter/experiments/models.py
```diff
@@ -1703,6 +1703,21 @@ def has_displayable_results(self):
 
         return False
 
+    def has_results_for_window(self, window):
```
mikewilli, line 1706: I'm getting a little confused about where different results functions like this should go now that some of them are here and others are in the results manager class. For example, here we have `has_results_for_window` and [there](https://github.com/mozilla/experimenter/blob/12cd177ef95e2f849db95121e4f73fb47cee90ea/experimenter/experimenter/jetstream/results_manager.py#L462) we have `get_window_results`.

This isn't a comment about this particular function, just a question more generally: is the convention that we want all the actual results fetchers and mutators to go in results_manager and all the higher level boolean-type logic to go here?

jaredlockhart: Yeah so, I think once we have the new results page launched I want us to go back through the infrastructure we have now for

1) Jetstream loading
2) Transforming into the visualization structure
3) Storing
4) Querying

Because we have many layers there that are probably unnecessary and can collapse, but I don't think we have the structure we'll want there in place yet, so yes littering methods like this around probably isn't ideal but I think we'll be able to clean all that up in a post launch cleanup phase.

PR #14617, opened 2026-02-10
feat(nimbus): Add advanced targeting for newtab trainhop 149

nathan-barrett:
This commit will allow us to create experiments targeting users of the latest New Tab trainhop that will be released (2026-02-09), users that have accepted v4 or higher of the terms of use and are on mac or windows.

Fixes mozilla#14614


jaredlockhart:
Our formatter is busted, fixing in https://github.com/mozilla/experimenter/pull/14630 then we can rebase and fix this 🙏

PR #14650, opened 2026-02-12
chore(nimbus): delete old results page

### LINKED ISSUE #14648: Delete Old Results Page
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6422)

moibra05:
Because

- The old results page and its references can be safely deleted after the new results page is cutover to

This commit

- Deletes all things related to the old results page

Fixes #14648

### DIFF
--- a/experimenter/experimenter/nimbus_ui/static/package.json
+++ b/experimenter/experimenter/nimbus_ui/static/package.json
@@ -10,7 +10,7 @@
     "watch": "webpack --watch"
   },
   "dependencies": {
-    "@cfworker/json-schema": "^4.1.1",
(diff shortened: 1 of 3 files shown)


mikewilli:
Can I request that we leave this up for a while (a few weeks maybe?) alongside the new version? It'd be nice to have in case we missed anything, and possibly as a point of comparison while we work on follow-up tasks.

mikewilli:
Alright discussed with Jared and it sounds like the existence of this page is blocking some other work, so feel free to ignore my request and go with whatever schedule is needed.

PR #14653, opened 2026-02-12
chore(cirrus): Add GHA workflow for Cirrus lint and tests

### LINKED ISSUE #14579: Migrate Cirrus Check to GHA (x86_64 + aarch64)
Migrate the check_cirrus_x86_64 and check_cirrus_aarch64 CircleCI jobs to GitHub Actions. These run 'make cirrus_check_and_report' on both architectures, uploading JUnit and coverage results to GCS.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6435)

jaredlockhart:
Because

* We are migrating CI from CircleCI to GitHub Actions (EXP-6320)
* The Cirrus check jobs need to run on PRs and pushes to main

This commit

* Adds a GitHub Actions workflow that runs `make cirrus_check` on both x86_64 and aarch64 architectures
* Converts `cirrus_build_test` from `docker compose build` to `docker buildx build` for consistent caching with `CIRRUS_BUILD_FLAGS`
* Adds cirrus pre-build with registry cache to the `setup-cached-build` composite action
* Adds `image: cirrus:dev` to `docker-compose-test.yml` so compose uses the pre-built image
* Triggers on changes to `cirrus/`, `application-services/`, and the workflow file itself

Fixes #14579

### DIFF
--- a/.github/workflows/check-cirrus.yml
+++ b/.github/workflows/check-cirrus.yml
@@ -1,0 +1,45 @@
+name: Cirrus Check
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "cirrus/**"
+      - "application-services/**"
+      - ".github/workflows/check-cirrus.yml"
+  pull_request:
+    paths:
+      - "cirrus/**"
+      - "application-services/**"
+      - ".github/workflows/check-cirrus.yml"
+  merge_group:
+    types: [checks_requested]
+
+jobs:
+  check:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-24.04
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+    name: "${{ matrix.arch }}"
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: ./.github/actions/setup-cached-build
+        with:
+          arch: ${{ matrix.arch }}
+          gcp-project-number: ${{ vars.GCPV2_WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER }}
+
+      - name: Run Cirrus tests and linting
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/check-cirrus.yml
```diff
@@ -0,0 +1,45 @@
+name: Cirrus Tests
```
yashikakhurana: should we name it cirrus check? 🤔

Makefile
```diff
@@ -286,7 +287,7 @@ cirrus_build_dev: build_megazords
 	$(CIRRUS_ENABLE) $(DOCKER_BUILD) --target dev -f cirrus/server/Dockerfile -t cirrus:dev --build-context=fml=experimenter/experimenter/features/manifests/ cirrus/server/
 
 cirrus_build_test: build_megazords
-	$(CIRRUS_ENABLE) $(COMPOSE_TEST) build cirrus
+	$(CIRRUS_ENABLE) $(DOCKER_BUILD) $(CIRRUS_BUILD_FLAGS) --target dev -f cirrus/server/Dockerfile -t cirrus:dev --build-context fml=experimenter/experimenter/features/manifests/ cirrus/server/
```
yashikakhurana: 🎉

.github/workflows/check-cirrus.yml
relud: either way i think we should match the filename

docker-compose-test.yml
relud: these lines should be removed if we aren't using `docker compose build` anymore

docker-compose-test.yml
```diff
@@ -20,6 +20,7 @@ services:
       POSTGRES_PASSWORD: postgres
 
   cirrus:
+    image: cirrus:dev
```
relud: Before this change, the target was the deploy image, which is meaningfully different from the dev image, and I don't think we should change that:
```suggestion
    image: cirrus:deploy
```

and then we should remove `cirrus_build_test` from the Makefile

PR #14657, opened 2026-02-12
fix(nimbus): fix experiments with only weekly data not displaying results

### LINKED ISSUE #14654: Exposure defaults with no data
I'm noticing a lot of results pages defaulting to exposures but then showing no data under branches

Image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6480)

moibra05:
Because

- Many of the functions used to parse the results data were defaulting to "overall" as no value for the analysis window was being provided
- This meant that in-progress experiments that were missing overall data but still contained relevant weekly data were not having the correct data displayed

This commit

- Refactored several functions that previously handled only overall data so they now also support weekly data processing

Fixes #14654

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -58,8 +58,9 @@ class ExperimentResultsManager:
         reference_branch,
         outcome_group,
         outcome_slug,
+        window="overall",
     ):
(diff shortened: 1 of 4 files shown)


moibra05:
> Tested locally works great, but where's the test changes? 🤔

forgot about that. cant ever escape those lol

PR #14672, opened 2026-02-13
feat(nimbus): Ignore some analysis errors to notify

### LINKED ISSUE #14671: Ignore some analysis error
We should not alert users if encountered the below errors

```
EndedException
NoEnrollmentPeriodException
HighPopulationException
EnrollmentLongerThanAnalysisException
ExplicitSkipException
NoStartDateException
RolloutSkipException
EnrollmentNotCompleteException
PreconditionFailed
UnsupportedApplicationException
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6489)

yashikakhurana:
Because

- We are notifying every errors of the analysis to the users, some errors are safe to ignore

This commit

- Filters out the errors which should not be send 

Fixes #14671

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -811,7 +811,8 @@ class NimbusConstants:
         EXPERIMENT_LAUNCHED = "experiment_launched", "Experiment Launched"
         ENROLLMENT_HEALTHY = "enrollment_healthy", "Enrollment Healthy"
 
-    # Error types from Jetstream that are expected for non-analysis reasons and don't want alerting
+    # Error types from Jetstream that are expected for non-analysis reasons and
+    # don't want alerting
     IGNORABLE_ANALYSIS_ERROR_TYPES = {
         "EndedException",
         "NoEnrollmentPeriodException",


experimenter/experimenter/experiments/constants.py
```diff
@@ -811,6 +811,20 @@ class AlertType(models.TextChoices):
         EXPERIMENT_LAUNCHED = "experiment_launched", "Experiment Launched"
         ENROLLMENT_HEALTHY = "enrollment_healthy", "Enrollment Healthy"
 
+    # Error types from Jetstream that are expected and don't want alerting
```
mikewilli:
```suggestion
    # Error types from Jetstream that are expected for non-analysis reasons and don't want alerting
```

experimenter/experimenter/slack/tests/test_tasks.py
```diff
@@ -391,8 +391,8 @@ def test_sends_alert_for_analysis_errors(self):
                     "errors": {
                         "experiment": [
                             {
-                                "exception_type": "NoEnrollmentPeriodException",
-                                "message": "No enrollment period",
+                                "exception_type": "UnexpectedException",
+                                "message": "Unexpected error occurred",
                                 "analysis_basis": "enrollments",
                                 "segment": "all",
                             }
```
mikewilli, line 398: Would it make sense to leave both of these errors in the array and then assert that the expected one is there but he other one is not?

mikewilli: I see you have a separate test for the ignored one, so I guess that's fine too.

mikewilli (approved):
Couple comments but LGTM thanks!

PR #14697, opened 2026-02-18
feat(nimbus): add daily metrics to metric popout card

### LINKED ISSUE #14691: UI changes to show daily metrics
* In the results page section,  for any daily guardrail metrics that are detected a user should be able to see them in real time under the overview section of the relevant metric (see screenshot below)
* Theme and styling should be conssitent with how the weekly metrics are shown
* Once the weekly metrics are avaialble, they should appear below between the overview and the daily metrics section.

!image-20260206-200918.png|width=394,alt="image-20260206-200918.png"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6380)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/112918">image-20260206-200918.png</a>

moibra05:
Because

- Daily metrics are now included in the results data object but are not being displayed anywhere

This commit

- Shows daily metric data when its available in the metric popout cards
- Adds missing `window_index` field to daily metrics

Fixes #14691

### DIFF
--- a/experimenter/experimenter/jetstream/models.py
+++ b/experimenter/experimenter/jetstream/models.py
@@ -306,7 +306,7 @@ class ResultsObjectModelBase(BaseModel):
                             significance
                         )
 
-                if window == AnalysisWindow.WEEKLY:
+                if window == AnalysisWindow.WEEKLY or window == AnalysisWindow.DAILY:
                     data_point.window_index = window_index
 
                 comparison_data = getattr(metric_data, branch_comparison)
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -84,10 +84,12 @@ class ExperimentResultsManager:
 
         return max_value
 
-    def get_weekly_metric_data(self, analysis_basis, segment, reference_branch):
+    def build_window_metric_breakdown(
+        self, analysis_basis, segment, reference_branch, window
+    ):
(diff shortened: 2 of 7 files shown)


moibra05:
image

moibra05:
to be merged after #14700

experimenter/experimenter/jetstream/models.py
mikewilli, line 1: Do we want to make changes to the ingestion to cut off after 7 days if that's all we want to show? I don't know how concerned we are about potentially aggregating a ton of data that we don't plan on using.

mikewilli (approved):
Curious what you/others think regarding cutting off daily results during ingestion vs just on the UI side. I assume data storage costs aren't significant, but in the past this had an effect on network response times. Maybe that's moot with the new UI though? And I don't know enough about how we're parsing the `results_data` for the new page to know if it would affect performance there.

experimenter/experimenter/jetstream/models.py
moibra05: any performance gains on the new results page would probably be marginal since the daily metric data is only parsed once, and even then we’re only operating on a fixed slice of it. given that, i don’t feel too strongly either way. if truncating at ingestion can simplify things elsewhere, i’m fine with that, but from the new UI’s perspective i don’t think it’ll make a meaningful difference.

PR #14700, opened 2026-02-18
fix(nimbus): remove incorrect weekly dates from metric popout

### LINKED ISSUE #14699: Weekly popout dates are wrong and should not be shown
Right now we show the dates for each week in the Weekly breakdown which are indexed off of the enrollment end date (see screenshot below). However, the week’s dates are actually relative to the individual client’s enrollment date (i.e., each client is monitored for the same amount of time in a window starting at enrollment and ending `enrollment date + observation period` days later. See [these docs](https://experimenter.info/workflow/configuring#the-observation-window) for more info.

!image-20260218-192900.png|width=411,alt="image-20260218-192900.png"!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6500)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/116346">image-20260218-192900.png</a>

moibra05:
Because

- The weekly date ranges shown in the metric popout were incorrect

This commit

- Removes the date ranges from the popout
- Fixes missing data block under reference branch for DAU in popout 
- Display error card if metric is missing a weekly data block

Fixes #14699

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -101,11 +101,31 @@ class ExperimentResultsManager:
                 weekly_data = {}
 
                 for branch_slug, branch_data in data.items():
+
+                    def append_missing_weeks(entries):
+                        if entries:
+                            last_week = int(entries[-1].get("window_index", 0))
+                            for missing_week in range(1, last_week + 1):
+                                if not any(
+                                    int(e.get("window_index")) == missing_week
+                                    for e in entries
+                                ):
+                                    entries.insert(
+                                        missing_week - 1,
+                                        {
+                                            "lower": None,
+                                            "upper": None,
+                                            "significance": MetricSignificance.NEUTRAL,
+                                            "window_index": str(missing_week),
+                                        },
+                                    )
+                        return entries
+
                     # Always produce a list of pairs by zipping absolute and relative
                     # When one side is missing or shorter, pad it with None so templates
                     # can iterate without complex conditionals.
-                    abs_list = branch_data.get("absolute") or []
-                    rel_list = branch_data.get("relative") or []
+                    abs_list = append_missing_weeks(branch_data.get("absolute") or [])
+                    rel_list = append_missing_weeks(branch_data.get("relative") or [])
 
                     if abs_list or rel_list:
                         weekly_data[branch_slug] = list(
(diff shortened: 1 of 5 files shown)


moibra05:
image

experimenter/experimenter/nimbus_ui/templates/common/metric_popout.html
```diff
@@ -75,12 +75,15 @@ <h2 class="fs-6 fw-bold">Weekly breakdown</h2>
...
-                           style="height: 100px">
-                        <small class="text-muted">Week {{ forloop.counter }}</small>
-                        <p class="mb-0">{{ week.0|date:"M j" }} - {{ week.1|date:"M j" }}</p>
-                      </div>
+                    {% for branch_slug, branch_weekly_metric_data in weekly_metric_data.data.items %}
+                      {% if forloop.first %}
+                        {% for week in branch_weekly_metric_data %}
+                          <div class="{% if forloop.last %}mb-4{% endif %} mb-3 d-flex text-center rounded-4 w-100 text-start position-relative d-flex flex-column align-items-center justify-content-center"
+                               style="height: 100px">
+                            <small class="text-muted">Week {{ forloop.counter }}</small>
```
mikewilli, line 83: What happens if a week is missing in the middle?

moibra05:
image

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -101,11 +101,31 @@ def get_weekly_metric_data(self, analysis_basis, segment, reference_branch):
                 weekly_data = {}
 
                 for branch_slug, branch_data in data.items():
+
+                    def append_missing_weeks(entries):
```
mikewilli, line 105: I don't think this needs to be a nested function. Unless I'm missing something?

moibra05: its only used within this method on both the absolute and relative data point lists, so keeping it nested makes the class a bit cleaner

mikewilli: Maybe it's because of my background outside of python, but it seems to me that the default should be to define non-nested functions, and only use nesting when necessary (e.g., if the function needs access to the scope inside the containing function). Either way, I'd also suggest that it doesn't need to be defined inside the loop, as there's likely some overhead involved in re-creating the function each loop iteration (but maybe that gets optimized away?).

Obviously it's already merged, and this is not a big enough deal to worry about reverting or immediately following up with a patch (and I could also just be wrong here), but wanted to provide some more context for my initial question.

PR #14701, opened 2026-02-18
feat(nimbus): Ingest 3 day retention data

### LINKED ISSUE #14694: 3 Day Retention Ingestion
Once we have the 3 Day Retention data computed by Jetstream we need to ingest it in Experimenter before we can move on to displaying it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6453)

yashikakhurana:
Because

- Add 3-day retention data to the Results page so experiment teams can see how many users returned to Firefox within 3 days of enrollment.

This commit

- Added code to extract 3-day retention data from Jetstream
- Updated metric configuration files to include the new metric

Fixes #14694

### DIFF
--- a/experimenter/experimenter/experiments/constants.py
+++ b/experimenter/experimenter/experiments/constants.py
@@ -933,7 +933,7 @@ Optional - We believe this outcome will <describe impact> on <core metric>
             "group": "other_metrics",
             "friendly_name": "3-Day Retention",
             "slug": RETENTION_3_DAYS,
-            "description": "Users who returned in first 3 days.",
+            "description": "Users who returned to Firefox within 3 days after enrollment.",  # noqa
             "display_type": "percentage",
         },
         {
--- a/experimenter/experimenter/jetstream/models.py
+++ b/experimenter/experimenter/jetstream/models.py
@@ -164,7 +164,7 @@ class JetstreamData(RootModel[JetstreamDataPoint]):
 
         self.extend(retention_data)
 
-    def append_days_3_retention(self, daily_data):
+    def append_retention_3_days(self, daily_data):
         # Extract the 3-day retention data (window index 4)
         # without falling back to earlier windows
         retention_data = self.get_retention_by_window(
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -219,13 +219,12 @@ class ExperimentResultsManager:
                 if branch.slug == reference_branch:
                     continue
 
-                branch_group_data = (
+                metric_data = (
                     window_results.get(branch.slug, {})
                     .get("branch_data", {})
                     .get(group, {})
+                    .get(metric_slug, {})
                 )
-
-                metric_data = branch_group_data.get(metric_slug, {})
 
                 abs_point_data = metric_data.get("absolute", {}).get("first", {})
                 rel_point_data = (
--- a/experimenter/experimenter/jetstream/tests/test_results_manager.py
+++ b/experimenter/experimenter/jetstream/tests/test_results_manager.py
@@ -433,7 +433,7 @@ class TestExperimentResultsManager(TestCase):
                         "group": "other_metrics",
                         "friendly_name": "3-Day Retention",
                         "slug": "active_in_last_3_days_legacy",
-                        "description": "Users who returned in first 3 days.",
+                        "description": "Users who returned to Firefox within 3 days after enrollment.",  # noqa
                         "display_type": "percentage",
                         "overall_change": MetricSignificance.NEUTRAL,
                         "has_data": False,
@@ -472,7 +472,7 @@ class TestExperimentResultsManager(TestCase):
                         "group": "other_metrics",
                         "friendly_name": "3-Day Retention",
                         "slug": "active_in_last_3_days_legacy",
-                        "description": "Users who returned in first 3 days.",
+                        "description": "Users who returned to Firefox within 3 days after enrollment.",  # noqa
                         "display_type": "percentage",
                         "overall_change": MetricSignificance.NEUTRAL,
                         "has_data": False,
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -270,7 +270,7 @@ def get_experiment_data(experiment: NimbusExperiment):
                     raw_data[AnalysisWindow.WEEKLY][AnalysisBasis.ENROLLMENTS][segment]
                 )
                 # Append 3-day retention from daily data
-                data.append_days_3_retention(
+                data.append_retention_3_days(
                     raw_data[AnalysisWindow.DAILY][AnalysisBasis.ENROLLMENTS][segment]
                 )
                 # Create the output object (overall data)
@@ -313,7 +313,7 @@ def get_experiment_data(experiment: NimbusExperiment):
                     raw_data[AnalysisWindow.WEEKLY][AnalysisBasis.EXPOSURES][segment]
                 )
                 # Append 3-day retention from daily data
-                data.append_days_3_retention(
+                data.append_retention_3_days(
                     raw_data[AnalysisWindow.DAILY][AnalysisBasis.EXPOSURES][segment]
                 )
 
--- a/experimenter/experimenter/jetstream/tests/test_jetstream_data.py
+++ b/experimenter/experimenter/jetstream/tests/test_jetstream_data.py
@@ -46,7 +46,7 @@ class TestJetstreamData(TestCase):
         self.assertIn(expected_control, data)
         self.assertIn(expected_variant, data)
 
-    def test_append_days_3_retention_extracts_data(self):
+    def test_append_retention_3_days_extracts_data(self):
         retention = JetstreamDataPoint(
             metric=Metric.RETENTION_3_DAYS,
             statistic=Statistic.BINOMIAL,
@@ -57,6 +57,6 @@ class TestJetstreamData(TestCase):
         )
 
         data = JetstreamData([])
-        data.append_days_3_retention([retention])
+        data.append_retention_3_days([retention])
 
         self.assertIn(retention, data)


yashikakhurana:
Screenshot 2026-02-18 at 3 57 34 PM

experimenter/experimenter/experiments/constants.py
```diff
@@ -904,6 +904,8 @@ class FirefoxLabsGroups(models.TextChoices):
     DAILY_ACTIVE_USERS = "client_level_daily_active_users_v2"
     DAYS_OF_USE = "days_of_use"
     RETENTION = "retained"
+    ACTIVE_IN_LAST_3_DAYS_LEGACY = "active_in_last_3_days_legacy"
+    ACTIVE_IN_LAST_3_DAYS = "active_in_last_3_days"
```
mikewilli: So the non-legacy version of this metric is the Glean version, and we'll only use one at a time. We will use legacy for now until data science finishes validating Glean analysis, and then just switch over. 

I'd suggest we only have one of these, name it "3 days retention" and then just swap the value of it whenever we do switch to the Glean version.
```suggestion
    3_DAYS_RETENTION = "active_in_last_3_days_legacy"
```

mikewilli: I also see you have these defined under the `Metric` enum, which seems like the right place to me. It seems odd to me that we have metrics defined here as well.

experimenter/experimenter/jetstream/models.py
```diff
@@ -161,6 +163,27 @@ def append_retention_data(self, weekly_data):
 
         self.extend(retention_data)
 
+    def append_active_in_last_3_days(self, daily_data):
```
mikewilli: This function doesn't seem to do anything other than pull out the metric, which should occur with the normal behavior anyway. What we want to do is get the 3rd day's value, similar to how we get the 2nd week's value. I think you could make some small changes to `append_retention_data` (and rename `get_week_x_retention` to something like `get_retention_by_window`) and then just call that instead of creating a new function here.

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -205,25 +205,47 @@ def metric_has_errors(self, metric_slug, analysis_basis, segment):
...
     ):
         if self.experiment.results_data:
-            window_results = (
-                self.get_window_results(analysis_basis, segment, "overall")
-                or self.get_window_results(analysis_basis, segment, "weekly")
-                or self.get_window_results(analysis_basis, segment, "daily")
-            )
+            # 3-day retention only exists in DAILY window
+            if self._is_3day_retention_metric(metric_slug):
+                window_results = self.get_window_results(analysis_basis, segment, "daily")
```
mikewilli: The changes to the jetstream client mean that this metric is in the OVERALL window, so I think it can be treated basically the same as retention, which I don't think has any special case logic. The main distinction between 3 days retention and retention is that 3days retention is in daily+overall and retention is in weekly+overall.

experimenter/experimenter/metrics/metric-hub-main/jetstream/defaults/firefox_desktop.toml
```diff
@@ -1,7 +1,7 @@
 # This defines the default metrics and statistics calculated for each experiment
 
 [metrics]
-daily = ["retained", "unenroll", "client_level_daily_active_users_v2"]
+daily = ["retained", "active_in_last_3_days", "active_in_last_3_days_legacy", "unenroll", "client_level_daily_active_users_v2"]
```
mikewilli: I think this directory is typically pulled in via external config update. Should we skip making this change here and just let it get picked up by that?

experimenter/experimenter/jetstream/results_manager.py
mikewilli: This also means that I don't think you need the other changes to this file.

experimenter/experimenter/metrics/metric-hub-main/jetstream/defaults/firefox_desktop.toml
yashikakhurana: I was not sure if it gets picked, @moibra05 helped me to add metric here, I am not sure, should I remove it?

moibra05: you shouldn't need to change anything here. this copy of the metric-hub definitions only serves to define metric areas for default metrics otherwise mike is correct

yashikakhurana: okay done, removed

experimenter/experimenter/jetstream/results_manager.py
yashikakhurana: so it prioritize weekly first as it returns {}, never reaches daily

yashikakhurana:
Hi @mikewilli, thanks for the feedback! I've implemented all your suggestions:

- Changed from `ACTIVE_IN_LAST_3_DAYS_LEGACY` and `ACTIVE_IN_LAST_3_DAYS` to a single `DAYS_3_RETENTION = "active_in_last_3_days_legacy" `that can be swapped when transitioning to Glean.

-  Refactored append_days_3_retention() to follow the same pattern as `append_retention_data()` - just filters and extends without special logic.

- Removed _is_3day_retention_metric() and _get_metric_data_from_branch() from results_manager.

Window Priority Logic
I discovered the metric data exists in the DAILY window (where Jetstream calculates it), not OVERALL yet. To ensure metric_has_data() finds it, I added a DAILY_FIRST_METRICS set to prioritize the daily window for metrics like 3-day retention (daily+overall) vs retention (weekly+overall).

experimenter/experimenter/jetstream/tests/test_results_manager.py
```diff
@@ -449,23 +465,71 @@ def test_get_branch_data_returns_correct_data(self):
...
                         "slug": "days_of_use",
                         "description": "Average number of days each client sent a main ping.",  # noqa E501
-                        "overall_change": "neutral",
+                        "overall_change": MetricSignificance.NEUTRAL,
+                        "has_data": False,
+                    },
+                ],
+            ),
+            (
+                "active_in_last_3_days_legacy",
```
moibra05: no need for this extra test. `test_get_kpi_metrics_returns_correct_metrics` is meant to test the case where older experiments make use of `days_of_use` (DOU) metric in place of `client_level_daily_active_users_v2` (DAU)

experimenter/experimenter/jetstream/tests/test_results_manager.py
```diff
@@ -1576,6 +1640,38 @@ def test_get_overall_change(self, results_data, expected):
                 },
                 False,
             ),
+            (
+                "active_in_last_3_days_legacy",
```
moibra05: this seems to be testing the same behaviour as the first case. maybe not needed?

experimenter/experimenter/jetstream/tests/test_results_manager.py
yashikakhurana: oh okay, I have removed this one

experimenter/experimenter/jetstream/tests/test_results_manager.py
yashikakhurana: removed

experimenter/experimenter/experiments/constants.py
```diff
@@ -928,6 +929,13 @@ class FirefoxLabsGroups(models.TextChoices):
             "description": "Percentage of users who returned to Firefox two weeks later.",
             "display_type": "percentage",
         },
+        {
+            "group": "other_metrics",
+            "friendly_name": "3-Day Retention",
+            "slug": DAYS_3_RETENTION,
+            "description": "Users who returned in first 3 days.",
```
mikewilli:
```suggestion
            "description": "Users who returned to Firefox within 3 days after enrollment.",
```

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -209,23 +215,33 @@ def metric_has_data(
...
-                metric_data = (
+                branch_group_data = (
                     window_results.get(branch.slug, {})
                     .get("branch_data", {})
                     .get(group, {})
-                    .get(metric_slug, {})
                 )
 
+                metric_data = branch_group_data.get(metric_slug, {})
+
```
mikewilli: Why this change?

experimenter/experimenter/experiments/constants.py
```diff
@@ -904,6 +904,7 @@ class FirefoxLabsGroups(models.TextChoices):
     DAILY_ACTIVE_USERS = "client_level_daily_active_users_v2"
     DAYS_OF_USE = "days_of_use"
     RETENTION = "retained"
+    DAYS_3_RETENTION = "active_in_last_3_days_legacy"
```
mikewilli: What do you think about `RETENTION_3_DAYS` so we maintain the "3 days" in order but avoid starting a constant with a number?

experimenter/experimenter/jetstream/models.py
```diff
@@ -161,6 +162,16 @@ def append_retention_data(self, weekly_data):
...
+    def append_days_3_retention(self, daily_data):
+        daily_data = daily_data or []
+        retention_data = [
+            jetstream_data_point
+            for jetstream_data_point in daily_data
+            if jetstream_data_point.metric == Metric.DAYS_3_RETENTION
+        ]
+
+        self.extend(retention_data)
+
```
mikewilli: This isn't quite right. What we want to do is take the relevant `daily` result and inject it into the `overall` results so it appears alongside those values. So that means we should extract window index 4 data (so we get days 1-3), basically the same as we do for week 2 above except that we don't want to fall back to earlier windows if it's not available.

If it were me, I'd refactor `append_retention_data` and `get_week_x_retention` to be more generic to handle both the weekly retention and 3 days retention cases.

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -209,23 +215,33 @@ def metric_has_data(
...
+                    self.get_window_results(analysis_basis, segment, "daily")
+                    or self.get_window_results(analysis_basis, segment, "overall")
+                    or self.get_window_results(analysis_basis, segment, "weekly")
+                )
+            else:
+                window_results = (
+                    self.get_window_results(analysis_basis, segment, "overall")
+                    or self.get_window_results(analysis_basis, segment, "weekly")
+                    or self.get_window_results(analysis_basis, segment, "daily")
+                )
```
mikewilli: I don't think this is right. We should be injecting the specific daily result (from window index 4) into `overall` and that's what we prefer. If that isn't available, we can decide whether we want to show the remaining daily results or not, but it doesn't matter if we check for weekly first because that will return `{}` and the `or` will just fall past that.

experimenter/experimenter/metrics/metric-hub-main/jetstream/defaults/firefox_desktop.toml
```diff
@@ -80,6 +80,8 @@ kpi = [
   "client_level_daily_active_users_v2",
   "days_of_use",
   "retained",
+  "active_in_last_3_days",
+  "active_in_last_3_days_legacy",
```
mikewilli: I still don't understand why we're making changes to this file, I thought this folder was managed like other external configs, see https://github.com/mozilla/experimenter/tree/main/experimenter/experimenter/segments and https://github.com/mozilla/experimenter/tree/main/experimenter/experimenter/outcomes.

Am I misunderstanding how updating external configs works?

yashikakhurana: I will remove it from here

experimenter/experimenter/experiments/constants.py
mikewilli: I think you changed this by accepting my suggestion, but then accidentally changed it back?

experimenter/experimenter/jetstream/models.py
```diff
@@ -144,31 +144,32 @@ def append_population_percentages(self):
...
+        retention_data = self.get_retention_by_window(2, weekly_data, Metric.RETENTION)
         if len(retention_data) == 0:
-            retention_data = self.get_week_x_retention(1, weekly_data)
+            retention_data = self.get_retention_by_window(
+                1, weekly_data, Metric.RETENTION
+            )
 
         self.extend(retention_data)
 
     def append_days_3_retention(self, daily_data):
```
mikewilli: Nit: can you update this to match the new naming (retention_3_days)?

mikewilli (approved):
r+wc

Logic all looks good so nothing blocking, but please address my comments below and the remaining one from last review. Thanks, should be very impactful updates we've gotten through in a short timeline!

PR #14728, opened 2026-02-23
bug(nimbus): Prevent JEXL errors from breaking the summary page

### LINKED ISSUE #14727: Cannot load experiments with invalid targeting
We try to pretty print JEXL expressions, but there are a few experiments in the database with invalid targeting expressions (due to being generated incorrectly and cached on the `published_dto`). The summary pages for these experiments cannot be loaded.

We should do our best to pretty-print JEXL expressions, but fall back to returning the expression as-is and report the error to Sentry so that it can still be tracked without bringing down the page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6515)

freshstrangemusic:
Because:

- some experiments in the prod database have invalid targeting cached on their published_dto; and
- we try to pretty-print these invalid expressions and raise an Exception

this commit:

- adds a try/except around the pretty-printing to fall back to the plain expression and report the error to sentry; and
- adds a validation to ensure that the entire expression parses before launch

Fixes #14727

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -32,6 +32,7 @@ from experimenter.experiments.constants import (
     NimbusConstants,
     TargetingMultipleKintoCollectionsError,
 )
+from experimenter.experiments.jexl_utils import JEXLParser
 from experimenter.experiments.models import (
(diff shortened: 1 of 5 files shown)


freshstrangemusic:
image

PR #14761, opened 2026-02-26
feat(nimbus): show percentage bounds on hover instead of high precision for relative metric values

### LINKED ISSUE #14756: Percent change hover text should show lower/upper range
I think that the percentages all have `lower` and `upper` bounds, and also that these are more useful than the high precision values that we currently show when hovering on the percentages (high precision can be useful for absolute values but less so for percentages). I propose we instead show the lower and upper bounds when hovering over the percent change values.

For example, this `20.3767776…` hover text would be something like `15.3% - 24.9%` based on what the `lower` and `upper` percent values are. 

!image-20260217-180730.png|width=163,alt="image-20260217-180730.png"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6496)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/116008">image-20260217-180730.png</a>

moibra05:
Because

- Hovering over the relative percent values would show a tooltip with the full precision value
- It is not very useful to display the full value and we would instead benefit from showing just the percent value range

This commit

- Updates the tooltip to show relative percent range value

Fixes #14756

### DIFF
--- a/experimenter/experimenter/nimbus_ui/templates/common/metric_card.html
+++ b/experimenter/experimenter/nimbus_ui/templates/common/metric_card.html
@@ -33,13 +33,15 @@
         {% if sig == "positive" %}
           <span class="fw-semibold"
                 data-bs-toggle="tooltip"
-                data-bs-title="{{ percentage|to_percentage }}">{{ percentage|to_percentage:1 }} increase</span>
+                data-bs-title="Range: {{ lower_percent|to_percentage:2 }} to {{ upper_percent|to_percentage:2 }}">{{ percentage|to_percentage:1 }} average increase</span>
         {% elif sig == "negative" %}
           <span class="fw-semibold"
                 data-bs-toggle="tooltip"
-                data-bs-title="{{ percentage|to_percentage }}">{{ percentage|to_percentage:1 }} decrease</span>
+                data-bs-title="Range: {{ lower_percent|to_percentage:2 }} to {{ upper_percent|to_percentage:2 }}">{{ percentage|to_percentage:1 }} average decrease</span>
         {% else %}
-          <span class="fw-semibold">No notable change</span>
+          <span class="fw-semibold"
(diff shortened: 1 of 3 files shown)


moibra05:
i opted to use the `(lower to upper)` convention here b/c i noticed it was a little confusing when either value was negative. alternatively we could take the absolute but i think that would also be a bit confusing if you saw something like `(bigger_num - smaller_num)` in the case of converting eg `(-2.0 - -1.0)`. maybe someone knows of a third option that could be better?

jaredlockhart:
@moibra05 Maybe you could include some screenshots to show examples? I think that'd make easier to see the options?

moibra05:
image

moibra05:
image

mikewilli:
I see the confusion here on how to display it. I wonder if it's worth prefixing with "Range: " so that it's clear what these numbers represent? Or something obvious like this?
```
upper: -0.96%
lower: -1.24%
```

moibra05:
> I see the confusion here on how to display it. I wonder if it's worth prefixing with "Range: " so that it's clear what these numbers represent? Or something obvious like this?
> 
> ```
> upper: -0.96%
> lower: -1.24%
> ```

oh good idea. i think prefixing with "Range: " helps a bit more in clearing up any ambiguity while keeping the tooltip fairly minimal

moibra05:
image

mikewilli:
Yea that looks better to me. I think it'd also help if we said "average increase" by the single number (so **0.8% average increase** in the screenshot example), what do you think?

moibra05:
not opposed to this. i don’t see any real downside to being more explicit about what that number represents

moibra05:
image

mikewilli (approved):
Nice, looks clear to me!

PR #14791, opened 2026-03-02
feat(nimbus): Only show 3 days retention data for desktop application

### LINKED ISSUE #14789: Only show 3 day retention for firefox desktop
We should show the 3 day retention for firefox desktop only

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6552)

yashikakhurana:
Because

- We recently added 3-day retention support the desktop application, but in the ui it is showing for every application and when data is not available it is showing metric data unavailable

This commit

- Only includes the 3 days retention for the desktop application

Fixes #14789

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -281,7 +281,6 @@ class ExperimentResultsManager:
     def get_kpi_metrics(
         self, analysis_basis, segment, reference_branch, window="overall"
     ):
-        from experimenter.experiments.models import NimbusExperiment
 
         kpi_metrics = NimbusConstants.KPI_METRICS.copy()
 


yashikakhurana:
> Is it possible to include it in the KPI_METRICS list but then just not show it if it isn't in the results? That seems cleaner than retroactively injecting it into a list that is ostensibly constant.



> Is it possible to include it in the KPI_METRICS list but then just not show it if it isn't in the results? That seems cleaner than retroactively injecting it into a list that is ostensibly constant.

I found the other way also to check when building kpi, let me know if you think its better

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -281,7 +281,18 @@ def check_valid_point(point):
     def get_kpi_metrics(
         self, analysis_basis, segment, reference_branch, window="overall"
     ):
+        from experimenter.experiments.models import NimbusExperiment
+
         kpi_metrics = NimbusConstants.KPI_METRICS.copy()
+
+        # 3-day retention is only available for Desktop experiments
+        if self.experiment.application != NimbusExperiment.Application.DESKTOP:
```
jaredlockhart:
```suggestion
        if self.experiment.application != self.experiment.Application.DESKTOP:
```

mikewilli:
This looks better and is totally fine for now, but it seems like we could just make it fully dynamic based on what results are available in the data. Isn't that how we build the rest of the page?

yashikakhurana:
> This looks better and is totally fine for now, but it seems like we could just make it fully dynamic based on what results are available in the data. Isn't that how we build the rest of the page?

for kpi we build it like this, but for others it is based on what's available

PR #14810, opened 2026-03-03
fix(nimbus): fix unhandled invalid metric data structure

### LINKED ISSUE #14675: TypeError: bad operand type for abs(): 'NoneType'
Sentry Issue: [EXPERIMENTER-PROD-186](https://mozilla.sentry.io/issues/7264490652/?referrer=github_integration)

```
TypeError: bad operand type for abs(): 'NoneType'
(2 additional frame(s) were not displayed)
...
  File "experimenter/nimbus_ui/views.py", line 825, in get_context_data
    extreme_bound = results_manager.get_max_metric_value(
  File "experimenter/jetstream/results_manager.py", line 83, in get_max_metric_value
    max_value = max(max_value, abs(lower), abs(upper))
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6494)

moibra05:
Because

- Some experiments were producing invalid metrics with data of the form `{"window_index": "1"}` missing values for `lower`, `upper` and `point`
- The previous validation logic considered a metric valid if any field was present, which allowed these incomplete data points to pass through.

This commit

- Updates the `metric_has_data` validation logic to properly mark metrics with incomplete data as erroneous

Fixes #14675

### DIFF
--- a/experimenter/experimenter/jetstream/results_manager.py
+++ b/experimenter/experimenter/jetstream/results_manager.py
@@ -270,11 +270,14 @@ class ExperimentResultsManager:
                 )
 
                 def check_valid_point(point):
-                    return any(value != 0 for value in point.values())
+                    return (
+                        point.get("lower") and point.get("upper") and point.get("point")
+                    )
 
-                if (not abs_point_data or not check_valid_point(abs_point_data)) and (
-                    not rel_point_data or not check_valid_point(rel_point_data)
-                ):
+                if (
+                    (not abs_point_data or not check_valid_point(abs_point_data))
+                    and metric_slug != NimbusConstants.DAILY_ACTIVE_USERS
+                ) or (not rel_point_data or not check_valid_point(rel_point_data)):
                     return False
         return True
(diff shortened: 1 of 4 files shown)


experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -521,6 +524,7 @@ def is_metric_notable(slug, group):
                 if (
                     is_metric_notable(metric["slug"], metric["group"])
                     and metric not in metric_areas[NimbusUIConstants.NOTABLE_METRIC_AREA]
+                    and metric["has_data"]
```
jaredlockhart: I guess we should be stricter about how we handle stuff like.  Key lookups like this can fail if the key isn't present, so we should prefer `.get()`, and `.get()` will fall back to `None` if no default is provided, so we should in general I think be doing something like `.get(..., <some default>)` as often as we can so we don't hit like, key errors, NoneType errors, etc

PR #14822, opened 2026-03-04
feat(nimbus): Format slack message

### LINKED ISSUE #14725: Refactor slack messages and add more emojis to the message
Lets move all the slack message to a constants file and add emojis to the message where necessary

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6512)

yashikakhurana:
Because

- Slack notifications are getting visually noisy. We want to increase the readability of the Slack messages.

This commit

- Formats the messages - @ name to after the experiment name 
:rocket: Requesting launch <slug> @ whoever
:double_vertical_bar:  Requesting pause <slug> @ whoever
:stop-:  Requesting end <slug> @ whoever
:arrows_counterclockwise:  Requesting update <slug> @ whoever

- Moves the messages into a constants 
- Add emojis in the results messages such as when results are available and when there is an error in analysis 

Fixes #14725


experimenter/experimenter/experiments/constants.py
```diff
@@ -1179,6 +1179,65 @@ class FirefoxLabsGroups(models.TextChoices):
         SLACK_ACTION_UPDATE_REQUEST: AlertType.UPDATE_REQUEST,
     }
 
+    SLACK_DM_PREFIX = "🔔 Join {channel} to get slack notifications: {message}"
```
jaredlockhart: Can't we put these in `slack/constants.py`?

yashikakhurana: good idea, let me move all

PR #14824, opened 2026-03-04
fix(nimbus): exclude Preview status experiments from v8 API

### LINKED ISSUE #14823: V8 API should exclude Preview status experiments
## Summary

The v8 API endpoint (`/api/v8/experiments/`) currently excludes only `Draft` status experiments from its queryset. It should also exclude `Preview` status experiments, since preview experiments are not intended to be served to clients and may be subject to change.

## Expected Behavior

The `/api/v8/experiments/` endpoint should exclude both `Draft` and `Preview` status experiments.

## Actual Behavior

Only `Draft` status experiments are excluded. `Preview` experiments are included in the response.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6562)

jaredlockhart:
Because

* Preview experiments are not intended to be served to clients and
  should be excluded from v8 like Draft experiments

This commit

* Adds `NimbusExperiment.Status.PREVIEW` to the exclude list in
  `NimbusExperimentViewSet` queryset
* Updates the test to expect Preview experiments to be excluded

Fixes #14823

### DIFF
--- a/experimenter/experimenter/experiments/api/v8/views.py
+++ b/experimenter/experimenter/experiments/api/v8/views.py
@@ -74,7 +74,12 @@ class NimbusExperimentDraftViewSet(NimbusExperimentViewSet):
 
     queryset = (
         NimbusExperiment.objects.with_related()
-        .filter(status=NimbusExperiment.Status.DRAFT)
+        .filter(
+            status__in=[
+                NimbusExperiment.Status.DRAFT,
+                NimbusExperiment.Status.PREVIEW,
+            ]
+        )
         .order_by("slug")
     )
 
--- a/experimenter/experimenter/experiments/tests/api/v8/test_views.py
+++ b/experimenter/experimenter/experiments/tests/api/v8/test_views.py
@@ -265,7 +265,10 @@ class TestNimbusExperimentDraftViewSet(
                 slug=lifecycle.name,
             )
 
-            if experiment.status == NimbusExperiment.Status.DRAFT:
+            if experiment.status in [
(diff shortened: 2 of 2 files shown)


freshstrangemusic (approved):

experimenter/experimenter/experiments/api/v8/views.py
```diff
@@ -54,7 +54,12 @@ class NimbusExperimentViewSet(
     lookup_field = "slug"
     queryset = (
         NimbusExperiment.objects.with_related()
-        .exclude(status__in=[NimbusExperiment.Status.DRAFT])
+        .exclude(
+            status__in=[
+                NimbusExperiment.Status.DRAFT,
+                NimbusExperiment.Status.PREVIEW,
+            ]
+        )
```
mikewilli, line 62: Would it make more sense to have this to be an inclusive filter on LIVE/COMPLETE? Not sure if we'd ever change the status options here where it'd be meaningful, but just a thought.

mikewilli (approved):
Looks good to me, thanks! One thought on the mechanics but not blocking.

experimenter/experimenter/experiments/api/v8/views.py
jaredlockhart: I know I've thought about it too.  I'm actually surprised I didn't either exclude preview or restrict it to live/complete from the get go I can't remember if there was a particular reason for it.  Anyways easy enough change if we wanna do it in the future.

PR #14829, opened 2026-03-04
feat(nimbus):Link to results page

### LINKED ISSUE #14753: Link to results page directly when results are available
When we send the slack notification regarding results such as results available or analysis errors, in that case we should link to the results page directly

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6537)

yashikakhurana:
Because

- We should send a notification to the users with the experiment result link instead of experiment link when results are available or if there is any analysis error.

This commit

- Sends the results page link for the results and analysis error

Fixes #14753

### DIFF
--- a/experimenter/experimenter/experiments/tests/test_email.py
+++ b/experimenter/experimenter/experiments/tests/test_email.py
@@ -57,6 +57,7 @@ class TestNimbusEmail(TestCase):
             action_text=SlackConstants.SLACK_EMAIL_ACTIONS[
                 NimbusExperiment.EmailType.EXPERIMENT_END
             ],
+            link_url=experiment.experiment_url,
         )
 
     @patch("experimenter.slack.tasks.nimbus_send_slack_notification.delay")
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/slack/notification.py
```diff
@@ -111,7 +112,9 @@ def send_slack_notification(
     if mentioned_user_ids:
         all_mentions.extend(f"<@{user_id}>" for user_id in mentioned_user_ids)
 
-    message = f"{action_text}: <{experiment.experiment_url}|{experiment.name}>"
+    # Use results_url if provided, otherwise use experiment_url
+    url = results_url if results_url else experiment.experiment_url
+    message = f"{action_text}: <{url}|{experiment.name}>"
     if all_mentions:
         message = f"{message} @ {' '.join(all_mentions)}"
```
jaredlockhart, line 118: Oh you know another way to do it would just be like

```py
def send_slack_notification(...., link_url, ....)
    ...
    message = f"{action_text}: <{link_url}|{experiment.name}>"



def launch_experiment(....):
    send_slack_notification(..., link_url=experiment.get_absolute_url, ...)

def results_available(...):
     send_slack_notification(..., link_url=experiment.get_results_url, ...)
```

and just remove the 

```py
    url = results_url if results_url else experiment.experiment_url
```

check altogether?

yashikakhurana: good suggestion let me do that

PR #14845, opened 2026-03-05
fix(nimbus): add pagination to YAML export endpoint

### LINKED ISSUE #14844: YAML export endpoint times out on production (needs pagination)
## Problem

The `/api/v5/yaml/` endpoint serializes all ~1000+ completed experiments into a single response. On production, an upstream proxy terminates the connection at 30s with a 502, making the endpoint unusable.

## Solution

Add pagination (100 experiments per page) to the YAML export endpoint. Each page returns well under the 30s timeout. The response includes `count`, `next`, `previous`, and `experiments` fields so clients can iterate through all pages.

Also moves sorting from Python `sorted()` to ORM `.order_by()` (required for pagination to work) and removes YAML from cache warming (paginated pages are fast enough without pre-warming).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6574)

jaredlockhart:
Because

* The `/api/v5/yaml/` endpoint serializes all ~1000+ completed experiments
  into a single response which times out at 30s on production (502)
* The upstream proxy has a 30s connection timeout that cannot be changed
* Cache warming does not help because the response is too large to
  transfer within the timeout even when served from cache

This commit

* Adds `YamlExportPagination` (100 experiments per page) to the YAML view
* Updates the YAML renderer to handle paginated response shape with
  count, next, previous, and experiments fields
* Moves sorting from Python `sorted()` to ORM `.order_by()` which is
  required for DRF pagination to work correctly
* Removes YAML from cache warming since paginated pages are fast enough
  without pre-warming
* Updates tests for the new paginated response format and adds
  pagination-specific tests

Fixes #14844

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/views.py
+++ b/experimenter/experimenter/experiments/api/v5/views.py
@@ -1,5 +1,7 @@
 import yaml
+from django.db.models import F
 from rest_framework.generics import ListAPIView, UpdateAPIView
+from rest_framework.pagination import PageNumberPagination
 from rest_framework.renderers import BaseRenderer
 from rest_framework_csv.renderers import CSVRenderer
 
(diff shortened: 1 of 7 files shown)


experimenter/experimenter/experiments/api/v5/views.py
```diff
@@ -84,20 +104,11 @@ class NimbusExperimentYamlListView(CachedListMixin, ListAPIView):
             "excluded_experiments",
         )
         .filter(is_archived=False, status=NimbusExperiment.Status.COMPLETE)
+        .order_by(F("_start_date").desc(nulls_last=True), "-id")
```
yashikakhurana, line 100: 🎉

PR #14853, opened 2026-03-05
feat(nimbus): Add cancel emoji if requests get cancelled

### LINKED ISSUE #14840: Cancelling launch request should add a cross emoji
if someone requests and then cancels request we should like put an 

!274c@2x.png|width=250,alt=":x:"!

on it or something

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6554)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/117460">274c@2x.png</a>

yashikakhurana:
Because

- Users can cancel launch requests and resubmit them. However, we don’t have any indication in Slack messages that a request has already been cancelled, so reviewers aren’t aware of this.

This commit

- Adds the cancel emoji to the requests if the request is already cancelled
- This also fixes to look for the latest request for the emoji to add as we have seen it tried to add emoji to the old requests

Fixes #14840 #14838

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -1386,7 +1386,9 @@ class CancelRequestMixin:
         experiment = super().save(commit=commit)
         if self.cancel_request_alert_type:
             add_emoji_to_message_async.delay(
-                experiment.id, self.cancel_request_alert_type, "x"
+                experiment.id,
+                self.cancel_request_alert_type,
+                SlackConstants.EmojiReaction.CANCEL,
             )
         return experiment
 
@@ -1596,7 +1598,9 @@ class ReviewToApproveForm(UpdateStatusForm):
         )
 
         add_emoji_to_message_async.delay(
-            experiment.id, NimbusConstants.AlertType.LAUNCH_REQUEST, "eyes"
+            experiment.id,
+            NimbusConstants.AlertType.LAUNCH_REQUEST,
+            SlackConstants.EmojiReaction.APPROVE,
         )
 
         return experiment
@@ -1815,7 +1819,9 @@ class ApproveUpdateRolloutForm(UpdateStatusForm):
             countdown=5, args=[experiment.kinto_collection]
         )
         add_emoji_to_message_async.delay(
-            experiment.id, NimbusConstants.AlertType.UPDATE_REQUEST, "eyes"
+            experiment.id,
+            NimbusConstants.AlertType.UPDATE_REQUEST,
+            SlackConstants.EmojiReaction.APPROVE,
         )
         return experiment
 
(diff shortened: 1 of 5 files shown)


experimenter/experimenter/slack/notification.py
```diff
@@ -246,21 +246,31 @@ def add_eyes_emoji_to_launch_message(experiment, alert_type):
...
         logger.error(f"{msg}: {e}")
         return False
+
+
+def add_eyes_emoji_to_launch_message(experiment, alert_type):
+    return add_emoji_to_slack_message(experiment, alert_type, "eyes")
+
+
+def add_cancel_emoji_to_request(experiment, alert_type):
+    return add_emoji_to_slack_message(experiment, alert_type, "x")
```
jaredlockhart: Is this really necessary?

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -1376,6 +1379,17 @@ def save(self, commit=True):
...
 
 
+class CancelRequestMixin:
+    cancel_request_alert_type = None
+
+    @transaction.atomic
+    def save(self, commit=True):
+        experiment = super().save(commit=commit)
+        if self.cancel_request_alert_type:
+            add_cancel_emoji_to_request(experiment, self.cancel_request_alert_type)
```
jaredlockhart: Actually I'm curious, why we have a special function for this and not just a more reusable function?

Also I realize now, this is making synchronous inline requests and we're not pushing communication to slack into a celery task?

In general it's not a great idea to put synchronous outgoing network calls in the middle of request flows because it could hang or fail and then block the Experimenter user.

Out of scope for this PR but we should think about whether we should move that out of the request flow.

jaredlockhart (changes requested):

experimenter/experimenter/nimbus_ui/forms.py
yashikakhurana: oh actually I was following the pattern one you yesterday suggested, I can make reusable one, yes the alerts like launch request etc etc are all async, emojis are not async, its in to do to make them async as well

yashikakhurana: I made emojis react async here too :)

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -1376,6 +1378,19 @@ def save(self, commit=True):
...
 
+class CancelRequestMixin:
+    cancel_request_alert_type = None
+
+    @transaction.atomic
+    def save(self, commit=True):
+        experiment = super().save(commit=commit)
+        if self.cancel_request_alert_type:
+            add_emoji_to_message_async.delay(
+                experiment.id, self.cancel_request_alert_type, "x"
```
jaredlockhart: Why do we have little emoji string constant keys littered around? Shouldn't that be an enum key or something?

jaredlockhart (approved):
Yeah this reads much better now ty @yashikakhurana 🎉 🎉 🎉 🎉

Issue #14860, opened 2026-03-09
Feature value fields vanish after saving on branch page


jaredlockhart:
Looks like when you save the branch page, the feature value editors disappear.  Probably just missing a codemirror reinitialize after an htmx swap somewhere.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6583)

PR #14862, opened 2026-03-09
bug(nimbus): Don't leak CodeMirror editors

### LINKED ISSUE #14861: Theme handler leaks all codemirror editors
We add our code mirror editors to a `Set`, which prevents them from being gc-ed when we save the page and replace them with new editors.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6584)

freshstrangemusic:
Because:

- we were storing all our CodeMirror editor views in a set so we can iterate over them in response to theme changes;
- this creates a strong reference, preventing the editors from being garbage collected

this commit:

- wraps each editor view in a WeakRef, which will allow them to be garbage collected.

Fixes #14861


freshstrangemusic:
Blocked: I want to test this once #14860 lands

PR #14866, opened 2026-03-09
fix(nimbus): sync is_paused from Remote Settings on rollback in handle_waiting_experiments

### LINKED ISSUE #14864: Experiment shows paused true but still show end enrollment button
Just encountered an experiment 

https://experimenter.services.mozilla.com/nimbus/tracker-resource-tailing-beta/summary/

that shows both 'end enrollment' and 'end experiment' buttons, but when clicking end enrollment it says

Cannot perform this action: experiment must be in state (NimbusConstants.Status.LIVE, None, NimbusConstants.PublishStatus.IDLE, False), but is currently in state ('Live', None, 'Idle', True).

the last bool there is the 'is_paused' bool, so if it is paused then we shouldn't be showing the end enrollment button at all.  

Let's figure out what's going on there.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6585)

jaredlockhart:
Because

* When a kinto publish for "end enrollment" fails via the catch-all
  `handle_waiting_experiments` path (manual rollback, missing rejection),
  `is_paused` was left as `True` even though kinto never accepted the change
* This left the experiment stuck showing both "End Enrollment" and
  "End Experiment" buttons, with "End Enrollment" producing an error
* The root cause is that `handle_waiting_experiments` reset `publish_status`
  and `status_next` but did not sync `is_paused` with the actual RS state

This commit

* Passes kinto main records to `handle_waiting_experiments`
* Syncs `is_paused` from the RS record's `isEnrollmentPaused` field on
  rollback, defaulting to `False` if the record doesn't exist
* Updates the End Enrollment (Approve/Reject+Manual Rollback) mermaid
  diagram to document the `is_paused` sync behavior
* Adds tests for pause rollback (`is_paused` resets to `False`) and
  end-experiment rollback (`is_paused` preserved as `True` from RS)

Fixes #14864

### DIFF
--- a/docs/experimenter/README.md
+++ b/docs/experimenter/README.md
@@ -1097,11 +1097,11 @@ A live experiment that is published in Remote Settings has passed its planned en
 
     Note over Experimenter Backend: The scheduled background task <br/> is invoked and finds the collection in <br/> to-sign with no record of the rejection
 
-
-    rect rgb(204,255,255) 
+    rect rgb(204,255,255)
         Experimenter Worker->>Remote Settings Backend: Check collection (timeout) <br/> RS status: to-sign
-        Experimenter Worker->>Experimenter Backend:  Status: Live <br/> Publish status: Idle <br/> Status next: <none> <br/> is_paused: False <br/> + changelog
-    end 
+        Note over Experimenter Worker: Worker syncs is_paused from <br/> RS record's isEnrollmentPaused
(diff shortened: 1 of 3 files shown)


yashikakhurana (approved):
Nice catch, and good test coverage, thank you @jaredlockhart

PR #14901, opened 2026-03-12
feat(nimbus): Add nimbus-devtools messaging integration

### LINKED ISSUE #14849: Experimenter: nimbus-devtools messaging integration
Experimenter needs to embed some metadata in markup on the branch page so that Nimbus Devtools can understand what application and feature ID each feature value is for. That will allow nimbus devtools to embed preview links in the html.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6577)

freshstrangemusic:
Because:

- nimbus-devtools is going to be integrating into experimenter;
- nimbus-devtools needs metadata about the experiment and page layout to support one-click message previews and force-enrollment; and
- our buttons on CodeMirror editors are slightly off and can render outside the editor if it is only a single line high

this commit:

- embeds metadata into experiment summary and branch edit pages; and
- fixes up some styling inconsistencies with the button overlays on our CodeMirror editors.

Fixes #14849

### DIFF
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/edit_branches.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/edit_branches.html
@@ -1,13 +1,14 @@
 {% extends "nimbus_experiments/experiment_base.html" %}
 
 {% load static %}
-{% load nimbus_extras %}
+{% load nimbus_devtools nimbus_extras %}
 {% load django_bootstrap5 %}
 {% load widget_tweaks %}
 {% load error_helpers %}
 
 {% block title %}{{ experiment.name }} - Branches{% endblock %}
 {% block main_content %}
+  {% devtools_metadata experiment %}
   <div class="position-relative">
     <!-- Full form overlay -->
     <div id="htmx-loading-overlay"
@@ -144,7 +145,19 @@
                             <div class="form-text">{{ NimbusUIConstants.COENROLLMENT_NOTE }}</div>
                           {% endif %}
                         </label>
-                        <div class="feature-value-editor">{{ branch_feature_values_form.value|add_error_class:"is-invalid" }}</div>
+                        <div class="feature-value-editor position-relative"
+                             data-nimbus-devtools-feature-editor
+                             data-feature-id="{{ branch_feature_values_form.instance.feature_config.slug }}">
+                          {{ branch_feature_values_form.value|add_error_class:"is-invalid" }}
+                          <div class="position-absolute top-0 end-0 m-1 py-1 px-1">
+                            <button class="btn btn-sm btn-outline-secondary ms-1 d-none"
+                                    type="button"
+                                    title="Preview with about:messagepreview"
+                                    data-nimbus-devtools-try-preview-button>
+                              <span class="fa fa-eye"></span>
+                            </button>
+                          </div>
+                        </div>
                         {% for error in branch_feature_values_form.value.errors %}
                           <div class="invalid-feedback d-block">{{ error }}</div>
                         {% endfor %}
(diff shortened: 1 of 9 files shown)


experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/edit_branches.html
```diff
@@ -144,7 +145,13 @@ <h4>Branches</h4>
                             <div class="form-text">{{ NimbusUIConstants.COENROLLMENT_NOTE }}</div>
                           {% endif %}
                         </label>
-                        <div class="feature-value-editor">{{ branch_feature_values_form.value|add_error_class:"is-invalid" }}</div>
+                        <div class="feature-value-editor position-relative"
+                             data-nimbus-devtools-feature-editor
+                             data-feature-id="{{ branch_feature_values_form.instance.feature_config.slug }}">
+                          {{ branch_feature_values_form.value|add_error_class:"is-invalid" }}
+                          <div class="position-absolute top-0 end-0 m-1 py-1 px-1"
+                               data-nimbus-devtools-json-controls></div>
```
jaredlockhart: Oh so you did put a little anchor point in.  Then yeah theoretically could put the whole button here.

experimenter/experimenter/nimbus_ui/templatetags/nimbus_devtools.py
```diff
@@ -0,0 +1,34 @@
...
+def devtools_metadata(experiment):
+    """Embed metadata about a given experiment in a page for nimbus-devtools to
+    consume.
+    """
+    if experiment.status == NimbusExperiment.Status.DRAFT:
+        recipe = NimbusExperimentSerializer(experiment).data
+        recipe_is_valid = NimbusReviewSerializer(
+            experiment, data=NimbusReviewSerializer(experiment).data
+        ).is_valid()
+
```
jaredlockhart: This should already be in both teh branches and summary page, since it's used to do the prelaunch checks and populate the sidebar with the 'missing details' etc sections.  It's costly to compute, adding it a second time here is probably not the best, I think we can either find the one that's already being computed in here

https://github.com/mozilla/experimenter/blob/8dbbd3e1e6a4b8ffeed11cb8ae03ef06270ac916/experimenter/experimenter/nimbus_ui/views.py#L114

So you could return the serializer right there and pass it into the context and then pass it along to the addon.

jaredlockhart: Actually same thing since we're already serializing the recipe as well

https://github.com/mozilla/experimenter/blob/8dbbd3e1e6a4b8ffeed11cb8ae03ef06270ac916/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail_card_audience.html#L141

could move that up into the context builder so we do it once and consume it in both places.

It's just costly to do all the nested fk lookups and json serialization, we should try to avoid doing it twice if we can.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/edit_branches.html
freshstrangemusic: We /could/ but we have to create the event listeners and muck with the dom from the extension anyway so :/

freshstrangemusic: Actually, on second thought, I think I agree with this. We get to keep the templates up-to-date instead of having to re-create them inside devtools. The code ends up being simpler and smaller, too.

experimenter/experimenter/nimbus_ui/templatetags/nimbus_devtools.py
freshstrangemusic: We actually don't need the recipe JSON for this patch so im removing that bit. All we need right now is the application -- the other stuff is for integrating preview

freshstrangemusic:
> Oh yeah much nicer, and then will the devtools addon change to look to find the recipe elsewhere?

we can probably just grab it from the readonly textarea in the summary page.

PR #14907, opened 2026-03-13
chore(nimbus): remove show_analysis flag from results data

### LINKED ISSUE #7375: Remove or rename show_analysis flag from experiment results data
The `show_analysis` flag was originally used for local development to prevent accidentally breaking other people's workflows. It should no longer be needed, so we should try removing it. If there is some dependency on this flag that would break if it is removed, we should at least rename the flag to give more context on its use. The current name seems to imply that it is a generated field based on whether analysis is ready to be shown, but if we need to keep the flag for some reason it should be named something to indicate that it is static and/or that it is for local dev environments.

moibra05:
Because

- The `show_analysis` is no longer useful to keep around

This commit

- Removes `show_analysis` from the results data

Fixes #7375

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -3,8 +3,8 @@ from collections import defaultdict
 from datetime import date, datetime, timedelta
 from itertools import chain
 from pathlib import Path
+from typing import Any
 
-from django.conf import settings
 from django.core.files.storage import storages
 from django.utils import timezone
 from mozilla_nimbus_schemas.jetstream import (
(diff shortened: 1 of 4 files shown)


experimenter/experimenter/jetstream/client.py
```diff
@@ -222,8 +222,7 @@ def get_experiment_data(experiment: NimbusExperiment):
     except RuntimeError as e:
         runtime_errors.append(str(e))
 
-    experiment_data = {
-        "show_analysis": settings.FEATURE_ANALYSIS,
```
mikewilli, line 226: Shouldn't we remove the setting itself too?

PR #14927, opened 2026-03-16
feat(nimbus): Handle emojis added reaction

### LINKED ISSUE #14921: enrollment ending notification failed: The request to the Slack API failed. (url: https://slack.com/api/reactions.add)
Sentry Issue: [EXPERIMENTER-PROD-198](https://mozilla.sentry.io/issues/7338917042/?referrer=github_integration)

```
enrollment ending notification failed: The request to the Slack API failed. (url: https://slack.com/api/reactions.add)
The server responded with: {'ok': False, 'error': 'already_reacted'}
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6615)

yashikakhurana:
### Because

In production, the enrollment ending notification was failing with a Slack API error already has that reaction.

The root cause is that `send_threaded_success_message()` is being called multiple times for the same enrollment ending request - either due to Kinto syncs running multiple times, task retries, or periodic checks. Each time it's called, it attempts to add the white_check_mark reaction to the original message. On subsequent calls, the reaction already exists, causing the API to reject the request with `already_reacted`.

Previously, any Slack API error (including `already_reacted`) was treated as a fatal failure, logging the error and returning False. This caused the entire notification flow to fail, even though the actual desired state (the message exists and the reaction is on it) was already achieved.

### This commit

Adds graceful error handling for the `already_reacted` Slack error. When `reactions_add` fails with `already_reacted`, the error is caught and ignored since the desired end state is already met - the reaction exists on the message. Other Slack API errors are still re-raised and properly logged as failures.

Additionally, this commit adds the underlying support for enrollment ending and experiment ending request notifications with dedicated alert types (`END_ENROLLMENT_REQUEST` and `END_EXPERIMENT_REQUEST`), consolidates duplicate Slack notification logic, and adds comprehensive tests to verify the fix works correctly.

Fixes #14921

### DIFF
--- a/experimenter/experimenter/slack/notification.py
+++ b/experimenter/experimenter/slack/notification.py
@@ -200,7 +200,10 @@ def send_threaded_success_message(
                 timestamp=thread_ts,
             )
         except SlackApiError as emoji_error:
-            if emoji_error.response.get("error") != "already_reacted":
+            if (
+                emoji_error.response.get("error")
+                != SlackConstants.ErrorCode.ALREADY_REACTED
+            ):
                 raise
 
         logger.info(success_log_message(experiment.slug))
--- a/experimenter/experimenter/slack/constants.py
+++ b/experimenter/experimenter/slack/constants.py
@@ -49,6 +49,10 @@ class SlackConstants:
         CANCEL = "x"
         APPROVE = "eyes"
 
+    # Slack API error codes
+    class ErrorCode:
+        ALREADY_REACTED = "already_reacted"
+
     # Slack message templates
     SLACK_DM_PREFIX = "🔔 Join {channel} to get slack notifications: {message}"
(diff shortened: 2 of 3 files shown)


experimenter/experimenter/slack/notification.py
```diff
@@ -193,11 +193,15 @@ def send_threaded_success_message(
...
-            timestamp=thread_ts,
-        )
+        try:
+            client.reactions_add(
+                channel=channel_id,
+                name="white_check_mark",
+                timestamp=thread_ts,
+            )
+        except SlackApiError as emoji_error:
+            if emoji_error.response.get("error") != "already_reacted":
```
jaredlockhart: Probably want an enum of slack error constants ya?

yashikakhurana: good point, done

PR #14958, opened 2026-03-18
feat(nimbus): Add opt-in integration with nimbus-devtools

### LINKED ISSUE #13330: Nimbus Devtools opt-in Integration (Experimenter)
We should support integration with Nimbus Devtools for opt-ins

See-also: https://github.com/mozilla-extensions/nimbus-devtools/issues/63

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5855)

freshstrangemusic:
Because:

- nimbus-devtools needs DOM controls to add opt-in support; and
- the pretty-formatted recipe JSON is not actually valid JSON

this commit:

- adds those elements; and
- removes the pretty formatting for recipe JSON.

Fixes #13330

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -2185,14 +2185,10 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
     def recipe_json(self):
         from experimenter.experiments.api.v6.serializers import NimbusExperimentSerializer
 
-        return (
-            json.dumps(
-                self.published_dto or NimbusExperimentSerializer(self).data,
-                indent=2,
-                sort_keys=True,
-            )
-            .replace("&&", "\n&&")  # Add helpful newlines to targeting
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -2185,14 +2185,10 @@ def conclusion_recommendation_labels(self):
...
-                indent=2,
-                sort_keys=True,
-            )
-            .replace("&&", "\n&&")  # Add helpful newlines to targeting
-            .replace("\\n", "\n")  # Handle hard coded newlines in targeting
+        return json.dumps(
+            self.published_dto or NimbusExperimentSerializer(self).data,
+            indent=2,
+            sort_keys=True,
         )
```
jaredlockhart, line 2192: Oh definitely yes, you're right those break the JSON validity altogether and probably we don't need that at all now since we have the nice formatting on the jexl only field for readability 👍

freshstrangemusic:
image

PR #14963, opened 2026-03-18
chore(project): add DOCKER_USER override for dev containers

### LINKED ISSUE #14962: Support configurable container user for dev containers (Linux/WSL2 file permissions)
## Problem

The dev containers in `docker-compose.yml` all run as root. On macOS this is fine because Docker Desktop's VM transparently handles UID translation for bind mounts. However, on Linux and Windows/WSL2, files created by root inside the container are owned by root on the host filesystem, causing permission issues for developers.

## Proposed Solution

Add a `DOCKER_USER` variable to the dev services in `docker-compose.yml` that defaults to empty (preserving current root behavior for macOS and CI), but can be set in `.env` by Linux/WSL2 developers:

```yaml
services:
  experimenter:
    user: "${DOCKER_USER:-}"
```

```env
# .env (Linux/WSL2 only)
DOCKER_USER=1000:1000
```

This is a zero-impact change for macOS developers and CI — an empty `user:` field is ignored by Docker Compose.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6638)

jaredlockhart:
Because

* Dev containers run as root, which works on macOS (Docker Desktop handles UID translation) but causes file permission issues on Linux and Windows/WSL2 where bind-mounted files end up owned by root
* Developers on those platforms need a way to match container UID/GID to their host user

This commit

* Adds `user: "${EXPERIMENTER_DOCKER_USER:-root}"` to dev services in `docker-compose.yml` (experimenter, yarn-nimbus_ui, worker, beat)
* Defaults to `root`, preserving current behavior for macOS developers and CI
* Linux/WSL2 developers set `EXPERIMENTER_DOCKER_USER=1000:1000` in their `.env` (replace with actual `$(id -u):$(id -g)`)

Fixes #14962

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,9 +73,6 @@ target/
 .envrc
 .psql_history
 
-# Local Docker Compose overrides (e.g., user: for Linux/WSL2 file permissions)
-docker-compose.override.yml
-
 # VS Code
 **/.vscode/*
 **/.devcontainer/*
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
   experimenter:
     image: experimenter:dev
     env_file: .env
+    user: "${EXPERIMENTER_DOCKER_USER:-root}"
     stdin_open: true
     tty: true
     depends_on:
@@ -30,6 +31,7 @@ services:
   yarn-nimbus_ui:
     image: experimenter:dev
     env_file: .env
+    user: "${EXPERIMENTER_DOCKER_USER:-root}"
     stdin_open: true
     tty: true
     volumes:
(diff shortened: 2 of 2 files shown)


relud:
this might also need to be set as the user id at image build time so that the user actually exists inside the container

jaredlockhart:
> this might also need to be set as the user id at image build time so that the user actually exists inside the container

Yeah I was wondering the same.

jaredlockhart:
Closing for now — the runtime `user:` override doesn't work when the UID doesn't exist in the container's /etc/passwd. A proper fix would need `adduser` calls at build or entrypoint time. We'll revisit.

PR #14978, opened 2026-03-19
fix(nimbus): check daily window in has_exposures for live experiments

### LINKED ISSUE #14976: Results page: exposure basis dropdown disabled for live experiments with only daily data
## Problem

The `has_exposures` property on `NimbusExperiment` only checks the `overall` and `weekly` result windows for exposure data. For live experiments that haven't completed a full analysis week yet, Jetstream only produces `daily` results — so the exposure basis dropdown on the results page is disabled even though exposure-basis data has been computed and ingested.

This means experiment owners can't view exposure-basis results until the experiment has been running long enough for weekly or overall windows to appear, even though Jetstream is computing exposure-basis daily statistics from day one.

## Fix

Add `"daily"` to the window list in the `has_exposures` property check:

```python
for window in ["overall", "weekly", "daily"]:
```

## How to test

1. Find a live experiment that has daily results but no weekly/overall results yet
2. Verify the "Exposures" option in the Analysis Basis dropdown is no longer disabled

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6649)

jaredlockhart:
Because

* The `has_exposures` property only checked `overall` and `weekly` result
  windows for exposure data
* For live experiments that haven't completed a full analysis week,
  Jetstream only produces daily results
* This caused the exposure basis dropdown on the results page to be disabled
  even though exposure-basis data was being computed and ingested by Jetstream

This commit

* Adds `daily` to the window list checked by `has_exposures`
* Adds a test for the daily-only exposure data case

Fixes #14976

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1718,7 +1718,7 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
 
         if self.results_data and "v3" in self.results_data:
             results_data = self.results_data["v3"]
-            for window in ["overall", "weekly"]:
+            for window in ["overall", "weekly", "daily"]:
                 if results_data.get(window):
                     exposures_branch_data = (
                         results_data[window].get("exposures", {}).get("all", {})
--- a/experimenter/experimenter/experiments/tests/test_models.py
+++ b/experimenter/experimenter/experiments/tests/test_models.py
@@ -3467,6 +3467,51 @@ class TestNimbusExperiment(TestCase):
             experiment.has_exposures, NimbusUIConstants.ExposuresStatus.NO_EXPOSURES
         )
 
+    def test_has_exposures_daily_only(self):
+        results_data = {
+            "v3": {
+                "daily": {
+                    "exposures": {
+                        "all": {
+                            "control": {
+                                "branch_data": {
+                                    "other_metrics": {
+                                        "identity": {
+                                            "absolute": {
+                                                "first": {
+                                                    "point": 120,
+                                                },
+                                            }
+                                        }
+                                    }
+                                }
+                            },
+                            "treatment": {
+                                "branch_data": {
+                                    "other_metrics": {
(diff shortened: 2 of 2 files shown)


experimenter/experimenter/experiments/tests/test_models.py
```diff
@@ -3467,6 +3467,51 @@ def test_has_exposures_results_false(self):
...
+                                    "other_metrics": {
+                                        "identity": {
+                                            "absolute": {
+                                                "first": {
+                                                    "point": 120,
+                                                },
+                                            }
+                                        }
+                                    }
+                                }
```
mikewilli, line 3487: Not related to this PR (and so not something to solve here), but seems like we could pull some of these things out so they're reusable. Like the selected portion is just repeated a bunch of places basically verbatim. The old UI tests had a file where some common things were defined and reused across various tests.

experimenter/experimenter/experiments/models.py
```diff
@@ -1718,7 +1718,7 @@ def has_exposures(self):
 
         if self.results_data and "v3" in self.results_data:
             results_data = self.results_data["v3"]
-            for window in ["overall", "weekly"]:
+            for window in ["overall", "weekly", "daily"]:
```
mikewilli, line 1721: Maybe we should have an `ALL_WINDOWS` constant for this? Or dynamically iterate the results data?

experimenter/experimenter/experiments/tests/test_models.py
jaredlockhart: Oh I know I absolutely want reusable data fixtures for these, definitely something we should do soon 🙏

experimenter/experimenter/experiments/models.py
jaredlockhart: Ya not a bad idea, I think there's a lot of room to clean up how we handle this data.  I'd say let's leave this like this for now and we can do some more rigorous cleaning soon.

PR #14999, opened 2026-03-20
fix(nimbus): capture branch feature value changes in history changelog

### LINKED ISSUE #14998: History page does not capture branch feature value changes
## Summary

When branch feature values are edited and saved via the Branches page, the experiment's changelog/history page does not capture the feature value changes. The changelog shows that branches were modified, but the recorded snapshot contains the **old** feature values instead of the updated ones.

## Steps to Reproduce

1. Create an experiment with branches and feature configs
2. Set initial feature values on the branches and save
3. Change the feature values and save again
4. Navigate to the History page

## Expected Behavior

The History page should show the feature value changes between the two saves, with the old and new feature values displayed in the diff.

## Actual Behavior

The changelog entry for the second save contains the **old** feature values in its snapshot. The feature value change is invisible in the History page (or appears delayed until the next save).

## Root Cause

In `NimbusBranchesForm.save()`, the MRO causes `NimbusChangeLogFormMixin.save()` to generate the changelog snapshot **before** `self.branches.save()` executes. The call chain is:

1. `NimbusBranchesForm.save()` → `super().save()` →
2. `NimbusChangeLogFormMixin.save()` → `super().save()` (ModelForm) → `generate_nimbus_changelog()` (snapshot taken here) → returns
3. Back in `NimbusBranchesForm.save()` → `self.branches.save()` (feature values updated here, too late)

The changelog snapshot is taken at step 2, but the feature values aren't saved until step 3.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6665)

jaredlockhart:
Because

* `NimbusBranchesForm.save()` called `super().save()` which triggered `NimbusChangeLogFormMixin.save()` to generate the changelog snapshot before `self.branches.save()` ran, causing the changelog to capture stale feature values instead of the updated ones
* Feature value changes were invisible on the History page

This commit

* Bypasses the mixin's `save()` in `NimbusBranchesForm` to call `ModelForm.save()` directly, then generates the changelog after all branch and feature value saves are complete
* Adds a test that saves an experiment with initial feature values, updates the values, and asserts the changelog captures the change

Fixes #14998

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -829,8 +829,8 @@ class NimbusBranchesForm(NimbusChangeLogFormMixin, forms.ModelForm):
 
     @transaction.atomic
     def save(self, *args, **kwargs):
-        experiment = super().save(*args, **kwargs)
         self.branches.save()
+        experiment = super().save(*args, **kwargs)
 
         if experiment.is_rollout:
             branches = experiment.branches.all()
--- a/experimenter/experimenter/nimbus_ui/tests/test_forms.py
+++ b/experimenter/experimenter/nimbus_ui/tests/test_forms.py
@@ -3226,6 +3226,18 @@ class TestNimbusBranchesForm(RequestFormTestCase):
 
         changelog = experiment.changes.get()
         self.assertIn("updated branches", changelog.message)
+        changelog_branches = changelog.experiment_data["branches"]
+        changelog_fv_values = [
+            fv["value"] for b in changelog_branches for fv in b["feature_values"]
+        ]
+        self.assertIn(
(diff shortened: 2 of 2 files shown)


experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -829,8 +829,8 @@ def clean(self):
 
     @transaction.atomic
     def save(self, *args, **kwargs):
-        experiment = super().save(*args, **kwargs)
         self.branches.save()
+        experiment = super().save(*args, **kwargs)
```
yashikakhurana, line 833: that's even more cleaner

jaredlockhart: Only the cleanest 🙏 🧹

PR #15035, opened 2026-03-25
feat(nimbus): Add enrollment alert data GCS reader and Celery fetch task

### LINKED ISSUE #15024: Add enrollment alert data GCS reader and Celery fetch task
### What to do
Fetch enrollment data from GCS and update experiments daily at 8am.

**Location:** `experimenter/experimenter/jetstream/`

### Deliverables
* [ ] Add to `client.py`:
```python
ENROLLMENT_ALERTS_FOLDER = "enrollment_alerts"

def get_enrollment_alert_data():
    """Fetch from GCS path: enrollment_alerts/enrollment_alerts_latest.json"""
```

* [ ] Add to `tasks.py`:
```python
@app.task
@metrics.timer_decorator("fetch_enrollment_alert_data")
def fetch_enrollment_alert_data():
    """Fetch data, update all experiments' enrollment_alert_data field"""
```

* [ ] Add to `settings.py` (CELERY_BEAT_SCHEDULE):
```python
"fetch_enrollment_alert_data": {
    "task": "experimenter.jetstream.tasks.fetch_enrollment_alert_data",
    "schedule": crontab(minute=0, hour=8),  # 8am UTC daily
}
```

* [ ] Add metrics: started, completed, failed
* [ ] Add logging and error handling

### Acceptance Criteria
* [ ] Reader fetches from GCS successfully
* [ ] Task updates experiments' enrollment_alert_data field
* [ ] Scheduled for 8am UTC daily (after 3am ETL)
* [ ] Metrics increment correctly
* [ ] Error handling in place

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6602)

yashikakhurana:
Because

- Now we have the enrollment count JSON available that we can use to set the monitoring data for each experiment, which can be used further to alert users for the thresholds and SRM.

This commit

- Fetches the JSON data, parses and set it to the experiment `monitoring_data` field

Fixes #15024 #15036

### DIFF
--- a/experimenter/experimenter/jetstream/tests/test_tasks.py
+++ b/experimenter/experimenter/jetstream/tests/test_tasks.py
@@ -3362,8 +3362,8 @@ class TestFetchJetstreamDataTask(MockSizingDataMixin, TestCase):
 
 
 @pytest.fixture
-def mock_monitoring_data():
-    return {
+def mock_monitoring_data(request):
+    data = {
         "total_enrollments": 15000,
         "total_unenrollments": 1200,
         "branches": {
@@ -3390,12 +3390,15 @@ def mock_monitoring_data():
             },
         },
     }
+    if request.instance:
+        request.instance.monitoring_data = data
+    return data
 
 
+@pytest.mark.usefixtures("mock_monitoring_data")
 class TestFetchMonitoringDataTask(TestCase):
     def setUp(self):
         super().setUp()
-        self.monitoring_data = mock_monitoring_data()
         patcher = patch("experimenter.jetstream.tasks.get_monitoring_data")
         self.mock_get_monitoring_data = patcher.start()
         self.addCleanup(patcher.stop)


experimenter/experimenter/settings.py
```diff
@@ -417,6 +417,11 @@
         "task": "experimenter.slack.tasks.check_experiment_alerts",
         "schedule": crontab(minute=0, hour=18),
     },
+    "fetch_monitoring_data": {
+        "task": "experimenter.jetstream.tasks.fetch_monitoring_data",
+        "schedule": crontab(minute=0, hour=8),
+        "options": {"expires": 3600},
```
mikewilli: What's this expiration option do?

experimenter/experimenter/settings.py
```diff
@@ -417,6 +417,11 @@
         "task": "experimenter.slack.tasks.check_experiment_alerts",
         "schedule": crontab(minute=0, hour=18),
     },
+    "fetch_monitoring_data": {
+        "task": "experimenter.jetstream.tasks.fetch_monitoring_data",
+        "schedule": crontab(minute=0, hour=8),
```
mikewilli, line 422: This is fine, but be aware that there will be times when the data isn't available by 8 (esp if this is UTC, which I think it is?). It might be nice to have a way to trigger this manually in those cases, but I think it's ok to worry about it later if it happens enough to be worth fixing.

experimenter/experimenter/slack/constants.py
mikewilli, line 1: It doesn't look like we're actually using these constants yet, can you wait to include them on the PR where they are actually needed?

experimenter/experimenter/jetstream/client.py
```diff
@@ -95,6 +99,17 @@ def get_sizing_data(suffix="latest"):
...
 
+def get_monitoring_data():
+    try:
+        filename = "enrollment_counts_latest.json"
+        path = Path(ENROLLMENT_COUNTS_FOLDER, filename)
+        data = load_data_from_gcs(str(path))
+        return data if data else {}
+    except Exception as e:
+        logger.error(f"Failed to load monitoring data from GCS: {e}")
+        return {}
```
mikewilli: I think we should just raise this exception and let the task handle it along with any logging. That fits with what we're doing in the other tasks, and it also doesn't seem useful to overwrite yesterday's data with blank data if there was an error getting the new data.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+@app.task
+@metrics.timer_decorator("fetch_monitoring_data")
+def fetch_monitoring_data():
+    metrics.incr("fetch_monitoring_data.started")
+    try:
+        data = get_monitoring_data()
+
+        if not data or "v1" not in data:
+            logger.warning("No enrollment alert data found in GCS")
+            metrics.incr("fetch_monitoring_data.completed")
```
mikewilli: Should this be an error log and a `.failed` metric status?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+                experiment = NimbusExperiment.objects.get(
+                    slug=exp_slug,
+                    status=NimbusConstants.Status.LIVE,
+                )
+
+                # Only update if data has changed
+                if experiment.monitoring_data != monitoring_data:
+                    experiment.monitoring_data = monitoring_data
+                    experiment.save(update_fields=["monitoring_data"])
+                    updated_count += 1
```
mikewilli, line 167: Should we also generate a changelog for this like we do for the `results_data`?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+    metrics.incr("fetch_monitoring_data.started")
+    try:
+        data = get_monitoring_data()
+
+        if not data or "v1" not in data:
+            logger.warning("No enrollment alert data found in GCS")
+            metrics.incr("fetch_monitoring_data.completed")
+            return
+
+        alert_data = data["v1"]
```
mikewilli:
```suggestion
        alert_data = data.get("v1")
```

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+                continue
+
+        logger.info(
+            f"Successfully updated monitoring data for {updated_count} experiments"
+        )
+        metrics.incr("fetch_monitoring_data.completed")
+
+    except Exception as e:
+        metrics.incr("fetch_monitoring_data.failed")
+        logger.error(f"Fatal error in fetch_monitoring_data task: {e}")
```
mikewilli:
```suggestion
        logger.exception(f"Fatal error in fetch_monitoring_data task: {e}")
```

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+                )
+
+                # Only update if data has changed
+                if experiment.monitoring_data != monitoring_data:
+                    experiment.monitoring_data = monitoring_data
+                    experiment.save(update_fields=["monitoring_data"])
+                    updated_count += 1
+
+            except NimbusExperiment.DoesNotExist:
+                logger.debug(f"Experiment {exp_slug} not found in database")
```
mikewilli: Do we keep debug level logs or is this just for local debugging?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+            try:
+                experiment = NimbusExperiment.objects.get(
+                    slug=exp_slug,
+                    status=NimbusConstants.Status.LIVE,
+                )
+
+                # Only update if data has changed
+                if experiment.monitoring_data != monitoring_data:
+                    experiment.monitoring_data = monitoring_data
+                    experiment.save(update_fields=["monitoring_data"])
```
mikewilli: Does `update_fields` prevent downstream properties from also updating?

experimenter/experimenter/jetstream/client.py
```diff
@@ -95,6 +99,17 @@ def get_sizing_data(suffix="latest"):
     return load_data_from_gcs(str(path))
 
 
+def get_monitoring_data():
+    try:
+        filename = "enrollment_counts_latest.json"
+        path = Path(ENROLLMENT_COUNTS_FOLDER, filename)
+        data = load_data_from_gcs(str(path))
```
mikewilli: Eventually it'd be nice if we could add a schema and validation for this data, like we have for the jetstream output.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,185 @@ def open_file(filename):
         mock_exists.return_value = True
         with self.assertRaises(Exception):
             tasks.fetch_population_sizing_data()
+
+
+class TestFetchMonitoringDataTask(TestCase):
+    def setUp(self):
+        super().setUp()
+        self.monitoring_data = {
```
mikewilli: I guess this works, but this looks more like it should be a fixture to me than a class var.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,185 @@ def open_file(filename):
...
+
+        mock_get_data.assert_called_once()
+
+    @patch("experimenter.jetstream.tasks.get_monitoring_data")
+    def test_fetch_monitoring_data_empty_v1(self, mock_get_data):
+        mock_get_data.return_value = {}
+
+        tasks.fetch_monitoring_data()
+
+        mock_get_data.assert_called_once()
```
mikewilli: Is it really useful to have both of these? Maybe we can combine them and parametrize the `return_value` to pass both options in?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+        for exp_slug, monitoring_data in alert_data.items():
+            try:
+                experiment = NimbusExperiment.objects.get(
+                    slug=exp_slug,
+                    status=NimbusConstants.Status.LIVE,
+                )
+
+                # Only update if data has changed
+                if experiment.monitoring_data != monitoring_data:
+                    experiment.monitoring_data = monitoring_data
```
mikewilli, line 160: It'd be nice to include a timestamp in the monitoring data as well (`"last_updated": datetime.now()`). But now that I think about it, it's probably better to add that on the ETL side... What do you think?

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,185 @@ def open_file(filename):
...
+
+    @patch("experimenter.jetstream.client.load_data_from_gcs")
+    def test_get_monitoring_data_no_data(self, mock_load):
+        mock_load.return_value = None
+
+        result = get_monitoring_data()
+
+        self.assertEqual(result, {})
+
+    @patch("experimenter.jetstream.client.load_data_from_gcs")
```
mikewilli, line 3556: I _think_ you can patch these for everything without annotating every function by creating a fixture like [this](https://github.com/mozilla/jetstream/blob/main/jetstream/tests/conftest.py#L293-L301), and having it take a parameter to set the return value dynamically.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,185 @@ def open_file(filename):
...
+    def test_get_monitoring_data_exception(self, mock_load):
+        mock_load.side_effect = Exception("GCS error")
+
+        result = get_monitoring_data()
+
+        self.assertEqual(result, {})
```
mikewilli: Yea see my other comment on this, but I really think this should result in an exception from `get_monitoring_data` that is handled by the task.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,185 @@ def open_file(filename):
...
+
+        with (
+            patch.object(experiment, "save", side_effect=Exception("DB error")),
+            patch(
+                "experimenter.jetstream.tasks.NimbusExperiment.objects.get"
+            ) as mock_get,
+        ):
+            mock_get.return_value = experiment
+            # Should not raise, should log and continue
+            tasks.fetch_monitoring_data()
```
mikewilli, line 3493: I'm not totally sure this test is necessary, but I guess it ensures that a random exception doesn't break the task? 

Can we test for something here, like that the status or log occurs?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -124,3 +128,49 @@ def fetch_population_sizing_data():
...
+            return
+
+        alert_data = data["v1"]
+        updated_count = 0
+
+        for exp_slug, monitoring_data in alert_data.items():
+            try:
+                experiment = NimbusExperiment.objects.get(
+                    slug=exp_slug,
+                    status=NimbusConstants.Status.LIVE,
```
mikewilli: This should work for COMPLETE also, no?

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,185 @@ def open_file(filename):
...
+        mock_get_data.return_value = {}
+
+        tasks.fetch_monitoring_data()
+
+        mock_get_data.assert_called_once()
+
+    @patch("experimenter.jetstream.tasks.get_monitoring_data")
+    def test_fetch_monitoring_data_updates_live_experiment(self, mock_get_data):
+        experiment = NimbusExperimentFactory.create(
+            status=NimbusExperiment.Status.LIVE,
```
mikewilli: Maybe parametrize this so it takes both LIVE and COMPLETE.

experimenter/experimenter/settings.py
yashikakhurana: timeout is 1 hour for this request

experimenter/experimenter/settings.py
yashikakhurana: makes sense, let me note down it somewhere so that we don't miss it

experimenter/experimenter/slack/constants.py
yashikakhurana: okay I can remove these, just added as was prepping for the next task

experimenter/experimenter/settings.py
mikewilli: Is this useful here? We don't use it for any of the other tasks, and I can't really tell from the [Celery docs](https://docs.celeryq.dev/en/main/userguide/calling.html#expiration) what scenario this parameter is trying to handle.

If we do want this, I think a more appropriate timeout would be ~5min. Were you seeing really long execution times in testing this? Or is there another reason this should be 1hr?

yashikakhurana: adding as a precautionary step so that it doesn't hang, if no other tasks uses it I can take that out, since others are working similarly

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: no but I have changed the pattern now

experimenter/experimenter/jetstream/client.py
yashikakhurana: that's true

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: yes good point

yashikakhurana:
Thank you for the review @mikewilli , hopefully I am not missing any changes 🤞 :D

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,201 @@ def open_file(filename):
         mock_exists.return_value = True
         with self.assertRaises(Exception):
             tasks.fetch_population_sizing_data()
+
+
+MONITORING_DATA_FIXTURE = {
```
mikewilli: What I meant by fixture is this:

```suggestion
@pytest.fixture
def mock_monitoring_data():
    return {
```

Then you can take it as a parameter in the tests that need it.

experimenter/experimenter/jetstream/tests/test_tasks.py
```diff
@@ -3358,3 +3358,201 @@ def open_file(filename):
...
+        self.mock_get_monitoring_data.return_value = return_value
+
+        tasks.fetch_monitoring_data()
+
+        self.mock_get_monitoring_data.assert_called_once()
+
+    @parameterized.expand(
+        [
+            ("live", NimbusExperiment.Status.LIVE),
+            ("complete", NimbusExperiment.Status.COMPLETE),
```
mikewilli: Why do these all include an unused `_name` parameter?

PR #15087, opened 2026-03-30
feat(nimbus): create targeting context fields parser

### LINKED ISSUE #14923: Parse targeting context source files to extract field names
Write a parser class that extracts valid targeting context field names from each platform's stored source file, with version awareness.

Following the pattern of experimenter/outcomes/__init__.py (lazy-loading class with caching), create a class in experimenter/targeting/ that:

* Reads each platform's targeting context file from the local manifests path (configured in settings.py, pointing into features/manifests/<app>/)
* Extracts field names using regex or simple string parsing:
* Desktop (JS): Keys from the ATTRIBUTE_TRANSFORMS object (camelCase: activeExperiments, addressesSaved, etc.)
* Android (Kotlin): Field names from RecordedNimbusContext (snake_case: app_version, is_first_run, etc.)
* iOS (Swift): Dictionary keys from the context dictionary (snake_case: is_first_run, region, etc.)
* Supports versioned lookup: TargetingContextFields.for_application(app, version=None) — defaults to latest (unversioned) if no version specified
* Caches results like Outcomes._outcomes

Acceptance Criteria:

* A parser class exists that extracts field names for each of the three platforms
* Field lists are accessible per application, optionally per version
* Parsers are tested with the current upstream file contents
* Parsers handle minor formatting changes (whitespace, comments) gracefully

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6618)

moibra05:
Because

- We needed a single, version-aware way to extract valid targeting context fields from each platform's source code

This commit

- Adds a dedicated targeting context parser class that follows the same lazy-loading cache pattern used by `Outcomes`, including per-app cache lifecycle and explicit cache clearing
- Introduces version-aware lookup via for_application(app, version=None) so callers can resolve fields from default versioned targeting field files or a specific versioned manifest path using the same method

Fixes #14923

### DIFF
--- a/experimenter/experimenter/targeting/targeting_context_parser.py
+++ b/experimenter/experimenter/targeting/targeting_context_parser.py
@@ -1,0 +1,129 @@
+import re
+
+from django.conf import settings
+
+from experimenter.experiments.constants import NimbusConstants
+
+
+class TargetingContextFields:
+    _desktop_targeting_fields = None
+    _fenix_targeting_fields = None
+    _ios_targeting_fields = None
+
+    _CACHE_FIELDS = {
+        NimbusConstants.Application.DESKTOP: "_desktop_targeting_fields",
+        NimbusConstants.Application.FENIX: "_fenix_targeting_fields",
+        NimbusConstants.Application.IOS: "_ios_targeting_fields",
+    }
+
+    @staticmethod
+    def _parse_desktop_targeting_fields(targeting_context_code):
+        found_block = False
+        targeting_fields = []
+        # This regex is meant to only match lines on the primary level of the object
+        # literal, so it looks for lines that start with two spaces followed by a letter.
+        # This is to avoid matching nested lines.
+        target_line_match = re.compile(r"^  [a-zA-Z]")
+
+        for line in targeting_context_code.splitlines():
+            stripped_line = line.strip()
+
+            if stripped_line == "export const ATTRIBUTE_TRANSFORMS = Object.freeze({":
+                found_block = True
+            elif found_block and stripped_line == "});":
+                break
+            elif found_block:
+                if target_line_match.match(line):
+                    targeting_fields.append(stripped_line.split(":")[0].strip())
+
+        return targeting_fields
+
+    @staticmethod
+    def _parse_fenix_targeting_fields(targeting_context_code):
+        found_block = False
+        targeting_fields = []
+        target_line_match = re.compile(r'"([^"]+)"')
+
+        for line in targeting_context_code.splitlines():
+            stripped_line = line.strip()
+
+            if stripped_line == "val obj = JSONObject(":
+                found_block = True
+            elif found_block and stripped_line == "),":
+                break
+            elif found_block:
+                res = target_line_match.search(stripped_line)
+                if res:
+                    targeting_fields.append(res.group(1))
+
+        return targeting_fields
+
+    @staticmethod
+    def _parse_ios_targeting_fields(targeting_context_code):
+        found_block = False
+        targeting_fields = []
+        target_line_match = re.compile(r'"([^"]+)"')
+
+        for line in targeting_context_code.splitlines():
+            stripped_line = line.strip()
+
+            if (
+                stripped_line
+                == "guard let data = try? JSONSerialization.data(withJSONObject: ["
+            ):
+                found_block = True
+            elif found_block and stripped_line == "]),":
+                break
+            elif found_block:
+                res = target_line_match.search(stripped_line)
+                if res:
+                    targeting_fields.append(res.group(1))
+
+        return targeting_fields
+
+    @classmethod
+    def _load_targeting_fields(cls, app, version=None):
+        app_path = (
+            settings.FEATURE_MANIFESTS_PATH / app / version
+            if version
+            else settings.FEATURE_MANIFESTS_PATH / app
+        )
+        targeting_context_path = (
+            app_path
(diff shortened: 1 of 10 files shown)


experimenter/experimenter/targeting/tests/mock-manifests/fenix/v100.0.0/RecordedNimbusContext.kt
```diff
@@ -0,0 +1,222 @@
...
+     * [toJson] is called by the Nimbus SDK Rust code after the event queries have been executed,
+     * and before experiment enrollments have been evolved. The value returned from this method
+     * will be applied directly to the Nimbus targeting context, and its keys/values take
+     * precedence over those in the main Nimbus targeting context.
+     *
+     * @return JSONObject
+     */
+    override fun toJson(): JsonObject {
+        val obj = JSONObject(
+            mapOf(
```
jaredlockhart, line 125: I was wrong, actually the keys we need are here in the JSONObject, not the RecordedTargetingContext below 🙏

experimenter/experimenter/targeting/tests/mock-manifests/ios/v100.0.0/RecordedNimbusContext.swift
```diff
@@ -0,0 +1,222 @@
...
+            "days_since_update": daysSinceUpdate as Any,
+            "is_default_browser": isDefaultBrowser,
+            "is_bottom_toolbar_user": isBottomToolbarUser,
+            "has_enabled_tips_notifications": hasEnabledTipsNotifications,
+            "has_accepted_terms_of_use": hasAcceptedTermsOfUse,
+            "is_apple_intelligence_available": isAppleIntelligenceAvailable,
+            "cannot_use_apple_intelligence": cannotUseAppleIntelligence,
+            "tou_experience_points": touExperiencePoints as Any
+        ]),
+            let jsonString = NSString(data: data, encoding: String.Encoding.utf8.rawValue) as? String
```
jaredlockhart, line 214: And this one for iOS

experimenter/experimenter/targeting/targeting_context_parser.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        targeting_context_path = (
+            app_path
+            / NimbusConstants.APPLICATION_CONFIGS[app].targeting_context_file_name
+        )
+
+        with targeting_context_path.open() as targeting_context_file:
+            targeting_context_code = targeting_context_file.read()
+
+            def parse_desktop_targeting_fields():
```
freshstrangemusic: Move these functions into classmethods that take the input as an argument.

experimenter/experimenter/targeting/targeting_context_parser.py
```diff
@@ -0,0 +1,124 @@
...
+            if version
+            else settings.FEATURE_MANIFESTS_PATH / app
+        )
+        targeting_context_path = (
+            app_path
+            / NimbusConstants.APPLICATION_CONFIGS[app].targeting_context_file_name
+        )
+
+        with targeting_context_path.open() as targeting_context_file:
+            targeting_context_code = targeting_context_file.read()
```
freshstrangemusic: You don't need to keep the file open for the entire duration.

Additionally, there is a helper on Path that does what you want.
```suggestion
        targeting_context_code = targeting_context_path.read_text()
```

experimenter/experimenter/targeting/targeting_context_parser.py
```diff
@@ -0,0 +1,124 @@
...
+                    ):
+                        found_block = True
+                    elif found_block and stripped_line == "});":
+                        break
+                    elif found_block:
+                        # This regex is meant to only match lines on the primary level of
+                        # the object literal, so it looks for lines that start with two
+                        # spaces followed by a letter. This is to avoid matching nested
+                        # lines.
+                        if re.match(r"^  [a-zA-Z]", line):
```
freshstrangemusic: compile these regexes. it will be much faster, especially if we are going to be processing multiple manifest files.

https://docs.python.org/3/library/re.html#re.compile

experimenter/experimenter/targeting/targeting_context_parser.py
```diff
@@ -0,0 +1,124 @@
...
+
+            def parse_desktop_targeting_fields():
+                found_block = False
+
+                for line in targeting_context_code.splitlines():
+                    stripped_line = line.strip()
+
+                    if (
+                        stripped_line
+                        == "export const ATTRIBUTE_TRANSFORMS = Object.freeze({"
```
freshstrangemusic: I would feel better about this if we had some sort of marker lines like

```
export const ATTRIBUTE_TRANSFORMS = Object.freeze({
    /* BEGIN_TARGETING_CONTEXT_ATRRIBUTES */
    "foo": bar,
    /* END_TARGETING_CONTEXT_ATTRIBTUES */
});
```

just in case the code changes so we can be durable.

moibra05: would this not require changes to each respective file for all firefox releases?

jaredlockhart: Yeah it's a good idea, we can add those and update the parser after, I wouldn't block on it though 👍

freshstrangemusic: Agreed

PR #15119, opened 2026-04-01
feat(nimbus): Send alert for unenrollments and srm ration on slack

### LINKED ISSUE #15115: Experimenter-Implement unenrollment spike and SRM mismatch alert functions
### What to do
Implement alert computation and logic in `slack/tasks.py`.

**Location:** `experimenter/experimenter/slack/tasks.py`

### Deliverables
* [ ] Add import: `from scipy.stats import chisquare`
* [ ] Add helpers:
```python
def _compute_unenrollment_rate(alert_data) → float
def _compute_srm_p_value(alert_data) → float | None  # chi-squared test
def _get_top_unenrollment_reason(alert_data) → str
```

* [ ] Add checks (with deduplication):
```python
def _check_unenrollment_spike(experiment)
    # If rate > 10% and no existing alert → send

def _check_srm_mismatch(experiment)
    # If p < 0.001 and no existing alert → send
```

* [ ] Add sends:
```python
def _send_unenrollment_spike_alert(experiment, rate)
def _send_srm_mismatch_alert(experiment, p_value)
    # Format message, call send_slack_notification(), create NimbusAlert
```

* [ ] Update `check_single_experiment_alerts()` to call both new checks
* [ ] Add metrics and logging

### Acceptance Criteria
* [ ] All 7 functions implemented
* [ ] Computation functions handle edge cases (0 enrollments, 1 branch, None)
* [ ] Check functions deduplicate (skip if NimbusAlert already exists)
* [ ] Send functions create NimbusAlert with slack_thread_id
* [ ] Wired into alert orchestration correctly
* [ ] Metrics increment on sent/failed
* [ ] Passes linting and tests

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6604)

yashikakhurana:
Because

- Experiment is fetching monitoring data and experiment field have now data available `monitoring_data` field, now we can use this data to check the unenrollments and srm ratios

This commit

- Notifies user on slack if the unenrollments is greater then 10%
- Sends alert for srm ratios
- Also adds the unenrollment reasons in the messages to inform users

Fixes #15115

### DIFF
--- a/experimenter/experimenter/slack/constants.py
+++ b/experimenter/experimenter/slack/constants.py
@@ -82,9 +82,9 @@ class SlackConstants:
         "Unenrollment rate: {rate:.1%} (threshold: {threshold:.1%})"
     )
     SLACK_SRM_MISMATCH_MESSAGE = (
-        "⚠️ Branches are misaligned (SRM detected) in *{experiment}*\n"
-        "P-value: {p_value:.6f} (threshold: {threshold})\n"
-        "This suggests unequal randomization or bias in enrollment."
+        "⚠️ *{experiment}* has a branch ratio mismatch.\n"
+        "Enrollment is not at the expected ratio across branches — "
+        "something may be wrong. Please review the experiment results"
     )
 
     # Slack notification log messages
--- a/experimenter/experimenter/slack/tasks.py
+++ b/experimenter/experimenter/slack/tasks.py
@@ -407,8 +407,6 @@ def _send_srm_mismatch_alert(experiment, p_value):
     try:
         message = SlackConstants.SLACK_SRM_MISMATCH_MESSAGE.format(
             experiment=experiment.name,
-            p_value=p_value,
-            threshold=SlackConstants.SRM_MISMATCH_P_VALUE_THRESHOLD,
         )
         email_addresses = [experiment.owner.email] if experiment.owner else []
         result = send_slack_notification(
--- a/experimenter/experimenter/slack/tests/test_tasks.py
+++ b/experimenter/experimenter/slack/tests/test_tasks.py
@@ -1042,7 +1042,9 @@ class TestCheckMonitoringAlerts(TestCase):
         ) as mock_send_slack:
             tasks._check_monitoring_alerts(experiment)
             mock_send_slack.assert_called_once()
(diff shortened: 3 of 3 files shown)


experimenter/experimenter/slack/tasks.py
```diff
@@ -301,6 +305,168 @@ def _send_error_alert(experiment, error_items):
...
+    enrollments = [b.get("enrollments", 0) for b in branches.values()]
+
+    if not enrollments or sum(enrollments) == 0:
+        return 1.0
+
+    expected_per_branch = sum(enrollments) / len(enrollments)
+    expected = [expected_per_branch] * len(enrollments)
+
+    _, p_value = chisquare(enrollments, expected)
+    return p_value
```
mikewilli, line 324: I thought we were going to put this in the ETL by just translating the SQL from the Looker version?

jaredlockhart: Oh I thought we said we'd do in Experimenter so it's easily modifiable?

mikewilli: Right, but I thought that was for the alert condition. I don't think there's any reason to redefine the SRM calculations in Experimenter in python when we already know what it needs to look like and have it defined in SQL.

experimenter/experimenter/slack/constants.py
```diff
@@ -72,6 +76,16 @@ class ErrorCode:
...
+    SLACK_UNENROLLMENT_SPIKE_MESSAGE = (
+        "⚠️ Unexpectedly large unenrollment in *{experiment}*\n"
+        "Primary reason: {reason}\n"
+        "Unenrollment rate: {rate:.1%} (threshold: {threshold:.1%})"
+    )
+    SLACK_SRM_MISMATCH_MESSAGE = (
+        "⚠️ Branches are misaligned (SRM detected) in *{experiment}*\n"
+        "P-value: {p_value:.6f} (threshold: {threshold})\n"
+        "This suggests unequal randomization or bias in enrollment."
+    )
```
jaredlockhart, line 88: I don't know if we need to explain the raw numbers when we send the alert, I think just something like 'the branches are not at the desired ratio, there may be something wrong, go check' (but worded nicely)

PR #15120, opened 2026-04-01
feat(nimbus): add randomly generated local results data to experiments

### LINKED ISSUE #6105: Add local dummy data for experiment results
Currently the nimbus-ui prepopulates dummy data for experiments with different lifecycles. Completed experiments have no results data locally. Devs working on the visualization results page need to complete multiple steps to have data to work with. Extending the `NimbusExperimentFactory` to included dummy results data we can view the results page without these extra steps.

The extra steps would still be worthwhile to be documented to help reproduce issues using production data locally.

moibra05:
Because

- Experiments have long been missing true accurate mock data
- The previous workflow required either needing access to the gcs buckets containing real data from live experiments to pull from the local admin page or manually copy and pasting an experiment results from the live admin page

This commit

- Adds randomly generated results data to experiments on local prod

Fixes #6105

### DIFF
--- a/experimenter/experimenter/experiments/tests/factories.py
+++ b/experimenter/experimenter/experiments/tests/factories.py
@@ -6,6 +6,7 @@ import random
 from collections.abc import Iterable
 from dataclasses import dataclass
 from enum import Enum
+from itertools import chain
 
 import factory
 from django.core.files.uploadedfile import SimpleUploadedFile
@@ -79,6 +80,8 @@ TEST_LOCALIZATIONS = """\
     }
 }
 """
+
+_UNSET = object()
 
 
 class NimbusFeatureConfigFactory(factory.django.DjangoModelFactory):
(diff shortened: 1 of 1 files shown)


experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -693,6 +964,18 @@ def create(
 
         experiment = super().create(*args, **kwargs)
 
+        results_data = kwargs.get("results_data", _UNSET)
+
+        if results_data is not _UNSET:
+            experiment.results_data = results_data
+        else:
+            experiment.results_data = build_random_results_data(
```
mikewilli, line 972: Why not just use `None` here instead of `_UNSET`? And in terms of the logic, couldn't you just do something like this?

```suggestion
        results_data = kwargs.get("results_data", None)

        experiment.results_data = results_data or build_random_results_data(
```

experimenter/experimenter/experiments/tests/factories.py
```diff
@@ -437,6 +440,275 @@ def random_slice(lst, *, min_items=0, max_items=2):
...
 
+def build_random_results_data(primary_outcomes, secondary_outcomes, application):
+    control_enrollment_pop = random.randint(100000, 150000)
+    control_exposure_pop = random.randint(50000, 100000)
+    treatment_enrollment_pop = random.randint(100000, 150000)
+    treatment_exposure_pop = random.randint(50000, 100000)
+
+    outcome_metrics = []
+
+    for outcome_slug in chain(primary_outcomes, secondary_outcomes):
```
mikewilli, line 451: Can you just add these instead of using `chain`, or what's the advantage of chain here?

```py
for outcome_slug in primary_outcomes + secondary_outcomes:
```

mikewilli (approved):
Couple questions but overall looks good, thanks! We've needed something like this for a long time.

It'd be nice if eventually we could do something like we have in the schemas package:
- define everything with pydantic model(s)
- create model factory that auto-generates randomly-filled in data that's valid for the schema (like [this](https://github.com/mozilla/experimenter/blob/main/schemas/mozilla_nimbus_schemas/jetstream/statistics.py#L59-L60))
- generate from the factory as needed ([example](https://github.com/mozilla/experimenter/blob/main/schemas/mozilla_nimbus_schemas/tests/jetstream/test_statistics.py#L109))

But in the meantime, this is definitely an improvement!

experimenter/experimenter/experiments/tests/factories.py
moibra05: there is no way to distinguish between an experiment that completely omits setting `results_data` (such as when we load in the dummy experiments) versus a test that is explicitly setting `results_data=None`. to solve this I created the `_UNSET` flag that allows me to represent the state where no value is assigned to `results_data` without needed to edit every test that passes `results_data=None` into the experiment creation methods

mikewilli: Ah fair enough, thanks!

experimenter/experimenter/experiments/tests/factories.py
moibra05: `chain` returns an iterator rather than creating a new list, so it’s marginally more memory efficient. that said, the performance gains here are negligible since `primary_outcomes` and `secondary_outcomes` together have at most 4 items. if there’s another advantage to using + instead, i'd be curious to know

PR #15127, opened 2026-04-02
chore(nimbus): replace summernote rte with prose

### LINKED ISSUE #14870: Takeaways overflow
The wysiwyg editors in the key takeaways modal are overflowing unneccarily

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6392)

moibra05:
Because

- The old django summernote rich text editor had unnecessary vertical overflow 
- This issue was particularily frustrating to find a fix for 

This commit

- Replaces the summernote rte with another editor called django prose which eliminates the overflow issue
- This new editor contains all the same important features provided by the old one including text formatting and image uploading

Fixes #14870


moibra05:
image

jaredlockhart:
Oh!!! One thing that came up recently was the editor we were using only emitted HTML which could contain malicious code and I wondered if we could find an editor that emits markdown instead.  If we're gonna swap out the editor can you see if there's one that's easily available that emits markdown? 🙏

moibra05:
> Oh!!! One thing that came up recently was the editor we were using only emitted HTML which could contain malicious code and I wondered if we could find an editor that emits markdown instead. If we're gonna swap out the editor can you see if there's one that's easily available that emits markdown? 🙏

Yea, I think [this](https://pypi.org/project/martor/) could be a viable option

jaredlockhart:
Oh yeah, that looks good, try it

moibra05:
sorry, i just noticed i didn't hit "comment" on my response i wrote days ago 🤦 . i spent a lot of time trying to get that markdown editor i linked to work but it refused to cooperate and ended up requiring way more code than expected to function properly (especially for image uploads), along with a bunch of compromises just to get it half working. i think the rte editor i’ve switched over to here should already be adequate from a security standpoint since it’s pretty strict on the html tags it allows, which are listed [here](https://github.com/withlogicco/django-prose/blob/main/README.md?plain=1#L157C1-L160C81)

PR #15142, opened 2026-04-06
fix(nimbus): add guardrail to prevent adding fetched targeting files into nonexistent directories

### LINKED ISSUE #15129: Missing test case for targeting file fetch
We are missing a test case to handle the situation where an app does not have its version correctly set. This is causing erroneous behaviour where we attempt to save a file to an invalid path:

FileNotFoundError: [Errno 2] No such file or directory: '/experimenter/experimenter/features/manifests/ios/vNone/RecordedNimbusContext.swift'

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6724)

moibra05:
Because

- After fetching the targeting context files we were impartially attempting to insert files into `manifest_dir / app_config.slug / f"v{version}"`
- This would not always work because `version` could occasionaly be set to `None` leading to errors when saving files to this path

This commit

- Adds a check to only fetch and insert targeting files if the app version is available

Fixes #15129

### DIFF
--- a/experimenter/manifesttool/fetch.py
+++ b/experimenter/manifesttool/fetch.py
@@ -37,19 +37,28 @@ class FetchResult:
 
 
 def fetch_targeting_files(
-    save_path: Path,
-    logging_msg: str,
+    manifest_dir: Path,
+    version: Optional[Version],
     app_config: AppConfig,
+    app_name: str,
     ref: Ref,
 ) -> None:
     targeting_files_path = app_config.targeting_files
     if targeting_files_path:
+        logging_msg = (
+            f"fetch: {app_name} at {ref} version {version} downloading targeting files"
+        )
+        path = manifest_dir / app_config.slug
+
+        if version:
+            path /= f"v{version}"
+
         try:
             github_api.fetch_file(
                 app_config.repo.name,
                 targeting_files_path[0],
                 ref.target,
-                save_path / Path(targeting_files_path[0]).name,
+                path / Path(targeting_files_path[0]).name,
             )
             print(logging_msg)
         except HTTPError as e:
@@ -132,9 +141,10 @@ def fetch_fml_app(
             )
 
         fetch_targeting_files(
-            manifest_dir / app_config.slug / f"v{version}",
-            f"fetch: {app_name} at {ref} version {version} downloading targeting files",
+            manifest_dir,
+            version,
             app_config,
+            app_name,
             ref,
         )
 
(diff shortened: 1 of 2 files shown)


yashikakhurana (approved):
Approving to unlock to update feature config

experimenter/manifesttool/fetch.py
```diff
@@ -131,12 +125,14 @@ def fetch_fml_app(
...
             )
 
-        fetch_targeting_files(
-            manifest_dir / app_config.slug / f"v{version}",
-            f"fetch: {app_name} at {ref} version {version} downloading targeting files",
-            app_config,
-            ref,
-        )
+        if version:
+            fetch_targeting_files(
```
freshstrangemusic: remove the top level fetch_targeting_files in fetch_releases. fetch_releases calls fetch_fml_app and fetch_legacy_app and so it doesnt have to be responsible for that. The responsibility should be entirely on fetch_{legacy,fml}_app

freshstrangemusic: Lets have `fetch_targeting_files` take `manifest_dir`, `app_config`, and `version` and have it build up the path itself.

experimenter/manifesttool/fetch.py
```diff
@@ -131,12 +125,14 @@ def fetch_fml_app(
...
+                manifest_dir / app_config.slug / f"v{version}",
+                f"fetch: {app_name} at {ref} version {version} downloading targeting"
+                "files",
+                app_config,
+                ref,
+            )
```
freshstrangemusic:
```suggestion
            fetch_targeting_files(
                manifest_dir,
                app_config,
                version,
                ref,
                (
                    f"fetch: {app_name} at {ref} version {version} downloading targeting"
                    "files"
                ),
            )
```

experimenter/manifesttool/fetch.py
```diff
@@ -200,12 +196,14 @@ def fetch_legacy_app(
             manifest_path,
         )
 
-        fetch_targeting_files(
-            manifest_dir / app_config.slug / f"v{version}",
-            f"fetch: {app_name} at {ref} version {version} downloading targeting files",
-            app_config,
-            ref,
-        )
+        if version:
```
freshstrangemusic: Same here.

freshstrangemusic (changes requested):

experimenter/manifesttool/fetch.py
freshstrangemusic: Actually if we're passing app_config, ref, and version, we do not need to pass the log string -- we can build it in fetch_targeting_files

experimenter/manifesttool/fetch.py
```diff
@@ -37,19 +37,25 @@ def __str__(self):  # pragma: no cover
 
 
 def fetch_targeting_files(
-    save_path: Path,
+    manifest_dir: Path,
+    version: Optional[Version],
     logging_msg: str,
```
freshstrangemusic: Lets build the logging msg in this function

PR #15155, opened 2026-04-07
fix(nimbus): add check for weekly results before appending error

### LINKED ISSUE #15033: Verify weekly results are expected before including error
In the jetstream client code, we inject an error into the `results_data` whenever we try to fetch results from GCS that don’t exist. We used to fetch only weekly/overall, and the results page was disabled until weekly was available, but now we are fetching daily and so results start becoming available earlier. This means that we need to ensure that we actually expect the `weekly` results to be available before we log an error that we can’t fetch them.

We already have a [check like this](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/jetstream/client.py#L361-L364) for the overall period, but now we need one for weekly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6581)

moibra05:
Because

- The results page used to only be available once an experiment had weekly results available
- To further enforce this we would inject weekly errors into the results data object regardless of whether they were actually ready or not
- Since we now fetch and display daily results we must change this so weekly errors are only included if we should be expecting them i.e. if it has been more than a week since the end of enrollment

This commit

- Adds a check to verify if weekly results should be expected before including an error
- Weekly results being expected is dependent on the `results_ready` field which is true whenever it has been more than a week since an experiment completes enrollment

Fixes #15033

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -242,7 +242,19 @@ def get_experiment_data(experiment: NimbusExperiment):
         try:
             data_from_jetstream = get_data(recipe_slug, window)
         except RuntimeError as e:
-            runtime_errors.append(str(e))
+            # only store runtime errors for weekly/overall windows if we expect those
+            # results (overall also lags by one day so there is time for analysis to
+            # complete)
+            if (
+                (window == AnalysisWindow.DAILY)
+                or (window == AnalysisWindow.WEEKLY and experiment.results_ready)
+                or (
+                    window == AnalysisWindow.OVERALL
+                    and experiment.end_date
+                    and experiment.end_date < (date.today() - timedelta(days=1))
+                )
+            ):
+                runtime_errors.append(str(e))
 
         segment_points_enrollments = defaultdict(list)
         segment_points_exposures = defaultdict(list)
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/jetstream/client.py
```diff
@@ -376,22 +376,26 @@ def get_experiment_data(experiment: NimbusExperiment):
...
-        if "overall.json" not in e or (
-            "overall.json" in e
-            and experiment.end_date
+        # only store runtime errors for weekly/overall windows if we expect those results
+        # (overall also lags by one day so there is time for analysis to complete)
+        if "weekly.json" in e and not experiment.results_ready:
+            continue
+
+        if "overall.json" in e and not (
+            experiment.end_date
```
mikewilli: So I know I did it this way and the logic you updated looks sound, but what do you think about moving this to the except block on line 244, and then here we could just create the AnalysisErrors?

Up in the except block, it would look very similar, but we could use the `window`. Something like:
```py
try:
    data_from_jetstream = get_data(recipe_slug, window)
except RuntimeError as e:
    if (
        (window == AnalysisWindow.DAILY)
        or (window == AnalysisWindow.WEEKLY and experiment.results_ready)
        or (window == AnalysisWindow.OVERALL and experiment.end_date and experiment.end_date < (date.today() - timedelta(days=1)))
    ):
        runtime_errors.append(str(e))
```

I dunno, is this better or worse? What do you think?

moibra05: yea i think it makes more sense keeping all the related error handling logic close together

experimenter/experimenter/jetstream/client.py
```diff
@@ -242,7 +242,16 @@ def get_experiment_data(experiment: NimbusExperiment):
         try:
             data_from_jetstream = get_data(recipe_slug, window)
         except RuntimeError as e:
-            runtime_errors.append(str(e))
+            if (
+                (window == AnalysisWindow.DAILY)
```
mikewilli, line 249: Yea I like this better, good choice!

Can you just add a comment here similar to what we had before?

PR #15197, opened 2026-04-09
feat(nimbus): Use threads to send alerts and reminders

### LINKED ISSUE #15192: Fold results available, errors and reminders in the threads
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6634)

yashikakhurana:
Because

- We are sending reminders to end enrollment, end experiment, alerting regarding the analysis errors and results availability and now we have the notifications too related to expected unenrollments and SRM, and all these notification are making it noisy

This commit

- Folds all the above notifications into the launch thread, and if the launch message is not available it will just send it is as it is doing now as top level notification

Fixes #15192


experimenter/experimenter/slack/notification.py
```diff
@@ -70,12 +70,25 @@ def _send_dm_to_user(client, user_id, message, channel_message_link=None):
...
 
+def get_launch_request_thread(experiment_id):
+    alert = NimbusAlert.objects.filter(
+        experiment_id=experiment_id,
+        alert_type=NimbusConstants.AlertType.LAUNCH_REQUEST,
+        slack_thread_id__isnull=False,
+        slack_channel_id__isnull=False,
+    ).last()
+    if alert:
+        return alert.slack_thread_id, alert.slack_channel_id
```
jaredlockhart: You could just return the alert object and unpack the fields you need at the callsite, looks like below we throw out the channel id anyways?

experimenter/experimenter/experiments/tests/test_email.py
```diff
@@ -293,3 +296,54 @@ def test_send_experiment_ending_email_experiment_subscribers_only(
...
+            message="Launch request",
+            slack_thread_id="1234567890.123456",
+            slack_channel_id="C123456",
+        )
+
+        nimbus_send_experiment_ending_email(experiment)
+
+        mock_slack_task.assert_called_once()
+        call_kwargs = mock_slack_task.call_args.kwargs
+        self.assertEqual(call_kwargs["thread_ts"], "1234567890.123456")
```
jaredlockhart: Couldn't you just fold these assertions under the similar test cases above?

jaredlockhart (dismissed):
Okay looks good, couple thoughts but nothing major, let's try it thnx @yashikakhurana 🎉

jaredlockhart (approved):
Very nice let's try it! Thnx @yashikakhurana 🎉 🎉 🎉

PR #15211, opened 2026-04-10
feat(nimbus): Enforce experiment key in messaging messages

### LINKED ISSUE #15210: Require "experiment": "{experiment}" in messaging feature
The messaging feature needs to enforce that the "experiment" field should exist. We can make this required in FML, but that will only affect versions going forward. Instead, lets make this required as a check in experimenter itself.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6770)

freshstrangemusic:
Because:

- iOS and Fenix have a "messaging" feature with a special "experiment" key in the message;
- this key needs to have a value of "{experiment}", but is optional; and
- if this key is not present, telemetry may not be submitted correctly

this commit:

- enforces that the key exists and is the correct value.

Fixes #15210

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -1133,15 +1133,18 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
         schemas_in_range: NimbusFeatureConfig.VersionedSchemaRange,
         loader: NimbusFmlLoader,
         feature_config: NimbusFeatureConfig,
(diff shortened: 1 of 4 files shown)


jaredlockhart (approved):
Oh yeah I remember this, it was the lesser evil of the possible approaches.  I wonder why we didn't add this validation a while ago? Was there ever a case where you didn't want to include the placeholder?  Anyways thanks for adding @freshstrangemusic 🙏

PR #15217, opened 2026-04-13
fix(project): close and recreate bot PRs to trigger GHA CI

### LINKED ISSUE #15215: Bot PR force-pushes don't trigger GHA CI runs
## Problem

When the cron job bot (experimenter-github-app) force-pushes new content to an existing PR branch, GHA workflows don't trigger. This is because GitHub App installation tokens don't generate push/synchronize events on existing PRs (anti-recursion protection).

CI only runs on the initial PR creation (`pull_request:opened`). Subsequent force-pushes with new content are untested.

Affected branches: `external-config`, `update-application-services`, `update_firefox_desktop_release`, `update_firefox_desktop_beta`

## Solution

Change the cron workflows to close the existing PR before pushing and creating a fresh one. This ensures every push gets a `pull_request:opened` event which does trigger CI.

```bash
gh pr close external-config --repo mozilla/experimenter 2>/dev/null || true
git push origin external-config -f
gh pr create ...
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6773)

jaredlockhart:
Because

* GitHub App installation tokens don't generate push/synchronize
  events on existing PRs, so subsequent force-pushes to bot PRs
  don't trigger GHA CI workflows
* Only pull_request:opened events trigger CI from App tokens
* See [Triggering a workflow from a workflow](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow)

This commit

* Closes the existing PR before pushing and creating a fresh one
  in the update-configs and update-firefox workflows
* This ensures every update gets a pull_request:opened event which
  triggers the full CI suite
* The close/recreate only fires when the content actually changed
  (gated by the existing git diff check)

Fixes #15215

### DIFF
--- a/.github/workflows/update-configs.yml
+++ b/.github/workflows/update-configs.yml
@@ -48,9 +48,9 @@ jobs:
             git add experimenter/ schemas/
             git commit -m 'chore(nimbus): Update External Configs'
             if (( $((git diff external-config origin/external-config || git diff HEAD~1) | wc -c) > 0 )); then
+              gh pr close external-config --repo mozilla/experimenter 2>/dev/null || true
               git push origin external-config -f
-              gh pr create -t "chore(nimbus): Update External Configs" -F /tmp/pr-body.txt --base main --head external-config --repo mozilla/experimenter || \
-                gh pr edit external-config -F /tmp/pr-body.txt
+              gh pr create -t "chore(nimbus): Update External Configs" -F /tmp/pr-body.txt --base main --head external-config --repo mozilla/experimenter
             else
               echo "Changes already committed, skipping"
             fi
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
I'm a little confused about this because the docs appear to be saying that the current method of triggering a workflow within a workflow via GH App installation token should work, while using `GITHUB_TOKEN` would not work.

But even if I'm reading that right, it seems like the current workflow isn't working as expected, and in the worst case I don't think this will break anything, so worth trying regardless.

jaredlockhart:
I am equally confused about what the docs say vs how it's behaving, but it seems to be the case that pushing to an open pr will not retrigger the checks and is a design choice of how GHA works.

PR #15219, opened 2026-04-13
fix(project): move path filters to job-level conditions in GHA workflows

### LINKED ISSUE #15218: Move GHA workflow path filters from trigger-level to job-level conditions
## Problem

GHA workflows that use `paths:` filters at the trigger level don't report any status when the paths don't match. GitHub docs confirm:

> If a workflow is skipped due to path filtering, then checks associated with that workflow will remain in a "Pending" state.

This blocks PRs from merging when the workflow's checks are required but the PR doesn't touch the relevant paths.

CircleCI handles this by running the job and exiting early, which reports a green check. GHA doesn't run the job at all.

## Solution

Move path filtering from the workflow `on:` trigger into job-level `if:` conditions. When a job is skipped via `if:`, it reports as success (green check) rather than staying pending.

Use a git diff check in the first step to determine changed paths — no third-party actions, just `git diff --name-only` against the base branch.

## Affected workflows

All check and integration test workflows that currently use `paths:` in their triggers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6774)

jaredlockhart:
Because

* Workflow-level `paths:` filters cause checks to stay in Pending state
  when the paths don't match, which blocks PRs when those checks are
  required
* [GitHub docs confirm](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks):
  "If a workflow is skipped due to path filtering, then checks
  associated with that workflow will remain in a Pending state"
* Changes to root-level files or `.github/` should trigger all workflows
  to catch CI config breakage

This commit

* Adds `check-changed-paths` composite action that uses git diff to
  determine if relevant paths changed
* Always runs on merge_group, workflow_dispatch, and push to main
* Always runs when root-level files or `.github/` are changed
* Removes `paths:` from all workflow triggers, replaced with job-level
  `if:` conditions that skip heavy steps but still report green

Fixes #15218

### DIFF
--- a/.github/actions/check-changed-paths/action.yml
+++ b/.github/actions/check-changed-paths/action.yml
@@ -1,0 +1,60 @@
+name: "Check Changed Paths"
+description: "Check if any files matching the given patterns were changed, always true for merge_group and push to main"
+
+inputs:
+  paths:
+    description: "Space-separated list of path patterns to check (e.g. 'experimenter/ application-services/')"
+    required: true
+
+outputs:
+  should-run:
+    description: "Whether the job should run based on changed paths"
+    value: ${{ steps.check.outputs.should_run }}
+
+runs:
+  using: composite
+  steps:
+    - name: Check changed paths
+      id: check
+      shell: bash
+      run: |
+        if [[ "${{ github.event_name }}" == "merge_group" || "${{ github.event_name }}" == "workflow_dispatch" ]]; then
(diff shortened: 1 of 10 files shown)


yashikakhurana (approved):
Lets goo 🎉 , thank you @jaredlockhart

PR #15220, opened 2026-04-13
perf(project): add GHA native cache for Docker images

jaredlockhart:
Because

* Docker image builds account for ~12 minutes of each 18-minute CI job
  (3.5 min in setup-cached-build + 8.5 min in build_prod inside make)
* The megazords, schemas, and cirrus images rarely change between PRs
* GHA native cache (actions/cache) is co-located with runners, avoiding
  the registry round-trip overhead that made GAR caching slower

This commit

* Uses docker save/load to cache built images as tar files
* Cache key is based on sha256 of Dockerfile + dependency lock files
* On cache hit, docker load replaces the full build (~seconds vs minutes)
* On cache miss, builds normally and saves for next run
* First run will be a cache miss (populates the cache), subsequent runs
  with the same dependencies will be cache hits

Fixes #15193


jaredlockhart:
## Cache Performance Analysis

Compared cached (this PR) vs non-cached (4 baseline PRs) across 3 representative jobs:

| Job | Baseline avg (n=4) | Baseline range | Cached | Delta |
|-----|-------------------|----------------|--------|-------|
| Enrollment (0) | 17.9m | 15m52s - 21m39s | 15m48s | -2.1m |
| Nimbus UI | 16.7m | 15m22s - 17m32s | 16m39s | -0.0m |
| RS Launch | 16.4m | 14m49s - 17m21s | 16m37s | +0.2m |

The cache saves ~1 minute on the setup-cached-build step (loading images vs building), but this is lost in 3-4 minute runner variance. The cached run falls within the normal baseline range for all jobs.

The experimenter images (4GB+) are too large for GHA's 10GB cache limit, so the 8.5 minute build_prod inside make can't be cached — that's where most of the time goes.

Not worth the added complexity. Closing.

PR #15224, opened 2026-04-13
Add 48 hour churned user targeting && lapsed user targeting for Windows

AllegroFox:
To support the [48h churned returning user](https://docs.google.com/document/d/1yIynQ67mVarSx0sy1287wL_gkeZIPF5uMfGLD9UuOG4/edit?tab=t.0) experiment,
This commit adds two targeting expressions. 

The first is a LAPSED_USER targeting expression updated to target only Windows users, which will be used to enroll users in a "dummy rollout" when they return to the browser after having been lapsed for 28 days. (Based on `userMonthlyActivity.)

The second expression will target the users enrolled in this first rollout, who have now lapsed for an additional 2 days after returning. (based on the `currentDate` of the default profile, seen from backgroundTask.) This is necessary since once the user has returned to the browser, their `userMonthlyActivity` will update and can no longer be used for the second expression.

### DIFF
--- a/experimenter/experimenter/targeting/constants.py
+++ b/experimenter/experimenter/targeting/constants.py
@@ -2633,21 +2633,7 @@ LAPSED_USER_WINDOWS_ONLY = NimbusTargetingConfig(
         "Users with a profile age of 28 days and 0 days of activity "
         "in the past 28 days, using Windows"
     ),
-    targeting=(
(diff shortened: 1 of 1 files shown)


jaredlockhart (approved):
Okay lgtm thmx @AllegroFox 🙏

Issue #14004, opened 2025-11-20
Experimenter not deploying


freshstrangemusic:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6198)

PR #15238, opened 2026-04-14
bug(project): Fix docker tags for deploys

### LINKED ISSUE #15237: Deploys broken
deployed images are generated with the wrong image tag and so argocd isnt picking them up

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6778)

freshstrangemusic:
Because:

- ArgoCD expects a image tag of the format `sha-{SHA9}` where SHA9 is *exactly* the first 9 characters of the SHA;
- the mozcloud publish action was added to use 10 characters of the SHA; and
- we recently removed our CircleCI action that was pushing with the correct format, causing deploys to stop

this commit:

- updates our tags to correctly use 9 characters of SHAs.

See-also #14004

Fixes #15237

### DIFF
--- a/.github/workflows/deploy-cirrus.yml
+++ b/.github/workflows/deploy-cirrus.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Generate image tag
         id: tag
         run: |
-          TAG="$(git rev-parse --short=10 HEAD)"
+          TAG="$(git rev-parse --short=9 HEAD)"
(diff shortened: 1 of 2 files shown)


jaredlockhart (approved):
Ah good catch ty 🙏

mikewilli (approved):

Issue #15050, opened 2026-03-26
3 Days Retention is not being display properly


data-sync-user:
3 Days Retention is considered “final” on Day 4 of results, and that is what should be displayed on the main results page. However, this seems to be displaying the most recent day of results, at least for the in-progress experiment I checked. You can see this behavior in the screenshots below:

Main page:

!image-20260319-203445.png|width=1073,alt="image-20260319-203445.png"!

Details View

!image-20260319-203512.png|width=566,alt="image-20260319-203512.png"!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/EXP-6657)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/120988">image-20260319-203445.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/120989">image-20260319-203512.png</a>

data-sync-user:
➤ Mike Williams commented:

Here’s the experiment I noticed this one: https://experimenter.services.mozilla.com/nimbus/week-1-privacy-focused-nudges-for-resurrected-users/results/ ( https://experimenter.services.mozilla.com/nimbus/week-1-privacy-focused-nudges-for-resurrected-users/results/|smart-link )

Issue #12647, opened 2025-05-19
Do not fall back to week-1 retention when week-2 is not available


mikewilli:
@danielkberry reported that this behavior is causing confusion, and the week-1 retention results are not useful so they should not be shown.

We currently retrieve this in the Jetstream client code that ingests analysis results into Experimenter. Instead of falling back to week 1, we should include a warning for the metric that the experiment was too short to compute 2-week retention results.

PR #15239, opened 2026-04-14
feat(nimbus): add analaysis window indicator inside metric name cards

### LINKED ISSUE #14850: Display which time window each metric is using on results
On the results page, any one metric may be displaying one of:

- A daily result
- A weekly result
- Overall result

And this can change over the course of the lifespan of the experiment.  

We should display somewhere on each metrich which time window is being used to show that metric, maybe on the metric title like

Daily Active Users
    (Week 1)

Retention
    (Day 3)

Clicks
   (Overall)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6578)

moibra05:
Because

- It's often times unclear which analysis window is currently being displayed for a metric's results

This commit

- Adds a text indicator that tells us which analysis window is being reflected in the main page results

Fixes #14850


moibra05:
image

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -387,6 +387,11 @@ def append_kpi_metric_fields(
                 kpi["slug"], kpi["group"], analysis_basis, segment, reference_branch
             )
 
+            if kpi["slug"] == NimbusConstants.RETENTION_3_DAYS:
+                kpi["displayed_window"] = "Day 4"
+            if kpi["slug"] == NimbusConstants.RETENTION:
+                kpi["displayed_window"] = "Week 2"
```
mikewilli, line 393: Are these always accurate based on the data available? Like what happens on Day 3? Do we show "Day 4" for RETENTION_3_DAYS, or do we just show nothing? And I'm fairly sure Week 2 for Retention could be wrong in some cases because we haven't done https://github.com/mozilla/experimenter/issues/12647 yet, so it could be Week 1 (the old Results page always said Week 2 Retention so maybe we don't need to block on this).

I just want to make sure we aren't going to be potentially showing the wrong thing for early or wonky (missing days/weeks) results.

moibra05: fair points. the changes im making in #15050 will make it so 3day retention will remain unavailable until day 4 data is available so it will show nothing on Day <4 or if the metric has no data (this goes for all metrics). in that case i think the most logical thing to do would be to complete #15050 and #12647 before this

mikewilli:
There was some recent confusion on what the Overview section of the metric details modal is showing -- can/should we tack something on to this PR to update that too? Or file a new ticket?

moibra05:
i think we can just include that change in this pr and broaden the scope/title to covering general improvements to analysis window representation

moibra05:
We discussed how 3DR and 2-week retention should be labeled in the UI. While there was some discussion around whether different time windows (e.g. Overall vs Day 4 for 3DR, Overall vs Week 2 for retention) should be shown, the consensus was to keep this ticket scoped to its original goal: adding a label that reflects the currently displayed window. Any broader changes to retention window semantics can be handled in a follow-up ticket. For the current implementation, we've decided labels such as "Day 4" for 3DR and "Week 2" for retention accurately describe the existing behavior.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-inner.html
```diff
@@ -333,6 +333,13 @@ <h4 class="mb-0">{{ area }}</h4>
                             data-bs-target="#{{ experiment.slug }}-{{ metric.slug }}-{{ area|slugify }}"
                             style="height: 100px">
                       <i class="fa-solid fa-arrow-up-right-from-square text-secondary fa-sm position-absolute top-0 end-0 p-3 px-2"></i>
+                      <small class="position-absolute bottom-0 end-0 p-2 fw-medium">
+                        {% if metric.displayed_window %}
+                          {{ metric.displayed_window|upper }}
+                        {% else %}
+                          {{ displayed_window|upper }}
```
mikewilli, line 340: This fallback is just the general period (like "week") instead of the specific window ("week 2"), right? If so, that seems like a good fallback. However...

I am a little concerned about this scenario where we don't know the displayed window for a metric. We want to make sure we aren't just displaying "week" just because that's what the other metrics are, if theoretically the actual result displayed could be "day" or "overall". Is that behavior possible with this logic or am I concerned over nothing here?

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -401,6 +401,14 @@ def append_kpi_metric_fields(
...
             )
 
+            if (
+                kpi["slug"] == NimbusConstants.RETENTION_3_DAYS
+                or kpi["slug"] == NimbusConstants.RETENTION_3_DAYS_DESKTOP
+            ):
+                kpi["displayed_window"] = "Day 4"
+            if kpi["slug"] == NimbusConstants.RETENTION:
+                kpi["displayed_window"] = "Week 2"
+
```
mikewilli, line 415: I think that since these follow the default behavior (for now), we should comment out this block and leave an accompanying comment that links to the follow-up ticket on specific retention semantics. Especially if I'm remembering right that the weekly retention still falls back to week 1 if week 2 is unavailable. (less of a concern for 3DR because I believe we just drop everything except day 4)

mikewilli (commented):
Super simple, love it! There are a couple potential scenarios I'd like to avoid that I left comments about. Please feel free to let me know if the concerns are unfounded, otherwise we should ensure we guard against them before merging to avoid potentially displaying incorrect info.

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-inner.html
moibra05: the only metrics that would have a value for `metric.displayed_window` are 3dr and retention so the fallback is really the default. also, the way the results page works now is that it decides which window from ["overall", "weekly", "daily"] (in that order) isn't empty from the results data and selects that as the primary source of data for what to show on the main page for ALL metrics. i mention this because unless we know ahead of time that a particular metric only computes week 1 data but everything else has overall available - the results page will attempt to show overall data for that metric which won't be present and so it'd appear as though it had an error

mikewilli: Ah ok, thanks for clarifying. In that case, this seems fine to me.

experimenter/experimenter/jetstream/results_manager.py
```diff
@@ -401,6 +401,18 @@ def append_kpi_metric_fields(
                 kpi["slug"], kpi["group"], analysis_basis, segment, reference_branch
             )
 
+            if (
+                kpi["slug"] == NimbusConstants.RETENTION_3_DAYS
+                or kpi["slug"] == NimbusConstants.RETENTION_3_DAYS_DESKTOP
+            ):
+                kpi["displayed_window"] = "Day 4"
```
mikewilli, line 408: Isn't this unnecessary too?

moibra05: sorry what did you mean by unnecessary? isn't it still true that 3dr only shows the day 4 window index? if that isn't there the displayed window for it will be either daily, weekly, or overall all of which presumably wouldn't be correct

mikewilli: Oh right, I keep forgetting that the label by default is not specific to each metric based on the data that comes in. Ok, let's go with this then.

mikewilli (approved):

PR #15248, opened 2026-04-14
fix(nimbus): omit monitoring data when cloning experiments

### LINKED ISSUE #15247: Omit monitoring data when cloning
When we clone an experiment we should omit the monitoring data.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6784)

jaredlockhart:
Because

* Cloned experiments carried over monitoring data from the parent
* Monitoring data is specific to the parent experiment's lifecycle and
  is not relevant to the new clone

This commit

* Nulls out `monitoring_data` during the clone step
* Adds a data migration to clear `monitoring_data` on non-Live/Complete
  experiments
* Updates clone test to set `monitoring_data` on parent and assert it is
  `None` on the child
* Updates migration test to cover the new data migration

Fixes #15247

### DIFF
--- a/experimenter/experimenter/experiments/migrations/0326_clear_monitoring_data_non_live_complete.py
+++ b/experimenter/experimenter/experiments/migrations/0326_clear_monitoring_data_non_live_complete.py
@@ -1,0 +1,18 @@
+from django.db import migrations
+
+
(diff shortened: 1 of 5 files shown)


mikewilli (approved):
LGTM

PR #15284, opened 2026-04-15
feat(nimbus): Should skip same day alert for unenrollments

### LINKED ISSUE #15221: wait atleast 1 day for the enrollment to start to start alerts on unenrollments
If the enrollment just started and the job task ran just after that, then it may notifies the users false positive about enrollment monitoring, we should wait at least a day and then should notify on the next job task run

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6775)

yashikakhurana:
Because

- If the enrollment just started and the job task ran just after that, then it may notifies the users false positive about enrollment monitoring, we should wait at least a day and then should notify on the next job task run

This commit

- Skips the alert and only sends the alert if its more than 1 day

Fixes #15221

### DIFF
--- a/experimenter/experimenter/slack/tasks.py
+++ b/experimenter/experimenter/slack/tasks.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 from datetime import timedelta
 
@@ -424,6 +425,11 @@ def _check_monitoring_alerts(experiment):
         return
 
     if not experiment.monitoring_data:
+        return
+
+    if experiment.start_date and (
+        datetime.date.today() - experiment.start_date
+    ) < datetime.timedelta(days=1):
         return
 
     try:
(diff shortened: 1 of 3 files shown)


experimenter/experimenter/slack/tasks.py
```diff
@@ -426,6 +427,11 @@ def _check_monitoring_alerts(experiment):
     if not experiment.monitoring_data:
         return
 
+    if experiment.start_date and (
+        datetime.date.today() - experiment.start_date
+    ) < datetime.timedelta(days=1):
```
jaredlockhart: Maybe that `1` should be a constant somewhere so we clearly know what it's for and what it's doing.

jaredlockhart (dismissed):

experimenter/experimenter/slack/tasks.py
yashikakhurana: yup good idea

jaredlockhart (approved):
Yeah that definitely reads better awesome ty @yashikakhurana 🎉

PR #15285, opened 2026-04-15
feat(nimbus): Skip SRM section on summary page and alerts for rollouts

### LINKED ISSUE #15282: Hide SRM section for the rollouts
Hide SRM section for the rollouts as it is not applicable in that case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6786)

yashikakhurana:
Because

- We should not be sending SRM alerts for the rollouts and similarly SRM section is not applicable to show on summary page

This commit

- Skips SRM alerts for the rollouts
- Does not dispaly SRM results for the rollouts

Fixes #15282

### DIFF
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail_card_monitoring.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail_card_monitoring.html
@@ -78,66 +78,68 @@
         </table>
       </div>
     </div>
-    <div class="card">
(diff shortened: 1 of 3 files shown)


jaredlockhart (approved):
Yep good enough for now, as we add more alerts we can think about if we want a more generic way to list all alerts out for each delivery type and just loop over them instead of checking like this, but no need to dig into that yet.

Looks good thnx @yashikakhurana 🙏

PR #15287, opened 2026-04-15
fix(nimbus): fix 3 days retention showing incorrect values

### LINKED ISSUE #15050: 3 Days Retention is not being display properly

moibra05:
Because

- We were displaying incorrect values for 3 day retention on the main page particularly affecting experiments with 4 < days <= 7 of results

This commit

- Ensures that only day 4 of the 3 day retention metric is inserted into the results data object
- Hides the weekly/daily breakdown for 3 day retention popout view
- This approach allows minimal changes to the results view which otherwise would require numerous edge case handling and ultimately make for much harder code to maintain

Fixes #15050

### DIFF
--- a/experimenter/experimenter/jetstream/models.py
+++ b/experimenter/experimenter/jetstream/models.py
@@ -77,6 +77,7 @@ GROUPED_METRICS = {
     Group.SEARCH: SEARCH_METRICS,
     Group.USAGE: USAGE_METRICS,
 }
+RETENTION_3_DAYS_WINDOW_INDEX = 4
 
 
 # A map of metric -> group for quick lookups.
@@ -168,9 +169,22 @@ class JetstreamData(RootModel[JetstreamDataPoint]):
         # Extract the 3-day retention data (window index 4)
         # without falling back to earlier windows
         retention_data = self.get_retention_by_window(
-            4, daily_data, Metric.RETENTION_3_DAYS
+            RETENTION_3_DAYS_WINDOW_INDEX, daily_data, Metric.RETENTION_3_DAYS
         )
 
+        self.extend(retention_data)
+
+    def replace_retention_3_days(self, daily_data):
+        # Extract and replace with the 3-day retention data (window index 4)
+        retention_data = self.get_retention_by_window(
+            RETENTION_3_DAYS_WINDOW_INDEX, daily_data, Metric.RETENTION_3_DAYS
+        )
+
+        self.root = [
+            jetstream_data_point
+            for jetstream_data_point in self.root
+            if jetstream_data_point.metric != Metric.RETENTION_3_DAYS
+        ]
         self.extend(retention_data)
 
 
--- a/experimenter/experimenter/nimbus_ui/templates/common/metric_popout.html
+++ b/experimenter/experimenter/nimbus_ui/templates/common/metric_popout.html
@@ -61,7 +61,7 @@
           </div>
         </div>
         {% with weekly_metric_data=all_weekly_metric_data|dict_get:metric_info.slug %}
-          {% if weekly_metric_data.has_weekly_data %}
+          {% if weekly_metric_data.has_weekly_data and not metric_info.slug == experiment.RETENTION_3_DAYS %}
             <div class="border border-1 rounded py-4 px-5 rounded-4">
               <h2 class="fs-6 fw-bold">Weekly breakdown</h2>
               <div class="d-flex mt-3">
@@ -123,7 +123,7 @@
           {% endif %}
         {% endwith %}
         {% with daily_metric_data=all_daily_metric_data|dict_get:metric_info.slug %}
-          {% if daily_metric_data.has_daily_data %}
+          {% if daily_metric_data.has_daily_data and not metric_info.slug == experiment.RETENTION_3_DAYS %}
             <div class="border border-1 rounded py-4 px-5 rounded-4">
               <h2 class="fs-6 fw-bold">Daily breakdown</h2>
               <div class="d-flex mt-3">
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -306,9 +306,25 @@ def get_experiment_data(experiment: NimbusExperiment):
 
                 if segment == Segment.ALL:
                     experiment_data["other_metrics"] = other_metrics
-            elif (data and window == AnalysisWindow.WEEKLY) or (
-                data and window == AnalysisWindow.DAILY
-            ):
+            elif data and window == AnalysisWindow.WEEKLY:
+                # Append 3-day retention from daily data
+                data.append_retention_3_days(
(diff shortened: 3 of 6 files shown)


experimenter/experimenter/jetstream/models.py
```diff
@@ -173,6 +173,19 @@ def append_retention_3_days(self, daily_data):
 
         self.extend(retention_data)
 
+    def replace_retention_3_days(self, daily_data):
+        # Extract and replace with the 3-day retention data (window index 4)
+        retention_data = self.get_retention_by_window(
+            4, daily_data, Metric.RETENTION_3_DAYS
```
jaredlockhart: Ah `4` here is we pull it out of day 4 eh?  That should be a well named constant otherwise it's easy to miss or not know why we're only pulling that one index out.

experimenter/experimenter/nimbus_ui/constants.py
```diff
@@ -255,6 +255,8 @@ class NimbusUIConstants:
     monitoring period is complete.
     <a href="https://experimenter.info/workflow/monitoring">Learn more</a>"""
 
+    RETENTION_3_DAYS = NimbusConstants.RETENTION_3_DAYS
```
jaredlockhart: What's this for?  Do we need this?

experimenter/experimenter/nimbus_ui/templates/common/metric_popout.html
```diff
@@ -61,7 +61,7 @@ <h2 class="fs-6 fw-bold">Overview</h2>
           </div>
         </div>
         {% with weekly_metric_data=all_weekly_metric_data|dict_get:metric_info.slug %}
-          {% if weekly_metric_data.has_weekly_data %}
+          {% if weekly_metric_data.has_weekly_data and not metric_info.slug == NimbusUIConstants.RETENTION_3_DAYS %}
```
jaredlockhart: That's a bit annoying.  Probably the best we can do for now, but inevitably we'll wind up needing to do a second special case and then we can think if there's a more robust solution.

jaredlockhart (changes requested):
Overall looks good, good test coverage, clear changes, just two small thoughts 🙏

mikewilli (commented):
I just want to clarify what we're doing with 3 days retention:
1. append 3 days retention to the OVERALL results data (added previously)
2. remove all 3 days retention DAILY results, append only the 4th day (added in this PR)

So I guess what I'd ask is: since we already appended the result we want to OVERALL, can we just skip 3 days retention for other periods and use the "OVERALL" results in the UI? I no longer know enough about the UI logic to say if this is possible, but that was the intention behind adding it to OVERALL, at least. And if the UI doesn't use the OVERALL 3 days retention, then we should consider removing that code alongside this change.

Related to these questions: how do we handle the weekly retention? It's doing the basically same thing (pulling out week 2 and putting it in the OVERALL results data).

moibra05:
1. Well experiments in the observation phase don’t have anything in the overall window so I can’t rely on that to pull 3dr data. that basically leaves parsing the daily data and grabbing day 4 as the only way to correctly show 3 day retention when there’s no overall data. i was initially doing this by using several one off if statements to handle 3dr in each step of the results data formatting pipeline which felt needlessly complicated as i could simply just not include any window index i wasnt using directly from the jetstream client. the way the ui decides the default window for all metrics in one place [here](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/nimbus_ui/views.py#L800-L806), so mixing different analysis windows isn’t really supported. That means we still need overall 3dr when that’s the correct window
2. I am still a little unclear on how we want to be handling retention. Are we expecting to keep the daily retention view, but then have weekly show as unavailable until week 2 data comes in, and only then have overall reflect that?

mikewilli:
1. Ah, I see. Sounds like it would require some substantial changes to handle it the way I was thinking, or at least a workaround that isn't necessarily better than what you're doing here. Makes sense, thanks!
2. Ok, I thought this might be a good example to follow here since that is an existing metric with a similar flow. But since that one is also unclear, let's discuss separately as part of https://github.com/mozilla/experimenter/issues/12647.

Appreciate the explanations. Seems like I'll just need to adjust my thinking here based on how the new UI works.

experimenter/experimenter/nimbus_ui/templates/common/metric_popout.html
moibra05: i've got something in mind as an alternative approach that i think makes most sense to try out here #12647

experimenter/experimenter/jetstream/models.py
```diff
@@ -77,6 +77,7 @@ class Group(StrEnum):
     Group.SEARCH: SEARCH_METRICS,
     Group.USAGE: USAGE_METRICS,
 }
+RETENTION_3_DAYS_WINDOW_INDEX = 4
```
jaredlockhart, line 80: Oh you know we should pull this and a couple other constants littered around the jetstream package out into a proper constants file so they're not just littered around.

experimenter/experimenter/nimbus_ui/templates/common/metric_popout.html
```diff
@@ -123,7 +123,7 @@ <h2 class="fs-6 fw-bold">Weekly breakdown</h2>
           {% endif %}
         {% endwith %}
         {% with daily_metric_data=all_daily_metric_data|dict_get:metric_info.slug %}
-          {% if daily_metric_data.has_daily_data %}
+          {% if daily_metric_data.has_daily_data and not metric_info.slug == NimbusConstants.RETENTION_3_DAYS %}
```
jaredlockhart: Oh sorry I see to use it in the template, the NimbusConstants are already available in the experiment instance like `experiment.RETENTION_3_DAYS`

jaredlockhart (changes requested):

experimenter/experimenter/jetstream/models.py
jaredlockhart: Not now I mean later

jaredlockhart (approved):
Yes yes yes, very nice looks great thnx @moibra05 🙏 🎉

PR #15319, opened 2026-04-16
feat(nimbus): Summary page improvements

### LINKED ISSUE #15316: SRM improvements on summary page
The the p-value icon could be made more clear in a few ways:
it's a bit unclear that you will get more info by hovering on this because it kind of blends in with the link to live monitoring, so it seems like the "info" is at the link
the 4.72e-07 doesn't strike me as too useful, maybe we should just have a threshold after which we just say p-value < 0.0001 , or have less precision, or something else?
it's unclear which direction the threshold is: might be helpful to say something more specific like "p-value 0.0001 (threshold: <0.001)" -- i guess you can figure this out from context, but i think it's worth the one extra character to clarify
(nit) we use little round icons everywhere else so this square one feels out of place
 we should add an alert that's like 'total enrollments unexpectedly low after N days
Also the borders are a bit wonky on the monitoring cards

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6790)

yashikakhurana:
Because

- The p-value tooltip icon blended into the dashboard link, making it unclear it was hoverable
- Scientific notation (e.g. 4.72e-07) was not user-friendly and the threshold direction was ambiguous
- The square ℹ️ emoji was inconsistent with the round icons used elsewhere in the UI
- Nested .card elements were causing double borders on the monitoring cards
-  We should add a days after which unernollment spikes happened to clear things up in the alert

This commit

- Replaces ℹ️ with fa-circle-question and moves it into the section header, clearly separated from the dashboard link
- Formats small p-values as < 0.0001 and clarifies threshold direction: p-value < 0.0001 (threshold: < 0.001)
- Removes nested .card wrappers to fix double borders
- Adds MONITORING_ALERT_MINIMUM_DAYS in the alert message

Fixes #15316

### DIFF
--- a/experimenter/experimenter/nimbus_ui/constants.py
+++ b/experimenter/experimenter/nimbus_ui/constants.py
@@ -45,7 +45,7 @@ Optional - We believe this outcome will <describe impact> on <core metric>
     )
     MONITORING_SRM_HEALTHY_DETAIL = "Branches are at the expected distribution."
     MONITORING_UNENROLLMENT_HEALTHY_DETAIL = (
-        "Unenrollment rate is within a healthy range — below the alert threshold."
+        "within a healthy range — below the alert threshold."
     )
 
     EXCLUDING_EXPERIMENTS_WARNING = """The following experiments are being excluded by
--- a/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail_card_monitoring.html
+++ b/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/detail_card_monitoring.html
@@ -52,7 +52,9 @@
         </thead>
         <tbody>
           <tr class="table-secondary">
-            <td><strong>Total</strong></td>
+            <td>
+              <strong>Total</strong>
+            </td>
             <td class="text-end">{{ summary.total_enrollments }}</td>
             <td class="text-end">{{ summary.total_unenrollments }}</td>
             <td></td>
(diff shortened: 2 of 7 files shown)


yashikakhurana:
Screenshot 2026-04-16 at 10 30 17 AM

jaredlockhart:
Oh good yeah, but you can probably just move those titles

image
image

right inside the coloured alert boxes instead of putting them above?

yashikakhurana:
@jaredlockhart here 
Screenshot 2026-04-16 at 12 18 46 PM

jaredlockhart:
@yashikakhurana Yeah exactly but now those two lines are redundant, just merge them into one 🙏

yashikakhurana:
@jaredlockhart  you mean like this? Screenshot 2026-04-16 at 1 57 36 PM

jaredlockhart:
Looks the same to me?

No I mean

✅ Unenrollment
Unenrollment rate is within healthy range....

Is just saying the same thing twice it just needs to be

✅ Unenrollment rate is within healthy range....

one line with the emoji and information and link

same for the other one

yashikakhurana:
ohh I see, here 
Screenshot 2026-04-16 at 4 01 45 PM

jaredlockhart (approved):
Beautiful!! Looks amazing thank you @yashikakhurana 🙏 ❤️ 🎉

PR #15329, opened 2026-04-16
fix(nimbus): render ConditionalExpression and ObjectLiteral nodes as jexl

### LINKED ISSUE #15328: fix(nimbus): conditional (ternary) expressions render as Python AST repr on detail page
## Observed

On the experiment detail page, the **Full targeting expression** formatted view leaks the Python AST repr of `ConditionalExpression` nodes into the output. Example:

```
browserSettings.update.channel == "release" &&
(
  experiment.slug in activeRollouts ||
  (
      !(hasActiveEnterprisePolicies) &&
      ConditionalExpression(test=BinaryExpression(operator=Operator('in'), left=Literal(value='e6eb0d1e856335fc'), right=Transform(name='mapToProperty', args=[Literal(value='id')], subject=Identifier(value='attachedFxAOAuthClients', subject=None, relative=False))), consequent=BinaryExpression(operator=Operator('>'), left=Identifier(value='lastAccessedDaysAgo', subject=FilterExpression(expression=BinaryExpression(operator=Operator('=='), left=Identifier(value='id', subject=None, relative=True), right=Literal(value='e6eb0d1e856335fc')), subject=Identifier(value='attachedFxAOAuthClients', subject=None, relative=False), relative=True), relative=False), right=Literal(value=29)), alternate=Literal(value=True)) &&
      ...
```

The raw targeting that produced this contains a ternary:

```
("e6eb0d1e856335fc" in attachedFxAOAuthClients|mapToProperty("id")) ? (attachedFxAOAuthClients[.id == "e6eb0d1e856335fc"].lastAccessedDaysAgo > 29) : true
```

## Cause

`format_jexl` in [experimenter/experiments/jexl_utils.py](experimenter/experimenter/experimenter/experiments/jexl_utils.py) walks the JEXL AST to produce the pretty-printed output, but the inner `node_to_str` and `format_node` helpers have no branch for `ConditionalExpression`. The node falls through to `str(node)`, which emits the dataclass repr.

The sibling `to_str` function in the same module does handle `ConditionalExpression`; `format_jexl` just missed it when it was added.

## Expected

`ConditionalExpression` should be rendered as `(test ? consequent : alternate)` (or an appropriately formatted multi-line variant).

jaredlockhart:
Because

* The Full targeting expression panel on the experiment detail page leaks the Python AST repr of `ConditionalExpression` (ternary) nodes into the output because `format_jexl`'s `node_to_str` / `format_node` helpers never learned how to serialize them back to jexl.
* A relative `Identifier` inside a filter (e.g. `things[.id == "x"]`) also leaks a literal `".{subject}"` into the output because `node_to_str` uses a plain string instead of an f-string when prepending the relative-access dot.
* `ObjectLiteral` nodes likewise fall through to `str(node)` since `node_to_str` never handles them.

This commit

* Adds a `ConditionalExpression` branch to `format_node` that renders `(test ? consequent : alternate)` using recursion so children with their own nested `&&` / `||` formatting still come out correctly.
* Fixes the relative `Identifier` prefix to build the dot-prefixed subject in a separate expression, matching the style of `to_str`.
* Adds an `ObjectLiteral` branch to `node_to_str` that serializes `{key: value, ...}` using the dict's string keys and recursing for values.
* Adds regression tests covering each case, plus one mirroring the real bug-report expression so the ternary + relative identifier interaction is locked in.

Fixes #15328

### DIFF
--- a/experimenter/experimenter/experiments/jexl_utils.py
+++ b/experimenter/experimenter/experiments/jexl_utils.py
@@ -102,11 +102,18 @@ def format_jexl(expression):
     def node_to_str(node):
         if isinstance(node, Identifier):
             subject = f"{node_to_str(node.subject)}." if node.subject else ""
-            return f"{'.{subject}' if node.relative else subject}{node.value}"
+            prefix = f".{subject}" if node.relative else subject
+            return f"{prefix}{node.value}"
         if isinstance(node, Literal):
             return json.dumps(node.value)
         if isinstance(node, ArrayLiteral):
             return f"[{', '.join(node_to_str(a) for a in node.value)}]"
+        if isinstance(node, ObjectLiteral):
+            items = ", ".join(
+                f"{k}: {node_to_str(v) or format_node(v, False, 0)}"
+                for k, v in node.value.items()
+            )
+            return f"{{{items}}}"
         if isinstance(node, Transform):
             args_list = [node_to_str(a) or format_node(a, False, 0) for a in node.args]
             args_str = ", ".join(args_list)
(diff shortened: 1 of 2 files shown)


yashikakhurana (approved):
ahh, nice catch, thank you @jaredlockhart

PR #15336, opened 2026-04-17
fix(nimbus): fix metric significance not displaying for overall results

### LINKED ISSUE #15335: Fix metric significance displaying incorrectly
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6797)

moibra05:
Because

- Changes were recently made to fix 3-day retention from displaying misleading values for experiments with only daily results
- Part of these changes included altering how we mapped metric significance to rely on the `window_index` field
- This field is sometimes missing particularly for data points in the overall window causing us to always incorrectly fall back to neutral significance even when the relevant data is there

This commit

- Adds a fixed `window_index: "1"` value inside of all points in the overall window

Fixes #15335


mikewilli (commented):
It seems to me that this might be a problem better fixed at ingestion time. It looks like we are intending to store the significance with the window index regardless of analysis period: https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/jetstream/models.py#L310-L321, so if that's not happening then maybe that's the thing to fix?

mikewilli:
Also, while we shouldn't ever really be in a situation where we don't know the significance, maybe it's worth adding a `UNKNOWN` to the enum as a fallback so we aren't misrepresenting the results?

jaredlockhart:
So then we just rely on the position of the data to know it's the right one?

moibra05:
> So then we just rely on the position of the data to know it's the right one?

cant do that because there's only 1 point in 3dr and it's window_index is 4

mikewilli (approved):
LGTM!

PR #15351, opened 2026-04-17
fix(nimbus): fix retention displaying incorrect values

### LINKED ISSUE #12647: Do not fall back to week-1 retention when week-2 is not available

moibra05:
Because

- We were not showing week 2 retention on the main page for experiments that ran for less than 2 weeks and for those still in the observation phase

This commit

- Adds an analysis error if an experiment was running for less than 2 weeks and was thus unable to get week 2 retention data
-  Applies a similar approach to #15287 where we only keep the second window index in the weekly analysis basis for retention so that we are not displaying incorrect values

Fixes #12647

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -280,9 +280,9 @@ def get_experiment_data(experiment: NimbusExperiment):
                 raw_data[window][AnalysisBasis.EXPOSURES] = {}
 
         for segment, segment_data in segment_points_enrollments.items():
-            data = raw_data[window][AnalysisBasis.ENROLLMENTS][segment] = JetstreamData(
-                segment_data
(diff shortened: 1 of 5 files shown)


moibra05:
Once merged, this unblocks #15239

jaredlockhart (approved):
Okay sorry for not getting back to this, let's just merge it so at least it's functioning, and if we find we need to change its behaviour we can do that as a followup.  Thanks for fixing this and reminding me to review 🙈 🙏

PR #15376, opened 2026-04-20
fix(nimbus): display results conditional on availability instead of experiment phase

### LINKED ISSUE #15341: Display results if available during enrollment
We discovered that results are already being ingested during enrollment, now we need to make the results page available conditional on results being present as opposed to experiment stage

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6798)

moibra05:
Because

- Results were conditionally showing based on experiment stage

This commit

- Changes results to appear as long as they exist

Fixes #15341


jaredlockhart (approved):
🎉 🎉 🎉 🎉  thnx @moibra05 🙏

PR #15406, opened 2026-04-21
feat(nimbus): add validation to nimbus review serializer for targeting context fields within version range

### LINKED ISSUE #14925: Validate targeting configs against targeting context fields within experiment version range
When a user creates an experiment, they select an application, a version range (min/max), and a targeting config. We should validate that the JEXL targeting expression only references targeting context fields that exist within the selected version range, similar to how we validate feature configs against versioned feature definitions.

Follow the pattern of feature config version-range validation:

* NimbusFeatureConfig.get_versioned_schema_range() queries versioned schemas between min/max version
* _validate_feature_configs() in NimbusReviewSerializer checks feature compatibility within the version range
* Unsupported versions produce errors or warnings depending on warn_feature_schema

For targeting configs:

* Use the versioned targeting context files fetched by the manifesttool (ticket 1) and parsed by the targeting context parser (ticket 2)
* When an experiment is submitted for review, extract the identifiers from the targeting config's JEXL expression
* For the selected application and version range, check that all referenced fields exist in the targeting context for every version in the range
* If a field exists in some versions but not others, report which versions are unsupported (similar to ERROR_FEATURE_CONFIG_UNSUPPORTED_IN_VERSIONS)
* Respect the warn_feature_schema flag for error vs warning behavior

Key files to reference:

* Serializer validation: experiments/api/v5/serializers.py (_validate_feature_configs, _validate_feature_value)
* Version range query: experiments/models.py (get_versioned_schema_range, schemas_between_versions)
* Constants: experiments/constants.py (MIN_VERSIONED_FEATURE_VERSION, error messages)

Acceptance Criteria:

* When an experiment is submitted for review, the targeting config JEXL is validated against the targeting context fields available in the selected version range
* If a targeting field is missing in some versions within the range, an error (or warning) identifies the unsupported versions
* If a targeting field is missing in all versions within the range, an error (or warning) reports the field is unsupported in the range
* Validation respects the existing warn_feature_schema flag
* Existing tests are extended or new tests cover version-range targeting validation

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6620)

moibra05:
Because

- Targeting configs were not validated against versioned targeting context fields, allowing JEXL expressions to reference fields that may not exist across the selected version range

This commit

- Adds version-range validation for targeting config JEXL expressions during experiment review
- Extracts referenced identifiers from the JEXL expression and validates them against targeting context fields for all versions in the selected range
- Reports unsupported fields with specific version breakdowns when they are only partially available

Fixes #14925

### DIFF
--- a/experimenter/experimenter/experiments/api/v5/serializers.py
+++ b/experimenter/experimenter/experiments/api/v5/serializers.py
@@ -22,7 +22,10 @@ from experimenter.experiments.constants import (
     NimbusConstants,
     TargetingMultipleKintoCollectionsError,
 )
-from experimenter.experiments.jexl_utils import JEXLParser
+from experimenter.experiments.jexl_utils import (
+    JEXLParser,
+    extract_targeting_fields,
+)
 from experimenter.experiments.models import (
     NimbusBranch,
     NimbusBranchFeatureValue,
@@ -34,6 +37,7 @@ from experimenter.experiments.models import (
     NimbusVersionedSchema,
 )
 from experimenter.features.manifests.nimbus_fml_loader import NimbusFmlLoader
+from experimenter.targeting.targeting_context_parser import TargetingContextFields
 
 logger = logging.getLogger()
 
@@ -1811,6 +1815,77 @@ class NimbusReviewSerializer(serializers.ModelSerializer):
 
         return data
 
+    def _validate_targeting_configs(self, data):
+        targeting_config_slug = data.get("targeting_config_slug")
+        targeting_config = NimbusExperiment.TARGETING_CONFIGS[targeting_config_slug]
+        result = self.ValidateFeatureResult()
+
+        if not targeting_config.targeting:
+            return data
+
+        min_version = None
+        max_version = None
+        assume_unversioned = False
+        if not NimbusExperiment.Application.is_web(self.instance.application):
+            raw_min_version = data.get("firefox_min_version", "")
+            raw_max_version = data.get("firefox_max_version", "")
+
+            min_version = NimbusExperiment.Version.parse(raw_min_version)
+            if raw_max_version:
+                max_version = NimbusExperiment.Version.parse(raw_max_version)
+
+        if min_supported_version := NimbusConstants.MIN_VERSIONED_FEATURE_VERSION.get(
+            data.get("application")
+        ):
+            min_supported_version = NimbusExperiment.Version.parse(min_supported_version)
+
+            if min_supported_version > min_version:
+                if max_version is not None and min_supported_version > max_version:
+                    assume_unversioned = True
+                elif max_version is None or min_supported_version < max_version:
+                    min_version = min_supported_version
+
+        extracted_root_fields = extract_targeting_fields(targeting_config.targeting)
+
+        versions = (
+            NimbusFeatureVersion.objects.filter(
+                NimbusFeatureVersion.objects.between_versions_q(min_version, max_version)
+            )
+            .order_by("-major", "-minor", "-patch")
+            .distinct()
+        )
+
+        unsupported_versions = defaultdict(list)
+        for version in versions:
+            targeting_context_version = None if assume_unversioned else f"v{version}"
+            fields = set(
+                TargetingContextFields.for_application(
+                    data.get("application"), targeting_context_version
+                )
+            )
+
+            unknown_fields = extracted_root_fields - fields
+
+            for field in unknown_fields:
+                unsupported_versions[field].append(str(version))
+
+        for field, versions in unsupported_versions.items():
+            result.append(
+                NimbusConstants.ERROR_TARGETING_FIELD_UNSUPPORTED_IN_VERSIONS.format(
+                    field=field,
+                    versions=versions,
+                ),
+                data.get("warn_feature_schema", False),
+            )
+
+        if result.errors:
+            raise serializers.ValidationError({"targeting_config_slug": result.errors})
+
+        if result.warnings:
+            self.warnings["targeting_config_slug"] = result.warnings
+
+        return data
+
     def validate(self, data):
         application = data.get("application")
         channel = data.get("channel")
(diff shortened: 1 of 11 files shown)


experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1811,6 +1812,85 @@ def _validate_firefox_labs(self, data):
...
+
+        extracted_root_fields = set()
+        for subexpr in collect_exprs(targeting_config.targeting):
+            try:
+                json.loads(subexpr)
+                continue
+            except json.JSONDecodeError:
+                if subexpr.startswith("."):
+                    continue
+                extracted_root_fields.add(subexpr.partition(".")[0])
```
jaredlockhart: Oh then we should pull this out into a utility somewhere since we're also doing it here:

https://github.com/mozilla/experimenter/blob/d34c14c24177b2e6f20e8e34b826a052ca0750f4/experimenter/experimenter/targeting/tests/test_targeting_configs.py#L69-L80

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1811,6 +1812,85 @@ def _validate_firefox_labs(self, data):
...
+        unsupported_versions = defaultdict(list)
+        for version in versions:
+            targeting_context_version = None if assume_unversioned else f"v{version}"
+            fields = set(
+                TargetingContextFields.for_application(
+                    data.get("application"), targeting_context_version
+                )
+            )
+
+            unknown_fields = extracted_root_fields - fields
```
jaredlockhart, line 1867: Set logic is the greatest thing ever 💯 💯 💯

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -4404,6 +4404,139 @@ def test_fml_validate_feature_versioned_unbounded_range_valid(self):
...
+        )
+
+        with (
+            patch(
+                "experimenter.experiments.api.v5.serializers.collect_exprs",
+                return_value=["knownField", "unknownField", ".id"],
+            ),
+            patch(
+                "experimenter.experiments.api.v5.serializers.TargetingContextFields.for_application",
+                return_value={"knownField"},
```
jaredlockhart: I'm not sure mocking out the function calls here is the right way to do this.  Usually we follow a pattern of having test fixture data and pointing the loader at it and forcing it to reload and then letting all the logic flow naturally.  I'd say see if you can do that rather than mocking it out at the function level.

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -4404,6 +4404,139 @@ def test_fml_validate_feature_versioned_unbounded_range_valid(self):
...
+            context={"user": self.user},
+        )
+
+        def fields_for_version(_app, version=None):
+            fields_by_version = {
+                "v120.0.0": {"knownField"},
+                "v121.0.0": {"knownField", "anotherKnownField"},
+                "v122.0.0": {"knownField", "anotherKnownField", "thirdKnownField"},
+                "v123.0.0": {"knownField"},
+            }
```
jaredlockhart: Yeah exactly like we could put these in fixture files and have the loader pick them by mocking out the file path.

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -4404,6 +4404,139 @@ def test_fml_validate_feature_versioned_unbounded_range_valid(self):
...
+            return fields_by_version.get(version, {"knownField"})
+
+        with (
+            patch(
+                "experimenter.experiments.api.v5.serializers.collect_exprs",
+                return_value=["knownField"],
```
jaredlockhart: And then we should be able to setup some dummy test targeting configs.

jaredlockhart (changes requested):
This is so so so sick I'm so pleased with this.  Very well written, very clear, very idiomatic with the project patterns, very clearly and well tested.

Try to do the test fixture file approach.  If it's a nightmare then we can fall back to mocking but it's worth at least trying to set it up now.

Great job @moibra05 🙏 🎉

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
moibra05: I couldnt think of a great way to create mock targeting configs. my only idea was defining one alongside the real production configs in the constants file, but that felt like it would unnecessarily pollute the global registry . im thinking it might be nicer approach to just mock the output of the targeting field extractor directly, rather than trying to mock the entire targeting config so that the test stays focused without needing any larger changes

jaredlockhart (approved):
Yep looking great.  Good test fixture files.  And yeah interesting point about mock targeting configs maybe we can think later about a better pattern there.  Thnx for thinking it through.  But yeah this looks great, great job 🎉 🙏

experimenter/experimenter/experiments/tests/api/v5/test_serializers/test_nimbus_ready_for_review_serializer.py
```diff
@@ -3371,11 +3375,13 @@ def test_invalid_targeting_expr(self):
         )
 
 
+@mock_targeting_manifests
```
jaredlockhart, line 3378: Yeah exactly 🙏

experimenter/experimenter/experiments/tests/api/v5/test_serializers/__init__.py
```diff
@@ -0,0 +1,7 @@
+from pathlib import Path
+
+from django.test import override_settings
+
+mock_targeting_manifests = override_settings(
+    FEATURE_MANIFESTS_PATH=(Path(__file__).parent.absolute() / "fixtures"),
+)
```
jaredlockhart, line 7: 💪 🤘

PR #15408, opened 2026-04-22
fix(nimbus): include region, language, locale in overlap warnings

### LINKED ISSUE #15405: Include region, language, locale in enrollment overlap calculations
If you have two different experiments/rollouts with identical targeting criteria *except* region, language, and locale you may get one of the following warnings:

- The following experiments are LIVE on a previous namespace and may reduce the eligible population for your experiment which may result in reduced statistical power and precision. Please check that the configured population proportion has accounted for this:
- The following multi-feature experiments are LIVE and may reduce the eligible population for your experiment which may result in reduced statistical power and precision. Please check that the configured population proportion has accounted for this: ...
- WARNING: A rollout already exists for this combination of application, feature, channel, and advanced targeting! If this rollout is launched, a client meeting the advanced targeting criteria will be enrolled in one and not the other and you will not be able to adjust the sizing for this rollout.

We should take these fields into account when computing overlap and not show these warnings in these cases.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6816)

jaredlockhart:
Because

* Two experiments or rollouts with identical targeting except for region, language, or locale trigger audience-overlap warnings even though their target populations cannot overlap.
* This drives reviewers to adjust population sizing that does not actually need adjusting on the previous-namespace, multi-feature, and duplicate-rollout warnings.

This commit

* Adds `NimbusExperiment.audiences_overlap`, computing per-dimension intersection across locales, countries, and languages while honoring the `exclude_*` booleans.
* Uses it as a post-filter on `live_experiments_in_namespace`, `feature_has_live_multifeature_experiments`, and `rollout_conflict_warning` so disjoint audiences no longer raise the warning.
* Leaves `bucket_namespace` unchanged so allocation for live experiments does not shift.

Fixes #15405

### DIFF
--- a/experimenter/experimenter/experiments/models.py
+++ b/experimenter/experimenter/experiments/models.py
@@ -1653,6 +1653,33 @@ class NimbusExperiment(NimbusConstants, TargetingConstants, FilterMixin, models.
             return not other_set.issubset(self_set)
         return not self_set.issubset(other_set)
 
+    def _audience_overlapping_candidates(self, candidates):
+        self_locales = [locale.code for locale in self.locales.all()]
+        self_countries = [country.code for country in self.countries.all()]
+        self_languages = [language.code for language in self.languages.all()]
+        return [
+            candidate
+            for candidate in candidates
+            if self._audience_dimension_overlap(
+                self_locales,
+                self.exclude_locales,
+                candidate["locale_codes"],
+                candidate["exclude_locales"],
+            )
+            and self._audience_dimension_overlap(
+                self_countries,
+                self.exclude_countries,
+                candidate["country_codes"],
+                candidate["exclude_countries"],
+            )
+            and self._audience_dimension_overlap(
+                self_languages,
+                self.exclude_languages,
+                candidate["language_codes"],
+                candidate["exclude_languages"],
+            )
+        ]
+
     @property
     def feature_has_live_multifeature_experiments(self):
         matching = []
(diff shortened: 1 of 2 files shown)


experimenter/experimenter/experiments/models.py
```diff
@@ -1638,6 +1639,60 @@ def allocate_bucket_range(self):
             ),
         )
 
+    @staticmethod
+    def audience_set_overlap(self_items, self_exclude, other_items, other_exclude):
+        self_set = set(self_items or ()) - {None}
+        other_set = set(other_items or ()) - {None}
```
freshstrangemusic, line 1651: Whats with `- {None} `?

experimenter/experimenter/experiments/models.py
```diff
@@ -1638,6 +1639,60 @@ def allocate_bucket_range(self):
             ),
         )
 
+    @staticmethod
+    def audience_set_overlap(self_items, self_exclude, other_items, other_exclude):
```
freshstrangemusic: I think type annotations would be very helpful here.

experimenter/experimenter/experiments/models.py
```diff
@@ -1638,6 +1639,60 @@ def allocate_bucket_range(self):
...
+            return True
+        if not self_exclude and not other_exclude:
+            return bool(self_set & other_set)
+        if self_exclude and other_exclude:
+            return True
+        if self_exclude:
+            return not other_set.issubset(self_set)
+        return not self_set.issubset(other_set)
+
+    def audience_overlap(self, candidates):
```
freshstrangemusic: And here too

freshstrangemusic (approved):

experimenter/experimenter/experiments/models.py
jaredlockhart: It's cuz the query annotation can return some Nones so they'll sneak into this list so this is just a handy dandy way to strip them out.

jaredlockhart: Often you might write it as something like

```py
things = [thing for thing in things if thing]
```

but `- {None}` does the same without a needless comprehension.

yashikakhurana: I wasn't aware of this that we can do this

jaredlockhart: Yeah it's a handy trick but only if you're using `set`s, if you're using a list still gotta just use the comprehension.

PR #15413, opened 2026-04-22
feat(nimbus): Reset slack notification state when cloning

### LINKED ISSUE #15412: Fix notifications state when we clone the experiment
It seems like in case of clone we are also cloning the notifications state too, so if its was turned off in the parent, then its turned off in cloned one too.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6817)

yashikakhurana:
Because

- By default we keep the slack notification state enabled, but when we clone the experiment we are copying the state from the parent, whereas we should reset the state and turned the slack notification by default.

This commit
-  Turns on the default state for the newly cloned experiment

Fixes #15412


moibra05 (approved):
lgtm 🎆

PR #15430, opened 2026-04-23
fix(nimbus): update targeting field validation to use additional preserved fields

### LINKED ISSUE #15429: Targeting field validation isnt considering the preserved targeting keys
We recently began validating targeting configs against targeting context fields within experiment version range but are not using the preserved targeting keys causing some experiments to be incorrectly deemed invalid

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6825)

moibra05:
Because

- When we added validation for targeting configs against targeting context fields within the experiment version range, we didn’t include preserved targeting fields in the set of fields being validated
- This caused some experiments to incorrectly show errors during review

This commit

- Adds preserved targeting fields to the set of fields validated per application

Fixes #15429


yashikakhurana (approved):
nice job, thank you @moibra05

PR #15433, opened 2026-04-23
feat(nimbus): Grafana proxy view

### LINKED ISSUE #15342: Show Grafana metric graphs on experimenter feature page
include [https://yardstick.mozilla.org/d/dtfz7xv/nimbus-feature-monitoring](https://yardstick.mozilla.org/d/dtfz7xv/nimbus-feature-monitoring) as an iframe on [https://experimenter.services.mozilla.com/nimbus/features/](https://experimenter.services.mozilla.com/nimbus/features/) in a way that allows specifying grafana template variables like rollout, metrics, dimensions, etc

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6730)

yashikakhurana:
Because

-Grafana is behind Google IAP, which sets session cookies on` yardstick.mozilla.org.` Browsers block these as third-party cookies when the dashboard is embedded in an iframe on a different origin, causing it to fail to load.

This commit

- Adds a server-side GrafanaProxyView that fetches Grafana content via internal k8s DNS (grafana.grafana-prod.svc.cluster.local:8080), bypassing IAP entirely
- Rewrites <base href> and appUrl in HTML responses so Grafana's assets and API calls also route through the proxy
- Adds `GRAFANA_INTERNAL_URL` and `GRAFANA_SERVICE_ACCOUNT_TOKEN` settings for the internal endpoint and service account auth
- Adds feature_monitoring_proxy_path on NimbusFeatureConfig to extract the path/query from the monitoring URL
Updates the feature monitoring iframe to use `/nimbus/grafana-proxy/ `instead of the direct Grafana URL

Fixes #15342 
https://github.com/mozilla/webservices-infra/pull/10728


jaredlockhart (dismissed):
Oh that's a clever idea!  I think there's a bunch to talk through on this so let's chat about it on Monday.

yashikakhurana:
Blocking this now as we discussed to do a security review first before adding this

yashikakhurana:
cleared from the security review, got approval

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -1143,6 +1147,67 @@ class NimbusFeatureMonitoringView(DetailView):
...
 
+# Grafana is behind Google IAP, which blocks third-party cookies in iframes.
+# Both Experimenter and Grafana run in the same k8s cluster, so the backend
+# can reach Grafana via internal DNS (bypassing IAP) and proxy it to the browser.
+_GRAFANA_PROXY_PREFIX = "/nimbus/grafana-proxy/"
+_SAFE_RESPONSE_HEADERS = ("Cache-Control", "ETag", "Last-Modified", "Content-Type")
+
+
+class GrafanaProxyView(LoginRequiredMixin, View):
+    def get(self, request, path=""):
```
jaredlockhart: We shouldn't pass a raw url in thorugh the proxy url, that's a security vulnerability.  Let's just pass the slug in and then we can construct the url, and to make sure the slug is valid do it as

```
class GrafanaProxyView(FeatureDetailView):
    model = NimbusFeatureConfig

    def get(self, request):
        response = super().get(request)
         
```

so that way it pre validates that the slug is a valid feature slug, so you can't just inject arbitrary url data into the slug field.


external url:
```
experimenter/grafana_proxy/?slug=some-feature-slug
```

construct in memory:

```
{grafana_root}/{feature_monitoring_dash_path}?orgId=1&from=now-7d&to=now&timezone=utc&var-application=firefox_desktop&var-feature={feature_slug}
```

yashikakhurana:
@jaredlockhart this is ready for review now

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -1137,10 +1140,64 @@ def get_context_data(self, **kwargs):
...
+        # User input (slug) is validated against the DB before use;
+        # the host and dashboard path come from settings only.
+        application = (feature_config.application or "").replace("-", "_")
+        target = "{base}/{path}".format(
+            base=settings.GRAFANA_INTERNAL_URL.rstrip("/"),
+            path=settings.GRAFANA_FEATURE_MONITORING_DASHBOARD_PATH,
+        )
+        params = {
+            "orgId": "1",
+            "from": "now-7d",
```
jaredlockhart: Let's try 30d

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -526,6 +526,34 @@ <h5 class="fw-semibold mt-3 d-flex align-items-center">
...
+              <i class="fa-solid fa-chart-line me-2"></i>
+              Feature Monitoring
+            </h5>
+            <a href="{{ selected_feature_config.feature_monitoring_url }}"
+               target="_blank"
+               rel="noopener noreferrer"
+               class="btn btn-outline-primary btn-sm">
+              <i class="fa-solid fa-arrow-up-right-from-square me-1"></i>
+              Open in Grafana
+            </a>
```
jaredlockhart, line 544: Yeah exactly I think it's better to link front and centre out to external dashboards, people will miss the links if they're in the sidebar.

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -1137,10 +1140,64 @@ def get_context_data(self, **kwargs):
...
+_SAFE_RESPONSE_HEADERS = ("Cache-Control", "ETag", "Last-Modified", "Content-Type")
+
+
+class GrafanaProxyView(LoginRequiredMixin, View):
+    def get(self, request):
+        slug = request.GET.get("slug")
+        if not slug:
+            return HttpResponse("Missing slug parameter", status=400)
+
+        feature_config = get_object_or_404(NimbusFeatureConfig, slug=slug)
```
jaredlockhart, line 1154: Yeah so we can do this like this, where we just receive GET and then naively unpack the slug field and do `get_object_or_404` ourselves.  But you can also mix in something like the [SingleObjectMixin](https://github.com/django/django/blob/main/django/views/generic/detail.py#L8) that does all that for you, and then it'll catch like extra cases and raise additional codes where necessary etc, it just does a lot of the plumbing for you.  This is totally fine to just leave as is I just wanted to highlight ways that Django gives you pieces that does things like this for you that you can easily drop in.

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -1137,10 +1140,64 @@ def get_context_data(self, **kwargs):
...
+            path=settings.GRAFANA_FEATURE_MONITORING_DASHBOARD_PATH,
+        )
+        params = {
+            "orgId": "1",
+            "from": "now-7d",
+            "to": "now",
+            "timezone": "utc",
+            "var-application": application,
+            "var-feature": feature_config.slug,
+            "kiosk": "",
```
jaredlockhart, line 1171: Are we gonna have to pass through the light/dark mode theme? 🤔

experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/features.html
```diff
@@ -526,6 +526,34 @@ <h5 class="fw-semibold mt-3 d-flex align-items-center">
...
+          </div>
+          <div class="card shadow-sm border-0 rounded-3 mb-3">
+            <div class="card-body p-0">
+              <iframe src="/nimbus/grafana-proxy/?slug={{ selected_feature_config.slug }}"
```
jaredlockhart, line 548: So yeah let's land this as a prototype just to show that the graph renders etc.  But next we'll probably have to

* Discover the list of metrics for a feature (which we can discover by pulling in teh feature metric config files that are in metric hub now, they might already even be pulled in via the config update? 🤔 )
* Loop over them and create one iframe per metric
* Pass that metric in through the url
But we an do that as a followup

Or if you can 'select all' by default so we get all the metrics for the feature automatically try that 🙏

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -1137,10 +1140,64 @@ def get_context_data(self, **kwargs):
...
 
-class NimbusFeatureMonitoringView(DetailView):
-    template_name = "nimbus_experiments/feature_monitoring.html"
-    model = NimbusFeatureConfig
-    context_object_name = "feature_config"
+# Grafana is behind Google IAP, which blocks third-party cookies in iframes.
+# Both Experimenter and Grafana run in the same k8s cluster, so the backend
+# can reach Grafana via internal DNS (bypassing IAP) and proxy it to the browser.
+_SAFE_RESPONSE_HEADERS = ("Cache-Control", "ETag", "Last-Modified", "Content-Type")
+
```
jaredlockhart: Does this need to be global? If it's only used there can it just be scoped within the view?

jaredlockhart (approved):
Alright let's give it a shot let's gooooo 🎉 🔥 🎉 🔥 🎉 🔥 🎉 🔥

PR #15454, opened 2026-04-27
fix(nimbus): add missing fenix targeting keys to preserved list

### LINKED ISSUE #15452: fix(nimbus): add missing Fenix runtime-only targeting keys to PRESERVED_TARGETING_KEYS_BY_APPLICATION
[`PRESERVED_TARGETING_KEYS_BY_APPLICATION[FENIX]`](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/targeting/constants.py#L41) is missing three keys that Fenix injects via [`CustomAttributeProvider.getCustomTargetingAttributes`](https://github.com/mozilla-firefox/firefox/blob/main/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/messaging/CustomAttributeProvider.kt#L48-L69) but doesn't emit from `RecordedNimbusContext.kt`. The validator in [`_validate_targeting_configs`](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/api/v5/serializers.py#L1819) currently rejects 50+ live recipes that reference them.

```diff
-    Application.FENIX: {"current_date", "nimbus_id"},
+    Application.FENIX: {
+        "current_date",
+        "nimbus_id",
+        "isFirstRun",
+        "is_large_device",
+        "number_of_app_launches",
+    },
```

Plus a serializer test analogous to `test_validate_targeting_config_uses_preserved_targeting_keys` from #15430, parameterized for Fenix with `MOBILE_FIRST_RUN_USER`.

Broader audit (full enumeration, iOS defensive entries) tracked in #15451.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6836)

moibra05:
Because

- We discovered that targeting keys are being sourced from multiple locations where there is some overlap of keys with the same name but with different casing (e.g. `isFirstRun` vs `is_first_run`)
- These inconsistencies cause certain valid targeting fields to be missed during validation, since casing differences prevent them from matching expected keys causing some experiments to incorrectly surface validation errors despite technically using valid targeting fields

This commit

- Adds several missing keys to the persevered targeting field list
- Adds tests for mobile experiments using preserved keys

Fixes #15452

### DIFF
--- a/experimenter/experimenter/experiments/tests/api/v5/test_serializers/fixtures/fenix/RecordedNimbusContext.kt
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_serializers/fixtures/fenix/RecordedNimbusContext.kt
@@ -1,0 +1,6 @@
+val obj = JSONObject(
+    mapOf(
+        "knownField" to knownField,
+    ),
+)
+return obj
--- a/experimenter/experimenter/experiments/tests/api/v5/test_serializers/fixtures/fenix/v120.0.0/RecordedNimbusContext.kt
+++ b/experimenter/experimenter/experiments/tests/api/v5/test_serializers/fixtures/fenix/v120.0.0/RecordedNimbusContext.kt
@@ -1,0 +1,6 @@
+val obj = JSONObject(
+    mapOf(
+        "knownField" to knownField,
+    ),
+)
+return obj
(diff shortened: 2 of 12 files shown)


jaredlockhart (approved):
Awesome ty @moibra05 🙏

PR #15455, opened 2026-04-27
feat(nimbus): For complete experiment do not update monitoring data as unenrollment spikes

### LINKED ISSUE #15441: Should not update monitoring date for the ended experiment
, when the experiment ends clients will all start unenrolling so now every summary page shows 'unexpectedly high unenrollment' after the experiment ends becuase of course all the clients are unenrolling at the end, so we should cut it off to like last day -1 or just hide the monitoring cards altogether after it ends or something so we don't give people the sense that something went wrong when it's just the normal flow at the end of an experiment we can do is stop notifying or updating the experiment monitoring data once the experiment is ended, if we stop updating it, it will never notify it

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6832)

yashikakhurana:
Because

- when the experiment ends clients will all start unenrolling so now every summary page shows 'unexpectedly high unenrollment' after the experiment ends because  all the clients are unenrolling at the end, we can do is stop notifying or updating the experiment monitoring data once the experiment is ended, if we stop updating it, it will never notify it

This commit

- Shows last updated time at summary page
- Does not update monitoring data for the completed experiments.

Fixes #15441


jaredlockhart (approved):
Ah yes very nice thnx @yashikakhurana 🙏

PR #15496, opened 2026-05-01
feat(nimbus): Rollouts overview forms

### LINKED ISSUE #15077: Rollout Overview Form/Card
Implement the Overview card for the rollout summary page. Contains name, slug, description, hypothesis, and owner fields.

* Fields: Name, Slug (read-only, click-to-copy), Public description, Hypothesis, Owner/Team.
* Slug is displayed with a copy-to-clipboard button.
* Platform field moves into this card from the old Configuration & customization section.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6681)

yashikakhurana:
Because

- We are working on new design, so adding the overview form on the new rollouts page

This commit

- Adds the new overview form on rollouts page

Fixes #15077

### DIFF
--- a/experimenter/experimenter/nimbus_ui/views.py
+++ b/experimenter/experimenter/nimbus_ui/views.py
@@ -2,7 +2,6 @@ import json
 
 from deepdiff import DeepDiff
 from django.conf import settings
-from django.contrib.auth import get_user_model
 from django.core.paginator import Paginator
 from django.db.models import Q
 from django.http import HttpResponse, HttpResponseRedirect
@@ -33,6 +32,8 @@ from experimenter.nimbus_ui.filtersets import (
     NimbusExperimentsHomeFilter,
     SortChoices,
     StatusChoices,
+    TagSearchFilterSet,
+    UserSearchFilterSet,
 )
 from experimenter.nimbus_ui.forms import (
     ApproveEndEnrollmentForm,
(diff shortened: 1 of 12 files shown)


yashikakhurana:
https://github.com/user-attachments/assets/6323a163-a9a5-4f04-bc5b-781ee4346176

experimenter/experimenter/nimbus_ui/views.py
```diff
@@ -1237,3 +1238,158 @@ class NimbusRolloutDetailView(
...
+    template_name = "nimbus_experiments/overview/tag_search_results.html"
+    m2m_attr = "tags"
+    context_key = "tags"
+
+    def get_search_results(self, q, excluded_ids):
+        return (
+            Tag.objects.filter(name__icontains=q)
+            .order_by("name")
+            .exclude(id__in=excluded_ids)[:15]
+        )
```
jaredlockhart: Doing filters like these could probably use the DjangoFilterSet library like we do on the list page, instead of writing the queries out manually.  Just makes it easier to modify in the future if you don't have to deal with the queries directly, just declare the fields:

https://django-filter.readthedocs.io/en/stable/ref/filterset.html#automatic-filter-generation-with-model

jaredlockhart (approved):
Awesome!!  Looking great already, can't wait to get this live 🤩 thnx @yashikakhurana 🙏  One thought about the filters but we don't need to block on it I don't think we can do it as a followup.

experimenter/experimenter/nimbus_ui/views.py
yashikakhurana: I will do that here

PR #15550, opened 2026-05-07
fix(nimbus): fix improper handling of missing relative data

### LINKED ISSUE #15548: TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
Sentry Issue: [EXPERIMENTER-PROD-1B8](https://mozilla.sentry.io/issues/7465873564/?referrer=github_integration)

```
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
(1 additional frame(s) were not displayed)
...
  File "experimenter/nimbus_ui/views.py", line 869, in get_context_data
    lower = relative[0].get("lower") * 100
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6882)

moibra05:
Because

- A recent PR fixed a recurring [issue](https://github.com/mozilla/experimenter/issues/15396) where missing `relative_uplift` datapoints were not being handled in the `results_manager`
- However the same case was still not being handled correctly in the view layer

This commit

- Adds proper handling of metric ui display properties when we are missing `relative_uplift` data for a particular metric

Fixes #15548

### DIFF
--- a/experimenter/experimenter/nimbus_ui/tests/test_views.py
+++ b/experimenter/experimenter/nimbus_ui/tests/test_views.py
@@ -3366,7 +3366,415 @@ class TestResultsView(AuthTestCase):
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, "nimbus_experiments/results-fragment.html")
 
-    def test_results_view_relative_ui_properties(self):
+    @parameterized.expand(
+        [
+            (
+                {
+                    "v3": {
+                        "overall": {
+                            "enrollments": {
+                                "all": {
+                                    "branch-a": {
+                                        "branch_data": {
+                                            "other_metrics": {
+                                                "urlbar_amazon_search_count": {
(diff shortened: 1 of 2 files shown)


mikewilli (approved):
Seems like a reasonable fix, thanks!

Do you think this is a scenario where Jetstream is passing invalid results, or just a valid scenario we didn't expect? If it's the former, we should probably update the schemas package if possible so that we can validate on both sides (jetstream export and experimenter ingest).

moibra05:
I suspect that its an issue with the jetstream ingestion step b/c what i'm seeing is that a single metric contains a mix of `relative_uplift` data points with all the expected fields and some without. that inconsistency doesn’t seem intentional, which is why i’m leaning toward jetstream being the source of the problem
image

mikewilli:
Hm, ok. I don't think there's much we can do with the schema to validate that, but good to know, thanks!

PR #15575, opened 2026-05-11
fix(nimbus): adjust precision of data points in overview popout cards

### LINKED ISSUE #15491: Fix data point precision in overview popout cards
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6852)

moibra05:
Because

- Data point values on hover in the popout overview cards were needlessly precise

This commit

- Reduce precision to 6 digits

Fixes #15491


mikewilli (approved):
🎉 Thanks!

Does this affect the absolute points too or just percentages?

moibra05:
> 🎉 Thanks!
> 
> Does this affect the absolute points too or just percentages?

both 👍

PR #15619, opened 2026-05-13
fix(nimbus): prevent kinto preview sync from clobbering concurrent user state transitions

### LINKED ISSUE #15618: Preview experiment can lose user state transition due to race with kinto preview sync task
The `nimbus_synchronize_preview_experiments_in_kinto` celery task races against user state transitions on Preview experiments.

In Section A ([experimenter/kinto/tasks.py#L498-L518](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/kinto/tasks.py#L498-L518)), the task:

1. Queries `should_publish_experiments` and hydrates each one in memory with `status=PREVIEW, publish_status=IDLE`.
2. Calls `kinto_client.create_record(data)` (slow network I/O).
3. Inside a `transaction.atomic()`, sets `published_dto` + `published_date` on the **stale in-memory instance** and calls `experiment.save()`, which writes **all fields** back.

If a user clicks "Request Launch" on the same Preview experiment between (1) and (3) — `PreviewToReviewForm` commits `status=DRAFT, status_next=LIVE, publish_status=REVIEW` — the task's `experiment.save()` overwrites those changes back to `status=PREVIEW, publish_status=IDLE`. The user sees the HTMX response render Review controls, then a hard refresh shows the experiment still in Preview.

### Fix

In Section A, re-fetch the row with `select_for_update()` filtered on `status=PREVIEW` inside the transaction. If the experiment is no longer in Preview, skip the DB update — the next sync run reconciles Kinto via Section C. Also narrow the `save()` call with `update_fields=["published_dto", "published_date"]` so we don't write back fields we didn't intend to change.

### Acceptance Criteria

- A concurrent user transition out of Preview during the Kinto push is preserved (not reverted by the task).
- Regression test in `TestNimbusSynchronizePreviewExperimentsInKinto` covers this case.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6889)

jaredlockhart:
Because

* The kinto preview sync task hydrated experiments in memory across slow Kinto network calls, then called experiment.save(), clobbering any concurrent write — most visibly a Request Launch on a Preview experiment, which appeared to succeed in the HTMX response but reverted on refresh
* All three sections (push, expire, unpublish) had the same lost-update shape

This commit

* Replaces experiment.save() with conditional QuerySet.update() in all three sections, gating each write on the status the task observed and writing only the columns it owns
* Adds regression tests for the push and expiry races

Fixes #15618

### DIFF
--- a/experimenter/experimenter/kinto/tasks.py
+++ b/experimenter/experimenter/kinto/tasks.py
@@ -505,10 +505,22 @@ def nimbus_synchronize_preview_experiments_in_kinto():
             kinto_client.create_record(data)
 
             with transaction.atomic():
-                experiment.published_dto = data
-                experiment.published_date = timezone.now()
-                experiment.save()
+                updated = NimbusExperiment.objects.filter(
+                    pk=experiment.pk,
+                    status=NimbusExperiment.Status.PREVIEW,
+                ).update(
+                    published_dto=data,
+                    published_date=timezone.now(),
+                )
 
+                if not updated:
+                    logger.info(
+                        f"{experiment.slug} left Preview during push; "
+                        f"next sync will reconcile Kinto"
+                    )
+                    continue
+
+                experiment.refresh_from_db()
                 generate_nimbus_changelog(
                     experiment,
                     get_kinto_user(),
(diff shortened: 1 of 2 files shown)


yashikakhurana (approved):
oh nice catch, thanks @jaredlockhart

PR #15622, opened 2026-05-13
fix(nimbus): isolate and stream warm_api_caches so v6/v7/v8 actually warm

### LINKED ISSUE #15621: warm_api_caches OOM-killed before v6/v7/v8 caches ever warm
**Problem**

The hourly [`warm_api_caches`](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/tasks.py) Celery beat task is OOM-killed at ~3.5 GB on every run, about 7 seconds after logging `Warmed v5:csv`. Because `v5:csv` is the first item in the warm loop, **v6/v7/v8 are never pre-warmed**. They get populated only when an ad-hoc cache-miss request happens to complete before gunicorn's worker timeout; when the cache is cold, real requests hit the timeout and return `SystemExit: 1` (visible in Sentry on `/api/v{6,7,8}/experiments/`).

GCP logs (`moz-fx-webservices-high-prod`, namespace `experimenter-prod`) show this happening every hour at HH:23:28 for the entire 72h window I checked:

```
17:23:21  Warmed v5:csv (1,805,252 bytes)
17:23:28  Memory cgroup out of memory: Killed process (celery) anon-rss:3594152kB
17:23:30  Started container experimenter   ← pod restart
```

**Proposed change**

1. **Split** the single `warm_api_caches` task into one task per endpoint, dispatched by a parent beat task via `apply_async`. A single endpoint OOMing can no longer prevent the others from warming.
2. **Reduce** the per-endpoint memory footprint in [`warm_api_cache()`](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/api/cache.py) and [`CachedListMixin.list()`](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/experiments/api/cache.py):

- Iterate the queryset with `qs.iterator(chunk_size=N)` instead of materializing it all at once (Django 5.2 supports `iterator(chunk_size=...)` with `prefetch_related`).
- Stream rendered JSON bytes per-experiment to a buffer so the full Python `data` list and the final JSON bytestring never co-exist in memory.
- Free intermediate refs between chunks.

**Acceptance criteria**

- `warm_api_caches` (or its successor) successfully warms all six endpoints — `v5:csv`, `v6:experiments`, `v6:draft-experiments`, `v7:experiments`, `v8:experiments`, `v8:draft-experiments` — within the celery worker pod's memory limit, every hour.
- A failure warming one endpoint does not prevent the others from warming.
- `SystemExit: 1` events on `/api/v{6,7,8}/experiments/` drop to ~0 in Sentry post-deploy.
- Local memory profile shows reduced peak RSS for at least the heaviest endpoint.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6890)

jaredlockhart:
Because

* The hourly `warm_api_caches` Celery task OOM-kills the worker (~3.5 GB) seven seconds after warming v5:csv, so v6/v7/v8 are never pre-warmed
* Cache-miss requests on the cold v6/v7/v8 caches risk timing out at the gunicorn worker limit (`SystemExit:1` on `/api/v{6,7,8}/experiments/` in Sentry)

This commit

* Splits `warm_api_caches` into a dispatcher that fires one sub-task per endpoint, so one failure can't tear down the others
* Streams `warm_api_cache` via `qs.iterator()` + `JSONEncoder.iterencode` so neither the full dict tree nor the full JSON string co-exist in memory. Output byte-identical to the non-streamed render, pinned by a new test
* Adds per-endpoint metrics so a silently-not-warming endpoint is alertable

Local profile (500 factory experiments, v8 viewset): peak RSS 71.75 → 49.62 MB (−31%).

Fixes #15621

### DIFF
--- a/experimenter/experimenter/experiments/api/cache.py
+++ b/experimenter/experimenter/experiments/api/cache.py
@@ -1,12 +1,13 @@
 import hashlib
-import io
 import logging
 from urllib.parse import urlencode
 
 from django.conf import settings
 from django.core.cache import cache
 from django.http import HttpResponse
+from rest_framework.compat import SHORT_SEPARATORS
 from rest_framework.renderers import JSONRenderer
+from rest_framework.utils.encoders import JSONEncoder as DRFJSONEncoder
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


experimenter/experimenter/experiments/api/cache.py
```diff
@@ -20,15 +24,68 @@ def get_api_cache_key(view_name, query_params=None):
...
     if sort_key is not None:
-        qs = sorted(qs, key=sort_key, reverse=True)
-    data = serializer_class(qs, many=True).data
-    rendered = renderer.render(data)
+        qs = sorted(queryset.all(), key=sort_key, reverse=True)
+        data = serializer_class(qs, many=True).data
+        rendered = renderer.render(data)
+    else:
+        rendered = stream_render_queryset(queryset, serializer_class)
+
```
mikewilli, line 88: So will the v5:csv still fail since it has a `sort_key` defined? But it just won't kill the tasks for the other endpoints now?

mikewilli (approved):

experimenter/experimenter/experiments/api/cache.py
jaredlockhart: No v5:csv wasn't failing, it was just succeeding before proceeding to the next ones which were then OOMing.  It should still proceed healthily.

PR #15687, opened 2026-05-20
feat(nimbus): rollout features form

### LINKED ISSUE #15080: Rollout Features Form/Card
Implement the Rollout Features card for the rollout summary page. This is the new Rollout Experience section that replaces the old Configuration & customization section.

* Feature Config selector (dropdown).
* Feature Value input — one input rendered per selected feature.
* JEXL Targeting displayed as a read-only field.
* Recipe JSON displayed as a read-only expandable field.
* Feature config is locked and non-editable once the rollout is Live.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6684)

moibra05:
Because

- The new design requires the features section for the rollout creation form

This commit

- Adds the features section of the rollout form

Fixes #15080

### DIFF
--- a/experimenter/experimenter/nimbus_ui/new/forms.py
+++ b/experimenter/experimenter/nimbus_ui/new/forms.py
@@ -14,10 +14,12 @@ from experimenter.base.models import Country, Language, Locale
 from experimenter.experiments.changelog_utils import generate_nimbus_changelog
 from experimenter.experiments.models import (
     NimbusBranch,
+    NimbusBranchFeatureValue,
     NimbusDocumentationLink,
     NimbusExperiment,
     NimbusExperimentBranchThroughExcluded,
     NimbusExperimentBranchThroughRequired,
+    NimbusFeatureConfig,
     Tag,
 )
 from experimenter.nimbus_ui.constants import NimbusUIConstants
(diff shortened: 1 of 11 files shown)


moibra05:
image
image

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -685,6 +685,101 @@ def get_queryset(self):
...
+            {
+                "hx-post": reverse(
+                    "nimbus-ui-new-update-rollout-features",
+                    kwargs={"slug": self.instance.slug},
+                ),
+                "hx-trigger": "change",
+                "hx-select": "#rollout-rollout-features-body",
+                "hx-target": "#rollout-rollout-features-body",
+            }
+        )
```
yashikakhurana: do we need to do this here? I believe we can do this in html files, would you take a look on the old forms and html files that how its done?

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -685,6 +685,101 @@ def get_queryset(self):
...
+            saved_branch_feature_configs - saved_experiment_feature_configs
+        )
+
+        reference_branch = experiment.reference_branch or experiment.branches.first()
+
+        reference_branch.feature_values.filter(
+            feature_config__in=deleted_feature_configs
+        ).delete()
+
+        for feature_config in new_feature_configs:
```
yashikakhurana: if we can reuse the existing form that would be great

moibra05: using the exisitng form complicates a lot of the logic in there so i thought for the sake of simplicity i'd create a new form for the rollout features

experimenter/experimenter/nimbus_ui/forms.py
moibra05: the old form does this in a [similar way](https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/nimbus_ui/forms.py#L793-L801)

experimenter/experimenter/nimbus_ui/forms.py
yashikakhurana: since we will be reusing this for the experiments too, and it should have same logic in the forms, whats different?

yashikakhurana (approved):

PR #15703, opened 2026-05-22
fix(nimbus): fix disconnect between main page metric results and details modal

### LINKED ISSUE #15696: Results page disconnect between main page metric results and details modal
The main page for https://experimenter.services.mozilla.com/nimbus/10-shortcuts-per-row/results/  is (currently) showing the Week 3 result, but the Overview on the details modal is showing a different result. Presumably the details modal overview shows the `overall` result, though it might be the Week 1 result. If `overall` is available, it should also be shown on the main page. If `overall` is *not *available, then what is the Overview showing?

Main page:

Detail modal:

It’s also confusing because this and other metrics have a discrepancy on the main page vs details modal on whether there is a “notable change” – you can see in the first screenshot that the main page is showing **No notable change**, despite the metric being located inside the Notable Changes section of the results page, and the details modal showing a change in the overview.

Reference: https://mozilla.slack.com/archives/CF94YGE03/p1779213540003239

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6898)

moibra05:
Because

- In the overview section of the metric details modal we were incorrectly showing data for the first window index whilst showing the latest on the main page
- Metrics were deemed notable as long as they had at least 1 window index with significant change  which would cause confusion if the latest index showed neutral change

This commit

- Fixes overview section in details modal so that it displays the last window point instead of the first
- Changes notable metric criteria to only display metrics whose latest window index shows significant change

Fixes #15696

### DIFF
--- a/experimenter/experimenter/jetstream/tests/test_results_manager.py
+++ b/experimenter/experimenter/jetstream/tests/test_results_manager.py
@@ -1,3 +1,5 @@
+from unittest import mock
+
 from django.test import TestCase
 from parameterized.parameterized import parameterized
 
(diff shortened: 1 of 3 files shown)


mikewilli (commented):
Changes make sense, but should there be any accompanying test updates here?

experimenter/experimenter/jetstream/tests/test_results_manager.py
```diff
@@ -1041,6 +1043,174 @@ def test_metric_areas_created_correctly(self):
...
+                                            }
+                                        }
+                                    },
+                                    "branch-b": {
+                                        "branch_data": {
+                                            "other_metrics": {
+                                                "retained": {
+                                                    "significance": {
+                                                        "branch-a": {
+                                                            "overall": {
```
mikewilli, line 1121: Nit but `overall` should never have multiple windows. Maybe better to use `weekly` so we don't confuse ourselves down the line.

mikewilli (approved):
Small suggestion but looks good, thanks!

PR #15787, opened 2026-06-02
feat(project): route Remote Settings tasks to a dedicated celery queue

### LINKED ISSUE #15773: Split celery worker affinity
Right now we lump all celery tasks into a single worker, but there are now many slower tasks for fetching jetstream/monitoring data that run frequently, and also the tasks that just broker interactions with Remote Settings.  Let's split those out so we have two workers: one dedicated to RS changes, and one that handles the slower moving jobs.  That way a slow moving job doesn't block fast turn around for RS reviews.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6927)

jaredlockhart:
Because

* All celery tasks share one worker, so a slow jetstream/monitoring fetch can block fast-turnaround Remote Settings brokering.
* Splitting RS work onto its own queue lets a dedicated worker drain it independently of the slower analysis jobs.

This commit

* Routes all `experimenter.kinto.tasks.*` to a `remote_settings` queue via `CELERY_TASK_ROUTES`; everything else stays on the default `celery` queue.
* Updates the local docker-compose worker to consume both queues so dev keeps processing kinto tasks.

The matching infra change (a second worker container consuming `remote_settings`) lands in webservices-infra and **must deploy first** — otherwise kinto tasks would route to a queue with no consumer. Hold this PR until the RS worker container is live in prod.

Fixes #15773


yashikakhurana (approved):
Lets goo 🎉 , thank you @jaredlockhart

PR #15788, opened 2026-06-02
feat(nimbus): Send alerts to the user if enrollment is let 1000

### LINKED ISSUE #15522: Alert: zero enrollments after N days post-launch
Alert recipe owners when a recipe has launched but enrolled zero clients after N days — usually a sign of broken targeting, mis-set bucketing, or a feature conflict that started before the recipe was published.

#### Trigger

- `Enrolled` count = 0 from launch through N days post-launch.
- Default N: placeholder — to be tuned from real data after launch.

#### Message

- The dominant funnel bucket absorbing the population so the owner has a starting point for diagnosis.

#### Delivery

- In-app badge + Slack, reusing scaffolding from the FeatureConflict alert ticket.

#### Depends on

- ETL (ticket 1)
- Ingestion (ticket 2)
- FeatureConflict alert scaffolding (ticket 4)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6875)

yashikakhurana:
Because

- Because if the experiment is misconfigured we might not see enrollments, so now we can add a check if enrollment numbers is less than 1000 after 3 days, then we can notify the users about it.

This commit

- Send the slack notification if the enrollment is let 1000

Fixes #15522


jaredlockhart (approved):
Very nice, very clear, good separation of concerns, great test coverage. All reads very clearly to me great job @yashikakhurana 🎉 👍

PR #15837, opened 2026-06-08
refactor(nimbus): use active_experiments for sticky enrollment

### LINKED ISSUE #8661: Re-write sticky flag for mobile in terms of active-experiments.
In https://github.com/mozilla/application-services/pull/5374 we refactored how enrollments were calculated.

In order to support a further refactor, experimenter needs to change how it expresses stickiness.

This should move from `is_already_enrolled` to `$slug in active_experiments`.

The second form works is already supported in the SDK, but is not used in the targeting string by experimenter.

_Originally posted by @jhugman in https://github.com/mozilla/application-services/pull/5374#discussion_r1158311719_

RJAK11:
Because

* mozilla/application-services#5374 changed how enrollments are calculated
* experimenter needs to express sticky enrollment using the SDK-supported <slug> in active_experiments format

This commit

* updates sticky enrollment targeting to use <slug> in active_experiments

Fixes #8661

### DIFF
--- a/cirrus/server/tests/test_feature_manifest.py
+++ b/cirrus/server/tests/test_feature_manifest.py
@@ -148,7 +148,8 @@ def test_compute_feature_configurations_targeting_doesnt_match(fml_setup):
         "appId": "test_app_id",
         "appName": "test_app_name",
         "channel": "developer",
-        "targeting": '(is_already_enrolled) || (username in ["test", "test2"])',
+        "targeting": "('experiment-slug' in active_experiments) "
+        "|| (username in ['test', 'test2'])",
         "bucketConfig": bucket_config,
         "isRollout": False,
         "isEnrollmentPaused": False,
--- a/cirrus/server/tests/test_feature_manifest.py
+++ b/cirrus/server/tests/test_feature_manifest.py
@@ -148,8 +148,9 @@ def test_compute_feature_configurations_targeting_doesnt_match(fml_setup):
         "appId": "test_app_id",
         "appName": "test_app_name",
         "channel": "developer",
-        "targeting": "('experiment-slug' in active_experiments) "
-        "|| (username in ['test', 'test2'])",
+        "targeting": (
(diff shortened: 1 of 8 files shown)


cirrus/server/tests/test_feature_manifest.py
```diff
@@ -148,7 +148,8 @@ def test_compute_feature_configurations_targeting_doesnt_match(fml_setup):
         "appId": "test_app_id",
         "appName": "test_app_name",
         "channel": "developer",
-        "targeting": '(is_already_enrolled) || (username in ["test", "test2"])',
+        "targeting": "('experiment-slug' in active_experiments) "
+        "|| (username in ['test', 'test2'])",
```
freshstrangemusic:
```suggestion
        "targeting": (
            "('experiment-slug' in active_experiments) "
            "|| (username in ['test', 'test2'])"
        ),
```

prefer multiline strings in () so that they reflow aligned and are easier to read

cirrus/server/tests/test_feature_manifest.py
```diff
@@ -148,7 +148,8 @@ def test_compute_feature_configurations_targeting_doesnt_match(fml_setup):
         "appId": "test_app_id",
         "appName": "test_app_name",
         "channel": "developer",
-        "targeting": '(is_already_enrolled) || (username in ["test", "test2"])',
+        "targeting": "('experiment-slug' in active_experiments) "
```
freshstrangemusic: I just checked and we dont have `experiment` in the targeting context on mobile so we'll have to do this in multiple stages. I'm going to file some more tickets and group them into an epic.

freshstrangemusic (changes requested):

cirrus/server/tests/test_feature_manifest.py
freshstrangemusic: Ignore -- I misread.

freshstrangemusic (approved):

Issue #15845, opened 2026-06-08
Fenix enrollment test: single log-state read cannot recover a late-persisted enrollment


jaredlockhart:
The Fenix Enrollment integration test (`experimenter/tests/integration/nimbus/android/test_fenix_enrollment.py`) ran `nimbus-cli log-state` **once** before the poll loop, and the loop then only re-dumped the logcat ring buffer — it never re-issued `log-state`. On a loaded CI emulator the enroll launch's `applyPendingExperiments` doesn't persist within the fixed wait, so the single `log-state` read sees an empty DB and the loop can never recover it.

### Fix (landed in #15846)

Re-issue `log-state` inside the poll loop (clear logcat, relaunch, dump, match) and raise the timeout, so a late-persisted enrollment still surfaces.

This is fixed in main. Remaining Fenix-enrollment flakiness is tracked in follow-up issues (#15986 and the bare-log-state recovery issue).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6962)

PR #15846, opened 2026-06-08
fix(nimbus): re-issue log-state in the fenix enrollment poll loop

jaredlockhart:
Because

* The single `log-state` launch can read an empty DB before the enroll's enrollment persists on a loaded CI emulator, and the poll loop only re-dumps logcat — it never re-issues `log-state` — so an early empty read can never recover.

This commit

* Re-issues `log-state` inside the poll loop (clear logcat, relaunch, dump, match) and raises the timeout, so a late-persisted enrollment still surfaces.

Refs #15845

---

Scoped to the `No log-state row found` flake. The separate `5554 / SeekableByteChannel` SDK-download flake tracked in #15845 needs a different approach (pre-download required cmdline-tools that aren't present pre-action on these runners) and is deferred to a follow-up.

### DIFF
--- a/experimenter/tests/integration/nimbus/android/test_fenix_enrollment.py
+++ b/experimenter/tests/integration/nimbus/android/test_fenix_enrollment.py
@@ -9,8 +9,8 @@ from nimbus.models.base_dataclass import BaseExperimentApplications
 
 FENIX_APP = BaseExperimentApplications.FIREFOX_FENIX.value
 APP_APPLY_WAIT = 15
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):

freshstrangemusic:
Looks like sdkmanager isnt installed?

```
/usr/local/lib/android/sdk/cmdline-tools/latest/bin/sdkmanager: No such file or directory
```

PR #15871, opened 2026-06-09
feat(nimbus): Always fetch newer results

### LINKED ISSUE #15737: Always fetch new results
In the results fetch task

- Loop over all files in results bucket
- Compare timestamps to what's stored
- Any newer just load

So forget things like

- Checking the status
- Checking the 3 day window

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6915)

RJAK11:
Because

* the fetch task used experiment status and a fixed time window to decide when to fetch results

This commit

* adds a field to track when an experiment's results were last loaded
* finds the newest modified time among each experiment's statistics, metadata, and errors files
* fetches results when that timestamp is newer than the stored one and saves the new timestamp

Fixes #15737


experimenter/experimenter/experiments/tests/test_changelog_utils.py
```diff
@@ -114,6 +114,7 @@ def test_outputs_expected_schema_for_empty_experiment(self):
                 "required_experiments": [],
                 "requires_restart": False,
                 "results_data": None,
+                "results_data_updated_at": None,
```
mikewilli: Maybe this?
```suggestion
                "results_data_last_updated": None,
```

experimenter/experimenter/experiments/migrations/0332_nimbusexperiment_results_data_last_updated.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('experiments', '0331_alter_nimbusalert_alert_type'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='nimbusexperiment',
+            name='results_data_updated_at',
+            field=models.DateTimeField(blank=True, null=True, verbose_name='Results Data Last Updated'),
```
mikewilli: What's going to be the behavior the first time we run the new version of the task? Refresh every experiment's results whether it needs them or not?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -41,9 +46,45 @@ def strip_errors(data):
...
+                continue
+
+            path = Path(folder, filename)
+            file_timestamp = analysis_storage.get_modified_time(str(path))
+
+            if latest_timestamp is None or file_timestamp > latest_timestamp:
+                latest_timestamp = file_timestamp
+
+    return latest_timestamp
+
```
mikewilli: I think these would be better inside client.py alongside the other functions that read from the GCS bucket.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -41,9 +46,45 @@ def strip_errors(data):
...
+
 @app.task
 @metrics.timer_decorator("fetch_experiment_data")
-def fetch_experiment_data(experiment_id):
+def fetch_experiment_data(experiment_id, results_data_updated_at=None):
```
mikewilli: Will this still work with the manual re-fetch from the admin console? I don't remember if it calls this or does something else.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -65,6 +105,12 @@ def fetch_experiment_data(experiment_id):
                     message=NimbusChangeLog.Messages.RESULTS_UPDATED,
                 )
 
+        if results_data_updated_at is not None:
+            experiment.results_data_updated_at = results_data_updated_at
+
+        if old_results_data != new_results_data or results_data_updated_at is not None:
+            experiment.save()
```
mikewilli: Should we check if the `results_data_updated_at` is different before saving?

experimenter/experimenter/jetstream/tasks.py
mikewilli: Oh oops I thought line 111 was an `and` when I wrote this. I'll leave the question just in case but I think it should be ok.

mikewilli (commented):
Will this work with our error checks? Right now we inject an error inside the client code if we expect results but the file doesn't exist. I think with this new feature we'd just skip over them.

RJAK11:
> Will this work with our error checks? Right now we inject an error inside the client code if we expect results but the file doesn't exist. I think with this new feature we'd just skip over them.

Thank you @mikewilli  for the review! You are right, the changes would've skipped the error checks and I didn’t notice that. I added a [new function](https://github.com/mozilla/experimenter/pull/15871/changes#diff-de8613f0beb69a9d58d3a8d9680a7ae8f9ab44cccccd798ade2a3a9f5467acc0R156) so that the task also re-fetches when we expect a window whose file isn't in the bucket, so now the client should still inject the error. It's bounded by the analysis in the exact way [we originally used](https://github.com/mozilla/experimenter/blob/5391b05ddd37034b5a19ea8e660b9c3c2c234c78/experimenter/experimenter/jetstream/tasks.py#L88C1-L96C39) so a result that never arrives doesn't just keep refetching forever. This might be a messy solution but I couldn’t think of another approach. Let me know what you think!

experimenter/experimenter/experiments/migrations/0332_nimbusexperiment_results_data_last_updated.py
RJAK11: Yes I believe so. I wasn't able to think of an approach that would safely prevent this.

experimenter/experimenter/jetstream/client.py
```diff
@@ -111,6 +111,64 @@ def get_enrollment_funnel_data():
...
+        < date.today()
+    ):
+        return False
+
+    recipe_slug = experiment.slug.replace("-", "_")
+    statistics_files = results_filenames[STATISTICS_FOLDER]
+    return any(
+        f"statistics_{recipe_slug}_{window}.json" not in statistics_files
+        for window in expected_windows(experiment)
+    )
```
jaredlockhart, line 167: So this is still doing the checks that look for dates and time windows and stuff, I was thinking more like

- If it's live or complete
- And it's an experiment

We just blanket expect results and check for them and compare dates and pull in whatever's newer.

The point of this is supposed to be more like, we throw out those old constraints and let jetstream decide what to compute and when and experimenter just says 'if you have any results newer than what I have i'll take them!'

And then show whatever we find.

That way later if we change jetstream to run under different conditions or more frequently or someone triggers a job we'll get the results without needing to manually refetch.

jaredlockhart (changes requested):

experimenter/experimenter/jetstream/client.py
RJAK11: [That was what I originally had](https://github.com/mozilla/experimenter/commit/ac8af58a88d4791c9eb3cdd76589b1148ce0a7f5) where Experimenter would just pull in newer results if Jetstream produced them but the reason I changed it was because Mike brought up that we currently inject errors when we expect a results file to exist but it is missing from the bucket. So this version was my attempt to still always fetch newer result files when they exist, but also preserve the existing error behaviour when Experimenter expects a file and Jetstream did not produce one. Should we just be skipping those error checks instead?

experimenter/experimenter/jetstream/client.py
```diff
@@ -111,6 +111,64 @@ def get_enrollment_funnel_data():
...
+    }
+
+    latest_timestamp = None
+    for folder, filenames in expected_filenames.items():
+        for filename in filenames:
+            if filename not in results_filenames[folder]:
+                continue
+
+            path = Path(folder, filename)
+            file_timestamp = analysis_storage.get_modified_time(str(path))
```
jaredlockhart: Instead of using the filesystem timestamp, we can/(should?) pull out the computed timestamp from inside the metadata like this:

https://github.com/mozilla/experimenter/blob/ecb1ae791d5af0240b48eb0a69261fc141c6811c/experimenter/experimenter/jetstream/client.py#L423-L427

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -81,25 +91,31 @@ def fetch_experiment_data(experiment_id):
...
+            )
+
+            if latest_results_timestamp is None and not needs_missing_results:
+                metrics.incr("fetch_jetstream_data.skipped")
+                continue
+
+            has_newer_results = latest_results_timestamp is not None and (
+                experiment.results_data_last_updated is None
+                or experiment.results_data_last_updated < latest_results_timestamp
+            )
```
jaredlockhart: And then right here instead of tracking the last updated time as a new field on the model, use the existing stored results timestamp 

https://github.com/mozilla/experimenter/blob/ecb1ae791d5af0240b48eb0a69261fc141c6811c/experimenter/experimenter/nimbus_ui/templates/nimbus_experiments/results-inner.html#L20

That way we don't have to introduce a new field

jaredlockhart (changes requested):
Okay I understand better, yes it makes sense to compute whether we 'expect' results to compute errors but if there's anything newer we always fetch it anyways so it still catches the 'long since ended but forced reran' catches so we don't have to manually reload.  This is good, great test coverage.  One idea about using the stored results date instead of adding a new field 🙏

jaredlockhart (approved):
Right yep we looked at this together I thought I had approved it sorry for the delay, yes let's ship this and then if we find there's some case we missed in prod we can update.  Thanks for tackling this @RJAK11 🙏 🎉  This will be such a help to not have to manually refresh results all the time 🙏 🙏 🙏 🙏

PR #15872, opened 2026-06-09
fix(nimbus): fix localizations reset by changes to audience form

### LINKED ISSUE #15867: Changing locale list removes localization value
For some reason it seems changing the locales in the audience page results in the localizations field being emptied.  This definitely shouldn't be happening let's see if we can recreate it in a test and get that fixed.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6966)

moibra05:
Because

- The new rollouts audience form was recently created but worked by using the audience form class used in the current workflow despite requiring an additional input for localization which we added as a hidden form input
- This decision broke the existing audience form which would always submit an empty value for localizations thus resetting it each time it was saved  

This commit

- Creates a new standalone subclass `RolloutAudienceForm` which inherits from `AudienceForm` and adds any additional fields it requires to better separate the new design form logic
- This in turn fixes the above stated issue

Fixes #15867

### DIFF
--- a/experimenter/experimenter/nimbus_ui/forms.py
+++ b/experimenter/experimenter/nimbus_ui/forms.py
@@ -1035,11 +1035,6 @@ class AudienceForm(NimbusChangeLogFormMixin, forms.ModelForm):
             key=lambda choice: choice[1].lower(),
         )
 
-    YES_NO_CHOICES = (
-        (True, "Yes"),
-        (False, "No"),
-    )
-
     channel = forms.ChoiceField(
         required=False,
         label="",
@@ -1108,7 +1103,6 @@ class AudienceForm(NimbusChangeLogFormMixin, forms.ModelForm):
         required=False,
         widget=MultiSelectWidget(),
     )
-    localizations = forms.CharField(required=False, widget=forms.HiddenInput())
     is_sticky = forms.BooleanField(required=False)
     is_first_run = forms.BooleanField(required=False)
     population_percent = forms.DecimalField(
@@ -1144,7 +1138,6 @@ class AudienceForm(NimbusChangeLogFormMixin, forms.ModelForm):
             "is_sticky",
             "languages",
             "locales",
-            "localizations",
             "population_percent",
             "proposed_duration",
             "proposed_enrollment",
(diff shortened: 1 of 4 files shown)


jaredlockhart (dismissed):
Perfect but let's also treat this like a bug ie the bug was that saving the audience form would empty out the localizations field for an experiment.  So let's add a test that exercises that exact case, and then without the class changes that test should fail and with it it should pass and that way we have it captured.  But other than that yes this all looks exactly right to me 🙏

experimenter/experimenter/nimbus_ui/forms.py
```diff
@@ -1315,6 +1295,32 @@ def get_changelog_message(self):
         return f"{self.request.user} updated audience"
 
 
+class RolloutAudienceForm(AudienceForm):
+    YES_NO_CHOICES = (
+        (True, "Yes"),
+        (False, "No"),
+    )
+    localizations = forms.CharField(required=False, widget=forms.HiddenInput())
```
freshstrangemusic, line 1303: What are we doing about the is_localized field?

moibra05: ah, good point thats missing here. i'll add that in

experimenter/experimenter/nimbus_ui/tests/test_views.py
```diff
@@ -3164,6 +3164,34 @@ def test_post_sets_rollout_dirty_on_population_change(self):
         experiment.refresh_from_db()
         self.assertTrue(experiment.is_rollout_dirty)
 
+    def test_post_does_not_update_localizations(self):
```
jaredlockhart, line 3167: Awesome thank you 🙏

experimenter/experimenter/nimbus_ui/forms.py
moibra05: will fix this in another pr https://github.com/mozilla/experimenter/issues/15947

yashikakhurana (approved):
Lets launch this to unblock the users, thank you @moibra05

PR #15897, opened 2026-06-10
chore(schemas): publish to npm via Trusted Publishing

### LINKED ISSUE #15896: ci(schemas): publish @mozilla/nimbus-schemas to npm via Trusted Publishing
## Problem

The npm half of the schemas deploy was never finished after the CircleCI→GHA migration (#15179 / #15180). `deploy-schemas.yml` references a `secrets.NPM_TOKEN` that was never added, so `@mozilla/nimbus-schemas` is not being published from CI. The PyPI half is working (TWINE secrets are configured).

Rather than re-create a long-lived npm automation token, npm now recommends [Trusted Publishing](https://docs.npmjs.com/trusted-publishers) (OIDC) for CI/CD and flags automation tokens as a security risk.

## Change

Move the npm publish step in `deploy-schemas.yml` to Trusted Publishing:

- Grant the job `id-token: write`.
- Publish from the runner (where the OIDC credential lives) with `npm publish --access public --provenance` on npm ≥ 11.5.1, instead of the Dockerized `yarn publish` + `.npmrc` authToken. The build step already produces the publishable artifacts on the host via the volume mount.
- No npm secret stored anywhere; provenance attestation generated automatically.
- PyPI step unchanged.

Requires a one-time trusted-publisher configuration on npmjs.com for `@mozilla/nimbus-schemas` (repo `mozilla/experimenter`, workflow `deploy-schemas.yml`).

## Acceptance criteria

- `deploy-schemas.yml` publishes to npm without any stored npm token.
- The next `schemas/VERSION` bump merged to `main` publishes the new version to npm with a provenance attestation.
- PyPI publishing continues to work unchanged.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6978)

jaredlockhart:
Because

* npm publishing was never finished after the CircleCI→GHA migration — `deploy-schemas.yml` referenced an `NPM_TOKEN` secret that was never added, so `@mozilla/nimbus-schemas` stopped publishing. The PyPI half works.
* npm now recommends [Trusted Publishing](https://docs.npmjs.com/trusted-publishers) (OIDC) over long-lived automation tokens for CI/CD.

This commit

* grants the deploy job `id-token: write`
* publishes from the runner with `npm publish --access public --provenance` on a recent npm, replacing the Dockerized `yarn publish` + `.npmrc` authToken (yarn classic can't do the OIDC exchange; the build step already produces the artifacts on the host)
* leaves PyPI publishing unchanged

Requires a one-time trusted-publisher config on npmjs.com for `@mozilla/nimbus-schemas` (repo `mozilla/experimenter`, workflow `deploy-schemas.yml`) before the next version bump merges.

Fixes #15896

### DIFF
--- a/.github/workflows/deploy-schemas.yml
+++ b/.github/workflows/deploy-schemas.yml
@@ -56,6 +56,10 @@ jobs:
         if: steps.check-paths.outputs.should-run == 'true' && steps.version-check.outputs.changed == 'true'
         working-directory: schemas
         run: |
-          npm install -g npm@latest
-          npm version --allow-same-version --no-git-tag-version "$(cat VERSION)"
+          PKG_VERSION="$(node -p "require('./package.json').version")"
+          FILE_VERSION="$(cat VERSION)"
+          if [ "$PKG_VERSION" != "$FILE_VERSION" ]; then
+            echo "package.json version ($PKG_VERSION) does not match VERSION ($FILE_VERSION); run 'make schemas_version' and commit" >&2
+            exit 1
+          fi
           npm publish --access public --provenance
--- a/.github/workflows/deploy-schemas.yml
+++ b/.github/workflows/deploy-schemas.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: read
+      id-token: write
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-schemas.yml
```diff
@@ -46,7 +47,15 @@ jobs:
...
+        if: steps.check-paths.outputs.should-run == 'true' && steps.version-check.outputs.changed == 'true'
+        with:
+          node-version: 24
+
       - name: Upload to NPM
         if: steps.check-paths.outputs.should-run == 'true' && steps.version-check.outputs.changed == 'true'
+        working-directory: schemas
         run: |
-          make schemas_deploy_npm SCHEMAS_ENV="-e NPM_TOKEN=${{ secrets.NPM_TOKEN }}"
+          npm install -g npm@latest
```
mikewilli: Wouldn't `npm` be installed as part of the `setup-node` action?

.github/workflows/deploy-schemas.yml
```diff
@@ -46,7 +47,15 @@ jobs:
...
+          npm version --allow-same-version --no-git-tag-version "$(cat VERSION)"
```
mikewilli: I don't know if we want to set the version here -- I think this has the potential to make the published version and the repo version (in package.json) get out of sync. If anything, I think we could check that the package.json version is already the same as the VERSION file specifies, but not sure if that's necessary.

mikewilli (approved):
Neat! I was thinking of a different way to do the check (something like `npm version "$(cat VERSION)" && exit_with_failure || echo "Publishing..."`), but this works and is much easier to parse so works for me.

PR #15900, opened 2026-06-10
feat(nimbus): replace feature monitoring iframe with Grafana dashboard link

### LINKED ISSUE #15898: SRE to update the grafana dashboard settings
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6946)

yashikakhurana:
## Because

- The feature monitoring is available, the iframe integration is taking time to verify, meanwhile lets launch the feature montioring dashboard as a link 
## This commit

- Removes the embedded iframe from the Feature Health Dashboard
- Adds a **Grafana dashboard link** that opens directly in Grafana if the feature is configured in metric-hub (`firefox_desktop.toml`)
- Shows a **warning** if the feature is not in metric-hub, with a direct link to the config file and the new [feature monitoring docs](https://experimenter.info/feature-monitoring)
- Moves the Feature Monitoring card to appear immediately after feature selection (before Deliveries / QA runs)
- Adds `var-metric=All` to the Grafana URL so all metrics are selected by default

Fixes #15898


experimenter/experimenter/experiments/models.py
```diff
@@ -2970,6 +2976,21 @@ def get_versioned_schema_range(
         )
 
 
+@functools.lru_cache(maxsize=1)
+def _get_featmon_slugs():
+    path = settings.METRIC_HUB_FEATMON_DESKTOP_PATH
+    try:
+        content = Path(path).read_text()
```
jaredlockhart: We shouldn't have to do it as text? It's a TOML file we can just load it?

https://docs.python.org/3/library/tomllib.html

jaredlockhart: Also I'd prolly put this utility in the jetstream folder with the other utilities for reading those configs.

jaredlockhart (dismissed):

experimenter/experimenter/experiments/models.py
yashikakhurana: good point

experimenter/experimenter/jetstream/client.py
```diff
@@ -454,3 +457,16 @@ def get_population_sizing_data():
...
+
+@functools.lru_cache(maxsize=1)
+def get_featmon_slugs():
+    path = settings.METRIC_HUB_FEATMON_DESKTOP_PATH
+    try:
+        data = tomllib.loads(Path(path).read_text())
+        return frozenset(
+            feature.get("slug", key.replace("_", "-"))
+            for key, feature in data.get("features", {}).items()
+        )
```
jaredlockhart, line 470: better 🙏

experimenter/experimenter/experiments/models.py
```diff
@@ -2846,6 +2846,12 @@ def feature_monitoring_url(self):
             slug=self.slug, application=application
         )
 
+    @property
+    def has_metric_hub_monitoring(self):
+        from experimenter.jetstream.client import get_featmon_slugs
```
jaredlockhart, line 2851: It'll be sweet when they land [Lazy Imports](https://peps.python.org/pep-0690/) so we don't have to do inline imports anymore 🙏 😭

experimenter/experimenter/jetstream/client.py
jaredlockhart: But also shouldn't we add tests for this?

jaredlockhart (dismissed):
Ya way better awesome ty, prolly just add some tests for the featmon loader but otherwise looks great 🎉 🎉 🎉

experimenter/experimenter/experiments/models.py
yashikakhurana: that is so true

jaredlockhart (approved):
Great looks great thnx @yashikakhurana let's gooooo 🚀 🚀 🚀 🚀 🚀

PR #15901, opened 2026-06-10
fix(nimbus): fix default metrics missing for experiments with missing exposure data

### LINKED ISSUE #15684: Experiment weekly/overall results not appearing
This [experiment](https://experimenter.services.mozilla.com/nimbus/smart-shortcuts-v4p2/results/) seems to have been successfully analyzed by jetstream (there are results in bigquery and GCS), but only a couple metrics are appearing in the results page.

Of note, I didn't notice any errors when manually triggering a results re-fetch from the admin console, either in prod or locally. However, the experiment object is quite large when exported from the admin (42MB), so I had to tweak the nginx config locally to allow for it to be imported.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6900)

moibra05:
Because

- `other_metrics` is defined by gathering all the metrics that aren't KPIs and are not associated with an outcome 
- It's expected that for an experiment with exposure data, the collected `other_metrics` would be the same as for enrollments so setting `experiment_data["other_metrics"] = other_metrics` is redundant for exposures data
- However, there exists experiments whose exposure data is in an erroneous state missing all the expected default metrics  meaning that when we set `experiment_data["other_metrics"] = other_metrics` what actually occurs is that the correct value for `other_metrics` is overridden to be incorrectly empty

This commit

- Changes `experiment_data["other_metrics"]` to be updated with any additional metrics instead of being overridden

Fixes #15684

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -357,7 +357,7 @@ def get_experiment_data(experiment: NimbusExperiment):
             )
 
             if segment == Segment.ALL:
-                experiment_data["other_metrics"] = other_metrics
+                experiment_data["other_metrics"].update(other_metrics)
 
             if data and window == AnalysisWindow.OVERALL:
                 # Append some values onto Jetstream data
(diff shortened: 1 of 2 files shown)


mikewilli (commented):
Incredible one line fix! Are there any test updates to make alongside the change?

mikewilli (approved):

PR #15989, opened 2026-06-16
fix(nimbus): re-enroll in fenix poll loop instead of bare log-state

### LINKED ISSUE #15988: Fenix enrollment test flaky: bare log-state in poll loop cannot recover a failed apply
Follow-up to #15845. After #15846 moved `log-state` into the poll loop, the Fenix Enrollment integration test still fails intermittently — mostly on the `beta` channel:

```
AssertionError: No log-state row found for fenix-<channel>-integration-test after 120s.
```

### Root cause

nimbus-cli's `enroll` launch is self-contained: in one app launch it resets the db, injects the experiment payload, applies it (enrolls), and prints the log-state table (`application-services/components/support/nimbus-cli/src/cmd.rs:367-373`). The recipe is `targeting:"true"` + bucket `count:10000/total:10000` (100%), so enrollment is deterministic — it cannot fail on eligibility.

The poll loop re-issues **bare `log-state`** (`cmd.rs:314-319`), which injects no experiment payload and does not reset. On a loaded beta emulator, the enroll launch's `applyPendingExperiments` doesn't commit to SQLite before the app is backgrounded — the `reset-db` already wiped the db, and the apply never landed. From then on, bare `log-state` has nothing to surface and nothing to recover from, so the loop reports an empty table for the full 120s.

So #15846 re-issued the right cadence but the wrong command.

### Fix

In the poll loop, re-issue the full `enroll` (reset-db + inject payload + apply + log-state) each iteration instead of bare `log-state`, so an incomplete apply self-heals on the next poll. Matches nimbus-cli's design (the enroll launch logs state itself).

### Acceptance criteria

- [ ] The poll loop re-injects the experiment payload each iteration so a late/incomplete apply still surfaces.
- [ ] No change to enrollment authority — Nimbus still owns assignment; this is read-reliability only.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7011)

jaredlockhart:
Because

* The nimbus-cli `enroll` launch is self-contained — it resets the db, injects the experiment payload, applies it, and prints log-state in one launch. The recipe is 100%/targeting=true, so enrollment is deterministic, yet beta intermittently shows an empty table for the full 120s window.
* The poll loop (post-#15846) re-issued bare `log-state`, which injects no payload and does not reset. When a loaded emulator backgrounds the enroll launch before `applyPendingExperiments` commits, the db is left wiped with nothing to surface — and bare `log-state` can never recover it, so the loop fails every iteration.

This commit

* Re-issues the full `enroll` (reset + inject + apply + log-state) each poll iteration so an incomplete apply self-heals on the next poll.
* Factors the enroll call into a local helper; renames the LOG_STATE_* constants to ENROLL_* to match the new behavior.

Fixes #15988

### DIFF
--- a/experimenter/tests/integration/nimbus/android/test_fenix_enrollment.py
+++ b/experimenter/tests/integration/nimbus/android/test_fenix_enrollment.py
@@ -8,9 +8,8 @@ import pytest
 from nimbus.models.base_dataclass import BaseExperimentApplications
 
 FENIX_APP = BaseExperimentApplications.FIREFOX_FENIX.value
-APP_APPLY_WAIT = 15
-LOG_STATE_TIMEOUT = 120
-LOG_STATE_POLL_INTERVAL = 5
+ENROLL_TIMEOUT = 120
+ENROLL_POLL_INTERVAL = 5
 
 
 @pytest.mark.fenix_enrollment
@@ -27,7 +26,6 @@ def test_fenix_enrollment(
     recipe_path.write_text(json.dumps(recipe))
 
     subprocess.check_call(["adb", "install", fenix_apk_path])
-    subprocess.check_call(["adb", "logcat", "-c"])
 
     # Prevent the real Nimbus RS fetch from overwriting our local enrollment
     # (without this, Nimbus evolves against the fresh fetch — which does not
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
Thank you @jaredlockhart

PR #16011, opened 2026-06-17
feat(nimbus): Include ration in the calculation for the SRM ratios alert

### LINKED ISSUE #16010: Adjust SRM ration for the alert
We have been noticing that due to such a small ratio mismatch it can still trigger a false ration alert and we should broaden up the ratio alert threshold

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7020)

yashikakhurana:
Because

- Experiments with intentionally unequal branch ratios (e.g. 95% delivery / 5%) were triggering false SRM mismatch alerts because the check assumed equal distribution across branches

This commit

- Passes each experiment's configured branch ratios to the SRM check so the chi-square test compares actual enrollment against the correct expected ratio, not 50/50

Fixes #16010


yashikakhurana:
@jaredlockhart Fixed two things that were causing false SRM alerts:

- The check assumed equal branch ratios even for intentional splits like 95/5 — now it uses the actual configured ratios
- With large populations, even a 0.2% difference was triggering the alert — added a minimum 5% deviation so it only fires when something is actually wrong

jaredlockhart (dismissed):
Oh yes good, taking the designed ratio into account good catch!  And actually yeah checking telemetry shows that 5% is probably exactly what we want.  One thing though I think this only updates slack, don't we also need to pass the ratios in here:

https://github.com/mozilla/experimenter/blob/4abc97868dd726584ec8841ccb5fac5cb8144dc4/experimenter/experimenter/experiments/models.py#L1589

To fix it in the UI?

yashikakhurana:
> Oh yes good, taking the designed ratio into account good catch! And actually yeah checking telemetry shows that 5% is probably exactly what we want. One thing though I think this only updates slack, don't we also need to pass the ratios in here:
> 
> https://github.com/mozilla/experimenter/blob/4abc97868dd726584ec8841ccb5fac5cb8144dc4/experimenter/experimenter/experiments/models.py#L1589
> 
> To fix it in the UI?

ahh very good catch @jaredlockhart

jaredlockhart (approved):
Oh sweet, thanks for fixing, looks good let's ship it! ⛵ 🎉

Issue #15917, opened 2026-06-11
Some metrics show error message, but have results inside the details modal


mikewilli:
For [this experiment](https://experimenter.services.mozilla.com/nimbus/android-second-and-seven-day-onboarding/results/?analysis_basis=enrollments&reference_branch=control&segment=all), some of the metrics for the Enrollments basis have the message on the main page that results are not available, but clicking in to the details modal shows the results (and an error). Examples: Widget Searches and Tagged Search Count.

For a (somewhat) educated guess, maybe this is because the metric is missing exposure results but has enrollment results?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6984)

mikewilli:
Found another example: this [experiment](https://experimenter.services.mozilla.com/nimbus/suggest-query-normalization-experiment/results/) is showing that results are not available for 3 Days Retention, but the error is only on Day 8 and doesn't seem to affect the results showing in the details modal. Similar for Firefox Desktop Client-Level DAU.

PR #16042, opened 2026-06-19
fix(nimbus): fix results page showing stale errors

### LINKED ISSUE #15946: Results page is showing errors from old/stale analysis runs
I'm pretty sure we have some filtering on both the Jetstream and Experimenter side to filter out old errors from previous analysis runs, but I'm currently seeing old errors on [this experiment](https://experimenter.services.mozilla.com/nimbus/suggest-query-normalization-experiment/results/?analysis_basis=exposures&reference_branch=treatment&segment=all) that are no longer relevant.

Filing here because I know that Jetstream hasn't changed anything in terms of what it exports, so I'm pretty sure this is an Experimenter regression, but a) I could be wrong, and b) we could still consider trying to fix on the Jetstream side anyway.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7003)

moibra05:
Because

- Metric level analysis errors could include stale errors from before the current analysis run

This commit

- Applies the existing `analysis_start_time` filtering to metric level errors as well as experiment level errors
- Adds test coverage for including fresh metric errors, excluding stale ones, and preserving malformed timestamp fallback behavior

Fixes #15946

### DIFF
--- a/experimenter/experimenter/jetstream/client.py
+++ b/experimenter/experimenter/jetstream/client.py
@@ -411,26 +411,31 @@ def get_experiment_data(experiment: NimbusExperiment):
 
     errors_by_metric = {}
     errors_experiment_overall = []
+
+    def include_error_after_analysis_start(err):
+        try:
+            analysis_start_time = datetime.fromisoformat(
+                experiment_metadata.get("analysis_start_time")
+                if experiment_metadata is not None
+                else ""
+            )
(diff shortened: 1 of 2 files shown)


moibra05:
Should also resolve #15917 once merged

jaredlockhart (approved):
Okay good, good find good fix thnx @moibra05 🙏

PR #16044, opened 2026-06-19
fix(nimbus): use Services.prefs instead of about:config UI in desktop integration tests

### LINKED ISSUE #16043: Desktop Enrollment integration test fails on Firefox 152 (about:config flip_pref no longer toggles)
## Problem

The `update_firefox_desktop_release` PRs fail `Desktop Enrollment (1)` once the release bump crosses Firefox 151 → 152, with a misleading `selenium ... TimeoutException: Summary Page: Unable to find live status`. Re-running does not help. This **blocks every desktop release-version bump from merging** (e.g. #16021, earlier #15983).

It is not flaky. The Desktop Enrollment workflow is hardcoded to `FIREFOX_CHANNEL: release`, so every run installs the release Firefox from `experimenter/tests/firefox_desktop_release_build.env`. Clean A/B: release 151.0.4 passes everywhere; release 152.0.x fails only on the PRs that change that file. Same commit, same test, only the browser binary differs.

## Real failure

The CI text log shows only the rerun's masked `TimeoutException`. The uploaded test-report artifact has the actual first-attempt traceback:

```
about_config.flip_pref(pref_name)
> assert about_config.get_pref_value(pref_name) == "false"
E AssertionError: assert 'true' == 'false'
test_desktop_client_integrations.py:217
```

Enrollment fires and the sticky experiment correctly sets `sticky.targeting.test.pref` to `true` (so Nimbus enrollment and sticky setPref both work on 152). `flip_pref()` is meant to toggle it back to `false`, but the toggle is a no-op, so the assert fails. The "Unable to find live status" timeout appears only on the rerun, because the failed first attempt leaves a half-launched experiment in the single-worker stack and the fresh rerun experiment can't sync to Live.

## Root cause

`flip_pref` in `experimenter/tests/integration/nimbus/pages/browser/__init__.py` single-clicks the cell:

```python
el.find_element(By.CSS_SELECTOR, ".cell-edit").click()   # clicks <td class="cell-edit">
```

Firefox 152's about:config click handler now ignores single-clicks that aren't on a button — `if (event.target.localName != "button") { return; }`. In 151 a single click on the cell toggled the boolean; in 152 you must click the inner `<button>` (or double-click). The test single-clicks the `<td>`, so the bool is never toggled.

This is a test-maintenance issue, not a Nimbus/product regression.

## Proposed fix

Click the toggle button inside the edit cell instead of the cell:

```python
el.find_element(By.CSS_SELECTOR, ".cell-edit button").click()
```

`flip_pref` is the shared pref-toggle helper, so this protects any other test that flips a pref.

## Acceptance criteria

- `test_check_telemetry_sticky_targeting[FIREFOX_DESKTOP]` passes against Firefox release 152.x.
- The desktop release-version bump PR (#16021) goes green once rebased on this fix.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7036)

jaredlockhart:
Because

* The desktop integration tests toggled and read prefs by scraping the about:config UI, which broke on Firefox 152 — a single click on the edit cell no longer toggles a boolean pref, so the sticky-targeting test failed on every release-version bump to 152+ (it surfaced misleadingly as a rerun `Unable to find live status` timeout).
* Driving the about:config DOM was always fragile; the chrome-context `Services.prefs` API (already used by `wait_for_pref_flip`) is version-stable.

This commit

* Rewrites `get_pref_value` and `flip_pref`, and adds `set_pref`, to use `Services.prefs.{get,set}{Bool,Int,String}Pref` via the chrome context.
* Drops the now-unneeded about:config navigation, DOM locators, and `wait_for_page_to_load`.

Fixes #16043

### DIFF
--- a/experimenter/tests/integration/nimbus/pages/browser/__init__.py
+++ b/experimenter/tests/integration/nimbus/pages/browser/__init__.py
@@ -3,8 +3,6 @@
 import time
 
 from pypom import Page
-from selenium.webdriver.common.by import By
-from selenium.webdriver.support import expected_conditions as EC
 
 
 class Browser:
@@ -19,16 +17,8 @@ class Browser:
 class AboutConfig(Page):
     URL_TEMPLATE = "about:config"
 
-    _search_bar_locator = (By.ID, "about-config-search")
-    _row_locator = (By.CSS_SELECTOR, "tr > td > span > span")
-    _pref_locator = (By.CSS_SELECTOR, "tr > th > span")
-
     def __init__(self, selenium, **kwargs):
         super().__init__(selenium, timeout=80, **kwargs)
-
-    def wait_for_page_to_load(self):
-        self.wait.until(EC.presence_of_element_located(self._search_bar_locator))
-        return self
 
     def wait_for_pref_flip(self, pref, pref_value, action=None, pref_type=str):
         timeout = time.time() + 60 * 5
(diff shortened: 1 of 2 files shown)


freshstrangemusic (changes requested):
r- on strategy. If we're here anyway we should update the pref setting logic to use `Services.prefs.{get,set}{Bool,Int,String}Pref`

experimenter/tests/integration/nimbus/pages/browser/__init__.py
```diff
@@ -61,44 +51,36 @@ def wait_for_pref_flip(self, pref, pref_value, action=None, pref_type=str):
...
+                return self.selenium.execute_script(
+                    "return Services.prefs.getBoolPref(arguments[0])", pref
+                )
+            elif pref_type is int:
+                return self.selenium.execute_script(
+                    "return Services.prefs.getIntPref(arguments[0])", pref
+                )
+            return self.selenium.execute_script(
+                "return Services.prefs.getStringPref(arguments[0])", pref
+            )
```
freshstrangemusic: A few things:

1. This is an excellent case for a match
2. If we call this with pref_type not in (bool, int) we try to read a string pref, but we should probably throw instead because that is programmer error
3. missing ; in script

```suggestion
            match pref_type:
                case bool:
                    method = "getBoolPref"
                case int:
                    script = "getIntPref"
                case str:
                    script = "getStringPref"
                case _:
                    raise TypeError("Unsupported pref type {pref_type}")

            return self.selenium.execute_script(
                f"return Services.prefs.{method}(arguments[0]);",
                pref,
            )
```

experimenter/tests/integration/nimbus/pages/browser/__init__.py
```diff
@@ -61,44 +51,36 @@ def wait_for_pref_flip(self, pref, pref_value, action=None, pref_type=str):
...
-                self.selenium.get("about:config")
-                self.wait_for_page_to_load()
             else:
-                return True
-        raise (error)
+                self.selenium.execute_script(
+                    "Services.prefs.setStringPref(arguments[0], arguments[1])",
+                    pref,
+                    value,
+                )
```
freshstrangemusic: 1. Again, another great case for match/case.
2. We shouldn't accept `pref_type` -- we can switch on `type(value)`, which prevents errors of the class `set_pref("foo", 123, pref_type="string")`
3. Likewise, not accepting pref_type means we won't end up calling setStringPref for unsupported pref types

```suggestion
    def set_pref(self, pref, value):
        with self.selenium.context(self.selenium.CONTEXT_CHROME):
            match type(value):
                case bool:
                    method = "setBoolPref"
                case int:
                    method = "setIntPref"
                case string:
                    method = "setStringPref"
                case _:
                    raise TypeError("Unsupported pref type {type(value)}")
            
            return self.selenium.executeScript(
                f"Services.prefs.{method}(arguments[0], arguments[1])",
                pref,
                value,
            )
```

freshstrangemusic (approved):

experimenter/tests/integration/nimbus/pages/browser/__init__.py
jaredlockhart: Done — converted to match/case, added the `case _` that raises `TypeError`, and added the missing `;`.

One semantic correction: here we're matching on the *type object* `pref_type`, and a bare `case bool:` is a capture pattern (it binds the name `bool` and matches everything), so the first case would always win. To compare type identity I used `is` guards:

```python
match pref_type:
    case _ if pref_type is bool:
        method = "getBoolPref"
    ...
    case _:
        raise TypeError(f"Unsupported pref type {pref_type}")
```

experimenter/tests/integration/nimbus/pages/browser/__init__.py
jaredlockhart: Done — dropped `pref_type` and switched on the value, so `set_pref("foo", 123, pref_type="string")` is impossible and unsupported types raise.

Switched on the value with class patterns rather than `type(value)` with bare cases (same capture-pattern gotcha — `case bool:` would match anything). `case bool()` does an isinstance check, and since `bool` is a subclass of `int` the `bool()` case is ordered first:

```python
match value:
    case bool():
        method = "setBoolPref"
    case int():
        method = "setIntPref"
    case str():
        method = "setStringPref"
    case _:
        raise TypeError(f"Unsupported pref type {type(value)}")
```

`flip_pref` now calls `set_pref(pref, not get_pref_value(pref, pref_type=bool))` (no `pref_type`).

PR #16063, opened 2026-06-22
feat(nimbus): Add Rollout Schedule Card

### LINKED ISSUE #15081: Rollout Schedule Form/Card
Implement the Rollout Schedule card for the rollout summary page. Replaces the existing Edit rollout schedule modal with an inline phase list.

* Remove the Edit rollout schedule modal entirely.
* Inline phase list: Phase 1, Phase 2, Phase N — each with date, duration, and population %.
* Add and remove phases inline with + and trash icons.
* When rollout is Live, display format: <Date> | <Duration> and Enrolling Phase N.
* Schedule remains editable even when Live to allow phase adjustments.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6685)

RJAK11:
Because

* We need to be able to manage rollout schedules from the rollout summary page

This commit

* Adds a rollout schedule card with phase dates and population percentages
* Adds support for adding and removing rollout phases
* Shows rollout phase progress with ability to pause/resume/start buttons

Fixes #15081

https://github.com/user-attachments/assets/fce834ef-b446-4588-9254-0ebc7a45291b

https://github.com/user-attachments/assets/3ecf4354-74c0-466e-9b73-76c7c86d7db9

https://github.com/user-attachments/assets/b60f58f5-3e8b-466d-91fb-b293f10dabec


experimenter/experimenter/experiments/models.py
```diff
@@ -2754,6 +2793,74 @@ def __str__(self):
         return f"{self.title} ({self.link})"
 
 
+class NimbusRolloutPhase(models.Model):
```
moibra05, line 2846: i like this approach 👍

moibra05 (commented):
found a couple bugs testing locally:

- its possible to set only the start or end date without the other. im not sure if this is intentional
- deleting/adding a phase then cancelling the form does not undo the phase action

besides that i think the code looks really good to me :)

yashikakhurana (changes requested):
Thank you @RJAK11  this is great, overall looks good, just some design changes

Image
Image
we will move all the controls like start phase etc etc left side on the sidebar, so this start here will not required

Image

RJAK11:
Thank you both for the reviews! I have addressed all the mentioned issues and you can see the changes in the video. For the bug regarding deleting/adding a phase then cancelling the form not undoing the action, I initially avoided that to prevent introducing some complicated code but I just added it with the other changes. Let me know if this approach looks okay or if you think there’s a better way to handle it 🙏🏼

https://github.com/user-attachments/assets/0373fa86-109a-4878-a5ee-cd8618e60bb5

experimenter/experimenter/nimbus_ui/urls.py
```diff
@@ -359,6 +366,41 @@
...
+    re_path(
+        r"^new/(?P<slug>[\w-]+)/update_schedule/$",
+        NewRolloutScheduleUpdateView.as_view(),
+        name="nimbus-ui-new-update-schedule",
+    ),
+    re_path(
+        r"^new/(?P<slug>[\w-]+)/create_rollout_phase/$",
+        NewRolloutPhaseCreateView.as_view(),
+        name="nimbus-ui-new-create-rollout-phase",
+    ),
```
yashikakhurana, line 376: cant we use create/update same url?

experimenter/experimenter/nimbus_ui/urls.py
```diff
@@ -359,6 +366,41 @@
...
+        name="nimbus-ui-new-delete-rollout-phase",
+    ),
+    re_path(
+        r"^new/(?P<slug>[\w-]+)/create_rollout_plan/$",
+        NewRolloutPlanCreateView.as_view(),
+        name="nimbus-ui-new-create-rollout-plan",
+    ),
+    re_path(
+        r"^new/(?P<slug>[\w-]+)/apply_rollout_plan/$",
+        NewRolloutPlanApplyView.as_view(),
```
yashikakhurana, line 389: I dont think we are using apply anywhere

experimenter/experimenter/nimbus_ui/urls.py
```diff
@@ -359,6 +366,41 @@
...
+    re_path(
+        r"^new/(?P<slug>[\w-]+)/start_rollout_phase/$",
+        NewRolloutStartPhaseView.as_view(),
+        name="nimbus-ui-new-start-rollout-phase",
+    ),
+    re_path(
+        r"^new/(?P<slug>[\w-]+)/pause_rollout/$",
+        NewRolloutPauseView.as_view(),
+        name="nimbus-ui-new-pause-rollout",
+    ),
```
yashikakhurana: do we need this here?

yashikakhurana (changes requested):
In general looks good, thank you @RJAK11, some suggestion, some views and urls are not needed now as we are not adding pause/start here

experimenter/experimenter/nimbus_ui/urls.py
RJAK11: I tried doing that but it made the logic super complicated because UpdateView persists the schedule to the db while CreateView creates a temporary preview before the user actually saves. I did this so that when the user cancels, any changes they made aren't actually saved.

experimenter/experimenter/nimbus_ui/urls.py
RJAK11: ApplyView is used [here](https://github.com/mozilla/experimenter/pull/16063/changes#diff-ca3de30905e42307f0e66a8f1dac365d84c5cdb06ca2203eceb8ff1b91840c06R960) and it just applies a template plan when the user chooses one from the dropdown.

moibra05 (commented):
im wondering if it makes sense to allow the user to create a phase and set the population ratio to 0. wouldn't this have the same effect as simply pausing/disabling the rollout? if that's the case maybe we should restrict the values for population ratio to (0, 100%] to avoid confusion

moibra05:
also thanks for your feedback in #16138. curious if what @yashikakhurana meant was to have the form always submit and go back to the "summary" page when you add/delete a phase similar to how its done for changing the project tags and subscribers in the overview form?

https://github.com/user-attachments/assets/f1a1ae93-f8b0-4c53-a381-634fe1fc45e8

yashikakhurana:
> also thanks for your feedback in #16138. curious if what @yashikakhurana meant was to have the form always submit and go back to the "summary" page when you add/delete a phase similar to how its done for changing the project tags and subscribers in the overview form?
> Screen.Recording.2026-07-08.at.5.08.50.PM.mov

oh this should not auto save closed model as in the recording it is doing for tags/subscribers, I meant if someone enter a tag they should be able to add more, but in this recording it is auto saving, but for the things like rollouts new phases it should create in the db, all add/delete should be instant, but all the selection can be temperoray until someone saves it

yashikakhurana (changes requested):
we can add a test- test_post_applies_plan_with_completed_phase_with_inconsistent_dates creates a phase where end_date < start_date — this is an invalid state that should be caught by validation but isn't.  also atetime.date.today() is used in several places — could cause issues if the server runs across timezones. Should use timezone.now().date() consistently.

yashikakhurana (commented):
Apart from adding a test for the missing test case and the date, I tested it locally looks good to me

RJAK11:
I’ve made the test changes, switched to `timezone.now().date()`, and blocked advancing to the next phase when its population is 0%.

I intentionally didn’t restrict the field itself to (0, 100] because a new phase is created in the database immediately with a default population of 0%. Enforcing that restriction at the field level would require us using a non-zero default, which could be confusing since it would not be a value chosen by the user.

yashikakhurana (approved):
perfect, tested locally everythings looks perfect, lets roll this out thank you @RJAK11 🎉

PR #16065, opened 2026-06-22
feat(nimbus): improve feature monitoring card messaging and links

### LINKED ISSUE #16064: Link Grafana dashboard to the features page
Fix the link to the correct doc if the feature is not available on the dashboard, also on that case we should show them some other feature example that they will get if they will write feature configuration, also if the link is available, show them little info this is the dashboard….

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6985)

yashikakhurana:
Because

- The metric-hub config and feature monitoring doc links in the Feature Monitoring card rendered empty, because the features view rebuilt its context without `EXTERNAL_URLS`.
- When no dashboard is available, users had no way to see what feature monitoring actually looks like.
- When a dashboard is available, there was no context explaining what it shows.

This commit

- Adds `EXTERNAL_URLS` to the features view context so the doc and metric-hub config links resolve.
- Shows a short info line describing the dashboard when monitoring is available.
- Adds an example dashboard link when monitoring is not available, so users can preview what they'll get after adding a featmon config.

fixes #16064


yashikakhurana:
Screenshot 2026-06-22 at 10 21 52 AM
Screenshot 2026-06-22 at 10 21 35 AM

jaredlockhart (approved):
Awesome 🎉 🎉 🎉 🎉

PR #16096, opened 2026-06-24
feat(nimbus): add is_holdback, do_rerun, do_rerun_timestamp fields to model and v8 API

yashikakhurana:
Because

- Holdback experiments need to be flagged in Experimenter and re-run by downstream analysis. Jetstream reads these fields from the v8 API to decide when to rerun an experiment, and metric-hub uses them for long-term holdback analysis.

This commit

- Adds `is_holdback`, `do_rerun`, and `do_rerun_timestamp` fields to the `NimbusExperiment` model and exposes them in the v8 API as `isHoldback`, `doRerun`, and `doRerunTimestamp`.
- Updates the v8 serializer and changelog serializer test expectations and regenerates the OpenAPI schema and Swagger docs.

Related

- mozilla/metric-hub#1502
- mozilla/jetstream#3258

fixes #16087

PR #16136, opened 2026-06-29
feat(nimbus): Celery tasks to run holdback analysis weekly

### LINKED ISSUE #16086: Weekly task to update holdback enrollment period and trigger Jetstream rerun
Add a Celery task that runs weekly and tells Jetstream when to rerun analysis for holdback experiments.

**What it does:** For each live holdback experiment past its enrollment end date — expands `proposed_enrollment` by the number of full weeks elapsed since enrollment ended, sets `do_rerun=True` and `do_rerun_timestamp=now()`. On the next hourly publish sync, Jetstream picks up the updated fields and triggers `rerun_config_changed`.

**Enrollment period formula:**

```
weeks_elapsed = max(1, (today - enrollment_end_date).days // 7)
new_enrollment_period = proposed_enrollment + (weeks_elapsed * 7)
new_enrollment_period = min(new_enrollment_period, proposed_duration)

```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7056)

yashikakhurana:
Because

- Holdback experiments run continuously and need Jetstream to rerun analysis weekly as the eligible population grows

This commit

- Adds a weekly Celery task that expands the proposed_enrollment window and sets do_rerun to signal Jetstream to rerun analysis

Fixes #16086


yashikakhurana:
Blocked on this https://github.com/mozilla/experimenter/pull/16096

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -203,3 +203,52 @@ def fetch_monitoring_data():
...
+                    "proposed_enrollment",
+                    "do_rerun",
+                    "do_rerun_timestamp",
+                ]
+            )
+            updated_count += 1
+
+        logger.info(
+            f"update_holdback_enrollment_period: updated {updated_count} experiments"
+        )
```
mikewilli, line 249: I'm not sure how useful it is to log the count of updated experiments, but I guess it doesn't hurt. I do think we should log the specific experiments and what changed -- perhaps just via changelog instead of an actual log.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -203,3 +203,52 @@ def fetch_monitoring_data():
...
+        experiments = NimbusExperiment.objects.filter(
+            is_holdback=True,
+            status=NimbusExperiment.Status.LIVE,
+        ).exclude(_start_date=None)
+
+        updated_count = 0
+        for experiment in experiments:
+            enrollment_end = experiment.computed_enrollment_end_date
+            if enrollment_end is None or today <= enrollment_end:
+                continue
```
mikewilli: Holdbacks have continuous enrollment, which is why we need to fake the enrollment end date. The check here should be whether the user has actually ended the holdback or not (so end date instead of enrollment end date).

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -203,3 +203,52 @@ def fetch_monitoring_data():
...
+        now = timezone.now()
+
+        experiments = NimbusExperiment.objects.filter(
+            is_holdback=True,
+            status=NimbusExperiment.Status.LIVE,
+        ).exclude(_start_date=None)
+
+        updated_count = 0
+        for experiment in experiments:
+            enrollment_end = experiment.computed_enrollment_end_date
```
mikewilli: The `enrollment_end` we want to use needs to be a rolling period of time so that the enrollment period continually grows throughout the experiment. It should be the number of days between the start of the experiment and the current date - 21 days, to allow for a 21-day observation period. Unless we decided on a different observation period length, in which case adjust 21 for the desired number.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -203,3 +203,52 @@ def fetch_monitoring_data():
...
+                continue
+
+            days_since_end = (today - enrollment_end).days
+            weeks_elapsed = max(1, days_since_end // 7)
+            new_enrollment_period = min(
+                experiment.proposed_enrollment + (weeks_elapsed * 7),
+                experiment.proposed_duration,
+            )
+
+            experiment.proposed_enrollment = new_enrollment_period
```
mikewilli: We should also update the `enrollment_end_date`, which is used by Jetstream.

mikewilli: And the `end_date`, because that will tell Jetstream to trigger `overall` analysis.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -203,3 +203,52 @@ def fetch_monitoring_data():
...
+            experiment.do_rerun = True
+            experiment.do_rerun_timestamp = now
+            experiment.save(
+                update_fields=[
+                    "proposed_enrollment",
+                    "do_rerun",
+                    "do_rerun_timestamp",
+                ]
+            )
```
mikewilli: As I look at this, it's not clear to me how this will work. If we update the actual experiment metadata, then how can we tell when the experiment actually ends? Maybe we could instead change the API serialization for `is_holdback` experiments that aren't ended to just send different data for these experiments? Or store additional mock fields with the fake values and use those in the response for `is_holdback` experiments?

Or let me know if I'm missing something that means this will work fine.

experimenter/experimenter/settings.py
```diff
@@ -421,6 +421,10 @@
         "task": "experimenter.experiments.tasks.warm_api_caches",
         "schedule": config("API_CACHE_WARMING_INTERVAL", default=3600, cast=int),
     },
+    "update_holdback_enrollment_period": {
+        "task": "experimenter.jetstream.tasks.update_holdback_enrollment_period",
+        "schedule": crontab(minute=0, hour=7, day_of_week=1),
+    },
```
mikewilli, line 431: I think this should run once per day because holdbacks can launch on any day of the week.

mikewilli (dismissed):

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -196,3 +196,50 @@ def fetch_monitoring_data():
         metrics.incr("fetch_monitoring_data.failed")
         logger.exception(f"Fatal error in fetch_monitoring_data task: {e}")
         raise
+
+
+HOLDBACK_OBSERVATION_DAYS = 21
```
mikewilli: Constants shouldn't appear in the middle of the file.

experimenter/experimenter/experiments/api/v8/serializers.py
```diff
@@ -161,6 +162,27 @@ class Meta:
             "requiresRestart",
         )
 
+    def get_enrollmentEndDate(self, obj):
+        enrollment_end = obj.actual_enrollment_end_date
+        if obj.is_holdback and not obj.end_date and enrollment_end:
+            return enrollment_end.isoformat()
```
mikewilli: I think this should be `today - 21 days`, right?

experimenter/experimenter/experiments/api/v8/serializers.py
```diff
@@ -161,6 +162,27 @@ class Meta:
             "requiresRestart",
         )
 
+    def get_enrollmentEndDate(self, obj):
+        enrollment_end = obj.actual_enrollment_end_date
+        if obj.is_holdback and not obj.end_date and enrollment_end:
```
mikewilli:
```suggestion
        if obj.is_holdback and not obj.end_date and not enrollment_end:
```

experimenter/experimenter/experiments/api/v8/serializers.py
```diff
@@ -161,6 +162,27 @@ class Meta:
...
+            return enrollment_end.isoformat()
+        return enrollment_end.isoformat() if enrollment_end else None
+
+    def get_endDate(self, obj):
+        enrollment_end = obj.actual_enrollment_end_date
+        if obj.is_holdback and not obj.end_date and enrollment_end:
+            return (enrollment_end + datetime.timedelta(days=21)).isoformat()
```
mikewilli: This 21 should also be a constant, ideally the same one that we use in the jetstream task code. Maybe it can go in the django constants/settings?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -196,3 +196,50 @@ def fetch_monitoring_data():
...
+    try:
+        today = timezone.now().date()
+        now = timezone.now()
+        enrollment_end = today - dt.timedelta(days=HOLDBACK_OBSERVATION_DAYS)
+
+        experiments = NimbusExperiment.objects.filter(
+            is_holdback=True,
+            status=NimbusExperiment.Status.LIVE,
+            _end_date=None,
+        ).exclude(_start_date=None)
```
mikewilli, line 214: Theoretically it shouldn't be possible for an experiment to be LIVE and have no start date, right? Not a big deal though if you think this could cover some edge case or something.

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -196,3 +196,50 @@ def fetch_monitoring_data():
...
+        updated_count = 0
+        for experiment in experiments:
+            if enrollment_end <= experiment.start_date:
+                continue
+
+            NimbusExperiment.objects.filter(pk=experiment.pk).update(
+                _enrollment_end_date=enrollment_end,
+                do_rerun=True,
+                do_rerun_timestamp=now,
+            )
```
mikewilli: Is changing the enrollment end date like this going to affect actual enrollments?

experimenter/experimenter/experiments/api/v8/serializers.py
mikewilli, line 1: I left some comments here and then went and looked at the task. I think what you have here makes sense IF it's valid to change the actual enrollment end date on the experiment, but I worry that this might break something else. I think it'd be better to assume the enrollment end date is NOT set until the experiment owner actually ends enrollment, and until then you can set a fake one in the serializer here the same way you do for the other dates.

If it is valid to change the actual enrollment end date, then maybe what you have here is ok.

mikewilli (dismissed):
I suggested a couple small changes, but more importantly I think there's still some confusion over which dates to fake, when to fake them, and what they should be set to. My understanding of what we want to do here isn't quite in line with what you have here, and I'm not totally sure if that's a misunderstanding on my part or if I'm just not quite explaining it clearly enough. 

Let me know if you want to chat about it to try and make sure we're on the same page.

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: yes, moved to settings

experimenter/experimenter/experiments/api/v8/serializers.py
yashikakhurana: correct, thank you for noticing this

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: that's correct

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: changed it now

yashikakhurana:
@mikewilli did the changes, I think its clear now, not updating the dates to not impact any working functionality, computing on the fly

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -196,3 +196,53 @@ def fetch_monitoring_data():
...
+            if (
+                days_since_start < minimum_days
+                or days_since_start % settings.HOLDBACK_RERUN_INTERVAL_DAYS != 0
+            ):
+                continue
+
+            update_fields = {"do_rerun_timestamp": now}
+            if not experiment.do_rerun:
+                update_fields["do_rerun"] = True
+            NimbusExperiment.objects.filter(pk=experiment.pk).update(**update_fields)
```
mikewilli: I don't know if it makes a difference, but should this be `.get` instead of `.filter`?

experimenter/experimenter/jetstream/tasks.py
```diff
@@ -196,3 +196,53 @@ def fetch_monitoring_data():
...
+            settings.HOLDBACK_OBSERVATION_DAYS + settings.HOLDBACK_MINIMUM_ENROLLMENT_DAYS
+        )
+        updated_count = 0
+        for experiment in experiments:
+            days_since_start = (today - experiment.start_date).days
+            if (
+                days_since_start < minimum_days
+                or days_since_start % settings.HOLDBACK_RERUN_INTERVAL_DAYS != 0
+            ):
+                continue
```
mikewilli, line 232: Want to log here so we know which experiments get skipped or think that'd be too noisy?

experimenter/experimenter/settings.py
```diff
@@ -421,6 +425,10 @@
         "task": "experimenter.experiments.tasks.warm_api_caches",
         "schedule": config("API_CACHE_WARMING_INTERVAL", default=3600, cast=int),
     },
+    "update_holdback_enrollment_period": {
+        "task": "experimenter.jetstream.tasks.update_holdback_enrollment_period",
+        "schedule": crontab(minute=0, hour=7),
```
mikewilli: Ok thinking about this a little more now that the logic looks right..

We probably want this task to run before Jetstream, so that everything is up-to-date when Jetstream starts. Jetstream's start time the last 25 days has typically been around 4:35am UTC, sometimes later than that but never earlier. The earliest possible time is 4am UTC based on the scheduling, but typically it's a little later because it waits for other data to be available. Assuming this scheduling is in UTC, maybe we do 3am to be safe? I don't think it hurts anything to do a little early (as long as it happens on the correct date).

mikewilli (approved):
Alright this is looking good now! Please change the schedule before merging, the other comments are up to you.

Thanks for working through everything.

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: I think we don't need this anymore as we are not updating any private property, we can directly access for the save

experimenter/experimenter/jetstream/tasks.py
yashikakhurana: I think that's the good idea, can do in debug mode

experimenter/experimenter/settings.py
yashikakhurana: that's a really good point

PR #16144, opened 2026-06-29
fix(nimbus): Refresh results using modified time

### LINKED ISSUE #16139: Auto-refresh doesn't update some results
Live/Complete experiments aren’t auto-refreshing when their Jetstream metadata has a missing or null `analysis_start_time`. In that case, `latest` becomes `None`. If all expected stats files already exist, then `has_missing_expected_results` returns `False` so the experiment gets skipped instead of refreshing. This means the experiment can get stuck with stale results even if Jetstream later writes fresh statistics. This is related to the changes from #15737

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7068)

RJAK11:
Because

* The fetch task relies on analysis_start_time to decide whether results are stale
* Jetstream does not reliably write analysis_start_time for every experiment so some Live/Complete experiments get latest = None
* Once those experiments have all expected results files they are treated as up to date and skipped forever even if there are fresh results

This commit

* Refreshes results when an result’s newest modified time is newer than the stored timestamp or when expected results are missing
* Backfill the stored timestamp to now() for already fetched experiments to avoid a mass re-fetch

Fixes #16139


RJAK11:
Small note on results_data_last_updated:

I set results_data_last_updated as now() for experiments that already have results_data so we don’t treat all of them as never fetched and end up re-fetching every Live/Complete experiment.

The one tradeoff is that if a result was updated after the last fetch but before this migration runs, that update could be missed until Jetstream writes newer files again or we manually fetch it.

mikewilli:
>Small note on results_data_last_updated:

>I set results_data_last_updated as now() for experiments that already have results_data so we don’t treat all of them as never fetched and end up re-fetching every Live/Complete experiment.

>The one tradeoff is that if a result was updated after the last fetch but before this migration runs, that update could be missed until Jetstream writes newer files again or we manually fetch it.

In case you pick this back up, I'd suggest just setting it back a day or two. We'll re-fetch a few extra experiments but way fewer than all of them, and more importantly we won't miss any.

PR #16097, opened 2026-06-24
feat(nimbus): add state machine for rollouts

moibra05:
Because

- Rollouts need dedicated lifecycle transitions, including a rollout specific `Disabled` state

This commit

- Adds rollout transition forms, views, routes, validation, and test coverage for all valid and invalid rollout state transitions
- Adds the Disabled status and updates rollout state filters
- Stages phase changes until Remote Settings approval, then advances or ends the active phase
- Supports disabling and re-enabling rollouts through the review workflow
- Updates API documentation and model choices

Fixes #15085

PR #16154, opened 2026-06-30
docs(nimbus): Add rollout mermaid diagrams

### LINKED ISSUE #16153: Rollout Mermaid Diagrams Update
Update the Mermaid diagrams to reflect the new rollout-specific state machine.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6945)

RJAK11:
Because

* we need to add Mermaid diagrams reflecting the new rollout-specific state machine

This commit

* adds the Mermaid diagrams for rollouts
* updates timeout flows to return to idle after timeout

Fixes #16153

### DIFF
--- a/docs/experimenter/rollouts.md
+++ b/docs/experimenter/rollouts.md
@@ -13,7 +13,7 @@ A new rollout which has yet to be sent for review or put into preview is marked 
     
     rect rgb(255,204,255) 
         Rollout Owner->>Experimenter UI: Create rollout
-        Experimenter UI->>Experimenter Backend: Status: Draft <br/> Publish status: Idle <br/> Status next: <none> <br/> Phase: None <br/> Phase Next: None <br/> + changelog
+        Experimenter UI->>Experimenter Backend: Status: Draft <br/> Publish status: Idle <br/> Status next: None <br/> Phase: None <br/> Phase Next: None <br/> + changelog
     end
 %%{init:{'themeCSS':'g:nth-of-type(1) .note { stroke: purple ;fill: white; };'}}%%
 ```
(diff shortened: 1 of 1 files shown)


RJAK11:
During our diagram review meeting and local testing, we found that when a Remote Settings review times out, the owner has to request review again. Because of that, we updated the timeout flows to reflect this behavior. We may need to make the same kind of update to the experiment diagrams as well.

yashikakhurana:
Also worth noting that @freshstrangemusic mentioned that rollout pause only works for the labs, what will happen for non labs rollouts, will it keep the pause behaviour as expected that means users who are already enrolled will still enrolled, but no users will be enrolled 👀 
cc @jaredlockhart and @freshstrangemusic

docs/experimenter/rollouts.md
```diff
@@ -0,0 +1,1281 @@
...
+    participant Remote Settings UI
+    participant Remote Settings Backend
+    title Enable (Approve/Approve)
+
+    Note over Rollout Owner: An owner is ready to enable <br/> their disabled rollout
+
+    rect rgb(255,204,255)
+        Note right of Rollout Owner: Owner enables in Experimenter
+        Rollout Owner->>Experimenter UI: Enable
+        Experimenter UI->>Experimenter Backend: Status: Disabled <br/> Publish status: Review <br/> Status next: Live <br/> Phase: Phase X <br/> Phase Next: None <br/> + changelog
```
jaredlockhart: After reviewing together we found that enabling must have `Phase Next: Phase X+1`

jaredlockhart (changes requested):

docs/experimenter/rollouts.md
jaredlockhart: Reasoning: 

If you disable on Phase N with size X%, then when you relaunch, you may want to launch to a different size (ie disabled at 100% and want to renable at some small percent).  If you force the user to go through two steps ie relaunch at same size THEN modify the size, you risk a period where clients will enroll/unenroll quickly.

If you allow the user to edit the current phase while it's disabled you lose the immutable record of how big that phase was while it was first live.

So the sensible thing to do is force enabling to ALWAYS move to the NEXT phase, which is editable while it's disabled, and if no next phase exists, we can warn the user that a next phase will be created with the same size as the current phase, and if they want to then edit it they can cancel and edit.

RJAK11:
I have switched all uses of `<none>` for `None`, changed all the Enable cases to have `Phase Next: Phase X+1`, and added the two cases for enabling with no next phase. Let me know if something seems off! 🙏🏼

jaredlockhart (approved):
Alright!!  Looks great to me.  Thank you so much for doing this @RJAK11.  This will make implementation MUCH smoother and will act as a reference for a long time.  Let's do it and then we can unblock https://github.com/mozilla/experimenter/pull/16097 🎉 🎉 🎉 🎉 🎉 🎉

PR #16173, opened 2026-07-02
feat(nimbus): add application field to rollouts overview form

### LINKED ISSUE #15920: Add Platform to Overview Card
The overview card should include the selected Platform which is equivalent to Application in the current UI.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6989)

moibra05:
Because

- Overview section in new rollouts form was missing the applications field

This commit

- Adds a field to specify application

Fixes #15920

### DIFF
--- a/experimenter/experimenter/nimbus_ui/new/forms.py
+++ b/experimenter/experimenter/nimbus_ui/new/forms.py
@@ -321,6 +321,15 @@ class RolloutOverviewForm(NimbusChangeLogFormMixin, forms.ModelForm):
     public_description = forms.CharField(
         required=False, widget=forms.Textarea(attrs={"class": "form-control", "rows": 3})
     )
+    application = forms.ChoiceField(
+        disabled=True,
+        choices=NimbusExperiment.Application.choices,
+        widget=forms.widgets.Select(
+            attrs={
+                "class": "form-select",
+            },
+        ),
+    )
 
     class Meta:
         model = NimbusExperiment
@@ -328,6 +337,7 @@ class RolloutOverviewForm(NimbusChangeLogFormMixin, forms.ModelForm):
             "name",
             "hypothesis",
             "public_description",
+            "application",
         ]
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


RJAK11 (commented):
I noticed that in the old UI, the Application field is disabled and has the note: “Application can not be changed after an experiment is created.” I was wondering if we should also disable this field in the new UI for consistency but I'm not too sure.

jaredlockhart:
@RJAK11 is exactly right, we put it there for reference but it's read only and can't be changed 🙏

moibra05:
ah are we still only going to be able to set the application through the " Create Experiment " modal?

jaredlockhart:
@moibra05 yeah exactly.  On the home page we'll need a new 'create rollout' button that does the same thing, title, hypothesis, application, that creates the rollout, then fill in the rest on the detail page, application field is present but locked.

experimenter/experimenter/nimbus_ui/templates/new/rollouts/overview/card.html
```diff
@@ -9,6 +9,11 @@
     <div class="text-secondary mb-1" style="font-size: 12px;">Observations &amp; Problem space</div>
     <div class="small text-body" style="white-space: pre-wrap;">{{ experiment.hypothesis|default:"—" }}</div>
   </div>
+  {# Application #}
+  <div>
+    <div class="text-secondary mb-1" style="font-size: 12px;">Application</div>
+    <div class="small text-body">{{ experiment.get_application_display|default:"—" }}</div>
```
jaredlockhart: I don't think we need a default here cuz we shouldn't even be able to enter this page with application not set ya?

experimenter/experimenter/nimbus_ui/new/forms.py
```diff
@@ -321,13 +321,23 @@ class RolloutOverviewForm(NimbusChangeLogFormMixin, forms.ModelForm):
     public_description = forms.CharField(
         required=False, widget=forms.Textarea(attrs={"class": "form-control", "rows": 3})
     )
+    application = forms.ChoiceField(
```
jaredlockhart, line 324: Ya we need to mark this as disabled.

yashikakhurana (approved):
Looks good to me, thank you @moibra05

PR #16175, opened 2026-07-02
chore(cirrus): migrate from poetry to uv

### LINKED ISSUE #16174: chore(cirrus): migrate cirrus from Poetry to uv
## Problem

Cirrus manages its Python dependencies with Poetry. We are standardizing the repo's Python projects on uv, which is now the best supported way to manage Python dependencies (faster installs, native Dependabot support). Cirrus is the pilot: its Poetry usage is confined to cirrus/server/Dockerfile, it publishes nothing, and it has its own CI and deploy pipelines.

## Change

- Convert cirrus/server/pyproject.toml from the legacy [tool.poetry] format to PEP 621 [project] plus [tool.uv] (package = false), converting caret constraints to PEP 440 ranges.
- Replace poetry.lock with uv.lock.
- Install dependencies in cirrus/server/Dockerfile via uv (uv export + uv pip install --system) instead of Poetry, keeping the existing system site-packages copy so the deploy stage is unchanged.
- Switch the /cirrus/server Dependabot entry from package-ecosystem: pip to uv.

## Acceptance criteria

- make cirrus_check passes (ruff, pyright, pytest, generate_docs).
- The cirrus:deploy image builds and runs.
- Dependabot tracks uv.lock via the uv ecosystem.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7075)

jaredlockhart:
Because

* uv is now the best supported way to manage Python dependencies, with faster installs and native Dependabot support
* We are standardizing the repo's Python projects on uv, starting with cirrus

This commit

* Converts cirrus/server/pyproject.toml to PEP 621 with [tool.uv] package = false
* Replaces poetry.lock with uv.lock
* Installs dependencies into a virtualenv via uv sync in cirrus/server/Dockerfile, with bytecode compilation and a uv cache mount
* Switches the /cirrus/server Dependabot entry from pip to uv

Fixes #16174


jaredlockhart:
Note for review on the dependency version ranges: the `>=x,<y` specifiers in `pyproject.toml` were produced by `uvx migrate-to-uv`, which mechanically expands the previous Poetry caret constraints (e.g. `^0.138.0` becomes `>=0.138.0,<0.139`), so they carry over the same bounds we already had. Exact versions for the full graph (direct and transitive) are pinned in `uv.lock`, and the image installs from the lock via `uv sync --frozen`.

This ranges-in-manifest / exact-in-lock split is uv's documented model: https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile

jaredlockhart:
Note on the venv in the Dockerfile: the builder stage runs `uv sync` into a venv at `/opt/venv` and the deploy stage copies that venv and puts it on `PATH`, rather than installing globally into system site-packages. This is because uv's lock-aware installer (`uv sync`) only targets a venv. There is no shipped `uv sync --system` (https://github.com/astral-sh/uv/issues/8085), and the clean "no venv, install straight to a system path for multi-stage copy" option (`uv sync --root`) is still an open request (https://github.com/astral-sh/uv/issues/13129).

The venv adds no isolation value in a single-purpose container, but it is harmless (a few KB) and is the pattern the uv maintainers point to: https://github.com/astral-sh/uv-docker-example. If `uv sync --root` ever lands we can drop the venv and install to the system path directly.

cirrus/server/Dockerfile
```diff
@@ -15,28 +15,30 @@ RUN apt-get update && \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | /bin/sh -s -- -y --no-modify-path
 ENV PATH=$PATH:/root/.cargo/bin
 
-# Install poetry package management tool
-RUN curl -sSL https://install.python-poetry.org | python3 - --version 2.2.1
+# Install uv package management tool
+COPY --from=ghcr.io/astral-sh/uv:0.10.0 /uv /uvx /usr/local/bin/
```
freshstrangemusic, line 19: neat!

cirrus/server/pyproject.toml
```diff
@@ -1,33 +1,33 @@
...
-pytest-asyncio = "^0.23.8"
-python-decouple = "^3.8"
-glean-parser = "^18.2.0"
-pytest-mock = "^3.15.0"
-mozilla-nimbus-shared = "^2.5.2"
-sentry-sdk = "^2.63.0"
+authors = [{ name = "Mozilla" }]
+requires-python = ">=3.11,<4"
+# Ranges generated by migrate-to-uv; uv.lock pins exact versions.
+# https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile
```
freshstrangemusic, line 8: We can probably remove this comment.

freshstrangemusic (approved):
🚀

cirrus/server/pyproject.toml
jaredlockhart: Well I was confused about the way they were expressed so I figured if anyone else ever is doesn't hurt to document

PR #16239, opened 2026-07-08
feat(nimbus): add rollout setup progress and issues modal

### LINKED ISSUE #15927: Add Setup Completion Progress Bar
Add a progress bar to the Set up section that shows how much percent of the setup is complete., warning signs, errors in case on the sidebar

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-6992)

RJAK11:
Because

* users need to know their rollout setup progress

This commit

* adds a setup completion progress bar to the rollout sidebar
* adds an issues modal that lists invalid fields with links to the relevant card

Fixes #15927

https://github.com/user-attachments/assets/c5393942-28eb-4213-95a9-d65a41bb02e3


RJAK11:
When testing locally with the "this is a rollout" checkbox checked, you'll hit 3 additional warnings not in the video for the population percent, total enrolled clients, and reference branch fields.

A few thoughts on handling these:
* Total enrolled clients doesn't apply to the new UI, and I was told we can just default it to 1 when we add the "create new rollout" flow, until we create a dedicated rollout serializer. I think the same idea works for Reference branch.
* For Population percent, the current serializer doesn't support the new rollout phases model, so I was thinking maybe we could also hardcode a value here too since it'll get overwritten anyway once the user requests to start a phase.

Separately, we might need to bring over the AI risk field from the old UI into the new Risks card, and Public description into the Overview card or we can handle them the same way since we don’t currently have those fields. Let me know what you think!

RJAK11:
Update: instead of hard-coding some values to prevent the warnings mentioned previously, I added a dedicated NimbusRolloutReviewSerializer that the new UI uses (the old UI keeps the base serializer, so this is all scoped to the new UI). That clears the 3 warnings:

* Total enrolled clients and Reference branch: changed to allow empty values for rollouts (since neither seems to apply to the new UI).
* Population percent: allows 0% since population now comes from the rollout phases model. To keep that meaningful, I added schedule validation (>= 1 phase, first phase non-zero) that shows up as a setup issue on the Rollout schedule card.

However, we may still want to add AI risk in the Risks card and Public description in the Overview card.

experimenter/experimenter/nimbus_ui/templates/new/rollouts/schedule/card.html
```diff
@@ -107,3 +107,4 @@
     </div>
   </div>
 </div>
+{% include "new/common/sidebar_setup_refresh.html" %}
```
yashikakhurana: can we move this to parent and inherit from there and then all the files will have it

RJAK11: I added a parent file which the rest of the templates inherit in the new commit. It seems like there are a lot of changes but its mostly whitespace 🙏🏼

experimenter/experimenter/nimbus_ui/new/views.py
```diff
@@ -94,7 +95,86 @@ def post(self, request, *args, **kwargs):
...
+        for fields in self.SETUP_SECTIONS.values():
+            tracked_fields.update(fields)
+        tracked_fields.update(self.NON_FIELD_ISSUE_CARDS)
+        tracked_fields.update(error_keys)
+
+        invalid_tracked = error_keys & tracked_fields
+        context["setup_completion_percent"] = round(
+            100 * (len(tracked_fields) - len(invalid_tracked)) / len(tracked_fields)
+        )
+
```
yashikakhurana: `tracked_fields.update(error_keys) `adds every erroring field into the denominator, which causes two issues:

error_keys & tracked_fields then always equals `error_keys,` so the & never filters anything (could just be len(error_keys)).
The total keeps changing — an unexpected error grows the denominator, so "100%" isn't a stable baseline.
Can we keep the denominator fixed to the known fields and only count broken ones?


tracked_fields = set()
for fields in self.SETUP_SECTIONS.values():
    tracked_fields.update(fields)
tracked_fields.update(self.NON_FIELD_ISSUE_CARDS)
`# don't add error_keys — keep total fixed`

broken = error_keys & tracked_fields
percent = round(100 * (len(tracked_fields) - len(broken)) / len(tracked_fields))

One decision for us to make first: should an unexpected error drag the progress bar down, or just show up in the issues list without affecting the percent? With the fixed denominator it does the latter. Right now test_setup_progress_untracked_issue_lowers_completion expects the former — so let's agree on the intended behavior and make the code + test match. My vote is the bar tracks only known setup fields, and unknown errors just appear as issues.

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1994,6 +1995,39 @@ def validate(self, data):
...
 
+class NimbusRolloutReviewSerializer(NimbusReviewSerializer):
+    population_percent = serializers.DecimalField(
+        7,
+        4,
+        min_value=0.0,
+        max_value=100.0,
+        required=True,
+        error_messages={"min_value": NimbusConstants.ERROR_POPULATION_PERCENT_MIN},
+    )
```
yashikakhurana, line 2005: min_value=0.0 is what we want for rollouts, but you're still passing the old "must be >= 0.0001" message — it can never fire

experimenter/experimenter/experiments/api/v5/serializers.py
```diff
@@ -1994,6 +1995,39 @@ def validate(self, data):
...
+    total_enrolled_clients = serializers.IntegerField(required=False, min_value=0)
+    rollout_phases = serializers.PrimaryKeyRelatedField(
+        many=True,
+        required=False,
+        queryset=NimbusRolloutPhase.objects.all(),
+    )
+
+    def validate_reference_branch(self, value):
```
yashikakhurana, line 2013: can we add a comment here why we are simply returning the value for the future purpose

experimenter/experimenter/nimbus_ui/new/views.py
```diff
@@ -94,7 +95,86 @@ def post(self, request, *args, **kwargs):
...
+        "Audience": "audience",
+    }
+
+    NON_FIELD_ISSUE_CARDS = {
+        "rollout_phases": {
+            "section": "Rollout schedule",
+            "card_id": "schedule",
+            "label": "Rollout Schedule",
+        },
+    }
```
yashikakhurana, line 123: `SETUP_SECTIONS, SECTION_CARD_IDS, NON_FIELD_ISSUE_CARDS`
All keyed by section, so easy to fall out of sync. Could we merge into one structure so a new section is added in one place?

yashikakhurana (changes requested):
Thank you @RJAK11 this is really great, nice approach to use separate serializer, and I have tested it locally, majorly works fine, some improvements to make, otherwise very good work

experimenter/experimenter/nimbus_ui/tests/test_new_views.py
```diff
@@ -137,6 +139,141 @@ def test_get_returns_new_rollout_detail_context(self):
...
+            issue
+            for issue in response.context["setup_issues"]
+            if issue["card_id"] == "schedule"
+        ]
+        self.assertEqual(len(schedule_issues), 1)
+        self.assertEqual(
+            schedule_issues[0]["messages"],
+            [NimbusConstants.ERROR_ROLLOUT_FIRST_PHASE_ZERO],
+        )
+
```
yashikakhurana, line 210: can we also assert here setup_completion_percent < 100 like the no-phases test doe

yashikakhurana (approved):
okay I have tested this locally, looks good to me, nice pattern, thank you @RJAK11

PR #16266, opened 2026-07-10
fix(nimbus): retry nimbus-cli fetch on transient GitHub 5xx

### LINKED ISSUE #13374: intermittent network errors when using manifesttool
manifesttool should retry HTTP requests that fail due to HTTP 4xx or 5xx errors with an exponential backoff. Three retries at something like 5s 30s 60s is probably sufficient.

see-also https://circleci.com/gh/mozilla/experimenter/359204

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-5881)

jaredlockhart:
Because

* The Update Configs cron shells out to `nimbus-cli`, which fetches each included FML file from the GitHub contents API in a burst, and GitHub intermittently fails individual files.
* With no retry, one transient failure drops an app-version into the PR FAILURES section, making that section effectively permanent noise (a different file fails almost every run).

This commit

* Retries the `nimbus-cli` subprocess with exponential backoff. The exit code cannot distinguish a transient failure from a genuine error, so retry on any failure; a real error still surfaces once the retries are exhausted.

Fixes #13374

### DIFF
--- a/experimenter/manifesttool/nimbus_cli.py
+++ b/experimenter/manifesttool/nimbus_cli.py
@@ -1,5 +1,4 @@
 import json
-import re
 import subprocess
 import sys
 import time
@@ -14,12 +13,6 @@ NIMBUS_CLI_PATH = "/application-services/bin/nimbus-cli"
 
 NIMBUS_CLI_MAX_ATTEMPTS = 3
 NIMBUS_CLI_RETRY_BACKOFF_SECONDS = 5
-
-TRANSIENT_ERROR_RE = re.compile(rb"returned (?:429|5\d\d)")
-
-
-def _is_transient_error(stderr: Optional[bytes]) -> bool:
-    return stderr is not None and TRANSIENT_ERROR_RE.search(stderr) is not None
 
 
 def _check_output(args: list[str]) -> bytes:
@@ -36,9 +29,10 @@ def nimbus_cli(args: list[str]) -> bytes:
     """Run nimbus-cli with the given arguments.
 
     nimbus-cli fetches each included FML file from the GitHub API, which
-    intermittently returns a 5xx/429 for individual files under the burst of
-    requests. Retry those transient responses so one flaky response does not
-    drop an app-version from the fetch.
+    intermittently fails for individual files under the burst of requests. The
+    exit code cannot distinguish a transient fetch failure from a genuine
+    error, so retry on any failure: a flaky response recovers, and a real error
+    still surfaces once the retries are exhausted.
     """
     print("nimbus-cli", " ".join(args))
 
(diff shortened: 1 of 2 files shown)


experimenter/manifesttool/tests/test_nimbus_cli.py
```diff
@@ -179,3 +179,71 @@ def test_generate_experimenter_yaml(
...
+
+        check_output.assert_called_once()
+        sleep.assert_not_called()
+
+    @parameterized.expand(
+        [
+            (b"GET https://api.github.com/... returned 502", True),
+            (b"GET https://api.github.com/... returned 500", True),
+            (b"GET https://api.github.com/... returned 503", True),
+            (b"GET https://api.github.com/... returned 429", True),
```
freshstrangemusic: If we hit a 409 with our token set we are doing something wrong (>= 5000 requests/hour) and we should probably not retry. That means something on a deeper level needs to change.

experimenter/manifesttool/nimbus_cli.py
```diff
@@ -24,6 +32,34 @@ def nimbus_cli(args: list[str]) -> bytes:
...
+
+            delay = NIMBUS_CLI_RETRY_BACKOFF_SECONDS * 2 ** (attempt - 1)
+            print(
+                f"nimbus-cli transient error on attempt {attempt} of "
+                f"{NIMBUS_CLI_MAX_ATTEMPTS}, retrying in {delay}s",
+                file=sys.stderr,
+            )
+            time.sleep(delay)
+
+    return _check_output(args)
```
freshstrangemusic, line 52: we're not always making HTTP requests with nimbus-cli. Do we always want to run this logic, even when we're calling `get_channels()` or `generate_experimenter_manifest()`, which operate entirely locally?

experimenter/manifesttool/nimbus_cli.py
```diff
@@ -24,6 +32,34 @@ def nimbus_cli(args: list[str]) -> bytes:
...
+
+    for attempt in range(1, NIMBUS_CLI_MAX_ATTEMPTS):
+        try:
+            return _check_output(args)
+        except subprocess.CalledProcessError as e:
+            if not _is_transient_error(e.stderr):
+                raise
+
+            delay = NIMBUS_CLI_RETRY_BACKOFF_SECONDS * 2 ** (attempt - 1)
+            print(
```
freshstrangemusic, line 45: Can we print the error that we hit?

freshstrangemusic (approved):

experimenter/manifesttool/tests/test_nimbus_cli.py
jaredlockhart: Yeah fair, I haven't seen 429s with token in the logs, it's just here as an example failure case.

experimenter/manifesttool/nimbus_cli.py
jaredlockhart: Well without a way to discriminate specific failures my thinking was, if it's a stable error, it'll survive the retries and then get surfaced, and it costs essentially nothing to reprocess some files locally, so it doesn't hurt to just blindly retry the whole pass.

PR #16301, opened 2026-07-14
chore(project): remove deprecated visualization package and v3 API

### LINKED ISSUE #16300: Remove deprecated visualization package / v3 API
The `experimenter/experimenter/visualization/` package exposes a single read-only endpoint, `/api/v3/visualization/<slug>/`, that returns `experiment.results_data["v3"]` as JSON. It is dead code and can be removed.

## Why it's safe to remove

- Not in `INSTALLED_APPS`; wired only via one URL include in `experimenter/experimenter/urls.py`.
- No Python module imports `experimenter.visualization` outside the package.
- The frontend does not call it. The old React nimbus-ui app (its only consumer) has been removed. The server-rendered `nimbus_ui` reads `results_data["v3"]` directly off the model, and the v5 serializer already surfaces the same blob.
- The endpoint was never public: the production nginx allowlist is `^/api/v[14678]`, so `/api/v3/` falls through to SSO auth. Verified in prod (`/api/v3/visualization/` redirects to SSO; `/api/v6/experiments/` returns 200). Only an authenticated experimenter session could call it, and that UI no longer exists.

Note: the `results_data["v3"]` data format itself stays. This removes only the stale endpoint/package.

## Changes

- Delete `experimenter/experimenter/visualization/`.
- Remove the `api/v3` include from `experimenter/experimenter/urls.py`.
- Regenerate the OpenAPI docs (drops the `/api/v3/visualization/{slug}/` path).
- Drop the `visualization` commit scope from `contributing.md`.

## Acceptance criteria

- The `visualization` package and its URL route are gone.
- `make check` passes (including `generate_docs --check`).
- No behavior change to the results UI, which reads `results_data["v3"]` from the model.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7135)

jaredlockhart:
Because

* The `/api/v3/visualization/<slug>/` endpoint is dead code: its only consumer was the removed React nimbus-ui app, and it was never public (nginx allowlists `^/api/v[14678]`, so v3 falls through to SSO).
* The results UI reads `results_data["v3"]` directly off the model, and the v5 serializer already surfaces the same blob.

This commit

* Deletes the `experimenter.visualization` package and its URL include.
* Regenerates the OpenAPI docs to drop the `/api/v3/visualization/{slug}/` path.
* Removes the now-unused `visualization` commit scope from `contributing.md`.

Fixes #16300


RJAK11 (approved):
This looks good to me!!

PR #16350, opened 2026-07-16
feat(nimbus): add jexl_to_sql.py — JEXL to BigQuery SQL translator

### LINKED ISSUE #16291: Build jexl_to_sql.py — JEXL to SQL Translator in Experimenter
A new file `jexl_to_sql.py` in Experimenter that:

1. Takes a JEXL string as input (e.g. `firefoxVersion >= 120 && locale == "en-US"`)
2. Uses the existing `JEXLParser` in `jexl_utils.py` to parse the JEXL into an AST (tree of nodes)
3. Walks the AST node by node
4. For each attribute, looks it up in a mapping dictionary to find the BigQuery column name
5. Builds a SQL WHERE clause from the mapped attributes
6. For any attribute not in the mapping, adds it to a `warnings` list and skips it — never errors out

### Column mapping example

```
ATTRIBUTE_TO_BQ_COLUMN = {
    "locale": "metrics.string.nimbus_targeting_context_locale",
    "firefoxVersion": "metrics.quantity.nimbus_targeting_context_firefox_version",
    "isDefaultBrowser": "metrics.boolean.nimbus_targeting_context_is_default_browser",
}

KNOWN_UNTRANSLATABLE = {
    "attachedFxAOAuthClients",  # privacy-sensitive, will never be recorded
}

```

### CI validation — 3 automated checks

- **Check 1 — Targeting config coverage:** Enumerates every attribute in all `NimbusConstants.TargetingConfig` values AND every attribute emitted by `NimbusExperiment.targeting`. Asserts each one has a mapping or is in `KNOWN_UNTRANSLATABLE`. PR fails if a new attribute appears without a mapping.
- **Check 2 — BigQuery schema drift:** Validates every BigQuery column referenced in the mapping exists in a checked-in schema snapshot of `nimbus_targeting_context`. PR fails if a column is renamed or removed.
- **Check 3 — JEXL generator changes:** Because Check 1 enumerates `NimbusExperiment.targeting` output, any change to how targeting is generated will produce new tokens that fail Check 1 automatically.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7127)

yashikakhurana:
Because

- Experiment owners have no way to know how many real clients will be eligible before launching — this is the first building block of pre-launch population sizing

This commit

- Adds `jexl_to_sql.py` that translates JEXL targeting expressions into BigQuery SQL WHERE clauses
- Maps 35+ JEXL attributes to `nimbus_targeting_context` BigQuery columns
- Handles` |date, |versionCompare, |length, |preferenceValue, |preferenceIsUserSet` transforms
- Supports `addonsInfo.addons['id']` filter expressions via IN UNNEST
- `os.isWindows derived from NOT isMac AND NOT isLinux` (not stored in BigQuery)
- Pref names use `__ separator `in BigQuery (e.g. `browser.newtabpage.enabled `→ `browser__newtabpage__enabled`)
- Partial translation returns SQL with warnings for untranslatable clauses — never blocks
- 75% of Desktop targeting configs produce usable SQL
- Adds CI test that fails if a new targeting config attribute is not in the column mapping or KNOWN_UNTRANSLATABLE — forces explicit handling of new attributes before merging

Fixes #16291

### DIFF
--- a/experimenter/experimenter/experiments/tests/test_jexl_to_sql.py
+++ b/experimenter/experimenter/experiments/tests/test_jexl_to_sql.py
@@ -7,8 +7,8 @@ from experimenter.experiments.jexl_to_sql import (
     jexl_to_sql,
 )
 from experimenter.targeting.constants import (
-    FX95_DESKTOP_USERS,
     FIRST_RUN_WINDOWS_1903_NEWER,
+    FX95_DESKTOP_USERS,
     NO_ENTERPRISE_MAC_WINDOWS_ONLY,
     WIN11_ONLY,
     NimbusTargetingConfig,
--- a/experimenter/experimenter/experiments/tests/test_jexl_to_sql.py
+++ b/experimenter/experimenter/experiments/tests/test_jexl_to_sql.py
@@ -1,0 +1,418 @@
+from django.test import TestCase
+from parameterized import parameterized
+
+from experimenter.experiments.jexl_to_sql import (
+    KNOWN_UNTRANSLATABLE,
+    jexl_to_sql,
+)
+
+_OS = "metrics.object.nimbus_targeting_context_os"
+_BS = "metrics.object.nimbus_targeting_context_browser_settings"
+_HP = "metrics.object.nimbus_targeting_context_home_page_settings"
+_AI = "metrics.object.nimbus_targeting_context_addons_info"
+_AD = "metrics.object.nimbus_targeting_context_attribution_data"
+_PREF = "metrics.object.nimbus_targeting_environment_pref_values"
+_USER_PREFS = "metrics.object.nimbus_targeting_environment_user_set_prefs"
+_UMA = "metrics.object.nimbus_targeting_context_user_monthly_activity"
+_FF = "metrics.quantity.nimbus_targeting_context_firefox_version"
+
+
+class TestJEXLToSQL(TestCase):
+    # --- Column mappings: (jexl_expression, expected_sql) ---
+
+    @parameterized.expand(
+        [
+            ("locale", "locale", "metrics.string.nimbus_targeting_context_locale"),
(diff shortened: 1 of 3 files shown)


jaredlockhart (dismissed):
Fantastic and awesome.  I think you just need to add one more line which is to invoke the jexl_to_sql right here:

https://github.com/mozilla/experimenter/blob/main/experimenter/experimenter/targeting/tests/test_targeting_configs.py#L37

So that it's hooked up to validating all existing targeting configs.  That way we know this translator works for all existing configs, and any future configs that dont' work will get caught.

experimenter/experimenter/experiments/tests/test_jexl_to_sql.py
```diff
@@ -0,0 +1,455 @@
...
+        self.assertEqual(result.warnings, [])
+
+    def test_transform_with_complex_subject_warns(self):
+        result = jexl_to_sql("(firefoxVersion + 1)|someTransform")
+        self.assertIsNone(result.sql)
+        self.assertIn("|someTransform", result.warnings)
+
+    # --- CI validation: all warnings must be accounted for ---
+
+    def test_all_targeting_config_warnings_are_accounted_for(self):
```
yashikakhurana: @jaredlockhart good point, I have this test here, but let me move to the file you mentioned, I think its better placed there

jaredlockhart: Oh oh I see, okay I missed that.

jaredlockhart (approved):

jaredlockhart:
Awesome thanks @yashikakhurana this is starting to get really interesting 🤩 🎉 🎉 🎉

PR #16354, opened 2026-07-16
feat(nimbus): Add targetingSql field to v8 API

### LINKED ISSUE #16352: Add targetingSql field to v8 API
bigquery-etl reads the v8 API to get experiment data. By including the translated SQL in the API response, bigquery-etl doesn’t need its own JEXL parser — it just receives ready-to-run SQL.

### What needs to be built

In the v8 serializer (`experiments/api/v8/serializers.py`), add a new `SerializerMethodField` called `targetingSql`:

```
targetingSql = serializers.SerializerMethodField()

def get_targetingSql(self, obj):
    if obj.status != NimbusExperiment.Status.DRAFT:
        return None
    result = jexl_to_sql(obj.targeting)
    return {
        "sql": result.sql,           # the WHERE clause string
        "warnings": result.warnings  # list of untranslatable attribute names
    }

```

### Field behavior

- Only populated for Draft experiments (Live/Complete don’t need sizing)
- Returns `null` if targeting expression is empty or produces no translatable clauses
- `sql` is the WHERE clause string ready to be inserted into a BigQuery query
- `warnings` is a list of attribute names that were skipped

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7128)

yashikakhurana:
Because

- `bigquery-etl` needs the translated SQL to run sizing queries against `nimbus_targeting_context` for the draft experiment to estimate the population

This commit

- Adds `targetingSql` field to `/api/v8/experiments/` API response
- For Draft experiments: returns the translated BigQuery SQL WHERE clause and any warnings for untranslatable attributes
- For Live/Complete experiments: returns null — they don't need sizing
- bigquery-etl reads this field directly, no JEXL parser needed on its side

Fixes #16352

### DIFF
--- a/experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
+++ b/experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
@@ -503,7 +503,6 @@ class TestNimbusExperimentSerializer(TestCase):
 
     def test_targeting_sql_populated_for_draft(self):
         _os = "metrics.object.nimbus_targeting_context_os"
-        # Create a minimal Draft with only the windows_only targeting config
         experiment = NimbusExperimentFactory.create_with_lifecycle(
             NimbusExperimentFactory.Lifecycles.CREATED,
             targeting_config_slug=WINDOWS_ONLY.slug,
--- a/experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
+++ b/experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
@@ -512,7 +512,7 @@ class TestNimbusExperimentSerializer(TestCase):
         self.assertIn("sql", targeting_sql)
         self.assertIn("warnings", targeting_sql)
         self.assertIsNotNone(targeting_sql["sql"])
-        # The composed targeting includes os.isWindows → derived as NOT isMac AND NOT isLinux
+        # os.isWindows is derived as NOT isMac AND NOT isLinux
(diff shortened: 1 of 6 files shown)


experimenter/experimenter/experiments/tests/api/v8/test_serializers.py
```diff
@@ -496,4 +497,42 @@ def _experiment_data_without_branches_and_featureIds(
...
+
+        self.assertIsNotNone(targeting_sql)
+        self.assertIn("sql", targeting_sql)
+        self.assertIn("warnings", targeting_sql)
+        self.assertIsNotNone(targeting_sql["sql"])
+        # os.isWindows is derived as NOT isMac AND NOT isLinux
+        self.assertIn("nimbus_targeting_context_os", targeting_sql["sql"])
+        self.assertIn("isMac", targeting_sql["sql"])
+        self.assertIn("isLinux", targeting_sql["sql"])
+        self.assertEqual(targeting_sql["warnings"], [])
```
jaredlockhart: It would be handy here to just affix all the inputs and then assert on the full structure of the return instead of just asserting fragments.  Just so we can see what the whole shape looks like.

jaredlockhart (dismissed):

jaredlockhart (approved):
Yeah exactly, better.  Awesome let's ship it thnx @yashikakhurana 🙏 🎉 🎉 🎉 🎉

PR #16397, opened 2026-07-20
feat(nimbus): add needsUpdate to targetingSql and sizing_data_updated field

### LINKED ISSUE #16387: Add needsUpdate flag to targetingSql API to optimize sizing job
Add a needsUpdate boolean to the targetingSql field in /api/v8/draft-experiments/ so bigquery-etl only runs BQ sizing queries for experiments that actually need updating.

Logic (computed by Experimenter):
- True if sizing_data_updated_at is null AND _updated_date_time >= now() - 7 days (never sized, recently active)
- True if _updated_date_time > sizing_data_updated_at (changed since last fetch)
- False otherwise (skip the BQ query)

This prevents running expensive BQ queries for all 1200+ Draft experiments daily. bigquery-etl reads the flag and skips experiments where needsUpdate = False.

Depends on: sizing_data_updated_at field on NimbusExperiment model (Ticket 3 of pre-launch population sizing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7162)

yashikakhurana:
Because

- Running BQ sizing queries for all 1200+ Draft experiments daily is expensive — many haven't changed and already have sizing data

This commit

- Adds `sizing_data_updated_at` field to NimbusExperiment to track when sizing was last computed
- Adds `sizing_needs_update` model property with the decision logic: True if never sized and recently active (last 7 days), or if the experiment changed after the last sizing fetch
- Adds`needsUpdate` to the `targetingSql`API response so bigquery-etl can skip experiments that don't need recomputing

Fixes #16387


yashikakhurana:
okay also it won't impact the cost- okay so based on queries- Ran benchmarks: 12 experiments = ~24s, 120 experiments = ~25s. Essentially no difference — the cost is entirely the table scan, not the number of experiment columns. So 1200 experiments in one query would be fine too. The needsUpdate flag still makes sense for data freshness (only recompute when targeting changes), but it's not needed for performance

experimenter/experimenter/experiments/models.py
```diff
@@ -2198,6 +2203,21 @@ def results_ready(self):
...
 
+    @property
+    def sizing_needs_update(self):
+        """
+        True if bigquery-etl should re-run the population sizing query.
+
+        - Never sized: only if the experiment was active in the last 7 days
+        - Already sized: only if the experiment changed after the last sizing fetch
+        """
+        last_updated = self._updated_date_time
```
mikewilli, line 2214: Do you think it's worth doing something more granular here, like just check if any of the targeting/audience info changed? Or is that more complex than it's worth?

mikewilli (approved):
Yep left one question but all looks good to me.

experimenter/experimenter/experiments/models.py
yashikakhurana: we thought about that, and thought we have updated date time already here, for that we need to store that info and then check it here, but worth doing it in the future

PR #16411, opened 2026-07-21
feat(nimbus): add rollout phase transition methods

### LINKED ISSUE #16406: Add phase transition methods
Add model methods for staging, advancing, copying, and ending rollout phases. Phase changes should only be committed after Remote Settings confirmation. Include focused model tests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/EXP-7165)

moibra05:
Because

- Rollout phases need to be staged and finalized independently as rollouts progress through their lifecycle

This commit

- Adds model methods for staging, advancing, copying, and ending rollout phases
- Adds tests covering phase progression, date updates, population changes, and edge cases

Fixes #16406


RJAK11 (commented):
This also looks good and great test coverage!! Just one minor thing I noticed:

From my understanding, the end_date <= today guard is trying to avoid overwriting a phase's real end date, but if a planned end date (set by user in the UI) passes and the phase gets ended later, it keeps the old date instead of the actual end date. 

Instead of checking the date we can always set today when we actually close the live phase. `end_current_rollout_phase` only ever seems to end the live phase, so it can probably drop the guard

`advance_rollout_phase` should probably do the same, except when re-enabling since there the phase already ended at disable and should keep its date. Status is still DISABLED at that point I think, so we can just probably check for it. Maybe something like:

```
resuming = self.status == NimbusExperiment.Status.DISABLED
if current_phase is not None and not resuming:
    current_phase.end_date = today
    if current_phase.actual_start_date:
        current_phase.start_date = current_phase.actual_start_date

```

Lemme know if I’m missing something or getting anything confused though! 🙏

moibra05:
oh yes, you're right. i think i interpreted the end_date field literally here

moibra05:
@RJAK11 i made those changes you pointed out and also added a test i was missing to ensure that advancing the rollout phase won't override the end date if the rollout is being advanced from a disabled state where the current phase was already forced to end

RJAK11:
> @RJAK11 i made those changes you pointed out and also added a test i was missing to ensure that advancing the rollout phase won't override the end date if the rollout is being advanced from a disabled state where the current phase was already forced to end

I looked over the changes and it looks good to me now!

yashikakhurana (approved):
Thank you @moibra05 looks good to me 🎉

